PR #1937, opened 2025-07-01
refactor(core): improvements to source management APIs for LSP integration

bitwalker:
This PR makes some breaking changes to the various APIs for managing source files/debug information, to better support upstream use cases, including integration into an LSP server.

The key things driving the breaking changes are:

* Source files may not always correspond to a file on disk. To better encode the semantics of source file "paths", a new `Uri` type was introduced, and usage of the "path" terminology was changed to use "URI" instead. This is of particular important to LSP integration, but in general we were very vague on what was considered a "valid" source file path. This new `Uri` type technically still allows mostly arbitrary names to be fed into it, mostly to preserve the current ergonomics, but in the future we may want to consider being stricter about what we accept.
* Source files may not always correspond to Miden Assembly sources - they might be Rust, or some other high-level language. As such, we now require specifying the source file language when creating them. A new `SourceLanguage` enum was added to make the common cases easy to specify.
* To ensure we avoid confusion between line/column indices and line/column numbers (basically zero-indexed vs one-indexed), I've introduced explicit types for both, and enforced their usages in critical APIs.

Aside from that, the changes are primarily about exposing more of the various source management APIs, and enhancing them with additional functionality needed by an LSP server:

* The ability to update a document with edits made by the user, while preserving the original `Arc<SourceFile>` if still referenced anywhere
* The ability to specify selection ranges in IDE/text editor terms. This is used by an LSP server to request ranges of the source file based on positions tracked by the editor.

The main thing this PR will require downstream is adding explicit `.into()` calls on the path parameter passed to some of the `SourceManager` APIs, addition of the `SourceLanguage` argument to `SourceManager::load`, and verifying that any places which rely on the "path" of a source file being translatable into a `LibraryPath`, be updated to be explicit about setting the source file path and the library path separately, as a path like `"foo::bar"` is not a valid URI, you must instead use `file://foo::bar`, or better yet, explicitly set the library path to `"foo::bar"` and the source file path to a valid URI.

PR #1945, opened 2025-07-03
refactor: split out debug/diagnostic infra into new crate

bitwalker:
This PR moves the debug information and diagnostics-related components from `miden-core` and `miden-assembly-syntax` respectively, into a new crate, `miden-core-diagnostics`.

**NOTE:** This should not be merged until #1937 is merged _and_ I've been able to update this branch to use the crates.io release of `serde_spanned`.

A few notes:

* We currently have a `miden-diagnostics` crate that is now only used by AirScript. I'd really like to move our diagnostics infrastructure there, but that's a bigger project than I'm interested in tackling right now.
* This leaves `miden-core` dependent on the new `miden-core-diagnostics` crate so that we can preserve the exports of the debug info types, and the `sync` module. AFAIK, we were only really using the `sync` module in the source management implementation, so if we don't really need to continue re-exporting that from `miden-core`, nor the `debuginfo` module and its contents, then we could remove the dependency on `miden-core-diagnostics` entirely
* The way things are exported from `miden-core-diagnostics` is a bit different, now that the debug info _and_ diagnostics components are mixed together. The top-level exports are, as you'd expect, primarily diagnostics-related. If we would prefer an alternative organization, I'm open to suggestions. The main point of annoyance is how pervasively things are re-exported from different crates (and different levels within those crates). If I had my way, we'd nuke all of that and stick to more explicit imports, but in an effort to minimize downstream effects, I've tried to preserve all of those for now.
* This temporarily includes a git dependency for `serde_spanned`, so that I can make the `miden-core-diagnostics` crate no-std compatible even with the `serde` feature enabled. I'm expecting a new release ASAP, but if one is not forthcoming in the next few days, I'll make the use of `serde_spanned` contingent on both `std` and `serde` features, and switch back to the crates.io release. This dependency is needed to make some of the source management types serializable with `serde`, which I rely on in subsequent changes I've got staged - it is introduced here for my convenience primarily.
* I've started moving dependency declarations used in multiple places into the root `Cargo.toml`, and using `dep.workspace = true` or `dep = { workspace = true, ... }` depending on the context. This drastically de-clutters the individual crate `Cargo.toml` files, and reduces the chance of us depending on divergent versions of the same dependencies. For now, I've focused only on a couple dependencies that are always needed by at least one crate in release builds; `dev-dependencies` and such I have not bothered with. I'd suggest we start to slowly move dependencies into the root `Cargo.toml` as we do further refactoring.


core-diagnostics/Cargo.toml
```diff
@@ -0,0 +1,58 @@
+[package]
+name = "miden-core-diagnostics"
```
plafer: Please add this crate to `publish-release.sh` 😉

core-diagnostics/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+pub use self::{
+    label::Label,
+    related::{RelatedError, RelatedLabel},
+};
+
+pub(crate) mod utils {
+    pub use miden_crypto::utils::{
+        ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
+    };
+}
```
plafer: What is the benefit of a pattern like this? Isn't it cleaner to, in this example, just use `miden_crypto::utils` in every file where it is used, as the reader can clearly see where these types are defined?

core/Cargo.toml
```diff
@@ -19,48 +19,27 @@ doctest = false
...
     "miden-crypto/std",
     "miden-formatting/std",
     "math/std",
-    "serde/std",
     "winter-utils/std",
     "thiserror/std",
 ]
+serde = ["miden-core-diagnostics/serde", "miden-crypto/serde", "math/serde"]
+# DEPRECATED
+diagnostics = []
```
plafer: Let's remove this - every release comes with breaking changes anyways. Plus, I see that `miden-processor` still needlessly requires this feature (just noise at this point).

core/src/lib.rs
```diff
@@ -50,14 +50,14 @@ assertion failed: `(left matches right)`
...
 pub mod chiplets;
-pub mod debuginfo;
 pub mod errors;
 
 mod program;
 pub use program::{Program, ProgramInfo};
 
 mod kernel;
 pub use kernel::Kernel;
+pub use miden_core_diagnostics::debuginfo;
```
plafer: Did you add this to prevent another breaking change, or do you see us keeping this long term? 

I'm still wrapping my head around the best way to manage types between our crates; where I'm landed so far is that this wouldn't be desirable. That is, every `miden` type should be exported from a single crate; if someone wants the `debuginfo` module, they should import that crate. This is how [Plonky3](https://github.com/Plonky3/Plonky3/tree/main) does it, and I've found it quite pleasant to use, as I find it reduces the mental overhead of where to find things (since every type is only defined/exported in one place). You find yourself importing a bunch of `p3-*` crates, but that's fine IMO.

core/src/utils/mod.rs
```diff
@@ -8,6 +6,7 @@ use core::{
 
 // RE-EXPORTS
 // ================================================================================================
+pub use miden_core_diagnostics::sync;
```
plafer: Similar question as in my previous comment

plafer (approved):
LGTM!

We should have a conversation about how exactly we want to manage crates/exports/reexports, as alluded to in this PR description, some of my comments, offline conversations, etc. As mentioned in my comments, I generally like how Plonky3 does it, but would be curious if there are any downsides to that approach.

core-diagnostics/Cargo.toml
bitwalker: Ah! Thanks, juggling too many different streams of work 😅

core-diagnostics/src/lib.rs
bitwalker: I want two things here:

1. The ability to qualify paths with `utils::whatever`
2. Clearly delineate the dependency on `miden_crypto::utils`, and strictly scope what is allowed to be used. If I could, I'd remove the dependency on `miden_crypto` entirely, but unfortunately we're still a long ways away from being able to standardize our serialization around `serde`, so this is a necessity for now.

But yes, more generally you could obviously just directly import from `miden_crypto`, I'm just trying to prevent more uses of it from leaking into this crate.

core/Cargo.toml
bitwalker: I'm totally fine with that, just be aware that it will break any downstream crate that depends on that feature directly or transitively when they upgrade, due to how Cargo handles features. I can remove all uses in this repo, but not sure to what extent we have downstream users of `miden-core` directly that use the feature.

core/src/lib.rs
bitwalker: > Did you add this to prevent another breaking change, or do you see us keeping this long term?

Purely for compatibility and avoiding more breaking changes, considering the laundry list of them at this point. Honestly, cleaning up the current disaster that is our legacy namespacing/re-exporting practices is going to be a long and/or painful process IMO.

> I'm still wrapping my head around the best way to manage types between our crates; where I'm landed so far is that this wouldn't be desirable. That is, every miden type should be exported from a single crate; if someone wants the debuginfo module, they should import that crate. 

Two things:

1. I absolutely 100% agree that we should not be re-exporting types from other crates _directly_ using e.g. `pub use external_crate::SomeType;` with very very few exceptions.
2. We _should_ re-export crates whose types appear in our public APIs, e.g. `pub use external_crate;`. This is to ensure that downstream users of that crate don't have to struggle with divergent dependency versions (which cause a caller's `SomeType` to differ from the callee's `SomeType`), they can simply import the type from the re-exported crate in their own code.

> This is how [Plonky3](https://github.com/Plonky3/Plonky3/tree/main) does it

I think this approach is really problematic for libraries as a general rule - it is really easy to get into a situation where your dependencies don't agree on a single version of a type. You pretty much should always re-export upstream crates for your users when types from those crates appear in your public API in order to avoid this.

core/src/utils/mod.rs
bitwalker: Yes, this is purely for backwards-compatibility. If we don't care about that, I'll gladly remove both of these re-exported modules from `miden-core` and remove the dependency on `miden-core-diagnostics` entirely

core/Cargo.toml
bobbinth: I think removing this is fine, as @plafer mentioned, we are still introducing very significant breaking changes and I would rather introduce these now than later.

core/src/utils/mod.rs
bobbinth: We actually do use `sync` module in a few places in `miden-base` (e.g. [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-tx/src/prover/mast_store.rs#L3) and [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-tx/src/auth/tx_authenticator.rs#L3)) primarily because we want a `non_std` implementation of `RwLock`.

But is also not clear to me why `sync` module needs to go into `miden-core-diagnostics`. Wouldn't it be more natural to keep it here, and then have `miden-core-diagnostics` depend on `miden-core`? If the issue was that we wanted to re-export `debuginfo` from `miden-core`, I'd rather get rid of this re-export, but keep `sync` in `miden-core`. Basically, we could do the following:

- Remove `debuginfo` re-export from `miden-core`.
- Bring back `sync` into `miden-core`.
- Make `miden-core-diagnostics` depend on `miden-core`.
- Remove `miden-crypto` dependency from `miden-core-diagnostics` (the serialization types we get from `miden-crypto` now could just come from `miden-core`).

core/src/lib.rs
bobbinth: As mentioned in another comment, I think it is fine to get rid of this re-export.

core/Cargo.toml
```diff
@@ -19,48 +19,27 @@ doctest = false
...
-paste = "1.0"
-serde = { version = "1.0", optional = true, default-features = false, features = [
-    "alloc",
-    "derive",
-    "rc",
-] }
-thiserror = { workspace = true }
-winter-utils = { package = "winter-utils", version = "0.13", default-features = false }
+thiserror.workspace = true
+winter-utils = { version = "0.13", default-features = false }
```
bobbinth: We could almost remove this dependency:

- It is primarily used to import serialization-related types, which could come from `miden-crypto` instead.
- One serialization-related type that isn't re-exported from `miden-crypto` is `ReadAdapter` - but it probably should be.
- The only other missing thing would be `group_slice_elements()` which as of Rust 1.88 could be replaced with `slice::as_chunks()`.

Cargo.toml
```diff
@@ -37,4 +38,22 @@ debug-assertions = true
 overflow-checks = true
 
 [workspace.dependencies]
```
bobbinth, line 40: In `miden-base` and other crates we've started to group dependencies by their categories (e.g., see [here](https://github.com/0xMiden/miden-base/blob/next/Cargo.toml#L40)). Maybe makes sense to do this here as well?

assembly-syntax/Cargo.toml
```diff
@@ -32,10 +31,8 @@ aho-corasick = { version = "1.1", default-features = false }
 env_logger = { version = "0.11", optional = true }
 lalrpop-util = { version = "0.22", default-features = false }
 log = "0.4"
```
bobbinth: nit: `log` could now be `log.workspace = true`.

core-diagnostics/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+miette = { package = "miden-miette", version = "8.0", default-features = false, features = [
+    "fancy-no-syscall",
+    "derive",
+] }
+miden-crypto.workspace = true
+miden-formatting.workspace = true
+parking_lot = { version = "0.12", optional = true }
+paste = "1.0"
+serde = { workspace = true, optional = true }
+serde_spanned = { version = "0.6", optional = true, default-features = false }
```
bobbinth: The PR description mentions that `serde_spanned` is a git dependency, but seems like we are able to rely on the crates.io version here?

bobbinth (commented):
Looks good! Thank you! (and really appreciate some of the dependency-related clean-up). I've replied to some comments above and also left a few more comments/questions below.

core-diagnostics/src/lib.rs
bobbinth: > we're still a long ways away from being able to standardize our serialization around `serde`

This hasn't been a goal so far and actually we've tried to stay away from `serde` as much as possible. I'm fine with revisiting our serialization approach, especially since we'll probably need to move away from winterfell dependencies over the next few months - but we should probably figure out if `serde` is what we want to go with (this is a separate discussion though).

core/Cargo.toml
bitwalker: It's probably best to depend on just `miden-crypto` for this, while we continue to need the serialization-related bits in `miden-core`; not sure if we really require `ReadAdapter` in this crate or not, but we also re-export things from `miden-core`, so it's a bit complicated.

A couple thoughts:

* I think we should bump the MSRV to latest stable, we'll pick up quite a few nice improvements, and there isn't really a compelling reason to hold our MSRV back on older releases at the moment. Especially if it enables further clean up of our deps.
* I'd like to revisit whether we actually need the `winter-utils` serialization primitives, or if we can replace them with `serde`. By using the latter, we can support serializing our types to any number and variety of formats using a more declarative specification via `serde`'s derive impl for `Serialize` and `Deserialize`. It would also decouple the actual handling of the encoding into something more re-usable (i.e. if you want efficient and compact binary serialization, there are crates like [`bitcode`](https://crates.io/crates/bitcode) that integrate with `serde` out of the box). I did exactly that in my initial implementation of the Miden package format in the compiler repo, and we only replaced that with the `winter-utils` based serialization when moving it into this repo, so AFAICT, we're not really held back from going that route due to any technical reason.
* Since the only real reason we have any serialization-related code in `miden-core` in the first place is for the serialization of MAST, which should probably be the responsibility of the Miden package format now anyway, I wonder if we even need it here at all. We could move the implementation into `miden-mast-package`, switch to the `serde`-based implementation we were originally using in the compiler, and remove the dependency on `winter-utils` as well as any of it's serialization bits re-exported from `miden-crypto`. We still require `miden-crypto` for other things in `miden-core`, but it would free up `miden-crypto` to remove the dependency on `winter-*` crates in the future if that became desirable at some point.

Cargo.toml
bitwalker: Sure!

assembly-syntax/Cargo.toml
bitwalker: Good catch! I should note that I haven't gone through and updated all the crates to use the workspace-level dependencies in this PR, only the crates where I was touching the dependencies in their `Cargo.toml`. I figure we can update all of the crates as we make changes to them, and avoid too many dependency-related conflicts between branches.

core-diagnostics/Cargo.toml
bitwalker: Yes, I went ahead and switched to the crates.io version for now - the new release is coming this week I believe, but to avoid holding up this PR, I made some adjustments. I have another PR pending where I can do the upgrade to that release when it happens.

core/Cargo.toml
bitwalker: I should also note that I think we can remove the serialization of `.masl` files - there isn't any reason to support that anymore IMO, and it is effectively just an incomplete subset of what gets serialized to a `.masp` file anyway.

That would remove the need for `ReadAdapter` in `miden-assembly-syntax` (and indeed the need for it to depend on any of the `winter-utils`/`miden-crypto` serialization bits).

core-diagnostics/src/lib.rs
bitwalker: Any particular reason to stay away from `serde`? IMO it is by far the most useful serialization library in the ecosystem, and nothing else even comes close. The biggest win is that we can trivially support serializing our types to different formats from a single description, which is of particular value with things like `Word`, `Felt`, `QualifiedProcedureName`, etc., that are used in many crates for different purposes.

core/src/utils/mod.rs
bitwalker: > But is also not clear to me why sync module needs to go into miden-core-diagnostics

It was required to maintain the re-export of the `debuginfo` module from `miden-core`. Without that, we can invert the dependency here, but...

Thinking on this further, I think I'm going to split out another crate 😅 - I think the usefulness of the `sync` module stands alone, and doesn't need to belong to `miden-core`. This removes the dependency between `miden-core`/`miden-core-diagnostics` regardless of what direction the dependency goes, and if downstream crates need the `sync` module, they don't have to pull in all of `miden-core` just for that functionality.

It's a bit annoying to have yet another crate, but it's not ideal to have amalgam crates like `miden-core` anyway because of these kinds of situations. Having more purpose-built crates with clear boundaries is going to reduce this kind of friction considerably IMO.

Thoughts?

core-diagnostics/src/lib.rs
bobbinth: I tried doing `serde` a while back (like more than 3 years ago) - so, things may have changed, but found that it was quite difficult to work with for the types of serialization I needed.

One reason for this was that we frequently have structs that are not trivially serializable - e.g., we don't need to serialize all the fields and on deserialization we can reconstruct these fields from other field. A simple example of this is a Merkle tree where we don't need to serialize internal nodes and can just re-construct them from leaf nodes during deserialization - but there are more complex examples too (I think a `StarkProof` is a good example).

So, I ended up with serialized objects that are just too big (my attempt to serialize `StarkProof` with `bincode` ended up like 2x the size of the hand-rolled serialization), or required a lot of custom serializaton/deserialization code which was quite complex (quite a bit more complicated than the winterfell-based serialization approach).

Another reason is the need for validation of deserialized data. Frequently we have structs which need to be validated after or during deserialization. The simplest example of this is `Felt` where we can't deserialize its internal representation without checking if the value is within the right range. There are also quite a bit more complex example, with perhaps https://github.com/facebook/winterfell/issues/377 being the most tricky (and it is not yet supported in winterfell-based serialization either). In this specific instance we want to protect ourselves from a potentially maliciously serialized object by putting bounds on the maximum size of a vector that can be deserialized.

But as I mentioned above, last time I really tried `serde` was quite a while ago - so, I don't mind trying it again. I do think we should check if it supports our requirements though:

- It should be relatively easy to write custom serialization/deserialization code.
- It should be relatively easy to run validation and add safeguards against maliciously serialized objects.

core/src/utils/mod.rs
bobbinth: My personal preference would be to keep `sync` in `miden-core` (at least for now). Places where we use it usually require `miden-core` anyway (though `miden-core-diagnostics` may be an exception here) and it would limit the amount of changes introduced in this PR.

This is a slight preference though - and if you feel strongly about it, I think it is fine to add another crate (though, we should probably re-organize directory structure of the repo as the number of crates grows - e.g., put them under `crates` directory as we do in `miden-base` and `miden-node`).

core/Cargo.toml
bobbinth: > I think we should bump the MSRV to latest stable, we'll pick up quite a few nice improvements, and there isn't really a compelling reason to hold our MSRV back on older releases at the moment. Especially if it enables further clean up of our deps.

Agreed! But maybe do it in another PR?

> I'd like to revisit whether we actually need the `winter-utils` serialization primitives, or if we can replace them with `serde`.

Agreed - I wrote in a bit more detail about it in https://github.com/0xMiden/miden-vm/pull/1945#discussion_r2188545005.

> Since the only real reason we have any serialization-related code in `miden-core` in the first place is for the serialization of MAST, which should probably be the responsibility of the Miden package format

Agreed - though I'm not sure `serde`-based serialization will really work for MAST. One reason is that we actually will need to serialize it in 2 different ways:

- Compact but slow to deserialize - in this format we won't be serializing internal node hashes and would derive them from the basic blocks + some metadata.
- Large but fast to deserialize - this would work largely as it works now, I estimate that this format would result in 5x - 10x bigger serialized files than the compact one.

But maybe there is a way to make it work with `serde` as well.

core-diagnostics/src/lib.rs
bitwalker: I see, I can certainly understand where you're coming from, but think the situation has probably changed. There are really two separate concerns in general here as well:

1. How we specify an object should be serialized in the abstract (i.e. whether a field should be skipped or not; whether it should be given a default value on deserialization, or perhaps computed from a function; deriving its serialization from another type, etc.). These are the parts that are format-agnostic. This is the stuff you get with `#[derive(Serialize, Deserialize)]` and things like `#[serde(skip)]`, `#[serde(default)]`, `#[serde(with = "some_module")]`, and so on. There are also crates like `serde_repr` which allow you to encode things like C-style enumerations efficiently without needing to know the underlying format.
2. The actual encoding of the format being serialized to/deserialized from. This is actually somewhat distinct from `serde` itself, hence crates like `serde_json`, which provide the encoding/decoding from `serde`'s data representation to a specific format, e.g. JSON. It is always an option to handle this separately from `serde` if truly necessary - but I have yet to find a compelling reason to do so. There are significantly better binary formats than `bincode` now (`bitcode` being a particularly good one, despite the _very_ similar sounding name). In pretty much any binary format, you are going to need to handle the kind of types present in `serde`'s data representation regardless, so it is rarely profitable to roll serialization completely manually, compared to only doing so for those edge cases (which is certainly possible with `serde`).

For example, I can say definitively that we could implement our MAST/package serialization with `serde` just as well (if not better) than it is currently. I can say that because I already did it.

That's not to say that there aren't occasional situations where one might want to manually handle serialization somewhere, but I just don't think any of those really apply to our current use cases honestly. The more pressing reality is that we are sharing a _lot_ of types across many different formats (TOML, JSON, our own ad-hoc binary encodings), and having all of that manually-implemented serialization code is not only fragile and error-prone, we have to implement it for each format independently (or make use of `serde` for a subset of them, in which case we're already using it, so we may as well lean into it).

In general, `serde` has evolved quite a lot too, so it's certainly worth a second look to see if the specific situations where we ran into problems might not be solved in a more elegant way now than was possible (or not) at that point in the past. The ecosystem has a lot of tooling around it as well, so even if something isn't natively baked into `serde` itself, there's a good chance that a crate exists to solve that specific problem on top of `serde` (e.g. `serde_repr`).

core/src/utils/mod.rs
bitwalker: I could go either way really - I'm just concerned about having things like a diagnostics crate take on a dependency to `miden-core` when it really is effectively standalone except for this one thing (the `sync` module). One of the problems we have at the moment (IMO) is the matter of where to place types in such a way that we can use them in all the places they are needed, without ending up in a situation where we can't depend on something somewhere due to circular dependencies.

Consider that `Library`, `LibraryPath`, etc., were defined in `miden-assembly`, then split out into `miden-assembly-syntax`, even though libraries aren't really a thing that that crate needs to know about, aside from paths as the means of qualifying module/procedure names. We would like that stuff to be the purview of `miden-mast-package`, but that crate depended on `miden-assembly` so that it could assemble packages in its tests, which mean that we couldn't move the types to `miden-mast-package` without ending up with a circular dependency (or having duplicate definitions of things like `LibraryPath`). Basically, the more entangled our dependency graph, the more frequently we're going to either end up with awkward placement of code in order to satisfy the needs of the graph, or needless duplication of types for the same reason. The narrower the scope of individual crates, the less entanglement of this kind we have, and we gain other benefits as well (better incremental compilation, faster compilation in general, more reuse, etc.). The downside is that it makes our already messy namespacing issue at the crate level, more so, at least for now.

Anyway, I think for now I'll leave it in `miden-core`, but just wanted to explain in more detail why more generally we probably should prefer splitting standalone things into standalone crates, rather than having chonky all-in-one crates.

core/Cargo.toml
bitwalker: > Agreed! But maybe do it in another PR?

For sure 👍 

> * Compact but slow to deserialize - in this format we won't be serializing internal node hashes and would derive them from the basic blocks + some metadata.
>
> * Large but fast to deserialize - this would work largely as it works now, I estimate that this format would result in 5x - 10x bigger serialized files than the compact one.

Certainly both can be done with `serde`, but supporting both simultaneously might be more of a pain, I'd need to consider it more to see if there is an elegant way to do so with `serde` out of the box, or whether it would be easier to support the fast format with `serde` and the slow format with something hand-rolled (that may re-use bits of the `serde` serialization where there is overlap).

Regardless, we only have the fast format currently, so it would make sense to use `serde` for that now and then make the decision on whether to implement the slow format with `serde` as well, or hand-roll that one, when we get to the point of implementing it, IMO.

core-diagnostics/src/lib.rs
plafer: Created #1962

core/src/lib.rs
plafer: > I think this approach is really problematic for libraries as a general rule - it is really easy to get into a situation where your dependencies don't agree on a single version of a type.

And what is the problem with that? Say `crate-a` depends on `miden-core v0.15.0`, while `crate-b` depends on `miden-core v0.16.0`; then cargo will just include both versions in the final build. So other than (maybe) binary size, what is the problem/situation that we're trying to avoid here?

bitwalker:
@bobbinth @plafer A few updates so you know what has changed since your review:

**NOTE:** All of the changes described below are made in separate commits so they are more easily reviewed independently.

* I split up the diagnostics and source-level debug information components into two crates, `miden-utils-diagnostics` and `miden-debuginfo` (I'll get to naming in a minute).
* I did end up splitting out the `sync` module into a new crate, `miden-utils-sync`. This was necessary to disentangle awkward dependencies between `miden-core`, `miden-utils-diagnostics`, and `miden-debuginfo`. Namely, `miden-core` requires the `miden-debuginfo` types `Location` and `Uri`, while `miden-debuginfo` requires the `sync` module for the `DefaultSourceManager` implementation, which we wanted to provide from `miden-core`. Rather than try to split up the `miden-debuginfo` types across `miden-core` and `miden-debuginfo`, it is more straightforward to extract the `sync` module to a new crate, make `miden-debuginfo` and `miden-core` depend on it, and then `miden-core` can depend on `miden-debuginfo` for access to `Location` and `Uri`.
* In the process of doing the above, I created a `crates` directory under which we can start to move crates, starting with the new ones I've added. This directory is meant to be organized to reflect the naming, e.g. `miden-utils-diagnostics` is in `crates/utils/diagnostics`, while `miden-debuginfo` is in `crates/debuginfo`. This is so we can start cleaning up the project root a bit with all of the crates we have.
* It also felt appropriate here to move `miden-test-utils` to `crates/utils/testing` and set up the workspace-level dependency so that it appears as `miden-utils-testing` (so the crate naming is consistent), while preserving the actual crate name of `miden-test-utils` for now, so we can continue to publish it under that name. We don't appear to be automatically publishing that crate though, and it isn't clear to me why we would be, other than perhaps to satisfy Cargo publishing checks - but regardless, my intent was to try and canonicalize the naming here around `miden-utils-<crate>` (for general utility crates).
* As a result of all this, I also undertook to make our workspace `Cargo.toml` files consistent about using workspace dependencies, the actual crate names rather than crate aliases, and simplify re-exports from `miden-core`, `miden-assembly-syntax`, and `miden-assembly`.
* I've also incorporated the other review feedback I hadn't yet addressed

The above changes have resulted in a fair amount of churn, mostly trivial renames, but hopefully this is a one-time annoyance and things will be easier from here on.

I'd suggest we merge this as soon as possible to avoid conflicts with other ongoing work, but it's pretty unavoidable that there will be conflicts. I'm happy to also resolve the conflicts myself, here, once all other work going into the `0.16.0` release is ready/merged, and only merge this PR then - let me know one way or the other.

miden-vm/benches/program_execution_fast.rs
```diff
@@ -54,7 +54,7 @@ fn program_execution_fast(c: &mut Criterion) {
                     let stack_inputs: Vec<_> = stack_inputs.iter().rev().copied().collect();
                     bench.to_async(Runtime::new().unwrap()).iter_batched(
                         || {
-                            let processor = FastProcessor::new_with_advice_inputs(
+                            let processor = Fastmiden_processor::new_with_advice_inputs(
```
plafer: Looks like a typo (not caught by our CI)

plafer: After investigating a bit, I forgot to add the `no_err_ctx` feature to the `ALL_FEATURES_BUT_ASYNC`. TLDR created #1964 for myself to fix in a subsequent PR

miden-vm/tests/integration/cli/cli_test.rs
```diff
@@ -63,8 +63,8 @@ fn cli_run() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-use assembly::Library;
-use vm_core::Decorator;
+use miden_assembly::Library;
+use miden_core::Decorator;
 
 #[test]
 fn cli_bundle_debug() {
```
plafer, line 65: The cli tests seem to hang in the CI - and I was able to reproduce locally (curiously also on the `next` branch so not sure what's going on)

core/src/lib.rs
bitwalker: @plafer The problem is when you want to pass a type from `miden-core v0.15.0` to an API that expects the `v0.16.0` version of that type, this is why you re-export the crate whose types are in your public API, so that one can always access the exact version of those types needed.

miden-vm/benches/program_execution_fast.rs
bitwalker: Whoops, good catch

miden-vm/tests/integration/cli/cli_test.rs
bitwalker: Thanks, I'll have a look!

bitwalker: @plafer I'm unable to reproduce with `make test` locally - and it looks like the tests are being run. Which test is hanging for you?

bobbinth: I've restarted the task in the CI but it seems like it is going to hang again. The tests that seem to cause this are:

- `cli_bundle_debug`.
- `cli_bundle_kernel`.
- `cli_bundle_kernel_noexports`.

bitwalker: @bobbinth I can't reproduce locally on any of those 🤷, what about you?

bitwalker: @plafer What system are you testing on, macOS or Linux? If the latter, what arch?

bobbinth: Yes - I can see this locally too (in fact, a few more tests get stack). I'm on Apple M4 Max, and just running `make test`.

Screenshot 2025-07-07 at 6 23 35 PM

I can try to go and test commit-by-commit, but it'll have to be late tonight.

bitwalker: Yeah I'm at a complete loss here - I wasn't seeing this at all on my Macbook Air M4. The entire test suite runs without any slow/hung tests, so there is something different about my local environment than either CI or you/@plafer's machines.

However, after running `cargo clean` and then re-running the test suite with `make test`, I'm now observing the slow tests, so it is entirely unclear to me how long this may have been lingering, or what the cause might be. I'll poke around a bit more tomorrow, but not sure this is actually caused by this PR, particularly since @plafer could reproduce with `next`

bitwalker: I think this may be due to the CLI test suite running `cargo build` in parallel to build the `miden-vm` executable. I suspect they are deadlocking `cargo` itself since each test is not being run in its own directory (which is probably good for other reasons, as otherwise they'd all rebuild `miden-vm` from scratch each time). I can semi-confirm this by the fact that `ps ax | grep miden` shows a ton of `cargo build` processes.

We should probably be building `miden-vm` only once, and then using the result of that build for all of the tests, e.g. using `OnceLock` or something.

bitwalker: It appears to be as I expected, but the problem is actually a bit worse than I thought - because the CLI tests are being run as integration tests, every test is being run in its own process completely independent of the others, so we can't use `OnceLock` to solve this in `cli_test.rs` without moving the test into the unit tests of the `miden-vm` crate instead (which we can certainly do here, we aren't actually using the integration test facility as it is intended in this case).

Regardless, this absolutely affects any branch (including `next`), and has done for some time. We likely get away without noticing it, since the integration test module rarely changes, and so rebuilds of the CLI aren't required frequently enough to trigger it (if I understand the exact conditions in which it is using a cached build of the `miden-vm` executable).

bitwalker: @bobbinth @plafer ^

bobbinth: Would the solution be to disable CLI tests for now? We have just a few of them (10, I think?) and while important, they don't test anything super critical.

plafer: Thank you for investigating! Your analysis makes sense (and is the best we've got so far) - I'm also on an M4, and was able to run each test independently on a clean build.

I think we should indeed disable them to not hold up this PR, and create an issue with the suggested fix (i.e. making them unit tests).

EDIT: Alternatively, we could serialize them either by merging them into one big test, or using `test_serial`, if it ends up being easier than converting them into unit tests (we'd just be serializing these 4-5 tests, so shouldn't be noticeable when running the full test suite).

bitwalker: I think I'll take a stab at moving them into unit tests (it is as simple as moving the file under `src` and gating the module under `#[cfg(test)]`) - if it works, we've solved the problem; if it doesn't, then I'll disable them for now and we can try one of your other suggested solutions (I think using `test_serial` is the next best option probably).

bitwalker: I've found the actual bug, but also modified our test configuration to run the CLI tests serially (so we only run `cargo build` for `miden-vm` once), and to enforce a timeout so that tests don't hang our CI runners for hours.

plafer: Interesting, so what was it about the environment variables (and `internal` feature?) that caused the bug? (based on [128f4ba](https://github.com/0xMiden/miden-vm/pull/1945/commits/128f4ba491a4ed3b0b30bfe8e40a908d43e3ed29))

bitwalker: * For the environment variables, `escargot` uses `--message-format=json` under the covers, but we were forcing Cargo to emit more verbose (and colored) output, which resulted in the parsing of messages by `escargot` to fail, and internally that was causing it to hang. The environment variables we were setting don't make sense to set here anyway, because not only do we not see the output, the output is already verbose because of the message format used.
* The `internal` feature is implied by `executable`, but furthermore, there appears to be an issue with how `escargot` is passing the features through to `cargo`. They are supposed to be comma-separated, not space-separated, and from what I can tell, `escargot` is passing them through unmodified. I can't tell if this was a change made in Cargo, or what. Regardless, we didn't need to specify the `internal` feature anyway.

crates/debuginfo/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "miden-debuginfo"
```
bobbinth: nit (feel free to ignore): would `miden-debug-types` be a more appropriate name?

crates/utils/diagnostics/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "miden-utils-diagnostics"
```
bobbinth, line 2: Question: the main reasons why this is `miden-utils-diagnostics` rather than `miden-diagnostics` is that `miden-diagnostics` is already taken by our other crate, right?

But if this is VM-specific, should it be `miden-vm-diagnostics`?

crates/utils/diagnostics/README.md
bobbinth, line 1: Not for this PR, but would be good to add a bit more info to the README (we do this for most other crates). Let's create an issue for this.

crates/utils/sync/.gitignore
bobbinth, line 1: Why do we need this separate `.gitigonore` file? (rather than modifying the root one)

crates/utils/sync/README.md
bobbinth, line 1: Same comment about the readme file as above - ideally, we'd provide a bit more here to be consistent with our other crates.

bobbinth (approved):
Looks good! Thank you! I left a few comments/questions inline. They are pretty much optional - so, feel free to merge even as is.

crates/debuginfo/Cargo.toml
bitwalker: That seems reasonable to me, `miden-debuginfo` was all I could come up with that seemed reasonable, but I like `miden-debug-types`

crates/utils/diagnostics/Cargo.toml
bitwalker: Correct - I would've used `miden-diagnostics` if it wasn't currently taken, and I would've moved this stuff to `miden-diagnostics` if it was relatively easy to do so. Unfortunately, the AirScript crates depend on the current implementation of `miden-diagnostics`, and migrating is a bit more involved than I wanted to tackle as part of this (and would've ended up imposing another dependency on releasing these changes).

This crate isn't really VM-specific though, or really even VM-related (to my knowledge, the VM doesn't really use it to nearly the same extent as the compiler and assembler). Ultimately the end goal would be to move this to `miden-diagnostics` in a future release, once we can migrate AirScript, thereby unifying most of our toolchain components around a common diagnostics infrastructure. So the naming here is really temporary IMO.

crates/utils/diagnostics/README.md
bitwalker: I kept this one to the bare minimum because it is really intended to be an internal-facing crate _and_ will be removed in favor of moving this stuff to `miden-diagnostics` in a future release.

crates/utils/sync/.gitignore
bitwalker: I think this was the doing of `cargo new`, and I must've missed it

crates/utils/sync/README.md
bitwalker: How much are we intending to document in `README.md` vs rustdoc? For bigger components that have more complex usage (and thus need more descriptive text), it makes sense to put that in `README.md` - but for simple components that already have fairly good documentation, it would be redundant to also put that documentation in `README.md` (not to mention require us to manually keep them in sync, which isn't ideal).

I can certainly expand the readme here, but would like to have a better intuition on how we view `README.md` with respect to documentation in general.

bobbinth: I don't think we need to add more than a couple of paragraphs. Maybe list the what structs are provided and some basic info about implementation (e.g., that we use spin-lock). I would also add license section at the bottom as we do in other crates (e.g., how we do in the [package](https://github.com/0xMiden/miden-vm/tree/next/package) crate).

But again, I think we can do it in a separate PR - so, I'll merge this one as is.

PR #1946, opened 2025-07-03
refactor: use semver::Version rather than hand-rolled impl

bitwalker:
This PR replaces our hand-rolled `Version` struct in `miden-assembly-syntax` with `semver::Version`, a much more complete implementation of semantic versions and requirements, which I plan to use in subsequent changes.

**NOTE:** This should be merged _after_ #1945


bobbinth:
@bitwalker - if you resolve the merge conflicts here, we could get this one into the release as well.

bitwalker:
I should have that done here in just a few

Issue #1815, opened 2025-05-14
Refactor `OpBatch` so that all NOOPs are inserted


plafer:
Currently, the processor inserts NOOPs which are not explicitly represented in an `OpBatch` (e.g. when the last operation in an operation group contains an immediate value). This slows down the fast processor due to the increased bookkeeping.

It would be better to instead construct operation batches directly with the NOOPs already inserted at assembly time.

plafer:
Working on #1824 made me realize that there are actually quite a few subtle problems that arise with this issue.

The main one is that adding `NOOP`s in the `Vec<Operation>` list of a basic block invalidates the `DecoratorList` (which uses operation indices internally). This is further complicated by the fact that the assembler will also merge basic blocks, which can in some cases remove some NOOPs, further invalidating any `DecoratorList`. For example, when merging `BB1 || BB2`, and the last operation of BB1 is a `PUSH`, there will be a `NOOP` after the `PUSH`. But when we append to `BB2` to `BB1`, we no longer need that `NOOP` since the `PUSH` is no longer the last operation in its group. This is only one example that shows that there's a decent amount of complexity to get right here, and I'm not sure it's worth it at this point, especially since we're foreseeing large changes to decorators with #1776. Nevertheless this is a good learning for #1776, in that if we'll have "external objects" (like `DebugInfo`) pointing into an operation stream, we really need to be precise about how an operation stream is defined (namely, with or without NOOPs?).

Hence I now think it's best to address #1776 first and then come back to this one. #1824 should be basically there once we've figured that out.

PR #1940, opened 2025-07-02
MastForest: separate debugging information into its own datastructure

paracetamolo:
**NOT READY FOR REVIEW**

This PR moves all the debugging information from the MastForest nodes into a dedicated DebugInfo structure.

The goal of this new data structure is to be able to easily remove the debug information from a MastForest in order to save space, e.g. when uploading on-chain, but also to be able to overlay again the same debug info into a stripped MastForest.

Tests passing:
- assembly
- processor
- stdlib

Tests failing:
- integration

Issue #1776, opened 2025-04-28
Investigate removing `Decorator` in favor of a new `DebugInfo`


plafer:
From #1693:

>It would be cleaner as well to have this debug information separately, instead of stored in decorators in the MastForest. For example, this struct could look like

```rs
struct DebugInfo {
  // Maps node_hash -> procedure_name
  proc_entrypoints: BTreeMap<RpoDigest, String>,
  proc_exits: BTreeMap<RpoDigest, String>,

  // some map "location -> assembly op" (the source mapping information"
  // some map "location -> debug option" (the debug decorator)
}
```

>which could be passed as an Option<DebugInfo> to Process:execute().
>
>This also has the benefit that we could implement the ability to provide an ExecutionTrace along with a DebugInfo to a debugger, and debug the ExecutionTrace with nice source mapping. This could be particularly useful to debug traces that occurred in production, where debug info is not available. This is why we said earlier that we prefer to use procedure hashes instead of MastNodeIds in DebugInfo - the execution trace doesn't contain node IDs, but it contains all the hashes (in the decoder).

Finally, in the processor, we would also then only execute the debug infos in debug mode (as mentioned in https://github.com/0xPolygonMiden/miden-vm/pull/1668#discussion_r2020240184)

plafer:
As touched on in https://github.com/0xMiden/miden-vm/issues/1815#issuecomment-2904783128, specifying a "location" in the code is a bit subtle, basically due to the ambiguity of whether we consider the NOOPs inserted at runtime or not in an operation stream.

Hence, one way to define a "unique location identifier" would be with a triplet `(MastNodeId, BatchIdx, OpIdxInBatch)`. For all non-basic block nodes, `BatchIdx` and `OpIdxInBatch` are set to 0. In this case, the representation itself makes it unambiguous that we consider the inserted NOOPs. We could even change the NOOP-insertion rules, and still we wouldn't need to change this representation - hence feels pretty robust to me.

One question that arises then is: if the fast processor is not tracking batches (for performance reasons) but is instead just executing an `impl Iterator<Item=Operation>` stream, how does it find the right `AssemblyOp` when an error occurs? One solution is to have that logic in the iterator. That is, at a high level, the `OperationIterator` could provide a `get_batch_and_op_idx()` method, where it would deduce that information from its internal state. For example,

```rs
struct OperationIterator<'a> {
  basic_block: &'a BasicBlockNode,
  op_batch_idx: usize,
  op_idx: usize
[...]

PR #1950, opened 2025-07-03
Introduce Noop in MAST basic blocks

paracetamolo:
Fixes https://github.com/0xMiden/miden-vm/issues/1815

When a basic block is built, it contains only Noops inserted by the user, however during execution more Noops are inserted by the processor following two specific rules. This complicates the relationship between an operation, a decorator and their clock cycle, making some things more complicated (e.g. https://github.com/0xMiden/miden-vm/pull/1940).

This PR introduces the Noop directly in the basic_blocks of the MastForest, simplifying the work of the processor.

**Only one rule is implemented for now. Making the number of groups 1,2,4 or 8.**

### DIFF
--- a/assembly/src/mast_forest_builder.rs
+++ b/assembly/src/mast_forest_builder.rs
@@ -8,8 +8,8 @@ use core::ops::{Index, IndexMut};
 use miden_core::{
     AdviceMap, Decorator, DecoratorList, Felt, Operation, Word,
     mast::{
-        DecoratorFingerprint, DecoratorId, MastForest, MastNode, MastNodeFingerprint, MastNodeId,
-        Remapping, SubtreeIterator,
+        BasicBlockNode, DecoratorFingerprint, DecoratorId, MastForest, MastNode,
+        MastNodeFingerprint, MastNodeId, Remapping, SubtreeIterator,
     },
 };
 
@@ -346,12 +346,7 @@ impl MastForestBuilder {
                 self.mast_forest.is_procedure_root(basic_block_id),
                 basic_block_node.num_op_batches(),
             ) {
-                for &(op_idx, decorator) in basic_block_node.decorators() {
-                    decorators.push((op_idx + operations.len(), decorator));
-                }
-                for batch in basic_block_node.op_batches() {
-                    operations.extend_from_slice(batch.ops());
-                }
+                merge_block(&mut operations, &mut decorators, &basic_block_node);
             } else {
                 // if we don't want to merge this block, we flush the buffer of operations into a
                 // new block, and add the un-merged block after it
@@ -376,6 +371,47 @@ impl MastForestBuilder {
         }
 
         Ok(merged_basic_blocks)
+    }
+}
+
+/// Appends to [operations] and [decorators] the content of [basic_block_node] except for Noops that
+/// don't have any decorator attached. For every skipped `Noop`, the decorators are shifted
+/// accordingly.
+fn merge_block(
+    operations: &mut Vec<Operation>,
+    decorators: &mut DecoratorList,
+    basic_block_node: &BasicBlockNode,
+) {
+    let offset = operations.len();
+    let mut skips = 0;
+    let mut max_dec_id = 0;
+    for (op_id, op) in basic_block_node.operations().enumerate() {
+        let decs: Vec<&DecoratorId> = basic_block_node
+            .decorators()
+            .iter()
+            .filter(|(id, _d)| *id == op_id)
+            .map(|(_, d)| d)
+            .collect();
+
+        if decs.is_empty() {
+            if matches!(op, Operation::Noop) {
+                skips += 1;
+            } else {
+                operations.push(*op);
+            }
+        } else {
+            max_dec_id = op_id;
+            for d in decs {
+                decorators.push((op_id + offset - skips, *d));
+            }
+            operations.push(*op);
+        }
+    }
+
+    for (op_id, dec) in basic_block_node.decorators().iter() {
+        if *op_id > max_dec_id {
+            decorators.push((op_id + offset - skips, *dec));
+        }
     }
 }
 
--- a/core/src/mast/node/basic_block_node/tests.rs
+++ b/core/src/mast/node/basic_block_node/tests.rs
@@ -52,7 +52,8 @@ fn batch_ops() {
     assert_eq!([2_usize, 0, 0, 0, 0, 0, 0, 0], batch.op_counts);
     assert_eq!(hasher::hash_elements(&batch_groups), hash);
 
-    // --- one group with 7 immediate values --------------------------------------------------
+    // --- one batch of 8 groups, the first with 8 operations and the following 7 with immediate
+    // values ---
     let ops = vec![
         Operation::Push(ONE),
         Operation::Push(Felt::new(2)),
@@ -85,7 +86,9 @@ fn batch_ops() {
     assert_eq!([8_usize, 0, 0, 0, 0, 0, 0, 0], batch.op_counts);
     assert_eq!(hasher::hash_elements(&batch_groups), hash);
 
-    // --- two groups with 7 immediate values; the last push overflows to the second batch ----
+    // --- two batches, the first with 7 groups, one with 9 operations, 6 with immediate values; it
+    // needs to be padded with a 8th group with a Noop so that they are 8 groups in total. The
+    // last push overflows to the second batch ----
     let ops = vec![
         Operation::Add,
         Operation::Mul,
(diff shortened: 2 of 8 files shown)


bobbinth:
I haven't looked too closely at the code yet, but a few thoughts/comments:

### Inserting `Noop`s
I think inserting `Noop`s in `OpBatchAccumulator::into_batch()` (as you have it) is probably the way to go for now. In the longer term, we should probably get rid of the concept of `OpBatch` (at least as a part of the MAST). If we insert all the relevant `Noop`s into batches, then `BasicBlockNode` could be simplified to:
```rust
pub struct BasicBlockNode {
    digest: Word,
    ops: Vec<Operation>,
    batch_lengths: Vec<usize>,
    decorators: DecoratorList,
}
```
And once https://github.com/0xMiden/miden-vm/issues/1776 is done, it could be further simplified to:
```rust
pub struct BasicBlockNode {
    digest: Word,
    ops: Vec<Operation>,
    batch_lengths: Vec<usize>,
}
```
Op batches, then, would be computed on the fly during trace generation.

### Merging basic blocks

I think one of the trickiest parts of this PR would be to properly handle merging basic blocks. I am not sure we can just filter out noops and then merge the resulting operations. The main reason for this is that we'd also be filtering out user-inserted noops, which we probably shouldn't do.

Overall, it may be cleaner to create something like a `BasicBlockMerger` struct to help manage this process (this struct could then be used from within `MastForestBuilder`).


### Serialization

One of the questions is if we want to serialize the "padded" noops. This would be tricky as we'd need to differentiate somehow between user-inserted noops and padded noops. I think for now we just serialize everything and later we can figure out if there is a good way to somehow change this.

core/src/mast/node/basic_block_node/tests.rs
```diff
@@ -183,10 +189,11 @@ fn batch_ops() {
     ];
     let (batches, hash) = super::batch_and_hash_ops(ops.clone());
     assert_eq!(1, batches.len());
-
+    std::dbg!(&batches);
```
plafer: Reminder to remove (if it was forgotten there)

plafer:
> I think one of the trickiest parts of this PR would be to properly handle merging basic blocks. I am not sure we can just filter out noops and then merge the resulting operations. The main reason for this is that we'd also be filtering out user-inserted noops, which we probably shouldn't do.

There's no way to explicitly add a NOOP from MASM, right? And we only do this block merging at assembly-time. So should be fine to filter out NOOPs during assembly, since there are no user-inserted noops?

bitwalker:
> There's no way to explicitly add a NOOP from MASM, right? And we only do this block merging at assembly-time. So should be fine to filter out NOOPs during assembly, since there are no user-inserted noops?

There is - the `nop` instruction can be written by hand. More generally, we must ensure that any MASM instruction be represented in the textual format so that the AST can be round-tripped (i.e. we should not emit MASM containing e.g. `nop` when the user can't write the same MASM by hand).

assembly/src/mast_forest_builder.rs
```diff
@@ -346,7 +346,13 @@ impl MastForestBuilder {
                     decorators.push((op_idx + operations.len(), decorator));
                 }
                 for batch in basic_block_node.op_batches() {
-                    operations.extend_from_slice(batch.ops());
```
plafer, line 353: (Continuation of the conversation in the "main PR thread" so that we get a single thread for it)

> More generally, we must ensure that any MASM instruction be represented in the textual format so that the AST can be round-tripped (i.e. we should not emit MASM containing e.g. nop when the user can't write the same MASM by hand).

If we agreed that there was no reason for a user to *want* to write a `nop` (are there?), and that the only 2 ways to get a `Operation::Noop` in a `BasicBlockNode` is if it was "inserted by the processor" (for the 2 reasons discussed in this PR's issue), then we could remove `Instruction::Nop` from MASM altogether. This would simplify how we handle block merging, etc.

cc @bitwalker

bitwalker: The compiler currently uses `nop` to attach trace events for capturing call frame information. I haven't explored whether that can be done with equivalent fidelity using other approaches, but AFAIK we don't yet have a way to attach a trace event to the _end_ of a block (only entering it), which is we use use `nop` for that purpose currently.

I agree though that in principle, we could remove `nop` from MASM if there is no reason to ever write it by hand - but are we sure there aren't valid reasons to do so? In traditional assemblers, it's usually to pad out code to a certain alignment, or to leave space for hot patching - I'm not sure what the equivalent in Miden would be, maybe padding out a procedure so it hits a certain number of cycles or something? 🤷

plafer: > but are we sure there aren't valid reasons to do so? In traditional assemblers, it's usually to pad out code to a certain alignment, or to leave space for hot patching - I'm not sure what the equivalent in Miden would be, maybe padding out a procedure so it hits a certain number of cycles or something?

Yeah *perhaps* in some cases someone might want to use it to differentiate the MAST roots of 2 otherwise identical procedures? Might be useful to workaround some VM issues or something. I guess it's probably safer to keep it in that case 😔

paracetamolo:
> The compiler currently uses nop to attach trace events for capturing call frame information.

AFAIK Noops are only used to attach decorators, which is why during merging I need to check if there are decorators, otherwise I can't remove the Noop.

However if the only case we need is to attach a decorator at the end of a block, then we should simply treat basic blocks like all other nodes and give them a `before_enter` and `after_exit` list of decorators. This would make everything more uniform and could allow us to be more precise with the use of Noop.

@bitwalker @plafer

bitwalker:
If something like the following:

```
export.foo
  trace.CALL_FRAME_START
  push.1
  trace.CALL_FRAME_END
end
```

Resulted in the `CALL_FRAME_START` event being attached to the `before_enter` decorators of the `export.foo` block, and the `CALL_FRAME_END` event being attached to the `after_exit` of that same block, _and_ that block structure is preserved in the MAST, then I would no longer need to use `nop` for that purpose.

Currently we emit code that looks like this instead:

```
export.foo
  trace.CALL_FRAME_START
  nop
  push.1
  trace.CALL_FRAME_END
  nop
end
```

plafer:
> then we should simply treat basic blocks like all other nodes and give them a before_enter and after_exit list of decorators.

This is going away with #1776 right? So yeah feel free to add before_enter/after_exit for basic blocks if helpful (with the expectation that it will be removed a few PRs down the line)

paracetamolo:
> > then we should simply treat basic blocks like all other nodes and give them a before_enter and after_exit list of decorators.
> 
> This is going away with #1776 right? So yeah feel free to add before_enter/after_exit for basic blocks if helpful (with the expectation that it will be removed a few PRs down the line)

It won't exactly go away, it will be moved to `DebugInfo`. Right now in the DebugInfo data structure, there is the exact same structure of `before_enter` and `after_exit` lists of decorators for all nodes, plus the "interleaving" decorators for basic blocks.

paracetamolo:
Regarding after/before decorators for blocks: https://github.com/0xMiden/miden-vm/issues/1979

bobbinth:
> ```
> export.foo
>   trace.CALL_FRAME_START
>   nop
>   push.1
>   trace.CALL_FRAME_END
>   nop
> end
> ```

From cycle-count perspective this is no different from using events - i.e.:
```
export.foo
    emit.CALL_FRAME_START
    push.1
    emit.CALL_FRAME_END
end
```
Though, for things like this we should probably have special decorators that are inserted by the assembler automatically.

bobbinth:
Trying to think about a staged approach of how we get get to the refactoring we want to do in #1776, I have the following:

- We probably can't get rid of the user facing `nop` operation in the near future (maybe after [#1776](https://github.com/0xMiden/miden-vm/issues/1776) is complete, we can revisit this).
- This also means that we probably don't need to rework handling of `before_enter` and `after_exit` decorators for basic blocks (i.e., [#1979](https://github.com/0xMiden/miden-vm/issues/1979)) right now.

The path then to get to #1776 could be as follows:

1. We modify the internal semantics of `OpBatches` so that they always contain padding `Noop`'s. This is in part covered by this PR, but we need to include all padding rules.
2. We modify how `OpBatches` are serialized so that they can be de-serialized unambiguously. This is what is being done in [#1974](https://github.com/0xMiden/miden-vm/pull/1974), but I think it would probably need to be in this PR.

One thing to double check is whether the above changes will require modifying the processor. If so, we'll need to make these updates as well, but I actually don't think we'll need to change anything there immediately.

Once the above is done, we can proceed to #1776 since at that point all operation references should be stable.

plafer:
> If so, we'll need to make these updates as well, but I actually don't think we'll need to change anything there immediately.

The part about the processor inserting noops indeed doesn't _need_ to be changed right away (those code paths will just never be taken anymore). But the `AsmOpInfo` decorator lookup on error to get the `SourceSpan` and `SourceFile` might break, since the meaning of "operation index" changes (it previously assumed no inserted NOOPs).

Not sure if this was addressed already in one of the PRs, but something to keep in mind.

plafer:
closing as stale

PR #1953, opened 2025-07-04
chore(processor): load `MastForest::advice_map` data lazily (#1949)

### LINKED ISSUE #1949: chore(processor): Cache the set of `MastForest`s whose `AdviceMap` has already been merged into the processors `AdviceProviver`
Yes, let's create a separate issue. The overall approach could be to keep an internal "cache" in the `Process` to keep track of already looked up forests (this would look simply like `BTreeMap<Digest, Arc<MastForest>>`). Then, on look ups, if the forest is in this cache, no need to add its advice map to the advice provider and no need to retrieve it from the host.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/1936#discussion_r2180632353_

adr1anh:
## Describe your changes

Fix #1949 

- Move `mast_forests` to `BaseHost`
- Load `MastForest`s into `AdviceProvider` at start of program execution
- Add lazy-loading of MastForests' AdviceMap in `AdviceProvider::get_mapped_value`
  - `get_mapped_value` is now mutable
  - Make `AdviceProvider::merge_advice_map` more private to avoid exposing it through the `ProcessState`
  - Store list of `Arc<MastForest>` in `AdviceProvider` to allow lazy loading of `AdviceMap`s
- Add `Host::mast_forests`
- Remove `AdviceMap::into_parts`
- Improve `AdviceMap::merge`
- Return error for `AdviceProvider::insert_into_map`
- Refactor `host::debug`

- Updated `CHANGELOG.md'


processor/src/host/mod.rs
```diff
@@ -27,6 +27,9 @@ pub trait BaseHost {
     // REQUIRED METHODS
     // --------------------------------------------------------------------------------------------
 
+    /// Returns the list of all available [MastForest]s.
+    fn mast_forests(&self) -> &[Arc<MastForest>];
```
bobbinth: I don't think we can add this method to the host as this would require the host to load all MAS forests it has into memory (which in theory could be a lot). But I also don't think we actually need it. The process could work as follows:

- When encountering an `External` node, the VM asks the host for MAST forest containing the required node via the `Host::get_mast_forest()` method (as it is done now).
- Then, the VM registers this MAST forest with the advice provider. This would require a new method on the advice provider - something like `add_mast_forest()` or `track_mast_forest()` (see next comment).
- The advice provider then adds the MAST forest to the internal list of MAST forests it is tracking.
- If a request for some key-value pair comes in, the advice provider would first check if the key is in the set of already loaded MAST forests, and if not, find and load the appropriate MAST forest (as it does in this PR).

processor/src/host/advice/mod.rs
```diff
@@ -36,9 +42,10 @@ type SimpleMerkleMap = BTreeMap<Word, StoreNode>;
...
 #[derive(Debug, Clone, Default)]
 pub struct AdviceProvider {
-    pub stack: Vec<Felt>,
-    pub map: AdviceMap,
-    pub store: MerkleStore<SimpleMerkleMap>,
+    pub(crate) stack: Vec<Felt>,
+    pub(crate) map: AdviceMap,
+    pub(crate) store: MerkleStore<SimpleMerkleMap>,
+    pub(crate) forests: Vec<Arc<MastForest>>,
 }
```
bobbinth: As mentioned in the previous comment, we could add something like `add_mast_forest()` method to the `AdviceProvier`. It could look something like this:
```rust
pub impl AdviceProvider {

    pub fn add_mast_forest(&mut self, forest: Arc<MastForest>) {
        if forest.has_advice_data() && self.has_forest(forest.clone()) {
            // TODO: add forest to internally tracked forests
        }
    }

}
```
The `self.has_forest()` call would need to tell us whether we've already seen this forest before. Just comparing `Arc`s here is probably not a good idea. One approach could be to compute an identifier for the forest (i.e., by hashing all procedure digests) - but this could get expensive to do every time - so, maybe we should do it once when a `MastForest` is instantiated.

The `AdviceProvider` can then look something like:
```rust
pub struct AdviceProvider {
    stack: Vec<Felt>,
    map: AdviceMap,
    store: MerkleStore<SimpleMerkleMap>,

    /// Forests for which we haven't loaded the advice data into the advice map yet
    pending_forests: Vec<Arc<MastForest>>,

    /// Commitments to all forests added to the advice provider
    forests: BTreeSet<Word>,
}
```
Then, `add_mast_forest()` method could look like:
```rust
pub impl AdviceProvider {

    pub fn add_mast_forest(&mut self, forest: Arc<MastForest>) {
        if forest.has_advice_data() && !self.forests.contains(forest.commitment()) {
            self.forests.insert(forest.commitment());
            self.pending_forests.push(forest.clone());
        }
    }

}
```

bobbinth (commented):
Thank you! Looks good. Not a review yet, but I left a couple of comments inline. The main one is that I don't think we should modify the host interface in this PR.

Also, if we want to make loading of the advice map as efficient as possible, we could consider changing the `AdviceMap` struct to something like:
```rust
pub struct AdviceMap(BTreeMap<Word, Arc<[Felt]>);
```
This way, copying data into the advice provider would, in most cases, but just cloning the `Arc` (this excludes the cases when the key is already in the map, and so we need to compare values).

adr1anh:
Thanks for the excellent suggestions!

The approach makes a lot sense. To summarize:
- load a MastForest into the processor via `Host::get_mast_forest` when a new external procedure is called
- adding this `MastForest` to the `AdviceProvider`
- merging a `MastForest`'s `AdviceMap` only when one of its entries is requested

In the future, we could more granular portions of the `MastForest` to the `AdviceProvider`, namely the procedure-specific advice or the shared advice used by all procedures in the Mast. 

Replacing the `Vec<Felt>` with `Arc<[Felt]>` would also be a good idea. But it also becomes harder to mutate, so I’m not sure it would work for the advice map contained in the provider. It makes sense for the `AdviceMap` inside the `MastForest`.

I also agree with storing a the sequential hash of all (sorted?) procedure roots in the `MastForest` to give them a unique identifier. 

I'll consider which parts make sense to address in this PR and create issues for the rest.

bobbinth:
> Replacing the `Vec<Felt>` with `Arc<[Felt]>` would also be a good idea. But it also becomes harder to mutate, so I’m not sure it would work for the advice map contained in the provider. It makes sense for the `AdviceMap` inside the `MastForest`

AFAIK, we never actually need to mutate individual entries in the advice map (we can add or overwrite entries, but not mutate the existing ones) - so, mutability shouldn't be an issue here.

> I also agree with storing a the sequential hash of all (sorted?) procedure roots in the `MastForest` to give them a unique identifier.

We already have similar logic implemented for the `Library` struct (see [here](https://github.com/0xMiden/miden-vm/blob/next/assembly-syntax/src/library/mod.rs#L209)). Though, there we hash only the roots of exported procedures - but I think this should be equivalent to hashing roots of all procedures in the MAST forest (presumably, if we have a non-exported procedure it is because at some point it is being called by an exported procedure). So, we should be able to just move this logic from `Library` to `MastForest` and then library could rely on its internal `MastForest` for its own digest (i.e., we won't need `Library::digest` field any more).

core/src/advice/map.rs
```diff
@@ -17,32 +21,47 @@ use crate::utils::{ByteReader, ByteWriter, Deserializable, DeserializationError,
...
-pub struct AdviceMap(BTreeMap<Word, Vec<Felt>>);
+pub struct AdviceMap(BTreeMap<Word, Arc<[Felt]>>);
 
 /// Pair representing a key-value entry in an [`AdviceMap`]
-type MapEntry = (Word, Vec<Felt>);
+type MapEntry = (Word, Arc<[Felt]>);
 
 impl AdviceMap {
     /// Creates a new advice map.
     pub fn new() -> Self {
```
adr1anh: Should we remove?

core/src/mast/mod.rs
```diff
@@ -458,6 +458,23 @@ impl MastForest {
             .expect("MAST forest contains more than 2^32 procedures.")
     }
 
+    /// Returns the [Word] representing the hash of the procedures in this forest.
+    pub fn digest(&self) -> Word {
+        self.digest_from_nodes(&self.roots)
+    }
+
+    /// Returns the [Word] representing the content hash of a subset of [`MastNodeId`]s.
+    pub fn digest_from_nodes<'a>(
```
adr1anh: Better name?

core/src/mast/mod.rs
```diff
@@ -458,6 +458,23 @@ impl MastForest {
             .expect("MAST forest contains more than 2^32 procedures.")
     }
 
+    /// Returns the [Word] representing the hash of the procedures in this forest.
+    pub fn digest(&self) -> Word {
```
adr1anh: Would it make sense to cache the digest in the forest?

processor/src/fast/mod.rs
```diff
@@ -144,6 +144,10 @@ pub struct FastProcessor {
     /// return. It is a stack since calls can be nested.
     call_stack: Vec<ExecutionContextInfo>,
 
+    /// Mapping of digests of external procedure digests with their corresponding node ID and
+    /// forest containing the procedure.
+    pub(crate) called_external_procedures: BTreeMap<Word, (MastNodeId, Arc<MastForest>)>,
```
adr1anh: Does a type alias make sense?

processor/src/host/advice/mod.rs
```diff
@@ -143,9 +150,48 @@ impl AdviceProvider {
...
+    /// `MastForest`s.
+    pub fn get_mapped_values(&mut self, key: &Word) -> Result<&[Felt], AdviceError> {
+        // Hot Path: Perform a single lookup.
+        if let Some(values) = self.map.get(key) {
+            // --- UNSAFE ---
+            // SAFETY: This is safe because this branch terminates immediately with a return.
+            // No mutation of `self` happens, so the returned reference cannot be
+            // invalidated by later operations in this function. We are manually
+            // telling the compiler that the aliasing it fears does not actually occur.
+            unsafe {
```
adr1anh: The alternative is for `AdviceMap::get` to return `Option<Arc[Felt]>` which would always clone the pointer.

processor/src/lib.rs
```diff
@@ -881,6 +887,22 @@ impl<'a> ProcessState<'a> {
             ProcessState::Fast(state) => state.processor.memory.get_memory_state(ctx),
         }
     }
+
+    //
+    // HELPERS
+    // --------------------------------------------------------------------------------------------
+
+    pub(crate) fn cached_external_procedure(
```
adr1anh: Better name?

processor/src/host/mod.rs
```diff
@@ -107,7 +107,6 @@ impl SyncHost for DefaultHost {
     fn get_mast_forest(&self, node_digest: &Word) -> Option<Arc<MastForest>> {
         self.store.get(node_digest)
     }
-
```
bobbinth: nit: I would revert this change.

processor/src/host/advice/mod.rs
plafer: Let's not have unsafe code here (unless we really benchmark that this ends up speeding up an important class of programs). A slightly less efficient but probably very fine option is:

```rs
    pub fn get_mapped_values(&mut self, key: &Word) -> Result<&[Felt], AdviceError> {
        if self.map.contains_key(key) {
            return Ok(self.map.get(key).unwrap());
        }

        if let Some(mast_forest) =
            self.pending_forests.pop_if(|mf| mf.advice_map().contains_key(key))
        {
            self.merge_advice_map(mast_forest.advice_map())?;
            Ok(self.map.get(key).unwrap())
        } else {
            Err(AdviceError::MapKeyNotFound { key: *key })
        }
    }
```

core/src/advice/map.rs
plafer: I think clippy complains if you have a `Default` implementation with a `new()`

core/src/advice/map.rs
```diff
@@ -17,32 +21,47 @@ use crate::utils::{ByteReader, ByteWriter, Deserializable, DeserializationError,
...
     }
 
     /// Returns the values associated with given key.
     pub fn get(&self, key: &Word) -> Option<&[Felt]> {
-        self.0.get(key).map(|v| v.as_slice())
+        self.0.get(key).map(|v| &**v)
+    }
+
+    /// Returns a pointer to the values associated with given key.
+    pub fn get_arc(&self, key: &Word) -> Option<&Arc<[Felt]>> {
```
plafer: If we're going to "lock" ourselves to exposing the underlying `Arc`, does `get()` provide any value anymore? i.e. should `get()` return `Option<&Arc<[Felt]>>`? `Arc<[Felt]` derefs to `&[Felt]`, so should work without breaking user code?

core/src/advice/map.rs
```diff
@@ -17,32 +21,47 @@ use crate::utils::{ByteReader, ByteWriter, Deserializable, DeserializationError,
...
+
+    /// Returns true if the key has a corresponding value in the map.
+    pub fn contains_key(&self, key: &Word) -> bool {
+        self.0.contains_key(key)
     }
 
     /// Inserts a key value pair in the advice map and returns the inserted value.
-    pub fn insert(&mut self, key: Word, value: Vec<Felt>) -> Option<Vec<Felt>> {
-        self.0.insert(key, value)
+    pub fn insert(&mut self, key: Word, value: Vec<Felt>) -> Option<Arc<[Felt]>> {
```
plafer: Should we make `value: Arc<[Felt]>`, so that if someone already has the value in this form, we can reuse the `Arc`? And if they have a `Vec`, they just need to `Arc::from()` it

core/src/advice/map.rs
```diff
@@ -53,87 +72,98 @@ impl AdviceMap {
...
         }
-        Ok(())
+    }
+
+    /// Finds the first key that exists in both `self` and `other` with different values.
+    ///
+    /// # Returns
+    /// - `Some` containing the conflicting key, its value from `self`, and the value from `other`.
+    /// - `None` if there are no conflicting values.
+    pub fn find_overlapping_entry(&self, other: &Self) -> Option<(MapEntry, Arc<[Felt]>)> {
```
plafer:
```suggestion
    pub fn find_conflicting_entry(&self, other: &Self) -> Option<(MapEntry, Arc<[Felt]>)> {
```
better?

core/src/advice/map.rs
```diff
@@ -53,87 +72,98 @@ impl AdviceMap {
...
+                    return Some(((*key, existing_value.clone()), new_value.clone()));
+                }
+            }
+        }
+        // No conflicts found.
+        None
     }
 }
 
-impl From<BTreeMap<Word, Vec<Felt>>> for AdviceMap {
```
plafer, line 79: Should we keep this version for convenience?

core/src/mast/mod.rs
```diff
@@ -458,6 +458,23 @@ impl MastForest {
...
+    /// Returns the [Word] representing the content hash of a subset of [`MastNodeId`]s.
+    pub fn digest_from_nodes<'a>(
+        &self,
+        node_ids: impl IntoIterator<Item = &'a MastNodeId>,
+    ) -> Word {
+        let digests = BTreeSet::from_iter(node_ids.into_iter().map(|&id| self[id].digest()));
+        digests
+            .into_iter()
+            .reduce(|a, b| crate::crypto::hash::Rpo256::merge(&[a, b]))
+            .unwrap()
```
plafer: This implementation runs the risk of creating a different hash depending on the order the `node_ids` are passed. I *assume* that this is mitigated in `BTreeSet`, but I can't find any such guarantees in the docs, so we shouldn't assume it. One option would be:

- Create a `Vec` from `node_ids` (or probably change interface to accept a `Vec`)
- sort `Vec`
- call `Vec::dedup()`
- run your `reduce(merge())` on that `Vec`

core/src/mast/mod.rs
plafer: I would only add that complexity if we measure it to be a problem (as it requires making sure we e.g. set a `dirty` bit every time the forest is mutated, i.e. a new source of bugs)

core/src/mast/mod.rs
plafer: `nodes_digest()`?

core/src/mast/mod.rs
plafer: Ah - noticed this was our previous `compute_content_hash()`. Well, we should fix it regardless :)

processor/src/fast/mod.rs
plafer: I personally can read this signature just fine, and so prefer it as is

core/src/mast/mod.rs
adr1anh: The set should guarantee ordering.

> Iterators returned by BTreeSet::iter and BTreeSet::into_iter produce their items in order, and take worst-case logarithmic and amortized constant time per item returned.

bitwalker: I think `BTreeSet`/`BTreeMap` would be nigh useless if they _didn't_ guarantee iteration order 😅

processor/src/utils.rs
```diff
@@ -54,11 +55,17 @@ pub(crate) fn split_u32_into_u16(value: u64) -> (u16, u16) {
 /// This helper function is extracted to ensure that [`crate::Process`] and
 /// [`crate::fast::FastProcessor`] resolve external nodes in the same way.
 pub(crate) fn resolve_external_node(
+    process_state: &mut ProcessState<'_>,
```
plafer: Just noticed this now, but `resolve_external_node()` and `resolve_external_node_async()` can become methods of `Process` and `FastProcessor` again (and take `&mut self` instead of `&mut ProcessState<'_>`). The previous reason for this function was to write that logic once between `Process` and `FastProcessor`, but with the `FastProcessor` move to `async`, we no longer reuse the code anyway.

We will also be able to remove `ProcessState::cached_external_procedure`.

processor/src/lib.rs
plafer: As mentioned in previous comment, if we adopt the suggestion there, this will be removed

processor/src/lib.rs
```diff
@@ -234,6 +234,7 @@ pub struct Process {
     max_cycles: u32,
     enable_tracing: bool,
     source_manager: Arc<dyn SourceManager>,
+    loaded_procedures: BTreeMap<Word, (MastNodeId, Arc<MastForest>)>,
```
plafer:
```suggestion
    loaded_forests_by_procedure: BTreeMap<Word, (MastNodeId, Arc<MastForest>)>,
```
Would something like this be better?

core/src/mast/mod.rs
plafer: Ah! Good to know.

core/src/advice/map.rs
adr1anh: Good point!

processor/src/lib.rs
adr1anh: I left it in as a private function to avoid duplicating the `match`

core/src/advice/map.rs
adr1anh: Made the value generic to take `Into<Arc<[T]>>`

processor/src/lib.rs
adr1anh: I initially misunderstood and moved the functions to `ProcessState`, but applied the correct suggestion in the latest commit.

processor/src/host/advice/mod.rs
adr1anh: Done

processor/src/fast/mod.rs
```diff
@@ -965,6 +968,42 @@ impl FastProcessor {
         Ok(())
     }
 
+    /// Analogous to [`Process::resolve_external_node`](crate::Process::resolve_external_node), but
+    /// for asynchronous execution.
+    async fn resolve_external_node_async(
```
```suggestion
    async fn resolve_external_node(
```
(I hate to do this, but we don't use `_async` pre/post-fixes in the fast processor)

processor/src/fast/mod.rs
```diff
@@ -965,6 +968,42 @@ impl FastProcessor {
...
+        &mut self,
+        external_node: &ExternalNode,
+        host: &mut impl AsyncHost,
+    ) -> Result<(MastNodeId, Arc<MastForest>), ExecutionError> {
+        let node_digest = external_node.digest();
+
+        // Check the cache if this procedure has already been called.
+        if let Some((node_id, forest)) = self.loaded_forests_by_procedure.get(&node_digest) {
```
plafer: Oops, I don't think we ever update the cache after we loaded a forest, right? (Same for `Process`)

adr1anh: Thanks for catching that!

plafer (approved):
LGTM!

core/src/mast/mod.rs
```diff
@@ -458,6 +458,23 @@ impl MastForest {
...
+    /// Returns the [Word] representing the content hash of a subset of [`MastNodeId`]s.
+    ///
+    /// # Panics
+    /// This function panics if `node_ids` is empty.
+    pub fn nodes_digest<'a>(&self, node_ids: impl IntoIterator<Item = &'a MastNodeId>) -> Word {
+        let digests: BTreeSet<_> = node_ids.into_iter().map(|&id| self[id].digest()).collect();
+        digests
+            .into_iter()
+            .reduce(|a, b| crate::crypto::hash::Rpo256::merge(&[a, b]))
+            .unwrap()
```
bobbinth: I'm thinking we should change the methodology here a bit:

1. Instead of using digests directly, we should use `LexicographicWord` (available now in `miden-crypto` v0.15.4).
2. Instead of using `Rpo256::merge()`, I think we should use `Rpo256::hash_elements()`. This way, commitment to the forest would be a simple sequential hash of its procedure roots (which should require half as many permutations).

Another thing I'm curious about: for something like standard library, how many roots are we hashing?

core/src/mast/mod.rs
```diff
@@ -458,6 +458,23 @@ impl MastForest {
             .expect("MAST forest contains more than 2^32 procedures.")
     }
 
+    /// Returns the [Word] representing the hash of the procedures in this forest.
+    pub fn digest(&self) -> Word {
+        self.nodes_digest(&self.roots)
+    }
```
bobbinth: Re-computing the digest every time is likely to take more time than redundantly merging advice maps. This is just my intuition, but we should verify (I'm thinking that computing a digest for something like `miden-stdlib` would probably take 20 - 30  µs and copying some memory should probably quite a bit faster than that).

Ideally, we'd add a field to the `MastForest` to cache the digest, but this is going to be tricky. The main reason is that we don't want to re-compute the digest every time the forest is mutated, and we have a bunch of methods that mutate the forest incrementally during construction. So, we may have to re-factor how `MastForestBuilder` works for this to happen correctly.

Also, I would rename this method to `commitment()` as what we are actually computing is a commitment to the MAST forest.

assembly-syntax/src/library/mod.rs
```diff
@@ -83,7 +78,7 @@ impl Library {
             }
         }
 
-        let digest = compute_content_hash(&exports, &mast_forest);
+        let digest = mast_forest.nodes_digest(exports.values());
```
bobbinth: I wonder if having library commitment be different from the commitment to the underlying `MastForest` is going to cause issues somehow. I can't think of a way it could, but it also feels odd to have this discrepancy.

processor/src/fast/mod.rs
```diff
@@ -144,6 +143,10 @@ pub struct FastProcessor {
     /// return. It is a stack since calls can be nested.
     call_stack: Vec<ExecutionContextInfo>,
 
+    /// Mapping of digests of external procedure digests with their corresponding node ID and
+    /// forest containing the procedure.
+    loaded_forests_by_procedure: BTreeMap<Word, (MastNodeId, Arc<MastForest>)>,
```
bobbinth: nit: I think the word "digests" may be used redundantly here. Should it be more like: "Mapping of external procedure digests..."?

processor/src/fast/mod.rs
```diff
@@ -965,6 +968,44 @@ impl FastProcessor {
...
+        &mut self,
+        external_node: &ExternalNode,
+        host: &mut impl AsyncHost,
+    ) -> Result<(MastNodeId, Arc<MastForest>), ExecutionError> {
+        let node_digest = external_node.digest();
+
+        // Check the cache if this procedure has already been called.
+        if let Some((node_id, forest)) = self.loaded_forests_by_procedure.get(&node_digest) {
+            return Ok((*node_id, forest.clone()));
+        }
```
bobbinth: It may be better to split the PR into two:

1. The PR that implements this structure - i.e., caching already loaded forests in the `FastProcessor`/`Process` structs.
2. The PR that adds commitment computation for `MastForest`.

I think with the first PR we will already get 80% of the benefit because:
- We would ever only need to merge advice maps once per loaded forest.
- Merging of maps with unique keys should be relatively fast (because we just need to copy an `Arc`).

The added benefit of the second PR would be that we could skip merging maps until a procedure tries to access their key. This is definitely useful, but if it comes with the complexity I mentioned in the previous comments, we could come back to it later.

core/src/advice/map.rs
```diff
@@ -53,87 +62,104 @@ impl AdviceMap {
...
+        if let Some(conflict) = self.find_conflicting_entry(other) {
+            Err(conflict)
+        } else {
+            self.merge_new(other);
+            Ok(())
+        }
+    }
+
+    /// Merges entries from `other`, but only for keys not already present in `self`.
+    pub fn merge_new(&mut self, other: &Self) {
```
bobbinth: Does this need to be public?

processor/src/host/advice/mod.rs
```diff
@@ -36,9 +41,11 @@ type SimpleMerkleMap = BTreeMap<Word, StoreNode>;
...
 #[derive(Debug, Clone, Default)]
 pub struct AdviceProvider {
-    pub stack: Vec<Felt>,
-    pub map: AdviceMap,
-    pub store: MerkleStore<SimpleMerkleMap>,
+    stack: Vec<Felt>,
+    map: AdviceMap,
+    store: MerkleStore<SimpleMerkleMap>,
+    pending_forests: Vec<Arc<MastForest>>,
+    loaded_forest_digests: BTreeSet<Word>,
```
bobbinth: As I mentioned in one of the previous comments, maybe it would be a good idea to move these changes into a separate PR.

core/src/advice/map.rs
adr1anh: I've left it public so that `resolve_external_node` can efficiently merge maps, though it doesn't check for conflicts.

adr1anh:
> It may be better to split the PR into two:
> 
> 1. The PR that implements this structure - i.e., caching already loaded forests in the `FastProcessor`/`Process` structs.
> 2. The PR that adds commitment computation for `MastForest`.
> 
> I think with the first PR we will already get 80% of the benefit because:
> 
> * We would ever only need to merge advice maps once per loaded forest.
> * Merging of maps with unique keys should be relatively fast (because we just need to copy an `Arc`).
> 
> The added benefit of the second PR would be that we could skip merging maps until a procedure tries to access their key. This is definitely useful, but if it comes with the complexity I mentioned in the previous comments, we could come back to it later.

This makes sense, but at the moment it means we're still "merging" the map every time we encounter a new external node, even from a forest who's map has already been merged. The question is whether we're okay with merging the map without checking if any keys overlap is safe. Could we make the assumption that the host is responsible for checking the validity of the `MastForest`s?

We're also caching calls to the Host with the `loaded_external_nodes` map. It's not clear how necessary this is, so I'm happy to remove it. For the follow up PR, we'll want to cache the map `MastDigest -> MastForest`, either in the processor or the `AdviceProvider` to keep track of unique forests and avoid checking for conflict unnecessarily. With this approach, checking for duplicates during a merge would only happen once.

adr1anh:
I ended up reverting a lot of the changes to keep this PR within the new scope. The actual changes are now 
- Improving the API and implementation of `AdviceMap`
  - Entries are stored as `Arc` instead of `Vec`.
  - Introduce `merge_new` allowing for faster but unchecked merging without comparing existing existing values.
- Introduce `MastForest::nodes_digest` for computing the digest to a set of nodes. This method replaces `compute_content_hash` from `assembly_syntax::library` and computes hashes more efficiently.
- Move `resolve_external_node` from `processor::utils` to `Process` and `FastProcessor`
- Adapt `AdviceProvider` and `AdviceInputs` to `Arc<[Felt]>`
  - Make `AdviceProvider::get_mapped_value` mutable in anticipation for actual lazy loading.
  - Return an error for `AdviceProvider::insert_into_map` if the key already exists.

core/src/mast/mod.rs
```diff
@@ -458,6 +456,13 @@ impl MastForest {
             .expect("MAST forest contains more than 2^32 procedures.")
     }
 
+    /// Returns the [Word] representing the content hash of a subset of [`MastNodeId`]s.
+    pub fn nodes_digest<'a>(&self, node_ids: impl IntoIterator<Item = &'a MastNodeId>) -> Word {
```
bobbinth: nit (feel free to ignore): maybe this could be `compute_nodes_digest()` or even `compute_nodes_commitment()`? I usually try to start a function name with an action verb.

core/src/advice/map.rs
bobbinth: I'm not seeing it being used there (maybe things have changed since this comment). But if it were, I'd probably use `AdviceMap::merge()` instead. The reason is: if, for w/e reason, two MAST forests which contain entries with the same key but different values were attempted to be loaded, we should error out - otherwise, we may get weird errors later on as at least one of the procedures would be executing under wrong assumptions.

And then, we could make this method private.

processor/src/host/advice/mod.rs
```diff
@@ -138,9 +141,25 @@ impl AdviceProvider {
...
+        self.map.contains_key(key)
+    }
+
     /// Returns a reference to the value(s) associated with the specified key in the advice map.
-    pub fn get_mapped_values(&self, key: &Word) -> Result<&[Felt], AdviceError> {
-        self.map.get(key).ok_or(AdviceError::MapKeyNotFound { key: *key })
+    ///
+    /// # Details
+    /// This function is mutable since we will want this method to lazily load advice maps
+    /// from external [`MastForest`](crate::MastForest)s.
```
bobbinth: I think this part of the comment can be deleted.

processor/src/host/advice/mod.rs
```diff
@@ -149,18 +168,23 @@ impl AdviceProvider {
     /// the [AdviceProvider::push_stack()] method.
```
bobbinth: Should this not be `push_from_map()`?

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -645,12 +655,12 @@ fn push_transformed_stack_top(
...
 
 fn get_smt_leaf_preimage(
-    process: &ProcessState,
+    process: &mut ProcessState,
     node: Word,
     err_ctx: &impl ErrorContext,
 ) -> Result<Vec<(Word, Word)>, ExecutionError> {
     let kv_pairs = process
-        .advice_provider()
+        .advice_provider_mut()
```
bobbinth: These changes could be reverted, right?

PR #1954, opened 2025-07-05
Use crate dependency in core for serialization types

bobbinth:
This small PR switches to using `crate::utils` as the source of serialization types in the `miden-core` create (instead of importing them from `winter-utils`).

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -15,9 +15,11 @@ pub use node::{
     BasicBlockNode, CallNode, DynNode, ExternalNode, JoinNode, LoopNode, MastNode, MastNodeExt,
     OP_BATCH_SIZE, OP_GROUP_SIZE, OpBatch, OperationOrDecorator, SplitNode,
 };
-use winter_utils::{ByteWriter, DeserializationError, Serializable};
 
-use crate::{AdviceMap, Decorator, DecoratorList, Felt, Operation, Word};
+use crate::{
+    AdviceMap, Decorator, DecoratorList, Felt, Operation, Word,
+    utils::{ByteWriter, DeserializationError, Serializable},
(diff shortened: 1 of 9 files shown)


bitwalker (approved):
/cc @plafer 

I'm approving this based on the idea that we're doing the same thing here that I did in #1945, i.e. scoping the set of things we use from e.g. `winter_utils` so that it is easier to enforce what is used from the upstream crate via code review - ideally with the goal to remove the dependency in the future. However, with the serialization bits from `winter-utils` the more likely scenario is that we would be replacing it with `serde` at some point, but that's not happening in the near term anyway.

I think in general though, both @plafer and I feel that being more explicit with imports is preferable vs re-exporting aliases from the crate root. The one exception to that IMO is what I outlined above. We definitely need to define a policy around it though and try to standardize on it, because we're all over the place with that stuff today.

bobbinth:
> I'm approving this based on the idea that we're doing the same thing here that I did in #1945, i.e. scoping the set of things we use from e.g. `winter_utils` so that it is easier to enforce what is used from the upstream crate via code review - ideally with the goal to remove the dependency in the future.

Basically yes - we will need to remove the dependency on `winter-utils` relatively soon (e.g., in 2 - 3 months) and having fewer places in the code that explicitly depend on it should simplify this task.

> I think in general though, both @plafer and I feel that being more explicit with imports is preferable vs re-exporting aliases from the crate root. The one exception to that IMO is what I outlined above. We definitely need to define a policy around it though and try to standardize on it, because we're all over the place with that stuff today.

Generally agree. The main motivation for re-exporting things through `miden-core` was to make it more flexible to move off of Winterfell dependencies in the future. But it didn't work out as well in practice as I had hoped.

PR #1957, opened 2025-07-06
Refactor `host::debug`

adr1anh:
## Describe your changes

Small refactor of the default debug event handler. 
- Replace `AdviceProvider::peek_stack` with `stack` returning the entire state and handler peeking in the debug implementation directly.
- Remove `Printer` struct by accessing the values directly from the `Process`
- Small simplification of `print_vm_adv_stack`

- Updated `CHANGELOG.md'

### DIFF
--- a/processor/src/host/debug.rs
+++ b/processor/src/host/debug.rs
@@ -1,36 +1,34 @@
 use alloc::vec::Vec;
-use std::{print, println};
+use std::{cmp::min, print, println};
 
-use miden_air::RowIndex;
 use vm_core::{DebugOptions, Felt};
 
 use super::ProcessState;
-use crate::{AdviceProvider, MemoryAddress, system::ContextId};
+use crate::MemoryAddress;
 
 // DEBUG HANDLER
 // ================================================================================================
 
 /// Prints the info about the VM state specified by the provided options to stdout.
 pub fn print_debug_info(process: &ProcessState, options: &DebugOptions) {
-    let printer = Printer::new(process.clk(), process.ctx(), process.fmp());
     match options {
         DebugOptions::StackAll => {
-            printer.print_vm_stack(process, None);
+            print_vm_stack(process, None);
         },
         DebugOptions::StackTop(n) => {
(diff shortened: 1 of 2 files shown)


processor/src/host/debug.rs
```diff
@@ -1,175 +1,170 @@
...
 
-        // we may have less elements than requested
-        let num_items = stack.len();
-        if num_items == 0 {
-            println!("Advice Stack empty before step {}.", self.clk);
-            return;
-        };
+/// Prints length items from the top of the  advice stack. If length is 0 it returns the whole
+/// stack.
+fn print_vm_adv_stack(process: &ProcessState, length: usize) {
```
bobbinth: nit: since we are refactoring anyways, could we rename `length` to `n` and pass it as `Option<usize>` to be consistent `print_vm_stack()`?

bobbinth (approved):
Looks good! Thank you! Not a very thorough review from me, but I left one comment inline.

processor/src/host/debug.rs
adr1anh: It's not much different to use `Option` and IMO is slightly less clear. 
```rust
fn print_vm_adv_stack(process: &ProcessState, n: Option<usize>) {
    let mut stack = process.advice_provider().stack();
    let num_items = if let Some(stack_len) = n {
        min(stack.len(), stack_len)
    } else {
        stack.len()
    };
    stack = &stack[..num_items];
```

we also have to add 
```rust
let n = (length != 0).then(|| length as usize);
```
to the caller. 

I'll change the argument to `n` but leave the rest as is

bitwalker: I think you'd typically do `Option<NonZeroUsize>` to make it clear that _if_ there is a length being specified, that it is non-zero.

plafer (approved):
LGTM - assuming that this doesn't break the printing (I didn't actually run it)

paracetamolo (approved):
LGTM. Nice simplification.

Issue #1960, opened 2025-07-07
Add `get_procedure_root_by_name` method to `Library`


Fumuran:
### Feature description

Add a new `get_procedure_root_by_name` method to the `Library` struct, which will allow to obtain the digest of some library procedure by its name.

### Why is this feature needed?

Basic protocol is checking whether a transaction could be executed against some account by checking whether some library is presented as a component of this account. This check could be simplified by moving the procedure digest obtaining to the `Library` itself. 

This feature will be also useful for well known components: https://github.com/0xMiden/miden-base/pull/1532#discussion_r2187286509

PR #1961, opened 2025-07-07
Add `get_procedure_root_by_name` method to `Library`

Fumuran:
This small PR adds a `get_procedure_root_by_name` method to the `Library` struct. 
This method allows to obtain the procedure digest from the Library using its `QualifiedProcedureName`.

Closes: #1960.

### DIFF
--- a/assembly-syntax/src/ast/procedure/name.rs
+++ b/assembly-syntax/src/ast/procedure/name.rs
@@ -78,6 +78,14 @@ impl TryFrom<&str> for QualifiedProcedureName {
     }
 }
 
+impl TryFrom<String> for QualifiedProcedureName {
+    type Error = Report;
+
+    fn try_from(name: String) -> Result<Self, Self::Error> {
+        Self::from_str(&name)
+    }
+}
+
 impl Eq for QualifiedProcedureName {}
 
 impl PartialEq for QualifiedProcedureName {
--- a/assembly-syntax/src/library/mod.rs
+++ b/assembly-syntax/src/library/mod.rs
@@ -140,7 +140,7 @@ impl Library {
     }
 
     /// Returns the digest of the procedure with the specified name, or `None` if it was not found
-    /// in the Library.
+    /// in the library or its library path is malformed.
     pub fn get_procedure_root_by_name(
         &self,
         proc_name: impl TryInto<QualifiedProcedureName>,


bitwalker (approved):
LGTM!

assembly-syntax/src/library/mod.rs
```diff
@@ -138,6 +138,13 @@ impl Library {
     pub fn mast_forest(&self) -> &Arc<MastForest> {
         &self.mast_forest
     }
+
+    /// Returns the digest of the procedure with the specified name, or `None` if it was not found
+    /// in the Library.
+    pub fn get_procedure_root_by_name(&self, proc_name: QualifiedProcedureName) -> Option<Word> {
+        let node_id = self.exports.get(&proc_name);
+        node_id.map(|id| self.mast_forest()[*id].digest())
+    }
```
bobbinth: Should we use something like `TryInto<QualifiedProcedureName>` for `proc_name` parameter? Then, if we have conversion from strings for `QualifiedProcedureName`, we could do something like:

```rust
let root = mylib.get_procedure_root_by_name("foo::bar::baz").unwrap();
```

bobbinth (approved):
Looks good! Thank you! I left a comment with a potential improvement proposal inline.

assembly/src/tests.rs
```diff
@@ -475,6 +476,51 @@ fn get_module_by_path() -> Result<(), Report> {
...
+    let bar_name = QualifiedProcedureName::new(
+        LibraryPath::new("test::names").unwrap(),
+        ProcedureName::new("bar").unwrap(),
+    );
+    let invalid_name = QualifiedProcedureName::new(
+        LibraryPath::new("test::names").unwrap(),
+        ProcedureName::new("baz").unwrap(),
+    );
+
+    assert_eq!(Some(library_procedure_digests[1]), library.get_procedure_root_by_name(foo_name));
```
plafer:
```suggestion
    assert!(library_procedure_digests.contains(library.get_procedure_root_by_name(foo_name)));
```
nit: just so that if we change the ordering of `exports()` somehow, then it won't break the test.

Same applies to next line

assembly-syntax/src/library/mod.rs
Fumuran: I implemented a `TryFrom<&str> for QualifiedProcedureName` to make it possible to use the strings in this procedure. Is it fine, or there is a way to somehow use `FromStr` conversion we had directly?

bobbinth: @bitwalker also suggested a slightly different approach - something like:
```rust
pub fn get_procedure_root_by_name(&self, proc_name: AsRef<str>) -> Option<Word>
```
Though, I think `TryInto<QualifiedProcedureName>` covers all the same options, and maybe is a bit more flexible (because we can pass in `QualifiedProcedureName` directly)?

assembly-syntax/src/library/mod.rs
```diff
@@ -138,6 +138,20 @@ impl Library {
     pub fn mast_forest(&self) -> &Arc<MastForest> {
         &self.mast_forest
     }
+
+    /// Returns the digest of the procedure with the specified name, or `None` if it was not found
+    /// in the Library.
+    pub fn get_procedure_root_by_name(
```
bobbinth: I would also mention that `None` is returned if the path is malformed.

assembly-syntax/src/ast/procedure/name.rs
```diff
@@ -70,6 +70,14 @@ impl FromStr for QualifiedProcedureName {
     }
 }
 
+impl TryFrom<&str> for QualifiedProcedureName {
+    type Error = Report;
+
+    fn try_from(name: &str) -> Result<Self, Self::Error> {
+        Self::from_str(name)
+    }
+}
```
bobbinth, line 79: Should we also add `TryFrom<String> for QualifiedProcedureName` for convenience?

bobbinth (approved):
Looks good! Thank you! I left a couple small comments inline.

assembly-syntax/src/library/mod.rs
Fumuran: I think that it will be useful to be able to pass the `QualifiedProcedureName` directly, because in the `miden-base` I will do exactly that. 

We can't use the string procedure name without the library path (the proc name like `foo` will return an error), so in case we want to get the proc without it (or with the default library path, to be precise) we can just create a `QualifiedProcedureName` directly using `Default::default()` as a library path.

Probably we can use `#anon` path to do the same, but IMO using `QualifiedProcedureName` will be more clean here.

PR #1884, opened 2025-06-16
Implement Event Handlers

adr1anh:
## Describe your changes

Fixes #1584

## Summary

This PR introduces a foundational framework for modular event handling in the VM, centered around a new `EventHandler` trait and `DefaultHost` implementation. It enables associating event handlers with MAST forests (via `HostLibrary`, formerly `MastForestSource`), which can be used by clients or standard library packages to define custom event behavior. These handlers are intended to be simple, stateless, and sync for now, with a pathway to richer or async handling in future iterations.

---

## Motivation

The core motivation is to reduce boilerplate and centralize event logic that would otherwise be embedded directly in `Host::on_event()` match arms. Specifically, system-level operations like `U64Div`, `FalconDiv`, and `Ext2Inv`, which belong more naturally in the standard library, can now be handled through composable, testable event handlers.

This lays groundwork for future precompile infrastructure (e.g., WASM-based handlers), without introducing premature complexity. By decoupling handler registration from core processor logic, this design helps streamline host development (e.g. `TransactionHost`) and opens the door for WASM packages to export their own logic cleanly.

---

## Highlights

- **New `EventHandler` trait**:
  - Can be implemented by free functions, closures, or structs with access to their own internal state.
  - Must be `Send + Sync + 'static` and implement `fn on_event(&self, &mut ProcessState) -> Result<(), EventError>`.

- **`EventError` and error propagation**:
  - Event handlers return a boxed error type.
  - Errors are wrapped at the processor level in an `ExecutionError`, preserving context and providing structured failure reporting.
  - Avoids the need for passing `ErrorContext` into handler methods directly.

- **`EventHandlerRegistry`**:
  - Stores handlers by event ID and dispatches them on request.
  - Fails fast on duplicate registration.

- **`DefaultHost`**:
  - Provides an extensible and reusable host for tests and real execution.
  - Supports loading MAST forests and event handlers via `HostLibrary`.
  - Allows setting custom debug handlers.

- **New `HostLibrary` trait**:
  - Allows bundling a `MastForest` with its associated handlers.
  - It is implemented for simple `Arc<MastForest>`s and the `Stdlib` (default: no handlers), or by richer "packages".

---

## Follow-ups & Open Questions

This PR doesn't finalize the full event handling story but offers a clean MVP. Discussion has raised the following open directions:

- **Naming and design**:
  - `HostLibrary` is a stopgap; candidates like `RichLibrary`, `CoreLibrary`, or `Package` may better reflect intended usage.
[...]

Issue #1942, opened 2025-07-02
Investigate making the `SourceManager` part of the `Host`


plafer:
This would involve transferring the `SourceManager` out of the processor, into the `Host`. This sounds like a good idea at a high level, but more investigation is needed.

[Context](https://github.com/0xMiden/miden-vm/pull/1936#discussion_r2179662852)

bobbinth:
Closed by the work that went into https://github.com/0xMiden/miden-vm/pull/2015.

PR #1967, opened 2025-07-08
Introduce `SourceManagerSync`

### LINKED ISSUE #1966: Introduce `SourceManagerSync` trait
We should introduce a `trait SourceManagerSync: SourceManager + Send + Sync`. This is meant to allow use cases that need a `Send + Sync` source manager (e.g. VM's processor, `miden-base`, & everything downstream), without forcing those that don't need to work in a multithreaded environment to pay the cost of synchronization primitives (e.g. the compiler and assembler).

As discussed offline, this will mean that we remove the `Assembler::assembler()` method (which currently returns a `dyn SourceManager + Send + Sync`) to allow the assembler to use just `dyn SourceManager` internally.

Further context: https://github.com/0xMiden/miden-vm/pull/1884#discussion_r2192680902

plafer:
Closes #1966 

This is a stacked PR on top of #1945 to avoid breaking changes when we merge that one.

Note that contrary to what was discussed in https://github.com/0xMiden/miden-vm/pull/1884#discussion_r2188948366, `ErrorContextImpl` uses `Arc<dyn SourceManager>` (instead of `SourceManagerSync`), and hence is not `Send` or `Sync` (and hence cannot be sent across threads, or be included in the returned `Future` of our `AsyncHost` methods). This is because it is actually not *needed* at this moment, so I prefer to only include the bounds if we find that we do indeed need to send error contexts across threads (which we currently don't).

### DIFF
--- a/assembly-syntax/src/testing/context.rs
+++ b/assembly-syntax/src/testing/context.rs
@@ -22,7 +22,7 @@ use crate::{
 /// Some of the assertion macros defined in this crate require a [SyntaxTestContext], so be aware of
 /// that.
 pub struct SyntaxTestContext {
-    source_manager: Arc<dyn SourceManager + Send + Sync>,
+    source_manager: Arc<dyn SourceManager>,
     warnings_as_errors: bool,
 }
 
@@ -66,7 +66,7 @@ impl SyntaxTestContext {
     }
 
     #[inline(always)]
-    pub fn source_manager(&self) -> Arc<dyn SourceManager + Send + Sync> {
+    pub fn source_manager(&self) -> Arc<dyn SourceManager> {
         self.source_manager.clone()
     }
 
--- a/assembly/src/testing.rs
+++ b/assembly/src/testing.rs
@@ -27,7 +27,7 @@ use crate::diagnostics::reporting::set_panic_hook;
 ///
 /// Some of the assertion macros defined above require a [TestContext], so be aware of that.
 pub struct TestContext {
-    source_manager: Arc<dyn SourceManager + Send + Sync>,
+    source_manager: Arc<dyn SourceManager>,
     assembler: Assembler,
 }
 
@@ -71,7 +71,7 @@ impl TestContext {
     }
 
     #[inline(always)]
-    pub fn source_manager(&self) -> Arc<dyn SourceManager + Send + Sync> {
+    pub fn source_manager(&self) -> Arc<dyn SourceManager> {
         self.source_manager.clone()
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - [BREAKING] Incremented MSRV to 1.88.
 
+#### Changes
+- [BREAKING] Introduce `SourceManagerSync` trait, and remove `Assembler::source_manager()` method [#1966](https://github.com/0xMiden/miden-vm/issues/1966)
+
 ## 0.16.0 (2025-07-08)
 
 #### Enhancements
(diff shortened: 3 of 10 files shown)


assembly-syntax/src/testing/context.rs
```diff
@@ -22,7 +22,7 @@ use crate::{
 /// Some of the assertion macros defined in this crate require a [SyntaxTestContext], so be aware of
 /// that.
 pub struct SyntaxTestContext {
-    source_manager: Arc<dyn SourceManager + Send + Sync>,
+    source_manager: Arc<dyn SourceManagerSync>,
```
bitwalker: *Question:* Does this actually need to be `SourceManagerSync` rather than `SourceManager` here? AFAIK all of the syntax-related tests are single-threaded and don't call into anything that requires a `Sync` source manager.

assembly/src/testing.rs
```diff
@@ -27,7 +27,7 @@ use crate::diagnostics::reporting::set_panic_hook;
 ///
 /// Some of the assertion macros defined above require a [TestContext], so be aware of that.
 pub struct TestContext {
-    source_manager: Arc<dyn SourceManager + Send + Sync>,
+    source_manager: Arc<dyn SourceManagerSync>,
```
bitwalker: Same question here with regard to whether the test context really needs `SourceManagerSync`.

crates/debuginfo/src/source_manager.rs
```diff
@@ -254,6 +254,14 @@ pub trait SourceManagerExt: SourceManager {
 #[cfg(feature = "std")]
 impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 
+/// [SourceManagerSync] is a [SourceManager] that can be used in a multi-threaded context.
+///
+/// This trait should not be implemented directly; there is a blanket implementation for all types
+/// that implement [SourceManager] + [Send] + [Sync].
```
bitwalker: I think I would just phrase this as:

> [SourceManagerSync] is a marker trait for [SourceManager] implementations that are also `Send + Sync`, and is automatically implemented for any [SourceManager] that meets those requirements.
>
> [SourceManager] is a supertrait of [SourceManagerSync], so you may use instances of the latter where the former is required, either implicitly or via explicit downcasting, e.g. `Arc<dyn SourceManagerSync> as Arc<dyn SourceManager>`.

The compiler won't let you implement `SourceManagerSync` manually anyway, so it isn't actually necessary to instruct consumers not to implement it.

bitwalker (approved):
LGTM! One question and one suggestion, but I'm marking this approved since neither are blockers.

assembly-syntax/src/testing/context.rs
plafer: That was a find-and-replace - fixed now

assembly/src/testing.rs
plafer: Fixed

crates/debuginfo/src/source_manager.rs
plafer: Done

plafer:
> Regarding making the ErrorContext sync, your comment makes sense but I’m not exactly sure what that means in the context of the error context passed to the handle_event method.

I think you're referring to e.g. `AsyncHost::on_event()` right? What I was hoping is that `on_event()` returns an error that doesn't need an `ErrorContext` (e.g. `Box<dyn Error>`), and when wrap the error with the proper error context in the thread where `on_event()` was called (such that `ErrorContext` doesn't need to be passed around). Also makes for a cleaner API - similar to what you started doing with `AdviceError`, etc. But indeed if that turns out not to work so well, we can certainly force `ErrorContext` to be `Send + Sync`.

bobbinth:
As a part of this refactoring, did we not want to make `FastProcessor` (and `Process`) to take `SourceManagerSync`? Or we are deferring on this because this will be reworked in https://github.com/0xMiden/miden-vm/issues/1942 ?

plafer:
Yeah in this PR I just went for what was strictly necessary - if at any point the `FastProcessor` needs to send the `SourceManager` across threads (or the `SourceManager` is moved into the `Host`), then we can switch to using `SourceManagerSync`.

Issue #1968, opened 2025-07-08
Add `hash_memory_double_words` and `copy_digest` procs to `std::crypto::hashes::rpo`


Fumuran:
We should add two new procedures to the `std::crypto::hashes::rpo` module. 
- `hash_memory_double_words` procedure should do the same as `absorb_double_words_from_memory`, but in addition internally preparing the hasher state. 
- `copy_digest` procedure should copy the output of the hashing procedures leaving the hasher state unchanged: `[A, B, C] -> [B, A, B, C]`.

PR #1971, opened 2025-07-09
Implement `hash_memory_double_words` and `copy_digest` procs in `rpo` module

Fumuran:
This small PR adds two new procedures to the `std::crypto::hashes::rpo` module.
- `hash_memory_double_words` works the same as `absorb_double_words_from_memory`, but it doesn't require the stack to be prepared for hashing (it should not contain the hasher state).
- `copy_digest` copies the result of the hash permutation (second word) leaving the state unchanged.

Closes: #1968.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,13 @@
 #### Changes
 
 - [BREAKING] Incremented MSRV to 1.88.
-- [BREAKING] Introduce `SourceManagerSync` trait, and remove `Assembler::source_manager()` method [#1966](https://github.com/0xMiden/miden-vm/issues/1966)
 - [BREAKING] Implement preliminary changes for lazy loading of external `MastForest` `AdviceMap`s ([#1949](https://github.com/0xMiden/miden-vm/issues/1949)).
+- [BREAKING] Introduce `SourceManagerSync` trait, and remove `Assembler::source_manager()` method [#1966](https://github.com/0xMiden/miden-vm/issues/1966).
+
+## 0.16.1 (2025-07-10)
+
+- Make `Process::state()` public and re-introduce `From<&Process> for ProcessState`.
+- Return `AdviceProvider` as part of the `ExecutionTrace`.
 
 ## 0.16.0 (2025-07-08)
 
--- a/assembly-syntax/README.md
+++ b/assembly-syntax/README.md
@@ -1,6 +1,6 @@
-# miden-utils-diagnostics
+# miden-assembly-syntax
 
-This crate provides diagnostic and reporting utilities used throughout Miden's toolchain components.
+This crate provides parsing and semantic analysis of the Miden Assembly language.
 
 ## License
 This project is dual-licensed under the [MIT](http://opensource.org/licenses/MIT) and [Apache 2.0](https://opensource.org/license/apache-2-0) licenses.
--- a/crates/debug/types/README.md
+++ b/crates/debug/types/README.md
@@ -1,6 +1,6 @@
-# miden-utils-diagnostics
+# miden-debug-types
 
-This crate provides diagnostic and reporting utilities used throughout Miden's toolchain components.
+This crate contains core source-level debugging information types used throughout the Miden toolchain.
 
 ## License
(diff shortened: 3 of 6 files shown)


stdlib/tests/crypto/rpo.rs
```diff
@@ -243,6 +243,43 @@ fn test_absorb_double_words_from_memory() {
     build_test!(even_words, &[]).expect_stack(&even_hash);
 }
 
+#[test]
+fn test_hash_memory_double_words() {
```
bobbinth, line 247: Could we also test the case the hashing an empty input (i.e., `start_addr == end_addr`) produces an `EMPTY_WORD`.

bobbinth (approved):
Looks good! Thank you!

Could we also add the `rpo` module as a sub-section [here](https://0xmiden.github.io/miden-vm/user_docs/stdlib/crypto/hashes.html) (I didn't realize we were missing the whole module there).

docs/src/user_docs/stdlib/crypto/hashes.md
```diff
@@ -9,6 +9,20 @@ Module `std::crypto::hashes::blake3` contains procedures for computing hashes us
 | hash_1to1   | Computes BLAKE3 1-to-1 hash.<br/><br/>Input: 32-bytes stored in the first 8 elements of the stack (32 bits per element).<br /> <br/>Output: A 32-byte digest stored in the first 8 elements of stack (32 bits per element). |
 | hash_2to1   | Computes BLAKE3 2-to-1 hash.<br/><br/>Input: 64-bytes stored in the first 16 elements of the stack (32 bits per element).<br /> <br/>Output: A 32-byte digest stored in the first 8 elements of stack (32 bits per element) |
 
+## RPO
```
bobbinth: Let's change this to `RPO256` and move this section after `SHA256`.

Also, Let's fix a typo on line 2 above: `std::crypto` -> `std::crypto::hashes`.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 0.17.0 (TBD)
 
+#### Enhancements
+
+- Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
```
plafer, line 7: nit: we now link to the issue (when there is one)

bobbinth: What was the reasoning for switching to issue links? Issues could have long discussions and potentially be covered by a couple of PRs. I think linking to a PR in which the change was made may be easier to figure out what actually happened.

plafer: > Issues could have long discussions and potentially be covered by a couple of PRs.

Most issues are handled by a single PR (or we'll often create a sub-issue), but when this doesn't apply for any reason then yes we would prefer linking the PR.

> What was the reasoning for switching to issue links?

> I think linking to a PR in which the change was made may be easier to figure out what actually happened.

"Historically", @adr1anh started doing it, and I happened to prefer that too, so we pushed to start linking issues instead of PRs.

Basically normally the issue contains the description of the work to be done, additional context for why we're doing it, etc, while the PR is at a much lower level. After reading the issue, anyone interested in the details can easily dig some more by opening the PR that closed the issue (which is clearly indicated in the Github UI). Personally, if a changelog entry I'm interested in links to the PR, the first thing I'll do is open the associated issue for the high level description.

But this is obviously a matter of personal preference; we could discuss it in our meeting next week. There might be a nice way to link both the issue and the PR in a standardized way.

bobbinth: > Basically normally the issue contains the description of the work to be done, additional context for why we're doing it, etc, while the PR is at a much lower level. After reading the issue, anyone interested in the details can easily dig some more by opening the PR that closed the issue (which is clearly indicated in the Github UI).

Right, but frequently during the implementation we come across things that sometimes change things slightly but sometimes change a lot as compared to the original issue description. So, we may end up linking to something that is no longer relevant, or at least not the latest in terms of what was actually done.

> Personally, if a changelog entry I'm interested in links to the PR, the first thing I'll do is open the associated issue for the high level description.

I think this may be an indication that the PRs we create should contain more detailed description of what was done and how - but yes, let's discuss this in a meeting.

Issue #1965, opened 2025-07-08
MastForest serialization should preserve basic_blocks groups


paracetamolo:
Currently we assume that a basic_block can uniquely be constructed starting from a vector of operations (and optionally decorators). For this reason the de/serializer does not bother with the internal structure of the basic_blocks but simply stores the vector of operations (and decorators).

In the context of adding Noop during assembly rather than during execution by the Processor (https://github.com/0xMiden/miden-vm/pull/1950), the introduced Noops cause ambiguities. They may be included in an existing group or have their dedicated group as required by the padding rules. This is further complicated by the fact that decorators may be attached to a Noop.

If the de/serializer stored the whole structure of the basic_blocks instead of reconstructing it, it would significantly simplify inserting Noop https://github.com/0xMiden/miden-vm/pull/1950 and the followup work on DebugInfo https://github.com/0xMiden/miden-vm/pull/1940.

bobbinth:
Serializing more information about the structure of the basic block would work here. Specifically, we could add one more vector containing number of operations per group for all op groups in the basic block. For example, (omitting the decorator part), what we'd serialize could conceptually look like so:

```rust
struct SerializedBasicBlock {
    ops: Vec<Operation>,
    group_sizes: Vec<u8>,
}
```
Using a `u8` for each group is sufficient because a group cannot have more than 9 operations. In fact, we could probably get away with using 4 bits per group to cut down the size - but it may not be worth the trouble.

-------

An alternative approach would be to change the definition of `Operation::Noop` variant to look like `Operation:Noop(bool)`, where the internal value would have the following meaning:

- `true`: the noop was inserted by the user.
- `false`: the noop was inserted by the assembler for padding purposes.

The execution semantics of the noop would remain the same as the are now. But during serialization, we'd also serialize the internal value (we could get clever hear and serialize it in such a way that doesn't add any extra space for serializing `Noop` nodes). This way, we could still serialize just `Vec<Operation>` and then during de-serialization we should be able to ambiguously tell how a sequence of operations is to be deserialized (in part, because `Operation::Noop(false)` can only be the last operation in a group).

My intuition is that the second approach could be both easier to implement (should result in minimal changes to the existing structure) and would be more efficient in general - but I also haven't thought through all possible implications.

paracetamolo:
I think I have a solution here https://github.com/0xMiden/miden-vm/pull/1974

[...]

PR #1974, opened 2025-07-10
Distinguish Noop required for padding when serializing

paracetamolo:
In principle a basic block can be serialized as a sequence of operations. However, when we meet a Noop during deserialization, it is unclear if it was inserted by the user or by the padding rules in [batch_ops]. 
If it was introduced by the user it should be preserved and treated like any other operation, for example it should be grouped together with other operations in a group. 
If it was introduced by padding then it should create its own group or be the trailing operation of a group.

In order to distinguish these two cases, during serialization each Noop is followed by an extra boolean flag. During deserialization, if a Noop is coming from padding, it is not inserted in the sequence of operations and it will be automatically inserted again by the padding rules. If the Noop is not padding, it is included in the sequence like any other operation.

Fixes https://github.com/0xMiden/miden-vm/issues/1965

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -422,6 +422,13 @@ fn batch_ops(ops: Vec<Operation>) -> Vec<OpBatch> {
         batches.push(batch);
     }
 
+    // If the block is empty, we insert a single batch with a Noop
+    if batches.is_empty() {
+        batch_acc = OpBatchAccumulator::new();
+        batch_acc.add_op(Operation::Noop);
+        batches.push(batch_acc.into_batch());
+    }
+
     batches
 }
 
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -107,9 +107,8 @@ impl BasicBlockNode {
 
     /// Returns a new [`BasicBlockNode`] from values that are assumed to be correct.
     /// Should only be used when the source of the inputs is trusted (e.g. deserialization).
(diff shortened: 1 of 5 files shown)


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,3 +448,68 @@ fn validate_decorators(operations: &[Operation], decorators: &DecoratorList) {
...
+*/
+
+pub fn write_batches_into<W: ByteWriter>(batches: &[OpBatch], target: &mut W) {
+    let num_ops: usize = batches
+        .iter()
+        .flat_map(|b| b.op_counts())
+        .cloned()
+        .reduce(|x, y| x + y)
+        .unwrap();
+    target.write_u16(num_ops.try_into().unwrap());
```
bobbinth, line 480: Using `u16` here may be overly conservative (there could be basic blocks with more than $2^{16}$ instructions). I would just use `usize` because it is serialized using variable-length encoding which could be as small as a single byte.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,3 +448,68 @@ fn validate_decorators(operations: &[Operation], decorators: &DecoratorList) {
...
+            let group_len = group.len();
+            for op in group {
+                op.write_into(target);
+                if matches!(op, Operation::Noop) {
+                    if group_len == 1 {
+                        target.write_bool(true);
+                    } else {
+                        target.write_bool(false);
+                    }
+                }
```
bobbinth, line 492: This would catch only `Noop`'s used to create padding groups, but not the `Noop`'s used to pad the end of the groups, right?

paracetamolo: Right, I still need to workout that other case.

bobbinth:
Closing as stale.

PR #1982, opened 2025-07-11
feat: Add extend_ methods on AdviceProvider

### LINKED ISSUE #1980: Add `extend_` methods on `AdviceProvider`
Consider adding `extend_` methods on `AdviceProvider`, analogous to those on `AdviceInputs`:

```rust
/// Extends the stack with the given elements.
pub fn extend_stack<I>(&mut self, iter: I)
where
    I: IntoIterator<Item = Felt>,
{}

/// Extends the map of values with the given argument, replacing previously inserted items.
pub fn extend_map<I>(&mut self, iter: I)
where
    I: IntoIterator<Item = (Word, Vec<Felt>)>,
{}

/// Extends the [MerkleStore] with the given nodes.
pub fn extend_merkle_store<I>(&mut self, iter: I)
where
    I: Iterator<Item = InnerNodeInfo>,
{}
```

Context: https://github.com/0xMiden/miden-base/pull/1564#discussion_r2199170923

varun-doshi:
## Describe your changes
Adds `extend_` methods on AdviceProvider.
Fixes #1980


processor/src/host/advice/mod.rs
```diff
@@ -138,6 +138,40 @@ impl AdviceProvider {
...
+    /// Extends the [MerkleStore] with the given nodes.
+    pub fn extend_merkle_store<I>(&mut self, iter: I)
+    where
+        I: Iterator<Item = InnerNodeInfo>,
+    {
+        self.store.extend(iter);
+    }
+
+    /// Extends the contents of this instance with the contents of the other instance.
+    pub fn extend(&mut self, other: Self) {
```
adr1anh: I would make this take `other: AdviceInputs` instead (reversing the stack to match the `From<AdviceInputs>` constructor, though we should see later if we can construct the stack in `AdviceInputs` in the correct order from the start).

An `AdviceProvider` represents the current state of the Advice, and shouldn't expose the same set of "builder" methods as the `AdviceInputs`. Perhaps a better name for the former would be `AdviceState`.

processor/src/host/advice/mod.rs
```diff
@@ -138,6 +138,40 @@ impl AdviceProvider {
         &self.stack
     }
 
+    // PUBLIC MUTATORS
+    // --------------------------------------------------------------------------------------------
```
bobbinth: nit: rather than putting them into a separate section, should we put each method into their respective sections?

processor/src/host/advice/mod.rs
varun-doshi: Understood
Should I rename `AdviceProvider` to `AdviceState` ?

bobbinth: I would keep the `AdviceProvider` name for now.

Also, I'd probably use some other name rather than `extend()` here (as `extend()` usually implies that `other` is `Self`). Maybe `extend_from_inputs()`?

processor/src/host/advice/mod.rs
```diff
@@ -138,7 +138,7 @@ impl AdviceProvider {
         &self.stack
     }
 
-    // PUBLIC MUTATORS
+    // STACK MUTATORS
     // --------------------------------------------------------------------------------------------
```
bobbinth: I don't think this heading is necessary - we already have an `ADVICE STACK` heading on line 48 above.

processor/src/host/advice/mod.rs
```diff
@@ -149,6 +149,9 @@ impl AdviceProvider {
         self.stack.extend(iter);
     }
 
+    // MAP MUTATORS
+    // --------------------------------------------------------------------------------------------
```
bobbinth: Similar comment here - we already have `ADVICE MAP` on line 182 below.

processor/src/host/advice/mod.rs
```diff
@@ -176,6 +184,14 @@ impl AdviceProvider {
         Ok(())
     }
 
+    /// Extends the map of values with the given argument, replacing previously inserted items.
+    pub fn extend_map<I>(&mut self, iter: I)
+    where
+        I: IntoIterator<Item = (Word, Vec<Felt>)>,
+    {
+        self.map.extend(iter);
+    }
```
bobbinth: Now that I think of it, we probably don't want to overwrite keys in the existing map - so, this should probably work like `merge_advice_map()`. Maybe what we should do is rename `merge_advice_map()` into `extend_map()` and make it public.

processor/src/host/advice/mod.rs
```diff
@@ -285,6 +309,14 @@ impl AdviceProvider {
...
         })
     }
+
+    /// Extends the contents of this instance with the contents of an `AdviceInputs`.
+    pub fn extend_from_inputs(&mut self, mut other: AdviceInputs) {
+        other.stack.reverse();
+        self.stack.extend(other.stack);
+        self.map.extend(other.map);
+        self.store.extend(other.store.inner_nodes());
+    }
```
bobbinth: Does this need to take ownership of `other`? I think we could write it like so:
```rust
pub fn extend_from_inputs(&mut self, inputs: &AdviceInputs) -> Result<(), AdviceError> {
    self.extend_stack(inputs.stack.iter().cloned().rev());
    self.extend_merkle_store(inputs.store.inner_nodes());
    self.extend_map(&inputs.map)
}
```

processor/src/host/advice/mod.rs
```diff
@@ -138,6 +138,14 @@ impl AdviceProvider {
         &self.stack
     }
 
+    /// Extends the stack with the given elements.
```
adr1anh, line 141: Would it be useful to explain that these elements are added to the top of the stack, and the last element of the iterator is the first one popped?

bobbinth: Yes, I think it's worth clarifying. This would be basically equivalent to calling `push_stack()` in a loop.

processor/src/host/advice/mod.rs
```diff
@@ -269,21 +293,22 @@ impl AdviceProvider {
...
 
+    /// Extends the [MerkleStore] with the given nodes.
+    pub fn extend_merkle_store<I>(&mut self, iter: I)
+    where
+        I: Iterator<Item = InnerNodeInfo>,
+    {
+        self.store.extend(iter);
+    }
+
     // HELPERS
```
adr1anh: I don't think we need this separator anymore.

processor/src/host/advice/mod.rs
```diff
@@ -138,6 +138,16 @@ impl AdviceProvider {
         &self.stack
     }
 
+    /// Extends the stack with the given elements.
```
bobbinth, line 141: nit: I would add a newline after this line - this way it'll render better in the docs.

processor/src/host/advice/mod.rs
```diff
@@ -269,21 +294,19 @@ impl AdviceProvider {
         self.store.get_node(root, NodeIndex::root()).is_ok()
     }
 
-    // HELPERS
-    // --------------------------------------------------------------------------------------------
+    /// Extends the [MerkleStore] with the given nodes.
+    pub fn extend_merkle_store<I>(&mut self, iter: I)
+    where
+        I: Iterator<Item = InnerNodeInfo>,
```
adr1anh: Nit: this should be `IntoIterator`. Any `Iterator` implements `IntoIterator` so the latter is more general.

adr1anh (approved):
Thanks for the patch!

One small nit but otherwise it's good to go.

processor/src/host/advice/mod.rs
varun-doshi: Got it
Fixed

bobbinth (approved):
All looks good! Thank you!

PR #1987, opened 2025-07-11
fix: remove unnecessary clone in SubtreeIterator usage

Fallengirl:
## Describe your changes
Removed redundant .clone() call when passing self.statically_linked_mast to SubtreeIterator::new. The iterator accepts a reference (&MastForest) and doesn't require ownership, while self.statically_linked_mast is already an Arc<MastForest>. This improves performance by eliminating an unnecessary cloning operation.


bitwalker:
Closing as spam.

@plafer It is actually a redundant clone, but I'm not down with entertaining these drive-by AI-generated PRs

Fallengirl:
Sorry, I'm trying to be useful for the project

plafer:
@Fallengirl thank you for your interest! However we're receiving a fair amount of spam, so to make sure we don't mistake you for an AI bot, I'd recommend tackling one of the issues marked as `good first issue` 🙂.

Fallengirl:
that would be right, thank you

PR #1988, opened 2025-07-11
chore(host::AdviceProvider): Revert `get_mapped_value` return signature (#1981)

### LINKED ISSUE #1981: Consider returning `Option` from `AdviceProvider::get_mapped_values`
We were previously using `AdviceInputs::mapped_values` which returns an `Option`. With the migration to 0.16 we had to switch to `AdviceProvider::get_mapped_values` which returns a `Result`. That `Result` is basically an `Option` because the function only ever returns `AdviceError::MapKeyNotFound`. This makes matching on the presence of a key less convenient, because we'd have to handle three instead of two cases:
- `Ok` -> key present
- `Err(AdviceError::MapKeyNotFound { .. })` -> key absent
- `Err(_)` -> other error

The latter case would never occur (right now) but we'd have to handle it anyway because another error could be added in the future. I think for that reason, it is preferable to return `Option` from `get`-like methods where the caller commonly wants to mach on presence/absence and not just bubble up the error. Of course that only makes sense if this wasn't done to retain the ability to return an error in the future without breaking the API, so it's clearly a trade-off.

Example for reference (showing the previous case): https://github.com/0xMiden/miden-base/blob/61d057bad853787b0421ec58e1d588aff4b6f038/crates/miden-tx/src/host/note_builder.rs#L55.

This could also be an opportunity to make the APIs of `AdviceInputs` and `AdviceProvider` more consistent, e.g. rename `get_mapped_values` to `mapped_values`.

adr1anh:
## Describe your changes

Fixes #1981

- Updated `CHANGELOG.md'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.2 (2025-07-11)
+
+- [BREAKING] Revert `get_mapped_value` return signature [(#1981)](https://github.com/0xMiden/miden-vm/issues/1981).
+
 ## 0.16.1 (2025-07-10)
 
 - Make `Process::state()` public and re-introduce `From<&Process> for ProcessState`.
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -139,8 +139,8 @@ impl AdviceProvider {
     // --------------------------------------------------------------------------------------------
 
     /// Returns a reference to the value(s) associated with the specified key in the advice map.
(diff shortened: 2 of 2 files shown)


plafer (approved):
LGTM

adr1anh:
> ... do we still need to keep `AdviceError::MapKeyNotFound` variant?

I can't check right now, but it could likely be unused now. For now though, I'd opt to leave it and I'll make a note to remove it when we tackle the second part about lazy advice loading.

bobbinth:
I merged this into `main` but it should have probably gone into `next` as this is a breaking change.

I will cherry-pick this commit into `next` and then revert it on `main`.

PR #1839, opened 2025-05-28
WIP parallel processor

plafer:
## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #1990, opened 2025-07-11
Convert `FastProcessor` to iterative execution

### LINKED ISSUE #1989: Convert `FastProcessor::execute()` to be iterative
As a follow-up to #1933.

For parallel trace generation (#1558), the trace generators need an explicit representation of the call stack at any point of execution - so we might as well make the fast processor natively use one (i.e. it would need to build it anyways).

plafer:
Closes #1989

Note: `ContinuationStack` is meant to replace/be merged with the `ExecutionTraversal` in #1839.

Unfortunately, we get the same performance on the `blake3_1to1` benchmark with this change (where we were hoping for an improvement). I didn't try to optimize too much yet, as I'd focus on fully merging #1839 first.

### DIFF
--- a/processor/src/continuation_stack.rs
+++ b/processor/src/continuation_stack.rs
@@ -1,0 +1,104 @@
+use alloc::{sync::Arc, vec::Vec};
+
+use miden_core::{
+    Program,
+    mast::{MastForest, MastNodeId},
+};
+
+/// A hint for the initial size of the continuation stack.
+const CONTINUATION_STACK_SIZE_HINT: usize = 64;
+
+/// Represents a unit of work in the continuation stack.
+///
+/// This enum defines the different types of continuations that can be performed on MAST nodes
+/// during program execution.
+#[derive(Debug)]
+pub enum Continuation {
+    /// Start processing a node in the MAST forest.
+    StartNode(MastNodeId),
+    /// Process the finish phase of a Join node.
+    FinishJoin(MastNodeId),
+    /// Process the finish phase of a Split node.
+    FinishSplit(MastNodeId),
+    /// Process the finish phase of a Loop node.
+    FinishLoop(MastNodeId),
+    /// Process the finish phase of a Call node.
+    FinishCall(MastNodeId),
+    /// Process the finish phase of a Dyn node.
+    FinishDyn(MastNodeId),
+    /// Enter a new MAST forest, where all subsequent `MastNodeId`s will be relative to this forest.
+    ///
+    /// When we encounter an `ExternalNode`, we enter the corresponding MAST forest directly, and
+    /// push an `EnterForest` continuation to restore the previous forest when done.
+    EnterForest(Arc<MastForest>),
+}
+
+/// [ContinuationStack] reifies the call stack used by the processor when executing a program made
+/// up of possibly multiple MAST forests.
+///
+/// This allows the processor to execute a program iteratively in a loop rather than recursively
+/// traversing the nodes. It also allows the processor to pass the state of execution to another
+/// processor for further processing, which is useful for parallel execution of MAST forests.
+pub struct ContinuationStack {
+    stack: Vec<Continuation>,
+}
+
+impl ContinuationStack {
+    /// Creates a new continuation stack for a program.
+    ///
+    /// # Arguments
+    /// * `program` - The program whose execution will be managed by this continuation stack
+    pub fn new(program: &Program) -> Self {
+        let mut stack = Vec::with_capacity(CONTINUATION_STACK_SIZE_HINT);
+        stack.push(Continuation::StartNode(program.entrypoint()));
+
+        Self { stack }
+    }
+
+    /// Pushes a continuation to enter the given MAST forest on the continuation stack.
+    ///
+    /// # Arguments
+    /// * `forest` - The MAST forest to enter
+    pub fn push_enter_forest(&mut self, forest: Arc<MastForest>) {
+        self.stack.push(Continuation::EnterForest(forest));
+    }
+
+    /// Pushes a join finish continuation onto the stack.
+    pub fn push_finish_join(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::FinishJoin(node_id));
+    }
+
+    /// Pushes a split finish continuation onto the stack.
+    pub fn push_finish_split(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::FinishSplit(node_id));
+    }
+
+    /// Pushes a loop finish continuation onto the stack.
+    pub fn push_finish_loop(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::FinishLoop(node_id));
+    }
+
+    /// Pushes a call finish continuation onto the stack.
+    pub fn push_finish_call(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::FinishCall(node_id));
+    }
+
+    /// Pushes a dyn finish continuation onto the stack.
+    pub fn push_finish_dyn(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::FinishDyn(node_id));
+    }
+
+    /// Pushes a continuation to start processing the given node.
+    ///
+    /// # Arguments
+    /// * `node_id` - The ID of the node to process
+    pub fn push_start_node(&mut self, node_id: MastNodeId) {
+        self.stack.push(Continuation::StartNode(node_id));
+    }
+
+    /// Pops the next continuation from the continuation stack, and returns it along with its
+    /// associated MAST forest.
+    pub fn pop_continuation(&mut self) -> Option<Continuation> {
+        self.stack.pop()
+    }
+}
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -1,13 +1,13 @@
-use alloc::{boxed::Box, sync::Arc, vec::Vec};
+use alloc::{sync::Arc, vec::Vec};
 use core::cmp::min;
 
 use memory::Memory;
 use miden_air::RowIndex;
 use miden_core::{
-    Decorator, DecoratorIterator, EMPTY_WORD, Felt, Kernel, ONE, Operation, Program, StackOutputs,
+    Decorator, DecoratorIterator, EMPTY_WORD, Felt, ONE, Operation, Program, StackOutputs,
     WORD_SIZE, Word, ZERO,
     mast::{
-        BasicBlockNode, CallNode, DynNode, ExternalNode, JoinNode, LoopNode, MastForest, MastNode,
+        BasicBlockNode, CallNode, ExternalNode, JoinNode, LoopNode, MastForest, MastNode,
         MastNodeId, OP_GROUP_SIZE, OpBatch, SplitNode,
     },
     stack::MIN_STACK_DEPTH,
@@ -17,7 +17,10 @@ use miden_debug_types::{DefaultSourceManager, SourceManager};
 
 use crate::{
     AdviceInputs, AdviceProvider, AsyncHost, ContextId, ErrorContext, ExecutionError, FMP_MIN,
-    ProcessState, SYSCALL_FMP_MIN, add_error_ctx_to_external_error, chiplets::Ace, err_ctx,
+    ProcessState, SYSCALL_FMP_MIN,
+    chiplets::Ace,
+    continuation_stack::{Continuation, ContinuationStack},
+    err_ctx,
 };
 
 mod memory;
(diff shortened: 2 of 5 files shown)


processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// pre-decorators, the node itself (in Start phase), and post-decorators.
+    ///
+    /// # Arguments
+    /// * `forest` - The MAST forest containing the nodes to be processed
+    /// * `first_node` - The ID of the first node to process
+    pub fn new(forest: Arc<MastForest>, first_node: MastNodeId) -> Self {
+        Self {
+            forest,
+            // Note: the steps are added in reverse order since they will be popped from the stack.
+            work_stack: vec![
```
bitwalker: I would recommend initializing the capacity of the stack with something like `Vec::with_capacity(64)` or so. This will avoid hitting reallocations + copies early during traversal, and make pushing new continuations much more efficient.

processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+/// [ContinuationStack] reifies the call stack used by the processor when executing a program made
+/// up of possibly multiple MAST forests.
+///
+/// This allows the processor to execute a program iteratively in a loop rather than recursively
+/// traversing the nodes.
+pub struct ContinuationStack {
+    forest_continuation_stack: Vec<ForestContinuationStack>,
```
bitwalker: It isn't clear to me why we manage our continuation stack as a stack of stacks - that fragments the continuations all over the heap rather than keeping them in a contiguous region of memory. 

Instead, this could be simplified by adding a `ProcessingStep::EnterForest(Arc<MastForest>)` variant, which is pushed on to the continuation stack with the current forest when switching to a new forest. When that step is later reached, it restores the previous forest as the current one. Then, all `ProcessingStep`s can be pushed on to the same `Vec`.

I suspect this will improve the performance a bit, but hard to say by how much - at a minimum it simplifies the structure here though.

processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,193 @@
...
+    mast::{MastForest, MastNodeId},
+};
+
+/// Specifies the phase of a processing step.
+///
+/// Specifically, it allows to distinguish between the start and finish phases of processing a node,
+/// which gives the processor sufficient information to execute the node correctly. Note that
+/// exactly how "start" and "end" are used depends on the node type being processed.
+#[derive(Debug, Default)]
+pub enum ProcessingStepPhase {
```
bitwalker: IMO this can just be combined with `ProcessingStep`, and then the prologue/epilogue of each node type handler be split out into separate functions.

This will be much friendlier to the branch predictor, because the dispatch loop will only call a given handler for a specific `ProcessingStep` variant, which in turn will lead to more predictable control flow patterns, i.e. every time the `execute_join_node_start` function is called from the call site in the match arm of `ProcessingStep::Prologue`, the branch predictor can predict with great certainty that the next branch in the dispatch loop will be to the `ProcessingStep::Node` arm and from there to `execute_join_node` - and so on.

processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,193 @@
...
+    Start,
+    Finish,
+}
+
+/// Represents a unit of work in the continuation stack.
+///
+/// This enum defines the different types of processing steps that can be performed on MAST nodes
+/// during program execution. Each variant represents a specific phase of node processing.
+#[derive(Debug)]
+pub enum ProcessingStep {
```
bitwalker: I find it a bit confusing and verbose to call this `ProcessingStep` versus `Continuation`, particularly given that this module is notionally a stack of  continuations, not processing steps. The fact that there a couple different types of continuations doesn't really change that IMO.

bitwalker (commented):
I took a look through with an eye to why performance hasn't improved, and I think I've identified a couple of issues that likely contribute, but even if they make no difference, may still simplify things in a couple ways.

processor/src/fast/mod.rs
```diff
@@ -349,294 +421,315 @@ impl FastProcessor {
...
         &mut self,
-        node_id: MastNodeId,
-        program: &MastForest,
-        kernel: &Kernel,
-        host: &mut impl AsyncHost,
+        join_node: &JoinNode,
+        current_node_id: MastNodeId,
+        phase: ProcessingStepPhase,
+        continuation_stack: &mut ContinuationStack,
     ) -> Result<(), ExecutionError> {
```
bobbinth: For this, and the other similar functions, it seems like we are combining two separate functions in one. Another way to do it to have something like:
```rust
fn start_join_node(
    &mut self,
    join_node: &JoinNode,
    current_node_id: MastNodeId,
    continuation_stack: &mut ContinuationStack,
) -> Result<(), ExecutionError> {
    ...
}

fn finish_join_node(&mut self) {
    ...
}
```
I would also probably move handling of pre and post decorators into these functions. This would eliminate the need to handle them explicitly via the continuations stack.

processor/src/fast/mod.rs
```diff
@@ -320,17 +323,86 @@ impl FastProcessor {
...
+                    for &decorator_id in node.before_enter() {
+                        self.execute_decorator(&current_forest[decorator_id], 0, host)?;
+                    }
+                },
+                ProcessingStep::PostDecorators(node_id) => {
+                    let node = current_forest.get_node_by_id(node_id).unwrap();
+                    for &decorator_id in node.after_exit() {
+                        self.execute_decorator(&current_forest[decorator_id], 0, host)?;
+                    }
+                },
```
bobbinth: Related to the previous comment, I would probably handle the decorators in node processing functions. It will add a bit more boilerplate code, but should also simplify continuation stack processing - i.e., the `ProcessingStep` could look something like this:
```rust
pub enum ProcessingStep{
    Start(MastNodeId),
    Finish(MastNodeId),
}
```
We could also consider making the `Finish` variant more explicit (because we always know what type of a node we are finishing) - e.g.:
```rust
pub enum ProcessingStep{
    Start(MastNodeId),
    FinishJoin, // don't care about node ID
    FinishSplit, // don't care about node ID
    FinishLoop(MastNodeId),
    ...
}
```

processor/src/fast/mod.rs
```diff
@@ -349,294 +421,315 @@ impl FastProcessor {
...
+                // an error.
+                match current_forest.find_procedure_root(callee_hash) {
+                    Some(callee_id) => {
+                        continuation_stack.push_node_with_decorators(callee_id);
+                    },
+                    None => {
+                        let mast_forest =
+                            host.get_mast_forest(&callee_hash).await.ok_or_else(|| {
+                                ExecutionError::dynamic_node_not_found(callee_hash, &err_ctx)
+                            })?;
```
bobbinth: This doesn't seem to handle loading of the advice data from the new MAST forests (i.e., similar to how it is done in `resolve_external_node()` on line 1090 - 1092 below). We should probably encapsulate loading of MAST forests into a small helper procedure which is used both here and there.

Also, do we not need to call  `continuation_stack.push_new_forest()` here similar to how we do it on line 731 below?

bobbinth (commented):
Looks good! Thank you! Not a review yet, but I left a few comments inline.

processor/src/continuation_stack.rs
plafer: Done

processor/src/continuation_stack.rs
plafer: Fixed

processor/src/continuation_stack.rs
plafer: Fixed

processor/src/continuation_stack.rs
plafer: Renamed to `Continuation`

processor/src/fast/mod.rs
plafer: fixed

processor/src/fast/mod.rs

processor/src/fast/mod.rs
plafer: Fixed - although I didn't create the function that encapsulates yet (will give another look tomorrow). The (small) challenge is that we return different errors when loading a mast forest depending on if we're processing an external node or dyn node - so I just have to think how to make this cleanest.

plafer:
@bitwalker @bobbinth addressed the comments, but will need to do a bit of cleanup in the morning - it's in a decent state, but I haven't explicitly updated the docstrings. Feel free to give it another look, or wait for when it's fully cleaned up tomorrow - up to you.

For some reason the CI fails to build, but it works fine locally, so I'll look into that as well.

bobbinth:
> addressed the comments, but will need to do a bit of cleanup in the morning - it's in a decent state, but I haven't explicitly updated the docstrings. Feel free to give it another look, or wait for when it's fully cleaned up tomorrow - up to you.

Thank you! Did these changes affect the performance in any way?

core/src/mast/mod.rs
```diff
@@ -794,3 +794,27 @@ pub enum MastForestError {
...
+/// let basic_block = BasicBlockNode::new(vec![Operation::Push(2_u32.into())], None).unwrap();
+/// let mast_node = MastNode::Block(basic_block.clone());
+///
+/// let downcasted = downcast_node!(mast_node => MastNode::Block);
+/// assert_eq!(downcasted, basic_block);
+/// ```
+///
+/// Adapted from the `downcast!` macro in the `ibc-relayer-types` crate.
+#[macro_export]
+macro_rules! downcast_node {
```
bitwalker: I would suggest using the following pattern instead:

```rust
impl MastNode {
    #[inline]
    #[track_caller]
    pub fn unwrap_basic_block(&self) -> &BasicBlockNode {
        match self {
            Self::Block(node) => node,
            other => unexpected_node_type(other, "basic block"),
        }
    }
}

#[cold]
#[inline(never)]
#[track_caller]
fn unexpected_node_type(node: &MastNode, expected: &str) -> ! {
    let actual = match node {
        MastNode::Block(_) => "basic block",
        MastNode::Join(_) => "join",
        ...
    };
    panic!("tried to unwrap {expected} node, but got {actual}");
}
```

A few reasons for this:

1. The `unwrap` terminology is conventional for this sort of thing. If you look at how `unwrap`/`expect` are implemented on types like `Option`/`Result`, they use this sort of approach. The `downcast` terminology is reserved specifically for casting trait objects to a more concrete type, e.g. `&dyn Any` to `&T`. When I initially saw this, I had to remind myself that `MastNode` wasn't a trait.
2. The actual panic is defined in a single place, but will still point to the actual place where the unwrap was attempted due to the `#[track_caller]` annotations, so a lot less code is generated.
3. The `#[cold]`/`#[inline(never)]` annotations on the panic path ensure that all of that code is outlined away from the hot path, hints to the compiler that the branch is rarely taken, and will keep it in a different page so that it isn't fetched into the cache with hot code paths.

plafer:
> Thank you! Did these changes affect the performance in any way?

Unfortunately not. My current hunch is that `ContinuationStack::pop_stack()` cloning the `Arc<MastForest>` (acquiring a mutex internally) is the main cause of performance degradation. But we are currently forced to do it for borrow-checker reasons. There's hopefully a way to re-design the API to avoid that, but I haven't spent enough time looking into it. But this would be in a subsequent PR.

core/src/mast/mod.rs
plafer: Ah - awesome suggestion! Now fixed.

processor/src/fast/mod.rs
plafer: > This doesn't seem to handle loading of the advice data from the new MAST forests (i.e., similar to how it is done in resolve_external_node() on line 1090 - 1092 below).

This was also missing in `Process::execute_dyn_node()`, so I fixed it there too. I did it as a quick fix though, since we're planning on removing `Process` soon.

plafer: > We should probably encapsulate loading of MAST forests into a small helper procedure which is used both here and there.

Now done too

processor/src/fast/mod.rs
```diff
@@ -320,17 +323,113 @@ impl FastProcessor {
...
     async fn execute_impl(
         &mut self,
         program: &Program,
         host: &mut impl AsyncHost,
     ) -> Result<StackOutputs, ExecutionError> {
-        self.execute_mast_node(program.entrypoint(), program.mast_forest(), program.kernel(), host)
-            .await?;
+        let mut continuation_stack = ContinuationStack::new(program);
+
+        while let Some((current_forest, processing_step)) = continuation_stack.pop_continuation() {
```
adr1anh: Would it make sense to extract this into an `execute_step` function? This was something I think was discussed async yesterday.

processor/src/fast/mod.rs
```diff
@@ -963,31 +1175,29 @@ impl FastProcessor {
         Ok(())
     }
 
-    /// Analogous to [`Process::resolve_external_node`](crate::Process::resolve_external_node), but
-    /// for asynchronous execution.
-    async fn resolve_external_node(
+    // HELPERS
+    // ----------------------------------------------------------------------------------------------
+
+    async fn load_mast_forest<E>(
```
adr1anh, line 1171: Below we can add the `err_ctx` to the `advice_error`

adr1anh (approved):
Minor nits, but otherwise look really good!

processor/src/fast/mod.rs
plafer: Good catch - fixed

processor/src/fast/mod.rs
plafer: Maybe - but I would wait until we have a clear use case for it, since the `ContinuationStack` is likely to change in #1839 and/or subsequent PR that address the issue that we're forced to clone the `Arc<MastForest>` on every continuation. So I wouldn't build more on top of it right now than is strictly necessary.

processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        self.stack.push(continuation);
+    }
+
+    /// Pops the next continuation from the continuation stack, and returns it along with its
+    /// associated MAST forest.
+    pub fn pop_continuation(&mut self) -> Option<(Arc<MastForest>, Continuation)> {
+        let continuation = self.stack.pop()?;
+        Some((self.current_forest.clone(), continuation))
+    }
```
bobbinth: Probably not for this PR, but I wonder if there is a way to avoid cloning `Arc` for MAST forest on every continuation pop.

processor/src/fast/mod.rs
```diff
@@ -320,17 +323,113 @@ impl FastProcessor {
...
+                },
+                Continuation::StartJoin(node_id) => {
+                    let join_node = current_forest[node_id].unwrap_join();
+                    self.start_join_node(
+                        join_node,
+                        node_id,
+                        &current_forest,
+                        &mut continuation_stack,
+                        host,
+                    )?
```
bobbinth: nit: we could look up the `join_node` inside the `start_join_node()` function. Then, this would look something like:
```rust
self.start_join_node(node_id, &current_forest, &mut continuation_stack, host)?
```
(i.e., passing both `join_node` and `node_id` into the function is a bit redundant)

This also applies to other node types.

processor/src/continuation_stack.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// A hint for the initial size of the continuation stack.
+const CONTINUATION_STACK_SIZE_HINT: usize = 64;
+
+/// Represents a unit of work in the continuation stack.
+///
+/// This enum defines the different types of continuations that can be performed on MAST nodes
+/// during program execution.
+#[derive(Debug)]
+pub enum Continuation {
```
bobbinth, line 16: Not for this PR: the current structure of this requires us to do a lookup into the MAST forest (via the `get_node_by_id()` function) twice for every node: first time when we push it into the continuation stack, and second time when we pop it.

Not sure if it will make practical difference, but we should be able to avoid this by removing node-specific `Start*` variants. It could look something like:
```rust
pub enum Continuation {
    StartNode(MastNodeId),
    FinishJoin(MastNodeId),
    FinishSplit(MastNodeId),
    FinishLoop(MastNodeId),
    FinishCall(MastNodeId),
    FinishDyn(MastNodeId),
    EnterForest(Arc<MastForest>),
}
```

processor/src/fast/mod.rs
bobbinth, line 1: Not for this PR, but this file has gotten pretty big. We should split it up into 2 - 3 files in subsequent PRs.

bobbinth (approved):
Looks good! Thank you. I left a few comments inline, but most of them are for subsequent PRs.

processor/src/continuation_stack.rs
plafer: Yup I think this is the main reason for performance degradation

bitwalker: I think the current forest can just be a local variable in the body of the main execution loop, AFAICT, it isn't actually needed as part of the continuation stack.

plafer: Went for it in this PR, and it made the `blake3_1to1` benchmark go from ~1.82-1.85ms down to ~1.79ms (still up from ~1.40ms before the change to async).

processor/src/fast/mod.rs

processor/src/continuation_stack.rs
plafer: Ah yes good suggestion! Went for it in this PR, since it reduces the number of variants quite a bit.

processor/src/fast/mod.rs
plafer: Ended up reverting this change due to your [other suggestion](https://github.com/0xMiden/miden-vm/pull/1990#discussion_r2211939045) for most nodes (all except `Dyn` and `External` which there were borrow-checker issues), since it made reduced the number of `MastForest` accesses.

Issue #1928, opened 2025-06-30
Add `word` module to stdlib


PhilippGackstatter:
@bobbinth mentioned in https://github.com/0xMiden/miden-base/pull/1428#discussion_r2174299643:

> It feels like we are missing an `std::word` module in the standard library. It could include things like:
>
> - `word::gt`, `word:lt` implemented here.
> - `word::is_empty` - which is now in `std::utils::is_empty_word`.
> - `word::eq` and maybe even `word::assert_eq` which are now dedicated VM instructions.
> - Probably some other things as well.

I guess one question is whether we want to define a standard ordering for words or not. I think that would be useful for use cases like the link map (https://github.com/0xMiden/miden-base/pull/1428).
Ideally we would then also update the [`Ord` impl for `Word`](https://github.com/0xMiden/crypto/blob/80389c4b8585af3454b4679c78e4fbee6bb7b376/miden-crypto/src/word.rs#L140) to match that implementation. That would avoid the need for custom wrappers like `LinkMapKey` introduced in https://github.com/0xMiden/miden-base/pull/1478.

Qyriad:
I can take this — it'll be handy for my work on #1218.

PhilippGackstatter:
For reference and in case it is useful, we have `gt`/`lt` implementations for word in [`link_map.masm`](https://github.com/0xMiden/miden-base/blob/e87f1f601696140d4035f720801a7874d4d047f7/crates/miden-lib/asm/kernels/transaction/lib/link_map.masm#L955), which I think could just be copied as-is and renamed.

PR #1996, opened 2025-07-15
WIP: feat(stdlib): init word module

Qyriad:
## Describe your changes

A draft PR to track my work on #1928.

At the time of this writing the functions are implemented, and `word::eq` has a test. The other exported functions still need tests, and all of them need docs written.

- Updated `CHANGELOG.md'


stdlib/asm/word.masm
```diff
@@ -0,0 +1,190 @@
...
+#! Returns a boolean indicating whether the input word is an empty word.
+#!
+#! Inputs: [INPUT_WORD]
+#! Outputs: [is_empty_word, INPUT_WORD]
+#!
+#! - INPUT_WORD is the word whose emptiness is to be determined.
+#! - is_empty_word is a boolean indicating whether INPUT_WORD is empty.
+#!
+#! Cycles: 11
+export.is_empty
```
bobbinth: Would be awesome to add cycle counts to other procedures here as well.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,190 @@
...
+#!         result |= gt & cont;
+#!         cont &= eq;
+#!     }
+#!
+#!     result
+#! }
+#!
+#! Inputs:  [LHS, RHS]
+#! Outputs: [is_lhs_greater]
+export.gt
```
bobbinth, line 75: For completeness, I would also add `gte` and `lte` procedures (which should be easy to add given the ones already existing here).

stdlib/asm/word.masm
```diff
@@ -0,0 +1,222 @@
...
+    movup.3
+end
+
+#! Returns a boolean indicating whether the input word is an empty word.
+#!
+#! Inputs: [INPUT_WORD]
+#! Outputs: [is_empty_word, INPUT_WORD]
+#!
+#! - INPUT_WORD is the word whose emptiness is to be determined.
+#! - is_empty_word is a boolean indicating whether INPUT_WORD is empty.
```
PhilippGackstatter: Nit: Not sure if this should apply here, but we have a style guide in [miden-base](https://github.com/0xMiden/miden-base/blob/next/crates/miden-lib/masm_doc_comment_fmt.md), where we use a `Where:` before explaining the variable names.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,222 @@
...
+# COMPARISON OPERATIONS
+# -------------------------------------------------------------------------------------------------
+
+#! Arranges the given words such that the corresponding elements are next to each other.
+#!
+#! Inputs:  [WORD1, WORD2]
+#! Outputs: [word2_3, word1_3, word2_2, word1_2, word2_1, word1_1, word2_0, word1_0]
+#!
+#! Cycles: 8
+proc.arrange_words_adjacent
```
PhilippGackstatter: Nit: Given that this is an internal procedure, I'd move it to the end of the file below a `HELPER PROCEDURES` separator.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,222 @@
...
+end
+
+#! Returns true if LHS is exactly equal to RHS, false otherwise.
+#! The implementation does not branch, and always performs the same number of comparisons.
+#!
+#! Inputs:  [LHS, RHS]
+#! Outputs: [lhs_eq_rhs]
+#!
+#! Cycles: 26
+export.eq
```
PhilippGackstatter, line 203: When would this be used instead of the `eqw` instruction?

stdlib/tests/word/mod.rs
```diff
@@ -0,0 +1,34 @@
+use core::iter;
+
+use miden_core::{Felt, Word};
+use miden_utils_testing::{ONE, StackInputs, ZERO, felt_slice_to_ints, rand::seeded_word};
+
+#[test]
```
PhilippGackstatter, line 50: I think we could move the `gt`/`lt` tests from miden-base here: https://github.com/0xMiden/miden-base/blob/8e88dc6b04429f67c685ce577f2afce6f8df757a/crates/miden-testing/src/kernel_tests/tx/test_link_map.rs#L370.

PhilippGackstatter (commented):
Looks good to me!

Left some initial comments.

stdlib/asm/word.masm
Qyriad: Possibly it may not be. But I got the impression that it *may* be desirable to remove the `eqw` instruction:

> word::eq and maybe even word::assert_eq which are now dedicated VM instructions.[^1]

I can update the docs to say that it is *currently* equivalent to the `eqw` instruction.

[^1]: https://github.com/0xMiden/miden-vm/issues/1928#issue-3187880636

stdlib/asm/word.masm
Qyriad: Now that I think about it — should this procedure be exported instead?

stdlib/asm/word.masm
bitwalker: AFAIK there are no plans to remove `eqw` and `assert_eqw`, nor desire to remove them - has this been discussed somewhere? I didn't see any mention of that in the linked issue.

bobbinth: My original thinking was that it may be awkward to have some word-related functionality in stdlib and other in the instruction set (e.g., `std::word::lt` and `std::word::gt`, but `eqw`). But now that I think about it, we can't really make `assert_eqw` a procedure because we can specify error messages for procedures.

Regarding `eqw` - I don't have a strong preference: we could have it in both places, only as an instruction, or only as a library procedure.

bitwalker: I would expect that as a library procedure only, it would be much more inefficient than as an instruction (currently 1 cycle). I don't know how much of a problem that is in practice, but to me that's the main reason to make something an instruction rather than a procedure (the other being assertions as you noted).

bobbinth: The instruction we currently have for this is just a "meta-instruction" which expands into basically the same sequence of actual VM operations. So, I don't think there is currently a cycle count difference here. Keeping it an instruction does leave the door open for potential future optimizations (e.g., if we decide to convert it to an actual single-cycle operation).

bitwalker: Oh actually you're right, I misread the cycle count when going through the docs and read the count for `assert_eqw`, which is 11, as 1. Sounds like it's less important than I thought then.

stdlib/tests/word/mod.rs
Qyriad: Excellent idea.

stdlib/asm/word.masm
PhilippGackstatter: Not a strong opinion, but I'd leave it private until we have a few cases where we think we would benefit from exposing this. I guess I haven't worked with "raw words" enough to say whether it'd be useful or not, so I default to leaving it private because making it public is easy while making it private is a breaking change.

Qyriad:
This is now ready for review!

crates/utils/testing/src/lib.rs
```diff
@@ -534,6 +534,14 @@ impl Test {
 // HELPER FUNCTIONS
 // ================================================================================================
 
+/// Appends a Word to an operand stack Vec.
+pub fn append_word_to_vec(target: &mut Vec<u64>, word: Word) {
+    target.push(word[0].as_int());
```
bitwalker: You could probably just make this `target.extend(word.into_iter().map(|felt| felt.as_int()))`.

Not sure it really matters, but it's a bit less noisy perhaps.

docs/src/user_docs/stdlib/word.md
```diff
@@ -0,0 +1,13 @@
+# Word procedures
+
+Module `std::word` contains utilities for manipulating *words* &mdash; sequences of four *fields*.
```
bitwalker:
```suggestion
Module `std::word` contains utilities for manipulating *words* &mdash; sequences of four *field elements*.
```

docs/src/user_docs/stdlib/word.md
```diff
@@ -0,0 +1,13 @@
...
+See [Terms and notations](./main.md#Terms-and-notations) for more information.
+
+| Procedure      | Description   |
+| -------------- | ------------- |
+| reverse        | Reverses order of the first four elements on the stack.<br /><br />Inputs: `[a, b, c, d, ...]`<br />Output: `[d, c, b, a, ...]`<br /><br />Cycles: 3|
+| is\_empty      | Returns a boolean indicating whether the input word is empty (all zeros).<br /><br />Inputs: `[INPUT_WORD]`<br />Outputs: `[is_empty_word, INPUT_WORD]`<br /><br />Where:<br />- `INPUT_WORD` is the word whose emptiness is to be determined.<br />- `is_empty_word` is a boolean indicating whether `INPUT_WORD` is empty.<br /><br />Cycles: 11|
+| gt             | Returns true if `LHS` is strictly greater than `RHS`, false otherwise.<br /><br />It compares words using the same ordering as Merkle tree key comparison. This implementation avoids branching for performance reasons.<br /><br />Inputs: `[LHS, RHS]`<br />Output: `[is_lhs_greater]`<br /><br />Cycles: 121 |
```
bitwalker: Could we link to the Merkle tree key comparison ordering semantics here? I would expect the ordering to be lexicographical, i.e. ordered by value starting from `word[0]`. If that's not the case, it will be important for the reader to understand the distinction (and perhaps the rationale for why it differs from the lexicographical approach, if it does).

stdlib/asm/word.masm
```diff
@@ -0,0 +1,229 @@
...
+#!
+#! Inputs:  [a, b, c, d, ...]
+#! Outputs: [d, c, b, a, ...]
+#!
+#! Cycles: 3
+export.reverse
+    swap
+    movup.2
+    movup.3
+end
```
bitwalker:
```suggestion
export.reverse
    movdn.3
    swap.2
end
```

One less instruction this way - not sure if there is any actual cycle count difference though.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,229 @@
...
+
+#! Returns a boolean indicating whether the input word is an empty word (all zeros).
+#!
+#! Inputs:  [INPUT_WORD]
+#! Outputs: [is_empty_word, INPUT_WORD]
```
bitwalker, line 34: It's a bit unusual to leave inputs on the operand stack (usually the convention is to always consume the input value, leaving it up to the caller to duplicate the input if needed).

AFAIK we only do so in rare cases (typically pairs of words IIRC), but I'm not sure this meets the bar given the input is only a single word.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,229 @@
...
+#!
+#! Where:
+#! - INPUT_WORD is the word whose emptiness is to be determined.
+#! - is_empty_word is a boolean indicating whether INPUT_WORD is empty.
+#!
+#! Cycles: 11
+export.is_empty
```
bitwalker: I think I prefer calling this `eqz` or `is_zeroed`, since `is_empty` implies the value is a variable-length collection. I think we generally treat words as scalar values, and only as aggregates when we're encoding data into them (i.e. viewing them as structs encoded using 4 field elements).

Not a hill I feel the need to die on, but I think it is probably best to err on the side of strict when it comes to terminology to avoid confusion.

stdlib/asm/word.masm
plafer: Agreed, I wouldn't leave `INPUT_WORD` on the stack - otherwise you are forced to e.g. `movdn.4 dropw` after each call to `is_empty` which you don't need `INPUT_WORD`

stdlib/tests/word/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        let expected = u64::from(valid_ords.contains(&expected_cmp));
+
+        build_test!(source, &operand_stack).expect_stack(&[expected]);
+    }
+}
+
+#[test]
+fn test_gt() {
+    execute_comparison_test("gt", &[Ordering::Greater]);
+}
```
plafer: I think [`rstest` cases](https://docs.rs/rstest/latest/rstest/#creating-parametrized-tests) would be a great fit here. We also already use it in the `miden-processor` crate

stdlib/asm/word.masm
bobbinth: Sometimes it may be useful to keep the input word on the stack (e.g., check if the word is empty, and if it isn't do something with it), but I agree that this is not the usual approach. I wonder if we can have two variants of this procedure:

- `eqz` which would consume the input.
- `eqz_non_consuming` which would leave the input on the stack (though, I'm hoping we can find some better way to say `non_consuming`).

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
 - Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
+- Add new stdlib module `std::word`, containing utilities for manipulating arrays of four fields (words) [#1928](https://github.com/0xMiden/miden-vm/issues/1928).
+- BREAKING: move `std::utils::is_empty_word` to `std::word::is_empty`, as part of the new word module.
```
bobbinth: nit: we usually write breaking as `[BREAKING]` (i.e., in square brackets).

bobbinth (commented):
Looks good! Thank you! Not a full review from me, but I left a couple of comments.

stdlib/asm/word.masm
Qyriad: This procedure was moved as-is from `std::utils` — it kept the input word on the stack, so I left it that way.

Should I change it?

Qyriad: > I'm hoping we can find some better way to say `non_consuming`

`preserving` is at least shorter, but if you think procedures that preserve the inputs might be more common going forward maybe a conventional suffix (e.g. `eqz_nc`) would be appropriate.

stdlib/asm/word.masm
Qyriad: I can confirm there's no difference in cycle count. It does look nicer, though.

docs/src/user_docs/stdlib/word.md
Qyriad: I don't believe that documentation quite exists. The ordering *is* [lexicographical](https://docs.rs/miden-crypto/latest/miden_crypto/word/struct.LexicographicWord.html), but `word[0]` in Miden assembly is `word[3]` in Rust code (… I think. It's very easy to get mixed up on this.). And neither [`LexicographicWord`](https://docs.rs/miden-crypto/latest/miden_crypto/word/struct.LexicographicWord.html) nor [`Word`](https://docs.rs/miden-crypto/latest/miden_crypto/word/struct.Word.html) specify a defined ordering.

bitwalker: I think we should specify then that the ordering is lexicographical, starting with the first element of the word in address order (i.e. the lowest-addressable element is compared first, and so on). For words on the operand stack, it is expected that the lowest-addressable element is furthest from the top of the operand stack, as if the value had just been loaded with `mem_loadw`.

I think that would make the ordering semantics clear regardless of whether the word is in memory or on the operand stack, and clearly lays out what is expected of the caller when invoking the comparison procedures.

If what I've described above differs from the actual implementation, I think it would be good to know what those differences are, and whether we can modify things to work as I've described.

stdlib/asm/word.masm
bitwalker: I think we could make this breaking, since all uses of `std::utils::is_empty` would have to be updated anyway, so this would be an ideal opportunity to update those uses as appropriate IMO.

docs/src/user_docs/stdlib/word.md
bobbinth: > I think we should specify then that the ordering is lexicographical, starting with the first element of the word in address order (i.e. the lowest-addressable element is compared first, and so on).

I think it is currently the reverse of this. In Rust, we compare elements at index 3 first, if they are equal, we compare elements at index 2 etc. (@PhilippGackstatter can correct me if I'm wrong). This is conceptually similar to interpreting the word as a little-endian 256-bit value, and then comparing the values of two words as integers.

stdlib/asm/word.masm
bobbinth: > `preserving` is at least shorter, but if you think procedures that preserve the inputs might be more common going forward maybe a conventional suffix (e.g. `eqz_nc`) would be appropriate.

Yes, `preserving` is better - but still a bit too long IMO. So, we could go with `eqz_nc` (or maybe `eqz_pres` - but `nc` is probably better).

docs/src/user_docs/stdlib/word.md
PhilippGackstatter: Yes, the [Rust impl](https://docs.rs/miden-crypto/latest/src/miden_crypto/word/lexicographic.rs.html#67-86) compares the felts of the words at index 3, 2, 1, 0 - in that order. For the use in the link map in miden-base it's not relevant that it is this order and not the reverse, so if we find this is counterintuitive, we can change it.

I think I did it that way because in the SMT we define [index 3](https://docs.rs/miden-crypto/latest/src/miden_crypto/merkle/smt/full/mod.rs.html#465) of the word as the most significant felt, we use this fact a lot in miden-base, and so it made sense (to me) to compare from most significant to least significant - according to that definition. I agree, it would be great to standardize/define this somewhere. The above makes sense to me, so I have a light preference towards that, but am happy to adapt as well.

stdlib/asm/word.masm
PhilippGackstatter: Nit: I find `nc` a bit hard to decipher.

I would suggest `eqz_keep`. It's short, not an abbreviation and also says "input preserving".

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+
+#! Returns true if LHS is strictly greater than RHS, false otherwise.
+#!
+#! This compares words using the same ordering as Merkle tree key comparisons.
+#!
+#! The implementation avoids branching for performance reasons.
+#!
+#! For reference, this is equivalent to the following Rust function:
+#!
+#! fn is_key_greater(key1: Word, key2: Word) -> bool {
```
PhilippGackstatter:
```suggestion
#! fn is_word_greater(word1: Word, word2: Word) -> bool {
```

Nit: I'd update this to say "word" instead of "key".

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+#!         let eq = key1[i].as_int() == key2[i].as_int();
+#!         result |= gt & cont;
+#!         cont &= eq;
+#!     }
+#!
+#!     result
+#! }
+#!
+#! Inputs:  [LHS, RHS]
+#! Output: [is_lhs_greater]
```
PhilippGackstatter:
```suggestion
#! Inputs:  [LHS, RHS]
#! Outputs: [is_lhs_greater]
```

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+    # => [is_lhs_greater, continue]
+
+    swap drop
+    # => [is_lhs_greater]
+end
+
+#! Returns true if LHS is greater than or equal to RHS.
+#!
+#! Inputs:  [LHS, RHS]
+#! Output: [is_lhs_greater_or_equal]
```
PhilippGackstatter:
```suggestion
#! Inputs:  [LHS, RHS]
#! Outputs: [is_lhs_greater_or_equal]
```

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+
+#! Returns true if LHS is strictly less than RHS, false otherwise.
+#!
+#! The implementation avoids branching for performance reasons.
+#!
+#! From an implementation standpoint this is exactly the same as `word::gt` except it uses
+#! `lt` rather than `gt`. See its docs for details.
+#!
+#! Inputs:  [LHS, RHS]
+#! Output: [is_lhs_lesser]
```
PhilippGackstatter:
```suggestion
#! Inputs:  [LHS, RHS]
#! Outputs: [is_lhs_lesser]
```

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+    # => [is_lhs_less, continue]
+
+    swap drop
+    # => [is_lhs_less]
+end
+
+#! Returns true if LHS is less than or equal to RHS.
+#!
+#! Inputs:  [LHS, RHS]
+#! Output: [is_lhs_less_or_equal]
```
PhilippGackstatter:
```suggestion
#! Inputs:  [LHS, RHS]
#! Outputs: [is_lhs_less_or_equal]
```

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+end
+
+#! Returns true if LHS is exactly equal to RHS, false otherwise.
+#!
+#! The implementation does not branch, and always performs the same number of comparisons.
+#!
+#! This is currently equivalent to the eqw instruction.
+#!
+#! Inputs:  [LHS, RHS]
+#! Output: [lhs_eq_rhs]
```
PhilippGackstatter:
```suggestion
#! Inputs:  [LHS, RHS]
#! Outputs: [lhs_eq_rhs]
```

PhilippGackstatter (approved):
Looks good to me!

docs/src/user_docs/stdlib/word.md
bitwalker: I think it would be surprising for word comparisons to work in reverse order to how most sorting algorithms would work over an aggregate/collection. When viewing a word in memory as just storage for a set of bytes, small integers, or felts, you'd expect them to be sorted in index order. It seems like we're treating the word as an integral value here, specifically a little-endian one, but that isn't really the default "type" of a word IMO, which is why performing comparisons that way is unintuitive.

That said, maybe that's OK for `lt`/`gt`, but I do think it has the potential to be a gotcha unless it is _very_ clear that the ordering is not what you'd expect based on intuition alone.

bobbinth: I don't remember if there was a specific reason for doing it this way (except for maybe it is slightly more efficient to do this in MASM) - but if the reason is not too strong, I think it shouldn't be too difficult to change (we'd just need to change a few lines of code in `LexicographicWord`). Again, cc @PhilippGackstatter in case I'm saying something wrong.

stdlib/asm/word.masm
bitwalker: Why have a suffix at all? I'm not sure I understand the need - the _default_ behavior for most procedures is to consume their inputs, but nowhere do we use a suffix like this to indicate stack behavior that I'm aware of, regardless of whether it is consumed or preserved.

bitwalker: Ah, I missed the part about having both procedures - I don't understand that either, just make the caller `dupw` before calling `eqz`?

bobbinth: Ah - I think I confused myself. I thought we'd be able to save some cycles by leaving the elements on the stack, but actually it doesn't help much. So, no need for another variant `dupw exec.word::eqz` works just as well.

bobbinth: Actually, thinking about it more, there is a cycle difference. Here is a non-consuming versions:
```
# 11 cyles
export.eqz_nc
    repeat.4
        dup.3 eq.0
    end
    and and and
end
```

Here is the consuming version:
```
# 10 cycles
export.eqz
    eqz
    repeat.3 
        swap eqz and
    end
end
```
So, `dupw exec.word::eqz` would take 3 cycles more than `exec.word::eqz`.

Whether 3 cycles are worth having another variant, I'm not sure - but I guess it doesn't hurt either.

bitwalker: Couldn't you do:

```
export.eqz           # [a, b, c, d]
    eq                     # [a_eq_b, c, d] (1 cycle)
    swap eq.0 and # [abc_is_zero, d] (3 cycles)
    swap eq.0 and # [is_zero] (3 cycles)
end
```

So `dupw exec.eqz` would be 10 cycles total, right?

bobbinth: Hmmm - `a_eq_b` doesn't imply that they are both `0`, right? So, something like `[2, 2, 0, 0]` would return true from the above (unless I'm missing something).

bitwalker: Makes sense, if we think there's a use case for it, then I think I'd name the consuming version `eqz` and the non-consuming version `testz` (which aligns with the `u32test`/`u32testw` instructions which also are non-consuming). 

I think my gut feeling is that procedure names which try to describe stack semantics is not great. Ideally we can find terminology that works for any instruction/procedure that leaves its input on the operand stack, so we can be consistent about it in the future when this sort of thing comes up; but that may not be feasible, naming is annoying ☹️. The "test" terminology is the best we've got, and it is somewhat associated with testing for valid values with the existing uses, but I don't think that is too strong of an association (hence `testz`, i.e. "test for zero")

bobbinth: Yep - let's go with `testz`.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+export.lte
+    exec.gt
+    not
+end
+
+#! Returns true if LHS is exactly equal to RHS, false otherwise.
+#!
+#! The implementation does not branch, and always performs the same number of comparisons.
+#!
+#! This is currently equivalent to the eqw instruction.
```
bobbinth, line 197: Actually, `eqw` instruction works differently - it keeps both words on the stack. Also, the implementation is:
```
dup.7, dup.4, eq,
dup.7, dup.4 eq, and,
dup.6, dup.3, eq, and,
dup.5, dup.2, eq, and,
```
which is 15 cycles.

I think for the `eq` procedure we should use the following implementation:
```
movup.4 eq swap
movup.4 eq and swap
movup.3 eq and
movdn.2 eq and
```
Which should be 13 cycles.

And I would also add `test_eq` procedure which don't consume the inputs and used the implementation equivalent to `eqw`.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
 - Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
+- Add new stdlib module `std::word`, containing utilities for manipulating arrays of four fields (words) [#1928](https://github.com/0xMiden/miden-vm/issues/1928).
+- [BREAKING]: move `std::utils::is_empty_word` to `std::word::eqz`, as part of the new word module.
```
Fumuran: nit: let's add a link to the issue here too (it is fine if they would be identical). That way it would be easier to locate the issue where this change was made.
```suggestion
- [BREAKING]: move `std::utils::is_empty_word` to `std::word::eqz`, as part of the new word module [#1928](https://github.com/0xMiden/miden-vm/issues/1928).
```

docs/src/user_docs/stdlib/word.md
```diff
@@ -0,0 +1,13 @@
+# Word procedures
+
+Module `std::word` contains utilities for manipulating *words* &mdash; sequences of four *field elements*.
+See [Terms and notations](./main.md#Terms-and-notations) for more information.
+
+| Procedure      | Description   |
+| -------------- | ------------- |
+| reverse        | Reverses order of the first four elements on the stack.<br /><br />Inputs: `[a, b, c, d, ...]`<br />Output: `[d, c, b, a, ...]`<br /><br />Cycles: 3|
+| eqz            | Returns a boolean indicating whether the input word `[0, 0, 0, 0]`.<br /><br />Inputs: `[INPUT_WORD]`<br />Outputs: `[is_word_zero, INPUT_WORD]`<br /><br />Where:<br />- `INPUT_WORD` is the word to compare against `[0, 0, 0, 0]`.<br />- `is_word_zero` is a boolean indicating whether `INPUT_WORD` is all zeros.<br /><br />Cycles: 11|
+| gt             | Returns true if `LHS` is strictly greater than `RHS`, false otherwise.<br /><br />It compares words using the same ordering as Merkle tree key comparison. This implementation avoids branching for performance reasons.<br /><br />Inputs: `[LHS, RHS]`<br />Output: `[is_lhs_greater]`<br /><br />Cycles: 121 |
```
Fumuran: nit: 
```suggestion
| gt             | Returns true if `LHS` is strictly greater than `RHS`, false otherwise.<br /><br />It compares words using the same ordering as Merkle tree key comparison. This implementation avoids branching for performance reasons.<br /><br />Inputs: `[LHS, RHS]`<br />Outputs: `[is_lhs_greater]`<br /><br />Cycles: 121 |
```

docs/src/user_docs/stdlib/word.md
```diff
@@ -0,0 +1,13 @@
...
+| gte            | Returns true if `LHS` is greater than or equal to `RHS`.<br /><br />Inputs: `[LHS, RHS]`<br />Outputs: `[is_lhs_greater_or_equal]`<br /><br />Cycles: 118 |
+| lt             | Returns true if `LHS` is strictly less than `RHS`, false otherwise.<br /><br />The implementation avoids branching for performance reasons.<br /><br />From an implementation standpoint this is exactly the same as `word::gt` except it uses `lt` rather than `gt`. See its docs for details.<br /><br />Inputs: `[LHS, RHS]`<br />Output: `[is_lhs_lesser]`<br /><br />Cycles: 117 |
```
Fumuran: nit:
```suggestion
| lt             | Returns true if `LHS` is strictly less than `RHS`, false otherwise.<br /><br />The implementation avoids branching for performance reasons.<br /><br />From an implementation standpoint this is exactly the same as `word::gt` except it uses `lt` rather than `gt`. See its docs for details.<br /><br />Inputs: `[LHS, RHS]`<br />Outputs: `[is_lhs_lesser]`<br /><br />Cycles: 117 |
```

and the same comment in the line below

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
+#! Reverses order of the first four elements on the stack
+#!
+#! Inputs:  [a, b, c, d, ...]
+#! Outputs: [d, c, b, a, ...]
+#!
+#! Cycles: 3
+export.reverse
+	movdn.3
+	swap.2
```
Fumuran: Probably the 8 spaces tab was used here? I would make it 4 spaces to be consistent with other procedures.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+end
+
+#! Returns a boolean indicating whether the input word is [0, 0, 0, 0].
+#!
+#! Inputs:  [INPUT_WORD]
+#! Outputs: [is_word_zero, INPUT_WORD]
+#!
+#! Where:
+#! - INPUT_WORD is the word to compare against [0, 0, 0, 0].
+#! - is_empty_word is a boolean indicating whether INPUT_WORD is all zeros.
```
Fumuran: I would use the `is_empty_word` instead of the `is_word_zero`, but it's not a strong preference. But the main idea is to make the name in `Outputs` and in the description the same.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,228 @@
...
+#! Outputs: [is_lhs_greater_or_equal]
+#!
+#! Cycles: 118
+export.gte
+    exec.lt
+    not
+end
+
+#! Returns true if LHS is strictly less than RHS, false otherwise.
+#!
```
Fumuran, line 130: I'm not sure about the comparison order.

In our other assembly instructions we often use the reverse comparison order (which sometimes could be confusing, I often double check it). For example, the `lt` [instruction](https://0xmiden.github.io/miden-vm/user_docs/assembly/field_operations.html#comparison-operations) works like so:
```masm
[b, a] -> [c], where c = (a < b)
```
In other words we compare the value deeper on the stack with the one on the top of the stack. This is also done in order to make usage of the instructions with immediate values more convenient: `lt.b` will assert that the value on the stack top is less that the one we provided. 

So to make the procedures consistent I would change the comparison order. It should be an easy change, essentially being just a renaming. @PhilippGackstatter please correct me if I'm missing something.

Fumuran (changes requested):
Looks great, thank you!

I left a few comments inline, but most of them are just minor formatting nits.

stdlib/asm/word.masm
bitwalker: I agree that it would be best to be consistent with operand ordering. I think it isn't quite as simple as a trivial rename in the docs though, rather I think you'd need to swap the implementations of `lt`/`gt`/etc. and swap the operand order in the docs, since the order of the operands affects the output.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,241 @@
...
+#! - INPUT_WORD is the word to compare against [0, 0, 0, 0].
+#! - is_empty_word is a boolean indicating whether INPUT_WORD is all zeros.
+#!
+#! Cycles: 11
+export.eqz
+    repeat.4
+        movup.3 eq.0
+    end
+    and and and
+end
```
bobbinth: This could be done with 10 cycles, right? Something like this:
```
export.eqz
    eq.0
    repeat.3 
        swap eq.0 and
    end
end
```

stdlib/asm/word.masm
```diff
@@ -0,0 +1,241 @@
...
+#!
+#! The implementation does not branch, and always performs the same number of comparisons.
+#!
+#! This is currently equivalent to the eqw instruction.
+#!
+#! Inputs:  [RHS, LHS]
+#! Outputs: [lhs_eq_rhs]
+#!
+#! Cycles: 13
+export.eq
```
bobbinth, line 203: Let's also add `test_eq` procedure which would do the same but would not consume the inputs. The implementation could look something like:
```
dup.7 dup.4 eq
dup.7 dup.4 eq and
dup.6 dup.3 eq and
dup.5 dup.2 eq and
```
This should be 15 cycles.

stdlib/asm/word.masm
Qyriad: You're correct! Good idea.

docs/src/user_docs/stdlib/word.md
```diff
@@ -0,0 +1,16 @@
...
+
+Module `std::word` contains utilities for manipulating *words* &mdash; sequences of four *field elements*.
+See [Terms and notations](./main.md#Terms-and-notations) for more information.
+
+| Procedure      | Description   |
+| -------------- | ------------- |
+| reverse        | Reverses order of the first four elements on the stack.<br /><br />Inputs: `[a, b, c, d, ...]`<br />Outputs: `[d, c, b, a, ...]`<br /><br />Cycles: 3|
+| eqz            | Returns a boolean indicating whether the input word `[0, 0, 0, 0]`.<br /><br />Inputs: `[INPUT_WORD]`<br />Outputs: `[is_word_empty]`<br /><br />Where:<br />- `INPUT_WORD` is the word to compare against `[0, 0, 0, 0]`.<br />- `is_word_empty` is a boolean indicating whether `INPUT_WORD` is all zeros.<br /><br />Cycles: 11|
+| testz          | Returns a boolean indicating whether the input word `[0, 0, 0, 0]` Unlike `eqz`, this does not consume the inputs.<br /><br />Inputs: `[INPUT_WORD]`<br />Outputs: `[is_word_empty, INPUT_WORD]`<br /><br />Where:<br />- `INPUT_WORD` is the word to compare against `[0, 0, 0, 0]`.<br />- `is_word_empty` is a boolean indicating whether `INPUT_WORD` is all zeros.<br /><br />Cycles: 10|
```
bobbinth: I think the cycle counts are are reversed: it should be 10 cycles for `eqz` and 11 cycles for testz`.

docs/src/user_docs/stdlib/word.md
PhilippGackstatter: No, I think this should work just fine. I just needed _some_ order, not this specific one. From my side, let's change it to be compared in index order.

bobbinth: I'm fine with this, but wondering if we should do this in a different PR as we'd need to thread the changes across 3 repos.

stdlib/asm/word.masm
```diff
@@ -0,0 +1,257 @@
...
+#! Inputs:  [RHS, LHS]
+#! Outputs: [is_lhs_greater]
+#!
+#! Cycles: 121
+export.gt
+    exec.arrange_words_adjacent
+    # => [2_3, 1_3, 2_2, 1_2, 2_1, 1_1, 2_0, 1_0]
+
+    push.1.0
+    # => [is_lhs_less, continue, 2_3, 1_3, 2_2, 1_2, 2_1, 1_1, 2_0, 1_0]
```
PhilippGackstatter: Since the inputs are no longer named `KEY1`/`KEY2` the naming here doesn't really make sense anymore. We should probably update it to use `r_3` and `l_3`, etc. Also applies to other procedures.

bobbinth:
I've merged this PR - thank you @Qyriad!

I think there are two potential follow-ups:

- Remove the `eqw` instruction. It is now redundant with `std::word::eq_test` and is different from `std::word::eq` (which could be confusing).
- Change how we do word comparison (to start with the least significant items first). This would require a change in `miden-crypto` first.

Also, we should probably start using functions from this module in `mdien-base` @PhilippGackstatter or @Fumuran - could you create an issue for this?

PR #1999, opened 2025-07-16
feat(processor): Add `AdviceProvider::advice_map` accessor

### LINKED ISSUE #1995: Add `AdviceMap` accessor for `AdviceInputs`
Add `AdviceMap` accessor for `AdviceInputs`, e.g. `AdviceInputs::map(&self) -> &AdviceMap`. Currently, the `AdviceInputs` does not enable access to the inner advice map which seems like it should be possible. 

We could change it to pass `AdviceInputs` to the function in question and use `AdviceInputs::mapped_values` but in other parts of the code we have an `AdviceProvider` whose map we want to pass to the same function, and in that case, `AdviceMap` would be the common API.

Not a pressing issue, but would be nice to have.

Related: https://github.com/0xMiden/miden-base/pull/1589#pullrequestreview-3020941742

adr1anh:
## Describe your changes

Fixes #1995 

- Updated `CHANGELOG.md'

### DIFF
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -1,7 +1,4 @@
-use alloc::{
-    collections::{BTreeMap, btree_map::Entry},
-    vec::Vec,
-};
+use alloc::{collections::BTreeMap, vec::Vec};
 
 use miden_core::{
     AdviceMap, Felt, Word,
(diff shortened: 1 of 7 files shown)


processor/src/host/advice/mod.rs
```diff
@@ -152,6 +152,11 @@ impl AdviceProvider {
     // ADVICE MAP
     // --------------------------------------------------------------------------------------------
 
+    /// Returns a reference to the inner [`AdviceMap`].
+    pub fn advice_map(&self) -> &AdviceMap {
```
plafer: Do we still need `contains_map_key()` and `get_mapped_values()` then if we expose `&AdviceMap`?

bobbinth: Yeah, if we have the `advice_map()` accessor, I don't really see a need for them.

processor/src/host/advice/mod.rs
```diff
@@ -152,39 +149,17 @@ impl AdviceProvider {
...
-                        prev_values: existing_values.to_vec(),
-                        new_values: values,
-                    });
-                }
-            },
-        }
-        Ok(())
+    /// the [`AdviceProvider::push_from_map()`] method.
+    pub fn map_mut(&mut self) -> &mut AdviceMap {
+        &mut self.map
```
bobbinth, line 162: Doesn't this change the existing behavior? Basically, now we'd be able to insert different values under the same key while previously this would have resulted in an error. If so, I'm not sure we should make this change.

bobbinth:
Since @adr1anh is going on vacation, I've replaced this PR with a simpler PR in #2009 (which only touches `AdviceInputs`).

PR #2000, opened 2025-07-16
chore: docs update, constants dont need to be before procedures

mmagician:
The docs currently state:
> Constants must be declared right after module imports and before any procedures or program bodies.

This is no longer true, as I can write:
```
use.kernel::account

export.foo
    ...
end

const.MY_CONST=1

export.BAR
    mem_load.MY_CONST
end
```

This PR removes the outdated sentence.


plafer (approved):
LGTM

Issue #1993, opened 2025-07-14
`ErrorContext` and `ProcessState` are not `Send`


PhilippGackstatter:
Version: 0.16.2

## `ErrorContext` is not `Send`

The following test does not compile:

```rust

#[cfg(test)]
mod async_tests {
    use alloc::{boxed::Box, string::String};

    use super::*;

    struct AsyncTransactionHost;

    #[derive(Debug, thiserror::Error)]
    enum ParseError {
        #[error("stub")]
        FailedToParse,
    }

    fn parse_event(_event_id: u32) -> Result<String, ParseError> {
        Err(ParseError::FailedToParse)
    }

    impl BaseHost for AsyncTransactionHost {}
    impl AsyncHost for AsyncTransactionHost{
        fn get_mast_forest(
            &self,
            _node_digest: &Word,
        ) -> impl Future<Output = Option<Arc<MastForest>>> + Send {
            async { None }
        }

        fn on_event(
            &mut self,
            process: &mut ProcessState<'_>,
            event_id: u32,
            err_ctx: &impl ErrorContext,
        ) -> impl Future<Output = Result<(), ExecutionError>> + Send {
            async move {
                let transaction_event = parse_event(event_id)
                    .map_err(|err| ExecutionError::event_error(Box::new(err), err_ctx))?;

                Ok(())
            }
        }
    }
}
```

results in:

```
error: future cannot be sent between threads safely
   --> processor/src/lib.rs:957:14
    |
957 |         ) -> impl Future<Output = Result<(), ExecutionError>> + Send {
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
    |
note: captured value is not `Send` because `&` references cannot be sent unless their referent is `Sync`
   --> processor/src/lib.rs:960:79
    |
960 |                     .map_err(|err| ExecutionError::event_error(Box::new(err), err_ctx))?;
    |                                                                               ^^^^^^^ has type `&impl ErrorContext` which is not `Send`, because `impl ErrorContext` is not `Sync`
help: consider restricting opaque type `impl ErrorContext` with trait `Sync`
    |
956 |             err_ctx: &impl ErrorContext + core::marker::Sync,
    |                                         ++++++++++++++++++++
```

## `ProcessState` is not `Send`

Similarly, replacing the above `on_event` impl with:

```rust
fn on_event(
  &mut self,
[...]

PR #2003, opened 2025-07-17
refactor: `on_event` should not require `&mut ProcessState` and error context

drahnr:
## Describe your changes

Removes the dependency of a `&mut ProcessState` by a layer of indirection, being `AdviceMutation` The mutations are applied after the respective call to the `(Sync|Async)Host`-trait implementors. The `handler` has been changed in the same fashion, returning a set of `AdivceMutation`s.  

Also remove the `impl err_ctx` to avoid another data dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
 - Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
+- [BREAKING] Change API on `on_event` and introduce the `AdviceMutation` ([#2003](https://github.com/0xMiden/miden-vm/pull/2003)).
 
 #### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
 - Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
-- [BREAKING] Change API on `on_event` and introduce the `AdviceMutation` ([#2003](https://github.com/0xMiden/miden-vm/pull/2003)).
+- [BREAKING] `{AsyncHost,SyncHost}::on_event` now returns a list of `AdviceProvider` mutations ([#2003](https://github.com/0xMiden/miden-vm/pull/2003)).
 
 #### Changes
 
--- a/processor/src/fast/sys_ops.rs
+++ b/processor/src/fast/sys_ops.rs
@@ -90,14 +90,14 @@ impl FastProcessor {
         host: &mut impl AsyncHost,
         err_ctx: &impl ErrorContext,
     ) -> Result<(), ExecutionError> {
-        let process = &mut self.state(op_idx);
+        let mut process = self.state(op_idx);
         // If it's a system event, handle it directly. Otherwise, forward it to the host.
         if let Some(system_event) = SystemEvent::from_event_id(event_id) {
-            handle_system_event(process, system_event, err_ctx)
+            handle_system_event(&mut process, system_event, err_ctx)
         } else {
             let clk = process.clk();
             let mutations = host
-                .on_event(&*process, event_id)
+                .on_event(&process, event_id)
                 .await
                 .map_err(|err| ExecutionError::event_error(err, event_id, err_ctx))?;
             for mutation in mutations {
--- a/processor/src/fast/sys_ops.rs
+++ b/processor/src/fast/sys_ops.rs
@@ -100,11 +100,9 @@ impl FastProcessor {
                 .on_event(&process, event_id)
                 .await
                 .map_err(|err| ExecutionError::event_error(err, event_id, err_ctx))?;
-            for mutation in mutations {
-                self.advice
-                    .apply_mutation(mutation)
-                    .map_err(|err| ExecutionError::advice_error(err, clk, err_ctx))?;
-            }
+            self.advice
+                .apply_mutations(mutations)
+                .map_err(|err| ExecutionError::advice_error(err, clk, err_ctx))?;
             Ok(())
         }
     }
--- a/processor/src/fast/sys_ops.rs
+++ b/processor/src/fast/sys_ops.rs
@@ -95,9 +95,17 @@ impl FastProcessor {
         if let Some(system_event) = SystemEvent::from_event_id(event_id) {
             handle_system_event(process, system_event, err_ctx)
         } else {
-            host.on_event(process, event_id)
+            let clk = process.clk();
+            let mutations = host
+                .on_event(&*process, event_id)
                 .await
-                .map_err(|err| ExecutionError::event_error(err, event_id, err_ctx))
+                .map_err(|err| ExecutionError::event_error(err, event_id, err_ctx))?;
+            for mutation in mutations {
+                self.advice
+                    .apply_mutation(mutation)
+                    .map_err(|err| ExecutionError::advice_error(err, clk, err_ctx))?;
+            }
+            Ok(())
         }
     }
 }
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -14,6 +14,8 @@ pub use inputs::AdviceInputs;
 mod errors;
 pub use errors::AdviceError;
 
+use crate::host::AdviceMutation;
+
 // TYPE ALIASES
 // ================================================================================================
 
@@ -45,6 +47,51 @@ pub struct AdviceProvider {
 }
 
 impl AdviceProvider {
+    pub fn apply_mutation(&mut self, mutation: AdviceMutation) -> Result<(), AdviceError> {
+        match mutation {
+            AdviceMutation::PopStack => {
+                self.pop_stack()?;
+            },
+            AdviceMutation::PopStackWord => {
+                self.pop_stack_word()?;
+            },
+            AdviceMutation::PopStackDword => {
+                self.pop_stack_dword()?;
+            },
+            AdviceMutation::PushStack { value } => {
+                self.push_stack(value);
+            },
+            AdviceMutation::PushStackWord { word } => {
+                self.push_stack_word(&word);
+            },
+            AdviceMutation::PushFromMap { key, include_len } => {
+                self.push_from_map(key, include_len)?;
+            },
+            AdviceMutation::ExtendStack { iter } => {
+                self.extend_stack(iter);
+            },
+            AdviceMutation::InsertIntoMap { key, values } => {
+                self.insert_into_map(key, values)?;
+            },
+            AdviceMutation::ExtendMap { other } => {
+                self.extend_map(&other)?;
+            },
+            AdviceMutation::UpdateMerkleNode { root, depth, index, value } => {
+                self.update_merkle_node(root, &depth, &index, value)?;
+            },
+            AdviceMutation::MergeRoots { lhs, rhs } => {
+                self.merge_roots(lhs, rhs)?;
+            },
+            AdviceMutation::ExtendMerkleStore { iter } => {
+                self.extend_merkle_store(iter);
+            },
+            AdviceMutation::ExtendFromInputs { inputs } => {
+                self.extend_from_inputs(&inputs)?;
+            },
+        }
+        Ok(())
+    }
+
     // ADVICE STACK
     // --------------------------------------------------------------------------------------------
 
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -47,46 +47,24 @@ pub struct AdviceProvider {
 }
 
 impl AdviceProvider {
-    pub fn apply_mutation(&mut self, mutation: AdviceMutation) -> Result<(), AdviceError> {
+    /// Apply the mutations given in order to the `AdviceProvider`.
+    pub fn apply_mutations(
+        &mut self,
+        mutations: impl IntoIterator<Item = AdviceMutation>,
+    ) -> Result<(), AdviceError> {
+        mutations.into_iter().try_for_each(|mutation| self.apply_mutation(mutation))
+    }
+
+    fn apply_mutation(&mut self, mutation: AdviceMutation) -> Result<(), AdviceError> {
         match mutation {
-            AdviceMutation::PopStack => {
-                self.pop_stack()?;
-            },
-            AdviceMutation::PopStackWord => {
-                self.pop_stack_word()?;
-            },
-            AdviceMutation::PopStackDword => {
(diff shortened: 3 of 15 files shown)


processor/src/host/default.rs
```diff
@@ -86,33 +86,33 @@ impl<D: DebugHandler> DefaultHost<D> {
 impl BaseHost for DefaultHost {
     fn on_debug(
         &mut self,
-        process: &mut ProcessState,
+        process: &ProcessState,
```
plafer: Was there a specific reason for this change? The main goal was for methods of `AsyncHost` to not require `&mut ProcessState`, but I don't think it's *needed* for methods of `BaseHost` (although it's up for discussion whether it's a good idea to allow the host to modify the advice provider on debug/trace/assert)

processor/src/fast/sys_ops.rs
```diff
@@ -95,9 +95,17 @@ impl FastProcessor {
         if let Some(system_event) = SystemEvent::from_event_id(event_id) {
             handle_system_event(process, system_event, err_ctx)
         } else {
-            host.on_event(process, event_id)
+            let clk = process.clk();
+            let mutations = host
+                .on_event(&*process, event_id)
```
plafer: nit: if we define `process` as

```rs
let mut process_state = self.state(op_idx);
```

Then we can take `&process_state` here (and `&mut process_state` in `handle_system_event()`). I actually didn't notice before we were initializing `process` that way, and would change it as above in any case, so might as well do it here :)

processor/src/host/advice/mod.rs
```diff
@@ -45,6 +47,51 @@ pub struct AdviceProvider {
 }
 
 impl AdviceProvider {
+    pub fn apply_mutation(&mut self, mutation: AdviceMutation) -> Result<(), AdviceError> {
```
plafer: This `AdviceMutation` enum is more a `Host`/`FastProcessor` concept than an `AdviceProvider` concept (i.e. no one having a reference to an `AdviceProvider` will want to use this method). So I would put it in a helper function next to `op_emit()`

processor/src/host/handlers.rs
```diff
@@ -119,13 +122,17 @@ impl EventHandlerRegistry {
...
+    pub fn handle_event(
+        &self,
+        id: u32,
+        process: &ProcessState,
+    ) -> Result<Option<Vec<AdviceMutation>>, EventError> {
         if let Some(handler) = self.handlers.get(&id) {
-            handler.on_event(process)?;
-            return Ok(true);
+            let mutations = handler.on_event(process)?;
+            return Ok(Some(mutations));
```
```suggestion
            return handler.on_event(process).map(Some);
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
 - Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
+- [BREAKING] Change API on `on_event` and introduce the `AdviceMutation` ([#2003](https://github.com/0xMiden/miden-vm/pull/2003)).
```
plafer:
```suggestion
- [BREAKING] `{AsyncHost,SyncHost}::on_event` now returns a list of `AdviceProvider` mutations ([#2003](https://github.com/0xMiden/miden-vm/pull/2003)).
```

plafer (commented):
That's great thank you! Left some nits.

As discussed offline, after this PR, we still are left with the issue that the returned `Future`s from `AsyncHost` methods are not `'static` (specifically since `ProcessState` is not `'static`), but this is a great first step.

processor/src/host/default.rs
drahnr: I think that was an overshot on my end. I can carve this out into a separate PR; regardless, it looked rather odd to me a debug or trace adapter should be able to modify the processor state.

processor/src/host/advice/mod.rs
drahnr: There are two places this is used, and I don't want to duplicate it, which means it sits either in `host` as a `pub(crate) fn` or I tag it onto either `AdviceProvider` or `AdviceMutation`. I'd prefer either of the latter two.

processor/src/host/handlers.rs
```diff
@@ -119,13 +122,16 @@ impl EventHandlerRegistry {
     ///
     /// Returns a bool indicating whether the event was handled. If the event was handled but
     /// returned an error, it is propagated to the caller.
-    pub fn handle_event(&self, id: u32, process: &mut ProcessState) -> Result<bool, EventError> {
+    pub fn handle_event(
```
adr1anh, line 127: We should probably keep the flag indicating whether the event was handled or not, along with the mutations. 

It's not used by the `DefaultHost` but is required for a custom `Host` implementation which would want to use the registry but fall back to other events.

If this is not required, the docstring needs to be updated.

processor/src/host/mod.rs
```diff
@@ -70,17 +116,11 @@ pub trait SyncHost: BaseHost {
...
-            "Event with id {} emitted at step {} in context {}",
-            event_id,
-            process.clk(),
-            process.ctx()
-        );
-        Ok(())
-    }
+    fn on_event(
+        &mut self,
+        process: &ProcessState,
```
adr1anh, line 89: More of a general comment. 

We should be able to pass `process: ProcessState` rather than by reference. It was changed to `&mut` to allow advice mutations, but that's no longer required with this PR.

processor/src/host/mod.rs
```diff
@@ -18,6 +21,53 @@ use handlers::DebugHandler;
...
+
+/// Any possible way an event can modify the advice map
+#[derive(Debug, PartialEq, Eq)]
+pub enum AdviceMutation {
+    PopStack,
+    PopStackWord,
+    PopStackDword,
+    PushStack {
+        value: Felt,
+    },
```
adr1anh:
```suggestion
    PushStack(Felt),
```
 The `value` seems clear from the context, but up to you!

processor/src/host/mod.rs
plafer: Although `ProcessState` was designed to be small, it's currently a enum wrapping `SlowProcessState`, which is 4 references (i.e. a bit big)... So one could argue there's a slight preference to pass it as `&ProcessState` still?

processor/src/host/advice/mod.rs
plafer: Ah yes actually making it an `AdviceMutation` method sounds ideal to me!

plafer:
We should also probably create a new `asynchost-send-experimental` branch (or whatever) off of `next`, and re-target this PR against it.

processor/src/host/default.rs
plafer: Probably fine to leave it as-is, as I don't think anyone would be using the `&mut` anyways

processor/src/host/mod.rs
adr1anh: That's true. My hunch was that passing by value might help the compiler bypass the enum altogether. But that's something that can be investigated later.

crates/utils/testing/src/host.rs
bobbinth, line 1: This should probably be reverted.

processor/src/host/mod.rs
```diff
@@ -18,6 +21,53 @@ use handlers::DebugHandler;
 mod mast_forest_store;
 pub use mast_forest_store::{MastForestStore, MemMastForestStore};
 
+// ADVICE MAP MUTATIONS
+// ================================================================================================
+
+/// Any possible way an event can modify the advice map
+#[derive(Debug, PartialEq, Eq)]
+pub enum AdviceMutation {
```
bobbinth, line 29: I don't think we actually need all of these variants:

- Variants that pop values from the stack are not useful in our context because the returned values would never make it back to the author of the mutation. But I don't think that's an issue - we just don't need them.
- Some of the variants can be collapsed because semantically, they'll be handle in the same way.

So, I think we can simplify this as:
```rust
pub enum AdviceMutation {
    /// Pushes the values onto the advice stack.
    ExtendStack(Vec<Felt>),
    /// Inserts an entry into the advice map.
    ExtendMap(Word, Vec<Felt>),
    /// Inserts the nodes into the Merkle store.
    ExtendMerkleStore (Vec<InnerNodeInfo>),
}
```

stdlib/tests/crypto/falcon.rs
```diff
@@ -72,10 +72,9 @@ pub fn push_falcon_signature(process: &mut ProcessState) -> Result<(), EventErro
     let result = falcon_sign(pk_sk, msg)
         .ok_or(FalconError::MalformedSignatureKey { key_type: "RPO Falcon512" })?;
 
-    for r in result {
-        process.advice_provider_mut().push_stack(r);
-    }
-    Ok(())
+    Ok(Vec::from_iter(
+        result.into_iter().map(|value| AdviceMutation::PushStack { value }),
+    ))
```
bobbinth: We should be able to do this as a single mutation: `AdviceMutation::ExtendStack(result)`.

processor/src/host/advice/mod.rs
bobbinth: I would maybe change this to `apply_mutations()` - something like:
```rust
pub fn apply_mutations(
    &mut self,
    mutations: impl IntoIter<Item = AdviceMutation>
) -> Result<(), AdviceError>
```
The primary use case for this is to be used with a list of mutations.

processor/src/host/default.rs
bobbinth: I'd keep as is - better to be consistent across both hosts.

processor/src/host/handlers.rs
bobbinth: I think maybe `None` serves the same purpose now?

Basically, if the registry didn't handle the event, it'll return `None`, but if it did, it would return `Some(Vec)`

But we do need to update the doc comment with this.

bobbinth (commented):
Looks good! Thank you. I left some comments inline - the main one is about simplifying the `AdviceMutation` enum.

bobbinth:
> We should also probably create a new `asynchost-send-experimental` branch (or whatever) off of `next`, and re-target this PR against it.

Yes, probably a good idea - though, I wonder if these changes are a net good regardless.

processor/src/host/handlers.rs
adr1anh: Oh right, never mind then. If an event was handled it would return `Some(vec![])`. The docstring still needs updating though.

processor/src/host/mod.rs
plafer: > Variants that pop values from the stack are not useful in our context because the returned values would never make it back to the author of the mutation. But I don't think that's an issue - we just don't need them.

But then we wouldn't allow `on_event()` to pop things off the advice stack; that seems restrictive?

plafer:
> Yes, probably a good idea - though, I wonder if these changes are a net good regardless.

My only "concern" was that some `on_event()` handlers could become quite complex, if they e.g. need to pop the stack, push some new values, write to the map based on the values popped from the stack, etc. But if that's not a real concern then we could merge this regardless indeed.

processor/src/host/handlers.rs
drahnr: Updated the doc string

crates/utils/testing/src/host.rs
drahnr: Not sure which thing this was pointed to, the comment lost its anchor I think

processor/src/host/mod.rs
drahnr: Removed the variant entirely :)

processor/src/host/mod.rs
drahnr: Ack, I'll leave as reference for now.

processor/src/host/mod.rs
drahnr: If I think of the stack as input params, the mutation code a function call, and the return value of that to be put back on the stack, then I'd expect popping the arguments off the stack would be a common pattern

bobbinth: > But then we wouldn't allow `on_event()` to pop things off the advice stack; that seems restrictive?

`on_event()` rarely needs to do it (in fact, in the uses we have now, we don't pop anything from the advice stack in the event handlers, I believe) - but maybe good to have nevertheless. In this case, I'd add one more variant - something like `TruncateStack(usize)` to tell how many elements to drop from the stack. But this could be a separate PR as well.

bobbinth: > If I think of the stack as input params, the mutation code a function call, and the return value of that to be put back on the stack, then I'd expect popping the arguments off the stack would be a common pattern

We'd still be popping values off the stack - but that would be done inside the VM itself (i.e., not in the event handlers).

processor/src/host/mod.rs
```diff
@@ -18,6 +21,17 @@ use handlers::DebugHandler;
...
+    ExtendStack { iter: Vec<Felt> },
+    ExtendMap { other: AdviceMap },
+    ExtendMerkleStore { iter: Vec<InnerNodeInfo> },
+}
```
bobbinth, line 33: I would add doc comments to the variants to explain how they affect the advice provider.

processor/src/host/advice/mod.rs
```diff
@@ -45,6 +47,29 @@ pub struct AdviceProvider {
 }
 
 impl AdviceProvider {
+    /// Apply the mutations given in order to the `AdviceProvider`.
+    pub fn apply_mutations(
```
bobbinth, line 51: We may be able to get rid of some of the mutation functions now. For example, I'm not sure `push_stack` and `push_stack_word` are needed any more. But we can do this also later.

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline.

drahnr:
@plafer @bobbinth shall we create the experimental branch as an intermediate step or merge to `next` as-is?

bobbinth:
> @plafer @bobbinth shall we create the experimental branch as an intermediate step or merge to `next` as-is?

Let's create an experimental branch for now. But unless we run into some problems with this in `miden-base`, I think we'll probably end up merging this into `next` regardless.

drahnr:
Pushed to branch: `experimental`, what are the next steps here? I'd continue with the `'static` requirement @plafer mentioned in a DM?

plafer:
> Pushed to branch: experimental, what are the next steps here?

I think what we had in mind was: create a new branch `async-host-experimental` from `next`, and then retarget this PR to that. This way when we're happy with the review we merge it, and then later on we can merge that branch into `next` (or cherry-pick commits).

> I'd continue with the 'static requirement @plafer mentioned in a DM?

Still not clear what the best way to do this one is - we can discuss offline.

bobbinth:
I've created an `async-experimental` branch and will merge this PR into it. @PhilippGackstatter, after this, you'll be able to use `async-experimental` in `miden-base` to see if you get unblocked with trying to make `TransactionAuthenticator` async.

I think the next steps here are:
- Address https://github.com/0xMiden/miden-vm/issues/1942.
- Add conditional compilation along the lines of what is described here https://github.com/0xMiden/miden-vm/issues/1993#issuecomment-3082701577.
  - As a part of this, we should also figure out what bounds we need for `std` vs. `no_std` options.

I will also delete the `experimental` branch because `async-experimental` will supersede it.

PR #2006, opened 2025-07-17
Exploration of using the sans-IO pattern in the `FastProcessor`

plafer:
**Disclaimer:** this PR is not meant to be merged. It is meant to get a feel for what the entire `FastProcessor` would look like if calls to the `AsyncHost` were replaced by the sans-IO pattern (first discussed [here](https://github.com/0xMiden/miden-base/issues/401#issuecomment-2658767755)).

I only implemented swapping the `AsyncHost::get_mast_forest()` call with the sans-IO alternative when processing an `ExternalNode`. The API is very rough, and very much not cleaned up.

The interesting public methods are `FastProcessor::{execute_sans_io, execute_sans_io_finish_external, execute_sync_host, execute_async_host}`. Internally, an external node is processed in `{start, finish}_external_node()` (and the new `MastForest` is fetched by the caller).

Observations:
- We are able to offer sync and `async` APIs effortlessly (👍👍👍)
- The performance is likely to be much better than what we have with the current async-only API, since the core of the functionality is synchronous
  - I am not able to actually benchmark it because this PR doesn't actually run due to the bunch of `unimplemented!()` I inserted 🙂
- We would expose the sans-io API on top of the sync/async (making @Mirko-von-Leipzig happy)

### DIFF
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -1,14 +1,14 @@
 use alloc::{sync::Arc, vec::Vec};
-use core::cmp::min;
+use core::{cmp::min, ops::ControlFlow};
 
 use memory::Memory;
 use miden_air::RowIndex;
 use miden_core::{
-    Decorator, DecoratorIterator, EMPTY_WORD, Felt, ONE, Operation, Program, StackOutputs,
+    Decorator, DecoratorIterator, EMPTY_WORD, Felt, Kernel, ONE, Operation, Program, StackOutputs,
     WORD_SIZE, Word, ZERO,
     mast::{
-        BasicBlockNode, CallNode, ExternalNode, JoinNode, LoopNode, MastForest, MastNode,
-        MastNodeId, OP_GROUP_SIZE, OpBatch, SplitNode,
+        BasicBlockNode, CallNode, JoinNode, LoopNode, MastForest, MastNode, MastNodeId,
+        OP_GROUP_SIZE, OpBatch, SplitNode,
     },
     stack::MIN_STACK_DEPTH,
     utils::range,
@@ -16,11 +16,7 @@ use miden_core::{
 use miden_debug_types::{DefaultSourceManager, SourceManager};
 
 use crate::{
-    AdviceInputs, AdviceProvider, ContextId, ErrorContext, ExecutionError, FMP_MIN, ProcessState,
-    SYSCALL_FMP_MIN, SyncHost,
-    chiplets::Ace,
-    continuation_stack::{Continuation, ContinuationStack},
-    err_ctx,
+    chiplets::Ace, continuation_stack::{Continuation, ContinuationStack}, err_ctx, AdviceInputs, AdviceProvider, AsyncHost, ContextId, ErrorContext, ExecutionError, ProcessState, SyncHost, FMP_MIN, SYSCALL_FMP_MIN
 };
 
 mod memory;
@@ -62,6 +58,25 @@ const WORD_SIZE_FELT: Felt = Felt::new(4);
 
 /// The size of a double-word.
 const DOUBLE_WORD_SIZE: Felt = Felt::new(8);
+
+pub enum ProcessingState {
+    /// When returned from `execute_sans_io()`, the caller must call `execute_sans_io()` again, this
+    /// time providing a `MastForest` that contains `node_digest` as a procedure.
+    ///
+    /// For this experimental branch, since we only implement the `GetMastForest` continuation for
+    /// the external node, this variant is assumed to come from an `ExternalNode`. In the real
+    /// implementation, it could also come from a `DynNode`.
+    GetMastForest {
+        node_digest: Word,
+    },
+    OnEvent {
+        event_id: u32,
+    },
+    /// Execution has successfully completed.
+    Done(StackOutputs),
+    /// An error has occurred, which terminates the execution early.
+    Error(ExecutionError),
+}
 
 /// A fast processor which doesn't generate any trace.
 ///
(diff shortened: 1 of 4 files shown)


processor/src/fast/mod.rs
```diff
@@ -668,86 +759,24 @@ impl FastProcessor {
...
-        }
-        Ok(())
+        _current_node_id: MastNodeId,
+        _current_forest: &Arc<MastForest>,
+        _continuation_stack: &mut ContinuationStack,
+    ) -> ControlFlow<ProcessingState> {
+        // Would be a similar solution than `ExternalNode` in the case where we need to fetch a new
+        // MastForest from the Host, so leaving unimplemented in this PR for simplicity.
+        //
+        // Q: Does `DynNode` really have to fetch a `MastForest` ever? i.e. doesn't the digest
```
bitwalker, line 776: It's not necessarily an `ExternalNode`, it could also be another concrete node type in the current forest as well - but I think the point stands that `DynNode` is never going to be the node that needs to fetch a new forest.

processor/src/fast/mod.rs
```diff
@@ -769,51 +797,70 @@ impl FastProcessor {
...
-        self.execute_before_enter_decorators(node_id, current_forest, host)?;
+        self.execute_before_enter_decorators(node_id, current_forest)?;
 
         let external_node = current_forest[node_id].unwrap_external();
-        let (root_id, new_mast_forest) = self.resolve_external_node(external_node, host).await?;
 
-        // Push current forest to the continuation stack so that we can return to it
         continuation_stack.push_enter_forest(current_forest.clone());
 
+        // NOTE: It is the responsibility of the caller to call `execute_sans_io_finish_external()`
```
bitwalker, line 816: I'm not sure there is - loading a new `MastForest` is almost certainly one of those things we'd want to make async in async contexts, so either the processor needs to call into code that is itself written in sans-io style, or we have to yield and have the caller pass us the new forest as you've done here.

processor/src/fast/mod.rs
```diff
@@ -1031,201 +1058,20 @@ impl FastProcessor {
...
-        Ok(())
+        _operation: &Operation,
+        _op_idx: usize,
+        _program: &MastForest,
+        _err_ctx: &impl ErrorContext,
+    ) -> ControlFlow<ProcessingState> {
+        // This method is not interesting in this exploration PR - we basically have to convert all
+        // callees to return `ControlFlow` instead of `ExecutionError`, so would only inflate the
+        // diff with no relevant information.
+        unimplemented!()
```
bitwalker, line 1069: My question would be whether there are any operations which might need to be `async` in async contexts - if so, then that will require the implementation here to yield to the caller with some new yield type (depends on what we're expecting the caller to do), and have the caller resume with whatever state/information we needed to obtain for the op.

bitwalker:
Looks more or less like I'd expect!

I think the trick here is going to be whether the processor needs to call into any code outside of the processor which we would need to make `async` for async contexts, as mentioned in some of my other comments. That code would need to also be written in this style, or we'd need to yield and make it the callers responsibility to handle those things, which may complicate the implementation here a fair bit (depending on where the suspension points are, we may end up needing to manually construct objects representing the state of the processor at that point, just like `async`/`await` do under the covers). That's not to say it isn't worth doing that for more control here - just something to bear in mind.

PhilippGackstatter (commented):
Nice! Definitely showcases that it's easy to call this in both sync and async style.

The API surface definitely becomes a lot larger, naturally because the internals are much more exposed. I think this approach requires really good documentation to let callers know how to correctly use the API, e.g. how to continue execution after being done with `GetMastForest` or `OnEvent` handling.

So while this is a good option, I would personally still prefer using the [`#[cfg(target)]` approach](https://github.com/0xMiden/miden-vm/issues/1993#issuecomment-3083752093) if it works.

processor/src/fast/mod.rs
plafer: The only such operation is `Emit`, and is already handled outside this function (and would return a `ProcessingState::GetEvent` to the caller, which would need to call back in). So this function would never yield back control to the caller even in the final implementation.

processor/src/fast/mod.rs
plafer: I see most callers using the `execute_sync_host()`/`execute_async_host()` methods, as opposed to the lower-level sans-IO API (which could arguably be private, but @Mirko-von-Leipzig would prefer using that directly IIUC). Hence I see this approach almost as an implementation detail of the `FastProcessor`, which allows it to execute faster, and easily expose `SyncHost` and `AsyncHost`-based APIs.

Then for the `AsyncHost`-based API, we are left with the `Send`/`'static` bounds problem, which we'd want to try based on std/no-std.

cc @PhilippGackstatter

PhilippGackstatter: Oh, I see. I think I misunderstood this then. I thought `execute_sync_host` was just to showcase how, e.g. miden-base would call into this, because I thought the goal was to expose the sans-IO-ness of the processor. If it's just an internal detail, then of course that alleviates my "documentation and complexity concerns" :sweat_smile:.

Mirko-von-Leipzig: I think the nice thing about this is that you can very easily choose and change as you see fit without having to refactor the world.

From the node's perspective we don't really see the VM - we just see `miden-base`. In terms of full performance, having the ability how to fetch the data at the moment itself is optimal. In the short term I just want an interface that is `Send` at least.

processor/src/fast/mod.rs
plafer: > but I think the point stands that DynNode is never going to be the node that needs to fetch a new forest.

Looked a bit more into it, and we don't currently assume that the procedure digest on the stack represents a procedure in the current MAST forest - we'll basically go look for a new forest that contains that procedure digest in the `Host`. So for example, someone could `dynexec` a procedure in `stdlib` if they wanted. I can't think of any reasons we'd want to explicitly disallow that though?

processor/src/fast/mod.rs
plafer: So then to start, we could leave the sans-IO API private for simplicity, and if we want to propagate the sans-IO pattern through the `TransactionExecutor` at some point, then we can make it public.

processor/src/fast/mod.rs
bitwalker: We definitely want to allow that IMO

processor/src/fast/mod.rs
```diff
@@ -315,99 +330,166 @@ impl FastProcessor {
...
+            ControlFlow::Break(processing_state) => processing_state,
+        }
+    }
+
+    fn execute_sans_io_impl(
+        &mut self,
+        continuation_stack: &mut ContinuationStack,
+        initial_forest: &Arc<MastForest>,
+        kernel: &Kernel,
+    ) -> ControlFlow<ProcessingState> {
```
bobbinth, line 430: Question: why do we return `ControlFlow<ProcessingState>` from here? From what I can tell, the `Continue` option is never really used. So, we could probably just return `ProcessingState>, no?

processor/src/fast/mod.rs
bobbinth: I think handling the `Emit` operation will be pretty tricky. We'll basically need to capture all the relevant state to be able to resume execution of a basic block - which I think currently is a lot (because of the op group/batch structure). This will be made much simpler once we insert padding `Noop`'s into basic blocks at assembly time. Once this is done, I think we may be able to get away with just tracking the operation index.

bobbinth (commented):
Looks great! Thank you for trying this out. Overall, I agree that we could first use this approach as an internal implementation, and then expose the sans-io API publicly later on. But I also think making it work right now will be pretty difficult, and so, it may be worth first simplifying how basic blocks are executed first (e.g., via padding `Noop` insertion at assembly time).

------

Some thoughts regarding the public interface.

First, as mentioned above, we'd always want to provide sync and async wrappers around the sans-io API (I would name them `execute()` and `execute_async()`). These wrappers would look like pretty much as you have them now.

For the sans-io API, I'm wondering if we can simplify it by making everything go through the `FastProcessor` struct. It could look something like this:

```rust
let mut processor = FastProcessor::new(...);

loop {
    match processor.next() {
        ProcessingState::GetMastForest { node_digest } => {
            let new_mast_forest = /* get the new MAST forest from somewhere */
            processor.set_current_forest(new_mast_forest);
        }
        ProcessingState::OnEvent { event_id } => {
            let advice_mutations = /* get the list of mutations from somewhere */ ;
            processor.apply_advice_mutations(advice_mutations)?;
        },
        ProcessingState::Done { stack_outputs, advice_provider } => {
            break Ok((stack_outputs, advice_provider))
        }
        ProcessingState::Error(execution_error) => break Err(execution_error),
    }
}
```

It is still a bit of boilerplate, but it doesn't seem too bad. Thought, I haven't thought it through completely to know if it will actually work.

processor/src/fast/mod.rs
plafer: This is so that we can use the `?` operator - this is "the root method" that uses `ControlFlow<>`, and then `execute_sans_io()` and others unwrap it into something nice for the public API

plafer:
Closing this as I feel like it served its purpose. Takeaways:
1. This is probably a good idea to be able to expose both sync & async interfaces with optimal performance and high code reuse,
2. We have the option to expose the sans-IO API, but can be internal only to start,
3. Probably best to simplify the `BasicBlockNode` structure first as mentioned by @bobbinth

PR #2008, opened 2025-07-18
Make recursive verifier in `stdlib` generic

Al-Kindi-0:
## Describe your changes
This is a PoC of how we can make the recursive verifier in `stdlib` generic. The main idea is to isolate the components of the existing recursive verifier that are "circuit"/AIR specific, and then use AirScript in order to generate these components using the ACE backend. These components are then dynamically called within the `stdlib` recursive verifier.

The `stdlib` recursive verifier now takes an additional input, namely the hash of all, actually 4, dynamically invoked procedures. The `stdlib` recursive verifier will unhash these and store them at some pre-defined memory region so that they can be latter dynamically invoked.

This PR contains an instantiation of this pattern using the Miden VM recursive verifier (in `stdlib/asm/crypto/miden_vm_verifier/`).

Note that there is a spectrum in how much dynamic invocation we use, more dynamic invocations will mean less code that needs to be generated per AIR but at the cost of more cycles due to the indirection, while less dynamic invocation will mean more code that needs to be generated per AIR but with the benefit of better performance.
This explains the choice to dynamically invoke `load_query_row` instead of the smaller called procedures `load_main_segment_execution_trace` and `load_constraints_composition_polys_trace`, as this part of the code is the most critical.

With this PR the cycle count for verifying a $2^{20}$ execution trace is $32500$ cycles compared to $31915$ in `next`.

One final note, if we decide to make the handling of variable length public inputs generic due to the acceptable performance cost, the codegened code per AIR could consist only of the code for ACE circuit, the code to computed the deep queries and the code to process the OOD evaluations.

- Updated `CHANGELOG.md'


stdlib/asm/crypto/miden_vm_verifier/mod.masm
```diff
@@ -0,0 +1,12 @@
+use.std::crypto::miden_vm_verifier::helper
+use.std::crypto::stark::verifier
+
+# Input: [log(trace_length), num_queries, grinding, ...]
+# Output: [log(trace_length), num_queries, grinding, num_constraints, trace_info, num_fixed_len_pi, num_ace_inputs, num_ace_gates, max_cycle_len_log, ROOT, ...]
+export.verify
```
bobbinth: I think the description of outputs here is incorrect. My understanding is that we are not returning anything from this procedure (i.e., if the procedure succeeds, the proof is verified, but if proof verification fails, we panic somewhere inside of the procedure.

stdlib/asm/crypto/miden_vm_verifier/mod.masm
```diff
@@ -0,0 +1,12 @@
...
+    exec.helper::prologue
+    # => [num_constraints, trace_info, num_fixed_len_pi, num_ace_inputs, num_ace_gates, max_cycle_len_log, ROOT, log(trace_length), num_queries, grinding, ...]
```
bobbinth: Given that the code in the `helper.masm` is fairly small, I would bring it into this module, and maybe even just integrate it into this procedure.

stdlib/asm/crypto/miden_vm_verifier/helper.masm
```diff
@@ -0,0 +1,68 @@
...
+    exec.constraints_eval::get_num_constraints
+    # => [num_constraints, trace_info, num_fixed_len_pi, num_ace_inputs, num_ace_gates, max_cycle_len_log, ROOT, log(trace_length), num_queries, grinding, ...]
+
+    # 3) Organize the stack and return
+    movup.12
+    movup.12
+    movup.12
+end
+
+proc.build_root_dynamic_procedures.20
```
bobbinth: Question: why do we need to "compress" all dynamic procedure roots into a single `ROOT`? Basically, why not return something like this from this procedure:
```
[LOAD_QUERY_ROW_RROC_ROOT, LOAD_ACE_CIRCUIT_PROC_ROOT, LOAD_PROCESS_OOD_ROW_PROC ROOT, LOAD_PROCESS_PUB_INPUTS_PROC_ROOT]
```

stdlib/asm/crypto/miden_vm_verifier/helper.masm
```diff
@@ -0,0 +1,68 @@
+use.std::crypto::miden_vm_verifier::deep_queries
+use.std::crypto::miden_vm_verifier::constraints_eval
+use.std::crypto::miden_vm_verifier::ood_frames
+use.std::crypto::miden_vm_verifier::public_inputs
+use.std::crypto::miden_vm_verifier::helper
+use.std::crypto::hashes::rpo
+
+const.TRACE_INFO=0x50010810
```
bobbinth: What does this encode?

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -26,28 +26,23 @@ const.MAX_CYCLE_LEN_LOG=3
 # CONSTANTS GETTERS
 # =================================================================================================
```
bobbinth: It is not clear to me why we need the constant getters below. It seems like we only use these constants internally? If so, we can just reference constants directly.

Also, we should now be able to declare a single constant for the evaluation circuit digest (we now support word-sized constants).

Lastly, shouldn't we also have a section with advice map inputs somewhere in this file?

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -26,28 +26,23 @@ const.MAX_CYCLE_LEN_LOG=3
...
+proc.get_num_constraints
+    push.NUM_CONSTRAINTS
 end
 
-export.get_max_cycle_length_log
-    push.MAX_CYCLE_LEN_LOG
-end
 
 # HELPER PROCEDURES
 # =================================================================================================
```
bobbinth: This comment is for the `load_ace_circuit_description` procedure below (but I can't comment on that line):

What are the inputs/outputs of this procedure. Initially, I assumed it would write the circuit description to some location in memory (specified as a parameter via the stack), but it seems to be doing something else.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
bobbinth, line 1: Is `load_query_row` the only procedure that is supposed to be exported from this module? I couldn't tell exactly because it is declared as `proc.load_query_row` while there are several other procedures declared with `export`.

Separately, there are several sequences like this in this file:
```
movup.4
assert_eq
movup.3
assert_eq
movup.2
assert_eq
assert_eq
```
I think these are equivalent to the `assert_eqw` instruction - no?

Lastly, and not for this PR, I wonder if there is a way to refactor the procedures in this file not to rely on temp addresses from the `constants` file. Naively, we can replaces these with locals, but that would add 7 extra cycles, I think.

stdlib/asm/crypto/miden_vm_verifier/ood_frames.masm
bobbinth, line 1: Would be great to add a more complete description to the `process_row_ood_evaluations` procedure (i.e., describing inputs/outputs).

stdlib/asm/crypto/miden_vm_verifier/public_inputs.masm
bobbinth, line 1: It doesn't seems like the information about fixed length public inputs is used anywhere in this file - do we need it?

Also, similar comment as in the other file re using constants directly rather than wrapping them in procedures (e.g., `KERNEL_OP_LABEL` is declared and used in this file).

Also, maybe this is for a different PR, but would be great to improve the readability of the `reduce_var_len_public_inputs_groups` procedure. For example, I'd move out the storing of the computed kernel digest input into the main `reduce_kernel_digests` procedure so that it looks like:
```
#! Inputs:  [num_kernel_procedures, ...]
#! Outputs: [prod_acc1, prod_acc0, ...]
proc.reduce_kernel_digests
    ...
end
```

stdlib/asm/crypto/stark/public_inputs.masm
```diff
@@ -86,7 +85,7 @@ export.load_public_inputs
     
     ## a) Compute the number of iterations.
     ##    We also check double-word alignment
-    exec.public_inputs::get_num_fixed_len_public_inputs
+    exec.constants::get_num_fixed_len_pi
```
bobbinth: If we "specialize" `load_public_inputs` procedure (i.e., move it under `miden_vm_verifier`), we should be able to avoid the while loops below, right?

stdlib/asm/crypto/stark/public_inputs.masm
bobbinth, line 1: Not for this PR, but I wonder if there is a good way to combine `load_public_inputs` and `reduce_variable_length_public_inputs` into a single procedure.

stdlib/asm/crypto/stark/verifier.masm
```diff
@@ -51,18 +51,22 @@ export.verify
     #
     # Cycles: 45
     exec.utils::validate_inputs
-    # => [log(trace_length), num_queries, grinding, ...]
+    # => [log(trace_length), num_queries, grinding, num_constraints, trace_info, num_fixed_len_pi, num_ace_inputs, num_ace_gates, max_cycle_len_log, ROOT, ...]
```
bobbinth: Question: why do we care about `num_fixed_len_pi`, `num_ace_inputs`, `num_ace_gates`, and `max_cycle_len_log` in the generic verifier?

bobbinth (commented):
Looks good! Thank you! I left some comments inline - but overall, I think this structure could work quite well.

In terms of trade-offs, I'm wondering if we should "specialized" a bit more of the code. Specifically, it feels like:

- Dealing with public inputs could be a bit more specialized (I left a comment about this inline).
- Constraint evaluation could be more specialized as well. For example, maybe `constraint_evaluation_check` procedure should be provided by the specific circuit module.

> With this PR the cycle count for verifying a $2^{20}$  execution trace is 32500  cycles compared to 31915 in `next`.

The difference here (i.e., ~600 cycles) is a bit more than what I expected. I was thinking it would be more like 100 - 200 cycles. Could we check it not against `next` but against the base branch (i.e., `al-refactor-recursive-verifier`)?

Also, not related to this, but what would be the cycle count for verifying a proof of a program with trace length $2^{16}$?

------

In terms of module locations, I wonder if the VM-specific verifier should go under the `sys` module. Then, we'd be able to do something like this:
```
use.std::sys::vm;

begin
    exec.vm::verify_proof
end
```

Separately, I don't remember why we don't include hash of public inputs as an input parameter into the VM verifier. Currently the inputs are:
```
[log(trace_length), num_queries, grinding]
```
But should they not be more like:
```
[PUB_INPUTS_HASH, log(trace_length), num_queries, grinding, ...]
```

stdlib/asm/crypto/miden_vm_verifier/public_inputs.masm
Al-Kindi-0: > It doesn't seems like the information about fixed length public inputs is used anywhere in this file - do we need it?

It is used by the top level `verify` procedure but it makes sense to have it in this module.

> Also, similar comment as in the other file re using constants directly rather than wrapping them in procedures (e.g., `KERNEL_OP_LABEL` is declared and used in this file).

Yes, we can do that.
Changed

 
> Also, maybe this is for a different PR, but would be great to improve the readability of the `reduce_var_len_public_inputs_groups` procedure. For example, I'd move out the storing of the computed kernel digest input into the main `reduce_kernel_digests` procedure so that it looks like:
> 
> ```
> #! Inputs:  [num_kernel_procedures, ...]
> #! Outputs: [prod_acc1, prod_acc0, ...]
> proc.reduce_kernel_digests
>     ...
> end
> ```

I was actually thinking of keeping the handling of each variable-length public input group self contained with an eye at the codegen part. For example, if we had an additional group, then things would look like

```
export.reduce_var_len_public_inputs_groups
    # 1) Reduce kernel procedure digests and store them.
    adv_push.1
    # => [num_ker_procedures, next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
    exec.reduce_kernel_digests
    # => [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]

    # 2) Reduce stack overflow table.
    adv_push.1
    # => [num_entries, next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
    exec.reduce_overflow_table
    # => [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
end
```

stdlib/asm/crypto/stark/public_inputs.masm
Al-Kindi-0: Indeed, in fact this is how it was before but I was thinking of going the opposite way, namely to make all the handling of public inputs part of the generic verifier.
The thinking is that, in most situations, the number of public inputs, both variable and fixed length, is going to be small, hence we can afford the overhead of the `while` for the benefit of making the specific verifier (e.g., `miden_vm_verifier`) smaller.

stdlib/asm/crypto/stark/public_inputs.masm
Al-Kindi-0: Will be decided based on the previous point

stdlib/asm/crypto/stark/verifier.masm
Al-Kindi-0: These are basically used by `stark::utils::constraint_evaluation_check` as they are AIR dependent.
If we move `stark::utils::constraint_evaluation_check` to be part of the generic verifier (e.g., `miden_vm_verifier`) then we would not need to pass these. The reason for not doing it is basically to keep the size of the generic verifier smaller, but we could also just move it here and accept the increase in the size.

stdlib/asm/crypto/miden_vm_verifier/ood_frames.masm
Al-Kindi-0: Done

stdlib/asm/crypto/miden_vm_verifier/helper.masm
Al-Kindi-0: The reason is that I wanted to keep the number of inputs to the generic `verify` procedure manageable i.e., under 16 elements.
Once we settle the question of API to the recursive verifier, we can change it if needed.

stdlib/asm/crypto/miden_vm_verifier/mod.masm
Al-Kindi-0: Done

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: Indeed, that's how I was planning to do it, namely with the new syntax for advice injection. Changed it now.

As for the getters, some of these constants are used in the main `verify` procedure and hence the use of the getters.

stdlib/asm/crypto/miden_vm_verifier/helper.masm
Al-Kindi-0: This encodes the trace info that is part of the proof context i.e., trace widths and number of auxiliary random elements.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: It is doing exactly that, loading from the advice tape while hashing and storing at the pointer. The pointer is fetched using a procedure and is not an input.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Yes, only `load_query_row` is dynamically called by the top `verify` procedure.
Indeed, I updated the code to use `assert_eqw`.
As for the use of temporary addresses, this is because this part of the code is the most critical and hence we would like to keep overheads as little as possible.

stdlib/asm/crypto/miden_vm_verifier/mod.masm
Al-Kindi-0: Indeed, fixed now

Al-Kindi-0:
> In terms of trade-offs, I'm wondering if we should "specialized" a bit more of the code. Specifically, it feels like:
> 
>     * Dealing with public inputs could be a bit more specialized (I left a comment about this inline).
> 
>     * Constraint evaluation could be more specialized as well. For example, maybe `constraint_evaluation_check` procedure should be provided by the specific circuit module.

I answered in the relevant places, but to summarize this will depend on how much code are confortable with being in the specific verifier e.g., miden_vm_verifier.

> The difference here (i.e., ~600 cycles) is a bit more than what I expected. I was thinking it would be more like 100 - 200 cycles. Could we check it not against `next` but against the base branch (i.e., `al-refactor-recursive-verifier`)?

That was my expectation as well. I have double checked and I actually tested against `al-refactor-recursive-verifier`.
Note that we are doing a `dynexec` for each FRI query, in addition to the constant overhead for the overall dynamic execution logic. Though still it seems that there is a discrepancy that I can't account for. Will investigate later.


> Also, not related to this, but what would be the cycle count for verifying a proof of a program with trace length 2 16 ?
> 

For a trace of size $2^{16}$ is costs us $26300$ stack trace rows and $37100$ hasher rows to verify (this is using this current branch)

> In terms of module locations, I wonder if the VM-specific verifier should go under the `sys` module. Then, we'd be able to do something like this:
> 
> ```
> use.std::sys::vm;
> 
> begin
>     exec.vm::verify_proof
> end
> ```

That seems like a good idea. We can do this once we finalize the overall design.

> Separately, I don't remember why we don't include hash of public inputs as an input parameter into the VM verifier. Currently the inputs are:
> 
> ```
> [log(trace_length), num_queries, grinding]
> ```
> 
> But should they not be more like:
> 
> ```
> [PUB_INPUTS_HASH, log(trace_length), num_queries, grinding, ...]
> ```


Indeed and this is part of a bigger question related to the API of recursive verification in the VM. So far, we have glossed over it but we should come to a decision on how things will work from a user perspective.

For example, how should we pass the proof to the verification procedure non-deterministically? How should we pass public inputs? Should it be using a pointer and a hash? ...
Once this is settled we will be able to clean up the interfaces and address many of the remaining above points

stdlib/asm/crypto/miden_vm_verifier/helper.masm
bobbinth: As discussed offline, since the generic `verify` procedure is invoked via `exec`, we are not limited to top 16 elements of the stack and can pass as many elements to it as needed.

stdlib/asm/crypto/miden_vm_verifier/helper.masm
bobbinth: I'd add a comment explaining this.

Al-Kindi-0:
Regarding the overhead we are incurring with the above solution, here is a breakdown of the costs:

1. `process_row_ood_evaluations`: (25253 - 25221) / 2 = 16 per invocation
2. `load_query_row`: (25253-24875)/27 = 14 per query
3. `reduce_var_len_public_inputs_groups`: (25253 - 25245) = 8
4. `load_ace_circuit_description`: 25253 - 25237 = 16
5. `process_dynamic_portion`: 56
6. Preparation before calling the generic verifier: 100

The total cycle count of the above is 590, which withing a couple of percentage points of the overhead we are seeing.
The above suggest that dynamically executing `load_query_row` is incurring a 16 cycles overhead per FRI query, and this is the biggest contributor to the overall overhead.

stdlib/asm/crypto/miden_vm_verifier/helper.masm
Al-Kindi-0: We now pass the digests on the stack

stdlib/asm/crypto/miden_vm_verifier/helper.masm
Al-Kindi-0: Added

Al-Kindi-0:
Went ahead and moved `compute_deep_composition_polynomial_queries` from the generic verifier. The overhead now is under 130 cycles.

The main point that is remaining in my mind is the handling of public inputs which I will address separately.

adr1anh:
Just a quick thought about the API and representing the proof, mainly stemming from previous work on recursive folding verifiers. The goal was to avoid as much non-determinism for safety (and a bit of efficiency).

If we think of a proof for a statement `y <- f(x)`, we'd ideally want our verifier to look like 
`(y, com(proof)) <- verify(preprocess(f), x)`. Optionally, we can return `com(x, y)` which would be computed anyway by the verifier. 

`com(proof)` is the final capacity of the FS sponge, which represents a commitments to all the non-deterministic proof elements. These would be provided as advice. 

`preprocess(f)` contains any relevant information required to verify `f`. Here, `f` could be our VM function executing the state transition from an input stack `x`, ending in the output stack `y`, along with the number of cycles it ran for. It contains the dynamic procedures defined in this PR, including ACE description, trace layout, number of queries etc. 
It doesn't need to be a commitment, and instead could be a pointer to a memory location containing the description. This is especially useful when verifying multiple proofs for the same `f`, since the program performing the verification can
- be initialized with the commitment to `preprocess(f)`
- decommit to the pre-processed data, storing it in memory, ensuring the decommitment only happens once.
- pass this pointer to the `verify` function every time we are verifying a proof for `f`.

The main reason why we would return the output `y` is to take the responsibility away from the caller of loading the output non-deterministically. This only makes sense if `f` actually "computes" something. 

If instead we consider `f(x, y, n)` as a predicate for "verify that running the VM for `n` steps with stack input `x` returns the stack `y`", it would make sense to return a Boolean indicating whether the predicate is true. We lose information however in the case where `y` or `n` is incorrect, since the caller can't check what the _actual_ output was.

In addition, it seems necessary in the future to check that a proof is invalid (for slashing purposes?). In this case, I don't think it makes sense to return any output at all so it could symbolically be wrapped in an `Option<Output>`. We would still return `com(input, output)` so the caller can prove that the proof is not valid for a committed IO.

For the above, our verifier can never fail. That means no assertions, and we need to slightly modify the ACE circuit to match the evaluation with the expected output (zero in the happy path).

There's many other ways we can design the API, so we should consider the different concrete requirements we want.

Al-Kindi-0:
This should be ready for a proper review.
As discussed offline, the logic for handling public inputs is now part of the specific verifier.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
+use.std::crypto::stark::constants
+use.std::crypto::stark::utils
+use.std::crypto::hashes::rpo
+
+# Constraint evaluation circuit description
+adv_map.DIGEST([9139186206676821480, 12763675724578443945, 7621207635344139731, 1122100627503939866])=[
```
bobbinth: nit: I would rename `DIGEST` into something like `CIRCUIT_COMMITMENT`.

Also, we should be able to skip specifying the digest - it would look something like:
```
adv_map.CIRCUIT_COMMITMENT=[
    ...
]
```
In this case, the value for `CIRCUIT_COMMITMENT` should be computed automatically, I believe.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+    1152921505680588801]
+
+# CONSTANTS
+# =================================================================================================
+
+# Number of constraints, both boundary and transitional
+const.NUM_CONSTRAINTS=6
+
+# Number of inputs to the constraint evaluation circuit
+const.NUM_INPUTS_CIRCUIT=244
```
bobbinth: It is probably explained somewhere, but why do we have so many inputs?

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+    # 5) Clean up the stack.
+    drop drop drop
+    # => [...]
+end
+
+#! Loads the description of the ACE circuit for the constraints evaluation check.
+#!
+#! Input: [...]
+#! Output: [...]
+export.load_ace_circuit_description
```
bobbinth: Does this need to be public? (i.e., could this be just `proc`?)

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+    adv.push_mapval
+    exec.constants::get_arithmetic_circuit_ptr
+    padw padw padw
+    repeat.14
+        adv_pipe
+        hperm
+    end
+    exec.rpo::squeeze_digest
+    movup.4 drop
+    assert_eqw
```
bobbinth: Would be great to add an error to the `assert_eqw` instruction. This way, in case something goes wrong here, it would be very easy to detect (error here is not very likely, but it is a good general practice that we've been using in `miden-base`).

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+    # => [...]
+
+    # 3) Set up the inputs to the "eval_circuit" op. Namely:
+    #    1. a pointer to the inputs of the circuit in memory,
+    #    2. the number of inputs to the circuit,
+    #    3. the number of evaluation gates in the circuit.
+    push.NUM_EVAL_GATES_CIRCUIT
+    push.NUM_INPUTS_CIRCUIT
+    exec.constants::public_inputs_address_ptr mem_load
+    # => [...]
```
bobbinth: I think the stat should not be empty after these instructions.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+
+    # 4) Perform the constraint evaluation check by checking that the circuit evaluates to zero, which
+    #    boils down to the "eval_circuit" returning.
+    eval_circuit
```
bobbinth: Could we add the expected stack state after this instruction?

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,195 @@
...
+end
+
+# HELPER PROCEDURES
+# =================================================================================================
+
+#! Executes the constraints evaluation logic based on arithmetic circuit evaluation.
+#!
+#! Input: [...]
+#! Output: [...]
+export.constraint_evaluation_check
```
bobbinth: nit: I would rename this to `check_constraint_evaluations` or something like this (to put the verb upfront).

Also, I would expand the doc comments for this procedure a bit. For example, I would state where the circuit is coming from (it is hard-coded based on the circuit commitment) and where the inputs are coming from (they are expected to be at `public_inputs_address_ptr`, I think).

stdlib/asm/crypto/fri/helper.masm
```diff
@@ -1,6 +1,77 @@
...
+    push.FRI_REMAINDER_CODEWORD_MAX_SIZE_LOG
+end
+
+export.get_fri_remainder_poly_max_degree_plus_1_half
+    push.FRI_REMAINDER_POLY_MAX_DEGREE_PLUS_1_HALF
+end
+
+export.get_fri_remainder_poly_max_degree_plus_1
+    push.FRI_REMAINDER_POLY_MAX_DEGREE_PLUS_1
+end
```
bobbinth: It seems like these constants are used only in this file. If so, we probably don't need to add wrapper procedures for them.

stdlib/asm/crypto/fri/helper.masm
```diff
@@ -1,6 +1,77 @@
...
+        adv_pipe hperm
+    end
+end
+
+#! Loads FRI remainder polynomial of degree strictly less than `max_degree + 1`.
+export.load_fri_remainder_poly_max_degree_plus_1
+    repeat.32
+        adv_pipe hperm
+    end
+end
```
bobbinth: Seems like these two procedures are used only locally. If so, we should change them to `proc`. In general, we should try not to export procedures that we don't need to.

stdlib/asm/crypto/stark/utils.masm
```diff
@@ -123,44 +125,22 @@ export.set_up_auxiliary_inputs_ace
...
-    drop drop drop
+#! Stores digests of dynamically executed procedures.
+#!
+#! Input: [D3, D2, D1, D0, ...]
+#! Output: [C, ...]
+export.store_dynamically_executed_procedures
+    exec.constants::get_dynamic_procedure_3_ptr mem_storew dropw
+    exec.constants::get_dynamic_procedure_2_ptr mem_storew dropw
+    exec.constants::get_dynamic_procedure_1_ptr mem_storew dropw
+    exec.constants::get_dynamic_procedure_0_ptr mem_storew dropw
```
bobbinth: Would it make sense to give specific names to the procedures rather than calling them `dynamic_procedure_x`? 

For example, instead of `get_dynamic_procedure_0_ptr` it could be `get_public_input_processor_ptr` or something like that.

stdlib/asm/crypto/stark/verifier.masm
```diff
@@ -8,50 +8,33 @@ use.std::crypto::stark::public_inputs
...
 #!   - To boost soundness, the protocol is run on a quadratic extension field and this means that
 #!     the OOD evaluation frame is composed of elements in a quadratic extension field i.e. tuples.
 #!     Similarly, elements of the auxiliary trace are quadratic extension field elements. The random
 #!     values for computing random linear combinations are also in this extension field.
-#!   - The following procedure makes use of global memory address beyond 3 * 2^30 and these are
-#!     defined in `constants.masm`.
 #!
-#! Input: [log(trace_length), num_queries, grinding,  ...]
+#! Input: [D3, D2, D1, D0, log(trace_length), num_queries, grinding, num_constraints, trace_info, num_fixed_len_pi, ...]
 #! Output: [...]
```
bobbinth: Are we still using `num_constraints`, `trace_info`, `num_fixed_len_pi` for something?

Also, I would probably describe what `D3`, `D2`, `D1`, `D0` are expected to be.

stdlib/asm/crypto/stark/public_inputs.masm
```diff
@@ -333,29 +69,29 @@ end
 #! the region storing, temporarily, the variable length public inputs that are to be reduced
 #! by the auxiliary randomness and, permanently, the results of the aforementioned reductions.
 #!
-#! Input: [...]
+#! Input: [num_var_len_pi_groups, num_fixed_len_pi, ...]
 #! Output: [...]
 proc.compute_and_store_public_inputs_address
```
bobbinth, line 74: This procedure seems to be used from another module - but it is declared as `proc` rather than `export` - how does this compile?

stdlib/asm/crypto/stark/random_coin.masm
```diff
@@ -2,6 +2,9 @@
 
 use.std::crypto::stark::constants
 use.std::crypto::stark::utils
+use.std::crypto::stark::helpers::random_coin
+use.std::crypto::stark::helpers::public_inputs
```
bobbinth: It doesn't seems like we have `use.std::crypto::stark::helpers` (unless I missed somehow) - but importing them doesn't cause an error?

stdlib/asm/crypto/stark/mod.masm
```diff
@@ -1,27 +1,15 @@
...
-#!    values for computing random linear combinations are also in this extension field.
-#!   - The following procedure makes use of global memory address beyond 3 * 2^30 and these are
-#!    defined in `constants.masm`.
+#!     the OOD evaluation frame is composed of elements in a quadratic extension field i.e. tuples.
+#!     Similarly, elements of the auxiliary trace are quadratic extension field elements. The random
+#!     values for computing random linear combinations are also in this extension field.
 #!
-#! Input: [log(trace_length), num_queries, log(blowup), grinding, ...]
+#! Input: [D3, D2, D1, D0, log(trace_length), num_queries, grinding, num_constraints, trace_info, num_fixed_len_pi, ...]
 #! Output: [...]
```
bobbinth: Similar to one of the previous comments: are we still using `grinding`, `num_constraints`, `trace_info`, `num_fixed_len_pi` for something? And also, I'd document what `D3`, `D2`, `D1`, `D0` are expected to be.

stdlib/asm/crypto/stark/mod.masm
```diff
@@ -1,27 +1,15 @@
 use.std::crypto::stark::verifier
 
-#!   Verify a STARK proof attesting to the correct execution of a program in the Miden VM.
+#!   Verifies a STARK proof.
```
bobbinth: I'd explain a bit more what this procedure does. I think it is more kind of like a "template" for verifying proofs which relies on the caller to supply procedures which will be used to fill int the missing parts.

stdlib/asm/crypto/stark/constants.masm
```diff
@@ -472,6 +434,58 @@ export.current_trace_row_ptr
...
+    push.MAX_CYCLE_LEN_LOG_PTR mem_store
+end
+
+export.get_num_ace_inputs
+    push.NUM_ACE_INPUTS_PTR mem_load
+end
+
+export.get_num_ace_gates
+    push.NUM_ACE_GATES_PTR mem_load
+end
```
bobbinth: It doesn't seems like these procedures are used anywhere. If so, we should probably remove them. Also, let's double-check if there are other such procedures.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,276 @@
...
+    # => [-gz1, -z0, -z1, query_ptr, query_end_ptr, W, query_ptr, ...]
+    swap
+    # => [-z0, -gz1, -z1, query_ptr, query_end_ptr, W, query_ptr, ...]
+    dup exec.constants::get_trace_domain_generator mul
+    # => [-gz0, -z0, -gz1, -z1, query_ptr, query_end_ptr, W, query_ptr, ...]
+    swap
+    # => [-z0, -gz0, -gz1, -z1, query_ptr, query_end_ptr, W, query_ptr, ...]
+    # Save to temporary location `tmp1` for later use when computing the denominators
+    exec.constants::tmp1 mem_storew
+    # => [Y, query_ptr, query_end_ptr, W, query_ptr, ...]                   # (Cycles: 28)
```
bobbinth: It seems like a bit part of this code is just loading values onto the stack and it doesn't seem like it is "computation-specific". If so, maybe we could pre-load the stack to before calling this procedure? Basically, maybe the signature could be something like:
```
#! Inputs:  []
#! Output: [-z0, -gz0, -gz1, -z1, query_ptr, query_end_ptr, W, query_ptr]
```

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,276 @@
...
+        #
+        # Store [poe, index, eval_1, eval_0] where poe := g^index = x / offset and prepare stack
+        # for next iteration.
+
+        ## a) Compute poe
+        ##
+        ## Cycles: 4
+        movup.3 movup.3
+        exec.constants::get_domain_offset_inv mul
+        #=> [poe, index, eval1, eval0, query_ptr, query_end_ptr, W, query_ptr, ...]
```
bobbinth: Similar to the last comment, it seems like this is also generic functionality. Maybe we could encapsulate it in a single procedure - e.g., `deep_queries::process_query_row`.

Basically, I'm trying to reduce the amount of boilerplate code by moving generic pieces into the `std::crypto::stark` modules.

stdlib/asm/crypto/miden_vm_verifier/public_inputs.masm
```diff
@@ -0,0 +1,351 @@
...
+#!
+#! Input: [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
+#! Output:  [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
+export.reduce_var_len_public_inputs_groups
+    # 1) Reduce kernel procedure digests and store them.
+    adv_push.1
+    # => [num_ker_procedures, next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
+    exec.reduce_kernel_digests
+    # => [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
+end
```
bobbinth: Seems like this procedure does little more than just call `reduce_kernel_digests`. Maybe we should get rid of it and just call `reduce_kernel_digests` directly?

bobbinth (commented):
Looks good! Thank you! I left some more comments inline.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: Agreed, changed it now

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: it doesn't, probably a leftover of shuffling things around.
Changed now

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: Added an error message

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: Thanks for catching it, corrected now

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: added

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
Al-Kindi-0: changed
expanded the comment with the suggestions

stdlib/asm/crypto/fri/helper.masm
Al-Kindi-0: Indeed, they are not used anymore after the shuffling 
Removed

stdlib/asm/crypto/fri/helper.masm
Al-Kindi-0: Agreed, changed now

stdlib/asm/crypto/stark/utils.masm
Al-Kindi-0: The decision was based on the desire to avoid very long identifiers for the procedures.
Changed it anyway to be more descriptive

stdlib/asm/crypto/stark/verifier.masm
Al-Kindi-0: Yes, they are used for initializing the seed for F-S transcript
Added a description as requested

stdlib/asm/crypto/stark/public_inputs.masm
Al-Kindi-0: Good catch and indeed it should compile.
Maybe @bitwalker has an idea?

stdlib/asm/crypto/stark/random_coin.masm
Al-Kindi-0: Indeed, the folder was removed but no errors were raised with the above import.
I have replicated this by making the import in `rpo_falcon512.masm`

@bitwalker

stdlib/asm/crypto/stark/mod.masm
Al-Kindi-0: Yes, we are. Added the requested description

stdlib/asm/crypto/stark/mod.masm
Al-Kindi-0: Added

stdlib/asm/crypto/stark/constants.masm
Al-Kindi-0: Removed, as well as other procedures

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Good point! I have factored out that code into a procedure and moved it to the generic verifier.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: totally agree, I have again factored the code out and put it in the generic verifier

stdlib/asm/crypto/miden_vm_verifier/public_inputs.masm
Al-Kindi-0: That is a valid point but the reason for doing it like this is related to the codegen part of this refactoring.
The above structure is aimed at supporting the case when we have more than one group/table of variable length public inputs.  This could look for example like this:

```
export.reduce_var_len_public_inputs_groups
    # 1) Reduce kernel procedure digests and store them.
    adv_push.1
    # => [num_ker_procedures, next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
    exec.reduce_kernel_digests
    # => [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]

    # 2) Reduce stack overflow table store the result.
    adv_push.1
    # => [num_rows, next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
    exec.reduce_overflow_table
    # => [next_var_len_pub_inputs_ptr, var_len_pub_inputs_res_ptr, ...]
end
```

stdlib/asm/crypto/fri/helper.masm
Al-Kindi-0: Note that there are a few constants which are used by the FRI verifier and hence they were not removed

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
+use.std::crypto::stark::constants
+use.std::crypto::stark::deep_queries
+
+#! Computes the DEEP composition polynomial FRI queries.
```
bobbinth: Could we expand the description here a bit? Specifically, since this functions takes no parameters and doesn't return anything, would be great to describe what it does (e.g., "computes X and saves it into some memory location").

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+#!
+#! Inputs:  []
+#! Outputs: []
+export.compute_deep_composition_polynomial_queries
+
+    # Prepare the stack for the loop computing the queries
+    exec.deep_queries::prepare_stack_deep_queries_computation
+
```
bobbinth: Let's add the expected stack state after this procedure is executed.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+    push.1
+    while.true
```
bobbinth: This loop is executed once per query, right? I would add a brief comment explaining this, and also explaining why we do `push.1` (I'm assuming this is because we assume there is at least 1 query).

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+        # Load the (main, aux, constraint)-traces rows associated with the current query and get
+        # the index of the query.
+        exec.load_query_row
+        #=> [Y, X, index, query_ptr, query_end_ptr, W, query_ptr, ...]
+
+        # Compute the current query and store the result
+        # We also re-arrange the stack for the next iteration of the loop
+        exec.deep_queries::compute_deep_query
```
bobbinth: Let's describe the expected stack state after this line.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+    padw swapw padw
+    #=> [Y, Y, 0, 0, 0, 0, ptr, y, y, y]
+
+    exec.load_main_segment_execution_trace
+    #=> [Y, L, C, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
+ 
+    # Load the leaf value we got using `mtree_get` and compare it against the hash we just computed
+    exec.constants::tmp3 mem_loadw
+    assert_eqw
+    #=> [Y, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
```
bobbinth: I kind of like that for the auxiliary trace we have it's own dedicated procedure (`process_aux_segment_execution_trace`) let's do the same for main trace segment and constraint poly.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+    # Load the 80 columns in 10 batches of 8 base field elements
+    padw swapw padw
+    #=> [Y, Y, 0, 0, 0, 0, ptr, y, y, y]
+
+    exec.load_main_segment_execution_trace
+    #=> [Y, L, C, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
+ 
+    # Load the leaf value we got using `mtree_get` and compare it against the hash we just computed
+    exec.constants::tmp3 mem_loadw
+    assert_eqw
```
bobbinth: We should try to attach error messages to these asserts - this would make debugging simpler in the future. Also, the convention we usually use for error messages is to first declare an error constant and then use. Something like this:
```
const.MY_ERROR="something bad happened"

being
    assert_eq.err=MY_ERROR
end
```

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+    #=> [Y, Y, C, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
+
+    # Load the first 4 columns as a batch of 4 quadratic extension field elements.
+    exec.load_aux_segment_execution_trace
+    #=> [Y, D, C, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
+
+    # Load the leaf value we got using `mtree_get` and compare it against the hash we just computed
+    exec.constants::tmp3 mem_loadw
+    assert_eqw
+    #=> [Y, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
```
bobbinth: I think the comment here is inconsistent with the code: as far as I can tell, we are not using `mtree_get` but rather loading data from memory.

stdlib/asm/crypto/miden_vm_verifier/constraints_eval.masm
```diff
@@ -0,0 +1,214 @@
...
+# --- constant getters --------------------------------------------------------
+
+proc.get_num_constraints
+    push.NUM_CONSTRAINTS
+end
+
+# ERRORS
+# =================================================================================================
+
+const.ERR_FAILED_TO_LOAD_CIRCUIT_DESCRIPTION="failed to load the circuit description for the constraints evaluation check"
```
bobbinth: Here is an example of how we usually write errors.

stdlib/asm/crypto/stark/ood_frames.masm
```diff
@@ -113,3 +105,7 @@ export.load_and_horner_eval_ood_frames
     exec.constants::ood_fixed_term_horner_evaluations_ptr add.2 mem_store
     # => [...]
 end
+
+proc.process_row_ood_evaluations
```
bobbinth, line 121: Could we add a comment with expected inputs and outputs for this procedure?

stdlib/asm/crypto/stark/deep_queries.masm
```diff
@@ -288,37 +137,40 @@ export.compute_deep_composition_polynomial_queries
...
+    # => [Y, query_ptr, query_end_ptr, W, query_ptr, ...]
+end
 
-        # II)
-        #
-        # Compute x := offset * domain_gen^index and denominators (x - z) and (x - gz)
+#! Compute the DEEP composition polynomial FRI query at `index`.
+#!
+#! Input: [Y, X, index, query_ptr, query_end_ptr, W, query_ptr, ...]
+#! Output: [?, Y, query_ptr+1, query_end_ptr, ...]
```
bobbinth, line 146: nit: I would replace `?` with something like `has_more_queries`.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
```diff
@@ -0,0 +1,206 @@
...
+use.std::crypto::stark::deep_queries
+
+#! Computes the DEEP composition polynomial FRI queries.
+#!
+#! Inputs:  []
+#! Outputs: []
+export.compute_deep_composition_polynomial_queries
+
+    # Prepare the stack for the loop computing the queries
+    exec.deep_queries::prepare_stack_deep_queries_computation
```
bobbinth: Could this be called by the caller of `compute_deep_composition_polynomial_queries`? This way, we won't have to call it here.

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Added

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Added

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Added

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Done

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Done

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Done

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: The data loaded from memory was obtained using `mtree_get` a few lines above (approx 15 lines), and it was stored temporally in order to liberate the stack for some intermediate manipulations

stdlib/asm/crypto/stark/ood_frames.masm
Al-Kindi-0: Added

stdlib/asm/crypto/stark/deep_queries.masm
Al-Kindi-0: done

stdlib/asm/crypto/miden_vm_verifier/deep_queries.masm
Al-Kindi-0: Good idea
done

stdlib/asm/crypto/stark/ood_frames.masm
```diff
@@ -16,100 +16,103 @@ use.std::crypto::stark::random_coin
...
-#! 2. S_i are the evaluations of columns in the main segment, auxiliary segment and constraint composition
-#!    at the the out-of-domain point z.
-#! 3. T_i are the evaluations of columns in the main segment, auxiliary segment and constraint composition
-#!    at the the out-of-domain point gz.
+#! 2. S_i are the evaluations of columns in the main segment, auxiliary segment and constraint
+#!    composition at the the out-of-domain point z.
+#! 3. T_i are the evaluations of columns in the main segment, auxiliary segment and constraint
+#!    composition at the the out-of-domain point gz.
 #! 4. alpha is the randomness used in order to build the DEEP polynomial.
 export.load_and_horner_eval_ood_frames
```
bobbinth, line 29: Let's add the expected input/output stack state to doc comments here.

stdlib/asm/crypto/stark/ood_frames.masm
```diff
@@ -16,100 +16,103 @@ use.std::crypto::stark::random_coin
...
+    ##    double-word aligned batches.
+    ## Note: the first word is the capacity, where its first element is initialized with the number
+    ## of elements to hash MODULO 8.
 
     ### a) Set up the hasher state
-    padw
-    padw     
-    # => [ZERO, 0, 0, 0, 0, U, Y, ...]
+    padw padw     
+    # => [ZERO, 0, 0, 0, 0, U, Y]
```
bobbinth: Seems like this code makes assumptions about trace width and constraint degrees, and so, is probably not generic enough. We have at least two potential solutions

- Have another dynamic function initialize the hasher state.
- Pass in trace width and constraint degree to figure out how to initialize the capacity in this procedure.
- Maybe something else?

stdlib/asm/crypto/stark/ood_frames.masm
Al-Kindi-0: done

stdlib/asm/crypto/stark/ood_frames.masm
Al-Kindi-0: This is actually just an out-of-place comment. There is nothing as far as I can see that is specific here to the VM verifier.

bobbinth: This means that always implicitly pad the number of columns to the next multiple of 8, right? If so, how does this work on the Rust side? I'm assuming we use `Hasher::hash_elements()` to hash the OOD rows, but if the number of columns is 73, would it apply the regular padding rule?

Al-Kindi-0: We have discussed this a while ago but, at the moment, we just pad the traces to the next multiple of 8 with 0 filled columns. Based on benchmarks, by myself and Phil, this doesn't incur any noticeable performance overhead.

Al-Kindi-0: In the future, we can do this virtually, in order to decrease (slightly) the proof size.

bobbinth: Makes sense! Yeah, I didn't think there would be a noticeable performance impact, except for maybe slightly higher memory requirements (if we hold the extra columns in memory).

Al-Kindi-0: Indeed, forgot memory as well, thanks for bringing it up

Al-Kindi-0:
@bobbinth , regarding this

> In terms of module locations, I wonder if the VM-specific verifier should go under the `sys` module. Then, we'd be able to do something like this:
```
use.std::sys::vm;

begin
    exec.vm::verify_proof
end
```

This is now done in the latest commit

stdlib/asm/crypto/miden_vm_verifier/public_inputs.masm
Al-Kindi-0: I actually went back and adapted your suggestion here in the latest commit

Al-Kindi-0:
@bobbinth , while working through the MASM extension to the ACE backend, I noticed that the `stdlib` verifier didn't handle the case when there are no auxiliary segments.
I fixed this in https://github.com/0xMiden/miden-vm/pull/2008/commits/2abbc4c7dbbb0a753e5bf3102505df8ab01dc151 , let me know what you think about the approach

Al-Kindi-0:
> One thing, for a different PR, may be move the stark module up to the root level.

Makes sense! My preference is to go with the first option, and once we have more than one LDT test (e.g., STIR) we can revisit this solution.

stdlib/asm/sys/mod.masm
```diff
@@ -1,3 +1,4 @@
+
```
adr1anh: is this intensional?

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+
+    # 3) Clean up the stack.
+    drop drop
+    # => [...]
+end
+
+#! Reduces the kernel procedures digests using auxiliary randomness.
+#!
+#! Inputs:  [num_ker_procedures, digests_ptr]
+#! Outputs: [next_ptr]
```
adr1anh, line 216: Isn't the output `[digests_ptr, var_len_pub_inputs_res_ptr, ...]` like in the last line?

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+
+    # Set up the stack for `mem_stream` + `horner_eval_ext`
+    swapw
+    padw padw
+    # => [Y, Y, alpha1, alpha0 + op_label, beta1, beta0,  digests_ptr, beta_ptr, acc1, acc0, ...]
+    # where `Y` is a garbage word.
+
+    exec.constants::tmp1 mem_loadw dup
+    push.0
+    neq
```
adr1anh, line 267: Took me a second to figure this out. Maybe add a comment explaining that we're pushing a flag equivalent to `num_ker_procedures > 1` which is stored in `tmp1`. And the same for the last part of the loop.

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+
+    while.true
+        repeat.1
```
adr1anh, line 274: Why do we need the `repeat.1` instead of just inlining?

adr1anh: I'm guessing this is because of the message length, and that this could be used to reduce other public inputs. Maybe worth explaining?

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+
+        push.0 movdn.6
+        push.0 movdn.6
+        # => [alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, 0, 0, Y, Y, term1', term0', ...]
+ 
+        swapdw
+        # => [Y, Y, alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, 0, 0, term1', term0', ...]
+
+        exec.constants::tmp1 mem_loadw sub.1
+        exec.constants::tmp1 mem_storew
```
adr1anh, line 310: Subtract 1 from `num_ker_digests`

stdlib/asm/sys/vm/public_inputs.masm
adr1anh: Comment: `acc = ∑ {0≤i<4} digest_ptr[i] ⋅ βⁱ`

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+        swapdw
+        # => [alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, acc1, acc0, Y, Y, ...]
+
+        movup.7 movup.7
+        # => [acc1, acc0, alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, Y, Y, ...]
+        
+        dup.5 dup.5
+        # => [beta1, beta0, acc1, acc0, alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, Y, Y, ...]
+        ext2mul
+        # => [tmp1', tmp0', alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, Y, Y, ...]
```
adr1anh: `tmp = β ⋅ acc =  ∑ {0≤i<4} digest_ptr[i] ⋅ βⁱ⁺¹`

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+
+        dup.3 dup.3
+        ext2add
+        # => [term1', term0', alpha1, alpha0 + op_label, beta1, beta0, digests_ptr, beta_ptr, Y, Y, ...]
```
adr1anh: `term = α + op_label + ∑{i = 0..4} digest_ptr[i] ⋅ βⁱ⁺¹`

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+    swapw
+    padw padw
+    # => [Y, Y, alpha1, alpha0 + op_label, beta1, beta0,  digests_ptr, beta_ptr, acc1, acc0, ...]
+    # where `Y` is a garbage word.
+
+    exec.constants::tmp1 mem_loadw dup
+    push.0
+    neq
+
+    while.true
```
adr1anh, line 269: I don't fully get this look. It's clear that we're reducing the digests, but where are they being stored? It feels like we just add them at the end of the stack, no?

adr1anh: Ah, I see below that this is the case. But a simple comment here would be useful.

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+#! the procedure uses the auxiliary randomness in order to reduce the variable-length public
+#! inputs to a single element in the challenge field. The resulting values are then stored
+#! contiguously after the fixed-length public inputs.
+#!
+#! Currently, the only variable-length public inputs are the kernel procedure digests.
+#!
+#! Input: 
+#!      - Operand stack: [...]
+#!      - Advice stack: [beta0, beta1, alpha0, alpha1, var_len_pi_1_len, ..., var_len_pi_k_len, ...]
+#! Output: [D, ...]
```
adr1anh: What is `D`?

stdlib/asm/sys/vm/public_inputs.masm
adr1anh: I think `digests_ptr` has been advanced to `next_ptr` but what is that?

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+        # => [loop, n - 1, acc1', acc0', term1_1, term1_0, ..., termn_1, termn_0, ...]
+    end
+
+    drop
+    exec.constants::tmp2 mem_load movdn.2
+    # since we are initializing the bus with "requests", we should invert the reduced result
+    ext2inv
+    # => [prod_acc1, prod_acc0, digests_ptr, ...]
+
+    # Store the result
```
adr1anh: at which address?

stdlib/asm/sys/vm/public_inputs.masm
```diff
@@ -0,0 +1,344 @@
...
+    # 1) Compute the address where the public inputs will be stored and store it.
+    #    This also computes the address where the reduced variable-length public inputs will be stored.
+    exec.get_num_fixed_len_public_inputs push.NUM_VAR_LEN_PI_GROUPS
+    exec.public_inputs::compute_and_store_public_inputs_address
+    # => [C, ...]
+
+    # 2) Load the public inputs.
+    #    This will also hash them so that we can absorb them in the transcript.
+    exec.load_public_inputs
+    # => [D, ...]
```
adr1anh: Maybe mention we pop them (`a_i`) from the advice?

adr1anh (commented):
First batch of comment relating to the public inputs. There's a lot of redundancies so feel free to resolve them and reduce clutter. 

My mindset is that this is code that will eventually get audited, so it should be as clear as possible. Even when it is possible to deduce what is going on from the code, small hints can drastically reduce the amount of time required to understand what is going on.

stdlib/asm/sys/vm/deep_queries.masm
```diff
@@ -0,0 +1,252 @@
...
+    exec.constants::tmp3 mem_storew
+    adv.push_mapval
+    #=>[LEAF_VALUE, MAIN_TRACE_TREE_ROOT, depth, index, query_ptr]
+
+    exec.constants::tmp2 mem_loadw
+    swapw
+    #=>[LEAF_VALUE, ptr_x, ptr_alpha_inv, acc1, acc0, depth, index, query_ptr]
+
+    # Load the values of the main segment of the execution trace at the current query. We also
+    # compute their hashing and the value of their random linear combination using powers of a
```
adr1anh:
```suggestion
    # compute their hash and the value of their random linear combination using powers of a
```

stdlib/asm/crypto/stark/ood_frames.masm
```diff
@@ -16,100 +16,104 @@ use.std::crypto::stark::random_coin
...
-#! 3. T_i are the evaluations of columns in the main segment, auxiliary segment and constraint composition
-#!    at the the out-of-domain point gz.
+#! 2. S_i are the evaluations of columns in the main segment, auxiliary segment and constraint
+#!    composition at the the out-of-domain point z.
+#! 3. T_i are the evaluations of columns in the main segment, auxiliary segment and constraint
+#!    composition at the the out-of-domain point gz.
 #! 4. alpha is the randomness used in order to build the DEEP polynomial.
+#!
+#! Inputs:  []
+#! Outputs: []
```
adr1anh, line 28: It's not really clear "where" this computation happens. Could we mention where in memory we store the results?

stdlib/asm/crypto/stark/ood_frames.masm
```diff
@@ -16,100 +16,104 @@ use.std::crypto::stark::random_coin
...
-    # I) Load the random challenge used in computing the DEEP polynomial.
-    #    We use this challenge to compute the constant terms needed in the computation of the DEEP queries.
-    #    Although this challenge is generated only after all the OOD evaluations are received by the verifier,
-    #    we use non-determinism to generate it before doing so. This is done so that we can hash, memory store
-    #    and Horner evaluate in parallel.
+    # --- Load the random challenge used in computing the DEEP polynomial -----
+
+    # We use this challenge to compute the constant terms needed in the computation of the DEEP
+    # queries. Although this challenge is generated only after all the OOD evaluations are received
+    # by the verifier, we use non-determinism to generate it before doing so. This is done so that
```
adr1anh, line 34: Mention where we do this check

stdlib/asm/crypto/stark/utils.masm
```diff
@@ -123,44 +125,22 @@ export.set_up_auxiliary_inputs_ace
...
+proc.execute_constraint_evaluation_check
+    exec.constants::get_procedure_digest_execute_constraint_evaluation_check_ptr dynexec
+end
 
-    # 5) Clean up the stack.
-    drop drop drop
+#! Stores digests of dynamically executed procedures.
+#!
+#! Input: [D3, D2, D1, D0, ...]
+#! Output: [C, ...]
```
adr1anh: Isn't this empty?

adr1anh (commented):
The other portions look really good, again just left a couple of nits around the comments.

stdlib/asm/sys/mod.masm
Al-Kindi-0: It feels to be congruent with the rule applied when having imports, but no hard opinions on this

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Indeed, this is now pointing to the next group of variable length public inputs, if such a group exists.
Updated the comment with this

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Yes, it is testing `num_ker_procedures > 0` to decide whether or not to enter the loop
Added a comment about it

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Exactly, here we could remove `repeat.1` but it is there for "templating" purposes.
Added a comment there as well as your comment

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Makes sense
Added the comment

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Copy-paste error, thanks for catching it!

stdlib/asm/sys/vm/deep_queries.masm
Al-Kindi-0: Fixed

stdlib/asm/crypto/stark/ood_frames.masm
Al-Kindi-0: Added a comment about where and how it is saved in memory

stdlib/asm/crypto/stark/ood_frames.masm
Al-Kindi-0: Added a comment about this

stdlib/asm/crypto/stark/utils.masm
Al-Kindi-0: Copy-paste typo again, thank you for catching it

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: At the address at the beginning of the region storing the group we just reduced.
Added a comment about this, so let me know if it is unclear or vague

stdlib/asm/sys/vm/public_inputs.masm
Al-Kindi-0: Done

Al-Kindi-0:
@adr1anh , thank you for the feedback, which I have implemented almost all of it

> My mindset is that this is code that will eventually get audited, so it should be as clear as possible. Even when it is possible to deduce what is going on from the code, small hints can drastically reduce the amount of time required to understand what is going on. 

Totally agree, and we will for sure have additional future PRs where we can make things even more clearer

bobbinth:
@Al-Kindi-0 - there seems to be a merge conflict. Once we resolve it, I think we can merge.

PR #2009, opened 2025-07-19
chore: add AdviceInputs::map() accessor

bobbinth:
This is a simpler version of https://github.com/0xMiden/miden-vm/pull/1999 to address https://github.com/0xMiden/miden-vm/issues/1995. The changes include:

- Added `AdviceInputs::map()` accessor.
- Made all fields in `AdviceInputs` private (under all features).
- Added `AdviceInputs::into_parts()` destructor.

### DIFF
--- a/processor/src/host/advice/inputs.rs
+++ b/processor/src/host/advice/inputs.rs
@@ -1,4 +1,4 @@
-use alloc::{sync::Arc, vec::Vec};
+use alloc::vec::Vec;
 
 use miden_core::{
     AdviceMap, Felt, Word,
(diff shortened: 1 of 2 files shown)


processor/src/host/advice/inputs.rs
```diff
@@ -109,20 +108,28 @@ impl AdviceInputs {
...
+        &self.map
     }
 
-    /// Returns the underlying [MerkleStore].
+    /// Returns the underlying [MerkleStore] of this advice inputs.
     pub const fn merkle_store(&self) -> &MerkleStore {
         &self.store
     }
+
+    // DESTRUCTORS
```
plafer: nit: I wouldn't call this a destructor as it's not a destructor in the conventional sense (i.e. the `Drop` trait). Since we only have one `into_parts()` maybe it can be merged in another section?

processor/src/host/advice/inputs.rs
```diff
@@ -21,12 +21,11 @@ use miden_core::{
...
 ///    with Merkle trees.
-#[cfg(not(feature = "testing"))]
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct AdviceInputs {
-    pub(super) stack: Vec<Felt>,
-    pub(super) map: AdviceMap,
-    pub(super) store: MerkleStore,
+    stack: Vec<Felt>,
+    map: AdviceMap,
+    store: MerkleStore,
```
plafer: Looking at the `AdviceInputs` methods, no method upholds any invariant of the struct, so is there a reason not to just make the fields public, and perhaps provide a few convenience constructors?

bobbinth: Hmmm - good point! Made the fields public and removed redundant accessor methods.

PR #2015, opened 2025-07-22
Tracking PR for enabling async functionality in the `FastProcessor`

bobbinth:
After #2003 got merged, we have two main tasks to accomplish on this front:

- Address https://github.com/0xMiden/miden-vm/issues/1942.
- Add conditional compilation along the lines of what is described here https://github.com/0xMiden/miden-vm/issues/1993#issuecomment-3082701577.
  - As a part of this, we should also figure out what bounds we need for `std` vs. `no_std` options.


processor/src/host/mod.rs
```diff
@@ -91,20 +119,34 @@ pub trait AsyncHost: BaseHost {
...
+pub trait AsyncHostFuture<O>: Future<Output = O> {}
+
+#[cfg(not(feature = "std"))]
+impl<T, O> AsyncHostFuture<O> for T where T: Future<Output = O> {}
+
+#[cfg(feature = "std")]
+pub trait AsyncHostFuture<O>: Future<Output = O> + Send {}
+
+#[cfg(feature = "std")]
+impl<T, O> AsyncHostFuture<O> for T where T: Future<Output = O> + Send {}
```
bobbinth: Before merging into `next`, we should revisit if we want to keep this as is or change the conditional compilation to be base on the target architecture.

plafer:
@PhilippGackstatter Are we ready to review & merge this now? I assume this ended up resolving the issues in downstream crates?

PhilippGackstatter:
> Are we ready to review & merge this now?

I think so, but I think @drahnr might have a better idea of the current state of async downstream, as he was working on https://github.com/0xMiden/miden-base/pull/1666 recently.

bobbinth:
I pushed a couple of small changes:

- Moved `get_mast_forest()` method into `BaseHost` since it is now sync. Once we make it async again, we can move it back to relevant hosts.
- Switched to using `#[cfg(not(target_family = "wasm"))]` feature instead of `no_std`.

On the last point, I still think we need to do something more explicit throughout the codebase to make it clear that, in our case WASM and `no_std` are basically the same. Even our `build-no-std` build target looks like so:
```
.PHONY: build-no-std
build-no-std: ## Builds without the standard library
	cargo build --no-default-features --target wasm32-unknown-unknown --workspace
```

Setting this aside though, would be great to have another set of eyes look through this. If all looks good, I'll merge this into `next` and will release the v0.17.0 VM tomorrow (Wednesday).

Since this branch has some breaking changes now as compared to a few hours ago, we'll also need to update https://github.com/0xMiden/miden-base/pull/1636. @PhilippGackstatter could you look into it and confirm that everything is still working there?

processor/src/host/mod.rs
```diff
@@ -91,20 +119,30 @@ pub trait AsyncHost: BaseHost {
...
-        process: &mut ProcessState<'_>,
+        process: &ProcessState<'_>,
         event_id: u32,
-    ) -> impl Future<Output = Result<(), EventError>> + Send;
+    ) -> impl FutureMaybeSend<Result<Vec<AdviceMutation>, EventError>>;
 }
+
+/// Alias for a `Future`
+///
+/// If feature `std` is enabled, we add `Send` to the required bounds, otherwise we do not. This
```
plafer: this still mentions the `std` feature - should be updated to `wasm` target family.

Should we also mention that this is automatically implemented for all `Future`s, similar to how the standard library [does for `Into`](https://doc.rust-lang.org/std/convert/trait.Into.html) (with respect to the comment about a blanket implementation for all types that implement `From`)?

processor/src/host/mod.rs
```diff
@@ -91,20 +119,30 @@ pub trait AsyncHost: BaseHost {
...
+/// impacts usability with a multithreaded executor.
+#[cfg(target_family = "wasm")]
+pub trait FutureMaybeSend<O>: Future<Output = O> {}
+
+#[cfg(target_family = "wasm")]
+impl<T, O> FutureMaybeSend<O> for T where T: Future<Output = O> {}
+
+#[cfg(not(target_family = "wasm"))]
+pub trait FutureMaybeSend<O>: Future<Output = O> + Send {}
```
plafer, line 153: there's no docs when `not(target_family = "wasm"))`, so I think the standard generated docs wouldn't have a docstring for this trait? Therefore we should copy over the docstring here

plafer (approved):
New changes LGTM (modulo a nit)!

drahnr:
> > Are we ready to review & merge this now?
> 
> I think so, but I think @drahnr might have a better idea of the current state of async downstream, as he was working on [0xMiden/miden-base#1666](https://github.com/0xMiden/miden-base/pull/1666) recently.

We are in a better place, everything compiles, `SourceManager`-refactors are a thing for follow up PRs. I'd hold off until we have `pgackst-async-experiemental` ready to merge as well before going ahead and merging. 

Remaining PRs to merge:
* https://github.com/0xMiden/miden-base/pull/1693
* https://github.com/0xMiden/miden-base/pull/1699

processor/src/host/mod.rs
bobbinth: done!

processor/src/host/mod.rs
bobbinth: fixed!

PR #1917, opened 2025-06-26
docs: clarify and unify documentation for hash, hperm, and hmerge operations

MozirDmitriy:
Fix #1730 

## Describe your changes

- Added detailed explanations of the differences between `hash`, `hperm`, and `hmerge` in both user and technical documentation.
- Clarified that `hmerge` is intended for merging two digests and correctness is only guaranteed for digest inputs.
- Explained the concepts of rate and capacity, including their initialization rules.
- Documented the procedure for extracting the final digest after `hperm` (using `squeeze_digest`).
- Provided Rust equivalents for each operation.
- All changes are strictly based on the codebase and requirements from issue #1730, with no speculative or non-existent information added.

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

Issue #1730, opened 2025-03-28
Document differences of hashing operations


PhilippGackstatter:
The existing docs on [hashing](https://0xpolygonmiden.github.io/miden-docs/imported/miden-vm/src/user_docs/assembly/cryptographic_operations.html#hashing-and-merkle-trees) assume pre-existing knowledge which would be good to add. Specifically it would be great if the following could be added: 
- Explain the differences between `hash`, `hperm` or `hmerge`.
- Clarify whether `hmerge` should only be used on two digests or if it can also be used on non-digests.
- Explain what the "rate" and "capacity" words are and how to initialize them.
- Explain how to get the final digest after an `hperm` (e.g. manually or using `rpo::squeeze_digest`) .
- State the equivalent operation in Rust (e.g. `hmerge` -> `miden_crypto::hash::rpo::Rpo256::merge`) and/or add examples.

bobbinth:
Closed by #2016.

PR #2016, opened 2025-07-22
Iimprove `hperm` documentation

bobbinth:
This PR addresses #1730 and supersedes #1917.

### DIFF
--- a/docs/src/user_docs/assembly/cryptographic_operations.md
+++ b/docs/src/user_docs/assembly/cryptographic_operations.md
@@ -20,3 +20,55 @@ mtree_verify.err=123
 mtree_verify.err=MY_CONSTANT
 ```
 If the error code is omitted, the default value of $0$ is assumed.
+
+#### Differences between `hash`, `hperm`, and `hmerge`
+
+- **hash**: 1-to-1 hashing, takes 4 elements (1 word), applies the RPO permutation to it, and returns a 4-element digest.
+- **hmerge**: 2-to-1 hashing, takes 8 elements (2 words), applies the RPO permutation to them, and returns a 4-element digest. This is frequently used to hash two digests (e.g., for Merkle trees), but could also be used to hash an arbitrary sequence of 8 elements. This is equivalent to `miden_crypto::hash::rpo::Rpo256::merge()` function.
+- **hperm**: Applies the RPO permutation to 12 stack elements (8 rate + 4 capacity), returns all 12 elements (the full sponge state). Used for intermediate operations or manual sponge state management. This is equivalent to `miden_crypto::hash::rpo::Rpo256::apply_permutation()` Rust function.
+
+#### `hperm` operation semantics
+
+As mentioned above, `hperm` instruction applies a single RPO permutation to the top 12 stack elements. This allows us to manually define the state of the sponge and incrementally absorb data into it. The sponge state consists of two parts:
+
+- `RATE` - two words specifying the data to be absorbed into the sponge.
+- `CAPACITY` - a single word that maintains the state of the sponge across permutations.
+
+The `hperm` instruction expects the sponge to be on the stacks as follows:
+
+```
+[RATE1, RATE0, CAPACITY]
+```
+Note that this is in the "stack-order" - i.e., the first element of the sponge is the deepest in the stack.
+
+According to the RPO specifications, the capacity should be initialized as follows:
+
+- The first capacity element (i.e., `stack[11]`) should be initialized to $n \mod 8$ where $n$ is the number of elements to be hashed across all permutations. For example, if we need to hash $100$ elements, the first capacity element must be initialized to $4$.
+- All other capacity elements must be initialized to zeros.
+
+The above rule is convenient as when we need to hash the number of elements divisible by $8$, all capacity elements should be initialized to zeros.
+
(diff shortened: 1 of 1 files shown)


docs/src/user_docs/assembly/cryptographic_operations.md
```diff
@@ -19,4 +19,56 @@ The `mtree_verify` instruction can also be parametrized with an error code which
...
+
+Once the capacity elements have been initialized, we need to put the data we'd like to hash onto the stack. As mentioned above, this is done by populating the `RATE` section of the sponge. The sponge can absorb exactly $8$ elements per permutation. Thus, if we have fewer than $8$ elements to absorb, we need to put our data onto the stack toward the "front" of the sponge (i.e., the deeper portion) and then pad the remaining elements with zeros. If we have more than $8$ to absorb, we'll need to iteratively load the rate portion of the sponge with $8$ elements, execute `hperm`, load the next $8$ elements, execute `hperm`, etc. - until all data has been absorbed.
+
+Once all the data has been absorbed, we can "squeeze" the resulting hash out of the sponge state by taking the first rate word (i.e., `RATE0`). To do this, we can use a convenience procedure from the standard library: `std::crypto::hashes::rpo::squeeze_digest`.
+
+For efficient hashing of long sequences of elements, `hperm` instruction can be paired up with `mem_stream` and `adv_pipe` instructions. For example, the following, will absorb 24 elements from memory and compute their hash:
```
Al-Kindi-0: `mem_stream` and `adv_pipe` -> `mem_stream` or `adv_pipe`

docs/src/user_docs/assembly/cryptographic_operations.md
```diff
@@ -19,4 +19,56 @@ The `mtree_verify` instruction can also be parametrized with an error code which
...
 ```
-If the error code is omitted, the default value of $0$ is assumed.
\ No newline at end of file
+If the error code is omitted, the default value of $0$ is assumed.
+
+#### Differences between `hash`, `hperm`, and `hmerge`
+
+- **hash**: 1-to-1 hashing, takes 4 elements (1 word), applies the RPO permutation to it, and returns a 4-element digest.
+- **hmerge**: 2-to-1 hashing, takes 8 elements (2 words), applies the RPO permutation to them, and returns a 4-element digest. This is frequently used to hash two digests (e.g., for Merkle trees), but could also be used to hash an arbitrary sequence of 8 elements. This is equivalent to `miden_crypto::hash::rpo::Rpo256::merge()` function.
+- **hperm**: Applies the RPO permutation to 12 stack elements (8 rate + 4 capacity), returns all 12 elements (the full sponge state). Used for intermediate operations or manual sponge state management. This is equivalent to `miden_crypto::hash::rpo::Rpo256::apply_permutation()` Rust function.
```
PhilippGackstatter: Should we mention the equivalent operation for `hash` as well?

docs/src/user_docs/assembly/cryptographic_operations.md
```diff
@@ -19,4 +19,56 @@ The `mtree_verify` instruction can also be parametrized with an error code which
...
+
+- The first capacity element (i.e., `stack[11]`) should be initialized to $n \mod 8$ where $n$ is the number of elements to be hashed across all permutations. For example, if we need to hash $100$ elements, the first capacity element must be initialized to $4$.
+- All other capacity elements must be initialized to zeros.
+
+The above rule is convenient as when we need to hash the number of elements divisible by $8$, all capacity elements should be initialized to zeros.
+
+Once the capacity elements have been initialized, we need to put the data we'd like to hash onto the stack. As mentioned above, this is done by populating the `RATE` section of the sponge. The sponge can absorb exactly $8$ elements per permutation. Thus, if we have fewer than $8$ elements to absorb, we need to put our data onto the stack toward the "front" of the sponge (i.e., the deeper portion) and then pad the remaining elements with zeros. If we have more than $8$ to absorb, we'll need to iteratively load the rate portion of the sponge with $8$ elements, execute `hperm`, load the next $8$ elements, execute `hperm`, etc. - until all data has been absorbed.
```
PhilippGackstatter: > - The first capacity element (i.e., `stack[11]`) should be initialized to $n \mod 8$ where $n$ is the number of elements to be hashed across all permutations. For example, if we need to hash $100$ elements, the first capacity element must be initialized to $4$.

> Thus, if we have fewer than $8$ elements to absorb, we need to put our data onto the stack toward the "front" of the sponge (i.e., the deeper portion) and then pad the remaining elements with zeros.

Just for my understanding. Is this basically saying that when I want to hash 3 felts felt0, felt1, felt2, then I have to do:

```
[pad(5), felt0, felt1, felt2, [0, 0, 3, 0]]
```

So the felts are arranged to start at the deepest part of the stack and extend to the shallower part, and not the other way around. Or am I missing the point?

Would it be problematic/incorrect/dangerous to do this?

```
[felt0, felt1, felt2, pad(5), [0, 0, 3, 0]]
```

In any case, it would be cool to add an example of how to do a permutation with a number of elements that is not cleanly divisible by 8.

PhilippGackstatter (approved):
Looks good to me! Thanks a lot for adding this. It clarifies a lot for me.

docs/src/user_docs/assembly/cryptographic_operations.md
bobbinth: Fixed!

docs/src/user_docs/assembly/cryptographic_operations.md
bobbinth: Done!

docs/src/user_docs/assembly/cryptographic_operations.md
bobbinth: This is how it would look like:
```
[pad(5), felt0, felt1, felt2, [0, 0, 0, 3]]
```
I've added the example - great suggestion!

bobbinth:
> I think it would be good to add an explanation, maybe with an example, to link the three operations `hperm`, `hmerge` and `hash`. Basically show how `hmerge` and `hash` can be recovered from `hperm`

Great idea! I've added the exact description of how `hash` and `hmerge` are implemented in the VM.

PR #2023, opened 2025-07-22
feat: Introduce `FastProcessor` checkpoints

### LINKED ISSUE #2022: `FastProcessor::execute` should return a list of checkpoints
This is part of the parallel execution line of work #1558.

This list of "checkpoints" is all the information that is needed by a trace generator to generate a trace. This is to be used by the `TransactionProver` in miden-base, as well as by the "trace fragment generators" in #1839.

plafer:
Partially closes #2022 (i.e. only the core trace checkpoints are generated)

Note that I didn't split the `FastProcessor`'s `mod.rs` file just yet, since this will cause pretty bad conflicts when merging the `async-experimental` branch. I would prefer to wait for that before we split the file up.

One of the challenges with this PR was in accommodating the 2 modes that we want the `FastProcessor` to run in: 
1. no trace states (i.e. run as fast as possible), and 
2. generating trace states for future parallel trace generation.

Mode 1 requires storing a subset of what is needed for mode 2. For example, for the overflow stack, mode 1 requires just the overflowed stack elements, whereas mode 2 requires the extra features provided by `OverflowTable`. Similarly, for execution context management, mode 1 requires the lightweight `ExecutionContextInfo` (defined in `FastProcessor`'s `mod.rs`), whereas mode 2 requires the beefier `BlockStack`. In order for mode 1 to indeed be as fast as it can, I decided to include both data structures. That is, in mode 1, we use the fastest version; however in mode 2, we use both versions "simultaneously" (and hence doing some duplicate work). I found that to be the best of both worlds. Note that the data structures needed for mode 2 are all stored in `CoreTraceStateBuilder`.

Finally, `FastProcessor::execute_op_batch()` is quite complex now due to the 2 simultaneous definitions of "operation batch index":
- the decorators still use the old definition that doesn't account for inserted NOOPs
- the `NodeExecutionPhase::BasicBlock` variant (stored in `CoreTraceState`) uses the new definition that takes the inserted NOOPs into account

This should be naturally cleaned up with #1815.


plafer:
@bobbinth this is ready for a high-level review. I'm still debugging the end-to-end tests with parallel trace generation, which provides interesting insights for this PR. My plan is to
1. Make all end-to-end trace generation tests pass (going well so far)
2. Cleanup this PR in light of any learnings, and add documentation

Specifically there are interesting edge cases & potential off-by-one errors in the interface between the `FastProcessor` and trace generators that need to be properly documented. But you can still look into it to make sure we're aligned on the high-level direction, and highlight if anything is missing.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+    block_stack: BlockStack,
+}
+
+/// Builder for recording the core trace state of the processor during execution.
+#[derive(Debug, Default)]
+pub struct CoreTraceStateBuilder {
+    // State that gets snapshotted at the start of each trace fragment
+    // TODO(plafer): Do we want to store this in a separate struct?
+    pub overflow: OverflowTable,
+    pub block_stack: BlockStack,
```
bobbinth: Cloning these every 1024 cycles may get quite expensive (especially if the MAST tree or overflow tables are deep). Not sure if these will work, but here are potential alternatives:

- For the overflow table, we could probably keep a "replay" of values being moved back onto the stack top from the overflow table (similar to how we have memory, advice etc. replays).
- For the block stack table, maybe we can keep a "transcript" of executed nodes - something similar to `NodeExecutionPhase` but with hasher addresses added to the internal variants.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    // State that gets snapshotted at the end of each trace fragment
+    pub hasher: HasherChipletShim,
+    pub memory: MemoryReplay,
+    pub advice: AdviceReplay,
+    pub external: ExternalNodeReplay,
+
+    // State stored at the start of a trace fragment
+    snapshot_start: Option<SnapshotStart>,
```
bobbinth: It feels like there is some redundancy between the data in `snapshot_start` and `block_stack` (especially between `block_stack` and `continuation_stack`) - though, I don't know yet what's the best way to reconcile them.

Also question: why is `snapshot_start` optional? i.e., under what circumstances it is `None`?

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    // Output
+    core_trace_states: Vec<CoreTraceState>,
```
bobbinth: Looking at the current code, it is not clear why this needs to be a field rather than just a local variable in `into_core_trace_states()` function.

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,464 @@
...
+        node_id: MastNodeId,
+        /// Index of the operation batch within the basic block
+        batch_index: usize,
+        /// Index of the operation within the batch
+        op_idx_in_batch: usize,
+        /// Whether a RESPAN operation needs to be added before executing this batch. When true,
+        /// `batch_index` refers to the batch to be executed *after* the RESPAN operation, and
+        /// `op_index_in_batch` MUST be set to 0.
+        needs_respan: bool,
+    },
```
bobbinth: In theory, we should be able to reduce this just to a tuple of values `(node_id, op_index)` where `op_index` is the first operation in this block to start executing from. But this would be much easier to implement after the batch execution refactoring.

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,464 @@
...
+    ContextId, continuation_stack::ContinuationStack, decoder::block_stack::BlockStack,
+    stack::OverflowTable,
+};
+
+// NODE EXECUTION PHASE
+// ================================================================================================
+
+/// Specifies the execution phase when starting fragment generation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum NodeExecutionPhase {
```
bobbinth: As mentioned in one of the previous comments, this feels very similar to the `Continuation` enum and maybe there is a way to combine them.

bobbinth (commented):
Thank you! Looks good! I took a very high-level look (and still not fully understanding everything), but overall, the approach should work. I left a few small comments inline.

I think the main thing is not too clear to me yet is whether the approach we've taken to record the execution of the program itself is optimal. For example, could we have something like `Vec<(RowIndex, Continuation)>` in `CoreTraceStateBuilder` instead of `BlockStack`? We'll probably need to track a bit more info with it, but it should be sufficient to create start/end rows for MAST nodes in the trace.

processor/src/fast/trace_state_builder.rs
plafer: Looking at `finish_current_snapshot()`, it shows why this needs to be a field. Visually at a high level, the `CoreTraceStateBuilder` works as:

```
 0                                 1024
|                                 | 
 ^                                 ^
start snapshot 0           finish snapshot 0, start snapshot 1        
```
The key insight is that for snapshot *i*, some information needs to be stored at the beginning (everything in `SnapshotStart`), and some at the end (e.g. the replays). We finish a snapshot by `mem::take()`ing the data out of the replays - e.g. everything that was recorded during the first 1024 cycles goes into snapshot 0. We reset the replays though, otherwise snapshot 1 would have all the replay data from snapshot 0.

To have `core_trace_states` be only a variable in `into_core_trace_states()`, we'd need then to store the index for each snapshot, in each replay - which is clearly more complex than a simple `mem::take()` to clear the replays on `finish_current_snapshot()` directly.

processor/src/fast/checkpoints.rs
plafer: We could combine `batch_index` and `op_idx_in_batch` into `op_index` (to the benefit of the fast processor after the refactoring), but I think `needs_respan` (or equivalent) needs to stay, because otherwise we can't distinguish between the 2 states
1. the first operation in batch *i*
2. the RESPAN right before the first operation in batch *i*

processor/src/fast/checkpoints.rs
plafer: They're different in a subtle way. `ContinuationStack` works at the granularity of `MastNodeId`s  in 2 states: start or finish. It encodes what comes after the *current* node (not explicitly encoded in the `ContinuationStack`). `NodeExecutionPhase` is there to encode what the `ContinuationStack` doesn't: exactly where we're at in the execution of the current node.

The reason we don't want to merge them is that we only want to create a `NodeExecutionPhase` at a fragment boundary - otherwise, for example, we'd be creating a new `NodeExecutionPhase` at each operation. Also they're different at a fundamental level, where `NodeExecutionPhase` doesn't encode a *continuation* - it encodes the "current state".

`CoreTraceFragmentGenerator::execute_fragment_generation()` in #1839 can provide more intuition for how the `NodeExecutionPhase` is used: we basically start a fragment by finishing the *current* node, and then keep going by looping over the `ContinuationStack` (in a way similar to the fast processor).

plafer: Maybe `NodeExecutionState` would be a better name

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Creates a new StackState with the provided parameters.
+    ///
+    /// `stack_top` should be the top 16 elements of the stack stored in reverse order, i.e.,
+    /// `stack_top[15]` is the topmost element (s0), and `stack_top[0]` is the bottommost element
+    /// (s15).
+    pub fn new(stack_top: [Felt; MIN_STACK_DEPTH], overflow: OverflowTable) -> Self {
+        Self { stack_top, overflow }
+    }
+
+    pub fn stack_top(&self) -> &[Felt; MIN_STACK_DEPTH] {
```
adr1anh: Could we make `stack_top` public instead?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,470 @@
...
+        &self.stack_top
+    }
+
+    pub fn stack_top_mut(&mut self) -> &mut [Felt; MIN_STACK_DEPTH] {
+        &mut self.stack_top
+    }
+
+    /// Derives the stack depth (b0 helper column) from the overflow table
+    pub fn stack_depth(&self) -> Felt {
```
adr1anh: Can we return an integer and let the caller convert to `Felt` if necessary?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,450 @@
...
+    /// Recorded hasher operations from Merkle path verification operations
+    /// Each entry contains (address, computed_root)
+    pub build_merkle_root_operations: VecDeque<(Felt, Word)>,
+
+    /// Recorded hasher operations from Merkle root update operations
+    /// Each entry contains (address, old_root, new_root)
+    pub mrupdate_operations: VecDeque<(Felt, Word, Word)>,
+}
+
+impl Default for HasherReplay {
```
adr1anh: Can we derive Default instead?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,450 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl HasherReplay {
+    pub fn new() -> Self {
```
adr1anh: Is this necessary with `Default`? And does it make sense to have `_with_capacity()` method?

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl CoreTraceStateBuilder {
+    /// Extracts the internal state out in order to create a `CoreTraceState`, and stores it
+    /// internally.
+    ///
+    /// The replay data is cleared after extraction, since each trace state is expected to be empty
+    /// at the beginning of a new trace fragment. The overflow table and block stack are not cleared
+    /// however, since they track the state of the computation since the beginning.
+    pub fn extract_new_state(
```
adr1anh: The name is a bit confusing as at first glance, I'd expect `extract` to return the new state. Perhaps  something like `begin_checkpoint` or `capture_trace_state`?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,450 @@
...
+
+impl Default for AdviceReplay {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl AdviceReplay {
+    /// Creates a new AdviceReplay with empty operation vectors
+    pub fn new() -> Self {
```
adr1anh: Could we replace with calls to `default`?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,450 @@
...
+
+impl Default for MemoryReplay {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl MemoryReplay {
+    /// Creates a new MemoryReplay with empty read vectors
+    pub fn new() -> Self {
```
adr1anh: Same comment about deriving `Default` and calling that instead

plafer:
This is now in a reviewable state - I will stop force pushing, but intend to clean it up further. Some things to be cleaned up still (which I'd gladly take suggestions for):
- reorganize `CoreTraceStateBuilder` (hasn't been cleaned up yet)
- naming
- docs
- maybe a bit of moving things around

@adr1anh since your last review, I removed `OverflowTable` and `BlockStack` from `CoreTraceState` in favor of dedicated replays (in order to avoid expensive clones on the `FastProcessor` side, as well as passing just the data that the core trace fragment builders will need).

Optionally, you can look at #1839 (stacked on top of this one) to look at how `CoreTraceState` is being used to generate the trace - or you can just ask me if you don't feel like sifting through thousands of lines of code 🙂

processor/src/fast/checkpoints.rs
plafer: Hmm I think clippy complains when you have `new()` without `Default`, but not the other way around - so just kept `Default`.

> And does it make sense to have _with_capacity() method?

Currently not really because of how we use `core::mem::take()` in `CoreTraceStateBuilder` - but might be an improvement to introduce one, and switch to `core::mem::swap()`. Will leave it as is for simplicity now, and revisit in the optimization phase.

processor/src/fast/trace_state_builder.rs
plafer: Right - it was a name from a previous implementation. Switched to `capture_trace_state` for now, because it both starts a new snapshot, and internally records a "previous" one. But I will be trying to clean up this struct further tomorrow.

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,219 @@
...
+    core_trace_states: Vec<CoreTraceState>,
+}
+
+impl CoreTraceStateBuilder {
+    /// Extracts the internal state out in order to create a `CoreTraceState`, and stores it
+    /// internally.
+    ///
+    /// The replay data is cleared after extraction, since each trace state is expected to be empty
+    /// at the beginning of a new trace fragment. The overflow table and block stack are not cleared
+    /// however, since they track the state of the computation since the beginning.
```
Al-Kindi-0: would this be better: "..since they track the state of the computation across trace fragments." or something similar
Edit: I think this comment no longer applies as they are now cleared as well, right?

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Convert the `CoreTraceStateBuilder` into the list of `CoreTraceState` built during
+    /// execution.
+    pub fn into_core_trace_states(mut self) -> Vec<CoreTraceState> {
+        // If there is an ongoing snapshot, finish it
+        self.finish_current_snapshot();
+
+        self.core_trace_states
+    }
+
+    fn finish_current_snapshot(&mut self) {
```
Al-Kindi-0: maybe a short comment?

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,219 @@
...
+
+// HASHER CHIPLET SHIM
+// =========================================================
+
+/// The number of hasher rows per permutation operation. This is used to compute the address for the
+/// next operation in the hasher chiplet.
+const NUM_HASHER_ROWS_PER_PERMUTATION: u32 = 8;
+
+#[derive(Debug)]
+pub struct HasherChipletShim {
```
Al-Kindi-0: maybe a comment explaining the purpose of the fields?

processor/src/fast/mod.rs
```diff
@@ -137,6 +152,9 @@ pub struct FastProcessor {
 
     /// Whether to enable debug statements and tracing.
     in_debug_mode: bool,
+
+    /// Shims used for tracing execution, if enabled.
```
Al-Kindi-0: would "... for trace building.." be better?

processor/src/fast/mod.rs
```diff
@@ -303,6 +327,23 @@ impl FastProcessor {
...
+        program: &Program,
+        host: &mut impl AsyncHost,
+    ) -> Result<(StackOutputs, AdviceProvider, Vec<CoreTraceState>), ExecutionError> {
+        self.trace_state_builder = Some(CoreTraceStateBuilder::default());
+        let stack_outputs = self.execute_impl(program, host).await?;
+
+        Ok((
+            stack_outputs,
+            self.advice,
+            self.trace_state_builder.unwrap().into_core_trace_states(),
```
Al-Kindi-0: maybe an `expect` instead?

processor/src/fast/mod.rs
```diff
@@ -375,18 +420,39 @@ impl FastProcessor {
...
+                    self.check_extract_trace_state(
+                        NodeExecutionPhase::End(node_id),
+                        &mut continuation_stack,
+                        &current_forest,
+                    );
                     self.finish_split_node(node_id, &current_forest, host)?
                 },
                 Continuation::FinishLoop(node_id) => {
                     self.finish_loop_node(node_id, &current_forest, &mut continuation_stack, host)?
                 },
```
Al-Kindi-0: Q: why don't we need to call `self.check_extract_trace_state(...)`?

processor/src/fast/mod.rs
```diff
@@ -579,25 +751,62 @@ impl FastProcessor {
...
-        self.clk += 1_u32;
-
         // call or syscall are not allowed inside a syscall
         if self.in_syscall {
             let instruction = if call_node.is_syscall() { "syscall" } else { "call" };
             return Err(ExecutionError::CallInSyscall(instruction));
         }
 
+        // In tracing mode, record the control block on the block stack
+        if self.trace_state_builder.is_some() {
```
Al-Kindi-0: maybe an if-let?

processor/src/fast/mod.rs
```diff
@@ -617,14 +826,18 @@ impl FastProcessor {
             self.in_syscall = true;
         } else {
             // set the system registers to the callee context
-            self.ctx = self.clk.into();
+            self.ctx = (self.clk + 1).into();
```
Al-Kindi-0: maybe worth explaining the +1 ?

processor/src/fast/mod.rs
```diff
@@ -655,18 +880,74 @@ impl FastProcessor {
...
+            NodeExecutionPhase::Start(current_node_id),
+            continuation_stack,
+            current_forest,
+        );
+
         // Execute decorators that should be executed before entering the node
         self.execute_before_enter_decorators(current_node_id, current_forest, host)?;
 
+        // In tracing mode, record the control block on the block stack
+        if self.trace_state_builder.is_some() {
```
Al-Kindi-0: same comment here

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+    /// Flag indicating whether currently executing within a SYSCALL block
+    pub in_syscall: bool,
+
+    /// Hash of the function that initiated the current execution context
+    /// - For root context: [ZERO; 4]
+    /// - For CALL/DYNCALL contexts: hash of the called function
+    /// - For SYSCALL contexts: hash remains from the calling function
+    pub fn_hash: Word,
+}
+
```
Al-Kindi-0: maybe add a comment?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+#[derive(Debug)]
+pub struct DecoderState {
+    pub current_addr: Felt,
+    pub parent_addr: Felt,
+}
+
+/// A checkpoint represents all the information for one row of the Stack trace.
```
Al-Kindi-0: should this be updated

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+        self.last_overflow_addr = last_overflow_addr;
+    }
+}
+
+/// Replay data for the block stack.
+#[derive(Debug, Default)]
+pub struct BlockStackReplay {
+    /// The parent address - needed for each node start operation (JOIN, SPLIT, etc).
+    node_start: VecDeque<Felt>,
+    /// The data needed recover the state on an END operation.
```
Al-Kindi-0: needed to recover

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+    node_end: VecDeque<NodeEndData>,
+    execution_contexts: VecDeque<ExecutionContextSystemInfo>,
```
Al-Kindi-0: maybe add comments to these two fields?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+///
+/// The shim records overflow values (from pop operations) and overflow addresses (representing
+/// the clock cycle of the last overflow update) and provides replay methods that return the
+/// pre-recorded values. This works naturally since the fast processor has exactly the same
+/// access patterns as the main trace generators.
+#[derive(Debug)]
+pub struct StackOverflowReplay {
+    /// Recorded overflow values and overflow addresses from pop_overflow operations. Each entry
+    /// represents a value that was popped from the overflow stack, and the overflow address of the
+    /// at the top of the overflow stack *after* the pop operation.
```
Al-Kindi-0: "...of the at the top..." needs a bit of simplification ?

processor/src/fast/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+        Self {
+            overflow_values: VecDeque::new(),
+            restore_context_info: VecDeque::new(),
+        }
+    }
+
+    // MUTATORS
+    // --------------------------------------------------------------------------------
+
+    /// Records the value returned by a pop_overflow operation, along as the overflow address stored
```
Al-Kindi-0: "... along as..." is it intentional, maybe can be re-written to simplify the phrasing

processor/src/fast/circuit_eval.rs
```diff
@@ -32,9 +32,10 @@ impl FastProcessor {
         let num_read = self.stack_get(1);
         let ptr = self.stack_get(0);
         let ctx = self.ctx;
+        let clk = self.clk;
         let circuit_evaluation =
             eval_circuit_fast_(ctx, ptr, self.clk, num_read, num_eval, &mut self.memory, err_ctx)?;
```
Al-Kindi-0: use `clk` instead

Al-Kindi-0 (commented):
Looks great to me!
Left a few comments (mostly nits).
My main question would be related to the branching in `check_extract_trace_state` which is called quite frequently and if we can do anything about it. I can't immediately see a global solution but maybe it is worth thinking about some solution nevertheless

plafer:
> I can't immediately see a global solution but maybe it is worth thinking about some solution nevertheless

Yes we have a solution for this in a following PR (TLDR that involves generics) - should have mentioned it earlier.

adr1anh:
This could be left for a future PR, but I noticed there are many instances where we use `Felt` to represent an address. We could instead use `RowIndex` which, as a `u32` could be slightly more optimal.

processor/src/fast/mod.rs
plafer: Rewrote the comment completely

processor/src/fast/trace_state_builder.rs
plafer: Indeed was outdated - cleaned up all docstrings related to `CoreTraceStateBuilder`

processor/src/fast/trace_state_builder.rs
plafer: done, and renamed method

processor/src/fast/trace_state_builder.rs
plafer: Done (and added other docstrings too)

processor/src/fast/mod.rs

processor/src/fast/mod.rs
plafer: I thought I had fixed this - but apparently I had done it only for the `Start` ones. I moved all the `check_extract_trace_state()` calls inside the `finish()` methods to remove the confusion when looking at that part of the code.

But to answer your question directly - it's not that we don't call it, it's that we call it with a different `NodeExecutionPhase` depending on the `condition` at the top of the stack, so we're forced to call it inside `finish_loop_node()`.

processor/src/fast/mod.rs
plafer: We need to do this for borrow-checker reasons - basically an `if let Some(trace_state_builder) = &mut self.trace_state_builder` would mutably borrow all of `self`, which then prevents us from calling e.g. `self.stack_depth()` (which attempts to immutably borrow it). But this will be naturally cleaned up soon when we implement #2089.

processor/src/fast/mod.rs
plafer: Same as https://github.com/0xMiden/miden-vm/pull/2023#discussion_r2294519435

processor/src/fast/mod.rs

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Updated the whole docstring

processor/src/fast/checkpoints.rs
plafer: Done

processor/src/fast/checkpoints.rs
plafer: Done - added an example too to be more clear

processor/src/fast/checkpoints.rs
plafer: Fixed

plafer:
> I noticed there are many instances where we use Felt to represent an address. We could instead use RowIndex which, as a u32 could be slightly more optimal.

The reason they're `Felt` (at least the ones in `CoreTraceState`) is that we rarely perform any operations on them (i.e. pretty much only on a `RESPAN`), and otherwise they're stored directly in the trace. So having them already as a `Felt` saves a Montgomery reduction during `Felt::new()` for each address on the trace generator side. And the `HasherChiplet` is where we do computations over addresses, so they're stored as a `u32` there.

processor/src/fast/checkpoints.rs
plafer: Renamed to `NodeExecutionState`, and added more docs

plafer:
Rebased on the latest `next` and squashed the "PR fix" commits

processor/src/fast/u32_ops.rs
```diff
@@ -192,6 +230,7 @@ impl FastProcessor {
         &mut self,
         f: impl FnOnce(u64, u64) -> u64,
         err_ctx: &impl ErrorContext,
+        _tracer: &mut impl Tracer,
```
bobbinth: Why do we need to pass the tracer into this function? Doesn't seem like i tis used here, and from what I can tell, we can safely remove this parameter.

processor/src/fast/u32_ops.rs
```diff
@@ -213,6 +252,7 @@ impl FastProcessor {
         err_code: Felt,
         f: impl FnOnce(u64, u64) -> Result<(u64, u64), ExecutionError>,
         err_ctx: &impl ErrorContext,
+        _tracer: &mut impl Tracer,
```
bobbinth, line 249: Same comment as above.

processor/src/fast/io_ops.rs
```diff
@@ -133,6 +178,8 @@ impl FastProcessor {
             .pop_stack_dword()
             .map_err(|err| ExecutionError::advice_error(err, self.clk, err_ctx))?;
 
+        tracer.record_advice_pop_stack_dword(words);
```
bobbinth, line 189: Not sure if it'll simplify things, but, at least in theory, we may be able to replace this with two calls to `record_advice_pop_stack_word()` and get rid of `record_advice_pop_stack_dword()`.

processor/src/fast/circuit_eval.rs
```diff
@@ -86,15 +87,22 @@ pub fn eval_circuit_fast_(
 
     let mut ptr = ptr;
     // perform READ operations
+    // Note: we pass in a `NoopTracer`, because the parallel trace generation skips the circuit
+    // evaluation completely
```
bobbinth, line 90: Is this temporary (i.e., until the chiplets trace is also being built)? Or do we not need to capture the memory read here at all?

processor/src/fast/trace_state_builder.rs
bobbinth, line 1: nit: I would probably have a separate "trace" module and move trace-related structs there. For example:

- This file could be under `src/fast/trace/state_builder.rs`.
- We could also have `src/fast/trace/state.rs` and in the future probably more sub-modules.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,493 @@
...
+/// well as recording the various values read during execution in the corresponding "replays" (e.g.
+/// values read from memory are recorded in [MemoryReplay], values read from the advice provider are
+/// recorded in [AdviceReplay], etc).
+///
+/// Then, to generate a trace fragment, we initialize the state of the processor using the snapshot
+/// stored in [StateSnapshot], and replay the recorded values as they are encountered during
+/// execution (e.g. when encountering a memory read operation, we will replay the value rather than
+/// querying the memory chiplet).
+#[derive(Debug, Default)]
+pub struct CoreTraceStateBuilder {
```
bobbinth: nit: as far as I can tell, this struct doesn't actually build the trace - maybe we should rename it? Following the `Tracer` and `NoopTracer` terminology, maybe this should be `ExecutionTracer`? (and in the future we could also have `DebugTracer`)

processor/src/fast/trace_state.rs
```diff
@@ -0,0 +1,697 @@
...
+
+    /// Returns the overflow address (b1 helper column) using the stack overflow replay
+    pub fn overflow_addr(&mut self) -> Felt {
+        self.last_overflow_addr
+    }
+
+    pub fn num_overflow_elements_in_current_ctx(&self) -> usize {
+        debug_assert!(self.stack_depth >= MIN_STACK_DEPTH);
+        self.stack_depth - MIN_STACK_DEPTH
+    }
```
bobbinth, line 170: nit: let's add doc comments to this and the following two methods.

Also, it is probably worth mentioning here that `self.stack_depth` is guaranteed to be >= 16, and so we should never have an underflow here.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,493 @@
...
+    // State stored at the start of a trace fragment
+    state_snapshot: Option<StateSnapshot>,
```
bobbinth: Under what circumstances this could be `None`? Maybe worth explaining this in the comment.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,493 @@
...
+        // check if we need to start a new trace state
+        if processor.clk.as_usize().is_multiple_of(NUM_ROWS_PER_CORE_FRAGMENT) {
+            self.start_new_trace_state(
+                SystemState {
+                    clk: processor.clk,
+                    ctx: processor.ctx,
+                    fmp: processor.fmp,
+                    in_syscall: processor.in_syscall,
+                    fn_hash: processor.caller_hash,
+                },
```
bobbinth: nit: it may be wroth implementing something like `SystemState::from_processor(&FastProcessor)`.

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,493 @@
...
+            );
+        }
+
+        // Update block stack
+        match &execution_state {
+            NodeExecutionState::BasicBlock {
+                node_id: _,
+                batch_index: _,
+                op_idx_in_batch: _,
+            } => {
```
bobbinth: nit: if we don't care about the fields, we could probably write this just as:
```rust
NodeExecutionState::BasicBlock { .. }
```

processor/src/fast/trace_state_builder.rs
```diff
@@ -0,0 +1,493 @@
...
+                MastNode::Join(_)
+                | MastNode::Split(_)
+                | MastNode::Loop(_)
+                | MastNode::Dyn(_)
+                | MastNode::Call(_) => {
+                    self.record_control_node_start(&current_forest[*mast_node_id], processor);
+                },
+                MastNode::Block(basic_block_node) => {
+                    let block_addr = self.hasher.record_hash_basic_block(basic_block_node);
+                    let parent_addr = self.block_stack.push(block_addr, BlockType::Span, None);
```
bobbinth: Not for this PR, but `BlockType::Span` is obsolete terminology and we should update it at some point.

processor/src/fast/trace_state.rs
```diff
@@ -0,0 +1,697 @@
...
+    pub stack: StackState,
+    pub block_stack_replay: BlockStackReplay,
+    pub stack_overflow: StackOverflowReplay,
+    pub memory: MemoryReplay,
+    pub advice: AdviceReplay,
+    pub hasher: HasherReplay,
+    pub external_node_replay: ExternalNodeReplay,
+    pub traversal: ContinuationStack,
+    pub execution_state: NodeExecutionState,
+    pub initial_mast_forest: Arc<MastForest>,
```
bobbinth: Probably not for this PR, but I wonder if it would make sense to group all the "replay" structs into a dedicated struct. For example, it could look something like:
```rust
pub struct ProcessorReplay {
    pub block_stack: BlockStackReplay,
    pub stack_overflow: StackOverflowReplay,
    pub memory: MemoryReplay,
    pub advice: AdviceReplay,
    pub hasher: HasherReplay,
    pub external_nodes: ExternalNodeReplay,
}
```
It will be a bit more code, but also may provide a bit cleaner separation between state and replay fields.

processor/src/fast/trace_state.rs
```diff
@@ -0,0 +1,701 @@
...
+
+use crate::{ContextId, continuation_stack::ContinuationStack};
+
+// CORE TRACE STATE
+// ================================================================================================
+
+/// The main state for the core processor, which captures all the necessary information to
+/// reconstruct the trace during parallel execution.
+#[derive(Debug)]
+pub struct CoreTraceState {
```
bobbinth: Would it be fair to say that this really consists for 3 parts:

1. Information about the first row in the trace fragment.
2. The replay information from various components that happened during this trace fragment execution.
3. Information about where we are in the global program execution (i.e., `traversal`, `execution_state`, and `initial_mast_forest`).

If this is correct, I'd update the doc comment to include it (maybe in better phrasing), and maybe in the future try to group fields into these 3 groups (e.g., something similar to what I mentioned in the previous comment).

processor/src/fast/mod.rs
```diff
@@ -215,6 +227,11 @@ impl FastProcessor {
         &self.stack[self.stack_bot_idx..self.stack_top_idx]
     }
 
+    /// Returns the top 16 elements of the stack.
+    pub fn stack_top(&self) -> &[Felt] {
+        &self.stack[self.stack_top_idx - MIN_STACK_DEPTH..self.stack_top_idx]
+    }
```
bobbinth, line 222: From what I can tell, this is used only in one place where we immediately convert the returned data into an array. It may be worth converting the data here - i.e., the signature would look like:
```rust
pub fn stack_top(&self) -> [Felt; MIN_STACK_DEPTH] {
    ...
}
```

processor/src/fast/mod.rs
```diff
@@ -302,30 +319,60 @@ impl FastProcessor {
...
+        mut self,
+        program: &Program,
+        host: &mut impl AsyncHost,
+        tracer: &mut impl Tracer,
+    ) -> Result<(StackOutputs, AdviceProvider), ExecutionError> {
+        let stack_outputs = self.execute_impl(program, host, tracer).await?;
         Ok((stack_outputs, self.advice))
     }
 
     async fn execute_impl(
```
bobbinth, line 347: Do we still need `execute_impl()`? It seems like now it could be replaced with `execute_with_tracer()`. If we do still need it, may be a good idea to add a brief doc comment explaining why.

processor/src/fast/mod.rs
```diff
@@ -417,26 +490,35 @@ impl FastProcessor {
...
     #[inline(always)]
     fn start_join_node(
         &mut self,
         join_node: &JoinNode,
         node_id: MastNodeId,
-        current_forest: &MastForest,
+        current_forest: &Arc<MastForest>,
         continuation_stack: &mut ContinuationStack,
         host: &mut impl AsyncHost,
+        tracer: &mut impl Tracer,
```
bobbinth: Not for this PR, but I wonder if it would make sene to create a wrapper struct to group some of these parameters together - e.g., something like `ExecutionContext`. This way, we could make these method signatures (and their call-sites) a bit more concise.

processor/src/fast/mod.rs
bobbinth, line 1: Not for this PR, but we should try to split this file up into several smaller files to make it easier to follow.

processor/src/fast/mod.rs
```diff
@@ -62,6 +71,9 @@ const WORD_SIZE_FELT: Felt = Felt::new(4);
 /// The size of a double-word.
 const DOUBLE_WORD_SIZE: Felt = Felt::new(8);
 
+/// The number of rows per core trace fragment.
+pub const NUM_ROWS_PER_CORE_FRAGMENT: usize = 1024;
```
bobbinth, line 58: At some point, we may want to make this a runtime setting - or at the very least try to benchmark performance for different values of this constant (I suspect that 1024 is way too low).

bobbinth (approved):
Looks great! Thank you! I left some comments inline - most of them are about improving code organization and many could be done in follow-up PRs.

Also, it may have been somewhere in the comments, but I'm curious what does the current performance look like for:

- Pure execution (with the `NoopTracer`).
- Execution with trace building.

processor/src/fast/u32_ops.rs
plafer: Oops - removed

processor/src/fast/u32_ops.rs
plafer: Removed

processor/src/fast/io_ops.rs
plafer: I personally prefer keeping it as is, since it maps clearly to `AdviceProvider::pop_stack_dword()`, both for clarity and debuggability.

processor/src/fast/circuit_eval.rs
plafer: This is not temporary - the parallel trace generators assume that the program was run correctly, and hence that the circuit evaluated to 0. So re-running the circuit evaluation would be wasteful, and not provide any useful data for trace generation purposes. Therefore they'll never make the memory queries, and so it's important that we don't record them here.

processor/src/fast/trace_state_builder.rs
plafer: Good idea, I was hoping to find a better name! Switched to `ExecutionTracer`.

processor/src/fast/trace_state.rs
plafer: > Also, it is probably worth mentioning here that self.stack_depth is guaranteed to be >= 16, and so we should never have an underflow here.

Since that assumption applies to a few methods, I added that comment to the field's docstring instead

processor/src/fast/trace_state_builder.rs
plafer: Added

>    // This field is only set to `None` at initialization, and is populated when starting a new
    // trace fragment with `Self::start_new_trace_state()`. Hence, on the first call to
    // `Self::start_new_trace_state()`, we don't extract a new `CoreTraceState`, but in every other
    // call, we do.

processor/src/fast/trace_state_builder.rs
plafer: https://github.com/0xMiden/miden-vm/issues/2126

processor/src/fast/trace_state.rs
plafer: Yes I'm keeping this in the back of my mind. However this struct will grow/change when we implement trace generation for chiplets and range checker, so I would leave the refactor for when we have all the pieces.

processor/src/fast/mod.rs
plafer: It's used to access words in the stack in the parallel trace generation, so I would leave this as is.

processor/src/fast/mod.rs
plafer: Yes I intend to change this in #1839 - since it's also needed for tests (where you want to e.g. set the value to 2).

processor/src/fast/trace_state_builder.rs
plafer: Done

processor/src/fast/trace_state.rs
plafer: Ended up going for it, but named it `ExecutionReplay`

processor/src/fast/trace_state.rs
plafer: I reworked `CoreTraceState`, namely in that I renamed it to `TraceFragmentContext`, and created sub-structs to better organize the fields. I also reworked the docstring a lot, so hopefully it's clearer now.

We can also rework this in #1839, but I think it's in a much better state now than it was.

plafer: Last thing is I would convert the `continuation` field into a replay - I will create an issue for this.

Created #2128

processor/src/fast/mod.rs
plafer: We need it so that tests can access the processor state after execution - added that to a docstring.

processor/src/fast/mod.rs
plafer: Gave it a first pass - we can always refine later

processor/src/fast/mod.rs
plafer: Agreed - but let's leave this for a subsequent PR

Issue #2018, opened 2025-07-22
Add accessors for `AdviceProvider` fields


PhilippGackstatter:
The fields in `AdviceProvider` used to be public and so we accessed them directly, but they no longer are on latest `next` and I don't see a way to access them now (please correct me if I'm missing them). It would be great if one or more of the following things could be done:
- `impl From<AdviceProvider> for AdviceInputs` could be added
- add accessors taking `&self` and returning a reference to stack, advice map and merkle store
- add an `into_parts` that returns the underlying parts so cloning can be avoided when possible.
- make the fields public again, as was recently done for `AdviceInputs`.

It would be great if this could eventually be merged into `async-experimental`, since that's what I'm building on in miden-base. Thanks!

bobbinth:
I've opened https://github.com/0xMiden/miden-vm/pull/2024 against `next` and also optimistically committed the same changes into `async-experimental` (in 6f1c807aeafd78b03c4b935b9683fce495e78b60).

adr1anh:
I think having accessors and `into_parts` makes it unnecessary to make the fields public. If we do though, accessors can just be replaced with calls to the underlying structures. We will need specialized accessors if/when we implement lazy loading of `AdviceMap`s. 

Are there any reasons why would want to prevent full access to the `AdviceProvider` by exposing fields? The only case I could imagine is if we want to prevent arbitrary handlers from reading sensitive data stored there.

plafer:
I briefly looked and didn't see any invariant upheld by the methods, so would also be in favor of making the fields public, and removing trivial methods (e.g. `extend_merkle_store()`). So basically anyone is free to access the inner state as needed, and the methods we provide are treated as making regular access patterns convenient.

bobbinth:
For `AdviceProvider`, I'd prefer to keep the fields private (at least for now). The main reason is that we may refactor how the internals work in not-too-distant future (e.g., to avoid merging maps too many times). If however, we end up in a state which doesn't really impose any conditions on how the fields can be mutated, we can revisit this again.

bobbinth:
Closed by #2024.

PR #2024, opened 2025-07-22
feat: add `AdviceProvider::into_parts()` method

bobbinth:
This small PR addresses https://github.com/0xMiden/miden-vm/issues/2018 by implementing `AdviceProvider::into_parts()` method. I went with this option as implementing `From<AdviceProvider> for AdviceInputs` is a bit more tricky as we'd need to reverse the advice stack, which I don't think we really care for.

Another option is to just discard the stack and return only `(AdviceMap, MerkleStore)` from this method, but I decided to be a bit more comprehensive in case the stack is also needed for some reason.

### DIFF
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -302,11 +302,22 @@ impl AdviceProvider {
         self.store.extend(iter);
     }
 
+    // MUTATORS
+    // --------------------------------------------------------------------------------------------
+
     /// Extends the contents of this instance with the contents of an `AdviceInputs`.
     pub fn extend_from_inputs(&mut self, inputs: &AdviceInputs) -> Result<(), AdviceError> {
         self.extend_stack(inputs.stack.iter().cloned().rev());
         self.extend_merkle_store(inputs.store.inner_nodes());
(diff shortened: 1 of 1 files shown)


PhilippGackstatter:
Thinking about it again, would it hurt to add accessors, e.g. `fn advice_map(&self) -> &AdviceMap`? Sometimes we only have an `&AdviceProvider` and cannot deconstruct it into its parts but still want to access the advice map. Similarly for stack and merkle store.

I don't think I need those specifically right now, but may be useful in the future.

plafer (approved):
LGTM

bobbinth:
> Thinking about it again, would it hurt to add accessors, e.g. `fn advice_map(&self) -> &AdviceMap`? Sometimes we only have an `&AdviceProvider` and cannot deconstruct it into its parts but still want to access the advice map. Similarly for stack and merkle store.

I mentioned this in https://github.com/0xMiden/miden-vm/issues/2018#issuecomment-3108899063 - but it is possible that some of the future refactorings may impose some restrictions on how we access (and mutate) specific fields - so, I'd rather keep them private for now.

Issue #2020, opened 2025-07-22
Make `AdviceProvider` accessible after `FastProcessor` execution


PhilippGackstatter:
`FastProcessor::execute` takes `self` and consumes the processor struct but only returns the stack outputs. For transaction execution, we also need access to the [advice provider after execution](
https://github.com/0xMiden/miden-base/blob/dc1036379b6bc5c7ad92309e7c2fd210ed29049c/crates/miden-tx/src/executor/mod.rs#L190) to access the values that were inserted during execution.

It would be good to return the `AdviceProvider` as part of the return values of `execute` or make it accessible in some other way.

bobbinth:
Ah - this is what i mentioned in https://github.com/0xMiden/miden-vm/issues/2022#issuecomment-3104667121. I believe this is blocking `miden-base` migration to the `FastProcessor`, and it may not be too difficult to enable.

@plafer - could you look into this?

plafer:
Echoing my comment [here](https://github.com/0xMiden/miden-base/pull/1636#discussion_r2225500948) - should this be implemented ASAP and merged into `async-experimental`?

bobbinth:
> Echoing my comment [here](https://github.com/0xMiden/miden-base/pull/1636#discussion_r2225500948) - should this be implemented ASAP and merged into `async-experimental`?

Yes, we should do this for `async-experimental` ASAP otherwise we can't progress on the async work.

bobbinth:
Closed by #2026 and I also cherry-picked the commit into `next`.

PR #2026, opened 2025-07-23
chore: `FastProcessor::execute()` now also returns the AdviceProvider`

plafer:
This commit is to unblock miden-base and merged in the async-experimental branch. It addresses #2020, but we should only close it once this makes it into `next`.


processor/src/fast/mod.rs
```diff
@@ -319,8 +319,10 @@ impl FastProcessor {
         mut self,
```
bobbinth, line 319: nit: I'd update the doc comment as well.

bobbinth (approved):
Looks good! Thank you! I left a small nit inline Also, I'd probably still add a changelog entry (if all works out well, we'll just merge the experimental branch into `next`).

PR #2027, opened 2025-07-24
feat: `SourceManager` make sync trait bound dependent on feature

drahnr:


crates/debug/types/src/source_manager.rs
```diff
@@ -260,10 +260,19 @@ impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
...
 /// downcasting, e.g. `Arc<dyn SourceManagerSync> as Arc<dyn SourceManager>`.
+
+#[cfg(feature = "syncbound")]
 pub trait SourceManagerSync: SourceManager + Send + Sync {}
 
+#[cfg(feature = "syncbound")]
 impl<T: ?Sized + SourceManager + Send + Sync> SourceManagerSync for T {}
 
+#[cfg(not(feature = "syncbound"))]
+pub trait SourceManagerSync: SourceManager + Send {}
```
bitwalker, line 271: Could you clarify why this is needed? It's hard to imagine a scenario where `SourceManagerSync` _not_ being `Sync` makes sense. Making it contingent on a Cargo feature is especially problematic because of how Cargo's feature unification works.

At this point, the type alias is almost useless - if we have as many variations of use cases as we do repos, we should be making the relevant types generic over the `SourceManager` type, rather than using trait objects and trying to find some kind of way to pretend all the instances look and work the same way, when they just don't.

processor/src/host/mod.rs
```diff
@@ -111,16 +111,17 @@ pub trait AsyncHost: BaseHost {
...
     /// the advice provider.
     fn on_event(
         &mut self,
         process: &ProcessState<'_>,
         event_id: u32,
-    ) -> impl Future<Output = Result<Vec<AdviceMutation>, EventError>> + Send;
+    ) -> impl SFuture<Result<Vec<AdviceMutation>, EventError>>;
 }
+
+pub trait SFuture<O>: Future<Output = O> + Send {}
```
bitwalker, line 125: If we _really_ need a type alias like this (and I would argue we don't), then we should call it `SendFuture` so it is at least immediately obvious what it is. I hadn't the slightest idea from just `SFuture` alone.

bitwalker (changes requested):
I don't think we should merge this, but would like to know more about the motivation so we can try to find a better overall solution to the underlying problem. Presumably we have somewhere that wants a `SourceManager` that is _just_ `Send`, so we now have three different "types" of `SourceManager` that we want to use in the same place - that is unequivocally a sign that we should be making use of generics for that type, not trait objects.

processor/src/host/mod.rs
plafer: I would remove this trait - `impl Future<...> + Send` is fine and used everywhere in the ecosystem

crates/debug/types/src/source_manager.rs
plafer: Agreed - I'm a bit lost as to what this PR is trying to achieve as well. My understanding of what we wanted to try was

1. The `Send` bound on `AsyncHost` methods would be conditional on the target (i.e. wasm doesn't have the bounds while all other targets do)
  - Maybe `+ 'static` as well but we'd hold off on that one until we're completely clear that it's necessary
2. I don't believe there needs to be any changes to the `SourceManager` traits, but I might be missing something on that one.

drahnr:
This was meant to be a draft, and not reviewed, but gh defaults to whatever was used _last_ and I missed that.

The idea was that it would have an effect on the `AsyncHost` `async fn` trait bounds. Now, if there is a reference held to the `dyn SourceManagerSync`. Now we need the `Sync + Send` for the `Arc<X>` to be `Send`.

Closing until we discuss another approach

processor/src/host/mod.rs
drahnr: Assuming the approach holds, the return is required, otherwise we'd have `#[cfg(..)]` not only in place of the trait declaration, but also the implementations.

crates/debug/types/src/source_manager.rs
drahnr: 1. after the previous PR I interpreted https://github.com/0xMiden/miden-vm/pull/2019#pullrequestreview-3050490013 as change from `#[cfg(target_arch = "..")]` into an actual feature. Maybe that was wrong.
* `'static` seems like an addenda for usability, but nothing vital
2. See the last comment

processor/src/host/mod.rs
plafer: Good point - my assumption was that downstream implementations were going to need to choose whether they target wasm, or everything else. But maybe that's the wrong assumption to make

crates/debug/types/src/source_manager.rs
drahnr: > At this point, the type alias is almost useless - if we have as many variations of use cases as we do repos, we should be making the relevant types generic over the `SourceManager` type, rather than using trait objects and trying to find some kind of way to pretend all the instances look and work the same way, when they just don't.

So a returned `Future` would depend on the usage of `self` and the concrete type `S` to determine the bounds on the generated future.

```
struct X<S> {
source_manager: Arc<S>,
}

impl<S> X where S: SourceManager {
  async fn y(&self) {
    // some impl using `self` across an await point
  }
}
```

To my understanding the generated future type will be contingent on the generic `S`. To be of any use `S` must be bounded on `SourceManager` as a minimum. For the _user_ of the generated future type `async fn y` we do not require a `Send` bound, it's autoimplemented by the compiler on the implicit, generated future type.

tl;dr it appears @bitwalker  approach would indeed work better, with a few caveats (internally we cannot spawn tasks with `Arc<S>`, but that's out of scope to my understanding, and not possible on `wasm32` anyways) and the cost of an extra generic. On a first quick review, it seems the impact wouldn't be too large.

PR #2028, opened 2025-07-24
feat: procedure type signature support

bitwalker:
This PR integrates the type system used by the compiler to reason about procedure arguments/results (defined in `midenc-hir-type`), and implements the following features:

* Syntax tree nodes to represent type declarations we intend to support in MASM source code; namely named type aliases for conveying semantic information about concrete types, and enum types for expressing a set of constant declarations that correspond to variants of a C-like enumeration, which has a corresponding type alias that can then be used in type signatures.
* Semantic analysis of the new declaration types
* The addition of an optional type signature to all procedure-related types throughout the parser and assembler.
* Resolution of type signatures for aliased procedures, i.e. though procedure aliases do not themselves provide a type signature, we do resolve the type signature of re-exported procedure aliases in the final assembly so that it appears as if they do.
* Serialization of type information in packages
* The addition of type signatures to the exports declared in a package  manifest
* Some modifications to how exports are represented in libraries/packages to make working with them more ergonomic.

**NOTE:** This PR does _not_ include changes to the MASM grammar for expressing types in MASM. That will be done in a subsequent change to avoid bikeshedding those details here.

Additionally, there are a few minor changes made here as well:

* I've modified the semantic analyzer to preserve constants defined in the original module, though I have not yet made the changes to stop performing constant propagation during parsing, and instead defer it to assembly-time. It made sense to make that change here while I was already adding support for enum type syntax which involves constants.
* I've added a new `LibraryExport` type to represent metadata about the exported procedures of a `Library`. Previously the only information we had was the `MastNodeId`, but now that it is expanding, a more general type was needed. This also affected some of the related APIs.
* I fixed the `Arbitrary` implementation in `miden-assembly` used for testing packages to actually take into account the procedures in the MAST artifact, rather than generate random manifests.
* I added some basic trait implementations to some types that were missing them in the AST.

**NOTE:** I haven't yet updated the changelog, I'll do that shortly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Add constraints evaluation check to recursive verifier ([#1997](https://github.com/0xMiden/miden-vm/pull/1997)).
 - Add `AdviceProvider::into_parts()` method ([#2024](https://github.com/0xMiden/miden-vm/pull/2024)).
 - Add `drop_stack_top` procedure in std::sys ([#2031](https://github.com/0xMiden/miden-vm/pull/2031))
+- Add type information to procedures in the AST, `Library`, and `PackageExport`
+  types ([#2028](https://github.com/0xMiden/miden-vm/pull/2028))
 
 #### Changes
 
@@ -23,6 +25,11 @@
 - Converted `FastProcessor::execute()` from recursive to iterative execution ([#1989](https://github.com/0xMiden/miden-vm/issues/1989)).
 - [BREAKING] made `AdviceInputs` field public and removed redundant accessors ([#2009](https://github.com/0xMiden/miden-vm/pull/2009)).
 - Enhancement for all benchmarks (incl. `program_execution_fast`) are built and run in a new CI job with required feature flags [(#https://github.com/0xMiden/miden-vm/issues/1964)](https://github.com/0xMiden/miden-vm/issues/1964)
+- [BREAKING] `Library::exports` now returns `(&QualifiedProcedureName, &LibraryExport)` rather than just `&QualifiedProcedureName`, to allow callers
+to extract more useful information. ([#2028](https://github.com/0xMiden/miden-vm/pull/2028))
+- [BREAKING] The serialized representation for `Package` was changed to include
+procedure type information. Older packages will not work with the new serialization code, and vice versa. The version of the binary format was incremented accordingly. ([#2028](https://github.com/0xMiden/miden-vm/pull/2028))
+- [BREAKING] Procedure-related metadata types in the `miden-assembly` crate in some cases now require an optional type signature argument. If that information is not available, you can simply pass `None` to retain current behavior. ([#2028](https://github.com/0xMiden/miden-vm/pull/2028))
 
 ## 0.16.4 (2025-07-24)
 


assembly-syntax/src/ast/type.rs
```diff
@@ -0,0 +1,278 @@
...
+/// See the [EnumType] docs for more information.
+#[derive(Debug, Clone)]
+pub struct Variant {
+    pub span: SourceSpan,
+    /// The documentation string attached to the constant derived from this variant.
+    pub docs: Option<DocString>,
+    /// The name of this enum variant
+    pub name: Ident,
+    /// The discriminant value associated with this variant
+    pub discriminant: ConstantExpr,
```
plafer, line 239: Isn't `ConstantExpr` too "broad" here, allowing e.g.

```
enum Color {
  RED = 1 + 2,
}
```

Even though the syntax probably wouldn't allow for that, is there a reason for using `ConstantExpr` rather than some representation of integer?

assembly-syntax/src/ast/type.rs
```diff
@@ -0,0 +1,278 @@
...
+/// of the enumeration when derived from a constant.
+#[derive(Debug, Clone)]
+pub struct EnumType {
+    span: SourceSpan,
+    /// The documentation string attached to this definition.
+    docs: Option<DocString>,
+    /// The enum name
+    name: Ident,
+    /// The concrete type of each enum variant
+    ty: Type,
```
plafer: How does `Type` represent "each enum variant"? Are we assuming we would be using e.g. `Type::Array` or `Type::List`?

My first instinct was that this should have been `Vec<Type>`

assembly-syntax/src/ast/type.rs
```diff
@@ -0,0 +1,278 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.docs == other.docs && self.ty == other.ty
+    }
+}
+
+impl core::hash::Hash for TypeAlias {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.docs.hash(state);
+        self.name.hash(state);
+        self.ty.hash(state);
```
plafer: nit: would destructure `self` here to get compiler warnings if the fields of `TypeAlias` change

assembly-syntax/src/ast/module.rs
```diff
@@ -198,6 +205,68 @@ impl Module {
...
+        self.types.push(ty.into());
+        Ok(())
+    }
+
+    /// Define a new enum type `ty`
+    ///
+    /// Returns `Err` if:
+    ///
+    /// * A type alias with the same name as the enum type is already defined
+    /// * A constant with the same name as any of the enum variants is already defined
```
plafer:
```suggestion
    /// * A constant or another enum's variant with the same name as any of the enum variants is already defined
```
minor nit - basically trying to convey that variants from any enum creates a constant and is indistinguishable from hand-written constants. Not sure if this is explicitly documented somewhere, but would be good to have

assembly-syntax/src/ast/module.rs
```diff
@@ -322,6 +391,26 @@ impl Module {
         &self.advice_map
     }
 
+    /// Get an iterator over the constants defined in this module.
+    pub fn constants(&self) -> core::slice::Iter<'_, Constant> {
```
plafer, line 397: nit: any reason to specify the current iterator used as opposed to `impl Iterator<...>`?

Applies to other methods too

package/src/package/manifest.rs
```diff
@@ -15,30 +15,34 @@ use crate::Dependency;
...
     pub digest: Word,
+    /// The type signature of the exported procedure
+    #[cfg_attr(feature = "arbitrary", proptest(value = "None"))]
+    pub signature: Option<FunctionType>,
 }
 
 impl fmt::Debug for PackageExport {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let Self { name, digest } = self;
+        let Self { name, digest, signature } = self;
```
plafer, line 96: we'd need a `#[cfg_attr(feature = "arbitrary", proptest(value = "None"))]` for the `signature` field here right?

plafer (approved):
LGTM! Left some questions and nits, but looks quite good overall.

assembly-syntax/src/ast/type.rs
bitwalker: The reason is to allow both literals and references to constant declarations, e.g. `RED = SOME_CONSTANT`.

But I could also imagine wanting to derive discriminants from some base value, e.g.:

```
const.WORD_SIZE=16

# Represents the various bucket sizes in a memory allocator
enum SizeClass : u32 {
    SMALL = WORD_SIZE,
    MEDIUM = WORD_SIZE * 8,
    LARGE = WORD_SIZE * WORD_SIZE,
}
```

So it might actually be desirable to support arbitrary expressions in addition to simple literals and constant names.

assembly-syntax/src/ast/type.rs
bitwalker: A better explanation for this field is that it is the type of the enum discriminant (i.e. the integral value associated with each variant). I can reword the comment here.

assembly-syntax/src/ast/module.rs
bitwalker: For slice iterators in particular, it allows for better optimizations due to all of the traits implemented by the concrete `Iterator` type, i.e. `DoubleEndedIterator`, `ExactSizeIterator`, `TrustedLen`, `FusedIterator`, etc. Additionally, it is possible to recover the underlying slice from it, so it is considerably more flexible.

package/src/package/manifest.rs
bitwalker: Remember it's `#[cfg_attr]`, not `#[cfg]`, i.e. what is conditionally enabled is the `#[proptest(value = "None")]` attribute, not the `signature` field itself.

package/src/package/manifest.rs
```diff
@@ -15,30 +15,34 @@ use crate::Dependency;
 #[cfg_attr(feature = "arbitrary", derive(proptest_derive::Arbitrary))]
 pub struct PackageManifest {
     /// The set of exports in this package.
-    pub exports: BTreeSet<PackageExport>,
+    pub exports: BTreeMap<QualifiedProcedureName, PackageExport>,
```
bobbinth: Why do we need to change this to be a `BTreeMap`? `PackageExport` already contains `QualifiedProcedureName` for the export, and from what I can tell, we are not really relying on the need to look up exports by name (though, I might have missed something).

Also, if we do want to make this into a map, it may make sense to make the field private and add methods to access/mutate it. This way, we'd be able to enforce that `exports` field is always internally consistent (i.e., the key in the map actually corresponds to the export's name).

assembly-syntax/src/library/mod.rs
```diff
@@ -102,9 +131,9 @@ impl Library {
         &self.digest
     }
 
-    /// Returns the fully qualified name of all procedures exported by the library.
-    pub fn exports(&self) -> impl Iterator<Item = &QualifiedProcedureName> {
-        self.exports.keys()
+    /// Returns the fully qualified name and metadata of all procedures exported by the library.
+    pub fn exports(&self) -> impl Iterator<Item = (&QualifiedProcedureName, &LibraryExport)> {
+        self.exports.iter()
```
bobbinth: Is the main reason for returning an iterator with `(&QualifiedProcedureName, &LibraryExport)` items because `LibraryExport` doesn't contain `QualifiedProcedureName`? If so, I think it would be clear to add `QualifiedProcedureName` to `LibraryExport` and then this becomes something like:
```rust
pub fn exports(&self) -> impl Iterator<Item = &LibraryExport> {
    ...
}
```

Cargo.toml
```diff
@@ -56,16 +56,30 @@ miden-verifier = { path = "./verifier", version = "0.17", default-features = fal
 # Miden crates
 miden-crypto = { version = "0.15", default-features = false }
 miden-formatting = { version = "0.1", default-features = false }
+midenc-hir-type = { version = "0.1", default-features = false }
```
bobbinth, line 59: This introduces a bit of a circular dependency between repos, right? That is, some `miden-vm` crates now depend on `midenc-hir-type` which is in the `miden-compiler` repo, and some `miden-compiler` crates depend on `miden-compiler` creates depend on `miden-vm` crates. If so, this may make the release process a bit more complicated.

An alternative could be to move `miden-hir-type` crate into `miden-vm` repo (maybe under a different name). It seems like it is a more appropriate place for it anyways since it is a part of how Miden assembly is defined (for only the AST part, but in the future, the syntax as well).

assembly-syntax/src/ast/procedure/procedure.rs
```diff
@@ -106,15 +116,15 @@ impl Procedure {
...
 
-    /// Modifies the visibility of this procedure.
-    ///
-    /// This is made crate-local as the visibility of a procedure is virtually always determined
-    /// by the source code from which it was derived; the only exception being kernel modules,
-    /// where exported procedures take on syscall visibility once the module is identified as
-    /// a kernel.
-    pub(crate) fn set_visibility(&mut self, visibility: Visibility) {
+    /// Override the visibility of this procedure.
+    pub fn set_visibility(&mut self, visibility: Visibility) {
```
bobbinth, line 120: Are there any implications to making this public? Previously, the comment said that we don't need this to be public because the visibility is determined by the source code. I'm assuming that's still the case, but now we may need to build `Procedure` not just from the source code?

assembly-syntax/src/ast/type.rs
```diff
@@ -0,0 +1,280 @@
...
+/// C-style enumerations are effectively a type alias for an integer type with a limited set of
+/// valid values with associated names (referred to as _variants_ of the enum type).
+///
+/// In Miden Assembly, these provide a means for a procedure to declare that it expects an argument
+/// of the underlying integral type, but that values other than those of the declared variants are
+/// illegal/invalid. Currently, these are unchecked, and are only used to convey semantic
+/// information. In the future, we may perform static analysis to try and identify invalid instances
+/// of the enumeration when derived from a constant.
+#[derive(Debug, Clone)]
+pub struct EnumType {
```
bobbinth, line 131: Question: why is `EnumType` defined here rather than in `midenc_hir_type`? It seems like all other types come from there, but I guess we need the enum type only here?

assembly-syntax/src/ast/constants.rs
```diff
@@ -15,6 +15,7 @@ use crate::{
 // ================================================================================================
 
 /// Represents a constant definition in Miden Assembly syntax, i.e. `const.FOO = 1 + 1`.
+#[derive(Clone)]
 pub struct Constant {
```
bobbinth, line 19: Not related to this PR, but I would love to add the ability to export constants from modules as well. The fact that we cannot do this right now requires pretty ugly workarounds (e.g., we wrap all constants in functions that just put these constants onto the stack).

bobbinth (commented):
Looks good! Thank you! I left a few comments/questions inline - the main one is about the location of the `midenc_hir_type` crate.

package/src/package/manifest.rs
bitwalker: We definitely want to be able to look up exports by name, but more to the point, impls of `PartialOrd`/`Ord` (required for values in a `BTreeSet` _MUST_ match the `PartialEq`/`Eq` implementation, and we can't meaningfully implement `PartialOrd` over a type signature (one of the fields of `PackageExport` - and we may add more fields in the future. Making this a map means we can rely on `PartialOrd`/`Ord` for the procedure name alone (which is really the "key" we're thinking in terms of anyway).

I don't have a problem with making the field private though - it makes the storage of exports an internal detail, and helps with future maintenance, so 👍 from me.

assembly-syntax/src/library/mod.rs
bitwalker: I agree, `LibraryExport` was a natural evolution of the old value, but I do think it is more useful to include the name in its definition, similar to `PackageExport`.

Cargo.toml
bitwalker: > This introduces a bit of a circular dependency between repos, right?

Between repos? I suppose - but not between any crates, as `midenc-hir-type` has no dependencies on any other compiler crates (or VM crates). It is a foundational crate of the compiler though, so most other crates depend on it, directly or indirectly.

> An alternative could be to move miden-hir-type crate into miden-vm repo (maybe under a different name). It seems like it is a more appropriate place for it anyways since it is a part of how Miden assembly is defined (for only the AST part, but in the future, the syntax as well).

I still think it is better left in the compiler repo (at least for now) - the type system is really a compiler IR concept, we're just re-using the IR type system for representing types in the MASM AST to avoid duplicate representations and effort. The compiler is the only thing actually concerned with types anyhow. To the extent we care about them in this repo, it would be for implementing static analyses for improved warnings/errors, and those would be implemented in `miden-assembly`/`miden-assembly-syntax` as well, but we don't have any immediate plans for that anyway.

The compiler on the other hand heavily uses the type system, and maintaining them there makes more sense IMO.

assembly-syntax/src/ast/procedure/procedure.rs
bitwalker: I just don't think there is any good reason to prevent downstream AST producers/consumers from mutating visibility (and that includes crates within this repo).

The previous comment wasn't _wrong_ per se, it is obviously the case that MASM source code always provides us the intended visibility; but that may not be true for other MASM producers. It may also be the case that analysis determines that a visibility change is appropriate, or perhaps the way the AST is constructed means that everything gets a specific visibility at construction time and is then updated at a later point. Bottom line - there isn't any technical reason why we need to keep this crate-local, which is why I'm making it public.

assembly-syntax/src/ast/type.rs
bitwalker: The notion of an enumerated type is a MASM-specific concept (all of the type definitions in this module correspond to syntax nodes, i.e. concrete syntax expected to be surfaced in some way in MASM itself). From the IR type system point of view, an enum type is equivalent to the underlying discriminant value type.

To elaborate a bit further - the IR type system is structural and low-level, but meant to be universal (i.e. any language can represent the type signature of a function written in it, in terms of the IR types expected by the compiled representation of that function). 

On the other hand, the purpose of higher-level type syntax (enumerated types, nominal types) is to provide more semantic information to readers of code that use those types to describe the signature of a procedure (and ultimately, we can provide some static checks to make those types useful for catching errors as well). In MASM, enum types allow expressing two intertwined concepts using a single intuitive syntax (defining a type alias for an integral type with a restricted set of valid values, and defining the constants corresponding to those valid values). Doing it this way keeps those two things in sync, and further, allows us to validate that the constant values are actually valid instances of the given integral type.

Type aliases (a form of nominal types, though we do not treat two nominal types with the same underlying type as unique), are used to provide the semantic information that isn't obvious from a type like `u32`, or `word`. Instead, you get something like `Address` or `AccountId`, and that will tell you (implicitly or explicitly) what is actually expected for a value of that type.

assembly-syntax/src/ast/constants.rs
bitwalker: I did actually add that in this PR, via `Module::constants`. We can add some convenience functions in another PR, e.g. `Module::get_constant` (to look up a constant by name).

package/src/package/manifest.rs
bobbinth: > I don't have a problem with making the field private though - it makes the storage of exports an internal detail, and helps with future maintenance, so 👍 from me.

Should we do this in this PR? Or leave for a follow-up?

assembly-syntax/src/library/mod.rs
```diff
@@ -66,18 +95,19 @@ impl Library {
...
         mast_forest: Arc<MastForest>,
-        exports: BTreeMap<QualifiedProcedureName, MastNodeId>,
+        exports: BTreeMap<QualifiedProcedureName, LibraryExport>,
     ) -> Result<Self, LibraryError> {
         if exports.is_empty() {
             return Err(LibraryError::NoExport);
         }
-        for (fqn, &proc_body_id) in exports.iter() {
-            if !mast_forest.is_procedure_root(proc_body_id) {
+        for (fqn, LibraryExport { node, .. }) in exports.iter() {
```
bobbinth: nit: we could write this as something like:
```rust
for LibraryExport { node, name, .. } in exports.values() {
    ...
}
```

assembly-syntax/src/ast/constants.rs
bobbinth: Ah nice! To use this in MASM (i.e., to import constants from other modules), would we need to change anything in the parser (or anywhere else)?

Cargo.toml
bobbinth: As discussed offline, let's leave as is for now and create an issue (in this repo) to investigate moving out `midenc-hir-type` out of the compiler repo either here or maybe to a stand-alone repo.

bobbinth (approved):
Looks good! Thank you! I left a few small comments/questions inline - but nothing blocking.

assembly-syntax/src/ast/constants.rs
bitwalker: There is some MASM language-level stuff that isn't present, and we don't currently attempt to resolve constants the same way we do procedure imports. So:

* Constants either need to be always public, or we need to modify the grammar to support specifying a visibility for constants. If constants can be exported, we probably should also surface those in `Library`/`PackageManifest`
* We need to disambiguate imports based on what is being imported (a constant or a procedure)
* We need to allow constants to remain unresolved until during assembly (something we probably should do anyway).

It's not a ton of work. You _can_ examine constants defined in a module, you just can't use them from other modules in the syntax at this point.

package/src/package/manifest.rs
bitwalker: I've made the change here

assembly-syntax/src/library/mod.rs
bitwalker: Done

assembly-syntax/src/ast/constants.rs
bobbinth: Makes sense! I'll create an issue to discuss how we could import constants from other modules without having to create procedure wrappers around them.

PR #2029, opened 2025-07-25
feat: Make AdviceInputs fields public

igamigo:
## Changes

Brings https://github.com/0xMiden/miden-vm/pull/2009 to `main`
More context here: https://github.com/0xMiden/miden-base/pull/1639#discussion_r2229722064

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.4 (2025-07-25)
+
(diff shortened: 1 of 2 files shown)


bobbinth:
I made a commit directly to `main` (fc368686bd1e6e171a51a1a5b365ef5400e4b8d5) which is similar to this but without the breaking changes.

The v0.16.4 crates with this change are now on crates.io.

PR #2031, opened 2025-07-26
feat: add drop_stack procedure

### LINKED ISSUE #2030: Create `drop_stack` procedure in `std::sys` module
At the end of the assembly programs we frequently clean the stack to remove the overflowing values using `dropw dropw dropw dropw` instruction sequence. 

For the convenience sake we could create a new `drop_stack` procedure which would do essentially the same but in a more clean way. This procedure should be placed in the `std::sys` module and essentially being just four `dropw`s under the hood.

varun-doshi:
## Describe your changes
Fixes #2030

- Adds `drop_stack` procedure in std::sys

- Updated `CHANGELOG.md'

### DIFF
--- a/stdlib/asm/sys.masm
+++ b/stdlib/asm/sys.masm
@@ -26,3 +26,8 @@ export.truncate_stack.4
     loc_loadw.0
     # => [A, B, C, D, ...]
 end
+
+#! Clean the stack to remove the overflowing values.
+export.drop_stack
+    dropw dropw dropw dropw
+end


varun-doshi:
Currently marked as draft since I could not find `dropw dropw dropw dropw` sequence in the codebase to replace this with and test.
Will proceed as advised.

cc: @Fumuran

stdlib/asm/sys.masm
```diff
@@ -26,3 +26,8 @@ export.truncate_stack.4
     loc_loadw.0
     # => [A, B, C, D, ...]
 end
+
+#! Clean the stack to remove the overflowing values.
```
plafer:
```suggestion
#! Drop 16 values from the stack.
```

plafer (approved):
LGTM, but would be good if @Fumuran you can confirm that this is what you wanted

stdlib/asm/sys.masm
```diff
@@ -26,3 +26,8 @@ export.truncate_stack.4
...
+export.drop_stack
```
bobbinth: Should this maybe be named `drop_stack_top` since we are not dropping the entire stack?

stdlib/asm/sys.masm
```diff
@@ -26,3 +26,8 @@ export.truncate_stack.4
...
+    dropw dropw dropw dropw
+end
```
Fumuran, line 33: nit: let's add a EOF newline.

Fumuran (commented):
Looks good, thank you! I left just one tiny nit inline.

Sorry for the confusion with the usage of this `dropw dropw dropw dropw` instructions sequence, it is commonly used in the `miden-base`, but rarely in the VM. But I think that this procedure is simple enough so we can leave it without tests.

Also I agree with  @bobbinth that it will be better to rename this procedure into the `drop_stack_top`, since we could have some other data deeper on the stack, which won't be removed (which is fine).

varun-doshi:
> it is commonly used in the miden-base

Would it be good idea to make the changes in `miden-base` once this PR is in? I can take up that task

PR #2033, opened 2025-07-28
draft: handle `Sync` requirements around `dyn SourceManager` for multithreaded executors

drahnr:
## Describe your changes

The current approach is making the `Sync` bound conditional on the `target_arch`, but it has implications in `miden-base`, i.e. `impl DataStore` is captured in closure scope and hence also needs to have `Sync`.

The next steps: Attempt to use a generic and see if this fares any better. 

Note: Previously https://github.com/0xMiden/miden-vm/pull/2027 which I can't seem to re-open.

Most relevant discussion items:

* https://github.com/0xMiden/miden-vm/pull/2027#discussion_r2228480630
* https://github.com/0xMiden/miden-vm/pull/2027#discussion_r2234137549

### DIFF
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -242,7 +242,7 @@ pub trait SourceManagerExt: SourceManager {
             .map(|s| SourceContent::new(lang, uri.clone(), s))
             .map_err(|source| {
                 SourceManagerError::custom_with_source(
-                    format!("failed to load filed at `{}`", path.display()),
+                    alloc::format!("failed to load filed at `{}`", path.display()),
                     source,
                 )
             })?;
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -242,7 +242,7 @@ pub trait SourceManagerExt: SourceManager {
             .map(|s| SourceContent::new(lang, uri.clone(), s))
             .map_err(|source| {
                 SourceManagerError::custom_with_source(
-                    format!("failed to load filed at `{}`", path.display()),
+                    std::format!("failed to load filed at `{}`", path.display()),
                     source,
                 )
             })?;
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -262,18 +262,9 @@ impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 /// [SourceManager] is a supertrait of [SourceManagerSync], so you may use instances of the
 /// [SourceManagerSync] where the [SourceManager] is required, either implicitly or via explicit
 /// downcasting, e.g. `Arc<dyn SourceManagerSync> as Arc<dyn SourceManager>`.
-
-#[cfg(target_arch = "wasm32")]
 pub trait SourceManagerSync: SourceManager + Send + Sync {}
 
-#[cfg(target_arch = "wasm32")]
 impl<T: ?Sized + SourceManager + Send + Sync> SourceManagerSync for T {}
-
-#[cfg(not(target_arch = "wasm32"))]
-pub trait SourceManagerSync: SourceManager + Send {}
-
-#[cfg(not(target_arch = "wasm32"))]
-impl<T: ?Sized + SourceManager + Send> SourceManagerSync for T {}
 
 // DEFAULT SOURCE MANAGER
 // ================================================================================================
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -242,7 +242,7 @@ pub trait SourceManagerExt: SourceManager {
             .map(|s| SourceContent::new(lang, uri.clone(), s))
             .map_err(|source| {
                 SourceManagerError::custom_with_source(
-                    std::format!("failed to load filed at `{}`", path.display()),
+                    format!("failed to load filed at `{}`", path.display()),
                     source,
                 )
             })?;
@@ -254,8 +254,6 @@ pub trait SourceManagerExt: SourceManager {
 #[cfg(feature = "std")]
 impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 
-// TODO rename to reflect the fact it's a "glue" trait that adapts per target, but only after
-// the `-base` has been validated to work with this approach.
 /// [SourceManagerSync] is a marker trait for [SourceManager] implementations that are also Send +
 /// Sync, and is automatically implemented for any [SourceManager] that meets those requirements.
 ///
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -260,9 +260,18 @@ impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 /// [SourceManager] is a supertrait of [SourceManagerSync], so you may use instances of the
 /// [SourceManagerSync] where the [SourceManager] is required, either implicitly or via explicit
 /// downcasting, e.g. `Arc<dyn SourceManagerSync> as Arc<dyn SourceManager>`.
+
+#[cfg(feature = "syncbound")]
 pub trait SourceManagerSync: SourceManager + Send + Sync {}
 
+#[cfg(feature = "syncbound")]
 impl<T: ?Sized + SourceManager + Send + Sync> SourceManagerSync for T {}
+
+#[cfg(not(feature = "syncbound"))]
+pub trait SourceManagerSync: SourceManager + Send {}
+
+#[cfg(not(feature = "syncbound"))]
+impl<T: ?Sized + SourceManager + Send> SourceManagerSync for T {}
 
 // DEFAULT SOURCE MANAGER
 // ================================================================================================
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -254,6 +254,8 @@ pub trait SourceManagerExt: SourceManager {
 #[cfg(feature = "std")]
 impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 
+// TODO rename to reflect the fact it's a "glue" trait that adapts per target, but only after
+// the `-base` has been validated to work with this approach.
 /// [SourceManagerSync] is a marker trait for [SourceManager] implementations that are also Send +
 /// Sync, and is automatically implemented for any [SourceManager] that meets those requirements.
 ///
--- a/crates/debug/types/src/source_manager.rs
+++ b/crates/debug/types/src/source_manager.rs
@@ -261,16 +261,16 @@ impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 /// [SourceManagerSync] where the [SourceManager] is required, either implicitly or via explicit
 /// downcasting, e.g. `Arc<dyn SourceManagerSync> as Arc<dyn SourceManager>`.
 
-#[cfg(feature = "syncbound")]
+#[cfg(target_arch = "wasm32")]
 pub trait SourceManagerSync: SourceManager + Send + Sync {}
 
-#[cfg(feature = "syncbound")]
+#[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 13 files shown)


processor/src/host/mod.rs
```diff
@@ -114,13 +127,30 @@ pub trait AsyncHost: BaseHost {
...
+    ) -> impl FutureAliasWrapper<Result<Vec<AdviceMutation>, EventError>>;
 }
+
+/// Alias for a `Future`
+///
+/// Depending on the the target being `wasm32-unknown-unknown` the
+/// trait bounds are _without_ `+ Send`, otherwise with `+ Send` to
+/// ensure functionality for mutli threaded executor.
+#[cfg(target_arch = "wasm32")]
+pub trait FutureAliasWrapper<O>: Future<Output = O> {}
```
PhilippGackstatter: Nit: Since we're using `target_arch = "wasm32"` I wouldn't describe this to only apply to `wasm32-unknown-unknown`. It would also apply to `wasm32-wasip1` or other targets.

PhilippGackstatter: I think I would call this `AsyncHostFuture` which seems more precise for the scope to which it applies, i.e. the `AsyncHost` trait. Since it's not more generic than that, I also wouldn't make the name more generic than necessary.

PhilippGackstatter: I assume we'll need `+ 'static` here, but since that would have quite an impact on the codebase, I'd just try it as is and see whether that ends up working in the client, before going through that work.

drahnr: Yeah, that'd be the next step, but I am not quite that far yet. Making `-base` compile is still not quite achieved.

drahnr: Does wasi preview 1 have a threading API ? I thought it didn't and hence https://doc.rust-lang.org/rustc/platform-support/wasm32-wasip1-threads.html existing? I am not sure we want to be overspecific until this is both standardized and implemented. I'd rather default to single threading to i.e. run wasm outside the browser with identical behaviour.
No strong opinion though.

bitwalker: Is there any reason not to use `#[cfg(target_family = "wasm")]` here? Whether "threading" is available or not is much more contingent on whether the `atomics` target feature is enabled anyway. None of the Wasm targets currently (and probably any time in the near future) will ever have threading "out of the box" - so IMO we should use `target_family = "wasm"` to gate this, to ensure that regardless of the Wasm target used, the behavior is consistent.

If we find that there is some use case involving a Wasm target where threading is available, we can cross that bridge when we get to it (I suspect it will involve something like `#[cfg(all(target_family = "wasm", not(target_feature = "foo")))]` where the `foo` target feature indicates that threading is enabled.

bitwalker: I should also note that `wasm32-wasip1` is effectively deprecated, `wasm32-wasip2` is the current WASI target, and `wasm32-wasip3` will become the current WASI target sometime in the fall, unless something changes.

There are also `wasm64-*` targets, but those are mostly experimental, so not something we need to actually concern ourselves with - but figured it is worth pointing out that the set of Wasm targets in Rust is actually quite varied.

processor/src/host/mod.rs
```diff
@@ -114,13 +127,30 @@ pub trait AsyncHost: BaseHost {
     fn get_mast_forest(
         &self,
         node_digest: &Word,
-    ) -> impl Future<Output = Option<Arc<MastForest>>> + Send;
+    ) -> impl FutureAliasWrapper<Option<Arc<MastForest>>>;
```
bobbinth: Should we change this back to a non-async function for now? Something like:
```rust
fn get_mast_forest(&self, node_digest: &Word) -> Option<Arc<MastForest>>;
```

processor/src/host/mod.rs
bobbinth: We currently have the following statement in our readme:
```
WARNING: For no_std, only the wasm32-unknown-unknown and wasm32-wasip1 targets are officially supported.
```
So, why not just make this gated by `no_std`? Basically, we already assume that if you compile in `no_std` context, you are targeting WASM - so, we could do the same thing here as well.

Separately, seems like we should now replace `wasm32-wasip1` with `wasm32-wasip2`- though, aside from updating the readme, I'm not sure there is much to do here.

bobbinth: I like `AsyncHostFuture` too.

processor/src/host/advice/mod.rs
```diff
@@ -41,9 +41,9 @@ type SimpleMerkleMap = BTreeMap<Word, StoreNode>;
 /// Advice data is store in-memory using [BTreeMap]s as its backing storage.
 #[derive(Debug, Clone, Default)]
 pub struct AdviceProvider {
-    stack: Vec<Felt>,
-    map: AdviceMap,
-    store: MerkleStore<SimpleMerkleMap>,
+    pub stack: Vec<Felt>,
+    pub map: AdviceMap,
+    pub store: MerkleStore<SimpleMerkleMap>,
 }
```
bobbinth: Why do we need to make this public? We may refactor the advice provider such that it would require some additional logic on how the internal data could be accessed or modified. So, I'd rather keep the fields private.

crates/debug/types/src/source_manager.rs
```diff
@@ -254,16 +254,27 @@ pub trait SourceManagerExt: SourceManager {
...
 pub trait SourceManagerSync: SourceManager + Send + Sync {}
 
+#[cfg(target_arch = "wasm32")]
 impl<T: ?Sized + SourceManager + Send + Sync> SourceManagerSync for T {}
 
+#[cfg(not(target_arch = "wasm32"))]
+pub trait SourceManagerSync: SourceManager + Send {}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<T: ?Sized + SourceManager + Send> SourceManagerSync for T {}
```
bobbinth: Do we still need these changes in this PR?

bobbinth (commented):
Looks good! Thank you. I left some small comments/questions inline.

processor/src/host/mod.rs
drahnr: Yes, I do have a branch for this, pending integration

crates/debug/types/src/source_manager.rs
```diff
@@ -254,6 +254,8 @@ pub trait SourceManagerExt: SourceManager {
 #[cfg(feature = "std")]
 impl<T: ?Sized + SourceManager> SourceManagerExt for T {}
 
+// TODO rename to reflect the fact it's a "glue" trait that adapts per target, but only after
+// the `-base` has been validated to work with this approach.
```
bobbinth: I think this `TODO` is now outdated. Maybe we should roll-back all changes in this file?

processor/src/host/mod.rs
bitwalker: There isn't a way to feature-gate on `#![no_std]` - you can use either target info or Cargo features, that's it.

The `wasm32-wasip2` target produces Wasm Component Model binaries, whereas `wasm32-wasip1` produces Core Wasm binaries, that's the main difference between those two targets. The `wasip1` variant is then typically wrapped up into a Component Model binary using external tooling. In both cases, the runtime is generally expected to be a Component Model one. I don't know that this really means anything for the VM either way though, AFAIK we're not targeting either `wasip1` OR `wasip2` anywhere yet.

processor/src/host/mod.rs
```diff
@@ -111,16 +124,30 @@ pub trait AsyncHost: BaseHost {
 
     /// Returns MAST forest corresponding to the specified digest, or None if the MAST forest for
     /// this digest could not be found in this [AsyncHost].
-    fn get_mast_forest(
-        &self,
-        node_digest: &Word,
-    ) -> impl Future<Output = Option<Arc<MastForest>>> + Send;
+    fn get_mast_forest(&self, node_digest: &Word) -> Option<Arc<MastForest>>;
```
plafer, line 127: Why is this no longer async? I think we want to use `AsyncHostFuture` right?

processor/src/host/mod.rs
```diff
@@ -111,16 +124,30 @@ pub trait AsyncHost: BaseHost {
...
         event_id: u32,
-    ) -> impl Future<Output = Result<Vec<AdviceMutation>, EventError>> + Send;
+    ) -> impl AsyncHostFuture<Result<Vec<AdviceMutation>, EventError>>;
 }
+
+/// Alias for a `Future`
+///
+/// Depending on the the target being `wasm32-unknown-unknown` the
+/// trait bounds are _without_ `+ Send`, otherwise with `+ Send` to
+/// ensure functionality for mutli threaded executor.
```
plafer: This comment is outdated - we not use `std` feature. But as discussed at some point offline, is it correct to depend on `std`? Basically AFAIU if any dependency depends on `std`, then the final binary will be compiled with `std` (due to cargo features being additive). This was my understanding why using `#[cfg(target_family = "wasm")]` as suggested by @bitwalker in https://github.com/0xMiden/miden-vm/pull/2033#discussion_r2240040800 was preferred

processor/src/host/mod.rs
drahnr: `AsyncHostFuture` ripples this all through `-base` and causes `&'store STORE` to be required to be `Sync`, since we return an `Arc<MastForest>`, which does not hold to be `Sync`. The intent is to do this the next PR/a follow-up but test against `-base` and `-client` first.

crates/debug/types/src/source_manager.rs
drahnr: Removed

processor/src/host/mod.rs
drahnr: If we only have `std` = `host`, and `wasm*-*-*` as `no_std`, then it's a matter of usage.

The feature based approach requires `default-features = false`, the target specific doesn't. But since `feature = "std"` already exists, I don't think it makes a difference.

processor/src/host/mod.rs
bobbinth: Yes, as @drahnr said, this is a temporary change to minimize the amount of downstream changes.

processor/src/host/mod.rs
bitwalker: As @plafer pointed out, using `#[cfg(feature = "...")]` for conditional compilation of trait bounds is problematic due to Cargo's feature unification. The feature could be enabled unexpectedly. I know there is some work in Cargo to allow for disjoint Cargo features, but it is not stable, or even close yet AIUI.

Technically, depending on libstd for the Wasm targets is possible. In practice, one typically compiles as `#[no_std]` to avoid calling APIs that will fail at runtime on the target machine, and to avoid linking in stuff that won't be used, but it isn't strictly required to do so. So you might have a crate in the crate graph that doesn't use `default-features = false`, or explicitly enables the `std` feature, and then compilation is going to fail because these trait bounds are probably not going to line up across crates.

Making the trait bounds target-specific doesn't present this problem because regardless of Cargo configuration, the results are always deterministic.

bobbinth (approved):
Looks good! Thank you. I think the only thing left from my end is https://github.com/0xMiden/miden-vm/pull/2033#discussion_r2240258054

processor/src/host/mod.rs
bobbinth: I guess my point was that we already have an explicit assumption: `no_std` is to be used only if we are targeting WASM and trying to compile to WASM without enabling `no_std` or trying to enable `no_std` for non-WASM targets is not supported (I think we settled on this after adding our own version of `no_std` locking primitives). So, to me both are equivalent.

I think we could make this even more explicit by using something like `#[cfg(all(target_family = "wasm", not(feature = "std")))]` - but I'd do this in a separate PR.

plafer: I think in theory given our docs they are equivalent - but in practice, you only need one crate in the dependency tree to enable `std` (even "by mistake") for this to fail. So although equivalent per our docs, depending on `std` is more brittle.

I'm happy to try it out - I just won't be surprised if someone ends up opening an issue about their build failing when targeting wasm due to this issue.

Issue #2046, opened 2025-08-01
Document `emit` in instruction reference


PhilippGackstatter:
The `emit` instruction is not documented in the [instruction reference](https://0xmiden.github.io/miden-docs/imported/miden-vm/src/user_docs/assembly/instruction_reference.html), but it would be nice if it was.

bobbinth:
Closed by https://github.com/0xMiden/miden-vm/pull/2047.

PR #2047, opened 2025-08-04
docs: add emit instruction to instruction reference

MozirDmitriy:
Add documentation for the emit instruction to the Miden VM instruction reference.

The emit instruction was already documented in the events.md file but was missing
from the comprehensive instruction reference. This commit adds complete
documentation including:

- Syntax: emit.<event_id>
- Stack behavior: no change to operand stack
- Cycle count: 5 cycles
- Detailed notes explaining purpose and usage
- Examples of usage with direct values and named constants

This resolves GitHub issue #2046 which requested documentation for the emit
instruction in the instruction reference.

The documentation is based entirely on existing code analysis and follows the
same format as other instructions in the reference.

### DIFF
--- a/docs/src/user_docs/assembly/instruction_reference.md
+++ b/docs/src/user_docs/assembly/instruction_reference.md
@@ -321,4 +321,17 @@ Instructions for inspecting VM state during execution. These do not affect VM st
 - **Cycles:** 0 (does not consume VM cycles).
 - **Notes:**
     - Prints the specified part of the VM state.
-    - Ignored if assembler is not in debug mode. 
+    - Ignored if assembler is not in debug mode.
+
+### `emit`
+
+- **Syntax:** `emit.<event_id>`
+- **Stack Input:** `[...]`
+- **Stack Output:** `[...]`
+- **Cycles:** 5
+- **Notes:**
+    - Emits an event with the specified `event_id` to the host.
+    - Does not change the state of the operand stack.
+    - The `event_id` can be any 32-bit value specified either directly or via a [named constant](./code_organization.md#constants).
+    - Events allow programs to communicate contextual information to the host for triggering appropriate actions.
+    - Example: `emit.123` or `emit.EVENT_ID_1` 


docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -321,4 +321,17 @@ Instructions for inspecting VM state during execution. These do not affect VM st
 - **Cycles:** 0 (does not consume VM cycles).
 - **Notes:**
     - Prints the specified part of the VM state.
-    - Ignored if assembler is not in debug mode. 
+    - Ignored if assembler is not in debug mode.
+
+### `emit`
```
bobbinth: `emit` is not really a debug instruction - so, it should go into a different section. Which section is a good question. Maybe Inputs/Outputs section is the right one? @plafer - what do you think?

plafer: Yeah Input/Outputs sounds right. Although I'm not sure why PUSH is in there?

bobbinth: > Although I'm not sure why PUSH is in there?

The original idea was that `push` puts data onto the stack. But I'm open moving it to another section.

MozirDmitriy: > Yeah Input/Outputs sounds right. Although I'm not sure why PUSH is in there?

corrected

bobbinth: Actually, thinking about this more, maybe it should go into a separate section. We currently have the detailed description of the instruction [here](https://0xmiden.github.io/miden-vm/user_docs/assembly/events.html), so, having "Events" section (before the "Debugging Operations" section) would probably be more consistent.

MozirDmitriy: > Actually, thinking about this more, maybe it should go into a separate section. We currently have the detailed description of the instruction [here](https://0xmiden.github.io/miden-vm/user_docs/assembly/events.html), so, having "Events" section (before the "Debugging Operations" section) would probably be more consistent.

Also done

bobbinth (approved):
Looks good! Thank you!

Issue #1218, opened 2024-01-29
`stdlib smt`: add support for leaves with multiple key/value pairs


plafer:
#1215 replaced the `TieredSmt` implementation with the new `Smt`. It only added support for empty and single leaves though. We still need to implement the case where multiple key-value pairs are stored in a leaf.

bobbinth:
Closed by https://github.com/0xMiden/miden-vm/pull/2048 and https://github.com/0xMiden/miden-vm/pull/2248.

Issue #2039, opened 2025-07-30
Create a `pipe_preimage_to_memory` for double words


Fumuran:
We should consider creating a new version of the `pipe_preimage_to_memory` procedure optimized for the double words, something like `pipe_double_words_preimage_to_memory`. This will decrease number of cycles required to store some advice data to memory in case we know that its length divides by 8.

It could be already used in `miden-base`, for example in the `miden::note::write_assets_to_memory` procedure.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-base/pull/1648#discussion_r2241225043_

gswirski:
This was implemented and merged here: https://github.com/0xMiden/miden-vm/pull/2048/files#diff-9ca2991c284a6d192772c69a86261a24b89d8220442dbcb0d9e4e223accc541fR190

We can probably close this issue, @Fumuran

bobbinth:
Closed by #2048.

PR #2048, opened 2025-08-04
stdlib: implement `smt::get` for leaves with multiple pairs

Qyriad:
## Describe your changes

One half of #1218, this adds support for _getting_ the value for a key in the standard library's sparse Merkle tree, where the key corresponds to a leaf which also has other key-value pairs.

- Updated `CHANGELOG.md'


.git-blame-ignore-revs
bobbinth, line 1: I'm not sure we need this file.

stdlib/asm/collections/smt.masm
```diff
@@ -370,7 +446,7 @@ end
 #! Cycles
 #!   Leaf empty: 48 cycles
 #!   Leaf single: 99 cycles
-#!   Leaf multiple: unimplemented
+#!   Leaf multiple: 184 cycles
```
bobbinth: The cycle count is not fixed at 184, right? It should be variable based on the number of key-value pairs in the leaf.

stdlib/asm/collections/smt.masm
```diff
@@ -431,9 +507,19 @@ export.get
...
             # => [V, R]
         else
-            # Multiple kv-pair case
-            # TODO (fail for now)
-            push.1 assertz
+            # => [leaf_size, NV, K, R]
+
+            # We don't care about the node's hash anymore.
+            movdn.12 dropw
+            # => [K, R, leaf_size]
```
bobbinth: I think we probably need the node value - otherwise we won't be able to verify that the data from the advice provider is correct (advice provider is an untrusted component and, in theory, could "lie").

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +353,82 @@ end
 # GET
 # =================================================================================================
 
+
+#! Inputs:
+#!   Operand stack: [pair_count, K, R]
+#!   Advice stack:  [V1, K2, V2, K2, ...]
+proc.get_multi_leaf_value
```
bobbinth: I haven't reviewed this in detail, but I think we could use a much simpler approach here. It could work something like this:

1. We "unhash" the node value into memory. For this, we could use `std::mem::pipe_preimage_to_memory` procedure, or maybe a more efficient version of it that relies on `pipe_double_words_to_memory` directly as we know that key-value pairs are always 2 words long.
  a. We actually have an issue for adding such a procedure to the standard library: https://github.com/0xMiden/miden-vm/issues/2039
2. Once we all key-value pairs laid out in memory, we can use non-deterministic searching to find the position where the key-value pair we are interested is located, or should be located in case it is not present.
  a. This would require adding a new instruction to the VM - something similar to `adv.push_smtpeek`, but much simpler. Let me know if you'd like help with this.
4. Then we verify that the position from the previous step is correct. This would require two word comparisons.

bobbinth (commented):
Thank you! I left a few comments inline - the main one is that I think we should take a different approach to how this procedure works. This approach would rely on non-determinism and should be much more efficient than the current approach.

Al-Kindi-0:
Pretty much agree with @bobbinth . I would just add that we would probably need to have the edge cases of first and last position handled separately, in addition to the typical middle position mentioned by Bobbin above.

stdlib/asm/collections/smt.masm
Qyriad: For the new instruction, are you thinking something like this?
> Inputs:
> > Operand stack: \[K\]
>>  Advice stack: \[K1, V1, Kn, Vn, …\]
>
> Outputs:
>> Operand stack: \[index where K == $K_n$, K\]
>> Advice stack: [K1, V1, Kn, Vn, …\]

bobbinth: I think the new instruction would need to take the key and the memory region as input and put the memory address of the found key and the flag indicating whether the key was found onto the advice stack. So, it would look roughly like this:

```
Inputs:
    Operand stack: [KEY, start_ptr, num_entries]
    Advice stack: [] - it doesn't really matter whats on the advice stack

Outputs:
    Operand stack: [KEY, start_ptr, end_ptr] - the operand stack does not change
    Advice stack: [was_key_found, key_ptr]
```

The instruction will assume that `start_ptr` is the start of memory region where they key-value pairs of a given leaf are laid out in memory and `num_entries` is the number of key-value pairs in this leaf. The returned values will indicate wither the key was found in this region (assuming the key is the first word in the tuple of words for each entry), and the memory pointer to the key if it was found, or the pointer to the place where the key should be inserted, if it wasn't found.

There may be some edge cases I'm not considering in the above - but I think the general approach should work.

Qyriad:
I've added `pipe_double_words_preimage_to_memory` (#2039), so now this is just waiting on #2114.

bobbinth:
@Qyriad - #2114 has now been merged. Do we need to merge the latest `next` into this PR?

Qyriad:
@bobbinth just did! This is ready for review now!

bobbinth:
> @bobbinth just did! This is ready for review now!

Great! Thank you! Seems like there are still some merge conflicts, could you try to resolve them?

@Al-Kindi-0, @plafer - could you do a first review pass?

crates/utils/testing/src/lib.rs
```diff
@@ -548,6 +548,12 @@ pub fn append_word_to_vec(target: &mut Vec<u64>, word: Word) {
     target.extend(word.iter().map(Felt::as_int));
 }
 
+/// Note that adding a word to the *beginning* of a Vec adds it to the *bottom* of the stack.
```
Al-Kindi-0: nit: I would write this comment in the same way we do for others, i.e., something like "Adds a word to ..."

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +357,56 @@ end
...
+#! Cycles: 124 + 3 * pair_count
+proc.get_multi_leaf_value
+    # First, we'll pipe all the pairs in the leaf from the advice stack.
+    # We'll need a pointer, the count of words to copy, and a "commitment" that it should hash to.
+    # The node value (its hash) is our commitment, and we can just divide `leaf_size` to get the
+    # word count.
+
+    # (2 cycles).
+    div.4
+    # => [num_words, COM, K, R]
```
Al-Kindi-0, line 388: NV is COM? If so, I would state that. Also would be good to add a few words in the procedure comment to explain things a bit.

stdlib/asm/collections/smt.masm
```diff
@@ -431,9 +485,12 @@ export.get
...
             # => [V, R]
         else
-            # Multiple kv-pair case
-            # TODO (fail for now)
-            push.1 assertz
+            # => [leaf_size, NV, K, R]
+
+            # Cycles: 124 + 3 * leaf_size / 4
+            exec.get_multi_leaf_value
+
```
Al-Kindi-0: nit: I would remove the newline

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +357,56 @@ end
...
+    # (3 cycles)
+    movdn.4 push.0xA0 movdn.4
+    # => [K, start_ptr, end_ptr, R]
+
+    # (56 cycles)
+    exec.sorted_array::find_key_value
+    # => [is_key_found, key_ptr, start_ptr, end_ptr, R]
+
+    # is_key_found will always be true, so we can reuse that bit of stack. Likewise we can reuse
```
Al-Kindi-0: Q: Why is this the case? I feel I am missing some context here. Also, if this is the case, then I would add an assertion, accepting the cycle cost for the resulting robustness

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
...
+#! Moves an even number of words from the advice stack to memory and asserts it matches the commitment.
+#!
+#! Input: [num_words, write_ptr, COM, ...]
+#! Output: [write_ptr', ...]
+#!
+#! Cycles: 56 + 3 * num_words / 2
+# Possible other name: pipe_pairs_preimage_to_memory
+export.pipe_double_words_preimage_to_memory
+  # Assert precondition (8 cycles).
+  # TODO: how precious are those cycles? Should we remove this assert?
```
Al-Kindi-0: I would say this is a negligible fraction of the total cost and it is worth that cost, in my opinion

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
...
 end
+
+#! Moves an even number of words from the advice stack to memory and asserts it matches the commitment.
+#!
+#! Input: [num_words, write_ptr, COM, ...]
+#! Output: [write_ptr', ...]
+#!
+#! Cycles: 56 + 3 * num_words / 2
+# Possible other name: pipe_pairs_preimage_to_memory
+export.pipe_double_words_preimage_to_memory
```
Al-Kindi-0, line 190: Other suggestion is: pipe_dword_check_preimage (the thinking is that pipe implies to memory)

stdlib/asm/mem.masm
Al-Kindi-0: Another option is to give the count in terms of double-words, which seems more logical to me, and hence avoid this check provided we do it in the calling function.
Also, in get_multi_leaf_value and when dividing by 4, are we sure things are divisible by 4?

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +357,56 @@ end
...
+    # First, we'll pipe all the pairs in the leaf from the advice stack.
+    # We'll need a pointer, the count of words to copy, and a "commitment" that it should hash to.
+    # The node value (its hash) is our commitment, and we can just divide `leaf_size` to get the
+    # word count.
+
+    # (2 cycles).
+    div.4
+    # => [num_words, COM, K, R]
+
+    # FIXME: is there any particular value I should use for the pointer?
```
Al-Kindi-0: You can use locaddr.0. This will require that get_multi_leaf_value declares usage of local procedure addresses and we can fix this number to the max leaf size we want to support.

Otherwise, you can use any address beyond `3 * 2^30` (see [here](https://0xmiden.github.io/miden-vm/user_docs/assembly/execution_contexts.html))

stdlib/asm/mem.masm
Al-Kindi-0: or to be consistent with the rest pipe_double_words_to_memory_check_preimage, though I prefer the first suggestion and we can change pipe_double_words_to_memory to match this (@bobbinth ?)

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
...
+  # => [write_ptr, end_ptr, COM, ...]
+
+  # Setup the initial hasher state, which starts with a capacity word `A`.
+  # For us, the capacity word is entirely empty.
+  # (12 cycles).
+  padw padw padw
+  # => [C, B, A, write_ptr, end_ptr, COM, ...]
+
+  # Cycles: 9 + 3 * num_words
+  # (e.g., 25 cycles for 4 words)
```
Al-Kindi-0: nit: I would stay consistent here, e.g.,
```
# Some comment
# (9 + 3 * num_words cycles)
```

stdlib/tests/mem/mod.rs
```diff
@@ -235,3 +235,78 @@ fn test_pipe_preimage_to_memory_invalid_preimage() {
     let res = build_test!(three_words, operand_stack, &advice_stack).execute();
     assert!(res.is_err());
 }
+
+#[test]
+fn test_pipe_double_words_preimage_to_memory() {
+    let mem_addr = 1000;
```
Al-Kindi-0, line 242: nit: I would add a comment that this is word-aligned

stdlib/asm/collections/smt.masm
Qyriad: No, actually I was mistaken. `is_key_found` will be false if we should be returning an empty word. I'll fix that.

stdlib/asm/collections/smt.masm
Qyriad: Is there an already-decided maximum leaf size?

stdlib/asm/collections/smt.masm
Qyriad: It's stated above: https://github.com/0xMiden/miden-vm/blob/816351ba229ea20959381dbc183f5b548ab6daaf/stdlib/asm/collections/smt.masm#L368

But not clearly enough! Will fix.

stdlib/asm/mem.masm
Qyriad: By `get_multi_leaf_value` it's already assumed that `leaf_size` is a multiple of 8: https://github.com/reilabs/miden-vm/blob/74ca5db776ff4a9122e51c244d34336c1098a204/stdlib/asm/collections/smt.masm#L474-L480

The only way it could not be a multiple of 4 is if someone inserted non-word-sized values into the advice map *over* the Merkle hash.

CHANGELOG.md
```diff
@@ -87,6 +91,15 @@
...
 - Added `drop_stack_top` procedure in `std::sys` ([#2031](https://github.com/0xMiden/miden-vm/pull/2031)).
+- Implement `copy_digest` and `hash_memory_double_words` procedures in the `std::crypto::hashes::rpo` module ([#1971](https://github.com/0xMiden/miden-vm/pull/1971)).
+- Add `extend_` methods on AdviceProvider [#1982](https://github.com/0xMiden/miden-vm/pull/1982).
+- Add new stdlib module `std::word`, containing utilities for manipulating arrays of four fields (words) [#1996](https://github.com/0xMiden/miden-vm/pull/1996).
+- [BREAKING]: move `std::utils::is_empty_word` to `std::word::eqz`, as part of the new word module [#1996](https://github.com/0xMiden/miden-vm/pull/1996).
+- Add constraints evaluation check to recursive verifier ([#1997](https://github.com/0xMiden/miden-vm/pull/1997)).
+- Add `AdviceProvider::into_parts()` method ([#2024](https://github.com/0xMiden/miden-vm/pull/2024)).
+- Add `drop_stack_top` procedure in std::sys ([#2031](https://github.com/0xMiden/miden-vm/pull/2031))
+- Add type information to procedures in the AST, `Library`, and `PackageExport`
+  types ([#2028](https://github.com/0xMiden/miden-vm/pull/2028))
```
bobbinth: I think this is may not have been rebased correctly - we shouldn't be adding any changes to v0.17 releases.

stdlib/asm/mem.masm
```diff
@@ -65,17 +65,17 @@ end
 #!     - C,B are the rate portion of the state
 #! - The value `words = end_ptr - write_ptr` must be positive and a multiple of 8
 #!
-#! Cycles: 10 + 9 * word_pairs
+#! Cycles: 9 + 6 * word_pairs
 export.pipe_double_words_to_memory.0
-  dup.13 dup.13 neq # (4 cycles)
+  dup.13 dup.13 neq # (3 cycles)
```
bobbinth: `neq` takes 2 cycles - so, I think `4 cycles` was correct here.

stdlib/asm/mem.masm
```diff
@@ -65,17 +65,17 @@ end
...
 
-  # loop until write_ptr reaches end_ptr (9 cycles per iteration + 1)
+  # loop until write_ptr reaches end_ptr (6 cycles per iteration + 1)
   # LOOP: [b, C, B, A, write_ptr, end_ptr, ...]
   while.true
     adv_pipe hperm # (2 cycles)
     # => [C', B', A', write_ptr', end_ptr, ...]
 
-    dup.13 dup.13 neq # (4 cycles)
+    dup.13 dup.13 neq # (3 cycles)
```
bobbinth: Same comment as above.

stdlib/asm/mem.masm
bobbinth: I actually don't mind `pipe_double_words_preimage_to_memory` - it is consistent with `pipe_preimage_to_memory` and `pipe_double_words_to_memory`.

We could rename `pipe_preimage_to_memory` to `pipe_word_preimage_to_memory` to make things even more consistent, but I'm not sure it's worth it.

stdlib/asm/mem.masm
bobbinth: I'd keep the assert - I think it is worth it because this procedure can be used in other contexts as well.

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
...
+  # => [num_words, write_ptr, COM, ...]
+
+  # Compute and move end_ptr (5 cycles)
+  mul.4 dup.1 add swap
+  # => [write_ptr, end_ptr, COM, ...]
+
+  # Setup the initial hasher state, which starts with a capacity word `A`.
+  # For us, the capacity word is entirely empty.
+  # (12 cycles).
+  padw padw padw
```
bobbinth: nit: I would use `exec.rpo::init_no_padding` here.

stdlib/asm/collections/smt.masm
bobbinth: Yes, I believe we allow at most 1024 entries per leaf. So, we'd need to allocate `8192` locals (this is fine as it doesn't cost us to allocate locals unless we actually use them).

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +357,73 @@ end
...
 # =================================================================================================
 
+#! Inputs:
+#!   Operand stack: [leaf_size, NV, K, R]
+#!   Advice stack:  [V1, K2, V2, K2, ...]
+#!
+#! Cycles:
+#!  Non-empty: 125 + 3 * pair_count
+#!  Empty: 129 + 3 * pair_count
+proc.get_multi_leaf_value
```
bobbinth: Let's add a brief description of this procedure in doc comments.

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +357,73 @@ end
...
+    # First, we'll pipe all the pairs in the leaf from the advice stack.
+    # We'll need a pointer, the count of words to copy, and a "commitment" that it should hash to.
+    # The node value (its hash) is our commitment, and we can just divide `leaf_size` to get the
+    # word count.
+
+    # => [leaf_size, COM=NV, K, R]
+
+    # (2 cycles).
+    div.4
```
bobbinth, line 387: I wonder if we should validate if leaf_size is in fact divisible by 8. We could do it here or in `get` (around line 479).

CHANGELOG.md
Qyriad: Yep, that's not right. Fixed!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 #### Enhancements
 
+- Added support for leaves with multiple pairs in `std::collections::smt::get` ([#2048](https://github.com/0xMiden/miden-vm/pull/2048)).
+
+#### Enhancements
```
bobbinth: The "Enhancement" title is duplicated (same as on line 5 above).

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 #### Enhancements
 
+- Added support for leaves with multiple pairs in `std::collections::smt::get` ([#2048](https://github.com/0xMiden/miden-vm/pull/2048)).
```
bobbinth, line 12: Maybe as a separate line I'd mention the new memory hashing procedure too.

stdlib/asm/mem.masm
bobbinth: Let's remove this TODO.

stdlib/tests/collections/smt.rs
```diff
@@ -1,23 +1,35 @@
...
+];
+
+/// Unlike the above `LEAVES`, these leaves use the same value for their most-significant felts, to
+/// test leaves with multiple pairs.
+const LEAVES_MULTI: [(Word, Word); 3] = [
+    (word(101, 102, 103, 69420), word(0x1, 0x2, 0x3, 0x4)),
+    // Most significant felt does NOT differ from previous.
+    (word(201, 202, 203, 69420), word(0xb, 0xc, 0xd, 0xe)),
+    // A key in the same leaf, but with no corresponding value.
+    (word(301, 302, 303, 49420), EMPTY_WORD),
```
bobbinth: For this to be int eh same leaf, should the most significant element also be `69420` here?

stdlib/asm/mem.masm
```diff
@@ -65,11 +65,11 @@ end
 #!     - C,B are the rate portion of the state
 #! - The value `words = end_ptr - write_ptr` must be positive and a multiple of 8
 #!
-#! Cycles: 10 + 9 * word_pairs
+#! Cycles: 9 + 6 * word_pairs
 export.pipe_double_words_to_memory.0
   dup.13 dup.13 neq # (4 cycles)
 
-  # loop until write_ptr reaches end_ptr (9 cycles per iteration + 1)
+  # loop until write_ptr reaches end_ptr (6 cycles per iteration + 1)
```
bobbinth, line 72: I think 9 cycles per iteration was correct here. For each iteration we take 1 cycle to enter the body of the loop, 1 cycle to exist the body of the loop, and 1 cycle to check the condition.

stdlib/asm/mem.masm
```diff
@@ -65,11 +65,11 @@ end
 #!     - C,B are the rate portion of the state
 #! - The value `words = end_ptr - write_ptr` must be positive and a multiple of 8
 #!
-#! Cycles: 10 + 9 * word_pairs
+#! Cycles: 9 + 6 * word_pairs
```
bobbinth, line 68: This may also need to be reverted.

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
...
   # => [write_ptr', ...]
 end
+
+#! Moves an even number of words from the advice stack to memory and asserts it matches the commitment.
+#!
+#! Input: [num_words, write_ptr, COM, ...]
+#! Output: [write_ptr', ...]
+#!
+#! Cycles: 56 + 3 * num_words / 2
+# Possible other name: pipe_pairs_preimage_to_memory
```
bobbinth: Let's remove this line.

stdlib/docs/collections/smt.md
```diff
@@ -7,4 +7,5 @@
...
-| get | Returns the value located under the specified key in the Sparse Merkle Tree defined by the<br />specified root.<br /><br />If no values had been previously inserted under the specified key, an empty word (i.e.,<br />[ZERO; 4]) is returned.<br /><br />Inputs:<br />Operand stack: [K, R, ...]<br /><br />Outputs:<br />Operand stack: [V, R, ...]<br /><br />Fails if the tree with the specified root does not exits in the VM's advice provider.<br /><br />Cycles<br />Leaf empty: 48 cycles<br />Leaf single: 99 cycles<br />Leaf multiple: unimplemented<br /> |
+| get_multi_leaf_value | Gets a value at the given get. The leaf we're retrieving from is guaranteed<br />to hold more than one key-value pair (provided on the advice stack).<br /><br />Inputs:<br />Operand stack: [leaf_size, NV, K, R]<br />Advice stack:  [V1, K2, V2, K2, ...]<br /><br />Outputs:<br />Operand stack: [V, R]<br /><br />Cycles:<br />Non-empty: 125 + 3 * pair_count<br />Empty: 129 + 3 * pair_count<br /> |
```
bobbinth: Is this procedure a part of public interface? If it doesn't need to be, I'd not include it in the docs.

stdlib/asm/collections/smt.masm
```diff
@@ -403,6 +484,10 @@ export.get
         adv.push_mapvaln adv_push.1
         # => [leaf_size, NV, K, R]
 
+        # Consistency check: assert this is actually a valid leaf.
+        dup lte.MAX_LEAF_SIZE assert.err="invalid leaf: larger than maximum size of 8192"
```
bobbinth: Since the value should be smaller than a `u32`, it may be more efficient to do something like `u32assert u32lte`.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - most of them are documentation related - so, I'm approving the PR.

It would be good to get another review on this though (cc @Al-Kindi-0 and maybe @gswirski).

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +360,80 @@ end
 # GET
 # =================================================================================================
 
+#! Gets a value at the given get. The leaf we're retrieving from is guaranteed
```
Al-Kindi-0: probably a typo here "Gets a value at the given get"

stdlib/asm/collections/smt.masm
Al-Kindi-0: and also because of the u32 op on the next line

stdlib/asm/collections/smt.masm
Al-Kindi-0: I would also format it as

```
#! xxxxxxx
#! 
#! The leaf we are retrieving ....
```

stdlib/asm/collections/smt.masm
Al-Kindi-0: I would add a comment on the safety of doing this, i.e., leaf_size is divisible by 4 and hence we can use field division safely here

stdlib/asm/collections/smt.masm
```diff
@@ -353,6 +360,80 @@ end
...
+
+        # (1 cycle)
+        mem_loadw
+        # => [V, R]
+    else
+        # If the key wasn't found, then the key's value is an empty word.
+        # => [x, x, x, R]
+
+        # (10 cycles)
+        push.0 padw swapw dropw
```
Al-Kindi-0: probably just `drop drop drop padw` is enough?

stdlib/asm/mem.masm
Al-Kindi-0: makes sense

stdlib/asm/mem.masm
```diff
@@ -179,3 +179,42 @@ export.pipe_preimage_to_memory.0
   assert_eqw
   # => [write_ptr', ...]
 end
+
+#! Moves an even number of words from the advice stack to memory and asserts it matches the commitment.
```
Al-Kindi-0: nit: I would write this as
```
#! Moves an even number of words from the advice stack to memory and asserts that their sequential hash matches a given commitment.
```

stdlib/asm/mem.masm
Qyriad: That's not consistent with my testing, in which 1 pair took 11 cycles, 2 pairs took 21, and 3 pairs took 27 (according to `miden-vm analyze`

stdlib/docs/collections/smt.md
Qyriad: This file is autogenerated docs, which is why `set_single_leaf`, `insert_single_leaf`, etc are also in there.

bobbinth: Ah! I thought the auto-generated docs have output only for publicly-exported procedures - but seems like this may not be the case. cc @huitseeker and @bitwalker.

stdlib/asm/collections/smt.masm
```diff
@@ -403,6 +487,11 @@ export.get
         adv.push_mapvaln adv_push.1
         # => [leaf_size, NV, K, R]
 
+        # Consistency check: assert this is actually a valid leaf.
+        # (18 cycles).
+        dup u32lte.MAX_LEAF_SIZE u32assert.err="invalid leaf: larger than maximum size of 8192" drop
```
bobbinth: I'm not sure this will work as intended. We can't use `u32lte` unless we know that `leaf_size` is a `u32`. So, I'd write this as:
```
dup u32assert2 u32lte.MAX_LEAF_SIZE assert.err="invalid leaf: larger than maximum size of 8192"
```
This should be 9 cycles.

bobbinth (approved):
Looks good! Thank you! There are some duplicate entries in the changelog, but I'll merge as is and will fix the changelog in a separate commit.

PR #2049, opened 2025-08-05
Add Miden Assembly Formatter

partylikeits1983:
This PR adds the Miden Assembly formatter as a separate binary to the Miden VM repo. This tool will be integrated into the CI checks on miden-base & miden-vm. There are many instances across both repositories where masm code is incorrectly formatted. This tool is the first version of the formatter until updates are made to the assembly parser to retain inline comments in assembly code. 

Adding this basic formatter will save time in PR reviews of masm code.

This tool both formats masm code and can be run in `--check` mode so that it can be used as a CI check on PRs.

Let me know if I should remove the test masm files, this is where the majority of the diff occurs. The formatting logic is only ~250 lines of code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,6 @@
 - [BREAKING] Procedure-related metadata types in the `miden-assembly` crate in some cases now require an optional type signature argument. If that information is not available, you can simply pass `None` to retain current behavior ([#2028](https://github.com/0xMiden/miden-vm/pull/2028)).
 - Remove basic block clock cycle optimization from `FastProcessor` ([#2054](https://github.com/0xMiden/miden-vm/pull/2054)).
 
-
 ## 0.16.4 (2025-07-24)
 
 - Made `AdviceInputs` field public.
--- a/miden-fmt/src/lib.rs
+++ b/miden-fmt/src/lib.rs
@@ -151,11 +151,12 @@ pub fn format_code(code: &str) -> String {
                                 let next_code_without_comment =
                                     next_line.split('#').next().unwrap().trim();
                                 let next_first_word = next_code_without_comment.split('.').next();
-                                if let Some(next_word) = next_first_word {
-                                    if next_word != "end" && next_word != "else" {
-                                        formatted_code.push('\n');
-                                        last_line_was_empty = true;
-                                    }
+                                if let Some(next_word) = next_first_word
+                                    && next_word != "end"
+                                    && next_word != "else"
+                                {
+                                    formatted_code.push('\n');
+                                    last_line_was_empty = true;
                                 }
                             }
                         }
@@ -243,7 +244,8 @@ pub fn format_code(code: &str) -> String {
             let should_skip_empty_line = if i > 0 && i + 1 < lines.len() {
                 let prev_line = lines[i - 1].trim();
                 let next_line = lines[i + 1].trim();
-                // Skip empty lines between regular comments and proc/export/const, but preserve them after section separators
+                // Skip empty lines between regular comments and proc/export/const, but preserve
+                // them after section separators
                 is_comment(prev_line)
                     && (is_proc_or_export(next_line) || next_line.starts_with("const."))
                     && !is_section_separator_comment(prev_line)
@@ -251,7 +253,8 @@ pub fn format_code(code: &str) -> String {
                 false
             };
 
-            // Allow up to 1 empty line, collapse 2+ into 1, but skip if between comment and proc/export
+            // Allow up to 1 empty line, collapse 2+ into 1, but skip if between comment and
+            // proc/export
             if consecutive_empty_count <= 1 && !should_skip_empty_line {
                 final_output.push_str(line);
                 final_output.push('\n');
--- a/miden-fmt/src/utils.rs
+++ b/miden-fmt/src/utils.rs
@@ -1,5 +1,4 @@
-use crate::constants::SINGLE_LINE_EXPORT_REGEX;
-use crate::types::LineType;
+use crate::{constants::SINGLE_LINE_EXPORT_REGEX, types::LineType};
 
 pub fn is_comment(line: &str) -> bool {
     line.trim_start().starts_with('#')
@@ -137,7 +136,7 @@ pub fn process_import_section(lines: &[&str]) -> (Vec<String>, usize) {
                 // If we have imports in the current group, sort and add them
                 if !current_import_group.is_empty() {
                     current_import_group.sort();
-                    result.extend(current_import_group.drain(..));
+                    result.append(&mut current_import_group);
                     // Add empty line after imports before comment
                     result.push(String::new());
                 }
@@ -146,8 +145,9 @@ pub fn process_import_section(lines: &[&str]) -> (Vec<String>, usize) {
                 end_index = i + 1;
             },
             LineType::Empty => {
-                // Empty lines are preserved in their position, but avoid multiple consecutive empty lines
-                if !result.is_empty() && !result.last().map_or(false, |s| s.is_empty()) {
+                // Empty lines are preserved in their position, but avoid multiple consecutive empty
+                // lines
+                if !result.is_empty() && !result.last().is_some_and(|s| s.is_empty()) {
                     result.push(String::new());
                     end_index = i + 1;
                 }
@@ -160,7 +160,7 @@ pub fn process_import_section(lines: &[&str]) -> (Vec<String>, usize) {
                 // If we have imports in the current group, sort and add them
                 if !current_import_group.is_empty() {
                     current_import_group.sort();
-                    result.extend(current_import_group.drain(..));
+                    result.append(&mut current_import_group);
                 }
                 break;
             },


formatter/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+[dependencies]
+glob = "0.3"
+clap = { version = "4.1", features = ["derive"] }
+regex = "1"
+once_cell = "1.20.3"
+tempfile = "3.17.1"
+
+[[bin]]
+name = "masm-fmt"
+path = "src/main.rs"
```
PhilippGackstatter: I think it would be good to add installation instructions (from crates.io if published and the local one) in the Readme or somewhere appropriate.

formatter/tests/formatter_tests.rs
```diff
@@ -0,0 +1,198 @@
+use std::{fs, fs::read_to_string, path::Path};
```
PhilippGackstatter: Not sure if there's an example/test for this, but I think the behavior in if-else statements is inconsistent. E.g. the formatter outputs this:

```
proc.test
    if.true
        push.5 add
        swap
    else
        push.6 add
        swap
    end
end
```

but also this:


```
proc.test
    if.true
        push.5 add
        swap
        # => [comment]

    else
        push.6 add
        swap
    end
end
```

The newline before else shouldn't be there imo, because it is also not added in the first case (and personally, I prefer it not being there in both cases).

PhilippGackstatter (commented):
Thank you for adding this! I just want to point out a small finding.

formatter/tests/formatter_tests.rs
partylikeits1983: Can definitely update. The reasoning for this was because I hardcoded it to always have an empty line after a stack comment. 

Essentially this is the current logic:
```
# => [...
must be empty line
```

but it would be super easy to update to:
```
# => [...
if `end` statement no empty line, else add empty line
```

PhilippGackstatter: > if `end` statement no empty line, else add empty line

I guess we'd want to avoid an empty line both when it's an `end` or an `else`, right? Maybe the logic could be "if there is a next instruction, insert empty line, otherwise don't", but not sure how easy it is?

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "assembly",
     "assembly-syntax",
     "core",
+    "formatter",
```
bitwalker: Let's call this `miden-fmt`

formatter/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "formatter"
```
bitwalker: Let's name the crate `miden-fmt`, since that's the convention for new CLIs as we transition to `midenup` for toolchain management. It also aligns with the `miden-vm` executable crate.

formatter/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+glob = "0.3"
+clap = { version = "4.1", features = ["derive"] }
+regex = "1"
+once_cell = "1.20.3"
+tempfile = "3.17.1"
+
+[[bin]]
+name = "masm-fmt"
```
bitwalker: With the crate rename I mentioned in my other comment, the explicit name here can be removed

formatter/tests/expected/example1_formatted.masm
plafer, line 1: I'm all for adding `masm` files for tests, but these seem unnecessarily long. I basically don't know what I should be looking at for each test. A better approach would have each file show/test one (or a few related) concepts.

For example, you could have a file with effectively just a long procedure body, and test how long bodies get formatted. We should probably be able to get away with a few hundred lines of test masm for formatting.

partylikeits1983: Yes good point, will reduce the lines of masm test code. These were just to cast a very wide net when testing :)

Cargo.toml
PhilippGackstatter: Btw. wouldn't it be slightly nicer if we didn't have `-` in executable names? I find `rustfmt` easier to type than `miden-fmt` (assuming that's the binary name). So `midenfmt` or `midenvm` as binary names, would be a bit more convenient, wdyt?

formatter/tests/formatter_tests.rs
partylikeits1983: Super easy, I updated it to match this logic.

formatter/tests/expected/example1_formatted.masm
partylikeits1983: Removed a bunch of test files. ~6100 added lines vs 1194 now.

partylikeits1983:
Addressed nits, added some extra logic, and minimized the test masm code. Additionally opened #2072 to show how `miden-fmt` works and which changes it will make.

formatter/src/lib.rs
```diff
@@ -0,0 +1,247 @@
+use std::{
```
PhilippGackstatter, line 1: Input:
```
# The slot in this component's storage layout where the public key is stored.
const.PUBLIC_KEY_SLOT=0

# The slot where the authentication configuration is stored.
const.AUTH_CONFIG_SLOT=1
```

Output:
```
# The slot in this component's storage layout where the public key is stored.

const.PUBLIC_KEY_SLOT=0

# The slot where the authentication configuration is stored.
const.AUTH_CONFIG_SLOT=1
```

That doesn't seem right - why does it insert a newline? I think it does this in every file for the first constant, but not the others.

miden-fmt/tests/expected/procedure_formatting_formatted.masm
```diff
@@ -0,0 +1,46 @@
...
+
+    if.true
+        add.1
+        # => [val]
+    else
+        sub.1
+        # => [val]
+    end
+    add
+    # => [val]
```
PhilippGackstatter:
```suggestion
    if.true
        add.1
        # => [val]
    else
        sub.1
        # => [val]
    end

    add
    # => [val]
```
Perhaps it would make sense to insert newlines after `end`, _iff_ there is an instruction following?

miden-fmt/tests/expected/stack_operations_formatted.masm
```diff
@@ -0,0 +1,34 @@
+export.test_stack_with_comments
+    # => []
+
+    push.1 push.2 push.3
+    # => [3, 2, 1]
+
+    dup.0
+    # => [3, 3, 2, 1]
+
+    swap
```
PhilippGackstatter, line 10: If I insert a comment before the `swap`, the newline will not be inserted, but I think it should be. E.g. this is the result:

```
    dup.0
    # => [3, 3, 2, 1]
    # swap the values
    swap
    # => [3, 3, 2, 1]
```

but I would want:

```
    dup.0
    # => [3, 3, 2, 1]

    # swap the values
    swap
    # => [3, 3, 2, 1]
```

We have this comment, instruction, stack state pattern a lot in miden-base, so I think formatting this as one unit would be best.

PhilippGackstatter (commented):
Thanks for addressing my previous newline comment :)! Looks good!

I only looked at the output of the formatter and how it currently formats the MASM files in miden-base and left some comments based on that.

I guess it would be nice if it auto-formatted comments to be at most 100 characters wide, but probably not necessary for the first version.

formatter/src/lib.rs
partylikeits1983: Yes, I also noticed this bug when formatting the multisig masm code, I fixed it.

miden-fmt/tests/expected/procedure_formatting_formatted.masm
partylikeits1983: Great point, I resolved this in the formatter!

miden-fmt/tests/expected/stack_operations_formatted.masm
partylikeits1983: Completely agree. Added a check that inserts a newline in between a stack comment and a regular comment.

partylikeits1983:
> I guess it would be nice if it auto-formatted comments to be at most 100 characters wide, but probably not necessary for the first version.

Added auto-formatting of long comments.

Cargo.toml
bitwalker: @PhilippGackstatter these executables are going to be invoked via the all-in-one `miden` CLI anyway, so the convention here is really for namespacing purposes, i.e. `miden fmt` will expect there to be a `miden-fmt` executable in the `bin` directory of the installed toolchain, but that directory may also be in the user's `PATH` if they have a need for directly working with the underlying binaries, in which case it is useful for all of our binaries to have a common naming convention.

Personally, I don't find the `-` to be all that annoying, but it entirely depends on one's keyboard layout and preferences. Since I expect most people will be using these tools via `miden` anyway, and shell aliases and/or symlinks are easy enough to create for anyone that wants to use `miden-fmt` directly and finds the full name annoying, I think sticking to the naming convention makes sense here.

miden-fmt/tests/unformatted/control_flow.masm
plafer, line 1: I think this file is unintentionally formatted right? Looks identical to the formatted version. Not sure if that was already your intention, but I would make the unformatted version have basically no indentation

miden-fmt/tests/expected/long_comments_formatted.masm
```diff
@@ -0,0 +1,12 @@
+# This is a very long comment that should be wrapped because it exceeds the 100 character limit and
+# contains multiple words that can be broken at appropriate boundaries
+export.test_procedure
+    # Another very long comment inside a procedure that should also be wrapped when it exceeds the
+    # maximum line length of 100 characters
+    push.1
+    # => [1]
```
PhilippGackstatter, line 7: Thanks for adding this formatting!

One issue I found with miden-base is that this input:

```
    drop
    # => [allow_unauthorized_input_notes, allow_unauthorized_output_notes, num_auth_trigger_procs, pad(16)]
```

is reformatted to

```
    drop
    # => [allow_unauthorized_input_notes, allow_unauthorized_output_notes, num_auth_trigger_procs,

    # pad(16)]
```

Thinking about it now, I don't think it should touch stack comments at all. Formatting this one correctly, with indentation and all seems quite complicated to get right and not something the formatter should even concern itself with. So maybe it's best to only let this "wrap at 100 characters" feature apply to doc comments (`#!`) for now?

It currently also wraps [this table](https://github.com/0xMiden/miden-base/blob/f21386ee057ac235ba497052a5cf201b064948c1/crates/miden-lib/asm/kernels/transaction/lib/tx.masm#L36-L41). Ideally, there would be some way to disable formatting for that like `rusftmt::skip`. For now it would also work to only let wrapping apply to `#!` since here `#` is used.

miden-fmt/tests/expected/basic_formatting_formatted.masm
```diff
@@ -0,0 +1,19 @@
+# Test basic formatting concepts: imports, constants, comments
```
PhilippGackstatter, line 1: One other thing I noticed in miden-base is that this input:

```
        add
        # => [stack]
        # 2nd comment
```

is formatted as:

```
        add
        # => [stack]

        # 2nd comment
```

Which it probably shouldn't do? It seems to only insert a newline if the comment that follows the instruction is `# => [...]`. Is there some kind of detection for `# => [...]`?

PhilippGackstatter (commented):
Thank you for adding comment wrapping. It works pretty well in miden-base, but it looks like there are a few edge cases (see inline comments).

miden-fmt/tests/expected/basic_formatting_formatted.masm
partylikeits1983: Ah! Yes. But the question is should these types of comments be allowed? I hardcoded the formatter to always add a new line after a stack comment. What do you think? It is possible to treat stack comments like regular comments, but stack comments are almost a part of the masm language syntax so that's why I almost treat them as their own "citizens" alongside opcodes.

PhilippGackstatter: One example where the current strategy is not optimal is [here](https://github.com/0xMiden/miden-base/blob/6fd9a1ef93064295e973ce1b7a4df27896b4f111/crates/miden-lib/asm/kernels/transaction/lib/account_delta.masm#L332-L336). I sometimes write comments like this to make it clearer that we're destructuring a word into its constituent felts:

```
# => [KEY, VALUE0, ...]
# => [[faucet_id_prefix, faucet_id_suffix, 0, 0], [delta_amount, 0, 0, 0], ...]
```

Inserting a newline is not the end of the world, but also not ideal. Ideally, though, the formatter would leave it up to the dev to format this appropriately.

> But the question is should these types of comments be allowed?

I think they should be allowed. Thinking about it more, I would make the formatter only format what is "officially" part of the language, e.g. instructions, if, while, etc. I think devs should be able to format their comments mostly without having to consider what the formatter will do with it.

The only exception may be wrapping at 100 characters, which seems like a reasonable thing to do - many formatters do something like that - but otherwise I would treat comments just as black box strings. I do think there is a distinction between doc comments and regular ones. `rustfmt` for example, can assume that doc comments use Markdown, which is well-specified, and so the formatter could apply Markdown formatting there. The stack state comments in MASM on the other hand are not well-specified, and they are also not doc comments, and so the formatter should probably not try to format them.

Even when trying to format them, one difficulty is the different styles they appear in. In miden base, we sometimes write `# OS => [...]` and sometimes just `# => [...]`, which we both consider valid styles, but are just a miden-base local convention and should not apply to other MASM code.

Stack state comments are very pervasive, and so I totally understand why you made them "first class citizens" in the formatter, but handling them correctly may need more discussion. So, maybe for this first version of the formatter, I would only suggest to:
- Wrap doc comments at 100 characters.
- Leave non-doc comments as they were written.

That does go against what I said earlier [here](https://github.com/0xMiden/miden-vm/pull/2049#discussion_r2276458905) with regard to inserting newlines. As a consequence of the above, I guess the simplest solution would be to leave it to MASM devs to insert newlines as appropriate.

But maybe you or others have a clearer perspective on this. I think the important part is that we have a consistent and robust way of formatting things, and stack state comments are a bit tricky.

partylikeits1983: That's true, I agree. Will make the stack comments follow the same rules as regular comments.

bobbinth:
@partylikeits1983 - I think MASM syntax got changed quite a bit since this PR. Does it need to be refreshed? Or maybe better to close it for now?

plafer:
Yes, this was discussed offline. @partylikeits1983 up to you if you want to take on this syntax update - if you do, please just open a new PR (or re-open this one).

Issue #2045, opened 2025-07-31
RFC: API for recursive verification in Miden


Al-Kindi-0:
We are planning, at least in principle, to use our (recursive) verifier from the `stdlib` as the core around which several other verifiers will be built e.g., the Miden VM recursive verifier.

This poses a few challenges:

1. How to define the core verifier in `stdlib` and its interface used by the specific verifier which will make use of it?
2. How do we define and identify proofs, so that they can be passed around and indexed?

The first point is mainly there for re-usability and minimization of the footprint in the blockchain. The second point is needed for most, if not all, components with interactions involving proofs e.g, network, advice provider,...

The first point is being addressed in [#2008](https://github.com/0xMiden/miden-vm/pull/2008), while the second point is the content of this issue.

## SNARGs

The standard methodology for building SNARGs, which are different from SNARKs in the **k**nowledge soundness part, for some relation, say the relation of all satisfiable AIRs, is to start from an interactive oracle proof (IOP) and then use the BCS (Ben-Sasson-Chiesa-Spooner) transform in order to basically remove the "I" and "O" parts, namely interaction and use of ideal oracles. The BCS transform is then thought of as a compiler from an information theoretic object to a cryptographic one.

This means, in the universe of (IOP + BCS) SNARGs, that a SNARG can be identified by the underlying IOP and the parameters of the BCS compiler, in addition of course to the relation.

For example, the current verifier, after the refactoring in [#2008](https://github.com/0xMiden/miden-vm/pull/2008), is a verifier for the relation of all satisfiable AIRs and it is an IOP that is called in the literature `DEEP-ALI + FRI`. Note that, at the protocol level, there is nothing preventing us from supporting SNARGs with the (TURBO)PLONK arithmetization instead of the AIR one or using the `ALI + FRI` IOP as the underlying IOP, or any other combinations, or even using multi-variate IOPs e.g., hyperPLONK, Binius, FRI-Binius ...

[...]

PR #2056, opened 2025-08-07
doc: advice provider layout for recursive verifier

### LINKED ISSUE #1918: Document the layout of the advice provided to the recursive verifier
### What should be done?

We should provide a detail account of the advice provided to the STARK recursive verifier in `stdlib`.

### How should it be done?

Add docs to the module responsible for building the advice provided to the STARK recursive verifier.

### When is this task done?

When the advice is well documented.

### Additional context

_No response_

Al-Kindi-0:
## Describe your changes
Closes #1918 

Putting in draft until we conclude the discussions in #2045 and #2008 

A few things left to fix:

1. We should probably not use the operand stack to pass in parameters. Instead, we should just use the advice stack to pass the log trace length, number of grinding bits and number of FRI queries.
2. We should think about how to change the logic in `stdlib/tests/crypto/stark/verifier_recursive/mod.rs` in order to accommodate the move to the generic verifier model. I think the main thing that we need to do is encapsulate the logic for handling of public inputs, which will need to be specific to each "specific" recursive verifier.

- Updated `CHANGELOG.md'

### DIFF
--- a/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
+++ b/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
@@ -48,11 +48,11 @@ pub struct VerifierData {
 /// | -------- | ---------------------- | -------------- | ----------------------------------------------------- |
 /// | 0        | variable_len_pi_size   | 1              | Size of variable length PI in Felt                    |
 /// | 1-n      | public_inputs_data     | varies         | Input/output stacks + Program digest + kernel digests |
-/// | n+1-n+4  | aux_randomness         | 4              | β = (β₀, β₁), ɑ = (ɑ₀, ɑ₁) (filled later)             |
+/// | n+1-n+4  | aux_randomness         | 4              | β = (β₀, β₁), ɑ = (ɑ₀, ɑ₁)                            |
 /// | n+5      | num_kernel_procedures  | 1              | Count of kernel procedure digests                     |
 /// | n+6-m    | trace_commitments      | 4 or 8         | Main/auxiliary trace segment commitments              |
 /// | m+1-p    | constraint_commitment  | 4              | Constraint composition commitment                     |
-/// | p+1-p+2  | alpha_deep_placeholder | 2              | Deep composition randomness (filled later)            |
+/// | p+1-p+2  | alpha_deep_placeholder | 2              | Deep composition randomness                           |
 /// | p+3-q    | ood_evaluations        | varies         | Out-of-domain trace and constraint evals              |
 /// | q+1-r    | fri_commitments        | varies         | FRI layer commitment digests                          |
 /// | r+1-s    | fri_remainder_poly     | varies         | FRI remainder polynomial coefficients                 |
@@ -73,12 +73,12 @@ pub struct VerifierData {
 /// # Advice Map Layout
 ///
 /// ```
-/// | Key (Word)                     | Value (Vec<Felt>)                   |
-/// | ------------------------------ | ----------------------------------- |
-/// | Query index                    | Main trace evaluations              |
-/// | Query index                    | Auxiliary trace evaluations         |
-/// | Query index                    | Constraint evaluations              |
-/// | FRI layer (folded) query index | FRI codeword evaluations on a coset |
+/// | Key (Word)                                               | Value (Vec<Felt>)                   |
+/// | -------------------------------------------------------- | ----------------------------------- |
+/// | Leaf hash of main trace segment tree at query index      | Main trace evaluations              |
+/// | Leaf hash of auxiliary trace segment tree at query index | Auxiliary trace evaluations         |
+/// | Leaf hash of constraint composition tree at query index  | Constraint evaluations              |
+/// | FRI layer evaluations at folded query index              | FRI codeword evaluations on a coset |
 /// ```
 pub fn generate_advice_inputs(
     proof: Proof,
--- a/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
+++ b/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
@@ -52,7 +52,7 @@ pub struct VerifierData {
 /// | n+5      | num_kernel_procedures  | 1              | Count of kernel procedure digests                     |
 /// | n+6-m    | trace_commitments      | 4 or 8         | Main/auxiliary trace segment commitments              |
 /// | m+1-p    | constraint_commitment  | 4              | Constraint composition commitment                     |
-/// | p+1-p+2  | alpha_deep_placeholder | 2              | Deep composition randomness                           |
+/// | p+1-p+2  | alpha_deep_randomness  | 2              | Deep composition randomness                           |
 /// | p+3-q    | ood_evaluations        | varies         | Out-of-domain trace and constraint evals              |
 /// | q+1-r    | fri_commitments        | varies         | FRI layer commitment digests                          |
 /// | r+1-s    | fri_remainder_poly     | varies         | FRI remainder polynomial coefficients                 |
--- a/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
+++ b/stdlib/tests/crypto/stark/verifier_recursive/mod.rs
@@ -24,6 +24,63 @@ pub struct VerifierData {
     pub advice_map: Vec<(Word, Vec<Felt>)>,
 }
 
+/// Generates advice inputs required for recursive STARK proof verification in Miden VM.
+///
+/// During the course of its execution, the recursive verifier will make use of non-deterministic
+/// advice data through the advice provider.
+/// In what follows is a description of the expected layout per component of the advice provider.
+/// We also include the expected layout of the operand stack for completness.
+///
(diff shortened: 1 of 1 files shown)


stdlib/tests/crypto/stark/verifier_recursive/mod.rs
```diff
@@ -24,6 +24,62 @@ pub struct VerifierData {
...
+/// ```
+///
+/// # Advice Stack Layout
+///
+/// ```
+/// | Position | Content                | Size (in Felt) | Description                                           |
+/// | -------- | ---------------------- | -------------- | ----------------------------------------------------- |
+/// | 0        | variable_len_pi_size   | 1              | Size of variable length PI in Felt                    |
+/// | 1-n      | public_inputs_data     | varies         | Input/output stacks + Program digest + kernel digests |
+/// | n+1-n+4  | aux_randomness         | 4              | β = (β₀, β₁), ɑ = (ɑ₀, ɑ₁) (filled later)             |
```
bobbinth: What does "filled later" mean here? (and same question for line 55 below).

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
```diff
@@ -24,6 +24,62 @@ pub struct VerifierData {
...
+///
+/// # Advice Map Layout
+///
+/// ```
+/// | Key (Word)                     | Value (Vec<Felt>)                   |
+/// | ------------------------------ | ----------------------------------- |
+/// | Query index                    | Main trace evaluations              |
+/// | Query index                    | Auxiliary trace evaluations         |
+/// | Query index                    | Constraint evaluations              |
+/// | FRI layer (folded) query index | FRI codeword evaluations on a coset |
```
bobbinth: Are we actually using query indexes s keys? I thought we'd be using leaf hashes as keys and leaf pre-images as values.

If we are using indexes, how can the same index be used for multiple values (i.e., for main trace, aux trace, and constraint evaluations).

bobbinth (commented):
Looks good! Thank you. I left a couple of questions inline.

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
Al-Kindi-0: What I was trying to say, in a succinct way, is that these values are out-of-order with respect to the interaction transcript. Thus, in the advice tape, we push place holders for these values and only later on in the interaction can we fill these place holders with the appropriate values.
Probably "filled late" is just confusing without enough context, so I have just removed it

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
Al-Kindi-0: Absolutely correct, this should be the leaf at the query index. Thanks for catching it.
Fixed it now

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
bobbinth: What does "placeholder" mean in this context? My understanding was that we put real values onto the advice stack, but then use the "optimistically" - and only later confirm that the values we used were indeed derived correctly. Or does it mean something else?

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
```diff
@@ -24,6 +24,62 @@ pub struct VerifierData {
...
+/// | 1-n      | public_inputs_data     | varies         | Input/output stacks + Program digest + kernel digests |
+/// | n+1-n+4  | aux_randomness         | 4              | β = (β₀, β₁), ɑ = (ɑ₀, ɑ₁)                            |
+/// | n+5      | num_kernel_procedures  | 1              | Count of kernel procedure digests                     |
+/// | n+6-m    | trace_commitments      | 4 or 8         | Main/auxiliary trace segment commitments              |
+/// | m+1-p    | constraint_commitment  | 4              | Constraint composition commitment                     |
+/// | p+1-p+2  | alpha_deep_placeholder | 2              | Deep composition randomness                           |
+/// | p+3-q    | ood_evaluations        | varies         | Out-of-domain trace and constraint evals              |
+/// | q+1-r    | fri_commitments        | varies         | FRI layer commitment digests                          |
+/// | r+1-s    | fri_remainder_poly     | varies         | FRI remainder polynomial coefficients                 |
+/// | s+1      | pow_nonce              | 1              | Proof-of-work nonce                                   |
```
bobbinth: Not for this PR, but a few thoughts on these (specifically in the context of https://github.com/0xMiden/miden-vm/issues/2045):

It is not yet clear to me how these would get onto the advice stack. Generally, there are two options for accomplishing this:

1. There could be a specialized event that, based on the data on the operand stack, would identify the relevant data in the advice map. Based on this data it would construct the advice stack.
2. There would be an entry in the advice map with this data. The key for this entry would be something we can derive deterministically from the data on the operand stack.

Some of the data seems to be specific to Miden VM proofs. For example, `num_kernel_procedures` is probably not something that's relevant for a generic STARK proof.

It may make sense to remove things like `variable_len_pi_size` and `public_inputs_data` from the advice stack. Instead, they these could be loaded onto the advice stack using their commitments (via the `adv.push_mapvaln` instruction). For example, if the commitment to public inputs is on the operand stack, and there is an entry in the advice map for `PUB_INPUTS_COMMITMENT |-> [PUB INPUTS DATA]`, we could load the public inputs onto the advice stack from within MASM.

Lastly, I wonder if it is possible to have the data on the advice stack basically be the transcript of the interaction between the prover and the verifier. Then, the verifier would be able to read it off the advice stack, absorb it into the hasher state, and then, at the end of proof verification, compare it to the transcript "commitment" that it may get via regular inputs (or in some other way).

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
Al-Kindi-0: This is used to signify that it is an out-of-sequence value on the stack that is filled later on in the interaction.
This is probably just confusing and not informative, so I have removed it

bobbinth: Ah - so it meant that we insert these values into the advice stack later (e.g., using something `adv.push_mapval` instruction)?

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
Al-Kindi-0: > Not for this PR, but a few thoughts on these (specifically in the context of #2045):
> 
> It is not yet clear to me how these would get onto the advice stack. Generally, there are two options for accomplishing this:
> 
>     1. There could be a specialized event that, based on the data on the operand stack, would identify the relevant data in the advice map. Based on this data it would construct the advice stack.
> 

Could you elaborate more here? Especially the "identify the relevant data" part?

>     2. There would be an entry in the advice map with this data. The key for this entry would be something we can derive deterministically from the data on the operand stack.
> 

Indeed, this would be similar to the proposal in #2045 , where this is derived from the public inputs, AIR commitment and protocol identifier.

> 
> Some of the data seems to be specific to Miden VM proofs. For example, `num_kernel_procedures` is probably not something that's relevant for a generic STARK proof.
> 

Not really, `num_kernel_procedures` is nothing but the number of messages/interactions in the first (and only) group/table of variable length public inputs. So as part of the transformation from `Proof` to its parts in the advice provider, we have to count the number of messages per table in the variable length public inputs and add this number to the advice stack.


> It may make sense to remove things like `variable_len_pi_size` and `public_inputs_data` from the advice stack. Instead, they these could be loaded onto the advice stack using their commitments (via the `adv.push_mapvaln` instruction). For example, if the commitment to public inputs is on the operand stack, and there is an entry in the advice map for `PUB_INPUTS_COMMITMENT |-> [PUB INPUTS DATA]`, we could load the public inputs onto the advice stack from within MASM.

I think it is cleaner to keep it here as it doesn't make sense, neither conceptually nor performance wise, to commit to them together with the public inputs. 

> Lastly, I wonder if it is possible to have the data on the advice stack basically be the transcript of the interaction between the prover and the verifier. Then, the verifier would be able to read it off the advice stack, absorb it into the hasher state, and then, at the end of proof verification, compare it to the transcript "commitment" that it may get via regular inputs (or in some other way).

Not sure I follow completely. Is the intention here to have the data on the advice stack be exactly the one, without any for example out of order advice data, resulting from the interactions between prover and verifier?
If so, then I think this is a significant limitation and in my mind unnecessary.

stdlib/tests/crypto/stark/verifier_recursive/mod.rs
Al-Kindi-0: Just on the Rust size. We basically, push say a `[0 ,0]` and record the index before this push. Later on when we can generate the random challenge, we can do something like `advice_tape[index] = alpha_0` and `advice_tape[index + 1] = alpha_1`.
Overall, this is just a Rust implementation detail i.e., it could be done differently

Al-Kindi-0:
Closing as stale

Issue #2012, opened 2025-07-21
Add ability to reference part of a word-sized constant in push instructions


PhilippGackstatter:
Add ability to reference part of a word-sized constant (in this context created with the mechanism from #2011) in push instructions, e.g.:

```
# see issue 2011
const.METADATA_SLOT_NAME="miden::basic_fungible_faucet::metadata"

begin
    # push the first two elements of the word onto the stack
    push.METADATA_SLOT_NAME[0..2]
end
```

Context: https://github.com/0xMiden/miden-base/discussions/1394#discussioncomment-13809514

Fumuran:
I've been thinking about this task, and it seems to me that it will require quite a lot of changes. Essentially this issue requires the support of arrays in assembly code, since a slice of a word is an array. So first we have to have something like this:

```masm
const.EXAMPLE_ARRAY=[1, 2, 3, 4, 5, 6, 7]

begin
    push.EXAMPLE_ARRAY
    # which will be equal to 
    push.1.2.3.4.5.6.7
end
```

After having the plain arrays, we could add a slicing feature for them. It could work in a similar way to how we resolve constant expressions in the `analyze()` function in the `assembly-syntax/src/sema/mod`: slice of an array should result in some other array. This will in fact solve this issue, but only if we remove the `IntValue::Word` in favor of array consisting of 4 elements. Otherwise we will have to adapt the slicing for the `IntValue::Word` but it should be much easier since it will result in an array anyway.

bobbinth:
I was thinking that the scope of this issue would be much less limited. We already have word-sized constants, and so we don't need to introduce a more general array type or enable manipulations of arrays/slices. So, what we need is:
```
const.WORD=[1, 2, 3, 4] # this already works

begin
    push.WORD # this already works

    push.WORD[0..2] # this would be translated to push.1.2, which also already works
end
```
So, really what we need is just a transformation this new syntax to already existing syntax - though, how difficult this is, I'm not sure.

bitwalker:
As @bobbinth noted, we only support felt or word-sized constants with `push`, and supporting specific slicing syntax for `push.CONST` is narrow enough to be a simple task.

No transformation is needed here - we'd just add a new `ConstantExpr` variant for slices, and then at expansion time validate that the constant being sliced is a word (or aggregate of felts, but I think we should be stricter initially).

[...]

Issue #2053, opened 2025-08-06
Remove unused `Push*List` instructions


Fumuran:
It seems like we don't use the instructions which push a list of values onto the stack, namely:
- `PushU8List`
- `PushU16List`
- `PushU32List`
- `PushFeltList`

We should decide will these procedures be used in the assembly in future, and if not — we should remove them.

 _Originally posted by @Fumuran in [#2012](https://github.com/0xMiden/miden-vm/issues/2012#issuecomment-3160314861)_

huitseeker:
Closed by 2057.

PR #2057, opened 2025-08-07
Implement slicing of the word constants

Fumuran:
This PR implements the slicing feature for the word constants, for example:
```masm
const.WORD=[1, 2, 3, 4]

begin
    push.WORD[1..3]
    push.WORD[2]
end
```

It was done by introducing a new `PushWordSlice` instruction, which holds not only the `Immediate<IntValue>` (like `Push` instruction), but also a Range.

Closes: #2012.


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Added `AdviceProvider::into_parts()` method ([#2024](https://github.com/0xMiden/miden-vm/pull/2024)).
 - Added type information to procedures in the AST, `Library`, and `PackageExport` types ([#2028](https://github.com/0xMiden/miden-vm/pull/2028)).
 - Added `drop_stack_top` procedure in `std::sys` ([#2031](https://github.com/0xMiden/miden-vm/pull/2031)).
+- Added slicing for the word constants ([#2057](https://github.com/0xMiden/miden-vm/pull/2057)).
```
bobbinth: This should go into v0.18.0 section.

assembly/src/tests.rs
```diff
@@ -1313,6 +1313,55 @@ fn const_conversion_failed_to_u32() -> TestResult {
...
+            "\
+    const.SAMPLE_WORD=[2, 3, 4, 5]
+
+    begin
+        push.SAMPLE_WORD[6..3]
+    end
+    "
+        )
+    );
+    let program = context.assemble(source_invalid_range)?;
```
bobbinth: If I'm understanding correctly, this currently compiles to an empty program. If so, I think what we should do instead is return an error.

assembly/src/instruction/env_ops.rs
```diff
@@ -36,6 +40,32 @@ where
     imms.iter().for_each(|imm| push_felt(block_builder, (*imm).into()));
 }
 
+/// Appends `PUSH` operations to the span block using the [Felt]s obtained from the Word value using
```
bobbinth: nit: we should say "basic block" instead of "span block".

assembly/src/instruction/env_ops.rs
```diff
@@ -36,6 +40,32 @@ where
...
+/// the provided range. If the range in malformed or empty, no operations will be appended to the
+/// span block.
```
bobbinth: Similar to my other comment, if the range is malformed or empty we should return an error.

assembly-syntax/src/ast/visit.rs
```diff
@@ -740,7 +740,7 @@ where
         | MTreeVerifyWithError(code) => visitor.visit_mut_immediate_error_message(code),
         AddImm(imm) | SubImm(imm) | MulImm(imm) | DivImm(imm) | ExpImm(imm) | EqImm(imm)
         | NeqImm(imm) => visitor.visit_mut_immediate_felt(imm),
-        Push(imm) => visitor.visit_mut_immediate_hex(imm),
+        Push(imm) | PushWordSlice(imm, _) => visitor.visit_mut_immediate_hex(imm),
```
bobbinth: Question: what does this do? I'm not too familiar with how this code works, but curious how hex immediate values are relevant here.

bobbinth (commented):
Looks good! Thank you! I left a few comments inline - the main one is about returning errors when the slice ranges are invalid.

Also, would it make sense to address https://github.com/0xMiden/miden-vm/issues/2053 as a part of this PR too? (only if it is straight-forward).

Fumuran:
>Also, would it make sense to address https://github.com/0xMiden/miden-vm/issues/2053 as a part of this PR too? (only if it is straight-forward).

It is very easy to remove these instructions, the only think which is I'm not sure about is whether we should preserve one such instruction as @bitwalker mentioned in his comment in the related issue. So I can remove them if we agree on which specific operations will be removed.

bobbinth:
> It is very easy to remove these instructions, the only think which is I'm not sure about is whether we should preserve one such instruction as @bitwalker mentioned in his comment in the related issue. So I can remove them if we agree on which specific operations will be removed.

My thinking is that we can remove them and if we need them later, shouldn't be too difficult to add.

btw, I also noticed that we have `PushU8`, `PushU16` etc. - are these unused as well? It seems like the same functionality is covered by `Instruction::Push(Immediate)`.

assembly-syntax/src/ast/visit.rs
Fumuran: It seems like because the word value could be encoded as a long hex constant, we use this `visit_mut_immediate_hex` procedure to handle constants which are both values, short and long hexes. But procedures naming looks a bit weird for me too, but probably only because of the changes which were made after the name was chosen.

Fumuran:
>btw, I also noticed that we have `PushU8`, `PushU16` etc. - are these unused as well? It seems like the same functionality is covered by `Instruction::Push(Immediate)`.

No, they are used, but in a slightly different scenario: we use concrete procedures like `Instruction::PushU8` in case we can immediately parse and identify the type of the provided value inside the `grammar.lalrpop`, like in case of `push.123.0x1234` — it will be transformed into the `[Instruction::PushU8(123), Instruction::PushU16(4660)]` instruction sequence.

`Instruction::Push(Immediate)` is used when we can't cay the type of immediate right now, for example when we use a constant as an immediate value. So if we have `push.CONST_FELT_VALUE.CONST_WORD`, it will me transformed into the `[Instruction::Push(Immediate::Constant(CONST_FELT_VALUE)), Instruction::Push(Immediate::Constant(CONST_WORD))]`, and only at the compilation stage after the constant values will be resolved we will be able to identify the exact type of the internal `IntValue` and use the dedicated `env_ops::push_one`/`env_ops::push_many` instruction (like we do so in the [`compile_instruction_impl`](https://github.com/0xMiden/miden-vm/blob/a2d1a0a85ffbf13da70641210a1518fb1597c65f/assembly/src/instruction/mod.rs#L346) function).

Probably we can somehow optimize this pipeline to use only `Push` instruction, but it will require further discussion, so for now I'll only remove unused `Push*List` instructions.

assembly/src/instruction/env_ops.rs
Fumuran: I made it to work the same way as the `.get()` method which is called  on the provided word. So I return a "malformed range" error only if `.get()` returns `None`. That means that using an empty range like `[2..2]` is valid, it just doesn't do anything. Is it fine, or should we return an error in that case too?

bitwalker:
> My thinking is that we can remove them and if we need them later, shouldn't be too difficult to add.

We need at least one of them for parsing - remember, that structure is for the AST, so we want instructions that map to valid syntax sugar, so as to preserve that structure when formatting source code.

> btw, I also noticed that we have `PushU8`, `PushU16` etc. - are these unused as well? It seems like the same functionality is covered by `Instruction::Push(Immediate)`.

They are ostensibly the same, but provide useful semantic information. I'm less concerned with preserving those examples, but it may still be useful to have them. The `Instruction` enum does not need to be very trim, it is used during parsing and semantic analysis, so information useful during those stages should be represented IMO

assembly-syntax/src/ast/visit.rs
bobbinth: So, does this enable the following syntax:
```
begin
    push.0x1234...5679[0..2]
end
```
Or does it do something else?

If the above, then I think it is fine to keep - just want to make sure I understand what it does.

assembly/src/instruction/mod.rs
```diff
@@ -355,10 +355,9 @@ impl Assembler {
             Instruction::PushU32(imm) => env_ops::push_one(*imm, block_builder),
             Instruction::PushFelt(imm) => env_ops::push_one(*imm, block_builder),
             Instruction::PushWord(imms) => env_ops::push_many(&imms.0, block_builder),
-            Instruction::PushU8List(imms) => env_ops::push_many(imms, block_builder),
-            Instruction::PushU16List(imms) => env_ops::push_many(imms, block_builder),
-            Instruction::PushU32List(imms) => env_ops::push_many(imms, block_builder),
-            Instruction::PushFeltList(imms) => env_ops::push_many(imms, block_builder),
```
bobbinth, line 361: As per @bitwalker's comment, let's keep `PushFeltList` in here.

assembly/src/instruction/mod.rs
```diff
@@ -355,10 +355,9 @@ impl Assembler {
             Instruction::PushU32(imm) => env_ops::push_one(*imm, block_builder),
             Instruction::PushFelt(imm) => env_ops::push_one(*imm, block_builder),
```
bobbinth, line 356: I didn't fully understand https://github.com/0xMiden/miden-vm/pull/2057#issuecomment-3167652127. Isn't `Push(Imm)` supposed capture the `u8`, `u16` and other similar cases? Basically, before this PR we have:

```rust
match &**instruction {
    ...
    Instruction::Push(imm) => match (*imm).expect_value() {
        IntValue::U8(v) => env_ops::push_one(v, block_builder),
        IntValue::U16(v) => env_ops::push_one(v, block_builder),
        IntValue::U32(v) => env_ops::push_one(v, block_builder),
        IntValue::Felt(v) => env_ops::push_one(v, block_builder),
        IntValue::Word(v) => env_ops::push_many(&v.0, block_builder),
    },
    Instruction::PushU8(imm) => env_ops::push_one(*imm, block_builder),
    Instruction::PushU16(imm) => env_ops::push_one(*imm, block_builder),
    Instruction::PushU32(imm) => env_ops::push_one(*imm, block_builder),
    Instruction::PushFelt(imm) => env_ops::push_one(*imm, block_builder),
    Instruction::PushWord(imms) => env_ops::push_many(&imms.0, block_builder),
    Instruction::PushU8List(imms) => env_ops::push_many(imms, block_builder),
    Instruction::PushU16List(imms) => env_ops::push_many(imms, block_builder),
    Instruction::PushU32List(imms) => env_ops::push_many(imms, block_builder),
    Instruction::PushFeltList(imms) => env_ops::push_many(imms, block_builder),
    ...
}
```
But I think the following should be sufficient to represent all use-cases:

```rust
match &**instruction {
    ...
    Instruction::Push(imm) => match (*imm).expect_value() {
        IntValue::U8(v) => env_ops::push_one(v, block_builder),
        IntValue::U16(v) => env_ops::push_one(v, block_builder),
        IntValue::U32(v) => env_ops::push_one(v, block_builder),
        IntValue::Felt(v) => env_ops::push_one(v, block_builder),
    },
    Instruction::PushWord(imms) => env_ops::push_many(&imms.0, block_builder),
    Instruction::PushWordSlice(imm, range) => {
        env_ops::push_word_slice(imm, range, block_builder)?
    }
    Instruction::PushFeltList(imms) => env_ops::push_many(imms, block_builder),
    ...
}
```
Or am I missing something?

We don't need to do all of this in this PR (especially if doing this is non-trivial) - but I think eventually we should try remove this duplication.

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -198,6 +198,49 @@ end
...
+
+```
+const.SAMPLE_WORD=[5,6,7,8]
+const.SAMPLE_VALUE=9
+
+begin
+    push.SAMPLE_WORD[1..3]  # is equivalent to push.6.7
+    push.SAMPLE_WORD[0]     # is equivalent to push.5
+
+    push.SAMPLE_WORD[2..2]  # do nothing
```
bobbinth: I would probably treat this as an error as well. We don't allow dynamic indexing, and so the only way this could happen is if the user somehow makes a mistake. It would be good to let the user know about this mistake early on.

bobbinth (commented):
Looks good! Thank you! I left a few more comments inline.

assembly-syntax/src/ast/visit.rs
Fumuran: ~~Unfortunately no: hexes and hex constants are parsed differently. The main idea (for now, in the current `grammar.lalrpop`) is to resolve as much as we can in the grammar itself, so any hex value got parsed into some `IntValue`, so we always falling into one of the `Immediate::Value(value)` variants.~~

I can change the type of the value which we expect to be word in the lalrpop (so we will expect `IntOrHexValue` instead of the `Constant`), so we will be able to take slices from the word _values_ as well.

In other words, these exact `visitor.visit_mut_immediate_hex(imm)` call just resolves constant names into their values, but we can also support slicing of the word values as well.

bobbinth: I more wanted to find out what this line does, but if it is not too difficult to enable slicing of hex literals, then we can enable it (but if it causes any complications, I think it is fine to skip this too).

assembly/src/instruction/mod.rs
Fumuran: Yes, you are correct, it is possible to capture all the values using only `Push(Imm)` instruction. AFAICS usage of the dedicated `PushU8`-like instructions could be a bit more efficient (but just a bit, essentially we're doing the same), and I thought that it was the reason for having them, but if it's not the case I can remove them in a follow up PR. Created an issue for that: https://github.com/0xMiden/miden-vm/issues/2066.

assembly-syntax/src/ast/instruction/print.rs
```diff
@@ -218,12 +218,13 @@ impl PrettyPrint for Instruction {
...
             },
-            Self::PushWord(value) => flatten(const_text("push") + const_text(".") + value.render()),
-            Self::PushU8List(values) => inst_with_pretty_params("push", values),
-            Self::PushU16List(values) => inst_with_pretty_params("push", values),
-            Self::PushU32List(values) => inst_with_pretty_params("push", values),
+            Self::PushWord(value) => flatten(const_text("push.") + value.render()),
+            Self::PushWordSlice(value, range) => flatten(
+                const_text("push.")
+                    + value.render()
+                    + text(format!("[{}..{}]", range.start, range.end)),
```
bitwalker: When formatting using the `PrettyPrint` trait, it's generally a good idea to emit delimiter tokens separately, so that the formatter can choose to break lines at those delimiters.

```suggestion
                    + const_text("[")
                    + display(range.start)
                    + const_text("..")
                    + display(range.end)
                    + const_text("]"),
```

assembly-syntax/src/ast/visit.rs
bitwalker: To answer your question @bobbinth, the parser (not the visitor impl we're commenting on here) is what determines what syntax is allowed, and we may narrow that during semantic analysis to prevent certain invalid constructions. So there isn't any specific reason why slicing of hex-encoded values (either written literally, or resolved from a constant name, e.g. `FOO`) can't be supported. 

An error would be raised if the hex-encoded value is not a valid word-sized constant though (and we could probably loosen that to only error if the slice indices would refer to something "out of bounds", so if you have a hex-encoded value that is representable as a single felt, but your slice is `[0..1]`, there isn't any reason why that needs to be an error, it's just a redundant slice (and is a bit goofy to begin with, but not an error per se).

assembly-syntax/src/parser/error.rs
```diff
@@ -283,6 +283,21 @@ pub enum ParsingError {
         #[label]
         span: SourceSpan,
     },
+    #[error("invalid slice constant")]
+    #[diagnostic()]
+    InvalidSliceConstant {
+        #[label("slicing could be applied only to the word constants")]
```
bitwalker:
```suggestion
        #[label("slices are only supported over word-sized constants")]
```

assembly-syntax/src/parser/error.rs
```diff
@@ -283,6 +283,21 @@ pub enum ParsingError {
...
+        #[label]
+        span: SourceSpan,
+    },
+    #[error("invalid word slice range")]
```
bitwalker:
```suggestion
    #[error("invalid slice: expected valid range")]
```

assembly-syntax/src/parser/error.rs
```diff
@@ -283,6 +283,21 @@ pub enum ParsingError {
...
+    #[diagnostic()]
+    InvalidWordSliceRange {
```
bitwalker: I think I'd prefer to split this into two variants:

* `EmptySlice` - raised when the slice range is empty, and therefore would produce no value.
* `InvalidRange` - raised when the range bounds are either inverted (i.e. start follows end) or one of the bounds is invalid (i.e. overflows `u8::MAX`)

assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -211,9 +212,7 @@ pub enum Instruction {
     PushU32(u32),
     PushFelt(Felt),
     PushWord(WordValue),
-    PushU8List(Vec<u8>),
-    PushU16List(Vec<u16>),
-    PushU32List(Vec<u32>),
+    PushWordSlice(Immediate<IntValue>, Range<usize>),
```
bitwalker: I think we should call this just `PushSlice`, as we don't need to necessarily require that the immediate is a word (though in practice that is likely to be the common case).

assembly-syntax/src/parser/grammar.lalrpop
```diff
@@ -1258,6 +1258,26 @@ InstWithBitSizeImmediate: Instruction = {
...
+        smallvec![Op::Inst(Span::new(
+            span!(source_file.id(), l, r),
+            Instruction::PushWordSlice(
+                imm_value,
+                Range { start: i as usize, end: i as usize + 1 }
+            )
+        ))]
+    },
+
+    <l:@L> "push" "." <imm_value:IntOrHexImm> "[" <l_range:uint> "." "." <r_range:uint> "]" <r:@R> => {
```
bitwalker: I would suggest we add a `..` token in the lexer, since that will produce more useful error messages if someone does something like `push.FOO[1.2]`.

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -198,6 +198,49 @@ end
...
+const.SAMPLE_WORD=[1,2,3,4]
+const.SAMPLE_HEX_WORD=0x0200000000000000030000000000000004000000000000000500000000000000
+
+begin
+    push.SAMPLE_WORD       # is equivalent to push.1.2.3.4
+    push.SAMPLE_HEX_WORD.6 # is equivalent to push.2.3.4.5.6
+end
+```
+
+#### Word constant slices
```
bitwalker:
```suggestion
#### Constant slices
```

I think it's fine for the content here to be specific about slices only working with word constants for now, but I think it's probably better to avoid making the title of this section word-specific, since the syntax itself doesn't _need_ to be word-specific.

bobbinth (approved):
Looks good! Thank you!

bobbinth:
@bitwalker I think this is ready to merge - could you take one last look?

bitwalker (approved):
LGTM!

Issue #1874, opened 2025-06-11
Document new syntax for declaring word constants


paracetamolo:
https://github.com/0xMiden/miden-vm/pull/1855 introduced support for word constants, the documentation should explain the syntax to define them.

Fumuran:
Closed by #2058.

PR #2058, opened 2025-08-07
Documentation for word constants

Fumuran:
This small PR adds a brief documentation for the word constants.

Closes: #1874.

### DIFF
--- a/docs/src/user_docs/assembly/code_organization.md
+++ b/docs/src/user_docs/assembly/code_organization.md
@@ -200,7 +200,7 @@ end
 
 #### Word constants
 
-Along with the regular value constants a word constants could be used. They could be declared as an array of four elements or as a long hex value, and then could be used in the `push` instructions referenced by their name. Notice that a word constant can not be used in a constant expression.
+Along with the regular value constants a _word_ constants could be used. They could be declared as an array of four elements or as a long hex value, and then could be used in the `push` instructions referenced by their name. Notice that a word constant can not be used in a constant expression.
 
 ```
 const.SAMPLE_WORD=[1,2,3,4]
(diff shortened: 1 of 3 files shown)


docs/src/user_docs/assembly/code_organization.md
```diff
@@ -198,6 +198,47 @@ end
...
+
+begin
+    push.SAMPLE_WORD       # is equivalent to push.1.2.3.4
+    push.SAMPLE_HEX_WORD.6 # is equivalent to push.2.3.4.5.6
+end
+```
+
+#### Word constant slices
+
+It is possible to get just some part of a word constant using a slice notation. This could be done by using a range in the square brackets right after a constant name. Attempt to get a slice from a constant which doesn't represent a word will return an error.
```
bobbinth:
```suggestion
It is possible to get just some part of a word constant using slice notation. This could be done by specifying a range in square brackets right after the constant's name. Attempt to get slices from constants which don't represent words will result in errors. 
```

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -198,6 +198,47 @@ end
 
 ```
 
+#### Word constants
+
+Along with the regular value constants a word constants could be used. They could be declared as an array of four elements or as a long hex value, and then could be used in the `push` instructions referenced by their name. Notice that a word constant can not be used in a constant expression.
```
bobbinth:
```suggestion
Along with the regular value constants a _word_ constants could be used. They could be declared as an array of four elements or as a long hex value, and then could be used in the `push` instructions referenced by their name. Notice that a word constant can not be used in a constant expression.
```

bobbinth (approved):
Looks good! Thank you! I left a couple of small suggestions inline.

bobbinth:
I'll merge this PR into the base PR as I think they changes are quite related.

PR #2065, opened 2025-08-10
refactor(dependency): Reuse existing serialization for DependencyName

gap-editor:
This PR refactors the serialization and deserialization logic for the Dependency struct to improve code reuse and encapsulation.
Previously, the `Serializable `and `Deserializable `implementations for `Dependency `accessed the inner field of `DependencyName` directly (i.e., `self.name.0`).
This approach had a couple of drawbacks:
- It breaks encapsulation by reaching into the internal representation of `DependencyName`.
- It ignores the existing `Serializable `/ `Deserializable `implementations that `DependencyName `already provides, leading to less maintainable code.
This change simplifies the code by delegating the serialization task to `DependencyName ` itself, which already knows how to handle it.

### DIFF
--- a/package/src/dependency/mod.rs
+++ b/package/src/dependency/mod.rs
@@ -54,14 +54,14 @@ pub struct Dependency {
 
 impl Serializable for Dependency {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
-        self.name.0.write_into(target);
+        self.name.write_into(target);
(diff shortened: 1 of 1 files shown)


plafer:
Thank you for your interest in the project - but we don't accept simple PRs like this. I would recommend looking into issues marked as `good first issue` instead.

PR #2069, opened 2025-08-11
doc: update miden-assembly README, use it as doctest

### LINKED ISSUE #2064: miden-assembly README is out of date
### Packages versions

miden-assembly: 0.17.0 

### Bug description

Readme https://github.com/0xMiden/miden-vm/tree/next/assembly/README.md , which is also shown on crates.io and linked to as miden-assembly docs in other places is outdated:

The full example doesn't work.
```
use miden_assembly::Assembler;
use miden_stdlib::StdLibrary;

// Source code of the kernel module
let kernel = "export.foo add end";

// Instantiate the assembler with multiple options at once
let assembler = Assembler::default()
    .with_debug_mode(true)
    .with_library(&StdLibrary::default())
    .and_then(|a| a.with_kernel_from_module(kernel))
    .unwrap();

// Assemble our program
assembler.assemble_program("
begin
    push.1.2
    syscall.foo
end
");
```

1.`with_library` not found
2. in `.and_then(|a| a.with_kernel_from_module(kernel))` `with_kernel_from_module` not found
 

_No response_

### Relevant log output

```shell

```

huitseeker:
Fixes #2064

## Describe your changes

Updates the miden-assembly Readme, makes it a doctest, and includes the README in the crate doc.
(`cd assembly && cargo doc --open`):

| Before | After|
 :-------------------------:|:-------------------------:
|Screenshot 2025-08-11 at 9 47 29 PM |  Screenshot 2025-08-11 at 9 47 11 PM|

### DIFF
--- a/assembly/README.md
+++ b/assembly/README.md
@@ -41,8 +41,7 @@ let program2 = assembler2.assemble_program(Path::new("../miden-vm/masm-examples/
     .unwrap();
 ```
 
-> [!NOTE]
-> The default assembler provides no kernel or standard libraries, you must
+> **Note:** The default assembler provides no kernel or standard libraries, you must
 > explicitly add those using the various builder methods of `Assembler`, as
 > described in the next section.
 
@@ -159,8 +158,7 @@ end
 ").unwrap();
 ```
 
-> [!NOTE]
-> An unqualified `syscall` target is assumed to be defined in the kernel module.
+> **Note:** An unqualified `syscall` target is assumed to be defined in the kernel module.
 > This is unlike the `exec` and `call` instructions, which require that callees
 > resolve to a local procedure; a procedure defined in an explicitly imported
 > module; or the hash of a MAST root corresponding to the compiled procedure.


assembly/README.md
```diff
@@ -59,26 +65,29 @@ to a common namespace, and which are packaged together. The
...
 ```rust
-use miden_assembly::Assembler;
-
-let assembler = Assembler::default()
-    .with_library(&miden_stdlib::StdLibrary::default())
+# use miden_assembly::Assembler;
+# use miden_assembly_syntax::debuginfo::DefaultSourceManager;
+# use miden_stdlib::StdLibrary;
+# use std::sync::Arc;
+#
```
plafer, line 80: Are the `#` a typo?

plafer: ah it hides those lines from rendered output - cool!

assembly/README.md
```diff
@@ -41,8 +41,7 @@ let program2 = assembler2.assemble_program(Path::new("../miden-vm/masm-examples/
     .unwrap();
 ```
 
-> [!NOTE]
-> The default assembler provides no kernel or standard libraries, you must
+> **Note:** The default assembler provides no kernel or standard libraries, you must
```
bitwalker, line 44: *Nit:* Personally, I'd prefer we preserve the use of alerts/annotations - it provides a much better reading experience on GitHub IMO, and they are much more likely to catch the attention of casual readers when calling attention to important details.

Is there a particular reason to remove them?

bitwalker: @plafer what are your thoughts on this

assembly/README.md
huitseeker: This is a rustdoc feature that lets you hide portions of the example.
https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example

Rustdoc will be used to render this page after release:
https://crates.io/crates/miden-assembly

assembly/README.md
huitseeker: Yes, there is: they are understood as links by the rustdoc parser, when grepping for doc in this file (due to the doc inclusion in src/lib). Run `cargo test --doc` on the penultimate commit of this PR to see [the error](https://github.com/0xMiden/miden-vm/actions/runs/16891763530/job/47853005594):

```
rror: unresolved link to `!NOTE`
  --> assembly/src/../README.md:44:4
   |
44 | > [!NOTE]
   |    ^^^^^ no item named `!NOTE` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`

error: unresolved link to `!NOTE`
   --> assembly/src/../README.md:162:4
    |
162 | > [!NOTE]
    |    ^^^^^ no item named `!NOTE` in scope
    |
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

error: could not document `miden-assembly`
```

So IIUC we can have either the admonition, xor running this README as a doc test (unless there is a workaround I don't know of).

assembly/Cargo.toml
```diff
@@ -34,5 +34,6 @@ thiserror.workspace = true
 
 [dev-dependencies]
 miden-mast-package = { workspace = true, features = ["arbitrary"] }
+miden-stdlib.workspace = true
```
bitwalker, line 37: I'm surprised this works, since it forms a circular dependency (`miden-stdlib` depends on `miden-assembly` for both its build script and the code itself). It seems like it works in CI, but that likely has both crates in its build cache, so that may allow it to succeed anyway - does a clean build still succeed?

assembly/README.md
bitwalker: The best of both worlds is likely using a build script to strip out the admonitions (and whatever other stuff is irrelevant to rustdoc tests), then writing the resulting markdown file to `$OUT_DIR/README.md`, and replacing `#![doc = include_str!("../README.md")]` with `#![doc = include_str!(concat!(env!("OUT_DIR"), "/README.md"))]` to read that instead. The downside is the addition of the build script for that purpose.

I guess I'd be interested in what the general consensus is around the use of admonitions in our READMEs, if people don't think they are all that valuable, then 🤷. I think they are, but it's also not a hill I'm interested in dying on, so I'll leave it up to you all.

assembly/Cargo.toml
huitseeker: I think because this is a `dev-dependency`, it produces a different node in the dependency graph (IOW, the `test` condition produces a different set of binaries). 

As to your last question, you can see for yourself with :
```
docker run --rm rust:trixie bash -c "apt-get update -qq && apt-get install -y -qq git && git clone --branch issue-2064 https://github.com/huitseeker/miden-vm.git && cd miden-vm && cargo check --all --tests"
```

See also
https://gist.github.com/huitseeker/16c418398d5c677aa29878d98d787e21

assembly/README.md
plafer: > @plafer what are your thoughts on this

I don't have strong opinions on this - and I'm fine with the current solution given that it avoids the rustdoc issue. Happy to revisit in the future if we really want to improve the rendering of our docs.

PR #2071, opened 2025-08-12
feat: implement custom sections in package format, prototype new serialization format

bitwalker:
This PR does the following:

* Implements support for custom sections in the package format, these can contain arbitrary data, and are either tagged with a well-known ID, or a user-defined name
* Prototype implementation of a new binary serialization format based on `serde`, which I've called BITE, see the `miden-bite` crate for more
* Implemented `serde`-based serialization for most core types to enable side-by-side comparison of existing serialization infra and `serde`. This hasn't been fully optimized yet, so some types are still not necessarily encoded as efficiently as they could be

This is in draft status until I finish adding some tests and cleaning up a few of the type serialization impls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ overflow-checks = true
 [workspace.dependencies]
 # Workspace crates
 miden-air = { path = "./air", version = "0.17", default-features = false }
+miden-bite = { path = "./crates/utils/bite", version = "0.17", default-features = false }
 miden-assembly = { path = "./assembly", version = "0.17", default-features = false }
 miden-assembly-syntax = { path = "./assembly-syntax", version = "0.17", default-features = false }
 miden-core = { path = "./core", version = "0.17", default-features = false }
@@ -72,6 +73,7 @@ serde = { version = "1.0", default-features = false, features = [
     "rc",
 ] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+serde-untagged = "0.1"
 smallvec = { version = "1.13", default-features = false, features = [
     "union",
     "const_generics",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ miden-formatting = { version = "0.1", default-features = false }
 midenc-hir-type = { version = "0.1", default-features = false }
 
 # Third-party crates
+env_logger = "0.11"
 log = "0.4"
 proptest = { version = "1.7", default-features = false, features = [
     "no_std",
--- a/assembly-syntax/src/ast/ident.rs
+++ b/assembly-syntax/src/ast/ident.rs
@@ -199,6 +199,27 @@ impl FromStr for Ident {
     }
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for Ident {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.as_str())
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for Ident {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let name = <&'de str as serde::Deserialize>::deserialize(deserializer)?;
+        Self::new(name).map_err(serde::de::Error::custom)
+    }
+}
+
 #[cfg(feature = "arbitrary")]
 pub(crate) mod testing {
     use alloc::string::String;
--- a/assembly-syntax/src/ast/procedure/name.rs
+++ b/assembly-syntax/src/ast/procedure/name.rs
@@ -14,6 +14,9 @@ use miden_core::utils::{
 use miden_debug_types::{SourceSpan, Span, Spanned};
 use miden_utils_diagnostics::{IntoDiagnostic, Report, miette};
 
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
 use crate::{
     LibraryNamespace, LibraryPath,
     ast::{CaseKindError, Ident, IdentError},
@@ -157,6 +160,103 @@ impl Deserializable for QualifiedProcedureName {
     }
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for QualifiedProcedureName {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
+            let name = format!("{}", self);
+            serializer.serialize_str(&name)
+        } else {
+            use serde::ser::SerializeStruct;
+
+            let mut builder = serializer.serialize_struct("QualifiedProcedureName", 2)?;
+            builder.serialize_field("module", &self.module)?;
+            builder.serialize_field("name", &self.name)?;
+            builder.end()
+        }
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for QualifiedProcedureName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Visitor;
+
+        if deserializer.is_human_readable() {
+            let name = <&'de str as serde::Deserialize>::deserialize(deserializer)?;
+            return Self::from_str(name).map_err(serde::de::Error::custom);
+        }
+
+        #[derive(Deserialize)]
+        #[serde(field_identifier, rename_all = "lowercase")]
+        enum Field {
+            Module,
+            Name,
+        }
+
+        struct QualifiedProcedureNameVisitor;
+        impl<'de> Visitor<'de> for QualifiedProcedureNameVisitor {
+            type Value = QualifiedProcedureName;
+
+            fn expecting(&self, formatter: &mut core::fmt::Formatter) -> core::fmt::Result {
+                formatter.write_str("struct QualifiedProcedureName")
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let module = seq
+                    .next_element()?
+                    .ok_or_else(|| serde::de::Error::invalid_length(0, &self))?;
+                let name = seq
+                    .next_element()?
+                    .ok_or_else(|| serde::de::Error::invalid_length(1, &self))?;
+                Ok(QualifiedProcedureName::new(module, name))
+            }
+
+            fn visit_map<A>(self, mut map: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::MapAccess<'de>,
+            {
+                let mut module = None;
+                let mut name = None;
+                while let Some(key) = map.next_key()? {
+                    match key {
+                        Field::Module => {
+                            if module.is_some() {
+                                return Err(serde::de::Error::duplicate_field("module"));
+                            }
+                            module = Some(map.next_value()?);
+                        },
+                        Field::Name => {
+                            if name.is_some() {
+                                return Err(serde::de::Error::duplicate_field("name"));
+                            }
+                            name = Some(map.next_value()?);
+                        },
+                    }
+                }
+                let module = module.ok_or_else(|| serde::de::Error::missing_field("module"))?;
+                let name = name.ok_or_else(|| serde::de::Error::missing_field("name"))?;
+                Ok(QualifiedProcedureName::new(module, name))
+            }
+        }
+
+        deserializer.deserialize_struct(
+            "QualifiedProcedureName",
+            &["module", "name"],
+            QualifiedProcedureNameVisitor,
+        )
+    }
+}
+
 // PROCEDURE NAME
 // ================================================================================================
 
@@ -196,6 +296,8 @@ impl Deserializable for QualifiedProcedureName {
 /// end
 /// ```
 #[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
 pub struct ProcedureName(Ident);
 
 impl ProcedureName {
--- a/assembly-syntax/src/library/mod.rs
+++ b/assembly-syntax/src/library/mod.rs
@@ -8,6 +8,9 @@ use miden_core::{
 use midenc_hir_type::{FunctionType, Type};
 use smallvec::SmallVec;
 
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
 use crate::ast::QualifiedProcedureName;
 
 mod error;
@@ -29,12 +32,14 @@ pub use self::{
 
 /// Metadata about a procedure exported by the interface of a [Library]
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct LibraryExport {
     /// The id of the MAST root node of the exported procedure
     pub node: MastNodeId,
     /// The fully-qualified name of the exported procedure
     pub name: QualifiedProcedureName,
     /// The type signature of the exported procedure, if known
+    #[cfg_attr(feature = "serde", serde(default))]
     pub signature: Option<FunctionType>,
 }
 
@@ -280,8 +285,12 @@ impl Library {
 /// - There must be at least one exported procedure.
 /// - The number of exported procedures cannot exceed [Kernel::MAX_NUM_PROCEDURES] (i.e., 256).
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "Library"))]
 pub struct KernelLibrary {
+    #[cfg_attr(feature = "serde", serde(skip))]
     kernel: Kernel,
+    #[cfg_attr(feature = "serde", serde(skip))]
     kernel_info: ModuleInfo,
     library: Library,
 }
@@ -407,6 +416,116 @@ impl Deserializable for Library {
         let digest = mast_forest.compute_nodes_commitment(exports.values().map(|e| &e.node));
 
         Ok(Self { digest, exports, mast_forest })
+    }
+}
+
+#[cfg(feature = "serde")]
+impl serde::Serialize for Library {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use serde::ser::SerializeStruct;
+
+        struct LibraryExports<'a>(&'a BTreeMap<QualifiedProcedureName, LibraryExport>);
+        impl serde::Serialize for LibraryExports<'_> {
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+            where
+                S: serde::Serializer,
+            {
+                use serde::ser::SerializeSeq;
+                let mut serializer = serializer.serialize_seq(Some(self.0.len()))?;
+                for elem in self.0.values() {
+                    serializer.serialize_element(elem)?;
+                }
+                serializer.end()
+            }
+        }
+
+        let Self { digest: _, exports, mast_forest } = self;
+
+        let mut serializer = serializer.serialize_struct("Library", 2)?;
+        serializer.serialize_field("mast_forest", mast_forest)?;
+        serializer.serialize_field("exports", &LibraryExports(exports))?;
+        serializer.end()
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for Library {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Visitor;
+
+        #[derive(Deserialize)]
+        #[serde(field_identifier, rename_all = "snake_case")]
+        enum Field {
+            MastForest,
+            Exports,
+        }
+
+        struct LibraryVisitor;
+
+        impl<'de> Visitor<'de> for LibraryVisitor {
+            type Value = Library;
+
+            fn expecting(&self, formatter: &mut core::fmt::Formatter) -> core::fmt::Result {
+                formatter.write_str("struct Library")
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                let mast_forest = seq
+                    .next_element()?
+                    .ok_or_else(|| serde::de::Error::invalid_length(0, &self))?;
+                let exports: Vec<LibraryExport> = seq
+                    .next_element()?
+                    .ok_or_else(|| serde::de::Error::invalid_length(1, &self))?;
+                let exports =
+                    exports.into_iter().map(|export| (export.name.clone(), export)).collect();
+                Library::new(mast_forest, exports).map_err(serde::de::Error::custom)
+            }
+
+            fn visit_map<A>(self, mut map: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::MapAccess<'de>,
+            {
+                let mut mast_forest = None;
+                let mut exports = None;
+                while let Some(key) = map.next_key()? {
+                    match key {
+                        Field::MastForest => {
+                            if mast_forest.is_some() {
+                                return Err(serde::de::Error::duplicate_field("mast_forest"));
+                            }
+                            mast_forest = Some(map.next_value()?);
+                        },
+                        Field::Exports => {
+                            if exports.is_some() {
+                                return Err(serde::de::Error::duplicate_field("exports"));
+                            }
+                            let items: Vec<LibraryExport> = map.next_value()?;
+                            exports = Some(
+                                items
+                                    .into_iter()
+                                    .map(|export| (export.name.clone(), export))
+                                    .collect(),
+                            );
+                        },
+                    }
+                }
+                let mast_forest =
+                    mast_forest.ok_or_else(|| serde::de::Error::missing_field("mast_forest"))?;
+                let exports = exports.ok_or_else(|| serde::de::Error::missing_field("exports"))?;
+                Library::new(mast_forest, exports).map_err(serde::de::Error::custom)
+            }
+        }
+
+        deserializer.deserialize_struct("Library", &["mast_forest", "exports"], LibraryVisitor)
     }
 }
 
--- a/assembly-syntax/src/library/namespace.rs
+++ b/assembly-syntax/src/library/namespace.rs
@@ -166,6 +166,17 @@ impl LibraryNamespace {
     pub fn to_ident(&self) -> Ident {
         Ident::from_raw_parts(Span::unknown(self.as_refcounted_str()))
     }
+
+    #[cfg(feature = "serde")]
+    const fn tag(&self) -> u8 {
+        // SAFETY: This is safe because we have given this enum a
+        // primitive representation with #[repr(u8)], with the first
+        // field of the underlying union-of-structs the discriminant
+        //
+        // See the section on "accessing the numeric value of the discriminant"
+        // here: https://doc.rust-lang.org/std/mem/fn.discriminant.html
+        unsafe { *(self as *const Self).cast::<u8>() }
+    }
 }
 
 impl core::ops::Deref for LibraryNamespace {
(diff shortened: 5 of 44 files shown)


package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// The section containing debug information (source locations, spans)
+    DebugInfo = 1,
+
+    //--- ROLLUP SECTIONS
+    /// This section provides the encoded metadata for a compiled account component
+    ///
+    /// Currently, this corresponds to the serialized representation of
+    /// `miden-objects::account::AccountComponentMetadata`, i.e. name, descrioption, storage, that
+    /// is associated with this package.
+    AccountComponentMetadata = 10,
```
bobbinth: It feels a bit weird that here in `miden-vm` we need to know about how custom sections may be used downstream in `miden-objects`. Ideally, `miden-vm` creates should not be aware of `miden-objects` or anything else in `miden-base`.

Could we not use the `Custom` section for these? That is, what is the main thing we gain from making `AccountComponentMetadata` special here?

package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+/// The following tag ranges are reserved for specific categories of use:
+///
+/// * `0` - unused
+/// * `1..10` - reserved for non-rollup use cases, e.g. debug information
+/// * `10..100` - reserved for rollup use cases, e.g. account component metadata
+/// * `100..254` - reserved for TBD use cases
+/// * `255` - assigned to custom user-defined sections, which must also be uniquely named
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum SectionId {
```
bobbinth: Why do we need `SectionId` to be a `u8`-based enum? Could we not make it string-based? This way, we won't need to assign some pre-defined ranges - each custom section could define its unique "namespace".

package/src/package/mod.rs
```diff
@@ -1,19 +1,26 @@
...
 
 // PACKAGE
 // ================================================================================================
 
 /// A package containing a [Program]/[Library], and a manifest (exports and dependencies).
 #[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Package {
     /// Name of the package
     pub name: String,
```
bobbinth, line 27: Should we add other things here like `description` and `version`? We currently have these in [AccountComponentMetadata](https://github.com/0xMiden/miden-base/blob/next/crates/miden-objects/src/account/component/template/mod.rs#L150), and it would seem a bit weird to have the name here and then description/version in a custom section.

We could also duplicate the name in both places, but that also seems a bit weird.

package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+}
+
+#[derive(Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Section {
+    pub id: SectionId,
+    pub data: Cow<'static, [u8]>,
+}
```
bobbinth, line 86: Related to the previous comment: i'm trying to think through how we'd use this downstream in `miden-base`.

One option is to serialize the whole `AccountComponentMetadata` as a custom section here. Though, this would lead to some duplication (e.g., of `name` and `description` field, which I think would be the same for both package and account component metadata). I'm assuming TOML representation of this would look something like:

```toml
# general package properties go here

[account_component]
name = "Fungible Faucet"
description = "Account component for fungible faucets"
version = "1.0.0"
supported-types = ["FungibleFaucet"]

[[account_component.storage]]
name = "token_metadata"
description = "Contains metadata about the token associated to the faucet account. The metadata
is formed by three fields: max supply, the token symbol and the asset's decimals"
slot = 0
value = [
    { type = "felt", name = "max_supply", description = "Maximum supply of the token in base units" },
    { type = "token_symbol", value = "TST" },
    { type = "u8", name = "decimals", description = "Number of decimal places for converting to absolute units", value = "10" },
    { value = "0x0" }
]

[[account_component.storage]]
name = "owner_public_key"
description = "This is a value placeholder that will be interpreted as a Falcon public key"
slot = 1
type = "auth::rpo_falcon512::pub_key"
```

Or we could assume that name/description/version (and maybe other things are a part of the generic package properties, and then this becomes:
```toml
[account_component]
supported-types = ["FungibleFaucet"]

[[account_component.storage]]
name = "token_metadata"
description = "Contains metadata about the token associated to the faucet account. The metadata
is formed by three fields: max supply, the token symbol and the asset's decimals"
slot = 0
value = [
    { type = "felt", name = "max_supply", description = "Maximum supply of the token in base units" },
    { type = "token_symbol", value = "TST" },
    { type = "u8", name = "decimals", description = "Number of decimal places for converting to absolute units", value = "10" },
    { value = "0x0" }
]

[[account_component.storage]]
name = "owner_public_key"
description = "This is a value placeholder that will be interpreted as a Falcon public key"
slot = 1
type = "auth::rpo_falcon512::pub_key"
```

------

Separately, my understanding is that "work" with this format, we'd still need to convert the `Package` into something like [AccountComponentTemplate](https://github.com/0xMiden/miden-base/blob/next/crates/miden-objects/src/account/component/template/mod.rs#L30) in `miden-base`. Basically, the `Package` "lives" on disc in this format, but when we want to use (e.g., to instantiate an `AccountComponent`), we'd do the following:

1. Deserialize it using the functionality contained here, at which point we'll get a set of sections which are basically just vectors of bytes.
2. Identify the relevant sections and convert them into their "typed" representation - e.g., `AccountComponentMetadata`.
3. Build the specialized struct - e.g., `AccountComponentTemplate`.

This all an be simplified via specialized constructors - e.g., `AccountComponentTemplate::from_package()`. Is this how you were thinking about it too?

bobbinth (commented):
Thank you! Looks good! This is not a review yet - I mostly looked only at the package-related changes and left some comments/questions above.

One thing that I wanted to clarify: with the things as they are here, we can only deserialize from bytes right? So, if we had a TOML file, for example (containing a custom section) we would need a deserialization logic to live somewhere else. If so, where would this logic live - e.g., in `miden-base`, in the compiler, somewhere else?

package/src/package/section.rs
bitwalker: The idea behind non-`Custom` variants here is to represent "well-known" custom sections, which are important in more than one place in the toolchain. By assigning a well-known section identifier to those, the encoding is more compact, and it is implied we can reason about the format of the data in that section (i.e. the section isn't completely opaque).

We don't _have_ to do that with account component metadata, but since that was previously the only actual custom section we were emitting anywhere (i.e. from the compiler), it seems a bit odd to treat it as completely opaque to packages in general (since any account package will effectively require it).

package/src/package/section.rs
bitwalker: IMO it makes the encoding more compact for all packages, only requiring overhead for storing section names when the section is ad-hoc/something not emitted by default for all or a subset of packages.

package/src/package/mod.rs
bitwalker: Makes sense to me to have those as optional fields, definitely

package/src/package/section.rs
bitwalker: Yep, more or less - you might work directly with `Package` if you don't need much more than what can be accessed from just the package itself, or you'd do something like what you've described, and "lift" the `Package` into a higher-level equivalent format that converts any custom sections into their concrete in-memory representations.

bitwalker:
> One thing that I wanted to clarify: with the things as they are here, we can only deserialize from bytes right? So, if we had a TOML file, for example (containing a custom section) we would need a deserialization logic to live somewhere else. If so, where would this logic live - e.g., in miden-base, in the compiler, somewhere else?

The format for a given custom section would be defined in whatever crate is the canonical owner of that custom section's format, IMO. For account component metadata, that would be `miden-base` presumably (or `miden-objects` to be more precise I suppose).

(De)serializers for any given custom section might not have a single canonical implementation, but I would assume that the crate which "owns" the format of the section would certainly provide both. To use account component metadata as an example, the compiler can't depend on `miden-objects` (it does now, but it causes a ton of problems), so the compiler will have its own serializer for the account component metadata section, and its implementation will need to track against the format described by `miden-objects`. This necessarily implies that changes to the account component metadata section format in `miden-objects` cannot be made arbitrarily without consideration for breaking changes, and ideally, we'd attempt to make all changes backwards-compatible (or upgradeable from older formats), where possible. Conversely, it implies that the compiler must remain aware of changes to that section format, and integrate changes to it as part of its own release process.

plafer:
@bitwalker I'll wait for the cleanup before I review - let me know if you're instead waiting for my review before cleanup 😄

bitwalker:
@plafer This should be ready for your review. Note that the `serde` serialization impl for MAST is naive currently, since I'm waiting for other changes to MAST (e.g. the `nop` changes and such) to land before attempting to improve on that, but I'll do so in a later PR.

package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+
+impl TryFrom<u8> for SectionId {
+    type Error = InvalidSectionIdError;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(Self::DebugInfo),
+            10 => Ok(Self::AccountComponentMetadata),
+            _ => Err(InvalidSectionIdError::Tag(value)),
```
plafer: Don't we want a variant for `255 => Custom`?

package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+}
+
+impl SectionId {
+    pub fn custom(name: impl AsRef<str>) -> Result<Self, InvalidSectionIdError> {
```
plafer, line 34: We should add a docstring which documents the naming rules

package/src/package/section.rs
plafer: > By assigning a well-known section identifier to those, the encoding is more compact,

I'm missing some context from the compiler, but aren't we just saving some 20-30 bytes for a section name such as `account_component_metadata` or similar? 

> and it is implied we can reason about the format of the data in that section (i.e. the section isn't completely opaque).

Can you expand a bit on how having a well-known ID allows us to reason about the format more than a well-known constant such as `account_component_metadata`?

So far, philosophically, the VM has been designed to be as independent in design from the rollup as possible (although naturally the constraints of the rollup has influenced some design decisions). Making some rollup concepts first-class such as this `AccountComponentMetadata` variant at the very least breaks that illusion - after which we should IMO outright say that the VM is designed solely to work well with the rollup, and abandon any hope that the VM could be adopted by any other project.

crates/utils/bite/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "miden-bite"
```
plafer: What was the impetus for defining a new encoding format? Was there [none available](https://david.kolo.ski/rust_serialization_benchmark/) that had the features we need?

package/src/package/section.rs
bitwalker: No, it's only valid to convert from an `u8` to a `SectionId` if the identifier is a well-known section, since otherwise we require the section name (and obviously can't get that from just a `u8`).

package/src/package/section.rs
bitwalker: > I'm missing some context from the compiler, but aren't we just saving some 20-30 bytes for a section name such as account_component_metadata or similar?

It has always been expressed to me that the size of the package format is paramount, i.e. avoiding encoding unnecessary strings as much as possible is something we wish to do.

So even if the size savings here is relatively small per section, and there are unlikely to be many "well-known" sections, every single package is going to have at least one of these, and possibly multiple of them. So instead of using 26 bytes for `account_component_metadata` in every single package containing that section, we just need 1. Similarly for `debug_info` (or whatever we'd want to call that one.

If we don't care about those savings, then I'm fine with simplifying this to always use strings for section identifiers.

> Can you expand a bit on how having a well-known ID allows us to reason about the format more than a well-known constant such as account_component_metadata?

It isn't the integer vs string distinction that I'm talking about there, but the distinction between having the notion of a "well-known" section _at all_. For example, debug information (and the layout we ultimately select for that section) is going to be effectively present in every package that gets produced; that section is "well-known" in the sense that we own it, and will document the encoding used, and provide tooling to decode its contents.

I agree that it is arguable whether or not the distinction is going to be useful - but I did want to make a distinction in this regard similar to what was done with the Wasm binary format (which has sections that are part of the base format, as well as support for custom sections that are not part of the base format).

> So far, philosophically, the VM has been designed to be as independent in design from the rollup as possible (although naturally the constraints of the rollup has influenced some design decisions). Making some rollup concepts first-class such as this AccountComponentMetadata variant at the very least breaks that illusion - after which we should IMO outright say that the VM is designed solely to work well with the rollup, and abandon any hope that the VM could be adopted by any other project.

To be clear, even though the account component metadata section is being considered "well-known" here, note that the _encoding_ of that section is not at all specified here (or indeed anywhere in this repo). That is the responsibility of `miden-base`.

The thing for me that distinguishes that section from say, some third-party defined section, is that it is a pretty fundamental piece of the Miden toolchain in general. It is true that the VM itself is agnostic to these sorts of details, but the VM is a very low-level piece of the pie. The package format is really the primary means of interchange between the various Miden toolchain components (compiler, client, etc.), and so I don't see it as a fundamental issue that we favor core Miden tools in a way that we don't third-party ones.

In any case, I think ultimately the tradeoff I was making here was primarily one of keeping the encoding format as compact as possible.

crates/utils/bite/Cargo.toml
bitwalker: The main design goals:

1. Designed specifically for use with `serde`'s data model
2. Retain control over details of the encoding so that we can optimize for the specific constraints of Miden's use case(s)

There are any number of good binary serialization formats out there, but they are either incompatible with some of `serde`'s API surface due to lack of self-description (e.g. `skip_serializing_if`, `default`, etc.), or are fully self-describing in a way that bloats the binary size unnecessarily.

I don't actually care if we use this format or another one, I mostly just wanted a proof of concept that demonstrated that we could implement a compact serialization format on top of `serde` that would be suitable with Miden (e.g. for use with MAST and/or the package format). Whether that is this or something else is something I don't have a strong opinion on, personally.

crates/utils/bite/README.md
```diff
@@ -0,0 +1,28 @@
...
+* _Booleans are intrinsically-tagged_. In other words, the encoding of `true` and `false`, as `1` and `0` respectively, are also valid type tags for the boolean type. So storing a boolean requires only a single byte for both type tag and value, rather than a byte for each.
+* _`Option<T>` is intrinsically tagged_. Similar to booleans, the value for `None` is just the type tag for `None`, while `Some` is encoded as the tag for `Some` followed by the encoded value of type `T`.
+
+## Structure
+
+The structure of a BITE-encoded stream is described using the Kaitai Struct language in [bite.ksy].
```
bobbinth: nit: we should probably link "Katai Struct language" to https://kaitai.io - right?

Also, let's add the "License" section at the end of this file - similar to how we have it in other README files.

package/src/package/section.rs
```diff
@@ -0,0 +1,267 @@
...
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// A unique identifier/tag for optional sections of the Miden package format
+///
+/// The following tag ranges are reserved for specific categories of use:
+///
+/// * `0` - unused
+/// * `1..10` - reserved for non-rollup use cases, e.g. debug information
+/// * `10..100` - reserved for rollup use cases, e.g. account component metadata
```
bobbinth: nit: I'd probably change "rollup" to "blockchain" here.

package/src/package/serialization.rs
bobbinth, line 1: We should probably update the description of the format on lines 1 - 35.

package/src/package/serialization.rs
```diff
@@ -164,6 +167,110 @@ impl Deserializable for MastArtifact {
 // PACKAGE MANIFEST SERIALIZATION/DESERIALIZATION
 // ================================================================================================
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for PackageManifest {
```
bobbinth, line 187: Question: we are implementing custom serializer/deserializer for `PackageManifest` mainly to skip serializing keys in the `exports` `BTreeMap`, right?

package/src/package/section.rs
bobbinth: I don't mind having a notion of "well-known" sections (and probably have a slight preference to keeping them) - but to clarify/comment on a couple of points:

> It has always been expressed to me that the size of the package format is paramount, i.e. avoiding encoding unnecessary strings as much as possible is something we wish to do.

I think the main concern is the size of the `MastForest` - with everything else we have a bit more flexibility. The reason for this is that MAST data goes on-chain (e.g., code of public accounts and scripts of public notes go into the blocks of the chain), while higher level data (like libraries, packages) does not need to live on-chain.

> For example, debug information (and the layout we ultimately select for that section) is going to be effectively present in every package that gets produced; that section is "well-known" in the sense that we own it, and will document the encoding used, and provide tooling to decode its contents.

Currently, debug info is part of the `MastForest`. With https://github.com/0xMiden/miden-vm/issues/1776, it would be split off from the rest of the actual code, but I was thinking it would still remain in the `MastForest` (e.g., as outlined in https://github.com/0xMiden/miden-vm/issues/1776#issuecomment-29735883650). We could move it to the package level, but then we'd need to make the VM "package-aware", which probably has its own pros and cons. So, basically, it isn't clear to me yet if there is benefit of moving debug info out of `MastForest`.

assembly-syntax/src/library/namespace.rs
```diff
@@ -239,3 +250,101 @@ impl Deserializable for LibraryNamespace {
         Self::new(name).map_err(|e| DeserializationError::InvalidValue(e.to_string()))
     }
 }
+
+#[cfg(feature = "serde")]
+impl serde::Serialize for LibraryNamespace {
```
bobbinth, line 259: The serialization/deserialization code below looks quite complex. What are the benefits of this vs. serializing and deserializing strings/ (like we do for the Winterfell-based approach)?

assembly-syntax/src/ast/procedure/name.rs
```diff
@@ -157,6 +159,103 @@ impl Deserializable for QualifiedProcedureName {
...
+            let mut builder = serializer.serialize_struct("QualifiedProcedureName", 2)?;
+            builder.serialize_field("module", &self.module)?;
+            builder.serialize_field("name", &self.name)?;
+            builder.end()
+        }
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for QualifiedProcedureName {
```
bobbinth, line 187: Similar comment as above: this deserialization logic looks quite complex (especially if we compare with the Winterfell-based approach) - is there some way to simplify it? Or are there specific properties we gain by introducing this complexity?

crates/utils/bite/src/varint.rs
```diff
@@ -0,0 +1,41 @@
...
+pub fn encode(mut value: u64, buf: &mut Vec<u8>) {
+    loop {
+        if value < 0x80 {
+            buf.push(value as u8);
+            break;
+        }
+        buf.push((value as u8 & 0x7f) | 0x80);
+        value >>= 7;
+    }
+}
```
bobbinth: In Winterfell, we use a different variable-length encoding technique which does not require loops (see [here](https://github.com/facebook/winterfell/blob/main/utils/core/src/serde/byte_writer.rs#L77) and [here](https://github.com/facebook/winterfell/blob/main/utils/core/src/serde/byte_reader.rs#L123)). Is there a reason to chose this approach over that one?

crates/utils/bite/bite.ksy
bobbinth, line 1: Question: I'm not too familiar with Kaitai Struct language, is this file used for documentation purposes or do we generate encoding/decoding logic from it?

crates/utils/bite/src/de/mod.rs
bobbinth, line 1: This is fine for a PoC, but if we want to make this the main serialization format we should document the code in this file (and other files in this crate) much more extensively.

core/src/kernel.rs
```diff
@@ -15,6 +17,8 @@ use crate::{
 /// The internally-stored list always has a consistent order, regardless of the order of procedure
 /// list used to instantiate a kernel.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
 pub struct Kernel(Vec<Word>);
```
bobbinth, line 22: Question: is there a way to tell `serde` that the vector in here can have at most 256 entries? With current Winterfell-based serialization, we get this "for free". I'm assuming for `serde` we'd need to write custom serializers/deserializers - or maybe there is a better way to do it?

bobbinth (commented):
Thank you! Looks good! I looked through most of the code though still need to go through the `bite` serialization/deserialization logic in much more detail. Overall, it looks really cool, but if we are to switch to it through out, it would need more eyes on it.

A couple of general questions/comments:

As mentioned in one of the comments, the main thing where the size is critical is the `MastForest` struct. Here we can assume that we'll be serializing it with all debug-related info stripped out and this will probably be done outside of the package serialization. So, for package serialization purposes, we can be more flexible on this front (e.g., we don't need to go the extra mile to squeeze out every last byte).

Is the plan for `bite` encoder to be used only for package serialization, or will we use it more broadly? If so, we might want to move it into a separate repo in the future.

Overall, I'd probably slit out the package-specific changes (i.e., adding custom sections) into a separate PR as we can probably merge them in the next day or two (assuming we align on the treatment of custom section identifiers). The serialization changes still have some open questions/comments and may take longer to merge.

assembly-syntax/src/library/namespace.rs
bitwalker: The string-based approach is used for human-readable formats, but for binary formats we can do much better, so the serialization code here picks the best method based on that criteria.

assembly-syntax/src/ast/procedure/name.rs
bitwalker: It's important to bear in mind that `serde` serialization code is used for _any_ format, human-readable or binary. It is also necessary to map types to/from `serde`'s data model, when implementing `Serialize`/`Deserialize` manually. One must also consider how a type will appear in human-readable formats, as well as how it will be encoded in binary formats. In human-readable formats, say JSON, you really want a representation that is convenient for humans to work with, which isn't always as rich/precise as the actual in-memory representation. In binary formats, you want a representation that will be compact as possible, while encoding the type accurately.

In the case of `QualifiedProcedureName`, a human wants to just see the path as a string, because that conveys all the information we care about. However, if we're encoding to a binary format that can supports string interning/de-duplication, then we can take advantage of that by encoding `QualifiedProcedureName` as a struct which separates the module path from the procedure name, ensuring that we only ever need to encode a given module path as a string once in the entire output. The same applies to the procedure name - because it is encoded separately from the module path, any occurrence of that procedure name only gets encoded as a string once. This adds up to a substantial reduction in binary size.

IMO, the serialization code here isn't actually complex, because it is pretty easy to understand and reason about once you are familiar with `serde`'s data model - but it does require more boilerplate when the representation of a type differs between human-readable and binary formats; or when you want tighter control over the serialized representation. However, the cost here is far outweighed by the benefits, namely that we can support serialization to essentially any arbitrary format that has a `serde` serializer.

crates/utils/bite/src/varint.rs
bitwalker: Nope, using vint64 for this works just as well (possibly better)

crates/utils/bite/bite.ksy
bitwalker: I'm using it for documentation purposes primarily, though one can use it to generate serializers/deserializers in various languages. I figured may as well kill two birds with one stone if I was going to try and write up higher-level documentation on the format.

crates/utils/bite/src/de/mod.rs
bitwalker: Yeah this is a prototype, largely to demonstrate the viability of using `serde` to replace our use of Winterfell serialization primitives. It is viable for production use, but figured we'd probably be iterating on it a bit before finalizing the format.

core/src/kernel.rs
bitwalker: Similar to the Winterfell-based (de)serializer, we could implement `Serialize`/`Deserialize` manually to assert that the encoded length is < 256. There isn't any declarative way to do that, but if it was a common-enough use case, it would make sense to define a wrapper type for `Vec` for which we implement `Serialize`/`Deserialize` manually to handle imposing some size limit on the vector, and use it in place of `Vec`, something like `SizedVec<T, N>` where `N` is a constant value indicating the max capacity. One benefit of doing that is we could also ensure that it is impossible to construct an instance of the vector which violates that rule period.

With the above approach, this type definition would remain almost the same as it is here, but using `SizedVec<Word, 255>` instead of `Vec<Word>`.

package/src/package/serialization.rs
bitwalker: Yeah, encoding as a sequence is more compact than as a map, and avoids redundantly encoding the same information in both the keys and values - but it's an optimization, not a requirement.

bitwalker:
> .. if we are to switch to it through out, it would need more eyes on it.

Yeah, as discussed I will split it out from this PR anyway, but the primary intent was to demonstrate the viability of `serde` as our sole serialization framework, rather than using both `serde` and Winterfell for different types of serialization.

> As mentioned in one of the comments, the main thing where the size is critical is the `MastForest` struct. Here we can assume that we'll be serializing it with all debug-related info stripped out and this will probably be done outside of the package serialization. So, for package serialization purposes, we can be more flexible on this front (e.g., we don't need to go the extra mile to squeeze out every last byte).

Agreed, though I think it is still a good idea to avoid needlessly bloating binary representations if we can avoid it - but we definitely don't need to go to great lengths for the package format (ignoring the MAST component).

> Is the plan for `bite` encoder to be used only for package serialization, or will we use it more broadly? If so, we might want to move it into a separate repo in the future.

My intent was for handling the binary serialization requirements in the VM repo (so, package format and MAST basically). I don't think it is necessary to try and use it everywhere, but it could be useful in other contexts where we don't already have an off-the-shelf format in mind.

package/src/package/section.rs
bitwalker: Update: I've reimplemented `SectionId` to always use a string identifier for now

Cargo.toml
```diff
@@ -40,6 +40,7 @@ overflow-checks = true
 [workspace.dependencies]
 # Workspace crates
 miden-air = { path = "./air", version = "0.18", default-features = false }
+miden-bite = { path = "./crates/utils/bite", version = "0.18", default-features = false }
```
plafer: This should go in #2130 right?

assembly-syntax/src/ast/ident.rs
```diff
@@ -199,6 +199,27 @@ impl FromStr for Ident {
     }
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for Ident {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.as_str())
```
plafer, line 208: Don't we want to serialize the `Ident.span` field too?

assembly-syntax/src/ast/procedure/name.rs
```diff
@@ -157,6 +159,103 @@ impl Deserializable for QualifiedProcedureName {
...
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for QualifiedProcedureName {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
+            let name = format!("{}", self);
+            serializer.serialize_str(&name)
```
plafer, line 174: Similar question to `Ident`: don't we care about serializing the span?

assembly-syntax/src/ast/procedure/name.rs
```diff
@@ -157,6 +159,103 @@ impl Deserializable for QualifiedProcedureName {
...
+                                return Err(serde::de::Error::duplicate_field("name"));
+                            }
+                            name = Some(map.next_value()?);
+                        },
+                    }
+                }
+                let module = module.ok_or_else(|| serde::de::Error::missing_field("module"))?;
+                let name = name.ok_or_else(|| serde::de::Error::missing_field("name"))?;
+                Ok(QualifiedProcedureName::new(module, name))
+            }
```
plafer, line 252: Why do we need both `visit_seq()` and `visit_map()` here? And more generally, how does the call to `serialize_struct()` that we make in `Serializer` map to those calls in `Deserializer`?

plafer (commented):
Just a partial review on my end.

I'm a little rusty on the specifics of serde, and will need to brush up before I finish this review.

package/src/package/serialization.rs
```diff
@@ -34,7 +34,13 @@
 //!       - `name` (`String`)
 //!       - `digest` (`Word`)
```
bobbinth: We should probably update the description above to mention things like version, description, and custom sections.

bobbinth (approved):
Looks good! Thank you! I left one doc-related comment inline.

I did not review all the custom serde serializers/deserializers - it would be great to get another set of eyes on them.

core/src/kernel.rs
bobbinth: I like this approach! I wonder if we should add `SizedVec<T, N>` in `miden-crypto` so that we could use it there anywhere else downstream.

greenhat: We could also add non-empty and lower bounds requirements and do something like https://github.com/ergoplatform/bounded-vec

assembly-syntax/src/library/mod.rs
```diff
@@ -280,8 +284,12 @@ impl Library {
 /// - There must be at least one exported procedure.
 /// - The number of exported procedures cannot exceed [Kernel::MAX_NUM_PROCEDURES] (i.e., 256).
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "Library"))]
```
huitseeker, line 318: I would love to see a roundtrip test here, because I would have used `#[cfg_attr(feature = "serde", serde(into = "Library"))]`. The worry is that deserializer will have to deal with a struct equipped with a single inner library field that it may not be able to deserialize into a `Library`.

huitseeker (commented):
This is in general super solid work! Serde attributes are well-used, and the deserializer visitors are simple enough.

I have found that it's super hard (for me at least) to remember everything and in particular figure the right embedding levels for structs to have their `Deserialize` impl match their `Serialize` impl well. See my comment on `KernelLibrary` for an instance of what I mean.

So I recommend relying on extensive round-trip tests. I have opened #2132 against this PR to help : this introduces a macro that generates roundtrip tests pegged on `Arbitrary` implems. With this, I have observed the following failing tests:

```
FAIL [   0.008s] miden-assembly-syntax ast::procedure::name::test_serde_roundtrip_QualifiedProcedureName_0
        FAIL [   0.007s] miden-assembly-syntax library::namespace::test_serde_roundtrip_LibraryNamespace_0
        FAIL [   0.007s] miden-assembly-syntax library::test_serde_roundtrip_LibraryExport_0
        FAIL [   0.008s] miden-assembly-syntax library::test_serde_roundtrip_Library_0
```

I hope this helps.

bitwalker:
> So I recommend relying on extensive round-trip tests. ..snip.. I have observed the following failing tests:
> 
> ```
> FAIL [   0.008s] miden-assembly-syntax ast::procedure::name::test_serde_roundtrip_QualifiedProcedureName_0
> FAIL [   0.007s] miden-assembly-syntax library::namespace::test_serde_roundtrip_LibraryNamespace_0
> FAIL [   0.007s] miden-assembly-syntax library::test_serde_roundtrip_LibraryExport_0
> FAIL [   0.008s] miden-assembly-syntax library::test_serde_roundtrip_Library_0
> ```

See my comment [here](https://github.com/0xMiden/miden-vm/pull/2132#issuecomment-3267392645). Agreed though that we should have a more extensive set of serialization tests to cover the new serde-based serialization code - though as noted in my linked comment, we should probably wait until we choose an appropriate binary format that supports all the necessary features (e.g. `miden-bite`).

assembly-syntax/src/library/mod.rs
bitwalker: Good call!

bitwalker: Note that the actual roundtrip test would only catch the issue if its implementation emits some kind of prelude/tag for structs, but that isn't typically the case in the implementations I've looked at.

That said, a strict interpretation of the data model would dictate that the approach you suggest here is used, so we should change it accordingly - I'm just not 100% sure the tests would catch a failure.

assembly-syntax/src/ast/procedure/name.rs
bitwalker: Because in some formats the structure will be  encoded as a sequence, and in others, a map - so which visitor method is called depends on the specific format being deserialized (or more precisely, the implementation of the deserializer). 

The serializer will have encoded a struct as either a sequence or a map (typically a sequence for binary formats, a map for human-readable ones). The deserializer will either know which visitor method to invoke (because it is expecting a specific type next in the input), or one of multiple possibilities is expected next, and the deserializer invokes `deserialize_any` to look for a type tag in the input that will tell the deserializer what to deserialize next, using the appropriate visitor method.

assembly-syntax/src/ast/procedure/name.rs
bitwalker: We never serialize these spans, no.

assembly-syntax/src/ast/ident.rs
bitwalker: No we never serialize the `SourceSpan` type, because it is not a stable value. We use the `Location` type for this, which is explicitly serialized as part of the `AssemblyOp` decorator.

huitseeker:
Ah, of course. I have pushed a commit that switches the test serializer to `serde_json` — which indeed fixes a number of issues —, and provides a more rigorous instance of `Arbitrary` for the `Library` type.

There is still one test failure on the `Library` roundtrip, which seems slightly more significant, as it exposes the serialization's field order. Have we documented what determinism we expect out of the serializers Miden would be compatible with (miden-bite, if we've made the decision this is the only such one)?

bitwalker:
For the JSON format specifically, field order doesn't actually matter (see the approach used when implementing `deserialize_map` for an example of this). For binary formats, I think it is reasonable to specify that the order of fields in structs corresponds to the order expected in the input, but we have not specified that anywhere - although to be clear, we haven't actually really specified any of these details anywhere (except to some degree for the package format, which we plan to change anyway, but it also omits details of how certain objects are encoded).

For `miden-bite` at least, the specification can be described for arbitrary Rust types, so then the encoding of a specific type depends on what `serde` attributes you add to customize the behavior for a given type.

bitwalker:
This should be ready to merge now, so anyone who wants to take a final pass through, feel free. I'd suggest we shoot to merge this by EOD tomorrow or Thursday at the latest

package/src/package/mod.rs
```diff
@@ -1,30 +1,45 @@
...
     /// The package manifest, containing the set of exported procedures and their signatures,
     /// if known.
     pub manifest: PackageManifest,
-    /// Serialized `miden-objects::account::AccountComponentMetadata` for the account component
-    /// (name, descrioption, storage,) associated with this package, if any.
-    pub account_component_metadata_bytes: Option<Vec<u8>>,
+    /// The set of custom sections included with the package, e.g. debug information, account
+    /// metadata, etc.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub sections: Vec<Section>,
```
lima-limon-inc, line 42: I know this is an old PR, but could a `Package` potentially contain two sections with the same `SectionId`?

If not, would it make sense to add an `add_section` function that checks that a `Package` does not have two sections with the same `SectionId`? As a way to "guarantee" that said property holds true

Issue #2011, opened 2025-07-21
Add ability to declare word-sized constants from strings


PhilippGackstatter:
We would like to be able to declare word-sized constants from strings, kind of how it already works for error messages. Writing:

```
const.METADATA_SLOT_NAME="miden::basic_fungible_faucet::metadata"
```

would declare a word-sized constant that is the hash of the given string. For error messages, I believe this is currently just one field element, but here we would like to get a hash that is a full word. Similar to error messages, this can be the BLAKE3 hash and doesn't have to be an RPO hash.

One thought is whether this should be explicit in which hash is used, in case we do want this mechanism to (at some later point) also be usable to create constants based on RPO hashes, in which case it would be possible to disambiguate right there. So something like the following is one idea:

```
const.METADATA_SLOT_NAME=hash.blake3("miden::basic_fungible_faucet::metadata")
```

Context: https://github.com/0xMiden/miden-base/discussions/1394#discussioncomment-13809514

bitwalker:
> `const.METADATA_SLOT_NAME="miden::basic_fungible_faucet::metadata"`

I don't think this should declare a word-sized constant on its own - just a string constant, as those are already used in other contexts (error messages).

> `const.METADATA_SLOT_NAME=hash.blake3("miden::basic_fungible_faucet::metadata")`

Something like this is more appropriate IMO, though I would strip the `hash.` prefix, as it complicates the grammar unnecessarily, and just do:

```
const.METADATA_SLOT_NAME=blake3("miden::basic_fungible_faucet::metadata")
```

bobbinth:
> Something like this is more appropriate IMO, though I would strip the `hash.` prefix, as it complicates the grammar unnecessarily, and just do:
> 
> ```
> const.METADATA_SLOT_NAME=blake3("miden::basic_fungible_faucet::metadata")
> ```

I wonder if we need to specify the hash algorithm used here - maybe something like this is better?
```
const.METADATA_SLOT_NAME=hash("miden::basic_fungible_faucet::metadata")
```

An alternative, could be to use the following syntax:

```
const.METADATA_SLOT_NAME="miden::basic_fungible_faucet::metadata"

begin
    push.hash(METADATA_SLOT_NAME)[0..2]
end
```
Though, this may be actually more confusing.

PhilippGackstatter:
> I wonder if we need to specify the hash algorithm used here

[...]

PR #2073, opened 2025-08-13
Declaration of word-sized constants from strings

Fumuran:
This PR implements a feature which allows to create a word-sized constants from the strings. This could be done using `word` syntax like so:
```masm
const.SAMPLE_WORD=word("lorem ipsum")
```

Closes: #2011.

### DIFF
--- a/assembly-syntax/src/ast/constants.rs
+++ b/assembly-syntax/src/ast/constants.rs
@@ -87,8 +87,8 @@ impl Spanned for Constant {
 /// Represents a constant expression or value in Miden Assembly syntax.
 #[derive(Clone)]
 pub enum ConstantExpr {
-    /// A literal integer value.
-    Literal(Span<Felt>),
+    /// A literal [`Felt`] value.
+    FeltLiteral(Span<Felt>),
     /// A reference to another constant.
     Var(Ident),
     /// An binary arithmetic operator.
@@ -98,9 +98,12 @@ pub enum ConstantExpr {
         lhs: Box<ConstantExpr>,
         rhs: Box<ConstantExpr>,
     },
+    /// A plain spanned string.
     String(Ident),
-    Word(Span<WordValue>),
-    HashWord(Ident, WordValue),
+    /// A literal [`WordValue`].
+    WordLiteral(Span<WordValue>),
+    /// A spanned string along with its hash represented by the [`WordValue`].
+    WordHash(Ident, WordValue),
 }
 
 impl ConstantExpr {
@@ -111,7 +114,7 @@ impl ConstantExpr {
     #[track_caller]
     pub fn expect_literal(&self) -> Felt {
         match self {
-            Self::Literal(spanned) => spanned.into_inner(),
+            Self::FeltLiteral(spanned) => spanned.into_inner(),
             other => panic!("expected constant expression to be a literal, got {other:#?}"),
         }
     }
@@ -132,10 +135,10 @@ impl ConstantExpr {
     pub fn try_fold(self) -> Result<Self, ParsingError> {
         match self {
             Self::String(_)
-            | Self::Literal(_)
+            | Self::FeltLiteral(_)
             | Self::Var(_)
-            | Self::Word(_)
-            | Self::HashWord(..) => Ok(self),
+            | Self::WordLiteral(_)
+            | Self::WordHash(..) => Ok(self),
             Self::BinaryOp { span, op, lhs, rhs } => {
                 if rhs.is_literal() {
                     let rhs = Self::into_inner(rhs).try_fold()?;
@@ -143,7 +146,7 @@ impl ConstantExpr {
                         Self::String(ident) => {
                             Err(ParsingError::StringInArithmeticExpression { span: ident.span() })
                         },
-                        Self::Literal(rhs) => {
+                        Self::FeltLiteral(rhs) => {
                             let lhs = Self::into_inner(lhs).try_fold()?;
                             match lhs {
                                 Self::String(ident) => {
@@ -151,7 +154,7 @@ impl ConstantExpr {
                                         span: ident.span(),
                                     })
                                 },
-                                Self::Literal(lhs) => {
+                                Self::FeltLiteral(lhs) => {
                                     let lhs = lhs.into_inner();
                                     let rhs = rhs.into_inner();
                                     let is_division =
@@ -162,18 +165,18 @@ impl ConstantExpr {
                                     }
                                     match op {
                                         ConstantOp::Add => {
-                                            Ok(Self::Literal(Span::new(span, lhs + rhs)))
+                                            Ok(Self::FeltLiteral(Span::new(span, lhs + rhs)))
                                         },
                                         ConstantOp::Sub => {
-                                            Ok(Self::Literal(Span::new(span, lhs - rhs)))
+                                            Ok(Self::FeltLiteral(Span::new(span, lhs - rhs)))
                                         },
                                         ConstantOp::Mul => {
-                                            Ok(Self::Literal(Span::new(span, lhs * rhs)))
+                                            Ok(Self::FeltLiteral(Span::new(span, lhs * rhs)))
                                         },
                                         ConstantOp::Div => {
-                                            Ok(Self::Literal(Span::new(span, lhs / rhs)))
+                                            Ok(Self::FeltLiteral(Span::new(span, lhs / rhs)))
                                         },
-                                        ConstantOp::IntDiv => Ok(Self::Literal(Span::new(
+                                        ConstantOp::IntDiv => Ok(Self::FeltLiteral(Span::new(
                                             span,
                                             Felt::new(lhs.as_int() / rhs.as_int()),
                                         ))),
@@ -183,7 +186,7 @@ impl ConstantExpr {
                                     span,
                                     op,
                                     lhs: Box::new(lhs),
-                                    rhs: Box::new(Self::Literal(rhs)),
+                                    rhs: Box::new(Self::FeltLiteral(rhs)),
                                 }),
                             }
                         },
@@ -207,7 +210,9 @@ impl ConstantExpr {
 
     fn is_literal(&self) -> bool {
         match self {
-            Self::Literal(_) | Self::String(_) | Self::Word(_) | Self::HashWord(..) => true,
+            Self::FeltLiteral(_) | Self::String(_) | Self::WordLiteral(_) | Self::WordHash(..) => {
+                true
+            },
             Self::Var(_) => false,
             Self::BinaryOp { lhs, rhs, .. } => lhs.is_literal() && rhs.is_literal(),
         }
@@ -225,9 +230,9 @@ impl Eq for ConstantExpr {}
 impl PartialEq for ConstantExpr {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
-            (Self::Literal(l), Self::Literal(y)) => l == y,
+            (Self::FeltLiteral(l), Self::FeltLiteral(y)) => l == y,
             (Self::Var(l), Self::Var(y)) => l == y,
-            (Self::HashWord(l_s, l_h), Self::HashWord(r_s, r_h)) => l_s == r_s && l_h == r_h,
+            (Self::WordHash(l_s, l_h), Self::WordHash(r_s, r_h)) => l_s == r_s && l_h == r_h,
             (
                 Self::BinaryOp { op: lop, lhs: llhs, rhs: lrhs, .. },
                 Self::BinaryOp { op: rop, lhs: rlhs, rhs: rrhs, .. },
@@ -241,11 +246,11 @@ impl core::hash::Hash for ConstantExpr {
     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
         core::mem::discriminant(self).hash(state);
         match self {
-            Self::Literal(value) => value.as_int().hash(state),
+            Self::FeltLiteral(value) => value.as_int().hash(state),
             Self::String(value) => value.hash(state),
-            Self::Word(value) => value.hash(state),
+            Self::WordLiteral(value) => value.hash(state),
             Self::Var(value) => value.hash(state),
-            Self::HashWord(string, hash) => {
+            Self::WordHash(string, hash) => {
                 string.hash(state);
                 hash.hash(state);
             },
@@ -261,10 +266,10 @@ impl core::hash::Hash for ConstantExpr {
 impl fmt::Debug for ConstantExpr {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Self::Literal(lit) => fmt::Debug::fmt(&**lit, f),
-            Self::Word(w) => fmt::Debug::fmt(&**w, f),
+            Self::FeltLiteral(lit) => fmt::Debug::fmt(&**lit, f),
+            Self::WordLiteral(w) => fmt::Debug::fmt(&**w, f),
             Self::Var(name) | Self::String(name) => fmt::Debug::fmt(&**name, f),
-            Self::HashWord(str, hash) => fmt::Debug::fmt(&(str, hash), f),
+            Self::WordHash(str, hash) => fmt::Debug::fmt(&(str, hash), f),
             Self::BinaryOp { op, lhs, rhs, .. } => {
                 f.debug_tuple(op.name()).field(lhs).field(rhs).finish()
             },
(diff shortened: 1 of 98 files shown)


assembly-syntax/src/ast/constants.rs
```diff
@@ -100,6 +100,7 @@ pub enum ConstantExpr {
     },
     String(Ident),
     Word(Span<WordValue>),
+    HashWord(Ident, WordValue),
```
bobbinth: Let's add doc comments for these variants.

Also, a couple of thoughts on naming:
- Should we rename `Word` into `WordLiteral` or maybe `LiteralWord`? I think this could make the naming more consistent.
- If we go with the above, we probably would need to rename `Literal` into `FeltLiteral` or `LiteralFelt`.
- And then, `HashWord` could be `HashedWord` or `DerivedWord` or `WordHash`.

assembly-syntax/src/ast/constants.rs
```diff
@@ -239,6 +245,10 @@ impl core::hash::Hash for ConstantExpr {
             Self::String(value) => value.hash(state),
             Self::Word(value) => value.hash(state),
             Self::Var(value) => value.hash(state),
+            Self::HashWord(string, hash) => {
+                string.hash(state);
+                hash.hash(state);
+            },
```
bobbinth: Why do we need to both to the `state`? Shouldn't `hash` (or `string` for that matter) uniquely define the object?

assembly-syntax/src/sema/context.rs
```diff
@@ -76,9 +77,17 @@ impl AnalysisContext {
...
-            ConstantExpr::Literal(_) | ConstantExpr::String(_) => Ok((*value).clone()),
-            ConstantExpr::Word(_) => Ok((*value).clone()),
+            ConstantExpr::Literal(_) | ConstantExpr::String(_) | ConstantExpr::Word(_) => {
+                Ok((*value).clone())
+            },
             ConstantExpr::Var(name) => self.get_constant(name).cloned(),
+            ConstantExpr::HashWord(string, _) => {
+                let digest_bytes: [u8; 32] = Blake3_256::hash(string.as_bytes()).into();
+                let hash_word = Word::try_from(digest_bytes)
+                    .expect("conversion of the Blake3_256 hash into the Word has failed");
```
bobbinth: There is a small probability that this could fail because the output of BLAKE3 is not guaranteed to produce valid field elements. One option to address this is to interpret each digest as four `u64` values and then use `Felt::new()` to convert them to filed elements (this will silently wrap around invalid `u64` values).

Also, I'd probably create a helper function for this as we could use it in other places as well (e.g., when we has strings into field elements for error codes).

assembly/src/tests.rs
```diff
@@ -1313,6 +1314,45 @@ fn const_conversion_failed_to_u32() -> TestResult {
...
+#[test]
+fn const_word_from_string() -> TestResult {
+    let context = TestContext::default();
+    let sample_source_string = "lorem ipsum";
+
+    let source = source_file!(
+        &context,
+        format!(
+            r#"
+    const.SAMPLE_WORD=to_word("{sample_source_string}")
```
bobbinth: We are using `to_word` here because making `word` a keyword causes issues as it collides with the `word` module in the standard library. I don't really love having `to_word` be a keyword either - so, curious what could be the alternatives.

Maybe we could use some special notation instead - e.g., something like:
```
const.MY_WORD=<"my_string">
```
But I'm not sure i like this approach either. @bitwalker, @adr1anh, @PhilippGackstatter - what do you think?

bobbinth (commented):
Looks good! Thank you! I left a few comments inline. The main one is about figuring out if we can replace `to_word` with anything else.

assembly-syntax/src/ast/constants.rs
Fumuran: I decided to preserve the original string to be able to get it if needed, as Paul mentioned. I'm not sure how exactly such data could be used, probably during the debugging or error handling, but if you think that in that case it is redundant — I can remove it.

Edit: probably I can move the process of the hash computation to the stage of the constant evaluation, in that case we will be able to get rid of the hash here and leave only the string ident.

adr1anh:
> Looks good! Thank you! I left a few comments inline. The main one is about figuring out if we can replace `to_word` with anything else.

It not clear how often we would use this feature, and if we expect it to mainly be defined as constants, a longer name shouldn't be too disturbing. What about `hash_to_word` or `map_to_word`?

Fumuran:
>It not clear how often we would use this feature, and if we expect it to mainly be defined as constants, a longer name shouldn't be too disturbing. What about `hash_to_word` or `map_to_word`?

I also think that slightly longer name would be fine in that case, other name variants could be `string_to_word`, `string_to_hash`, or even something like `wordify` (by analogy with `stringify!`).

bobbinth:
> I also think that slightly longer name would be fine in that case, other name variants could be `string_to_word`, `string_to_hash`, or even something like `wordify` (by analogy with `stringify!`).

I think out of these `string_to_word` probably makes the most sense. Though, @bitwalker mentioned that it may be possible to use `word` in this context without introducing ambiguity with the `std::word` module name. @bitwalker could you take a look?

assembly/src/tests.rs
bitwalker: I don't understand the bit about this colliding with the `word` module - it sounds to me like the issue is that the `word` token was added without also adding it as a valid identifier token (in a different part of the grammar, and is needed for any token which is also a valid identifier).

bobbinth: @Fumuran - could we try what @bitwalker is suggesting to see if we can make it work with just `word(..)` notation?

Fumuran: @bitwalker But what do you mean by the valid identifier token? So I added the `word`/`to_word` string as a token to the `parser/token.rs` and to the related places in the grammar, but where else it should be added?

assembly-syntax/src/parser/grammar.lalrpop
```diff
@@ -8,6 +8,7 @@ use alloc::{
 use core::{marker::PhantomData, num::NonZeroU32, ops::Range};
 
 use miden_debug_types::{SourceFile, SourceSpan, Span, Spanned};
+use miden_core::ZERO;
```
adr1anh: Should we more this to the `crate` import below?

assembly/src/tests.rs
bitwalker: @Fumuran If you look at the `BareIdent` rule in the grammar, you'll notice that it handles the `bare_ident` token, as well as all of the tokens which are also valid bare identifiers (which currently happens to be just opcode names, so we're delegating to the `OpcodeName` rule for that), we should add cases to `BareIdent` for things which are _not_ valid opcode names, but which are tokens that should be permitted as valid bare identifiers (e.g. `word`). If something _is_ also an opcode name, it would need to be added to `OpcodeName` instead, but that's not the case here.

Fumuran: Ah, now I see, thank you!

assembly-syntax/src/parser/grammar.lalrpop
Fumuran: `ZERO` is not imported to the `assembly-syntax`, so I had to import it directly from the `miden_core`. As an alternative we could import it in the `parser/mod.rs` to get it from the `super`, but it will require to import it from the core anyway either here or there. But I'm not sure if it's worth it considering that `ZERO` is used only in the grammar

assembly-syntax/src/ast/constants.rs
```diff
@@ -87,8 +87,8 @@ impl Spanned for Constant {
 /// Represents a constant expression or value in Miden Assembly syntax.
 #[derive(Clone)]
 pub enum ConstantExpr {
-    /// A literal integer value.
-    Literal(Span<Felt>),
+    /// A literal [`Felt`] value.
+    FeltLiteral(Span<Felt>),
```
bitwalker: Let's just call this `Felt`

assembly-syntax/src/ast/constants.rs
```diff
@@ -98,8 +98,12 @@ pub enum ConstantExpr {
         lhs: Box<ConstantExpr>,
         rhs: Box<ConstantExpr>,
     },
+    /// A plain spanned string.
     String(Ident),
-    Word(Span<WordValue>),
+    /// A literal [`WordValue`].
+    WordLiteral(Span<WordValue>),
```
bitwalker: Similarly, I think calling this just `Word` is descriptive enough.

assembly-syntax/src/ast/constants.rs
```diff
@@ -98,8 +98,12 @@ pub enum ConstantExpr {
...
+    /// A spanned string along with its hash represented by the [`WordValue`].
+    WordHash(Ident, WordValue),
```
bitwalker: I wonder if we really need to compute the hash here, but additionally, I think it might be cleaner to instead have something like:

```
enum ConstantExpr {
    Hash(HashKind, Ident),
    ...
}

enum HashKind {
    Word,
    Blake3,
    ...
}
```

This keeps the handling of `ConstantExpr` variants simple, while still providing semantic information about how to hash the given string.

It doesn't seem particularly worthwhile to compute the hash into the AST itself, particularly because it bloats the size of `ConstantExpr` significantly in order to hold the array of 4 felts. It's much simpler to just compute it on demand (and it is unlikely that hashes would be re-computed all that frequently).

bitwalker (changes requested):
Looks good overall! I have a few nits/suggestions that I think are worth making before this is merged, but I think we're basically there. Sorry for all the back and forth!

assembly-syntax/src/ast/constants.rs
Fumuran: Indeed, that would be better. But do you mean by the `Word` and `Blake3` variants in the `HashKind `? As far as I know, error messages, events and words constructed from strings (so, in fact, every user provided input) will use the Blake3 hash, which in turn returns a word.

bobbinth: I don't think we need the support for multiple hash functions. So, just `WordHash(Ident)` is probably fine.

> It doesn't seem particularly worthwhile to compute the hash into the AST itself, particularly because it bloats the size of `ConstantExpr` significantly in order to hold the array of 4 felts.

Just to clarify this point: hash is just 32 bytes - so, it won't add much more to the size - or at lest not much more than a simple `Word` constant.

assembly-syntax/src/sema/context.rs
```diff
@@ -171,3 +180,18 @@ impl AnalysisContext {
     #[cfg(not(feature = "std"))]
     fn emit_warnings(self) {}
 }
+
+// HELPER FUNCTIONS
+// ================================================================================================
+
+/// Converts the digest obtained from the [`Blake3_256::hash`] procedure into the [`Word`].
+fn blake3_digest_to_word(blake3_digest: Blake3Digest<32>) -> Word {
```
bobbinth: I would probably make this function more generic. Maybe something like:
```rust
fn hash_string_to_word(value: impl Into<&str>) -> Word {
    ...
}
```
Also, should probably move this into the `utils` module of `miden-core` and export it from there.

Lastly, can we use this function for hashing error messages too? (I think we use the same methodology there). We don't need to change how error messages work, but somewhere, to get the error code, we hash the error string using BLAKE3, I believe.

bobbinth (commented):
Looks good! Thank you! I left one comment inline.

assembly-syntax/src/ast/constants.rs
bobbinth: Though, thinking about it a bit more, maybe including the `HashKind` makes sense, though, I'd make it not `Word` vs. `Blake3`. Maybe something like this:

```rust
enum ConstantExpr {
    Hash(HashKind, Ident),
    ...
}

enum HashKind {
    Felt, // not needed now, but in the future could be used to reduce a string to a single felt
    Word, // reduces a string to a word
    Event, // in the future reduces a string to event
    ...
}
```

bitwalker: Yes, my understanding is that we will have various different "types" of hashes, depending on the thing being hashed. I think personally, I'd have the constant just be a string, and have the value type be determined by the way the string is used, e.g. `emit.CONST` would reduce it to an event identifier; `push.CONST` would reduce it to a word, and so on.

In any case, structuring it as I described above allows us future extensibility without requiring a bunch of changes to the AST, just the addition of the `HashKind`, appropriate handling for that, and whatever syntax we require.

I should note, my choice of `Word` and `Blake3` above were just examples that have come up in comments/discussion, not the actual hash kinds we need - something more along the lines of what @bobbinth mentioned seems more appropriate. That said, I think `Word` _should_ be `Blake3`, to be explicit about the hash algorithm used, so that if we change the defaults in the future, we have a path for backwards-compatibility.

assembly-syntax/src/sema/context.rs
bitwalker: Maybe `fn digest_to_word(value: impl AsRef<[u8; 32]>) -> Word`? This ensures that the input is something we can represent as a `Word`, but doesn't imply any specific hash algorithm (i.e. the caller has already hashed the input to some 32 byte digest).

assembly-syntax/src/ast/constants.rs
bitwalker: > Just to clarify this point: hash is just 32 bytes - so, it won't add much more to the size - or at lest not much more than a simple `Word` constant.

The need for the `Ident` field means that the `WordValue` portion is an _extra_ 32 bytes to the overall size of `ConstantExpr`, i.e. it is not the same size as the `Word` variant.

assembly-syntax/src/sema/context.rs
bobbinth: I was thinking of having a "canonical" way of reducing strings to words so that we can use it in other places. For example, we may want to compute storage slot names in `miden-base` in Rust (e.g., for testing purposes or otherwise) - and we'd use this function there to get the same result as we'd get in MASM.

It was also be awesome if this function was `const fn` - this way, we could even initialize constants in Rust that correspond to words we'd get in MASM by hashing the same string.

assembly-syntax/src/ast/constants.rs
```diff
@@ -325,3 +342,27 @@ impl fmt::Display for ConstantOp {
...
+
+/// Represents the type of the final value to which some string value should be converted.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum HashKind {
+    /// Reduce a string to a single felt value (unused right now)
+    Felt,
+    /// Reduce a string to a word using Blake3 hash function
+    Blake3,
+    /// Reduce a string to event (unused right now)
+    Event,
```
adr1anh: This would support hashing to a `Felt` right?

adr1anh (approved):
Looks good to me!

assembly-syntax/src/ast/constants.rs
Fumuran: Probably yes, it will depend on our decision how do we want an event value to be computed ¯\_(ツ)_/¯

bobbinth: Let's use `Word` here instead of `Blake3` as all these variants will use BLAKE3 for hashing.

Also, it is not clear to me if we need to add the currently unused variants - especially the `Felt` variant as I'm not sure what we'll do with it.

core/src/utils/mod.rs
```diff
@@ -39,6 +42,21 @@ impl ToElements for Vec<u64> {
     }
 }
 
+// TO WORD
+// ================================================================================================
+
+/// Converts the digest obtained from the [`Blake3_256::hash`] procedure into the [`Word`].
+pub fn hash_string_to_word<'a>(value: impl Into<&'a str>) -> Word {
```
bobbinth: nit: I would write this as something like: "Hashes the provided string using the BLAKE3 hash function and converts the resulting digest into a [`Word`]."

Also, can we update the `error_code_from_msg()` function to use this function under the hood?

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline, once these are addressed, I think we should be good to merge.

assembly-syntax/src/ast/constants.rs
bobbinth: > That said, I think `Word` _should_ be `Blake3`, to be explicit about the hash algorithm used, so that if we change the defaults in the future, we have a path for backwards-compatibility.

I actually think just `Word` here is fine because all other variants will use the same hash function as well, and we also implicitly use BLAKE3 as the hash function for hashing error messages as well. So, it is kind of like a "canonical" way of hashing string in the assembler.

If we do want to make the hash function explicit, this should be applied to other variants too - and we'd have something like `Blake3Word`, `Blake3Felt`, `Blake3Event`, and probably should also somehow make it explicit for error message hashing. But as mentioned above, I don't think that's necessary.

Also, I'm not sure backward-compatibility reasons apply here because, AFAIK, we don't serialize the AST anywhere, and we don't want to surface the underlying hash function in the source code. So, if there is a breaking change, it would be something like v1 assembler used BLAKE3 for hashing strings and v2 assembler is using something else.

assembly-syntax/src/ast/constants.rs
Fumuran: Make sense. I'll remove the unused variants in that case.

Fumuran:
@bitwalker I hope this PR is ready to merge, could you confirm and approve if so?

Issue #2068, opened 2025-08-11
Add support for emitting events from stack values


partylikeits1983:
### Feature description

Currently, to emit an event, the syntax looks like this:

```
emit.SOME_EVENT
```

However, it would be extremely useful to be able to emit an event from a value on the stack. For example:

```
push.SOME_EVENT emit
```

CC-ing @mmagician @bobbinth 

### Why is this feature needed?

This functionality would be useful in this context: https://github.com/0xMiden/miden-base/pull/1599#discussion_r2265479452

plafer:
If we took the emitted value from the stack (where `emit` would be equivalent to a `drop` with respect to its effect on the stack), I believe the operation could be only degree 2, and we could set its opcode to `001_1111` (as suggested in #2043). This would then free up a high-degree slot.

Syntactically in MASM, we could have 2 forms `emit` and `emit.VALUE` which would work similar to e.g. `mem_store` vs `mem_store.addr`.

bobbinth:
> Syntactically in MASM, we could have 2 forms `emit` and `emit.VALUE` which would work similar to e.g. `mem_store` vs `mem_store.addr`.

Given that `emit` is used rather sparsely, I think spending 2 cycles on `emit.CONST` is totally fine. So, I'm +1 on this.

PR #2075, opened 2025-08-15
Refactor and simplify codebase

### LINKED ISSUE #1583: Consider spliting event IDs into "source" and "event" space
As a follow-up to #1457.

Context: 
- main issue comment: https://github.com/0xPolygonMiden/miden-vm/issues/1457#issue-2471210549
- PR discussion: https://github.com/0xPolygonMiden/miden-vm/pull/1581#discussion_r1853114792

adr1anh:
DRAFT

## Describe your changes

Fixes #1583.
Implemented with help from Claude Code.

- Updated `CHANGELOG.md'


assembly-syntax/src/ast/visit.rs
```diff
@@ -322,8 +325,8 @@ where
         | MemLoadWImm(imm)
         | MemStoreImm(imm)
         | MemStoreWImm(imm)
-        | Emit(imm)
         | Trace(imm) => visitor.visit_immediate_u32(imm),
+        Emit(event_value) => visitor.visit_event_value(Span::new(span, event_value)),
```
bitwalker, line 329: I'm not sure why this is being wrapped in a `Span` here - both variants of `EventValue` contain an `Immediate<T>`, which already wraps its inner value in a `Span`

adr1anh: I’m not super familiar with the assembler, and still trying to work through what makes sense or not. Thanks for taking a look!

As a more general question though, does it make sense to support "legacy" IDs in the form of a `u32`, or transition fully to named events? That would break backwards compatibility and it's not clear what our policy around that is.

However if we do want to support both, does it make sense to have an enum for both possible types? When playing around with the grammar it seems like it makes it ambiguous.

assembly/src/instruction/mod.rs
```diff
@@ -547,8 +549,43 @@ impl Assembler {
...
+                        // Legacy numeric event ID - convert directly to Felt
+                        EventID::from(id.expect_value())
+                    },
+                    EventValue::Name(name) => {
+                        // String-based event name - parse and register in EventTable
+                        let event_name = match name {
+                            Immediate::Value(s) => s.inner().clone(),
+                            Immediate::Constant(_) => {
+                                return Err(Report::new(
+                                    RelatedLabel::error("event names cannot be constants yet")
```
bitwalker, line 564: The actual error here is that the constant name was not resolved to a concrete definition/value prior to lowering - that's the responsibility of semantic analysis (and soon the linker). If we hit an unresolved constant at this point, we should raise a panic/assertion that says we expected the constant to have been resolved by this point.

bitwalker (commented):
Just out of curiosity - has there been any discussion around how this stuff will be surfaced in packages? Are the set of events "exported" from, say, the standard library part of the metadata in some way?

Since constants are not yet exportable/importable, it would be kind of an all-or-nothing thing at the moment, but it's probably something we should figure out before releasing the functionality.

assembly-syntax/src/ast/visit.rs
bitwalker: @bobbinth would be the best person to ask regarding whether to keep around the raw u32 event identifiers - but to answer your question, yes, the approach you've taken here is more or less the right one.

The reason the grammar is ambiguous is because of how the rules for the two variations are being expressed - both the legacy and new formats allow for a reference to a constant declaration containing the actual value, and so in the case where a constant name is the next token after `emit.`, it can't determine which of the rules match, as it could be either the legacy or new format.

What you need to do to address the ambiguity is to have one rule for when a constant name is encountered (in which case we don't know yet whether the event value is in legacy or new form), and then a rule each for a u32 literal and string literal (in which case the format is known by definition).

With that in mind, you want something more like the following:

```rust
enum EventValue {
    Legacy(u32),
    New(Arc<str>),
}

enum Instruction {
    Emit(Immediate<EventValue>),
    ...,
}
```

`Immediate` takes care of the "is this a constant reference, or a literal; while `EventValue` takes care of representing the case where the value is a literal.

bobbinth:
Initially, I was thinking of this PR simply as changing `Operation::Emit(u32)` to `Operation::Emit(Felt)`, but in light of #2068 we should probably change the approach slightly.

Specifically, if we go with #2068, the enum variant becomes just `Operation::Emit` (without the inner value). This, in turn, would mean that we cannot restrict `Emit` operation to work with special type of constants. The result is that we can split the work into two PRs:

The first PR would address #2068 (and this would make #1583 obsolete). This PR would allow the following syntax to invoke trigger events:
```
const.MY_EVENT=1234

begin
    # push an immediate value onto the stack and emit the event
    push.1234 emit

    # same as above, but use the constant instead of immediate value
    push.MY_EVENT emit

     # more concise form, but gets transformed into `push.1234 emit` by the assembler
    emit.MY_EVENT
end
```

The second PR would allow defining event codes from strings. This could look something like this:
```
const.MY_EVENT=event("my_library::my_event")

begin
    # both of the approaches below would be allowed
    push.MY_EVENT emit
    emit.MY_EVENT
end
```
This would require making `event` a keyword in the grammar, which I think is OK.

The above implies that we'd have a way to convert the provided string into the a felt and for this we can use the approach similar to what @Fumuran did in https://github.com/0xMiden/miden-vm/pull/2073 (i.e., we just hash the user-provided string and convert the resulting digest into a felt).

Regarding this, there are two open questions:
- When converting a string into a felt, should we use the un-modified string or should we transform it somehow? One potentially useful transformation could be to prepend library namespace to the user-provide string. This would reduce the chance of collisions if people try to use the same event name in different libraries (and don't follow the "my_library::my_event" convention). My thinking is that if this doesn't create too much extra work, it is probably a good thing to do.
- Should we save the mapping from strings to their corresponding felts in the `MastForest`. I think this could be a good addition, but I would probably leave it for a follow-up (i.e. 3rd) PR.

> @bobbinth would be the best person to ask regarding whether to keep around the raw u32 event identifiers

As mentioned above, I don't think we need to keep the legacy format around and this will become meaningless anyway if we go with the proposal from #2068.

> Just out of curiosity - has there been any discussion around how this stuff will be surfaced in packages? Are the set of events "exported" from, say, the standard library part of the metadata in some way?
> 
> Since constants are not yet exportable/importable, it would be kind of an all-or-nothing thing at the moment, but it's probably something we should figure out before releasing the functionality.

My current thinking is that events are something internally to a given library and so they don't need to be exported. Basically, if we define events in some library, then this library should provide handlers for handling these events (right now this would done in Rust, but in the future each event handler would probably be complied to WASM and included with the package). Ideally, the users of the library should not even know that the library defines/uses some events.

adr1anh:
Thanks for the feedback! I'll leave this PR on hold for now as I agree with the implementation plan you suggested.

PR #2075 is soon ready and implements the first issue by removing the immediate from the Emit operation and pulling the value from the stack instead.

bobbinth:
This has been superseded by #2083, right?

adr1anh:
Mostly, yes. The only remaining task is to define the the syntax for events to allow them to be defined as strings. That will happen in PR #2085

PR #2076, opened 2025-08-16
feat: Remove unused Push(U8|U16|U32|Felt)List

### LINKED ISSUE #2053: Remove unused `Push*List` instructions

huitseeker:
Closes #2053


bitwalker (changes requested):
We need to retain `PushFeltList`, otherwise 👍

huitseeker:
This was already addressed in #2057 .

PR #2077, opened 2025-08-16
feat: Remove Push(U8|U16|U32|Felt|Word)

### LINKED ISSUE #2066: Investigate removal of the dedicated `Push*` instructions
It seems like in all our use cases we could use the `Push<Immediate>` instruction instead of the dedicated (typed) ones, so they could be safely removed. The list of the instructions which could be removed is here:
- `PushU8`
- `PushU16`
- `PushU32`
- `PushFelt`

huitseeker:
Closes #2066

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Changes
 
 - Removed unused `PushU8List`, `PushU16List`, `PushU32List` and `PushFeltList` instructions ([#2057](https://github.com/0xMiden/miden-vm/pull/2057)).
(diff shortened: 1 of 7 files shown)


bitwalker (approved):
Still not sure how I feel about removing some of these, but there isn't really a compelling reason to keep them atm, so 👍

bobbinth (approved):
Looks good! Thank you!

> Still not sure how I feel about removing some of these, but there isn't really a compelling reason to keep them atm,

Yeah, if we need them for some reason in the future, we can always bring them back.

PR #2078, opened 2025-08-17
fix: report *all* erroneous values on a NotU32Values

### LINKED ISSUE #1327: When both values are not u32 the error for instruction u32assert2 should carry both
When execution code like the following:

```
begin
    push.4294967297
    push.4294967296
    u32assert2
end
```

The VM error should carry information about all invalid values. Currently the top of the stack "wins":

https://github.com/0xPolygonMiden/miden-vm/blob/9abd8d17483dc7a3b23ebb72ab30b86426ddf19d/processor/src/operations/u32_ops.rs#L35-L40

Having both values makes a little bit easier to debug some situations.

huitseeker:
Closes #1327

### DIFF
--- a/processor/src/fast/u32_ops.rs
+++ b/processor/src/fast/u32_ops.rs
@@ -1,9 +1,36 @@
 use alloc::vec::Vec;
 
 use miden_core::{Felt, ZERO};
+use paste::paste;
 
 use super::FastProcessor;
 use crate::{ErrorContext, ExecutionError, utils::split_element};
+
+const U32_MAX: u64 = u32::MAX as u64;
+
+macro_rules! require_u32_operands {
+    ($processor:expr, [$($idx:expr),*], $err_ctx:expr) => {
+        require_u32_operands!($processor, [$($idx),*], ZERO, $err_ctx)
+    };
+    ($processor:expr, [$($idx:expr),*], $errno:expr, $err_ctx:expr) => {{
+        let mut invalid_values = Vec::new();
+
+        paste!{
+            $(
+                let [<operand_ $idx>] = $processor.stack_get($idx);
+                if [<operand_ $idx>].as_int() > U32_MAX {
+                    invalid_values.push([<operand_ $idx>]);
+                }
+            )*
+
+            if !invalid_values.is_empty() {
+                return Err(ExecutionError::not_u32_values(invalid_values, $errno, $err_ctx));
+            }
+            // Return tuple of operands based on indices
+            ($([<operand_ $idx>].as_int()),*)
+        }
+    }};
+}
 
 impl FastProcessor {
     /// Analogous to `Process::op_u32split`.
@@ -30,30 +57,10 @@ impl FastProcessor {
     /// The size of the stack is decremented by 1.
     #[inline(always)]
     pub fn op_u32add3(&mut self, err_ctx: &impl ErrorContext) -> Result<(), ExecutionError> {
-        let (sum_hi, sum_lo) = {
-            let c = self.stack_get(0).as_int();
-            let b = self.stack_get(1).as_int();
-            let a = self.stack_get(2).as_int();
+        let (c, b, a) = require_u32_operands!(self, [0, 1, 2], err_ctx);
 
-            // Collect all invalid u32 values before returning an error
-            let mut invalid_values = Vec::new();
-
-            if a > u32::MAX as u64 {
-                invalid_values.push(Felt::new(a));
-            }
-            if b > u32::MAX as u64 {
-                invalid_values.push(Felt::new(b));
-            }
-            if c > u32::MAX as u64 {
-                invalid_values.push(Felt::new(c));
-            }
-
-            if !invalid_values.is_empty() {
-                return Err(ExecutionError::not_u32_values(invalid_values, ZERO, err_ctx));
-            }
-            let result = Felt::new(a + b + c);
-            split_element(result)
-        };
+        let result = Felt::new(a + b + c);
+        let (sum_hi, sum_lo) = split_element(result);
 
         // write the high 32 bits to the new top of the stack, and low 32 bits after
         self.decrement_stack_size();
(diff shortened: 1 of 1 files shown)


processor/src/errors.rs
```diff
@@ -213,6 +213,15 @@ pub enum ExecutionError {
         value: Felt,
         err_code: Felt,
     },
+    #[error("operation expected u32 values, but got values: {values:?} (error code: {err_code})")]
+    NotU32Values {
```
plafer, line 208: While the issue only refers to `u32assert2`, I think we'd want this to be fixed for all u32 operations - i.e. replace all uses of `NotU32Value` with `NotU32Values`

processor/src/fast/u32_ops.rs
```diff
@@ -234,12 +236,18 @@ impl FastProcessor {
...
+            invalid_values.push(Felt::new(first_old));
         }
         if second_old > u32::MAX as u64 {
-            return Err(ExecutionError::not_u32_value(Felt::new(second_old), err_code, err_ctx));
+            invalid_values.push(Felt::new(second_old));
+        }
+
+        if !invalid_values.is_empty() {
+            return Err(ExecutionError::not_u32_values(invalid_values, err_code, err_ctx));
         }
```
plafer: This makes the error a bit hard to parse, since we lose the information of which value is incorrect (e.g. if only 1 value is incorrect, was the first one or second one incorrect?). I think we should prioritize encoding that information somehow, even if it makes the error message not as clean, e.g. `"u32 operation expected u32 inputs, but at least one was not a u32 value. First: OK, Second: not OK"` (except better). Haven't thought about how easy/hard this would be with miette.

What do you think?

processor/src/errors.rs
bobbinth: Maybe we could change `require_u32_operand!` macro into `require_u32_operands!` macro. This could work like so:
```rust
let (a, b) = require_u32_operands!(self.stack, err_ctx);
```
This should handle all cases except for `op_u32add3` and `op_u32madd` - which is annoying - but maybe there is a way to handle these ones too.

adr1anh: This might be a good moment to consolidate `NotU32StackValue`

huitseeker: I've added a commit that replaces `NotU32Value` with `NotU32Values`

huitseeker: Added `NotU32Value` removal and adapted the macro.

processor/src/fast/u32_ops.rs
huitseeker: If we want to push such positional information, I don't see how to do it without embarking an explicitly indexed list in the error (`[(idx, val),*]`). This seems like a meaningfully increased scope of work, is this something we want to tackle in this PR?

plafer: No, let's leave it as is then and fix it in the future if users complain about it

processor/src/operations/u32_ops.rs
```diff
@@ -61,8 +91,9 @@ impl Process {
     /// Pops two elements off the stack, adds them, splits the result into low and high 32-bit
     /// values, and pushes these values back onto the stack.
     pub(super) fn op_u32add(&mut self, err_ctx: &impl ErrorContext) -> Result<(), ExecutionError> {
-        let b = require_u32_operand!(self.stack, 0, err_ctx).as_int();
-        let a = require_u32_operand!(self.stack, 1, err_ctx).as_int();
+        let (b, a) = require_u32_operands!(self.stack, [0, 1], err_ctx);
```
adr1anh, line 77: Could we return the converted `u64` values to avoid having to convert them again?

processor/src/errors.rs
```diff
@@ -204,13 +204,13 @@ pub enum ExecutionError {
         source_file: Option<Arc<SourceFile>>,
         value: Felt,
     },
-    #[error("operation expected a u32 value, but got {value} (error code: {err_code})")]
-    NotU32Value {
+    #[error("operation expected u32 values, but got values: {values:?} (error code: {err_code})")]
+    NotU32Values {
```
adr1anh, line 208: Copying the comment here. Can we merge this error with `NotU32Stack` value?

processor/src/operations/u32_ops.rs
plafer: We can probably start considering `Process` as legacy - especially in that any kind of optimization shouldn't be prioritized.

Sort of applies here, but also more generally for future PRs.

Cargo.toml
```diff
@@ -60,6 +60,7 @@ midenc-hir-type = { version = "0.1", default-features = false }
 
 # Third-party crates
 log = "0.4"
+paste = "1.0.15"
```
bobbinth: nit: we usually don't specify patch versions - i.e., `1.0.15` -> `1.0`.

processor/src/fast/u32_ops.rs
```diff
@@ -33,15 +35,21 @@ impl FastProcessor {
...
+                invalid_values.push(Felt::new(b));
             }
             if c > u32::MAX as u64 {
-                return Err(ExecutionError::not_u32_value(Felt::new(c), ZERO, err_ctx));
+                invalid_values.push(Felt::new(c));
+            }
+
+            if !invalid_values.is_empty() {
+                return Err(ExecutionError::not_u32_values(invalid_values, ZERO, err_ctx));
             }
```
bobbinth: Question: if all values are valid, this shouldn't result in vector allocation, right? Also, if we wanted to optimize it even more for the happy path, we could probably do something like this:
```rust
if a > u32::MAX as u64 || b > u32::MAX as u64 || c > u32::MAX as u64 {
    // TODO: build the error here
}
```

processor/src/fast/u32_ops.rs
```diff
@@ -93,14 +101,20 @@ impl FastProcessor {
...
-                return Err(ExecutionError::not_u32_value(Felt::new(b), ZERO, err_ctx));
+                invalid_values.push(Felt::new(a));
             }
             if c > u32::MAX as u64 {
-                return Err(ExecutionError::not_u32_value(Felt::new(c), ZERO, err_ctx));
+                invalid_values.push(Felt::new(c));
+            }
+            if !invalid_values.is_empty() {
+                return Err(ExecutionError::not_u32_values(invalid_values, ZERO, err_ctx));
             }
```
bobbinth: Similar comment as above.

processor/src/operations/u32_ops.rs
```diff
@@ -46,8 +60,24 @@ impl Process {
...
+        }
+
+        // Check second value (a at stack[1])
+        if a.as_int() > U32_MAX {
+            invalid_values.push(a);
+        }
+
+        if !invalid_values.is_empty() {
+            return Err(ExecutionError::not_u32_values(invalid_values, err_code, err_ctx));
+        }
```
bobbinth: Question: why not use the macro here as well?

bobbinth (approved):
Looks good! Thank you! I left a few small questions/comments inline.

processor/src/fast/u32_ops.rs
huitseeker: `Vec::new` doc, second line:
> The vector will not allocate until elements are pushed onto it.

processor/src/operations/u32_ops.rs
huitseeker: Yes, of course!

processor/src/errors.rs
huitseeker: What are you suggesting here? Deleting `NotU32StackValue` and losing the specificity of the error message, notably in terms of the clock cycle report? Adding the clock cycle everywhere?

adr1anh: The latter was my thinking indeed. The clock should be available from the `process` when errors are constructed. But I had not taken that into account when writing the comment, and it may lead to a larger refactor than I thought.

This should be easy to do once th3 processor refactors are merged, and we can tackle it when removing the `ErrorContext`.

huitseeker: Opened https://github.com/0xMiden/miden-vm/issues/2087

PR #2081, opened 2025-08-19
feat(CLI): add `multicall` support

lima-limon-inc:
## Describe your changes

This PR makes the VM's CLI multicall compatible. With this, the CLI is now aware of whether it is being called under the `miden-vm` name or a different one. This is used in midenup's invocation of the vm, which is done under the `miden vm` command.

The `Cli` struct now stores this information, in the `external` field; however, this is not currently being used.

Multicall awareness is being tracked in this issue (https://github.com/0xMiden/midenup/issues/70) in midenup.

- [X] Repo forked and branch created from `next` according to naming convention.
- [X] Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- [X] Relevant issues are linked in the PR description.

### DIFF
--- a/miden-vm/src/main.rs
+++ b/miden-vm/src/main.rs
@@ -1,4 +1,6 @@
-use clap::Parser;
+use std::ffi::OsString;
+
+use clap::{FromArgMatches, Parser, Subcommand};
 use miden_assembly::diagnostics::Report;
 #[cfg(feature = "tracing-forest")]
 use tracing_forest::ForestLayer;
@@ -20,9 +22,50 @@ pub(crate) mod utils;
     version,
     rename_all = "kebab-case"
 )]
+#[command(multicall(true))]
+pub struct MidenVmCli {
+    #[command(subcommand)]
+    behavior: Behavior,
+}
+
+impl From<MidenVmCli> for Cli {
+    fn from(value: MidenVmCli) -> Self {
+        match value.behavior {
+            Behavior::MidenVm { cli } => cli,
+            Behavior::External(args) => Cli::parse_from(args).set_external(),
+        }
+    }
+}
+
+#[derive(Debug, Subcommand)]
+#[command(rename_all = "kebab-case")]
+enum Behavior {
+    /// The Miden VM CLI.
+    MidenVm {
+        #[command(flatten)]
+        cli: Cli,
+    },
+
+    /// Used when the Miden VM CLI is called under a different name, like
+    /// when it is called from [Midenup](https://github.com/0xMiden/midenup).
+    /// Vec<OsString> holds the "raw" arguments passed to the command line,
+    /// analogous to `argv`.
+    #[command(external_subcommand)]
+    External(Vec<OsString>),
+}
+
+#[derive(Parser, Debug)]
+#[command(name = "miden-client")]
 pub struct Cli {
     #[command(subcommand)]
     action: Actions,
+
+    /// Indicates whether the vm's CLI is being called directly, or externally
+    /// under an alias (like in the case of
+    /// [Midenup](https://github.com/0xMiden/midenup).
+    #[arg(skip)]
+    #[allow(unused)]
+    external: bool,
 }
 
 /// CLI actions
@@ -54,12 +97,20 @@ impl Cli {
             Actions::Repl(repl) => repl.execute(),
         }
     }
+
+    fn set_external(mut self) -> Self {
+        self.external = true;
+        self
+    }
 }
 
 /// Executable entry point
 pub fn main() -> Result<(), Report> {
     // read command-line args
-    let cli = Cli::parse();
+    let cli = <MidenVmCli as clap::CommandFactory>::command();
+    let matches = cli.get_matches();
+    let parsed = MidenVmCli::from_arg_matches(&matches).unwrap_or_else(|err| err.exit());
+    let cli: Cli = parsed.into();
 
     initialize_diagnostics();
 
@@ -105,5 +156,39 @@ fn initialize_diagnostics() {
     #[cfg(not(feature = "std"))]
     {
         let _ = reporting::set_hook(Box::new(|_| Box::new(ReportHandlerOpts::new().build())));
+    }
+}
+
+// TESTS
+// ================================================================================================
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    /// Check that the VM recognizes when it is being called under an alias.
+    fn test_mutlicall() {
+        // Direct call
+        let miden_vm_command =
+            MidenVmCli::try_parse_from(["miden-vm", "repl"]).expect("failed to parse commands");
+
+        assert!(matches!(
+            miden_vm_command.behavior,
+            Behavior::MidenVm { cli: Cli { external: false, .. } }
+        ));
+
+        let cli: Cli = miden_vm_command.into();
+        assert!(matches!(cli, Cli { external: false, .. }));
+
+        // External call
+
+        // This recreates how the vm would be called from midenup
+        let external =
+            MidenVmCli::try_parse_from(["miden vm", "repl"]).expect("failed to parse commands");
+
+        assert!(matches!(external.behavior, Behavior::External(_)));
+
+        let cli: Cli = external.into();
+        assert!(matches!(cli, Cli { external: true, .. }));
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Removed dedicated `PushU8`, `PushU16`, `PushU32`, `PushFelt`, and `PushWord` assembly instructions. These have been replaced with the generic `Push<Immediate>` instruction which supports all the same functionality through the `IntValue` enum (U8, U16, U32, Felt, Word) ([#2066](https://github.com/0xMiden/miden-vm/issues/2066)).
 - [BREAKING] Update miden-crypto dependency to v0.16 (#[2079](https://github.com/0xMiden/miden-vm/pull/2079))
 - Made `get_mast_forest()` async again for `AsyncHost` now that basic conditional async support is in place ([#2060](https://github.com/0xMiden/miden-vm/issues/2060)).
+- Added `multicall` support for the CLI ([#1141](https://github.com/0xMiden/miden-vm/pull/2081))
 
 ## 0.17.0 (2025-08-06)
 


miden-vm/src/main.rs
```diff
@@ -20,9 +22,50 @@ pub(crate) mod utils;
...
+    /// Used when the Miden VM CLI is called under a different name, like
+    /// when it is called from [Midenup](https://github.com/0xMiden/midenup).
+    /// Vec<OsString> holds the "raw" arguments passed to the command line,
+    /// analogous to `argv`.
+    #[command(external_subcommand)]
+    External(Vec<OsString>),
+}
+
+#[derive(Parser, Debug)]
+#[command(name = "miden-client")]
```
huitseeker: Is this the right name to use for the vm's CLI?

miden-vm/src/main.rs
```diff
@@ -20,9 +22,50 @@ pub(crate) mod utils;
...
+        }
+    }
+}
+
+#[derive(Debug, Subcommand)]
+#[command(rename_all = "kebab-case")]
+enum Behavior {
+    /// The Miden VM CLI.
+    MidenVm {
+        #[command(flatten)]
```
huitseeker, line 71: Should we use `#[command(visible_alias = "miden-vm")]` so we can get routed here when called from the right (sym)link?

huitseeker (commented):
This makes sense to me. Some inline comments.

Is there a risk of 3rd party users starting to depend on the `external_command` routing from clap early on before we figure out what structure we want to give to our CLI tools here? If so, and if we think it may be too early to commit to backwards compatibility, should we filter / sanity-check the args being passed on the `Cmd::External`, and return a nice error message if it's not what we expect?

miden-vm/src/main.rs
lima-limon-inc: Great catch! 
Fixed here: https://github.com/0xMiden/miden-vm/pull/2081/commits/886e690c0cd4a8e6fa25e59eb865479483f2a005

miden-vm/src/main.rs
lima-limon-inc: > Should we use #[command(visible_alias = "miden-vm")] so we can get routed here when called from the right (sym)link?

I believe this is the case with the code as it stands. 
When the executable is called under the "miden-vm" name, it will fall under the `Behavior::MidenVm` after `MidenVmCli::from_arg_matches` gets executed. 

Here's a snippet from clap's cargo expansion 

```rust
    fn from_arg_matches_mut(
        __clap_arg_matches: &mut clap::ArgMatches,
    ) -> ::std::result::Result<Self, clap::Error> {
        #![allow(deprecated)]
        if let Some((__clap_name, mut __clap_arg_sub_matches)) = __clap_arg_matches
            .remove_subcommand()
        {
            let __clap_arg_matches = &mut __clap_arg_sub_matches;
            if __clap_name == "miden-vm" && !__clap_arg_matches.contains_id("") {
                return ::std::result::Result::Ok(Self::MidenVm {
                    cli: <Cli as clap::FromArgMatches>::from_arg_matches_mut(
                        __clap_arg_matches,
                    )?,
                });
            }
```


The thing with the `Behavior` enum is that, despite being labeled as a subcommand, it is not a "visible" subcommand. It only exists to serves as a way for `multicall` to do the routing between the different possible executable names.

I've pushed this commit which adds documentation to the `Behavior` enum to explain its purpose: https://github.com/0xMiden/miden-vm/pull/2081/commits/0ac5492e8d4ef57c01a6e0e5419555e9c5aaa090

Do let me know if I'm missing something! Thanks in advance

lima-limon-inc:
> If so, and if we think it may be too early to commit to backwards compatibility, should we filter / sanity-check the args being passed on the `Cmd::External`, and return a nice error message if it's not what we expect?

Regarding alias parsing, I've pushed this commit https://github.com/0xMiden/miden-vm/pull/2081/commits/8f7423ef3484705bf0273e5d55f0780b82a876d6 where the used alias is parsed before executing. 

Sidenote: as far as I know, `midenup`'s used aliased are not yet formally decided, since it is still under development (ATTOW, this is being worked on this PR https://github.com/0xMiden/midenup/pull/73 and to a lesser extent here https://github.com/0xMiden/midenup/pull/84).

Was something along these lines similar to what you had in mind? Thanks in advance!

miden-vm/src/main.rs
huitseeker: That makes sense, thanks!

huitseeker:
> What something along these lines similar to what you had in mind? 

Yup, precisely. For as long as we're unsure about our own aliases / CLI ecosystem, I'm happier with a world where folks are incited to open an issue and chat with us before integrating the CLI. This achieves that.

miden-vm/src/main.rs
```diff
@@ -107,3 +184,37 @@ fn initialize_diagnostics() {
...
+        ));
+
+        let cli: Cli = miden_vm_command.try_into().expect("Failed to turn MidenVmCli into Cli.");
+        assert!(matches!(cli, Cli { external: false, .. }));
+
+        // External call
+
+        // This recreates how the vm would be called from midenup
+        let external =
+            MidenVmCli::try_parse_from(["miden vm", "repl"]).expect("failed to parse commands");
```
bitwalker, line 213: I'm not sure I understand why this would be the case, I'd expect that for `midenup`, it translates `miden vm repl` to `miden-vm repl` - there is little to no reason to make this binary multi-call just so that we can invoke it with arg[0] as `miden vm`.

The main reason to make a binary multi-call is so it can be invoked as two _different_ executables. Take `midenup` and `miden` - they are both implemented by the same binary, and `miden` is symlinked - the behavior you get is then inferred from arg[0].

In the case of the `miden-vm` binary, the reason to make it multi-call would be so that we can create symlinks that correspond to some/all of its subcommands, that can then be invoked directly, e.g. using `miden-vm repl` _or_ `repl`.

That's not what is implemented here though - there is no reason why we can't invoke the `miden-vm` executable any way we like (so long as we can contrive to execute it that way). From what I can tell, we're still expecting all the same arguments and subcommands though, so the actual behavior of the executable doesn't change based on whether you invoke it as `miden-vm repl` or `miden vm repl` (and the only way you can execute it as the latter, is if you use a syscall to create the process with "miden vm" as arg[0], which I'm not sure is even portable across all OS that we may need to support).

If we don't plan to surface subcommands as symlinks on disk, I don't think we should make this binary multi-call.

@lima-limon-inc Could you remind me what the context was with making this multi-call for `midenup`? I thought it was so we could create symlinks corresponding to parts of various executables that we wanted as part of the `miden` top-level, e.g. `miden repl` would invoke a `repl` symlink in the toolchain bin directory that would link to the `miden-vm` binary. It seems like the purpose of this PR is different though, and I think calls into question if we actually need to do this or not.

lima-limon-inc: > I'm not sure I understand why this would be the case, I'd expect that for `midenup`, it translates `miden vm repl` to `miden-vm repl` - there is little to no reason to make this binary multi-call just so that we can invoke it with arg[0] as `miden vm`.

Originally, `midenup` did translate `miden vm repl` to `miden-vm repl`, and it would later invoke the `miden-vm` executable with the passed in arguments.

However, if I'm not mistaken, this implementation has an unintended side effect related to how clap displays its help/"call format" messages.
Clap, as far as I can tell, uses the name of the file from which it was invoked as its `argv[0]`. This meant, that if a user ran `miden vm`, `midenup` would invoke the `miden-vm` binary; which would then produce the following output:

```
The Miden virtual machine

Usage: miden-vm <COMMAND>

Commands:
  analyze  Analyze a miden program
```

Notice the `miden-vm`, instead of `miden vm` in the "Usage:" section.

These messages are technically not "midenup" compatible, since the `miden-vm` executable, when installed via midenup, is not exposed to the user's PATH and thus copying and pasting clap's message would result in failure.

So, in order to circumvent this, I created a set of symlinks that follow midenup's `miden` format, such as: `miden vm`, `miden client`, etc. All of these symlinks point to their respective executables.

With this setup, when a user invokes `miden vm`, `midenup` will actually execute the `miden vm` symlink; which will produce this output:

```
The Miden virtual machine

Usage: miden vm <COMMAND>

Commands:
  analyze  Analyze a miden program
```

This means that all of clap's messages are now "midenup" compatible. Adittionally, while minor, this also comes with the benefit of being "executable name agnostic".
The downside with this setup is the prescense of a space character in the filename, although, these would only be invoked by midenup itself and not exposed to the user.


> The main reason to make a binary multi-call is so it can be invoked as two _different_ executables. Take `midenup` and `miden` - they are both implemented by the same binary, and `miden` is symlinked - the behavior you get is then inferred from arg[0].
>
> In the case of the `miden-vm` binary, the reason to make it multi-call would be so that we can create symlinks that correspond to some/all of its subcommands, that can then be invoked directly, e.g. using `miden-vm repl` _or_ `repl`.

I see; I did not have that model in mind. I do believe that implementation still has the help message "issue" if I'm not mistake; where clap diplays its messages according to the used file name and will circumvent `midenup`'s `miden` functionality. Such as:

```
$~ miden vm repl --help

Launch the Miden REPL

Usage: repl [OPTIONS]
```

But do let me know if this is not an issue. 

> (and the only way you can execute it as the latter, is if you use a syscall to create the process with "miden vm" as arg[0], which I'm not sure is even portable across all OS that we may need to support).

I understand; I didn't take OS compatibility into consideration. For context this has been mainly tested on MacOS. 

> @lima-limon-inc Could you remind me what the context was with making this multi-call for midenup? I thought it was so we could create symlinks corresponding to parts of various executables that we wanted as part of the miden top-level, e.g. miden repl would invoke a repl symlink in the toolchain bin directory that would link to the miden-vm binary. It seems like the purpose of this PR is different though, and I think calls into question if we actually need to do this or not.

The main motivation behind this implmentation was to make clap's help messages as "midenup compatible/friendly" as possible; but do let me if you'd like me to rework it with your comments in mind.

As an aside, that implementation can be seen here: https://github.com/0xMiden/midenup/pull/73


Thanks for the feedback!

bitwalker: @lima-limon-inc Thanks for the explanation!

I am a bit concerned about having symlinks with spaces in the name, in terms of portability to Windows, but I don't actually know if it is an issue in practice (I'd love to test, but I don't think we've tested hardly _anything_ on Windows at this point, so I'm not going to send you down that rabbit hole). That said, I think we can probably deal with that issue, if it arises, at that point in time.

lima-limon-inc: > @lima-limon-inc Thanks for the explanation!
> 
> I am a bit concerned about having symlinks with spaces in the name, in terms of portability to Windows, but I don't actually know if it is an issue in practice (I'd love to test, but I don't think we've tested hardly _anything_ on Windows at this point, so I'm not going to send you down that rabbit hole). That said, I think we can probably deal with that issue, if it arises, at that point in time.

Noted! I was also very hesitant about the presence of spaces in the file names. For reference, I found these two stack exchange posts which would appear to indicate that Windows does support filenames with spaces:
- https://superuser.com/questions/239565/can-i-use-the-start-command-with-spaces-in-the-path
- https://stackoverflow.com/questions/43510438/how-to-run-an-executable-that-contains-a-space-in-path-from-command-line-on-wind

lima-limon-inc:
> We can merge this once the conflicts have been resolved, thanks!

Conflicts resolved. I do believe that the error raised by the lint task (https://github.com/0xMiden/miden-vm/actions/runs/17551429997/job/49873567012?pr=2081#step:3:413) is not related to this PR itself.
It seems to have been fixed in this PR: https://github.com/0xMiden/miden-vm/pull/2136

PR #2083, opened 2025-08-20
Add support for emitting events from stack values

### LINKED ISSUE #2068: Add support for emitting events from stack values

adr1anh:
## Describe your changes

Fixes #2068

This PR changes the semantic of `emit` by reading the event ID from the stack rather than from the immediate value in the instruction. As a consequence, `emit.ID` now gets lowered to a push instruction followed by an emit, taking 2 cycles instead of one. 

Due to the current implementation of `SlowProcessorState`, we also change when the clock is advanced when handling an event. Previously, the clock would increase *after* executing the event, whereas we now increment if *before*. This ensures that handlers have access to the full stack without the event ID at the top. 

- Updated `CHANGELOG.md'


processor/src/operations/mod.rs
```diff
@@ -63,7 +63,17 @@ impl Process {
             Operation::Caller => self.op_caller()?,
 
             Operation::Clk => self.op_clk()?,
-            Operation::Emit => self.op_emit(host, err_ctx)?,
+            Operation::Emit => {
+                // Get the event ID from the top of the stack and drop it before executing emit.
+                let event_id = self.stack.peek().as_int() as u32;
+                self.op_drop()?;
+                // We manually advice the clock *before* calling the operation since the stack is
```
Al-Kindi-0: advance

processor/src/fast/mod.rs
```diff
@@ -891,10 +891,21 @@ impl FastProcessor {
...
             match op {
-                Operation::Emit(event_id) => self.op_emit(*event_id, host, &err_ctx).await?,
+                Operation::Emit => {
+                    let event_id = self.stack_get(0).as_int() as u32;
+
+                    // The event ID is removed from the stack before calling handlers via op_emit.
+                    // For consistency with SlowProcessor, we also increment the clock cycle before
+                    // executing the emit operation.
+                    self.decrement_stack_size();
+                    self.clk += 1_u32;
```
plafer: I understand why you went with this solution, but I don't think it's a viable one, since the user gets the incorrect clock cycle in the event handler (i.e. off by one, pointing to the next operation). Looking a bit more into it, the root of the issue is that `ProcessState` is designed to return the stack state *at the start* of the clock cycle before the operation is executed (tailored to use cases like debugger, where you want the debugger to stop and observe the state before the selected operation is executed), whereas in this case we want the state of the stack *after* the operation is executed. Hence your solution to return the state at the beginning of the following clock cycle.

However, thinking about this more, I'm starting to think it would be better to revisit the idea of making `EMIT` a degree-2 "no stack shift" operation.

As I understand it, the benefits are:
- free up a high degree operation slot
- the handlers are still "as powerful" as before, since they get access to the entire stack, not just the top 16 elements
  - I initially thought during the call that they only had access to the top 16 elements of the stack, so we'd be restraining them from using one of those slots - but this is incorrect

The downsides are:
- we break all existing handlers
- pretty much every `emit` will be followed by a `drop` (hence taking an extra cycle)

During the call, it seems like we favored not breaking existing handlers in favor of freeing a high degree operation slot - but I no longer believe this is the right call. Especially since fixing the handlers should be rather trivial, and we're still "allowed" to make such breaking changes.

Finally, circling back to the issue here, if `emit` didn't touch the stack, we'd get the added benefit that it would be simpler to implement correctly (i.e. no `clk` hack as is done here).

bobbinth: > The downsides are:
> 
> * we break all existing handlers
> * pretty much every `emit` will be followed by a `drop` (hence taking an extra cycle)

I think my main concern here is not the handlers that we control, but that any custom event would now need to account for an "extra" value being on top of the stack that should be ignored in basically all cases.

But I wonder if there is a way to have our cake and eat it too. Specifically, maybe the `ProcessState` struct can account for this and basically ignore the top element of the stack. For example, `ProcessState::get_stack_word()` would read the word starting with the second element on the stack.

I think the main difficulty here is that we use `ProcessState` not only with event handlers but also in other contexts as well. But it may be possible to deal with this relatively cleanly:

- In the long run, we should probably minimize uses of `ProcessState` outside of event handlers. For example, it is not clear to me that we need things like `Host::on_debug()` or `Host::on_assert_failed()` any more.
- In the short term, we could create another variant of `ProcessState` that is specifically tailored for event handlers. We could even maybe re-use the existing variants just add an extra field to them - something like `is_for_event_handler` - and based on this adjust the return values for the stack.

So, to summarize: assuming this "fiddling" with `ProcessState` can be done relatively quickly, I think we can change the `emit` instruction which gets converted to `push.EVENT_ID emit drop` sequence of operations.

Also, in the interest of time, we could probably keep the instruction in the same bucket as it is now, and move it to the low-degree bucket later on (assuming we change the semantics of the instruction, it should be relatively easy).

adr1anh: I agreed with @plafer's comment about modifying the clock. It was the easiest "fix" compared to what I was [doing before](https://github.com/0xMiden/miden-vm/commit/5985fe05599fa06d930ee669edbe5a94384f5495). The issue was really with the slow processor, since we could easily decrement the stack in the fast processor before calling the event handler. From what I can tell, the checkpoint PR would be compatible with this change, though I may be missing something.

I also agree with @bobbinth's comment that handlers should have access to the full stack, since they're only used to modify the advice state. From a quick glance, we're only using the `ProcessState` for handling events (trace, debug, `op_assert`, and system/user event handlers). All of these should be allowed to access the full stack. 

For now and in the interest of time, moving `EMIT` to another bucket would take more time than leaving the hack in place for the slow processor. I don't have the full picture around our current constraints and how to modify them safely. In particular, I'd revert the changes from the linked commit but clean it up to avoid allocating the full stack vector at every stack access.

adr1anh:
I've gone ahead and reverted the change that incremented the clock _before_ executing the event. It improves on the previous hack where we would get the full stack state whenever we read an item. I also went ahead and allow reading the stack beyond the minimum depth.

This should be good for a final review.

processor/src/fast/mod.rs
plafer: > I think my main concern here is not the handlers that we control, but that any custom event would now need to account for an "extra" value being on top of the stack that should be ignored in basically all cases.

I wouldn't say it's useless - we already provide that information [here](https://github.com/0xMiden/miden-vm/blob/e961d413e40f8380ec30e00451923ecb3d0128c3/processor/src/host/mod.rs#L110). In the new version, it would be on top of the stack instead of being passed as a parameter to `on_event()`. Therefore that value has a purpose IMO (i.e. reusing the same handlers for multiple event IDs), even if we don't currently make use of it. I don't see our current handlers simply ignoring it as a downside - especially given the benefits that making emit a no stack shift operation provides.

> But I wonder if there is a way to have our cake and eat it too. Specifically, maybe the ProcessState struct can account for this and basically ignore the top element of the stack.

I think this would radically increase the complexity of event handling, and for a reason I don't understand. I prefer the value to be *explicit* on top of the stack, and explicitly ignored by handlers that don't need it, than us trying to hide it somehow.

> I also agree with @bobbinth's comment that handlers should have access to the full stack, since they're only used to modify the advice state.

They currently *do* have access to the whole stack (i.e. stack top + overflow) in both cases - what am I missing?

> So, to summarize: assuming this "fiddling" with ProcessState can be done relatively quickly, I think we can change the emit instruction which gets converted to push.EVENT_ID emit drop sequence of operations.

I don't think we should fiddle with `ProcessState` as described above, but I do agree that we should make `emit` a no stack shift. I'm okay with leaving it in the high degree bucket for now in the interest of time - it'll be simple to move it to the "no stack shift" bucket later *if it already has the no stack shift semantics*. @adr1anh your latest commit still has `emit` drop the element - given all the above, I believe it should
1. keep its opcode for now
2. not shift the stack

bobbinth: > I wouldn't say it's useless - we already provide that information [here](https://github.com/0xMiden/miden-vm/blob/e961d413e40f8380ec30e00451923ecb3d0128c3/processor/src/host/mod.rs#L110). In the new version, it would be on top of the stack instead of being passed as a parameter to `on_event()`. Therefore that value has a purpose IMO (i.e. reusing the same handlers for multiple event IDs), even if we don't currently make use of it. I don't see our current handlers simply ignoring it as a downside - especially given the benefits that making emit a no stack shift operation provides.

Ah - good point! Then it is quite a bit simpler than I thought. Basically, `on_event()` handler doesn't need to take the `event_id` as a separate parameter and can read it as the first element from the stack. I think this could work pretty nicely.

bobbinth:
@adr1anh - how difficult would be to be update this PR to reduce make `emit` a non-stack-shifting operation (while still keeping it in the high-degree bucket)? Basically, we'd need to do the following:

First, transform `emit` to `emit drop` and `emit.imm` into `push(immi) emit drop`.

Then, update the `Host::on_event()` handler to be to look as follows:
```rust
fn on_event(&mut self, process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError>;
```

Then, update the existing event handlers to account for `event_id` being on top of the stack.

Thinking more about this, I think this is probably the long-term solution that should avoid the hacks - so, maybe good to do it sooner rather than later (and moving `emit` to a low-degree bucket can be done at some later point in time).

adr1anh:
> @adr1anh - how difficult would be to be update this PR to reduce make `emit` a non-stack-shifting operation (while still keeping it in the high-degree bucket)? Basically, we'd need to do the following:
> 
> First, transform `emit` to `emit drop` and `emit.imm` into `push(immi) emit drop`.

This is relatively easy, though I would say that the user should manually `push.event_id` and `drop` when calling just `emit` since doing one without the other could be confusing. 

It's not yet clear to me though how to move it to the other bucket so I'll leave that to another issue.

> Then, update the `Host::on_event()` handler to be to look as follows:
> 
> ```rust
> fn on_event(&mut self, process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError>;
> ```
> 
> Then, update the existing event handlers to account for `event_id` being on top of the stack.
> 
> Thinking more about this, I think this is probably the long-term solution that should avoid the hacks - so, maybe good to do it sooner rather than later (and moving `emit` to a low-degree bucket can be done at some later point in time).

After a discussion with @plafer and experimenting, I agree this is the "right" and best solution. I tried adding something like `ProcessState::with_stack_offset(offset) -> ProcessState` but ran into lifetime issues. The goal was that we could keep our existing handler implementations. It'll be simpler to just update indices instead.

Perhaps once we only have a single processor and can simplify the `ProcessState` this will become trivial, though we can explore that option later as well.

core/src/operations/mod.rs
```diff
@@ -169,7 +169,7 @@ pub enum Operation {
     ///
     /// Similar to Noop, this operation does not change the state of user stack. The immediate
     /// value affects the program MAST root computation.
-    Emit(u32) = OPCODE_EMIT,
+    Emit = OPCODE_EMIT,
```
bobbinth, line 173: Doc comments for this variant need to be updated.

core/src/sys_events.rs
```diff
@@ -318,7 +320,8 @@ impl SystemEvent {
 
     /// Returns a system event corresponding to the specified event ID, or `None` if the event
     /// ID is not recognized.
-    pub fn from_event_id(event_id: u32) -> Option<Self> {
+    pub fn from_event_id(event_id: Felt) -> Option<Self> {
+        let event_id: u32 = event_id.as_int().try_into().ok()?;
```
bobbinth, line 324: Do we need to downcast to `u32` here? We could probably change the constants below to a `u64` type - but also, if this is non-trivial (or leads to lots of changes), we can do it in a follow-up.

miden-vm/tests/integration/operations/decorators/mod.rs
```diff
@@ -57,11 +57,8 @@ impl SyncHost for TestHost {
         None
     }
 
-    fn on_event(
-        &mut self,
-        _process: &ProcessState,
-        event_id: u32,
-    ) -> Result<Vec<AdviceMutation>, EventError> {
+    fn on_event(&mut self, process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError> {
+        let event_id = process.get_stack_item(0).as_int() as u32;
```
bobbinth: If changing event IDs to a `Felt` is difficult, I would probably do `try_into().unwrap()` here so that we don't convert to a `u32` silently.

miden-vm/tests/integration/operations/decorators/mod.rs
```diff
@@ -78,9 +75,9 @@ impl AsyncHost for TestHost {
     #[allow(clippy::manual_async_fn)]
     fn on_event(
         &mut self,
-        _process: &ProcessState<'_>,
-        event_id: u32,
+        process: &ProcessState<'_>,
     ) -> impl FutureMaybeSend<Result<Vec<AdviceMutation>, EventError>> {
+        let event_id = process.get_stack_item(0).as_int() as u32;
```
bobbinth: Same comment as above.

docs/src/design/stack/op_constraints.md
```diff
@@ -300,7 +300,7 @@ $$
 The immediate value flag $f_{imm}$ is set to 1 when an operation has an immediate value, and 0 otherwise:
 
 $$
-f_{imm} = f_{push} + f_{emit} \text{ | degree} = 4
+f_{imm} = f_{push} \text{ | degree} = 3
```
bobbinth: Shouldn't the degree go up to $5$? $f_{push}$ has degree $5$, I believe. Previously, it was paired with $f_{emit}$, which has encoding different in the least significant bit - and I think these canceled out to reduce the degree to $3$.

Regardless of the impact though, we should make sure update degrees of other constraints involving $f_{imm}$ flag.

docs/src/user_docs/assembly/events.md
```diff
@@ -2,11 +2,13 @@
...
 
-An event can be emitted via the `emit.<event_id>` assembly instruction where `<event_id>` can be any 32-bit value specified either directly or via a [named constant](./code_organization.md#constants). For example:
+An event can be emitted via the `emit.<event_id>` assembly instruction where `<event_id>` can be any 32-bit value specified either directly or via a [named constant](./code_organization.md#constants). Alternatively, the `emit` instruction can be used without an immediate value - in this case it will use the value from the top of the stack as the event ID. For example:
```
bobbinth: Similar to other comments - I don't think we should allow full `Felt` event IDs - but could be done in another PR.

docs/src/design/stack/system_ops.md
```diff
@@ -11,15 +11,9 @@ s'_i - s_i = 0 \ \text{ for } i \in [0, 16) \text { | degree} = 1
...
+The `EMIT` operation pops an element off the stack and emits it as an event. The event ID is taken from the top of the stack and dropped before executing the event.
 
-The `EMIT` operation does not impose any constraints besides the ones needed to ensure that the entire state of the stack is copied over. This constraint looks like so:
-
->$$
-s'_i - s_i = 0 \ \text{ for } i \in [0, 16) \text { | degree} = 1
-$$
-
-Additionally, the prover puts `EMIT`'s immediate value in the first user op helper register non-deterministically. The [Op Group Table](../decoder/main.md#op-group-table) is responsible for ensuring that the prover sets the appropriate value.
+This operations therefore has the same semantics as the [`DROP`](./stack_ops.md#drop) operation for the stack.
```
bobbinth: I think this is no longer accurate and probably should be updated.

docs/src/design/decoder/constraints.md
```diff
@@ -10,21 +10,21 @@ We assume that the VM exposes a flag per operation which is set to $1$ when the
...
+| `RESPAN`  | $f_{respan}$  |   4    | Stack remains unchanged.                                                                         |
+| `DYN`     |   $f_{dyn}$   |   5    | Stack remains unchanged.                                                                         |
+| `CALL`    |  $f_{call}$   |   4    | Stack remains unchanged.                                                                         |
+| `SYSCALL` | $f_{syscall}$ |   4    | Stack remains unchanged.                                                                         |
+| `END`     |   $f_{end}$   |   4    | When exiting a loop block, top stack element is dropped; otherwise, the stack remains unchanged. |
+| `HALT`    |  $f_{halt}$   |   4    | Stack remains unchanged.                                                                         |
+| `PUSH`    |  $f_{push}$   |   5    | An immediate value is pushed onto the stack.                                                     |
+| `EMIT`    |  $f_{emit}$   |   5    | Top stack element is dropped.                                                                    |
```
bobbinth: I think the effect on the stack here is still "Stack remains unchanged."

docs/src/design/decoder/constraints.md
```diff
@@ -628,10 +628,10 @@ Where $i \in [1, 8)$. Thus, $v_1$ defines row value for group in $h_1$, $v_2$ de
 We compute the value of the row to be removed from the op group table as follows:
 
 $$
-u = \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot gc + \alpha_3 \cdot ((h_0' \cdot 2^7 + op') \cdot (1 - f_{imm}) + s_0' \cdot f_{push} + h_2 \cdot f_{emit}) \text{ | degree} = 6
+u = \alpha_0 + \alpha_1 \cdot a + \alpha_2 \cdot gc + \alpha_3 \cdot ((h_0' \cdot 2^7 + op') \cdot (1 - f_{imm}) + s_0' \cdot f_{push}) \text{ | degree} = 6
```
bobbinth: Related to one of the previous comments: could we check if the degree here stays the same or goes up by $1$?

processor/src/stack/mod.rs
```diff
@@ -161,13 +161,38 @@ impl Stack {
     /// stack will be at the last position in the word.
     ///
     /// Creating a word does not change the state of the stack.
-    pub fn get_word(&self, word_idx: usize) -> Word {
-        let offset = word_idx * WORD_SIZE;
+    pub fn get_word(&self, idx: usize) -> Word {
+        [self.get(idx + 3), self.get(idx + 2), self.get(idx + 1), self.get(idx)].into()
+    }
```
bobbinth: Doc comments for this function need to be updated.

Also, changing semantics in this way may lead to difficult-to-detect bugs downstream. At the very least, we should clearly document this as a breaking change in the changelog. But I also wonder if a different name should be used instead.

processor/src/stack/mod.rs
```diff
@@ -161,13 +161,38 @@ impl Stack {
...
+
+    /// Returns the value at the specified position on the stack, including overflow items.
+    ///
+    /// This method can access items beyond the top 16 positions by reading from the overflow table.
+    /// Position 0 is the top of the stack, positions 0-15 are in the trace, and positions 16+ are
+    /// in the overflow table.
+    pub fn get_with_overflow(&self, pos: usize) -> Felt {
```
bobbinth: I wonder if we need a special function for this, or if we should just update the logic of `Stack::get()` method to handle overflow items as well.

processor/src/stack/mod.rs
```diff
@@ -161,13 +161,38 @@ impl Stack {
...
+            self.overflow.get_element_at(overflow_index).unwrap_or(ZERO)
+        }
+    }
+
+    /// Returns a word from the stack at the specified word index, including overflow items.
+    ///
+    /// This method can access words that span into the overflow table.
+    /// Word 0 is defined by stack positions 0-3, word 1 by positions 4-7, etc.
+    /// The words are created in reverse order (top element at the last position).
+    pub fn get_word_with_overflow(&self, start_idx: usize) -> Word {
```
bobbinth: Similar comment as above - maybe there could be just a single function for getting words from the stack and it would handle both cases.

processor/src/host/default.rs
```diff
@@ -154,11 +154,8 @@ where
         self.store.get(node_digest)
     }
 
-    fn on_event(
-        &mut self,
-        process: &ProcessState,
-        event_id: u32,
-    ) -> Result<Vec<AdviceMutation>, EventError> {
+    fn on_event(&mut self, process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError> {
+        let event_id = process.get_stack_item(0).as_int() as u32;
```
bobbinth: Similar to one of the other comments: I would rather use `try_into().unwrap()` than `as u32`.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -12,7 +12,8 @@ use miden_core::{
 use crate::{ExecutionError, MemoryError, ProcessState, errors::ErrorContext};
 
 /// The offset of the domain value on the stack in the `hdword_to_map_with_domain` system event.
-pub const HDWORD_TO_MAP_WITH_DOMAIN_DOMAIN_OFFSET: usize = 8;
+/// TODO: Mention we add 1 to account for event ID at the top
+pub const HDWORD_TO_MAP_WITH_DOMAIN_DOMAIN_OFFSET: usize = 9;
```
bobbinth: We should probably address this `TODO`.

bobbinth (commented):
Looks great! Thank you! I left some comments inline - but nothing major.

core/src/sys_events.rs
adr1anh: I’ve left it for now, this could be overhauled when we move system events to be just normal handlers.

miden-vm/tests/integration/operations/decorators/mod.rs
adr1anh: Good catch, let me check the other places where we might be doing this conversion

docs/src/design/stack/system_ops.md
adr1anh: I’m reworking all comments and will be pushing soon. But thanks for highlighting

docs/src/user_docs/assembly/events.md
adr1anh: With these changes, the event ID is really an assembly-level construct rather than a VM one. `emit` really just interrupts the execution to handover control for one cycle to the host. 

For system events, we use the convention that the top of the stack contains the ID, but user-defined handlers are a lot more flexible now. 

Obviously, following the "convention" makes things easier. For this, we'd use `blake3("namespace/event")%p` (or some variation), but without the underlying string we don't know if the `Felt` is valid. 

TLDR: I think we should allow arbitrary event IDs at the VM level, but enforce correct naming at the assembly level.

bobbinth: Ah! It was a typo on my end. I think I wanted to say is that we **should** allow full felt event IDs. Sorry!

adr1anh:
Apologies for the force push, due to the changes yesterday this was the easiest on my end. 

Thanks for the feedback @bobbinth , I've gone through all the documentation portion of the PR and addressed your comments.

The main breaking change was making `Stack::get` return elements from the overflow stack. This matches the behavior for the FastProcessor as well. Maybe @plafer has comments though.

core/src/operations/mod.rs
bobbinth: I think lines 164 - 168 still need to be updated.

docs/src/user_docs/assembly/events.md
```diff
@@ -1,13 +1,24 @@
...
+Events interrupt VM execution for one cycle and hand control to the host. The host can read VM state and modify the advice provider. From the VM's perspective, `emit` has identical semantics to `noop` - the operand stack and registers remain unchanged.
 
-An event can be emitted via the `emit.<event_id>` assembly instruction where `<event_id>` can be any 32-bit value specified either directly or via a [named constant](./code_organization.md#constants). For example:
+By convention, programs place a 32-bit `event_id` at the top of the stack before executing `emit`. The VM doesn't enforce this - the host reads and interprets the event ID.
```
bobbinth, line 5: I think we should remove the "32-bit" part from here (or replace it with "Felt").

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -307,10 +307,11 @@ High-level constructs for controlling the execution flow.
...
+|--------------------|-------------------|--------------|--------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `emit.<event_id>`  | `[...]`           | `[...]`      | 3      | Emits an event with the specified `event_id` to the host. The net effect on the operand stack is no change (internally expands to `push.<event_id> emit drop`). The `event_id` can be any 32-bit value specified either directly or via a [named constant](./code_organization.md#constants). Events allow programs to communicate contextual information to the host for triggering appropriate actions. Example: `emit.123` or `emit.EVENT_ID_1` |
```
bobbinth, line 312: Similar to the other comment - I'd remove the "32-bit" part (I don't think we are enforcing this, are we?).

bobbinth (approved):
Looks good! Thank you! I left a few small (mostly doc-related) comments inline.

adr1anh:
Thanks @bobbinth. I made these modifications in #2085, so if you're okay with it, I'll merge this one as-is.

bitwalker (approved):

PR #2084, opened 2025-08-20
docs: document hasher chiplet modules

Galoretka:
docs: document hasher chiplet modules

Add module documentation for hasher-related components:
- Hash chiplet interface (core/src/chiplets/hasher.rs)
- ProcessorAir constraints (air/src/lib.rs) 
- Hasher trace definitions (air/src/trace/chiplets/hasher.rs)

Documentation covers 16-column trace structure and core functionality
while maintaining project's concise documentation style.


adr1anh:
Thanks for your interest, but we don't accept these kind of documentation PRs.

Issue #1582, opened 2024-11-22
Move some system events to the standard library


plafer:
As a follow-up to #1457, we should move some of the system events (formerly "advice injectors") introduced in #1581 to the standard library.

Context: https://github.com/0xPolygonMiden/miden-vm/issues/1457#issuecomment-2491157516 and https://github.com/0xPolygonMiden/miden-vm/issues/1457#issuecomment-2491951354

adr1anh:
Tackling this now, and wondering which events we should move. 

The following seem like they are clearly "system" events since they are tightly linked to the `AdviceProvider`
- `MAP_VALUE_TO_STACK`
- `MAP_VALUE_TO_STACK_N`
- `HAS_MAP_KEY`
- `HDWORD_TO_MAP`
- `HDWORD_TO_MAP_WITH_DOMAIN`
- `HPERM_TO_MAP`
- `MEM_TO_MAP`

These involve Merkle trees, and likely need supporting masm to work. These probably should go in the `stdlib`.
- `MERKLE_NODE_MERGE`
- `MERKLE_NODE_TO_STACK`
- `SMT_PEEK`


Not sure about all of these "Integer" related events. The first two should probably be taken out.
- `U64_DIV`
- `FALCON_DIV`

I think the others are simple and useful enough that we could leave them in.
- `EXT2_INV`
- `U32_CLZ`
- `U32_CTZ`
- `U32_CLO`
- `U32_CTO`
- `ILOG2`

Al-Kindi-0:
> Tackling this now, and wondering which events we should move.
> 
> The following seem like they are clearly "system" events since they are tightly linked to the `AdviceProvider`
> 
>     * `MAP_VALUE_TO_STACK`
> 
>     * `MAP_VALUE_TO_STACK_N`
> 
>     * `HAS_MAP_KEY`
> 
>     * `HDWORD_TO_MAP`
> 
>     * `HDWORD_TO_MAP_WITH_DOMAIN`
> 
>     * `HPERM_TO_MAP`
> 
>     * `MEM_TO_MAP`
> 

Agreed

> 
> These involve Merkle trees, and likely need supporting masm to work. These probably should go in the `stdlib`.
> 
>     * `MERKLE_NODE_MERGE`
> 
>     * `MERKLE_NODE_TO_STACK`
> 
>     * `SMT_PEEK`
> 

[...]

PR #2085, opened 2025-08-20
feat(masm): Add support for `event()` syntax

adr1anh:
## Describe your changes

This PR implements human-readable event names in Miden Assembly, replacing numeric event IDs with descriptive string-based identifiers.

### New Syntax

```
# Event constants
const.MY_EVENT=event("miden::transfer::initiated")
emit.MY_EVENT

# Inline events
emit.event("miden::transfer::initiated")

# No longer allowed
# const.MY_EVENT_U32=42 emit.MY_EVENT_U32
# emit.42

# Backwards compatibility with u32 events 
const.MY_EVENT_U32=42 push.MY_EVENT_U32 emit drop
push.42 emit drop
```

#### Parser & AST:
- Added `Event` token and `HashKind::Event` for `event("...")` expressions
- Changed `EmitImm` from `ImmU32` to `ImmFelt` to support field element event IDs
- Extended grammar to parse both constant and inline event forms

#### Event ID Generation:
- Deterministic mapping: first 64 bits of `blake3(event_name)` as little-endian u64, mod p
- Implement `string_to_event_id` in `core::utils` for stable mapping
- Ensures identical names always produce the same event ID

#### Validation:
- `emit.CONST` now requires `CONST` to be defined via `event("...")`
- Raw numeric immediates (`emit.42`) are no longer supported
- Added comprehensive error checking for invalid event constants

#### Documentation & Examples

- Updated event documentation with naming guidelines
- Recommends hierarchical naming: `project_name::library_name::event_name`

#### Breaking Changes

- `emit.<number>` syntax is no longer supported
- Event constants must be defined using `event("...")` syntax
- `Operation::Emit` no longer contains an immediate

### Testing

- Added validation tests for syntax restrictions
- Equivalence tests confirming all forms produce identical MAST
- Updated integration tests to use new syntax

The implementation maintains backward compatibility for stack-based event emission while providing a developer-friendly interface for compile-time event definitions.

- Updated `CHANGELOG.md'

### DIFF
--- a/docs/src/user_docs/assembly/events.md
+++ b/docs/src/user_docs/assembly/events.md
@@ -2,7 +2,9 @@
 
 Events interrupt VM execution for one cycle and hand control to the host. The host can read VM state and modify the advice provider. From the VM's perspective, `emit` has identical semantics to `noop` - the operand stack and registers remain unchanged.
 
-Event identifiers are field elements. Use a stable mapping: `event_id = blake3("<name>") mod p`, where `p` is the Goldilocks prime. The VM doesn’t enforce structure for stack-provided IDs, but immediate forms restrict inputs to this mapping.
+Event identifiers are field elements. Use a stable mapping: first 64 bits of `blake3("<name>")` as little-endian u64, mod p. The VM doesn't enforce structure for stack-provided IDs, but immediate forms restrict inputs to this mapping.
+
+Event names should be as unique as possible to avoid collisions with other libraries. Use a hierarchical naming convention like `project_name::library_name::event_name`. Generic names may cause conflicts in multi-library environments.
 
 ### Event Instructions
 
@@ -11,11 +13,11 @@ Event identifiers are field elements. Use a stable mapping: `event_id = blake3("
 
 ```miden
 # Using a constant
-const.MY_EVENT=event("transfer::initiated")
+const.MY_EVENT=event("miden::transfer::initiated")
 emit.MY_EVENT
 
 # Inline form
-emit.event("transfer::initiated")
+emit.event("miden::transfer::initiated")
 
 # Equivalent manual stack form (any Felt – not validated):
 push.<felt> emit drop
--- a/assembly-syntax/src/sema/passes/const_eval.rs
+++ b/assembly-syntax/src/sema/passes/const_eval.rs
@@ -63,7 +63,7 @@ impl VisitMut for ConstEvalVisitor<'_> {
                     // NOTE: This function only validates the kind; the actual resolution to a Felt
                     // happens below in `visit_mut_immediate_felt` just like other Felt immediates.
                     // Enabled syntax:
-                    //   const.EVT = event("my::evt")
+                    //   const.EVT = event("...")
                     //   emit.EVT
                 },
                 Ok(_) => {
@@ -121,7 +121,7 @@ impl VisitMut for ConstEvalVisitor<'_> {
                     Ok(ConstantExpr::Hash(HashKind::Event, string)) => {
                         // CHANGE: resolve `event("...")` to a Felt when a Felt immediate is
                         // expected (e.g. enables `emit.EVENT`):
-                        //   const.EVT = event("my::evt")
+                        //   const.EVT = event("...")
                         //   emit.EVT
                         let event_id = string_to_event_id(string.as_str());
                         *imm = Immediate::Value(Span::new(span, event_id));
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 - [BREAKING] `Operation::Emit` no longer contains a `u32` parameter, affecting pattern matching and serialization (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - [BREAKING] Host `on_event` methods no longer receive `event_id` parameter; event ID must be read from stack position 0 (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - [BREAKING] `get_stack_word` uses element-aligned indexing instead of word-aligned indexing (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
+- [BREAKING] Implement support for `event("event_name")` in MASM (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
```
bobbinth, line 22: Why is this a breaking change?

docs/src/user_docs/assembly/events.md
```diff
@@ -2,16 +2,23 @@
 
 Events interrupt VM execution for one cycle and hand control to the host. The host can read VM state and modify the advice provider. From the VM's perspective, `emit` has identical semantics to `noop` - the operand stack and registers remain unchanged.
 
-By convention, programs place a 32-bit `event_id` at the top of the stack before executing `emit`. The VM doesn't enforce this - the host reads and interprets the event ID.
+Event identifiers are field elements. Use a stable mapping: `event_id = blake3("<name>") mod p`, where `p` is the Goldilocks prime. The VM doesn’t enforce structure for stack-provided IDs, but immediate forms restrict inputs to this mapping.
```
bobbinth: nit: I'm not sure if `blake3("<name>") mod p` is correct. We basically take the first element of the hash and do `mod p` on it - but is this the same as doing `mod p` on the whole hash?

core/src/operations/mod.rs
```diff
@@ -161,15 +161,19 @@ pub enum Operation {
...
+    ///
+    /// Semantics:
+    /// - Reads the event id from the top of the stack (as a `Felt`) without consuming it; the
+    ///   caller is responsible for pushing and later dropping the id.
+    /// - User-defined events are conventionally derived from strings via
+    ///   `hash_string_to_word(name)[0]` (Blake3-based) and may be emitted via immediate forms in
+    ///   assembly (`emit.event("...")` or `emit.CONST` where `CONST=event("...")`).
+    /// - System events are still identified by specific 32-bit codes; the VM attempts to interpret
+    ///   the stack `Felt` as `u32` to dispatch known system events, and otherwise forwards the
+    ///   event to the host.
```
bobbinth, line 174: Is using `u32` IDs for system events temporary? (i.e., do we plan to remove this restriction in future PRs?)

docs/src/user_docs/assembly/events.md
```diff
@@ -2,16 +2,23 @@
...
 
 ```miden
-emit.42                    # Immediate form
-push.42 emit drop          # Equivalent manual sequence
+# Using a constant
+const.MY_EVENT=event("transfer::initiated")
+emit.MY_EVENT
+
+# Inline form
+emit.event("transfer::initiated")
```
bobbinth: I would probably add a couple of sentences on how users should specify event names. For example, maybe it should be something like `project_name::library_name::event_name` - basically, encourage people to make names as unique as possible (and I would mention that if the names are too generic, this may cause collisions with other libraries).

And then, here in examples, I'd maybe use something like `miden::transfer::initiated`.

bobbinth (approved):
Looks great! Thank you! I left a few small comments inline - once these are addressed, I think we are good to merge.

CHANGELOG.md
adr1anh: We don't allow `emit.42` anymore, and instead require the `event` keyword. It's still possible using `push.42 emit drop` though.

bobbinth: Ah yes - makes sense!

core/src/operations/mod.rs
adr1anh: The system event IDs should probably be defined as `event("adv.has_map_key")` in which case it would be a full Felt. We can make this part of the effort to move events to the stdlib.

One small annoyance I wasn't able to fully figure out was how to create `const` IDs, since blake3 isn't const currently.

bobbinth: Yep, we can do it in a follow-up.

> The system event IDs should probably be defined as `event("adv.has_map_key")` in which case it would be a full Felt.

I'm wondering if we could just assign constant values to them - e.g., `0`, `1`, `2` etc. I think with this approach a collision is equality unlikely as if we derived the values using BLAKE3.

adr1anh: Linked in #1582

PR #2086, opened 2025-08-20
test: replace pretty_assertions with insta when matching string literals

huitseeker:
**EDIT**: post review, chose to use [insta](https://insta.rs/) and per-file snapshots instead of [expect_test](https://docs.rs/expect-test/latest/expect_test/).

The rationale is reducing the friction of reviewing changes in file based snapshot tests, which `cargo insta` does really well ( `cargo insta review` isn't supported in expect, only `cargo insta approve` is)

----

This commit replaces the use of pretty_assertions with ~~expect_test~~ insta (where possible) in assembly tests for easier snapshot test edits.

- Replaced `pretty_assertions` dependency with ~~`expect-test`~~ `insta` when matching against string literals,
- Updated all string-based snapshot assertions from `assert_str_eq!` to ~~`assert_eq!`~~ `assert_snapshot!`,
- Maintained backward compatibility - all existing tests pass unchanged,
- ~~Minimal API changes for developers - same test behavior, better implementation~~ expected output needs a file export, better implementation

## 1. auto updates (better)
**Instead of manually editing expected strings in tests**, developers can now:

*Edit:*
```bash
cargo insta review
```
(after `cargo install cargo-insta`)

This automatically updates expected values in the source code, eliminating tedious manual alignment and formatting issues.

## 2. nice diffs (the same)
~~expect_test~~ insta provides readable diffs with clear visual separation between expected and actual

## 3. more advanced features (future)
While this commit maintains the simple ~~`assert_eq`~~ `assert_snapshot` interface for backward compatibility, developers can sometimes leverage advanced features, see [documentation](https://insta.rs/docs/)

### DIFF
--- a/assembly/Cargo.toml
+++ b/assembly/Cargo.toml
@@ -25,7 +25,6 @@ logging = ["dep:env_logger"]
 
 [dependencies]
 env_logger = { version = "0.11", optional = true }
-insta = "1.43.1"
 log.workspace = true
 miden-assembly-syntax.workspace = true
 miden-core.workspace = true
@@ -36,6 +35,5 @@ thiserror.workspace = true
 [dev-dependencies]
 miden-mast-package = { workspace = true, features = ["arbitrary"] }
 miden-stdlib.workspace = true
-expect-test = "1.5"
-pretty_assertions = "1.4"
+insta = "1.43.1"
 proptest = { workspace = true, features = ["std"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -606,16 +606,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "expect-test"
-version = "1.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "63af43ff4431e848fb47472a920f14fa71c24de13255a5692e93d4e90302acb0"
-dependencies = [
- "dissimilar",
- "once_cell",
-]
-
-[[package]]
 name = "fixedbitset"
 version = "0.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1102,14 +1092,12 @@ name = "miden-assembly"
 version = "0.18.0"
 dependencies = [
  "env_logger",
- "expect-test",
  "insta",
  "log",
  "miden-assembly-syntax",
  "miden-core",
  "miden-mast-package",
  "miden-stdlib",
- "pretty_assertions",
  "proptest",
  "smallvec",
  "thiserror 2.0.12",
--- a/assembly/src/snapshots/miden_assembly__tests__assert_eq_with_code.snap
+++ b/assembly/src/snapshots/miden_assembly__tests__assert_eq_with_code.snap
@@ -1,0 +1,14 @@
+---
+source: assembly/src/tests.rs
+expression: program
+---
+begin
+    basic_block
+        eq
+        assert(0)
+        eq
+        assert(15491226248792286710)
+        eq
+        assert(15491226248792286710)
+    end
+end
--- a/assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
+++ b/assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
@@ -1,0 +1,41 @@
+---
+source: assembly/src/tests.rs
+expression: program
+---
+begin
+    basic_block
+        movup4
+        eq
+        assert(0)
+        movup3
+        eq
+        assert(0)
+        movup2
+        eq
+        assert(0)
+        eq
+        assert(0)
+        movup4
+        eq
+        assert(15491226248792286710)
+        movup3
+        eq
+        assert(15491226248792286710)
+        movup2
+        eq
+        assert(15491226248792286710)
+        eq
+        assert(15491226248792286710)
+        movup4
+        eq
+        assert(15491226248792286710)
+        movup3
+        eq
+        assert(15491226248792286710)
+        movup2
+        eq
+        assert(15491226248792286710)
+        eq
+        assert(15491226248792286710)
+    end
+end
--- a/assembly/src/snapshots/miden_assembly__tests__assert_with_code.snap
+++ b/assembly/src/snapshots/miden_assembly__tests__assert_with_code.snap
@@ -1,0 +1,11 @@
+---
+source: assembly/src/tests.rs
+expression: program
+---
+begin
+    basic_block
+        assert(0)
+        assert(15491226248792286710)
+        assert(15491226248792286710)
+    end
+end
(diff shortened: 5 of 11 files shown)


bitwalker (approved):
Nice!

I would suggest we extract all of the expected strings into separate files though - I've found that updating the expected output fails in odd ways sometimes, corrupting the file in the process (i.e. leaving delimiters unclosed, etc.). I suspect the issues are due to concurrency, but haven't bothered to try and confirm/fix it upstream myself. Aside from avoiding that issue though, using separate files also allows us to read the source files using appropriate syntax highlighting, and it side steps git conflicts when two PRs both update different sets of tests.

Take a look at the compiler's test suite if you want to see how we make use of this style of tests (we actually have our own implementation of `expect-test` now with other features, but the API is basically identical).

/cc @greenhat in case he has anything to add

huitseeker:
@bitwalker I'm quite familiar with [insta](https://github.com/mitsuhiko/insta), which uses the mode of snapshot testing you suggest (and actually, `expect-test` has `expect_file!`). Changing the assembly tests to use split files for every test is a much larger refactoring, though. Should we open an issue?

Re: automated updates not working, while I have not experienced the kind of issue you suggest, I've found that an intermediate commit + `git (revert|commit -p|amend -p)` always gets me out of trouble.

bitwalker:
Since it is purely a mechanical change, I'd do it in this PR. I will say that this is something LLMs do really really well - I did the same extraction in the compiler test suite when I discovered the issue I mentioned with the update process corrupting the test suite itself, simply by asking Claude to do it for me, and telling it the pattern to follow. It took about 10 minutes to write up the prompt, let it run, and review the output. I'd do it that way just to avoid the excruciating pain in the ass that is doing it by hand 😅.

huitseeker:
It's beyond the mere inconvenience:
- the snapshots being tested here are often in two parts, first a source, then an expected value, then a program build from the source is compared against the expected value. Comparing the source to the expected value if the later is in a different file and the former isn't is going to involve some friction (the reviewer has to reconstruct the test-to-file mapping and open a few files to compare properly).
- in several tests, the expected value is not a string literal, but is constructed in an interpolated string from various values in the test. These cases can't be tested with expect, so we keep the default pretty-assertions, and will therefore keep the expected value right next to the source it's present in. Moving string literals out of this file will make _these_ tests look very different from _those_ tests.

I'm ok doing the export to files, but my bet is we're likely to have a PR reverting everything to pretty-assertions ("for uniformity") in a couple of weeks. 

----
*Edit*: while trying to use your nice unified-diff in midenc-expect-test, I've encountered an issue, lmkwyt: https://github.com/0xMiden/compiler/issues/642

*Edit*: have tried to give up on fancy interpolated strings for expected values, and reach a wider extent of file-based snapshots.

bitwalker:
> * the snapshots being tested here are often in two parts, first a source, then an expected value, then a program build from the source is compared against the expected value. Comparing the source to the expected value if the later is in a different file and the former isn't is going to involve some friction (the reviewer has to reconstruct the test-to-file mapping and open a few files to compare properly).

IMO, you move both the source and the expected output to files on disk if appropriate (almost always the case), otherwise all that needs to go on disk is the expected output, as that is the only part that needs to be automatically modified by `expect-test`.

In the case where just the expected output is on disk, and the input isn't, sure, there is some friction there during review; which is why I generally feel that they both go on disk with the same name, differing only in the extension, thus keeping them side-by-side) - but this is a minor issue IMO. The mapping is straightforward, as the file names are derived from the name of the test in the typical case, and if you are in a position to review the code, you would presumably also be familiar with the test suite and this convention.

Regardless, it is going to almost always be the case that we are modifying multiple tests at once with `UPDATE_EXPECT=1`, and that is going to cause the issue I mentioned with the update process trying to modify the file concurrently, corrupting it in the process. This happened frequently with a subset of the compiler tests that used `expect!` vs `expect_file!`, and I found that far more frustrating than having sources and expected outputs being in separate files.

I suppose an alternative here is to determine if that issue has been fixed upstream, and if not, fixing it ourselves and upstreaming the fix, so that we can continue to use `expect!` - but aside from that, I just don't think waving the bug away is an option here. I'll defer to @bobbinth @plafer and @Al-Kindi-0 though.

> * in several tests, the expected value is not a string literal, but is constructed in an interpolated string from various values in the test. These cases can't be tested with expect, so we keep the default pretty-assertions, and will therefore keep the expected value right next to the source it's present in.

I don't think we need to change the tests using `pretty_assertions` at all. I'm talking specifically about those using `expect-test`. I also think it is perfectly fine that tests using `pretty_assertions` look/work differently than those using `expect-test` - they are different types of tests after all.

> but my bet is we're likely to have a PR reverting everything to pretty-assertions ("for uniformity") in a couple of weeks.

Uniformity isn't an important property of the test suite - consistency, sure, but consistency in terms of how we're testing a given part of the system. It doesn't make sense for us to have severals ways of testing how MASM diagnostics are rendered, for example. However, we might have a few different ways of testing expected outputs in general, depending on the needs of different types of tests.

To be honest, I don't see why separating the expected outputs into files is any different in this scenario than using `expect!` while still using `pretty_assertions` in some places. The latter serves a purpose that is not easily served by `expect-test`. If we're using them interchangeably, that's different, but doesn't mean we need to use one or the other, just that we need to be consistent in our usage.

> _Edit_: while trying to use your nice unified-diff in midenc-expect-test, I've encountered an issue, lmkwyt: [0xMiden/compiler#642](https://github.com/0xMiden/compiler/issues/642)

Thanks for opening the issue! Definitely an easy one for us to ship a fix for ASAP.

huitseeker:
> Regardless, it is going to almost always be the case that we are modifying multiple tests at once with UPDATE_EXPECT=1, and that is going to cause the issue I mentioned with the update process trying to modify the file concurrently, corrupting it in the process. This happened frequently with a subset of the compiler tests that used expect! vs expect_file!, and I found that far more frustrating than having sources and expected outputs being in separate files.

FWIW, `UPDATE_EXPECT=1 cargo test -- --test-threads=1` is an option (though in the current state of the PR this is a bit moot)

assembly/Cargo.toml
```diff
@@ -35,5 +35,5 @@ thiserror.workspace = true
 [dev-dependencies]
 miden-mast-package = { workspace = true, features = ["arbitrary"] }
 miden-stdlib.workspace = true
-pretty_assertions = "1.4"
+insta = "1.43.1"
```
plafer, line 38: The [quickstart docs](https://insta.rs/docs/quickstart/) recommend compiling insta in release mode for faster test times - probably a good idea to do so? From the docs:

```toml
[profile.dev.package]
insta.opt-level = 3
similar.opt-level = 3
```

Al-Kindi-0 (approved):
LGTM!

PR #2090, opened 2025-08-22
fix: script crashes and improve variable defaults

mdqst:
## Describe your changes

updated the script to prevent crashes when `set -u` is enabled.

* `${NO_CHANGELOG_LABEL}` → `${NO_CHANGELOG_LABEL:-}` to avoid errors if the variable is unset.
* `${BASE_REF}` → `${BASE_REF:-main}` so the script defaults to the main branch.
* Replaced multiline `echo` with `cat <<EOF ... EOF` for proper line breaks.

now the script runs safely even with strict variable checking.

### DIFF
--- a/scripts/check-changelog.sh
+++ b/scripts/check-changelog.sh
@@ -3,17 +3,19 @@ set -u
 
 CHANGELOG_FILE="${1:-CHANGELOG.md}"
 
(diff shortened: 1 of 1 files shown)


plafer:
Has this script been crashing in some of our PRs?

bobbinth:
Closing as I don't think these changes are needed.

PR #2092, opened 2025-08-22
Sparse iteration for OpBatches

huitseeker:
This introduces: 
1. a few more proptests for the construction of batches,
2. a simplified [CSR (aka Yale)](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)) representation for batches, where the sparse element is `Operation::Noop`. 
3. this representation is used to implement two traits:
- the [Index trait](https://doc.rust-lang.org/std/ops/trait.Index.html), that allows indexing in the fully-padded OpBatch representation, as if it was a vector,
- the [Iterator trait](https://doc.rust-lang.org/std/iter/trait.Iterator.html) that allows iterating through the fully padded representation. This is accessed through `(opb: OpBatch).ops.iter_dense()`
4. we then use this in the `OperationOrDecoratorIterator` iterator to iterate over the padded representation. The load-bearing change is the following:
```rust
 // If no decorator needs to be executed, then execute the operation
 if let Some(batch) = self.node.op_batches.get(self.batch_index) {
-    if let Some(operation) = batch.ops.get(self.op_index_in_batch) {
-        ...
+    if self.op_index_in_batch < batch.ops.num_rows() * batch.ops.num_cols() {
+        let operation = &batch.ops[self.op_index_in_batch];
         self.op_index_in_batch += 1;
 }
```

One test to read is `test_groups_ops_coherence`, which basically guarantees the padding operations and the group Felts agree on the contents of the dense iteration (3. above)

### DIFF
--- a/assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
+++ b/assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
@@ -37,5 +37,8 @@ begin
         assert(15491226248792286710)
         eq
         assert(15491226248792286710)
+        noop
+        noop
+        noop
     end
 end
--- a/assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
+++ b/assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
@@ -4,11 +4,11 @@ expression: program
 ---
 begin
     join
-        basic_block push(2) push(3) end
+        basic_block push(2) push(3) noop noop noop noop noop noop noop end
         if.true
-            basic_block add end
+            basic_block add noop noop noop noop noop noop noop noop end
         else
-            basic_block noop end
+            basic_block noop noop noop noop noop noop noop noop noop end
         end
     end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Enhancements
 
 - Added slicing for the word constants ([#2057](https://github.com/0xMiden/miden-vm/pull/2057)).
+- Adds sparse representation for the operation batches ([#1815](https://github.com/0xMiden/miden-vm/issues/1815))
 
 #### Changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1136,6 +1136,7 @@ dependencies = [
  "num-derive",
  "num-traits",
  "proptest",
+ "proptest-derive 0.3.0",
  "thiserror 2.0.12",
  "winter-math",
  "winter-rand-utils",
@@ -1207,7 +1208,7 @@ dependencies = [
  "miden-assembly-syntax",
  "miden-core",
  "proptest",
- "proptest-derive",
+ "proptest-derive 0.6.0",
 ]
 
 [[package]]
@@ -1233,7 +1234,7 @@ dependencies = [
  "supports-color",
  "supports-hyperlinks",
  "supports-unicode",
- "syn",
+ "syn 2.0.104",
  "terminal_size",
  "textwrap",
  "thiserror 2.0.12",
@@ -1247,9 +1248,9 @@ version = "8.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "86a905f3ea65634dd4d1041a4f0fd0a3e77aa4118341d265af1a94339182222f"
 dependencies = [
- "proc-macro2",
- "quote",
- "syn",
+ "proc-macro2 1.0.95",
+ "quote 1.0.40",
+ "syn 2.0.104",
 ]
 
 [[package]]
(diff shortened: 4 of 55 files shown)


assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
```diff
@@ -37,5 +37,44 @@ begin
...
+        noop
+        noop
+        noop
+        noop
+        noop
+        noop
+        noop
+        noop
+        noop
+        noop
```
bobbinth: Without having looked into this in detail, something looks off here: there should never be this many padded noops in a basic block. As a very rough guess, I think no basic block should end up with more than 8 padded noops.

assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
```diff
@@ -4,11 +4,11 @@ expression: program
 ---
 begin
     join
-        basic_block push(2) push(3) end
+        basic_block push(2) push(3) noop noop noop noop noop noop noop end
         if.true
-            basic_block add end
+            basic_block add noop noop noop noop noop noop noop noop end
```
bobbinth, line 9: I don't think there should be any padded noops in this block.

assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
```diff
@@ -4,11 +4,11 @@ expression: program
 ---
 begin
     join
-        basic_block push(2) push(3) end
+        basic_block push(2) push(3) noop noop noop noop noop noop noop end
```
bobbinth, line 7: I think this block should have 2 noops.

assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
plafer: This (and other similar places) points to what we discussed in an offline call. The CSR representation is equivalent to the representation of a basic block *from the perspective of the hasher*; that is, a hasher will pad unused bytes with 0's (which map to NOOPs when interpreted as operations).

However from the perspective of the decoder, the exact number of inserted NOOPs is important - and differentiates between "unused bytes" and a NOOP. For example, given the "dense" operation group `[SWAP, PUSH]`, the "sparse" representation from the point of view of the decoder needs to be `[SWAP, PUSH, NOOP]` (i.e. exactly one NOOP inserted, not 7). Similarly, for the dense group `[ADD]`, the "sparse" representation from the decoder's POV is exactly `[ADD]`. From our conversation though, it seems like this is handled in how the iterators are defined - but I haven't look into the details of this test to see why this is happening.

huitseeker: Also, @bobbinth was looking at a variant of this PR where I was still padding to const BATCH_SIZE == 8 rows, whereas I now pad to the nearest power of two.

I think the padding pattern is incorrect for now, but we can handle this differently.

assembly/src/snapshots/miden_assembly__tests__assert_eqw_with_code.snap
huitseeker: I think this is an old version of the PR that padded blocks to specifically contain 8 batches. We're now padding to the nearest power of two.

assembly/src/snapshots/miden_assembly__tests__basic_block_and_simple_if_true-2.snap
huitseeker: Oh, the *groups* (not batches) are padded to a power of two? That's a bit surprising (though @plafer 's [comment](https://github.com/0xMiden/miden-vm/pull/2092#discussion_r2294213775)) helps, but it can be done.

plafer: EDIT: nevermind, indeed there should be 2 NOOPs: 1 at the end of the op group (since it ends with a PUSH), and 1 to pad the number of groups from 3 to 4 in the batch.

huitseeker:
OK, so two reasons why I'm closing this PR:
1. The rules for padding (or more exactly group-splitting) are precisely adding the *minimal* number of noops per group so that:
    - we can't end a group with an immediate value and
    - we want the number of batches rounded to the nearest power of two
    However, the current PR pads the group *to a fixed length*
2. With a little bit of metadata, padded and unpadded iterations can be views of each other. This implements the unpadded view as the major one — from which the padded iteration is deduced—, and we would like to implement the padded view as the major (or 'reference') one, to provide strong guarantees of iteration speed.

plafer:
Also just to make sure you keep this in the back of your mind when designing the next iteration, we would like a way to iterate over the operations of a basic block, with `batch_idx` and `operation_idx_in_batch` metadata for each operation - basically the indices to populate [this](https://github.com/0xMiden/miden-vm/blob/df49ba71143b2046672cc5d8ebb31aa64ddd988b/processor/src/fast/mod.rs#L1264-L1268) `NodeExecutionPhase` (taken from #2023) without needing to compute anything at runtime.

PR #1824, opened 2025-05-16
`OpBatch` refactor

plafer:
Closes #1815 

Inserts all the NOOPs due to op batch constraints at assembly-time rather than at runtime.

Left to do:
- Fix `DecoratorList` representation
  - The index associated with a decorator is in the list of operations *before* NOOPs are appended, and so are currently broken by this change
- Fix serialization/deserialization
- Think: return an error when group ends with an operation containing an immediate value?
  - keeping track of this will slow the fast processor down
- Check if some source spans are broken after we insert noops (so that the `op_idx` would be invalidated)

PR #2094, opened 2025-08-25
feat: Add Noops to OpBatches

### LINKED ISSUE #1815: Refactor `OpBatch` so that all NOOPs are inserted

### LINKED ISSUE #1965: MastForest serialization should preserve basic_blocks groups

huitseeker:
## Summary

This refactoring addresses issue #1815 by eliminating the dynamic insertion of NOOPs during fast processor execution. Instead, NOOPs are now inserted at assembly time within the `OpBatch` representation, significantly reducing bookkeeping overhead in the fast path.

## Notes to reviewers

1. Introduction of `indptr` and `padding` arrays in `OpBatch`
- **`indptr`**: An array of indexes marking the beginning and end of each operation group, following sparse matrix format conventions (name from SciPy's CSR matrix `indptr`)
- **`padding`**: Tracks which groups have been padded with NOOPs to maintain proper group sizing

2. Separation into distinct commits is load-bearing for review

3. New methods in `OpBatch`:
- `raw_ops()`: Iterator over operations excluding padding NOOPs
- `end_indices()`: Returns end indexes of each group. This array is relative to the index of the op in the batch, and replaces old `op_counts()`, which was relative to an index of the same op in its group.

4. Assembly integration:
- Updated `MastForestBuilder` to use `raw_ops()` instead of `ops()` for correct iteration in merging and serialization

<details><summary> Technical details</summary>

### Sparse Matrix Analogy
The `indptr` array follows the same pattern as sparse matrix representations (e.g., SciPy's CSR format), where:
- `indptr[i]` is the start index of group `i`
- `indptr[i+1]` is the end index of group `i`  
- Groups with padding have their end index adjusted to exclude the trailing NOOP

### NOOP Insertion Logic
NOOPs are inserted when:
1. An operation group ends with an immediate value operation
2. An operation group would otherwise be empty
3. During power-of-two padding of batch sizes (reusing 2)

### Prior work
- https://github.com/0xMiden/miden-vm/pull/1824
- https://github.com/0xMiden/miden-vm/pull/1950/
- https://github.com/0xMiden/miden-vm/pull/2092

## Testing
- Comprehensive snapshot testing ensures correct batch construction
- Property-based testing validates accumulator behavior across various operation sequences
</details>

## TODO
- [x] clean up processor padding
- [x] adjust decorators <-> op alignment in merging

Closes #1815, closes #1965


core/Cargo.toml
```diff
@@ -40,5 +40,6 @@ winter-math.workspace = true
 winter-utils.workspace = true
 
 [dev-dependencies]
+insta = "1.43.1"
```
bobbinth: nit: `1.43.1` -> `1.43`.

Also, should we make it a workspace dependency (since it is used on a coupe of crates now)?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,44 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
 /// operations or a single immediate value.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct OpBatch {
+    /// A list of operations in this batch, including decorators.
     pub(super) ops: Vec<Operation>,
```
bobbinth: What does "including decorators" means here? Operations don't usually include decorators.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,44 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
+    pub(super) indptr: [usize; Self::BATCH_SIZE_PLUS_ONE],
+    /// An array of bits representing whether a group had undergone padding
+    pub(super) padding: [bool; BATCH_SIZE],
+    /// Values of operation groups, including immediate values.
     pub(super) groups: [Felt; BATCH_SIZE],
-    pub(super) op_counts: [usize; BATCH_SIZE],
+    /// Number of non-decorator groups in this batch.
+    /// The arrays above are meaningful in their [0..self.num_groups] prefix
+    /// (or [0..self.num_groups + 1] in the case of the indptr array).
     pub(super) num_groups: usize,
```
bobbinth: Similar question as above: what does "non-decorator groups" mean here? Do you perhaps mean non-noop groups (i.e., groups that don't solely consist of a single noop?)

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,44 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
 }
 
 impl OpBatch {
+    /// The size of the indptr array, made to maintain its invariant
+    const BATCH_SIZE_PLUS_ONE: usize = BATCH_SIZE + 1;
+
     /// Returns a list of operations contained in this batch.
+    pub fn ops(&self) -> &[Operation] {
```
bobbinth: nit: I would maybe mention in the doc comment that this returns padding noops as well.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,44 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
+        &self.ops
+    }
+
+    /// Returns a list of operations contained in this batch, without any decorators.
```
bobbinth: Was "without any decorators" meant to be "without padding noops"?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -34,11 +58,38 @@ impl OpBatch {
         &self.groups
     }
 
-    /// Returns the number of non-decorator operations for each operation group.
+    /// Returns a list of indexes in the ops array, marking the beginning and end of each group.
+    /// The array maintains the invariant that the i-th group (i <= BATCH_SIZE-1) is at
```
bobbinth: nit: I would insert a line break between this lines (it'll render better in the docs this way).

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -181,11 +181,24 @@ impl BasicBlockNode {
...
+    }
+
     /// Returns an iterator over the operations in the order in which they appear in the program.
     pub fn operations(&self) -> impl Iterator<Item = &Operation> {
         self.op_batches.iter().flat_map(|batch| batch.ops())
     }
 
+    /// Returns an iterator over the un-decorated (un-padded) operations in the order in which they
+    /// appear in the program.
+    pub fn raw_operations(&self) -> impl Iterator<Item = &Operation> {
```
bobbinth: Similar comment re some other ones "un-decorated" and "un-padded" mean two different things: decorators are "extra instructions" that don't change the VM state (and don't advance the clock cycle) but may be useful primarily for debugging purposes. Padding operations are basically just noops which don't change the state of the VM but do advance the clock cycle.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -181,11 +181,24 @@ impl BasicBlockNode {
         DecoratorIterator::new(&self.decorators)
     }
 
+    /// Returns a [`RawDecoratorIterator`] which allows us to iterate through the decorator list of
+    /// this basic block node with op indexes aligned to the "raw" (unpadded / undecorated) op
+    /// batches of the basic block node
+    pub fn raw_decorator_iter(&self) -> RawDecoratorIterator<'_> {
```
bobbinth: Similar comment as above re "unpadded" vs. "undecorated".

assembly/src/snapshots/miden_assembly__tests__can_push_constant_word.snap
```diff
@@ -3,5 +3,5 @@ source: assembly/src/tests.rs
 expression: program
 ---
 begin
-    basic_block push(2) push(3) push(4) push(5) end
+    basic_block push(2) push(3) push(4) push(5) noop noop noop end
```
bobbinth: Should this not have 4 `noop`s at the end? 

I'm assuming the source program here is `begin push.2 push.3 push.4 push.5 end`, and then the op groups look like so:

```
[push, push, push push, noop] // need to add a noop so that the group does not end with a push
[2]                           // immediate value
[3]                           // immediate value
[4]                           // immediate value
[5]                           // immediate value
[noop]                        // need to make sure we round up to 8 groups
[noop]                        // need to make sure we round up to 8 groups
[noop]                        // need to make sure we round up to 8 groups
```
But also, since the tests are passing, maybe I'm not remembering some rules correctly?

assembly/src/snapshots/miden_assembly__tests__const_word_from_string.snap
```diff
@@ -0,0 +1,15 @@
...
+---
+begin
+    basic_block
+        push(10613494933894083877)
+        push(4225041690115989743)
+        push(1168993639339072050)
+        push(13186528713821658146)
+        noop
+        noop
+        noop
```
bobbinth, line 13: Similar comment as above: shouldn't here be 4 `noop`s at the end?

processor/src/fast/mod.rs
```diff
@@ -928,12 +923,10 @@ impl FastProcessor {
...
-        // make sure we execute the required number of operation groups; this would happen when the
-        // actual number of operation groups was not a power of two. In this processor, this
-        // corresponds to incrementing the clock by the number of empty op groups (i.e. 1 NOOP
-        // executed per missing op group).
-
-        self.clk += (num_batch_groups - group_idx) as u32;
+        // by construction, batch.num_groups() is a power of two
+        for _ in group_idx..batch.num_groups() {
+            self.clk += 1_u32;
+        }
```
bobbinth: Could we avoid the `for` loop here? Basically, do something like:
```rust
self.clk += batch.num_noop_groups()`
```
The motivation is that we'd be executing this at the end of every batch, and the fewer branches we add to the program, the better.

bobbinth (commented):
Looks good! Thank you! This is not a full review, but I left some comments/questions inline. Overall, the structure makes sense.

It would be good to rebase this from the latest `next` as there are some small conflicts.

processor/src/fast/mod.rs
plafer: Should we be removing it entirely though, if batches are now always padded to be a power of 2 in the assembler?

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Yeah, I had written those comments about 10 days ago or so, and they were badly in need of a rewording.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -34,11 +59,39 @@ impl OpBatch {
...
+            _ => unreachable!("windows invariant violated"),
+        })
+    }
+
+    /// Returns the end indexes of each group.
+    pub fn end_indices(&self) -> &[usize; BATCH_SIZE] {
+        // SAFETY:
+        // - indptr is an array of length BATCH_SIZE+1, so elements 1..=BATCH_SIZE form exactly
+        //   BATCH_SIZE contiguous `usize`s.
+        // - `as_ptr().add(1)` is in-bounds and properly aligned.
```
plafer:
```suggestion
        // - `as_ptr().add(1)` is in-bounds and properly aligned, since `[T; N]` has the same alignment requirements as `T` (see [layout.array] in the reference).
```
I would explicitly document the alignment requirements of arrays, since I wasn't sure if they needed to be aligned on `size_of::<T>` or `N * size_of::<T>`.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -34,11 +59,39 @@ impl OpBatch {
...
+    pub(super) fn group_chunks(&self) -> impl Iterator<Item = &[Operation]> {
+        self.indptr[..=self.num_groups].windows(2).map(|slice| match slice {
+            [start, end] => &self.ops[*start..*end],
+            _ => unreachable!("windows invariant violated"),
+        })
+    }
+
+    /// Returns the end indexes of each group.
+    pub fn end_indices(&self) -> &[usize; BATCH_SIZE] {
+        // SAFETY:
```
plafer, line 101: is there a debug/const assertion we can add that makes sure that if we change the size of `indptr` (however unlikely), then the assertion fails?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -52,13 +105,17 @@ impl OpBatch {
...
+    /// A list of operations in this batch, including noops.
     ops: Vec<Operation>,
+    /// An array of indexes in the ops array, marking the beginning and end of each group.
+    /// The array maintains the invariant that the i-th group (i <= BATCH_SIZE-1) is at
+    /// `self.ops[self.indptr[i]..self.indptr[i+1]]`.
+    indptr: [usize; OpBatch::BATCH_SIZE_PLUS_ONE],
+    /// An array of bits representing whether a group had undergone padding with a
+    /// noop at the end of the group.
+    padding: [bool; BATCH_SIZE],
     /// Values of operation groups, including immediate values.
```
plafer:
```suggestion
    /// Value of groups in the batch, which includes operations and immediate values.
```
Let's fix this comment (from before this PR)

assembly/src/snapshots/miden_assembly__tests__can_push_constant_word.snap
huitseeker: Yep, there was an OB1 in the materialization of the last noop in a padding (empty) group.

processor/src/fast/mod.rs
huitseeker: @plafer correct, this loop was a manifestation of the OB1 in the [above comment](https://github.com/0xMiden/miden-vm/pull/2094#discussion_r2305804039). Removed.

assembly/src/snapshots/miden_assembly__tests__const_word_from_string.snap
huitseeker: And now there are.

processor/src/fast/mod.rs
```diff
@@ -910,7 +905,7 @@ impl FastProcessor {
             }
 
             // determine if we've executed all non-decorator operations in a group
-            if op_idx_in_group == op_counts[group_idx] - 1 {
+            if op_idx_in_batch + 1 == end_indices[group_idx] {
                 // if we are at the end of the group, first check if the operation carries an
                 // immediate value
                 if has_imm {
```
plafer: I think we can remove this block, since all NOOPs should already be inserted (and similarly in `Process`)

plafer (commented):
Looks good! Only a partial review so far - but will leave comments as I go since you're currently working on it

processor/src/fast/mod.rs
```diff
@@ -932,13 +927,6 @@ impl FastProcessor {
             self.clk += 1_u32;
         }
 
-        // make sure we execute the required number of operation groups; this would happen when the
-        // actual number of operation groups was not a power of two. In this processor, this
-        // corresponds to incrementing the clock by the number of empty op groups (i.e. 1 NOOP
-        // executed per missing op group).
-
-        self.clk += (num_batch_groups - group_idx) as u32;
-
```
plafer, line 941: Now that the processor no longer enforces the padding NOOP invariants, we should add a `MastForest::validate()` method that ensures that the `MastForest` is properly built (including but not limited to the padding NOOP invariants). We can then document in the `FastProcessor::execute_*()` methods that it is the responsibility of the caller to ensure that the program's `MastForest` is valid.

Similarly for `MastForest`s that we get from the host when executing an external node, we should offer the option to validate those or not (probably through some `ExecutionOptions` struct we pass to the execute methods).

This should be done as a follow-up PR though.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -167,13 +238,137 @@ impl OpBatchAccumulator {
...
+
+    /// Saves the start index of the upcoming group (at self.next_group_idx), corrects any groups
+    /// created through immediate values.
+    #[inline]
+    fn finalize_indptr(&mut self) {
+        // we are finalizing a group, we now know the start of the upcoming group
+        self.indptr[self.next_group_idx] = self.ops.len();
+        // we also need to correct the start indexes of groups carrying immediate values, if any,
+        let mut uninit_group_idx = self.next_group_idx - 1;
+        while uninit_group_idx > 0 && self.indptr[uninit_group_idx] == Self::INVALID_IDX {
```
plafer:
```suggestion
        while uninit_group_idx > self.group_idx && self.indptr[uninit_group_idx] == Self::INVALID_IDX {
```
I think this better communicates that we're going over all the immediate values associated with the current group

core/src/operations/decorators/mod.rs
```diff
@@ -120,3 +120,114 @@ impl<'a> Iterator for DecoratorIterator<'a> {
...
+    }
+}
+
+/// Iterator used to iterate through the decorator list of a span block
+/// while executing operation batches of a span block.
+///
+/// This lets the caller iterate through a Decorator list with indexes that match the
+/// raw (unpadded) representation of a basic block.
+pub struct RawDecoratorIterator<'a> {
+    decorators: &'a DecoratorList,
```
plafer: We should specify in the `DecoratorList` docstring that the operation index does *not* consider padding NOOPs

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,45 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
-    pub(super) op_counts: [usize; BATCH_SIZE],
+    /// Number of groups in this batch.
+    /// The arrays above are meaningful in their [0..self.num_groups] prefix
+    /// (or [0..self.num_groups + 1] in the case of the indptr array).
     pub(super) num_groups: usize,
 }
 
 impl OpBatch {
-    /// Returns a list of operations contained in this batch.
+    /// The size of the indptr array, made to maintain its invariant
```
Al-Kindi-0: Wouldn't mind re-stating the invariant

assembly/src/mast_forest_builder.rs
```diff
@@ -615,3 +615,187 @@ fn should_merge(is_procedure: bool, num_op_batches: usize) -> bool {
...
+                    // Determine which original operation this should point to
+                    // if op_idx < 3 {
+                    //     // This is from block1
+                    //     // Block1 operations: 0 -> Trace(1), 1 -> Trace(2), 2 -> Trace(3)
+                    //     assert_eq!(*trace_value, op_idx as u32 + 1);
+                    // } else {
+                    //     // This is from block2, adjust for block1's operations
+                    //     // Block2 operations: 0 -> Trace(4), 1 -> Trace(5)
+                    //     assert_eq!(*trace_value, (op_idx - 3) as u32 + 4);
+                    // }
```
plafer: Does this need to be uncommented?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -34,11 +59,39 @@ impl OpBatch {
...
-    /// Returns the number of non-decorator operations for each operation group.
+    /// Returns a list of indexes in the ops array, marking the beginning and end of each group.
+    ///
+    /// The array maintains the invariant that the i-th group (i <= BATCH_SIZE-1) is at
+    /// `self.ops[self.indptr[i]..self.indptr[i+1]]`.
+    pub fn indptr(&self) -> &[usize; Self::BATCH_SIZE_PLUS_ONE] {
+        &self.indptr
+    }
+
+    /// Returns a list of flags marking if each group of the batch contains a final padding Noop
```
Al-Kindi-0: if -> whether ?

assembly/src/mast_forest_builder.rs
```diff
@@ -615,3 +615,187 @@ fn should_merge(is_procedure: bool, num_op_batches: usize) -> bool {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use miden_core::Operation;
+
+    use super::*;
+
+    #[test]
+    fn test_merge_basic_blocks_preserves_decorator_links_with_padding() {
```
plafer, line 626: Great test! I think we should add a similar one for decorators across a padding NOOP, for example, in MASM for demonstration purposes,

```
begin
  add add add add add add add add trace.1
  # <NOOP INSERTED> due to upcoming PUSH at the end of group
  # <new op group begins>
  push.4
  trace.2 mul
end
```
We want to make sure that `trace.2` is associated with `mul` (i.e. operation index 10, due to the inserted NOOP).

I currently suspect this might fail, since our `DecoratorList` representation still uses operation indices without padding NOOPs, whereas in `FastProcessor` we use `op_idx_in_block`, which does consider padding NOOPs. But I might be missing something; and would be good to have the test regardless.

huitseeker: I've rewritten the test (and the fix), please take a look. As expected the order of commits matters, the test precedes the fix, and fails before — and only before — it.

core/src/mast/node/basic_block_node/op_batch.rs
huitseeker: Added.

processor/src/fast/mod.rs
huitseeker: Can you be a bit more specific as to what the scope of this validation method would entail? Noop padding rules at the BasicBlockNode level are an obvious item, but the scope beyond that is a bit unclear to me. I would note that padding-preserving merging is still managed by the forest builder, for example.

plafer: I was going more for a "catch-all"/future-proof `validate()` method. For today's struct though, off the top of my head I can think of
1. NOOP padding rules
2. all `MastNodeId`s point to a valid `MastNode` in the forest

> I would note that padding-preserving merging is still managed by the forest builder, for example.

That's right, but someone could always bypass the assembler and build a (potentially malicious) `MastForest` by hand. Hence the processor (or rollup node in general) should ensure that the `MastForest` users provide are valid.

Cargo.toml
```diff
@@ -59,6 +59,7 @@ miden-formatting = { version = "0.1", default-features = false }
 midenc-hir-type = { version = "0.1", default-features = false }
 
 # Third-party crates
+insta = "1.43"
 log = "0.4"
 paste = "1.0"
```
Al-Kindi-0: Should we use the same convention used throughout the rest of the file here as well?

assembly/src/mast_forest_builder.rs
```diff
@@ -615,3 +615,193 @@ fn should_merge(is_procedure: bool, num_op_batches: usize) -> bool {
...
+        let mut builder = MastForestBuilder::new(&[]).unwrap();
+
+        // We need to create a benchmark with a removed Noop operation *in the middle* of the batch
+        // (not at the end). That's because across batches, decorators are re-indexed (shifted) by
+        // the amount of concrete operations in the previous batch in the sequence, and that
+        // re-indexing remains valid whether or not *final* pading is elided.
```
Al-Kindi-0: padding

assembly/src/mast_forest_builder.rs
```diff
@@ -615,3 +615,193 @@ fn should_merge(is_procedure: bool, num_op_batches: usize) -> bool {
...
+
+        // Create first block with operations that will cause padding (ending with Push)
+        // Block1: [Push(1), Add, Push(2)]
```
Al-Kindi-0: should this be updated?

assembly/src/mast_forest_builder.rs
```diff
@@ -615,3 +615,193 @@ fn should_merge(is_procedure: bool, num_op_batches: usize) -> bool {
...
+                    // Verify that the decorator points to the expected operation type
+                    // Get the raw operations to check what's at this index
+                    let merged_ops: Vec<Operation> = merged_block.operations().cloned().collect();
+
+                    if op_idx < merged_ops.len() {
+                        match op_idx {
+                            0 => {
+                                // Should be Push(1) from block1
+                                match &merged_ops[op_idx] {
+                                    Operation::Push(_) => {
```
Al-Kindi-0: nit: wouldn't hurt having a check for the inner value of the enum as well

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +469,83 @@ impl<'a> Iterator for OperationOrDecoratorIterator<'a> {
...
+            operations_len >= decorators.last().expect("empty decorators list").0,
+            "last op index in decorator list should be less than or equal to the number of ops"
+        );
+    }
+}
+
+// Indexes into an operations-long sequence of decorators. Should not be user-facing.
+//
+// a [`DecoratorList`] is a sequence of (op_idx, decorator_id) tuples such that the op_idx
+// can extend up to and including the length of the operations list it is meant to index in.
```
Al-Kindi-0: index into ?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,53 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
 pub struct OpBatch {
+    /// A list of operations in this batch, including padding noops.
     pub(super) ops: Vec<Operation>,
+    /// An array of indexes in the ops array, marking the beginning and end of each group.
+    ///
+    /// The array maintains the invariant that the i-th group (i <= BATCH_SIZE-1) is at
+    /// `self.ops[self.indptr[i]..self.indptr[i+1]]`.
+    ///
+    /// By convention, the groups containing immediate values have a zero-length slice of the ops
+    /// arrya.
```
Al-Kindi-0: array

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,53 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
     /// of operations in the batch may be larger than the number of operations reported by this
     /// method.
-    pub fn ops(&self) -> &[Operation] {
-        &self.ops
+    pub fn raw_ops(&self) -> impl Iterator<Item = &Operation> {
+        debug_assert!(self.num_groups == 0 || self.indptr[self.num_groups] != 0, "{:?}", self);
+        (0..self.num_groups).flat_map(|group_idx| {
+            let padded = self.padding[group_idx];
+            let start_idx = self.indptr[group_idx];
+            let end_idx = self.indptr[group_idx + 1] - usize::from(padded);
```
Al-Kindi-0, line 60: It could be helpful to explain the rational behind the subtraction

assembly/src/mast_forest_builder.rs
plafer: This new version does capture what I was saying indeed - but now the problem is with `BasicBlockBuilder` in the assembler. Basically, I expect that compiling the same `MastForest` but from MASM would fail - since `BasicBlockBuilder` doesn't know about padding NOOPs, and uses the [wrong operation index](https://github.com/0xMiden/miden-vm/blob/30d5bd4382aebc16440fc3d4d7493b6738c3440b/assembly/src/basic_block_builder.rs#L111) for decorators.

core/src/mast/node/basic_block_node/op_batch.rs
huitseeker: Added a comment.

assembly/src/mast_forest_builder.rs
huitseeker: Can you make the following parts of your statement more precise, I'm not sure I get what you mean:
- "know about padding NOOPs",
- "wrong" in "the wrong operation index for decorators".
?

As mentioned offline, post this PR any place that eventually calls `BasicBlockBuilder::new`, (as the `BasicBlockBuilder` does through `ensure_block`) with an `(operations, decorators)` pair that does not know about padding rules but *which match each other* will have:
- its operations sequence modified to include padding, and
- its decorators re-indexed to take the inserted padding into account

This preserves the invariant that the `BasicBlockNode` that comes out of `new` has operations and decorators that match each other.

processor/src/fast/mod.rs
```diff
@@ -910,35 +904,14 @@ impl FastProcessor {
             }
 
             // determine if we've executed all non-decorator operations in a group
-            if op_idx_in_group == op_counts[group_idx] - 1 {
-                // if we are at the end of the group, first check if the operation carries an
-                // immediate value
-                if has_imm {
```
plafer, line 916: we should also remove this NOOP-insertion block from `Process::execute_op_batch()`

processor/src/fast/tests/mod.rs
```diff
@@ -83,7 +83,7 @@ fn test_stack_underflow_and_overflow_bounds_failure() {
 
         // program 3: just enough dups to get 1 away from the stack buffer overflow error. Since we
         // can't drop the elements, we expect to end the program with a stack output overflow.
-        const NUM_DUPS_WITH_OVERFLOW: usize = NUM_DUPS_NO_OVERFLOW + 1;
+        const NUM_DUPS_WITH_OVERFLOW: usize = NUM_DUPS_NO_OVERFLOW + 2; // TODO(@plafer): confirm this is correct taking padding into account
```
plafer: The problem now with this test is that every time we change `STACK_BUFFER_SIZE` (which we will), then the number of inserted NOOPs will change as well, and therefore so will these constants. Also as it stands, it's very hard to tell that the `+ 1 - 2` and `+ 2` offsets are correct. And computing the correct values for these constants at runtime is not trivial either (i.e. the problem is: give me the number of `DUP` operations, if any, which amount to `NUM_DUPS_NO_OVERFLOW + 1` operations after padding NOOP are inserted, and that do not end in a NOOP).

I think I would revert the changes to the test, and `#[ignore]` it for now. We still need to fix the `FastProcessor` issue that once our stack buffer is full, we panic - we can revisit this test once we address that.

core/src/mast/node/basic_block_node/tests.rs
```diff
@@ -1,57 +1,153 @@
...
+        Just(Operation::Ext2Mul),
+        Just(Operation::Expacc),
+        Just(Operation::HPerm),
+        // Note: We exclude Assert here because it has an immediate value (error code)
+    ]
+}
+
+// Strategy for operations with immediate values
+fn op_with_imm_strategy() -> impl Strategy<Value = Operation> {
+    prop_oneof![any_felt().prop_map(Operation::Push)]
```
plafer, line 91: I would add a comment in `Operation::imm_value()` to come and fix this strategy (and the `no_imm_strategy()` if needed) whenever we add a new operation with an immediate value

plafer (approved):
LGTM! Beautiful solution to the problem.

assembly/src/mast_forest_builder.rs
plafer: As discussed offline, I had missed the logic in `BasicBlockNode` - and retract my previous comment.

processor/src/fast/mod.rs
plafer: @huitseeker if you agree with the above, can you create an issue from the above discussion so we can track it?

huitseeker: https://github.com/0xMiden/miden-vm/issues/2106

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -181,11 +202,24 @@ impl BasicBlockNode {
...
+    /// this basic block node with op indexes aligned to the "raw" (un-padded)) op
+    /// batches of the basic block node
+    pub fn raw_decorator_iter(&self) -> RawDecoratorIterator<'_> {
+        RawDecoratorIterator::new(&self.decorators, &self.op_batches)
+    }
+
     /// Returns an iterator over the operations in the order in which they appear in the program.
     pub fn operations(&self) -> impl Iterator<Item = &Operation> {
         self.op_batches.iter().flat_map(|batch| batch.ops())
     }
```
bobbinth, line 215: Question: this now returns operations with padding noops, right? If so, may be good to update the comment to make this clear.

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -11,20 +11,55 @@ use super::{BATCH_SIZE, Felt, GROUP_SIZE, Operation, ZERO};
...
+    /// An array of bits representing whether a group had undergone padding
+    pub(super) padding: [bool; BATCH_SIZE],
+    /// Value of groups in the batch, which includes operations and immediate values.
     pub(super) groups: [Felt; BATCH_SIZE],
-    pub(super) op_counts: [usize; BATCH_SIZE],
+    /// Number of groups in this batch.
+    ///
+    /// The arrays above are meaningful in their [0..self.num_groups] prefix
+    /// (or [0..self.num_groups + 1] in the case of the indptr array).
     pub(super) num_groups: usize,
```
bobbinth, line 32: At some point in the future (after we get rid of the slow processor), I wonder if we could get rid of these fields and maybe keep only the `ops` field as I think this is really the only field we need to execute programs in the fast processor.

Or, if we wanted to optimize this even further, we could have a single vector of operations for the entire basic block. It could look something like this:

```rust
pub struct BasicBlockNode {
    /// operations here would contain Respan operations to mark boundaries of operation batches
    ops: Vec<Operation>,
    digest: Word,
}
```

This structure would make serialization/deserialization and hash computations more difficult, but would make executing programs via the fast processor as fast as possible.

bitwalker: I definitely think that would be a lot easier to reason about, and the perf boost around raw execution would probably be significant. I don't think serialization/deserialization is much of a concern (i.e. it is fine if they are complex, since it is relatively rare), but since hashing is important for performance, it isn't clear to me whether it would be a net win or not. I think as long as it breaks even, or is close enough to not matter, then it would be nicer this way.

bobbinth: Created https://github.com/0xMiden/miden-vm/issues/2109 for this.

PR #2098, opened 2025-08-27
Adds the option to prove using Poseidon2

Al-Kindi-0:
## Describe your changes
The PR also changes the options to the prove command by adding a hasher parameter.
A separate point that is worth considering is whether or not we want to remove the option `blake192` from the whole repo. Though I see the rational for having it at the 96 bit security level, I think that it will be rather useless in the medium term and in the short term it is just a nuisances in relation to uniformity. Moreover, performance-wise, I am not sure there is any significant difference between the `192` and `256` variants (probably they use the same core and truncation for `192`).

- Updated `CHANGELOG.md'


air/src/options.rs
```diff
@@ -94,65 +94,51 @@ impl ProvingOptions {
...
-        } else {
-            Self {
-                exec_options: ExecutionOptions::default(),
-                proof_options: Self::REGULAR_96_BITS,
-                hash_fn: HashFunction::Blake3_192,
-            }
+    /// In this setting, proofs will be generated using an arithmetization-friendly hash
+    /// function (RPO). Such proofs are well-suited for recursive proof verification, but may
+    /// take significantly longer to generate.
+    pub fn with_96_bit_security_recursive() -> Self {
```
bobbinth: nit: I wonder if a better name could be `with_96_bit_security_recursion_friendly()`. It is a bit longer but I think "recursive proof" term is somewhat imprecise.

air/src/options.rs
```diff
@@ -94,65 +94,51 @@ impl ProvingOptions {
...
+            hash_fn: HashFunction::Rpo256,
         }
     }
 
-    /// Creates a new preset instance of [ProvingOptions] targeting 96-bit security level,
-    /// using the RPX hashing function.
-    pub fn with_96_bit_security_rpx() -> Self {
+    /// Creates a new preset instance of [ProvingOptions] targeting 96-bit security level, given
+    /// a choice of a hash function, in the non-recursive setting.
+    pub fn with_96_bit_security_non_recursive(hash_fn: HashFunction) -> Self {
```
bobbinth: Similar comment as above: maybe like `with_96_bit_security_not_recursion_friendly()` is a bit better naming? Or we could even shorten it to just `with_96_bit_security()`.

miden-vm/src/cli/prove.rs
```diff
@@ -50,9 +50,11 @@ pub struct ProveCmd {
...
     recursive: bool,
 
-    /// Specifies if the RPX Hash should be used. Conflicts with the recursive flag
-    #[arg(long = "rpx", conflicts_with("recursive"))]
-    rpx: bool,
+    /// Specifies the hash function to be used in case we are not in the recursive setting.
+    /// Conflicts with the recursive flag. Valid options: blake3-192, blake3-256, rpo, rpx,
+    /// poseidon2
+    #[arg(long = "hasher", conflicts_with("recursive"), default_value = "blake3-192")]
+    hasher: String,
```
bobbinth: I'm wondering if we actually need both of these. If we are introducing `hasher` option, maybe we should just get rid of `recursive` option?

If we do that, we can also simplify `ProvingOptions` constructors - e.g., we could have just `ProvingOptions::with_128_bit_security(hasher)`. And internally, we can assume that anything besides BLAKE3 should use the recursive settings.

bobbinth (approved):
Looks good! Thank you! I left some small comments inline.

One thing I'm wondering is how performance looks like for Poseidon2. E.g., if we run it against RPX and RPO, would we get 2x and 4x improvement respectively?

> A separate point that is worth considering is whether or not we want to remove the option `blake192` from the whole repo. Though I see the rational for having it at the 96 bit security level, I think that it will be rather useless in the medium term and in the short term it is just a nuisances in relation to uniformity. Moreover, performance-wise, I am not sure there is any significant difference between the `192` and `256` variants (probably they use the same core and truncation for `192`).

The main purpose for the 192-bit version is reduction in proof size (there should be no difference in proving speed). The proof size should probably be almost 20% smaller - thought, it's been a while since I actually checked. Could you check what the delta actually is for our current setting? If it is not too big, we should definitely get rid of the 192-bit variant.

air/src/options.rs
Al-Kindi-0: Changed

air/src/options.rs
Al-Kindi-0: Done

miden-vm/src/cli/prove.rs
Al-Kindi-0: Agreed and changed

Al-Kindi-0:
> Looks good! Thank you! I left some small comments inline.
> 
> One thing I'm wondering is how performance looks like for Poseidon2. E.g., if we run it against RPX and RPO, would we get 2x and 4x improvement respectively?

Here are some numbers on the `blake3_1to1` example with number of iterations equal to 20

```
RPO          22.8s
RPX          12.6s
Poseidon2     7.6s
Blake3-256    1.5s
```

> The main purpose for the 192-bit version is reduction in proof size (there should be no difference in proving speed). The proof size should probably be almost 20% smaller - thought, it's been a while since I actually checked. Could you check what the delta actually is for our current setting? If it is not too big, we should definitely get rid of the 192-bit variant.

I can confirm that it is around 17% (for different trace lengths). Should we still keep it?

bobbinth (approved):
Looks good! Thank you!

> Here are some numbers on the `blake3_1to1` example with number of iterations equal to 20
> 
> ```
> RPO          22.8s
> RPX          12.6s
> Poseidon2     7.6s
> Blake3-256    1.5s
> ```

What was the cycle count for the above benchmarks?

Running `blake3_1to1` example for $2^{20}$ cycles on an M4 Max gives me the following results:

- RPO: 16.6 sec (with Metal acceleration)
- RPX: 10.2 sec (with Metal acceleration)
- Poseidon2: 11.7 sec (no acceleration)
- Blake3-256: 3.3 sec (no acceleration)

It is interesting that Metal-accelerated RPX is just a little bit faster than plain Poseidon2.

> I can confirm that it is around 17% (for different trace lengths). Should we still keep it?

Let's create an issue for this. 17% is not a lot, but also not negligible - so, I'm kind of on the fence about this.

Al-Kindi-0:
> What was the cycle count for the above benchmarks?

This is for $2^{16}$ cycles

> It is interesting that Metal-accelerated RPX is just a little bit faster than plain Poseidon2.

Seems congruent with previous conclusions, namely that RPO benefits more from acceleration (in relative term). This may change though if one adopts the approach of Plonky3 of accelerating across hash invocations instead of the intra approach adopted in our crypto crate.

> Let's create an issue for this. 17% is not a lot, but also not negligible - so, I'm kind of on the fence about this.

Done

Issue #1978, opened 2025-07-10
Investigate removing `ErrorContext`


plafer:
See https://github.com/0xMiden/miden-vm/pull/1884#issuecomment-3056286215 and https://github.com/0xMiden/miden-vm/pull/1884#issuecomment-3058246443

bobbinth:
I think this is probably going to be related to https://github.com/0xMiden/miden-vm/issues/1942.

adr1anh:
I was playing around with this idea using Gemini, where the `ExecutionError` is just a wrapper around another error with the source and span. The main difficulty is enabling the wrapped error to define custom label. 

```rust
// --- Core Trait ---
// This trait defines the contract for any error that can be part of an ExectionError.
// It's responsible for providing the specific text for its own label.
trait ExecErrorTrait: Diagnostic + Send + Sync + 'static + Error {
    fn custom_label(&self) -> Option<String>;
}

// --- Outer Error Container ---
#[derive(Debug)]
struct ExecutionError {
    pub source_file: Arc<SourceFile>,
    pub span: SourceSpan, // Renamed from `label` for clarity.
    pub err: Box<dyn ExecErrorTrait>,
}

impl Display for ExecutionError {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        write!(f, "execution error")
    }
}

// Manual `Error` implementation to correctly expose the source.
impl Error for ExecutionError {
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(self.err.as_ref())
    }
}
// Manual miette implementation which allows the inner error to define its own label
impl Diagnostic for ExecutionError {
    fn code<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
        // We can forward the code from the inner error.
        self.err.code()
    }

    fn source_code(&self) -> Option<&dyn miette::SourceCode> {
        Some(&*self.source_file)
    }

    fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
        // Use the trait method to get the label text, no downcasting needed here.
        self.err.custom_label().map(|text| {
            let labeled_span = LabeledSpan::new_with_span(Some(text), self.span.clone());
            Box::new(std::iter::once(labeled_span)) as Box<dyn Iterator<Item = _>>
        })
    }
}
```

Then, we can take as example the `MemoryError` which would then be implemented as 
```rust
// --- Example Concrete Error Implementation ---
[...]

PR #2101, opened 2025-08-27
Introduce `Tracer` trait

### LINKED ISSUE #2089: Introduce `Tracer` trait for `FastProcessor` performance and debugging
In this issue I will describe a follow-up improvement to #2023 that simultaneously fixes the performance issues incurred by all the `if` statements (on the [`CoreTraceStateBuilder`](https://github.com/0xMiden/miden-vm/blob/1b33b08b978747d453dfba940604252ff6095a2f/processor/src/fast/trace_state_builder.rs#L34-L53)), as well as providing an API for a debugger to be able to step through a program operation by operation.

The main idea is to create a trait `Tracer` to abstract away `CoreTraceStateBuilder`. It would look approximately as

```rs
pub trait Tracer {
  // This is called at the beginning of each clock cycle, and effectively provides
  // the node ID and stable operation index (when in a basic block) through `exec_phase`,
  // as well as the `FastProcessor` directly to interrogate it about its state.
  // It is roughly analogous to `FastProcessor::check_trace_state()` in #2023.
  // This is the only method the debugger would care about.
  fn capture_trace_state(&mut self, 
    processor: &FastProcessor, 
    exec_phase: NodeExecutionPhase, 
    mast_forest: Arc<MastForest>
  );

  // There is one of these for each replay-recording operation on the `CoreTraceStateBuilder`.
  fn record_memory_read_element(&mut self, element: Felt, addr: RowIndex);

  // ...
}

impl Tracer for () {
  // We implement every method to be empty, and force the #[inline(always)].
  // This will have the effect of compiling away all `Tracer` calls for the `()`
  // implementation (which we would use in `FastProcessor::execute()` when we don't
  // care about generate trace states for trace generation.
  #[inline(always)]
  fn capture_trace_state(&mut self, 
    processor: &FastProcessor, 
    exec_phase: NodeExecutionPhase, 
    mast_forest: Arc<MastForest>
  ) {
    // do nothing
  }
 
  // ...
}

impl Tracer for CoreTraceStateBuilder {
  // This is implemented essentially as in #2023
}
```

## Fixing performance issues introduced in #2023

As mentioned above, we fix the performance issues by compiling away all the calls to `CoreTraceStateBuilder` when they're not needed (i.e. when using `FastProcessor::execute()`). The `FastProcessor` execute API would then be modified as

```rs
impl FastProcessor {
    pub async fn execute(
        mut self,
        program: &Program,
        host: &mut impl AsyncHost,
    ) -> Result<(StackOutputs, AdviceProvider), ExecutionError> {
        // pass in &mut () for the tracer, compiling away all calls to its methods
        // in `execute_impl()`
        let mut tracer = ();
        let stack_outputs = self.execute_impl(program, host, &mut tracer).await?;

        Ok((stack_outputs, self.advice))
    }

    pub async fn execute_for_trace(
        mut self,
        program: &Program,
        host: &mut impl AsyncHost,
    ) -> Result<(StackOutputs, AdviceProvider, Vec<CoreTraceState>), ExecutionError> {
        let mut tracer = CoreTraceStateBuilder::default();
        let stack_outputs = self.execute_impl(program, host, &mut tracer).await?;

        Ok((
            stack_outputs,
            self.advice,
            tracer.into_core_trace_states(),
        ))
    }

    async fn execute_impl(
        &mut self,
        program: &Program,
        host: &mut impl AsyncHost,
        tracer: &mut impl Tracer,
    ) -> Result<StackOutputs, ExecutionError> {
      // as in #2023, but uses tracer instead of `self.trace_state_builder`
    }
}
```

## Debugger

The debugger would also implement the `Tracer` trait, roughly as

```rs
// This is a dummy example pseudocode that would store the stack state and operation executed at every clock
struct DebuggerTracer {
  // (clk, operation, stack)
  states: Vec<(RowIndex, Operation, Vec<Felt>)>
}

impl Tracer for DebuggerTracer {
  fn capture_trace_state(
    &mut self, 
    processor: &FastProcessor, 
    exec_phase: NodeExecutionPhase, 
    mast_forest: Arc<MastForest>
  ) {
    // exec_phase encodes the node id + operation index, so along with the MAST forest
    // we have all the necessary information to derive the Operation
    let operation: Operation = self.get_operation(&exec_phase, mast_forest);
    self.states.push((processor.clk, operation, processor.stack().clone()));
  }

  // We don't need any of the other methods
  // ...
}
```

Then, the `FastProcessor` would expose a new low-level method that allows to run a single continuation off its internal `ContinuationStack` (e.g. starting/finishing a node, or executing a full basic block). This is so that the debugger gets back control at many points during execution, even when user programs are stuck in an infinite loop (where `FastProcessor::execute()` would never return).

```rs
impl FastProcessor {
    pub async fn step_continuation(
        mut self,
        program: &Program,
        host: &mut impl AsyncHost,
        tracer: &mut impl Tracer,
    ) -> Result<(StackOutputs, AdviceProvider), ExecutionError> {
       // this is `execute_impl()`, except just one step of its internal `while` loop
    }
}
```

To drive execution, the debugger would then call `step_continuation()`, passing it a `DebuggerTracer`, and then go through the collected information in `DebuggerTracer`. For example,

```rs
let mut tracer = DebuggerTracer::new();
let program = Program::default();
let mut processor = FastProcessor::new(stack_inputs);

while processor.step_continuation(&program, &mut host, &mut tracer).await.is_continue() {
    // processor called `DebuggerTracer::capture_trace_state()` at the beginning of each clock
    // cycle that it executed during the continuation
    for (clk, operation, stack) in tracer.states() {
        println!("stack at clock {clk} was {stack:?} before executing {operation:?}");
    }
    
    // in this example, we wipe the debugger tracer at each continuation, because
    // we don't need past states. In a real implementation, we might want to keep
    // them around
    tracer = DebuggerTracer::new();
}
```

cc @bobbinth @bitwalker

plafer:
Closes #2089

Stacked on #2023

Introduces the `Tracer` trait, as discussed in #2089. This also cleaned up the `CoreTraceBuilder`, which was left over from #2023.

This is a fairly mechanical change; the general idea is that in the fast processor, we swapped all "if we have a core trace builder, record X" calls to a `tracer.record_x()` call. We then pass in `NoopTracer` (with an empty body for every method) when executing from `execute()`, and pass in `CoreTraceStateBuilder` when executing from `execute_from_trace()`. 

Note that we added a new method `FastProcessor::execute_with_tracer()`, but leaves the discussed `FastProcessor::step_continuation()` (needed for the debugger) for a subsequent PR.

Performance-wise, we recovered the benchmark performance prior to #2023 (i.e. the hypothesis that the compiler would compile away all calls to the `NoopTracer` is correct).

### DIFF
--- a/processor/src/fast/field_ops.rs
+++ b/processor/src/fast/field_ops.rs
@@ -1,17 +1,19 @@
 use miden_core::{Felt, FieldElement, ONE, ZERO};
 
 use super::{ExecutionError, FastProcessor};
-use crate::{ErrorContext, operations::utils::assert_binary};
+use crate::{ErrorContext, fast::tracer::Tracer, operations::utils::assert_binary};
 
 const TWO: Felt = Felt::new(2);
 
 impl FastProcessor {
     /// Analogous to `Process::op_add`.
-    pub fn op_add(&mut self) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| Ok(a + b))
+    #[inline(always)]
+    pub fn op_add(&mut self, tracer: &mut impl Tracer) -> Result<(), ExecutionError> {
+        self.pop2_applyfn_push(|a, b| Ok(a + b), tracer)
     }
 
     /// Analogous to `Process::op_neg`.
+    #[inline(always)]
     pub fn op_neg(&mut self) -> Result<(), ExecutionError> {
         let element = self.stack_get(0);
         self.stack_write(0, -element);
@@ -19,8 +21,9 @@ impl FastProcessor {
     }
 
     /// Analogous to `Process::op_mul`.
-    pub fn op_mul(&mut self) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| Ok(a * b))
+    #[inline(always)]
+    pub fn op_mul(&mut self, tracer: &mut impl Tracer) -> Result<(), ExecutionError> {
+        self.pop2_applyfn_push(|a, b| Ok(a * b), tracer)
     }
 
     /// Analogous to `Process::op_inv`.
@@ -35,6 +38,7 @@ impl FastProcessor {
     }
 
     /// Analogous to `Process::op_inc`.
+    #[inline(always)]
     pub fn op_incr(&mut self) -> Result<(), ExecutionError> {
         *self.stack_get_mut(0) += ONE;
         Ok(())
@@ -42,24 +46,38 @@ impl FastProcessor {
 
     /// Analogous to `Process::op_and`.
     #[inline(always)]
-    pub fn op_and(&mut self, err_ctx: &impl ErrorContext) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| {
-            assert_binary(b, err_ctx)?;
-            assert_binary(a, err_ctx)?;
+    pub fn op_and(
+        &mut self,
+        tracer: &mut impl Tracer,
+        err_ctx: &impl ErrorContext,
+    ) -> Result<(), ExecutionError> {
+        self.pop2_applyfn_push(
+            |a, b| {
+                assert_binary(b, err_ctx)?;
+                assert_binary(a, err_ctx)?;
 
-            if a == ONE && b == ONE { Ok(ONE) } else { Ok(ZERO) }
-        })
+                if a == ONE && b == ONE { Ok(ONE) } else { Ok(ZERO) }
+            },
+            tracer,
+        )
     }
 
     /// Analogous to `Process::op_or`.
     #[inline(always)]
-    pub fn op_or(&mut self, err_ctx: &impl ErrorContext) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| {
-            assert_binary(b, err_ctx)?;
-            assert_binary(a, err_ctx)?;
+    pub fn op_or(
+        &mut self,
+        tracer: &mut impl Tracer,
+        err_ctx: &impl ErrorContext,
+    ) -> Result<(), ExecutionError> {
+        self.pop2_applyfn_push(
+            |a, b| {
+                assert_binary(b, err_ctx)?;
+                assert_binary(a, err_ctx)?;
 
-            if a == ONE || b == ONE { Ok(ONE) } else { Ok(ZERO) }
-        })
+                if a == ONE || b == ONE { Ok(ONE) } else { Ok(ZERO) }
+            },
+            tracer,
+        )
     }
 
     /// Analogous to `Process::op_not`.
@@ -77,11 +95,13 @@ impl FastProcessor {
     }
 
     /// Analogous to `Process::op_eq`.
-    pub fn op_eq(&mut self) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| if a == b { Ok(ONE) } else { Ok(ZERO) })
+    #[inline(always)]
+    pub fn op_eq(&mut self, tracer: &mut impl Tracer) -> Result<(), ExecutionError> {
+        self.pop2_applyfn_push(|a, b| if a == b { Ok(ONE) } else { Ok(ZERO) }, tracer)
     }
 
     /// Analogous to `Process::op_eqz`.
+    #[inline(always)]
     pub fn op_eqz(&mut self) -> Result<(), ExecutionError> {
         let top = self.stack_get_mut(0);
         if (*top) == ZERO {
@@ -93,6 +113,7 @@ impl FastProcessor {
     }
 
     /// Analogous to `Process::op_expacc`.
+    #[inline(always)]
     pub fn op_expacc(&mut self) {
         let old_base = self.stack_get(1);
         let old_acc = self.stack_get(2);
@@ -123,6 +144,7 @@ impl FastProcessor {
     /// where c0 = b0 * a0 - 2 * b1 * a1 and c1 = (b0 + b1) * (a1 + a0) - b0 * a0. It pushes 0 to
     /// the first and second positions on the stack, c1 and c2 to the third and fourth positions,
     /// and leaves the rest of the stack unchanged.
+    #[inline(always)]
     pub fn op_ext2mul(&mut self) {
         let [a0, a1, b0, b1] = self.stack_get_word(0).into();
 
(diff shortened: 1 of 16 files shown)


processor/src/fast/field_ops.rs
```diff
@@ -77,11 +95,12 @@ impl FastProcessor {
     }
 
     /// Analogous to `Process::op_eq`.
-    pub fn op_eq(&mut self) -> Result<(), ExecutionError> {
-        self.pop2_applyfn_push(|a, b| if a == b { Ok(ONE) } else { Ok(ZERO) })
+    pub fn op_eq(&mut self, tracer: &mut impl Tracer) -> Result<(), ExecutionError> {
```
Al-Kindi-0: Is it missing `#[inline(always)]`

processor/src/fast/fri_ops.rs
```diff
@@ -20,7 +20,7 @@ const TAU3_INV: Felt = Felt::new(281474976710656); // tau^{-3}
 
 impl FastProcessor {
     /// Analogous to `Process::op_fri_ext2fold4`.
-    pub fn op_fri_ext2fold4(&mut self) -> Result<(), ExecutionError> {
+    pub fn op_fri_ext2fold4(&mut self, tracer: &mut impl Tracer) -> Result<(), ExecutionError> {
```
Al-Kindi-0: Is it missing `#[inline(always)]`

processor/src/fast/mod.rs
```diff
@@ -1791,62 +1639,11 @@ impl FastProcessor {
...
-                stack_top,
-                continuation_stack.clone(),
-                execution_state,
-                current_forest.clone(),
-            );
-        }
-    }
-
     // TESTING
     // ----------------------------------------------------------------------------------------------
```
Al-Kindi-0, line 1649: Maybe add a comment to `execute_for_trace_sync`

processor/src/fast/trace_state_builder.rs
```diff
@@ -172,6 +267,151 @@ impl CoreTraceStateBuilder {
...
+        if processor.clk.as_usize().is_multiple_of(NUM_ROWS_PER_CORE_FRAGMENT) {
+            self.start_new_trace_state(
+                SystemState {
+                    clk: processor.clk,
+                    ctx: processor.ctx,
+                    fmp: processor.fmp,
+                    in_syscall: processor.in_syscall,
+                    fn_hash: processor.caller_hash,
+                },
+                processor.stack_top().try_into().unwrap(),
```
Al-Kindi-0: Can we avoid the unwrap?

processor/src/fast/tracer.rs
```diff
@@ -0,0 +1,189 @@
...
+    // -----------------------------------------------
+
+    /// Signals that the processor clock is being incremented.
+    fn increment_clk(&mut self);
+
+    /// Signals that the stack depth is incremented as a result of pushing a new element.
+    fn increment_stack_size(&mut self, processor: &FastProcessor);
+
+    /// Signals that the stack is depth is decremented as a result of popping an element off the
+    /// stack.
```
Al-Kindi-0: the stack is => the stack's

Al-Kindi-0: or as we usually say "stack depth"

Al-Kindi-0 (approved):
LGTM!

processor/src/fast/field_ops.rs
plafer: fixed

processor/src/fast/fri_ops.rs
plafer: Fixed (and all other functions in file)

processor/src/fast/mod.rs
plafer: fixed

processor/src/fast/trace_state_builder.rs
plafer: Switched to an `expect()` - which AFAIK is unavoidable without `unsafe` code

processor/src/fast/tracer.rs
plafer: Fixed (to "stack depth")

huitseeker (approved):
This will work, though I'm worried about the threading of context arguments in the processor (error context, tracer ...).

plafer:
> This will work, though I'm worried about the threading of context arguments in the processor (error context, tracer ...).

Note that `ErrorContext` will be going away with #1978, so only the `Tracer` will be left. 

The alternative is to put the `Tracer` as a field of the `FastProcessor`, which we could investigate. Basically you would specify the type of execution you want at construction time (where we would store the proper `Tracer` for the execution type), and then provide just one `execute()` that would use the internal tracer. The downside though is that `FastProcessor` would become generic, which has caused headaches in the past (e.g. with `ProcessState`).

adr1anh:
> The downside though is that `FastProcessor` would become generic, which has caused headaches in the past (e.g. with `ProcessState`).

This would go away when the slow processor is removed?

processor/src/fast/trace_state_builder.rs
```diff
@@ -64,42 +72,138 @@ pub struct CoreTraceStateBuilder {
...
+                    // Note: the stack depth to record is the `current_stack_depth - 1` due to the
+                    // semantics of DYNCALL. That is, the top of the stack contains the memory
+                    // address to where the address to dynamically call is
+                    // located. Then, the DYNCALL operation performs a drop, and
+                    // records the stack depth after the drop as the beginning of
+                    // the new context. For more information, look at the docs for how the
+                    // constraints are designed; it's a bit tricky but it works.
+                    let stack_depth_after_drop = processor.stack_depth() - 1;
+                    (
+                        Some(ExecutionContextInfo::new(
```
adr1anh: Small nit: can we factor out the `ExecutionContextInfo`?

adr1anh (approved):
Looks good!

plafer:
> This would go away when the slow processor is removed?

Not really - you would still need to make `ProcessState` generic over the `Tracer`, and therefore everything that touches a `ProcessState` is as well. Not a deal breaker necessarily, but I currently prefer passing the `Tracer` around over making `FastProcessor` generic over `Tracer` (or at least until we find a good reason to switch).

processor/src/fast/trace_state_builder.rs
plafer: Done

plafer:
Will merge this one into #2023 since it is a clear improvement.

PR #2103, opened 2025-08-28
feat: Implement keccak handler

adr1anh:
## Describe your changes

This PR introduces a Keccak256 precompile system for the Miden standard library. The precompile consists of a Rust event handler that reads preimages from memory, computes Keccak256 hashes, and stores results on the advice stack along with cryptographic commitments in the advice map. A MASM wrapper provides the user interface and handles RPO-based commitment generation for both the preimage and resulting hash. The system includes comprehensive testing that validates memory reads, advice stack contents, and commitment correctness. Future work will add kernel communication to signal precompile usage.

To support this architecture, the PR migrates event handler IDs from u32 to Felt throughout the system (internally mapped to u64 for comparisons). This change affects the testing framework, default host interface, and event registry. Additionally, the PR demonstrates a testing methodology where event handlers can be added to perform custom assertions during MASM code execution, which may be useful for other testing scenarios beyond this precompile.

*TLDR:*
- Implement `hash_mem` and `merge_stack` handlers with accompanying masm wrappers
- Deprecate old keccak implementations
- Refactor some event handling code to support adding handlers to standard library

- Updated `CHANGELOG.md'


stdlib/src/precompiles/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+        return Err(());
+    }
+
+    let len: u32 = len.try_into().map_err(|_| ())?;
+
+    let end = ptr.checked_add(len).ok_or(())?;
+
+    let ctx = process.ctx();
+
+    (ptr..end).map(|addr| process.get_mem_value(ctx, addr).ok_or(())).collect()
```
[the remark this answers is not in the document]
adr1anh: Good observation, though I think it's simpler to return an `Option`, since the precompile can return the range of memory which could not be read.

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#!
+#! Final stack state:
+#! [commitment, ...]
+#!
+#! Where:
+#! - ptr_in:     memory address where input bytes start
+#! - len:        number of bytes to hash
+#! - ptr_out:    memory address where to write the 32-byte hash
+#! - commitment: RPO hash of [RPO(input), RPO(keccak256(input))]
+export.keccak256_precompile
```
bobbinth: I'm wondering if we need the distinction between the "precompile" and not "precompile" version. Basically, I'm thinking maybe this module should just replace the existing `keccak256` module as eventually there would be no reason to use that module instead of this one. (and if we do want to keep the old code for reference, we could rename it to something like `keccak256_deprecated`)

Also, I wonder if we should follow somewhat the general approach of the `rpo` module here as well. Basically, we could have a handful of functions that cater to specific use cases. For example:

- `std::crypto::hashes::keccak::hash` - a "1-to-1" hash: takes 8 elements from the stack (assumes each element contains a 32-bit value), and computes their hash.
- `std::crypto::hashes::keccak::merge` - a "2-to-1" hash: takes 16 elements from the stack and hashes them.
- `std::crypto::hashes::keccak::hash_memory` - similar interface in `std::crypto::hashes::rpo::hash_memory` (which is slightly different from the current interface).

Under the hood, these will use the same precompile - just pre-processing of inputs would be a bit different.

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+    # Emit keccak precompile event - this triggers the event handler which:
+    # 1. Reads input bytes from memory [ptr_in, ptr_in+len)
+    # 2. Computes keccak256 hash
+    # 3. Pushes hash bytes to advice stack
+    # 4. Stores commitment in advice map for prover recovery
+    # => [ptr_in, len, ptr_out, ...]
+    emit.event("miden_stdlib::hash::keccak")
+
+    # Compute RPO hash of the input memory range
+    exec.rpo::hash_memory
```
bobbinth: We should probably check that the input address is word-aligned before calling this (I believe it will be checked inside `rpo::hash_memory` as well, but we could return a nicer error from here).

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+    end
+    exec.rpo::squeeze_digest
+    # => [RPO(keccak_hash), ptr_out, RPO(input), ...]
+    movup.4 drop
+    # => [RPO(keccak_hash), RPO(input), ...]
+
+    # Merge the two RPO hashes to create the final commitment
+    # This commitment serves as fingerprint of the precompile computation.
+    hmerge
+    # => [commitment, ...] where commitment = RPO([RPO(input), RPO(keccak_hash)])
```
bobbinth: I think the output here should be just Keccak hash (on the stack). In the next iteration, we'll need to process the RPO commitment too - but I don't think it will be returned to the user (IIRC, it would be added to the "registry" of deferred computations).

stdlib/src/precompiles/mod.rs
bobbinth, line 1: nit: I would maybe call this module "handlers" (i.e., `src/handlers/mod.rs`) as it will probably contain handlers for u64 division, smt events etc. in the future.

stdlib/src/precompiles/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+pub mod keccak;
+
+// # HELPERS
+// =================================================================================================
+
+/// Reads a contiguous region of memory from the VM process state.
+///
+/// This function returns `None` if the provided range is invalid or if it references an
+/// uninitialized address.
+fn read_memory(process: &ProcessState, ptr: u64, len: u64) -> Option<Vec<Felt>> {
```
bobbinth: I wonder if we should also enforce here that `ptr` is word-aligned.

stdlib/src/precompiles/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+    let preimage = witness
+        .iter()
+        .enumerate()
+        .map(|(idx, felt)| {
+            u8::try_from(felt.as_int()).map_err(|_| KeccakError::InvalidByteValue {
+                value: felt.as_int(),
+                byte_index: idx,
+            })
+        })
+        .collect::<Result<Vec<_>, _>>()?;
```
bobbinth: Why do we assume that each element contains just 1 byte? I think we can safely assume that each element has 4 bytes. Or does this create some problems?

bobbinth (commented):
Looks great! Thank you! Not a full review yet (I mostly looked at `miden-stdlib` crate for now), but I left some comments inline.

stdlib/asm/crypto/hashes/keccak_precompile.masm
adr1anh: > I'm wondering if we need the distinction between the "precompile" and not "precompile" version. Basically, I'm thinking maybe this module should just replace the existing `keccak256` module as eventually there would be no reason to use that module instead of this one. (and if we do want to keep the old code for reference, we could rename it to something like `keccak256_deprecated`)

While I agree it would be mostly irrelevant, there is one use case where we might still want to use it: when it's cheaper to prove keccak in MASM than verifying a precompile proof of it. The tradeoff is when the cost of recursion (~32k cycles ATM) is higher than the native keccak computation. 

It's a pretty niche situation though and may not be worth the effort to support it.

stdlib/asm/crypto/hashes/keccak_precompile.masm
adr1anh: Good point, it's at least worth a mention.

stdlib/asm/crypto/hashes/keccak_precompile.masm
adr1anh: Totally agreed, the commitment isn't useful to the user, but it's useful for debugging. The solution here could be to factor the implementation into smaller chunks, which should also help with exporting all the different hashing modes you suggested.

Regarding the placement of the hash on the stack, is this usually more practical than having it in memory? We could support both modes depending on the situation. The only constraint is that we need to RPO-hash the keccak-hash while popping the latter from the advice stack.

stdlib/src/precompiles/keccak.rs
adr1anh: Using `u32` should work fine, I think using bytes was pretty arbitrary.

stdlib/asm/crypto/hashes/keccak_precompile.masm
bobbinth: I thought the native Keccak computation currently takes something like 120K cycles (and so should always be worse then the precompile). Or am I remembering this incorrectly?

adr1anh: > I thought the native Keccak computation currently takes something like 120K cycles (and so should always be worse then the precompile). Or am I remembering this incorrectly?

In that case, we should definitely deprecate the native one.

adr1anh: Created #2117

adr1anh: > Also, I wonder if we should follow somewhat the general approach of the `rpo` module here as well. Basically, we could have a handful of functions that cater to specific use cases. For example:
> 
> * `std::crypto::hashes::keccak::hash` - a "1-to-1" hash: takes 8 elements from the stack (assumes each element contains a 32-bit value), and computes their hash.
> * `std::crypto::hashes::keccak::merge` - a "2-to-1" hash: takes 16 elements from the stack and hashes them.
> * `std::crypto::hashes::keccak::hash_memory` - similar interface in `std::crypto::hashes::rpo::hash_memory` (which is slightly different from the current interface).
> 
> Under the hood, these will use the same precompile - just pre-processing of inputs would be a bit different.

Created #2118

bobbinth: I would maybe do https://github.com/0xMiden/miden-vm/issues/2117 in this PR (unless it is more than just simple renaming).

Regarding https://github.com/0xMiden/miden-vm/issues/2118, we'd implement `hash_memory` in this PR and then https://github.com/0xMiden/miden-vm/issues/2118 would add `hash` and `merge` procedures, right?

adr1anh: Makes sense! I'll give it a shot and examine whether it's worth factoring into smaller chunks.

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#! This module provides a wrapper function around the keccak precompile event handler,
+#! allowing efficient keccak256 computation by deferring to the native verifier.
+
+use.std::mem
+use.std::crypto::hashes::rpo
+
+# Constants
+# ================================================================================================
+
+
```
Al-Kindi-0: nit: I would remove one newline

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#! Size of keccak256 hash output in bytes
+const.KECCAK_HASH_SIZE=32
+
+#! Compute keccak256 hash using precompile event handler.
```
Al-Kindi-0: nit: I would add a section separator between these

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#!
+#! This function takes a memory slice as input, computes its keccak256 hash using the
+#! precompile event handler, writes the hash to the specified output location, and
+#! returns a cryptographic commitment to the precompile call on the stack.
```
Al-Kindi-0: nit: I would define what is meant by a slice (i.e., a pointer and a length). I would also mention that pointers need to be word aligned

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#!
+#! The commitment is computed as RPO([RPO(input), RPO(keccak256(input))]) and can be
+#! used for proof verification and precompile call validation.
```
Al-Kindi-0: Question: Does `precompile call validation` mean debugging? If so I would do two things:

1. Define the concept of deferred computations (somewhere),
2. Change this comment to say something like `...proof verification of deferred Keccak precompile computations`
3. Change this comment to say something like `...precompile call validation (for debuggin purposes)`

stdlib/asm/crypto/hashes/keccak_precompile.masm
```diff
@@ -0,0 +1,70 @@
...
+#! [ptr_in, len, ptr_out, ...]
+#!
+#! Final stack state:
+#! [commitment, ...]
+#!
+#! Where:
+#! - ptr_in:     memory address where input bytes start
+#! - len:        number of bytes to hash
+#! - ptr_out:    memory address where to write the 32-byte hash
+#! - commitment: RPO hash of [RPO(input), RPO(keccak256(input))]
```
Al-Kindi-0: nit: we can try to follow the conventions for this used throughout the code bases

stdlib/asm/crypto/hashes/keccak_precompile.masm
Al-Kindi-0: I would probably just create a wrapper for now that drops `commitment` and loads the Keccak digest onto the operand stack.
Later on we can create different "real" wrappers that do something useful with `commitment`

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum KeccakError {
+    #[error("failed to read memory at ptr {ptr}, len {len}")]
+    MemoryReadFailed { ptr: u64, len: u64 },
+    #[error("invalid byte at index {byte_index} with value: {value}")]
+    InvalidByteValue { value: u64, byte_index: usize },
+}
+
+/// Event handler which pushes keccak256 hash to the advice stack.
```
Al-Kindi-0: nit: should we use throughout Keccak256 or keccak256?

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// This handler is used to defer expensive keccak computation to the native verifier.
```
Al-Kindi-0: nit: I think here Keccak is the right call?

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+/// The verifier will re-execute the keccak operation and verify the result.
```
Al-Kindi-0: nit: I would add a bit more e.g., the result is verified outside the VM and this verification passing together with that of the proof guarantee the correctness of the execution of the program that made these deferred computations

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+/// Inputs:
+///   Operand stack: [event_id, ptr, len, ...]
+///   Memory: bytes at [ptr, len)
+///
+/// Outputs:
+///   Advice stack: [hash_bytes...] (32 bytes as individual Felts)
+///   Advice map: commitment -> preimage mapping for prover recovery
+///
+/// Where:
+/// - ptr is the memory address where the input bytes start
```
Al-Kindi-0: I would again mention the restriction regarding word-alignment

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,82 @@
...
+    ]);
+
+    // Reverse the keccak hash before extending the advice stack, so that
+    // the first popped element is the first byte of the hash.
+    keccak_hash_felt.reverse();
+    let advice_stack_extension = AdviceMutation::extend_stack(keccak_hash_felt);
+
+    // store the calldata in the advice map to be recovered later on by the prover
+    let entry = (calldata_commitment, witness);
+    let advice_map_extension = AdviceMutation::extend_map(AdviceMap::from_iter([entry]));
```
Al-Kindi-0: Maybe it is not the right time nor place for this, but maybe we should also add the event id as the first element of the value associated to the key `calldata_commitment`.
Though I think this will become clearer once we start working on the kernel side of things

stdlib/src/handlers/mod.rs
```diff
@@ -0,0 +1,26 @@
+use alloc::vec::Vec;
+
+use miden_core::Felt;
+use miden_processor::ProcessState;
+
+pub mod keccak;
+
+// # HELPERS
```
Al-Kindi-0: nit: would remove # to stay consistent

stdlib/src/handlers/keccak.rs
adr1anh: I'm starting to think we should add an extra `deferred: Vec<(Tag, Calldata)>` field to the advice provider to keep track of the precompile calls. The tag would then represent which precompile was called (either the event ID or something similar). I'm starting to think we should add an extra `deferred: Vec<(Tag, Calldata)>` field to the advice provider to keep track of the precompile calls. The tag would then represent which precompile was called (either the event ID or something similar).

stdlib/src/precompiles/keccak.rs
adr1anh: Actually, this might be a bit more involved than I thought since we'll have to include the length in bytes of the input. It's not 100% clear to me what would be the best way to handle this. For now, I'll leave the preimage as bytes but return the hash as `u32`s.

stdlib/src/handlers/keccak.rs
Al-Kindi-0: Could be a valid approach as well. Repurposing the advice map as it is currently implemented to achieve our objectives seems a bit problematic

processor/src/host/handlers.rs
```diff
@@ -36,6 +37,15 @@ where
     }
 }
 
+/// A handler which ignores the process state and leaves the `AdviceProvide` unchanged.
```
```suggestion
/// A handler which ignores the process state and leaves the `AdviceProvider` unchanged.
```

processor/src/host/handlers.rs
```diff
@@ -36,6 +37,15 @@ where
...
+pub struct TrivialEventHandler;
```
plafer: Nit: I started using `Noop*` for things that don't do anything - e.g. `NoopEventHandler`. In the name of consistency, we should try to adopt a single convention. I personally find `Noop*` more descriptive than `Trivial*` - what do you think?

processor/src/host/handlers.rs
```diff
@@ -91,21 +101,21 @@ pub type EventError = Box<dyn Error + Send + Sync + 'static>;
 /// ```
 #[derive(Default)]
 pub struct EventHandlerRegistry {
-    handlers: BTreeMap<u32, Box<dyn EventHandler>>,
+    handlers: BTreeMap<u64, Arc<dyn EventHandler>>,
```
plafer, line 104: As discussed [here](https://github.com/0xMiden/miden-vm/issues/2112#issuecomment-3245381608), we could define a new `EventId` type that wraps `Felt`

processor/src/host/default.rs
```diff
@@ -80,21 +80,21 @@ where
...
-        self.event_handlers.register(id, Box::new(handler))
+        self.event_handlers.register(id, handler)
     }
 
     /// Unload a handler with the given id, returning a flag indicating whether a handler
     /// was previously registered with this id.
-    pub fn unload_handler(&mut self, id: u32) -> bool {
+    pub fn unload_handler(&mut self, id: Felt) -> bool {
         self.event_handlers.unregister(id)
     }
```
plafer: nit: should we stick with the `register`/`unregister` convention, instead of using both that and `load/unload`?

plafer (commented):
Quick review on my end with mainly nits

processor/src/host/default.rs
adr1anh: Agreed, I went with `register` since it seems more appropriate.

processor/src/host/handlers.rs
adr1anh: Will leave this as follow-up.

adr1anh:
The latest commit implements support for merging two digests on the stack. This was a lot more involved than expected due to some stack/word ordering problems. From now, I'll finish this off by 
- deprecating the old keccak implementation
- creating wrappers that drop the call_data commitment

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - [BREAKING] `get_stack_word` uses element-aligned indexing instead of word-aligned indexing (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - [BREAKING] Implement support for `event("event_name")` in MASM (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - Improved representation of `OPbatches` to include padding Noop by default, simplifying fast iteration over program instructions in the processor ([#1815](https://github.com/0xMiden/miden-vm/issues/1815)).
+- Initial support for Keccak precompile ([#2103](https://github.com/0xMiden/miden-vm/pull/2103))
```
bobbinth: Question: is this a breaking change? Or did all external interfaces remain the same?

stdlib/docs/crypto/hashes/keccak256.md
bobbinth, line 1: In addition to this file, we should also update [this file](https://github.com/0xMiden/miden-vm/blob/next/docs/src/user_docs/stdlib/crypto/hashes.md) so that Keccak256 description appears in the [mdBook](https://github.com/0xMiden/miden-vm/blob/next/docs/src/user_docs/stdlib/crypto/hashes.md) docs. There we can probably skip some of the internal procedures.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-    movup.2
-    u32and
-    swap
+# Constants
+# ================================================================================================
 
-    movup.3
-    movup.3
+const.KECCAK_MEM_EVENT=event("miden_stdlib::hash::keccak::hash_mem")
+const.KECCAK_STACK_EVENT=event("miden_stdlib::hash::keccak::merge_stack")
```
bobbinth: We are using two separate events for optimization purposes, right? My original thinking was that we'd only have one event (i.e., just the `KECCAK_MEM_EVENT`) and then the other keccak procedures would just write inputs to memory and then invoke the procedure that computes keccak hash from memory. I think if the incremental cycle count is within 20 cycles or so - it's probably fine to have a single event.

Also, I wonder if the event should be named `std::crypto::hash::keccak256::compute_hash` - basically the module name followed by the event name.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-
-#! Keccak-p[1600, 24] permutation's ι ( iota ) function, which is
-#! implemented in terms of 32 -bit word size ( bit interleaved form );
-#! imagine https://github.com/itzmeanjan/merklize-sha/blob/1d35aae9da7fed20127489f362b4bc93242a516c/include/sha3.hpp#L288-L306
-#! invoked with (c0, c1) as template arguments
+#! Computes Keccak256 hash of memory data.
 #!
-#! Expected stack state :
+#! Input: [ptr, len_bytes, ...]
+#! Output: [keccak_lo, keccak_hi, ...]
```
bobbinth: nit: by convention, we write word-size variables in upper case. So, this would be something like:
```
#! Inputs:  [ptr, len_bytes]
#! Outputs: [KECCAK_LO, KECCAK_HI]
```
Also, the current convention is that words on the stack in the "big-endian" order - though, this is likely to change. My understanding is that both `KECCAK_LO` and `KECCAK_HI` are in "little-endian". If so, we should clarify this in the docs.

We should also make it clear that each element contains a 32-bit value.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-
-    movup.5
-    u32xor
-
-    swap
+#! Where:
+#! - ptr: starting memory address (must be word-aligned)
+#! - len_bytes: number of bytes to hash
+#! - keccak_lo, keccak_hi: 256-bit digest as two words
+export.hash_mem
```
bobbinth: I'd name this procedure `hash_memory` to keep the naming consistent with the `std::crypto::hashes::rpo` module.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-#! After application of bit interleaving, stack looks like
+#! Internal implementation of memory-based Keccak256 computation via deferred computation.
 #!
-#! [even, odd, ...]
+#! Emits event to populate advice stack with digest, computes commitment for tracking
+#! deferred computations, and stores witness in advice map for later proof generation.
 #!
-#! Read more about bit interleaved representation in section 2.1 of https://keccak.team/files/Keccak-implementation-3.2.pdf
+#! Input: [ptr, len_bytes, ...]
+#! Output: [commitment, keccak_lo, keccak_hi, ...]
```
bobbinth: Same comment re input/output doc formatting as in one of the previous comments.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-        # ---
-
-        dup.3
-        dup.3
+#! Where:
+#! - ptr: starting memory address (must be word-aligned)
+#! - len_bytes: number of bytes to hash
+#! - commitment: RPO(RPO(input) || RPO(hash)) for kernel tracking
+#! - keccak_lo, keccak_hi: 256-bit digest as two words
+export.hash_mem_impl
```
bobbinth: Is this `export` rather than `proc` for testing purposes?

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-#! [even, odd, ...]
-#!
-#! After application of logic, stack looks like
-#!
-#! [hi, lo, ...]
+#! Merges two 256-bit digests via Keccak256 hash.
 #!
-#! This function reverts the action done by `to_bit_interleaved` function implemented above.
+#! Input: [digest_left_lo, digest_left_hi, digest_right_lo, digest_right_hi, ...]
+#! Output: [keccak_lo, keccak_hi, ...]
```
bobbinth: Same comment re input/output formatting and explanations as above.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -1,3503 +1,127 @@
...
-
-        movup.2
-        u32xor
-        swap
-
-        # ---
+#! Where:
+#! - digest_left_lo/hi, digest_right_lo/hi: two digests to merge (as four words)
+#! - keccak_lo, keccak_hi: Keccak256(left || right) as two words
+export.merge_stack
```
bobbinth: As mentioned in one of the above comments, this procedure could probably just write the stack to memory and then invoke `hash_memory` procedure. Since the inputs on the stack are in little-endian order, this would require some amount of manipulation.

Alternatively, we could put inputs in big-endian order, and then saving to memory should be pretty straight-forward.

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,222 @@
...
+use miden_core::{AdviceMap, Felt, FieldElement, Word, crypto::hash::Digest};
+use miden_crypto::hash::{keccak::Keccak256, rpo::Rpo256};
+use miden_processor::{AdviceMutation, EventError, ProcessState};
+
+use crate::handlers::read_memory;
+
+/// Event ID for memory-based Keccak256 computation
+pub const KECCAK_HASH_MEM_EVENT_ID: &str = "miden_stdlib::hash::keccak::hash_mem";
+/// Event ID for stack-based Keccak256 merge computation
+pub const KECCAK_MERGE_STACK_EVENT_ID: &str = "miden_stdlib::hash::keccak::merge_stack";
```
bobbinth: I would rename these into `KECCAK_HASH_MEM_EVENT_NAME` and `KECCAK_MERGE_STACK_EVENT_NAME`. 

Would also be great to introduce `KECCAK_MERGE_STACK_EVENT_ID` and `KECCAK_MERGE_STACK_EVENT_ID` which would be felts. We can hard-code them to the actual IDs and then have a tests that ensures that names and IDs are consistent (by using `string_to_event_id()` function in the tests).

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Keccak256 event handler that merges two digests.
+///
+/// - Input: Reads two 256-bit digests from stack as four words
+/// - Output: Returns Keccak256(left || right) via advice stack, stores witness in advice map
+///
+/// Stack: [event_id, digest_left_lo, digest_left_hi, digest_right_lo, digest_right_hi, ...]
+pub fn handle_keccak_merge_stack(
+    process: &ProcessState,
+) -> Result<Vec<AdviceMutation>, EventError> {
```
bobbinth: As mentioned one of the previous comments, I think we can probably avoid having two handlers.

bobbinth (commented):
Looks good! Thank you! I left some comments inline - the main ones are about reducing the number of keccak handlers from two to one (i.e., just the one that operates with memory).

CHANGELOG.md
adr1anh: Good catch! With the deprecation of the old keccak implementation as well as the changes to the event handling infra, this is definitely a breaking change.

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: Performance was not the main issue here, but rather the awkwardness of having to pass a pointer to a temporary memory region. This issue is related to the problem we face with the kernel keeping track of all precompile calls, since we need a "clean" address that is ideally isolated from whatever else the user is doing. Though here things are easier here since we can just ask the caller for an unused region.

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: I'll add some docs to make this clearer, but for context
For a 32-byte digest in LE order, we have `bytes = [b_0, ..., b_{31}]`.
We then pack each chunk of 4 bytes into one `u32`: `packed = [h_0, ..., h_7]` and `h_i = u32::from_le_bytes(bytes[4*i..4*(i-1)])`
We then cast `packed` into two words `[H_LO, H_HI] = [h_0, ..., h_3], [h_4, ..., h_7]`.
On the stack though, the digest we output on the stack is
`=> [H_LO, H_HI] = [[h_3, ..., h_0], [h_7, ..., h_4]` which ensures that we can efficiently write the words sequentially using `mem_storew` without having to change the order of the words.

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: Yes, I'm not sure if there's another better way though.

stdlib/src/handlers/keccak.rs
adr1anh: Fixed in latest commit

stdlib/src/handlers/keccak.rs
adr1anh: Fixed in latest commit

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: Fixed in latest commit. I explained more throughly the expected ordering.

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: Fixed in latest commit

adr1anh:
I've updated the relevant documentation as well as removed the two handlers.

stdlib/asm/crypto/hashes/keccak256.masm
bobbinth: The way to do it is to allocate procedure locals. This should work because we know how much memory we need in each case. For example, `hash_1to1` could look something like this:
```
#! Input: [INPUT_LO, INPUT_HI, ...]
#! Output: [KECCAK_LO, KECCAK_HI, ...]
export.hash_1to1.8
    # get the absolute address of the first element in the procedure's local memory
    locaddr.0
    # => [ptr, INPUT_LO, INPUT_HI]

    dup movdn.5 mem_storew dropw add.4
    # => [ptr+4, INPUT_HI]

    dup movdn.5 mem_storew dropw
    # => [ptr+4]

    sub.4 push.32 swap
    # => [ptr, 32]

    exec.hash_memory
end
```

docs/src/user_docs/stdlib/main.md
```diff
@@ -24,13 +24,14 @@ For an example of how to invoke procedures from imported modules see [this secti
...
+| [std::crypto::hashes::sha256](./crypto/hashes.md#sha256)            | Contains procedures for computing hashes using SHA256 hash function.                                                                                             |
+| [std::math::u64](./math/u64.md)                                     | Contains procedures for working with 64-bit unsigned integers.                                                                                                   |
+| [std::mem](./mem.md)                                                | Contains procedures for working with random access memory.                                                                                                       |
+| [std::sys](./sys.md)                                                | Contains system-level utility procedures.                                                                                                                        |
+| [std::word](./word.md)                                              | Contains utilities for working with words.                                                                                                                       |
```
bobbinth, line 37: Not for this PR, but I think this table needs to be updated. For example, we should mention `std::collections::smt` and `std::crypto::hashes::rpo` modules here as well (maybe some other modules too). Let's create an issue for this.

stdlib/src/handlers/keccak.rs
```diff
@@ -0,0 +1,247 @@
...
+//! ### Digest Representation
+//! A Keccak256 digest (256 bits) is represented as [`KeccakFeltDigest`]: 8 field elements,
+//! each containing a u32 value in little-endian order.
+//!
+//! #### Example Encoding
+//! For a 32-byte digest, the encoding process is:
+//! 1. **Bytes to u32s**: `[b0,b1,...,b31]` → `[h0,h1,h2,h3,h4,h5,h6,h7]` where each `hi` is a
+//!    little-endian u32 (e.g., `h0 = u32::from_le_bytes([b0,b1,b2,b3])`)
+//! 2. **Group into words**: `KECCAK_LO = [h0,h1,h2,h3]`, `KECCAK_HI = [h4,h5,h6,h7]`
+//! 3. **Stack representation**: `[[h3,h2,h1,h0], [h7,h6,h5,h4]]` (each word reversed for LIFO)
```
bobbinth: I am not sure about this "mixed" representation here. I think two potential alternatives are:

1. "Little-endian": `[[h0, h1, h2, h3], [h4, h5, h6, h7]]`.
2. "Big-endian": `[[h7, h6, h5, h4], [h3, h2, h1, h0]]`. 

The latter would be more consistent with how we do things now, but the former is more consistent with where we'd like to go in the future. The main downside of the former representation is a handful of extra cycles that we'd need to pay for saving data to memory - but I think that's probably not a big concern.

This issue is very much related to the discussion in https://github.com/0xMiden/miden-vm/pull/2120#discussion_r2329566753 and so cc @bitwalker in case he he has any comments.

docs/src/user_docs/stdlib/main.md
adr1anh: Created #2143

stdlib/src/handlers/keccak.rs
adr1anh: If I understand correctly, we want the precompiles to return with `[h_0, ..., h_7]` on the stack right?

bobbinth: Yes, I think so.

bobbinth: It may make sense to add procedures like `std::word::reverse` and `std::word::reverse_double` to help with stack manipulations.

adr1anh: Everything has been updated to match this ordering.

adr1anh:
Before merging, I need to confirm that the output of the wrappers is in the correct order.

EDIT: the latest commit fixes the output stack ordering. This issue came from the fact that `hmerge` computes `Rpo([d_7, .., d_0])` since the digest is popped onto the operand stack as `[d_0, ..., d_7]`. To fix this issue without additional reversing in the masm, the native digest commitment computation now reverses the elements before hashing.

stdlib/src/handlers/keccak256.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Creates an commitment of the digest using Rpo256.
+    ///
+    /// When the digest is popped from the advice stack, it appears as
+    /// `[d_0, ..., d_7]` on the operand stack. In masm, the `hmerge` operation computes
+    /// `Rpo256([d_7, ..., d_0])`, so we reverse the order here to match that behavior.
+    pub fn to_commitment(&self) -> Word {
+        let mut rev = self.0;
+        rev.reverse();
+        Rpo256::hash_elements(&rev)
+    }
```
bobbinth, line 175: I think reversing the order is fine for now, but depending on how the pre-compile circuit is implemented, we may need to perform hashing in the "normal order".

bobbinth (approved):
All looks good! Thank you!

PR #2111, opened 2025-08-31
perf(stdlib): optimize Vec allocation in FRI verifier

strmfos:
## Describe your changes

Use Vec::with_capacity in FRI proof verification to avoid reallocations.The final size is known to be positions.len() since each iteration pushes exactly one element to final_pos_eval vector.


adr1anh:
Thank you for your interest, but we don't accept one-line change PRs.

PR #2114, opened 2025-09-01
sorted_array/map fast lookup

gswirski:
- Updated `CHANGELOG.md'


adr1anh:
This seems like a really good idea as it's a perfect use case of non-determinism. We're currently working on better support for custom events/handlers -- a mechanism where the execution of the VM is interrupted to call a custom function that can mutate the AdviceProvider. The goal is to move most of the current SystemEvents to the standard library using this framework. This feature would be a good addition. 

I think it would make sense to merge this once #1582 is tackled, as this feature seems more appropriate for the standard library rather than a core VM operation.

gswirski:
@adr1anh this PR is a result of a comment here: https://github.com/0xMiden/miden-vm/pull/2048#discussion_r2261250290

I know @bobbinth wanted to land SMT access in September - this PR would be a prerequisite/an important optimization for that work. Depending on how fast we can move with #1582 , we could merge this first to unblock #2048 (once I add support for sorted key-values) and refactor later? What do you think?

adr1anh:
Totally fine with me, adding an extra event won't affect the migration to the `stdlib`. The one thing we'll want to sort out is the MASM syntax. Adding stdlib-specific syntax seems wrong, and we would start favoring `emit.U64_DIV` for example. We'll figure this out when the time comes, but for now just be aware that the syntax (and likely the MAST forest) will change once we move it to the stdlib.

gswirski:
Fair enough, thanks for the heads-up.

gswirski:
This is now ready for an initial review. I have a few outstanding questions:
- what is the policy for code duplication in MASM? `sorted_array::find_word` and `sorted_map::find_key` are extremely similar (they differ in stride value, find_key needs to do an additional alignment check) but I'm not sure if we prefer precise cycles optimization, or smaller code size
- strictly speaking, we only need `sorted_map` version for smt::get/set methods, but I thought `sorted_array` would be generally useful for others. Can remove it if we don't want to pollute stdlib too much (which would also solve the above dillema).
- the `start_ptr, end_ptr` in the result are there because they tend to be useful until the end of the algorithm. should we drop them before returning, or let callee decide if they want to clean them up?

I know **I need to add cycles count to instructions**, didn't want to spend too much time on it before we finalize the approach.

stdlib/tests/collections/sorted_map.rs
```diff
@@ -0,0 +1,31 @@
+use super::*;
+
+#[test]
+fn test_array_find_sorted_word() {
```
bobbinth: The name here should probably be `test_map_find_sorted_word()`.

stdlib/asm/collections/sorted_map.masm
```diff
@@ -0,0 +1,103 @@
...
+
+                # there was no match, the last map key must be smaller than KEY
+                exec.word::gt
+                assert.err="push_array_lowerbound invariant: last element not smaller than KEY"
+
+                push.0
+            else
+                # make sure maybe_str_ptr within array bounds
+                dup dup.6 u32gt assert.err="maybe_key_ptr < start_ptr"
+                dup dup.7 u32lt assert.err="maybe_key_ptr > end_ptr"
```
bobbinth: When using `u32gt`/`u32lt`, we need to be sure that the values we re comparing are in fact in `u32` range (otherwise, the behavior of these operations is undefined). We usually do this via the `u32assert2` instruction. So, these lines could look like:
```
dup dup.6 u32assert2 u32gt assert.err="maybe_key_ptr < start_ptr"
dup dup.7 u32assert2 u32lt assert.err="maybe_key_ptr > end_ptr"
```

stdlib/asm/collections/sorted_map.masm
```diff
@@ -0,0 +1,103 @@
+use.std::word
+
+# ===== ARRAY UTILITIES ============================================================================
+
+#! Finds a key in a sorted array of (key, value) word tuples.
+#!
+#! Input:  [KEY, start_ptr, end_ptr]
+#! Output: [is_key_found, key_ptr, start_ptr, end_ptr]
+export.find_key
```
bobbinth: I think the overall approach works, but I wonder if we can rely a bit more on on-determinism to reduce the number and depth of `if`-statements. For example, instead of just placing `maybe_key_ptr` onto the advice stack, the event handler could place `[maybe_key_ptr, is_key_found]` onto it. And then, the overall structure of this procedure could be something like:

```
export.find_key
    adv.push_map_lowerbound
    adv_push.2
    #=> [is_key_found, maybe_key_ptr, KEY, start_ptr, end_ptr]

    if.true
        # verify that key at maybe_key_ptr is in fact equals to KEY
        # and that start_ptr <= maybe_key_ptr <= end_ptr
    esle
        dup.6 dup.1 eq
        #=> [is_end_ptr, maybe_key_ptr, KEY, start_ptr, end_ptr]

        dup.6 dup.2 eq
        #=> [is_start_ptr, is_end_ptr, maybe_key_ptr, KEY, start_ptr, end_ptr]

        # we pre-compute both is_start_ptr and is_end_ptr here to reduce the depth of the
        # generated MAST

        if.true
            # the key was not found, and the maybe_key_ptr points to the start of the list
            # make sure that either the list is empty or the key at start_ptr > KEY
        else
            if.true
                # the key was not found, and the maybe_key_ptr points to the end of the list
                # make sure that the key at end_ptr < KEY
            else
                # the key was not found, and the maybe_key_ptr is neither the start nor the end
                # of the list
                # make sure that the key at maybe_key_ptr < KEY and at maybe_key_ptr+1 > KEY
            end
        end
    end
end
```

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+            .unwrap_or([ZERO, ZERO, ZERO, ZERO].into());
+
+        // Compare mid_word with key lexicographically
+        let cmp = {
+            let [l0, l1, l2, l3]: [Felt; 4] = mid_word.into();
+            let [r0, r1, r2, r3]: [Felt; 4] = key.into();
+            let mid_tuple = (l0.as_int(), l1.as_int(), l2.as_int(), l3.as_int());
+            let key_tuple = (r0.as_int(), r1.as_int(), r2.as_int(), r3.as_int());
+            mid_tuple.cmp(&key_tuple)
+        };
```
bobbinth: We have `LexicographicWord` in `miden-crypto` which should simplify this.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+
+    // If range is empty, result is end_ptr
+    if lo >= hi {
+        process.advice_provider_mut().push_stack(Felt::from(end_addr));
+        return Ok(());
+    }
+
+    let mut len = (hi - lo) / stride; // count of elements
+
+    while len > 0 {
```
bobbinth: I think the code below is using binary search, right? If so, a quick comment stating that would be helpful.

bobbinth (commented):
Looks good! Thank you! Not a full review yet, but I left some comments inline.

> what is the policy for code duplication in MASM? `sorted_array::find_word` and `sorted_map::find_key` are extremely similar (they differ in stride value, find_key needs to do an additional alignment check) but I'm not sure if we prefer precise cycles optimization, or smaller code size

I think given that these procedures are not too big, duplication is fine. But also, if there is away to find how to avoid it without complicating things too much, that, of course be better.

> strictly speaking, we only need `sorted_map` version for smt::get/set methods, but I thought `sorted_array` would be generally useful for others. Can remove it if we don't want to pollute stdlib too much (which would also solve the above dillema).

I think having both is good - though, I'm not too sure about naming. Both of these are technically arrays, they just different in the type of entries. So, maybe we should just have a single `sorted_array` module with two procedures. Though, I haven't thought about the naming of these procedures too much.

> the `start_ptr, end_ptr` in the result are there because they tend to be useful until the end of the algorithm. should we drop them before returning, or let callee decide if they want to clean them up?

No strong preference here. I think more recently we started dropping the inputs and then if the caller wants them, then can always `dup` them before invoking the procedure, but there are plenty of examples where we return the inputs as well.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+///
+/// Outputs:
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [key_ptr, ...]
+fn push_lowerbound_result(
+    process: &mut ProcessState,
+    stride: u32,
+    err_ctx: &impl ErrorContext,
+) -> Result<(), ExecutionError> {
+    assert_eq!(stride % 4, 0);
```
Al-Kindi-0: nit: would probably test against just 4 and 8, after probably introducing constants for both strides

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+
+    // Read inputs from the stack
+    let key = process.get_stack_word(1);
+    let (start_addr, end_addr) =
+        get_mem_addr_range(process, 5, 6).map_err(ExecutionError::MemoryError)?;
```
Al-Kindi-0: nit: would introduce, maybe local, constants for 5 and 6, or at least add a comment

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+/// mem[word_ptr] >= KEY in lexicographic order of words. If all words are < KEY, returns end_ptr.
+/// Supports sorted arrays (stride = 4) and sorted maps (stride = 8).
+///
+/// Inputs:
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [...]
+///
+/// Outputs:
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [key_ptr, ...]
```
Al-Kindi-0: Should we add a comment saying that the result is undefined if the sortedness assumption is not satisfied.

Also should we specify here and throughout that they are sorted in non-decreasing order

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
     Ok(())
 }
 
+/// Pushes onto the advice stack the first pointer in [start_ptr, end_ptr) such that
+/// mem[word_ptr] >= KEY in lexicographic order of words. If all words are < KEY, returns end_ptr.
```
Al-Kindi-0: Would also add the other case i.e.,  all words are greater than KEY

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+#! Input:  [KEY, start_ptr, end_ptr]
+#! Output: [is_key_found, key_ptr, start_ptr, end_ptr]
+export.find_word
+    # check if empty array (3 cycles)
+    dup.5 dup.5 eq
+    #=> [is_empty, KEY, start_ptr, end_ptr]
+
+    if.true
+        dropw dup push.0
+        #=> [is_key_found = 0, key_ptr = end_ptr, start_ptr, end_ptr]
```
Al-Kindi-0: Would write key_ptr = start_ptr, or at least add a comment

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+            else
+                # dereference maybe_key_ptr (8 cycles)
+                dup.4 padw movup.4 mem_loadw
+                #=> [DEREF, KEY, maybe_key_ptr, start_ptr, end_ptr]
+
+                # there was no match, the first array element must be larger than KEY
+                exec.word::lt
+                assert.err="push_array_lowerbound invariant: start_ptr not greater than KEY"
+                
+                push.0
```
Al-Kindi-0, line 85: would add the stack state after this op

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+            end
+        else
+            # check if pointing to end_ptr (3 cycles)
+            dup dup.7 eq
+            if.true
+                # fetch the last word in the array (exists, array not empty)
```
Al-Kindi-0: nit: I would write
```
# fetch the last word in the array
# Being in this branch of the outer if-statement, the array is guaranteed to be non-empty and hence the last word must exist
```
or something to the same effect

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+                dup movdn.5 sub.4 padw movup.4 mem_loadw
+                #=> [LAST_WORD, KEY, maybe_key_ptr, start_ptr, end_ptr]
+
+                # there was no match, the last array element must be smaller than KEY
+                exec.word::gt
+                assert.err="push_array_lowerbound invariant: last element not smaller than KEY"
+
+                push.0
```
Al-Kindi-0, line 106: nit: would add a stack description after this

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+            else
+                # make sure maybe_str_ptr within array bounds
+                dup dup.6 u32gt assert.err="maybe_key_ptr < start_ptr"
+                dup dup.7 u32lt assert.err="maybe_key_ptr > end_ptr"
```
Al-Kindi-0: nit: I would add a stack description after this, even if it seems redundant, it will help readers a bit

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+                else
+                    # dereference maybe_key_ptr (8 cycles)
+                    dup.4 padw movup.4 mem_loadw
+                    #=> [DEREF, KEY, maybe_key_ptr, start_ptr, end_ptr]
+
+                    # there was no match, the dereferenced element must be larger than KEY
+                    dupw.1 exec.word::gt
+                    assert.err="push_array_lowerbound invariant: *ptr not larger than KEY"
+
+                    # there was no match, the previous element must be smaller than KEY
```
Al-Kindi-0: nit: I would write

```
# there was no match, the word preceding MAYBE_KEY  must be smaller than KEY
```

or something to the same effect

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
...
+                    dup.4 sub.4 padw movup.4 mem_loadw
+                    exec.word::gt
+                    assert.err="push_array_lowerbound invariant: previous not smaller than KEY"
+
+                    push.0
```
Al-Kindi-0: nit: would also add the stack state after this line

stdlib/asm/collections/sorted_map.masm
Al-Kindi-0: Same comments as for the other procedure, since they are very similar.
Though I think both will change given the comment of @bobbinth , which I agree with

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+
+    // Binary search over word-aligned addresses in [start_addr, end_addr)
+    let mut lo = start_addr;
+    let hi = end_addr;
+
+    // If range is empty, result is end_ptr
+    if lo >= hi {
+        process.advice_provider_mut().push_stack(Felt::from(end_addr));
+        return Ok(());
+    }
```
PhilippGackstatter: Nit: We already validate `start_addr > end_addr` earlier, so `lo == hi` would be sufficient and clearer here, I think. We can never enter the `>` case, though it is more defensive to write it that way.

Nit: I'd probably use `start_addr` and `end_addr` for this comparison, and only introduce `lo` and `hi` after this statement.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -612,6 +614,108 @@ fn push_smtpeek_result(
...
+    let mut len = (hi - lo) / stride; // count of elements
+
+    while len > 0 {
+        let half = len / 2;
+        let mid_addr = lo + half * stride;
+        // Read word at mid_addr
+        let mid_word = process
+            .get_mem_word(process.ctx(), mid_addr)
+            .map_err(ExecutionError::MemoryError)?
+            .unwrap_or([ZERO, ZERO, ZERO, ZERO].into());
```
PhilippGackstatter:
```suggestion
        let mid_word = process
            .get_mem_word(process.ctx(), mid_addr)
            .map_err(ExecutionError::MemoryError)?
            .unwrap_or(Word::empty());
```

Nit: `Word::empty` is a bit more concise.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,100 @@
+use.std::word
+
+# ===== ARRAY UTILITIES ============================================================================
+
+#! Finds a key in a sorted array of words.
+#!
+#! Input:  [KEY, start_ptr, end_ptr]
+#! Output: [is_key_found, key_ptr, start_ptr, end_ptr]
+export.find_word
```
PhilippGackstatter: Nit: Since this is an array, I think the input may be more accurately described as a `WORD` or `VALUE` rather than `KEY`? For the array case, maybe naming this just `find` is sufficient.

processor/src/operations/sys_ops/sys_event_handlers.rs
gswirski: The other case (all words are greater than KEY) is already covered by the comment - the first pointer (start_ptr) satisfies `mem[word_ptr] >= KEY`.

stdlib/asm/collections/sorted_map.masm
gswirski: Interesting, I have not thought about balancing the MAST or not mixing two MAST nodes within an if branch.

I've implemented your suggestion. I think the total number of `if` branches is the same, but MAST depth is indeed lower.

stdlib/asm/collections/sorted_array.masm
gswirski: Agreed on the `KEY` -> `VALUE`  change. Fixed.

As for `find_word` vs `find`, I'd generally prefer to have a "word" _somewhere_ in the fully-qualified name. With functions for words and kv pairs now living in a single `sorted_array` module, I think it makes sense to have them named `find_word` and `find_key_value`.

stdlib/asm/collections/sorted_array.masm
gswirski: I now changed it to

```# there was no match, MAYBE_VALUE_PREV must be smaller than VALUE```

to make it consistent whenever with other memory loads. What do you think?

**Comment by gswirski** [comment]
Addressed most of the feedback, some documentation nits (and cycle counting) left until we agree on the general structure.

I renamed functions to `sorted_array::find_word` and `sorted_array::find_key_value` - let me know what you think of these names. I keep them as separate methods because `find_word` can skip alignment checks in two places, and handling additional `stride` element on the stack is more annoying than helpful.

**One thing I'm not sure about is the binary search.** I suspect we're not going to deal with large buffers in MASM. We could replace binary search with linear search to detect unsorted arrays passed to push_xxx_lowerbound and avoid a potential footgun.

File: stdlib/asm/collections/sorted_array.masm (hunk 19 above)
Comment (PhilippGackstatter) [reply stdlib/asm/collections/sorted_array.masm]: I'd maybe use `find_value` to match the input variable name `VALUE`, and be analogous to `find_key_value` with input `KEY`, but not a strong opinion.

File: stdlib/asm/collections/sorted_array.masm [hunk 20]
```diff
@@ -0,0 +1,230 @@
...
+                assert.err="push_map_lowerbound invariant: last key-value pair not smaller than KEY"
+
+                push.0
+                #=> [is_key_found = 0, maybe_key_ptr = end_ptr, start_ptr, end_ptr]
+            else
+                # The key was not found, and the maybe_key_ptr is neither the start nor the end
+                # of the list. Make sure that the key at maybe_key_ptr-1 < KEY and at maybe_key_ptr > KEY
+                #=> [maybe_key_ptr, KEY, start_ptr, end_ptr]
+
+                # make sure maybe_str_ptr is pointing to a key, not value
```
adr1anh: Did you mean `maybe_key_ptr`?

gswirski: good catch, fixed.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,230 @@
...
+        end
+    end
+end
+
+# ===== MAP UTILITIES ============================================================================
+
+#! Finds a key in a sorted array of (key, value) word tuples.
+#!
+#! Input:  [KEY, start_ptr, end_ptr]
+#! Output: [is_key_found, key_ptr, start_ptr, end_ptr]
```
adr1anh, line 155: I would specify that this function is undefined if the array is not sorted, or if `(end_ptr - start_ptr) % 8 != 0`, or if `end_ptr < start_ptr`. This means the caller must ensure these preconditions hold

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,230 @@
+use.std::word
+
+# ===== ARRAY UTILITIES ============================================================================
+
+#! Finds a value in a sorted array of words.
+#!
+#! Input:  [VALUE, start_ptr, end_ptr]
+#! Output: [is_value_found, value_ptr, start_ptr, end_ptr]
```
adr1anh, line 18: Same point as below about mentioning the expected preconditions

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,230 @@
...
+
+                # dereference maybe_key_ptr
+                padw dup.12 mem_loadw
+                #=> [MAYBE_KEY, KEY, KEY, maybe_key_ptr, start_ptr, end_ptr]
+                
+                # there was no match, *maybe_key_ptr must be larger than KEY
+                exec.word::lt
+                assert.err="push_map_lowerbound invariant: MAYBE_KEY not greater than KEY"
+
+                # dereference maybe_key_ptr-4
```
adr1anh:
```suggestion
                # dereference maybe_key_ptr-8
```

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,230 @@
...
+                padw dup.8 sub.8 mem_loadw
+                #=> [MAYBE_KEY_PREV, KEY, maybe_key_ptr, start_ptr, end_ptr]
+                
+                # there was no match, *(maybe_key_ptr-4) must be smaller than KEY
```
adr1anh:
```suggestion
                # there was no match, *(maybe_key_ptr-8) must be smaller than KEY
```

adr1anh:
I've just merged #2145 where we moved some system events to the standard library. These also fit the pattern so we should try adapt this PR in the same way. Let me know if you have any questions around the new structure, I'd be happy to help out!

assembly-syntax/src/ast/instruction/advice.rs
```diff
@@ -15,6 +15,8 @@ pub enum SystemEventNode {
     PushU64Div,
     PushFalconDiv,
     PushSmtPeek,
+    PushArrayLowerBound,
+    PushMapLowerBound,
```
bobbinth: I'm not sure about this naming here: I think `PushArrayLowerBound` is fine, but maybe instead of `PushMapLowerBound` it should be `PushKeyValueLowerBound`?

bobbinth (commented):
Looks good! Thank you! I did a high-level review (i.e., didn't review MASM line-by-line) - and I think this is the right structure. I left one small comment inline about naming.

> I've just merged #2145 where we moved some system events to the standard library. These also fit the pattern so we should try adapt this PR in the same way. Let me know if you have any questions around the new structure, I'd be happy to help out!

I'm thinking that this functionality is generally useful, and so, maybe we should keep these as system events. We should still though merge the latest `next` into this PR to make sure everything still works fine.

> **One thing I'm not sure about is the binary search.** I suspect we're not going to deal with large buffers in MASM. We could replace binary search with linear search to detect unsorted arrays passed to push_xxx_lowerbound and avoid a potential footgun.

Very good point! Yes, let's switch to linear scan and maybe even go as far as to scan the whole specified memory range to ensure that all entries are in increasing order. As you've mentioned, cost-wise, this should be negligible, but could save many hours on debugging.

adr1anh:
> I'm thinking that this functionality is generally useful, and so, maybe we should keep these as system events.

I agree it's useful, but the utility comes from both the handler and the MASM wrapper being paired together. If I wanted to use this new handler without importing the stdlib, I'd have to pretty much reimplement the wrapping procedure. With the fairly large number of checks happening here, it's really not something easy to get right.

If this functionality is generally useful then it should be available to use in a safe way, which means calling the handler through a reliable(/audited) way. 

Finally, if some power users really needed to squeeze some extra performance out of their program, they would probably want to optimize the handler as well as the wrapper, so they'd just copy both and include them as a pair in their own library.

bobbinth:
> I agree it's useful, but the utility comes from both the handler and the MASM wrapper being paired together. If I wanted to use this new handler without importing the stdlib, I'd have to pretty much reimplement the wrapping procedure. With the fairly large number of checks happening here, it's really not something easy to get right.

Ah yes - good point! Agreed, the instructions should go into the standard library and be transformed into events (with accompanying handlers). Sorry for the confusion!

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,158 @@
+use alloc::{vec, vec::Vec};
+use miden_core::{EventId, Felt, LexicographicWord, Word};
+use miden_processor::{AdviceMutation, ErrorContext, EventError, ExecutionError, MemoryError, ProcessState};
+
+/// Qualified event name for the `lowerbound` events.
+pub const LOWERBOUND_ARRAY_EVENT_NAME: &str = "stdlib::collections::sorted_array::lowerbound_array";
+pub const LOWERBOUND_KEY_VALUE_EVENT_NAME: &str = "stdlib::collections::sorted_array::lowerbound_key_value";
+
+/// Constant Event ID for the `lowerbound` events, derived via `EventId::from_name(EVENT_NAME)`.
+pub const LOWERBOUND_ARRAY_EVENT_ID: EventId = EventId::from_u64(2382974753388103136);
```
adr1anh, line 12: Can we add a test to check this at test-time?

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,158 @@
...
+///   Advice stack: [...]
+///
+/// Outputs:
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [maybe_key_ptr, was_key_found, ...]
+fn push_lowerbound_result(
+    process: &ProcessState,
+    stride: u32,
+    err_ctx: &impl ErrorContext,
+) -> Result<Vec<AdviceMutation>, ExecutionError> {
```
adr1anh: If you modify the returned error to `EventError`, the question mark operator will automatically convert any of your errors to that type (it's just a `Box`). This should avoid the extra wrapper at the top.

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,158 @@
...
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [...]
+///
+/// Outputs:
+///   Operand stack: [KEY, start_ptr, end_ptr, ...]
+///   Advice stack: [maybe_key_ptr, was_key_found, ...]
+fn push_lowerbound_result(
+    process: &ProcessState,
+    stride: u32,
+    err_ctx: &impl ErrorContext,
```
adr1anh: You don't need any error context. It is automatically added by the processor when it wraps the `EventError`

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,158 @@
...
+
+
+// HELPER METHODS
+// --------------------------------------------------------------------------------------------
+
+// TODO: copied from sys_event_handlers.rs, extract to utils
+
+/// Reads (start_addr, end_addr) tuple from the specified elements of the operand stack (
+/// without modifying the state of the stack), and verifies that memory range is valid.
+fn get_mem_addr_range(
```
adr1anh: I think we can use this for the keccak handler as well. Let's put this helper in the module.

adr1anh (commented):
Looks good! Thanks for updating to the new event structure. Left a couple of small nits.

stdlib/src/handlers/sorted_array.rs
bobbinth: Yes, let's put it somewhere [here](https://github.com/0xMiden/miden-vm/blob/69ebe68851e7bc3179fd8499c966a02a951744de/stdlib/src/handlers/mod.rs#L10).

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,235 @@
...
+
+                push.0
+                #=> [is_key_found = 0, maybe_key_ptr = end_ptr, start_ptr, end_ptr]
+            else
+                # The key was not found, and the maybe_key_ptr is neither the start nor the end
+                # of the list. Make sure that the key at maybe_key_ptr-1 < KEY and at maybe_key_ptr > KEY
+                #=> [maybe_key_ptr, KEY, start_ptr, end_ptr]
+
+                # make sure maybe_key_ptr is pointing to a key, not value
+                dup dup.6 sub u32assert2 u32mod.8 assertz.err="push_map_lowerbound invariant: key_ptr must be double-word aligned with start_ptr"
```
bobbinth: Error message here and in a few other places need to be updated.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,235 @@
...
+                exec.word::lt
+                assert.err="push_array_lowerbound invariant: start_ptr not greater than VALUE"
+                
+                push.0
+                #=> [is_value_found = 0, maybe_value_ptr = start_ptr, start_ptr, end_ptr]
+            end
+        else
+            if.true
+                # the value was not found, maybe_value_ptr = end_ptr and the list is not empty
+                # assert that the value before end_ptr < VALUE
```
bobbinth: I would maybe explain here that we know that the list is not empty because if it was empty `start_ptr == end_ptr`, and this would imply that `maybe_value_ptr == start_ptr` and we've handled this case above.

gswirski:
Given a large volume of comments, I'll push changes and mark conversations as resolved as I go. This PR is not yet ready for another review, I'll ping you when it's done. :)

stdlib/src/handlers/sorted_array.rs
gswirski: Given this method was lifted from miden_processor (and is used in that crate), I added it here: https://github.com/0xMiden/miden-vm/pull/2114/files#diff-f02be9e480cda12fea6e1e68a47195f743fabdacd5f456e520cc59a56b174ecaR956

gswirski:
All comments addressed, all features implemented, docs (including cycle counts) provided. This is now ready for another round of reviews.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+            if.true
+                # the list is empty
+                #=> [maybe_value_ptr = start_ptr = end_ptr, VALUE, start_ptr, end_ptr]
+
+                # Prepare output (6 cycles)
+                movdn.4 dropw push.0
+                #=> [is_value_found = 0, value_ptr = start_ptr, start_ptr, end_ptr]
+            else
+                # the value was not found, maybe_value_ptr = start_ptr, and the list is not empty
+                # assert that the value at start_ptr > VALUE
```
Al-Kindi-0: nit: assert that the value at start_ptr is greater than VALUE

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+
+                # No value was found (1 cycle)
+                push.0
+                #=> [is_value_found = 0, maybe_value_ptr = start_ptr, start_ptr, end_ptr]
+            end
+        else
+            if.true
+                # the value was not found, maybe_value_ptr = end_ptr
+                # the list is not empty, otherwise maybe_value_ptr = start_ptr and handled above
+                # assert that the value before end_ptr < VALUE
```
Al-Kindi-0: nit: assert that the value before end_ptr is strictly less than VALUE

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+                # (119 cycles)
+                exec.word::gt
+                assert.err="lowerbound_array invariant: last word not smaller than VALUE"
+
+                # No value was found (1 cycle)
+                push.0
+                #=> [is_value_found = 0, maybe_value_ptr = end_ptr, start_ptr, end_ptr]
+            else
+                # The value was not found, and the maybe_value_ptr is neither the start nor the end
+                # of the list. Make sure that the value at maybe_value_ptr-1 < VALUE and at maybe_value_ptr > VALUE
```
Al-Kindi-0: nit: same comment here and would probably write them as two bullet points i.e., 

... make sure that 
1. the value at maybe_value_ptr-1 is strictly less than VALUE, and that
2. the value at maybe_value_ptr is strictly greater than VALUE

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+                assert.err="lowerbound_array invariant: start_ptr not greater than VALUE"
+
+                # No value was found (1 cycle)
+                push.0
+                #=> [is_value_found = 0, maybe_value_ptr = start_ptr, start_ptr, end_ptr]
+            end
+        else
+            if.true
+                # the value was not found, maybe_value_ptr = end_ptr
+                # the list is not empty, otherwise maybe_value_ptr = start_ptr and handled above
```
Al-Kindi-0: nit: either exchange = for "equals" or put the equation between quotation marks

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+#! Input:  [KEY, start_ptr, end_ptr]
+#! Output: [is_key_found, key_ptr, start_ptr, end_ptr]
+#!
+#! Cycles:
+#!   Key exists: 56 cycles
+#!   Key doesn't exist and the array is empty: 25 cycles
+#!   Key doesn't exist and is smaller than all stored keys: 151 cycles
+#!   Key doesn't exist and is larger than all stored keys: 149 cycles
+#!   Key doesn't exist: 298 cycles
+export.find_key_value
```
Al-Kindi-0, line 163: same comments about using math operators in comments as above

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    // Read inputs from the stack
+    let key_offset = 1;
+    let start_addr_offset = 5;
+    let end_addr_offset = 6;
+    let key = LexicographicWord::new(process.get_stack_word(key_offset));
+    let (start_addr, end_addr) = get_mem_addr_range(process, start_addr_offset, end_addr_offset)?;
+
+    // Validate the start_addr is word-aligned (multiple of 4)
+    if start_addr % 4 != 0 {
```
Al-Kindi-0: nit: using constants, even if just internally defined, would be a bit better in my opinion

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,189 @@
...
+    let mut result = None;
+
+    // Test the first element
+    let mut previous_word = get_word(start_addr)?;
+    if previous_word >= key {
+        was_key_found = previous_word == key;
+        result = Some(start_addr);
+    }
+
+    // Validate the entire array is non-decreasing and find the first element >= key
```
Al-Kindi-0: nit: >= -> greater than or equal

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,189 @@
...
+///
+/// # Errors
+/// Returns an error if the provided word array is not sorted in ascending order.
+pub fn handle_lowerbound_array(process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError> {
+    push_lowerbound_result(process, 4).map_err(EventError::from)
+}
+
+/// Pushes onto the advice stack the first pointer in [start_ptr, end_ptr) such that
+/// mem[word_ptr] >= KEY in lexicographic order of words. If all keys are < KEY, returns end_ptr.
+/// The array must be a list of key-value pairs (word tuples) where all keys are in ascending order.
```
Al-Kindi-0: nit: I would also use non-decreasing order to stay consistent throughout

Al-Kindi-0 (approved):
Looks great!
Left a few nits, but feel free to ignore

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+
+        dup.6 dup.2 eq
+        #=> [is_start_ptr, is_end_ptr, maybe_value_ptr, VALUE, start_ptr, end_ptr]
+
+        if.true
+            # the value was not found, and maybe_value_ptr = start_ptr
+
+            if.true
+                # the list is empty
+                #=> [maybe_value_ptr = start_ptr = end_ptr, VALUE, start_ptr, end_ptr]
```
PhilippGackstatter: The handler says the list is empty, but we have to check that claim from the MASM side, afaict. So wouldn't we have to assert that `start_ptr == end_ptr` here to check the handler's claim is true?

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+                # the value was not found, maybe_value_ptr = start_ptr, and the list is not empty
+                # assert that the value at start_ptr > VALUE
+                #=> [maybe_value_ptr = start_ptr, VALUE, start_ptr, end_ptr]
+
+                movdn.4
+                #=> [VALUE, maybe_value_ptr, start_ptr, end_ptr]
+
+                # dereference maybe_value_ptr (7 cycles)
+                dup.4 padw movup.4 mem_loadw
+                #=> [MAYBE_VALUE, VALUE, maybe_value_ptr, start_ptr, end_ptr]
```
PhilippGackstatter: Nit: I think for me it would be clearer if we wrote `start_ptr` instead of `maybe_value_ptr` and `FIRST_WORD` instead of `MAYBE_VALUE` (roughly the same applies to the other branch but with `end_ptr`).

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+
+        dup.6 dup.2 eq
+        #=> [is_start_ptr, is_end_ptr, maybe_key_ptr, KEY, start_ptr, end_ptr]
+
+        if.true
+            # the key was not found, maybe_key_ptr = start_ptr
+
+            if.true
+                # the list is empty
+                #=> [maybe_key_ptr = start_ptr = end_ptr, KEY, start_ptr, end_ptr]
```
PhilippGackstatter: Same comment here regarding checking that start and end pointer match.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -0,0 +1,279 @@
...
+                # the key was not found, maybe_key_ptr = start_ptr and the list is not empty
+                # assert that the key at start_ptr > KEY
+                #=> [maybe_key_ptr = start_ptr, KEY, start_ptr, end_ptr]
+
+                movdn.4
+                #=> [KEY, maybe_key_ptr, start_ptr, end_ptr]
+                
+                # dereference maybe_key_ptr (7 cycles)
+                dup.4 padw movup.4 mem_loadw
+                #=> [DEREF, KEY, maybe_key_ptr, start_ptr, end_ptr]
```
PhilippGackstatter:
```suggestion
                #=> [FIRST_KEY, KEY, maybe_key_ptr, start_ptr, end_ptr]
```

Nit: Should this be called first key to match `LAST_KEY` in the other branch?

stdlib/tests/collections/sorted_array.rs
```diff
@@ -0,0 +1,218 @@
...
+                push.124 push.100 push.[{key}]
+
+                exec.sorted_array::find_key_value
+                exec.truncate_stack
+            end
+        "
+        );
+
+        let program = build_test!(source, &[]);
+        println!("testing {key}");
```
PhilippGackstatter, line 133:
```suggestion
```

Nit: Probably a leftover?

stdlib/tests/collections/sorted_array.rs
```diff
@@ -0,0 +1,218 @@
...
+                push.112 push.100 push.[{key}]
+
+                exec.sorted_array::find_word
+                exec.truncate_stack
+            end
+        "
+        );
+
+        let program = build_test!(source, &[]);
+        println!("testing {key}");
```
PhilippGackstatter, line 38:
```suggestion
```

stdlib/asm/collections/sorted_array.masm
gswirski: We already check that on the MASM side but I agree it's not obvious. To enter this branch of code we have: `maybe_key_ptr == start_ptr && maybe_key_ptr == end_ptr` which implies `start_ptr == end_ptr`. In other words: if `start_ptr != end_ptr`, we cannot be here.

I'll extend the comment to explain this.

stdlib/tests/collections/sorted_array.rs
gswirski: I left it intentionally. By default `cargo test` does not show stdout of successful tests. When the test fails, you'll see "testing A,B,C,D" and you'll know which scenario caused trouble.

I don't feel strongly either way, can remove it.

stdlib/asm/collections/sorted_array.masm
gswirski: @Al-Kindi-0 I find English (and natural language in general) to be extremely annoying to describe logical relationships leading to verbose descriptions and ambiguity. How would you feel about:

```
# assert that "value at start_ptr" > VALUE
```

or even more pedantic:

```
# assert that `"value at start_ptr" > VALUE`
```

or maybe:

```
# assert that `value_at_start_ptr > VALUE`
```

In my mind ">" and ">=" leave no room for interpretation vs in English it's useful to say "strictly greater" rather than just "greater".

Logical operators are shorter, faster to scan, leave no room for interpretation, and are easier to spot inconsistencies with the actual code implementation. English looks nicer, but the intended audience for these comments are other engineers, auditors and people who probably prefer denser descriptions.

Same consideration for all related comments. In my mind quoting all formulas (preferably with `) makes a ton of sense. English - not so much.

stdlib/asm/collections/sorted_array.masm
gswirski: Agreed about bullet points. See my comment about logical relationships: https://github.com/0xMiden/miden-vm/pull/2114#discussion_r2348341757

stdlib/tests/collections/sorted_array.rs
```diff
@@ -0,0 +1,218 @@
...
+
+            push.8456.415.4922.593 mem_storew.108 dropw
+            push.3015.7211.2002.5143 mem_storew.112 dropw
+
+            push.3015.7211.2002.5143 mem_storew.116 dropw
+            push.8595.8794.8303.7256 mem_storew.120 dropw
+
+            push.124 push.100 push.[8675,5816,5458,2767]
+
+            exec.sorted_array::find_word
```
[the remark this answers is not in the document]
gswirski: Great catch, Cursor. Fixed

gswirski:
Addressed all lints, rustdoc errors, comments and nits except for:
1. `println!("testing {key}")`
2. Replacing logical operators with their English descriptions

I think in both instances the current version provides additional value but I don't feel very strongly either way.

stdlib/tests/collections/sorted_array.rs
PhilippGackstatter: Ah, well I don't work so much in this code base so I don't have strong opinions. But I often use `--no-capture` when trying something and then the output would be interleaved by unrelated `println`s that I didn't add, so I usually avoid them to keep things cleaner.

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,192 @@
...
+            result = Some(current_addr);
+        }
+        previous_word = current_word;
+        current_addr += stride;
+    }
+
+    Ok(vec![AdviceMutation::extend_stack(vec![
+        Felt::from(was_key_found),
+        Felt::from(result.unwrap_or(end_addr)),
+    ])])
```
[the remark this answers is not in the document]
gswirski: Not true. Advice stack is in reverse order than the rest of the MASM code. `adv_push.2` flips it back to the proper order.

stdlib/asm/collections/sorted_array.masm
Al-Kindi-0: sounds good, thanks

stdlib/asm/collections/sorted_array.masm
Al-Kindi-0: Fully agree that natural language can get in the way of fast and easy parsing, though I have a slight dislike of mixing natural language with math symbols without any separator, because then symbols starting getting in the way of reading natural language.
Your solution, the pedantic one, makes me happy, but any of your solution will do the job, as long as we separate math equations from natural language.

processor/src/operations/sys_ops/sys_event_handlers.rs
Al-Kindi-0: Missed the equality sign, thanks for explaining

Al-Kindi-0:
@gswirski , is this good to go?

gswirski:
> @gswirski , is this good to go?

I think so!

I addressed all the feedback except for 2 small things I didn’t agree with (but don’t feel strongly about and can easily update): https://github.com/0xMiden/miden-vm/pull/2114/#issuecomment-3291338504

processor/src/lib.rs
```diff
@@ -950,3 +950,27 @@ pub(crate) fn add_error_ctx_to_external_error(
...
     }
 }
+
+/// Reads (start_addr, end_addr) tuple from the specified elements of the operand stack (
+/// without modifying the state of the stack), and verifies that memory range is valid.
+pub fn get_mem_addr_range(
+    process: &ProcessState,
+    start_idx: usize,
+    end_idx: usize,
+) -> Result<(u32, u32), MemoryError> {
```
bobbinth, line 960: A potential small improvement: maybe this could be a method on the `ProcessState`? It could look something like:
```rust
impl ProcessState {
    pub fn get_mem_addr_range(
        &self,
        start_idx: usize,
        end_idx: usize,
    ) -> Result<Range<MemoryAddress>, MemoryError> {
        ...
    }
}
```
In the above, I've also used `Range<MemoryAddress>` instead of `(u32, u32)` for the return value as I think it is a bit more self-describing (and type-safe).

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    // Validate address range is valid
+    if start_addr > end_addr {
+        return Err(MemoryError::InvalidMemoryRange {
+            start_addr: start_addr as u64,
+            end_addr: end_addr as u64,
+        }
+        .into());
+    }
```
bobbinth, line 90: This is already validated by `get_mem_addr_range()` - right? If so, and if we still want to double-check this, maybe an `assert!` would be better?

stdlib/src/handlers/sorted_array.rs
```diff
@@ -0,0 +1,192 @@
...
+    // Test the first element
+    let mut previous_word = get_word(start_addr)?;
+    if previous_word >= key {
+        was_key_found = previous_word == key;
+        result = Some(start_addr);
+    }
+
+    // Validate the entire array is non-decreasing and find the first element where `element >= key`
+    let mut current_addr = start_addr + stride;
+    while current_addr < end_addr {
```
bobbinth, line 133: nit: a slightly more expressive way to write this could be:
```rust
for addr in (start_addr..end_addr).step_by(stride).skip(1) {
    ...
}
```

processor/src/lib.rs
gswirski: Created a followup PR here: https://github.com/0xMiden/miden-vm/pull/2194

Issue #1747, opened 2025-04-03
RFC: Providing type information for Miden packages


bitwalker:
This is a proposal to define a type system for the Miden VM that makes it possible to precisely specify the ABI of Miden Assembly programs and libraries. This will further enable reasoning about the binary compatibility of Miden packages, and the ability to provide stable interfaces for interoperability regardless of source language or toolchain version.

This proposal focuses on defining the type system, and describing its application to the Miden Assembly AST, the `MastForest`, and by extension, the Miden package format.

Lastly, it sketches out how we can surface the type system in Miden Assembly syntax, so that we can begin to document the type signatures and calling conventions of existing Miden Assembly libraries, and in some cases, make changes to enable procedures to be called from higher-level languages like Rust.

## Terminology

Since this document contains a fair amount of language related to types in terms of both their type-theoretic properties, and their target-specific physical properties (i.e. data layout) - I'd like to clarify up front some terminology used throughout, so that it is unambiguous what is meant by those terms:

- **_type_**, a name associated with some specific data representation, typically in the context of a _type system_.
- **_type system_**, a (sometimes closed) set of types, and their ontology, i.e. their categorization (_type kinds_), relationships (subtyping, valid casts/conversions), and properties
- **_type kind_**, a family of types, e.g. function types, integral types, etc.
- **_data type_**, a type which has a defined physical representation in memory
- **_immediate_**, a data type which can be used as an "immediate" operand of a machine instruction, i.e. the operand _is_ the data, not a reference to data located in memory or somewhere else. For example, a value whose type is a pointer to an array of 64 bytes is a valid immediate, because the data represented by the value is the _address_ of the array, not the array itself, which is _not_ a valid immediate due to its size.
- **_primitive type_**, a data type which is fundamental and not composed of any other type, e.g. fixed-width integer types. Note that "primitive" in this context does not mean "built-in", but rather is a reference to the fact that such types are used as building blocks for specifying the data representation of all other types. Most machines only understand integers (and often floats), and will interpret integers as memory addresses when used as a memory operand of a machine instruction. Aggregate types are thus formed by either splitting the physical representation of the type across a subset of registers/operands, in the case of small aggregate types; OR, by interpreting some region of memory as containing a value of that type. Notably, there are no machine instructions that understand aggregate types, instead, operations on those types are decomposed into operations on the primitive types that comprise them (or by interpreting some part of its physical representation as one).
[...]

PR #2120, opened 2025-09-02
feat(assembly): implement type signature syntax

bitwalker:
This PR implements the syntax for type signatures initially described in #1747, with some compromises given the current parser implementation. I'll summarize the syntax below, but see that RFC for more background and context. I expect that changes to the proposed syntax here are going to be minor, but if there are major issues identified, we can use this PR to address those before merging.

This PR also reworks the parsing and evaluation of constants to address some issues and tighten up the syntax a bit. The only significant change here is that it is no longer valid to mix word literals and dot-delimited `push` syntax - you can either push a word literal, or push one or more integer literals, but not both at the same time.

## Syntax Tour

```text
# A type alias for an array type
type Word = [felt; 4]
# A type alias for a struct type
type Foo = struct { a: u16, b: i1 }
# A type alias for a pointer type
type ByteRef = *u8
# A type alias for a primitive type
type Int8 = i8

# An enumerated type definition, equivalent to `type Tag = u8`, which
# also declares constants corresponding to each variant, the value of which
# is the discriminant for that variant, and which must be a valid instance of the
# declared type
enum Tag : u8 {
    A,         # Discriminants are derived from 0, or the prev variant, if not explicitly specified
    B = 2,
    C = B * 2, # Arbitrary expressions are permitted, including referencing earlier variants
    D          # The discriminant here is effectively 5, i.e. 2 * 2 + 1
}

# A procedure using all of the basic syntax (`pub` visibility modifier, type signature)
pub fn add(a: felt, b: felt) -> felt
    add
end

# Same as above, but also demonstrating the syntax for specifying calling convention
@callconv(C)
pub fn add2(a: felt, b: felt) -> felt
    add
end

# Type signatures are optional, as are dot-delimited visibility modifiers
pub fn.mul
    mul
end

# The old `export` visibility modifier is still supported for backwards-compatibility
export.mul2
    mul
end
```

A quick summary of the non-type changes:

* A new visibility modifier is introduced, `pub`, to replace the use of `export`
* A new keyword for procedure definitions is introduced, `fn`, to replace `proc`
* The `.` between visibility modifier and `fn`/`proc` is made optional
* The `.` between `fn`/`proc` and procedure name is made optional
* The order in which constants are declared no longer matters for evaluation

---

I have a few minor tasks to wrap up before this is completely finished (adding some more tests, updating the changelog), but this is essentially ready for review.


assembly-syntax/src/ast/type.rs
```diff
@@ -241,20 +444,87 @@ pub struct Variant {
...
+
+    /// Used to validate that this variant's discriminant value is an instance of `ty`,
+    /// which must be a type valid for use as the underlying representation for an enum, i.e. an
+    /// integer type up to 64 bits in size.
+    ///
+    /// It is expected that the discriminant expression has been folded to an integer value by the
+    /// time this is called. If the discriminant has not been fully folded, then an error will be
+    /// returned.
+    pub fn assert_instance_of(&self, ty: &Type) -> Result<(), crate::SemanticAnalysisError> {
+        use miden_core::StarkField;
```
Al-Kindi-0: A general remark: I think it would be a bit better for us to define a `MODULUS` constant at the crate level in order to avoid having to import this trait every time we need `Felt::MODULUS`

assembly-syntax/src/ast/type.rs
```diff
@@ -241,20 +444,87 @@ pub struct Variant {
...
+                });
+            },
+        };
+
+        match ty {
+            Type::I1 if value > 1 => Err(SemanticAnalysisError::InvalidEnumDiscriminant {
+                span: self.discriminant.span(),
+                repr: ty.clone(),
+            }),
+            Type::I1 => Ok(()),
```
Al-Kindi-0, line 806: nit: would it be better to merge these?

Al-Kindi-0: Maybe not in order to keep consistent with the other arms

assembly-syntax/src/sema/context.rs
```diff
@@ -47,53 +47,136 @@ impl AnalysisContext {
...
+                        evaluating.push(name.clone());
+                    }
+                    continuations.push(Cont::Return(name.clone()));
+                    continuations.push(Cont::Eval(self.get_constant(&name)?.clone()));
+                },
+                Cont::Eval(ConstantExpr::BinaryOp { span, op, lhs, rhs, .. }) => {
+                    continuations.push(Cont::Apply(Span::new(span, op)));
+                    continuations.push(Cont::Eval(*lhs));
+                    continuations.push(Cont::Eval(*rhs));
+                },
```
Al-Kindi-0, line 140: Q: Is the order of pushes correct here? Or maybe I am missing something?

assembly-syntax/src/sema/context.rs
```diff
@@ -47,53 +47,136 @@ impl AnalysisContext {
...
+                Cont::Apply(op) => {
+                    let lhs = stack.pop().unwrap().expect_int();
+                    let rhs = stack.pop().unwrap().expect_int();
```
Al-Kindi-0, line 143: Here as well ?

assembly-syntax/src/sema/context.rs
```diff
@@ -47,53 +47,136 @@ impl AnalysisContext {
...
+                    if should_simplify {
+                        *original = stack.last().unwrap().clone();
+                    }
+                },
+            }
+        }
+
+        // When we reach here, we should have exactly one expression on the operand stack
+        assert_eq!(stack.len(), 1, "expected constant evaluation to produce exactly one output");
+        Ok(unsafe { stack.pop().unwrap_unchecked() })
```
Al-Kindi-0, line 181: Q: Is the unsafe block necessary here ?

assembly/src/linker/mod.rs
```diff
@@ -727,6 +737,71 @@ impl Linker {
...
+        &self,
+        gid: GlobalProcedureIndex,
+    ) -> Result<Option<Arc<types::FunctionType>>, LinkerError> {
+        // Fetch procedure metadata from the graph
+        let module = match &self[gid.module] {
+            ModuleLink::Ast(module) => module,
+            ModuleLink::Info(module) => {
+                let proc = module
+                    .get_procedure_by_index(gid.index)
+                    .expect("invalid global procedure index");
```
Al-Kindi-0, line 751: Q: Is the expect here fine ?

Al-Kindi-0 (commented):
Looks great to me!
Did an initial pass and left mostly questions for my own understanding.

assembly-syntax/src/ast/type.rs
bitwalker: Good suggestion! I think we maybe had one at one point (maybe we still do, and I need to look for it 😅)

assembly-syntax/src/ast/type.rs
bitwalker: I generally prefer to express things like this in `match` guards if I can, just to avoid the tendency of code to shift right due to indentation when nesting conditionals in match bodies - but I could go either way here.

assembly-syntax/src/sema/context.rs
bitwalker: The order we evaluate `lhs` or `rhs` doesn't matter, so long as it agrees with the order in which we expect the operands when evaluating the `Apply` (which is the case currently, i.e. we expect `lhs` on top of the stack).

We could invert the order though to be more in line with how we generally lay these things out on the operand stack though to be more consistent with how we lay things out on the operand stack for these operations in MASM in general. I'll make a note to do that.

assembly-syntax/src/sema/context.rs
bitwalker: See my comment above - basically it doesn't matter as long as they agree, but for the sake of consistency I think it makes sense to expect `rhs` on top.

assembly-syntax/src/sema/context.rs
bitwalker: Yes, since `unwrap_unchecked` is marked `unsafe`. I meant to add a `// SAFETY` comment here explaining why it is used and why it is safe, so I'll make a note to do that.

assembly/src/linker/mod.rs
bitwalker: Yes - within the linker we expect that any `GlobalProcedureIndex` is one the linker returned to the caller when they added a module or library to the linker's module graph. If it is invalid, we always assert, since it reflects a bug somewhere that can't be reasonably handled by a caller anyway.

We could make it a non-panicking error, though IMO, for internal errors of the sort that would produce an invalid handle here, assertions are preferable.

assembly-syntax/src/ast/type.rs
Al-Kindi-0: Makes sense

assembly-syntax/src/sema/context.rs
Al-Kindi-0: Thank you for the explanation!

assembly-syntax/src/parser/grammar.lalrpop
bobbinth, line 1: Some comments/questions about naming/syntax/semantics (not really specific to this file, but attaching here to have threaded comments).

#### `fn` vs. `proc`
Almost universally we've been using "procedure" when referring Miden assembly procedures/functions. I think it may create some confusion if we now start using the word "function" (i.e., our documentation will likely keep using the word "procedure" when referring to `fn` declarations) . Not sure if the confusion will be big - we'll basically be using "function" and "procedure" interchangeably - but at the same time, I'm not seeing a strong reason to change things.

For example, I think we could have the following syntax just as easily:
```
# public visiblity
pub proc mul
    mul
end

# private visibility
proc mul2
    mul
end
```

#### `export` keyword
I think it may be fine to get rid of this. We are introducing breaking changes anyways, and doing a find-and-replace for `export.` is not that difficult. The main idea is to reduce "optionality" - i.e., having multiple ways to do the same thing.

#### `.` separators
It is not clear to me if we should make `.` optional in visibility and name modifiers. For example, we could keep things close to the current syntax:
```
pub.proc.mul
    mul
end
```
The idea again is to keep just one way to do things. We could of course switch completely to using spaces instead of `.`, but that would be more work for relatively little gain IMO.

#### Number of procedure locals
How would we specify the number of procedure locals with the new syntax? Would it be something like:
```
pub.proc.foo.2(a: u32, b: u32) -> u32
    ...
end
```
Or is there some other way?

#### Parameter names
Parameter names are purely for informational purposes, right? i.e., we won't really use them in procedure bodies. If so, are they optional? Could we do something like:
```
pub.proc.foo.2(u32, u32) -> u32
    ...
end
```
If they are optional, I think there is an argument to be made that the above syntax shouldn't be allowed.

#### Constant declaration syntax
Given that we now allow something like this:
```
type Word = [felt; 4]
```
Should we also allow declaring constants in the same way? i.e.:
```
const FOO = 10;
const BAR = [1, 2, 3, 4];
```

#### Calling convention options
In the example there is `@callconv(C)` modifier, what other conventions can we specify? In the `miden-lib` code we frequently annotate procedures (for now just in the comments) as `Invocation: exec`, `Invocation: dynexec`, `Invocation: call`, `Invocation: syscall` - how would we express this using `@callconv` attribute?

#### Parameter and word layouts
My understanding is that procedure parameters are laid out from the top of the stack. That is:
```
pub.proc.foo.2(a: u32, b: u32) -> u32
     # => [a, b] - i.e., `a` is at the top of the stack
    ...
end
```
This is consistent with how we describe the parameters for procedures in the comments as well. But do we assume that words are laid out in big endian or little endian order? Should we have two primitive types for words - `BeWord` and `LeWord`?

#### Reference type semantics
In the example there is `type ByteRef = *u8` declaration. How is it handled? To get a felt address from this value, would be basically do `byte_address / 4`?

Similarly, if we had something like `type WordRef = *Word`, would we need to something like `word_address * 4` to get the element address? If so, this is a bit different to how we use addresses right now - i.e., even when referring to words, we use their "element address" (so, no need to multiply by 4). Is there a way to express this using the type system?

#### Alternative signature declaration syntax
We could also consider specifying procedure signatures via attributes. For example, instead of:
```
@callconv(exec)
pub.proc.foo.2(a: u32, b: u32, c: Word) -> u32, Word
    ...
end
```
We could do something like:
```
@inputs(a: u32, b: u32, c: Word)
@outputs(d: u32, e: Word)
@callconv(exec)
pub.proc.foo.2
    ...
end
```
The main advantages are:
- This doesn't introduce too much of a new syntax and is very similar to how we already specify inputs/outputs in procedure comments.
- It allows us to name the outputs.

The main disadvantage is that it is different from traditional ways of specifying procedure signatures and is probably a bit more difficult to read.

Curious if @PhilippGackstatter, @Al-Kindi-0, @partylikeits1983, @Fumuran or anyone else have an opinion on this.

bitwalker: > Almost universally we've been using "procedure" when referring Miden assembly procedures/functions. I think it may create some confusion if we now start using the word "function" (i.e., our documentation will likely keep using the word "procedure" when referring to `fn` declarations) . Not sure if the confusion will be big - we'll basically be using "function" and "procedure" interchangeably - but at the same time, I'm not seeing a strong reason to change things.

I recall Alex being a bit confused about the distinction, and I'm not sure there is a particularly good reason to continue emphasizing "procedure" as the terminology we use here, versus something more familiar to the broader developer audience. That said, I don't have a strong opinion on that specific aspect. I do think that "function" is actually strictly more accurate for what we have in MASM, as procedure is - if distinct at all - a narrower subset of functions in general (i.e. those for which no result value is produced), but even that narrow distinction is up for debate.

I'd say if there is any reason to change, it's probably simply because it is going to be more familiar for most developers moving forward. Those already working in MASM are almost certainly not going to be a group of people confused by the switch, but that is a presumption on my part. Like I said though, it's not something I feel super strongly about, I just think the syntax looks nicer 😆.

> For example, I think we could have the following syntax just as easily:

Sure, `pub proc` and `proc` also would work fine - I find `proc` to be a bit awkward as a keyword, but it's already the one we use, so not something that matters here.

> #### `export` keyword
> I think it may be fine to get rid of this. We are introducing breaking changes anyways, and doing a find-and-replace for `export.` is not that difficult. The main idea is to reduce "optionality" - i.e., having multiple ways to do the same thing.

My primary goal here was to allow for a deprecation period for the old syntax, since we have a non-trivial amount of existing MASM code, and supporting the old syntax for a release cycle is low-effort (and we can emit warning diagnostics if we wish to do so).

Ultimately though, yes, I think we should remove `export` entirely.

> #### `.` separators
> It is not clear to me if we should make `.` optional in visibility and name modifiers. For example, we could keep things close to the current syntax:

I want to remove `.` as a separator almost entirely, only using them for specifying immediate instruction operands, e.g. `push.0`. They serve no purpose otherwise IMO, and just add noise to the syntax, and make the grammar more awkward.

> The idea again is to keep just one way to do things. We could of course switch completely to using spaces instead of `.`, but that would be more work for relatively little gain IMO.

To be clear, the optionality here is to provide for a deprecation period for the older syntax - support for the old syntax would be removed after an appropriate period of time has been provided to switch.

The grammar will become less verbose if we can remove the need to support `.` separators all over the place, so aside from updating code (which could be automated using the formatter), I see this as strictly a win IMO.

> How would we specify the number of procedure locals with the new syntax?

Yes, basically. Specifically (assuming we use `proc` instead of `fn`):

```
pub proc foo.2(a: u32, b: u32) -> u32
    ...
end
```

> Parameter names are purely for informational purposes, right? 

Yes, correct - the names are never used.

> If so, are they optional? 

They aren't now, but we could allow them to be optional. Personally, I think parameter names should be required, because it will result in clearer documentation.

> If they are optional, I think there is an argument to be made that the above syntax shouldn't be allowed.

Sorry, which syntax, the version with or without names? There is no support for the optional form currently, so if you mean disallowing it, I agree 😄.

> Given that we now allow something like this:
> 
> ```
> type Word = [felt; 4]
> ```
> 
> Should we also allow declaring constants in the same way? i.e.:
> 
> ```
> const FOO = 10;
> const BAR = [1, 2, 3, 4];
> ```

That is indeed supported in this PR 🙂. In fact, all of the top-level declarations allow eliding the `.` separators (if they had them before), e.g. `const A = 1`; `use ::std::math::u64->myu64`, etc.

> In the example there is `@callconv(C)` modifier, what other conventions can we specify? 

To be clear, the syntax is supported here to allow for representing the full information that can be represented about a function type in the IR type system, but I don't expect the ability to specify a calling convention to be used all that often, if at all.

The actual list corresponds to the variants of the `CallConv` enumeration, which documents their semantics, but the main ones are:

* `fast` - ad-hoc convention where everything is passed by value
* `C` - the rules for passing a given type are described by mapping the C ABI for WebAssembly to Miden, according to the rules for mapping byte-addressable memory to Miden's felt-addressable memory. The primary significance here has to do with what types must be passed by reference, and how that is done.

There are others, but they aren't relevant outside the compiler currently. The default is `fast`, and is the convention I expect will be used for most hand-written MASM procedures (though `C` might be a good fit in some cases as well).

> In the `miden-lib` code we frequently annotate procedures (for now just in the comments) as `Invocation: exec`, `Invocation: dynexec`, `Invocation: call`, `Invocation: syscall` - how would we express this using `@callconv` attribute?

You don't, at least, that's not what calling convention means in this case. The convention for a procedure is distinct from the instruction used to invoke it, because we cannot control how a procedure is invoked in general. Instead, there are rules that must hold at call sites for any given procedure:

* Invoking a procedure via `call` requires that none of its argument/result types are pointers/references.
* Taking the "address" of a procedure for use with `dynexec`/`dyncall` requires that no more than 15 elements of the operand stack are required to hold its arguments/results, since one element is required to hold the address of the callee hash.
* Invoking a procedure via `call` or `dyncall` where arguments or results are heap-allocated values, requires use of the advice provider to pass those values between caller and callee. The precise details of this I'm not going to get into here.

All calling conventions must obey those rules, and simply specify the precise protocol for how arguments and results of specific types are passed from caller to callee and back. Some calling conventions may choose to put everything on the operand stack, while others may choose to write anything larger than a felt to memory, and pass them by reference. As a specific well-known example, the `C` calling convention dictates that values returned by reference require the caller to allocate the memory for those values, and pass the return value address as the first parameter of the function.

> My understanding is that procedure parameters are laid out from the top of the stack. That is:
> 
> ```
> pub.proc.foo.2(a: u32, b: u32) -> u32
>      # => [a, b] - i.e., `a` is at the top of the stack
>     ...
> end
> ```

Correct. The first parameter in the parameter list is expected on top of the operand stack on entry to the procedure.

> This is consistent with how we describe the parameters for procedures in the comments as well. But do we assume that words are laid out in big endian or little endian order? 

The type system assumes all values are laid out as bytes (or felts) in order of lowest-addressed to highest-addressed byte, i.e. in memory order (we've been calling this little-endian, but that's a bit of a misnomer, because endianness refers to the way in which the bytes of some specific value are interpreted, not the address order of bytes in general, which is AFAIK always least to greatest).

So `[felt; 4]` (i.e. a "word") is laid out with the first felt of the word being on top of the operand stack, because this corresponds to how a value of that type is laid out in memory.

> Should we have two primitive types for words - `BeWord` and `LeWord`?

This wouldn't work, but I think the motivation for this question isn't actually about words in general (for which the lowest-address-first ordering is fine), but actually about specific types returned by APIs of the transaction kernel which store information in a word-sized value in such a way that loading/storing from memory doesn't require reversing the word (i.e. these APIs operate on an implicit struct type that happens to be word-sized, but whose layout on the operand stack differs from the layout in memory).

The problem with having a different layout on the operand stack than in memory, is that it is only really viable to deal with such types in hand-written MASM. Low-level type systems assume a consistent memory layout determined by the target architecture, in order to be able to determine how to store and interact with those values in registers (if permitted) as well as in memory. In our case, that's a little-endian 32-bit machine.

IMO the problem/question here is really - what do we do about the type(s?) in the existing transaction kernel API which exhibit this context-sensitive layout property? I need to see if there is _maybe_ a path here where we could describe them as struct types with a special annotation that indicates this, but I'm not 100% sure if that will actually work or not without running some scenarios.

More generally, I think we need to ensure that all of our type layout is done with a consistent, lowest-addressed-bytes-first approach. Migrating the current implementation of the transaction kernel is a big lift though, so we'll need some kind of short-term solution for binding against the transaction kernel from compiler-generated code. I'm still investigating this.

> In the example there is `type ByteRef = *u8` declaration. How is it handled? 

This represents a byte address to a `u8` value, so you need to convert the byte address to an element address + offset (i.e. the offset of the desired byte within the 4-byte chunk represented by any given element of Miden's linear memory).

The compiler has intrinsics for this, but obviously outside the compiler, it would be up to the procedure to do the right thing here.

The IR type system does have the notion of "address spaces", and we have an address space for Miden memory, which means that a pointer in the Miden address space represents an element address, not a byte address. I haven't added syntax for that here, but that might be worth doing actually.

With that in mind, I think I'd be inclined to represented pointer types as:

```
type ByteRef = ptr<u8> # or ptr<u8, addrspace(byte)>
type FeltRef = ptr<felt, addrspace(felt)>
```

In other words, `ptr<T>` is a simple pointer to a value of type `T` in byte-addressable space, while `ptr<T, addrspace(A)>` is a pointer to a value of type `T` in address space `A`.

> We could also consider specifying procedure signatures via attributes. 
> ```
> @inputs(a: u32, b: u32, c: Word)
> @outputs(d: u32, e: Word)
> @callconv(exec)
> pub.proc.foo.2
>     ...
> end
> ```
> 
> The main advantages are:
> 
> * This doesn't introduce too much of a new syntax and is very similar to how we already specify inputs/outputs in procedure comments.

Most of the new syntax actually isn't the procedure type signature itself, but in declaring types at all - so I'm not sure we save much here.

> * It allows us to name the outputs.

You can do this using types as well, for example, to riff off your example above:

```
pub proc foo.2(a: u32, b: u32, c: Word) -> struct { d: u32, e: Word }
   ...
end
```

Obviously you don't have to inline the struct type like I did above, but I'm mostly just trying to demonstrate that the ability to name your output values exists in both approaches.

> The main disadvantage is that it is different from traditional ways of specifying procedure signatures and is probably a bit more difficult to read.

A more practical issue is a technical one, in that we do not currently support representing types as attribute metadata, nor extracting types from attribute metadata - and that's more complex implementation-wise than just parsing first-class syntax for type signatures.

I personally find the aesthetics of first-class syntax a lot nicer to read, and more ergonomic, but that's just my personal preference.

Al-Kindi-0: Personally, I don't see a significant issue with change brought by the new syntax in terms of switching effort. Also, given the remarks of @bitwalker , it seems that the tradeoff is on the side of the new syntax.
One additional point, I think, is that it seems to me that we should make judicious use of the (limited) "real-estate" above the function/procedure, because, otherwise, code can turn quite quickly into something not aesthetically pleasing, though this might be just a subjective view from my part.

bobbinth: > I think, is that it seems to me that we should make judicious use of the (limited) "real-estate" above the function/procedure, because, otherwise, code can turn quite quickly into something not aesthetically pleasing, though this might be just a subjective view from my part.

Agree with this, thought in many cases we'd probably end up with quite a few inputs/outputs and they won't fit on a single line. For example, in `miden-lib` we have `miden::account::set_map_item` procedure that looks like so:
```
#! Sets a map item in the account storage.
#!
#! Inputs:  [index, KEY, VALUE]
#! Outputs: [OLD_MAP_ROOT, OLD_MAP_VALUE]
#!
#! Where:
#! - index is the index of the map where the KEY VALUE should be set.
#! - KEY is the key to set at VALUE.
#! - VALUE is the value to set at KEY.
#! - OLD_MAP_ROOT is the old map root.
#! - OLD_MAP_VALUE is the old value at KEY.
#!
#! Panics if:
#! - the index for the map is out of bounds, meaning > 255.
#! - the slot item at index is not a map.
#!
#! Invocation: exec
export.set_map_item
    ...
end
```
The new syntax for this would look something like this:
```
pub proc set_map_item(index: u8, key: Word, value: Word) -> (old_map_root: Word, old_map_value: Word)
    ...
end
```
And maybe we'd want to split this into multiple lines anyways - maybe something like:
```
pub proc set_map_item
    (index: u8, key: Word, value: Word) 
    -> (old_map_root: Word, old_map_value: Word)
    ...
end
```
There are probably other ways to format it - but I'm not sure they would look much better.

-----

An alternative syntax, without using attribute notation could be something like:
```
pub proc set_map_item:
    in: index: u8, key: Word, value: Word
    out: old_map_root: Word, old_map_value: Word
begin
    ...
end
```

Al-Kindi-0:
@bobbinth fair point, though for the example you provided I think that a single line is still acceptable.

For situations where we have very complex signatures, we would probably just use the same conventions used by the Rust formatter, with maybe even some warnings about complex return types and too many arguments extreme cases for some notion of "extreme" TBD, if we think it is a good idea.

assembly-syntax/src/parser/grammar.lalrpop
Al-Kindi-0: @bobbinth fair point, though for the example you provided I think that a single line is still acceptable.

For situations where we have very complex signatures, we would probably just use the same conventions used by the Rust formatter, with maybe even some warnings about complex return types and too many arguments extreme cases for some notion of "extreme" TBD, if we think it is a good idea.

Al-Kindi-0: Also, thinking about your last proposal, I think it might be better to stay closer to Rust than to Python, especially given the smart-contract language design and our relation to Rust in general, though again this might be more subjective than substantial.

PhilippGackstatter: Just to summarize my opinion:
- I agree that we should end up with having one way to do something, e.g. either `pub` or `export`, either `fn` or `proc` and either `.` or space as a separator.
  - I think the plan to phase one of these out over time sounds great.
- Agreed with staying closer to Rust than Python :) Just a preference towards how Rust looks.
  - This includes a preference for `fn` over `proc`, `pub` over `export` and space over `.` separators.
- I'd prefer requiring that parameter names are present, just as a way to document things. We already specify the `Inputs` and `Outputs` to a procedure in MASM docs anyway, and if we have this as part of the procedure declaration, then we may be able to get rid of the doc part of this, which seems like a benefit.
- I think the `rustfmt`-style of the `set_map_item` example would probably look like the following, which seems fine (assuming it can be parsed correctly):

```
pub proc set_map_item(
  index: u8, key: Word, value: Word
) -> (old_map_root: Word, old_map_value: Word)
    ...
end
```
- I would prefer that over the annotation style.

bitwalker: I believe `rustfmt` would do something more like:

```
pub proc set_map_item(
  index: u8, 
  key: Word, 
  value: Word,
) -> (
  old_map_root: Word, 
  old_map_value: Word
)
    ...
end
```

But it is one of those things that depends on the `rustfmt` config being used, the width of the line, etc.

bobbinth: To first address relatively simple items:

- I'm fine with moving away from using the `.` as a substitute for the whitespace. It does feel a bit more "natural" to me, but that's probably just my familiarity.
- For `proc` vs. `fn`, I have a slight preference for sticking with `proc` mostly because it would mean we don't need to comb through the existing docs and rename things.
- Agreed that naming of input parameters should be required, and I'd probably make naming of output paramters required too.
- I think for this PR, we can make the changes "additive" (i.e., not break existing MASM), but I would follow up with a PR which introduces breaking changes pretty soon. These breaking changes would include things like getting rid of `export` keyword, removing `.` separators for declarations, and maybe a few more.

> Taking the "address" of a procedure for use with `dynexec`/`dyncall` requires that no more than 15 elements of the operand stack are required to hold its arguments/results, since one element is required to hold the address of the callee hash.

For `dynexec`, there is actually no restriction - we could pass as many elements via the stack as needed (this is similar to `exec`).

For `dyncall`, I think we made it so that 16 elements (following the address) would be accessible to the callee. @plafer can correct me if I'm wrong here.

-----

> ```
> pub proc set_map_item(
>   index: u8, 
>   key: Word, 
>   value: Word,
> ) -> (
>   old_map_root: Word, 
>   old_map_value: Word
> )
>     ...
> end
> ```

This does feel a bit like mixing of different styles - i.e., the declaration is Rust-like, but then we have the `end` keyword. If we do want to go with Rust-like syntax, we should probably go all the way and add curly braces.
```
pub proc set_map_item(
    index: u8,
    key: Word,
    value: Word.
) -> (
    old_map_root: Word,
    old_map_value: Word.
) {
    ...
}
```
Though, personally, I like that syntaxes are different enough (makes context switching between Rust and MASM a bit simpler for me). And maybe going with something like this is a bit better:
```
pub proc set_map_item(
    index: u8,
    key: Word,
    value: Word.
) -> (
    old_map_root: Word,
    old_map_value: Word.
)
begin
    ...
end
```

I do think that syntax for specifying the number of procedure locals would look a bit odd in both cases - maybe there is another way to handle it? (e.g., would we want to specify types for locals too somehow - though, maybe an overkill for now).

------

> > In the `miden-lib` code we frequently annotate procedures (for now just in the comments) as `Invocation: exec`, `Invocation: dynexec`, `Invocation: call`, `Invocation: syscall` - how would we express this using `@callconv` attribute?
> 
> You don't, at least, that's not what calling convention means in this case. The convention for a procedure is distinct from the instruction used to invoke it, because we cannot control how a procedure is invoked in general. Instead, there are rules that must hold at call sites for any given procedure:

Maybe this is not for this PR, but I think we should be able to specify what is the intended invocation method for a given procedure. For example, if a procedure is meant to be invoked via `call` or `dyncall` and someone is trying to invoke it via `exec`, the assembler should, at the very least, show a warning (and I think now we'll have enough info to do that in most cases - i.e., excluding innovations by hash).

We could do it via a different attribute, but it does feel like there is some overlap with the `@callconv`. Specifically:

- It seems like `fast` is basically synonymous with `exec` and `dynexec` invocations.
- And `C` is very similar to what we'd do for `call` or `dyncall` - though, there may be some differences.

Though, maybe it does make sense to keep things separate because a procedure that is meant for `exec` or `dynexec` could pass parameters by reference too. But a procedure that is meant for `call` or `dyncall` needs to pass "complex" parameters via the advice provider.

----------------

> I think we need to ensure that all of our type layout is done with a consistent, lowest-addressed-bytes-first approach. Migrating the current implementation of the transaction kernel is a big lift though, so we'll need some kind of short-term solution for binding against the transaction kernel from compiler-generated code. I'm still investigating this.

Yes, agreed that switching to the "lowest-addressed-bytes-first" is the goal, but we can't do immediately. This will require significant changes in the transaction kernel, the VM, and the hash function we use. As discussed offline, we have a few options - none ideal:

1. We could write wrappers around all publicly exposed procedures from the transaction kernel, stdlib, and miden protocol library. These wrappers would reverse the words on the stack so that we can express them in the type system. This wouldn't be too difficult to do, but the maintenance may become an issue.
2. We could build some "pre-processing" step that would allow us to use something like `BeWord` to describe the types as they currently are, and the "pre-processor" would generate the wrappers somehow automatically. This would simplify maintenance, but writing (and debugging) such a tool will take some time.
3. We could define a special struct `BeWord` that the compiler will treat somehow specially. This would be the least amount of work for the existing MASM code, but may create complications in the compiler.

To elaborate on the 3rd approach above. Let's say as a part of https://github.com/0xMiden/miden-vm/issues/2125 we introduce `mem_loadw_be` / `mem_loadw_le` and `mem_storew_be` / `mem_storew_le`. For almost everything, the compiler would use `mem_loadw_le` / `mem_storew_le`. But could we create a special struct (via annotation or otherwise), that will always consist of 4 elements (we could call this struct `BeWord`), and then when the compiler needs to save this struct to memory (or read it from memory) it would use `mem_storew_be` and `mem_loadw_be`?

Here, I'm assuming the complier always knows the type of the struct it is dealing with when moving it to/from memory - but I'm not sure that's true.

bitwalker: Just a few follow ups/responses to various points here:

* I'll preserve `proc` vs `fn`
* I don't think we should use braces for blocks, and `begin` already has special significance (and I think is overly verbose for the purpose proposed here anyway), _but_ we could use `do` to indicate the start of a block if we feel that a start-of-block token is needed. This has prior art in other languages, e.g. Elixir, Ruby.
* While I'm very much keen on requiring naming of arguments, naming results feels a bit out of place to me, as the vast majority of procedures are likely to only produce a single result or none. I think it would probably overburden the syntax unnecessarily to support/require named results. I think this is also consistent with what most developers expect with regard to function definition syntax, but that's somewhat secondary. Open to additional discussion on this though.
* I think we could define calling conventions in terms of their relationship to `call` vs `exec`, but we will need to ensure that the specification of `call`-oriented conventions allows for both `call` and `dyncall`, likewise for `exec` vs `dynexec` (i.e. the convention should work whether the procedure is invoked statically or dynamically). We could then verify any static call sites adhere to the correct convention. However, it is not _in general_ possible for us to verify all call sites, because of things like the ability to invoke a procedure using its MAST root, and even dynamic procedure invocations may occur far away from the point at which the original `procref` takes the address. Most of the existing calling conventions are unused, except the Wasm-oriented ones, which are quite important, and I could see value in revisiting them to provide a better set of conventions that are all actually used.
* This PR now provides the syntax support for the new `@bigendian` struct representation, which will be used by the compiler-generated Rust bindings to bind against uses of that type in procedure type signatures. I see this as a temporary workaround, but it should be a stable one for the time being.

bobbinth: > I don't think we should use braces for blocks, and `begin` already has special significance (and I think is overly verbose for the purpose proposed here anyway), _but_ we could use `do` to indicate the start of a block if we feel that a start-of-block token is needed. This has prior art in other languages, e.g. Elixir, Ruby

I do think it would be good to have the "start of block" keyword, and I still have a slight preference for using `begin` here (there is also some [prior art](https://en.wikipedia.org/wiki/Pascal_(programming_language)#Procedures_and_functions) for this) - but maybe `do` is fine too. Curious what others think on this.

> While I'm very much keen on requiring naming of arguments, naming results feels a bit out of place to me, as the vast majority of procedures are likely to only produce a single result or none. I think it would probably overburden the syntax unnecessarily to support/require named results. I think this is also consistent with what most developers expect with regard to function definition syntax, but that's somewhat secondary. Open to additional discussion on this though.

I think it's fine to make naming of outputs optional - but I think we should have this option. In the current docs for the standard library and the protocol library procedures we almost always specify output names, and if we don't have this ability, we'd still need to maintain auxiliary doc comments that describe the outputs (e.g., see [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-lib/asm/miden/account.masm#L301-L302) and [here](https://github.com/0xMiden/miden-base/blob/next/crates/miden-lib/asm/miden/account.masm#L377)).

> I think we could define calling conventions in terms of their relationship to `call` vs `exec`, but we will need to ensure that the specification of `call`-oriented conventions allows for both `call` and `dyncall`, likewise for `exec` vs `dynexec` (i.e. the convention should work whether the procedure is invoked statically or dynamically). We could then verify any static call sites adhere to the correct convention.

I believe `call` and `dyncall` have the same invocation semantics - i.e., the same procedure can be invoked via `call` and `dyncall` and no special handling is needed in either case. This applies `exec` and `dynexec`. @plafer - could you confirm?

> However, it is not _in general_ possible for us to verify all call sites, because of things like the ability to invoke a procedure using its MAST root, and even dynamic procedure invocations may occur far away from the point at which the original `procref` takes the address. Most of the existing calling conventions are unused, except the Wasm-oriented ones, which are quite important, and I could see value in revisiting them to provide a better set of conventions that are all actually used.

Yes, understood. I think initially we could handle only static invocations and leave dynamic ones for the future (e.g., we could check dynamic invocations at runtime assuming debug info is enabled).

PhilippGackstatter: > I do think it would be good to have the "start of block" keyword

I'm a bit on the fence about this. The syntax looks fine or verbose, depending on the concrete example:

Here it looks overly verbose and better without it:
```
pub proc set_item(index: u8, value: Word) -> Word
begin
    swap drop add
end
```

And here it looks mostly fine:

```
pub proc set_map_item(
    index: u8,
    key: Word,
    value: Word,
) -> (
    old_map_root: Word,
    old_map_value: Word,
)
begin
    ...
end
```

I think this applies regardless of whether it's `begin` or `do`. I'm leaning a bit more towards not requiring it, both because I think it will look fine most of the time, and it seems more pleasant to write less verbose code. Whenever we can format the procedure in one line, or the return values on one line it should be fine, and I would be okay with accepting the slightly odd look in rarer cases like `set_map_item`. Since this seems quite formatting specific to me, maybe this is also something that can be improved through the way we format this. We have a formatter now, so that's something we can do.

assembly/src/linker/mod.rs
```diff
@@ -104,6 +104,16 @@ impl ModuleLink {
...
         }
     }
+
+    /// Resolves a user-expressed type, `ty`, to a concrete type
+    pub fn resolve_type(&self, ty: &ast::TypeExpr) -> Option<types::Type> {
+        match self {
+            Self::Ast(module) => module.resolve_type(ty),
+            // TODO(pauls): Ideally we ship user-defined types in packages so that we can
+            // reference them from dependencies, but we're punting on that for now.
+            Self::Info(_) => None,
```
bobbinth, line 114: Should we create a follow-up issue for this? In general, what does this limitation imply? (i.e., users can still define types but they cannot be exported from packages? Or something else?)

assembly-syntax/src/parser/grammar.lalrpop
bitwalker: There is actually a concrete reason we cannot use `begin` for this purpose - it is ambiguous whether it is part of the definition of a procedure, or starts the definition of a program entrypoint. For example:

```
pub proc foo|
            ^-- cursor is here

begin
    exec.foo
end
```

We're in the middle of defining the `foo` procedure in our editor, but it is incomplete - we want a syntax error to indicate that the body of `foo` is missing, be able to recover from that parse error by implicitly closing the definition with an empty body, and continuing - but instead, the source file here would parse successfully because it just looks like a valid procedure definition, even though it is merging together two definitions that should be separate.

If we strongly feel we need something to indicate start-of-block, I think `do` is the way to go here - it is far more common for this purpose than `begin` (and less verbose), and would keep the grammar unambiguous. Note that we would also need to use `do` with the control flow instructions, e.g. `if.true do`/`while.true do`/etc.

assembly/src/linker/mod.rs
```diff
@@ -727,6 +737,71 @@ impl Linker {
         resolver.resolve_target(caller, target)
     }
 
+    /// Resolves a user-defined type `signature` to concrete types
+    pub(super) fn resolve_signature(
+        &self,
+        gid: GlobalProcedureIndex,
+    ) -> Result<Option<Arc<types::FunctionType>>, LinkerError> {
```
bobbinth: Is the doc comment here correct? Seems like this looking up signature for a procedure rather than converting `signature` to a concrete type.

assembly/src/linker/mod.rs
```diff
@@ -727,6 +737,71 @@ impl Linker {
...
+                        } else {
+                            let span = result.span();
+                            return Err(LinkerError::UndefinedType {
+                                span,
+                                source_file: self.source_manager.get(span.source_id()).ok(),
+                            });
+                        }
+                    }
+                    Ok(Some(Arc::new(types::FunctionType::new(cc, args, results))))
+                },
```
bobbinth: nit: maybe this could be converted into a helper method to improve readability a bit? Maybe something like:
```rust
impl Linker {
    fn resolve_function_types(
        &self,
        sig: types::FunctionType,
        module: &Module,
    ) -> Result<types::FunctionType, LinkerError> {
        ...
    }
}
```

assembly-syntax/src/ast/type.rs
```diff
@@ -52,6 +54,348 @@ impl From<EnumType> for TypeDecl {
...
+}
+
+/// A procedure type signature
+#[derive(Debug, Clone)]
+pub struct FunctionType {
+    pub span: SourceSpan,
+    pub cc: types::CallConv,
+    pub args: Vec<TypeExpr>,
+    pub results: Vec<TypeExpr>,
+}
```
bobbinth, line 79: Given how we use it, would a more appropriate for `FunctionType` be `FunctionSignature`? Or maybe even `ProcedureSignature`?

assembly-syntax/src/ast/module.rs
```diff
@@ -553,6 +577,33 @@ impl Module {
     pub fn import_paths(&self) -> impl Iterator<Item = &LibraryPath> + '_ {
         self.imports.iter().map(|import| &import.path)
     }
+
+    /// Resolves a user-expressed type, `ty`, to a concrete type
+    pub fn resolve_type(&self, ty: &ast::TypeExpr) -> Option<types::Type> {
```
bobbinth, line 582: Question: is this basically to resolve type aliases? Or is it doing more than that?

assembly-syntax/src/parser/grammar.lalrpop
bobbinth, line 1: Not for this PR, but this file is pretty huge. I wonder if there is a way to split this up into multiple logical parts.

assembly/src/linker/mod.rs
bitwalker: Yep! Basically, MASM allows for assigning names to types for ergonomic convenience, but this is not a feature of the HIR type system (and doesn't really make sense to support there). If we wanted to allow importing these type definitions from dependencies, we'd need to store user-defined type aliases in the metadata of a `Library`/`Package`, so that we know what aliases exist and are available for import.

Secondly, we'd need to add support for importing type aliases like we do procedures. We do not currently support this, but could.

assembly/src/linker/mod.rs
bitwalker: Ah, it's a bit off, but the gist is correct - we're looking up the AST type signature and resolving it to its equivalent HIR type signature. I'll update it.

assembly-syntax/src/ast/type.rs
bitwalker: Currently these structs are basically AST equivalents of the underlying HIR type system structs, so I'd prefer to keep them named this way, but it's not super important. I have a habit of referring to `FunctionType` as both a signature and a type interchangeably, and there isn't really a difference - a function type is only complete if you also know its calling convention, and that overlaps with what a signature is anyway.

assembly-syntax/src/ast/module.rs
bitwalker: It's converting from the AST `TypeExpr` to HIR `Type`, but that's basically a 1:1 conversion, with the exception of type aliases (which don't exist in HIR). So this method is really just about resolving the aliases so that it can be converted to the equivalent HIR representation - but it also does the conversion since we perform both steps at the same time.

assembly-syntax/src/parser/grammar.lalrpop
bitwalker: Unfortunately the LALRPOP grammar has to be defined in a single source file and can't be broken up - but that's a pretty common problem with parser generators.

When I get around to rewriting the parser for better IDE integration, we'll be able to organize things much more cleanly, but for now we're sort of stuck with this as-is.

assembly-syntax/src/ast/type.rs
```diff
@@ -52,6 +57,454 @@ impl From<EnumType> for TypeDecl {
...
+
+// POINTER TYPE
+// ================================================================================================
+
+#[derive(Debug, Clone)]
+pub struct PointerType {
+    pub span: SourceSpan,
+    pub pointee: Box<TypeExpr>,
+    addrspace: Option<AddressSpace>,
+}
```
bobbinth, line 240: Not for this PR, but would be good to add doc comments for this struct (as well as other structs in this module). One thing that is not immediately clear is what is the default address space (after looking further in the code, I think it is `Element`).

Also, nit: why make `addrspace` optional? If we assume that the default is `Element` - might as well just set it at construction time.

bobbinth (approved):
Looks good! Thank you! I left one small comment for the future inline.

Also, copying some of the times for the future from https://github.com/0xMiden/miden-vm/pull/2120/#pullrequestreview-3245226573, I think we need to:

1. Update documentation - both mdBook (probably requires a new section) and probably some inline docs as well.
2. Finalize "start of block" syntax (e.g., begin vs. do vs. nothing).
3. Finalize calling convention options.
4. Remove "temporary" syntax - e.g., export, . notation etc.
5. In the longer term, we should refactor the parser definition to make it more modular.
6. Anything else?

assembly-syntax/src/ast/type.rs
bitwalker: It's optional syntax, but the actual address space is always known - this type is an AST structure, so it reflects the syntax used.

As for why we might allow the syntax to be optional - it makes for much more succinct type signatures when pointers are involved, e.g. `ptr<T>` instead of `ptr<T, addrspace(felt)>`

PR #2121, opened 2025-09-02
Fix `OpBatchAccumulator`

plafer:
I believe we missed a bug with #2094. I currently only have the failing tests, and will work on a fix tomorrow.

Please confirm though that the test fixes are appropriate.


plafer:
Please disregard - I was confused.

huitseeker:
@plafer one trick that may help: the test-only method [group_chunks](https://github.com/0xMiden/miden-vm/blob/cb5cfe0b179a3b668cd497949701ef349a913eb3/core/src/mast/node/basic_block_node/op_batch.rs#L91-L96) will give you the split of a batch into groups.

PR #2124, opened 2025-09-03
Improve Makefile Testing Infrastructure for AI Development

adr1anh:
## Overview

Restructures Makefile testing commands to provide an efficient two-phase development workflow optimized for AI-assisted development and rapid iteration.

## Problem

- Broken commands: test-loom and test-package were failing
- Invalid references to non-existent test-release profile
- No clear workflow for rapid iteration vs detailed debugging
- Different packages need different feature combinations

### New Cargo Profiles

Added two optimized profiles:

- `test-fast`: opt-level=2, no debug symbols, thin LTO for speed
- `test-dev`: full debug info, rich stack traces for debugging

### Improved `make test-fast`

Uses more optimized `test-fast` profile, with the goal being of detecting which tests are failing across the workspace.

### Package-Specific Test Commands

Created individual commands for each package with appropriate features. The `test-dev` configuration enables better error reporting. Use these to fix specific failing tests detected by `make fast`

```
make test-air           # --features testing
make test-processor     # --features concurrent,testing
make test-stdlib        # --features with-debug-info
make test-miden-vm      # --features concurrent,executable,metal,internal
```

All commands support optional filtering:
```
make test-processor test="decoder"            # Tests matching pattern
make test-processor test="fast::tests::all_ops"  # Specific function
```

### Two-Phase Workflow (for agents)

**Phase 1: Fast Iteration**

`make test-fast  # Quick feedback across all packages`
- Uses optimized test-fast profile for speed

**Phase 2: Targeted Debugging**

`make test-processor [test=<pattern>]  # Deep debugging`
- Uses test-dev profile with full debug info
- Optional test filtering for precise investigation

- Updated `CHANGELOG.md'

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,12 +30,26 @@ inherits = "release"
 codegen-units = 1
 lto = true
 
+# Test profile with full debug information for detailed debugging of test failures.
+# Use via `make test-PACKAGE` commands for investigating specific test failures.
+# Provides rich stack traces, source locations, and variable names in panics.
 [profile.test-dev]
 inherits = "dev"
 opt-level = 1
 debug = true
 debug-assertions = true
 overflow-checks = true
+
+# Fast test profile optimized for test execution speed over compilation speed.
+# Use via `make test-fast` for rapid iterative development feedback.
+# Prioritizes faster test runs at the cost of slightly slower compilation.
+[profile.test-fast]
+inherits = "test"
+opt-level = 2          # Higher optimization for faster test execution
+debug = false          # No debug symbols for smaller binaries
+debug-assertions = true
+overflow-checks = true
+lto = "thin"           # Thin LTO for better optimization without full LTO cost
 
 [workspace.dependencies]
 # Workspace crates
--- a/Makefile
+++ b/Makefile
@@ -64,20 +64,52 @@ test-docs: ## Run documentation tests
 	cargo test --doc $(ALL_FEATURES_BUT_ASYNC)
 
 .PHONY: test-fast
-test-fast: ## Runs all tests with the debug profile
-	$(DEBUG_ASSERTIONS) cargo nextest run --features testing
+test-fast: ## Runs all tests quickly for rapid iterative development feedback
+	$(DEBUG_ASSERTIONS) cargo nextest run --cargo-profile test-fast --features testing,no_err_ctx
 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -30,13 +30,27 @@ inherits = "release"
...
+# Use via `make test-PACKAGE` commands for investigating specific test failures.
+# Provides rich stack traces, source locations, and variable names in panics.
 [profile.test-dev]
 inherits = "dev"
 opt-level = 1
 debug = true
 debug-assertions = true
 overflow-checks = true
 
+# Fast test profile optimized for test execution speed over compilation speed.
```
huitseeker: Both these profiles are noisy. They only differ on opt-level, debug, and lto. The test profile inherits dev, the debug-assertions are overridden to the same value as their default, the overflow-checks as well.

huitseeker (commented):
I think `test-loom` might need a feature tweak, here is the output I get running it:
https://gist.github.com/huitseeker/4a6895ac571244ba2c0bcbf64f88d6d4

Also, I think the proliferation of crate-specific targets calls for a [make pattern](https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html). Here's an example of what that might looks like:
https://gist.github.com/huitseeker/420e57333586c576201f75feb49a0bdc

To me that looks more maintainable.

bitwalker (changes requested):
One thing to be aware of with using different profiles for the two "phases" you describe @adr1anh - and that is that changing profile is going to recompile _all_ the crates in the workspace, which is almost as time consuming as just running the tests (maybe worse).

I would recommend that we use a single profile for running tests, using the `dev` profile, any changes we want to make for the test profile should probably just be made to the `dev` profile anyway (e.g. `opt-level = 1`, `lto = thin`), to make iteration faster (as no rebuild of the workspace is required after the first clean build). The `test-fast` target should then correspond to running non-integration tests (or alternatively, inexpensive tests that finish on the order of milliseconds). 

The most expensive test by far is the `cli_bundle_debug` test, but the CLI tests more generally are integration tests, and expected to be more expensive (though currently only the above-mentioned one is actually all that bad). That one test alone takes 30s on my machine, out of 43s total for `make test`, and the first 2k or so tests finish in just 6-8s.

bitwalker (approved):

PR #2108, opened 2025-08-29
Add ability to remove decorators from `MastForest`

bobbinth:
This PR implements `MastForest::strip_decorators()` function which removes all decorators from the underlying MAST forest. This is needed to help address https://github.com/0xMiden/miden-base/issues/1812. For this reason, the PR is against `main` and I'll make a patch release once this is merged (the changes are non-breaking).

This is not meant as a long-term solution and may change as a result of https://github.com/0xMiden/miden-vm/issues/1580.

PR #2127, opened 2025-09-04
refactor: behavior of MastNode instances as a trait, existing boilerplate macro-generated

huitseeker:
This PR implements a `MastNodeTrait` to gather key methods that were previously implemented as instance methods on the `MastNode` enum's variants, like `digest`, the `append_X` methods, the `Y_children` methods, and `domain`. This change leverages the `enum_dispatch` macro to make dispatching to those variants automatic, providing a more maintainable and extensible architecture.

## The Problem

Previously, methods related to node children, digest, domain, etc were implemented individually, and then led to match expressions on the `MastNode` enum which dispatched to conventional methods on each of the variants. This approach had several limitations:

1. **Implicit Logic**: All MastNode variants had a bunch of methods they were all required to implement, but by convention only, see for example #2108,
2. **Maintenance Burden**: Adding new node types required updating multiple match expressions, and similar dispatch logic was repeated across different operations,
3. **Extensibility Challenges**: Extending the system with new node types required touching existing dispatch code.

## Solution

We've moved these methods into the `MastNodeTrait` and used `#[enum_dispatch]` to automatically generate the boilerplate dispatch implementations for all node types. This preserves type safety, and involves no dynamic dispatch (`dyn`) and therefore pays none of the associated costs.

### Adding New Node Types

When adding a new MAST node type in the future:

1. Define the struct with its specific fields in its own file,
2. Implement `MastNodeTrait` for the new node type,
3. Add to the `MastNode` enum under a new variant name, profit.

The compiler will automatically ensure all trait methods are implemented, eliminating the need to update central match expressions, or modify existing dispatch logic.

### Example: Adding a hypothetical `ParNode`

```rust
pub struct ParNode {
    children: [MastNodeId; 4],  // 4 children in parallel
    // ... other fields ...
}

impl MastNodeTrait for ParNode {
    // ... other methods
    fn has_children(&self) -> bool { true }
    fn append_children_to(&self, target: &mut Vec<MastNodeId>) {
        target.extend_from_slice(&self.children);
    }
    fn remap_children(&self, remapping: &Remapping) -> Self {
        // Apply remapping to all 4 children
        // ...
    }
    fn domain(&self) -> Felt { Self::DOMAIN }
}

// Add to enum
#[enum_dispatch(MastNodeTrait)]
pub enum MastNode {
    // ... existing variants ...
    Par(ParNode),
}
```

### Adding new behaviors (methods)

1. add the method to the trait,
2. add the implementations to the trait implementations,
3. use the method on `MastNode`, profit.

## Technical details

- (trivia) This is a practical application of solving the expression problem in Rust, and old OP/FP problem originally by Phil Wadler in 1998 in a [java-genericity email](http://homepages.inf.ed.ac.uk/wadler/papers/expression/expression.txt). 
- The `MastNodeExt` trait can not be merged with `MastNodeTrait`, because it involves a generic return type, which makes it impossible to have an enum (`MastNode`) dispatch to several distinct instantiations of this type => the enum_dispatch macro is irrelevant for it.
- In a very similar situation, I have assumed that resorting to dynamic dispatch for pretty-print/display is OK from a performance PoV. This might not actually be the case, please let me know.

## TODO
- [ ] bike-shed on a proper name for `MastNodeTrait`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - [BREAKING] Implement support for `event("event_name")` in MASM (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - Improved representation of `OPbatches` to include padding Noop by default, simplifying fast iteration over program instructions in the processor ([#1815](https://github.com/0xMiden/miden-vm/issues/1815)).
 - Rename `program_execution` benchmark to `program_execution_for_trace`, and benchmark `FastProcessor::execute_for_trace()` instead of `Process::execute()` (#[2131](https://github.com/0xMiden/miden-vm/pull/2131))
+- Refactored `MastNode` to eliminate boilerplate dispatch code ([#2127](https://github.com/0xMiden/miden-vm/pull/2127)).
 
 ## 0.17.1 (2025-08-29)
 
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -35,7 +35,7 @@ miden-debug-types.workspace = true
 miden-formatting.workspace = true
 num-derive = { version = "0.4", default-features = false }
 num-traits = { version = "0.2", default-features = false }
-enum_dispatch = "0.3.12"
+enum_dispatch = { version =  "0.3" }
 thiserror.workspace = true
 winter-math.workspace = true
 winter-utils.workspace = true
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -35,6 +35,7 @@ miden-debug-types.workspace = true
 miden-formatting.workspace = true
 num-derive = { version = "0.4", default-features = false }
 num-traits = { version = "0.2", default-features = false }
+enum_dispatch = "0.3.12"
 thiserror.workspace = true
 winter-math.workspace = true
 winter-utils.workspace = true
--- a/core/src/mast/node/mod.rs
+++ b/core/src/mast/node/mod.rs
@@ -6,6 +6,7 @@ pub use basic_block_node::{
     BATCH_SIZE as OP_BATCH_SIZE, BasicBlockNode, GROUP_SIZE as OP_GROUP_SIZE, OpBatch,
     OperationOrDecorator,
 };
+use enum_dispatch::enum_dispatch;
 
 mod call_node;
 pub use call_node::CallNode;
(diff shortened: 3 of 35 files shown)


core/src/mast/node/mod.rs
```diff
@@ -33,9 +34,49 @@ use crate::{
     mast::{MastForest, MastNodeId, Remapping},
 };
 
+#[enum_dispatch]
+pub trait MastNodeTrait {
```
plafer: We should merge the existing `MastNodeExt` trait into this one - and perhaps use that name instead? It's not exactly an `Ext` in the conventional sense, but probably still better than `MastNodeTrait`?

huitseeker: See the PR description, section Technical details, second paragraph, and TODO. Thanks for the name suggestion! Happy to go with it if no objections, would rename `MastNodeExt` to something more specialized instead.

plafer: Ah sorry about that, apparently I skipped over that section. I would agree with calling this one `MastNodeExt`, and renaming the current `MastNodeExt` to something more "error/debug" related (since its only current use is in `ErrorContextImpl`).

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - [BREAKING] `get_stack_word` uses element-aligned indexing instead of word-aligned indexing (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - [BREAKING] Implement support for `event("event_name")` in MASM (#[2068](https://github.com/0xMiden/miden-vm/issues/2068)).
 - Improved representation of `OPbatches` to include padding Noop by default, simplifying fast iteration over program instructions in the processor ([#1815](https://github.com/0xMiden/miden-vm/issues/1815)).
+- refactored `MastNode` to eliminate boilerplate dispatch code ([#2127](https://github.com/0xMiden/miden-vm/pull/2127)).
```
Al-Kindi-0: nit: Refactored

core/Cargo.toml
```diff
@@ -35,6 +35,7 @@ miden-debug-types.workspace = true
 miden-formatting.workspace = true
 num-derive = { version = "0.4", default-features = false }
 num-traits = { version = "0.2", default-features = false }
+enum_dispatch = { version =  "0.3.12" }
```
Al-Kindi-0: nit: "0.3"

PR #2129, opened 2025-09-04
`FastProcessor::execute_*()` methods now return an `ExecutionOutput`

### LINKED ISSUE #2052: Allow access to `FastProcessor` memory after execution
If I understood correctly, the idea is to get rid of the `Process` in favor of `FastProcessor` mid-term (?). In miden-base we'd need one additional thing to move away from the `Process`. The current `Process` allows access to the memory after `execute` (because it takes `&mut self`) but `FastProcessor::execute` takes `self` and so it wouldn't be possible to execute a program and then access memory afterward. The primary use case is to assert the [memory layout of the transaction kernel](https://github.com/0xMiden/miden-base/blob/72c432012abf5bc2892c0038f92d2c165022362d/crates/miden-testing/src/kernel_tests/tx/test_prologue.rs#L153-L160).

So the request here would be to have some way to access a process' memory after `execute`. What that API looks like exactly doesn't seem important, e.g. whether it's changing `execute` to not consume `self` or to return the memory state as part of `execute` or something else entirely.

Note: This is not time-critical, just something I noticed and wanted to document. As long as `Process` exists, we can continue to use it in test code.

plafer:
Closes #2052


processor/src/fast/mod.rs
```diff
@@ -684,6 +686,21 @@ impl FastProcessor {
...
+// EXECUTION OUTPUT
+// ===============================================================================================
+
+/// The output of a program execution, containing the state of the stack, advice provider, and
+/// memory at the end of the execution.
+#[derive(Debug)]
+pub struct ExecutionOutput {
+    pub stack_outputs: StackOutputs,
+    pub advice: AdviceProvider,
+    pub memory: Memory,
```
bobbinth: nit: I would probably rename `stack_outputs` to just `stack` as the "outputs" part is implied by the struct name as well as field type. Also, this would be more consistent with other fields (e.g., we don't say `advice_outputs`).

bobbinth (approved):
Looks good! Thank you! I left one small nit inline.

huitseeker (approved):
👍

processor/src/fast/mod.rs

PR #2130, opened 2025-09-04
feat: implement new binary format, BITE

bitwalker:
Split off from #2071

This PR implements a prototype implementation of a new binary serialization format based on serde, which I've called BITE, see the miden-bite crate for more info.


bitwalker:
I don't believe this is needed any longer, so closing as stale

PR #2131, opened 2025-09-04
Create program_execution_for_trace benchmark

plafer:
Follow-up to https://github.com/0xMiden/miden-vm/pull/2023#pullrequestreview-3178901680

Introduces the `program_execution_for_trace` benchmark, which benchmarks `FastProcessor::execute_for_trace()` - the method used to generate trace fragments for parallel trace generation.

Note that we currently still use a trace fragment length of 1024, which negatively impacts the benchmark results. Nevertheless, the `blake3_1to1` benchmark sits at:

- `program_execution_fast`: 1.85ms
  - Representing 262 MHz
- `program_execution_for_trace` (1024 rows per fragment): 4.45ms
  - Representing 109 MHz
- `program_execution_for_trace` (5000 rows per fragment): 4ms
  - Representing 121 MHz


bobbinth (approved):
Looks good! Thank you!

The `blake3_1to1` benchmark runs for about 0.5M rows, right? If so, we'd get around 0.5ms for generating checkpoints a trace of $2^{16}$ steps and around 8ms for $2^{20}$ steps - which I think is quite nice.

Curious if multi-threaded trace generation can do the rest in under 50ms for the $2^{20}$ case - but we'll know this soon enough.

> `program_execution_for_trace` (5000 rows per fragment): 4ms
> * Representing 121 MHz

I would probably use powers of two for rows per fragment as this will split the trace into even chunks. My intuition is that we'll probably settle either at 4096 or 8192 for as the "default" fragment size.

PR #2132, opened 2025-09-05
feat: macro-generated serialization tests

huitseeker:
This PR against #2071 implements a `serde_test` macro that can generate round-trip serialization tests pegged on an implementation of `proptest::prelude::Arbitrary`. This macro:

- supports generic arguments passed as attributes,
- uses ~~`bincode`~~ `serde_json` as the test serializer (I expect this will be replaced by #2130)

=> no more boilerplate tests! (and future tests are easier to define)

## Note

Tests failing here are signal for #2071 (unless I messed up something in my `Arbitrary` impls).

## TODO (after merging this PR)
- [x] make sure the import of the `miden-serde-test-macros` crate is sparse and follows best practices re: features. (I've been fast & loose),
- [x] check CI / Makefile to ensure those feature-gated tests are run (i.e. the failure in the present PR's CI is still observed after tweaking features as per the above),
- [ ] have a better implementation of `Arbitrary` for `LibraryExports` (where the last field isn't always `None`), add one in missing places, e.g. `KernelLibrary`,
- [x] fix failing tests (see #2071 review)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,6 +159,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "bit-set"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


bitwalker:
Note that our use of serde requires a self-describing format to work properly, and `bincode` [is not](https://docs.rs/bincode/2.0.1/bincode/serde/index.html#known-issues). This was one of the main motivations behind the `miden-bite` design.

I'd suggest we either pick a binary format this is self-describing for these tests, or punt on merging this until `miden-bite` (or whatever format we land on there) is merged.

PR #2138, opened 2025-09-08
Introduce `EventID` type

adr1anh:
## Describe your changes

NOTE: based on the `keccak-precompile` branch while waiting to merge it.

The type of an event is a `Felt` though we often needed to convert it to a `u64` in order to use it as key for a map. 

The wrapper also makes it clearer when we are using an event id.

- Updated `CHANGELOG.md'

### DIFF
--- a/core/src/event_id.rs
+++ b/core/src/event_id.rs
@@ -1,8 +1,6 @@
 use core::fmt::{Display, Formatter};
 
-use miden_crypto::hash::blake::Blake3_256;
-
-use crate::Felt;
+use crate::{Felt, utils::hash_string_to_word};
 
 /// A type-safe wrapper around a [`Felt`] that represents an event identifier.
 ///
@@ -22,13 +20,15 @@ impl EventId {
     /// 3. Interpreting these bytes as a little-endian u64
     /// 4. Reducing modulo the field prime to create a valid Felt
     ///
+    /// Note that this is the same procedure performed by [`hash_string_to_word`], where we take
+    /// the first element of the resulting [`Word`](crate::Word).
+    ///
     /// This ensures that identical event names always produce the same event ID, while
     /// providing good distribution properties to minimize collisions between different names.
     pub fn from_name(name: impl AsRef<str>) -> Self {
-        let digest_bytes: [u8; 32] = Blake3_256::hash(name.as_ref().as_bytes()).into();
-        let event_bytes: [u8; 8] = digest_bytes[0..8].try_into().unwrap();
+        let digest_word = hash_string_to_word(name.as_ref());
 
-        Self(Felt::new(u64::from_le_bytes(event_bytes)))
+        Self(digest_word[0])
     }
 
     /// Creates a new event ID from a [`Felt`].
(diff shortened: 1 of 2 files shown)


core/src/event_id.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::Felt;
+
+/// A type-safe wrapper around a [`Felt`] that represents an event identifier.
+///
+/// Event IDs are used to identify events that can be emitted by the VM or handled by the host.
+/// This newtype provides type safety and ensures that event IDs are not accidentally confused
+/// with other [`Felt`] values.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct EventID(Felt);
```
bobbinth: nit: I would rename this into `EventId` - I think this follows Rust conventions better (and we follow a similar approach in other places - e.g., `AccountId`).

core/src/event_id.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// 3. Interpreting these bytes as a little-endian u64
+    /// 4. Reducing modulo the field prime to create a valid Felt
+    ///
+    /// This ensures that identical event names always produce the same event ID, while
+    /// providing good distribution properties to minimize collisions between different names.
+    pub fn from_name(name: impl AsRef<str>) -> Self {
+        let digest_bytes: [u8; 32] = Blake3_256::hash(name.as_ref().as_bytes()).into();
+        let event_bytes: [u8; 8] = digest_bytes[0..8].try_into().unwrap();
+
+        Self(Felt::new(u64::from_le_bytes(event_bytes)))
```
bobbinth: Should we use `hash_string_to_word()` here as a building block? Then, this could look like:
```rust
Self(hash_string_to_word(name.as_ref())[0])
```

processor/src/decoder/tests.rs
```diff
@@ -1522,7 +1522,8 @@ fn set_user_op_helpers_many() {
 fn build_trace(stack_inputs: &[u64], program: &Program) -> (DecoderTrace, usize) {
     let stack_inputs = StackInputs::try_from_ints(stack_inputs.iter().copied()).unwrap();
     let mut host = DefaultHost::default();
-    host.register_handler(EMIT_EVENT_ID, Arc::new(NoopEventHandler)).unwrap();
+    host.register_handler(EventID::from_felt(EMIT_EVENT_ID), Arc::new(NoopEventHandler))
```
bobbinth: nit: if instead of `EventId::from_felt()` we have `impl From<Felt> for EventId`, here, we may be able to do something like `EMIT_EVENT_ID.into()`.

bobbinth (approved):
Looks good! Thank you! I left just a couple of small nits inline.

processor/src/decoder/tests.rs
adr1anh: The `From` trait cannot be used in a `const` context, so we need a manual constructor.

For the others, `from_{name, u64}` I felt it was better to be explicit since we often create event IDs from constants and this makes it clear what their type is.

adr1anh: Though in this case we could make the event constant an `EventId` directly.

PR #2140, opened 2025-09-08
fix(core): normalize decorator index to u32 in MastNodeFingerprint

radik878:
## Describe your changes

- Use u32 for decorator operation index when hashing in MastNodeFingerprint.
- This prevents architecture-dependent fingerprints caused by usize::to_le_bytes() producing 4 or 8 bytes depending on pointer width.
- Other fingerprint inputs already use fixed-width types; this makes decorator-related fingerprints deterministic across 32/64-bit platforms.


bobbinth:

PR #2141, opened 2025-09-08
Introduce `Processor` trait

plafer:
Abstracts away the `FastProcessor` operation execution into a `Processor` trait, that will be shared with the parallel trace generators. 

The huge win that this PR presents is that we no longer duplicate the vast majority of the processing logic between the fast processor, and the parallel trace generators. In a future PR, we can investigate abstracting away the entire `FastProcessor::execute_impl()` logic (i.e. driving the continuation stack), where we would effectively have just about one abstract processor implementation.

Notes:
- #1839 has already been updated to use this trait, and all tests pass there as well.
- The `program_execution_fast::blake3_1to1` benchmark is as unaffected by this change.


processor/src/fast/basic_block.rs
```diff
@@ -0,0 +1,235 @@
...
+
+                // Corresponds to the RESPAN operation added to the trace.
+                self.increment_clk(tracer);
+            }
+
+            self.execute_op_batch(
+                basic_block_node,
+                node_id,
+                op_batch,
+                batch_index_minus_1 + 1,
```
Al-Kindi-0: nit: should we define this once at the start of the loop

processor/src/fast/memory.rs
```diff
@@ -214,3 +208,45 @@ fn enforce_word_aligned_addr(
...
+    }
+
+    fn read_word(
+        &mut self,
+        ctx: ContextId,
+        addr: Felt,
+        clk: RowIndex,
+        err_ctx: &impl ErrorContext,
+    ) -> Result<Word, MemoryError> {
+        Self::read_word(self, ctx, addr, clk, err_ctx)
```
Al-Kindi-0, line 229: nit: can we be consistent with the rest here

processor/src/host/advice/mod.rs
```diff
@@ -346,3 +346,44 @@ impl From<AdviceInputs> for AdviceProvider {
...
+        &mut self,
+        root: Word,
+        depth: &Felt,
+        index: &Felt,
+        value: Word,
+    ) -> Result<(MerklePath, Word), AdviceError> {
+        self.update_merkle_node(root, depth, index, value)
+
+        // TODO(plafer): return error if depth is invalid, equivalent to:
+        // assert_eq!(path.len(), depth.as_int() as usize);
```
Al-Kindi-0: Q: is this the right place for this?

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -0,0 +1,245 @@
...
+pub(super) fn op_hperm<P: Processor>(
+    processor: &mut P,
+    tracer: &mut impl Tracer,
+) -> [Felt; NUM_USER_OP_HELPERS] {
+    let state_range = range(MIN_STACK_DEPTH - STATE_WIDTH, STATE_WIDTH);
+
+    // Compute the hash of the input
+    let (addr, hashed_state) = {
+        let input_state: [Felt; STATE_WIDTH] =
+            processor.stack().top()[state_range.clone()].try_into().unwrap();
```
Al-Kindi-0: nit: should we use expect here instead?

Al-Kindi-0 (approved):
LGTM!
A few nits and a question from me.

processor/src/fast/memory.rs
plafer: I don't know why, but with `self.read_word()`, Rust sees it as an infinite loop (i.e. calling itself, the trait method, as opposed to the method defined directly on `Memory`). And we don't get that with e.g. `read_element()`, even though it is defined next to `read_word()` with the same visibility.

Would it be better to just use the `Self::method_name(self, ...)` syntax everywhere for consistency?

processor/src/host/advice/mod.rs
plafer: Yes, although currently unintuitive. In the main body of `op_mrupdate()`, we currently can't assume that we have a valid `MerklePath` (e.g. see how this method is implemented for `AdviceReplay`). So if we were to run that check there, it would always fail in the fast processor.

I wanted to investigate using `Option<MerklePath>` instead as the return type to more properly model this in a follow-up PR, but I might just do it in this one because I agree that it's confusing.

processor/src/processor/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+    ///
+    /// top                                                       bottom
+    /// v                                                           v
+    /// a | b | c | d | e | f | g | h | i | j | k | l | m | n | o | p
+    ///
+    /// Then
+    /// - `stack_get_word(0)` returns `[d, c, b, a]`,
+    /// - `stack_get_word(1)` returns `[e, d, c ,b]`,
+    /// - etc.
+    fn get_word(&self, start_idx: usize) -> Word;
```
bobbinth, line 131: Not for this PR, but in the context of recent discussions, it may make sense to start naming things more explicitly. For example, this could be `get_word_be()` and then later on we could also add `get_word_le()` that would return `[a, b, c, d]` for `get_word_le(0)`.

processor/src/processor/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+    fn depth(&self) -> u32;
+
+    /// Writes an element to the stack at the given index.
+    fn write(&mut self, idx: usize, element: Felt);
+
+    /// Writes a word to the stack starting at the given index.
+    ///
+    /// The index is the index of the first element of the word, and the word is written in reverse
+    /// order.
+    fn write_word(&mut self, start_idx: usize, word: &Word);
```
bobbinth: nit: would `set()` and `set_word()` be more consistent with `get()` and `get_word()`?

processor/src/processor/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+pub trait HasherInterface {
+    /// Applies a single permutation of the hash function to the provided state and records the
+    /// execution trace of this computation.
+    ///
+    /// The returned tuple contains the hasher state after the permutation and the row address of
+    /// the execution trace at which the permutation started.
+    ///
+    /// The address is only needed for operation helpers in trace generation, and thus an
+    /// implementation might choose to return a default/invalid address if it is not needed.
+    fn permute(&mut self, state: HasherState) -> (Felt, HasherState);
```
bobbinth, line 293: nit (and probably not for this PR): should we return `(RowIndex, HasherState)` from here instead?

processor/src/processor/mod.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// an implementation might choose to return a default/invalid address if it is not needed.
+    fn update_merkle_root(
+        &mut self,
+        claimed_old_root: Word,
+        old_value: Word,
+        new_value: Word,
+        path: &MerklePath,
+        index: Felt,
+        on_err: impl FnOnce() -> ExecutionError,
+    ) -> Result<(Felt, Word), ExecutionError>;
```
bobbinth, line 337: Same nit as above.

bobbinth (approved):
Looks good! Thank you! I mostly reviewed the overall structure - I'm assuming most of the code here is just copy&pasted.

processor/src/fast/basic_block.rs
plafer: Done

processor/src/host/advice/mod.rs
plafer: Added a commit that uses `Option<MerklePath>`, and moves the check outside

processor/src/processor/mod.rs
plafer: Probably good because it's more readable, but indeed let's keep it for a future PR

processor/src/processor/mod.rs
plafer: Done

processor/src/fast/memory.rs
Al-Kindi-0: Not a big deal in my opinion

PR #2152, opened 2025-09-13
refactor: MastNode quality of life improvements

huitseeker:
This checkpoints some uncontroversial quality-of-life improvements I've found on the way to #1776. I expect no behaviour changes from this PR. The commits each stand on their own, but to summarize:
- rename `DecoratorIterator`, `RawDecoratorIterator` to better reflect their `Self::Item`,
- removes instance methods when there is a trait method that's an exact replacement (often implemented using that instance method),
    - technically removes the `BasicBlockNode::(append|prepend)decorators` methods, happy to deprecate instead,
- uses `MastNodeErrorContext::decorators` instead of `BasicBlockNode::decorators`
    - technically removes the slice-returning `BasicBlockNode::decorators`, happy to deprecate instead,
- relies on `From<VariantX> for MastNode` instances instead of ad-hoc constructors,
    - the former are more familiar and do not require learning new APIs,
    - removes the `MastNode::new_variant_x` methods, happy to deprecate instead,
- uses ad-hoc destructors `MastNode::unwrap_x` where reimplemented

I'll mark this change as breaking in the CHANGELOG, however borderline (the changed functions aren't really used much).


core/src/mast/merger/mod.rs
```diff
@@ -411,27 +412,29 @@ impl MastForestRootMap {
...
+        }
     }
 
     /// Maps the given key to the given value.
     ///
     /// It is the caller's responsibility to only pass keys that belong to the forest for which this
     /// map was originally created.
     fn insert(&mut self, key: DecoratorId, value: DecoratorId) {
-        self.inner[key.as_usize()] = Some(value);
+        self.inner[key.as_usize()].set(value).expect("DecoratorIdMap initialized twice");
```
plafer: Can't we use `match &mut self.inner[key]` here instead, and keep the `Option`? I don't think `OnceCell` is necessary here, since we don't need interior mutability. Any specific other reason to use it?

huitseeker: No worries, the goal of this PR is to fly through review with non-controversial changes, this one is debatable, so it's out.

huitseeker:
Closing and re-opening because this will need stacking now.

PR #2156, opened 2025-09-15
fix: CI checks stdlib doc is rebuilt and changes checked in

### LINKED ISSUE #2155: CI check that library was rebuilt before merging
### Packages versions

c99d793db362525181a3c1f01bd5453e65752fb9

### Bug description

In https://github.com/0xMiden/miden-vm/pull/2153 we see an update of the stdlib's doc, but no accompanying rebuild of the library, reinstated in https://github.com/0xMiden/miden-vm/pull/2105

CI will in this case rebuild the library, but not check that the state of the local checkout is clean (i.e. that there are no uncommitted modified files in `stdlib/docs`, and no new untracked files there as well). Yet it should, since this PR being merged will lead to every dev checkout being rebuilt having new uncommitted files.

_No response_

### Relevant log output

```shell

```

huitseeker:
Checks that stdlib doc is committed.

Closes #2155


bobbinth (approved):
Looks good! Thank you!

PR #2157, opened 2025-09-15
Renumber SystemEvent IDs

### LINKED ISSUE #2151: Renumber system event IDs
Context:

> For all remaining system events, should we just assign simple values from 0 to 15?

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2145#discussion_r2342913811_

adr1anh:
## Describe your changes

Closes #2151

This PR refactors `SystemEvent` to use integers 0..15 (offset by 2^31 when mapping to `EventId`) rather than the current randomly chosen ones. 

- Updated `CHANGELOG.md'


core/src/sys_events.rs
```diff
@@ -261,50 +237,35 @@ pub enum SystemEvent {
...
-            SystemEvent::ILog2 => EVENT_ILOG2,
-            SystemEvent::MemToMap => EVENT_MEM_TO_MAP,
-            SystemEvent::HdwordToMap => EVENT_HDWORD_TO_MAP,
-            SystemEvent::HdwordToMapWithDomain => EVENT_HDWORD_TO_MAP_WITH_DOMAIN,
-            SystemEvent::HqwordToMap => EVENT_HQWORD_TO_MAP,
-            SystemEvent::HpermToMap => EVENT_HPERM_TO_MAP,
-        }
+impl From<SystemEvent> for EventId {
+    fn from(system_event: SystemEvent) -> Self {
+        EventId::from_u64(system_event as u32 as u64)
```
plafer: With this approach, we will no longer get a compile-time error when we add a system event, and therefore the `try_from()` implementation will be wrong but will compile.

I don't think there's a stable way to query the number of variants in an enum at compile-time (to `const_assert` on). So if we can't find a trick to get a compile-time error when that happens to remind us to update `try_from()`, the next best thing would be to add a comment where we're likely to see it (perhaps at the end of `SystemEvent`, after `HpermToMap`).

stdlib/docs/word.md
```diff
@@ -5,10 +5,10 @@
...
 | gte | Returns true if LHS is greater than or equal to RHS.<br /><br />Inputs:  [RHS, LHS]<br />Outputs: [is_lhs_greater_or_equal]<br /><br />Cycles: 118<br /> |
 | lt | Returns true if LHS is strictly less than RHS, false otherwise.<br /><br />The implementation avoids branching for performance reasons.<br /><br />From an implementation standpoint this is exactly the same as `word::gt` except it uses<br />`lt` rather than `gt`. See its docs for details.<br /><br />Inputs:  [RHS, LHS]<br />Outputs: [is_lhs_lesser]<br /><br />Cycles: 117<br /> |
-| lte | Returns true if LHS is less than or equal to RHS, false otherwise.<br /><br />Inputs:  [RHS, LHS]<br />Outputs: [is_lhs_less_or_equal]<br /><br />Cycles: 122<br /> |
+| lte | Returns true if LHS is less than or equal to RHS, false otherwise.<br /><br />Inputs:  [RHS, LHS]<br />Outputs: [is_lhs_less_or_equal]<br /><br />Cycles: 118<br /> |
```
plafer: I assume those were intended? Not sure why they show up in this PR though

adr1anh: This is due to #2158 (or rather the reason for the PR). The docs weren't updated properly.. This change should disappear when the we merge the fix.

core/src/sys_events.rs
adr1anh: I've added some guardrails inside `try_from` which will trigger a compilation failure if a new variant is not handled.

adr1anh:
> LGTM, except for the fact that with this approach we run the risk of people hardcoding their event IDs to these same "nice" values (avoid the MASM `emit.EVENT_NAME` syntax, e.g. using `push.3 emit` instead), and not understanding why their event is not called.

I've changed the encoding to use a fixed offset when mapping to an `EventId` (arbitrarily set to 2^31).

bobbinth:
> > LGTM, except for the fact that with this approach we run the risk of people hardcoding their event IDs to these same "nice" values (avoid the MASM `emit.EVENT_NAME` syntax, e.g. using `push.3 emit` instead), and not understanding why their event is not called.
> 
> I've changed the encoding to use a fixed offset when mapping to an `EventId` (arbitrarily set to 2^31).

I think it is fine to keep the "small" values here as if someone goes out of their way to use numeric constants (via `push`) instruction, they should really know what they are doing - and then it is kind of on them.

But thinking about it more, i wonder if we should have a "reserved" range for event IDs. That is, let's say we reserve 0 - 255 range for "system" events, and if someone tries to register an event handler for such an event ID, we'd return an error. This way, even if someone does try to use such values in custom event handlers, they wouldn't be able to do that.

plafer:
> I've changed the encoding to use a fixed offset when mapping to an EventId (arbitrarily set to 2^31).

This goes back to my original question of what exactly the purpose of this change is. Because if the purpose is helping us debug things with system events (such that we see small numbers instead of big ones), then this offset change doesn't address the problem. And if we're having big numbers again, I'd go for random ones instead - again, since I'm not sure what having consecutive ones achieves.

> But thinking about it more, i wonder if we should have a "reserved" range for event IDs. That is, let's say we reserve 0 - 255 range for "system" events, and if someone tries to register an event handler for such an event ID, we'd return an error. This way, even if someone does try to use such values in custom event handlers, they wouldn't be able to do that.

How would we detect that though? Basically I'd see the user do something like

```
const.MY_EVENT=3

... push.MY_EVENT emit ...
```

I don't know *why* someone would do that - perhaps they got annoyed at looking at big event IDs, or want to use powers of 2 (and do bit-related things on event IDs) - who knows. But *if* they do that for whatever reason, then things break - and I don't think they should.

bobbinth:
> How would we detect that though? Basically I'd see the user do something like
> 
> ```
> const.MY_EVENT=3
> 
> ... push.MY_EVENT emit ...
> ```
> 
> I don't know _why_ someone would do that - perhaps they got annoyed at looking at big event IDs, or want to use powers of 2 (and do bit-related things on event IDs) - who knows. But _if_ they do that for whatever reason, then things break - and I don't think they should.

Presumably, the user would also provide a handler for this event, which they would need to then register with the processor. In the `EventHandlerRegistry` methods, we can check if someone is trying to register an event handler with code between `0 - 255` and return an error.

This doesn't completely eliminate possibility of someone shooting themselves in the foot - i.e., they could write their own event handler registry struct rather than using the default one. But I would say that if someone does this, they should be well aware of all the invariants as this is pretty advanced usage scenario, and if we document things well, it shouldn't be a problem.

adr1anh:
IMHO, we shouldn't focus too much on the risks of users calling `emit` without an event immediate value. Using `push.ID emit` is an anti-pattern that we really should discourage, but did make it easier for the existing transaction kernels to be updated. Since an `event` type in masm can be treated as an arbitrary `Felt`, the users should really only be pushing event IDs which were originally defined using the `event("...")` syntax. 

There is an inherent brittleness around event IDs stemming from the decision to use a single Felt which we derive pseudo-randomly. I still think this is better than assigning magic "random numbers" to the system events as constants. Using increasing integers does indeed help somewhat for debugging, but to me the point of this PR is really about making the separation between system and user events more explicit.

I've reverted the change around the offset, while adding the extra guardrails that prevents users from registering events with reserved IDs.

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -309,7 +309,7 @@ Instructions for communicating with the host through events and tracing.
...
-| `emit`             | `[event_id, ...]` | `[event_id, ...]` | 1      | Emits an event using the `event_id` from the top of the stack. The stack remains unchanged as the event_id is read without consuming it. This instruction reads the event ID from the stack but does not modify the stack depth. Example: with `push.123` on stack, `emit` reads the event ID 123 and executes the corresponding event handler.                                                                                 |
+| `emit`             | `[event_id, ...]` | `[event_id, ...]` | 1      | Emits an event using the `event_id` from the top of the stack. The stack remains unchanged as the event_id is read without consuming it. This instruction reads the event ID from the stack but does not modify the stack depth. Example: with `push.1230` on stack, `emit` reads the event ID 123- and executes the corresponding event handler. Note that event IDs in the range `0..256` are reserved for system events.     |
```
bobbinth: Should `123-` be `1230`?

core/src/sys_events.rs
```diff
@@ -38,6 +13,7 @@ mod constants {
 /// All actions, except for `MerkleNodeMerge`, `Ext2Inv` and `UpdateMerkleNode` can be invoked
 /// directly from Miden assembly via dedicated instructions.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[repr(u8)]
 pub enum SystemEvent {
```
bobbinth, line 39: Should we explicitly set the enum values here? For example:
```rust
pub enum SystemEvent {
    MerkleNodeMerge = 0,
    MerkleNodeToStack = 1,
    ...
}
```

core/src/sys_events.rs
```diff
@@ -261,54 +237,65 @@ pub enum SystemEvent {
...
+            7 => Ok(SystemEvent::U32Ctz),
+            8 => Ok(SystemEvent::U32Clo),
+            9 => Ok(SystemEvent::U32Cto),
+            10 => Ok(SystemEvent::ILog2),
+            11 => Ok(SystemEvent::MemToMap),
+            12 => Ok(SystemEvent::HdwordToMap),
+            13 => Ok(SystemEvent::HdwordToMapWithDomain),
+            14 => Ok(SystemEvent::HqwordToMap),
+            15 => Ok(SystemEvent::HpermToMap),
+            other => Err(other),
```
bobbinth: This implicitly relies on the current ordering of enums. Giving explicit values to enums as per previous comment would make this a bit less error prone, but also maybe it is still worth to have constants which we use to set enum values and to parse the numerical values into enum variants.

core/src/sys_events.rs
```diff
@@ -261,54 +237,65 @@ pub enum SystemEvent {
...
+            | SystemEvent::U32Ctz
+            | SystemEvent::U32Clo
+            | SystemEvent::U32Cto
+            | SystemEvent::ILog2
+            | SystemEvent::MemToMap
+            | SystemEvent::HdwordToMap
+            | SystemEvent::HdwordToMapWithDomain
+            | SystemEvent::HqwordToMap
+            | SystemEvent::HpermToMap => {},
+        };
```
bobbinth: Could this be put into a test?

bobbinth (commented):
Looks good! Thank you! I left some comments inline.

Also, should we update the [default event handler](https://github.com/0xMiden/miden-vm/blob/018253cbb6fe1794476d6314ba4e5489a5f38d8f/processor/src/host/handlers.rs#L113) to make sure we return an error if someone tries to register a handler for `EventId` corresponding to a system event?

adr1anh:
Thanks for the feedback @bobbinth. I've addressed all remaining feedback. It should be good to go.

plafer:
> Using push.ID emit is an anti-pattern that we really should discourage

Is it though? From what I recall, we first moved the event ID on the stack due to #2068, which specifically asked to be able to do `push.EVENT_ID emit`.

But regardless, I agree that docs + `EventHandlerRegistry` checks should be good enough.

plafer (approved):
LGTM

adr1anh:
> Is it though? From what I recall, we first moved the event ID on the stack due to #2068, which specifically asked to be able to do `push.EVENT_ID emit`.

I was thinking specifically of `push.128 emit` which is a stopgap solution for reusing existing event numbers, until transitioning to the string based approach. You would still use the non-immediate syntax for cases when you want the event ID dynamically, usually when the event on the stack originated from an event. For example, you can select which event to call depending on a condition
`push.event("b").event("a").condition drop emit `.

PR #2159, opened 2025-09-15
Log deferred precompile calls in AdviceProvider

### LINKED ISSUE #2158: Log deferred precompile calls in AdviceProvider
The current precompile framework uses the AdviceMap to store the call-data for the precompile. This makes it impossible to track the order in which the precompiles were called, which means we need to keep track of the list some other way.

By introducing a field `deferred: Vec<(PrecompileId, Vec<Felt>)>`, the precompile handler could return an advice mutation which pushes the call to `deferred`. Once the proof is done, we will be able to either
- Add all calldata to the VM proof so the verifier can manually check each call.
- Generate a recursive proof for all these calls.

adr1anh:
## Describe your changes

Closes #2158

This PR adds a `deferred` field to the AdviceProvider for storing call-data for all precompiles which were called during a program execution. 

Previously, we were storing this data in the AdviceMap which made it impossible to recover the order of the entries. This behavior is necessary to be able to recover the digest of all logged precompile calls. 

The change is breaking because we now return the list of calls in `AdviceProvider::into_parts`. This is currently unused but will be necessary when we will want attach the witness data to the VM execution proof.

- Updated `CHANGELOG.md'


processor/src/host/advice/mod.rs
```diff
@@ -37,6 +42,7 @@ pub struct AdviceProvider {
     stack: Vec<Felt>,
     map: AdviceMap,
     store: MerkleStore,
+    deferred: Vec<(EventId, Vec<Felt>)>,
```
plafer, line 48: My issue with adding this field is that it is not first-class as the other ones, in that we don't have any MASM functions to access it. 

Can't we solve the precompile ordering some other way? What about storing the precompile call order under some other key (and still put the data in the advice map)? Baiscally right now we have a clean definition of the advice provider being 3 things, and ideally we wouldn't tack on more single-use stuff to it.

adr1anh: > My issue with adding this field is that it is not first-class as the other ones, in that we don't have any MASM functions to access it. 
>
> Can't we solve the precompile ordering some other way? What about storing the precompile call order under some other key (and still put the data in the advice map)? Baiscally right now we have a clean definition of the advice provider being 3 things, and ideally we wouldn't tack on more single-use stuff to it.

While not currently implemented, I think we could have MASM functions to push to the `deferred` list. Currently it's only being modified by the keccak handler, but we would make use of the new instruction when it comes to verifying signatures. Since the verification doesn't require any computation to be pushed on the advice stack, the handler would really just be pushing the signature to the deferred list, which could be done entirely at the assembly level. 

The less we consider precompiles "first-class" in the VM, the more complexity is pushed downstream, e.g. in the kernel and stdlib, which IMO leads to much more coupling across the stack and creates more places where soundness could break. While we are currently working through a solution that relies on the kernel, it's a stop-gap solution until we get the specialized VM running. When that happens, precompile management will have to take place in the VM, and we'll likely have other situations where we need to include precompile-related code.

I agree that we should try to keep concepts as self-contained as possible in general, and not place things places just because it is practical. I'll list out a couple of more reasons why I think it makes sense to include it here

- The `deferred` list is a log of all data pushed to the advice which the VM was not able to verify. In general, it should be empty because the VM is actually able to verify everything injected in the advice. It's a way of signaling that because a certain handler was called, the proof is now unsound unless the data in that list is externally verified.
- Using the map is awkward here, because we would have to assign a "special" slot which can only be used in the context of precompiles and because the entry should be an append-only list. I don't see an elegant way of enforcing this which makes things more brittle.
- The `deferred` list is really an append-only, so we'd want to make sure that it can only be updated in this way. 
- Using the AdviceProvider is the right layer of abstraction between the VM and kernel which allows the prover to easily create a proof of all deferred statements either for native verification (where we put additional data in the proof) or recursion. 

My main point here is that we **should** treat this list as "first-class" in the VM. Using the AdviceMap will be asking for awkward handling later down the road both for the VM and down-stream users of it. It really seems to me like this is the right place for it since it's tightly linked to pushing unverified data to the advice provider.

plafer: As discussed offline, this makes sense to me! I would just make sure that we clearly document that "deferred" refers to "deferred computations not verified", etc (as you just explained).

bobbinth, line 48: I think it makes sense to have specialized tracking of deferred computations here, but a couple of questions:

Would we want to make this a bit more specific to precompiles? For example, we could have something like a `PrecompileRequest` struct (a better name is probably needed):
```rust
pub struct PrecompileRequest {
    label: Word, // or maybe some specialized type
    calldata: Vec<Felt>,
    // maybe other fields?
}
```

Then, the field would be: `deferred: Vec<PrecompileRequest>`.

Should we have a system event that adds entries to the this list? Or is the intent to make it more broad the pre-compiles? And if we introduce a dedicated instruction for tracking precompile accumulator, I'm imagining that that instruction would also push a new request onto this list - but either way, it is not clear to me if we need to add the `AdviceMutation::ExtendDeferred` variant.

adr1anh: The instruction would track the precompile _hashes_ while this field tracks their _witness_. A precompile handler will have to read the witness (either from the stack or from memory), and either check the validity or output a result to the advice stack/map. It then writes the entire state it has read into this `deferred` field.

While I mentioned in another comment we could potentially have a masm instruction for pushing data to the stack, I don't think it would make sense since we would always want to invoke a handler to verify the result.

adr1anh: Regarding a specialized struct, I really don't expect we'd be adding any fields for the moment so I'll leave it as is for now.

bobbinth: I may be misunderstanding the flow here a bit, but I'm imagining there something like `log_precompile` instruction that would populate this vector. In case of in-memory accumulator, this could be just a system event, and in case of aux-column based accumulator this could be the instruction that updates the accumulator as well.

The instruction would expect the stack to look like this:
```
[PRECOMPILE_LABEL, CALL_DATA_HASH]
```
And by the time we call this instruction, we'd also have the following data in the advice map:
```
CALL_DATA_HASH |-> [CALL_DATA]
```
So, to add an entry to the `deferred` vector, we'll just need to look up the calldata in the advice map. But maybe I'm missing something here?

adr1anh: Adding a vector of deferred calls also gives us the _ordering_ of the calls which is required by the native verifier/precompile VM prover to prove the actual calls.

To clarify the flow:
- The precompile event is called and the handler is able to read the calldata from the ProcessState.
- The handler computes the expected result, pushes it to the advice stack, and pushes the calldata to the `deferred` vector.
- The wrapper computes the `CALL_DATA_HASH` by hashing the calldata and the result it just popped from the advice stack.
- The wrapper calls the `log_precompile` instruction which absorbs `CALL_DATA_HASH` into the running capacity portion of the hash.
- At the end of the program execution, the VM outputs (either in the stack or via the auxiliary column) the squeezed hash.
- The deferred data is extracted from the AdviceProvider and can either be included in the proof, or used as input to the precompile VM for generating a proof. 
  - For native verification, the verifier iterates through all the deferred calls, verifies them, and recomputes the commitment output by the VM
  - For recursive verification, the commitment is a public input to the precompile VM which checks that each entry is valid and that their individual commitments match the one output by the Miden VM.

If I understand your suggestion, the handler first stores the calldata in the AdviceMap, and let the `log_precompile` operation then push the contents at `CALL_DATA_HASH` into the `deferred` vector. The end result is the same, but this adds a level of indirection which IMHO complicates the overall flow. At the moment, it's easier to just let the handler push the call data to `deferred` since it reads this data anyway. 

Finally, and this is something I am changing in the upcoming #2183 PR, it will likely be easier and more efficient to store the data as bytes rather than `Vec<Felt>`. This was left-over from when I was originally using the AdviceMap to store the calldata.

adr1anh:
I've had to modify most of the code here in #2183, which is already based on this branch. At this point, I think it would make more sense to close this one and just focus on the latter.

processor/src/host/advice/mod.rs
bobbinth: > If I understand your suggestion, the handler first stores the calldata in the AdviceMap, and let the `log_precompile` operation then push the contents at `CALL_DATA_HASH` into the `deferred` vector. The end result is the same, but this adds a level of indirection which IMHO complicates the overall flow. At the moment, it's easier to just let the handler push the call data to `deferred` since it reads this data anyway.

Agreed that the end result is the same, though, I'm somewhat hesitant about exposing another component of the advice provider publicly. If we keep it as an internal implementation, we may have a bit more flexibility to modify it down the road. I don't have any specific thoughts of what such modifications could be - so, this is more of a general desire to keep the surface of public APIs small (or at least as small as it is now).

---------

To refresh my memory, what would calldata for the Keccak precompile look like? Is it basically:

- RPO hash of the input data.
- Keccak hash of the input data.

Or is there anything else?

adr1anh: > Agreed that the end result is the same, though, I'm somewhat hesitant about exposing another component of the advice provider publicly. If we keep it as an internal implementation, we may have a bit more flexibility to modify it down the road. I don't have any specific thoughts of what such modifications could be - so, this is more of a general desire to keep the surface of public APIs small (or at least as small as it is now).

The only thing we would be exposing is a new AdviceMutation for storing call data, at least from the perspective of a precompile handler implementation. The rest of the API doesn't really need to be made public, as long as we can extract the `deferred` vector at the end of program execution.

On advantage of the current approach is we can store the witness as bytes rather than field elements (not currently implemented here but in #2183). For keccak, this requires half as much space. Having to go through the AdvideMap means we can only consider Felts. Bytes allow us to use more elaborate serialization for the witness which may be sent over the network in certain situations.

Moreover, the handler is slightly less efficient since it now also needs to compute the RPO hash of the input when it could just dump the witness from memory to `deferred`. We don't need to compute the AdviceMap key. I don't think this is a big win necessarily but does simplify the handler implementation.

Overall, while I agree with the principle of keeping the public API surface small whenever we can. In this case though, I think the benefit in terms of implementation simplicity outweighs the additional API surface.

adr1anh: > To refresh my memory, what would calldata for the Keccak precompile look like?

The calldata is just a the preimage as bytes, it's the only thing we need to recompute the full commitment `COMM = Rpo(Rpo(digest_u32) || Rpo(preimage_u32))`. Due to the padding necessary for packing bytes as u32, we will be using a tag containing both the keccak event ID and the length of the preimage in bytes.

Issue #1746, opened 2025-04-03
REPL: add ability to define procedures


plafer:
The current REPL executes everything typed by the user wrapped in a `begin ... end`.

This means, amongst other things, that a user cannot define procedures and execute them. This is unintuitive coming from other REPLs (e.g. `ipython`). I believe a better UX would look like:

```
>>> proc.foo add end
>>> push.1.2
>>> exec.foo
>>> debug.stack.1
3
>>>
```

bobbinth:
If we have #1748, this may no longer be needed.

plafer:
I think in an ideal world we'd have both - basically recreating the `ipython` experience. So if you just want to spin up a REPL and try out something quick and define a function, you don't have to first write in a file.

bobbinth:
Closing this as REPL functionality has been recently removed. Re-introducing it is covered by #2490.

PR #2161, opened 2025-09-15
fix: improve !use command parsing to prevent false matches

Snezhkko:
## Describe your changes

Fix !use command parsing in REPL to use exact matching instead of starts_with

The previous implementation used line.starts_with("!use") which caused
commands like !user, !used, or !use_something to be incorrectly parsed
as !use commands, resulting in unexpected behavior where these invalid
commands would display the list of available modules instead of being
treated as regular Miden assembly instructions.

This change ensures that only exact !use commands (with or without
module name argument) are processed as REPL commands, maintaining
consistency with other REPL commands that use exact string matching
and preventing user confusion from false command matches.

The fix uses a two-part validation:
- line.trim() == "!use" for commands without arguments
- line.trim().starts_with("!use ") && token_count == 2 for commands with arguments

### DIFF
--- a/miden-vm/src/repl/mod.rs
+++ b/miden-vm/src/repl/mod.rs
@@ -267,7 +267,10 @@ pub fn start_repl(library_paths: &Vec<PathBuf>, use_stdlib: bool) {
                     };
                 } else if line == "!stack" {
                     should_print_stack = true;
-                } else if line.starts_with("!use") {
+                } else if line.trim() == "!use"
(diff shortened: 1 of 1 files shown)


plafer:
Thank you for this, but I think we'll pass on it for now. Our REPL needs a decent rework (e.g. see #1746), and it's not a priority at the moment. I think the current implementation is good enough for now, as the REPL is a "best effort" anyways.

PR #2166, opened 2025-09-16
refactor: MastNode quality of life improvements

huitseeker:
**Note: this is a re-opening of #2152 for stacking**

This checkpoints some uncontroversial quality-of-life improvements I've found on the way to #1776. I expect no behaviour changes from this PR. The commits each stand on their own, but to summarize:
- rename `DecoratorIterator`, `RawDecoratorIterator` to better reflect their `Self::Item`,
- removes struct methods where there is a trait method that's an exact replacement (often implemented using that method),
    - technically removes the `BasicBlockNode::(append|prepend)_decorators` methods, happy to deprecate instead,
- uses `MastNodeErrorContext::decorators` instead of `BasicBlockNode::decorators`
    - technically removes the slice-returning `BasicBlockNode::decorators`, happy to deprecate instead,
- relies on existing `From<VariantX> for MastNode` impls instead of ad-hoc constructors,
    - the former are more familiar and do not require learning new APIs,
    - removes the `MastNode::new_variant_x` methods, happy to deprecate instead,
- uses ad-hoc destructors `MastNode::unwrap_x` where reimplemented

I'll mark this change as breaking in the CHANGELOG, however borderline (the changed functions aren't really used much).


adr1anh (approved):
Looks good to me!

Al-Kindi-0 (approved):
LGTM!

PR #2167, opened 2025-09-16
feat: add before_enter and after_exit decorators to BasicBlockNode

### LINKED ISSUE #1979: Add `before_enter` and `after_exit` to `basic_block` in order restrict `Noop` usage.
Right now all nodes in Mast have two lists of decorators to be executed one before "entering" the node and one after "exiting". The only exception is BasicBlock that has a single list of decorators that can be interleaved with the operations of the block.

However BasicBlock needs to execute some decorators at "exit" too and the only hacky way to do it right now is to [add a Noop at the end of the block and attach a decorator to it](https://github.com/0xMiden/miden-vm/pull/1950#issuecomment-3057709586).

This in turn seems to be the only usage of `Noop` besides padding, which means that if we add the before/after lists to BasicBlock we might restrict the use of Noop.

If all Noop are only padding, then https://github.com/0xMiden/miden-vm/issues/1965 https://github.com/0xMiden/miden-vm/pull/1974 would be obsolete.

It would also simplify:
- https://github.com/0xMiden/miden-vm/pull/1950
- https://github.com/0xMiden/miden-vm/pull/1940

This is more speculative but we might be able to remove `Instruction::Nop` altogether and only let the compiler/assembler insert `Operation::Noop` when needed.

huitseeker:
On top of #2166.

This change adds two decorator lists to `BasicBlockNode`: `before_enter` and `after_exit` using the usual `DecoratorId` pointers (same as the other block types).

Why this is good:
- It makes execution clear. No more adding no-ops or special “sentinel” indexes (see #1979).
- It makes decorator storage simpler. We will be able to keep the *operation-indexed* decorators in an *immutable* and *non-BasicBlock* place *after* the block is created, but such a block would still have easy methods to add or change the before and after decorators (see #1776).

Closes #1979


core/src/mast/node/external.rs
```diff
@@ -45,7 +45,7 @@ impl ExternalNode {
 }
 
 impl MastNodeErrorContext for ExternalNode {
-    fn decorators(&self) -> impl Iterator<Item = (usize, DecoratorId)> {
+    fn decorators(&self) -> impl Iterator<Item = DecoratedOpLink> {
```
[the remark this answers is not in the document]
huitseeker: This would technically be an issue (and one that would not be introduced in this PR, mind you) if the 6 other node types than `BasicBlock` had any operations. In this case 🤷

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +403,128 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
...
+                },
+                Segment::Middle => {
+                    if let Some(&(p, _)) = self.middle.clone().next() {
+                        should_yield = pos == p;
+                        break 'segwalk;
+                    }
+                    self.seg = Segment::After;
+                },
+                Segment::After => {
+                    if self.after.clone().next().is_some() {
```
adr1anh: While the iterator should be cheap to clone, we can use `peekable()`.

```suggestion
                    if self.after.peekable().next().is_some() {
```

assembly/src/basic_block_builder.rs
```diff
@@ -181,9 +181,9 @@ impl BasicBlockBuilder<'_> {
         if !self.ops.is_empty() {
             let ops = self.ops.drain(..).collect();
             let decorators = if !self.decorators.is_empty() {
```
adr1anh:
```suggestion
            let decorators = self.decorators.drain(..).collect();
```

core/src/mast/node/mod.rs
```diff
@@ -262,6 +262,10 @@ pub trait MastNodeErrorContext: Send + Sync {
     }
 }
 
+// Links an operation dex in a block to a decoratorid, to be executed right before this operation's
```
plafer:
```suggestion
// Links an operation index in a block to a decoratorid, to be executed right before this operation's
```

processor/src/fast/basic_block.rs
```diff
@@ -45,7 +45,7 @@ impl FastProcessor {
 
         let mut batch_offset_in_block = 0;
         let mut op_batches = basic_block_node.op_batches().iter();
-        let mut decorator_ids = basic_block_node.decorator_iter();
+        let mut decorator_ids = basic_block_node.indexed_decorator_iter();
```
plafer, line 47: Should we remove the call to `self.execute_before_enter_decorators(node_id, program, host)?;` on line 41? Otherwise IIUC this would re-execute the `before_enter` decorators before the first operation in the block.

plafer (commented):
Looks good so far! But just a partial review - will finish tomorrow

processor/src/fast/basic_block.rs
huitseeker: This specific iterator does not include `before_enter` (or `after_exit` decorators).

So as to make the non-duplication clear, I have a commit on top of this PR that counts the details of the decorator execution: https://github.com/0xmiden/miden-vm/commit/813c07e81a751f2612ccba71516fe1e989a04411

It's not included in this PR because it's a bit pedantic (and we've stated elsewhere that we may not want to slave ourselves to the legacy processor's behavior), but have a look, and tell me what you think @plafer. Maybe you like the tests, and I can find somewhere to include them. Maybe you like e.g. the `TraceCollector`, which looks marginally bit nicer to me than the `TestHost`.

plafer: Ah yes I see it now - looks great. It's a *bit* unintuitive because looking at `DecoratorOpLinkIterator::new()`, you expect `before_enter` & friends to map exactly to the fields of `BasicBlockNode` - but have to go look at how `indexed_decorator_iter()` initializes it to realize that `&[]` is passed to those (i.e. the source of my initial confusion). But honestly this is a problem that we haven't had a solution for in a while, and this quite clean and a massive improvement over what we had - so all good on my end.

re: new commit + `TraceCollector`: let's include those tests for sure! As for `TraceCollector`, this could maybe be merged with our `ConsistencyHost` (that could be renamed)? Take a look - but if `ConsistencyHost` doesn't fit well with what `TraceCollector` is doing, then we can leave them separate.

huitseeker: I've included the processor tests. Note the behavior of `indexed_decorator_iter` was also covered by `test_indexed_decorator_iter_excludes_before_enter_after_exit`.

I've made the doc comments on all iterator-returning methods more clear as well re: what they do or do not include.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -187,17 +196,31 @@ impl BasicBlockNode {
...
-    /// Returns a [`DecoratorIdIterator`] which allows us to iterate through the decorator list of
-    /// this basic block node while executing operation batches of this basic block node.
-    pub fn decorator_iter(&self) -> DecoratorIdIterator<'_> {
-        DecoratorIdIterator::new(&self.decorators)
+    /// Returns a [`DecoratorOpLinkIterator`] which allows us to iterate through the decorator list
+    /// of this basic block node while executing operation batches of this basic block node.
+    ///
+    /// This iterator is intended for e.g. processor consumption, as such a component iterates
+    /// differently through block operations: contrarily to e.g. the implementation of
+    /// [`MastNodeErrorContext`] this does not include the before_enter or after_exit
```
adr1anh:
```suggestion
    /// [`MastNodeErrorContext`] this does not include the `before_enter` or `after_exit`
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -187,17 +196,31 @@ impl BasicBlockNode {
...
     /// Returns an iterator which allows us to iterate through the decorator list of
     /// this basic block node with op indexes aligned to the "raw" (un-padded)) op
-    /// batches of the basic block node
-    pub fn raw_decorator_iter(&self) -> RawDecoratorIdIterator<'_> {
-        RawDecoratorIdIterator::new(&self.decorators, &self.op_batches)
+    /// batches of the basic block node.
+    ///
+    /// Though this adjusts the indexation of op-indexed decorators, this iterator returns all
+    /// decorators of the [`BasicBlockNode`] in the order in which they appear in the program.
+    /// This includes before_enter, op-indexed decorators, and after_exit.
```
adr1anh:
```suggestion
    /// This includes `before_enter`, op-indexed decorators, and `after_exit`.
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -239,8 +262,15 @@ impl BasicBlockNode {
 }
 
 impl MastNodeErrorContext for BasicBlockNode {
-    fn decorators(&self) -> impl Iterator<Item = (usize, DecoratorId)> {
-        self.decorators.iter().copied()
+    /// This iterator returns all decorators of the [`BasicBlockNode`] in the order in which they
+    /// appear in the program. This includes before_enter, op-indexed decorators, and after_exit.
```
adr1anh:
```suggestion
    /// appear in the program. This includes `before_enter`, op-indexed decorators, and `after_exit`.
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +409,128 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
     }
 }
 
+// DECORATOR ITERATION
+// ================================================================================================
+
+/// Iterator used to iterate through the decorator list of a span block
```
adr1anh: Should this be "basic block" rather than span?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +409,128 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
...
+/// while executing operation batches of a span block.
+///
+/// This lets the caller iterate through a Decorator list with indexes that match the
```
adr1anh, line 419:
```suggestion
/// This lets the caller iterate through a Decorator list with indices that match the
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +409,128 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
...
+            seg: Segment::Before,
+        }
+    }
+
+    /// Optional: yield only if the next item corresponds to the given op index.
+    /// - before_enter items map to op 0
+    /// - middle items use their stored position
+    /// - after_exit items map to `total_ops`
+    //
+    // some decorators are pegged on an operation index equal to the total number of
```
adr1anh:
```suggestion
    // Some decorators are pegged on an operation index equal to the total number of
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -384,6 +409,128 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
...
+    // operations since decorators are meant to be executed before the operation
+    // they are attached to, this allows them to be executed after the last
```
adr1anh:
```suggestion
    // they are attached to. This allows them to be executed after the last
```

core/src/mast/serialization/tests.rs
```diff
@@ -408,3 +410,77 @@ fn mast_forest_serialize_deserialize_advice_map() {
     let parsed = MastForest::read_from_bytes(&forest.to_bytes()).unwrap();
     assert_eq!(forest.advice_map, parsed.advice_map);
 }
+
+/// Test that BasicBlockNode serialization doesn't duplicate before_enter/after_exit decorators.
```
adr1anh:
```suggestion
/// Test that [`BasicBlockNode`] serialization doesn't duplicate `before_enter`/`after_exit` decorators.
```

core/src/mast/node/basic_block_node/mod.rs
huitseeker: https://english.stackexchange.com/a/3126 
I'm ok with indexes for now, just to keep the signal-to-noise ratio high. Happy to do a pass in another PR.

adr1anh (approved):
Looks good to me!

PR #2169, opened 2025-09-16
fix: run after-exit decorators when skipping loop (condition == 0)

VolodymyrBg:
Ensure execute_after_exit_decorators is called in start_loop_node when the loop condition is ZERO (skip body). Previously, after-exit decorators ran only in the finish path or when the loop executed at least once, causing inconsistency with:
- slow executor which always runs after-exit decorators after a node,
- other fast node handlers (join/call/dyn/split/external),
- documented loop semantics (LOOP followed by END when skipping).

### DIFF
--- a/processor/src/fast/loop.rs
+++ b/processor/src/fast/loop.rs
@@ -65,6 +65,9 @@ impl FastProcessor {
 
             // Increment the clock, corresponding to the END operation added to the trace.
             self.increment_clk(tracer);
+
+            // Execute decorators that should be executed after exiting the node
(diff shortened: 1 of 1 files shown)


plafer (approved):
LGTM - good catch!

plafer:
@VolodymyrBg can you add a changelog entry? Then we're good to go.

VolodymyrBg:
> @VolodymyrBg can you add a changelog entry? Then we're good to go.

Thanks for reviw, changelog is done

bobbinth (approved):
Looks good! Thank you!

PR #2172, opened 2025-09-16
feat: introduce `reversew` and memory ordered instruction

### LINKED ISSUE #2125: RFC: Introduce `reversew` instruction
I'd like to propose the addition of a new MASM instruction, `reversew`, which reverses the elements of the word on top of the operand stack. This would be initially implemented as a macro instruction (i.e. it expands to multiple "native" VM instructions, rather than being implemented as a single VM instruction itself). In the future, we might decide to promote it to a VM instruction for performance reasons, but at least initially that isn't necessary.

For some additional background context, this is related to the discussion [here](https://github.com/0xMiden/miden-vm/discussions/2059).

Note that this is implemented currently in the standard library as `std::word::reverse` - this issue proposes to introduce an instruction for this, due to its fundamental nature, and the possibility of implementing it natively in the future.

## Rationale

Per the above-linked discussion, there are some reasons we might wish to make `mem_loadw` and `mem_storew` work with words on the operand stack in the same order as they appear in Miden's linear memory. Currently, the elements of a word loaded via `mem_loadw` are in reverse order, i.e. the highest-addressable element is on top of the stack. There are some downsides to this that I'm not going to get into here. However, one way in which we could alleviate some of the pain points of this behavior, is to introduce an instruction that swaps the order of the elements in a word, so that one can do `mem_loadw reversew` and/or `reversew mem_storew` so that you can work with a word on the operand stack in the same order as it is in memory, without needing to change the actual `mem_loadw`/`mem_storew` semantics.

In conversations on this topic, it was also noted that such an instruction would also be useful more generally than just in conjunction with loads/stores. There are many situations in which you might want to shuffle around the elements of a word, or swap the order to make operating on the contents of the word easier. For this reason, I believe introducing a dedicated instruction for this is worthy of its own proposal.

Once introduced, this would also provide a pathway for introducing new word-oriented load/store operations that lay things out in memory order on the operand stack, as macro instructions which leverage `reversew`. We could then change which set of load/store ops are implemented as native VM instructions by having the assembler emit `reversew` after whichever set has the opposite behavior as the native set.

## Semantics

The `reversew` instruction consumes a word from the operand stack, and pushes back a word with the elements of the input word in reverse order, i.e. `[a, b, c, d]` becomes `[d, c, b, a]`.

In practice, the implementation of this instruction simply mutates the order of the top 4 elements on top of the operand stack, and is equivalent to the following procedure (again, currently implemented as `std::word::reverse`):

```text
export.reversew # [a, b, c, d]
    movdn.3         # [b, c, d, a]
    swap.2            # [d, c, b, a]
end
```

## Implementation

For the time being, the assembler would treat `reversew` as a macro which expands to the sequence `movdn.3 swap.2`.

In the future, we could promote this to a native VM instruction, if it is possible to implement it in few enough cycles to make that worthwhile.

## Future Work

A couple of additional things that might fall out from this, or might be similarly useful:

* Introduce `mem_loadw_be`/`mem_loadw_le` instructions, i.e. load a word in either big-endian or little-endian element order. These would expand to `mem_loadw`/`mem_loadw reversew` depending on the semantics of the native `mem_loadw` (with `mem_storew` equivalents as well).
* Introduce a `reversedw` instruction, which reverses the order of the top 8 elements of the operand stack, i.e. `[a, b, c, d, e, f, g, h]` becomes `[h, g, f, e, d, c, b, a]`, essentially equivalent to `reversew swapw reversew`.

adr1anh:
## Describe your changes

Closes #2125

Intruduces a new `reversew` instruction which desugards to `movdn.3 swap.2`.

We also introduce `mem_storew_{le, be}` and `mem_loadw_{le, be}` as convenience functions.

- Updated `CHANGELOG.md'


assembly/src/instruction/mod.rs
```diff
@@ -336,6 +336,7 @@ impl Assembler {
             Instruction::MovDn15 => block_builder.push_ops([MovDn8, SwapDW, MovDn7, SwapDW]),
             Instruction::MovDnW2 => block_builder.push_ops([SwapW2, SwapW]),
             Instruction::MovDnW3 => block_builder.push_ops([SwapW3, SwapW2, SwapW]),
+            Instruction::Reversew => block_builder.push_ops([MovDn3, MovUp2]),
```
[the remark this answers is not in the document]
adr1anh: Good catch, thanks

stdlib/asm/word.masm
```diff
@@ -2,13 +2,17 @@
...
+#!
 #! Inputs:  [a, b, c, d, ...]
 #! Outputs: [d, c, b, a, ...]
 #!
 #! Cycles: 3
 export.reverse
     movdn.3
+    # => [b, c, d, a, ...]
     swap.2
+    # => [d, c, b, a, ...]
```
bobbinth: Should we just use use `reversew` here?

assembly/src/instruction/mod.rs
```diff
@@ -419,8 +440,8 @@ impl Assembler {
                 )?
             },
             Instruction::MemStore => block_builder.push_ops([MStore, Drop]),
-            Instruction::MemStoreW => block_builder.push_ops([MStoreW]),
-            Instruction::MemStoreImm(v) => mem_ops::mem_write_imm(
+            Instruction::MemStoreW | Instruction::MemStoreWLe => block_builder.push_ops([MStoreW]),
+            Instruction::MemStoreImm(v) | Instruction::MemStoreWLeImm(v) => mem_ops::mem_write_imm(
```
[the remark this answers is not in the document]
adr1anh: Good catch

assembly/src/instruction/mod.rs
```diff
@@ -428,14 +449,33 @@ impl Assembler {
...
+                push_reversew(block_builder);
+                mem_ops::mem_write_imm(
+                    block_builder,
+                    proc_ctx,
+                    v.expect_value(),
+                    false,
+                    false,
+                    span,
+                )?
+            },
```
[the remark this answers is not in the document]
adr1anh: This is wrong

adr1anh:
> The `mem_loadw_be`, `mem_loadw_le` etc. changes would go into a different PR - right?

I already had started work on these yesterday and just pushed them in this PR.

assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -218,14 +220,22 @@ pub enum Instruction {
     MemLoadImm(ImmU32),
     MemLoadW,
     MemLoadWImm(ImmU32),
+    MemLoadWLe,
+    MemLoadWLeImm(ImmU32),
+    MemLoadWBe,
+    MemLoadWBeImm(ImmU32),
```
bobbinth: Do we still need to keep `MemLoadW` and `MemLoadWImm`? I thought these would be replaced by the `Be` variants.

assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -218,14 +220,22 @@ pub enum Instruction {
...
     MemStore,
     MemStoreImm(ImmU32),
-    LocStore(ImmU16),
     MemStoreW,
     MemStoreWImm(ImmU32),
+    MemStoreWLe,
+    MemStoreWLeImm(ImmU32),
+    MemStoreWBe,
+    MemStoreWBeImm(ImmU32),
+    LocStore(ImmU16),
```
bobbinth: Same comment as above but about `MemStoreW` and `MemStoreWImm`.

bobbinth (commented):
Looks good! Thank you! One thing is that I was thinking that we'd replace the current word-related IO instructions with the `Be` variants. Or does this create issues somehow?

adr1anh:
> Looks good! Thank you! One thing is that I was thinking that we'd replace the current word-related IO instructions with the Be variants. Or does this create issues somehow?

I _think_ this is a matter of find-and-replace, I'll try it out to see what effect this has, but maybe this is worth doing as a followup

adr1anh:
> Looks good to me! The main thing I am not sure about is the le vs be in this PR, it seems to me that we want the current mem_loadw to be equivalent to mem_loadw_be, and not the other way around. Basically, in memory, a word will be stored in le and then loading onto the stack, in our current configuration, would reverse it such that its highest limb is on the stack-top, i.e., it is placed on the stack in be.

Ah yes, I think you're right. This also should be rather simple but worth doing in this PR.

adr1anh:
I've implemented @Al-Kindi-0's comment and reversed the previous definitions, and added more explicit tests to demonstrate the new semantics.

docs/src/user_docs/assembly/io_operations.md
```diff
@@ -65,13 +65,17 @@ System events fall into two categories: (1) events which push new data onto the
...
+| mem_storew_be <br> - *(1 cycles)*  <br> mem_storew_be.*a* <br> - *(2-3 cycles)* | [a, A, ... ]          | [A, ... ]           | $A \rightarrow mem[a..(a+4)]$ <br> Stores the top four elements of the stack in big-endian (reversed) order in memory starting at address $a$, such that the top of stack is placed at `mem[a+3]`. Equivalent to `mem_storew`. If $a$ is provided via the stack, it is removed from the stack first. <br> Fails if $a \ge 2^{32}$, or if $a$ is not a multiple of 4 |
+| mem_storew_le <br> - *(6 cycles)*  <br> mem_storew_le.*a* <br> - *(5-6 cycles)* | [a, A, ... ]          | [A, ... ]           | $A \rightarrow mem[a..(a+4)]$ <br> Stores the top four elements of the stack in little-endian (memory) order in memory starting at address $a$, such that the top of stack is placed at `mem[a]`. Equivalent to `reversew mem_storew`. If $a$ is provided via the stack, it is removed from the stack first. <br> Fails if $a \ge 2^{32}$, or if $a$ is not a multiple of 4 |
```
bobbinth, line 77: I think this is equivalent to `reversew mem_storew reversew` (rather than just `reversew mem_storew`) - no? If so, they cycle count would also be different (i.e., 9 - 10 cycles).

adr1anh: Oops, that change got lost somehow. Yes you're right, I'll create a small PR to fix

Issue #1701, opened 2025-03-17
Fast processor improvements


plafer:
This is a tracking issue of all the improvements to try/make for `FastProcessor` after #1668.

- [ ] Refactor `BasicBlockNode` (and specifically `OpBatch`) to make iterating over operations fast
    - Specifically, the processor shouldn't have to infer where to insert `Noop`s; they should already be inserted in the `OpBatch`.
- [ ] Investigate reducing the stack's buffer size, with a fallback mechanism to a `Vec` (https://github.com/0xPolygonMiden/miden-vm/pull/1668#discussion_r1997769752)
- [ ] Investigate using `hashbrown`'s `HashMap` instead of a `BTreeMap` for memory (https://github.com/0xPolygonMiden/miden-vm/pull/1668#discussion_r1997774871)
- [ ] Investigate using `u64` instead of `(ContextId, u32)` as the key to the memory's `BTreeMap` (https://github.com/0xPolygonMiden/miden-vm/pull/1668#discussion_r1997774871)
- [ ] Investigate changing how bounds checks work https://github.com/0xPolygonMiden/miden-vm/pull/1668#discussion_r2020248213

bobbinth:
@plafer - seems like at least some of the things mentioned in this issue may have been addressed. Should we close it and create more targeted issues for w/e remains?

plafer:
I checked off the 2 things that are done - otherwise I would just leave this one open. Some of those points are a bit too minor to have their own issue, but it would still be good to try them out in the future.

plafer:
Superseded by #2818 and #2819

PR #2173, opened 2025-09-16
Attempt to move stack in stack buffer when reaching boundaries

plafer:
Fixes issues we've observed in the node/client/etc regarding stack underflows. For a detailed description of the issue, see https://github.com/0xMiden/miden-base/issues/1902#issuecomment-3303083370.

In this PR, we move the stack back up in the stack buffer when we reach the boundaries. We will only error-out if the stack buffer is truly full (or just about). This will still need to be improved in the future, either by falling back to a `Vec` for the stack buffer (as described in #1701), or some other solution.


bobbinth:
Do we still need this PR? Or would merging `main` into `next` resolve the issue?

plafer:
We still need this PR - merging `main` into next will cause some pretty bad merge conflicts. Ideally 0.17.2 becomes a branch of its own, and `main` is effectively reverted back to 0.17.1 before merging `next` into it for v0.18.

This is basically the proper fix PR (with tests, etc), whereas 0.17.2 was more a band-aid to unblock production.

bobbinth (approved):
Looks good! Thank you!

PR #2174, opened 2025-09-16
fix: make MastForest::merge behave as identity for single forests

### LINKED ISSUE #1644: `MastForest::merge` for one forest does not behave like identity
### Packages versions

miden-vm: 0.12.0

### Bug description

When `MastForest::merge` takes a single element it should return it unchanged. In some cases like the program built in `miden-vm/assembly/src/assembler/tests.rs:nested_blocks` the returned forest is different than the original.

This is a minor issue, just documenting it because it caused an unrelated piece of code to fail (the mast vendoring).

_No response_

### Relevant log output

```shell

```

huitseeker:
This fixes issue #1644 where merging a single forest would change node digests
due to ID remapping, even though logically it should be an identity operation.

This introduces a test explaining the issue (NodeID remappings), then detects when MastForest::merge is called with a single forest and returns that forest unchanged with an identity root map, preserving node IDs and digests while maintaining behavior for multi-forest merges.

TBH, I'm not convinced this is such a big deal, but I was in the neighborhood.

Closes #1644


adr1anh:
After looking a bit into the functionality, it seems a bit odd to handle this case separately. The function performs "normalization" after merging multiple forests, so if the input forest isn't then it would be normal that it doesn't behave as the identity. Here, by "normalizing" it seems to 
- Maintain the invariant that child node IDs < parent node IDs
- Create a clean, deduplicated forest structure
- Provide consistent node ordering regardless of input

I'd be in favor of closing the issue, and either update the PR to better explain the semantics of `merge` or just close it.

huitseeker:
@adr1anh Agreed, though I would think that part of the consistent node ordering (beyond the child-parent relationship) is a bit of an idiosyncracy.

core/src/mast/merger/tests.rs
```diff
@@ -846,3 +849,56 @@ fn mast_forest_merge_advice_maps_collision() {
...
+    let new_root_id = merged_roots[0];
+
+    let new_root = merged_forest.get_node_by_id(*new_root_id).unwrap();
+    let new_digest = new_root.digest();
+
+    // even for identity operations, the merge process:
+    // 1. Remaps all node IDs to new indices in the merged forest
+    // 2. Recomputes fingerprints using the NEW node IDs
+    // 3. This changes the hash input, even though the logical structure is identical
+    assert_eq!(original_digest, new_digest, "Single-forest merge should not preserve digests");
```
adr1anh: Shouldn't this be 

```suggestion
    assert_ne!(original_digest, new_digest, "Single-forest merge should not preserve digests");
```

core/src/mast/merger/mod.rs
```diff
@@ -37,10 +37,27 @@ pub(crate) struct MastForestMerger {
 impl MastForestMerger {
     /// Creates a new merger with an initially empty forest and merges all provided [`MastForest`]s
     /// into it.
+    ///
+    /// # Normalizing Behavior
```
adr1anh, line 41: Might be worth mentioning that this normalization is an idempotent operation?

core/src/mast/merger/tests.rs
huitseeker: Yep, I missed something on this test — draft until I've had the time to come back to it.

huitseeker: I've added a much better test here, which does show the node order is remapped when performing even a single-forest merge, but that this is idempotent, which was the intent all along.

PR #2176, opened 2025-09-17
doc: address various documentation issues

### LINKED ISSUE #2134: Docs: Minor fixes to improve explanations & fix broken link
### Main docs page

Page: https://0xmiden.github.io/miden-docs/imported/miden-vm/src/intro/main.html#planned-features

**Issue:** The "Recursive proofs" term is used without definition on main docs page
**Action needed:** Add explanation of what recursive proofs are

### Overview page

Page: https://0xmiden.github.io/miden-docs/imported/miden-vm/src/intro/overview.html

**Issues:**
- Typo: "MV ≠ VM" in first sentence should be "VM ≠ MV"
-  Missing disclaimer about advice provider/host trustworthiness. Include that data should only be read from but not written and re-read.
- "Vector" term used without explanation, which might be confusing for non Rust-native readers. A simple one-word explanation for the term should be added.

### Development Tooling Page

Page: https://0xmiden.github.io/miden-docs/imported/miden-vm/src/tools/main.html

**Issue:** The "Scaffold repo" link is broken
**Action needed:** Make the link point to the new [project-template](https://github.com/0xmiden/project-template/) repository.

huitseeker:
This closes #2134 and adds a link-checking job in CI, which runs every day and opens an issue on error.


docs/src/intro/overview.md
```diff
@@ -48,7 +48,7 @@ The *advice provider* component is responsible for supplying nondeterministic in
...
+* **Advice map** which is a key-value map where keys are words and values are vectors (dynamic arrays) of field elements. The VM can copy values from the advice map onto the advice stack as well as insert new values into the advice map (e.g., from a region of memory).
 * **Merkle store** which contain structured data reducible to Merkle paths. Some examples of such structures are: Merkle tree, Sparse Merkle Tree, and a collection of Merkle paths. The VM can request Merkle paths from the Merkle store, as well as mutate it by updating or merging nodes contained in the store.
 
-The prover initializes the advice provider prior to executing a program, and from that point on the advice provider is manipulated solely by executing operations on the VM.
+The prover initializes the advice provider prior to executing a program, and from that point on the advice provider is manipulated solely by executing operations on the VM. **Security note:** Data from the advice provider should only be read from and not written back and re-read, as this could potentially create trust assumptions about the data source.
```
bobbinth: Technically, this is not entirely accurate: we actually do write data to the advice provider and read it back later on (we use it to pass large amounts of data across context calls). The important thing is that when reading the data from the advice provider, we can't assume that it is valid - we always have to verify this (e.g., when writing and then reading the data, we need to make sure that the data we read hashes to some expected value).

docs/src/intro/main.md
```diff
@@ -22,7 +22,7 @@ Miden VM is a fully-featured virtual machine. Despite being optimized for zero-k
 ### Planned features
 In the coming months we plan to finalize the design of the VM and implement support for the following features:
 
-* **Recursive proofs.** Miden VM will soon be able to verify a proof of its own execution. This will enable infinitely recursive proofs, an extremely useful tool for real-world applications.
+* **Recursive proofs.** Miden VM will soon be able to verify a proof of its own execution. This will enable infinitely recursive proofs, an extremely useful tool for real-world applications. Recursive proofs allow one to prove the validity of a proof by recursively proving the validity of the proof's validity, and so on. This enables composition of proofs in layers, where each layer proves the soundness of the layer below it.
```
bobbinth: I think "Recursive proofs allow one to prove the validity of a proof by recursively proving the validity of the proof's validity, and so on." is a bit confusing.

I'm actually wondering if we should re-write this whole point to be about something else. Basically, this is about enabling efficient STARK-verifiers which will enable verifying of VM proofs recursively, but also will enable pre-compiles, and potentially many other things.

@al @adr1anh - what do you think?

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,56 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Open issue on failure (only if fail-fast is false)
+        if: steps.lychee.outcome == 'failure' && steps.flags.outputs.fail_fast != 'true'
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Link Checker Report
+          content-filepath: ./lychee/out.md
+          labels: report, automated issue
```
bobbinth: nit: EOF newline.

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,56 @@
+name: Check documentation links
```
bobbinth, line 2: Should we add a brief comment explaining what this workflow does (and maybe we should start doing this for other workflows too). Maybe something like: "Checks documentation for broken links once per day and creates issues if broken links are found".

.github/workflows/link-checker.yml
bobbinth: I don't think we have these labels in the repo. Should we create them? But also, maybe the only label we need is `documentation`?

Also also, could we set the issue type to "task"?

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,56 @@
...
+            --verbose
+            --exclude-mail
+            --exclude "localhost"
+            --exclude "127.0.0.1"
+            --exclude "^file://"
+            --exclude "github.com/0xMiden/.*"
+            docs/src/**/*.md
+            **/*.md
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
bobbinth, line 49: We cannot run this as a `make` command because we want to open an issue based on the result, right?

huitseeker: Oh, we're running a full dedicated Github action, here, which is more efficient than installing the (non-default) utility, and configuring the outputs, and running it through Make. You'd also have the issue that we'd need developers to have yet another non-installed by default tool.

.github/workflows/link-checker.yml
huitseeker: Those will be created upon first use and do not need to be pre-filled. The issue type is a project-specific *template•, IIUC, not something the automated issue creation understand.

docs/src/intro/overview.md
huitseeker: Re-phrased.

docs/src/intro/main.md
huitseeker: Rephrased.

bobbinth (approved):
Looks good! Thank you!

PR #2182, opened 2025-09-17
hotfix 0.17: remove all underflow errors

plafer:
This is a backport of #2173 on top of 0.17.1.


processor/src/fast/mod.rs
```diff
@@ -1202,9 +1186,13 @@ impl FastProcessor {
...
-    fn increment_stack_size(&mut self) {
-        self.stack_top_idx += 1;
-        self.update_bounds_check_counter();
+    fn increment_stack_size(&mut self) -> Result<(), ExecutionError> {
+        if self.stack_top_idx < STACK_BUFFER_SIZE - 1 {
+            self.stack_top_idx += 1;
+            Ok(())
+        } else {
+            Err(ExecutionError::FailedToExecuteProgram("stack overflow"))
+        }
```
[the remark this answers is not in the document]
plafer: No, `stack_top_idx` is about to get incremented. If it is `== STACK_BUFFER_SIZE - 1`, then it will point outside the buffer after the increment.

processor/src/fast/mod.rs
```diff
@@ -1319,6 +1276,62 @@ impl FastProcessor {
...
+    fn reset_stack_in_buffer(&mut self, new_stack_top_idx: usize) {
+        debug_assert_eq!(self.stack_depth(), MIN_STACK_DEPTH as u32);
+
+        let new_stack_bot_idx = new_stack_top_idx - MIN_STACK_DEPTH;
+
+        // Copy stack in new position
+        let current_stack_top: [Felt; MIN_STACK_DEPTH] = self.stack
+            [self.stack_bot_idx..self.stack_top_idx]
+            .try_into()
+            .expect("stack depth must be 16 in restore_overflow_stack");
```
bobbinth: nit: the expect message should probably refer to `reset_stack_in_buffer` rather than `restore_overflow_stack`.

processor/src/fast/mod.rs
```diff
@@ -1319,6 +1276,62 @@
...
+                core::cmp::min(INITIAL_STACK_TOP_IDX + target_overflow_len, STACK_BUFFER_SIZE - 1);
+
+            self.reset_stack_in_buffer(new_stack_top_idx);
+        }
+
+        // Restore the overflow
+        self.stack[range(self.stack_bot_idx - target_overflow_len, target_overflow_len)]
+            .copy_from_slice(&ctx_info.overflow_stack);
+        self.stack_bot_idx -= target_overflow_len;
+    }
```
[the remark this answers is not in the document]
plafer: This is mitigated by the fact that we choose the `new_stack_top_idx` as

```rs
let new_stack_top_idx =
    core::cmp::min(INITIAL_STACK_TOP_IDX + target_overflow_len, STACK_BUFFER_SIZE - 1);
```

We know that `target_overflow_len < STACK_BUFFER_SIZE` (i.e. fits on the stack), since the overflow was previously on the stack. By taking the `min` with `STACK_BUFFER_SIZE - 1`, this basically allows us to put the stack top index as far into the buffer as possible - but still "try" to put it near `INITIAL_STACK_TOP_IDX`.

processor/src/fast/mod.rs
```diff
@@ -1319,6 +1276,62 @@ impl FastProcessor {
...
+        debug_assert_eq!(self.stack_depth(), MIN_STACK_DEPTH as u32);
+
+        let new_stack_bot_idx = new_stack_top_idx - MIN_STACK_DEPTH;
+
+        // Copy stack in new position
+        let current_stack_top: [Felt; MIN_STACK_DEPTH] = self.stack
+            [self.stack_bot_idx..self.stack_top_idx]
+            .try_into()
+            .expect("stack depth must be 16 in restore_overflow_stack");
+        self.stack[new_stack_bot_idx..new_stack_top_idx].copy_from_slice(&current_stack_top);
```
bobbinth: It may be easier to use [copy_within()](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within) method here. This way, we won't need to use the `current_stack_top` array as in intermediate buffer.

bobbinth (approved):
Looks good! Thank you! I left a couple of small nits inline.

processor/src/fast/mod.rs
plafer: Removed, since switched to use `copy_within`

processor/src/fast/mod.rs
plafer: Ah good idea! Done

Issue #2147, opened 2025-09-11
Precompile logging implementation


adr1anh:
The current precompile handlers are not yet sound since there is no proper mechanism in the VM for keeping track of all precompile calls that were requested from the host. A correct implementation would 
1. Get the non-deterministic result from the host (e.g. a keccak hash of data stored in memory).
2. Compute a commitment to the call-data required to recompute the result (e.g. the hash pre-image and claimed output). The pre-image is stored in the AdviceProvider.
3. Log the call by accumulating the call-data commitments
4. Output a commitment to the call-data of all precompile calls, alongside the usual public inputs.
5. Pack call-data in the VM proof, and update the verifier to
	1. Verify the correctness of each precompile call
	2. Recompute the call-data commitment, ensuring it equals the commitment output by the VM

This issue revolves around point 3. and 4. which we are planning on implementing as two kernel procedures.

#### `log_calldata`

The implementation itself is rather straightforward. The procedure is called by a precompile wrapper which will have emitted the handler event, and them computed the `CALL_DATA` commitment. It will then absorb the commitment into a running sponge from which we can extract a commitment to the list of all precompile calls.

```
use.std::crypto::hashes::rpo

const.PRECOMPILE_SPONGE_PTR=0 #?

#! Inputs: [CALL_DATA, ...]
#! Outputs: [CALL_DATA, ...]
export.log_calldata
	# Load the running capacit 
	padw memloadw.PRECOMPILE_SPONGE_PTR
	# => [C, CALL_DATA]
	
	# Prepare sponge state
	swapw padw
	# => [ZERO, CALL_DATA, C]
	
	# Absorb into sponge and extract next state
	hperm dropw dropw
	# => [C']
	
	# Write the next state to memory
	mem_storew.PRECOMPILE_SPONGE_PTR
end
```

The main issue with this implementation is the reliance on a well-known address, which may only be written to by this specific kernel procedure. If this address was maliciously overwritten with `[0, 0, 0, 0]`, any previously logged call-data would be discarded and no longer need to be verified.  

Given that a kernel is considered "trusted", we can use an address in the kernel memory, and specify that no other procedure may write to it. It does not seem like this is statically enforceable. 

The other issue is that any precompile wrapper is now tied to this specific procedure, since it must be explicitly called. We would have to update our standard library whenever this procedure changes (unlikely). If we were to transition to a different way of handling precompiles, the wrappers themselves will likely also need to be updated.

[...]

PR #2183, opened 2025-09-17
feat: native precompile verification

adr1anh:
- Updated `CHANGELOG.md'


adr1anh:
Some notes:

- I've expanded on #2159 and changed it to store the calldata as a byte vector rather than field elements ([context](https://github.com/0xMiden/miden-vm/pull/2159#discussion_r2359374108)).
- All precompile related structures and traits are stored in `core` since it doesn't rely on the processor, and its usage spans multiple crates.
- I've refactored the keccak tests quite a bit to handle verification of the changes made to the proof verification. In particular, I've moved the testing `Preimage` struct to the handler module since it avoids code duplication. 
- The verifier currently "verifies" the precompiles, but the actual verification would occur when comparing commitments. This is because the calldata does not store the calldata hash computed by the VM. We check this manually in a keccak test.

adr1anh:
This should now be in a reviewable state for those interested. I'll comb over all the changes tomorrow to make sure I didn't miss anything.

adr1anh:
This PR is now ready for review. The main changes are the following
- Added a `precompile` module to `core` which includes new primitives required to verify precompiles and recompute their commitments during native verification. We also introduce a new `PrecompileVerifier` trait.
- Extending upon #2159 by changing `AdviceProvider::deferred` to `precompile_requests: Vec<PrecompileRequest>`. In particular, we now allow precompiles to use an additional word to store any auxiliary data that can be computed from the request. 
- Adds two methods `prove_with_precompiles` and `verify_with_precompiles` which support native verification. **NOTE:** The current approach is missing a check that ensures the verifier derives the same commitment to all calls like in the VM. This will require #2147.
- Refactor of the keccak handler to support the new structure
  - The preimage is now stored as bytes in the `precompile_requests` field, rather than as Felts in the advice map. This both reduces the request size by a factor of two. It also avoids having to recompute the commitment in the handler, leading to faster execution in the FastProcessor.
  - Introduce helper struct `KeccakPreimage` to unify handler and testing code, and avoids duplicating code. 
  - Adds a verifier which recompute the expected commitment to the precompile call. 
- The standard library now exports a list of verifiers for the precompiles it exposes.

core/src/precompile.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub event_id: EventId,
+    /// Raw byte data for the precompile operation
+    pub data: Vec<u8>,
+}
+
+impl Serializable for PrecompileRequest {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        target.write_u64(self.event_id.as_felt().as_int());
+        target.write_usize(self.data.len());
+        target.write_bytes(&self.data);
```
bobbinth: nit: I would write this as:
```rust
self.event_id.as_felt().write_into(target);
self.data.write_into(target);
```

core/src/precompile.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+}
+
+impl Deserializable for PrecompileRequest {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let event_id = EventId::from_u64(source.read_u64()?);
+        let len = source.read_usize()?;
+        let data = source.read_vec(len)?;
+        Ok(Self { event_id, data })
```
bobbinth: I would write this as:
```rust
let event_id = EventId::from_felt(Felt::read_from(source)?);
let data = Vec::<u8>::read_from(source)?;
Ok(Self { event_id, data })
```
This is a bit more compact, but more importantly, it avoids a slight ambiguity in event ID deserialization (e.g., a `u64` could be great than field modulus and we'd silently perform the reduction).

stdlib/src/handlers/keccak256.rs
```diff
@@ -101,56 +94,34 @@ fn read_witness(process: &ProcessState, ptr: u64, len_bytes: u64) -> Option<Vec<
...
-    witness.push(Felt::new(len_bytes));
-
     // Read each memory location in the range [start_addr, end_addr) and append to the witness.
     let ctx = process.ctx();
-    for addr in start_addr..end_addr {
-        let value = process.get_mem_value(ctx, addr)?;
-        witness.push(value);
-    }
-    Some(witness)
+    (start_addr..end_addr).map(|addr| process.get_mem_value(ctx, addr)).collect()
```
bobbinth, line 111: Maybe we should add a method on `ProcessState` to encapsulate this. Something like `get_memory_range()`.

stdlib/src/handlers/keccak256.rs
```diff
@@ -101,56 +94,34 @@ fn read_witness(process: &ProcessState, ptr: u64, len_bytes: u64) -> Option<Vec<
...
+// ================================================================================================
+
 /// Keccak256 digest representation in the Miden VM.
 ///
 /// Represents a 256-bit Keccak digest as 8 field elements, each containing a u32 value
 /// packed in little-endian order: `[d_0, ..., d_7]` where
 /// `d_0 = u32::from_le_bytes([b_0, b_1, b_2, b_3])` and so on.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub struct KeccakFeltDigest([Felt; 8]);
+pub struct KeccakFeltDigest(pub [Felt; 8]);
```
bobbinth: nit: does the inner value need to be `pub`? Maybe instead of making it `pub` we could implement `Deref` or maybe just add `as_elements()` method?

Also, not related to this PR, but I'd probably modify the `to_commitment()` method on lines 142 - 146 below to avoid reversing the elements. To do this, we'd need to reverse the elements on the stack before computing the RPO hash. For this, I think we should be able to use the `reversedw` instruction.

stdlib/src/handlers/keccak256.rs
```diff
@@ -173,14 +144,152 @@ impl KeccakFeltDigest {
...
+    /// Returns the digest formatted as 8 field elements, each containing a u32 value
+    /// in little-endian byte order. This matches the format expected by the VM
+    /// and can be directly used on the operand stack.
+    pub fn digest(&self) -> KeccakFeltDigest {
+        let hash_u8 = Keccak256::hash(&self.0);
+        KeccakFeltDigest::from_bytes(&hash_u8)
+    }
+
+    /// Computes the precompile commitment: `RPO(RPO(input) || RPO(hash))`, along with the tag for
+    /// the computation.
```
bobbinth: nit: I would maybe write it as `RPO(RPO(input) || RPO(keccak_hash))`.

stdlib/src/handlers/keccak256.rs
```diff
@@ -173,14 +144,152 @@ impl KeccakFeltDigest {
...
+/// Keccak256 preimage structure representing the raw input data to be hashed.
+///
+/// This structure encapsulates the raw bytes that will be passed to the Keccak256
+/// hash function, providing utilities for:
+/// - Converting between bytes and field element representations
+/// - Computing the Keccak256 digest
+/// - Generating precompile commitments for verification
+/// - Handling the data packing format used by the VM
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct KeccakPreimage(pub Vec<u8>);
```
bobbinth: Similar question to one of the above: do we need to make the internal value public?

core/src/precompile.rs
```diff
@@ -0,0 +1,272 @@
...
+///
+/// This structure encapsulates the call data for a precompile operation, storing
+/// the raw bytes that will be processed by the precompile function.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileRequest {
+    /// Event ID identifying the type of precompile operation
+    pub event_id: EventId,
+    /// Raw byte data for the precompile operation
+    pub data: Vec<u8>,
+}
```
bobbinth, line 104: I wonder if a 64-bit value is enough here. My understanding is that this will be a part of the input into the precompile VM, and so, if we have a precompile that where inputs/outputs are the same (may be possible if the output is something simple - like a binary flag), it may be possible to find a collision to trick the precompile VM to check some other precompile.

Granted, doing so is not going to be easy - but I still wonder if it is enough of an attack vector to think about. If it is, we may need to define "precompile ID" somehow differently from the event ID.

------

More broadly, I'd like to align on the intended uses/properties of this struct. Here is how I was thinking about it:

Given a list of precompile requests (i.e., `Vec<PrecompileRequest>`, we should be able to compute a commitment to all precompile requests - i.e., the commitment that would be output by the VM. For this, we'd need to be able to do something like `PrecompileRequest::to_commitment() -> PrecompileCommitment` - but I don't think the request currently has enough data to do this. I think to address this, we now have the `PrecompileVerifiers` structure - but it may be simpler just putting all the required data into the request.

For example, the struct could look like so:
```rust
pub struct PrecompileRequest {
    pub tag: Word,
    pub commitment: Word,
    pub data: Vec<u8>,
}
```

I'm also assuming that the precompile VM will use the request to populate its equivalent of the advice provider. For example, for Keccak, we'll need to have something like `RPO(input) |-> [input elements]`. To accommodate this, we could track data as elements, and the struct would look like:

```rust
pub struct PrecompileRequest {
    pub tag: Word,
    pub commitment: Word,
    pub data: Vec<Felt>,
}
```

Then "loading" of precompile requests into the precompile VM could look something like:

1. Take all precompile requests and compute the overall commitment to deferred computations. For this, we'll need to compute `RPO([tag, commitment])` value for each request, and then hash the results sequentially (potentially, appending `[ZERO, ZERO]` to the last has).
2. The value computed above will be the public input into the precompile VM, and we'll also need to load `overall_hash |-> [request_hashes]` into the advice provider equivalent of the precompile VM.
3. For each request, we'll also need to load the following into the advice provider:
  a. `request_hash |-> [tag, commitment]`.
  b. `RPO(data) |-> [data]`

With this, the precompile VM should have all the info needed to generate the proof, assuming the data commitment can always be reduced to a sequential hash. If the data is more complex in nature, we could provide more flexibility:

```rust
pub struct PrecompileRequest {
    pub tag: Word,
    pub commitment: Word,
    pub data: BTreeMap<Word, Vec<Felt>>,
}
```
Where, the `BTreeMap` could track a set of inputs that would need to be provided to the precompile VM non-deterministically.

-------

To facilitate precompile verification outside of the precompile VM, we could still make `PrecompileRequest` convertable into concrete types. For example, for Keccak we could implement something like `TryFrom<PrecompileRequest> for KeccakPreimage` as well as `From<KeccakPreimage> for PrecompileRequest`.

stdlib/src/handlers/keccak256.rs
adr1anh: I agree, a method was recently introduced to `ProcessState` but this requires `start` and `end` pointers to be on the stack. I'll change that one to accept `u64` which are manually read from the stack and converted to an integer.

stdlib/src/handlers/keccak256.rs
adr1anh: The main reason was for testing purposes, but I'll see what I can do to make the inner field private. If we can put any value there it is not really a digest anymore.

EDIT: Both of these were fixed in the latest commits

stdlib/src/handlers/keccak256.rs
adr1anh: In this case I think it's reasonable because any `Vec<u8>` is a valid preimage so there is no need for any constructor. It avoid any `new` methods which would just populate this field.

core/src/precompile.rs
adr1anh: I've implemented `From<KeccakPreimage> for PrecompileRequest` since it was a very simple change. For `TryFrom` though we'd want to take a reference since we otherwise drop the preimage, as well as perform extra allocations due to the Felt conversion, both of which I'm not really a fan of.  I'd rather introduce another kind of trait which makes the conversion explicit.

adr1anh: I'll respond below in the conversations to make it easier to discuss the points above.

adr1anh:
## Tagging precompiles

I think you may be right that in the current form, the `EventId` is not enough and could lead to ambiguity when it comes to identifying requests. At the moment, its purpose is to map a request to a specific `PrecompileVerifier`, but it would be good to enforce this mapping. 

One solution would be to use the final rate portion of the final hash (which we currently set to `[ZERO, ZERO]`) to store a commitment to the mapping in the form of a sequential hash `[VERIFIER_COMMITMENT, event_id, 0, 0, 0]`. If we consider only the recursive verification scenario, then the commitment could be the ACE circuit for that particular precompile, or even some form of on-chain commitment. The main point is that we can keep a short identifier to each request by maintaining a map of `VERIFIER_COMMITMENT |-> precompile_id`. This would be similar to the way strings are encoded in the upcoming BITE serialization format. 

I haven't fully thought this through though, and maybe we do need to include the actual commitment inside each individual precompile commitment. 

## `PrecompileRequest` struct

This was something I considered originally, but I realized that we really only need the `data` portion. It should be exactly the information we need to compute the mapping from `PrecompileRequest` to `PrecompileCommitment`. It should always be "enough" since we can store more data in it if necessary.  The idea is that this information would be serialized natively to ensure the request is small since these would sometimes be sent over the network. For signatures and hashes it would be done as follows:

**Hashes**:
We have a mapping from `(len_bytes, input_felt_u32) <-> input_u8`, where `input_u8` corresponds to `data`. We need `len_bytes` since we would otherwise have ambiguity in the last `u32` limb due to the packing rule. The `PrecompileCommitment` is then defined as 
```
RPO(
	TAG = [hash_id, len_bytes, 0, 0],
	RPO(
		RPO(input_felt_u32),
		RPO(digest_felt_u32),
	)
)
```

This seems to be the "easiest" representation to compute within the VM (Miden and Precompile), as well as natively. 

I’m not against putting auxiliary information in the second component though, or even using capacity of the sponge, which would free up the entire `TAG` for the precompile ID. 

**Signatures**:

While we have not implemented this yet, I imagine the commitment would look something like this:

```
RPO(
	TAG = [sig_id, is_verified, 0, 0],
	RPO(
		message_digest,
		public_key,
		signature_data,
	)
)
```

When a signature verification is requested through an event wrapper, the handler pushes the resulting `is_verified` of the verification to the advice stack. The caller however can ignore this result and compute the commitment setting `is_verified = 1` ensuring that the signature must be valid, but it could ask the VM to check if the signature is invalid.

The `data` portion of the request would be the serialization of a struct containing the 3 components of the signature payload, from which we can compute the expected `TAG`. Again though, we could easily include the result in the precompile commitment commitment and use the full `TAG` for the signature precompile ID.

---
In both situations, if any portion of the commitment is incorrect, then the resulting commitment would be different, and therefore would fail the overall verification which computes the aggregated commitment of all precompile calls. From this point of view, the mapping `PrecompileRequest -> PrecompileCommitment` should be injective and actually never fail. The reason why we may want to include `commitment` inside `PrecompileRequest` would be to detect if an individual request is invalid, rather than only reporting that "there exists one invalid request". I think whether we want this depends on our adversarial model or debugging needs, since an adversary could make either an individual request fail or the entire set. The only thing we can trust from the VM execution is the final commitment, so we can't accurately report which inputs where incorrect.

If we think of a model where a network prover would like to prove that a malicious client has sent it an invalid set of precompiles, we could require the client to supply a signature of `(Vec<PrecompileRequest>, VmProof)` in order to slash the client. This would be a heavy computation to prove and may not really be worth it, and I'm not event sure it's possible to accurately account for the computation the prover would have to do in order to recover some form of bounty. This might be a situation where we would rather use the consensus layer and gossip the full data with signature to all validators so that a majority can attest that the signature/data is invalid.

## Integration with the Precompile VM

This point is definitely up for discussion as it touches on the design of the precompile VM which we still need to figure out. I think the approach we'd want would be logUp-based, in which we prove a collection of statements that make up the aggregated commitment. 

In particular, we would just initialize the prover with the collection of requests as the only advice. During trace generation, each request would populate one or more rows which would then be connected with lookups. The responsibility of mapping a precompile request to rows is on the chip.



## Handling more complex data structures

Since the `data` portion of a request is really a "native" concept, we could very well just serialize the `BTreeMap` using our existing serialization framework. When we are populating the Precompile VM trace, the chiplet would map the `PrecompileRequest.data` into whatever rows are required.

processor/src/lib.rs
```diff
@@ -897,16 +897,13 @@ impl<'a> ProcessState<'a> {
...
-    /// without modifying the state of the stack), and verifies that memory range is valid.
+    /// Returns a verified memory address range given untrusted `u64` values `start_addr` and
+    /// `end_addr`.
     pub fn get_mem_addr_range(
         &self,
-        start_idx: usize,
-        end_idx: usize,
+        start_addr: u64,
+        end_addr: u64,
     ) -> Result<core::ops::Range<u32>, MemoryError> {
```
bobbinth, line 947: Rather than modifying this method, I was thinking we'd add another method that would look something like this:
```rust
pub fn get_memory_range(&self, range: Range<u32>) -> Result<Vec<Felt>, MemoryError> {
    ...
}
```
This would basically be used to read a contiguous region of memory and return a vector of field elements. The method would be pretty simple (like 2 lines of code) - so, whether it is useful is debatable.

As for `get_mem_addr_range()` - I'd probably keep it as it was.

bobbinth:
> This was something I considered originally, but I realized that we really only need the `data` portion. It should be exactly the information we need to compute the mapping from `PrecompileRequest` to `PrecompileCommitment`. It should always be "enough" since we can store more data in it if necessary. The idea is that this information would be serialized natively to ensure the request is small since these would sometimes be sent over the network. 

Yes, agreed that there is enough data to convert `PrecompileRequest` to `PrecompileCommitment` - but I think the main difference is in the design that I outlined we don't need to rely on any extra code to do this conversion. Basically, we'd be able to do something like:

```rust
impl PrecompileRequest {
    pub fn to_commitment(&self) -> PrecompileCommitment {
        ...
    }
}
```

In the current approach, we need `PrecompileVerifier` (and associated structs) to make this conversion. So, the main thing I'm trying to understand is whether we need the structure associated with `PrecompileVerifiers` for other reasons - because otherwise, we can make this PR quite a bit smaller. So, if the main reason here is to reduce the serialized size of `PrecompileRequest`, I'm not sure it is worth the complexity. The requests we'll be sending are likely to be pretty small (e.g., probably less than 100KB in vast majority of cases) and definitely smaller than proofs we'd be getting in response.

To summarize, I think the approaches are:

#### Approach 1 (in this PR)

Precompile request looks like:
```rust
pub struct PrecompileRequest {
    pub event_id: EventId,
    pub data: Vec<u8>,
}
```

Pros:
- The serialization is as compact as it can be.
- Maybe better way to check validity of the requests in the precompile VM.

Cons:
 - We need `PrecompileVerifier` trait and `PrecompileVerifiers`.
 - The precompile VM may need to do specialized pre-processing for each request type.

#### Approach 2

Precompile requests looks something like:
```rust
pub struct PrecompileRequest {
    pub tag: Word,
    pub commitment: Word,
    pub data: Vec<Felt>,
}
```

Pros:
- No need for `PrecompileVerifier` trait and `PrecompileVerifiers`.
- No need to do any pre-processing of the requests in the precompile VM.

Cons:
 - The sterilized size would be bigger - potentially up to 2x bigger than in the first approach.
 - May not catch invalid requests before precompile VM execution.

----------------

> One solution would be to use the final rate portion of the final hash (which we currently set to `[ZERO, ZERO]`) to store a commitment to the mapping in the form of a sequential hash `[VERIFIER_COMMITMENT, event_id, 0, 0, 0]`. If we consider only the recursive verification scenario, then the commitment could be the ACE circuit for that particular precompile, or even some form of on-chain commitment. The main point is that we can keep a short identifier to each request by maintaining a map of `VERIFIER_COMMITMENT |-> precompile_id`. This would be similar to the way strings are encoded in the upcoming BITE serialization format.

I'm not sure I fully followed this part. My understanding is that to compute the hash of all precompile commitments, for each request we absorb the following tuple into the precompile accumulator:
```
[TAG, COMMITMENT]
```
Where `TAG` contains the data needed to uniquely identify the precompile, and `COMMITMENT` is hash of the calldata (maybe I should refer to it as `CALLDATA_HASH`) - which uniquely defines the instance of the computation identified by the `TAG`.

So, as long as we make the precompile ID of the tag big enough (maybe 2 elements is enough?) - it probably doesn't really matter way the actual value is. So, for example, we could use the first two elements of the word to which the event name hashes (i.e., 1 more element as compared to the event ID). The important thing would be that both, the main VM and the precompile VM use the same ID - but I think that should be relatively easy to ensure.

adr1anh:
> > This was something I considered originally, but I realized that we really only need the `data` portion. It should be exactly the information we need to compute the mapping from `PrecompileRequest` to `PrecompileCommitment`. It should always be "enough" since we can store more data in it if necessary. The idea is that this information would be serialized natively to ensure the request is small since these would sometimes be sent over the network.
> 
> Yes, agreed that there is enough data to convert `PrecompileRequest` to `PrecompileCommitment` - but I think the main difference is in the design that I outlined we don't need to rely on any extra code to do this conversion. Basically, we'd be able to do something like:
> 
> ```rust
> impl PrecompileRequest {
>     pub fn to_commitment(&self) -> PrecompileCommitment {
>         ...
>     }
> }
> ```
> 
> In the current approach, we need `PrecompileVerifier` (and associated structs) to make this conversion. So, the main thing I'm trying to understand is whether we need the structure associated with `PrecompileVerifiers` for other reasons - because otherwise, we can make this PR quite a bit smaller. So, if the main reason here is to reduce the serialized size of `PrecompileRequest`, I'm not sure it is worth the complexity. The requests we'll be sending are likely to be pretty small (e.g., probably less than 100KB in vast majority of cases) and definitely smaller than proofs we'd be getting in response.

I think we might be looking at the problem in different ways and arriving at different conclusions about the way we should structure the requests and corresponding commitments. I'll elaborate on my perspective which I'm basing off a more theoretical model of what statements we need to prove in native or with the Precompile VM. I'll abbreviate Miden/Precompile VM to MVM/PVM for brevity in what follows. 

We can model a precompile as a function which reads some input $x$ from the current processor state, and returns some result $y = f(x)$ via the advice provider. For hashes, the result is just the hash of the input, while for signatures the result is just a bit indicating the validity of the result. 

We then transform this into a precompile commitment $(t, c)$ composed of a tag and a commitment to a witness $w = (x,y)$. The exact shape of the commitment may vary since it could depend on what is easiest to compute in both the MVM and PVM. We may for example set it to `RPO(RPO(x) | RPO(y))` rather than $RPO(x | y)$. We'll encapsulate this as another function $(t,c) \gets g(x, y)$.

The NP relation for this single precompile call is
*"there exists an input $x$ such that $(t,c) = C(x, f(x))$"*. On the MVM side, we are given $y \gets f(x)$ non-deterministically, so we have $(t,x) \gets C(x,y)$ so the commitment function must be efficient to compute in MASM. Natively, we can write the composition as $V(x) = C(x, f(x))$, which would be implemented as the `PrecompileVerifier`. More specifically, it's the mapping from `PrecompileRequest` to a `PrecompileCommitment`. To ensure the output commitment is correct, the verifier must actually recompute the result $y \gets f(x)$. 

Now if we look at the list of precompile requests made by the MVM, we will return a list $\[ t\_i, x\_i\]\_i$  with the tag and input. For each entry $i$, we will compute $(t\_i, c\_i) \gets V\_{t\_i}(x\_i)$, which is the same as what the MVM computes, and then the aggregated precompile commitment is given by $c_{pc} = \text{RPO}( t_1 | c_1 | t_2 | c_2 | \ldots)$. Since this is the only value that is output by the MVM, we must be able to recompute it exactly, and can only do so from the inputs $x_1, x_2, \ldots$ using the `PrecompileVerifier`s. 

More concretely, the `KeccakPreimage::to_precompile_commitment` which is called by the `keccak_verifier` is exactly the function $V(x)$ since it recomputes the keccak hash and then commits to the input and the result in the same way as the MASM wrapper does. Therefore, at least for the native verification scenario, we need an actual "native" function to be able to verify that the final commitment output by the MVM is correct. I can't really think of a scenario where where we'd need to use the cached `commitment` since we would have to verify it's validity by recomputing it anyway. 


The section above only really touches on how a native verifier would function, so I'll focus on how I imagine the PVM would work. My main point of focus here is the map from a `PrecompileRequest` to a set of rows to be populated in the trace. I'm also basing my assumptions on the row-generation that are defined in plonky3 for keccak, blake3 and Poseidon.

My assumption is that we would initialize the PVM with the list of `PrecompileRequest` in the same order they were requested. For each request, we need a native function which generates one or more rows in the trace(s) corresponding to that particular precompile. In [`blake3-air`](https://github.com/Plonky3/Plonky3/blob/e6d473caf696d523670d67c5e81b8549d5f431d6/blake3-air/src/generation.rs#L49) for example, the input for generating a row is a list of `u32` which is easily obtainable from just the `data` portion of the request. If it was represented as `Vec<Felt>`, then we would need to convert them to `u64`s and check they are correctly packed. As a side-effect of the row generation, we also obtain the result of the precompile. 

We would also have to account for the fact that the constraints would need to request the hash of the input and output of the precompile using a common RPO hasher chip. The details are a bit murky here, but again we end up with the side-effect that generating the witness for the RPO hash gives us the commitment of the precompile call which we would then be able to accumulate into the aggregated commitment. This latter word will be the output of the PVM, and should match the commitment computed by the MVM. 


> #### Approach 1 (in this PR)
> Precompile request looks like:
> 
> ```rust
> pub struct PrecompileRequest {
>     pub event_id: EventId,
>     pub data: Vec<u8>,
> }
> ```
> 
> Pros:
> 
> * The serialization is as compact as it can be.
> * Maybe better way to check validity of the requests in the precompile VM.
> 
> Cons:
> 
> * We need `PrecompileVerifier` trait and `PrecompileVerifiers`.
> * The precompile VM may need to do specialized pre-processing for each request type.

- I agree compactness should not be something we necessarily want to optimize for, but it is a nice side-benefit of having a raw representation. 
- Checking the validity of precompiles is especially useful when the PVM prover doesn't trust the client and wants to make sure it will even be able to generate a valid proof before starting it.
- The mapping of a request to its commitment needs to be a native function and I'm not sure we find a way to do this without traits, unless we're able to find a uniform representation like WASM which won't require dynamic traits.
- Most precompiles will usually treat their inputs as bytes/`u32` anyway, and in the case where it accepts `Felt`s, the conversion from bytes will usually be cheaper. 
	- If the pre-processing you're thinking about is about the computation of the commitment, we can always obtain these by recomputing the commitment while we're performing the verification.

> #### Approach 2
> Precompile requests looks something like:
> 
> ```rust
> pub struct PrecompileRequest {
>     pub tag: Word,
>     pub commitment: Word,
>     pub data: Vec<Felt>,
> }
> ```
> 
> Pros:
> 
> * No need for `PrecompileVerifier` trait and `PrecompileVerifiers`.
> * No need to do any pre-processing of the requests in the precompile VM.
> 
> Cons:
> 
> * The sterilized size would be bigger - potentially up to 2x bigger than in the first approach.
> * May not catch invalid requests before precompile VM execution.

- I'm not sure what the alternative would be here since we pretty much need to compute the result and the commitment.
- We don't really need preprocessing, it doesn't seem like it would provide much benefit when it comes to generating the PVM trace which will recompute the result anyway when it generates the row.
- Agreed about serializing not being a priority
- Catching invalid requests seems essential for a network prover to prevent DDoS.

----

### Alternative #3

There may be another perspective in which the `data` portion of the request actually contains the PVM row which is generated when the handler is called. This would make the handler invocation slightly slower since we're doing more work, but we end up with a more generic structure which could be verified by evaluating the AIR constraints of the precompile over the row. 

**Pros**:
- Easier to fill the PVM trace.
- Generic `PrecompileRequest` structure where the `data` portion is just a row of `Felt`s.
- Maybe easier to check the validity of `commitment` since we just hash *some* of the field elements in the request.
**Cons**:
- Slower handler
- Verification is more expensive since we evaluate the AIR constraints rather than performing the native computation
- Verification still requires a trait that evaluates the AIR constraints (this is a native function)
- More complex precompiles are harder to encode as a single vector
	- For an ECC computation, we would fill some rows with base field computations, as well as separate rows for the group operations.

--- 

While I may be looking at this problem from a different perspective, I'm not seeing a way where we can avoid traits when it comes to verifying the requests natively. Even if we're in a situation where we only prove precompiles using the PVM, it seems like we still need an efficient way to assert the requests are valid before starting the proof. Moreover, we will currently need to support native verification until the PVM is implemented.

prover/Cargo.toml
```diff
@@ -23,6 +23,7 @@ std = ["miden-air/std", "miden-debug-types/std", "miden-processor/std", "winter-
 [dependencies]
 # Miden dependencies
 miden-air.workspace = true
+miden-core.workspace = true
```
bobbinth: Is adding this dependency needed primarily to import `PrecompileRequest` from `miden-core`? If so, maybe we can re-export it from `miden-processor` and here just get it from the existing `miden-processor` dependency?

verifier/src/lib.rs
```diff
@@ -52,12 +56,37 @@ pub use miden_air::ExecutionProof;
...
+/// Returns any error produced by [`verify`], as well as any errors resulting from precompile
+/// verification
+#[tracing::instrument("verify_program", skip_all)]
+pub fn verify_with_precompiles(
+    program_info: ProgramInfo,
+    stack_inputs: StackInputs,
+    stack_outputs: StackOutputs,
+    proof: ExecutionProof,
+    precompile_requests: &[PrecompileRequest],
+    precompile_verifiers: &PrecompileVerifiers,
```
bobbinth: I wonder if it is a bit premature to introduce this function. Some reasons for this:

- We need a generic way for the VM to output precompile commitment for this to work. And while we are moving in this direction, we are not there yet. If, on the other hand, we end up not having such a generic way, this function would not be possible here (it would need to be implemented downstream - e.g., in [TransactionVerifier](https://github.com/0xMiden/miden-base/blob/next/crates/miden-tx/src/verifier/mod.rs#L16)).
- To make this function a bit more ergonomic, it may be a good idea to include the list of precompiles into `ExecutionProof` (as described in https://github.com/0xMiden/miden-vm/issues/2187).
- We may also want to consider making the verifier "stateful" so that it is instantiated with a set of available verifiers. Though, this can probably also be done downstream - e.g., transaction verifier could look like:

```rust
pub struct TransactionVerifier {
    tx_program_info: ProgramInfo,
    precompile_verifiers: PrecompileVerifiers,
    proof_security_level: u32,
}
```

To be clear, I don't mind having this function here with the "caveat" that it'll probably evolve or maybe even will be removed entirely in the future.

verifier/src/lib.rs
```diff
@@ -14,16 +14,20 @@ use miden_core::crypto::{
...
+};
 pub use winter_verifier::{AcceptableOptions, VerifierError};
 use winter_verifier::{crypto::MerkleTree, verify as verify_proof};
 pub mod math {
     pub use miden_core::{Felt, FieldElement, StarkField};
 }
 pub use miden_air::ExecutionProof;
-
+use miden_core::precompile::PrecompileRequest;
 // VERIFIER
```
bobbinth, line 32: nit: I'd add a newline above this line.

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -85,40 +85,49 @@ end
...
+    dup.1 add.3 u32shr.2 swap
+    # => [ptr, len_u32, len_bytes, ...]
 
     exec.rpo::hash_memory
-    # => [COMM_INPUT, ...]
+    # => [COMM_INPUT, len_bytes, ...]
+
+    # Prepare TAG = [0, 0, len_bytes, KECCAK_HASH_MEMORY_EVENT]
+    push.KECCAK_HASH_MEMORY_EVENT movup.5 push.0.0 swapw
+    # => [COMM_INPUT, TAG, ...]
```
bobbinth, line 113: I think this approach to computing the `TAG` is fine for now, but as discussed in some other comments, we probably should make the precompile identifier bigger than 64 bits. Let's create an issue for this (or maybe just add to one of the existing issues?).

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// PRECOMPILE VERIFIERS REGISTRY
+// ================================================================================================
+
+/// Registry of precompile verifiers.
+///
+/// This struct maintains a map of event IDs to their corresponding verifiers.
+/// It is used to verify precompile requests during proof verification.
+#[derive(Default, Clone)]
+pub struct PrecompileVerifiers {
```
bobbinth: Should we call this `PrecompileVerifierRegistry`? Also, I wonder if there is a better term than "verifier" - we usually use "verifier" in the context of verifying ZK proofs - and it may get a bit confusing if we use it here as well.

bobbinth (commented):
Looks good! Thank you! I left a few more comments line. Also, I'd probably rever the change to `get_mem_addr_range()` as mentioned in https://github.com/0xMiden/miden-vm/pull/2183#discussion_r2374385739.

In regards to https://github.com/0xMiden/miden-vm/pull/2183#issuecomment-3328096794 - I think we are largely on the same page, though I think there are maybe two main differences in how we think about precompiles:

First, I view "native verification" as a temporary solution that could go away in the medium term (e.g., in the next 4 - 5 months) because we'd transition to the recursive verification approach once PVM is available. This is mostly because I'm not seeing too much use for "native verification" in the context of the blockchain protocol. That is, even if we allow "native verification" at the transaction level, we still would need to recursively verify all of the batch level. This means we need the PVM to work regardless, and once it is available, there is little reason to use "native verification".

There may, however, be one exception to this: we could collect a bunch of transactions with deferred precompiles, and then recursively verify a single PVM proof for all transactions at the batch level. This could be a very interesting use case, and so, maybe I should re-think my initial view. (also, cc @PhilippGackstatter and @mmagician to double-check my thinking on this).

So, if we do want to keep "native verification" long-term because of the above use case or maybe because it would be useful outside of the blockchain protocol context, then I think the direction you've taken makes much more sense.

Second, I think of a set of precompiles as relatively static (at least in the context of the blockchain protocol). We'll start out with a handful of precompiles (e.g., Keccak, ECDSA, and EdDSA), and any time we'd need to add a new precompile that would require a kernel modification. This is because the epilogue of either the transaction or the batch kernel would need to run a specific PVM verifier, and as expand capabilities to the PVM, we'd need to make corresponding updates to the transaction/batch kernels. These things for sure will happen, but I'm thinking it'll be more like adding 1 - 2 precompiles every 6 - 12 months (though, maybe I'm being pessimistic on this).

So, I'm not seeing a use case where, in the context of the blockchain protocol, we'd be able to handle precompiles dynamically. And so, from this perspective, having a structure that supports a dynamic set of precompiles *feels* a bit like an overkill.

But maybe here as well I'm taking a bit too narrow of a view and the structure is not that complicated, and even if we only have a handful of precompiles it is still pretty useful.

mmagician:
> we could collect a bunch of transactions with deferred precompiles, and then recursively verify a single PVM proof for all transactions at the batch level

I'm not sure I fully understand the use case. Do you view this as an _optional alternative_ to verifying transactions in a batch kernel, or as a _replacement_?

core/src/precompile.rs
adr1anh: Good call on the struct name. I agree that "verifier" could be a bit misleading, would `PrecompileChecker` work?

verifier/src/lib.rs
adr1anh: I think this gets automatically changed by `rustfmt` but I'll see if I can get it to do the correct thing here.

bobbinth:
> > we could collect a bunch of transactions with deferred precompiles, and then recursively verify a single PVM proof for all transactions at the batch level
> 
> I'm not sure I fully understand the use case. Do you view this as an _optional alternative_ to verifying transactions in a batch kernel, or as a _replacement_?

I'm thinking of this more like of a complement to batch building. For example, let's say a user builds a transaction which requires a bunch of Keccaks (and maybe some ECDSAs) - but they don't want to (or can't) use delegated proving. So, they send the transaction with precompile request commitment to the network (i.e., they don't bother to build a proof for these requests neither for recursive purposes nor for native verification purposes).

The batch builder can then do the following:
- Collect a set of transactions with "unverified precompile requests".
- Combine all of these requests into a single precompile request commitment. This would require some logic in the batch kernel - e.g., to "unhash" all individual requests and then combine them into a single "big" request - but I don't think this logic is going to be too complicated.
- Then, the batch builder runs the precompile VM and generates a single proof that covers all transactions in the batch.
- Then, as a part of the batch kernel, the batch builder:
  - Verifies all individual transaction proofs (as it would do anyway).
  - Verifies the single PVM proof that covers all deferred precompile requests from the transactions in the batch.

To summarize this model: instead of using 3rd-party delegated provers, users are using the network (batch builders) for proof delegation. This probably leaks a bit more privacy - but depending on the use case, this could be acceptable.

mmagician:
> The batch builder can then do the following:
>
> - Collect a set of transactions with "unverified precompile requests".
> - Combine all of these requests into a single precompile request commitment

I see, so the need for native verification would be at the batch builder level, where they natively verify the individual precompile requests from users, right?
If that's the case, then this is not really protocol-level verification. Rather, I see this as a sanity check / spam protection that the builder performs to make sure they don't waste resources on generating proofs for invalid precompile requests.
The proof generation for the big request later guarantees the validity at the protocol level, but for this the batcher only needs recursive verification.

bobbinth:
> I see, so the need for native verification would be at the batch builder level, where they natively verify the individual precompile requests from users, right?
> If that's the case, then this is not really protocol-level verification. Rather, I see this as a sanity check / spam protection that the builder performs to make sure they don't waste resources on generating proofs for invalid precompile requests.
> The proof generation for the big request later guarantees the validity at the protocol level, but for this the batcher only needs recursive verification.

I think it is both: the question is whether we need to have the capability for transaction proofs not to include precompile proofs (i.e., only include a commitment to used precompiles). In the use case I've described, the batch builder needs the info about used (but not proven) precompiles so that they can generate the proof for them. And the same info can be used to do "native verification" as a spam protection mechanism.

verifier/src/lib.rs
adr1anh: I originally did put the precompiles in the execution proof but wanted to minimize the API modifications in this PR. It should be straightforward to remove later as you mention.

adr1anh:
The ability to delegate precompiles to the network was something I had in mind. There are clearly some issues to iron out here which merit a separate discussion IMHO. 

Regarding the set of precompiles we are aiming to support, I agree that we'll likely only consider a fixed set of them. This is the approach taken by SP1 and we can see that their [set of precompile](https://github.com/succinctlabs/sp1/tree/dev/crates/core/machine/src/syscall/precompiles) has not changed in a while. We might support a few more options, but overall I think there is little benefit to making them dynamic or considering the use case where users would want to write them themselves.

If we go this route, the trait might be slightly overkill but it might internally be useful for debugging, and perhaps simplifying some other parts of the code. For now though, I'd vote for leaving it in this PR and consider removing it later when the full picture becomes clearer. It would unblock EdDSA/ECDSA implementation PR for this release.

Finally, another benefit of having fixed precompiles defined by the VM is that we can make the tags simpler. I'd propose using two field elements for the precompile ID (or even just an index in the list of verifiers) along with a version. Again, not fully fledged out but I have a hunch this is enough.

processor/src/lib.rs
adr1anh: We could fix this more generally but I'll leave it as is for this PR. It's very little work for the caller to get the address from the stack before calling this method and would make this method more generic since we can pass an arbitrary range.

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: Linked to this comment in #2147

verifier/src/lib.rs
adr1anh: Fixed using a different syntax. Let me know if this works. If so, I think this might be worthwhile to apply across the codebase to make it clearer what we're exporting

```rust
mod exports {
  use crate::Type;
}
pub use exports::*;
```

core/src/precompile.rs
adr1anh: In some sense, these are "verifiers" in terms of the theoretical NP definition, though I would also call it a "decider" since it's performing computation. I'll leave verifier for now until we find a better name.

verifier/src/lib.rs
bobbinth: Ah - interesting! Yes, I think the new syntax works.

core/src/precompile.rs
bobbinth: I guess what throws me off a bit is that the verifiers don't actually verify the validity. For example, for Keccak, we are just computing the hash, but the `verify()` function doesn't have enough info to figure out if the computed result is valid (the verification step comes later after we combine all commitments and check the result against the commitment output by the VM).

To me, this feels more like "precompile commitment builder" as it is main job is to convert precompile request into precompile commitment. But that's not a great name either.

bobbinth (commented):
Looks good! Thank you! I think the only thing left to address here is https://github.com/0xMiden/miden-vm/pull/2183#discussion_r2390503794 - but I would love for @Al-Kindi-0 and @huitseeker to look through this as well.

> If we go this route, the trait might be slightly overkill but it might internally be useful for debugging, and perhaps simplifying some other parts of the code. For now though, I'd vote for leaving it in this PR and consider removing it later when the full picture becomes clearer. It would unblock EdDSA/ECDSA implementation PR for this release.

Agreed. An alternative could be to define an enum with all supported precompiles - but it is not clear to me that that's better. So, let's keep traits for now, and we can refactor later, if needed.

> Finally, another benefit of having fixed precompiles defined by the VM is that we can make the tags simpler. I'd propose using two field elements for the precompile ID (or even just an index in the list of verifiers) along with a version. Again, not fully fledged out but I have a hunch this is enough.

I think i'd probably go with two field elements. They could even be derived in the same way as the precompile event ID, but instead of taking a single element, we could take two elements of the resulting hash. But we can address this in a follow-up PR.

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+/// Represents a single precompile request consisting of an event ID and byte data.
+///
+/// This structure encapsulates the call data for a precompile operation, storing
+/// the raw bytes that will be processed by the precompile function.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileRequest {
+    /// Event ID identifying the type of precompile operation
+    pub event_id: EventId,
+    /// Raw byte data for the precompile operation
+    pub data: Vec<u8>,
```
Al-Kindi-0: Q: Did we use to call this calldata? If yes, why did we change? I guess now the call data is the combination of `event_id` + `data`, but I wonder if it should be just the latter that we call `call_data`

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+impl Serializable for PrecompileRequest {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        self.event_id.as_felt().write_into(target);
```
Al-Kindi-0: nit: should we implement serialization on `EventId`? This way it will be smoother to change the representation in the future

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// PRECOMPILE COMMITMENT
+// ================================================================================================
+
+/// Represents the result of a verified precompile computation.
+///
+/// This structure contains both the tag (which includes metadata like event ID)
+/// and the commitment word that represents the verified computation result.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileCommitment {
```
Al-Kindi-0: Would `PrecompileRequestDigest` be more appropriate? Also the first line of the comment is a bit incongruent, at least to me, but maybe I am missing something else

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+        // We add 2 empty words to account for the finalization of the hash inside the VM.
+        // The VM keeps track of the sponge's capacity only, so once all precompile request
+        // commitments have been absorbed, the finalization requires one last permutation where
+        // we set the rate portion of the state to the zeros.
+        // This slot could be used to encode auxiliary metadata for the entire list of precompile
+        // requests.
+        final_commitments.extend([Felt::ZERO; 8]);
+        Rpo256::hash_elements(&final_commitments)
+    }
```
Al-Kindi-0: Q: Would it make sense replacing these two functions with a `defer_precompile_calls` function which takes an initial (optional) digest and a number of (ordered) `PrecompileRequest`-s, and produces a digest representing the digest of all differed precompile calls we need to verify?
In the same spirit, verifying + "committing" to a precompile becomes just one atomic operation, namely deferring a precompile request, which both checks the validity of the request and reduces to a digest it for succinctness.

core/src/precompile.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// PRECOMPILE VERIFIER TRAIT
+// ================================================================================================
+
+/// Trait for verifying precompile computations.
+///
+/// Each precompile type must implement this trait to enable verification of its
+/// computations during proof verification. The verifier validates that the
+/// computation was performed correctly and returns a precompile commitment.
+pub trait PrecompileVerifier: Send + Sync {
```
Al-Kindi-0, line 252: Related to the above is, `PrecompileDeferrer`? (Although not very standard :) )

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -85,40 +85,49 @@ end
 
 #! Internal implementation of memory-based Keccak256 computation via deferred computation.
 #!
-#! Emits event to populate advice stack with digest, computes commitment for tracking
-#! deferred computations, and stores witness in advice map for later proof generation.
+#! Emits an event to populate advice stack with digest, and storing the preimage for deferred verification.
```
Al-Kindi-0: nit 
```
Emits an event to populate advice stack with digest and store the preimage for deferred verification.
```

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -85,40 +85,49 @@ end
...
+#! It returns the commitment to the precompile call and its tag.
```
Al-Kindi-0: nit: remove `it`

Al-Kindi-0 (commented):
Looks awesome!
Did an initial pass and left mostly questions

core/src/precompile.rs
adr1anh: That's a good catch, I think calldata is better

core/src/precompile.rs
adr1anh: I'd maybe go for `PrecompileRequestCommitment` but that's also a bit verbose. Agreed about the first line.

core/src/precompile.rs
bobbinth: Some potential options to replace "verifier":

- `PrecompileCommitmentBuilder` - probably the most accurate, but quite verbose.
- `PrecompileHasher` - short, but probably not very accurate.
- `PrecompileProcessor` - relatively short, but maybe too generic.
- `PrecompileDecoder` - also short, but not sure if accurate.

Overall, I think this trait could look something like this:

```rust
pub trait TBD: Send + Sync {

    /// Returns a unique identifier of the precompile.
    fn id(&self) -> PrecompileId;

    /// Computes a precompile commitment for the specified calldata.
    ///
    /// The commitment is equivalent to the commitment output by the Miden VM for a deferred
    /// precompile computation.
    ///
    /// # Errors
    /// Returns an error if the provided calldata is malformed or is not a valid calldata for
    /// for this precompile.
    fn get_commitment(&self, calldata: &[u8]) -> Result<PrecompileCommitment, PrecompileError>;
}
```

core/src/precompile.rs
```diff
@@ -0,0 +1,282 @@
...
+// ================================================================================================
+
+/// Represents a single precompile request consisting of an event ID and byte data.
+///
+/// This structure encapsulates the call data for a precompile operation, storing
+/// the raw bytes that will be processed by the precompile function.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileRequest {
+    /// Event ID identifying the type of precompile operation
+    pub event_id: EventId,
```
huitseeker: Could these fields be made private?

stdlib/src/handlers/keccak256.rs
```diff
@@ -169,18 +140,160 @@ impl KeccakFeltDigest {
...
```
huitseeker: Could this be a private with an accessor method rather than exposing the inner vec  directly?

core/src/precompile.rs
```diff
@@ -0,0 +1,282 @@
...
+
+// PRECOMPILE REQUEST
+// ================================================================================================
+
+/// Represents a single precompile request consisting of an event ID and byte data.
+///
+/// This structure encapsulates the call data for a precompile operation, storing
+/// the raw bytes that will be processed by the precompile function.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileRequest {
```
huitseeker, line 85: @bobbinth left a [comment](https://github.com/0xMiden/miden-vm/pull/2183/files#r2371585256) on how binding the `EventId` is, and how we make sure the right verification function gets used for each request.

It's completely fine to use this PR as scaffolding for those concerns to be addressed later, but I'd like to manage expectations for future readers: could you add a comment explaining what security concerns we should keep in mind when using this struct?

core/src/precompile.rs
```diff
@@ -0,0 +1,282 @@
...
+// TESTS
+// ================================================================================================
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::utils::SliceReader;
+
+    #[test]
+    fn test_precompile_data_serialization() {
```
huitseeker: You may want to use one of the macros which allow you to do a serialization roundtrip (given `Arbitrary`) from #2224.

core/src/precompile.rs
```diff
@@ -0,0 +1,282 @@
...
+
+// PRECOMPILE VERIFIER TRAIT
+// ================================================================================================
+
+/// Trait for verifying precompile computations.
+///
+/// Each precompile type must implement this trait to enable verification of its
+/// computations during proof verification. The verifier validates that the
+/// computation was performed correctly and returns a precompile commitment.
+pub trait PrecompileVerifier: Send + Sync {
```
huitseeker, line 252: Security design thought: Since we're still figuring out the verification correctness (as Bobbinth mentioned in his comment), should we consider making this trait sealed for now?

Two options:
1. [Sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html) pattern - only allow implementations in this crate while we nail down the security model
2. [enum_dispatch](https://docs.rs/enum_dispatch/latest/enum_dispatch/) approach - list known precompiles in an enum and generate static dispatch, replace your `dyn Trait` with the enum. 

Both would limit the attack surface until we're confident about the security properties. Right now folks outside our code base can implement `PrecompileVerifier`. Granted, the `Registry` would allow us to check what we're passed at runtime, but that's something we intend to keep for gating, rather than something meant to tell developers "we're in charge of the precompile development for now, this is an alpha, nothing to see there".

I think one option could be a good temporary measure that we can relax later once the verification model is more mature. What do you think?

core/src/event_id.rs
```diff
@@ -74,3 +76,15 @@ impl Display for EventId {
         core::fmt::Display::fmt(&self.0, f)
     }
 }
+
+impl Serializable for EventId {
```
huitseeker, line 95: An implementation of Arbitrary with a serialization round-trip would do nicely.

core/src/precompile.rs
```diff
@@ -0,0 +1,282 @@
...
+
+// PRECOMPILE VERIFICATION STATE
+// ================================================================================================
+
+/// Tracks the RPO256 sponge capacity for aggregating [`PrecompileCommitment`]s.
+///
+/// This structure mirrors the VM's implementation of precompile commitment tracking. During
+/// execution, the VM maintains only the capacity portion of an RPO256 sponge, absorbing each
+/// precompile commitment as it's produced. At the end of execution, the verifier recomputes
+/// this same aggregation and compares the final digest.
```
huitseeker, line 274: The documentation here could be more helpful with concrete examples and internal links. Could you add rustdoc links to the commitment tracking implementation this mirrors?

stdlib/src/handlers/keccak256.rs
adr1anh: In this case it seems fine, since any underlying vector of bytes is a valid pre-image. Are there any negatives associated with making the field public?

core/src/precompile.rs
adr1anh: I guess we could, but same question as below: are there benefits to making them private?

adr1anh: I just saw [Structs have private fields](https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private). I'll add a constructor since the representation here is likely to change

core/src/precompile.rs
adr1anh: I like the idea of the sealed trait here, which we can easily loosen if we continue this current approach.

For later though, an enum dispatch seems more appropriate given that we're wanting to support only a fixed set of precompiles, at least in the medium term. 

I'll create an issue to investigate what that would look like more concretely so as to not change this PR too much.

verifier/src/lib.rs
```diff
@@ -52,20 +63,52 @@ pub use miden_air::ExecutionProof;
...
 /// - The protocol parameters used to generate the proof are not in the set of acceptable
 ///   parameters.
-#[tracing::instrument("verify_program", skip_all)]
+/// - The proof contains any precompile requests.
 pub fn verify(
     program_info: ProgramInfo,
     stack_inputs: StackInputs,
     stack_outputs: StackOutputs,
     proof: ExecutionProof,
 ) -> Result<u32, VerificationError> {
```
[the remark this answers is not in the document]
adr1anh: This is done automatically when verifying the precompiles in the function below

stdlib/docs/crypto/hashes/keccak256.md
```diff
@@ -5,4 +5,4 @@
...
+| hash_memory_impl | Internal implementation of memory-based Keccak256 computation via deferred computation.<br /><br />Emits an event to populate the advice stack with digest and store the preimage for deferred verification.<br />Returns the commitment to the precompile call and its tag.<br /><br />Input: [ptr, len_bytes, ...]<br />Output: [COMM, TAG, DIGEST_U32[8], ...]<br /><br />Where:<br />- ptr: word-aligned memory address containing INPUT_U32[len_u32] where len_u32=⌈len_bytes/4⌉<br />- len_bytes: number of bytes to hash<br />- COMM: Rpo256(Rpo256(INPUT_U32[..]) \|\| Rpo256(DIGEST_U32[..])) for kernel tracking<br />- TAG = [0, 0, len_bytes, KECCAK_HASH_MEMORY_EVENT] for precompile verification<br />- DIGEST_U32[8] = [d_0, ..., d_7] = Keccak256(INPUT_U8[len_bytes])<br /> |
```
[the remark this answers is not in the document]
adr1anh: Even Copilot gets confused by our ordering

core/src/precompile.rs
adr1anh: Done

core/src/precompile.rs
adr1anh: This doesn't actually exist yet, but will be in #2249. I made a note there to link it later.

core/src/precompile.rs
adr1anh: Hmm upon further inspection, it seems non-trivial to use a sealed trait since we're working across different crates. 

Overall though, I think the best solution at this time is to just warn developers that the precompile framework is unstable. If they do decide to implement one, then it's their responsiblity to both ensure it's secure and to deal with any breaking changes.

core/src/precompile.rs
adr1anh: I'm not sure what you meant by example here though.

core/src/precompile.rs
adr1anh: Shall we brainstorm a different name in a separate issue?

Al-Kindi-0: Sounds good and probably low priority until things stabilize across

stdlib/src/handlers/keccak256.rs
huitseeker: Yes: inner mutability which you don't need at the moment. "No is temporary, yes is forever".

core/src/precompile.rs
huitseeker: > I'm not sure what you meant by example here though.

In my wildest dreams: a doc page narrating "the lifecycle of the new Keccak256 precompile in the MidenVM". The choice of precompile being the example.

Makefile
```diff
@@ -46,7 +46,7 @@ FEATURES_LOG_TREE        := --features concurrent,executable,tracing-forest
 FEATURES_air             := testing
 FEATURES_assembly        := testing
 FEATURES_assembly-syntax := testing
-FEATURES_core            :=
+FEATURES_core            := testing,serde,arbitrary
```
huitseeker: I think the convention in most crates is the `testing` feature implies `arbitrary` (and crates with the later should have the earlier available) ?

core/src/precompile.rs
adr1anh: Created #2269 to address this thoroughly

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+    ///
+    /// # Arguments
+    /// * `data` - The byte data used for the precompile computation
+    ///
+    /// # Returns
+    /// Returns a precompile commitment containing both tag and commitment word on success.
+    ///
+    /// # Errors
+    /// Returns an error if the verification fails.
+    fn verify(&self, data: &[u8]) -> Result<PrecompileCommitment, PrecompileError>;
```
plafer:
```suggestion
    fn verify(&self, calldata: &[u8]) -> Result<PrecompileCommitment, PrecompileError>;
```
nit: name `calldata` to be consistent with `PrecompileRequest`. Helps communicate the link with `PrecompileRequest`

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+// ================================================================================================
+
+/// Tracks the RPO256 sponge capacity for aggregating [`PrecompileCommitment`]s.
+///
+/// This structure mirrors the VM's implementation of precompile commitment tracking. During
+/// execution, the VM maintains only the capacity portion of an RPO256 sponge, absorbing each
+/// precompile commitment as it's produced. At the end of execution, the verifier recomputes
+/// this same aggregation and compares the final digest.
+///
+/// The exact shape of the commitment is described in TODO(adr1anh)
```
plafer: should this be resolved in this PR?

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// # Details:
+/// - The capacity of the sponge is set to `ZERO`.
```
plafer:
```suggestion
/// - The capacity of the sponge is initialized to `ZERO`.
```

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+/// Tracks the RPO256 sponge capacity for aggregating [`PrecompileCommitment`]s.
+///
+/// This structure mirrors the VM's implementation of precompile commitment tracking. During
+/// execution, the VM maintains only the capacity portion of an RPO256 sponge, absorbing each
+/// precompile commitment as it's produced. At the end of execution, the verifier recomputes
+/// this same aggregation and compares the final digest.
+///
+/// The exact shape of the commitment is described in TODO(adr1anh)
+///
+/// # Details:
```
plafer: nit: These 3 steps map 1-to-1 with the 3 methods, so I would indicate that here, as a guideline on how to use the struct, e.g.

```
/// - `new()`: The capacity ...
/// - `absorb()`: We absorb ...
/// - `finalize()`: ...
```

verifier/src/lib.rs
```diff
@@ -52,20 +63,53 @@ pub use miden_air::ExecutionProof;
 /// - The provided proof does not prove a correct execution of the program.
 /// - The protocol parameters used to generate the proof are not in the set of acceptable
 ///   parameters.
-#[tracing::instrument("verify_program", skip_all)]
+/// - The proof contains any precompile requests.
```
plafer:
```suggestion
/// - The proof contains one or more precompile requests.
```
Is this true though? How do we get an error when there are precompiles? At least this isn't clear to me at this point in my read through.

prover/src/lib.rs
```diff
@@ -65,6 +65,25 @@ pub fn prove(
...
+/// of the program.
+#[instrument("prove_program", skip_all)]
+#[maybe_async]
+pub fn prove_with_precompiles(
+    program: &Program,
+    stack_inputs: StackInputs,
+    advice_inputs: AdviceInputs,
+    host: &mut impl SyncHost,
+    options: ProvingOptions,
+) -> Result<(StackOutputs, ExecutionProof, Vec<PrecompileRequest>), ExecutionError> {
```
plafer: Should `Vec<PrecompileRequest>` be part of `ExecutionProof`? Essentially encoding the things that are explicitly not proven yet by that proof.

processor/src/trace/mod.rs
```diff
@@ -123,6 +126,11 @@ impl ExecutionTrace {
         &self.stack_outputs
     }
 
+    /// Returns the precompile requests generated during program execution.
+    pub fn precompile_requests(&self) -> &[PrecompileRequest] {
+        self.advice.precompile_requests()
+    }
```
plafer, line 162: Note: just realizing with this how `ExecutionTrace` has grown to represent a lot more than just a trace. It looks more like an "execution output" to me. This type should be considered legacy (I don't think it's ever used with the `FastProcessor` and parallel trace generation), and to be removed when we remove `Process`.

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+
+/// A commitment to the evaluation of [`PrecompileRequest`], representing both the input and result
+/// of the request.
+///
+/// This structure contains both the tag (which includes metadata like event ID)
+/// and the commitment word that represents the verified computation result.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileCommitment {
+    /// Tag containing metadata including the event ID in the first element
+    pub tag: Word,
```
plafer: Can we be specific about what is included in the tag? And I assume it is used as the "unique identifier" to a precompile request? I would expand on why the `tag` is there in the `PrecompileCommitment` docstring

core/src/precompile.rs
```diff
@@ -0,0 +1,304 @@
...
+// PRECOMPILE COMMITMENT
+// ================================================================================================
+
+/// A commitment to the evaluation of [`PrecompileRequest`], representing both the input and result
+/// of the request.
+///
+/// This structure contains both the tag (which includes metadata like event ID)
+/// and the commitment word that represents the verified computation result.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PrecompileCommitment {
```
plafer, line 138: Should this be named `TaggedPrecompileCommitment`? Basically just the commitment doesn't seem to be sufficient, and it might be a good idea for the name to reflect that?

adr1anh: The tag serves as metadata here and can contain some extra information that is not practical to include in the `commitment` (e.g. `len_bytes`). The `commitment` alone is not enough.

plafer: Can we document & expand on that somewhere in the code? I currently don't have a good intuition for why it's needed, and providing that somewhere would be very helpful in understanding the precompile infra.

prover/src/lib.rs
adr1anh: This makes sense. It should also allow use to remove the extra verifier implementation.

core/src/precompile.rs
adr1anh: I'll address this #2269

verifier/src/lib.rs
adr1anh: I added a comment below, explaining that we would fail since there are no matching verifiers in the registry.

core/src/precompile.rs
```diff
@@ -0,0 +1,307 @@
...
+use miden_crypto::{Felt, Word, hash::rpo::Rpo256};
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    EventId,
+    utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
+};
+
+// PRECOMPILE REQUEST
```
huitseeker, line 67: Here I would add a more comprehensive module-level overview documentation here that explains the deferred verification workflow.

- Workflow Summary
    1. **VM Execution**: When precompile operations are executed, capture raw input data and create `PrecompileRequest` objects
    2. **Request Storage**: Store requests in `AdviceProvider` using `AdviceMutation::ExtendPrecompileRequests`  
    3. **Proof Generation**: Include precompile requests in `ExecutionProof` serialization
    4. **Deferred Verification**: During proof verification, use registered verifiers to recompute and validate commitments
- Concepts
    - `PrecompileRequest`: Captures raw input bytes and event identifier for deferred verification
    - `PrecompileCommitment`: Cryptographic commitment proving computation was performed correctly  
    - `PrecompileVerifierRegistry`: Maps event IDs to verification functions
    - `PrecompileVerificationState`: RPO256 sponge for aggregating commitments
- Note
    Reference the Keccak256 implementation in stdlib/src/handlers/keccak256.rs for a complete example of implementing a precompile handler and verifier.

core/src/precompile.rs
```diff
@@ -0,0 +1,307 @@
...
+// ================================================================================================
+
+/// Represents a single precompile request consisting of an event ID and byte data.
+///
+/// This structure encapsulates the call data for a precompile operation, storing
+/// the raw bytes that will be processed by the precompile function.
+///
+/// # Note:
```
huitseeker: I would expand on the Event ID selection guidelines. After reading this, developers should know after reading just those few lines here:

1. How to choose unique Event IDs for their precompiles 
2. Security considerations for ID selection, now and in the future
3. Guidelines for avoiding collisions with system events (0..256)
4. Recommended approach to ID generation (?)

stdlib/src/handlers/keccak256.rs
```diff
@@ -23,8 +26,8 @@ pub const KECCAK_HASH_MEMORY_EVENT_ID: EventId = EventId::from_u64(5779517439479
 /// Keccak256 event handler that reads data from memory.
```
huitseeker: I would take the occasion of defining `Keccak256` to add documentation on the more general handler implementation pattern. 

I'd love these lines to give me a sense of:
1. How to capture raw input bytes from memory
2. How to create `PrecompileRequest` from input data  
3. How to return both advice stack and precompile mutations
4. Where else to go read about the complete workflow for using a precompile handler

core/src/precompile.rs
```diff
@@ -0,0 +1,307 @@
...
+/// };
+/// ```
+impl<F> PrecompileVerifier for F
+where
+    F: Fn(&[u8]) -> Result<PrecompileCommitment, PrecompileError> + Send + Sync + 'static,
+{
+    fn verify(&self, calldata: &[u8]) -> Result<PrecompileCommitment, PrecompileError> {
+        self(calldata)
+    }
+}
```
huitseeker: This blanket (not "default" technically) implementation for closures  makes it annoying to restrict the implementers of `PrecompileVerifier` at compile-time .. and is permissive, since now all code paths that provide such a function "work", as opposed to one that would require specifically laying out a more explicit `impl Trait` block.

Do we really need it? (I nuke it in #2279)

Makefile
adr1anh: Right! Should we add `serde` to the `testing` feature so we also test for the derived serialization tests? I’m not 100% how this works, maybe for another PR?

core/src/precompile.rs
adr1anh: Fair, it was a lot more useful for EventHandlers because we can add simple handlers for testing. In particular, it allows the test to inject an event which can make assertions about the test using Rust rather than MASM.

core/src/precompile.rs
adr1anh: The event ID should, by convention, be the same one as the corresponding EventHandler. Given the current plan, it seems very unlikely we'll want to support a mode in which an arbitrary set of precompiles can be used as this would drastically complicate any implementation of a recursive verifier.

core/src/precompile.rs
adr1anh: This makes a lot of sense, module documentation with your suggestions have been added.

Makefile
adr1anh: I've reverted the changes, so we can do this more thoroughly in a future PR.

verifier/src/lib.rs
```diff
@@ -52,20 +60,63 @@ pub use miden_air::ExecutionProof;
...
 
     // build public inputs and try to verify the proof
     let pub_inputs = PublicInputs::new(program_info, stack_inputs, stack_outputs);
-    let (hash_fn, proof) = proof.into_parts();
+    let (hash_fn, proof, precompile_requests) = proof.into_parts();
+
+    // TODO: Check that this corresponds to the commitment output by the VM
+    // compute the commitment to the list of all precompile requests.
+    // if no verifiers were provided (e.g. when this function was called from `verify()`),
+    // but the proof contained requests anyway, returns a `NoVerifierFound` error.
```
huitseeker, line 115: Shall we open an issue?

huitseeker (commented):
Thanks a lot for this PR, the current recision looks great! Your [documentation](https://hackmd.io/@adr1anh/Byxn949aeg) clarified a *lot* about the intended design behind this PR (and the following, e.g. #2249, #2277) — much appreciated. I think I can summarize the concerns that are still relevant this way:

1. Documentation / Implementation Mismatch

  The core code documentation still describes the framework as if it supported arbitrary verifiers (`PrecompileVerifierRegistry` with dynamic registration), but the actual security model (as explained in [the doc](https://hackmd.io/@adr1anh/Byxn949aeg)) assumes a fixed set controlled by the network. This may create confusion about the intended security boundaries.

  2. Event ID Collision Still Theoretical

  While I agree this isn't an actual security issue because networks control which verifiers they accept, the framework still uses 64-bit Event IDs that could theoretically collide. In a future with many precompiles, this could become a practical concern — at least from an error handling PoV — even if it's not currently exploitable.

  3. Error Handling for Network Enforcement

  The code doesn't clearly differentiate yet between "no verifier registered" (lack of constraints the match an event, development issue) vs "unsupported precompile" (constraint mismatch, network policy). This distinction matters for the security model described in [the doc](https://hackmd.io/@adr1anh/Byxn949aeg)

  4. Precompile VM Future Compatibility

  The current [design](https://hackmd.io/@adr1anh/Byxn949aeg) assumes the "Precompile VM" will use "blessed" verification keys, but there's no obvious path for how the current `PrecompileVerifier` trait maps to those constraints yet. This could create future compatibility issues.

All of this is fine at this stage: I'm open to merging this PR as scaffolding to progress towards the full design, as soon issues are open towards bridging those apparent gaps (which the design does address).

verifier/src/lib.rs
adr1anh: this is being tackled in #2249

adr1anh:
Created https://github.com/0xMiden/miden-vm/issues/2293 which combines the discussion points. Once some details clear up, we can add sub-issues to address individual points with more exhaustive descriptions.

huitseeker (approved):
Thank you so much!

adr1anh:
Thank you all for your invaluable feedback on getting this into a much better shape. I'll merge it now to unblock other work, but feel free to comment on #2249 and I can address things there

core/src/precompile.rs
```diff
@@ -0,0 +1,359 @@
...
+///
+/// # Aggregation Process
+///
+/// - **`new()`**: Initialize capacity to `ZERO`
+/// - **`absorb(comm)`**: Apply RPO256 permutation to `[capacity, tag, commitment]` and update
+///   capacity to the first word of the result
+/// - **`finalize()`**: Apply RPO256 permutation to `[capacity, ZERO, ZERO]` and extract the second
+///   word as the final digest
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+struct PrecompileVerificationState {
```
bobbinth, line 284: We could probably mention that this process is basically the same as building a single `Vec<Felt>` vector from the values returned from `PrecompileCommitment::to_elements()` and then calling `Rpo256::hash_elements()` on this vector (i.e., the commitment is just a sequential hash of all precompile commitments).

It may also be good to mention that this works because each precompile commitment is exactly 8 elements in size, and so we are guaranteed to have the number of elements divisible by 8 (which means that we don't need to do anything special with the initial capacity elements for padding).

PR #2184, opened 2025-09-17
fix: advice_map loading in test framework for the fast processor

### LINKED ISSUE #1886: Advice Map bug in test-utils
When testing a script with an advice map using the test-utils framework, even if the map is correctly loaded by the host, the event handler is unable to find the key because the map is empty.
See `processor/src/tests.rs::test_advice_map_inline`.

There is a trait with several implementations, some just for testing so there is probably a bug in one of them or in the way they are linked.

The same script can be successfully tested using the cli where the advice map is correctly populated and queried, see `miden-vm/tests/integration/cli/cli_test.rs::test_advmap_cli`.

huitseeker:
The test framework was not properly loading advice maps from compiled programs' Mastforests into the execution context, causing "advice provider error" failures in tests that use advice maps defined in assembly source.

The slow processor was working correctly because it automatically calls `self.advice.extend_map(program.mast_forest().advice_map())` in `execute_mast_root()`. The fast processor was failing because `assert_result_with_fast_processor()` was creating the FastProcessor with only manually provided advice inputs, without merging the program's compiled advice map.

This merges that advice map. Closes #1886.

### DIFF
--- a/crates/utils/testing/src/lib.rs
+++ b/crates/utils/testing/src/lib.rs
@@ -14,6 +14,7 @@ use alloc::{
     vec::Vec,
 };
 
+use miden_air::RowIndex;
 use miden_assembly::{KernelLibrary, Library, Parse, diagnostics::reporting::PrintDiagnostic};
 pub use miden_assembly::{
     LibraryPath,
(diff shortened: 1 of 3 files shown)


adr1anh:
I haven't tested this fully, but I wonder if we could solve this by adding `self.advice.extend_map(current_forest.advice_map());` in [`execute_impl`](https://github.com/0xMiden/miden-vm/blob/487879423d25781d91ff41fed634c84bbc7ac874/processor/src/fast/mod.rs#L353) instead.

plafer:
Agreed with @adr1anh - the issue is that `Process::execute()` correctly merges the mast forest's advice map at the start of execution, but we don't do that in `FastProcessor::execute_impl()`. I double-checked that we do it properly when loading new MAST forests though (from an external node or dyn/dyncall).

huitseeker:
Thanks @adr1anh & @plafer !

PR #2185, opened 2025-09-18
fix: complete rename Span -> Basic Block

### LINKED ISSUE #2126: Fix all remaining uses of "span" (renamed to "basic block")
Not for this PR, but `BlockType::Span` is obsolete terminology and we should update it at some point.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2023#discussion_r2317860092_

huitseeker:
Closes #2126. I think I haven't missed anything (?).

### DIFF
--- a/docs/src/design/decoder/main.md
+++ b/docs/src/design/decoder/main.md
@@ -25,8 +25,8 @@ Managing control flow in the VM is accomplished by executing control flow operat
 | `SPLIT`   | Initiates processing of a new [Split block](../programs.md#split-block).     |
 | `LOOP`    | Initiates processing of a new [Loop block](../programs.md#loop-block).       |
 | `REPEAT`  | Initiates a new iteration of an executing loop.                              |
-| `SPAN`    | Initiates processing of a new [Span block](../programs.md#span-block).       |
-| `RESPAN`  | Initiates processing of a new operation batch within a span block.           |
+| `SPAN`    | Initiates processing of a new [Basic block](../programs.md#basic-block). (historically called "span block") |
+| `RESPAN`  | Initiates processing of a new operation batch within a basic block. (historically called "span block") |
 | `DYN`     | Initiates processing of a new [Dyn block](../programs.md#dyn-block).         |
 | `CALL`    | Initiates processing of a new [Call block](../programs.md#call-block).       |
 | `SYSCALL` | Initiates processing ofa new  [Syscall block](../programs.md#syscall-block). |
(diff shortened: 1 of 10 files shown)


adr1anh:
I'll recuse myself here since I’m not super familiar with this part of the VM/codebase.

My only question is whether we should just use Span everywhere instead of BasicBlock. IIRC the docs only use the former, so is there any reason why we don't use that terminology in the code?

bobbinth (approved):
Thank you! Looks good!

There are some more references in the docs (e.g., [here](https://0xmiden.github.io/miden-vm/design/programs.html#span-block), [here](https://0xmiden.github.io/miden-vm/design/decoder/main.html), and [here](https://0xmiden.github.io/miden-vm/design/decoder/constraints.html)) - but maybe that's for a different issue.

bobbinth:
> My only question is whether we should just use Span everywhere instead of BasicBlock. IIRC the docs only use the former, so is there any reason why we don't use that terminology in the code?

I've moved from "span" to "basic block" terminology a while back, but haven't updated it consistently everywhere yet. The main reasons are:

- "Basic block" is a standard term which is used to refer to sequences of instructions without control flow (i.e., in compilers) - I just didn't know about this when first building the VM.
- We do have `Span` struct as well, but that refers to code location identifiers (for debugging) - so, having two "span" structs would be confusing.

huitseeker:
@bobbinth I've added a commit for the doc updates.

docs/src/design/decoder/main.md
```diff
@@ -25,8 +25,8 @@ Managing control flow in the VM is accomplished by executing control flow operat
 | `SPLIT`   | Initiates processing of a new [Split block](../programs.md#split-block).     |
 | `LOOP`    | Initiates processing of a new [Loop block](../programs.md#loop-block).       |
 | `REPEAT`  | Initiates a new iteration of an executing loop.                              |
-| `SPAN`    | Initiates processing of a new [Span block](../programs.md#span-block).       |
-| `RESPAN`  | Initiates processing of a new operation batch within a span block.           |
+| `SPAN`    | Initiates processing of a new [Basic block](../programs.md#basic-block). (historically called "span block") |
+| `RESPAN`  | Initiates processing of a new operation batch within a basic block. (historically called "span block") |
```
bobbinth, line 29: I wonder if at some point we'd want to rename `SPAN` and `RESPAN` into something else - but I don't have good suggestions for now.

bobbinth (approved):
Looks good! Thank you!

PR #2188, opened 2025-09-18
Plafer full parallel trace gen

plafer:
Stacked on #1839

Implements the rest of parallel trace generation.


processor/src/fast/execution_tracer.rs
```diff
@@ -303,6 +412,11 @@ impl ExecutionTracer {
             self.fragment_contexts.push(trace_state);
         }
     }
+
+    /// Extracts the memory reads replay from the current memory reads replay and resets it.
+    fn extract_memory_reads_replay(&mut self) -> MemoryReadsReplay {
```
adr1anh: Any reason for including this when we don't have it for the other replay structs?

processor/src/fast/trace_state.rs
```diff
@@ -33,11 +38,12 @@ use crate::{
 ///    forest to execute when the current node is done executing,
 /// 4. initial state: some information about the state of the execution at the start of the
 ///    fragment. This includes the [MastForest] that is being executed at the start of the fragment
-///    (which can change when encountering an [miden_core::mast::ExternalNode]), and the current
-///    node's execution state, which contains additional information to pinpoint exactly where in
-///    the processing of the node we're at when this fragment begins.
+///    (which can change when encountering an [miden_core::mast::ExternalNode] or
+///    [miden_core::mast::DynNode]), and the current node's execution state, which contains
+///    additional information to pinpoint exactly where in the processing of the node we're at when
+///    this fragment begins.
```
adr1anh: These make the types clickable in the docs
```suggestion
///    (which can change when encountering an [`ExternalNode`](miden_core::mast::ExternalNode) or
///    [`DynNode`](miden_core::mast::DynNode)), and the current node's execution state, which contains
///    additional information to pinpoint exactly where in the processing of the node we're at when
///    this fragment begins.
```

processor/src/parallel/mod.rs
```diff
@@ -65,13 +77,134 @@ mod tests;
...
+    // Compute the final main trace length, after accounting for random rows
+    let main_trace_len = {
+        // Get the trace length required to hold all execution trace steps
+        let max_len = range_table_len.max(core_trace_len).max(chiplets.trace_len());
+
+        // Pad the trace length to the next power of two and ensure that there is space for random
+        // rows
+        let trace_len = (max_len + NUM_RAND_ROWS).next_power_of_two();
+        core::cmp::max(trace_len, MIN_TRACE_LEN)
+    };
```
Al-Kindi-0: nit: I would encapsulate this logic

processor/src/parallel/mod.rs
```diff
@@ -1704,3 +1970,89 @@ impl OperationHelperRegisters for TraceGenerationHelpers {
         ]
     }
 }
+
+// HELPERS
+// ================================================================================================
+
+// TODO(plafer): If we want to keep this strategy, then move the `op_eval_circuit()` method
```
Al-Kindi-0, line 1978: Q: Could you elaborate on this?

Al-Kindi-0 (commented):
Did an initial pass and looks great!
Will do another one once the TODOs are addressed or converted

processor/src/fast/trace_state.rs
```diff
@@ -436,55 +442,105 @@ impl MastForestResolutionReplay {
...
-        let (stored_addr, word) = self.words_read.pop_front().expect("No words read from memory");
+        let (stored_addr, word, _ctx, _clk) =
+            self.words_read.pop_front().expect("No words read from memory");
         debug_assert_eq!(stored_addr, addr, "Address mismatch: expected {addr}, got {stored_addr}");
         word
     }
+
+    /// Returns an iterator over all recorded memory element reads, yielding tuples of
+    /// (address, element, context ID, clock cycle).
+    pub fn iter_read_elements(&self) -> impl Iterator<Item = (Felt, Felt, ContextId, RowIndex)> {
```
adr1anh, line 501: Are these necessary when the struct fields are public?

processor/src/parallel/mod.rs
adr1anh: We discussed offline that the fast processor could just record all the memory reads, and let the trace generator handle the evaluation

adr1anh (commented):
Looks great! Just left a couple of nits.

processor/src/parallel/mod.rs
plafer: Done

processor/src/parallel/mod.rs
plafer: This `TODO` is related to the second bullet point in the issue description: 

> We now record the memory reads in ExecutionTracer when executing the EvalCircuit operation. This is now required so that we can populate the Memory chiplet properly, as well as the RangeChecker (since all memory reads come with a range check).
>
>    We currently also store the CircuitEvaluation struct in AceReplay (which could be computed during parallel trace generation instead). In principle this allows us to build the chiplets concurrently with the core trace, but at the cost of a larger message to send between the processor and trace generator. We can change this in a subsequent PR if needed.

> We discussed offline that the fast processor could just record all the memory reads, and let the trace generator handle the evaluation

This is currently what the fast processor does. To elaborate further, the current way we handle the `EvalCircuit` operation is two-fold:

1. The `FastProcessor` records all memory reads that occur during the operation
  - This is necessary in order for the Memory chiplet to have all the reads that occurred during program execution.
2. We store separately the `CircuitEvaluation` struct that is fed to the `Ace` chiplet

(1) is uncontroversial; we need to store that information one way or another. But (2) is not strictly necessary. As @adr1anh was alluding to, we could have the core trace generators build the `CircuitEvaluation` struct to be passed to the `Ace` chiplet. The downside of that approach though is that in order to build the `Ace` chiplet, we need to wait for the core trace generation to be done (so that they can return all the `CircuitEvaluation` structs that were built when re-executing the program in parallel).

So the upside of the current approach is that the ace chiplet can be built in parallel with the core trace fragments, at the cost of a larger `TraceGenerationContext` being sent from the fast processor. But this needs to be experimented with/benchmarked; thinking about it more, we expect the core trace generation to fully saturate the available cores, and so there shouldn't be a big cost (if any) to waiting for that to be done in order to start Ace chiplet trace generation.

**Conclusion**: For this PR, I think we should keep the current strategy just because it's already implemented, and we intend to improve on the current PR anyways. I will look into cleaning up that `TODO` with that in mind (by not having to copy/paste the `eval_circuit_fast_()` implementation between the fast processor and core trace generators)

processor/src/fast/execution_tracer.rs
plafer: Yeah not sure; removed and inlined the implementation instead.

processor/src/fast/trace_state.rs
plafer: Done

processor/src/fast/trace_state.rs
plafer: Didn't mean to make the fields public (similar to other replays); made private. And similarly for `MemoryWritesReplay`.

processor/src/parallel/mod.rs
plafer: Actually there's a few ways to do fix this TODO, which requires further investigation. I'll leave it for a subsequent PR (opened #2255).

plafer:
This is ready for final review. I

- updated the PR description with the benchmark results
- Opened a bunch of follow-up issues
- fixed the CI problem in #1839 (see https://github.com/0xMiden/miden-vm/pull/1839#issuecomment-3366019089)
- addressed all remaining `TODO(plafer)`, except for one (see https://github.com/0xMiden/miden-vm/pull/2188#discussion_r2402392032)

processor/src/parallel/mod.rs
Al-Kindi-0: Makes sense, thanks for the elaboration

processor/src/processor/operations/u32_ops.rs
```diff
@@ -32,7 +32,7 @@ macro_rules! require_u32_operands {
                 return Err(ExecutionError::not_u32_values(invalid_values, $errno, $err_ctx));
             }
             // Return tuple of operands based on indices
-            ($([<operand_ $idx>].as_int()),*)
+            ($([<operand_ $idx>]),*)
```
adr1anh, line 35: Ideally this would return the `u64` to avoid converting them again. But let's do this in a different PR.

plafer: Hmm, what do you mean by "again"? Here, the macro no longer calls `as_int()` on each operand, so effectively leaves it up to the caller to call `as_int()`. 

I ended up needing this because in some cases I need the `Felt` value, and so I was forced to do a spurious `Felt::new()` on the `u64`, e.g. in `op_u32add3`. And in all the functions, I was careful to only call `as_int()` once, preserving the previous behavior.

PR #2189, opened 2025-09-19
feat: propagate procedure annotations/attrs to library/package metadata

bitwalker:
This surfaces annotations attached to exported procedure definitions in the `Library` and `Package` metadata. This annotation information comes in the form of the AST corresponding to the annotation/attribute in the original MASM source (or that was generated from a high-level source language equivalent).

This PR implements the serialization for all the `AttributeSet`/`Attribute`-adjacent types so that they can be serialized with a `Package` (or `Library`, but serialization of `Library` is going away eventually anyway, so that's not really noteworthy here).

There are a couple of small fixes made in separate commits as well that I encountered during testing.

**NOTE:** This should be merged _after_ the type signatures PR, as it is based on that branch.

### DIFF
--- a/assembly-syntax/src/parser/token.rs
+++ b/assembly-syntax/src/parser/token.rs
@@ -2,7 +2,7 @@ use alloc::string::String;
 use core::fmt;
 
 use miden_core::{
-    Felt, FieldElement,
+    Felt, FieldElement, StarkField,
     utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
 };
 #[cfg(feature = "serde")]
@@ -355,7 +355,13 @@ impl Serializable for IntValue {
 impl Deserializable for IntValue {
     fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
         let raw = source.read_u64()?;
-        Ok(super::lexer::shrink_u64_hex(raw))
+        if raw >= Felt::MODULUS {
+            Err(DeserializationError::InvalidValue(
+                "int value is greater than field modulus".into(),
+            ))
+        } else {
+            Ok(super::lexer::shrink_u64_hex(raw))
+        }
     }
 }
 


bobbinth:
Not a review yet, but it may be good to update documentation (e.g., somewhere [here](https://0xmiden.github.io/miden-vm/user_docs/assembly/code_organization.html#procedures)) to explain how annotations work.

assembly-syntax/src/parser/grammar.lalrpop
```diff
@@ -1868,12 +1869,6 @@ WordValue: WordValue = {
     hex_word,
 }
 
-PushImm: Immediate<PushValue> = {
-    <l:@L> <value:WordValue> <r:@R> => Immediate::Value(Span::new(span!(source_file.id(), l, r), value.into())),
-    <l:@L> <value:IntValue> <r:@R> => Immediate::Value(Span::new(span!(source_file.id(), l, r), value.into())),
-    <ConstantName> => Immediate::Constant(<>),
-}
```
bobbinth, line 1875: Is this no longer needed? Does this change the functionality in some way?

assembly/src/linker/mod.rs
```diff
@@ -808,6 +809,41 @@ impl Linker {
...
+                    module: gid.module,
+                    kind: InvokeKind::ProcRef,
+                };
+                match self.resolve_target(&caller, &target)? {
+                    ResolvedTarget::Phantom(_) => Ok(AttributeSet::default()),
+                    ResolvedTarget::Exact { gid } | ResolvedTarget::Resolved { gid, .. } => {
+                        self.resolve_attributes(gid)
+                    },
+                }
+            },
```
bobbinth, line 843: Does this mean we can attach attributes to re-export? Or are attributes for re-exports inherited from the underlying procedures?

bobbinth:
@bitwalker - I'm going to merge as is to include the annotations in the v0.18.0 release - but let's do a follow-up PR with docs relatively shortly after this.

assembly-syntax/src/parser/grammar.lalrpop
bitwalker: It was dead code, no change to existing functionality

assembly/src/linker/mod.rs
bitwalker: They're inherited here, we do not currently allow attaching attributes to the alias/re-export itself.

I think it's debatable whether we should allow attaching additional attributes to procedure aliases at all, but if we have a use case where it would be important to do so, we'll need to discuss the semantics of how that works. For example, do attributes at the point an alias is declared override attributes of the underlying definition? What if you re-export a re-export - do the attributes accumulate at each re-export?

If we do allow them, then we probably should also dictate that attributes cannot affect semantics of the program at all, because it could be a source of bugs if a re-export overrides an attribute of the original definition that changes the semantics in a way that breaks it.

bobbinth: This was more just a question to make sure I understand how it work - but I agree that it probably doesn't make sense to allow attaching attributes to aliases. We should just make this clear when we write up the docs for procedure attributes.

PR #2191, opened 2025-09-19
fix: avoid panics by propagating errors in outputs parsing and proof writing

VolodymyrBg:
Replace unwraps that could crash the CLI on invalid user data or I/O errors.
OutputFile::stack_outputs: map parse::<u64> errors to Result for malformed .outputs.
ProofFile::write: return error on write_all failure instead of panicking.
Aligns with existing Result-based error handling


plafer (approved):
LGTM

PR #2194, opened 2025-09-20
sorted_array::push_lowerbound_result refactor

gswirski:
No changelog required. Addresses a few comments after #2114 was merged:
- https://github.com/0xMiden/miden-vm/pull/2114#discussion_r2365583544
- https://github.com/0xMiden/miden-vm/pull/2114#discussion_r2365584280
- https://github.com/0xMiden/miden-vm/pull/2114#discussion_r2365585634

> A potential small improvement: maybe this could be a method on the ProcessState? It could look something like:
> ```
> impl ProcessState {
>     pub fn get_mem_addr_range(
>         &self,
>         start_idx: usize,
>         end_idx: usize,
>     ) -> Result<Range<MemoryAddress>, MemoryError> {
>         ...
>     }
> }
> ```
> In the above, I've also used Range<MemoryAddress> instead of (u32, u32) for the return value as I think it is a bit more self-describing (and type-safe).

I used `Range<u32>` because `Range<MemoryAddress>` was notoriously difficult to work with: you cannot `for addr in range_object` without implementing `Step` trait on the `MemoryAddress` type (which is unstable).

### DIFF
--- a/processor/src/lib.rs
+++ b/processor/src/lib.rs
@@ -899,7 +899,7 @@ impl<'a> ProcessState<'a> {
         &self,
         start_idx: usize,
         end_idx: usize,
-    ) -> Result<(u32, u32), MemoryError> {
+    ) -> Result<core::ops::Range<u32>, MemoryError> {
         let start_addr = self.get_stack_item(start_idx).as_int();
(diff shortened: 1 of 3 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #2196, opened 2025-09-21
feat: allow entering odd-length representations of hex numbers

### LINKED ISSUE #1302: allow odd number of characters in hex constants
the following code:

```
begin
    push.0x100
end
```

fails to assemble with the following error code:

```
Failed to parse program file `t.masm` - parsing error at [2:5]: malformed instruction 'push.0x100', parameter 0x100 is invalid: hex string '100' does not contain an even number of characters
```

However, `0x100` is a valid representation of the number `128`. As an example, the following program:

```rust
fn main() {
    println!("{}", format!("0x{:02x}", 256));
}
```

Will print `0x100` when executed. This is causing test failures in `miden-base` when similar logic is used to format an account id for testing.

huitseeker:
Closes #1302.


assembly-syntax/src/parser/lexer.rs
```diff
@@ -615,10 +615,18 @@ fn parse_hex<'input>(
     hex_digits: &'input str,
 ) -> Result<Token<'input>, ParsingError> {
     use miden_core::{FieldElement, StarkField};
+
+    // Handle odd-length hex strings by padding with a leading zero
+    let hex_digits = if !hex_digits.len().is_multiple_of(2) {
```
bitwalker: *Nit*: We should use `alloc::borrow::Cow<'input, str>` as the type being bound here, to avoid allocating needlessly when the input string doesn't need to be padded.

bitwalker (approved):
Looks good! One small change needed to avoid this negatively impacting parser performance in the even-length case (the most common), but otherwise this is ready to merge. I'm marking this approved contingent on that to avoid the need for another round of review on my end.

PR #2197, opened 2025-09-21
Tracking PR for v0.19.0 release

bobbinth:
This is a tracking PR for v0.19.0 release


bobbinth:
Not sure why `no-std` build on `nightly` is failing. Seems like CI is trying to build it with Rust 1.89 - but I can't find where this is set (and why this is happening specifically for `nightly` which should be Rust 1.92).

PR #2201, opened 2025-09-22
Fix misleading expect message in range aux trace builder

viktorking7:
### Description
- Replaces an invalid "{}" placeholder in an `expect` message with a clear, static string.
- Improves error diagnostics; `expect` does not perform formatting.


viktorking7:
@huitseeker everything is ok? ready to merge?

PR #2202, opened 2025-09-22
fix(cli): own output path to avoid borrowing temporary PathBuf

radik878:
- Fix a lifetime bug in miden-vm/src/cli/bundle.rs where the default output_file path was created as a reference to a temporary PathBuf (&parent.join(...).with_extension(...)). This temporary is dropped at the end of the match arm, leading to a “temporary value dropped while borrowed” compile error.
- Change output_file to be an owned PathBuf and update write_to_file calls to pass &output_file. This aligns with write_to_file’s AsRef<Path> API and matches similar patterns elsewhere in the repo.


radik878:
> There's no compilation error, and the PR summary seems to make a false assumption about how long the lifetime of the `output_file` reference needs to be valid for.

Yes, i'm wrong, sorry for taking time

PR #2203, opened 2025-09-23
fix: make line_column_to_offset character-based

VolodymyrBg:
This change updates SourceContent::line_column_to_offset to interpret columns as Unicode character positions (Rust char) rather than raw bytes. The previous implementation treated columns as byte offsets and could panic when a column landed in the middle of a UTF-8 code point. It also contradicted location(), which already reports columns based on character counts.
With this change:

- line_column_to_offset now computes a safe byte offset via char_indices(), aligning semantics with location() and editor/LSP expectations.
- select() and update() become correct for non-ASCII input since they build on line_column_to_offset.
- A new unit test verifies round-trip consistency on a line containing multi-byte characters (aβ中💖).

These fixes prevent panics on valid UTF-8, ensure consistent column semantics across APIs, and better support LSP-style operations. If UTF-16 positions are needed by a client, conversion should occur at the integration boundary rather than within these core types.

### DIFF
--- a/crates/debug/types/src/source_file.rs
+++ b/crates/debug/types/src/source_file.rs
@@ -632,9 +632,14 @@ impl SourceContent {
             .expect("invalid line boundaries: invalid utf-8");
 
         let column_index = column_index.to_usize();
-        // Determine byte offset within the line corresponding to the character column
-        let byte_in_line =
-            line_src.char_indices().nth(column_index).map(|(byte_index, _)| byte_index)?;
+        // Determine byte offset within the line corresponding to the character column.
+        // Include end-of-line position so that col == num_chars maps to line_src.len(),
+        // and col == 0 on an empty line maps to 0.
+        let byte_in_line = line_src
+            .char_indices()
+            .map(|(i, _)| i)
+            .chain(core::iter::once(line_src.len()))
+            .nth(column_index)?;
 
         Some(line_span.start + ByteOffset::from_str_len(&line_src[..byte_in_line]))
     }
@@ -1333,5 +1338,21 @@ end
             assert_eq!(flc.line, line_index.number());
             assert_eq!(flc.column.to_u32(), col as u32 + 1); // FileLineCol is one-indexed
         }
+    }
+
+    #[test]
+    fn selection_end_at_next_empty_line_start() {
+        // Single line with trailing newline creates an empty next line
+        let content_str = "abc\n";
+        let content = SourceContent::new("masm", "newline.masm", content_str);
+
+        // Selection from line 0 to line 1 should resolve end to the byte after '\n'
+        let start = content.line_column_to_offset(LineIndex(0), ColumnIndex(0)).expect("start");
+        let end = content.line_column_to_offset(LineIndex(1), ColumnIndex(0)).expect("end");
+
+        // start at 0
+        assert_eq!(start.to_usize(), 0);
+        // end at position after 'abc\n' (i.e., 4)
+        assert_eq!(end.to_usize(), content_str.len());
     }
 }


crates/debug/types/src/source_file.rs
```diff
@@ -625,18 +625,27 @@ impl SourceContent {
...
     ) -> Option<ByteIndex> {
-        let column_index = column_index.to_usize();
         let line_span = self.line_range(line_index)?;
         let line_src = self
             .content
             .get(line_span.start.to_usize()..line_span.end.to_usize())
             .expect("invalid line boundaries: invalid utf-8");
-        if line_src.len() < column_index {
+
+        let col_chars = column_index.to_usize();
```
bitwalker: *Nit:* This name is less clear than `column_index` (the previous name of this binding) - the `ColumnIndex` type is a 0-indexed value representing the index of the character on the line, but `col_chars` could be interpreted a number of different ways. I would keep the old name.

crates/debug/types/src/source_file.rs
```diff
@@ -625,18 +625,27 @@ impl SourceContent {
...
+        let num_chars = line_src.chars().count();
+
+        if col_chars > num_chars {
```
bitwalker: It isn't necessary to check this here, as  `.char_indices().nth()` will produce `None` if the index is out of bounds of the line.

crates/debug/types/src/source_file.rs
```diff
@@ -625,18 +625,27 @@ impl SourceContent {
...
             return None;
         }
-        let (pre, _) = line_src.split_at(column_index);
-        let start = line_span.start;
-        Some(start + ByteOffset::from_str_len(pre))
+
+        // Determine byte offset within the line corresponding to the character column
+        let byte_in_line = if col_chars == num_chars {
```
bitwalker: This is also redundant - `char_indices().nth()` already handles this case, and removing this conditional avoids the need for us to compute `num_chars` at all (and thus iterate the characters of the line twice).

bitwalker (changes requested):

crates/debug/types/src/source_file.rs
VolodymyrBg: > This is also redundant - `char_indices().nth()` already handles this case, and removing this conditional avoids the need for us to compute `num_chars` at all (and thus iterate the characters of the line twice).

all done

crates/debug/types/src/source_file.rs
```diff
@@ -625,18 +625,30 @@ impl SourceContent {
...
+            count += 1;
+        }
+        if count != column_index {
+            // Out of bounds: requested column is greater than number of characters in line
             return None;
         }
-        let (pre, _) = line_src.split_at(column_index);
-        let start = line_span.start;
-        Some(start + ByteOffset::from_str_len(pre))
+
```
bitwalker: All of this can be replaced with just:

```rust
let byte_in_line = line_src.char_indices().nth(column_index).map(|(byte_index, _)| byte_index)?;
```

bitwalker:
It looks like this exposes an issue with `SourceContent::update`, namely that having a `Selection` whose `end` specifies a line that doesn't exist/is empty, will fail, presumably because the attempt to resolve a `ColumnIndex` of 0 on that line will fail (`char_indices()` on that line will be an empty iterator, so `None` is returned, which causes `SourceContent::update` to fail).

We want the semantics of `Selection::from(LineIndex(0)..LineIndex(1))` on a source file that has a single line of text with a trailing newline, to produce a `Selection` whose `end` specifies the byte offset of the trailing newline. I think we probably need to handle that case in `line_column_to_offset`, by checking that when `char_indices().nth(column_index)` produces `None`, if `column_index == 0`, then we should return the byte index of the start of the line.

VolodymyrBg:
> It looks like this exposes an issue with `SourceContent::update`, namely that having a `Selection` whose `end` specifies a line that doesn't exist/is empty, will fail, presumably because the attempt to resolve a `ColumnIndex` of 0 on that line will fail (`char_indices()` on that line will be an empty iterator, so `None` is returned, which causes `SourceContent::update` to fail).
> 
> We want the semantics of `Selection::from(LineIndex(0)..LineIndex(1))` on a source file that has a single line of text with a trailing newline, to produce a `Selection` whose `end` specifies the byte offset of the trailing newline. I think we probably need to handle that case in `line_column_to_offset`, by checking that when `char_indices().nth(column_index)` produces `None`, if `column_index == 0`, then we should return the byte index of the start of the line.

seems like failling tests are unrelated

bitwalker:
Can you rebase on `next`? That should clear up the CI failures

VolodymyrBg:
> Can you rebase on `next`? That should clear up the CI failures

Yes, CI feels good now

bobbinth:
Closing this as stale. @bitwalker - is this something that will be addressed in the current assembler refactoring?

bitwalker:
No, but if the CI wasn't passing, then closing this is fine. If it was, then we should reopen and merge.

VolodymyrBg:
> No, but if the CI wasn't passing, then closing this is fine. If it was, then we should reopen and merge.

CI was green, but there was a conflict, names of the files changed, idk if there is no sense of this pr, then closed

PR #2205, opened 2025-09-23
fix(processor): fix debug.adv_stack print order

Qyriad:
## Describe your changes

This makes `debug.adv_stack` print in the same order elements are pushed to the operand stack in. e.g., makes `debug.adv_stack.1` print the same value that `adv_push.1` pushes to the operand stack. I believe this was the case in an earlier version, but I haven't tried to bisect it.

### Given:

```json
{
  "operand_stack": [ ],
  "advice_stack": [
      "1",
      "2",
      "3",
      "4"
  ]
}
```

```ruby
begin
	debug.adv_stack
	adv_push.1 debug.stack.1
	dropw
end
```

### Before:

```
Advice Stack state before step 1:
├──  0: 4
├──  1: 3
├──  2: 2
└──  3: 1

Stack state before step 2:
└──  0: 1

└── (0 more items)
```

### After:

```
Advice Stack state before step 1:
├──  0: 1
├──  1: 2
├──  2: 3
└──  3: 4

Stack state before step 2:
└──  0: 1

└── (0 more items)
```

- Updated `CHANGELOG.md'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - [BREAKING] Incremented MSRV to 1.90.
 
+#### Fixes
+
+- Fixed `debug.adv_stack` printing in reverse order ([#2205](https://github.com/0xMiden/miden-vm/pull/2205)).
+
 ## 0.18.0 (2025-09-21)
 
 #### Enhancements
--- a/processor/src/host/debug.rs
+++ b/processor/src/host/debug.rs
@@ -80,10 +80,11 @@ fn print_vm_adv_stack(process: &ProcessState, n: u16) {
 
     let stack = &stack[..num_items];
 
-    if let Some((last, front)) = stack.split_last() {
+    // Note: `stack` is in reverse order. e.g., `adv_push.1` pushes `stack.last()`.
+    if let Some((last, front)) = stack.split_first() {
         // print all items except for the last one
         println!("Advice Stack state before step {}:", process.clk());
-        for (i, element) in front.iter().enumerate() {
+        for (i, element) in front.iter().rev().enumerate() {
             println!("├── {i:>2}: {element}");
         }
 


plafer:
Interesting... Would be good to understand the core issue of why this changed though (*if* this changed) - in case there's a bigger issue underlying this.

Qyriad:
Hm… I may be wrong about the order previously being different, given that this test relies on the existing ordering… https://github.com/0xMiden/miden-vm/blob/bed310d9626a9e27d24d4c4faf56a53f590d597a/miden-vm/tests/integration/cli/cli_test.rs#L156-L174

adr1anh:
Looking at the flow, when we initialize a program advice stack `[1, 2, 3, 4]`, the internal stack in the `AdviceProvider` is reversed and we get `[4,3,2,1]`. When we call `adv_push`, we `pop` the element at the end of the provider's stack vector, which in this case is `1`. Calling `debug.adv_stack` iterates over the stack "in-order", so it would print `[4,3,2,1]` and after the pop, we'd get `[4,3,2]`. 

Given that the current semantics of `debug.stack` prints the elements from top to bottom, the current implementation of `debug.adv_stack` is inconsistent. It seems much more intuitive to keep the representation of the initial input. 

The test you linked to has advice stack inputs `[21, 42]`, but prints it out in reverse order. It seems natural for the debug command to print it out in the input order as well.

Overall, I think the semantics introduced in this PR are correct.

adr1anh:
As an aside -- and we should probably convert this to an issue -- I don't understand why we are using a CLI test for checking the correctness of this result. The fact that the test program and inputs and outputs are in 3 different locations makes it harder to check that we're testing the right thing.

I haven't used it yet, but it seems like we could use `insta` to check the diagnostics message matches the expected ordering.

bobbinth (approved):
Looks good! Thank you!

I do agree that the way we specify advice stack inputs is counter-intuitive and we may want to revisit this and other input ordering issues comprehensively. Let's create a separate issue for this.

adr1anh:
I've been digging into the all the debug options, and have found several other mistakes. I have a PR which is almost ready and which should address all issues and includes exhaustive tests that show the expected behavior. It also brought to light an edge case which is not addressed here (`debug.adv_stack.n` is not handled properly). I suggest we close this one if that's alright with everyone, so we can fix all issues at once.

adr1anh:
The PR is ready for review #2218.

PR #2206, opened 2025-09-23
fix: proof file existence check in verify CLI

Galoretka:
Replace incorrect Path::try_exists().is_err() with a straightforward exists() check in VerifyCmd::infer_defaults(). The previous logic only flagged I/O errors and failed to detect a missing file (Ok(false)), allowing nonexistent proof paths to pass the pre-check. This change ensures we fail early with a clear “Proof file does not exist” message, aligns with the project’s existing style (InputFile::read uses .exists()), and avoids misleading behavior before ProofFile::read.

### DIFF
--- a/miden-vm/src/cli/verify.rs
+++ b/miden-vm/src/cli/verify.rs
@@ -81,3 +81,41 @@ impl VerifyCmd {
         Ok((input_file, output_file))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::{fs, fs::File};
+
+    use super::*;
+
+    #[test]
+    fn infer_defaults_uses_proof_file_basename_for_defaults() {
+        // prepare a unique temp directory
+        let base =
+            std::env::temp_dir().join(format!("miden_vm_verify_test_{}", std::process::id()));
+        fs::create_dir_all(&base).expect("create temp test dir");
+
+        // create a dummy proof file
+        let proof_path = base.join("proof_file");
+        File::create(&proof_path).expect("create proof file");
+
+        // build command with no explicit input/output
+        let cmd = VerifyCmd {
+            input_file: None,
+            output_file: None,
+            proof_file: proof_path.clone(),
+            program_hash: "00".to_string(),
+        };
+
+        // exercise
+        let (input, output) = cmd.infer_defaults().expect("infer defaults");
+
+        // verify: defaults are proof file with replaced extensions
+        assert_eq!(input, proof_path.with_extension("inputs"));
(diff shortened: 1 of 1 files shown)


huitseeker:
A rebase would be lovely.

miden-vm/src/cli/verify.rs
```diff
@@ -69,11 +66,10 @@ impl VerifyCmd {
     }
 
     fn infer_defaults(&self) -> Result<(PathBuf, PathBuf), Report> {
```
huitseeker, line 68: While we're here, couldn't we refactor the repetition, and avoid extra clones?
```rust
fn infer_defaults(&self) -> Result<(PathBuf, PathBuf), Report> {
    if !self.proof_file.exists() {
        return Err(Report::msg("Proof file does not exist"));
    }

    let infer_path = |ext: &str| {
        let mut path = self.proof_file.clone();
        path.set_extension(ext);
        path
    };

    let input_file = self.input_file.clone().unwrap_or_else(|| infer_path("inputs"));
    let output_file = self.output_file.clone().unwrap_or_else(|| infer_path("outputs"));

    Ok((input_file, output_file))
}
```

Galoretka: Yes we can

huitseeker (changes requested):
While the latest revision is certainly better when it comes to path cloning, it changes the behaviour of the CLI from setting extensions to using the "input" and "output" strings as a default.

Galoretka:
> While the latest revision is certainly better when it comes to path cloning, it changes the behaviour of the CLI from setting extensions to using the "input" and "output" strings as a default.

so I should revert last changes?

huitseeker:
@Galoretka you should have a unit test that checks that when the input and output files don't exist, but that the proof file does and is at `/path/to/proof_file`, the resulting input is `/path/to/proof_file.inputs`, and the output is `/path/to/proof_file.outputs`, not `/path/to/inputs` and `/path/to/outputs`.

Galoretka:
> @Galoretka you should have a unit test that checks that when the input and output files don't exist, but that the proof file does and is at `/path/to/proof_file`, the resulting input is `/path/to/proof_file.inputs`, and the output is `/path/to/proof_file.outputs`, not `/path/to/inputs` and `/path/to/outputs`.

Yes, it's done

plafer:
Hey @Galoretka I'm blocked from merging this because it contains commits that are not signed. Can you sign those and re- (force) push?

bobbinth (approved):
Looks good! Thank you!

PR #2207, opened 2025-09-24
fix: mismatched Push expectations in decoder syscall_block test

GarmashAlex:
## Describe your changes

The syscall_block test expected Push(TWO) in the program’s first SPAN and Push(ONE) in bar’s SPAN, while the constructed MAST nodes use Push(ONE) for the initial +1 FMP update and Push(TWO) for bar’s +2 FMP update. This mismatch caused the test to assert against values that cannot occur given the program structure and FmpUpdate semantics. The change aligns the expected op decoding with the actual program and documented FMP behavior (CALL resets to FMP_MIN, SYSCALL resets to SYSCALL_FMP_MIN, and FmpUpdate adds the pushed value).

### DIFF
--- a/processor/src/decoder/tests.rs
+++ b/processor/src/decoder/tests.rs
@@ -122,8 +122,8 @@ fn basic_block_small() {
 
     // --- check block address, op_bits, group count, op_index, and in_span columns ---------------
     check_op_decoding(&trace, 0, ZERO, Operation::Span, 4, 0, 0);
-    check_op_decoding(&trace, 1, INIT_ADDR, Operation::Push(ONE), 3, 0, 1);
-    check_op_decoding(&trace, 2, INIT_ADDR, Operation::Push(TWO), 2, 1, 1);
+    check_op_decoding_with_imm(&trace, 1, INIT_ADDR, ONE, 1, 3, 0, 1);
+    check_op_decoding_with_imm(&trace, 2, INIT_ADDR, TWO, 2, 2, 1, 1);
     check_op_decoding(&trace, 3, INIT_ADDR, Operation::Add, 1, 2, 1);
     check_op_decoding(&trace, 4, INIT_ADDR, Operation::Swap, 1, 3, 1);
     check_op_decoding(&trace, 5, INIT_ADDR, Operation::Drop, 1, 4, 1);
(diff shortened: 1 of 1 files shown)


huitseeker (commented):
The original test was passing because the test framework checks the operation, but not which values were pushed. Would you consider enhancing the test framework to validate the actual pushed values?

You could grab the immediate values in the `hasher_trace`columns, and create a `check_op_decoding_with_imm` which would check values, as well as the operation, and update tests to use it.

GarmashAlex:
> The original test was passing because the test framework checks the operation, but not which values were pushed. Would you consider enhancing the test framework to validate the actual pushed values?
> 
> You could grab the immediate values in the `hasher_trace`columns, and create a `check_op_decoding_with_imm` which would check values, as well as the operation, and update tests to use it.

corrected

huitseeker (approved):
Very nice, thanks a bunch!

processor/src/decoder/tests.rs
bobbinth: I think there is a typo here: this should probably be `call.bar` - right?

processor/src/decoder/tests.rs
```diff
@@ -1634,6 +1634,39 @@ fn check_op_decoding(
...
+    }
+    let span_row = span_row.expect("no preceding SPAN/RESPAN row found for PUSH");
+
+    let mut found = false;
+    for column in trace[HASHER_STATE_RANGE].iter() {
+        if column[span_row] == imm {
+            found = true;
+            break;
+        }
+    }
```
bobbinth: Instead of trying to find the column with the right value, we could also pass the expected index of the value as a parameter to this function. That is, the function could look something like this:
```rust
fn check_op_decoding_with_imm(
    trace: &DecoderTrace,
    row_idx: usize,
    addr: Felt,
    imm: Felt,
    imm_idx: usize,
    group_count: u64,
    op_idx: u64,
    in_span: u64,
)
```
And then instead of this loop, we'd change the current like 1654 above to something like:
```rust
assert_eq!(trace[HASHER_STATE_RANGE.start + imm_idx][r], imm);
```

bobbinth (approved):
Looks good! Thank you for finding this!

We can merge as is, but I also left a couple of comments with potential improvements.

processor/src/decoder/tests.rs
GarmashAlex: > Instead of trying to find the column with the right value, we could also pass the expected index of the value as a parameter to this function. That is, the function could look something like this:
> 
> ```rust
> fn check_op_decoding_with_imm(
>     trace: &DecoderTrace,
>     row_idx: usize,
>     addr: Felt,
>     imm: Felt,
>     imm_idx: usize,
>     group_count: u64,
>     op_idx: u64,
>     in_span: u64,
> )
> ```
> 
> And then instead of this loop, we'd change the current like 1654 above to something like:
> 
> ```rust
> assert_eq!(trace[HASHER_STATE_RANGE.start + imm_idx][r], imm);
> ```

Made these changes

PR #2218, opened 2025-09-25
fix: Make `DefaultDebugHandler` consistent

adr1anh:
@Qyriad discovered in #2205 that the current debugging commands did not return the advice in the correct order. Upon further investigation, it turns out there were many issues with the current implementation which were hard to test for since the information was logged via `println`. 

The issues include:
- `debug.local` would not print anything because the range end was incorrectly set by the compiler
- `debug.adv` was printing elements in reverse order
- `debug.stack` was handling a case where the stack was empty, which could never happen
- `debug.local` was checking weird bonds `[0, u16::MAX]`
- Even with @Qyriad's fix, we were incorrectly printing the truncated advice stack

We also make some improvements:
- If a single value is requested, it is printed as a single line
- When printing the stack, we only report the number of remaining elements when we request elements from the overflow stack (this one may be contentious, it might make sense to report the number of remaining elements when we have more than 16 elements)
- We report the number of remaining elements on the advice stack
- If we request more elements of the stack than its size, we print `EMPTY`

The main goal and benefit of this PR is to exhaustively test all variations of the debug options, in a way where the source code and expected output is contained in the same test.

To acheive this, we modified the `DefaultDebugHandler` to write the output to an arbitrary buffer. By default, we write to `stdout` using `println!`, though in the new `execute_with_debug_buffer` testing function, we now record all outputs in a buffer which is extracted from the process once the program has executed. 

Finally, we removed the CLI tests that were testing the some debugging behavior.

We also reduce the number of `#[cfg(feature = "std")]` by limiting it to the `StdoutWriter`.

- Updated `CHANGELOG.md'


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 #### Changes
 
 - [BREAKING] Incremented MSRV to 1.90.
+- [BREAKING] Fix inconsistencies in debugging instructions ([2205](https://github.com/0xMiden/miden-vm/pull/)).
```
huitseeker: URL and PR number need adjustment.

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,372 @@
...
+            # => [5, 4, 3, 2, 1, 0]
+
+            # Shrink stack to 12 elements and print the whole output
+            # We should get 16 elements since zeros are pushed until the size is 16
+            drop dropw debug.stack
+            # => [0, ..., 0]
+
+        end";
+
+    let expected_output = "\
```
huitseeker: Please use [insta](https://insta.rs/docs/quickstart/), it will make this so much more manageable!

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+
+        end";
+
+    // Execute with debug buffer
+    let test = build_debug_test!(source, &stack_inputs);
+    let (_stack, output) = test.execute_with_debug_buffer().expect("execution failed");
+
+    insta::assert_snapshot!(output, @r"
+    Stack state before step 1:
+    Stack state before step 1:
```
bobbinth: Question: why is this line duplicated here?

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+    ├── 11: 42
+    ├── 12: 4
+    ├── 13: 3
+    ├── 14: 2
+    ├── 15: 1
+    ├── 16: 0
+    ├── 17: 0
+    ├── 18: 0
+    └── 19: 0
+    └── (8 more items)
```
bobbinth, line 89: nit: should this be printed out as:
```
    ├── 18: 0
    ├── 19: 0
    └── (8 more items)
```

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+    ├── 0x40000005: 0
+    ├── 0x40000006: 5
+    └── 0x40000007: 0
+    ");
+}
+
+#[test]
+fn test_debug_adv_stack() {
+    let stack_input = [0, 1, 2, 3, 4]; // 4 is on top
+    let advice_stack = [8, 7, 6, 5, 4, 3, 2, 1]; // 8 is on top
```
bobbinth, line 284: Not related to this PR, but it is definitely confusing that we specify operand stack inputs in one way and advice stack inputs in another way. I think maybe the fast processor fixed this already?

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+    ├──  2: 6
+    ├──  3: 5
+    ├──  4: 4
+    ├──  5: 3
+    ├──  6: 2
+    └──  7: 1
+    Advice Stack state before step 1:
+    ├──  0: 8
+    └──  1: 7
+    └── (6 more items)
```
bobbinth, line 340: Same nit as in one of the above comments - would be great if this was:
```
    ├──  0: 8
    ├──  1: 7
    └── (6 more items)
```

processor/src/tests/debug.rs
```diff
@@ -0,0 +1,365 @@
...
+    └──  2: 1
+    Stack state before step 28:
+    ├──  0: 1
+    ├──  1: 2
+    ├──  2: 3
+    ├──  3: 3
+    ├──  4: 2
+    ├──  5: 1
+    └──  6: 0
+    Advice Stack empty before step 45.
```
bobbinth: Would be great to print something out here - otherwise, it may not be immediately clear that the advice stack is empty.

bobbinth (approved):
Looks great! Thank you for doing this! I left some comments inline - but these can also be addresses in a separate PR.

processor/src/tests/debug.rs
adr1anh: This was to test `debug.stack.0`, or does it make more sense to output the full stack in that case?

processor/src/tests/debug.rs
adr1anh: Something like this?
```
Stack state before step 28:
└──  EMPTY
```

PR #2219, opened 2025-09-25
Stdlib: fix FALCON_DIV advice stack docs and clarify falcon_sign casting

sashaodessa:
### Description
- Summary:
  - Align `falcon_div` advice stack docs with implementation: output is `[r_lo, q_lo, q_hi, ...]`.
  - Add concise English comments explaining invariants and I/O in `falcon_div`.
  - Clarify in `falcon_sign` that `u64 → u8` conversion is safe post range-check.

- Rationale:
  - Prevents confusion/misuse by documenting the exact stack layout.
  - Makes cryptographic code intent explicit, reducing maintenance risk.

### DIFF
--- a/stdlib/src/handlers/falcon_div.rs
+++ b/stdlib/src/handlers/falcon_div.rs
@@ -29,14 +29,14 @@ pub const FALCON_DIV_EVENT_ID: EventId = EventId::from_u64(9834516640389212175);
 ///   Advice stack: [...]
 ///
 /// Outputs:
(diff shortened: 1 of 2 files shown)


bobbinth:
I actually think the comments were correct before this PR. cc @adr1anh and @Al-Kindi-0 just in case.

adr1anh:
Yes, I don't seen an issue with the previous comments. Let's close this.

PR #2223, opened 2025-09-26
fix(racy-lock): tie Send/Sync to T and F; add compile-time assertions

radik878:
Add explicit unsafe impls:
- Send for RacyLock<T, F> where T: Send, F: Send + Fn() -> T
- Sync for RacyLock<T, F> where T: Sync, F: Sync + Fn() -> T

Add positive compile-time trait assertions for Send and Sync.

Rationale: Without explicit bounds, auto-trait derivation could mark RacyLock as Send/Sync even when T is not, because AtomicPtr<T> is Send + Sync regardless of T. Since force/deref expose &T across threads and F may be invoked concurrently, we must require T: Sync for shared access, T: Send for cross-thread ownership transfer, and F: Send + Sync to safely move/share the initializer. This aligns behavior with std::sync::LazyLock and once_cell and prevents unsound cross-thread access.

### DIFF
--- a/crates/utils-sync/src/racy_lock.rs
+++ b/crates/utils-sync/src/racy_lock.rs
@@ -196,10 +196,7 @@ where
     }
 }
 
-impl<T, F> Drop for RacyLock<T, F>
-where
-    F: Fn() -> T,
-{
+impl<T, F> Drop for RacyLock<T, F> {
     /// Drops the underlying pointer.
     fn drop(&mut self) {
         let ptr = *self.inner.get_mut();


crates/utils/sync/src/racy_lock.rs
```diff
@@ -26,6 +26,97 @@ where
...
+    use alloc::boxed::Box;
+    use core::{
+        cell::RefCell,
+        ptr,
+        sync::atomic::{AtomicPtr, Ordering},
+    };
+
+    use loom::{hint, model::Builder, sync::Arc, thread};
+
+    // Deliberately unsound lock that ignores `T` in Sync/Send bounds to demonstrate the failure.
```
huitseeker: Nice touch adding the `unsound_demo` module! This really helps illustrate why the bounds are needed. The loom test with `RefCell` clearly shows the runtime panic that can happen without proper bounds.

crates/utils/sync/src/racy_lock.rs
```diff
@@ -184,4 +284,16 @@ mod tests {
         fn assert_traits<T: Send + Sync>() {}
         assert_traits::<RacyLock<Vec<i32>>>();
     }
+
+    #[test]
+    fn is_send() {
```
huitseeker: Good addition of the explicit trait assertion tests! These will help catch regressions if someone accidentally changes the bounds in the future.

crates/utils/sync/src/racy_lock.rs
```diff
@@ -26,6 +26,97 @@ where
...
+
+    // UNSOUND: `Sync` and `Send` do not depend on `T`.
+    unsafe impl<T, F: Fn() -> T + Sync> Sync for BadLock<T, F> {}
+    unsafe impl<T, F: Fn() -> T + Send> Send for BadLock<T, F> {}
+
+    // This test demonstrates the failure mode: sharing `&RefCell<_>` across threads via
+    // an unsound `Sync` impl allows concurrent `borrow_mut`, which panics at runtime.
+    #[test]
+    #[should_panic]
+    fn bad_sync_allows_cross_thread_refcell_borrow_mut_panic() {
```
huitseeker: I've renamed the test to include 'loom' in its name, so it will now be picked up by the existing `make test-loom` target.

huitseeker:
I think the `T: Sync` bound is correct in this case. 
- `Lazy/Once` (`OnceLock`, `LazyLock`, `::once_cell::sync::OnceCell`): these expose shared borrows `&T` after initialization → they’re `Sync` iff `T: Sync`. That’s our case.
    - [sync bound for OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#impl-Sync-for-OnceLock%3CT%3E)
    - [sync bound for LazyLock](https://github.com/rust-lang/rust/blob/377a9310457f4e92d28fd0bef1fe9bfde47abe3b/library/std/src/sync/lazy_lock.rs#L413)
    - [sync bound for ::once_cell::sync::OnceRef](https://github.com/matklad/once_cell/blob/fda60a6c70eb2e25a0c73bf515ec4779df3d2e6c/src/race.rs#L238)
- `Mutex/RwLock`: these don’t hand out `&T` that can be active concurrently across threads; access is gated via a lock/guard. They’re `Sync` iff `T: Send` (because ownership can move across threads through the guard), not `T: Sync`.
    - [sync bound for Mutex](https://github.com/rust-lang/rust/blob/377a9310457f4e92d28fd0bef1fe9bfde47abe3b/library/std/src/sync/poison/mutex.rs#L257)
    - [sync bound for RwLock](https://github.com/rust-lang/rust/blob/377a9310457f4e92d28fd0bef1fe9bfde47abe3b/library/std/src/sys/sync/rwlock/solid.rs#L14)
    - [sync bound for MutexGuard](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#impl-Sync-for-MutexGuard%3C'_,+T%3E)
    - [sync bound for RwLockGuard](https://doc.rust-lang.org/std/sync/struct.RwLockReadGuard.html#impl-Sync-for-RwLockReadGuard%3C'_,+T%3E)

Since `RacyLock` [exposes `&T`](https://github.com/0xMiden/miden-vm/blob/04336eefa82241192c73295a08a5c45c1d619de5/crates/utils/sync/src/racy_lock.rs#L50), it’s in the first camp, so `T: Sync` is required. The loom test is worth reading: we make `RacyLock` `Sync` through [auto-derive](https://doc.rust-lang.org/nomicon/send-and-sync.html) without requiring `T: Sync`, stash a `RefCell` inside and then share `&RefCell<_>` across threads through `RacyLock`. That enables concurrent `borrow_mut`/interior mutation without synchronization.

crates/utils/sync/src/racy_lock.rs
```diff
@@ -127,6 +218,15 @@ where
...
 }
 
+// Ensure `RacyLock` only implements auto-traits when it is sound to do so.
+// `Send` requires ability to move the owned initializer and the (possibly
+// newly allocated) `T` across threads safely.
+unsafe impl<T: Send, F: Send + Fn() -> T> Send for RacyLock<T, F> {}
+
+// `Sync` requires that shared access through `&self` is safe, which implies
+// both the stored `T` and the initializer `F` can be shared across threads.
+unsafe impl<T: Sync, F: Sync + Fn() -> T> Sync for RacyLock<T, F> {}
```
plafer: Based on [huitseeker's comment](https://github.com/0xMiden/miden-vm/pull/2223#issuecomment-3419238164), shouldn't this be:

```suggestion
unsafe impl<T: Send + Sync, F: Send> Sync for RacyLock<T, F> {}
```

Taken from both [`LazyLock`](https://github.com/rust-lang/rust/blob/377a9310457f4e92d28fd0bef1fe9bfde47abe3b/library/std/src/sync/lazy_lock.rs#L413) and [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#impl-Sync-for-OnceLock%3CT%3E) in the stdlib. Haven't fully thought through the implications of `F: Send` instead of `F: Sync` yet.

crates/utils/sync/src/racy_lock.rs
```diff
@@ -127,6 +218,15 @@ where
     }
 }
 
+// Ensure `RacyLock` only implements auto-traits when it is sound to do so.
+// `Send` requires ability to move the owned initializer and the (possibly
+// newly allocated) `T` across threads safely.
+unsafe impl<T: Send, F: Send + Fn() -> T> Send for RacyLock<T, F> {}
```
plafer:
```suggestion
unsafe impl<T: Send, F: Send> Send for RacyLock<T, F> {}
```
Similarly here, we don't need the `Fn() -> T` bound for the purposes of `Send`

plafer:
In the meantime @radik878 could you fix the merge conflict? Sorry about that, should be trivial to fix (we just changed the directory structure).

radik878:
> In the meantime @radik878 could you fix the merge conflict? Sorry about that, should be trivial to fix (we just changed the directory structure).

didn't notice it immediately, done

crates/utils/sync/src/racy_lock.rs
radik878: > Based on [huitseeker's comment](https://github.com/0xMiden/miden-vm/pull/2223#issuecomment-3419238164), shouldn't this be:
> 
> Taken from both [`LazyLock`](https://github.com/rust-lang/rust/blob/377a9310457f4e92d28fd0bef1fe9bfde47abe3b/library/std/src/sync/lazy_lock.rs#L413) and [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#impl-Sync-for-OnceLock%3CT%3E) in the stdlib. Haven't fully thought through the implications of `F: Send` instead of `F: Sync` yet.

looks like tests failed because of this change

plafer: Ah, we should remove the trait bounds on the struct definition [here](https://github.com/0xMiden/miden-vm/blob/f3f53881b8fbc2c048f69537a722c048c8d4e2d3/crates/utils-sync/src/racy_lock.rs#L22-L23)

PR #2200, opened 2025-09-22
test: instances of Arbitrary for MastNodeForest & BasicBlockNode

huitseeker:
- This creates instances of `Arbitrary` for `BasicBlockNode`, reusing and extending the strategies introduced in #2094.
- Incidentally, this uses the above to add a few serde roundtrips, reusing the macros introduced in #2132.
- This creates a benchmark leveraging the above to generate a handful of random `MastForest`s, and benches their merge.

The goal of this PR is to then power refactors of the `mast::merger` module (one of the later writers into the `MastForest` nodes in their life cycle).

PR #2224, opened 2025-09-27
test: Arbitrary + {winter,serde} serialization, up to Program (+ small bugfix)

huitseeker:
**This PR:**
- is stacked on #2200,
- extends the round-trip macros started in #2132 to also generate winterfell round-trip tests,
- adds a few instances of Arbitrary for types that were missing one (up to and including `Program`), fixing an issue in the instance for `WordValue` 
- adds ~35 roundtrip serialization tests,
- fixes an enum-tagging bug in the serialization of `Attribute`

**Also (unrelated)**:
- adds a rust-toolchain because CI (based on rustup atm) isn't picking up the latest stable.

**Why?** I am working on a feature which is blocked by a (winter) serialization issue in the space *between* `MastForest` (ok) and `Program` (not so). This lets me rule out a range of issues.

### DIFF
--- a/core/src/mast/node/basic_block_node/arbitrary.rs
+++ b/core/src/mast/node/basic_block_node/arbitrary.rs
@@ -1,9 +1,12 @@
+use alloc::{collections::BTreeMap, sync::Arc};
 use core::ops::RangeInclusive;
 
 use proptest::{arbitrary::Arbitrary, prelude::*};
 
 use super::*;
-use crate::{Decorator, Operation, mast::DecoratorId};
+use crate::{
+    AdviceMap, AssemblyOp, DebugOptions, Decorator, Felt, Operation, Word, mast::DecoratorId,
+};
 
 // Strategy for operations without immediate values (non-control flow)
 pub fn op_no_imm_strategy() -> impl Strategy<Value = Operation> {
@@ -203,16 +206,19 @@ pub fn mast_forest_strategy(params: MastForestParams) -> impl Strategy<Value = M
     };
 
     // 1) Generate a Vec<BasicBlockNode> with length in `params.blocks`
-    prop::collection::vec(any_with::<BasicBlockNode>(bb_params), params.blocks.clone())
+    (
+        prop::collection::vec(any_with::<BasicBlockNode>(bb_params), params.blocks.clone()),
+        prop::collection::vec(any::<Decorator>(), params.decorators as usize..=params.decorators as usize)
+    )
         // 2) Map concrete blocks -> build a concrete MastForest
-        .prop_map(move |blocks| {
+        .prop_map(move |(blocks, decorators)| {
             let mut forest = MastForest::new();
 
             // Pre-populate the decorator ID space so referenced IDs are valid.
-            // TODO: Replace Decorator::Trace(i) with Arbitrary for Decorator
-            for i in 0..params.decorators {
+            // Generate all decorator types for more comprehensive testing
+            for decorator in decorators {
                 forest
-                    .add_decorator(Decorator::Trace(i))
+                    .add_decorator(decorator)
                     .expect("Failed to add decorator");
             }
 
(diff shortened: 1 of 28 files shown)


core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -194,42 +198,214 @@ impl Default for MastForestParams {
...
+}
+
+impl Arbitrary for AssemblyOp {
+    type Parameters = ();
+    type Strategy = BoxedStrategy<Self>;
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+        (
+            any::<bool>(),
+            prop::collection::vec(any::<char>(), 1..=20)
```
bitwalker, line 269: There are better primitives for generating strings in `proptest` IMO - but we should ensure that the chars are at least printable, e.g. using `is_ascii_graphic` or generating from a specific range (see how we do it for `Ident` for example).

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -194,42 +198,214 @@ impl Default for MastForestParams {
...
+                .prop_map(|chars| chars.into_iter().collect()),
+            prop::collection::vec(any::<char>(), 1..=20)
```
bitwalker, line 271: *Suggestion:* Here, a handy trick we could use would be to generate a random `u8` and then treat that as the discriminant of the `Operation` enum from `miden_core` using a transmute. We can then render the `Operation` as a string so that we are always generating a valid operation name. We may also need to filter out some `u8` values, since I think a few are unused, but that's a minor detail.

core/src/mast/serialization/decorator.rs
```diff
@@ -3,6 +3,8 @@ use alloc::vec::Vec;
 use miden_debug_types::{Location, Uri};
 use num_derive::{FromPrimitive, ToPrimitive};
 use num_traits::{FromPrimitive, ToPrimitive};
+#[cfg(all(feature = "arbitrary", feature = "std"))]
```
bitwalker, line 6: Do we actually require gating on libstd for this? AFAIK we only require liballoc at most, but maybe I'm forgetting somewhere that needs libstd functionality in the `Arbitrary` impl.

package/Cargo.toml
```diff
@@ -21,7 +21,7 @@ doctest = false
 default = ["std", "serde"]
 arbitrary = ["dep:proptest-derive", "dep:proptest", "miden-assembly-syntax/arbitrary"]
 std = ["miden-assembly-syntax/std", "miden-core/std", "serde/std", "thiserror/std"]
-serde = ["dep:serde", "dep:serde-untagged", "miden-assembly-syntax/serde", "miden-core/serde"]
+serde = ["dep:serde", "dep:serde-untagged", "dep:serde_json", "miden-assembly-syntax/serde", "miden-core/serde"]
```
bitwalker: I don't think we should be adding `serde_json` here except via `dev-dependencies` - AFAIK, it is only used for testing, and we don't want to impose it as an actual dependency of the published crate.

package/Cargo.toml
```diff
@@ -33,5 +33,9 @@ derive_more = { version = "2.0", features = ["from"] }
 proptest = { workspace = true, optional = true }
 proptest-derive = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
```
bitwalker: I think we want to move this to `dev-dependencies` and make it non-optional there

bitwalker (changes requested):
Looks good! Nice to have tests covering more of the winterfell serialization for sure. The only reason I'm requesting changes is due to the addition of `serde_json` as a non-dev dependency of `miden-mast-package` - I don't think we want that; but if you think it _is_ needed, I'd like to know more about why.

Other than that, the rest is just nits/suggestions, so 👍

core/src/mast/serialization/decorator.rs
huitseeker: This is required in `proptest_derive` for instances of `Arbitrary`. Which should not be an actual issue since we actually use those `Arbitrary` instances in a context that has std imported through testing.

package/Cargo.toml
bitwalker: It looks like `serde_json` is still being added as a regular dependency, rather than a dev-dependency.

bitwalker (changes requested):
A fresh review was requested, but it looks like the `serde_json` dependency issue I raised in my last round is still present, so I'm assuming that was maybe unintentional? Let me know

huitseeker:
@bitwalker Yeah 🤦  my apologies, I had back-ported the change to #2200 and must have lost the changes to `package/Cargo.toml` in a rebase.

bitwalker (approved):

bobbinth:
Is there anything outstanding on this PR? Or just waiting for another review?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -194,42 +198,214 @@ impl Default for MastForestParams {
...
-    // 1) Generate a Vec<BasicBlockNode> with length in `params.blocks`
-    prop::collection::vec(any_with::<BasicBlockNode>(bb_params), params.blocks.clone())
-        // 2) Map concrete blocks -> build a concrete MastForest
-        .prop_map(move |blocks| {
-            let mut forest = MastForest::new();
+    fn arbitrary_with(params: Self::Parameters) -> Self::Strategy {
+        // BasicBlockNode generation must reference decorator IDs in [0, decorators)
+        let bb_params = BasicBlockNodeParams {
+            max_decorator_id_u32: params.decorators,
+            ..Default::default()
```
plafer, line 209: Not from this PR, but the default value for `BasicBlockNodeParams::max_ops_len` is `32`. We should probably make this value greater than the max number of ops in a batch (72) - maybe `72 * 2`? So that we can catch bugs that only surface when we have multiple batches.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -194,42 +198,214 @@ impl Default for MastForestParams {
...
 }
 
-/// Strategy for generating MastForest instances
-pub fn mast_forest_strategy(params: MastForestParams) -> impl Strategy<Value = MastForest> {
-    // BasicBlockNode generation must reference decorator IDs in [0, decorators)
-    let bb_params = BasicBlockNodeParams {
-        max_decorator_id_u32: params.decorators,
-        ..Default::default()
-    };
+impl Arbitrary for MastForest {
```
plafer, line 201: Q: This works for now, but a "full" `Arbitrary` implementation for `MastForest` would be able to generate more complicated trees that contain more than just basic blocks, right? Should we create an issue for this?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -194,42 +198,214 @@ impl Default for MastForestParams {
...
+                // Generate all decorator types for more comprehensive testing
+                for decorator in decorators {
+                    forest
+                        .add_decorator(decorator)
+                        .expect("Failed to add decorator");
+                }
+
+                // Insert the generated blocks into the forest
+                for block in blocks {
+                    forest.add_node(block).expect("Failed to add block");
```
plafer: Should we make each of those procedure roots (with `MastForest::make_root()`)? Both for the serde tests, as well as future proptests that rely on MAST forest procedures.

core/src/mast/node/basic_block_node/arbitrary.rs
plafer: But if we accept that the default strategy (or "arbitrary") for `AssemblyOp` is pretty much just for the purposes of `serde`, this is probably fine. Any test that relies on the generated values' semantics would instantly fail for a number of reasons (e.g. `dummy.rs` doesn't exist) - so we'd need to implement a different `Strategy` for those anyways.

crates/utils/serde-test-macros/src/lib.rs
```diff
@@ -22,25 +22,70 @@ use syn::{AttributeArgs, Ident, Item, Lit, Meta, MetaList, NestedMeta, Type, par
...
-/// `types` can be used multiple times.
+/// # Configuration Attributes
 ///
-/// # Example
+/// The macro supports configuration attributes to control test generation:
+///
+/// | Attribute   | Type | Default | Purpose | Features Required |
+/// |-------------|------|---------|---------|-------------------|
+/// | `serde_test` | `bool` | `true` | Generate standard Serde round-trip tests | `arbitrary`, `serde`, `test` |
+/// | `winter_serde` | `bool` | `false` | Generate Winterfell-style round-trip tests | `arbitrary`, `test` |
```
plafer, line 35: By "winterfell-style", we mean "using winterfell's `Serializable` and `Deserializable` traits, right? I would clarify that here, as this was not clear to me before reading the implementation

plafer (approved):
LGTM! Left some questions and nits separately

core/src/mast/node/basic_block_node/arbitrary.rs
huitseeker: Yes, of course. #2243

core/src/mast/node/basic_block_node/arbitrary.rs
huitseeker: So that's the reason that the `BasicBlockNodeParams` are configurable: both generation and shrinking on failure are costly, so this is meant to minimize CI testing costs with a conservative default. 
We should definitely run this on larger forests in targeted tests, and this parameter allows us to do so.

PR #2225, opened 2025-09-27
feat(cli): add --kernel flag to run command

### LINKED ISSUE #1321: Allow setting the kernel via the CLI
### Task

Add a flag to the run command, when used compiles and configures a kernel to the VM.

avorylli:
Add support for specifying a kernel file when running Miden Assembly programs via CLI.

- Add `--kernel` flag to `miden-vm run` command
- Implement kernel compilation and linking functionality
- Support custom kernel procedures in program execution

Closes #1321


miden-vm/src/cli/run.rs
```diff
@@ -166,12 +170,26 @@ fn run_masm_program(params: &RunCmd) -> Result<(ExecutionTrace, [u8; 32]), Repor
         }
     }
 
+    // validate kernel file if provided
+    if let Some(kernel_path) = &params.kernel_file {
```
bitwalker, line 174: We should support specifying a kernel in either `.masp` (i.e. `Package`) or `.masm` forms. More often than not, the kernel will already be compiled to a package, so specifying the `.masp` will be more useful.

The same is true of any libraries passed in - I would expect most of them to be already assembled to packages.

bitwalker (changes requested):
Looks good! We need to make sure packages are supported as well as MASM sources though (see my other comment), other than that 👍

avorylli:
@bitwalker fixed clippys, probably everything good now

bobbinth:
@avorylli - I tried to resolve merge conflicts but seems like it broke things. Would you be able to bring this branch up to date?

avorylli:
@bobbinth I dont think I'll be able to bring this branch up to date,

Maybe I should open a new PR and fix this issue from scratch? If you like this option could you please assign me to this issue?

bobbinth:
@avorylli - sounds good! could you comment on the issue? Otherwise, it doesn't seem like I can assign it to you.

PR #2228, opened 2025-09-28
Ci: drive CI with rust-toolchain.yml, add MSRV adjustment automation

huitseeker:
This is essentially a re-do of https://github.com/0xMiden/crypto/pull/547 on this repo. Copying the PR description from there.
----
This adds a CI job (locally runnable) that will check the MSRV of a PR is set correctly on a crate-by-crate basis. If the MSRV is set incorrectly, due to e.g. some feature being used, it will compute the correct MSRV for that crate and indicate it in its error. 

The script's error message can be emulated by manually changing the MSRV to an incorrect version (e.g. "1.87") and running it (`./scripts/check-msrv.sh`).
<details><summary> Sample output (click to unfold) </summary>

```
Checking MSRV for workspace members...
Checking miden-crypto in /Users/huitseeker/tmp/crypto/miden-crypto
   Current MSRV: 1.87
ERROR: MSRV check failed for miden-crypto
Searching for correct MSRV for miden-crypto...
   Found actual MSRV: 1.88.0

ERROR SUMMARY for miden-crypto:
   Package:   miden-crypto
   Directory: /Users/huitseeker/tmp/crypto/miden-crypto
   Current (incorrect) MSRV: 1.87
   Correct MSRV:             1.88.0

TO FIX:
   Update rust-version in /Users/huitseeker/tmp/crypto/miden-crypto/Cargo.
toml from "1.87" to "1.88.0"

   Or run this command (portable in-place edit):
     sed_i 's/^\s*rust-version\s*=\s*"1.87"/rust-version = "1.88.0"/' "/Us
ers/huitseeker/tmp/crypto/miden-crypto/Cargo.toml"
--------------------------------------------------------------------------
-----
```
</details>

This further adds a rust-toolchain file that will make CI tooling (and local rustup dev envs) follow the latest stable as it updates.

As a consequence, the following behaviors can change from being proactive to being on-demand:
- bumping the MSRV.

As a side effect, we can expect the MSRV setting to (over time) accurately reflect what we can compile with. Right now, the project or any of its dependents cannot be compiled with Rust 1.88 (and there is not technical reason to impose this restriction).

----

The local difference is this project has an MSRV of 1.90 (and the current PR does not modify it). The innovation here is the toolchain file bumps CI to 1.90.


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "stable"
```
bitwalker: My only suggestion is that we set this to our current desired MSRV toolchain, e.g. `"1.90"` or `"1.88"`, or whatever makes the most sense.

This will ensure that building/testing the project from the repo will always use the exact same toolchain, as opposed to whatever `stable` happens to mean at the moment on the persons machine (or in CI)

PR #2217, opened 2025-09-25
Fix EXT2 ops docs

Al-Kindi-0:
## Describe your changes
Specifies the irreducible polynomial used to define our quadratic extension field.

Closes #2179

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](../CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #2229, opened 2025-09-28
doc: escape brackets in rustdoc

huitseeker:
Fixes a rust doc typo from #2217.


bobbinth (approved):
Looks good! Thank you!

huitseeker:
The fact that CI passes on #2228 (which only interacting change is that it stacks one commit which adds a ` rust-toolchain.toml`  file on top of this) shows that this change is good. Merging in the interest of making next green again.

PR #2230, opened 2025-09-29
chore: fix link checker job 

huitseeker:
- Add GNU make manual URL exclusion to lychee.toml to prevent 429 errors
- Set fail-fast=false in CI workflow to allow job to continue (and open an issue) after link failures


bobbinth (approved):
Looks good! Thank you!

PR #2232, opened 2025-09-29
WIP: implement `smt::set` for leaves with multiple pairs

Qyriad:
## Describe your changes

This is a draft but working implementation for supporting leaves with multiple pairs in `smt::set`. Promotion (from single to multi), demotion (opposite), insertion, and removal are all working.

This branch is also based atop #2048.

Docs and further cleanup coming shortly.

- Updated `CHANGELOG.md'


bobbinth:
Thank you! Now that https://github.com/0xMiden/miden-vm/pull/2048 is merged, could you rebase this?

Also, after taking quick look at the code, is my understanding roughly correct that it works as follows:

- First, we detect the type of update. There are two categories:
  - We are adding an entry into a single-entry leaf, making it a multi-entry leaf.
  - We are modifying an existing multi-entry leaf.
- In the first case (starting with a single-entry leaf), we just need to figure out the order. After this, we lay the leaf out in memory, compute its hash, update the advice map, and finally update the tree root.
- In the second case (modifying the existing leaf), we have cases:
  - Updating a value of an existing entry. This one is the simplest to handle, we just write the entries in memory, find and update the required entry, and then update the tree based on this.
  - Adding a new entry. This is a bit more tricky - we need to find a place to insert, then we need to "splice" a new entry at this location. This would require "moving" (using memory copying) the "tail" for the entry by 2 words. Once this is done, we hash the leaf and update the tree.
  - Removing an entry. This is also a bit tricky. We find the place where we want to remove the entry, and then we need to "move" the tail again by 2 words - but now we move it "forward". This also requires mem copies. After this, we hash the leaf and update the tree.

Is this how it works, or did I miss something?

gswirski:
@Qyriad is no longer working on this. Let's close this PR (I cannot change its status to Open) and move to https://github.com/0xMiden/miden-vm/pull/2248

Your understanding is roughly correct, but:
1. We divide it into even more cases, the full list below
2. We don't "splice" the array in place; instead, we copy a prefix, insert/delete a key-value pair and then copy a suffix. You'll find this logic in `insert_into_leaf`, `remove_in_multi_leaf`

Full branching list:
1. `set`: check if the existing leaf is empty
	1. *empty*: `set_empty_leaf`: check if inserting an empty value (existing implementation)
		1. *empty*: verify adv_provider claims and don't modify
		2. *not_empty*: `insert_hdword`
	2. *not_empty*: check if inserting into a single leaf
		1. *single*: `set_single_leaf`: check if inserting an empty value (existing implementation)
			1. *empty*: `remove_single_leaf`: check if we are removing the same key
				1. `insert_hdword`
				2. verify adv_provider claims and don't modify
			2. *not_empty*: `insert_single_leaf`: check if we are inserting the same key
				1. `insert_hdword`
				2. `insert_single_to_multi_leaf`: check if new key smaller/larger than the old one. layout memory accordingly and call `hash_and_insert_mem`
		2. *multiple*: `set_multi_leaf`: pipe preimage: `insert_or_replace_in_multi_leaf`. check if key exists in the preimage
			1. *key exists*: check if inserting an empty value
				1. `remove_in_multi_leaf`: copy array without a key and `rpo::hash_memory_words`
				2. `replace_in_multi_leaf`: update memory in place and `rpo::hash_memory_words`
			2. *key not found*: check if inserting an empty value
				1. *empty*: no op
				2. not empty: `insert_into_leaf`: copy array with a new key and `rpo::hash_memory_words`

I'll copy this comment into the new PR for reference.

Issue #2236, opened 2025-09-30
Build failure of `build.rs` `fs::write`


drahnr:
### Packages versions

miden-stdlib: 0.18

### Bug description

I occassionall (locally, CI https://github.com/0xMiden/miden-base/actions/runs/18130495930/job/51595736500?pr=1832 ) get the following error.

```
thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/miden-stdlib-0.18.0/build.rs:64:32:
  called `Result::unwrap()` on an `Err` value: Os { code: 17, kind: AlreadyExists, message: "File exists" }
```

There are two things
* use `fs-err` to get a path for almost free
* ensure `truncate` + `file lock` are used to ensure a coherent version exists

### Relevant log output

```shell

```

bobbinth:
I wonder if this is related to #2231 somehow. Both seems to be issues in `build.rs` of `miden-stdlib`.

bobbinth:
Should be closed by https://github.com/0xMiden/miden-vm/pull/2239 - but feel free to re-open if not.

PR #2239, opened 2025-10-01
fix: improve stdlib build.rs robustness for filesystem errors

### LINKED ISSUE #2231: Doc build for `miden-vm` and `miden-stdlib` 0.18 fails on docs.rs
The docs for `miden-vm` and `miden-stdlib` 0.18 (probably related) don't build on docs.rs.

```
[INFO] [stderr]   thread 'main' (15) panicked at build.rs:63:36:
[INFO] [stderr]   called `Result::unwrap()` on an `Err` value: Os { code: 30, kind: ReadOnlyFilesystem, message: "Read-only file system" }
```

We had a read-only error in miden-lib builds some time ago - there is some info in this PR for reference: https://github.com/0xMiden/miden-base/pull/970.

huitseeker:
Resolves build failures in CI/CD environments by adding:

- Read-only filesystem detection w/ skip behavior
- Safer directory/file operations w/ error handling
- Better error messages with file path (fs_err)

This addresses issues where documentation generation would fail with:
- "File exists" errors (Os code: 17)
- "Read-only file system" errors (Os code: 30)

Fixes #2231 #2236


bobbinth:
It would be good to make this PR against the `main` branch (assuming these are non-breaking changes). This way, we can make a patch release that would immediately fix things downstream.

stdlib/build.rs
```diff
@@ -40,6 +42,26 @@ impl MarkdownRenderer {
 // HELPER FUNCTIONS
 // ================================================================================================
 
+/// Check if the filesystem at the given path is read-only
+static READ_ONLY_CACHE: OnceLock<bool> = OnceLock::new();
+
+fn is_readonly_filesystem(target_dir: &Path) -> bool {
```
drahnr: Can't we match against `std::io::ErrorKind::ReadOnlyFilesystem` in the actual file operation we want to do?

stdlib/build.rs
```diff
@@ -59,9 +81,34 @@ fn markdown_file_name(ns: &str) -> String {
 /// Writes Miden standard library modules documentation markdown files based on the available
 /// modules and comments.
 pub fn build_stdlib_docs(asm_dir: &Path, output_dir: &str) -> io::Result<()> {
-    // Remove docs folder to re-generate
-    fs::remove_dir_all(output_dir).unwrap();
-    fs::create_dir(output_dir).unwrap();
+    let output_path = Path::new(output_dir);
+
+    // Check if filesystem is read-only by probing the actual target directory
+    fs::create_dir_all(output_path).ok(); // ignore here; probe will decide
```
drahnr:
```suggestion
    match fs::create_dir_all(output_path) { Err(e) if e.kind() == std::io::ErrorKind::ReadOnlyFileSystem => return Ok(()), Ok(_) => {}, e => e?, }
```

stdlib/build.rs
```diff
@@ -76,11 +123,17 @@ pub fn build_stdlib_docs(asm_dir: &Path, output_dir: &str) -> io::Result<()> {
...
 
-        let mut f = fs::OpenOptions::new()
-            .write(true)
-            .create(true)
-            .truncate(true)
-            .open(output_file_path)?;
+        let mut f = match fs::File::create(&output_file_path) {
+            Ok(file) => file,
+            Err(e) => {
+                eprintln!(
```
drahnr: I think we need to use `println!("cargo::warn=Your message here!")` to make it visible to the user

stdlib/build.rs
```diff
@@ -76,11 +123,17 @@ pub fn build_stdlib_docs(asm_dir: &Path, output_dir: &str) -> io::Result<()> {
             fs::create_dir_all(parent)?;
         }
 
-        let mut f = fs::OpenOptions::new()
-            .write(true)
-            .create(true)
-            .truncate(true)
-            .open(output_file_path)?;
+        let mut f = match fs::File::create(&output_file_path) {
```
drahnr: nit: `inspect_err` might be more concise way to express this

stdlib/build.rs
huitseeker: I'm just trying to capture a much broader set of issues that will pop up as `PermissionDenied`. `ErrorKind::ReadOnlyFilesystem` will come up reliably on a ROFS on Linux, but on Windows (ACLs, file locks, ..), or a network fs (sccache used?), I expect a semantically similar error may come up under other error codes.

drahnr: > network fs (sccache used?)

I am not aware of a network fs in the traditional sense used with sccache, it's all request response based, for all backends.

> I'm just trying to capture a much broader set of issues that will pop up as

Not sure we want to skip docs generation entirely in all cases over showing an actionable error message, do we?

huitseeker:
@drahnr The protection against `FileAlreadyExists` comes from a particularity of create_dir_all (instead of the previously used create_dir), which is that [it does not error when the directory that is passed as argument already exists](https://doc.rust-lang.org/beta/std/fs/fn.create_dir_all.html#errors). `File::create` [creates or truncates](https://doc.rust-lang.org/std/fs/struct.File.html#method.create). I suspect this is being hit in concurrent rebuilds of this library *as a dependency*.

Now, one could definitely argue the current fix does not prevent concurrency errors properly, since this still follows a non-atomic removal-then-recreate pattern, but it has the advantage of being a smaller patch, which is easy to audit. The canonical pattern if we are dealing with *serious* concurrency errors is to write to a temporary filesystem, then atomically replace the other docs directory with our newly written one.

However that atomicity is a bit annoying to achieve, [renaming doesn't work if the new name is on a different mount point](https://doc.rust-lang.org/beta/std/fs/fn.rename.html) and we would need to fall back to a copy in this case, leading to a significantly larger diff. For that there's also specialized crates that can help, or a different policy for how we build the doc that would not rely on writing outside of the crate's `OUT_DIR`.

stdlib/build.rs
huitseeker: See comment [above](https://github.com/0xMiden/miden-vm/pull/2239/files#r2393569989)

stdlib/build.rs
huitseeker: > Not sure we want to skip docs generation entirely in all cases over showing an actionable error message, do we?

This is probably hit when this library is built as a dependency, a case where I'm happy skipping doc rebuilds

stdlib/build.rs
```diff
@@ -59,9 +81,34 @@ fn markdown_file_name(ns: &str) -> String {
...
-    fs::remove_dir_all(output_dir).unwrap();
-    fs::create_dir(output_dir).unwrap();
+    let output_path = Path::new(output_dir);
+
+    // Check if filesystem is read-only by probing the actual target directory
+    fs::create_dir_all(output_path).ok(); // ignore here; probe will decide
+    if is_readonly_filesystem(output_path) {
+        eprintln!("Warning: Read-only filesystem detected, skipping documentation generation");
+        return Ok(());
+    }
```
PhilippGackstatter: For the docs.rs build issue, maybe the simplest solution here would be to do:

```suggestion
    if std::env::var("DOCS_RS").is_ok() {
        eprintln!("Warning: Read-only filesystem detected, skipping documentation generation");
        return Ok(());
    }
```

Reference: https://docs.rs/about/builds#detecting-docsrs

But I haven't yet understood what exactly the cause of the "file exists" error was, so if the current solution is necessary to fix this, maybe this point is moot.

bobbinth:
What's outstanding on this PR? I'd like to merge it sooner rather than later (i.e., ideally today).

In `miden-base`, I believe we've taken a slightly different approach to addressing this. Specifically, we have an [env variable](https://github.com/0xMiden/miden-base/blob/next/crates/miden-lib/build.rs#L31) which controls whether we write anything to the filesystem. If this variable is set (as we do in our `Makefile` command), then we generate extra files. Otherwise (as it would happen in all cases when `Makefile` command are not used) we don't. I think this addresses both the issue with read-only filesystem and also building as a dependency.

But I'm fine with the approach in this PR too (assuming it addresses the problem).

stdlib/build.rs
bobbinth, line 191: Do we need to add something like:
```rust
println!("cargo::rerun-if-env-changed=MIDEN_BUILD_STDLIB_DOCS");
```
Somewhere here?

stdlib/build.rs
bobbinth: Do we still need this check (and the associated code)?

bobbinth (commented):
Looks good! Thank you! I added a couple questions inline.

stdlib/build.rs
huitseeker: Arguably not, but I'd note `docs.rs` is just *one* builder that sandboxes the build environment in a setup where everything but the `$OUT_DIR` is readonly. Others include [nix's crane](https://crane.dev/faq/sandbox-unfriendly-build-scripts.html), [bazel](https://bazel.build/docs/sandboxing), etc.

But I'll clean that up, sure.

bobbinth (approved):
Looks good! Thank you!

stdlib/build.rs
```diff
@@ -59,28 +60,32 @@ fn markdown_file_name(ns: &str) -> String {
...
     // Render the modules into markdown
     for (label, file_path) in modules {
-        let relative_path = markdown_file_name(&label);
-        let output_file_path = Path::new(output_dir).join(relative_path);
+        let relative = markdown_file_name(&label);
+        let out = output_path.join(&relative);
 
         // Create directories if needed
-        if let Some(parent) = output_file_path.parent() {
+        if let Some(parent) = out.parent() {
```
drahnr, line 84: nit: By definition every file path has a parent.

stdlib/build.rs
```diff
@@ -203,7 +213,9 @@ fn main() -> io::Result<()> {
     stdlib.write_to_file(output_file).map_err(|e| io::Error::other(e.to_string()))?;
 
     // Generate documentation
-    build_stdlib_docs(&asm_dir, DOC_DIR_PATH)?;
+    if std::env::var("MIDEN_BUILD_STDLIB_DOCS").is_ok() {
```
drahnr, line 216: https://docs.rs/about/builds#detecting-docsrs I think it'd be good to use `DOCS_RS` either exclusively or in addition, to my understanding the current approach es explicitly opting into building the docs. This is good for using makefiles but editor integrations might not be aware of this, so requires user action.

huitseeker: I don't understand. 

This is static generation of files that get checked into the repo, and that we verify on CI are indeed checked into the repo (#2156). 

Docs.rs is welcome to use those checked in files, we shouldn't need to refresh them, and in fact `cargo publish` will prevent us from pushing source to crates.io that creates any unclean state (outside its $OUT_DIR) on rebuild — meaning we won't pass `cargo publish` if our docs need refreshing.

PR #2240, opened 2025-10-01
fix(library): add Optional FunctionType generation in Arbitrary for LibraryExport

sashass1315:
- Implement proptest strategy to generate Option<FunctionType> for LibraryExport
- Use limited CallConv variants and simple Type set (Felt, U32, U64)
- Convert Vec to SmallVec for params/results to match existing types

### DIFF
--- a/crates/assembly-syntax/src/library/mod.rs
+++ b/crates/assembly-syntax/src/library/mod.rs
@@ -88,13 +88,8 @@ impl Arbitrary for LibraryExport {
         let params = prop_vec(simple_type.clone(), 0..=4);
         let results = prop_vec(simple_type, 0..=2);
 
-        // Limited ABI space sufficient for roundtrip coverage
-        let abi = prop_oneof![
-            Just(midenc_hir_type::CallConv::Fast),
-            Just(midenc_hir_type::CallConv::SystemV),
-            Just(midenc_hir_type::CallConv::Wasm),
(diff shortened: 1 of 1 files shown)


crates/assembly-syntax/src/library/mod.rs
```diff
@@ -74,19 +74,43 @@ impl LibraryExport {
...
+
+        // Generate a small set of simple types for params/results to keep strategies fast/stable
+        let simple_type = prop_oneof![Just(Type::Felt), Just(Type::U32), Just(Type::U64),];
+
+        // Small vectors of params/results
+        let params = prop_vec(simple_type.clone(), 0..=4);
+        let results = prop_vec(simple_type, 0..=2);
+
+        // Limited ABI space sufficient for roundtrip coverage
+        let abi = prop_oneof![
```
bitwalker: Just use `Fast` here, we don't need to use the others for now

bitwalker (changes requested):
Looks good, just one tweak I'd like to make before merging

plafer (approved):

crates/assembly-syntax/src/library/mod.rs
sashass1315: > Just use `Fast` here, we don't need to use the others for now

Ok, thanks for review

PR #2244, opened 2025-10-02
fix: replace unsafe expect("TODO") with proper error handling in ACE circuit tests

crStiv:
Replace expect("TODO") calls with proper error handling in Circuit::evaluate method

These changes are necessary because:
1. expect("TODO") with placeholder messages is dangerous in production code
2. Panics from expect() calls can cause DoS attacks or system crashes
3. Proper error handling provides better debugging information and graceful failure modes
4. The existing validation in Circuit::new() ensures these errors should never occur, but 
   defensive programming requires proper error handling


adr1anh:
The code is only used in testing.

PR #2245, opened 2025-10-02
fix: guard parsing of hex words against missing 0x prefix

strmfos:
## Describe your changes
- prevent panic in InputFile::parse_word when inputs omit the 0x prefix
- return descriptive error for malformed word strings and keep behavior otherwise unchanged
- verified with cargo test internal::tests::test_merkle_data_parsing and cargo test miden_vm::internal::tests::test_merkle_data_parsing


strmfos:
Thanks for the review! Added the requested unit tests in test_merkle_data_parsing module:

- `test_parse_word_missing_0x_prefix()` - verifies error handling for missing "0x" prefix
- `test_parse_word_edge_cases()` - covers empty strings, just "0x", and short hex
- `test_parse_word_valid_hex()` - ensures valid hex parsing still works

bitwalker (approved):

miden-vm/src/internal.rs
```diff
@@ -264,7 +264,9 @@ impl InputFile {
 
     /// Parse a `Word` from a hex string.
     pub fn parse_word(word_hex: &str) -> Result<Word, String> {
```
bobbinth, line 266: Theoretically, we should be able to replace this function with `Word::try_from()` - though, this would not support "incomplete strings" (i.e., the string would need to be 66 characters). Maybe it is worth updating how `Word::try_from()` works in `miden-crypto`? cc @huitseeker.

PR #2247, opened 2025-10-02
fix: after_exit executed incorrectly in External nodes

### LINKED ISSUE #2005: `FastProcessor` executes "after exit" decorators incorrectly for external nodes
#1990 introduced a bug where the `after_exit` decorators are executed instantly rather than after executing the external node ([here](https://github.com/0xMiden/miden-vm/blob/0772ffb599f898502c728d2c80576ae3aa29b80b/processor/src/fast/mod.rs#L796)).

huitseeker:
Modified continuation_stack.rs: Added FinishExternal((MastNodeId, Arc<MastForest>)) continuation type to store both node ID and forest reference for after_exit decorator execution.

Fixes #2005

### DIFF
--- a/processor/src/fast/external.rs
+++ b/processor/src/fast/external.rs
@@ -34,7 +34,9 @@ impl FastProcessor {
         // Push the root node of the external MAST forest onto the continuation stack.
         continuation_stack.push_start_node(resolved_node_id);
 
-        self.execute_after_exit_decorators(external_node_id, current_forest, host)?;
+        // Push a continuation to execute after_exit decorators when we return from the external
+        // forest
+        continuation_stack.push_finish_external(external_node_id, current_forest.clone());
 
         // Update the current forest to the new MAST forest.
         *current_forest = new_mast_forest;
--- a/processor/src/continuation_stack.rs
+++ b/processor/src/continuation_stack.rs
@@ -26,6 +26,8 @@ pub enum Continuation {
     FinishCall(MastNodeId),
     /// Process the finish phase of a Dyn node.
     FinishDyn(MastNodeId),
+    /// Process the finish phase of an External node (execute after_exit decorators).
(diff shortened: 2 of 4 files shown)


processor/src/fast/external.rs
```diff
@@ -34,7 +34,9 @@ impl FastProcessor {
         // Push the root node of the external MAST forest onto the continuation stack.
         continuation_stack.push_start_node(resolved_node_id);
 
-        self.execute_after_exit_decorators(external_node_id, current_forest, host)?;
+        // Push a continuation to execute after_exit decorators when we return from the external
+        // forest
+        continuation_stack.push_finish_external(external_node_id, current_forest.clone());
```
plafer: I think this pushes it in the wrong order; as is, the next continuation we'll pop is the `FinishExternal` (instead of executing `resolved_node_id`). I think the order needs to be

```rs
// note: no need to push the `current_forest`
continuation_stack.push_finish_external(external_node_id);

continuation_stack.push_enter_forest(current_forest.clone());
continuation_stack.push_start_node(resolved_node_id);
```

Such that the order of execution becomes

1. execute the (resolved) external node (and all its children)
2. re-enter `current_forest`
3. finish external node (which can now use `current_forest` in `execute_impl`, due to (2) running before)

Didn't check the details of why the test didn't fail - but I suspect that one where some traces are attached to the resolved external node, we wouldn't be seeing all the traces in the expected order (i.e. the after_exit of the external node would run before the traces attached to the resolved node).

bitwalker: +1

huitseeker: Yes 🤦 
Writing a test that actually tracks something done in the external would have helped, rather than just positioning the `before_enter` w.r.t `after_exit`.
Thanks!

plafer (approved):
LGTM!

PR #2248, opened 2025-10-02
implement `smt::set` for leaves with multiple pairs

gswirski:
## Describe your changes
Support leaves with multiple pairs in smt::set. Promotion (from single to multi), demotion (opposite), insertion, and removal are all working.

This PR supersedes https://github.com/0xMiden/miden-vm/pull/2232

Full branching list:
1. `set`: check if the existing leaf is empty
	1. *empty*: `set_empty_leaf`: check if inserting an empty value (existing implementation)
		1. *empty*: verify adv_provider claims and don't modify
		2. *not_empty*: `insert_hdword`
	2. *not_empty*: check if inserting into a single leaf
		1. *single*: `set_single_leaf`: check if inserting an empty value (existing implementation)
			1. *empty*: `remove_single_leaf`: check if we are removing the same key
				1. `insert_hdword`
				2. verify adv_provider claims and don't modify
			2. *not_empty*: `insert_single_leaf`: check if we are inserting the same key
				1. `insert_hdword`
				2. `insert_single_to_multi_leaf`: check if new key smaller/larger than the old one. layout memory accordingly and call `hash_and_insert_mem`
		2. *multiple*: `set_multi_leaf`: pipe preimage: `insert_or_replace_in_multi_leaf`. check if key exists in the preimage
			1. *key exists*: check if inserting an empty value
				1. `remove_in_multi_leaf`: copy array without a key and `rpo::hash_memory_words`
				2. `replace_in_multi_leaf`: update memory in place and `rpo::hash_memory_words`
			2. *key not found*: check if inserting an empty value
				1. *empty*: no op
				2. not empty: `insert_into_leaf`: copy array with a new key and `rpo::hash_memory_words`

I'll copy this comment into the new PR for reference.

- Updated `CHANGELOG.md'


stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,572 @@
...
+    # By nature, if an insertion transforms a single-leaf to a multi-leaf, then the old value at
+    # this key is an empty word.
+
+    padw swapw dropw
+    # => [Z, R_new]
+end
+
+
+# MAX_LEAF_SIZE + 4
+proc.insert_into_leaf.8196
```
huitseeker: Noticed the O(n) memory copying pattern in `insert_into_leaf` and `remove_in_multi_leaf`. Curious about your thoughts on when we might need to optimize for leaves with many pairs - thinking ahead now could help avoid breaking changes later if we need a hybrid approach with advice provider. 

Not asking for any changes to this PR, just wondering if it makes sense to have e.g. caps in input sizes here.

gswirski: We already limit the number of k-v pairs to 1024: https://github.com/0xMiden/crypto/pull/455

One idea that Bobbin floated was to "splice" the original. Instead of copying the whole thing, identify the insertion/removal place and only shift elements after it. This wouldn't help with the worst-case scenario, but would improve average case. I'm a little reluctant to implement this because: it's making hidden side-effects and callers need to make sure to pad their array if they store any data right after the affected memory.

The way I currently understand it, we are still hashing in O(n), see this fragment in stdlib/asm/crypto/hashes/rpo.masm:

```
export.absorb_double_words_from_memory
  dup.13 dup.13 neq  # (4 cycles )
  while.true
    mem_stream hperm # (2 cycles)
    dup.13 dup.13 neq  # (4 cycles )
  end
```

We could potentially be smarter here. Instead of copying the array and then calling `mem_stream`, we could rebuild the `exec.rpo::hash_memory_words` method to call `hperm` with an option to skip/insert a k-v pair in the middle. Not sure if it's worth the effort.

gswirski:
> it would be valuable to add a few tests for error cases to ensure the validation assertions in the assembly code bounce them correctly. For example: malformed leaf data (size not multiple of 8), corrupted advice provider data.

Do we have any facility to inject invalid advice data in tests?

plafer:
> Do we have any facility to inject invalid advice data in tests?

Not at the moment. But I don't think we've tested for (and try to detect) corrupted advice data anywhere else; basically it's expected that at the very least, proof verification would fail. Unless @huitseeker you had other thoughts, we could leave open an issue to think about how we want to handle corrupted advice data more generally.

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,572 @@
...
+
+    exec.hash_and_insert_mem
+    # => [NV_old, R_new, V_old, K]
+
+    dropw movupw.2 dropw swapw
+    # => [V_old, R_new]
+end
+
+# Inputs: [leaf_size, NV, V, K, R]
+proc.set_multi_leaf.8192
```
plafer: We could check that `leaf_size < 8192` to make sure we don't overflow the buffer with `pipe_double_words_preimage_to_memory`

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,572 @@
...
+
+        # cur_read += num_elems
+        swap.2 add
+        # => [cur_read', cur_write']
+    else
+        drop
+    end
+end
+
+#! Hashes an array of SMT leaves, then updates both the advice map and Merkle store.
```
plafer: Why do we also update the advice map? I see no mention of this in the docs for `get` or `set`. Does code elsewhere rely on this?

stdlib/asm/collections/smt.masm
plafer: > We already limit the number of k-v pairs to 1024

That's in the `Smt` implementation used by our `AdviceProvider`, but anyone using a different prover (not written by us) wouldn't have this restriction, and we could get buffer overflows. As mentioned in my other comment, we should check explicitly for leaf size in the masm code.

> I'm a little reluctant to implement this because: it's making hidden side-effects and callers need to make sure to pad their array if they store any data right after the affected memory.

This is a function local to this file (currently only called once), so as long as it's properly documented, I think it's fine to ask the caller to pass in a buffer that is large enough for us to do this splicing thing here, instead of copying the whole leaf twice. This could be left as an improvement though if you want, since we expect leaves to contain very few elements in the the vast majority of cases.

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,572 @@
...
+
+    # cur_write = &new_pairs[2]
+    add.4 loc_store.3
+    # => [i, R]
+
+    #
+    # STEP 3: copy from cur_read..pairs_array_end
+    #
+
+    loc_load.3 # cur_write
```
plafer: nit: we could avoid remove the `loc_store.3 loc_load.3`; it's "less clean" because the `cur_write` wouldn't be updated, but since this is local to this function, it's probably fine to remove the read/write here. We could add a comment to say where `local[3]`'s scope ends (since we don't have variable scopes in MASM 🤷‍♂️), but basically in general we should try to make `stdlib` functions as performant as possible as we expect them to be used widely.

gswirski:
I'm still working on some of the feedback. Below notes on testing and memcopy.

**Testing**
I've created a follow-up task: https://github.com/0xMiden/miden-vm/issues/2288

**Splice in place**
Removing key-value pair in place is rather easy: reuse the same buffer, memcopy from key_ptr+8...end to key_ptr. It's already done.

Adding a pair is more difficult because current memcopy will not work when the target buffer is overlapping and shifted right. We need to duplicate mamcopy into memcopy_from_end so that we don't override values that we will read next. This causes a fair bit of code duplication. I did it in this commit (outside of this PR for now): https://github.com/0xMiden/miden-vm/commit/c56c0dceae5c212e4f339f1ac2805f2e557c7728

I'm not entirely happy with that solution. I'll see if it wouldn't be easier and cleaner to just `hperm` exactly what we need.

stdlib/asm/collections/smt.masm
gswirski: Hard for me to say why this exists, but single leaf implementation was updating the advice map, too. See here: https://github.com/0xMiden/miden-vm/blob/next/stdlib/asm/collections/smt.masm#L111-L113 (+ 2 other locations in that file)

plafer: cc @bobbinth

gswirski:
A'right, I need an opinion. I've rewritten insert and remove functions so that they don't do any memory copying at all (and wouldn't need any locals either). The resulting code is both faster and more concise. All of these changes are in [the most recent commit](https://github.com/0xMiden/miden-vm/pull/2248/commits/1005446828032e5c6cbda61e4e5a9d406ebb8b01) so if we don't like it it's easy to revert.

The issue I'm having is the advice map. Setters for single leaves (and original implementation by @Qyriad) were providing `advice_map[leaf_hash] = array of key-value pairs` during the procedure. [@plafer indicated that they _might_ not be useful](https://github.com/0xMiden/miden-vm/pull/2248#discussion_r2420042359).

Now, if we don't need the advice map, I think the current implementation is superior and I'll proceed to stripping the rest of locals and advice map setters.

However, if we do need the advice map, I think we have two paths forward:
1. Use the implementation I showed yesterday that shifts the second part of the array left (removal) or right (insertion). It's a little messy and causes a lot of code duplication but works as is: https://github.com/0xMiden/miden-vm/commit/c56c0dceae5c212e4f339f1ac2805f2e557c7728
2. Continue with the new in-place hashing approach. To prevent memory copying, we could introduce a new instruction: `adv.append_mem` that would do the following: `advice_map[K] <- advice_map[K] || mem[a..b]`

Let me know what you think @bobbinth @plafer .

plafer:
*If* we don't need the advice map, then your latest changes as described sound very desirable (without actually looking at the commit). But I personally don't know if they're ever used; perhaps @Al-Kindi-0 or @PhilippGackstatter would know?

**The question ([context](https://github.com/0xMiden/miden-vm/pull/2248#discussion_r2420042359)):** when we set a leaf using `smt::set`, we also store the (key,value) pair in the advice map. AFAIK this behavior is not documented. Does anyone rely on this anywhere?

PhilippGackstatter:
> when we set a leaf using `smt::set`, we also store the (key,value) pair in the advice map. AFAIK this behavior is not documented. Does anyone rely on this anywhere?

I don't think miden-base or other components I know of rely on being able to _extract_ leaf data in this way after they have been set. We do rely on being able to _provide_ (key, value) data via the advice map ([example usage](https://github.com/0xMiden/miden-base/blob/0e4abf7a3629b9d0494685e58ea287128881d614/crates/miden-lib/src/transaction/inputs.rs#L271-L279)), but that's probably expected. Just mentioning for completeness.

The protocol extracts data via events, e.g. when a storage map (read SMT) is updated, we let the host know that (key, value) is being updated, and so we don't need to extract that from the advice map after the fact.

I guess changing this also does not have any consequences for transaction replayability, because the `smt::set` logic will just be re-executed anyway, right?

bobbinth:
Just to make sure I understand: currently, but the end of the `smt::set` we have `advice_map[new_leaf_hash] = array of key-value pairs` in the advice map, and the question is about whether we can remove this - right?

If so, it seems to me that we do need this value in the advice map because it is required for `smt::get`. Basically, if we do `smt::set` on a leaf that results in a leaf with multiple key-value pairs, and then try to do `smt::get` on the same leaf, the `smt::get` would need to write the pre-image of the leaf into memory. But if the pre-image is not in the map, it'll fail.

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
-use.std::mem
+use.std::crypto::hashes::rpo
```
huitseeker, line 1: Looks great! Consider adding a randomized round-trip test to strengthen coverage and couple with PR #2048. A test that creates non-empty SMTs with random key-value pairs, uses `smt::set` to insert values, then `smt::get` to verify round-trip correctness would help validate multi-leaf functionality.

stdlib/asm/collections/smt.masm
huitseeker: ~~IIUC both set and get expect to find leaf pre-images in the advice map with get reading (and checking) those expected inputs? IOW, a subsequent get call would fail if the advice map was not updated?~~
~~The usage of the advice map makes sense to me to keep the leaf pre-images unbounded. Or at least that's what I could gather looking at tests/collections/smt while thinking about the roundtrip test I suggest above.~~

Edit: the above is just parroting the comment by @bobbinth [here](https://github.com/0xMiden/miden-vm/pull/2248#issuecomment-3410652033)

stdlib/asm/collections/smt.masm
huitseeker: @gswirski https://github.com/huitseeker/miden-vm/commit/5a1ae04c2fde5f4f454383df06445b13cd739a6d

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+
+#! Hashes an array of SMT leaves, then updates both the advice map and Merkle store.
+#!
+#! Returns the node value (hash) of the previous leaf at the given index, as well as
+#! the root for the new SMT.
+#!
+#! Inputs: [start_addr, end_addr, leaf_index, R]
+#! Outputs: [NV_old, R_new]
+proc.hash_and_insert_mem
+    # => [start_addr, end_addr, i, R, ...]
```
Al-Kindi-0: nit: i or leaf_index?

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+
+    dup.1 dup.1
+    # => [start_addr, end_addr, start_addr, end_addr, i, R, ...]
+
+    exec.rpo::hash_memory_words
+    # => [NV, start_addr, end_addr, i, R, ...]
+
+    # We're already ready for adv.insert_mem.
+    adv.insert_mem
+
```
Al-Kindi-0: nit: output stack here would be consistent

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+    mtree_set
+    # => [NV_old, R_new, ...]
+end
+
+#! Inserts a key-value pair into a leaf currently containing a single, non-empty pair,
+#! converting the leaf to a multi-leaf.
+#!
+#! Inputs: [K_in_leaf, K, V_in_leaf, V, R]
+#! Outputs: [V_in_leaf, R_new]
```
Al-Kindi-0: nit: probably we can call it K_l and add a comment to explain what it is

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+    # Prepare for hash_memory_words,
+    locaddr.12 add.4
+    locaddr.0
+    # => [start_addr, end_addr, i, R]
+
+    exec.hash_and_insert_mem
+    # => [NV_old, R_new]
+
+    # By nature, if an insertion transforms a single-leaf to a multi-leaf, then the old value at
+    # this key is an empty word.
```
Al-Kindi-0: Q: Could you explain this a bit more?

stdlib/asm/collections/smt.masm
Al-Kindi-0: @plafer I think we do test against the limit 1024 in MASM

Regarding optimizations, I agree that we can improve the average case complexity of this, both for hashing and operand stack, though I would leave for future PRs and only after a thorough discussion

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+
+    padw swapw dropw
+    # => [Z, R_new]
+end
+
+
+# MAX_LEAF_SIZE + 4
+proc.insert_into_leaf.8196
```
Al-Kindi-0: Would be good to document the +4 in the locals

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+    exec.rpo::absorb_double_words_from_memory
+    # => [C"', B"', A"', end_ptr, end_ptr, i, R]
+
+    exec.rpo::squeeze_digest
+    # => [NV, end_ptr, end_ptr, i, R]
+
+    # Step 4, prepare for mtree_set
+
+    movup.4 drop movup.4 drop movup.8 movup.8 movup.8 movup.8 movup.8
+    # => [i, R, NV]
```
Al-Kindi-0: Would "pad movdn.4 movdnw.2 drop drop drop" be better?
Would in any case split it up and add comments

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+    mtree_set
+    # => [NV_old, R_new]
+ 
+    # By nature, if this insertion put a value where there wasn't one, then the
+    # old value at this key is an empty word.
+    dropw padw
+    # => [Z, R_new]
+end
+
+proc.insert_or_replace_in_multi_leaf
```
Al-Kindi-0: Would add documentation to explain what each helper procedure is doing

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+    # => [is_key_found, key_ptr, start_ptr, end_ptr; V, K, R]
+    if.true
+        # => [key_ptr, start_ptr, end_ptr, V, K, R]
+
+        # We're replacing or removing an existing value.
+        # To check which, we'll word-align the stack for a moment.
+        push.0
+        # => [X, V, K, R]
+        swapw padw
+        # => [Z, V, X, K, R]
```
Al-Kindi-0: nit: would define Z

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+
+        eqw movdn.8
+        # => [Z, V, V==EMPTY, 0, key_ptr, start_ptr, end_ptr, K, R]
+
+        dropw
+        # => [V, V==EMPTY, 0, key_ptr, start_ptr, end_ptr, K, R]
+
+        # Now we can use V==EMPTY for word-aligning the stack instead.
+        # Convenient!
```
Al-Kindi-0: nit: would remove this

stdlib/asm/collections/smt.masm
```diff
@@ -1,14 +1,421 @@
...
+        swapw padw
+        # => [Z, V, X, K, R]
+
+        eqw movdn.8
+        # => [Z, V, V==EMPTY, 0, key_ptr, start_ptr, end_ptr, K, R]
+
+        dropw
+        # => [V, V==EMPTY, 0, key_ptr, start_ptr, end_ptr, K, R]
+
+        # Now we can use V==EMPTY for word-aligning the stack instead.
```
Al-Kindi-0: nit: this is also probably redundant as we have already used 0 to word-align and we now will replace this with V==EMPTY to do that
I would just remove

stdlib/asm/collections/smt.masm
plafer: > I think we do test against the limit 1024 in MASM

Can you expand? Do you mean somewhere in this `smt.masm` file, we run that check (in which case I missed it on my first read)?

stdlib/asm/collections/smt.masm
gswirski: I had to make some changes to your test: https://github.com/0xMiden/miden-vm/pull/2248/commits/82ba7338ce3e36faf0e82cf447ef6fbe76c44231

- It was not exercising multi-leaves at all. For now I hard-coded it to leaf_index 42 but will work on adding more cases.
- Set and get were not connected - the test didn't catch that advice_map was wired incorrectly.

stdlib/asm/collections/smt.masm
Al-Kindi-0: It is enforced in set as 
```
# Ensure that the leaf size does not exceed the maximum size of 1024 k-v pairs.
dup u32assert u32lte.MAX_LEAF_SIZE assert.err="invalid leaf: larger than maximum size of 8192"
```
but I am not sure it is enforced everywhere

gswirski:
Summary of the changes done so far:
- Incorporated @huitseeker's round-trip test and extended it to exercise multi-leaves. This expectedly failed due to advice_map not being updated in multi-leaf set.
- Reverted my multi-leaf set optimizations and returned to Qyriad's version to pass the test above. The current implementation copies the entire leaf buffer (rather than splicing in place) as it is the easiest thing to review and audit (per @Al-Kindi-0 comments).
- Reorganized procedures so that they go from exports through high-level procedures all the way to smaller helper procedures.
- Added procedure comments and addressed other doc comments.

At this point, I'm not sure how much we care about performance in this PR vs having an obviously correct implementation that can be improved at a later date. I see three potential paths forward:
1. Focus on micro-optimizations (like removing one or two local variables that maybe are not strictly necessary).
2. Improve average-case scenario and splice in place. This introduces a bit of code duplication (if we insert and need to shift elements right, we need to memcpy from the end). Code for this exists, I can bring it in here, or we can do it in a follow-up PR: https://github.com/0xMiden/miden-vm/commit/c56c0dceae5c212e4f339f1ac2805f2e557c7728
3. Further improve speed by not doing any memory copying at all. This would require a new instruction to update advice_map (something like `adv.append_mem: advice_map[K] <- advice_map[K] . mem[a..b]`). This also drastically simplifies the code: https://github.com/0xMiden/miden-vm/pull/2248/commits/7495f964adbb2d8afb08216abe9da2d4b55d56b9

I agree with @Al-Kindi-0 that this is something that requires a discussion first. It might be valuable to merge something that works first (this PR), agree on the direction, and then push a follow-up PR which improves performance.

stdlib/asm/collections/smt.masm
```diff
@@ -1,232 +1,351 @@
...
 #! Cycles
-#!   Insert empty value: X cycles
-#!   Insert non-empty value: X cycles
-proc.set_empty_leaf
-    # Check if we're inserting the empty value (X cycles)
+#!   Leaf empty: 66 cycles
+#!   Leaf single: 115 cycles
+#!   Leaf multiple, non-empty value: 189 + 3 * pair_count
+#!   Leaf multiple, empty value: 186 + 3 * pair_count
+export.get
```
huitseeker: TODO: Consider mentioning this handles the full multi-leaf state machine (empty ↔ single ↔ multi) for clarity in the top-level export documentation.

huitseeker (approved):
Thanks for the comprehensive exploration and next steps.
Merging and iterating makes sense to me here!

plafer:
hey @gswirski can you fix the merge conflicts? Then we're good to merge

gswirski:
@plafer done

PR #2097, opened 2025-08-26
Implement `adv.insert_hqword` instruction

Fumuran:
This PR adds a new `adv.insert_hqword` instruction (decorator) which hashes top 4 words on the stack and stores them in the advice map using the resulting hash as a key. This procedure will be extensively used in the `miden-base` for the authentication procedures. 

Closes: #2032.

PR #2249, opened 2025-10-02
feat: implement `log_precompile` opcode

adr1anh:
This PR adds a new `log_precompile` instruction with a new `OP_LOGPRECOMPILE` opcode.

On the VM side, we now include the "precompile transcript state" in the VM and implement instructions to update it. The transcript state at the end of the execution should be the same as the one computed when the verifier recomputes the transcript from the raw requests.

Trace generation and constraints were updated to include the new bus operations 
- remove/insert precompile state into vtable column
- hasher request for the state permutation
- boundary constraints now take into account the initial and final transcript state, the latter which is now a public input for the proof.

We changed the `PrecompileVerificationState` to a `PrecompileTranscript` which more accurately describes how the "logging" is implemented. The transcript can record new precompile commitments, and be finalized to extract an actual digest. The usage of transcript/state is now preferred over sponge/capacity since the latter is an implementation detail. 

Documentation was enhanced, most notably by including a "precompile flow" section, along with explanations of the constraints and stack transition.

- Updated `CHANGELOG.md'


docs/src/design/stack/crypto_ops.md
```diff
@@ -296,4 +296,99 @@ Then, using the above value, we can describe the constraint for the chiplets' bu
 
 >$$
 b_{chip}' \cdot v_{ace} = b_{chip} \text{ | degree} = 2
-$$
\ No newline at end of file
+$$
+
+## LOG_PRECOMPILE
```
adr1anh, line 306: TODO: Link from https://github.com/0xMiden/miden-vm/blob/7adce6181c32c0bc8c79b53623a3493ae88b72c8/core/src/precompile.rs#L172

air/src/lib.rs
```diff
@@ -321,10 +322,20 @@ pub struct PublicInputs {
 }
 
 impl PublicInputs {
+    /// Creates a new instance of `PublicInputs` from program information, stack inputs and outputs,
+    /// and the precompile sponge state.
+    ///
+    /// # Note on Precompile Sponge
+    ///
+    /// The `precompile_sponge` parameter is currently ignored because the Winterfell backend does
+    /// not yet support verifying it as a public input. However, this parameter is included to make
```
adr1anh: I wrote this because it seems like we're not handling the kernel procedure digests either here.

air/src/trace/mod.rs
```diff
@@ -44,6 +44,9 @@ pub const STACK_TRACE_OFFSET: usize = DECODER_TRACE_RANGE.end;
 pub const STACK_TRACE_WIDTH: usize = 19;
 pub const STACK_TRACE_RANGE: Range<usize> = range(STACK_TRACE_OFFSET, STACK_TRACE_WIDTH);
 
+/// Label for log_precompile transcript state messages on the virtual table bus.
+pub const LOG_PRECOMPILE_LABEL: u8 = 200;
```
adr1anh: I need to double check, but we should be able to use the `LOG_PRECOMPILE` opcode here? The issue is that the messages are shared with all chiplet messages which use the chiplet selector labels.

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -200,6 +200,63 @@ pub(super) fn op_horner_eval_base<P: Processor>(
...
+    tracer: &mut impl Tracer,
+) -> [Felt; NUM_USER_OP_HELPERS] {
+    // Read TAG and COMM from stack
+    let comm = processor.stack().get_word(0);
+    let tag = processor.stack().get_word(4);
+
+    // Get the current precompile sponge capacity
+    let cap_prev = processor.precompile_transcript_state();
+
+    // Build the full 12-element hasher state for RPO permutation
```
huitseeker: The comment mentions writing [CAP_NEXT, R0, R1] but the actual order based on the iteration is [R1, R0, CAP_NEXT] due to the reverse indexing. The comment should accurately reflect the actual stack layout.

processor/src/chiplets/aux_trace/virtual_table.rs
```diff
@@ -175,3 +243,77 @@ where
...
+        [1, 2, 3, 4].map(|idx| main_trace.helper_register(idx, row)).into();
+
+    let message = LogPrecompileMessage { state };
+    let value = message.value(alphas);
+
+    #[cfg(any(test, feature = "bus-debugger"))]
+    _debugger.add_request(alloc::boxed::Box::new(message), alphas);
+
+    value
+}
```
huitseeker: The hardcoded indices [11, 10, 9, 8] should be replaced with named constants or derived from a base offset to improve maintainability.

processor/src/chiplets/aux_trace/bus/hasher.rs
```diff
@@ -175,6 +176,74 @@ pub(super) fn build_hperm_request<E: FieldElement<BaseField = Felt>>(
...
+    _debugger: &mut BusDebugger<E>,
+) -> E {
+    // Read helper registers
+    let addr = main_trace.helper_register(0, row);
+
+    // Input state [CAP_PREV, TAG, COMM]
+    // Helper registers store capacity in sequential order [e0, e1, e2, e3]
+    let cap_prev: Word = [1, 2, 3, 4].map(|idx| main_trace.helper_register(idx, row)).into();
+    // Stack stores words in big-endian order: stack[0..3] = [e3, e2, e1, e0]
+    // Therefore we read in reverse order to reconstruct the word correctly
```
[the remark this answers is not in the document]
huitseeker: Even Copilot is confused about #2059 😆

processor/src/chiplets/aux_trace/virtual_table.rs
huitseeker: The hardcoded indices `[11, 10, 9, 8]` should be replaced with named constants or derived from a base offset to improve maintainability.

For example:
```rust
const CAPACITY_BASE_OFFSET: usize = 8;
const WORD_LEN: usize = 4;

let capacity_indices: [usize; WORD_LEN] = [
    CAPACITY_BASE_OFFSET + WORD_LEN - 1,
    CAPACITY_BASE_OFFSET + WORD_LEN - 2,
    CAPACITY_BASE_OFFSET + WORD_LEN - 3,
    CAPACITY_BASE_OFFSET + WORD_LEN - 4,
];

let capacity: Word = capacity_indices
    .map(|idx| main_trace.stack_element(idx, row + 1)).into();
```

This would make it clearer that we're accessing the CAP_NEXT word at stack indices 8-11 in reverse order.

huitseeker (approved):
This LGTM, but:

- You'll probably want to share in the PR why we may avoid an integration test (à la #2097) here.
- I still think most of the comments by Copilot on avoiding magic numbers are warranted: it's more useful to document the registry layout by naming the start offset at which you start reading a value (e.g. `CAP_PREV_OFFSET=1`), than to read `[1..4)`.
- I'd love to see an issue mentioned in the TODOs re: updating e.g. the transcript state once the recursive verifier supports it (and other similar comments in the code). Searching for mentions of that issue in the code would then serve as a checklist.

air/src/lib.rs
```diff
@@ -339,6 +353,9 @@ impl miden_core::ToElements<Felt> for PublicInputs {
         let mut result = self.stack_inputs.to_vec();
         result.append(&mut self.stack_outputs.to_vec());
         result.append(&mut self.program_info.to_elements());
+        // TODO: Append transcript state to PI elements after recursion supports it.
+        // let cap: [Felt; 4] = self.precompile_transcript_state.into();
+        // result.extend_from_slice(&cap);
```
Al-Kindi-0: Q: Should we probably just uncomment these and deactivate the recursive verifier test, Phil is also probably doing the same for his PR which changes the number of committed columns

air/src/lib.rs
```diff
@@ -351,6 +368,8 @@ impl Serializable for PublicInputs {
         self.program_info.write_into(target);
         self.stack_inputs.write_into(target);
         self.stack_outputs.write_into(target);
+        // TODO: Serialize transcript state after recursion supports it.
+        // self.precompile_transcript_state.write_into(target);
```
Al-Kindi-0: here also and everywhere related

core/src/precompile.rs
```diff
@@ -263,57 +291,72 @@ pub trait PrecompileVerifier: Send + Sync {
...
+/// - **Finalization**: Squeeze with zero rate to extract a transcript digest (the sequential
+///   commitment)
 ///
-/// - **`new()`**: Initialize capacity to `ZERO`
-/// - **`absorb(comm)`**: Apply RPO256 permutation to `[capacity, tag, commitment]` and update
-///   capacity to the first word of the result
-/// - **`finalize()`**: Apply RPO256 permutation to `[capacity, ZERO, ZERO]` and extract the second
-///   word as the final digest
+/// # Implementation Note
+/// We store only the 4-element capacity portion between absorptions since the rate is ephemeral.
```
Al-Kindi-0: nit: maybe better "... since the rate is always overwritten when absorbing blocks that are a multiple of the rate width" or something similar

processor/src/processor/mod.rs
```diff
@@ -42,6 +45,16 @@ pub trait Processor: Sized {
     /// Returns a mutable reference to the internal hasher subsystem.
     fn hasher(&mut self) -> &mut Self::Hasher;
 
+    /// Returns the current precompile transcript state (sponge capacity).
+    ///
+    /// Used by `log_precompile` to thread the transcript across invocations.
+    fn pc_transcript_state(&self) -> PrecompileTranscriptState;
```
plafer:
```suggestion
    fn precompile_transcript_state(&self) -> PrecompileTranscriptState;
```
Let's aim for descriptive names over "C-style" shortened names

processor/src/processor/mod.rs
```diff
@@ -42,6 +45,16 @@ pub trait Processor: Sized {
...
+
+    /// Sets the precompile transcript state (sponge capacity) to a new value.
+    ///
+    /// Called by `log_precompile` after recording a new commitment.
+    fn set_pc_transcript_state(&mut self, state: PrecompileTranscriptState);
```
plafer:
```suggestion
    fn set_precompile_transcript_state(&mut self, state: PrecompileTranscriptState);
```

processor/src/operations/crypto_ops.rs
```diff
@@ -186,6 +190,58 @@ impl Process {
...
+            output_state[STATE_CAP_RANGE.start + 1],
+            output_state[STATE_CAP_RANGE.start + 2],
+            output_state[STATE_CAP_RANGE.start + 3],
+        ]);
+        self.pc_transcript_state = cap_next;
+
+        // Write the output to stack (top 12 elements) in reverse order: [R1, R0, CAP_NEXT].
+        for i in 0..STATE_WIDTH {
+            self.stack.set(i, output_state[STATE_WIDTH - 1 - i]);
+        }
```
plafer, line 241: This loop makes it hard to understand the ordering of the words on the stack (a la #2059). The commitment and tag are retrieved using `get_word()` - can't we use 3 calls to `set_word()` here? This would make it much clearer what goes where, as well as the ordering.

stdlib/docs/crypto/hashes/keccak256.md
```diff
@@ -5,4 +5,4 @@
...
+| hash_memory_impl | Internal implementation of memory-based Keccak256 computation via deferred computation.<br /><br />Emits an event to populate the advice stack with digest and store the preimage for deferred verification.<br />Returns the commitment to the precompile call and its tag.<br /><br />Input: [ptr, len_bytes, ...]<br />Output: [COMM, TAG, DIGEST_U32[8], ...]<br /><br />Where:<br />- ptr: word-aligned memory address containing INPUT_U32[len_u32] where len_u32=⌈len_bytes/4⌉<br />- len_bytes: number of bytes to hash<br />- COMM: the calldata commitment `Rpo256(Rpo256(INPUT_U32[..]) \|\| Rpo256(DIGEST_U32[..]))` consumed by `log_precompile`<br />- TAG = [KECCAK_HASH_MEMORY_EVENT, len_bytes, 0, 0] encodes the precompile identifier and the byte length as metadata<br />- DIGEST_U32[8] = [d_0, ..., d_7] = Keccak256(INPUT_U8[len_bytes])<br /> |
```
plafer, line 8: Was this ever meant to be public? Or should it have been addressed with #2309?

air/src/lib.rs
```diff
@@ -144,6 +147,7 @@ impl Air for ProcessorAir {
             constraint_ranges,
             program_digest: pub_inputs.program_info.program_hash().to_owned(),
             kernel_digests: pub_inputs.program_info.kernel_procedures().to_owned(),
+            pc_transcript_state: pub_inputs.pc_transcript_state,
```
plafer, line 150:
```suggestion
            precompile_transcript_state: pub_inputs.precompile_transcript_state,
```

docs/src/design/stack/crypto_ops.md
```diff
@@ -301,4 +301,120 @@ Then, using the above value, we can describe the constraint for the chiplets' bu
...
+
+The `log_precompile` operation logs a precompile event by recording two user-provided words (`TAG` and `COMM`) into the precompile transcript (implemented via an RPO sponge). Initialization and boundary enforcement are handled via variable‑length public inputs; see [Precompile flow](./precompiles.md) for a high‑level overview. This section concentrates on the stack interaction and bus messages.
+
+### Operation Overview
+
+The stack is expected to be arranged as `[COMM, TAG, PAD, ...]`, where
+- `COMM` is the precompile commitment to the calldata (precompile-specific),
+- `TAG` is a word identifying the precompile, which may contain request-specific metadata,
+- `PAD` is a word that will get overwritten in the next cycle.
```
plafer: My current model is that a precompile is a triplet (inputs, function, outputs). `COMM` encodes the inputs, and `TAG` encodes the function (i.e. an event ID), as well as the outputs.

Is this correct? If so I would reword this part to provide this intuition, and be more specific rather than general (e.g. using "event ID" instead of "identifying the precompile" - since AFAIU event IDs is how we identify precompiles in our system).

EDIT: I just noticed this is more or less explained like this in `precompiles.md` - instead of re-explaining here, I would refer to `precompiles.md` for a thorough/centralized explanation of what each of these are.

air/src/trace/mod.rs
```diff
@@ -42,6 +42,55 @@ pub const STACK_TRACE_OFFSET: usize = DECODER_TRACE_RANGE.end;
...
+    pub const STACK_TAG_RANGE: Range<usize> = STACK_R0_RANGE;
+
+    // HASHER STATE LAYOUT
+    // --------------------------------------------------------------------------------------------
+    // The hasher permutation uses a 12-element state. For `log_precompile` the state is interpreted
+    // differently for the input (`[CAP_PREV, TAG, COMM]`) and output (`[CAP_NEXT, R0, R1]`) words.
+
+    pub const STATE_CAP_RANGE: Range<usize> = range(0, CAPACITY_LEN);
+    pub const STATE_R0_RANGE: Range<usize> = range(STATE_CAP_RANGE.end, DIGEST_LEN);
+    pub const STATE_R1_RANGE: Range<usize> = range(STATE_R0_RANGE.end, DIGEST_LEN);
```
plafer:
```suggestion
    pub const STATE_RATE_0_RANGE: Range<usize> = range(STATE_CAP_RANGE.end, DIGEST_LEN);
    pub const STATE_RATE_1_RANGE: Range<usize> = range(STATE_R0_RANGE.end, DIGEST_LEN);
```
I didn't know/recall what `R0` and `R1` meant for most of the time I reviewed this

docs/src/design/stack/precompiles.md
```diff
@@ -0,0 +1,79 @@
...
+3. **Wrapper constructs commitment** – The wrapper pops result(s) from advice, computes `(TAG, COMM)` per the precompile’s convention, and prepares to log the operation.
+4. **`log_precompile` records the commitment** – The wrapper invokes `log_precompile` with `[COMM, TAG, PAD, ...]`. The instruction:
+   - Reads the previous transcript capacity `CAP_PREV` (non‑deterministically via helper registers).
+   - Applies the RPO permutation to `[CAP_PREV, TAG, COMM]`, producing `[CAP_NEXT, R0, R1]`.
+   - Writes `[R1, R0, CAP_NEXT]` back onto the stack; programs typically drop these words immediately.
+5. **Capacity tracking via vtable** – Capacity is tracked inside the VM via the chiplets’ virtual table; the host never tracks capacity. The table always stores the current capacity (the transcript state). On each `log_precompile`:
+   - The previous capacity is removed from the table.
+   - The permutation links `CAP_PREV --[TAG,COMM]--> CAP_NEXT`.
+   - The next capacity is inserted back into the table.
+   This enforces that updates can only occur by applying the permutation.
```
plafer, line 28: I'm still trying to convince myself that the "sibling vtable" + hasher chiplet properly enforce this relationship as currently implemented - will sleep on it and update this comment tomorrow.

stdlib/docs/crypto/hashes/keccak256.md
adr1anh: In this case, we want to be able to test the wrapper _before_ calling `log_precompile` since that clears the precompile commitment on the stack. @bitwalker mentioned we should be able to add annotations later to mark it as hidden publicly.

processor/src/operations/crypto_ops.rs
adr1anh: There's no method `set_word` unfortunately. I think the issue is about documentation. We always represent a hasher state on the stack in big-endian order (reversed). I added a comment explaining why we do this.,

docs/src/design/stack/crypto_ops.md
adr1anh: The doc is already linked in the section header. Should it be more explicit here?

air/src/lib.rs
adr1anh: I opted to use `pc_*` for fields as things got very verbose, and the docstrings explain the abbreviation.

processor/src/operations/crypto_ops.rs
plafer: Yes there is - [here](https://github.com/0xMiden/miden-vm/blob/b0254acaefb50ce64ba8502a10f533759bfcfa94/processor/src/processor/mod.rs#L138)

docs/src/design/stack/crypto_ops.md
plafer: Yeah - or at least when I read "a high-level overview", I didn't expect that level of detail to be there. I guess my point is: I find it better explained in `precompiles.md`, and since there was also an explanation of each here, I didn't look for another one. IMO saying something like:

> The stack is expected to be arranged as `[COMM, TAG, PAD, ...]`; see precompiles.md for a detailed explanation of each field.

would be better, as the reader would go read the best explanation (+ the convention detailed there), and there would only be 1 place to make updates in the future.

processor/src/operations/crypto_ops.rs
adr1anh: `Processor` doesn't seem to be implemented for `Process` though, unless I'm missing something

plafer: Ah sorry I put the comment on the legacy `Process` implementation, but I really meant to put it [for the `Processor` implementation](https://github.com/0xMiden/miden-vm/blob/476f24b73a664bbe580c5b7a74da51d063e23224/processor/src/processor/operations/crypto_ops.rs#L252-L256). We don't need to do anything about the `Process`'s implementation, indeed!

core/src/precompile.rs
```diff
@@ -134,19 +150,38 @@ impl Deserializable for PrecompileRequest {
...
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct PrecompileCommitment {
-    /// Tag containing metadata including the event ID in the first element. The remaining 3
-    /// elements can be used for precompile-defined associated data.
-    pub tag: Word,
-    /// Commitment word representing the inputs and result of the request.
-    pub commitment: Word,
+    tag: Word,
+    comm: Word,
```
plafer, line 156: In light of our offline discussion, I'd argue for removing the TAG/COMM concepts, in favor of modeling the precompile commitment as "event ID + 7 Felts", where the precompile is free to encode precompile parameters, inputs and outputs however they wish in those 7 Felts.

I find that to be much clearer than the current system. For example, after our offline discussion, I thought I understood how TAG and COMM are sufficiently general concepts, in that:
- TAG contains the event ID, and 3 Felts for precompile parameters
- COMM is a commitment to the inputs and outputs

However, when re-reading `precompile.md`, in the "signature scheme" example, the `is_valid` "output" of the signature is placed in the TAG. This provides strong evidence that the TAG and COMM abstractions are more distracting than helpful, and we should instead leave it as general (and simple) as possible.

For example, I would represent this type as

```rs
struct PrecompileCommitment {
  event_id: Felt,
  blob: [Felt; 7]
}
```

and then provide a few constructors that suggest some common/useful layouts for the commitment:

```rs
impl PrecompileCommitment {
    pub fn new_parameterized(
        event_id: Felt, 
        precompile_params: [Felt; 3],
        input_outputs_commitment: Word) -> Self {
        ...
    }

    pub fn new_simple_output(
        event_id: Felt,
        output: Felt,
        precompile_params: [Felt; 2],
        input_commitment: Word) -> Self {
        ...
    }

    /// catch-all, make-your-own layout
    pub fn new_raw(
       event_id: Felt,
       blob: [Felt; 7]
    ) -> Self {
       ...
    }
}
```

If we agree on this direction, this could be done in a follow-up PR.

docs/src/design/stack/precompiles.md
plafer, line 28: On top of this explanation, I would add a section `## Soundness Argument` that explains how the hasher chiplet and the virtual table come together to ensure that the transcript is properly built. I see the outline being roughly

- High level
  - The VM builds a transcript (RPO256 sponge) which encodes all precompile calls $f_{\theta}(x) = y$ (one per call to `log_precompile`)
  - The final transcript digest is included in the public inputs
  - The verifier is given all the necessary information to rebuild the transcript; that is, the list of `(precompile_id, params, inputs, outputs)`
- Ensuring that the VM builds the transcript correctly
  - Each `log_precompile` operation includes in the trace the previous transcript capacity `cap_prev` (nondeterministically), all the information for the precompile call (`(precompile_id, params, inputs, outputs)`), and the resulting new capacity `cap_new`
  - a chiplets bus call to the hasher chiplet is made to ensure that `RPO256(cap_prev, new_precompile_call) = cap_new`
  - Finally, we need to somehow link the `cap_new` of the last `log_precompile` call to a boundary constraint that checks that it's consistent to what's in the public inputs.
    - We do that with a new virtual table, where for each `log_precompile` operation, we remove from the table `cap_prev`, and add `cap_new`.
    - The table is initialized with `initial_cap` inserted, and `final_capacity` removed.
    - The virtual table will be empty at the end of the program if and only if the prover properly set the `cap_prev` field in all `log_precompile` calls

plafer (approved):
Looks great!

IMO the main future improvement is to get rid of the TAG/COMM concepts, as outlined in the first comment. If we agree, then we should open an issue and implement it in a subsequent PR.

core/src/precompile.rs
Al-Kindi-0: If the only motivation for collapsing the data commitment (COMM) and the precompile selector (TAG) is that the current signature verifier exposes an is_valid bit, I’d rather revisit that example than the interface. We can model is_valid as a call parameter: when the flag is set we run the “verify” branch, when it’s clear we run the complementary branch. It’s still part of the precompile’s control plane, not a return value.

  I’m on board with renaming COMM/TAG—the terms are clunky—but the split itself keeps the code paths clean by separating payload from behavior. Unless we have a stronger use case that forces us to blend them, I’d keep the two-word structure and just improve the terminology.

processor/src/operations/crypto_ops.rs
adr1anh: I've updated the processor, it does seem more clear there.

core/src/precompile.rs
adr1anh: I really agree that the names are not really helpful and have led to confusion in the review. Something along the lines of `(instruction, immediates[3], payload_comm[4])` seems more descriptive already, but other suggestions are welcome. 

The structure however should probably stay the same (modulo splitting the current tag into `instruction, meta[3]`). The (possibly premature) motivation here is that imposing a stricter structure could help simplify the design of the "decoder" in the precompile VM. This is still a hunch for now, and if it turns out to be irrelevant, it will be easier to loosen the requirements later down the line than to constrain them afterwards.

docs/src/design/stack/precompiles.md
adr1anh: I've linked this comment to the ongoing issue #2269 but we can also create a sub-issue for it.

adr1anh:
I've addressed the last remaining comments, and added to the #2269 issue.

core/src/precompile.rs
plafer: Since we don't all agree on the best way forward, let's leave it as is for now - and probably open an issue to track this? We can revisit after we've designed the precompile VM.

IMO we're all good to merge!

processor/src/operations/crypto_ops.rs
plafer: I still don't see the change though - was that commit pushed yet?

adr1anh: Oops, this was pushed to the ECDSA branch accidentally.

core/src/precompile.rs
bobbinth: Not directly related to the discussion above, but I think using just one element to identify the precompile is probably not sufficient. I'd feel much more comfortable if we used 2 elements as a "precompile ID". This is not something we need to do immediately, but let's open an issue to discuss this (unless one exists already).

PR #2253, opened 2025-10-02
refactor: use &str instead of &String in ProgramHash::read

maradini77:
## Summary
Replace `&String` parameter with `&str` in `ProgramHash::read()` for better API ergonomics.

- **Change:** Parameter type `hash_hex_string: &String` → `hash_hex_string: &str`


bobbinth:

Issue #2227, opened 2025-09-28
Automate release process


bobbinth:
Currently, our release process is "semi-automated". That is, when a new version of the VM is to be released, I do the following:

1. Check if we need to make any changes to the `next` branch. Usually, there isn't much and I just run `cargo update` to refresh the `Cargo.lock` file. But sometimes I discover that the repo is not in a publishable state and need to make small updates to fix that.
2. Merge `next` into `main`.
3. Run `publish-release.sh` script.
4. Create a Github release.

While this process is simple enough, it would be good to make it a bit more automated. This is what we've been doing in `miden-base` and `miden-node` repos. There, we are using `release-plz`, but I think with Rust 1.90 we should be able to get away with just `cargo` (via the `cargo publish --workspace` command).

Mimicking how things work in `miden-node` and `miden-base`, I think we can do the following:

 - On merging to `next` or `main` we dry-run the publishing (similar to how it works in [miden-base](https://github.com/0xMiden/miden-base/blob/next/.github/workflows/release-plz-dry-run.yml)).
 - On creating a Github release, we publish the crates to crates.io (also similar to how it works in [miden-base](https://github.com/0xMiden/miden-base/blob/next/.github/workflows/release-plz.yml)).

But I'm open to other suggestions too.

bitwalker:
I've been meaning to look into switching the compiler to just using our own process on top of `cargo publish --workspace`, so 👍 on something in that vein.

The two other things we use `release-plz` for (in the compiler):

1. Generation of CHANGELOG entries for each crate. This doesn't hold a lot of value IMO, but it's a bit better than no CHANGELOG, and it avoids us having to do it by hand. I think `release-plz` just uses some other tool for this though, so we could probably use that directly.
2. Bumping the version based on the current version and the types of changes made. I believe this also builds on top of other tools. This one is more valuable, and ensures we're following semantic versioning correctly. Some form of this would be needed for fully automating releases though.

bobbinth:
I tried to publish crates using `cargo publish --workspace`, but ran into some Cargo bug (it just crushes out). Manual publishing via the publish script works fine though.

bitwalker:
Out of curiosity, what was the bug you hit?

bobbinth:
Unfortunately I didn't record it because I was getting the same panic when running with `--dry-run` and so I thought we'd be able to get later as well. But now if I run `cargo publish --workspace --dry-run` on `next` - everything seems to work fine.

PR #2254, opened 2025-10-03
feat: systematize u32 Vector Indexes

huitseeker:
## This PR

Our workspace is littered with the pattern of using a [new type](https://doc.rust-lang.org/book/ch20-03-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction) to index into a vector, where that new type wraps a u32 (instead of the native usize used by Rust to index those vectors). 

This approach sometimes requires a bit of care to make sure:
- the bounds checks are implemented properly against `u32::MAX`, and that 
- the `Vec` cannot be accidentally indexed into using anything else than its dedicated new type.

As we seem to be using this pattern in *numerous* places (to my personal interest: `MastNodeId`, `DecoratorId`, but also addressed in this PR: `RowIndex` and `SourceId`), we centralize the implementation, [as it is in the rust compiler's source](https://doc.rust-lang.org/beta/nightly-rustc/rustc_index/vec/struct.IndexVec.html), into an `IndexedVec` which lays out the correct behavior.

While we're here, we also move the generic `DenseIdMap` used for dense remappings (where we know we're creating a map from a full specific range) to that indexing crate, and slightly extend its use in the `MastForestBuilder`, which slightly improves benchmarks below.

**Note for reviewers:**

The more meaty part of this PR is in `mast/src/merge/mod.rs`. Hence the prior test PRs (e.g. #2200).

### mast_forest_merge

|           | `next`                    | `indexed_vec`                     |
|:----------|:--------------------------|:--------------------------------- |
| **`8`**   | `307.76 us` (✅ **1.00x**) | `294.91 us` (✅ **1.04x faster**)  |
| **`16`**  | `578.54 us` (✅ **1.00x**) | `560.11 us` (✅ **1.03x faster**)  |
| **`32`**  | `1.23 ms` (✅ **1.00x**)   | `1.19 ms` (✅ **1.03x faster**)    |
| **`64`**  | `2.26 ms` (✅ **1.00x**)   | `2.20 ms` (✅ **1.03x faster**)    |
| **`128`** | `4.73 ms` (✅ **1.00x**)   | `4.61 ms` (✅ **1.03x faster**)    |
| **`256`** | `9.37 ms` (✅ **1.00x**)   | `9.19 ms` (✅ **1.02x faster**)    |

## Next steps

We'll open an issue for the remaining indexes I didn't have the time to tackle in this PR: ~~`ByteIndex`~~, `MemoryAccess` (maybe?). We could also extend this to the `u16`-based indexes : `ModuleIndex` and `ProcedureIndex`.


core/src/mast/merger/mod.rs
```diff
@@ -242,7 +245,7 @@ impl MastForestMerger {
                 // which has descendants (Call, Loop, Split, ...), then their descendants need to be
                 // in the indices.
                 self.node_id_by_hash.insert(node_fingerprint, new_node_id);
-                self.hash_by_node_id.insert(new_node_id, node_fingerprint);
+                let _ = self.hash_by_node_id.push(node_fingerprint);
```
plafer: This ignores the possible error occurring in `push()` - we should instead propagate. I think this applies everywhere else we do `let _ = ...`.

Also, IIUC, we expect `_ == new_node_id`, right (where `_` is the return value of `push()`)? In that we assume that `MastForest::add_node()` returns contiguous node IDs. I would add a `debug_assert!()` here to make sure we catch it if ever that changes.

crates/utils/indexing/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+/// efficient dense ID remapping.
+#[derive(Clone)]
+pub struct DenseIdMap<From: Idx, To: Idx> {
+    inner: IndexVec<From, Option<To>>,
+}
+
+impl<From: Idx, To: Idx> DenseIdMap<From, To> {
+    /// Create a new dense ID mapping with the specified capacity.
+    #[inline]
+    pub fn with_capacity(capacity: usize) -> Self {
```
plafer: nit: since the `DenseIdMap` docs mention that this is equivalent to a `IndexVec`, my mental model for what `with_capacity()` is basically the `Vec`'s capacity. But this initializes the inner vector, so I would call this something like `new(size: usize)` or `with_len(size: usize)`.

crates/utils/indexing/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+            panic!("source ID {} exceeds DenseIdMap capacity {}", idx, self.inner.len());
+        }
+        self.inner.insert_at(k, Some(v));
+    }
+
+    /// Get the target ID for the given source ID.
+    #[inline]
+    pub fn get(&self, k: From) -> Option<To> {
+        *self.inner.get(k)?
+    }
```
plafer, line 226: Should we panic if `From` is out of bounds (either `assert` or `debug_assert`)? Since that would indicate something is pretty wrong.

crates/utils/indexing/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Panics if the source ID is beyond the capacity of this DenseIdMap.
+    /// This DenseIdMap should be created with sufficient capacity to accommodate
+    /// all expected source IDs.
+    #[inline]
+    pub fn insert(&mut self, k: From, v: To) {
+        let idx = k.to_usize();
+        if idx >= self.inner.len() {
+            panic!("source ID {} exceeds DenseIdMap capacity {}", idx, self.inner.len());
+        }
+        self.inner.insert_at(k, Some(v));
```
plafer, line 219:
```suggestion
        let idx = k.to_usize();
        let len = self.len();

        assert!(idx < len, "source ID {idx} exceeds DenseIdMap capacity {len}",);
        self.inner.insert_at(k, Some(v));
```
nit: would replace if/panic with assert

bitwalker:
Just a note: `ByteIndex` is _not_ an instance of this pattern (nor `(Column|Line)Index`).

core/src/mast/merger/mod.rs
huitseeker: Yep!

crates/utils/indexing/src/lib.rs
huitseeker: It would, but I'm loathe to introduce panics where not absolutely necessary. The irrecoverability also makes testing the edge cases a bit harder (e.g. `mast_forest_merge_invalid_decorator_index`). If you insist, I'd add a `debug_assert` like in the preceding `insert`.

plafer: > It would, but I'm loathe to introduce panics where not absolutely necessary.

What consists of "absolutely necessary" is up for debate - one we should have. There's an argument to be made for asserting all preconditions, so that we can catch when those things happen and actually fix them, instead of silently failing. This in turn forces the caller to ensure that proper sanitization of input is made.

This is a broader discussion that we should have - for this one, I'm okay with either leaving as-is or a `debug_assert`, up to you.

core/src/mast/merger/mod.rs
```diff
@@ -271,95 +281,103 @@ impl MastForestMerger {
...
+    /// Returns the ID of the node in the merged forest that matches the given
+    /// fingerprint, if any.
+    fn lookup_node_by_fingerprint(&self, fingerprint: &MastNodeFingerprint) -> Option<MastNodeId> {
+        self.node_id_by_hash.get(fingerprint).copied()
+    }
+
+    /// Builds a new node with remapped children and decorators using the provided mappings.
+    fn build_node_with_remapped_children(
+        &self,
+        _forest_idx: usize,
```
plafer: nit: can get rid of this argument

core/src/mast/mod.rs
```diff
@@ -69,12 +70,31 @@ pub struct MastForest {
...
+        Self {
+            nodes: IndexVec::new(),
+            roots: Vec::new(),
+            decorators: IndexVec::new(),
+            advice_map: AdviceMap::default(),
+            error_codes: BTreeMap::new(),
+        }
+    }
+}
+
```
plafer: Can't we derive this?

crates/utils/indexing/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    /// Check if the mapping is empty.
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+}
+
+/// A trait for looking up fingerprints by node ID with generic fingerprint type.
+pub trait FingerPrintLookup<ID, F>
```
plafer: The name suggests that it specializes to MAST node IDs and fingerprints, but the definition is one of a generic mapping where the key is an `Idx`.

Renaming it to something like `Map`, `MapByIdx`, `Lookup`, `LookupByIdx` or something of the sort, would remove the confusion for me. Basically `Map<MastNodeId, MastNodeFingerprint>` is a "fingerprint lookup", but `Map<SomeOtherIdxType, SomethingElse>` is not.

processor/src/stack/trace.rs
plafer, line 1: For future PRs: I wouldn't bother modifying the deprecated `Process`. It adds to the diff of things to review, and we're getting rid of it anyways.

core/src/mast/merger/mod.rs
```diff
@@ -271,95 +281,101 @@ impl MastForestMerger {
...
+    // ================================================================================================
+
+    /// Remaps a child node ID using the node ID map, returning the original ID if not found.
+    fn remap_child(
+        &self,
+        child_id: MastNodeId,
+        nmap: &DenseIdMap<MastNodeId, MastNodeId>,
+    ) -> MastNodeId {
+        nmap.get(child_id).expect("every node id should have an entry")
+    }
```
adr1anh, line 301: Should this be a free function? I’m not sure I quite understand the function though, since I'd associate "remap" with a mutation.

core/src/mast/mod.rs
```diff
@@ -48,18 +49,18 @@ mod tests;
 ///
 /// A [`MastForest`] does not have an entrypoint, and hence is not executable. A [`crate::Program`]
 /// can be built from a [`MastForest`] to specify an entrypoint.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
```
adr1anh, line 51: Is this a local lint you have? If so it would be nice if this could be done with `rustfmt`

adr1anh (approved):
LGTM

core/src/mast/merger/mod.rs
huitseeker: The existence of this private function is to make the point that the map changes the node IDs involved in, precisely, the mutation described a little below in `build_node_with_remapped_children`.

core/src/mast/mod.rs
huitseeker: Just a mistake in this case.

crates/utils/indexing/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "miden-utils-indexing"
```
bobbinth, line 2: Since we added a new crate (and size we don't yet have https://github.com/0xMiden/miden-vm/issues/2227), we should also update the publish [script](https://github.com/0xMiden/miden-vm/blob/next/scripts/publish-release.sh).

bitwalker: @bobbinth Seems to me that we can get rid of that script and just do `cargo publish --workspace` now

bobbinth: Do you mean even before addressing #2227? I think this could work be we'd probably need to make some minor updates (e.g., update `Cargo.toml` for `miden-test-utils` so that we don't try to publish it).

Also, I just noticed that running `cargo publish --workspace --dry-run` currently returns an error because we don't have a README file for the `miden-utils-indexing` create. @huitseeker - could you look into it?

PR #2262, opened 2025-10-04
refactor(processor): remove unused `is_first_child` field from BlockInfo

prestoalvarez:
## Summary

Removes the unused `is_first_child` field from the `BlockInfo` struct in the block stack decoder.

## Changes

- **Removed** unused `is_first_child: bool` field from `BlockInfo` struct
- **Removed** corresponding `#[allow(dead_code)]` attribute and TODO comment  
- **Updated** `BlockStack::push()` method to no longer compute or assign the unused field

## Rationale

The `is_first_child` field was marked with `#[allow(dead_code)]` and a TODO comment indicating it should be removed. Analysis shows:

- No methods exist to access this field (unlike other fields that have corresponding getter methods)
- The field is never read anywhere in the codebase
- Similar functionality exists in `BlockHashTableRow` struct, but that's a separate concern

### DIFF
--- a/processor/src/decoder/block_stack.rs
+++ b/processor/src/decoder/block_stack.rs
@@ -39,23 +39,22 @@ impl BlockStack {
         }
 
         // determine additional info about the new block based on its parent
-        let (parent_addr, is_loop_body, is_first_child) = match self.blocks.last() {
+        let (parent_addr, is_loop_body) = match self.blocks.last() {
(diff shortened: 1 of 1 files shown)


bitwalker:
Sorry, we do not accept PRs this small without an associated issue created by us.

Additionally, this code is already being refactored, so clean up lie this will be handled as part of those efforts.

PR #2263, opened 2025-10-06
fix: avoid needless collects in mast merger tests

sky-coderay:
- replace temporary Vec allocations in mast merger tests with iterator `any` checks
- continue verifying the same fingerprints without allocating or collecting


bitwalker:
Sorry, we do not accept PRs this small without an associated issue created by us.

PR #2264, opened 2025-10-07
Parallel trace generation tweaks

huitseeker:
Some tweaks on #1839 / #2188 .


processor/src/fast/trace_state.rs
```diff
@@ -245,6 +245,47 @@ impl StackState {
...
     }
+
+    /// Shifts the stack left by one position (removing the top element) and starts a new context.
+    ///
+    /// This method performs two operations:
+    /// 1. Removes the top element of the stack by shifting all other elements up
+    /// 2. Starts a new execution context, hiding the overflow table
+    ///
+    /// Returns a tuple of (stack_depth_before_context, overflow_addr_before_context)
+    pub fn shift_left_and_start_context(
```
plafer, line 256: The reason I left it as a TODO was that I wanted to avoid copy/pasting the `decrement_size()` implementation in here. 

Looking into it more, I think a cleaner way to do it would be to implement `StackInterface` on `StackState` instead of `CoreTraceGenerator`, and then we'll be able to reuse the `decrement_size()` implementation, so that it becomes:

```rs
self.decrement_size();
self.start_context();
```

Then the `impl Processor for CoreTraceGenerator` will need to be adjusted as well.

processor/src/parallel/mod.rs
```diff
@@ -314,17 +314,16 @@ fn fixup_stack_and_system_rows(
     }
 
     // Initialize subsequent fragments with their corresponding rows from the previous fragment
-    // TODO(plafer): use zip
-    for (i, fragment) in fragments.iter_mut().enumerate().skip(1) {
-        if fragment.row_count() > 0 && i - 1 < stack_rows.len() && i - 1 < system_rows.len() {
+    for (fragment, (system_row, stack_row)) in
+        fragments.iter_mut().skip(1).zip(system_rows.iter().zip(stack_rows.iter()))
+    {
+        if fragment.row_count() > 0 {
```
plafer, line 317: I see that this was present before, but I think we should remove it as well for better communication of intent. That is, this suggests that empty fragments are possible, but actually they are not. And we should make that more clear (either through docs, checking that as a precondition, etc) - but checking  the row count in this loop IMO is not the right place for it.

processor/src/parallel/tests/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+    for col_idx in 0..miden_air::trace::PADDED_TRACE_WIDTH {
+        let col_from_fragments = trace_from_fragments.main_segment().get_column(col_idx);
+        let col_from_single_fragment =
+            trace_from_single_fragment.main_segment().get_column(col_idx);
+
+        // Since the parallel trace generator only generates core traces, its column length will
+        // be lower than the slow processor's trace in the case where the range checker or
+        // chiplets column length exceeds the core trace length. We also ignore the last element
+        // in the column, since it is a random value inserted at the end of trace generation,
+        // and will not match when the 2 traces don't have the same length.
```
plafer, line 121: This is no longer true since #2188; `build_trace()` builds the full trace. So we should check the entire columns (including the random value since they should match). I see that this comment was copy/pasted from `Test`, so we should update that too.

processor/src/parallel/tests/loop_tests.rs
```diff
@@ -0,0 +1,88 @@
...
+        host.load_library(create_simple_library()).unwrap();
+        let (execution_output, trace_fragment_contexts) =
+            processor.execute_for_trace_sync(&program, &mut host, fragment_size).unwrap();
+        build_trace(
+            execution_output,
+            trace_fragment_contexts,
+            program.hash(),
+            program.kernel().clone(),
+        )
+    });
```
plafer, line 62: We could have `test_fragment_boundary_trace_consistency()` return the trace, which we can use in `insta` rather than regenerating it (for both test performance, and reducing boilerplate).

Applies to all files

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -463,6 +557,73 @@ mod op_batch_tests {
...
+        let mut acc = OpBatchAccumulator::new();
+        acc.add_op(Operation::Noop);
+        let single_op_batch = acc.into_batch();
+
+        let results: Vec<_> = single_op_batch.iter_with_groups().collect();
+        assert_eq!(results.len(), 1);
+        assert_eq!(results[0], (0, 0, &Operation::Noop));
+    }
+
+    #[cfg(feature = "arbitrary")]
```
plafer, line 577: I missed this from previous PRs, but what was the reason behind the `arbitrary` feature? Don't we always want these tests? I also checked our `Makefile` (and hence what the CI runs), and I don't think we enable the `arbitrary` feature? Ideally I would like `cargo nextest run` to run all of our tests; I know I will forget to include `arbitrary` from time to time.

processor/src/fast/operation.rs
```diff
@@ -77,29 +76,13 @@ impl Processor for FastProcessor {
...
-        self.ace.add_circuit_evaluation(self.clk, circuit_evaluation.clone());
-        tracer.record_circuit_evaluation(self.clk, circuit_evaluation);
-
-        Ok(())
+    // Use the default op_eval_circuit implementation from the Processor trait
+    // and override handle_circuit_evaluation to store the result
+
+    /// Override handle_circuit_evaluation to store the result in ACE and tracer
+    fn handle_circuit_evaluation(&mut self, evaluation: CircuitEvaluation) {
+        self.ace.add_circuit_evaluation(self.clk, evaluation.clone());
```
plafer, line 84: This `clone()` is not necessary

processor/src/fast/operation.rs
```diff
@@ -77,29 +76,13 @@ impl Processor for FastProcessor {
...
-            &mut self.memory,
-            err_ctx,
-            tracer,
-        )?;
-        self.ace.add_circuit_evaluation(self.clk, circuit_evaluation.clone());
-        tracer.record_circuit_evaluation(self.clk, circuit_evaluation);
-
-        Ok(())
+    // Use the default op_eval_circuit implementation from the Processor trait
+    // and override handle_circuit_evaluation to store the result
```
plafer, line 80: I would remove this comment, and move the docstring from `op_eval_circuit()` to `Processor::op_eval_circuit()`. However the last part about not recording memory reads is outdated and should be removed.

processor/src/processor/mod.rs
```diff
@@ -47,12 +50,123 @@ pub trait Processor: Sized {
...
+        }
+
+        Ok(evaluation_context)
+    }
+
+    /// Record circuit evaluation for later use in ACE chiplet.
+    /// Different processors can implement this differently:
+    /// - FastProcessor: Store in AceReplay
+    /// - CoreTraceFragmentGenerator: Store in AceReplay or compute later
+    /// - Regular Processor: Use directly
```
plafer, line 164: Does "Regular Processor" refer to `Process`? If so, I think we should remove that reference, since it's going away, and isn't using this `Processor` architecture.

processor/src/processor/mod.rs
```diff
@@ -47,12 +50,123 @@ pub trait Processor: Sized {
...
+        let clk = self.system().clk();
+
+        let circuit_evaluation =
+            self.eval_circuit_shared(ctx, ptr, clk, num_read, num_eval, err_ctx, tracer)?;
+
+        // Record circuit evaluation for tracer
+        tracer.record_circuit_evaluation(self.system().clk(), circuit_evaluation.clone());
+
+        // Allow processors to handle the result differently
+        self.handle_circuit_evaluation(circuit_evaluation);
```
plafer, line 67: My plan for moving this to `Processor` was to add a `AceInterface` (similar to `HasherInterface` or `MemoryInterface`), and here we would call `self.ace.add_circuit_evaluation()`. Then different implementations can tune that behavior by how they implement the `AceInterface`.

We then also wouldn't need the `eval_circuit_shared()`; just the `op_eval_circuit()`, and if a `Processor` wants to do something different (e.g. read the `CircuitEvaluation` from a replay, or do nothing because it assumes the circuit was evaluated elsewhere already), then it only has `op_eval_circuit()` to override. What I don't like about the current state is there are 3 `Processor` methods related to `op_eval_circuit()`, so it's not as clean/clear exactly how to override the entire default behavior.

core/src/mast/node/basic_block_node/op_batch.rs
huitseeker: > I missed this from previous PRs, but what was the reason behind the arbitrary feature? Don't we always want these tests?

We do, that's a mistake - the goal of the arbitrary feature is to allow access to the `Arbitrary` impls in a cross-crate context.

> I also checked our Makefile (and hence what the CI runs), and I don't think we enable the arbitrary feature?

We do, as part of `testing`.


> Ideally I would like cargo nextest run to run all of our tests; I know I will forget to include arbitrary from time to time.

Can't promise anything for `cargo nextest run`, and the issue of controlling visibility of `Arbitrary` definitions while having access to it in tests is old and well-known in the Rust world. But `make test` will work well!

huitseeker:
@plafer I will re-open the PR for fixing the operation iteration, i.e. https://github.com/0xMiden/miden-vm/pull/2264/commits/0fb9cc1a9b50b29803dcfaa2c8a0b6007f84c616 — the rest is less important an you seem to have other ideas on how to approach them, as well as moving fast on #2188, which makes maintaining this hard.

plafer:
Ah, to be fair, I merged #2188 assuming we would merge most of what's in here (after reviews). There were a lot of improvements in this PR - Are you sure you want to close this?

PR #2266, opened 2025-10-07
Remove redundant clone in exec test to avoid unnecessary copy

maradini77:
### PR Description
- **What**: Replace a double-clone pattern by making the initial `mast_forest` binding mutable and removing the subsequent `clone`.
- **Where**: `miden-vm/tests/integration/exec.rs`
- **Why**: The second `clone` created an unnecessary allocation/copy with no behavioral impact. Using a single mutable owner is clearer and cheaper.

#### Change Summary
- Made the first `mast_forest` binding mutable.
- Removed the redundant `mast_forest.clone()` before `advice_map_mut().insert(...)`.


bobbinth:

PR #2267, opened 2025-10-08
add pre/post decorators to `BasicBlockNode` fingerprint

huitseeker:
After #2167, `BasicBlockNode` has `before_enter` and `after_exit` decorators in addition to the existing operation-indexed decorators. However, `MastNodeFingerprint::from_mast_node` was only hashing the operation-indexed decorators for basic blocks, which meant that basic blocks with identical operations but different pre/post decorators would have the same fingerprint.

This fix updates the fingerprint calculation to include all three types of decorators in the correct order:
1. before_enter decorators
2. operation-indexed decorators (existing behavior)
3. after_exit decorators

Added unit tests to verify that basic blocks with different decorator configurations produce different fingerprints.

Bonuses:
- in the first commit, we add `for_each_children` to the `MastNodeExt` trait, which cuts down on a bunch of boilerplate,
- in the last commit, we update expected fingerprints in a snapshot test, a consequence of changing the fingerprinting logic.


core/src/mast/node_fingerprint.rs
```diff
@@ -58,13 +58,34 @@ impl MastNodeFingerprint {
...
             MastNode::Block(node) => {
+                // Check if we have any decorators at all (before_enter, op-indexed, or after_exit)
+                let has_before_decorators = !node.before_enter().is_empty();
+                let has_after_decorators = !node.after_exit().is_empty();
+                let has_op_decorators = node.decorators().next().is_some();
+
+                // If no decorators, return simple fingerprint
+                if !has_before_decorators && !has_after_decorators && !has_op_decorators {
+                    return Ok(MastNodeFingerprint::new(node.digest()));
+                }
```
plafer: This makes us skip the `Assert`, `U32assert2` and `MpVerify` opcodes, even if present (lines 89-109). I think we should revert to the previous behavior (i.e. remove this check, and reinstate the block

```rs
if bytes_to_hash.is_empty() {
    Ok(MastNodeFingerprint::new(node.digest()))
} else {
    let decorator_root = Blake3_256::hash(&bytes_to_hash);
    Ok(MastNodeFingerprint::with_decorator_root(node.digest(), decorator_root))
}
```

We could add a test (or modify one) to test this case explicitly: no decorators, but different `Assert` opcodes, and make sure that the fingerprints are different.

crates/assembly/src/snapshots/miden_assembly__tests__asserts_and_mpverify_with_code_in_duplicate_procedure.snap
plafer, line 1: I think it's incorrect to change this file - I expect this to be unchanged after the previous comment is implemented. IIUC this currently shows that there are collisions between `u32assert2`/`assert` operations with different opcodes.

plafer (changes requested):

core/src/mast/node_fingerprint.rs
huitseeker: Yes, I'm a dunce 🤦 , thanks for catching this!

CHANGELOG.md
```diff
@@ -23,10 +23,13 @@
 - Place the `FastProcessor` stack on the heap instead of the (OS thread) stack (#[2271](https://github.com/0xMiden/miden-vm/pull/2271)).
 - `after_exit` decorators execute in the correct sequence in External nodes in the Fast processor ([#2247](https://github.com/0xMiden/miden-vm/pull/2247)).
 - Remove O(n log m) iteration in parallel processor (#[2273](https://github.com/0xMiden/miden-vm/pull/2273)).
```
adr1anh, line 25: Nit: formatting

PR #2268, opened 2025-10-08
feat: Allow searching for a partial key in `sorted_array`

PhilippGackstatter:
## Describe your changes

Implements `sorted_array::find_half_key_value` procedure that works similar to `find_key_value` except it searches only for a match on half the `KEY` in the array. This is useful in the context of named storage slots https://github.com/0xMiden/miden-base/issues/1724#issuecomment-3380152079, where the memory layout of storage is:

```
[[0, slot_type, name_id_suffix, name_id_prefix], SLOT_VALUE]
```

When searching for a named storage slot, we only want to find the slot that has the two felts `[name_id_prefix, name_id_suffix]`.
We do not need or want to match on the `slot_type`. Having to match the entire key also restricts how we can evolve this structure in the future, e.g. storing more data in the felt that contains `slot_type` or the `0` felt.

This change is done by introducing a `use_half_key` boolean parameter for the actual implementation of `find_key_value`, which is now called `find_partial_key_value`:

Assume a KEY is laid out as `[key3, key2, key1, key0]`.

- If `use_half_key = 0`, then the entry whose key matches the entire KEY is searched for.
- If `use_half_key = 1`, then the entry whose elements key3 and key2 match the corresponding elements
in KEY is searched for.

An earlier commit of this PR generalized this to `key_size` between `1..=4` but for now it was deemed unnecessary to have 1 and 3.

See also the docs of that procedure for more details.

This is a non-breaking change (unless we consider changed MAST roots a breaking change at this stage) and targeted at main.

There is a cycle count TODO left in the code. Could someone let me know the best way to update the cycle counts for the changed procedures? I haven't done this before.

Also applies https://github.com/0xMiden/miden-vm/pull/2284 to `main` to fix the clippy lints.

- Updated `CHANGELOG.md'

### DIFF
--- a/stdlib/asm/collections/sorted_array.masm
+++ b/stdlib/asm/collections/sorted_array.masm
@@ -221,14 +221,12 @@ end
 #!     - `(end_ptr - start_ptr)` must be divisible by 8
 #! - `start_ptr > end_ptr`
 #!
-#! TODO: Update cycle counts.
-#!
 #! Cycles:
-#!   Key exists: 56 cycles
+#!   Key exists: 70 cycles
 #!   Key doesn't exist and the array is empty: 25 cycles
-#!   Key doesn't exist and is smaller than all stored keys: 151 cycles
-#!   Key doesn't exist and is larger than all stored keys: 149 cycles
-#!   Key doesn't exist: 298 cycles
+#!   Key doesn't exist and is smaller than all stored keys: 166 cycles
+#!   Key doesn't exist and is larger than all stored keys: 162 cycles
+#!   Key doesn't exist: 322 cycles
 proc.find_partial_key_value.1
     # Call the non-deterministic lowerbound advisor (5 cycles)
     emit.LOWERBOUND_KEY_VALUE_EVENT
@@ -253,7 +251,7 @@ proc.find_partial_key_value.1
         dup dup.6 sub u32mod.8 assertz.err="lowerbound_key_value invariant: key_ptr must be double-word aligned with start_ptr"
         #=> [maybe_key_ptr, KEY, start_ptr, end_ptr]
 
-        # dereference maybe_key_ptr (8 cycles)
+        # dereference maybe_key_ptr (22 cycles)
         dup movdn.5
         #=> [maybe_key_ptr, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
@@ -288,10 +286,10 @@ proc.find_partial_key_value.1
                 # assert `the key at start_ptr > KEY`
                 #=> [maybe_key_ptr = start_ptr, KEY, start_ptr, end_ptr]
 
+                # load key (22 cycles)
                 dup movdn.5
                 #=> [maybe_key_ptr, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
-                # load key (?? cycles)
                 loc_load.0 exec.load_key
                 #=> [FIRST_KEY, KEY, maybe_key_ptr = start_ptr, start_ptr, end_ptr]
 
@@ -312,7 +310,7 @@ proc.find_partial_key_value.1
 
                 #=> [maybe_key_ptr = end_ptr, KEY, start_ptr, end_ptr]
 
-                # fetch the last key in the array (10 cycles)
+                # fetch the last key in the array (23 cycles)
                 dup movdn.5 sub.8
                 #=> [maybe_key_ptr-8, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
@@ -342,7 +340,7 @@ proc.find_partial_key_value.1
                 movdn.4 dupw
                 #=> [KEY, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
-                # load key (?? cycles)
+                # load key (21 cycles)
                 dup.8 loc_load.0 exec.load_key
                 #=> [MAYBE_KEY, KEY, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
@@ -351,7 +349,7 @@ proc.find_partial_key_value.1
                 assert.err="lowerbound_key_value invariant: MAYBE_KEY not greater than KEY"
                 #=> [KEY, maybe_key_ptr, start_ptr, end_ptr]
 
-                # dereference `maybe_key_ptr-8` (previous key) (11 cycles)
+                # dereference `maybe_key_ptr-8` (previous key) (22 cycles)
                 dup.4 sub.8
                 #=> [maybe_key_ptr-8, KEY, maybe_key_ptr, start_ptr, end_ptr]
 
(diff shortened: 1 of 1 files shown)


stdlib/docs/collections/sorted_array.md
```diff
@@ -3,4 +3,7 @@
...
+| load_key | Loads the KEY at the pointer from memory and conditionally sets its elements to zero.<br /><br />Given the KEY laid out as [key3, key2, key1, key0]:<br /><br />if key_size == 4: return KEY unchanged<br />if key_size <= 3 -> is_key_size_3 or is_key_size_2 or is_key_size_1: set key0 = 0<br />if key_size <= 2 -> is_key_size_2 or is_key_size_1: set key1 = 0<br />if key_size == 1 -> is_key_size_1: set key2 = 0<br /><br />Key sizes outside the range 1..=4 are unsupported.<br /><br />Inputs:  [key_size, key_ptr]<br />Outputs: [KEY]<br /> |
```
PhilippGackstatter: `load_key` is defined as `proc`, not `export`. Are internal procedure actually intended to be part of the docs? I've built them with `make build-no-std`. cc @huitseeker

mmagician:
Some comments without reviewing the code and only the description:

> I originally thought that just half and full key would be sufficient, and maybe they are. But it turned out that supporting all key sizes was possible with a few more cycles and this may become useful in the future.

At this stage of the project I'd rather optimize for code simplicity (better maintainability for us, easier reviewability for audits) rather than support for hypothetical use cases. So, if the code is simpler with `full` and `half`-key lookups, I'd favor that option.

stdlib/docs/collections/sorted_array.md
PhilippGackstatter: This is still a problem, but for now I've manually removed the private procedures. If someone can confirm that this is an actual unintended bug, we should open an issue.

plafer: Created #2307

stdlib/src/handlers/sorted_array.rs
```diff
@@ -145,6 +164,30 @@ fn push_lowerbound_result(
     ])])
 }
 
+trait LexicographicWordExt {
+    /// Selectivel zeroizes the felts in a [`LexicographicWord`] based on the provided [`KeySize`].
```
bobbinth: typo: "Selectivel" -> "Selectively"?

stdlib/src/handlers/sorted_array.rs
```diff
@@ -105,9 +124,9 @@ fn push_lowerbound_result(
...
     let get_word = {
         |addr: u32| {
-            process
-                .get_mem_word(process.ctx(), addr)
-                .map(|word| LexicographicWord::new(word.unwrap_or(Word::empty())))
+            process.get_mem_word(process.ctx(), addr).map(|word| {
+                LexicographicWord::new(word.unwrap_or(Word::empty())).zeroize_by_key_size(key_size)
+            })
         }
     };
```
bobbinth, line 131: nit (feel free to ignore): I wonder if instead of the extension trait, we could just have a simple function for constructing `LexicographicWord` - something like:
```rust
fn word_to_search_key(value: Option<Word>, key_size: KeySize) -> LexicographicWord {
    ...
}
```

stdlib/src/handlers/sorted_array.rs
```diff
@@ -38,19 +44,31 @@ pub fn handle_lowerbound_array(process: &ProcessState) -> Result<Vec<AdviceMutat
...
 /// Returns an error if the keys are not sorted in non-decreasing order.
 pub fn handle_lowerbound_key_value(
     process: &ProcessState,
 ) -> Result<Vec<AdviceMutation>, EventError> {
-    push_lowerbound_result(process, 8)
+    let use_half_key = process.get_stack_item(7);
+    debug_assert!(
+        use_half_key == Felt::ONE || use_half_key == Felt::ZERO,
+        "use_half_key must be bool, was {use_half_key}"
+    );
```
bobbinth: Should this be an error?

stdlib/asm/collections/sorted_array.masm
```diff
@@ -144,43 +146,121 @@ end
...
+#! - end_ptr is not double-word-aligned with the start_ptr:
+#!     - `(end_ptr - start_ptr)` must be divisible by 8
+#! - `start_ptr > end_ptr`
+export.find_half_key_value
+    push.0 movdn.2 push.0 movdn.2
+    # => [key_prefix, key_suffix, 0, 0, start_ptr, end_ptr]
+    # => [KEY, start_ptr, end_ptr]
+
+    push.1
+    # => [use_half_key = 0, KEY, start_ptr, end_ptr]
```
bobbinth: In the comment, it should be `use_half_key = 1` - right?

stdlib/asm/collections/sorted_array.masm
```diff
@@ -281,3 +369,38 @@ export.find_key_value
...
+#! Loads a key from the pointer.
+#!
+#! If use_half_key is 1, the key is loaded with the two least significant elements set to zero,
+#! e.g.: [key3, key2, 0, 0], while if it is 0, the full key [key3, key2, key1, key0] is loaded.
+#!
+#! This procedure avoids branches as an optimization.
+#!
+#! Inputs:  [use_half_key, key_ptr]
+#! Outputs: [KEY]
+proc.load_key
```
bobbinth, line 382: If we invert the meaning of `use_half_key` - i.e., change it to `use_full_key` - then, I think this could be done much simpler - e.g. something like:
```
#! Inputs:  [use_full_key, key_ptr]
#! Outputs: [KEY]
proc.load_key
    padw movup.5 mem_loadw
    # => [FULL_KEY, use_full_key]

    # multiple the two least significant elements by use_full_key
    movup.3 dup.4 mul
    movup.3 movup.4 mul
    movup.3 movup.3
    # => [KEY]
end
```

This should take 14 cycles (but would be good to double-check).

PhilippGackstatter: Nice optimization, thank you! Refactored `use_half_key` to `use_full_key` and addressed your other comments as well.

stdlib/asm/collections/sorted_array.masm
```diff
@@ -144,43 +146,121 @@ end
...
+#! # Panics
+#!
+#! Panics if:
+#! - keys are not sorted in non-decreasing order,
+#! - start_ptr is not word-aligned
+#! - end_ptr is not double-word-aligned with the start_ptr:
+#!     - `(end_ptr - start_ptr)` must be divisible by 8
+#! - `start_ptr > end_ptr`
+#!
+#! TODO: Update cycle counts.
```
bobbinth: We should update the cycle counts below.

plafer:
This is targeting main - is the goal to do another minor release? We should release 0.19 soon, wouldn't it make more sense to include this in it?

bobbinth:
> This is targeting main - is the goal to do another minor release? We should release 0.19 soon, wouldn't it make more sense to include this in it?

Yes, the goal to make it a patch release because releasing v0.19.0 will probably not happen until Friday, and there is some work in `miden-base` that relies on this (which we hope to be merged into main before v0.19.0 is available).

PR #2270, opened 2025-10-08
chore: place the fast processor stack on the heap instead of the stack

plafer:
Places the `FastProcessor` stack on the heap to reduce the risk of (OS thread) stack overflow, as observed in e.g. https://github.com/0xMiden/miden-base/issues/1927.

Benchmarks are unaffected by this change.

Also related: https://github.com/0xMiden/miden-base/issues/1936#issuecomment-3380244534

@bobbinth I could rebase this against `main` as well if we want to release a patch version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Fixed hex word parsing to guard against missing 0x prefix ([#2245](https://github.com/0xMiden/miden-vm/pull/2245)).
 - Systematized u32-indexed vectors ([#2254](https://github.com/0xMiden/miden-vm/pull/2254)).
 - Introduce a new `build_trace()` which builds the trace in parallel from trace fragment contexts ([#1839](https://github.com/0xMiden/miden-vm/pull/1839)) ([#2188](https://github.com/0xMiden/miden-vm/pull/2188))
+- Place the `FastProcessor` stack on the heap instead of the (OS thread) stack (#[2270](https://github.com/0xMiden/miden-vm/pull/2270))
  
 ## 0.18.1 (2025-10-02)
 
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -1,4 +1,4 @@
-use alloc::{sync::Arc, vec::Vec};
+use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use core::cmp::min;
 
 use memory::Memory;
@@ -87,7 +87,7 @@ const INITIAL_STACK_TOP_IDX: usize = 250;
 #[derive(Debug)]
 pub struct FastProcessor {
     /// The stack is stored in reverse order, so that the last element is at the top of the stack.
-    pub(super) stack: [Felt; STACK_BUFFER_SIZE],
+    pub(super) stack: Box<[Felt; STACK_BUFFER_SIZE]>,
     /// The index of the top of the stack.
     stack_top_idx: usize,
     /// The index of the bottom of the stack.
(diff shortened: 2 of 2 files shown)


adr1anh:
Looks good!

I recall we were using a `Vec` initially but that performance wasn't great. It doesn't seem like there should be much different though with `Box`. Do you think it's worth considering `Vec` again, since that would avoid imposing a hard limit on the stack size (even if in practice it doesn't seem like we would reach it)?

adr1anh (approved):

plafer:
Good point! The `Vec` experiment that was slower was one specifically using `push()` and `pop()`. But this is probably the way to go to solve the current ["panic on overflow" behavior](https://github.com/0xMiden/miden-vm/blob/be759680eafa233650a0be3da54ff2416ff21f0a/processor/src/fast/operation.rs#L281) that we have: use a `Vec` initialized as `vec![ZERO; STACK_BUFFER_SIZE]`. When the reach the "right end" (i.e. an overflow), then move the data "leftwards" and/or double the size of the allocation (initialized to 0) (i.e. `stack.resize(stack.len() * 2, ZERO)`).

plafer:
I tried the above, and for some reason, all else equal, using a `Vec<Felt>` instead of `Box<[Felt; STACK_BUFFER_SIZE]>` decreases the performance. On the `blake3_1to1` benchmark, we have

- `Box<[Felt; STACK_BUFFER_SIZE]>`: 2.61ms
- `Vec<Felt>`: 2.70ms **+ 3.7%**
  - Note: initialized as `vec![ZERO; STACK_BUFFER_SIZE]`, so the underlying buffer is never resized.

Somehow, memory accesses from a `Vec<Felt>` are slower than a `Box<[Felt; STACK_BUFFER_SIZE]>`, and I'm not sure why. We would probably need to dig into the generated llvm IR next, but I prefer to leave it as a `Box` for now until we understand this better.

plafer:
Merging this; @bobbinth we can cherry-pick this commit on `main` if you want to release a 0.18 patch as well.

bitwalker:
> I tried the above, and for some reason, all else equal, using a `Vec<Felt>` instead of `Box<[Felt; STACK_BUFFER_SIZE]>` decreases the performance. On the `blake3_1to1` benchmark, we have
> 
> * `Box<[Felt; STACK_BUFFER_SIZE]>`: 2.61ms
> * `Vec<Felt>`: 2.70ms **+ 3.7%**
>   
>   * Note: initialized as `vec![ZERO; STACK_BUFFER_SIZE]`, so the underlying buffer is never resized.
> 
> Somehow, memory accesses from a `Vec<Felt>` are slower than a `Box<[Felt; STACK_BUFFER_SIZE]>`, and I'm not sure why. We would probably need to dig into the generated llvm IR next, but I prefer to leave it as a `Box` for now until we understand this better.

It's likely because the allocated memory backing a `Box<[T]>` can never change, while pushing on a `Vec<T>` may reallocate new memory for the contents - which means that every store to a `Vec<T>` invalidates previous reads of the data pointer, requiring the pointer value to be reloaded before attempting to load from the vector, or compute pointers to individual elements. The same is _not_ true for `Box<[T]>`, as even though writes may invalidate prior loads of individual elements, the pointer to the slice itself can be reused across stores, as can pointers to individual elements. That can add up to a quite a few instructions saved.

bobbinth:
> Merging this; @bobbinth we can cherry-pick this commit on `main` if you want to release a 0.18 patch as well.

Yes, that would be great. I think we now need to cherry-pick two commits, right?

plafer:
> Yes, that would be great. I think we now need to cherry-pick two commits, right?

Right, you ultimately want the behavior from #2271

PR #2271, opened 2025-10-08
chore(processor): properly allocate the fast processor stack directly on heap

plafer:
Fixes a mistake made in #2270, pointed out by @Mirko-von-Leipzig.

`Box::<[T; N]>::new()` will first allocate the array on the stack, and then move it to the heap (https://github.com/rust-lang/rust/issues/53827). In this PR, we use `Vec::into_boxed_slice()` as a workaround. I verified that this does indeed allocate directly on the heap compared to `Box::new()`.


bobbinth:
@plafer - could you open a PR to cherry-pick this commit into `main`?

PR #2273, opened 2025-10-09
feat: add an Operation iterator which spits out group indexes

### LINKED ISSUE #2170: `CoreTraceFragmentGenerator::execute_op_batch`: avoid linear scan on every op
Ah, I would recommend leaving a TODO to add an operation iterator that gives you those group indexes, since the complexity of performing that is `O(1)` per operation (as opposed to the current `O(log N)` where `N` is the batch length).

It's because that iterator is easy to make (and that group indexes weren't really used elsewhere) that I didn't add a group indexation api already. You can see an example with `group_chunks`, which can be tweaked to support your `next_op_group_index` approach as well.

This doesn't have to be done in this PR, of course.

_Originally posted by @huitseeker in https://github.com/0xMiden/miden-vm/pull/1839#discussion_r2351163516_

huitseeker:
Uses it in processor/src/parallel/mod.rs. Closes #2170.

### DIFF
--- a/core/src/mast/node/basic_block_node/op_batch.rs
+++ b/core/src/mast/node/basic_block_node/op_batch.rs
@@ -126,7 +126,7 @@ impl OpBatch {
     /// # Performance Consideration
     /// For iterating over all operations in a batch, prefer using `iter_with_groups()`
     /// which tracks group boundaries incrementally in O(1) per operation rather than
-    /// O(log m) per operation.
+    /// O(log m) per operation, where m is the number of groups.
     #[must_use]
     pub fn op_idx_in_batch_to_group(&self, op_idx_in_batch: usize) -> Option<(usize, usize)> {
         if op_idx_in_batch >= self.ops.len() {
@@ -161,12 +161,8 @@ impl OpBatch {
     ///
     /// This is significantly more efficient than calling `op_idx_in_batch_to_group()` for each
     /// operation because it tracks group boundaries incrementally instead of using binary search
-    /// for each operation (which would be O(n * log m) total).
-    ///
-    /// This method returns an iterator that yields tuples of `(group_idx, op_idx_in_group,
-    /// operation)` for each operation in the batch, making it easy to iterate over operations
-    /// while keeping track of their group positions without the overhead of repeated binary
-    /// searches.
+    /// for each operation (which would be O(n * log m) total, where m is the number of groups,
+    /// and n is the number of operations).
     pub fn iter_with_groups(&self) -> OpBatchIterator<'_> {
         OpBatchIterator::new(self)
     }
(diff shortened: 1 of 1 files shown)


core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -138,6 +148,29 @@ impl OpBatch {
...
+    /// # Returns
+    /// An iterator that produces tuples containing:
+    /// - `group_idx`: Index of the operation group within the batch (0-7)
+    /// - `op_idx_in_group`: Index of the operation within its group (0-8)
+    /// - `operation`: Reference to the operation itself
+    ///
+    /// This is significantly more efficient than calling `op_idx_in_batch_to_group()` for each
+    /// operation because it tracks group boundaries incrementally instead of using binary search
+    /// for each operation (which would be O(n * log m) total).
+    ///
```
Al-Kindi-0: nit: would be nice to introduce what are `m` and `n` here

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -138,6 +148,29 @@ impl OpBatch {
...
+    /// This method returns an iterator that yields tuples of `(group_idx, op_idx_in_group,
+    /// operation)` for each operation in the batch, making it easy to iterate over operations
+    /// while keeping track of their group positions without the overhead of repeated binary
+    /// searches.
```
Al-Kindi-0: nit: seems like there is a bit of redundancy with the previous paragraph, maybe we can merge them?

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -345,9 +378,73 @@ impl OpBatchAccumulator {
...
+/// in the batch, tracking group boundaries incrementally without binary search.
+///
+/// # Fields
+/// - `batch`: Reference to the batch being iterated
+/// - `current_op_idx`: Current operation index within the batch
+/// - `grp_idx`: Current group index tracked incrementally
+///
+///
+/// This is more efficient than calling `op_idx_in_batch_to_group()` for each
+/// operation (O(log m) per operation) because it tracks group boundaries incrementally
```
Al-Kindi-0: nit: same point regarding `m`'s definition

core/src/mast/node/basic_block_node/op_batch.rs
```diff
@@ -345,9 +378,73 @@ impl OpBatchAccumulator {
...
+        if self.current_op_idx >= self.batch.ops().len() {
+            return None;
+        }
+
+        let op = &self.batch.ops()[self.current_op_idx];
+        let indptr = self.batch.indptr();
+
+        // Increment grp_idx if we've crossed into the next group
+        // Check if current_op_idx >= indptr[self.grp_idx + 1] (start of next group)
+        while self.grp_idx + 1 < self.batch.num_groups()
```
Al-Kindi-0, line 419: nit(feel free to ignore): I would love a short reminder of why we use the `while` loop here.

PR #2274, opened 2025-10-10
MastNode builders extension (2/4)

huitseeker:
## Changes
* After PR 2259, you still needed two steps: build node → `forest.add_node(built_node)`
* After this PR, you can add nodes directly with `builder.add_to_forest(&mut forest)` - single unified workflow
* Node mutations operations (e.g. children remapping) are moved progressively to the builder API for consistency
* This builder API is meant to become the single (external) point of interaction for creating and adding nodes.

### DIFF
--- a/core/src/mast/node/basic_block_node/arbitrary.rs
+++ b/core/src/mast/node/basic_block_node/arbitrary.rs
@@ -6,7 +6,10 @@ use proptest::{arbitrary::Arbitrary, prelude::*};
 use super::*;
 use crate::{
     AdviceMap, AssemblyOp, DebugOptions, Decorator, Felt, Kernel, Operation, Program, Word,
-    mast::DecoratorId,
+    mast::{
+        CallNodeBuilder, DecoratorId, DynNodeBuilder, ExternalNodeBuilder, JoinNodeBuilder,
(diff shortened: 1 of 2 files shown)


core/src/mast/node/mast_forest_contributor.rs
```diff
@@ -0,0 +1,6 @@
+use crate::mast::{MastForest, MastForestError, MastNodeId};
+
+#[allow(dead_code)]
+pub trait MastForestContributor {
+    fn add_to_forest(self, forest: &mut MastForest) -> Result<MastNodeId, MastForestError>;
```
plafer, line 5: It's a bit hard to know without seeing the future PRs if this is the right API. From what I recall, you were hoping to change the `MastForestBuilder` so that it accumulates state, and only creates the `MastForest` at the end. However, this API forces us to mutate the `MastForest` directly, without any option to add state to a `MastForestBuilder` (e.g. the decorators).

Am I reading into this right? I'm also happy to wait for the future PRs to see how it ends up being used.

huitseeker: Waiting for the future PRs is recommended, sorry. the builders will definitely have mutation capabilities.

adr1anh (approved):
Looks good!

PR #2277, opened 2025-10-10
feat: ECDSA precompile

adr1anh:
# ECDSA Precompile Implementation - PR Summary

## Summary

This PR adds ECDSA secp256k1 signature verification as a precompile, following the deferred verification pattern established in the `log-precompile` branch. The new `std::crypto::dsa::ecdsa::secp256k1::verify` procedure provides a memory-based API that delegates cryptographic verification to the host, reducing cycle count from ~10,000 to ~100 cycles. During execution, the host performs native signature verification and provides the result via the advice stack, while logging the verification inputs for deferred validation at proof verification time. The verifier recomputes a hierarchical commitment `Rpo256(Rpo256(Rpo256(pk) || Rpo256(digest)) || Rpo256(sig))` that binds the inputs and result together, ensuring soundness.

## Structural Changes

To improve maintainability, we've consolidated the deprecated native implementations within their respective modules rather than maintaining separate `_deprecated` files. The old MASM-native ECDSA implementation (which performed elliptic curve operations in-circuit) is now preserved as `verify_deprecated` in the main `secp256k1.masm` file. Similarly, Keccak256's native implementation procedures (theta, rho, pi, chi, etc.) have been moved from `keccak256_deprecated.masm` into the main `keccak256.masm` under a "DEPRECATED PROCEDURES" section. This consolidation eliminates module fragmentation while clearly marking legacy code paths. We also introduced `std::sys::log_precompile_request`, a helper procedure that wraps the `log_precompile` instruction and cleans up the stack by dropping the three internal helper words, providing a simpler API for precompile wrappers.

## Test Infrastructure

The test suite has been unified across precompiles with a consistent three-tier validation strategy: unit tests for component-level validation (commitment computation, serialization), integration tests that execute MASM wrappers with handlers to validate stack outputs and commitment matching, and end-to-end tests that exercise the full prove/verify cycle with transcript validation. We introduced shared test helpers in `stdlib/tests/helpers.rs` (notably `masm_store_packed_bytes`) that handle the common pattern of setting up memory with packed u32 data, eliminating code duplication between Keccak and ECDSA tests. The new `stdlib/tests/sys/mod.rs` module provides comprehensive integration testing for `log_precompile_request`, including a full prove/verify cycle with dummy handlers that validates the entire precompile infrastructure end-to-end.

## Migration Path

The old stack-based ECDSA API (48 elements with Montgomery form coordinates) remains available as `verify_deprecated`, allowing gradual migration. The new API accepts three memory pointers to packed u32 byte arrays (33-byte compressed public key, 32-byte digest, 66-byte signature) and returns a boolean result (1 for valid, 0 for invalid) rather than aborting on failure. All deprecated procedures remain fully functional and tested to ensure backward compatibility while clearly signaling the recommended migration path through naming and documentation.

### DIFF
--- a/core/src/precompile.rs
+++ b/core/src/precompile.rs
@@ -24,8 +24,7 @@
 //!
 //! 3. **Proof Generation**: The prover generates a STARK proof of the VM execution. The final
 //!    [`PrecompileTranscript`] state (sponge capacity) is a public input. The verifier enforces the
-//!    initial (empty) and final state via variable‑length public inputs. TODO(#2045): These will be
-//!    enforced after updates to the recursive verifier.
+//!    initial (empty) and final state via variable‑length public inputs.
 //!
 //! 4. **Verification**: The verifier:
 //!    - Recomputes each precompile commitment using the stored requests via [`PrecompileVerifier`]


stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -35,10 +36,7 @@ export.hash_memory
     # => [COMM, TAG, DIGEST_U32[8], ...]
 
     # Log the precompile request for deferred verification
-    # Requires a placeholder word (PAD) so the top 3 words are [COMM, TAG, PAD]
-    padw movdnw.2
-    log_precompile
-    dropw dropw dropw
+    exec.sys::log_precompile_request
```
adr1anh, line 38: this is the only change in the file. I've appended the deprecated functions to avoid adding a new module.

stdlib/asm/sys/mod.masm
```diff
@@ -33,3 +33,16 @@ end
 export.drop_stack_top
     dropw dropw dropw dropw
 end
+
+#! Logs a precompile commitment and removes the helper words produced by `log_precompile`.
+#!
+#! Input: `[COMM, TAG, ...]`
+#! Output: Stack without the three helper words `[R1, R0, CAP_NEXT]`.
+#! Cycles: 1 (plus cost of the underlying `log_precompile` instruction).
+export.log_precompile_request
```
adr1anh, line 42: This is the procedure that is thoroughly tested to ensure that the log-precompile op-code works. It allows us to avoid the complicated end-to-end tests for the precompiles and only focus on the implementations of the wrappers, the handlers and verifiers.

stdlib/src/handlers/ecdsa.rs
```diff
@@ -0,0 +1,299 @@
...
+
+/// Qualified event name for the ECDSA signature verification event.
+pub const ECDSA_VERIFY_EVENT_NAME: &str = "stdlib::crypto::dsa::ecdsa::verify";
+/// Constant Event ID for the ECDSA verification event, derived via
+/// `EventId::from_name(ECDSA_VERIFY_EVENT_NAME)`
+pub const ECDSA_VERIFY_EVENT_ID: EventId = EventId::from_u64(7877366289282707984);
+
+const PUBLIC_KEY_LEN_BYTES: usize = 33;
+const MESSAGE_DIGEST_LEN_BYTES: usize = 32;
+const SIGNATURE_LEN_BYTES: usize = 66; // r (32) + s (32) + v (1) + padding (1)
```
adr1anh, line 54: These could be imported from crypto, but they aren't public (only public key is)

stdlib/asm/crypto/hashes/keccak256.masm
```diff
@@ -134,3 +132,3510 @@ export.hash_memory_impl
     movupw.3 swapw
     # => [COMM, TAG, DIGEST_U32[8], ...]
 end
+
+# DEPRECATED PROCEDURES
+# ================================================================================================
```
plafer: I don't think we need to keep those around - if ever we want to bring some of this stuff back, we can go look in the git history

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -1,3 +1,108 @@
...
+#! All data must be stored in memory as packed u32 values (little-endian), with unused bytes
+#! in the final u32 set to zero.
+export.verify
+    exec.verify_impl
+    # => [COMM, TAG, result, ...]
+
+    # Log the precompile request for deferred verification
+    exec.sys::log_precompile_request
+    # => [result, ...]
+end
```
Al-Kindi-0, line 48: Q: What is the rational for logging a request when result is `false`?

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -1,3 +1,108 @@
...
+#!
+#! Where:
+#! - `ptr_pk`: word-aligned memory address containing 33-byte public key
+#! - `ptr_digest`: word-aligned memory address containing 32-byte digest
+#! - `ptr_sig`: word-aligned memory address containing 66-byte signature
+#! - `COMM`: commitment to calldata computed as
+#!       `Rpo256(Rpo256(Rpo256(pk) || Rpo256(digest)) || Rpo256(sig))`
+#! - `TAG`: `[ECDSA_VERIFY_EVENT, result, 0, 0]`
+#! - `result`: 1 if signature is valid, 0 if invalid
+export.verify_impl
```
Al-Kindi-0, line 73: Q: Should we mention that this is the equivalent of `sign_prehash`? This will also clarify `ptr_digest`?

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -1,3 +1,108 @@
...
+
+# INTERNAL PROCEDURES
+# ================================================================================================
+
+#! Internal implementation of ECDSA signature verification via deferred computation.
+#!
+#! Emits an event to trigger the precompile handler, reads the verification result from
+#! the advice stack, and computes the commitment and tag for tracking deferred verification.
+#!
+#! Input: `[ptr_pk, ptr_digest, ptr_sig, ...]`
```
Al-Kindi-0: Nit: we usually write `pk_ptr` instead of `ptr_pk`

stdlib/src/handlers/ecdsa.rs
Al-Kindi-0: Good idea

stdlib/src/handlers/mod.rs
```diff
@@ -15,3 +19,155 @@ fn u64_to_u32_elements(value: u64) -> (Felt, Felt) {
...
+/// let bytes = read_memory_packed_u32(process, 0x100, 5)?;
+/// // Returns: [0x01, 0x02, 0x03, 0x04, 0x05]
+/// ```
+pub(crate) fn read_memory_packed_u32(
+    process: &ProcessState,
+    start: u64,
+    len_bytes: usize,
+) -> Result<Vec<u8>, MemoryReadError> {
+    // Validate word alignment
+    if !start.is_multiple_of(4) {
```
Al-Kindi-0: nit: would it make sense to define a constant for 4?

Al-Kindi-0 (commented):
Looks great!
Left a couple of questions, but otherwise this is ready to merge in my opinion.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
adr1anh: It allows you to prove that a signature is invalid. This could be useful for situations like slashing where you could prove malicious behavior. 

It doesn't seem like there is much overhead to providing this more general interface and letting the caller assert the result is true.

stdlib/src/handlers/keccak256.rs
plafer, line 1: For future reference, updates to this file could have been done in a separate PR (in light of Bobbin's comment about reducing PR size)

stdlib/tests/crypto/ecdsa_secp256k1_deprecated.rs
plafer, line 1: I would remove this file

stdlib/tests/crypto/keccak256_deprecated.rs
plafer, line 1: I would also remove this file

Cargo.toml
```diff
@@ -53,7 +53,7 @@ miden-utils-testing = { path = "./crates/test-utils", version = "0.1", package =
 miden-verifier = { path = "./verifier", version = "0.19", default-features = false }
 
 # Miden crates
-miden-crypto = { version = "0.17", default-features = false }
+miden-crypto = { version = "0.17.1", default-features = false }
```
plafer: nit: I think a `cargo update` would update the `Cargo.lock` file, without requiring to change this file?

stdlib/asm/crypto/hashes/keccak256.masm
adr1anh: done

CHANGELOG.md
huitseeker, line 33: I thought this was #2249 ?

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -200,6 +206,68 @@ pub(super) fn op_horner_eval_base<P: Processor>(
...
+///
+/// Stack transition:
+/// `[COMM, TAG, PAD, ...] -> [R1, R0, CAP_NEXT, ...]`
+///
+/// Where:
+/// - The hasher computes: `[CAP_NEXT, R0, R1] = Rpo([CAP_PREV, TAG, COMM])`
+/// - `CAP_PREV` is the previous sponge capacity provided non-deterministically via helper
+///   registers.
+#[inline(always)]
+pub(super) fn op_log_precompile<P: Processor>(
```
huitseeker, line 223: This PR mixes two distinct features: `log_precompile` VM instruction and ECDSA API changes. Since PR #2249 already introduced the `log_precompile` instruction, could you rebase this PR to separate:

1. `log_precompile` improvements (if any) from PR #2249
2. ECDSA API modernization

This would make it much easier to review the cryptographic VM instruction changes independently from the stdlib API changes. The diff is currently hard to follow because of the mixing.

adr1anh: The git history is messed up since I was branched off log-precompile. I'll rework the history and create an intermediary PR.

adr1anh: This PR now only contains ECDSA-related code. Other fixes are included in #2325

stdlib/src/handlers/ecdsa.rs
bobbinth: Let's add this to the list of follow-up.

stdlib/src/handlers/ecdsa.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    /// Computes the precompile commitment for this request.
+    ///
+    /// The commitment is `RPO(RPO(RPO(pk) || RPO(digest)) || RPO(sig))` with tag
+    /// `[event_id, result, 0, 0]`, where `result` is 1 for valid signatures and 0 for
+    /// invalid ones. Each component is hashed over u32‑packed field elements.
+    ///
+    /// This is called by the [`PrecompileVerifier`] at verification time and must match
+    /// the commitment generated during execution.
+    pub fn as_precompile_commitment(&self) -> PrecompileCommitment {
```
bobbinth, line 197: nit: I would probably rename this to `to_precompile_commitment()`. By convention, methods starting with `as_` imply little or now work, and when there is some work `to_` is more appropriate.

stdlib/src/handlers/ecdsa.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Returns a reference to the signature.
+    pub fn sig(&self) -> &Signature {
+        &self.sig
+    }
+
+    /// Converts this request into a [`PrecompileRequest`] for deferred verification.
+    ///
+    /// Serializes the request data (public key || digest || signature) and wraps it in a
+    /// PrecompileRequest with the ECDSA event ID.
+    pub fn as_precompile_request(&self) -> PrecompileRequest {
```
bobbinth, line 175: nit: same comment as above - I'd rename this into `to_precompile_request()`.

stdlib/src/handlers/ecdsa.rs
```diff
@@ -0,0 +1,285 @@
...
+        let mut calldata = Vec::with_capacity(PRECOMPILE_REQUEST_LEN);
+        self.write_into(&mut calldata);
+        PrecompileRequest::new(ECDSA_VERIFY_EVENT_NAME.to_event_id(), calldata)
+    }
+
+    /// Performs ECDSA signature verification and returns the result.
+    ///
+    /// Returns `true` if the signature is valid for the given public key and digest,
+    /// `false` otherwise.
+    pub fn result(&self) -> bool {
```
bobbinth, line 185: nit: we usually use non-action-verb method names for "properties" (i.e., simple accessors). Here, I'd probably use something like `compute_result()` or `perform_verification()` or something similar.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -1,285 +1,105 @@
...
-    movup.4
-    assert_eq
-
-    movup.3
-    assert_eq
+#! Verifies an ECDSA signature over the secp256k1 curve.
+#!
+#! This procedure uses deferred verification via a precompile. The actual cryptographic
+#! verification is performed by the host, and the result is provided via the advice stack.
+#!
```
bobbinth, line 29: I would maybe mention somewhere here that this procedure is agnostic as to what hash algorithm was used to generate the message digest.

bobbinth (approved):
Looks great! Thank you! I left some nits inline - mostly related to naming. These can be addressed in follow-up PRs.

PR #2278, opened 2025-10-10
fix(processor): disallow same-clock write conflicts; allow multiple reads

### LINKED ISSUE #2276: `FastProcessor` memory allows for multiple writes in same clock cycle
This should be disallowed, similar to how it's done with `Process`'s `Memory` on write [here](https://github.com/0xMiden/miden-vm/blob/09b7b70c6a592a9e6e8696d145dc546d5960d590/processor/src/chiplets/memory/segment.rs#L192-L195). Specifically, multiple reads are allowed, but only 1 access is allowed to a given memory location when one of the accesses is a write (see #1626).

Forostovec:
## Describe your changes

Fixes #2276 

Align FastProcessor memory semantics with Process memory: multiple reads to the same (context, word address) are allowed in the same clock cycle, but if any access is a write, only that single access is allowed that cycle.

Why: Match the rule enforced in Process memory and the policy introduced in “allow more than 1 memory reads in the same clock cycle” so the fast path can’t produce illegal traces.

What changed:
processor/src/fast/memory.rs: Track last access per (ctx, word_addr) with clock and kind. On read_word, reject if the last access in the same clk was a write; otherwise record a read. On write_word, reject if there was any access in the same clk; otherwise record a write.

Scope: word-level accesses (e.g., MLoadW/MStoreW) in the fast path.

### DIFF
--- a/processor/src/fast/memory.rs
+++ b/processor/src/fast/memory.rs
@@ -1,4 +1,5 @@
 use alloc::{collections::BTreeMap, vec::Vec};
+use core::cell::RefCell;
 
 use miden_air::RowIndex;
 use miden_core::{EMPTY_WORD, Felt, WORD_SIZE, Word, ZERO};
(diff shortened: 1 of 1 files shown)


plafer:
Thank you for your interest, but this is not quite what I had in mind, and I'll prefer to handle that one myself. For example, we don't need to use a `BTreeMap`, which would grow as the program progresses. Rather, in principle we can only use a constant amount of memory.

PR #2279, opened 2025-10-11
feat(precompiles): implement sealed trait pattern for PrecompileVerifier

huitseeker:
- Removes blanket implementation of PrecompileVerifier for function types in core
- Move PrecompileVerifier trait and associated types to stdlib with sealed pattern
- Create private Sealed marker trait to restrict implementations to stdlib
- Implement KeccakVerifier struct to replace function-based verifier
- Add Sealed bound to PrecompileVerifier trait

 Not really meant to be merged, meant as a PoC of the sealed trait pattern for #2183.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1653,6 +1653,7 @@ version = "0.19.0"
 dependencies = [
  "miden-air",
  "miden-core",
+ "miden-stdlib",
  "thiserror",
  "tracing",
  "winter-verifier",
--- a/core/src/precompile.rs
+++ b/core/src/precompile.rs
@@ -1,7 +1,7 @@
-use alloc::{boxed::Box, collections::BTreeMap, sync::Arc, vec::Vec};
+use alloc::{boxed::Box, vec::Vec};
 use core::error::Error;
 
-use miden_crypto::{Felt, Word, hash::rpo::Rpo256};
+use miden_crypto::{Felt, Word};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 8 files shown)


adr1anh:
Thanks a lot @huitseeker! I've taken the liberty of including some of the changes in the existing branch. However, there are some points that are not necessary for the current implementation.

> * Removes blanket implementation of PrecompileVerifier for function types in core

This makes sense, and I've removed it in the latest commit.

> * Move PrecompileVerifier trait and associated types to stdlib with sealed pattern

The end goal is to factor the stdlib into a separate crate or event repo, since the VM should be usable without it. Precompiles on the other hand are built-in the VM and controlled by us. 

> * Create private Sealed marker trait to restrict implementations to stdlib

The more I think of it, sealed traits don't really make sense, and the only way to implement them without creating cyclic or unwanted dependencies would be to centralize all precompile functionality in a single trait that would be imported by `processor`. In this PR, the `verifier` crate imports `stdlib` which is not compatible with the goal of factoring the latter crate out of this repo. 

If we add a new `precompile`, then we could make it sealed, preventing external implementations. 

But allowing external implementations doesn't affect soundness. The entity which will be executing the verifier is entirely in control of which PrecompileVerifiers it is able or wants to support, and it is the verifier implementation of these precompiles that dictate how the original handler should've been executed. Moreover, PR #2249 will ensure that precompiles called by the VM must be verified natively. Once it is merged, the only scenarios the result in unsound precompile invocations are 
- The user calls the precompile's EventHandler without logging the commitment to the request. Here, this would be caught by a node which re-executes the program, since the processor logs the event handler call, and therefore the VM and verifier would end up with different aggregated commitments, thereby making the verification fail.
- The user uses a custom implementation of the precompile event handler. The only way this would work is if their implementation is equivalent to the handler that is expected by the verifier.

Allowing users to implement their own precompiles isn't necessarily a bad thing, they just need to make sure that their code is correct, in the same way that writing an Event handler could be unsafe if the wrapper doesn't properly constrain the advice output by the handler.

Finally when we will have implemented the Precompile VM, the same arguments as above hold in a similar way. To ensure overall soundness of the network, in which users run programs which call precompiles, any recursive verification calls that verify a Precompile VM proof will use a "blessed" verification key. This key is a commitment to the constraints we have validated and allow on the network. If any unsupported precompiles are called, then it is impossible to create a proof that verifies against that key. 

> * Implement KeccakVerifier struct to replace function-based verifier

Good point, but instead I implemented EventHandler and PrecompileVerifier for a new KeccakPrecompile struct

huitseeker:
Yep, I agree that with the removal of the blanket impl for a function type, the PR is much less of a footgun. The sealed trait pattern isn't necessary for security because verification integrity is enforced at the network level - nodes/verifiers control which precompiles they support and will reject any proof with unsupported precompile types. External implementations can't compromise soundness because they'd need matching verifier implementations that network operators explicitly choose to accept — and in the _current_ state of #2183, it's become harder to make an erroneous one.

PR #2284, opened 2025-10-14
fix: Add #[allow(unused_assignments)] surrounding the miette Diagnostic macros

huitseeker:
This is a stopgap to CI failing post nightly update.

The long-term fixes would be to fix this in https://github.com/0xmiden/miette, or contribute the `no_std` support upstream and update.


adr1anh:
This should soon be fixed, see https://github.com/rust-lang/rust/issues/147648

PR #2287, opened 2025-10-14
feat: add `EventName` for better event handling errors & unify SytemEvent ID derivation (#2150)

### LINKED ISSUE #2150: Host should resolve an event's name from when a handler fails
At the moment, if a precompile fails, the error only displays the EventID which is an arbitrary Felt. This is because Libraries only export lists of `(EventId, Handler)`, mainly because this is the information required by the Host. For the Host to be able to produce better error messages, the Library would need to export an Event Name instead of just the ID.

Moreover, an `EventName` type would also be able to enforce consistent name-spacing rules.

Extra context:
- https://github.com/0xMiden/miden-vm/pull/2145#discussion_r2342971711
- https://github.com/0xMiden/miden-base/issues/1563#issuecomment-3269322275

adr1anh:
**Summary**
Implements event name resolution for better diagnostics (fixes #2150) by separating identity from presentation: `EventId` remains a lightweight `Felt`, while a new `EventName` enum carries human-readable names and converts to/from IDs where appropriate.

**Highlights**
- Adds `EventName` with constructors for static, dynamic, system, and unknown events; derive an ID via `EventName::to_event_id()`.
- `SystemEvent` publishes canonical names via `name_str`/`event_name` and is tested to map enum discriminants directly to IDs.
- Hosts can surface names via `BaseHost::resolve_event(event_id)`, and the processor enriches errors with names or falls back to `EventName::unknown`.
- Handler registry stores `(EventName, handler)` so duplicate/reserved checks and diagnostics use readable names.
- Test utilities and handler wiring accept `EventName`, enabling clear, name-first usage across crates.

**What This Replaces**
- No embedded names in `EventId`.
- No `unsafe from_static` constructor.
- No `declare_event!` macro or lazy-init scaffolding.

**Breaking Changes**
- Error surfaces now include `EventName` instead of a raw `EventId` in relevant variants (e.g., `ExecutionError::EventError`).
- Handler registration APIs use `EventName` instead of `EventId`.
- Test helpers expecting event identifiers switch to `EventName`.

**Migration**
- When registering handlers, pass `EventName::new("my::event")` (or `EventName::from_string(...)`) instead of an `EventId`.
- When an `EventId` is required (e.g., stack comparisons), call `event_name.to_event_id()`.
- Update code that matched on error variants to handle `event: EventName` instead of `event_id: EventId`.
- For system events, use `SystemEvent::event_name()` to get an `EventName`, or `EventId::from(system_event)` for the ID.

**Why This Direction**
- It resolves prior concerns about equality/`Display` ambiguity by keeping IDs and names separate.
- It removes unsafe and macro-based const gymnastics in favor of simple, explicit APIs.
- It answers “how do I get names at runtime?” via `resolve_event`, improving error messages without forcing names into hot-path data.


core/src/event_id.rs
```diff
@@ -28,35 +36,73 @@ impl EventId {
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the ID correctly corresponds to `hash_string_to_word(name)[0]`.
+    /// Providing an incorrect ID can lead to event handler mismatches and incorrect program
+    /// behavior.
+    ///
+    /// **Prefer using the [`declare_event!`](crate::declare_event) macro instead**, which generates
+    /// both the constant and a compile-time test to validate the ID is correct.
+    pub const unsafe fn from_static(name: &'static str, id: u64) -> Self {
```
plafer: `unsafe` relates to memory safety, where the caller must adhere to some preconditions manually for the program not to have undefined behavior. This isn't the case here, so the `unsafe` keyword should be removed. We should also convert `# Safety` to e.g. `# Warning` or something.

Also, wouldn't `from_name_and_value()` or similar be a better name?

core/src/sys_events.rs
```diff
@@ -289,7 +315,9 @@ impl TryFrom<EventId> for SystemEvent {
 
 impl From<SystemEvent> for EventId {
     fn from(system_event: SystemEvent) -> Self {
-        Self::from_u64(system_event as u64)
+        // SAFETY: SystemEvent's event_name() is hardcoded to match the enum discriminant,
+        // so the ID is guaranteed to be correct for the name.
```
plafer: Similarly, this shouldn't be a `SAFETY` comment once `unsafe` is removed.

core/src/sys_events.rs
```diff
@@ -289,7 +315,9 @@ impl TryFrom<EventId> for SystemEvent {
...
+        unsafe { Self::from_static(system_event.event_name(), system_event as u64) }
```
plafer: `EventId::from_static()` claims that the (first element of the) blake3 hash of the name must be the `id`. But the IDs are 1 to 15, which are clearly not related to the hash. Why does this work then? (I don't understand the explanation as-is).

Also, if `from_static()` is only meant for system events, we could make it `pub(crate)` to force everyone else *not* to use it.

core/src/event_id.rs
```diff
@@ -65,12 +111,142 @@ impl PartialOrd for EventId {
...
+/// - A `pub const` [`EventId`] with the given name
+/// - A test module with the same name containing validation logic
+///
+/// # Note on Naming
+///
+/// The generated test module uses the constant's name. This is safe because Rust naming
+/// conventions naturally separate these: constants use `SCREAMING_SNAKE_CASE` while modules
+/// use `snake_case`. Conflicts would only occur if violating both conventions simultaneously.
+#[macro_export]
+macro_rules! declare_event {
```
plafer: If we keep this macro, I would rename it to `declare_top_level_event` or something, so that the name indicates that this isn't the only valid way to create an event, and that it shouldn't be used e.g. in a function body.

However, the alternative to this macro is to let users use [`lazy_static!`](https://docs.rs/lazy_static/latest/lazy_static/) as e.g.

```rs
lazy_static::lazy_static! {
    pub static ref SOME_EVENT_ID: EventId = EventId::from_name("some::event");
}
```

The pros of this approach is that `lazy_static!` is a well-known canonical solution to the "static initialization" problem (i.e. doesn't require the reader to go understand how our custom `declare_event!` works), and is arguably simpler (since it doesn't require running a test to make work). The downside is that we don't get a true `const`, and accesses to `SOME_EVENT_ID` requires an extra (hidden) atomic variable read. IMO though, since reading event ids is never going to be in a hot loop, the "performance drawback" is negligible.

What do you think?

core/src/event_id.rs
adr1anh: Agreed, the unsafe is a bit overkill, and the name could be better.

core/src/event_id.rs
```diff
@@ -28,35 +36,73 @@ impl EventId {
...
-        Self(digest_word[0])
+        Self {
+            id: digest_word[0],
+            name: Some(Cow::Owned(name_str)),
+        }
     }
 
-    /// Creates a new event ID from a [`Felt`].
+    /// Creates a new event ID from a [`Felt`] without an associated name.
     pub const fn from_felt(value: Felt) -> Self {
```
PhilippGackstatter:
```suggestion
    pub const fn from_felt(event_id: Felt) -> Self {
```

Nit: Descriptive variable names make things clearer for callers (who will see that name in docs or in auto-completed code).

core/src/event_id.rs
PhilippGackstatter: I usually use `_unchecked` for such cases, clearly indicating at call-sites that a `SAFETY` comment should be expected and explain why bypassing the check is ok.

I thought that this was common Rust usage, but I can't back that up now actually and it looks like SAFETY comments are expected for unsafe code, and maybe fine in other places? Then again, I'm asking myself how else logic invariants or `expect`s are supposed to be justified.

Nit: Use `id` instead of `value` to be more precise.

core/src/event_id.rs
```diff
@@ -65,12 +111,142 @@ impl PartialOrd for EventId {
...
 
 impl Display for EventId {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
-        core::fmt::Display::fmt(&self.0, f)
+        match &self.name {
+            Some(name) => write!(f, "{name}"),
+            None => write!(f, "unknown event"),
+        }
+    }
+}
```
PhilippGackstatter, line 95: Nit: I'm not sure if it's a good idea to implement `Display` when a useful implementation only exists for some of the instances. I think I would implement `as_str(&self) -> Option<&str>` instead, giving callers the ability to differentiate the "has name" and "has no name" cases.

core/src/event_id.rs
PhilippGackstatter: I think I agree that the lazy static approach is more familiar and a bit more transparent than `declare_event`, although it is a really cool idea for declaring it in a const way. In miden-base at least, we extract `EventId`s in build.rs, and so we can make sure they are consistent in that way.

(Instead of lazy static, I think `LazyLock` is now the way to go as it is defined in the stdlib, and we also have our own no-std compatible impl in `miden-utils-sync`.)

core/src/event_id.rs
```diff
@@ -8,10 +9,17 @@ use crate::{Felt, utils::hash_string_to_word};
...
+///
 /// While not enforced by this type, the values 0..256 are reserved for
 /// [`SystemEvent`](crate::sys_events::SystemEvent)s.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct EventId(Felt);
+#[derive(Debug, Clone)]
+pub struct EventId {
+    id: Felt,
+    name: Option<Cow<'static, str>>,
+}
```
PhilippGackstatter: Combining name and ID, and a `const` context needs quite a bit of machinery to work, e.g. `declare_event!`. What is the motivation for needing the ID at compile time or in a constant context? Would it not work to have a `struct EventName(Cow<'static, str>>)` type, and a `EventName::to_id(&self) -> EventId` function to compute the event ID when necessary? Maybe the `event` can still contain the name optionally, but the thing that is declared in `const` contexts would be `EventName` instead of `EventId`.

Not saying it's the right approach here, as I could be missing some context. This is the approach I've taken in https://github.com/0xMiden/miden-base/pull/1987 where `SlotName` is a string wrapper and `SlotNameId` is the derived ID. This solution has its own "problems", namely that we need to sort another type (`NamedStorageSlot`) by the `SlotNameId`, and to make the `Ord` impl not constantly hash the name, the `SlotNameId` is cached in `NamedStorageSlot`. This is just to illustrate that this is not a perfect solution either, but if you don't have some hot loop use cases like this one, this solution could work fairly nicely here.

Another approach I was now thinking about is to redefine `SlotName`/`EventName` to be:

```rust
pub struct SlotName {
    name: Cow<'static, str>,
    id: Option<SlotNameId>,
}
```

And in a non-const context, the `id` could be initialized at construction time while it would be `None` for `const` constructors. The `SlotName::id` function could then return the cached value or compute it, and there could be another function to compute it and cache it internally (which requires `&mut self`). Have to think this through some more though before 🙂 

Anyway, just a couple of loud thoughts.

Edit: Ended up trying the cached approach and it is reasonable, but I slightly prefer the old approach. [Diff for reference](https://github.com/0xMiden/miden-base/compare/pgackst-named-storage-slots...pgackst-named-storage-slots-cached-id).

core/src/event_id.rs
```diff
@@ -28,35 +36,73 @@ impl EventId {
     ///
     /// This ensures that identical event names always produce the same event ID, while
     /// providing good distribution properties to minimize collisions between different names.
-    pub fn from_name(name: impl AsRef<str>) -> Self {
-        let digest_word = hash_string_to_word(name.as_ref());
+    ///
+    /// The event name is stored and will be displayed in error messages.
+    pub fn from_name(name: impl Into<String>) -> Self {
```
PhilippGackstatter: Looking at `FastProcessor::op_emit`, the name is never present on `EventId`s passed to the host's `on_event` function, right? Is there a way to change this or make it otherwise possible for miden VM users to lookup the name from an ID? This would be just for debugging purposes or error messages, so in branches where performance is not critical.

core/src/event_id.rs
```diff
@@ -65,12 +111,142 @@ impl PartialOrd for EventId {
...
+/// 3. Update the declaration with the correct ID
+/// 4. The test will pass, confirming the ID is correct
+///
+/// # Example
+///
+/// ```rust,ignore
+/// use miden_core::declare_event;
+///
+/// // Step 1: Write with placeholder
+/// declare_event!(MY_EVENT, "miden::foo::bar", 0);
```
drahnr: When would we want to call those? I.e. do we want a minimal sharable crate that exposes just event constants to be shared across all repositories to remain in sync, so both emission and consumption stay in sync? Currently we to check in a `build.rs` what is emitted vs what is covered, which is obviously somewhat incomplete/brittle.

PhilippGackstatter (commented):
Looks good!

I left a few comments/questions regarding the structure.

core/src/event_id.rs
```diff
@@ -8,10 +9,17 @@ use crate::{Felt, utils::hash_string_to_word};
...
+/// debugging. The name can be either a static string (for compile-time defined events) or an
+/// owned string (for runtime-generated events).
+///
 /// While not enforced by this type, the values 0..256 are reserved for
 /// [`SystemEvent`](crate::sys_events::SystemEvent)s.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct EventId(Felt);
+#[derive(Debug, Clone)]
+pub struct EventId {
+    id: Felt,
```
drahnr: in which cases would we want only the `id` field to be populated?

adr1anh: Internally when we get the event id from the stack when handling an event, we only have a felt. This is a good point though, as we need to make sure that the error we return is a named event. 

Thinking about it more, I think it makes sense to have a separate NamedEvent to avoid confusion. I'll put this PR as draft to fix it.

adr1anh:
Thank you all for your comments!

I put this back into draft more as I forgot to consider the flow when we are generating an event purely from a Felt.

core/src/event_id.rs
plafer: > I thought that this was common Rust usage, but I can't back that up now actually and it looks like SAFETY comments are expected for unsafe code, and maybe fine in other places? Then again, I'm asking myself how else logic invariants or expects are supposed to be justified.

I've only ever seen it used with `unsafe` code, and the convention is documented [here](https://std-dev-guide.rust-lang.org/policy/safety-comments.html). Basically these "safety" comments are associated with "memory safety", and IMO it's confusing if they're used for documenting preconditions that don't relate to memory safety.

> Then again, I'm asking myself how else logic invariants or expects are supposed to be justified.

AFAIK there isn't a standard way to do it, but we could come up with our own keyword e.g. `# PRECONDITIONS` (except better).

> Nit: Use id instead of value to be more precise.

👍

core/src/event_id.rs
plafer: > (Instead of lazy static, I think LazyLock is now the way to go as it is defined in the stdlib, and we also have our own no-std compatible impl in miden-utils-sync.)

Ah yes! Good point, I keep forgetting about that one.

However my opinion stands that I think `static` + `LazyLock` is a more canonical approach than a custom macro, but won't die on that hill either. If more people prefer the `declare_event!` macro then that works for me.

core/src/event_id.rs
plafer: *More loud thoughts*

Right, I assumed that the only benefit of doing everything statically was to avoid running a `blake3` hash every time we want to get the ID, but that might be fine given that it would simplify the implementation, and that events are relatively rare. BUT, between the "run the hash at runtime" and the `LazyLock` solutions, I still prefer the `LazyLock`.

...*And* probably to allow for events to be defined without names - however I'm not sure if that's ever desirable.

> The SlotName::id function could then return the cached value or compute it, and there could be another function to compute it and cache it internally (which requires &mut self). 

We wouldn't require a `&mut self` if we used interior mutability, e.g. a `RefCell<Option<SlotNameId>>` (or `Mutex`/`RwLock` if we need it to be `Send`).

PhilippGackstatter: > We wouldn't require a `&mut self` if we used interior mutability, e.g. a `RefCell<Option<SlotNameId>>` (or `Mutex`/`RwLock` if we need it to be `Send`).

Yeah, I'd try to avoid `RefCell` because for sure this type will end up in a `Send` `Future` somewhere, and then we'll have to change it anyway :sweat_smile:

I tried changing `SlotName` to this:

```rust
pub struct SlotName {
    name: Cow<'static, str>,
    id: crate::utils::sync::rw_lock::RwLock<Option<SlotNameId>>,
}
```

I also kept the `const` constructors like `SlotName::from_static_str("abc::def")`.

This compiles, but triggers the following lint whenever declaring a `const NAME: SlotName`: https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const. This lint would probably also be triggered if `RefCell` was used.

So, based on this, and if inlining the `id` into the name struct is desired, it indeed makes more sense to avoid `const` altogether and use a `static` declaration + `LazyLock`.

core/src/event_id.rs
adr1anh: Yes, this was the main missing piece in this PR. I added a new `Host` method that resolves `EvenId -> EventName`, which later on could look up names stored in the MastForest. For now though, the DefaultHost implementation can resolve them from the registry. 

From my understanding, event handling is not necessarily performance critical (which is why we also decided to let it take 3 cycles instead of 1), but I may be wrong here.

adr1anh:
This is a general reply to earlier review threads. We’ve shifted the design to keep event IDs small and move human-readable details into a separate type. Many previous comments about equality, `Display`, unsafe constructors, and macro choices are addressed by this new direction.

**What Changed**
- `EventId` stays a plain `Felt`. Names are represented by a new `EventName` enum with constructors for static, dynamic, system, and unknown cases (`core/src/event_id.rs`).
- `SystemEvent` now provides canonical names via `name_str`/`event_name`, with tests that verify discriminant-to-ID mapping. The old unsafe static path is gone (`core/src/sys_events.rs`).
- Hosts expose `BaseHost::resolve_event`, letting the processor attach a name to errors when available, or use `EventName::unknown` if only an ID is present (`processor/src/host/mod.rs`, `processor/src/fast/basic_block.rs`).
- The handler registry stores `(EventName, handler)` pairs, so duplicate/reserved checks and diagnostics reference meaningful names (`processor/src/host/handlers.rs`, `processor/src/errors.rs`).
- Test utilities and handler wiring accept `EventName`, so downstream crates can register named events without custom macros or lazy initialization scaffolding (`crates/test-utils/src/lib.rs`).

**Feedback comments**
- Equality/`Display` surprises: resolved by separating identity (`EventId`) from presentation (`EventName`). IDs don’t carry names anymore.
- Unsafe constructor and “Safety” docs: removed; system IDs are defined by enum discriminants with explicit tests and clear docs.
- Macro vs `LazyLock`: no custom macro needed. Static names use `EventName::new`, and IDs are derived as needed; no lazy global required. The VM doesn't require constant IDs. Downstream users still have the ability to define constant names/ids if required (as is done in `miden-base` using `build.rs`)
- Name lookup during diagnostics: handled via `BaseHost::resolve_event` so errors show names when the host knows them, or degrade to `unknown` when not.
- System events and hashes: clarified that system IDs are discriminants; human-readable names come from `SystemEvent::name_str`.

core/src/event_id.rs
```diff
@@ -86,12 +101,129 @@ impl Ord for EventId {
...
+    miden_test_serde_macros::serde_test(winter_serde(true))
+)]
+pub enum EventName {
+    /// A named event
+    Event(Cow<'static, str>),
+    /// A system event (IDs 0-255)
+    System(EventId),
+    /// An unknown event with only an ID available
+    Unknown(EventId),
+}
```
plafer: What I don't like about this PR is not so much the implementation details, but with how it forces me to think about events.

**How I want to think about events**: `name --(blake3)--> id`

**How this system forces me to think about events**:
- if *system event*, then `name --(special case)--> id`
- if named, then `name --(blake3)--> id`
- otherwise, `name -> id` map doesn't exist

For one, I've argued for this in the past, but I still believe adding a special case to how system events are derived adds unnecessary complexity to the system (as outlined above). Additionally, this PR introduces a new concept: an event is allowed not to have a name.

**My question**: Why don't we just enforce the `name --(blake3)--> id` everywhere, and have just one way to think about events?

Concretely speaking, `BaseHost::resolve_event(EventId) -> Option<String>` would now return `Option<String>`, meaning "all events have a name, but I don't know the name of that specific one". For readability, we could define `EventName(String)` (or `EventName(Cow<'static, str>)`).

> Compare this with the current `EventName::Unknown`, which tells me that conceptually speaking, events are allowed to be unnamed.

Then, we'd have a `LazyLock<BTreeMap<EventId, EventName>>` for system events, where we name our system events as e.g. `::miden::sys::event_merkle_node_merge`. To get a `SystemEvent` for the purposes of `handle_system_event()`, we'd do `event_id -> name -> SystemEvent::from(name)`.

What I'm pushing for here is really to simplify how we model events. I want to be able to explain events to someone as "all events have a name, and we derive the event ID by blake3 hashing the name." I can't do that from how events are modeled in this PR, and therefore IMO we created complexity that doesn't need to be there.

adr1anh: You raise a good point. From the start of this PR I was under the impression that having defining SystemEvents as fixed constants was something we needed but it's becoming clearer that it introduces more complexity overall. When making events read the ID from the stack and taking 3 cycles instead of 1, the argument was that events are rare enough that the potential performance hit was worth the overall simplification. 

While implementing this feature, I realized we weren't actually making use of constant event definitions, and that the overhead of regenerating event IDs is negligible since it only happens during initialization or failure cases. The functionality is still there for downstream use cases since both EventName and EventId can be defined at compile time if necessary. 

Regarding the `EventName::Unknown`, I think we can remove it as it was mainly used for error reporting and `Option<EventName>` is more explicit. 

I'll let others chime in before changing the PR.

processor/src/host/default.rs
```diff
@@ -149,6 +150,10 @@ where
 
     /// Handles the failure of the assertion instruction.
     fn on_assert_failed(&mut self, _process: &ProcessState, _err_code: Felt) {}
+
+    fn resolve_event(&self, event_id: EventId) -> Option<&EventName> {
```
drahnr, line 154: Practically speaking, in `miden-base` this would imply carving out a well-known prefix for system events in https://github.com/0xMiden/miden-base/pull/1954/files#diff-43c3f3d1c06c35941320e0a796d34b4d86ab6bea225526a1a52510bc3799ce80R840 for name resolution, and making sure they exist in this table, correct me if I am wrong

drahnr:
I generally do like the approach, although I agree that the system events could be flattened/avoid the special casing, but I am uncertain about the incompatibilities and consequences.

core/src/event_id.rs
huitseeker: I think there is a possible convergence here based on 3 principles (the string syntax is not important):

- All events follow "name → blake3 → EventId" rule
  - System events: `"miden::sys::merkle_node_merge"` → blake3 → ID
  - User events: `"my::custom::event"` → blake3 → ID
  - No unnamed events: every EventId can be resolved to a canonical name

- System events reserved by namespace prefix, not special ID ranges
  - `"miden::sys::"` prefix reserves system event namespace
  - Prevents naming conflicts while maintaining unified hashing rule

- Performance via compile-time constants verified by tests
  ```rust
  const SYS_EVENT_MERKLE_NODE_MERGE: u64 = [computed_hash]; // verified by test
  const SYS_EVENT_MERKLE_NODE_TO_STACK: u64 = [computed_hash]; // verified by test
  Tests ensure each constant matches blake3("miden::sys::{event_name}")[0]. This gives both performance (constants) and correctness (verified semantics).

adr1anh: I agree with all the points above. We were previously using the approach from the third point for regular events, which was why this PR originally included a macro for simplifying the derivation of the test. However, the VM doesn't require this kind of optimization since event handling is not necessarily in the hot-path. 

However, the methodology is still available for downstream users if necessary.

processor/src/host/default.rs
adr1anh: We still don't impose any namespacing requirements, the library is free to choose it. But in miden-base, you'd want a table (dynamic, as resolving events only happens in failure paths and not performance sensitive) to lookup the name.

adr1anh:
I've gone ahead and made `SystemEvent`s follow the same convension as normal events. It uses a constant lookup table (array) which should be fast for most use cases. Overall I think this is a lot simpler conceptually. Happy to revert if others have different opinions.

core/src/event_id.rs
```diff
@@ -86,12 +82,84 @@ impl Ord for EventId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+
+    /// Returns the [`EventId`] for this event name.
+    ///
+    /// The ID is computed by hashing the name using blake3.
+    pub fn to_event_id(&self) -> EventId {
+        EventId::from_name(self.as_str())
+    }
```
plafer, line 145: 🏄🎉🥳

core/src/sys_events.rs
```diff
@@ -301,71 +328,222 @@ impl crate::prettier::PrettyPrint for SystemEvent {
...
-            | SystemEvent::ILog2
-            | SystemEvent::MemToMap
-            | SystemEvent::HdwordToMap
-            | SystemEvent::HdwordToMapWithDomain
-            | SystemEvent::HqwordToMap
-            | SystemEvent::HpermToMap => {},
-        };
+            // Verify name has correct "sys::" prefix
+            assert!(
+                entry.name.starts_with("sys::"),
```
plafer, line 505: nit: users could realistically use `sys::` as a namespace for their events. Should we make our events start with `miden::sys::` or `miden::vm::sys::` - or basically a namespace that only the VM is likely to use?

processor/src/host/default.rs
```diff
@@ -169,7 +174,7 @@ where
 
         // EventError is a `Box` so we can define the error anonymously.
```
plafer:
```suggestion
        // EventError is a `Box<dyn Error>` so we can define the error anonymously.
```

stdlib/src/handlers/smt_peek.rs
```diff
@@ -7,16 +7,13 @@
...
 };
 use miden_processor::{AdviceMutation, EventError, ProcessState};
 
-/// Qualified event name for the `smt_peek` event.
-pub const SMT_PEEK_EVENT_NAME: &str = "stdlib::collections::smt::smt_peek";
-/// Constant Event ID for the `smt_peek` event, derived via
-/// `EventId::from_name(SMT_PEEK_EVENT_NAME)`.
-pub const SMT_PEEK_EVENT_ID: EventId = EventId::from_u64(3580205917336794987);
+/// Event name for the smt_peek operation.
+pub const SMT_PEEK_EVENT_NAME: EventName = EventName::new("stdlib::collections::smt::smt_peek");
```
plafer, line 16: A breath of fresh air

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@
 - [BREAKING] Add precompile requests to proof ([#2187](https://github.com/0xMiden/miden-vm/issues/2187)).
 - `after_exit` decorators execute in the correct sequence in External nodes in the Fast processor ([#2247](https://github.com/0xMiden/miden-vm/pull/2247)).
 - Remove O(n log m) iteration in parallel processor (#[2273](https://github.com/0xMiden/miden-vm/pull/2273)).
+- [BREAKING] `BaseHost` now exposes `resolve_event` so hosts can provide event names for diagnostics. Support `NamedEvent` ([#2150](https://github.com/0xMiden/miden-vm/issues/2150)).
```
plafer: Should update this now (e.g. `NamedEvent` is outdated)

core/src/sys_events.rs
adr1anh: I chose to avoid extra restrictions as we don't really have strong enforcement of name spacing. For now, nothing really bad would happen if a user did chose the "sys" namespace.

CHANGELOG.md
adr1anh:
```suggestion
- [BREAKING] `BaseHost` now exposes `resolve_event` so hosts can provide event names for diagnostics. Unify `SystemEvent` ID derivation ([#2150](https://github.com/0xMiden/miden-vm/issues/2150)).
```

PR #2259, opened 2025-10-04
feat: introduce MastNode builders (1/4)

huitseeker:
This PR implements a builder pattern for `MastNode`s, eliminating the create-mutate-add anti-pattern used throughout the repo with `before_enter`/`after_exit` decorators.

### Changes

- builder pattern with straightforward API: All builders support with_before_enter(), with_after_exit(), and build() methods,
- all direct `MastNode` constructors are now `pub(in crate::mast)`, favoring public builders,
- eliminated create-mutate-add patterns,
- `add_node()` calls replaced with specific `add_block()`, `add_join()`, etc. methods where appropriate, for reduced verbosity,
- the footprint of mutablility in `MastNode`s shrinks: the only two places using `before_enter`/`after_exit` mutation meaningfully are serialization & assembler.

<details><summary>Technical Details</summary>

### Before
```
let mut block = BasicBlockNode::new(ops, decorators).unwrap();
block.append_before_enter(&before_ids);
block.append_after_exit(&after_ids);
let id = mast_forest.add_node(block).unwrap();
```

### After
```
let id = mast_forest.add_block_with_decorators(ops, decorators, &before_ids, &after_ids).unwrap();
// Or using builder:
let id = BasicBlockNodeBuilder::new(ops, decorators)
    .with_before_enter(before_ids)
    .with_after_exit(after_ids)
    .build().unwrap();
```
</details>

### TL;DR & next steps

*External* code now has two clean options for node construction:
1. Specific `MastForest` methods for simple cases: `add_block()`, `add_join()`, etc.
2. Builder pattern for complex scenarios with decorators: `BasicBlockNodeBuilder::new(...).with_before_enter(...).build()`.

**Next**:
This established the pattern before adding more complex forest-building features in subsequent PRs.

PR #1539, opened 2024-10-21
Rename `EqHash` to `MastNodeFingerprint`

PhilippGackstatter:
## Describe your changes

Renames `EqHash` to `MastNodeFingerprint` as a follow-up to #1534 and moves the fingerprint to its own module.

## Checklist before requesting a review

- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.

PR #2291, opened 2025-10-15
MastNode builder extension (3/4)

### LINKED ISSUE #2243: test: improve Arbitrary for MastForest to include other block types

huitseeker:
This PR continues the builder pattern work for MAST nodes from PRs #2259 and #2274. In the previous PRs, we introduced builder patterns for creating MAST nodes and extended them with forest-building capabilities. This current PR focuses on creating a unified system where builders can handle all node operations consistently.

The main changes in this PR are:

- created a `MastNodeBuilder` enum that wraps all the different node builder types, giving you one simple way to work with any node type,
- correspondingly, created a custom enum dispatch system called `enum_thispatch` to replace the external `enum_dispatch` dependency, giving us more control over how the builders work (the builder trait, `MastForestContributor`, now takes an associated type which requires this),
- added the ability to convert nodes back to builders using `to_builder()`, so you can modify existing nodes and rebuild them,
- moved key operations like `remap_children` and fingerprint computation into the builders, making it possible to do these operations during construction rather than after (the fingerprint computation is a strict duplication, but it's proven equivalent by proptests, and will eventually be removed — incidentally this fixes #2243 at the builder level ),
- made `add_node` private to the crate, so external code must use the builder patterns instead of creating nodes directly (mutability footprint shrinks),

The next PR will build on this foundation by completely removing the old `add_node` method and adding more advanced features like digest forcing and complete the migration to builder-based decorator management. This will complete the transition to using builders, and more importantly, unlock immutable `MastForest`s post assembly.

### DIFF
--- a/core/src/mast/node/mod.rs
+++ b/core/src/mast/node/mod.rs
@@ -6,7 +6,8 @@ pub use basic_block_node::{
     BATCH_SIZE as OP_BATCH_SIZE, BasicBlockNode, BasicBlockNodeBuilder, DecoratorOpLinkIterator,
     GROUP_SIZE as OP_GROUP_SIZE, OpBatch, OperationOrDecorator,
 };
-use miden_utils_core_derive::{FromVariant, MastNodeExt};
+use derive_more::From;
+use miden_utils_core_derive::MastNodeExt;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -89,7 +90,7 @@ pub trait MastNodeExt {
 // MAST NODE
 // ================================================================================================
 
-#[derive(Debug, Clone, PartialEq, Eq, MastNodeExt, FromVariant)]
+#[derive(Debug, Clone, PartialEq, Eq, MastNodeExt, From)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[mast_node_ext(builder = "MastNodeBuilder")]
 pub enum MastNode {
--- a/crates/utils-core-derive/src/lib.rs
+++ b/crates/utils-core-derive/src/lib.rs
@@ -5,7 +5,6 @@
 //! - `MastNodeExt` derive macro: generates MastNodeExt trait implementations for enums
 //! - `MastForestContributor` derive macro: generates MastForestContributor trait implementations
 //!   for enums
-//! - `FromVariant` derive macro: generates `From<VariantType> for EnumType` implementations
 //!
 //! This crate provides enum dispatch functionality with:
 //! - Zero-cost enum dispatch without external dependencies
@@ -296,44 +295,6 @@ pub fn derive_mast_forest_contributor(input: TokenStream) -> TokenStream {
     TokenStream::from(trait_impl)
 }
 
-/// Derive From implementations for converting each variant type to the enum.
-#[proc_macro_derive(FromVariant)]
-pub fn derive_from_variant(input: TokenStream) -> TokenStream {
-    let input = parse_macro_input!(input as DeriveInput);
-
-    let enum_name = &input.ident;
-    let generics = &input.generics;
-
-    // Parse the data to ensure it's an enum
-    let enum_data = match &input.data {
-        Data::Enum(data) => data,
-        _ => panic!("FromVariant can only be derived for enums"),
-    };
-
-    // Extract variant information
-    let variants: Vec<_> = enum_data.variants.iter().collect();
-    let variant_names: Vec<_> = variants.iter().map(|v| &v.ident).collect();
-    let variant_types: Vec<_> = variants.iter().map(|v| extract_variant_type(v)).collect();
-
-    // Generate From implementations
-    let from_impls =
-        variant_names.iter().zip(variant_types.iter()).map(|(variant, variant_type)| {
-            quote! {
-                impl From<#variant_type> for #enum_name #generics {
-                    fn from(node: #variant_type) -> Self {
-                        #enum_name::#variant(node)
-                    }
-                }
-            }
-        });
-
-    let expanded = quote! {
-        #(#from_impls)*
-    };
-
-    TokenStream::from(expanded)
-}
-
 /// Generate MastForestContributor trait implementation for enum dispatch
 fn generate_mast_forest_contributor_impl(
     enum_name: &Ident,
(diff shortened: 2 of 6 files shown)


crates/utils-core-derive/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+        &trait_path,
+        &variant_names,
+        &variant_fields,
+    );
+
+    TokenStream::from(trait_impl)
+}
+
+/// Derive From implementations for converting each variant type to the enum.
+#[proc_macro_derive(FromVariant)]
```
plafer: Couldn't we use `derive_more::From` instead?

core/src/mast/node/mod.rs
```diff
@@ -83,14 +79,19 @@ pub trait MastNodeExt {
 
     /// Returns the domain of this node.
     fn domain(&self) -> Felt;
+
+    /// Converts this node into its corresponding builder, reusing allocated data where possible.
+    type Builder: MastForestContributor;
+
+    fn to_builder(self) -> Self::Builder;
```
plafer, line 87: Couldn't we change `to_builder()` to be `fn to_builder(self) -> impl MastForestContributor)`? And then IIUC we could go back to using `enum_dispatch`?

crates/utils-core-derive/src/lib.rs
huitseeker: Probably, though I didn't look into it. 

I'm replacing the `enum_dispatch` dependency here, it would be sad to re-introduce one for a menial side-macro.

core/src/mast/node/mod.rs
huitseeker: > Couldn't we change to_builder() to be fn to_builder(self) -> impl MastForestContributor)? 

```
  type Builder: MastForestContributor;
  fn to_builder(self) -> Self::Builder;
```
is the precise de-sugaring of `fn to_builder(self) -> impl MastForestContributor);` in the strong sense of the word, that is the later is a syntactic facility for the former. 

> IIUC we could go back to using enum_dispatch?

We could not: https://gitlab.com/antonok/enum_dispatch/-/issues/50

There's also a fundamental algebraic ambiguity that explains this. The contract is: _if I give you a way to compute N > 1 implementations of trait T, please give me a single type that can contain those N values and be an implementation of trait T_. It so happens that you can naturally define this within-macro by defining the co-product of all N alternatives. The macro *could* define that type in an ad-hoc manner, and that would be *the thing you want* if that was *universal* in some sense, i.e. if Rust had anonymous coproducts, rather than named enums — and therein lies the rub: we don't want some (heretofore non-existent) naturally-defined co-product, we want a specific enum that has a pre-defined meaning elsewhere in the code, `MastNodeBuilder` in this case. That's why we rely on (and require that) every variant of enum `E` on which we `#[derive(MastNodeExt)]` (is) satisfying not only :
```rust
<Variant as MastNodeExt>::Builder: MastForestContributor;
```
but in fact:
```rust
<Variant as MastNodeExt>::Builder: MastForestContributor + Into<<E as MastNodeExt>::Builder>;
```

There's other ways to solve this, e.g. the `either` crate offers another take on co-products. `enum_dispatch` does not choose between the many ways to address this. We do, specifically with the convention above.

plafer: The following compiles and runs properly:

```rs
use enum_dispatch::enum_dispatch;

trait Builder {}

struct JoinNodeBuilder;
impl Builder for JoinNodeBuilder {}

#[enum_dispatch]
trait MastForestContributor {
    fn to_builder(&self) -> impl Builder;
}

struct JoinNode;

impl MastForestContributor for JoinNode {
    fn to_builder(&self) -> impl Builder {
        println!("in JoinNode's to_builder");
        JoinNodeBuilder
    }
}


#[enum_dispatch(MastForestContributor)]
enum MastNode {
    Join(JoinNode)
}

#[test]
fn test() {
    let node = MastNode::Join(JoinNode);
    let _builder = node.to_builder();
}
```

I confirm that associated types are still unsupported by `enum_dispatch`. However, the linked issue is from 2021, before the ["return-position `impl Trait` in traits"](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/) syntactic form was introduced, and it looks to me like `enum_dispatch` handles that properly. So my suggestion was essentially that if we used `-> impl Builder`, we wouldn't need to roll our own macro, which IMO is preferable given how widely used `enum_dispatch` is.

crates/assembly/src/mast_forest_builder.rs
plafer, line 403: Not from this PR, but I would rename `hash_by_node_id` -> `fingerprint_by_node_id` (and all other places where the `hash_by_node_id` name was reused in the new PRs). This would make it clear from the name that it's the inverse map of `node_id_by_fingerprint`.

core/src/mast/node/mod.rs
huitseeker: Thanks. The following does not compile.

```rust
use enum_dispatch::enum_dispatch;

trait Builder {}

struct JoinNodeBuilder;
impl Builder for JoinNodeBuilder {}

struct SplitNodeBuilder;
impl Builder for SplitNodeBuilder {}

#[enum_dispatch]
trait MastNodeExt {
    fn to_builder(&self) -> impl Builder;
}

struct JoinNode;

impl MastNodeExt for JoinNode {
    fn to_builder(&self) -> impl Builder {
        println!("in JoinNode's to_builder");
        JoinNodeBuilder
    }
}

struct SplitNode;

impl MastNodeExt for SplitNode {
    fn to_builder(&self) -> impl Builder {
        println!("in SplitNode's to_builder");
        SplitNodeBuilder
    }
}

#[enum_dispatch(MastNodeExt)]
enum MastNode {
    Join(JoinNode),
    Split(SplitNode),
}

#[test]
fn test() {
    let node = MastNode::Join(JoinNode);
    let _builder = node.to_builder();
}
```

Error message [here](https://gist.github.com/huitseeker/5dc80da17604f60350b99e8bc8ae8397)

plafer: Ah I get it now! Thank you bearing with me 😅

crates/assembly/src/mast_forest_builder.rs
huitseeker: Agreed. Due to the effect of renames on a stack of PRs, I'd rather do this (and other renames) in a small cleanup PR after we've merged a good chunk.

crates/utils-core-derive/src/lib.rs
huitseeker: Thanks! I noticed `derive_more` is already used in the workspace, and used it here.

adr1anh (approved):

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -985,4 +987,109 @@ impl MastForestContributor for BasicBlockNodeBuilder {
...
+            bytes_to_hash.extend(forest[*decorator_id].fingerprint().as_bytes());
+        }
+
+        // Hash after_exit decorators last
+        for decorator_id in &self.after_exit {
+            bytes_to_hash.extend(forest[*decorator_id].fingerprint().as_bytes());
+        }
+
+        // Add any `Assert`, `U32assert2` and `MpVerify` opcodes present, since these are
+        // not included in the MAST root.
```
Al-Kindi-0, line 1024: This is probably not from this PR so I wonder, @bobbinth , is this accurate as stated?

huitseeker: It's not. Last edited by @PhilippGackstatter in #1539, introduced by @plafer in #1466 :
https://github.com/0xMiden/miden-vm/pull/1466/files#diff-e8de68365191c0d8f682c77290041c4c907051d34135f3864d5acc1209897d0aR460-R461

Issue #2295, opened 2025-10-16
Incorrect decorator index mapping due to NOOPs


plafer:
I believe the core issue from #2181 hasn't been fully resolved with #2195. Consider the following program:

```
begin
    push.4
    push.8
    drop
    drop

    repeat.12
        push.42
    end

    debug.stack.15  # <=== incorrect
    dropw dropw dropw dropw
end
```

We would expect the `debug.stack` decorator to print out the first *12* elements to be `42`, but it only prints 11:

```
Stack state in interval [0, 14] before step 21:
├──  0: 42
├──  1: 42
├──  2: 42
├──  3: 42
├──  4: 42
├──  5: 42
├──  6: 42
├──  7: 42
├──  8: 42
├──  9: 42
├── 10: 42
├── 11: 0
├── 12: 0
├── 13: 0
├── 14: 0
└── (12 more items)
```

I'm not sure if this is completely minimal, or what exactly is the edge case that makes the current implementation fail. I double-checked that this was also failing after [bd9564c](https://github.com/0xMiden/miden-vm/commit/bd9564c03a9841e66abb314d6f210831dcdd253f).

PR #2296, opened 2025-10-16
feat: allow `CALL` and `DYNCALL` in a `SYSCALL` context

### LINKED ISSUE #1681: Call should be allowed in kernel modules
While looking into #1680, I noticed that the assembler [rejects](https://github.com/0xPolygonMiden/miden-vm/blob/0db3e00e9f959b87902a6c11d3d7f7730f4e6dfd/assembly/src/sema/passes/verify_invoke.rs#L116-L118) `syscall`s and `call`s in kernel modules. If I am not mistaken, this is too strict: we should allow calls/syscalls in the root context if we're not processing a syscall (i.e. `in_syscall` flag is false). It is also inconsitent that `dyncall` is currently allowed, but not `call` and `syscall`.

Hence, I think we should remove that check from the assembler.

plafer:
Closes #1681

From https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3387252809:

>To recap, the main issue with allowing `CALL` from `SYSCALL` is to properly manage the `fmp` register in this example:
>
```
operation    ctx         fmp
BLOCK         0          FMP_MIN
CALL          2          FMP_MIN     // fmp reset in new execution context

// FIRST SYSCALL
SYSCALL       0          SYSCALL_FMP_MIN
BLOCK         0          SYSCALL_FMP_MIN
FMPUPDATE     0          SYSCALL_FMP_MIN + 4 // emulating situation where a procedure was exec'd
CALL          10         FMP_MIN

// SECOND SYSCALL
// This is the interesting one: we need the fmp to be restored to its value when servicing the first system call
SYSCALL       0          SYSCALL_FMP_MIN + 4

...
```
>
>Specifically, the issue with allowing `CALL` in a syscall context is that the call'd procedure can then again `SYSCALL` once more. In the current system, the `fmp` register would be reset to `SYSCALL_FMP_MIN`, possibly overwriting the values of the procedure in the first syscall (which might be accessed when eventually returning to the first syscall context).

The approach taken in this PR to solve this problem is to get rid of the separate `SYSCALL_FMP_MIN` zone, and instead store the fmp in memory. Hence, when we re-enter execution context 0 (due to a syscall), the fmp value is naturally restored to what it was the last time we left context 0 (from either a `CALL` or `DYNCALL`). Specifically, the high-level changes are
- Assembler: on program start, inject a sequence of operations to initialize the `FMP` (stored at address `FMP_ADDR = 2^32 - 1`)
- on `CALL` and `DYNCALL`: write to memory in the new context, to initialize `FMP_ADDR = 2^32 - 1`
  - This requires updating the processor to do that write, as well as the chiplet bus
- Assembler: when resolving the address of a procedure local, instead of using `FMPADD`, load the fmp value from memory

It is also worth noting that syscalls were also previously disallowed in a syscall context. In this PR we allow them, as there is no difference between
- a syscall from context 0 at the start of the program (previously allowed), and 
- a syscall from context 0, where context 0 was entered due to a SYSCALL.

In a subsequent PR, I will remove the now unused `FMPADD` & `FMPUPDATE` operations, as well as the now-unused `fmp` (and `in_syscall`) column.

Left to do:
- [x] fix remaining 4 tests (blocked on #2295)
- [x] Remove the checks that disallow `CALL` from a `SYSCALL` context (both processor and assembler)
- [x] Update docs
  - Memory: document how the new `fmp` address is "reserved"
  - Update bus equation (with new memory writes from CALL and DYNCALL)
  - Update docs about how accessing a memory local incurs a memory read
- [x] Agree on what to do with `in_syscall` column (currently incorrect) - see https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3415807040
  - Conclusion: we no longer need the `in_syscall` column

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added `std::mem::pipe_double_words_preimage_to_memory`, a version of `pipe_preimage_to_memory` optimized for pairs of words ([#2048](https://github.com/0xMiden/miden-vm/pull/2048)).
 - Added missing endianness-aware memory instructions (`mem_loadw_be`, `mem_loadw_le`, `mem_storew_be`, `mem_storew_le`) to Instruction Reference documentation ([#2285](https://github.com/0xMiden/miden-vm/issues/2285)).
 - Made `miden-vm analyze` output analysis even if execution ultimately errored. ([#2204](https://github.com/0xMiden/miden-vm/pull/2204)).
+- Allow `CALL` and `DYNCALL` from a syscall context ([#2296](https://github.com/0xMiden/miden-vm/pull/2296))
 
 #### Changes
 


miden-vm/tests/integration/operations/decorators/asmop.rs
```diff
@@ -3,6 +3,8 @@ use miden_debug_types::Location;
 use miden_processor::{AsmOpInfo, RowIndex, VmStateIterator};
 use miden_utils_testing::{assert_eq, build_debug_test};
 
+// Ignoring the now deprecated `VmStateIterator` tests
+#[ignore]
```
plafer: I ignored those for now, but I personally would prefer removing them, as IMO they will never be updated. Thoughts?

bitwalker: Probably not worth keeping them if they are ignored - about the only reason to keep them is to catch regressions between now and when `VmStateIterator` is removed, but I'm not sure that is a real risk we need to care about at this point.

plafer: Removed

plafer:
This is ready for review. There are 4 failing tests, which will be fixed by rebasing on `next` after #2295 is fixed. Please also see https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3415807040 for a discussion of what to do with the `in_syscall` column (either in this PR or the next, where we will be removing the `FMPADD` and `FMPUPDATE` operations, as well as the now unused `fmp` column).

Al-Kindi-0:
> It is also worth noting that syscalls were also previously disallowed in a syscall context. In this PR we allow them, as there is no difference between a syscall from context 0 (previously allowed), and one from a "syscall context 0". We also allow them in the assembler, since it's not possible to CALL into methods of a kernel module, which could then syscall back.

This reads a bit dense to me, could you unpack it a bit? Especially, not sure what "syscall context 0" mean here?

crates/assembly/src/tests.rs
```diff
@@ -1867,7 +1868,13 @@ fn ensure_correct_procedure_selection_on_collision() -> TestResult {
         MastNodeId::from_u32_safe(0_u32, program.mast_forest().as_ref()).unwrap();
 
     let (exec_f_node_id, exec_g_node_id) = {
-        let split_node_id = program.entrypoint();
+        let split_node_id = {
+            // Note: the program starts with a join node, which joins:
+            // - left: the fmp initialization sequence,
+            // - right: the actual entrypoint of the program (the if statement).
```
Al-Kindi-0, line 1903: Q: not sure about the meaning of "the if statement" here in the context of JOIN

Al-Kindi-0: I see, this is probably in the context of a test (I can't see) which is in principal just and IF node

docs/src/design/decoder/constraints.md
```diff
@@ -166,12 +166,24 @@ $$
...
 $$
 
 The above value sends both the hash initialization request and the kernel procedure access request to the chiplets bus when the `SYSCALL` operation is executed.
 
-Similar to `SYSCALL`, `DYN` and `DYNCALL` are handled separately, since in addition to communicating with the hash chiplet they must also issue a memory read operation for the hash of the procedure being called. 
+Similar to `SYSCALL`, `CALL` is handled separately, since in addition to communicating with the hash chiplet, it must also initialize the frame memory pointer (stored in memory at constant address `fmpaddr` with constant value `fmpinit`):
+
+$$
+m_{fmpwrite} = \alpha_0 + \alpha_1 \cdot m_{write} + \alpha_2 \cdot ctx' + \alpha_3 \cdot fmpaddr + \alpha_4 \cdot clk + alpha_5 \cdot fmpinit
+$$
```
Al-Kindi-0: Just double checking, this is the correct label here `m_{write}`

docs/src/design/decoder/constraints.md
```diff
@@ -182,10 +194,13 @@ m_{dynordyncall} = \alpha_0 + \alpha_1 \cdot m_{read} + \alpha_2 \cdot ctx + \al
 $$
 
 $$
-u_{dynordyncall} = (f_{dyn} + f_{dyncall}) (h_{dynordyncall} \cdot m_{dynordyncall})
+u_{dyn} = f_{dyn} \cdot h_{dynordyncall} \cdot m_{dynordyncall} \text{ | degree} = 7
+$$
+$$
+u_{dyncall} = f_{dyncall} \cdot h_{dynordyncall} \cdot m_{dynordyncall} \cdot m_{fmpwrite} \text{ | degree} = 8
```
Al-Kindi-0, line 205: Maybe we can still combine them and send the write request only when `f_dyncall` is set?

crates/assembly/src/assembler.rs
```diff
@@ -697,22 +710,25 @@ impl Assembler {
...
-            // - to allocate procedure locals we need to increment fmp by the number of locals
-            //   (rounded up to the word size), and
-            // - to deallocate procedure locals we need to decrement it by the same amount.
-            let locals_frame = Felt::from(num_locals.next_multiple_of(WORD_SIZE as u16));
-            let wrapper = BodyWrapper {
-                prologue: vec![Operation::Push(locals_frame), Operation::FmpUpdate],
-                epilogue: vec![Operation::Push(-locals_frame), Operation::FmpUpdate],
-            };
-            self.compile_body(proc.iter(), &mut proc_ctx, Some(wrapper), mast_forest_builder)?
+        let body_wrapper = if proc_ctx.is_program_entrypoint() {
```
adr1anh, line 713: Instead of Option<BodyWrapper> would it make sense to use `else {BodyWrapper::default()}`?

docs/src/user_docs/assembly/execution_contexts.md
```diff
@@ -42,35 +42,22 @@ In Miden Assembly, it is used to execute procedures without switching execution
...
 
-For the root context we have the following:
+- The first $2^{31}$ addresses are assumed to be global memory, freely usable by programs.
+- All addresses including and above $2^{31}$ are reserved for procedure locals (i.e., via the `exec` instruction).
 
-- The first $2^{30}$ addresses are assumed to be global memory.
-- The next $2^{30}$ addresses are reserved for memory locals of procedures executed in the root context.
-- The next $2^{30}$ addresses are reserved for memory locals of procedures executed from within a `syscall`.
-- The remaining address space has no special meaning.
+> Note that the highest accessible address ($2^{32} - 1$) is reserved by the VM to store the frame memory pointer.
```
adr1anh, line 61: This is not enforced right? What happens if it gets overwritten?

processor/src/system/mod.rs
```diff
@@ -21,6 +21,7 @@ mod tests;
 // address in a given context, regardless of which memory segment it belongs to.
 
 /// Memory addresses for procedure locals start at 2^30.
+/// DEPRECATED: use FMP_INIT_VALUE instead.
 pub const FMP_MIN: u64 = 2_u64.pow(30);
```
adr1anh, line 25: Should we just remove?

plafer: I left it there because some tests rely on it, and ultimately is more cleanly done when removing the `fmp` column in the next PR. i.e. This is going away very shortly.

docs/src/user_docs/assembly/execution_contexts.md
plafer: Correct, it is currently not enforced. In https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3391474721 I discussed a system to enforce that a given memory region is only accessed by designated instructions. But then as [suggested by Bobbin](https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3404826186) it's not *critical* to enforce that `fmp` is not overwritten. There are 3 cases to consider:
1. If a program explicitly manipulates the fmp, and they mess up, then it's on them.
2. If a program allocates so much memory that it overwrites the `fmp`. However at this point they'll have overwritten all the memory locals, so our previous system is as broken as the new one proposed in this PR.
3. a buffer overrun could let a hacker manipulate the `fmp`, and hence change where memory local accesses of future procedure calls land.

(3) is the biggest risk to consider, as decades of buffer overruns have shown. I'd argue that placing the `fmp` is marginally worse (if at all) than what we currently have. That is, if a buffer overrun is able to modify the `fmp`, then it's probably also able to modify the memory locals themselves (which similarly are not "protected"). Or any memory location, really - memory locals aren't special when it comes to the undesirable effects of buffer overruns. In other words, even if we protect the `fmp`, a hacker is still able to screw with the memory locals. So we figured it doesn't really change the attack surface to put the `fmp` in (unprotected) memory.

crates/assembly/src/tests.rs
plafer: > (I can't see)

What do you mean by you can't see? 😆

The program's `begin` block is:

```
begin
    if.true
        exec.f
    else
        exec.g
    end
end
```

But we now inject a basic block to initialize the `fmp` at the beginning of the program, as

```
begin
    push.FMP_INIT_VALUE push.FMP_ADDR mem_store drop
    if.true
        exec.f
    else
        exec.g
    end
end
```

That basic block and the `if.true ... end` blocks get `JOIN`'d together.

Should I update the comment?

docs/src/design/decoder/constraints.md
plafer: We want the [write_element](https://github.com/0xMiden/miden-vm/blob/5e901450170b789cc185712e3d0ab74cbe7cb63f/air/src/trace/chiplets/memory.rs#L32) label, but I dread using `_` in latex because from what I recall rendering it properly (on all platforms) can be tricky.

Switched to `m_{writeele}` for now, as I'm confident of how it gets rendered, and is more precise

crates/assembly/src/assembler.rs
plafer: I don't have strong opinions; we're using `Option<BodyWrapper>` everywhere so I kept it as is, but happy to switch to that if you prefer. Let me know

crates/assembly/src/tests.rs
Al-Kindi-0: Hahaha, my second comment was an answer to my question, and your answer confirms it, so all good for me
Thanks!

docs/src/design/decoder/constraints.md
Al-Kindi-0: makes sense

docs/src/design/decoder/constraints.md
plafer: IIUC, something like 

```
u_{dynordyncall} = (f_{dyn} + f_{dyncall}) * h_{dynordyncall} * m_{dynordyncall}
u_{dyncall} = f_dyncall * m_{fmpwrite}
```

IMO this makes seeing the degree of the "effective" `u_{dyncall}` (i.e. 8) more difficult. When reading quickly, I would see 
- `u_{dynordyncall}` -> degree 7
- `u_{dyncall}` -> degree 6

What do you think?

Al-Kindi-0: We can keep as is then

crates/assembly/src/fmp.rs
```diff
@@ -0,0 +1,78 @@
...
+/// pointer is always word-aligned for operations that require it.
+pub(crate) fn fmp_start_frame_sequence(num_locals: u16) -> Vec<Operation> {
+    let locals_frame = Felt::from(num_locals.next_multiple_of(WORD_SIZE as u16));
+
+    [Operation::Push(locals_frame)]
+        .into_iter()
+        .chain(add_fmp_to_stack_top())
+        .chain(store_stack_top_to_fmp())
+        .collect()
+}
```
plafer, line 34: *(For completeness)*

The operations that currently get injected at the start of a procedure with locals are:

```
# Put updated fmp value on stack top
Push(locals_frame)
Push(FMP_ADDR)
Mload
Add

# Push stack top to fmp
Push(FMP_ADDR)
Mstore
Drop
```

Hence we inject **7 operations** containing **3 immediate values**.

Alternatively, we could avoid pushing `FMP_ADDR` twice, and instead duplicate it on the stack, resulting in the following:

```
Push(locals_frame)

# Push fmp addr, and store it on stack
Push(FMP_ADDR)
Dup
MovDn2

# Compute new fmp on stack top (as before)
Mload
Add

# Push stack top to fmp - but get the FMP_ADDR from the stack
Swap
Mstore
Drop
```

This results in **9 operations** containing **2 immediate values**.

So the tradeoff is: add 2 operations to save 1 immediate value.

I think the current set of operations is preferable, since 
- procedures which do *not* start with a basic block will save 2 clock cycles
- similarly for procedures starting with a small basic block

Basically the only case where adding 2 operations to save an immediate value is preferable is for procedures that start with a basic block where adding the extra immediate value would force to create a new batch. Anyhow, it's a minor point, but let me know if you disagree and that we should switch to this other set of operations.

plafer:
Rebased on `next`, and all tests now pass.

> This reads a bit dense to me, could you unpack it a bit? Especially, not sure what "syscall context 0" mean here?

Reworded

crates/assembly/src/assembler.rs
adr1anh: all good!

Al-Kindi-0 (approved):
LGTM!

Issue #1873, opened 2025-06-11
Reduce the serialized size of the compiled MAST


greenhat:
### Feature description

Following https://github.com/0xMiden/miden-base/issues/1441#issuecomment-2959614805:

> Our current serialization methodology is fast but "fat". We can implement a much more efficient serialization approach. I'm expecting the "slow but thin" serialization to give us somewhere between 5x and 10x in size reduction.

### Why is this feature needed?

The MASM code compiled from Rust is larger than the equivalent manually written code.

bobbinth:
Closing this as we now have a way to strip decorators and this reduces the size of the MAST forest quite a bit. We can do more compression, but this is tracked by #2628.

Issue #1497, opened 2024-09-14
Performance of current decorator representation


bitwalker:
## Issue

While investigating 0xPolygonMiden/compiler#317, I discovered via flamegraph analysis that up to 30% of program runtime was being spent just `drop`ing vectors used to store decorators.

The conditions in which this occurs are:

* A program is compiled with debug information enabled, thus every node with source spans attached gets an `AsmOp` decorator at a minimum. For programs compiled from Rust, this is basically every node in the program.
* One or more `Library` are loaded into memory, potentially cloned up to two times for reasons of ownership, and then dropped when no longer used.

The issue was noticed in our test suite, _prior_ to the change in this repo to allocating `Library`s in `Arc`, and cloning those rather than the actual `Library`. Even though the compiler was caching instances of `Library` and only cloning it (similar to what is now done in `next`), we observed the excessive time spent in `drop_in_place`.

## Impact

However, even though the recent changes in `next` makes this a non-issue for `miden-stdlib` (and I believe `miden-base`, not sure if that library is also allocated with `'static` lifetime), for the time being, it will affect any other library, even if allocated via `Arc`. The use of `Arc` does mean that clones are cheap and don't require dropping, however the actual `Library` still gets dropped when all of its referencing `Arc`s are dropped. In one-shot compiler-like scenarios, this is a small difference, and the time spent dropping the data will still impact user-perceived execution time of the compiler/assembler and executor.

The compiler is also planning to switch to a `rustup`-like toolchain management approach to the libraries and tools it uses, which means that there will be no benefit to the `miden-stdlib` strategy of caching the `Library` allocations in a static. All libraries, regardless of whether they are fundamental or user-defined, will be treated the same, so this affects all programs, not just a subset.

## Background

**NOTE:** The way decorators are managed was recently changed in `next`, but it may actually make the problem worse rather than better, as we now have two `Vec` for decorators (to partition before and after decorators), rather than just one.

As mentioned above, the problem is that every `MastNode` must be dropped individually when dropping a `MastForest`. If each `MastNode` also has two `Vec`, both vectors must be dropped as well, if they are non-empty, which is rarely the case when debug info is enabled.

More generally though, any heap-allocated type stored in a `MastNode` would present a similar issue. `Vec` is particularly bad because all of the elements must be dropped as well, but it's less about that, and more about the amount of drop glue that is executed when a `MastForest` is dropped - it ends up being a lot.

[...]

PR #2301, opened 2025-10-19
MastNode builder extension (4/4)

huitseeker:
### Context
This is the 4th and final PR in the MAST node builder modernization series, completing the transition from direct construction patterns to a unified builder-based architecture. Building on PRs #2259, #2274 and #2291, this PR removes all remaining mutable MastNode legacy APIs and adds immutability we need to solve #1776.

**Complete Legacy API Elimination**
- Removed: All `add_node` methods and `new_unsafe` constructors from MastForest and node types
- Removed: Dead constructor methods (`JoinNode::new`, `SplitNode::new`, `CallNode::new`, etc.)
- Removed: Duplicate `MastNodeFingerprint::from_mast_node` API (replaced with builder-based fingerprinting)
- Builders are now the only way to create and add nodes to MAST forests

**New Builder Capabilities**
- **Digest Forcing** for serialization: All builders now support `with_digest(digest)` for cryptographic control
- **Decorator Appending**: New `append_before_enter()`/`append_after_exit()` methods for incremental decorator management in the builders
- **Serialization Support**: `MastNodeInfo::try_into_mast_node_builder()` enables extracting builders from serialization data

<details><summary> Technical details for reviewers</summary>

**1. Digest Forcing Implementation** 
- Added `digest: Option<Word>` field to all builder structs
- Modified `build()` methods to use forced digests when provided
- Updated fingerprint computation to respect forced digests
- Critical for serialization round-trips where original digests must be preserved

**2. Decorator Management Refactoring**
- Replaced post-creation decorator appending with builder-time specification
- Updated `ensure_*` methods in MastForestBuilder to accept decorator parameters
- Refactored instruction compilation pipeline to pass decorators through builders
- Eliminated potential fingerprint inconsistencies from delayed decorator addition

**3. MAST Merger Modernization**
- Replaced `build_with_remapped_children` to return builders instead of nodes
- Updated merger to use `to_builder()` and `remap_children()` methods
- Simplified merge logic by working directly with builders

**4. Serialization Infrastructure**
- Relaxed `read_block_decorators` to work without full MastForest reference
- Added `MastNodeInfo::try_into_mast_node_builder()` for deserialization workflows
</details>

### Migration Path for Users
```rust
// Old way (no longer works)
let node = BasicBlockNode::new(ops, decs);
let node_id = forest.add_node(node);

// New way (required)
let node_id = BasicBlockNodeBuilder::new(ops, decs)
    .with_before_enter(decorators_before)
    .with_after_exit(decorators_after)
    .with_digest(some_digest)
    .add_to_forest(&mut forest);
```

### DIFF
--- a/core/src/mast/node/call_node.rs
+++ b/core/src/mast/node/call_node.rs
@@ -264,12 +264,16 @@ impl proptest::prelude::Arbitrary for CallNode {
         // Generate callee, digest, and whether it's a syscall
         (any::<MastNodeId>(), any::<[u64; 4]>(), any::<bool>())
             .prop_map(|(callee, digest_array, is_syscall)| {
-                // Use new_unsafe since we're generating arbitrary nodes
-                // The digest is also arbitrary since we can't compute it without a MastForest
+                // Generate a random digest
                 let digest = Word::from(digest_array.map(Felt::new));
-                let mut node = CallNode::new_unsafe(callee, digest);
-                node.is_syscall = is_syscall;
-                node
+                // Construct directly to avoid MastForest validation for arbitrary data
+                CallNode {
+                    callee,
+                    is_syscall,
+                    digest,
+                    before_enter: Vec::new(),
+                    after_exit: Vec::new(),
+                }
             })
             .no_shrink()  // Pure random values, no meaningful shrinking pattern
             .boxed()
--- a/core/src/mast/node/dyn_node.rs
+++ b/core/src/mast/node/dyn_node.rs
@@ -223,9 +223,9 @@ impl proptest::prelude::Arbitrary for DynNode {
         any::<bool>()
             .prop_map(|is_dyncall| {
                 if is_dyncall {
-                    DynNode::new_dyncall()
+                    DynNodeBuilder::new_dyncall().build()
                 } else {
-                    DynNode::new_dyn()
+                    DynNodeBuilder::new_dyn().build()
                 }
             })
             .no_shrink()  // Pure random values, no meaningful shrinking pattern
--- a/core/src/mast/node/external.rs
+++ b/core/src/mast/node/external.rs
@@ -204,7 +204,7 @@ impl proptest::prelude::Arbitrary for ExternalNode {
         any::<[u64; 4]>()
             .prop_map(|[a, b, c, d]| {
                 let word = Word::from([Felt::new(a), Felt::new(b), Felt::new(c), Felt::new(d)]);
-                ExternalNode::new(word)
+                ExternalNodeBuilder::new(word).build()
             })
             .no_shrink()  // Pure random values, no meaningful shrinking pattern
             .boxed()
--- a/core/src/mast/node/join_node.rs
+++ b/core/src/mast/node/join_node.rs
@@ -225,10 +225,15 @@ impl proptest::prelude::Arbitrary for JoinNode {
         // Generate two MastNodeId values and digest for the children
         (any::<MastNodeId>(), any::<MastNodeId>(), any::<[u64; 4]>())
             .prop_map(|(first_child, second_child, digest_array)| {
-                // Use new_unsafe since we're generating arbitrary nodes
-                // The digest is also arbitrary since we can't compute it without a MastForest
+                // Generate a random digest
                 let digest = Word::from(digest_array.map(Felt::new));
-                JoinNode::new_unsafe([first_child, second_child], digest)
+                // Construct directly to avoid MastForest validation for arbitrary data
+                JoinNode {
+                    children: [first_child, second_child],
+                    digest,
+                    before_enter: Vec::new(),
+                    after_exit: Vec::new(),
+                }
             })
             .no_shrink()  // Pure random values, no meaningful shrinking pattern
             .boxed()
--- a/core/src/mast/node/loop_node.rs
+++ b/core/src/mast/node/loop_node.rs
@@ -210,10 +210,15 @@ impl proptest::prelude::Arbitrary for LoopNode {
         // Generate one MastNodeId value and digest for the body
         (any::<MastNodeId>(), any::<[u64; 4]>())
             .prop_map(|(body, digest_array)| {
-                // Use new_unsafe since we're generating arbitrary nodes
-                // The digest is also arbitrary since we can't compute it without a MastForest
+                // Generate a random digest
                 let digest = Word::from(digest_array.map(Felt::new));
-                LoopNode::new_unsafe(body, digest)
+                // Construct directly to avoid MastForest validation for arbitrary data
+                LoopNode {
+                    body,
+                    digest,
+                    before_enter: Vec::new(),
+                    after_exit: Vec::new(),
+                }
             })
             .no_shrink()  // Pure random values, no meaningful shrinking pattern
             .boxed()
--- a/core/src/mast/node/split_node.rs
+++ b/core/src/mast/node/split_node.rs
@@ -221,10 +221,15 @@ impl proptest::prelude::Arbitrary for SplitNode {
         // Generate two MastNodeId values and digest for the children
         (any::<MastNodeId>(), any::<MastNodeId>(), any::<[u64; 4]>())
             .prop_map(|(true_branch, false_branch, digest_array)| {
-                // Use new_unsafe since we're generating arbitrary nodes
-                // The digest is also arbitrary since we can't compute it without a MastForest
+                // Generate a random digest
                 let digest = Word::from(digest_array.map(Felt::new));
-                SplitNode::new_unsafe([true_branch, false_branch], digest)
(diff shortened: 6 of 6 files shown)


core/src/mast/node/call_node.rs
```diff
@@ -48,79 +48,7 @@ impl CallNode {
...
-        Self {
-            callee,
-            is_syscall: true,
-            digest,
-            before_enter: Vec::new(),
-            after_exit: Vec::new(),
-        }
-    }
-}
+impl CallNode {}
```
plafer: I would remove this

core/src/mast/node/mast_forest_contributor.rs
```diff
@@ -6,30 +6,55 @@ use super::{
     BasicBlockNodeBuilder, CallNodeBuilder, DynNodeBuilder, ExternalNodeBuilder, JoinNodeBuilder,
     LoopNodeBuilder, SplitNodeBuilder,
 };
-use crate::mast::{MastForest, MastForestError, MastNodeId};
+use crate::{
+    LookupByIdx,
+    mast::{MastForest, MastForestError, MastNodeId},
+};
 
 pub trait MastForestContributor {
```
plafer, line 14: Would `MastNodeBuilderInterface` be a better name? This would be consistent with the naming for the `Processor` trait.

plafer (commented):
Not sure how others feel about this, but maybe a single PR would be easier to review? Currently, I find myself looking through the PRs to find when a given change was made to comment on it. It also spreads the conversation.

The diff for a single PR would be +4196 -1765, which is big but still manageable IMO (especially since all the changes are part of a single whole).

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -489,27 +544,40 @@ impl MastForestBuilder {
...
+        &mut self,
+        node_id: MastNodeId,
+        decorator_ids: Vec<DecoratorId>,
+    ) -> Result<(), MastForestError> {
+        // Extract the existing node and convert it to a builder
+        let mut decorated_builder = self.mast_forest[node_id].clone().to_builder();
+        decorated_builder.append_before_enter(decorator_ids);
+        let new_node_fingerprint =
+            decorated_builder.fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)?;
+        self.mast_forest[node_id] = decorated_builder.build(&self.mast_forest)?;
```
plafer, line 557: Don't we also need to update `self.node_id_by_fingerprint` with the new mapping (and possibly remove the previous/stale one)?

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -489,27 +544,40 @@ impl MastForestBuilder {
...
+    ) -> Result<(), MastForestError> {
+        // Extract the existing node and convert it to a builder
+        let mut decorated_builder = self.mast_forest[node_id].clone().to_builder();
+        decorated_builder.append_after_exit(decorator_ids);
+        let new_node_fingerprint =
+            decorated_builder.fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)?;
+        self.mast_forest[node_id] = decorated_builder.build(&self.mast_forest)?;
 
-        let new_node_fingerprint = self.fingerprint_for_node(&self[node_id]);
         self.hash_by_node_id.insert(node_id, new_node_fingerprint);
```
plafer, line 576: Same comment as above

huitseeker:
@plafer The feedback that I've *universally* received is that folks want diffs in smaller chunks, to the point each of these PRs passes tests and checks _on every commit_, and that every commit is thematically self-contained.

If it's a matter of seeing the diff, opening the global PR is already easy, at https://github.com/0xMiden/miden-vm/compare/next...huitseeker/node_builder_extension_3

Happy to do whatever reviewers prefer, if there's s/o else committing to review that shares your preferences.

crates/assembly/src/mast_forest_builder.rs
huitseeker: I think the reason this works without it is that by the point this function is called in the recursive descent, that lookup table is no longer used. But you're right from a coherence / defensive programming PoV (and I keep eyeing #1873 from the corner of my eye), so I'll add the update for sure.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Same ack.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -489,27 +544,40 @@ impl MastForestBuilder {
     ///
     /// If other decorators are already present, the new decorators are added to the end of the
     /// list.
-    pub fn append_before_enter(&mut self, node_id: MastNodeId, decorator_ids: &[DecoratorId]) {
-        self.mast_forest[node_id].append_before_enter(decorator_ids);
+    pub fn append_before_enter(
+        &mut self,
+        node_id: MastNodeId,
+        decorator_ids: Vec<DecoratorId>,
+    ) -> Result<(), MastForestError> {
```
plafer, line 551: Thank you for taking on this massive refactoring!

Though I think there's a disconnect between what I expected this series of PRs to achieve, and what it presently does. As I understood it, the main evil we were trying to banish is: **no more mutating MAST nodes after they've been inserted into a `MastForest`**. This goal can be reduced down to: remove all `&mut` methods from `MastForest` - except maybe a few innocuous ones like `make_root()`. But critically, no more `index_mut()`. I recall an earlier conversation where you stated that using a `MastForest` in the `MastForestBuilder` was wrong - it would be better to hold all that state internally and only at the end, once we have all the information, create the `MastForest` in one go. And hence there would never be a need to mutate the `MastForest` again.

In contrast, the infrastructure introduced in this PR ensures that we no longer mutate *MastNode*s. I think though that this is not enough to achieve our goal, by the fact that we still use the pattern `mast_forest[node_id] = old_node.to_builder().mutate().build()`. Conceptually, since we use `MastNodeId` effectively as pointers, this is equivalent to mutating the `MastNode` pointed to by its `MastNodeId` (even though we didn't mutate the *object* that was in memory - we still mutate that slot in the `MastForest`).

With this in mind, what I was expecting was a bigger change to the assembler where it would no longer be necessary to go back to any node that was built to mutate it given new information (e.g. appending decorators). Essentially, whenever a node is created, all the information needed to *fully* create it would already exist.

> Personally, whether or not we use a `MastForest` as a backing store to the `MastForestBuilder` is less important, as long as we never `index_mut()` back to a node that was previously added to mutate that slot.

Am I reading this right? Or did I misunderstand the ultimate goal that this series of PRs was aiming to achieve? Also, now that I have the context from these PRs, it would be helpful if you reiterated how you would use this for #1776 🙂

huitseeker: The short version of #1776, and the gist of #1497 is that we're uncomfortable with the representation of decorators indexes owned by each node (multiple reasons - primarily for the `BasicBlockNode` ' s op-indexed decorator indexes, but for `before_enter` / `after_exit` indexes as well). We would like to replace that representation with one where access to decorator information is, in a nutshell, _shared_ across all nodes.

There's no _shared_ reference to a representation that allows mutability under Rust's memory model without acrobatics. Those PRs remove this mutability (i.e. the mutability of the decorator representation), and thus the need for acrobatics.

There is a world in which we could:
- refactor the whole assembler to not mutate a `MastForest`, displacing e.g. merging *and* incremental block construction,
- *not* offer APis for accessing decorators from a node (working exclusively from a ` MastForest` and a `NodeId`  key),

These worlds are on principle probably good, and I we may want to incarnate them at some point. They are also more complex to engineer than this PR proposes, and I think they are not necessary for the eventual purpose. Arguably, this PR proposes a stepping stone in that direction should we one day want to make this happen.

With that said, I hear you on wanting more detail on the #1776 work.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Done!

crates/assembly/src/mast_forest_builder.rs
huitseeker: Done!

core/src/mast/node/mast_forest_contributor.rs
huitseeker: It's a bit pedantic but I'm open to it. Due to the effect of renames on a stack of PRs, I'd rather do this (and other renames) in a small cleanup PR after we've merged a good chunk.

core/src/mast/node/call_node.rs
huitseeker: Done!

Al-Kindi-0 (approved):
LGTM

bobbinth (approved):
Also a pretty light review from me - but looks good! thank you!

PR #2305, opened 2025-10-21
fix: fix padding direction bug in BasicBlockNode

### LINKED ISSUE #2295: Incorrect decorator index mapping due to NOOPs

huitseeker:
The issue was that `DecoratorPaddingOffsets` was being used in "reverse" for both raw -> padded and padded -> raw conversions, causing sign/direction confusion. The padding and un-padding operations were tested to be inverses of each other, but both code paths were using offsets in the wrong direction, which did not allow that round-trip to catch the bug.

Fixes #2295 

## Solution 

Implemented two explicit, unambiguous prefix arrays:

### RawToPaddedPrefix
- Domain: raw operation indices
- Meaning: count of padding ops strictly before raw index r
- Usage:  (addition)

### PaddedToRawPrefix
- Domain: padded operation indices (equivalent to old DecoratorPaddingOffsets)
- Meaning: count of padding ops strictly before padded index p
- Usage:  (subtraction)

## Tests

- tested in variants using proptests for each direction.
- maintained the round-trips so that un-padding is the inverse of padding.
- specific test reproducing issue 2295.

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -785,11 +785,6 @@ impl RawToPaddedPrefix {
         }
 
         // Sentinel for r == raw_ops
-        //
-        // Some decorators have an operation index equal to the length of the
-        // operations array, to ensure they are executed at the end of the block
-        // (since the semantics of the decorator index is that it must be executed
-        // before the operation index it points to).
         v.push(pads_so_far);
         RawToPaddedPrefix(v)
     }
@@ -802,6 +797,13 @@ impl RawToPaddedPrefix {
 }
 
 /// Get the number of padding operations before raw index r.
+///
+/// ## Sentinel Access
+///
+/// Some decorators have an operation index equal to the length of the
+/// operations array, to ensure they are executed at the end of the block
+/// (since the semantics of the decorator index is that it must be executed
+/// before the operation index it points to).
 impl core::ops::Index<usize> for RawToPaddedPrefix {
     type Output = usize;
     #[inline]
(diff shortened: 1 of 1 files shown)


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -759,62 +747,142 @@ pub(crate) fn validate_decorators(operations_len: usize, decorators: &DecoratorL
...
 }
 
-#[doc(hidden)]
-impl Index<usize> for DecoratorPaddingOffsets {
+/// Get the number of padding operations before raw index r.
+impl core::ops::Index<usize> for RawToPaddedPrefix {
     type Output = usize;
+    #[inline]
+    fn index(&self, idx: usize) -> &Self::Output {
+        &self.0[idx]
```
adr1anh, line 811: Should we panic if accessing the sentinel?

huitseeker: We do in fact voluntarily access it, something we document a few lines above when populating a value for it. I should perhaps move that comment ...

plafer: Do we want to move the comment? Otherwise this is good to merge

PR #2308, opened 2025-10-23
Remove `FmpUpdate` + `FmpAdd` operations, and `fmp` + `in_syscall` columns

plafer:
Follow-up cleanup to #2296. This is a mechanical cleanup, no new functionality is introduced.

Notable mentions:
- The `Caller` operation is now allowed is all execution contexts (rather than just the kernel while servicing a syscall)
  - This is derived from the fact that we no longer have the `in_syscall` column. The reason for removing it is discussed in https://github.com/0xMiden/miden-vm/issues/1681#issuecomment-3415807040
- The recursive verifier tests are disabled, due to the change in trace layout, and constraints.
  - We can fix and re-enable it when we're done writing all the constraints in airscript

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added missing endianness-aware memory instructions (`mem_loadw_be`, `mem_loadw_le`, `mem_storew_be`, `mem_storew_le`) to Instruction Reference documentation ([#2285](https://github.com/0xMiden/miden-vm/issues/2285)).
 - Made `miden-vm analyze` output analysis even if execution ultimately errored. ([#2204](https://github.com/0xMiden/miden-vm/pull/2204)).
 - Allow `CALL` and `DYNCALL` from a syscall context ([#2296](https://github.com/0xMiden/miden-vm/pull/2296))
+- Remove operations `FmpUpdate` and `FmpAdd`, as well as columns `fmp` and `in_syscall` ([#2308](https://github.com/0xMiden/miden-vm/pull/2308))
 
 #### Changes
 


docs/src/design/stack/op_constraints.md
```diff
@@ -78,7 +78,7 @@ This group contains $32$ operations which do not shift the stack (this is almost
 | `INV`     |     $3$      |   `000_0011`    |  [Field ops](./field_ops.md)  |     $7$     |
 | `INCR`    |     $4$      |   `000_0100`    |  [Field ops](./field_ops.md)  |     $7$     |
 | `NOT`     |     $5$      |   `000_0101`    |  [Field ops](./field_ops.md)  |     $7$     |
-| `FMPADD`  |     $6$      |   `000_0110`    | [System ops](./system_ops.md) |     $7$     |
+| `<unused>`|     $6$      |   `000_0110`    |                               |     $7$     |
```
adr1anh, line 81: nice 😁

docs/src/user_docs/assembly/io_operations.md
```diff
@@ -34,7 +34,7 @@ In both case the values must still encode valid field elements.
...
 | clk <br /> - _(1 cycle)_             | [ ... ]      | [t, ... ]    | $t \leftarrow clock\_value()$ <br /> Pushes the current value of the clock cycle counter onto the stack.                                                                                                              |
 | sdepth <br /> - _(1 cycle)_          | [ ... ]      | [d, ... ]    | $d \leftarrow stack.depth()$ <br /> Pushes the current depth of the stack onto the stack.                                                                                                                             |
-| caller <br /> - _(1 cycle)_          | [A, b, ... ] | [H, b, ... ] | $H \leftarrow context.fn\_hash()$ <br /> Overwrites the top four stack items with the hash of a function which initiated the current SYSCALL. <br /> Executing this instruction outside of SYSCALL context will fail. |
+| caller <br /> - _(1 cycle)_          | [A, b, ... ] | [H, b, ... ] | $H \leftarrow context.fn\_hash()$ <br /> Overwrites top 4 stack items with hash `H` of the function that called into the current context, or initiated the current `SYSCALL`. Returns `[0, 0, 0, 0]` when in context 0 and not servicing a `SYSCALL`. |
```
bobbinth, line 37: I think this would return `[0, 0, 0, 0]` in two cases:

1. As described, when we are in context `0` but not in a `syscall`.
2. In the first context that was started from context `0`.

Or am I missing something?

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -173,7 +173,7 @@ Instructions for moving data between the stack and other sources like program co
 | -------------- | ------------ | ------------ | ------ | --------------------------------------------------------------------------------------------------------------------------- |
 | `clk`          | `[ ... ]`    | `[t, ... ]`  | 1      | Pushes current clock cycle `t`.                                                                                             |
 | `sdepth`       | `[ ... ]`    | `[d, ... ]`  | 1      | Pushes current stack depth `d`.                                                                                             |
-| `caller`       | `[A, b,...]` | `[H, b,...]` | 1      | Overwrites top 4 stack items with hash `H` of the function that initiated the current `SYSCALL`. Fails if not in `SYSCALL`. |
+| `caller`       | `[A, b,...]` | `[H, b,...]` | 1      | Overwrites top 4 stack items with hash `H` of the function that called into the current context, or initiated the current `SYSCALL`. Returns `[0, 0, 0, 0]` when in context 0 and not servicing a `SYSCALL`. |
```
bobbinth, line 176: Same comment as above.

docs/src/img/design/decoder/decoder_call_operation.png
bobbinth, line 1: We should also probably update the [trace diagram](https://github.com/0xMiden/miden-vm/blob/next/docs/src/img/design/vm_trace.png) image and the surrounding text.

docs/src/user_docs/assembly/io_operations.md
plafer: Hmmm, actually, when in any other context than 0, `caller` returns the hash of the function that entered the context. For example,

```
proc.f
  # returns hash of `f`
  caller
end

begin
  call.f
end
```

I'll open a PR to fix the docs.

plafer: #2321

docs/src/img/design/decoder/decoder_call_operation.png
plafer: Fixed in #2321

docs/src/user_docs/assembly/io_operations.md
bobbinth: Makes sense. So, to summarize:

- When we execute a `call` (or `dyncall`), `fn_hash` is set to the function that we call.
- When we execute `syscall`, fn_hash` remains unchanged.

This way the `caller` instruction returns hash of the "current" function when we are inside a call, but hash of the caller when we are inside a syscall.

With #2321 I think this is documented well, but I wonder if we should consider changing this in the future so that `caller` always returns the caller of the current function. This may be something we could do together with moving `fn_hash` to memory.

Issue #1821, opened 2025-05-16
always include debug info when assembling


plafer:
Context: https://github.com/0xMiden/miden-base/pull/1353#pullrequestreview-2835612467

Note that without any mitigation this will slow down the fast processor, which executes all decorators. Hence this is related to the issue of only executing the decorators in the fast processor in some "debug mode" (or not at all if that ends up really not being possible).

bitwalker:
Technically, that option would still exist, it would just be handled by the Miden equivalent of `strip`.

I think either approach is fine though.

bobbinth:
Removing the option would simplify the internal of the assembler a bit - so, I think it is still preferred. Functionality wise, we could offer an option that first assembles a program and then strips debug info. From the user standpoint this could be equivalent to `--release` build.

plafer:
While working on this issue, I noticed a difficulty with stripping the decorators from a `MastForest`. That is, in the `MastForestBuilder`, the decorators on a node are considered when deciding whether 2 nodes are equal (with our `MastNodeFingerprint`). Therefore, when we strip the decorators, some nodes that were previously unequal become equal (if they're equal except for their decorators).

We should probably do #1776 before then - after which the `AssemblyOp`/source mapping information will no longer be considered in the equality check between nodes. However, we'll still have the subtle problem that a `MastForest` built with decorators + strip them out can result in a different `MastForest` if it was built directly without decorators if we're not careful. For example,


```
proc.f
  add
end

prof.g
  add debug.stack
end
```

If we build without decorators, both `f` and `g` will be compiled down to the same `MastNodeId`; but if we build *with* decorators, they will be compiled to different `MastNode`s, and (blindly) stripping decorators will result in a `MastForest` with 2 nodes instead of 1.

plafer:
Given the above, I think we should just switch the default to be `debug_mode=true` for release 0.15, and only fully remove `debug_mode` once we have a better idea of how to properly strip decorators.

bobbinth:
> If we build without decorators, both `f` and `g` will be compiled down to the same `MastNodeId`; but if we build _with_ decorators, they will be compiled to different `MastNode`s, and (blindly) stripping decorators will result in a `MastForest` with 2 nodes instead of 1.

Ah - good point! Yes, simply removing decorators could result in an odd state of the `MastForest` (it shouldn't cause any issues, but still will look somewhat unexpected). So, I agree, stripping out decorators would be a bit more involved and should be done as a separate task before this one.

Issue #1580, opened 2024-11-19
Allow to truncate decorator data when serializing MAST


yasonk:
### Feature description

For use cases described in [this comment](https://github.com/0xPolygonMiden/miden-vm/pull/1531#discussion_r1837037857), need to have the capability to truncate the decorator data when serializing MAST.

[This comment](https://github.com/0xPolygonMiden/miden-vm/pull/1531#discussion_r1837883671) describes one approach for implementing this feature.

Should probably introduce CLI flag for the compile command to write the compiled program with truncated optional data.

### Why is this feature needed?

Per the comment on Pull Request #1531 :
... general idea is that producers, e.g. the compiler, will emit a fully-fleshed out package containing all the things that could be useful for interacting with that package (e.g. debug info, documentation, sources in some caes, etc.), and the MAST contained in the package would have all of the decorators (naturally required to support debugging, etc.). There are contexts where all of that stuff is simply dead weight though, or you want to optimize for binary size, in which case you can strip all of that, resulting in a minimal package that only contains exactly what is needed to execute what it contains. One such potential use case, would be publishing packages on-chain, which would only be feasible if they are as small as possible.

yasonk:
With the advice map, calculating the decorator offset becomes a little more complicated. Potentially a SizeCalculator is warranted. You would first use SizeCalculator to dry-run all writing operations that you want to calculate, then write the decorator data offset, then write the data:
```rust
struct SizeCalculator {
    total_size: usize,
}

impl SizeCalculator {
    fn new() -> Self {
        Self { total_size: 0 }
    }

    fn size(&self) -> usize {
        self.total_size
    }
}

impl ByteWriter for SizeCalculator {
    fn write_u8(&mut self, _value: u8) {
        // Increment size for a single byte
        self.total_size += 1;
    }

    fn write_bytes(&mut self, bytes: &[u8]) {
        // Increment size by the length of the byte slice
        self.total_size += bytes.len();
    }

    // All other methods from ByteWriter can remain unimplemented
    // because they eventually rely on write_bytes or write_u8.
}
```

I can see some brittleness of this approach if the concrete class of `ByteWriter` used for writing the program changes without changing the `SizeCalculator`. But not sure how to couple them in a graceful manner.

Using size_hints seems even more brittle because then any change in structure/implementation may require modifying size_hint, which seems like a bigger mental burden compared to `ByteWriter` changes.

[...]

PR #2310, opened 2025-10-26
refactor: hoist BasicBlock op-indexed decorators to the MastForest (1/2)

huitseeker:
On top of #2259, #2274, #2291, #2301, which make `MastNode`s immutable after insertion in a `MastForest`.

# BasicBlockNode Op-Indexed Decorators

This PR adds a new way to store decorator index information in `BasicBlockNode` that lets both old and new systems work together. Think of it like having two types of storage - one for blocks that stand alone, and one for blocks that are part of a `MastForest`.

## What Changed

We added a `DecoratorStore` that has two options:
- `Owned`: Stores decorator index pointers directly in the block (for standalone blocks)
- `Linked`: Points to shared decorator index information (for blocks in a `MastForest`)

This means blocks inside a `MastForest` can now represent their decorator index information in a single sparse vector (with a few side indexes), which saves memory and decreases allocator pressure.

## Key Improvements

**Memory Efficiency**: When many blocks contain little or no decorator index information, we represent them as a few indices in CSR representation, rather than with a per-node COO representation.

**Systematic Representation**: The new representation follows the doctrine (adapted to our context) of compressed, sparse, row-major "matrix" representations, which is a standard way to store sparse values efficiently.

**Custom Serialization**: Blocks can still be saved & loaded correctly, even with the new storage system.

<details><summary>**How It Works**</summary>

When you create a block by itself, via a Builder or manually, it uses `Owned` storage:

```rust
// legacy, don't do this outside tests
let test_block = BasicBlockNode::new(operations, decorators)?;
// Do this instead
let block = BasicBlockNodeBuilder::new(operations, decorators).build()?;
```

When you add that block to a MastForest, it automatically switches to `Linked` storage which defers to the `MastForest`'s centralized storage:

```rust
let node_id = BasicBlockNodeBuilder::new(operations, decorators).add_to_forest(&mut forest)?;
```

No need to change your code - everything works the same way.

### Testing

Unit tests make sure:
- All existing code still works
- Decorators are saved and loaded correctly
- Memory is laid out properly between blocks
- Edge cases like empty blocks work fine
- Several decorator arrangements (like multiple decorators on the same operation) work correctly

### References
- Contributes to #1776 and #1497.
- [plan](https://hackmd.io/@huitseeker/HkkhyiUsge)

</details>

## Benefits & Next steps

This change makes the basic block decorators uses less memory & iterate faster in some cases.

The next step would be:
- to eventually remove the old `Owned` storage completely, but for now both systems work together,
- to remove the `before_enter` and `after_exit` decorators out of the `MastNode`s, in a similar way.

### DIFF
--- a/core/src/mast/decorator_storage.rs
+++ b/core/src/mast/decorator_storage.rs
@@ -137,6 +137,7 @@ impl DecoratorIndexMapping {
     /// - Last value in `node_indptr_for_op_idx` must be <= `op_indptr_for_dec_ids.len() - 1`
     /// - Both `op_indptr_for_dec_ids` and `node_indptr_for_op_idx` must be strictly monotonic (each
     ///   successive value must be >= the previous one)
+    #[cfg(test)]
     pub fn from_components(
         decorator_ids: Vec<DecoratorId>,
         op_indptr_for_dec_ids: Vec<usize>,


core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,945 @@
...
+
+impl Default for DecoratorIndexMapping {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Immutable view over all `(op_idx, DecoratorId)` pairs for a node.
+/// Uses the two-level CSR encoded by `node_indptr_for_op_idx` and `op_indptr_for_dec_idx`.
+pub struct DecoratedLinks<'a> {
```
plafer, line 431: This docstring (and others) talk about operation indices, but then refer to the general "node" (rather "basic block", the only node which has operations). FWIU this decorator storage will grow to include all nodes - so how does this work for non-basic block nodes for which `op_idx` doesn't apply?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -987,10 +1266,39 @@ impl BasicBlockNodeBuilder {
...
+            }))
+            .map_err(|_| MastForestError::TooManyNodes)?;
+
+        // Add decorator info to the forest storage
+        Arc::make_mut(&mut forest.decorator_storage)
+            .add_decorator_info_for_node(node_id, decorators_info)
+            .map_err(MastForestError::DecoratorError)?;
+
+        // Convert the block to Linked variant (this is the core of Step 2)
+        if let MastNode::Block(block_node) = &mut forest.nodes[node_id] {
```
plafer: I would create a `MastNode::unwrap_basic_block_mut()` to be used here

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -987,10 +1266,39 @@ impl BasicBlockNodeBuilder {
...
+            block_node.decorators = DecoratorStore::Linked {
+                id: node_id,
+                storage: Arc::clone(&forest.decorator_storage),
+            };
```
plafer: IIUC, using `Arc::make_mut()` followed by `Arc::clone()` means that for every `BasicBlock` (and every node in the subsequent PR?), `Arc::make_mut()` will cause a `DecoratorIndexMapping::clone()` (such that we'll end up with many copies in memory).

Fixing this might need to wait for a larger refactor - at least for the obvious fix to use a plain `&mut DecoratorIndexMapping` during construction, and only put it behind an `Arc` at the end of `MastForest` construction. But maybe you see another way around it?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -27,6 +28,65 @@ use op_batch::OpBatchAccumulator;
...
+    /// which have not yet been inserted into a MAST forest.
+    Owned(DecoratorList),
+    /// The decorators are stored in a MAST forest and can be accessed via
+    /// this block's ID. The `Arc` provides shared, read-only access to the
+    /// forest's decorator storage.
+    Linked {
+        id: MastNodeId,
+        storage: Arc<DecoratorIndexMapping>,
+    },
+}
```
plafer: Is this meant as a long term solution, or a temporary one while waiting for the next PR? I'm basically trying to understand the path towards our ultimate goal of #1580 (and https://github.com/0xMiden/miden-vm/issues/1821#issuecomment-2920621174).

Or to recap, with this new `DecoratorStore`, we end up still storing decorator-related data within each node. But as I recall (one of) our ultimate goal is to be able to serialize all the decorator-related data at the end of the serialized representation, so that to remove this decorator/debug info data, we just need to truncate the serialized representation. As I understood it, #1776 is a step towards this goal: we first fully "separate" the decorator data from node data, which makes achieving that desired serialized representation more or less trivial.

Are we still working in this direction with this set of changes?

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Ooh, boy this was 🤡 . Thanks for pointing this out, I've advanced the schedule on the switch of all decorator-accessing methods to require a `&MastForest` in my latest addition to the PR, which incidentally revealed a bug on an edge case of the DecroatorIndexMapping data structure.

Good news: there's no `Arc` involved any more.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: At a high level, because we discussed this offline: #1580 is not in scope of the present PR, which resolutely takes the stance of being backwards compatible (take the custom `impl` of `serde::{Serialize, Deserialize}  for MastForest` as a witness of that strategy).

We're no longer storing decorator-related data within each node, because once the `MastForest` is finalized, all blocks are `Linked`. At that point, decorator information is a simple integer tag (their `MastNodeId`), which is the minimal amount of information needed to link that individual node to their decorator metadata in the `MastForest`-level `decorators` (which is keyed by that `MastNodeId`). 

Once we do decide to tackle changes in serialization, we really only have to work with the `MastForest::decorators` field: this is a `DecoratorIndexMapping` which only contains trivially (de)serializable information (vectors of small integers). 

So I would say this PR is a meaningful step towards #1580. I would wait to deal with `before_enter`/`after_exit` decorators before changing the serialization in one pass.

core/src/mast/decorator_storage.rs
huitseeker: > FWIU this decorator storage will grow to include all nodes

We will definitely use the same notion of CSR in the future, but in a simplified and separate data structure.
That is, `DecoratorIndexMapping` is a two-level CSR presenting the following mapping:
```
MastNodeId -> op_idx slice,
op_idx -> decorator_id slice,
```
We will have a simplified data structure for `before_enter`/`after_exit` decorators that will work with much of same algorithmics, but only represent a one-level mapping:
```
MastNodeIf -> decorator_id slice
```

I _might_ look into abstracting over the structure involved and see if I can nest two instances of a 1-level CSR to represent the nested CSR that `DecoratorIndexMapping` fundamentally is. But I haven't gone down that road so far.

core/src/mast/merger/mod.rs
plafer, line 120: Are we properly updating the `MastForest.decorator_storage`? I basically want to make sure that we're taking into account that `DecoratorIndexMapping.decorator_ids` will now be stale, and somehow we have to make sure we properly map those decorator IDs too. Might already be taken care of, but want to make sure we don't forget about it.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -85,14 +89,29 @@ pub struct BasicBlockNode {
     /// Multiple batches are used for blocks consisting of more than 72 operations.
     op_batches: Vec<OpBatch>,
     digest: Word,
-    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Vec::is_empty"))]
-    decorators: DecoratorList,
-    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Vec::is_empty"))]
+    /// This replaces the old `decorators` field
```
plafer: I would remove this sentence, since it's no longer clear what the "old" decorators field was

core/src/mast/mod.rs
```diff
@@ -71,6 +72,10 @@ pub struct MastForest {
     /// codes, so they are stored in order to provide a useful message to the user in case a error
     /// code is triggered.
     error_codes: BTreeMap<u64, Arc<str>>,
+
+    /// Provides efficient access to decorators per operation per node during execution and
+    /// debugging.
+    decorator_storage: DecoratorIndexMapping,
```
plafer, line 78: How about merging this field with `decorators` into a new type? The type would basically encapsulate everything needed to manage decorators in one place.

Alternatively, moving the `decorators` field down here would at least put them visually together.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Done!

core/src/mast/mod.rs
huitseeker: Your first remark is approached in #2323. The second is done in an additional commit here.

core/src/mast/merger/mod.rs
huitseeker: The `decorator_id_mappings` vector in `MastForestMerger` tracks all remappings, and these are used to remap decorator IDs in nodes via `build_with_remapped_children()`. `remove_nodes` and its caller clean up the old decorator info. I've added a dedicated test `mast_forest_merge_op_indexed_decorators_preservation` to this stack of PRs, lmk if you like it.

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1129 @@
...
+                while it.peek().is_some_and(|(i, _)| *i == op) {
+                    self.decorator_ids.push(it.next().unwrap().1);
+                }
+            }
+            // final sentinel for this node
+            self.op_indptr_for_dec_ids.push(self.decorator_ids.len());
+
+            // Push end pointer for this node (index of last op pointer)
+            let end_ops = self.op_indptr_for_dec_ids.len() - 1;
+            let _ = self.node_indptr_for_op_idx.push(end_ops);
```
plafer: We should `expect()` (or handle error if appropriate) instead of silently ignoring the error. Applies to every other place where we use this pattern

core/src/mast/node/basic_block_node/mod.rs
huitseeker: This has been refactored to stg w/o any cloning.

core/src/mast/decorator_storage.rs
huitseeker: good catch, of course!

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1135 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+pub struct DecoratorIndexMapping {
+    /// All the decorator IDs per operation per node, in a CSR relationship with
+    /// node_indptr_for_op_idx and op_indptr_for_dec_ids
+    decorator_ids: Vec<DecoratorId>,
+    /// For the node which operation indices are in op_indptr_for_dec_ids[node_start, node_end],
+    /// the indices of its i-th operation are at decorator_ids[op_indptr_for_dec_ids[node_start
+    /// + i]..op_indptr_for_dec_ids[node_start + i + 1]]
```
adr1anh: I think this would render better using quotes and perhaps a separate line?

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1135 @@
...
+/// operation per node.
+///
+/// This structure provides efficient access to decorator IDs in a hierarchical manner:
+/// 1. First level: Node -> Operations (operations belong to nodes)
+/// 2. Second level: Operation -> Decorator IDs (decorator IDs belong to operations)
+///
+/// The data layout follows CSR format at both levels:
+/// - `decorator_ids`: Flat storage of all DecoratorId values
+/// - `op_indptr_for_dec_ids`: Pointer indices for operations within decorator_ids
+/// - `node_indptr_for_op_idx`: Pointer indices for nodes within op_indptr_for_dec_ids
```
Al-Kindi-0: The description of the fields is very terse, I would try to write this in a more simple way, even if at the cost of being verbose (unless this is done somewhere else)

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1135 @@
...
+/// node_indptr_for_op_idx: [0, 2, 3]   // Node 0: [0,2], Node 1: [2,3]
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+pub struct DecoratorIndexMapping {
+    /// All the decorator IDs per operation per node, in a CSR relationship with
+    /// node_indptr_for_op_idx and op_indptr_for_dec_ids
+    decorator_ids: Vec<DecoratorId>,
+    /// For the node which operation indices are in op_indptr_for_dec_ids[node_start, node_end],
```
Al-Kindi-0: which -> who's ?

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1135 @@
...
+            }
+        }
+
+        for window in node_slice.windows(2) {
+            if window[0] > window[1] {
+                return Err(DecoratorIndexError::InternalStructure);
+            }
+        }
+
+        Ok(Self {
```
Al-Kindi-0, line 185: Q: aren't there some restrictions on admissible inputs to this function? I am thinking not per component but between components

core/src/mast/decorator_storage.rs
huitseeker: addressed in 44308ee423c9f5bc38b6878ca4849bc2f23e3e7d

core/src/mast/decorator_storage.rs
huitseeker: addressed in 44308ee423c9f5bc38b6878ca4849bc2f23e3e7d

core/src/mast/decorator_storage.rs
huitseeker: Yes, good remark, I've made its purpose implicit by making it test-only.

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1174 @@
...
+///
+/// The data layout follows CSR format at both levels:
+///
+/// - `decorator_ids` contains all the DecoratorId values in a single flat array. These are the
+///   actual decorator identifiers that need to be accessed. We store them contiguously to minimize
+///   memory overhead and improve cache locality when iterating.
+///
+/// - `op_indptr_for_dec_ids` stores pointer indices that map operations to their position within
+///   the `decorator_ids` array. For each operation, it contains the start index where that
+///   operation's decorator IDs begin in the flat storage.
```
Al-Kindi-0, line 26: nit: what about the final value in this array, which is a bit special, right?

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1174 @@
...
+///
+/// - `node_indptr_for_op_idx` stores pointer indices that map nodes to their position within the
+///   `op_indptr_for_dec_ids` array. For each node, it contains the start index where that node's
+///   operation pointers begin.
```
Al-Kindi-0, line 30: same comment as above

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1174 @@
...
+/// Node 0, Op 0: [decorator_id_0, decorator_id_1]
+/// Node 0, Op 1: [decorator_id_2]
+/// Node 1, Op 0: [decorator_id_3, decorator_id_4, decorator_id_5]
+/// ```
+///
+/// This would be stored as:
+/// ```text
+/// decorator_ids:         [0, 1, 2, 3, 4, 5]
+/// op_indptr_for_dec_ids: [0, 2, 3, 6]  // Node 0: ops [0,2], [2,3]; Node 1: ops [3,6]
+/// node_indptr_for_op_idx: [0, 2, 3]   // Node 0: [0,2], Node 1: [2,3]
```
Al-Kindi-0, line 49: the comment on the comment seem to need updating as well

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1174 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+pub struct DecoratorIndexMapping {
+    /// All the decorator IDs per operation per node, in a CSR relationship with
+    /// node_indptr_for_op_idx and op_indptr_for_dec_ids
+    decorator_ids: Vec<DecoratorId>,
+    /// For the node whose operation indices are in
+    /// `op_indptr_for_dec_ids[node_start..node_end]`,
+    /// the indices of its i-th operation are at:
```
Al-Kindi-0, line 60: nit: I would add something like "where `node_end` is `next_node_start` ..." with the convention about the final node index

core/src/mast/decorator_storage.rs
```diff
@@ -0,0 +1,1174 @@
...
+    /// ```text
+    /// decorator_ids[op_indptr_for_dec_ids[node_start + i]..
+    ///               op_indptr_for_dec_ids[node_start + i + 1]]
+    /// ```
+    op_indptr_for_dec_ids: Vec<usize>,
+    /// The decorated operation indices for the n-th node are at
+    /// `op_indptr_for_dec_ids[n..n+1]`
```
Al-Kindi-0, line 67: does this need updating as well?

Al-Kindi-0 (commented):
Looks great

core/src/mast/decorator_storage.rs
huitseeker: This is detailed on the comment on the field `op_indptr_for_dec_ids` you're looking at the struct-level comment.

core/src/mast/decorator_storage.rs
huitseeker: Same answer.

core/src/mast/decorator_storage.rs
huitseeker: Yes!

core/src/mast/decorator_storage.rs
huitseeker: Will add unit test elucidating this in my next PR.

PR #2314, opened 2025-10-28
fix: resolve testing feature gating issue in miden-assembly

huitseeker:
- Add miden-assembly as dev-dependency with testing feature to ensure testing functionality is available during test runs
- Fix feature gating for parse_forms method to only require testing feature

This resolves a compilation error when running tests without explicit testing feature: `cargo test -p miden-assembly`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1391,6 +1391,7 @@ dependencies = [
  "env_logger",
  "insta",
  "log",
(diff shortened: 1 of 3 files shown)


adr1anh (approved):
LGTM

plafer (approved):
🙏

PR #2322, opened 2025-10-29
In-VM AEAD Implementation using crypto_stream instruction

Al-Kindi-0:
## Description

  This PR implements AEAD (Authenticated Encryption with Associated Data) in the Miden standard library using the `crypto_stream` instruction. The implementation follows the `MonkeySpongeWrap` construction with RPO permutation.

  ## Design

  ### Automatic Padding
  - Encryption automatically appends padding block `[1,0,0,0,0,0,0,0]`
  - `num_blocks` parameter specifies data blocks only (padding not included)
  - Memory layout: data blocks → padding block → authentication tag

  ### Non-Deterministic Decryption
  Decryption uses two-phase verification:
  1. Event handler reads full ciphertext (data + padding + tag), verifies tag, provides data blocks via advice
  2. MASM procedure loads data blocks, calls encrypt to re-add padding, re-computes and verifies tag
  3. Tag serves as a commitment to both plaintext and ciphertext, in addition to its use in authentication

  ## Limitations

  ### Block Alignment
  - Operations work on 8-element blocks only
  - Memory addresses must be word-aligned (divisible by 4)
  - Callers must pad data to 8-element boundaries

  ### Associated Data
  Associated data (AD) is not currently supported.

  ## Follow-Up

  The `CRYPTOSTREAM` instruction has high algebraic degree in AIR constraints. #2328 vacates the opcode value of `HORNERBASE` after reducing its constraint degree so that it can be used by `CRYPTOSTREAM`

### DIFF
--- a/libcore/asm/crypto/aead.masm
+++ b/libcore/asm/crypto/aead.masm
@@ -10,7 +10,7 @@
 #! Note: Associated data (AD) is currently NOT supported. Only empty AD is handled,
 #! which is represented by the padding block [1,0,0,0,0,0,0,0].
 
-use std::crypto::hashes::rpo
+use miden::core::crypto::hashes::rpo256
 
 # Constants
 # ================================================================================================
@@ -209,8 +209,8 @@ pub proc encrypt
     hperm
     # => [rate_final(8), capacity_final(4), src_ptr_final, dst_ptr_final, ...]
 
-    # Extract the authentication tag (first word of rate) using rpo::squeeze_digest
-    exec.rpo::squeeze_digest
+    # Extract the authentication tag (first word of rate) using rpo256::squeeze_digest
+    exec.rpo256::squeeze_digest
     # => [tag(4), src_ptr_final, dst_ptr_final, ...]
 
     # Clean up stack - remove the two pointers
--- a/libcore/tests/crypto/aead.rs
+++ b/libcore/tests/crypto/aead.rs
@@ -8,7 +8,7 @@ fn test_encrypt_zero_blocks_roundtrip() {
     // Verifies that aead::encrypt handles num_blocks = 0 by encrypting only the padding block
     // and producing a tag, and that aead::decrypt accepts it and succeeds.
     let source = r#"
-    use std::crypto::aead
+    use miden::core::crypto::aead
 
     begin
         # No plaintext needed; num_blocks = 0
@@ -70,7 +70,7 @@ fn test_encrypt_with_known_values() {
     // Build MASM test dynamically with extracted values
     let source = format!(
         "
-    use std::crypto::aead
+    use miden::core::crypto::aead
 
     begin
         # Store plaintext [10,11,12,13,14,15,16,17] at address 1000
@@ -148,7 +148,7 @@ fn test_decrypt_with_known_values() {
     // Build MASM test for decryption
     let source = format!(
         "
-    use std::crypto::aead
+    use miden::core::crypto::aead
 
     begin
         # Store ciphertext at address 1000 (data + padding + tag)
@@ -228,7 +228,7 @@ fn test_decrypt_with_wrong_key() {
     // Build MASM test that uses wrong key for decryption
     let source = format!(
         "
-    use std::crypto::aead
+    use miden::core::crypto::aead
 
     begin
         # Store ciphertext at address 1000
--- a/libcore/src/lib.rs
+++ b/libcore/src/lib.rs
@@ -81,6 +81,7 @@ impl CoreLibrary {
             (FALCON_DIV_EVENT_NAME, Arc::new(handle_falcon_div)),
             (LOWERBOUND_ARRAY_EVENT_NAME, Arc::new(handle_lowerbound_array)),
             (LOWERBOUND_KEY_VALUE_EVENT_NAME, Arc::new(handle_lowerbound_key_value)),
+            (AEAD_DECRYPT_EVENT_NAME, Arc::new(handle_aead_decrypt)),
         ]
     }
 
--- a/libcore/tests/crypto/falcon.rs
+++ b/libcore/tests/crypto/falcon.rs
@@ -13,10 +13,7 @@ use miden_processor::{
 };
 use miden_utils_testing::{
     Word,
-    crypto::{
-        MerkleStore, Rpo256,
-        falcon512_rpo::{Polynomial, SecretKey},
-    },
+    crypto::{MerkleStore, Rpo256},
     expect_exec_error_matches,
     proptest::proptest,
     rand::rand_value,


stdlib/tests/crypto/aead.rs
```diff
@@ -0,0 +1,318 @@
...
+    begin
+        # Store plaintext [10,11,12,13,14,15,16,17] at address 1000
+        push.10.11.12.13 push.1000 mem_storew_be dropw
+        push.14.15.16.17 push.1004 mem_storew_be dropw
+
+        # Encrypt 1 block with key and nonce from reference
+        push.1           # num_blocks = 1
+        push.2000        # dst_ptr
+        push.1000        # src_ptr
+        push.{}.{}.{}.{}     # key
```
adr1anh: I _think_ we can do this
```suggestion
        push.{key_elements:?}     # key
```

core/src/operations/mod.rs
```diff
@@ -516,6 +517,25 @@ pub enum Operation {
...
+    /// - The resulting ciphertext is written to destination memory.
+    /// - The top 8 elements of the stack are updated with the ciphertext (becomes new rate).
+    /// - Elements 8-11 (capacity) remain unchanged.
+    /// - Source address (position 12) is incremented by 8.
+    /// - Destination address (position 13) is incremented by 8.
+    /// - All other stack elements remain the same.
+    ///
+    /// This instruction is designed for AEAD encryption using the RPO hash function as a stream
+    /// cipher. After this operation, an application of `hperm` is expected in order to refresh
+    /// the keystream for the next block.
```
adr1anh: simplify?

processor/src/operations/crypto_ops.rs
```diff
@@ -186,6 +188,129 @@ impl Process {
...
+        let src_ptr = self.stack.get(SRC_PTR_IDX);
+        let dst_ptr = self.stack.get(DST_PTR_IDX);
+
+        // Calculate addresses for the two words
+        let src_word1_addr = src_ptr;
+        let src_word2_addr = src_ptr + Felt::from(WORD_SIZE as u32);
+        let dst_word1_addr = dst_ptr;
+        let dst_word2_addr = dst_ptr + Felt::from(WORD_SIZE as u32);
+
+        if src_ptr == dst_ptr {
```
adr1anh: Doesn't this get enforced by the memory component?

processor/src/operations/crypto_ops.rs
```diff
@@ -186,6 +188,129 @@ impl Process {
...
+            .map_err(ExecutionError::MemoryError)?;
+
+        // Update stack[0..7] with ciphertext (becomes new rate for next hperm)
+        // Elements are in reverse order on stack (stack[0] is top)
+        for (i, &elem) in ciphertext_words.iter().flat_map(|word| word.iter()).rev().enumerate() {
+            self.stack.set(i, elem);
+        }
+
+        // Copy capacity elements (stack[8..11]) to preserve them
+        for i in 8..SRC_PTR_IDX {
```
adr1anh, line 295: Should we use the range defined by the hasher chiplet?

processor/src/operations/crypto_ops.rs
```diff
@@ -186,6 +188,129 @@ impl Process {
...
+            let value = self.stack.get(i);
+            self.stack.set(i, value);
+        }
+
+        // Update source pointer: src_ptr + 8
+        self.stack.set(SRC_PTR_IDX, src_ptr + Felt::from((WORD_SIZE * 2) as u32));
```
adr1anh: nit: `Felt::new((WORD_SIZE * 2) as u64)` as this is now a constant

processor/src/operations/crypto_ops.rs
```diff
@@ -186,6 +188,129 @@ impl Process {
...
+            self.stack.get(5),
+            self.stack.get(4),
+            self.stack.get(3),
+            self.stack.get(2),
+            self.stack.get(1),
+            self.stack.get(0),
+        ];
+
+        // Encrypt: ciphertext = plaintext + rate (element-wise addition)
+        let mut ciphertext_words = [[ZERO; WORD_SIZE]; 2];
```
adr1anh: nit: I would split this into 2 loops for each word with explicit word idx 0,1

processor/src/processor/operations/io_ops.rs
```diff
@@ -255,3 +257,109 @@ pub(super) fn op_pipe<P: Processor>(
...
+    const DST_PTR_IDX: usize = 13;
+
+    let ctx = processor.system().ctx();
+    let clk = processor.system().clk();
+
+    // Get source and destination pointers
+    let src_addr = processor.stack().get(SRC_PTR_IDX);
+    let dst_addr = processor.stack().get(DST_PTR_IDX);
+
+    if src_addr == dst_addr {
```
adr1anh: I'd be more in favor of removing this and letting the memory catch the concurrent access. We're also missing the case where there is only a 1-word overlap.

processor/src/processor/operations/io_ops.rs
```diff
@@ -255,3 +257,109 @@ pub(super) fn op_pipe<P: Processor>(
...
+        processor.stack().get(5),
+        processor.stack().get(4),
+        processor.stack().get(3),
+        processor.stack().get(2),
+        processor.stack().get(1),
+        processor.stack().get(0),
+    ];
+
+    // Encrypt: ciphertext = plaintext + rate (element-wise addition in field)
+    let ciphertext_word1 = [
```
adr1anh: this is a lot clearer than the slow processor.

stdlib/tests/crypto/aead.rs
```diff
@@ -0,0 +1,318 @@
...
+    let ciphertext = encrypted.ciphertext();
+
+    // Build MASM test that uses wrong key for decryption
+    let source = format!(
+        "
+    use.std::crypto::aead
+
+    begin
+        # Store ciphertext at address 1000
+        push.{}.{}.{}.{} push.1000 mem_storew_be dropw
```
adr1anh: PR #2325 provides some helper functions for generating MASM code that initializes memory.

Cargo.toml
```diff
@@ -53,7 +53,7 @@ miden-utils-testing = { path = "./crates/test-utils", version = "0.1", package =
 miden-verifier = { path = "./verifier", version = "0.19", default-features = false }
 
 # Miden crates
-miden-crypto = { version = "0.18", default-features = false }
+miden-crypto = { git = "https://github.com/0xMiden/crypto.git", branch = "al-aead-constructors", default-features = false }
```
adr1anh: Note that we should update this when we update `miden-crypto`

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+/// - This handler reads ALL elements: data blocks + padding + tag
+///
+/// The MASM decrypt procedure will then:
+/// 1. Load the plaintext data blocks from advice and write to dst_ptr
+/// 2. Call encrypt which reads the data blocks and adds padding automatically
+/// 3. Re-encrypt data + padding to compute authentication tag
+/// 4. Compare computed tag with expected tag and halt if they don't match
+///
+/// Security: Tag verification happens TWICE in this design:
+/// 1. Once in this event handler (via reference implementation)
```
adr1anh: Would it make sense to add a feature where this can be disabled for performance? I'm not sure the "defense-in-depth" argument is totally accurate as it only really helps when we are re-executing the program.

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+/// provides defense in depth and ensures cryptographic soundness.
+///
+/// Non-determinism soundness: Using advice for decryption is cryptographically sound
+/// because:
+/// 1. The event handler verifies the tag before providing plaintext
+/// 2. The MASM procedure re-verifies the tag via re-encryption
+/// 3. The deterministic encryption creates a bijection between plaintext and ciphertext
+/// 4. A malicious prover cannot provide incorrect plaintext without causing tag mismatch
+pub fn handle_aead_decrypt(process: &ProcessState) -> Result<Vec<AdviceMutation>, EventError> {
+    // Stack: [nonce(4), key(4), src_ptr, dst_ptr, num_blocks, ...]
```
adr1anh: missing `event_id` in the first slot

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    //   src_ptr = ciphertext + encrypted_padding + tag location (input)
+    //   dst_ptr = plaintext destination (output)
+    //   num_blocks = number of plaintext data blocks (NO padding)
+
+    // Read parameters from stack
+    // Note: Stack position 0 contains the Event ID when the handler is called,
+    // so the actual parameters start at position 1
+    // Also note: Words on the stack are stored in reverse element order
+    let nonce_word = Word::new([
+        process.get_stack_item(4),
```
adr1anh: `get_word`?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    let key_word = Word::new([
+        process.get_stack_item(8),
+        process.get_stack_item(7),
+        process.get_stack_item(6),
+        process.get_stack_item(5),
+    ]);
+
+    let src_ptr = process.get_stack_item(9).as_int();
+    let _dst_ptr = process.get_stack_item(10).as_int(); // Not needed for decryption
```
adr1anh: remove or comment out?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    let num_blocks = process.get_stack_item(11).as_int();
+
+    // Read full ciphertext from memory including padding block
+    // Total: (num_blocks + 1) * 8 elements (data blocks + padding block)
+    let num_ciphertext_elements = ((num_blocks + 1) * 8) as usize;
+    let mut ciphertext = Vec::with_capacity(num_ciphertext_elements);
+    let ctx = process.ctx();
+    for i in 0..num_ciphertext_elements {
+        let addr = (src_ptr + i as u64) as u32;
```
adr1anh: This technically could panic

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    let src_ptr = process.get_stack_item(9).as_int();
+    let _dst_ptr = process.get_stack_item(10).as_int(); // Not needed for decryption
+    let num_blocks = process.get_stack_item(11).as_int();
+
+    // Read full ciphertext from memory including padding block
+    // Total: (num_blocks + 1) * 8 elements (data blocks + padding block)
+    let num_ciphertext_elements = ((num_blocks + 1) * 8) as usize;
+    let mut ciphertext = Vec::with_capacity(num_ciphertext_elements);
+    let ctx = process.ctx();
+    for i in 0..num_ciphertext_elements {
```
adr1anh: I _think_ we have a helper method for this in mod.rs, but if not we should probably add one.

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // Read the authentication tag (4 elements at src_ptr + (num_blocks + 1) * 8)
+    let tag_addr = (src_ptr + (num_blocks + 1) * 8) as u32;
+    let mut tag_elements = [miden_core::Felt::ZERO; 4];
+    for i in 0..4 {
+        let addr = tag_addr + i as u32;
+        let value =
+            process
+                .get_mem_value(ctx, addr)
```
adr1anh: `get_mem_word`?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+                .ok_or_else(|| AeadDecryptError::MemoryReadFailed {
+                    addr,
+                    reason: alloc::string::String::from("memory read returned None"),
+                })?;
+        ciphertext.push(value);
+    }
+
+    // Read the authentication tag (4 elements at src_ptr + (num_blocks + 1) * 8)
+    let tag_addr = (src_ptr + (num_blocks + 1) * 8) as u32;
+    let mut tag_elements = [miden_core::Felt::ZERO; 4];
```
adr1anh: We can just import `ZERO`

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    let mut ciphertext = Vec::with_capacity(num_ciphertext_elements);
+    let ctx = process.ctx();
+    for i in 0..num_ciphertext_elements {
+        let addr = (src_ptr + i as u64) as u32;
+        let value =
+            process
+                .get_mem_value(ctx, addr)
+                .ok_or_else(|| AeadDecryptError::MemoryReadFailed {
+                    addr,
+                    reason: alloc::string::String::from("memory read returned None"),
```
adr1anh: nit: import `String`

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    // Extract only the data blocks (without padding) to insert into advice
+    // The MASM encrypt procedure will add padding automatically during re-encryption
+    let data_blocks_count = (num_blocks * 8) as usize;
+    let plaintext_data: Vec<_> =
+        plaintext_with_padding.into_iter().take(data_blocks_count).collect();
+
+    // Insert plaintext data (WITHOUT padding) into advice map with key = nonce (4 elements)
+    // The padding will be added by the MASM encrypt procedure during re-encryption
+    // Create an AdviceMap from the iterator
+    let advice_map: miden_core::AdviceMap =
```
adr1anh: Not for this PR, but we should probably add a constructor to AdviceMap to create one from a single entry. 

Nit: import AdviceMap

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+    let plaintext_with_padding = secret_key.decrypt_elements(&encrypted_data).map_err(|e| {
+        AeadDecryptError::DecryptionFailed {
+            reason: alloc::format!("decryption failed: {:?}", e),
+        }
+    })?;
+
+    // Extract only the data blocks (without padding) to insert into advice
+    // The MASM encrypt procedure will add padding automatically during re-encryption
+    let data_blocks_count = (num_blocks * 8) as usize;
+    let plaintext_data: Vec<_> =
```
adr1anh:
```suggestion
    let plaintext_data = plaintext_with_padding[..data_blocks_count].to_vec();
```

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+/// Error types that can occur during AEAD decryption.
+#[derive(Debug, thiserror::Error)]
+enum AeadDecryptError {
+    /// Memory read failed.
+    #[error("failed to read memory at address {addr}: {reason}")]
+    MemoryReadFailed { addr: u32, reason: alloc::string::String },
+
+    /// Decryption failed.
+    #[error("decryption failed: {reason}")]
+    DecryptionFailed { reason: alloc::string::String },
```
adr1anh: This is only used once it seems, can we just include the error with `#[source]`?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// TESTS
+// ================================================================================================
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
adr1anh, line 141: not sure this is necessary

stdlib/tests/crypto/aead.rs
Al-Kindi-0: good point

core/src/operations/mod.rs
Al-Kindi-0: done

processor/src/operations/crypto_ops.rs
Al-Kindi-0: yes, though I think it is helpful to have it here as well as the error would be more informative
Though happy to remove it if you think otherwise

processor/src/operations/crypto_ops.rs
Al-Kindi-0: I think that includes the chiplet offset and also the coupling between the two is not tight in my opinion

processor/src/operations/crypto_ops.rs
Al-Kindi-0: done

processor/src/processor/operations/io_ops.rs
Al-Kindi-0: indeed, switched to this implementation there as well

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: removed the comment as it was misleading
The tag verification happens naturally as part of the decryption outside the VM, and as part of the MASM encryption procedure inside the VM. The comments have been updated to express this

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: done, thanks

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: done

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: done

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: Indeed, fixed it to use checked addition

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: created, let me know if it is lacking in any aspect

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: simplified

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: not relevant anymore

core/src/operations/mod.rs
```diff
@@ -516,6 +517,23 @@ pub enum Operation {
...
+    /// - Two consecutive words (8 elements) are loaded from source memory.
+    /// - Each element is added to the corresponding element in the rate (top 8 stack elements).
+    /// - The resulting ciphertext is written to destination memory.
+    /// - The top 8 elements of the stack are updated with the ciphertext (becomes new rate).
+    /// - Elements 8-11 (capacity) remain unchanged.
+    /// - Source address (position 12) is incremented by 8.
+    /// - Destination address (position 13) is incremented by 8.
+    /// - All other stack elements remain the same.
+    ///
+    /// After this operation, `hperm` should be applied to refresh the keystream for the next block.
```
adr1anh, line 536: This could be a bit clearer with an explicit stack transition

processor/src/operations/crypto_ops.rs
adr1anh: Fair, I ended up re-defining the constants in the log-precompile PR but maybe we can find a way later to consolidate

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,136 @@
...
+/// This handler is called when the VM emits an AEAD_DECRYPT_EVENT. It reads the full
+/// ciphertext (including padding block) and tag from memory, performs decryption and
+/// tag verification using AEAD-RPO, then provides the plaintext via the advice map.
+///
+/// Process:
+/// 1. Reads full ciphertext from memory at src_ptr ((num_blocks + 1) * 8 elements)
+/// 2. Reads authentication tag from memory at src_ptr + (num_blocks + 1) * 8
+/// 3. Constructs EncryptedData and decrypts using AEAD-RPO
+/// 4. Extracts only the data blocks (first num_blocks * 8 elements) from plaintext
+/// 5. Inserts the data blocks (WITHOUT padding) into the advice map (keyed by nonce)
```
adr1anh: Would we want to push it to the AdviceStack instead?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,136 @@
...
+        EncryptedData::from_parts(DataType::Elements, ciphertext, auth_tag, nonce.clone());
+
+    // Decrypt using the standard reference implementation
+    // This performs tag verification internally
+    let plaintext_with_padding = secret_key.decrypt_elements(&encrypted_data)?;
+
+    // Extract only the data blocks (without padding) to insert into advice
+    // The MASM encrypt procedure will add padding automatically during re-encryption
+    let data_blocks_count = (num_blocks * 8) as usize;
+    let plaintext_data = plaintext_with_padding[..data_blocks_count].to_vec();
```
adr1anh: Nit: truncate to avoid re-allocation

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Decryption failed (wraps EncryptionError from miden-crypto).
+    #[error(transparent)]
+    DecryptionFailed(#[from] EncryptionError),
+}
+
+// TESTS
+// ================================================================================================
+
+#[test]
+fn test_event_name() {
```
adr1anh: I think this is self-explanatory

stdlib/src/handlers/mod.rs
```diff
@@ -15,3 +17,42 @@ fn u64_to_u32_elements(value: u64) -> (Felt, Felt) {
...
+/// `Some(Vec<Felt>)` with `len` elements, or `None` if any check fails
+///
+/// # Example
+/// ```ignore
+/// let elements = read_memory_region(process, ctx, src_ptr, num_elements)
+///     .ok_or(MyError::MemoryReadFailed)?;
+/// ```
+pub(crate) fn read_memory_region(
+    process: &ProcessState,
+    ctx: ContextId,
```
adr1anh: Do we need to support `ctx != process.ctx()`?

core/src/operations/mod.rs
adr1anh:
```rust
    /// - The resulting ciphertext `ct` is written to destination memory and replaces the rate.
    /// - Source and destination addresses are incremented by 8.
    ///
    /// Stack transition:
    /// ```text
    /// [rate(8), cap(4),  src,   dst,   ...]
    ///     ↓
    /// [ct(8),   cap'(4), src+8, dst+8, ...]
    /// ```
    /// where ct = mem[src] + rate.
```

processor/src/processor/operations/io_ops.rs
```diff
@@ -255,3 +257,107 @@ pub(super) fn op_pipe<P: Processor>(
...
+        plaintext_word1[0] + rate[0],
+        plaintext_word1[1] + rate[1],
+        plaintext_word1[2] + rate[2],
+        plaintext_word1[3] + rate[3],
+    ];
+    let ciphertext_word2 = [
```
adr1anh: nit: Convert to `Word` here to avoid `into()` later

stdlib/src/handlers/mod.rs
```diff
@@ -15,3 +17,42 @@ fn u64_to_u32_elements(value: u64) -> (Felt, Felt) {
...
+/// This is a safe wrapper around memory reads that:
+/// - Validates the starting address fits in u32
+/// - Validates the length doesn't overflow when converted to u32
+/// - Uses checked arithmetic to compute the end address
+/// - Returns `None` if any validation fails or if any memory location is uninitialized
+///
+/// # Arguments
+/// * `process` - Process state to read memory from
+/// * `ctx` - Memory context ID
+/// * `start_ptr` - Starting address (u64 from stack)
```
adr1anh: should we make this word aligned?

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,136 @@
...
+    let ciphertext = read_memory_region(process, ctx, src_ptr, num_ciphertext_elements).ok_or(
+        AeadDecryptError::MemoryReadFailed {
+            addr: src_ptr,
+            len: num_ciphertext_elements,
+        },
+    )?;
+
+    // Read authentication tag: 4 elements immediately after ciphertext
+    let tag_ptr = src_ptr + num_ciphertext_elements;
+    let tag_vec = read_memory_region(process, ctx, tag_ptr, 4)
```
adr1anh: we could use `process.get_mem_word` but we have to sanitize `tag_ptr` first

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,363 @@
...
+    # Move num_blocks to accessible position
+    movup.14
+    # => [num_blocks, rate(8), capacity(4), src_ptr, dst_ptr, ...]
+
+    # Encrypt each block using crypto_stream
+    dup neq.0
+    while.true
+
+        # Decrement the counter
+        push.1 sub
```
adr1anh: `sub.1`?

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,363 @@
...
+    crypto_stream
+    # => [encrypted_padding(8), capacity'(4), src_ptr_final+8, dst_ptr_final+8, ...]
+
+    # Generate authentication tag from current sponge state
+    # Apply final permutation
+    hperm
+    # => [rate_final(8), capacity_final(4), src_ptr_final, dst_ptr_final, ...]
+
+    # Extract the authentication tag (first word of rate)
+    dropw       # Drop rate[7..4]
```
adr1anh: We have `rpo::squeeze_digest` which does the same

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,363 @@
...
+    # Push nonce onto advice stack to use as map key for loading plaintext
+    adv.push_mapval
+    # This pushes the plaintext elements from advice map onto advice stack
+    # The advice contains data blocks only (no padding) - padding is added by encrypt
+    # => [nonce(4), key(4), src_ptr, dst_ptr, num_blocks, tag_ptr, ...]
+
+    # Prepare to load plaintext data blocks from advice
+    dup.9   # dst_ptr
+    dup.11  # num_blocks
+    swap
```
adr1anh: can we reorder the above to avoid the swap?

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,363 @@
...
+    dup.13 neq.0
+    # => [continue?, 0(4), 0(4), 0(4), dst_ptr, num_blocks, nonce(4), ...]
+
+    # Load plaintext data blocks from advice using adv_pipe
+    while.true
+        adv_pipe  # Reads 8 elements from advice, writes to dst_ptr, increments dst_ptr by 8
+        # => [DATA1(4), DATA0(4), 0(4), dst_ptr+8, num_blocks, nonce(4), key(4), src_ptr, dst_ptr, num_blocks, tag_ptr, ...]
+
+        movup.13 sub.1  # Decrement counter
+        movdn.13 dup.13 # Move counter back and check if more blocks remain
```
adr1anh: `dup movdn.14` ?

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,363 @@
...
+    exec.encrypt
+    # => [computed_tag(4), tag_ptr, ...]
+
+    # Verify computed tag matches expected tag from ciphertext
+    padw            # Prepare stack for loading expected tag
+    movup.8         # Get tag_ptr
+    mem_loadw_be    # Load expected tag from memory
+    # => [expected_tag(4), computed_tag(4), ...]
+
+    eqw assert      # Assert tags match (halts if different)
```
adr1anh: `assert_eqw.err="tag mismatch"`

core/src/operations/mod.rs
Al-Kindi-0: Done

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: Probably easier and better
Changed

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: You mean the section separator? Using just for consistency with the others

stdlib/src/handlers/mod.rs
Al-Kindi-0: not really
forgot to use `process.ctx()` and now fixed it

processor/src/processor/operations/io_ops.rs
Al-Kindi-0: good point

stdlib/src/handlers/mod.rs
Al-Kindi-0: yep, done

stdlib/src/handlers/aead_decrypt.rs
Al-Kindi-0: yep, done

stdlib/asm/crypto/aead.masm
Al-Kindi-0: yep, changed

stdlib/asm/crypto/aead.masm
Al-Kindi-0: I always forget to use it as finding the import path is more work than doing it as above :)
Changed it now to use the procedure

stdlib/asm/crypto/aead.masm
Al-Kindi-0: yep, should've been cleaned up after debugging

stdlib/asm/crypto/aead.masm
Al-Kindi-0: yep, should've been cleaned up as well
Changed now

stdlib/asm/crypto/aead.masm
Al-Kindi-0: Done

core/src/operations/mod.rs
```diff
@@ -118,6 +118,7 @@ pub(super) mod opcode_constants {
     pub const OPCODE_PUSH: u8           = 0b0101_1011;
     pub const OPCODE_DYNCALL: u8        = 0b0101_1100;
     pub const OPCODE_EVALCIRCUIT: u8    = 0b0101_1101;
+    pub const OPCODE_CRYPTOSTREAM: u8   = 0b0101_1110;  // TODO: swap with HORNERBASE (0x64) once HORNERBASE constraints are reduced to degree ≤4
```
plafer: Just a note to remember to address this before merging

core/src/operations/mod.rs
```diff
@@ -517,6 +518,24 @@ pub enum Operation {
...
+    /// ciphertext is written to destination memory and replaces the rate. Source and destination
+    /// addresses are incremented by 8.
+    ///
+    /// Stack transition:
+    /// ```text
+    /// [rate(8), cap(4), src, dst, ...]
+    ///     ↓
+    /// [ct(8), cap(4), src+8, dst+8, ...]
+    /// ```
+    /// where `ct = mem[src] + rate`.
```
plafer:
```suggestion
    /// where `ct = mem[src..src+8] + rate` (element-wise addition).
```

libcore/asm/crypto/aead.masm
```diff
@@ -0,0 +1,352 @@
+#! AEAD (Authenticated Encryption with Associated Data) implementation using RPO hash.
+#!
+#! This implementation uses the `crypto_stream` instruction for optimal performance.
```
plafer, line 3: How "optimal" do we really *need* this to be? Or, what part of the system is this to be used in, and how much leeway do we have?

Otherwise, if we have a bit of "performance leeway", we could either
- not add any new instruction at all
- add a more general-purpose one (e.g. `loadw_add`, which would load a word from memory, and add it to the word present in stack).

And basically we would implement the current `crypto_stream` instruction as a procedure. This is basically very specialized (unlikely to be used in any other context), and takes up a valuable "very high-degree" operation - so it would be good to make sure that we really *need* it.

Al-Kindi-0: As Bobbin mentioned offline, this will be used in the context of encrypted notes. Provided we get some typical numbers on the sizes of data we will be encrypting, we can make some back-of-the-envelop calculations to get some data points to inform our decision.
Intuitively, it seems to me that this is a very basic/fundamental primitive (i.e., encryption within a VM) and hence should have broad applicability and unlock applications which we may not be able to envision currently. Though, this is speculative from my side, I admit.

core/src/operations/mod.rs
Al-Kindi-0: done

core/src/operations/mod.rs
Al-Kindi-0: noted and removed

miden-vm/tests/integration/operations/crypto_ops.rs
```diff
@@ -257,3 +258,65 @@ fn mtree_update() {
...
+        crypto_stream
+    ";
+
+    let test = build_op_test!(asm_op, &[]);
+    let err = test.execute().expect_err("crypto_stream should reject in-place encryption");
+    assert!(matches!(
+        err,
+        ExecutionError::MemoryError(MemoryError::IllegalMemoryAccess { .. })
+    ));
+}
```
plafer, line 322: Will we be able to ensure that this is the case with constraints? Also separately, I can't remember if this is enforced by constraints in the context of the memory chiplet.

stdlib/src/handlers/aead_decrypt.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Decryption failed (wraps EncryptionError from miden-crypto).
+    #[error(transparent)]
+    DecryptionFailed(#[from] EncryptionError),
+}
+
+// TESTS
+// ================================================================================================
+
+#[test]
+fn test_event_name() {
```
plafer: We should put the test under a `#[cfg(test)] mod tests { ... }` module, so that the test functions don't get included in the main library.

libcore/asm/crypto/aead.masm
plafer: > Intuitively, it seems to me that this is a very basic/fundamental primitive (i.e., encryption within a VM) and hence should have broad applicability and unlock applications which we may not be able to envision currently.

And are all ciphers likely to use addition of new inputs with the rate? In any case I trust your "cryptography compass" more than mine, so happy to go with this instruction as designed.

Al-Kindi-0: I meant encryption in general as a cryptographic primitive, though this might warrant a discussion...

As for the first point, we evaluated several designs and settled on this one because it’s a minimal, standard instantiation: RPO as the permutation with a duplex schedule, combining plaintext with the keystream via field addition. So in principle, this op can be combined with any cryptographic permutation (e.g., Poseidon2) to obtain an AEAD.

Though, it might make sense to think a bit if we can make the op more general at the cost of some performance loss to the encryption use case. Will look into it

miden-vm/tests/integration/operations/crypto_ops.rs
Al-Kindi-0: I don't think we do currently, but maybe we should?
Reading and then writing, or vice-versa, is not correct semantically as this will break the map form `(context, address, clock) -> Value`.
The fix will be to enforce that when `(context, address, clock)` is constant across two rows, then both operations should be a READ.

plafer: Yup, makes sense. Should we create an issue? We should basically write that constraint down somewhere in the docs, and ideally in Airscript as well.

processor/src/processor/operations/io_ops.rs
huitseeker: How about doing a traditional overlap check?
```
  let src_end = src_addr + DOUBLE_WORD_SIZE;
  let dst_end = dst_addr + DOUBLE_WORD_SIZE;
  if src_addr < dst_end && dst_addr < src_end {
      return Err(ExecutionError::MemoryError(...));
  }
```
Note : the same issue is present in crypto_ops.

Whichever solution is adopted in the end, I'd highly advise unit tests covering this overlap case!

libcore/tests/crypto/aead.rs
huitseeker, line 1: This looks like a great place to add a few tests:
  - Empty plaintext (num_blocks = 0)
  - Partial overlap scenarios (should fail with proper error, see [thread](https://github.com/0xMiden/miden-vm/pull/2322/files#r2481449310))
  - Single-element/word memory access edge cases(?)

huitseeker (changes requested):
I'd like to understand the outcomes of this op in cases of overlap better.

processor/src/processor/operations/io_ops.rs
Al-Kindi-0: Good point!
I think letting the memory chiplet handle concurrent accesses is the write solution but since we are doing 2 reads and 2 writes and hence the potential for overlap, and as this is specific only to this op, I think that adding checks at op level is the way to go.

Went ahead and encapsulated the checks into a helper function to aid readability, and also added some tests to catch overlaps.

libcore/tests/crypto/aead.rs
Al-Kindi-0: Add overlap tests at the level of the op, as this where things can be problematic. Also, the unaligned memory access cases are tested for at the op level.

Added a test in `stdlib` for the case of an empty plaintext encryption.

stdlib/tests/crypto/ecdsa_secp256k1.rs
```diff
@@ -139,7 +139,7 @@ fn test_ecdsa_verify_impl_commitment() {
 /// Generates a valid signature using deterministic seed
 fn generate_valid_signature() -> EcdsaRequest {
     let mut rng = StdRng::seed_from_u64(42);
-    let secret_key = SecretKey::with_rng(&mut rng);
+    let mut secret_key = SecretKey::with_rng(&mut rng);
```
bobbinth: Question: why do we need secret keys to be mutable here? (and same question for line 160 below).

Cargo.toml
bobbinth: Do we have a PR for this in `miden-crypto`? I'm planning to release the new version by the end of the week and it would probably be good to include the changes in that release.

core/src/operations/mod.rs
```diff
@@ -118,6 +118,7 @@ pub(super) mod opcode_constants {
     pub const OPCODE_PUSH: u8           = 0b0101_1011;
     pub const OPCODE_DYNCALL: u8        = 0b0101_1100;
     pub const OPCODE_EVALCIRCUIT: u8    = 0b0101_1101;
+    pub const OPCODE_CRYPTOSTREAM: u8   = 0b0101_1110;
```
bobbinth: This will change once https://github.com/0xMiden/miden-vm/pull/2328 is merged, right? We should probably merge https://github.com/0xMiden/miden-vm/pull/2328 sooner rather than later and then update this PR.

processor/src/processor/operations/io_ops.rs
```diff
@@ -255,3 +256,94 @@ pub(super) fn op_pipe<P: Processor>(
 
     Ok(())
 }
+
+#[inline(always)]
+pub(super) fn op_crypto_stream<P: Processor>(
+    processor: &mut P,
+    err_ctx: &impl ErrorContext,
+    tracer: &mut impl Tracer,
+) -> Result<(), ExecutionError> {
```
bobbinth: We have this function defined twice (here and in `src/operations/crypto-ops.rs`) to support both the slow and the fast processors, right?

If so, why is it in `io_ops` here and in `crypto_ops` there?

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,353 @@
...
+    # => [AD_padding(8), capacity'(4), ...]
+end
+
+#! Encrypts plaintext data from memory using the `crypto_stream` instruction.
+#!
+#! This procedure encrypts plaintext and automatically adds a padding block at the end.
+#! The padding block [1, 0, 0, 0, 0, 0, 0, 0] is written to memory and encrypted, ensuring
+#! proper AEAD operation without requiring the caller to handle padding manually.
+#! This, however, requires the plaintext to be double-word aligned. This assumption is made
+#! both in this procedure as well as in the decryption procedure.
```
bobbinth: Is "double-word aligned" correct here? I though it just had to be word-aligned.

I think maybe what we wanted to say here is that the length of plaintext must be a multiple of 8 elements?

stdlib/asm/crypto/aead.masm
```diff
@@ -0,0 +1,353 @@
...
+#! - Blocks must be stored contiguously in memory
+#! - src_ptr and dst_ptr MUST be different (in-place encryption not supported)
+#!   This is because crypto_stream reads and writes in the same clock cycle
+#!
+#! Padding:
+#! - Padding is AUTOMATIC - caller should NOT pad the plaintext
+#! - The procedure writes [1, 0, 0, 0, 0, 0, 0, 0] to dst_ptr + (num_blocks * 8)
+#! - This padding block is then encrypted along with the data
+#! - For empty plaintext (num_blocks = 0), only the padding block is encrypted
+export.encrypt
```
bobbinth: This syntax is no longer supported - it should be `pub proc encrypt`.

Separately could we add some info about how many cycles it takes to encrypt data? Doesn't have to be exact, but would be good do know rough magnitude (e.g., encrypting 100 words will take 3K cycles).

stdlib/asm/crypto/aead.masm
Al-Kindi-0: Indeed, it is more about the length than memory alignment.
Updated

processor/src/processor/operations/io_ops.rs
Al-Kindi-0: Indeed, I was unsure which one to choose and this resulted in this inconsistency.
Changed it now so that both are in `crypto_ops`

core/src/operations/mod.rs
Al-Kindi-0: Makes sense

stdlib/tests/crypto/ecdsa_secp256k1.rs
Al-Kindi-0: Switched to `next` for `miden-crypto` dependency and now it is not needed

Cargo.toml
Al-Kindi-0: It is already part of `next`.
Changed the branch to `next` and then once we have a new release of `miden-crypto` we can come back and update this part

stdlib/asm/crypto/aead.masm
Al-Kindi-0: Done, and added rough estimates for the cycle count

libcore/asm/crypto/aead.masm
```diff
@@ -121,7 +121,12 @@ end
 #! - The procedure writes [1, 0, 0, 0, 0, 0, 0, 0] to dst_ptr + (num_blocks * 8)
 #! - This padding block is then encrypted along with the data
 #! - For empty plaintext (num_blocks = 0), only the padding block is encrypted
-export.encrypt
+#!
+#! Cycles (estimate): 77 + 2 * n
```
bobbinth, line 125: Ah nice! So, to encrypt 100 words we'd need under 1K cycles, right? I was way off with my guess :)

bobbinth:
Now that https://github.com/0xMiden/miden-vm/pull/2328 is merged, we can update this PR.

libcore/asm/crypto/aead.masm
Al-Kindi-0: A rare event :) !
Yeah, this is actually pessimistic but a good upper bound. The main work happens in the `while` block and there we can see that the cycles per field element encrypted are not more than 2

miden-vm/tests/integration/operations/crypto_ops.rs
Al-Kindi-0: I think this deserves a more thorough investigation, I created an issue for this

Al-Kindi-0:
> Now that #2328 is merged, we can update this PR.

This should be ready now.

bobbinth:
> > Now that #2328 is merged, we can update this PR.
> 
> This should be ready now.

The only thing left before we can merge is to release the next version of `miden-crypto`, right?

Al-Kindi-0:
Yes, once that is done, we can change the dependency in this PR and that should be it

huitseeker (approved):

PR #2323, opened 2025-10-30
refactor: hoist before_enter/after_exit decorator info to the MastForest (2/2)

huitseeker:
On top of #2310, and transitively, #2259, #2274, #2291, #2301, which make `MastNode`s immutable after insertion in a `MastForest`.

## Overview

The PR finishes a comprehensive forest-borrowing decorator setup for access to the decorators of Mast Nodes. This particular PR changes how node-level decorators (`before_enter`/`after_exit`) are stored & accessed, moving from node local storage to centralized forest-based storage.

- `NodeDecoratorStorage`: a CSR based storage system for node-level decorators, stored in the `MastForest`
- Single contiguous storage for all node decorators instead of per-node Vec allocations
- single `DecoratorStore` enum to handle both operation-level and node-level decorators
  - `Owned` variant: Local storage before nodes are added to forest
  - `Linked` variant: Reference to forest's centralized storage (i.e. the MastNodeId) after nodes are added to forest
- Linked nodes access decorators via forest reference, enabling shared decorator data

### API Changes
- `MastNodeExt::before_enter()` now requires `&MastForest` parameter
- `MastNodeExt::after_exit()` now requires `&MastForest` parameter
- Callers must update to pass forest reference

### Migration Guide
```rust
// Before
let decorators = node.before_enter();

// After
let decorators = node.before_enter(&forest);
```

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -389,6 +389,7 @@ impl BasicBlockNode {
     ///
     /// Unlike the derived PartialEq, this method works correctly with both owned and linked
     /// decorator storage by accessing the actual decorator data from the forest when needed.
+    #[cfg(test)]
     pub fn semantic_eq(&self, other: &BasicBlockNode, forest: &MastForest) -> bool {
         // Compare operations by collecting and comparing
         let self_ops: Vec<_> = self.operations().collect();
--- a/core/src/mast/node/join_node.rs
+++ b/core/src/mast/node/join_node.rs
@@ -155,6 +155,7 @@ impl JoinNode {
     ///
     /// Unlike the derived PartialEq, this method works correctly with both owned and linked
     /// decorator storage by accessing the actual decorator data from the forest when needed.
+    #[cfg(test)]
     pub fn semantic_eq(&self, other: &JoinNode, forest: &MastForest) -> bool {
         // Compare children
         if self.first() != other.first() || self.second() != other.second() {


CHANGELOG.md
```diff
@@ -36,7 +36,8 @@
 - [BREAKING] add pre/post decorators to BasicBlockNode fingerprint ([#2267](https://github.com/0xMiden/miden-vm/pull/2267)).
 - [BREAKING] Added explicit endianness methods `get_stack_word_be()` and `get_stack_word_le()` to stack word accessors, deprecated ambiguous `get_stack_word()` ([#2235](https://github.com/0xMiden/miden-vm/issues/2235)).
 - Fix decorator offset bug in `BasicBlockNode` padding ([#2305](https://github.com/0xMiden/miden-vm/pull/2305)).
-- [BREAKING] Update miden-crypto dependency to v0.18 (#[2311](https://github.com/0xMiden/miden-vm/pull/2311))
+- [BREAKING] Update miden-crypto dependency to v0.18 (#[2311](https://github.com/0xMiden/miden-vm/pull/2311)).
+- [BREAKING] hoist before_enter and after_exit decorators to MastForest ([#2323](https://github.com/0xMiden/miden-vm/pull/2323)).
```
plafer: Note: now that 0.19 is released, we should update this to put this under 0.20 (and similarly for the other PRs)

core/src/mast/mod.rs
```diff
@@ -76,6 +79,10 @@ pub struct MastForest {
     /// Provides efficient access to decorators per operation per node during execution and
     /// debugging.
     decorator_storage: DecoratorIndexMapping,
+
+    /// Storage for node-level decorators (before_enter and after_exit). This uses CSR format
+    /// for efficient storage and access of node-level decorators.
+    node_decorator_storage: NodeDecoratorStorage,
```
plafer: IIUC,
- `DecoratorIndexMapping`: only relevant for operations *within* basic blocks (i.e. no before_enter/after_exit),
- `NodeDecoratorStorage`: the `before_enter`/`after_exit` decorators for all node types.

Therefore I would rename the following:
- `decorator_storage` -> `op_decorator_storage`
- `DecoratorIndexMapping` -> `OpToDecoratorIdsMapping` or `OpToDecoratorIds`
- `NodeDecoratorStorage` -> `NodeToDecoratorIdsMapping` or `NodeToDecoratorIds`


Or any other naming schemes that make that relationship more explicit.

core/src/mast/node/decorator_store.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// which have not yet been inserted into a MAST forest.
+    Owned {
+        /// Operation-level decorators indexed by operation position
+        /// (Note: Only used by BasicBlockNode, other nodes will have empty decorators)
+        decorators: DecoratorList,
+        /// Node-level decorators executed before entering the node
+        before_enter: Vec<DecoratorId>,
+        /// Node-level decorators executed after exiting the node
+        after_exit: Vec<DecoratorId>,
+    },
```
plafer, line 26: In principle, we have the `*NodeBuilder` types so that any building-related concerns can be encapsulated there, without being surfaced in the "clean" `*Node` types. However the `DecoratorStore::Owned` variant here is never meant to appear on a `*Node` type once it's inserted in a `MastForest`, and so in principle shouldn't exist.

If we take `LoopNodeBuilder::add_to_forest()` as an example, we
- `build()` the `LoopNode` (using `DecoratorStorage::Owned`)
- Update the `MastForest.node_decorator_storage` with the decorator ids for the node
- Add a modified version of the `LoopNode` to the `MastForest`, using the `DecoratorStore::Linked` variant.

IIUC, we could get rid of the `DecoratorStore::Owned` variant if we
- Update the `MastForest.node_decorator_storage` with the decorator ids for the node, present in the `LoopNodeBuilder` state
- `build()` the `LoopNode`, using the `Linked` variant directly (but really `DecoratorStore` wouldn't even be an `enum` anymore)
- insert that node unmodified in the `MastForest`

I haven't looked at them all, but I think this pattern would work for all node types, and clean up `DecoratorStore`?

plafer (commented):
Looks good!

Left a suggestion that I think could simplify the `DecoratorStorage` (such that we no longer need the `Owned` variant).

core/src/mast/node/decorator_store.rs
huitseeker: We can indeed evict the usage of `DecoratorStore::Owned` from `add_to_forest`, at the cost of some code complexity (the `Owned` variant of the builder forms a convenient place where to figure e.g. the `digest`): 
https://github.com/0xMiden/miden-vm/compare/dc240024b...huitseeker:miden-vm:no-owned-in-addforest?expand=1 

However, that's not enough to "clean up `DecoratorStore`" because the `Owned` variant is still needed for:
 - Deserialization (nodes come back from disk with `Owned` decorators),
 - Round-trip conversions (node → builder → node) are essential for forest transformations (merge, rewrite)
 - Tests need to create nodes without forests

Refactoring the points above is a larger (i.e. future) endeavour.

CHANGELOG.md
huitseeker: Will do at the next rebase (updating all 6 PRs).

core/src/mast/mod.rs
huitseeker: Done in 7dc705bdd !

core/src/mast/node/decorator_store.rs
plafer: Makes sense - thank you for the explanation

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -0,0 +1,212 @@
...
+        // Ensure sentinel pointers exist for all nodes
+        let max_len = self.node_indptr_for_before.len().max(self.node_indptr_for_after.len());
+
+        // Add final sentinel pointers if needed
+        if self.node_indptr_for_before.len() == max_len {
+            let _ = self.node_indptr_for_before.push(self.before_enter_decorators.len());
+        }
+        if self.node_indptr_for_after.len() == max_len {
+            let _ = self.node_indptr_for_after.push(self.after_exit_decorators.len());
+        }
```
plafer, line 182: Similar to comment in previous PR, I don't think we should silently ignore the error here.

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -0,0 +1,212 @@
...
+        // as it would require shifting all subsequent data. This is a no-op.
+    }
+
+    /// Clears all decorators and mappings.
+    pub fn clear(&mut self) {
+        self.before_enter_decorators.clear();
+        self.after_exit_decorators.clear();
+        self.node_indptr_for_before = IndexVec::new();
+        self.node_indptr_for_after = IndexVec::new();
+    }
```
plafer, line 203: It's not immediately clear to me the difference between `clear()` and `remove_decorators()` - since "remove decorators" is a no-op, but "clear all decorators" isnt.

This might just be a documentation/naming issue.

plafer (approved):
LGTM!

As discussed offline, this doesn't fully close #1776 since `AssemblyOp`s are still part of the `Decorator` enum. Were you planning on working on this next?

I think this work though is related to #1978, so I would probably coordinate with @adr1anh to make sure you're not both making big changes to the same areas.

core/src/mast/mod.rs
```diff
@@ -75,7 +77,11 @@ pub struct MastForest {
 
     /// Provides efficient access to decorators per operation per node during execution and
     /// debugging.
-    decorator_storage: DecoratorIndexMapping,
+    op_decorator_storage: OpToDecoratorIds,
+
+    /// Storage for node-level decorators (before_enter and after_exit). This uses CSR format
+    /// for efficient storage and access of node-level decorators.
+    node_decorator_storage: NodeToDecoratorIds,
```
bobbinth, line 84: Not for this PR, but ideally, these 4 fields should be combined into a single `DebugInfo` struct. The main public interface of this struct could be something like:

```rust
impl DebugInfo {

    pub fn get_before_enter_decorators(
        &self,
        nodeId: MastNodeId,
    ) -> &impl Iterator<Item = &Decorator> {
        ...
    }

    pub fn get_after_exit_decorators(
        &self,
        nodeId: MastNodeId,
    ) -> &impl Iterator<Item = &Decorator> {
        ...
    }

    pub fn get_op_decorators(
        &self,
        node_id: MastNodeId,
        local_op_idx: usize,
    ) -> &impl Iterator<Item = &Decorator> {
        ...
    }

    pub fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>> {
        ...
    }
}
```

And within `MastForest`, this could be stored as `Option<DebugInfo>`.

processor/src/fast/basic_block.rs
```diff
@@ -37,7 +37,7 @@ impl FastProcessor {
         );
 
         // Execute decorators that should be executed before entering the node
-        self.execute_before_enter_decorators(node_id, program, host)?;
+        self.execute_before_enter_decorators(node_id, current_forest, host)?;
```
bobbinth, line 40: Not for this PR, but we should only execute decorators if we are running in debug mode. Using the something like the API from the pervious comment, this could look like:
```rust
if self.in_debug_mode {
    for decorator in current_forest.get_node_decorators(node_id) {
        self.execute_decorator(decorator, host)?;
    }
}
```

processor/src/fast/basic_block.rs
```diff
@@ -119,7 +119,7 @@ impl FastProcessor {
             self.execute_decorator(decorator, host)?;
         }
 
-        self.execute_after_exit_decorators(node_id, program, host)
+        self.execute_after_exit_decorators(node_id, current_forest, host)
```
bobbinth, line 122: Same comment as above.

processor/src/fast/basic_block.rs
```diff
@@ -119,7 +119,7 @@ impl FastProcessor {
             self.execute_decorator(decorator, host)?;
         }
```
bobbinth, line 120: Is this not redundant with the "after exist" decorators?

processor/src/fast/basic_block.rs
bobbinth, line 153: Similar to some of the previous comments: this code should run only if `self.in_debug_mode` is true.

Also, could we not just look up decorators by their operation index? Something like:
```rust
if self.in_debug_mode {
    for decorator in program.get_op_decorators(node_id, op_idx_in_block) {
        self.execute_decorator(decorator, host)?;
    }
}
```

core/src/mast/node/join_node.rs
```diff
@@ -24,10 +24,7 @@ use crate::{
 pub struct JoinNode {
     children: [MastNodeId; 2],
     digest: Word,
-    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Vec::is_empty"))]
-    before_enter: Vec<DecoratorId>,
-    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Vec::is_empty"))]
-    after_exit: Vec<DecoratorId>,
+    decorator_store: DecoratorStore,
 }
```
bobbinth, line 28: Not for this PR, but one of the goals of this refactoring was to remove any decorator-related data from MAST nodes. What would it take to get rid of `decorator_store` from her and from other nodes?

core/src/mast/node/join_node.rs
```diff
@@ -140,6 +145,41 @@ impl fmt::Display for JoinNodePrettyPrint<'_> {
...
+// ================================================================================================
+
+#[cfg(test)]
+impl JoinNode {
+    /// Checks if two JoinNodes are semantically equal (i.e., they represent the same join
+    /// operation).
+    ///
+    /// Unlike the derived PartialEq, this method works correctly with both owned and linked
+    /// decorator storage by accessing the actual decorator data from the forest when needed.
+    pub fn semantic_eq(&self, other: &JoinNode, forest: &MastForest) -> bool {
```
bobbinth, line 159: Not related to this PR, but now that we have this method, should we redefine `PartialEq` to basically just be `self.digest == other.digest`?

core/src/mast/node/join_node.rs
```diff
@@ -160,19 +200,18 @@ impl MastNodeExt for JoinNode {
...
+    fn before_enter<'a>(&'a self, forest: &'a MastForest) -> &'a [DecoratorId] {
+        self.decorator_store.before_enter(forest)
     }
 
     /// Returns the decorators to be executed after this node is executed.
-    fn after_exit(&self) -> &[DecoratorId] {
-        &self.after_exit
+    fn after_exit<'a>(&'a self, forest: &'a MastForest) -> &'a [DecoratorId] {
+        self.decorator_store.after_exit(forest)
     }
```
bobbinth, line 211: Not for this PR, but related to one of the previous comments: I'd love to get rid of these methods and from all decorator-related methods on MAST nodes. From processor standpoint that should be pretty straight-forward (we should be looking up decorators always but `MastNodeId` or `MastNodeId`+ operation index). Not sure if it causes much complexity elsewhere though.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -89,22 +83,9 @@ pub struct BasicBlockNode {
     /// Multiple batches are used for blocks consisting of more than 72 operations.
     op_batches: Vec<OpBatch>,
     digest: Word,
+    /// Stores both operation-level and node-level decorators
     /// Custom serialization is handled via Serialize/Deserialize impls
     decorators: DecoratorStore,
```
bobbinth, line 88: Not for this PR, but as mentioned in some of the past comments, I'd love to simplify this to be something like this:
```rust
pub struct BasicBlockNode {
    digest: Word,
    ops: Arc<[Operation]>,
}
```
Though, this is probably not something we can do before we get rid of the slow processor.

core/src/mast/node/dyn_node.rs
```diff
@@ -57,9 +54,17 @@ impl DynNode {
 impl MastNodeErrorContext for DynNode {
```
bobbinth, line 54: Do we actually need the `MastNodeErrorContext` trait? AFAICT, this is used primarily in `ErrorContextImpl` and I think we can pass just `MastNodeId` there to get a list of decorators related to a given operation. We would need to move the `get_assembly_op()` implementation into `MastForest` - but it shouldn't be too difficult. It could look something like:

```rust
impl MastForest {
    fn get_assembly_op(
        &self,
        node_id: MastNodeId,
        target_op_idx: Option<usize>,
    ) -> Option<AssemblyOp> {
        ...
    }
}
```

bobbinth (commented):
Looks good! Thank you! This is not a super in-depth review, but I left some comments inline. A lot of them are probably better suited for follow-up PRs, but I put them here to make sure we are aligned on the ultimate goal.

processor/src/fast/basic_block.rs
huitseeker: You're absolutely right - we should only execute decorators in debug mode. This is exactly what issue #1821 is about. The goal there is to remove the debug mode flag from the assembler and always assemble with debug info, but the processor should only execute decorators when in debug mode.

Your suggested API looks great:
```rust
if self.in_debug_mode {
    for decorator in current_forest.get_node_decorators(node_id) {
        self.execute_decorator(decorator, host)?;
    }
}
```

This work is tracked in #1821 "always include debug info when assembling" and we plan to implement the debug mode conditional execution as part of that issue. We'll make sure the fast processor only pays the decorator execution cost when debugging.

core/src/mast/node/dyn_node.rs
huitseeker: You're right - we should be able to simplify this! Passing just `MastNodeId` to `ErrorContextImpl` and looking up decorators in the forest would be cleaner.

The main reason we currently need the trait is to handle the different ways nodes store decorator information (Owned vs Linked). But with:
1. This PR completing the decorator hoisting to forest (done)
2. DebugInfo struct centralizing all debug data (#1776)
3. Potential removal of decorator_store from nodes (#1497)

The trait becomes unnecessary. Everything can be looked up via `MastForest` methods like:

```rust
impl MastForest {
    fn get_assembly_op(
        &self,
        node_id: MastNodeId,
        target_op_idx: Option<usize>,
    ) -> Option<AssemblyOp> {
        ...
    }
}
```

This simplification is another benefit of #1776. Once we have all debug info in a centralized location, the error context logic becomes straightforward forest lookups without needing a trait abstraction over different node types.

We'll remove the trait as part of #1776 when we implement the DebugInfo struct.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Some variant of that simplified structure is exactly where we want to end up — even if I'm not sure what the `Arc` intends to convey in this case.

This would eliminate some of the complexity of op_batches and decorator storage from the node itself. This is part of the broader vision in #1497, which proposes dynamic node objects with arena allocation.

The main blockers are:
1. The slow processor - it relies on the current BasicBlockNode structure
2. Need to complete the decorator hoisting (which this PR does)
3. Need finalized `DebugInfo` struct (#1776) to handle all debug data centrally

Once those are done, we can refactor BasicBlockNode to this simpler form as part of #1497. The operations would still be accessible via the forest context, and all debug info would come from `DebugInfo` rather than being stored in the node.

core/src/mast/node/join_node.rs
huitseeker: Removing `decorator_store` from the nodes is definitely the ultimate goal of this refactoring series. The main obstacle is that we still need the Owned variant for three key scenarios:

The path forward is likely:

core/src/mast/node/join_node.rs
huitseeker: In the current state of the PR, we have a synchronous moment where we transform all nodes to a `Linked` format, which makes their `DecoratorStore` carry just a `MastNodeId`: insertion in a `MastForest`. Nodes are therefore either not in a `MastForest`, and all `Owned`, or all in a `MastForest`, and all `Linked`, and because of this sharp boundary, any difference in representation is informative (and often indicative of a bug). Hence my approach of favoring a careful default `PartialEq` and a loose explicit comparison (the present function). 

I'll keep this in mind as we work through #1497 - once decorator_store is gone, we can revisit PartialEq to see if it makes sense to simplify it to digest comparison or just compare children + digest.

core/src/mast/node/join_node.rs
huitseeker: I agree - removing decorator-related methods from MAST nodes would make the API much cleaner. From the processor standpoint, lookups should always be by `MastNodeId` or `MastNodeId` + operation index.

The complexity elsewhere would mainly be in places which rely on simultaneously modifying decorators and nodes:
1. The `MastForest` builder patterns - would need to update decorators by node ID, which can be decomposed in merger and assembly,
2. Serialization/deserialization - would need to reconstruct the DebugInfo separately,
3. Testing utilities - would need forest context to look up decorators

All of these are manageable and align with the goals of #1776. We'll tackle this as part of that issue once the `DebugInfo` struct is in place.

core/src/mast/mod.rs
huitseeker: We'll implement this as part of #1776 after this PR lands, since it will be easier to refactor once we have both decorator types hoisted to the forest level (which this PR completes).

processor/src/fast/basic_block.rs
huitseeker: You're right - we should only execute decorators in debug mode. This is exactly what issue #1821 is about. The goal there is to remove the debug mode flag from the assembler and always assemble with debug info, but the processor should only execute decorators when in debug mode.

This work is tracked in #1821 "always include debug info when assembling" and we plan to implement the debug mode conditional execution as part of that issue. We'll make sure the fast processor only pays the decorator execution cost when debugging.

processor/src/fast/basic_block.rs
huitseeker: The files `processor/src/tests/decorator_execution_tests.rs` and `processor/src/fast/tests/fast_decorator_execution_tests.rs` have been made in #2167 to answer [this question](https://github.com/0xMiden/miden-vm/pull/2167#discussion_r2353671745). I think you're looking for `test_no_duplication_between_inner_and_before_exit_decorators_fast` in the later file.

core/src/mast/node/dyn_node.rs
huitseeker: You're right - we should be able to simplify this. Passing just `MastNodeId` to `ErrorContextImpl` and looking up decorators in the forest by operation index would be cleaner.

However, the `decorators` iterator is suprisingly enough used to find source information form more than the `BasicBlockNode`s, and `ErrorContextImpl` will fail to find appropriate source info (in `precalc_label_and_source_file`) if this is removed naively — as will be shown by our dozen of diagnostic tests.

processor/src/fast/basic_block.rs
huitseeker: Same answer, I think (?).

processor/src/fast/basic_block.rs
huitseeker: Agreeed, this is just separating concerns and maintaining the defensive iterator-style of `next_filtered`, to not have this PR wade too far into the processor. Will do once #1776 is implemented, because as of now, the `Owned` representation of decorators (in COO) has no ability to directly index into an operation's decorators.

core/src/mast/debuginfo/node_decorator_storage.rs
huitseeker: Addressed in 061acdb0900fa5d93d39d1a6e756a407fe0713c9!

core/src/mast/debuginfo/node_decorator_storage.rs
huitseeker: Good remark. The idea is that we want to strip *all* decorator in a `MastForest` recursively, a feature introduced in #2108, and currently served better by `clear()`. 

However, we had introduced `remove_decorators` in various traits and it found its way there to allow a `Linked` `MastNode` to have something sensible to call in this case. This later part doesn't really make sense any more.

huitseeker: Opened https://github.com/0xMiden/miden-vm/issues/2356

processor/src/fast/basic_block.rs
bobbinth: This may be for a separate PR, but we probably shouldn't have both "inner" decorators which get executed after the last operation and "after exit" decorators. This may require changes in the assembler (if we haven't done them already).

Basically, with basic blocks, we have two scenarios:

1. We either have "inner" decorators that get executed after the last operation. In this case, we don't need "before enter" and "after exist" decorators for basic blocks.
2. Or we have "after exit" decorators and then we don't need to have "inner" decorators after the last op.
  a. Technically, we don't need "before enter" decorators for basic blocks in this scenario - but maybe it is easier to keep them in.

core/src/mast/node/join_node.rs
bobbinth: Makes sense. I wonder if it would be possible to do this after steps 1 - 3, without doing all the work mentioned in 4. Specifically:

1. **Deserialization** - should be sorted out in step 3.
2. **Testing** - I'm not sure if we have a lot of tests that test decorators and require building `MastNodes` from scratch. If the number of such tests is small, we could consider rewriting them to use `MastForest`s.
3. **Round-trip conversion** - this one I'm not sure about - but maybe it is possible but maybe it would be possible to construct builders from node + MAST forest objects.

core/src/mast/node/join_node.rs
bobbinth: > I'll keep this in mind as we work through #1497 - once decorator_store is gone, we can revisit PartialEq to see if it makes sense to simplify it to digest comparison or just compare children + digest.

I'm not sure there is any reason to compare children+digest as digest is already a commitment to children. Basically, if digest are equal that means children are equal as well (unless the hash function is broken, but then we have bigger problems). So, we really only need 2 methods of comparison:

- Just digests - to check that the nodes have the same execution semantics.
- Digests + decorators - to check that nodes are identical (i.e., same execution semantics and same decorators).

processor/src/fast/basic_block.rs
huitseeker: > This may be for a separate PR, but we probably shouldn't have both "inner" decorators which get executed after the last operation and "after exit" decorators. This may require changes in the assembler

This does. We're moving towards scenario 2, but the assembler is not ready for it yet — the logic of *the entire block builder* relies on being able to push decorators at an index strictly posterior to that of the last operation it has so far. This is how every op-indexed decorator is added:
https://github.com/0xMiden/miden-vm/blob/17cd44ff7997c341916b6060ed85be436b8de167/crates/assembly/src/basic_block_builder.rs#L108-L114

<details><summary> Trivia </summary>
Technically, we will may still need `before_enter` decorators in this scenario, to e.g. track entry into a procedure by attaching a decorator to the first node of its body *irrespective of whether that's a basic block*.
</details>

core/src/mast/node/join_node.rs
huitseeker: Sorry, I misspoke when mentioned comparing children + digest, which may have led you astray here.

Digests, contrary to fingerprints[^1]  — or the function you are commenting on — do not include any information about decorators, do they? This function does need to discriminate base on these decorators, but I'm happy to make its purpose clear by making it test-only; addressed in 9dbf67f45df03372230b2eb42d0e4ffae90240cf

[^1]: post #2259, #2274, #2291, #2301, #2310, #2323, because prior to that series of PRs, the answer to "do fingerprints take decorators into account?" was "Sometimes."

bobbinth: > do not include any information about decorators, do they?

Yes, correct. My comment was more conceptual in nature. We have these two ways of comparing MAST nodes (with or without decorators) - the current `PartialEq` is kind of in-between. On the one hand, it does include comparison of decorators, on the other hand, this comparison is not always done correctly (hence the need for `semantic_eq()`).

huitseeker: Opened https://github.com/0xMiden/miden-vm/issues/2358

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Removes all decorators for a given node.
+    ///
+    /// # Arguments
+    /// * `_node_id` - The ID of the node to remove decorators for
+    ///
+    /// Note: This operation is not supported in the CSR structure as it would require
+    /// shifting all subsequent data. This method is a no-op.
+    pub fn remove_decorators(&mut self, _node_id: MastNodeId) {
+        // For CSR structure, removing individual node decorators is not supported
+        // as it would require shifting all subsequent data. This is a no-op.
```
Al-Kindi-0, line 194: Q: wouldn't we want to panic if this method is called? Otherwise, why not just remove it altogether?

huitseeker: Done in #2360

Issue #2293, opened 2025-10-15
precompiles: fix set of allowed precompiles


adr1anh:
The current precompile architecture provides more more flexibility than what we want in the final design (current temporary architecture documentation can be found [here](https://hackmd.io/@adr1anh/Byxn949aeg)). In particular, the VM and network will only support a fixed/predefined set of precompiles decided by the VM. The following points will need to be addressed to ensure our implementation, documentation and future deployment is clear, sound, and prevents any misuse.

**Remove dynamic precompile verifier registry**
Solution: create an enum with all supported precompiles which are always included in the verifier implementation. 

**Avoid `EventId` collisions**
Solution: use the discriminant of the precompile (with a version number allowing for upgrades) as the identifier.

**Error Handling for Network Enforcement**
When encountering a precompile whose ID does not match the current set of supported precompile, the verification fails indicating that the precompile is either unsupported or was generated with a deprecated prover version.

** Precompile VM Future Compatibility**
Extend the precompile verification trait (which could just become a `Precompile` trait), each precompile will include 
- a handler providing the non-deterministic result and logging the request
- a verifier to generate a commitment from the request
- a set of AirScript constraints
- a witness generator to populate the Precompile VM trace

The details here are still a bit murky, but a solution would likely incorporate the following components:
- The list of precompiles supported by the VM at its current version defines a unique verification key that is currently supported by the network.
- The verification key would be posted to the network, where the commitment includes the version matching the VM. The contract would have to check that this version is strictly larger that the previous one.
- The contract could expose committed procedure for verifying such precompile proof, essentially committing to a "blessed" verifier. This could perhaps become a kernel procedure.

----

This issue is related to #2282 and #2269 as the points mentioned would be addressed by centralizing the precompile architecture and documentation.

_Originally suggestion by @huitseeker in https://github.com/0xMiden/miden-vm/pull/2183#pullrequestreview-3339889583_

<details>
Thanks a lot for this PR, the current recision looks great! Your [documentation](https://hackmd.io/@adr1anh/Byxn949aeg) clarified a *lot* about the intended design behind this PR (and the following, e.g. #2249, #2277) — much appreciated. I think I can summarize the concerns that are still relevant this way:

1. Documentation / Implementation Mismatch

[...]

PR #2324, opened 2025-10-30
feat(precompiles): fix allowed set via enum, improve errors, keep compatibility

DeVikingMark:
## Describe your changes

- Add core::precompile_ids::SupportedPrecompile to define the fixed/preapproved precompile set.
- Map enum to canonical event IDs (keeps network compatibility with existing programs).
- Replace ambiguous verifier lookup error with UnsupportedPrecompile.
- Update Keccak precompile to use enum-based ID for tags and requests.
- Expose fixed precompile verifier registry from StdLibrary.
- Minor build fixes for prover call sites.

### Tests
- Core, stdlib (Keccak), verifier, prover: pass.
### Closes: #2293


adr1anh:
Closing, since the issue requires deeper investigation.

PR #2325, opened 2025-10-30
refactor(precompiles): simplify interfaces and tests

adr1anh:
The series streamlines how precompiles are handled end to end: the verifier registry now keeps both event names and implementations so errors surface clear context, the keccak precompile path is rebuilt on shared memory/packing utilities with the deprecated MASM dropped, and the new `miden_vm::verify` wrapper plus `exec.sys::log_precompile_request` helper smooth out verification for CLI users and MASM authors alike. Commit stack is linear—each change is reviewable on its own.

- Added `with_verifier`/`merge` API for `PrecompileVerifierRegistry`, storing `(EventName, Arc<…>)` for better diagnostics.
- CLI `verify` now calls `miden_vm::verify`, which wires in the stdlib registry before delegating to `verify_with_precompiles`.
- `log_precompile` processor op writes `[R1, R0, CAP_NEXT]` explicitly; MASM helper `exec.sys::log_precompile_request` wraps the opcode.
- Shared helper `read_memory_packed_u32` (with overflow-safe padding) powers the keccak handler and new MASM test utilities.
- Deprecated keccak MASM/tests/docs removed; keccak tests consolidated around the shared helpers.
- Documentation updated for hashes/sys modules; new sys integration test exercises deferred transcript verification path.

### DIFF
--- a/core/src/precompile.rs
+++ b/core/src/precompile.rs
@@ -24,8 +24,7 @@
 //!
 //! 3. **Proof Generation**: The prover generates a STARK proof of the VM execution. The final
 //!    [`PrecompileTranscript`] state (sponge capacity) is a public input. The verifier enforces the
-//!    initial (empty) and final state via variable‑length public inputs. TODO(#2045): These will be
-//!    enforced after updates to the recursive verifier.
+//!    initial (empty) and final state via variable‑length public inputs.
 //!
 //! 4. **Verification**: The verifier:
 //!    - Recomputes each precompile commitment using the stored requests via [`PrecompileVerifier`]


stdlib/asm/crypto/hashes/keccak256_deprecated.masm
plafer, line 1: I'm a bit confused how this PR is related to #2277, since this file is also removed in #2277. Should then we wait for #2277 to be merged, and you to rebase this PR before reviewing?

adr1anh: ECDSA is built on top of this one, so this one should go first

stdlib/asm/sys/mod.masm
```diff
@@ -33,3 +33,16 @@ end
 export.drop_stack_top
     dropw dropw dropw dropw
 end
+
+#! Logs a precompile commitment and removes the helper words produced by `log_precompile`.
+#!
+#! Input: `[COMM, TAG, ...]`
+#! Output: Stack without the three helper words `[R1, R0, CAP_NEXT]`.
+#! Cycles: 1 (plus cost of the underlying `log_precompile` instruction).
+export.log_precompile_request
```
bobbinth, line 42: The way to write inputs/outputs here would be:
```
Inputs:  [COMM, TAG]
Outputs: []
```
Also, the cycle count is not correct. By my count this procedure takes 19 cycles (including the 1 cycle for the `log_precompile` instruction).

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -245,15 +245,20 @@ pub(super) fn op_log_precompile<P: Processor>(
...
     processor.set_precompile_transcript_state(cap_next);
 
     // Write the output to the stack (top 12 elements): [R1, R0, CAP_NEXT, ...]
-    // The stack stores elements in reverse order relative to the permutation output.
-    for i in 0..STATE_WIDTH {
-        processor.stack().set(i, output_state[STATE_WIDTH - 1 - i]);
-    }
+    processor.stack().set_word(0, &r1);
+    processor.stack().set_word(4, &r0);
+    processor.stack().set_word(8, &cap_next);
```
bobbinth, line 261: Should we rename `set_word()` into `set_word_be()` to make things more explicit (the way we did in other cases)?

miden-vm/src/lib.rs
```diff
@@ -18,10 +18,30 @@ pub use miden_prover::{
...
 // (private) exports
 // ================================================================================================
 
 #[cfg(feature = "internal")]
 pub mod internal;
+
+/// Verifies a Miden proof.
+///
+/// See [miden_verifier::verify] for more details.
+pub fn verify(
```
bobbinth, line 32: I would mention in the comments here that this uses the verifier registry from the standard library to verify precompiles. This would make it clear that if the proof contains any "custom" precompiles, the proof will fail to verify because we don't have precompiles for them.

stdlib/src/lib.rs
```diff
@@ -77,10 +77,11 @@ impl StdLibrary {
         ]
     }
 
-    /// List of all `PrecompileVerifier` required to verify precompile requests.
-    pub fn verifiers(&self) -> Vec<(EventId, Arc<dyn PrecompileVerifier>)> {
-        let keccak_event_id = KECCAK_HASH_MEMORY_EVENT_NAME.to_event_id();
-        vec![(keccak_event_id, Arc::new(KeccakPrecompile))]
+    /// Returns a [`PrecompileVerifierRegistry`] containing all verifiers required to validate
+    /// standard library precompile requests.
+    pub fn verifier_registry(&self) -> PrecompileVerifierRegistry {
```
bobbinth, line 82: I do still wonder about the naming. For one, I think `precompile_verifiers()` would probably have been better as it makes it clear that this method is related to precompiles. But I'm also not sure about using "verifier" name in this context.

stdlib/src/handlers/mod.rs
```diff
@@ -15,3 +24,157 @@ fn u64_to_u32_elements(value: u64) -> (Felt, Felt) {
...
+/// // Read 5 bytes from address 0x100
+/// // Memory layout: addr[0x100] = 0x04030201, addr[0x104] = 0x00000005
+/// let bytes = read_memory_packed_u32(process, 0x100, 5)?;
+/// // Returns: [0x01, 0x02, 0x03, 0x04, 0x05]
+/// ```
+pub(crate) fn read_memory_packed_u32(
+    process: &ProcessState,
+    start: u64,
+    len_bytes: usize,
+) -> Result<Vec<u8>, MemoryReadError> {
```
bobbinth, line 67: It feels like maybe this (and some other functions here) could be methods on the `ProcessState` struct.

PR #2326, opened 2025-10-30
Updates the kernel ROM docs

plafer:
This PR

- Removes outdated comments
- Makes the docs concrete rather than general
    - i.e. assumes the current chiplet order rather than any order, which makes it easier to follow
- Explicitly multiplies $f_{krom}$ in so that we see the full degree
    - rather than relying on surrounding text to know which equation needs to be multiplied by it

### DIFF
--- a/docs/src/design/chiplets/kernel_rom.md
+++ b/docs/src/design/chiplets/kernel_rom.md
@@ -32,12 +32,15 @@ The meaning of columns in the above is as follows:
 
 ## Constraints
 
-> Note: the following assumes the ACE chiplet is included in the previous slot, whose documentation will be included
-> in a subsequent PR.
+We first define the selector flag $f_{krom}$ that is active in all rows of the kernel ROM chiplet.
+
+>$$
+> f_{krom} = s_0 \cdot s_1 \cdot s_2 \cdot s_3 \cdot (1 - s_4) \text{ | degree} = 5
+>$$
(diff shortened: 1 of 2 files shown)


docs/src/design/chiplets/kernel_rom.md
```diff
@@ -32,45 +32,48 @@ The meaning of columns in the above is as follows:
...
+We disable this constraint in the last row of the kernel ROM chiplet trace by using the kernel ROM chiplet selector $s_4'$, since the latter transitions from 0 to 1 in the first row of the next chiplet.
+
+For $i \in \{0,1,2,3\}$,
 
 > $$
-> (1 - s_4') \cdot (1 - s_{first}') \cdot (r_i' - r_i) = 0 \text{ | degree} = 3
+> f_{krom} \cdot (1 - s_4') \cdot (1 - s_{first}') \cdot (r_i' - r_i) = 0 \text{ | degree} = 8
 > $$
 
-_**Note**: we could technically remove the selector $(1-s_4')$ since $s_4$ and $s_{first}$ correspond to the same column. We include it here for completeness though._
```
plafer, line 72: $s_4$ is not the same as $s_{first}$ so I removed this comment - or maybe I'm missing something?

Al-Kindi-0 (approved):
LGTM!

PR #2327, opened 2025-10-31
fix: prevent integer underflow in OpBatchAccumulator::finalize_indptr

Patrici97:
Fixes a potential panic in `finalize_indptr()` where `uninit_group_idx` could underflow to `usize::MAX` when decrementing past zero.


huitseeker:
Thanks, the while gate makes this a non-issue, if you want us to look into this, please add a unit test demonstrating this is not the case.

PR #2328, opened 2025-10-31
Reduce the degree of HORNERBASE constraints

Al-Kindi-0:
## Describe your changes
Moves it from the very high degree bucket in order to vacate it for the op in #2322 

The general idea is to use two extra helper registers in order to reduce the degree of constraints. In addition, we switch from doing a memory word read per invocation of HORNERBASE to two element reads per invocation. This is works out as the degree of the op flag after this PR is 5 and, with 2 bus requests, the bus constraint equation becomes of degree 8.

- Updated `CHANGELOG.md'


docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,94 @@ The effect on the rest of the stack is:
...
-    &+ 8 \cdot \mathsf{acc}_1\cdot\alpha_0\cdot\alpha_1^3 + 2\cdot\mathsf{acc}_0\cdot\alpha_1^4  
-    &+ 6\cdot\mathsf{acc}_1\cdot\alpha_1^4 + s_7\cdot\alpha_0^3 - 6 \cdot s_7\cdot\alpha_0\cdot\alpha_1^2  
-    &- 2 \cdot s_7\cdot\alpha_1^3 + s_6\cdot\alpha_0^2 - 2\cdot s_6\cdot\alpha_1^2 + s_5\cdot\alpha_0 + s_4  \text{ | degree} = 5
+    \mathsf{tmp0}_0 &= s_1 + s_0 \cdot \alpha_0 + s_{15} \cdot (\alpha_0^2 - 2 \cdot \alpha_1^2)  \\
+    &\quad - 2 \cdot s_{14} \cdot (2 \cdot \alpha_0 \cdot \alpha_1 + \alpha_1^2)  \; \text{ | degree} = 3
 \end{align*}
 $$
 
 $$
 \begin{align*}
```
adr1anh: Should we combine in a single LaTeX block?

docs/src/design/stack/op_constraints.md
```diff
@@ -194,12 +194,13 @@ This group contains operations which require constraints with degree up to $3$.
 | `JOIN`        |     $87$     |   `101_0111`    | [Flow control ops](../decoder/index.md) |     $5$     |
 | `DYN`         |     $88$     |   `101_1000`    | [Flow control ops](../decoder/index.md) |     $5$     |
 | `HORNEREXT`   |     $89$     |   `101_1001`    |     [Crypto ops](./crypto_ops.md)      |     $5$     |
+| `HORNERBASE`  |     $90$     |   `101_1010`    |     [Crypto ops](./crypto_ops.md)      |     $5$     |
 | `LOGPRECOMPILE` |     $90$     |   `101_1010`    | [Crypto ops](./crypto_ops.md#log_precompile) |     $5$     |
```
adr1anh: Remove?

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -159,51 +159,64 @@ pub(super) fn op_horner_eval_base<P: Processor>(
...
+    let coef: [Felt; 8] = core::array::from_fn(|i| processor.stack().get(i));
+
+    let c0 = QuadFelt::from(coef[0]);
+    let c1 = QuadFelt::from(coef[1]);
+    let c2 = QuadFelt::from(coef[2]);
+    let c3 = QuadFelt::from(coef[3]);
+    let c4 = QuadFelt::from(coef[4]);
+    let c5 = QuadFelt::from(coef[5]);
+    let c6 = QuadFelt::from(coef[6]);
+    let c7 = QuadFelt::from(coef[7]);
```
adr1anh: Should the order be reversed?

docs/src/design/stack/crypto_ops.md
Al-Kindi-0: done

docs/src/design/stack/op_constraints.md
Al-Kindi-0: indeed, thanks for catching it

processor/src/processor/operations/crypto_ops.rs
Al-Kindi-0: good catch, fixed now

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,74 @@ The effect on the rest of the stack is:
...
-    $$\mathsf{tmp} = (((\mathsf{acc} \cdot \alpha + a_7) \cdot \alpha + a_6) \cdot \alpha + a_5) \cdot \alpha + a_4$$
+The `HORNERBASE` operation performs $8$ steps of the Horner method for evaluating a polynomial with coefficients over the base field at a point in the quadratic extension field. More precisely, it performs the following updates to the accumulator on the stack
+    $$\mathsf{tmp0} = ((\mathsf{acc} \cdot \alpha + c_0) \cdot \alpha) + c_1$$
 
-   $$\mathsf{acc}^{'} = (((\mathsf{tmp} \cdot \alpha + a_3) \cdot \alpha + a_2) \cdot \alpha + a_1) \cdot \alpha + a_0$$
-where $a_i$ are the coefficients of the polynomial, $\alpha$ the evaluation point, $\mathsf{acc}$ the current accumulator value, $\mathsf{acc}^{'}$ the updated accumulator value, and $\mathsf{tmp}$ is a helper variable used for constraint degree reduction.
+   $$\mathsf{tmp1} = ((((\mathsf{tmp0} \cdot \alpha) + c_2) \cdot \alpha + c_3) \cdot \alpha) + c_4$$
+
+   $$\mathsf{acc}^{'} = ((((\mathsf{tmp1} \cdot \alpha + c_5) \cdot \alpha + c_6) \cdot \alpha) + c_7)$$
```
adr1anh: Formatting nit: 

```suggestion
The `HORNERBASE` operation performs $8$ steps of the Horner method for evaluating a polynomial with coefficients over the base field at a point in the quadratic extension field. More precisely, it performs the following updates to the accumulator on the stack:

$$
\begin{align*}
\mathsf{tmp0}    &= ((\mathsf{acc} \cdot \alpha + c_0) \cdot \alpha) + c_1 \\
\mathsf{tmp1}    &= ((((\mathsf{tmp0} \cdot \alpha) + c_2) \cdot \alpha + c_3) \cdot \alpha) + c_4 \\
\mathsf{acc}^{'} &= ((((\mathsf{tmp1} \cdot \alpha + c_5) \cdot \alpha + c_6) \cdot \alpha) + c_7)
\end{align*}
$$
```

Al-Kindi-0: Done

processor/src/fast/tests/masm_consistency.rs
```diff
@@ -182,8 +182,8 @@ use super::*;
 )]
 // ---- horner ops --------------------------------
 #[case(None,
-    "begin
-        push.1.2.3.4 mem_storew_be.40 dropw
+    "begin 
+        push.1.2.0.0 mem_storew_be.40 dropw
```
plafer: Why was this change necessary?

processor/src/processor/operations/crypto_ops.rs
```diff
@@ -159,51 +159,66 @@ pub(super) fn op_horner_eval_base<P: Processor>(
...
     // Return the user operation helpers
-    Ok(P::HelperRegisters::op_horner_eval_registers(alpha, k0, k1, acc_tmp))
+    Ok([
+        alpha.to_base_elements()[0],
+        alpha.to_base_elements()[1],
+        tmp1.to_base_elements()[0],
+        tmp1.to_base_elements()[1],
+        tmp0.to_base_elements()[0],
+        tmp0.to_base_elements()[1],
+    ])
```
plafer: In principle, we want to use a `P::HelperRegisters` method, since with the `NoopHelperRegisters` implementation we want to be able to ignore all the arguments, and just return `[ZERO; NUM_USER_OP_HELPERS]` (used with the `FastProcessor` which just wants to go fast and doesn't care about helper register values). Though note that in this specific case, the computation to generate the registers is cheap and won't make a concrete performance difference - we should do it still for consistency with the rest of the operations.

So we should split `OperationHelperRegisters::op_horner_eval_registers()` into `op_horner_eval_base_registers()` and `op_horner_eval_ext_registers()`, and use `Ok(P::HelperRegisters::op_horner_eval_base_registers())` here.

docs/src/user_docs/assembly/cryptographic_operations.md
```diff
@@ -98,7 +98,7 @@ The following instructions are designed mainly for use in recursive verification
...
-| horner_eval_base <br /> - *(1 cycle)* | [c7,  c6,  c5,  c4,  c3,  c2,  c1,  c0, - , - , - , - , - , alpha_addr, acc1, acc0, ...]          | [c7,  c6,  c5,  c4,  c3,  c2,  c1,  c0, - , - , - , - , - , alpha_addr, acc1', acc0', ...]          | Performs 8 steps of the Horner evaluation method on a polynomial with coefficients over the base field, i.e., it computes `acc' = (((acc_tmp * alpha + c4) * alpha + c5) * alpha + c6) * alpha + c7` where `acc_tmp := (((acc * alpha + c0) * alpha + c1) * alpha + c2) * alpha + c3`                                          |
+| horner_eval_base <br /> - *(1 cycle)* | [c7,  c6,  c5,  c4,  c3,  c2,  c1,  c0, - , - , - , - , - , alpha_addr, acc1, acc0, ...]          | [c7,  c6,  c5,  c4,  c3,  c2,  c1,  c0, - , - , - , - , - , alpha_addr, acc1', acc0', ...]          | Performs 8 steps of the Horner evaluation method using 3-level computation: `tmp0 = ((acc * alpha + c0) * alpha) + c1`, `tmp1 = (((tmp0 * alpha + c2) * alpha + c3) * alpha) + c4`, `acc' = (((tmp1 * alpha + c5) * alpha + c6) * alpha) + c7`. Reads alpha from memory at `alpha_addr` and `alpha_addr + 1`.                                          |
```
plafer: I think `tmp0` and `tmp1` are implementation details right? We could document it just as a standard horner eval, similar to how [wikipedia writes it](https://en.wikipedia.org/wiki/Horner%27s_method) - i.e. `c7 + alpha * (c6 + alpha * (...))`.

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,76 @@ The effect on the rest of the stack is:
...
-
-   $$\mathsf{acc}^{'} = (((\mathsf{tmp} \cdot \alpha + a_3) \cdot \alpha + a_2) \cdot \alpha + a_1) \cdot \alpha + a_0$$
-where $a_i$ are the coefficients of the polynomial, $\alpha$ the evaluation point, $\mathsf{acc}$ the current accumulator value, $\mathsf{acc}^{'}$ the updated accumulator value, and $\mathsf{tmp}$ is a helper variable used for constraint degree reduction.
+The `HORNERBASE` operation performs $8$ steps of the Horner method for evaluating a polynomial with coefficients over the base field at a point in the quadratic extension field. More precisely, it performs the following updates to the accumulator on the stack:
+$$
+\begin{align*}
+\mathsf{tmp0}    &= ((\mathsf{acc} \cdot \alpha + c_0) \cdot \alpha) + c_1 \\
+\mathsf{tmp1}    &= ((((\mathsf{tmp0} \cdot \alpha) + c_2) \cdot \alpha + c_3) \cdot \alpha) + c_4 \\
+\mathsf{acc}^{'} &= ((((\mathsf{tmp1} \cdot \alpha + c_5) \cdot \alpha + c_6) \cdot \alpha) + c_7)
+\end{align*}
```
plafer, line 167: We should also update the chiplets bus equation. Though it currently doesn't even talk about the memory requests done by the horner operations - I guess we missed that when we introduced the operations?

plafer (changes requested):
Looks good! The idea makes sense to me.

Though we should update the chiplets bus equation, and there's a small change needed in the processor.

processor/src/fast/tests/masm_consistency.rs
Al-Kindi-0: Leftover from when I was making the op work with word-reads before switching to element-reads. Restored it now

processor/src/processor/operations/crypto_ops.rs
Al-Kindi-0: Makes sense!
Changed to your suggestion

docs/src/user_docs/assembly/cryptographic_operations.md
Al-Kindi-0: Good point, removed and updated

docs/src/design/stack/crypto_ops.md
Al-Kindi-0: Seems like we forgot it the first time
Added now, for both

Al-Kindi-0:
> LGTM!
> 
> Though let's just hold off until our meeting to be sure that this is the direction we want to go in.

I think this change is good regardless as we gain one high-degree op without paying much, though we can wait till the parallel PR is agreed on.

adr1anh (approved):
The changes look good to me too!

plafer:
> I think this change is good regardless as we gain one high-degree op without paying much, though we can wait till the parallel PR is agreed on.

I agree that this is good to merge regardless

core/src/operations/mod.rs
```diff
@@ -114,13 +114,13 @@ pub(super) mod opcode_constants {
     pub const OPCODE_JOIN: u8           = 0b0101_0111;
     pub const OPCODE_DYN: u8            = 0b0101_1000;
     pub const OPCODE_HORNEREXT: u8      = 0b0101_1001;
-    pub const OPCODE_LOGPRECOMPILE: u8  = 0b0101_1010;
+    pub const OPCODE_HORNERBASE: u8     = 0b0101_1010;
```
bobbinth: nit: I would maybe swap these places so that "base" operation comes before the "ext" operation. If we do this, we should update other parts (e.g., docs) to be the same).

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,82 @@ The effect on the rest of the stack is:
...
+Using the above values, we can describe the constraint for the memory bus column as follows:
 
 $$
-u_{mem} = \alpha_0 + \alpha_1 \cdot op_{mem\_readword} + \alpha_2 \cdot ctx + \alpha_3 \cdot s_{13} + \alpha_4 \cdot clk + \alpha_{5} \cdot h_{0} + \alpha_{6} \cdot h_{1} + \alpha_{7} \cdot h_{3} + \alpha_{8} \cdot h_{4}
+b_{mem}' = b_{mem} \cdot u_{mem,0} \cdot u_{mem,1} \text{ | degree} = 3
 $$
 
+The effect on the rest of the stack is:
+* **No change.**
 ## HORNEREXT
```
bobbinth, line 239: nit: let's add a newline above this line.

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,82 @@ The effect on the rest of the stack is:
 * **Left shift** starting from position $16$.
 
 ## HORNERBASE
-The `HORNERBASE` operation performs $8$ steps of the Horner method for evaluating a polynomial with coefficients over the base field at a point in the quadratic extension field. More precisely, it performs the following update to the accumulator on the stack
-    $$\mathsf{tmp} = (((\mathsf{acc} \cdot \alpha + a_7) \cdot \alpha + a_6) \cdot \alpha + a_5) \cdot \alpha + a_4$$
-
-   $$\mathsf{acc}^{'} = (((\mathsf{tmp} \cdot \alpha + a_3) \cdot \alpha + a_2) \cdot \alpha + a_1) \cdot \alpha + a_0$$
-where $a_i$ are the coefficients of the polynomial, $\alpha$ the evaluation point, $\mathsf{acc}$ the current accumulator value, $\mathsf{acc}^{'}$ the updated accumulator value, and $\mathsf{tmp}$ is a helper variable used for constraint degree reduction.
+The `HORNERBASE` operation performs $8$ steps of the Horner method for evaluating a polynomial with coefficients over the base field at a point in the quadratic extension field. More precisely, it performs the following updates to the accumulator on the stack:
+$$
```
bobbinth, line 162: nit: let's add a newline above this line (it'll fix rendering on Github).

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,82 @@ The effect on the rest of the stack is:
...
+\begin{align*}
+\mathsf{tmp0}    &= ((\mathsf{acc} \cdot \alpha + c_0) \cdot \alpha) + c_1 \\
+\mathsf{tmp1}    &= ((((\mathsf{tmp0} \cdot \alpha) + c_2) \cdot \alpha + c_3) \cdot \alpha) + c_4 \\
+\mathsf{acc}^{'} &= ((((\mathsf{tmp1} \cdot \alpha + c_5) \cdot \alpha + c_6) \cdot \alpha) + c_7)
+\end{align*}
+$$
```
bobbinth, line 168: nit: let's add a newline below this line (it'll fix rendering on Github).

docs/src/design/stack/crypto_ops.md
```diff
@@ -157,65 +157,82 @@ The effect on the rest of the stack is:
...
 - The next $5$ stack elements are irrelevant for the operation and unaffected by it.
-- The next stack element contains the value of the memory pointer `alpha_ptr` to the evaluation point $\alpha$. The word address containing $\alpha = (\alpha_0, \alpha_1)$ is expected to have layout $[\alpha_0, \alpha_1, k_0, k_1]$ where $[k_0, k_1]$ is the second half of the memory word containing $\alpha$. Note that, in the context of the above expressions, we only care about the first half i.e., $[\alpha_0, \alpha_1]$, but providing the second half of the word in order to be able to do a one word memory read is more optimal than doing two element memory reads.
+- The next stack element contains the memory address `alpha_ptr` pointing to the evaluation point $\alpha = (\alpha_0, \alpha_1)$. The operation reads $\alpha_0$ from `alpha_ptr` and $\alpha_1$ from `alpha_ptr + 1`.
 - The next $2$ stack elements contain the value of the current accumulator $\textsf{acc} = (\textsf{acc}_0, \textsf{acc}_1)$.
 
 The diagram below illustrates the stack transition for `HORNERBASE` operation.
 
 ![horner_eval_base](../../img/design/stack/crypto_ops/HORNERBASE.png)
```
bobbinth, line 180: I think the diagram needs to be updated - my understanding is that we no longer use $k_0$ and $k_1$ values because we read individual elements from memory (rather than the entire word).

A question related to this: with this setup we now perform 2 memory read operations which would double the number of rows in memory chiplet. This is probably fine, but I wonder if we could still use a single word read and just assume that values of $k_0$ and $k_1$ are zeros. Or does this complicate things in other places? (no need to change anything in this PR - just thoughts for potential future optimizations).

Al-Kindi-0: Yes, it needs updating but the source diagram is on my other machine, so I think I will just create an issue to fix this in a followup PR.

Regarding your question, this is a very good point, I initially started with that solution, namely using word-reads but restricting that such reads should only happen at words which are half-empty. I then switched to the current solution because the aforementioned restriction on read words seemed a bit awkward to work with (and probably required some changes to some MASM code.
Given the effect on the memory chiplet length, the first option (namely word-reads) is probably the most preferable.

Should we create an issue and fix this in a subsequent PR?

core/src/operations/mod.rs
Al-Kindi-0: Done

docs/src/design/stack/crypto_ops.md
bobbinth: > Yes, it needs updating but the source diagram is on my other machine, so I think I will just create an issue to fix this in a followup PR.

Sounds good!

> Regarding your question, this is a very good point, I initially started with that solution, namely using word-reads but restricting that such reads should only happen at words which are half-empty. I then switched to the current solution because the aforementioned restriction on read words seemed a bit awkward to work with (and probably required some changes to some MASM code. Given the effect on the memory chiplet length, the first option (namely word-reads) is probably the most preferable.
> 
> Should we create an issue and fix this in a subsequent PR?

Yes, let's create an issue, but this would be a fairly low priority for now.

PR #2330, opened 2025-10-31
refactor(assembly): item import/exports, symbol resolution, path handling

bitwalker:
This is, unfortunately, a mega PR that I haven't had time to properly stage in smaller changes, but nonetheless contains some important stuff for 0.19.

**NOTE:** This is ready for review, but should not be merged yet, until the following are checked off:

- [x] Rebase against latest changes in `next`, I'm sure that'll be fun
- [x] Fix whatever thing in the `serde_test` macro is barfing on the new `Arbitrary` implementation for paths (the old one just had three hardcoded paths, basically useless). I'm using an attribute-defined strategy that I'm not sure is being handled properly by `serde_test`, but the only remaining broken tests are the same error in all of the package-related serialization tests.
- [x] It can be punted to a point release, but I would like to finish up the const evaluation handling in the linker that kicks in when you have a constant expression involving an imported constant. Without it, using imported constants is not going to actually work properly. This is an easy thing to add now, but I just hadn't gotten to it yet.

A summary of the changes:

* Hard deprecates the `.`-separated keyword syntax for top-level items, e.g. `const.FOO` is now `const FOO`
* Exporting (and importing) constant and type declaration items are now supported, e.g. `pub const FOO = 1` will export `FOO`. Exported constants and types are included in the `Library` and `Package` metadata so they can be extracted for use.
* Exports are now exclusively handled with visibility modifiers, i.e. `pub` makes an item publicly visible, while no modifier uses the default visibility of private. There is no explicit private modifier. The modifier is prepended in front of the keyword to introduce an item, i.e. `pub proc`, `pub const`, etc.
* Imports and re-exported imports are now exclusively handled with `use path::to::item` and `pub use path::to::item` respectively. You can import items into scope directly as local symbols, e.g. `use path::to::CONSTANT` will bring `CONSTANT` into scope, and semantic analysis catches if there are symbol conflicts.
* Syntax for procedure locals is now handled using procedure annotations, i.e. `export foo.4` becomes `@locals(4)  pub proc foo` (note: the annotation does not have to be on the same line, I just did that here for brevity). The old syntax is hard deprecated.
* `LibraryNamespace` is removed, as it was no longer meaningful, and `LibraryPath` was completely rewritten into two new types `Path` and `PathBuf` (mirroring the libstd types), these are now much more ergonomic to use, support both owned and borrowed use cases, and has a simpler implementation. `PathBuf` is basically used for building up a path imperatively, while `Path` provides the bulk of the actual functionality, and is the primary way to pass them around. Path handling in general now is centered around treating paths as a hierarchical namespace, and makes them suitable for representing individual item symbols, module-qualified items, and fully-qualified absolute paths.
* The symbol resolution logic in the semantic analyzer and linker phases has been heavily reworked, and now shares a lot of the implementation details. Additional simplifications can be made, but I didn't have the time to go further down that road.
* The `Import` type in the AST has been removed, the `Alias` type supercedes it.
* The `Export` type now servers a broader purpose, and so has been moved into a new submodule of `miden-assembly-syntax`, along with other item-generic types and logic.
* I've removed many uses of the AST types `QualifiedProcedurePath` and `ProcedureName` from public types and APIs, as they are much less useful now after the path handling changes. They still serve a purpose in early validation, but can probably be removed at some point in the future. Most places now take `Arc<Path>` instead.
* I've simplified the `AliasTarget` and `InvocationTarget` types to take advantage of the new path handling and symbol resolution infra
* I've added some updated documentation that introduces types and type signatures. Further documentation is still needed, but it at least covers the basics.
* I've updated all the standard library MASM, and all of the test code where deprecated syntax was used
* I've added some more tests to cover basic path handling bits
* I've simplified some of the grammar rules a bit

There are a handful of trivial things added in the course of all this work that aren't really worth mentioning here, just things like missing functionality on `Span`, etc. If you see anything you have a question about, let me know.

### DIFF
--- a/crates/assembly-syntax/src/ast/ident.rs
+++ b/crates/assembly-syntax/src/ast/ident.rs
@@ -129,7 +129,14 @@ impl Ident {
 
     /// Returns true if this identifier is a valid constant identifier
     pub fn is_constant_ident(&self) -> bool {
-        self.name.chars().all(|c| c.is_ascii_uppercase() || c == '_')
+        self.name
+            .chars()
+            .all(|c| c.is_ascii_uppercase() || c.is_ascii_digit() || c == '_')
+    }
+
+    /// Returns true if this identifier must be quoted in Miden Assembly syntax
+    pub fn requires_quoting(&self) -> bool {
+        !self.name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '$')
     }
 
     /// Applies the default [Ident] validation rules to `source`.
@@ -258,8 +265,8 @@ impl Deserializable for Ident {
 }
 
 #[cfg(feature = "arbitrary")]
-pub(crate) mod testing {
-    use alloc::string::String;
+pub mod arbitrary {
+    use alloc::{borrow::Cow, string::String};
 
     use proptest::{char::CharStrategy, collection::vec, prelude::*};
 
@@ -302,6 +309,22 @@ pub(crate) mod testing {
     const PREFERRED_RANGES: &[core::ops::RangeInclusive<char>] = &['a'..='z', 'A'..='Z'];
     const EXTRA_RANGES: &[core::ops::RangeInclusive<char>] = &['0'..='9', 'à'..='ö', 'ø'..='ÿ'];
 
+    const PREFERRED_CONSTANT_RANGES: &[core::ops::RangeInclusive<char>] = &['A'..='Z'];
+    const EXTRA_CONSTANT_RANGES: &[core::ops::RangeInclusive<char>] = &['0'..='9'];
+
+    prop_compose! {
+        /// A strategy to produce a random character from a more restricted dictionary for bare
+        /// identifiers
+        fn bare_ident_chars()
+                      (c in CharStrategy::new_borrowed(
+                          &['_'],
+                          PREFERRED_RANGES,
+                          &['0'..='9']
+                      )) -> char {
+            c
+        }
+    }
+
     prop_compose! {
         /// A strategy to produce a random character from our valid dictionary, using the rules
         /// for selection provided by `CharStrategy`
@@ -316,14 +339,68 @@ pub(crate) mod testing {
     }
 
     prop_compose! {
+        /// Like `ident_chars`, but for constants
+        fn const_ident_chars()
+                      (c in CharStrategy::new_borrowed(
+                          &['_'],
+                          PREFERRED_CONSTANT_RANGES,
+                          EXTRA_CONSTANT_RANGES
+                      )) -> char {
+            c
+        }
+    }
+
+    prop_compose! {
         /// A strategy to produce a raw String of no more than length `length` bytes, containing any
         /// characters from our dictionary.
         ///
         /// The returned string will always be at least 1 characters.
         fn ident_raw_any(length: u32)
-                        (chars in vec(ident_chars(), 1..=(length as usize))) -> String {
+                        ((leading_char, rest) in (
+                            proptest::char::ranges(Cow::Borrowed(&['a'..='z', '_'..='_'])),
+                            vec(ident_chars(), 0..=(length as usize))
+                        )) -> String {
             let mut buf = String::with_capacity(length as usize);
-            for c in chars {
+            buf.push(leading_char);
+            for c in rest {
+                if !buf.is_empty() && buf.len() + c.len_utf8() > length as usize {
+                    break;
+                }
+                buf.push(c);
+            }
+            buf
+        }
+    }
+
+    prop_compose! {
+        /// Like `ident_raw_any`, but for bare identifiers
+        fn bare_ident_raw_any(length: u32)
+                        ((leading_char, rest) in (
+                            proptest::char::range('a', 'z'),
+                            vec(bare_ident_chars(), 0..=(length as usize))
+                        )) -> String {
+            let mut buf = String::with_capacity(length as usize);
+            buf.push(leading_char);
+            for c in rest {
+                if !buf.is_empty() && buf.len() + c.len_utf8() > length as usize {
+                    break;
+                }
+                buf.push(c);
+            }
+            buf
+        }
+    }
+
+    prop_compose! {
+        /// Like `ident_raw_any`, but for constants
+        fn const_ident_raw_any(length: u32)
+                        ((leading_char, rest) in (
+                            proptest::char::range('A', 'Z'),
+                            vec(const_ident_chars(), 0..=(length as usize))
+                        )) -> String {
+            let mut buf = String::with_capacity(length as usize);
+            buf.push(leading_char);
+            for c in rest {
                 if !buf.is_empty() && buf.len() + c.len_utf8() > length as usize {
                     break;
                 }
@@ -347,10 +424,81 @@ pub(crate) mod testing {
     }
 
     prop_compose! {
+        /// Generate a random bare identifier of `length` containing any characters from our
+        /// dictionary
+        pub fn bare_ident_any(length: u32)
+                    (raw in bare_ident_raw_any(length)
+                        .prop_filter(
+                            "identifiers must be valid",
+                            |s| Ident::validate(s).is_ok()
+                        )
+                    ) -> Ident {
+            Ident::from_raw_parts(Span::new(SourceSpan::UNKNOWN, raw.into_boxed_str().into()))
+        }
+    }
+
+    prop_compose! {
+        /// Generate a random constant identifier of `length` containing any characters from our
+        /// constant dictionary
+        pub fn const_ident_any(length: u32)
+                    (raw in const_ident_raw_any(length)
+                        .prop_filter(
+                            "identifiers must be valid",
+                            |s| Ident::validate(s).is_ok()
+                        )
+                    ) -> Ident {
+            let id = Ident::from_raw_parts(Span::new(SourceSpan::UNKNOWN, raw.into_boxed_str().into()));
+            assert!(id.is_constant_ident());
+            id
+        }
+    }
+
+    prop_compose! {
+        /// Generate a random type identifier corresponding to of one of the built-in types
+        pub fn builtin_type_any()
+                                (name in prop_oneof![
+                                    Just(crate::ast::types::Type::I1),
+                                    Just(crate::ast::types::Type::I8),
+                                    Just(crate::ast::types::Type::U8),
+                                    Just(crate::ast::types::Type::I16),
+                                    Just(crate::ast::types::Type::U16),
+                                    Just(crate::ast::types::Type::I32),
+                                    Just(crate::ast::types::Type::U32),
+                                    Just(crate::ast::types::Type::I64),
+                                    Just(crate::ast::types::Type::U64),
+                                    Just(crate::ast::types::Type::I128),
+                                    Just(crate::ast::types::Type::U128),
+                                    Just(crate::ast::types::Type::Felt),
+                                ]) -> Ident {
+            Ident::from_raw_parts(Span::new(SourceSpan::UNKNOWN, name.to_string().into_boxed_str().into()))
+        }
+    }
+
+    prop_compose! {
         /// Generate a random identifier of `length` containing any characters from our dictionary
         pub fn ident_any_random_length()
             (length in 1..u8::MAX)
             (id in ident_any(length as u32)) -> Ident {
+            id
+        }
+    }
+
+    prop_compose! {
+        /// Generate a random bare identifier of `length` containing any characters from our
+        /// dictionary
+        pub fn bare_ident_any_random_length()
+            (length in 1..u8::MAX)
+            (id in ident_any(length as u32)) -> Ident {
+            id
+        }
+    }
+
+    prop_compose! {
+        /// Generate a random constant identifier of `length` containing any characters from our
+        /// dictionary
+        pub fn const_ident_any_random_length()
+            (length in 1..u8::MAX)
+            (id in const_ident_any(length as u32)) -> Ident {
             id
         }
     }
--- a/crates/assembly-syntax/src/ast/path/path_buf.rs
+++ b/crates/assembly-syntax/src/ast/path/path_buf.rs
@@ -8,8 +8,6 @@ use core::{
 use miden_core::utils::{
     ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
 };
-#[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
 
 use super::{Path, PathError};
 use crate::ast::Ident;
@@ -19,7 +17,6 @@ use crate::ast::Ident;
 
 /// Path to an item in a library, i.e. module, procedure, constant or type.
 #[derive(Default, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(
     all(feature = "arbitrary", test),
     miden_test_serde_macros::serde_test(winter_serde(true))
@@ -313,41 +310,32 @@ impl Deserializable for PathBuf {
     }
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for PathBuf {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.inner.as_str())
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for PathBuf {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let inner = <&'de str as serde::Deserialize<'de>>::deserialize(deserializer)?;
+
+        PathBuf::new(inner).map_err(serde::de::Error::custom)
+    }
+}
+
 impl fmt::Display for PathBuf {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         fmt::Display::fmt(self.as_path(), f)
     }
-}
-
-// ARBITRARY IMPLEMENTATION
-// ================================================================================================
-
-#[cfg(any(test, feature = "arbitrary"))]
-impl proptest::prelude::Arbitrary for PathBuf {
-    type Parameters = ();
-
-    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
-        use proptest::prelude::*;
-
-        let wasm_cm_style =
-            Just(PathBuf::new("namespace-kebab:package-kebab/interface-kebab@1.0.0").unwrap());
-        let component_strategy = crate::ast::ident::testing::ident_any_random_length();
-        let components_strategy = proptest::collection::vec(component_strategy, 1usize..8usize)
-            .prop_map(|components| {
-                let mut buf = String::with_capacity(256);
-                for (i, component) in components.iter().enumerate() {
-                    if i > 0 {
-                        buf.push_str("::");
-                    }
-                    buf.push_str(component.as_str());
-                }
-                PathBuf { inner: buf }
-            });
-
-        prop_oneof![wasm_cm_style, components_strategy].boxed()
-    }
-
-    type Strategy = proptest::prelude::BoxedStrategy<Self>;
 }
 
 // TESTS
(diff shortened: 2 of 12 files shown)


crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+    Local(Span<ItemIndex>),
+    /// The name was resolved to an path referring to an item exported from another module
+    External(Span<Arc<Path>>),
+    /// The name was resolved to a procedure MAST root
+    MastRoot(Span<Word>),
+    /// The name was resolved to a known definition in the same module at the given index
+    Exact {
+        gid: GlobalItemIndex,
+        path: Span<Arc<Path>>,
+    },
```
plafer: Is the `path` pointing to the symbol? So the `gid` and `path` are equivalent representations of the symbol?

Also, the docstring is basically the same as `Local` - is this right?

crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+    fn from(module: &crate::library::ModuleInfo) -> Self {
+        Self { symbols: SymbolTable::from(module) }
+    }
+}
+
+struct SymbolTable {
+    symbols: BTreeMap<Arc<str>, ItemIndex>,
+    imports: BTreeMap<Arc<str>, Result<SymbolResolution, LocalSymbolResolutionError>>,
+    items: Vec<Symbol>,
+}
```
plafer: Maybe some docs? For example, it's not clear what the `Arc<str>` represent.

crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+
+#[derive(Debug)]
+enum Symbol {
+    Item { name: Ident, resolved: SymbolResolution },
+    Import(Span<Arc<str>>),
+}
```
plafer: Why do we need the `Import` variable? Isn't it equivalent to `SymbolResolution::External`, and hence could be encoded in the `Symbol::Item` variant? Might be worth documenting in a docstring

crates/assembly-syntax/src/ast/ident.rs
```diff
@@ -305,13 +316,20 @@ pub(crate) mod testing {
     }
 
     prop_compose! {
-        /// A strategy to produce a raw String of length `length`, containing any characers from
-        /// our dictionary.
+        /// A strategy to produce a raw String of no more than length `length` bytes, containing any
+        /// characers from our dictionary.
```
Al-Kindi-0: typo: Characters?

crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+};
+
+use super::{GlobalItemIndex, ModuleIndex};
+
+/// Represents the result of resolving a symbol
+#[derive(Debug, Clone)]
+pub enum SymbolResolution {
+    /// The name was resolved to a definition in the same module at the given index
+    Local(Span<ItemIndex>),
+    /// The name was resolved to an path referring to an item exported from another module
```
Al-Kindi-0: an -> a

crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+                            // not already
+                            if rest.is_empty() {
+                                Ok(SymbolResolution::External(
+                                    resolved.map(|p| p.to_absolute().into_owned().into()),
+                                ))
+                            } else {
+                                Ok(SymbolResolution::External(
+                                    resolved.map(|p| p.to_absolute().join(rest).into()),
+                                ))
+                            }
```
Al-Kindi-0: Q: could these two branches be collapsed?

crates/assembly-syntax/src/ast/item/resolver.rs
```diff
@@ -0,0 +1,390 @@
...
+                    Err(err) => {
+                        log::error!(target: "symbol-table", "resolution of '{name}' failed: {err}");
+                        Err(err)
+                    },
+                }
+            },
+        }
+    }
+
+    fn expand(
```
Al-Kindi-0, line 264: nit: maybe some further comments would help following the case-analysis logic in this function

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -347,25 +345,25 @@ impl Module {
 
 /// Metadata
 impl Module {
-    /// Get the name of this specific module, i.e. the last component of the [LibraryPath] that
+    /// Get the name of this specific module, i.e. the last component of the [Path] that
     /// represents the fully-qualified name of the module, e.g. `u64` in `std::math::u64`
     pub fn name(&self) -> &str {
-        self.path.last()
+        self.path.last().unwrap()
```
Al-Kindi-0: nit: would it make sense, effort wise, to replace the unwrap with an expect containing a message explaining why this is safe/acceptable?

Al-Kindi-0: I am thinking both here and in several other places involving `self.path`

crates/assembly-syntax/src/ast/path/components.rs
```diff
@@ -0,0 +1,558 @@
...
+                Some(Err(PathError::InvalidComponent(crate::ast::IdentError::InvalidLength {
+                    max: Path::MAX_COMPONENT_LENGTH,
+                })))
+            },
+            next => next,
+        }
+    }
+}
+
+struct Components<'a> {
```
Al-Kindi-0, line 142: nit: maybe a comment explaining the purpose of this struct. Similar comment about all other internal/helper structs

crates/assembly-syntax/src/ast/path/components.rs
```diff
@@ -0,0 +1,558 @@
...
+                },
+                State::Done => break,
+            }
+        }
+        None
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Al-Kindi-0, line 389: Should we also exercise the error cases, for example
  - Empty component ("foo::" or "foo::::bar")
  - Missing separator after a quoted component ("foo""bar")
  - Un-terminated quoted component ("foo::"bar)
  - Quoted component with empty content (foo::""::bar)
  - Component exceeding Path::MAX_COMPONENT_LENGTH.
  - Others?

crates/assembly-syntax/src/ast/path/path.rs
```diff
@@ -0,0 +1,438 @@
...
+    borrow::{Borrow, Cow, ToOwned},
+    string::{String, ToString},
+};
+use core::fmt;
+
+use crate::ast::Ident;
+
+use super::{Iter, PathBuf, PathComponent, PathError, StartsWith};
+
+/// A borrowed reference to a subset of a [Path]
```
Al-Kindi-0: nit: self-reference is a bit confusion here

crates/assembly-syntax/src/ast/path/path_buf.rs
```diff
@@ -0,0 +1,463 @@
...
+        let inner = self.inner.into_boxed_str();
+        let inner = alloc::boxed::Box::into_raw(inner);
+        // SAFETY: This cast is safe because *mut Path is equivalent to *mut str
+        unsafe { alloc::boxed::Box::from_raw(inner as *mut Path) }
+    }
+}
+
+/// Mutation
+impl PathBuf {
+    /// Override the parent prefix of this path.
```
Al-Kindi-0: overrides

crates/assembly-syntax/src/ast/path/path_buf.rs
```diff
@@ -0,0 +1,463 @@
...
+    pub fn into_boxed_path(self) -> alloc::boxed::Box<Path> {
+        let inner = self.inner.into_boxed_str();
+        let inner = alloc::boxed::Box::into_raw(inner);
+        // SAFETY: This cast is safe because *mut Path is equivalent to *mut str
+        unsafe { alloc::boxed::Box::from_raw(inner as *mut Path) }
+    }
+}
+
+/// Mutation
+impl PathBuf {
```
Al-Kindi-0, line 134: should we add tests for `push`, `pop`, `set_parent` and `set_namespace`?

crates/assembly-syntax/src/ast/type.rs
```diff
@@ -1,10 +1,23 @@
...
 pub use midenc_hir_type as types;
 use midenc_hir_type::{AddressSpace, Type, TypeRepr};
 
-use super::{ConstantExpr, DocString, Ident};
+use super::{
+    ConstantExpr, DocString, GlobalItemIndex, Ident, ItemIndex, LocalSymbolResolutionError, Path,
+    PathBuf, SymbolResolution, Visibility,
+};
+
+pub trait TypeResolver<E> {
```
Al-Kindi-0, line 21: maybe a bit of documentation, could also be at the module level

crates/assembly-syntax/src/ast/visibility.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Represents the visibility of an item (procedure, constant, etc.) globally.
+#[derive(Default, Debug, Copy, Clone, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum Visibility {
+    /// The item is visible outside its defining module
+    Public = 0,
+    /// The item is visible only within its defining module
+    #[default]
+    Private = 2,
+}
```
Al-Kindi-0, line 15: Q: why hope over 1?

crates/assembly-syntax/src/parse.rs
```diff
@@ -49,11 +49,11 @@ impl ParseOptions {
...
     ///
     /// This is primarily useful when compiling a module from source code that has no meaningful
-    /// [LibraryPath] associated with it, such as when compiling from a `str`. This will override
+    /// [Path] associated with it, such as when compiling from a `str`. This will override
     /// the default name derived from the given [ModuleKind].
     pub fn new<P, E>(kind: ModuleKind, path: P) -> Result<Self, E>
     where
-        P: TryInto<LibraryPath, Error = E>,
+        P: TryInto<PathBuf, Error = E>,
     {
```
Al-Kindi-0: I wonder if we can switch the names `Path` and `PathBuf`? Maybe rename `PathBuf` to `Path` and `Path` to `PathRef` or something else?

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -241,11 +245,110 @@ begin
...
+const DEBUG = 1
+const WARN = DEBUG + 1
+const ERROR = WARN + 1
+
+pub proc set_level(level: Level)
+    ...
+end
+```
+
+Note that nothing currently validates that a u8 value passed as an instance of `Level` is actually one of those three constants, that is up to you. Instead, the `enum` syntax is intended to provide more semantic information for readers of the code, and to better express the relationship between the type and the constants in question.
```
Al-Kindi-0: u8 -> `u8`

stdlib/docs/crypto/stark/mod.md
```diff
@@ -2,4 +2,3 @@
...
... AIR<br />3. `trace_info` which is a field element encoding the layout of the AIR<br />4. `num_fixed_len_pi` is the number of fixed length public inputs of the AIR<br /><br />In addition to the above parameters, the verifier expects the following auxiliary proof parameters:<br /><br />1. `log(trace_length)`, the logarithm base 2 of the trace length<br />2. `num_queries`, the number of FRI queries<br />3. `grinding`, the number of bits of grinding i.e., proof-of-work<br /><br />The following simplifying assumptions are currently made and hardcoded:<br /><br />- The blowup factor is set to 8.<br />- The maximal allowed degree of the remainder polynomial is 127.<br />- To boost soundness, the protocol is run on a quadratic extension field and this means that<br />the OOD evaluation frame is composed of elements in a quadratic extension field i.e. tuples.<br />Similarly, elements of the auxiliary trace are quadratic extension field elements. The random<br />values for computing random linear combinations are also in this extension field.<br /><br />Inputs:  [D3, D2, D1, D0, log(trace_length), num_queries, grinding, num_constraints, trace_info, num_fixed_len_pi]<br />Outputs: []<br /> |
```
Al-Kindi-0, line 5: Q: Is this intentional?

crates/assembly-syntax/src/ast/item/resolver.rs
bitwalker: I made some adjustments to `join` to make it safe to `.join(rest)` even if `rest.is_empty()` is true, so now these can be collapsed.

crates/assembly-syntax/src/ast/path/components.rs
bitwalker: Definitely a good idea

crates/assembly-syntax/src/ast/path/path_buf.rs
bitwalker: We certainly can, though they are exercised by other tests currently, but more localized tests is definitely preferable

crates/assembly-syntax/src/ast/visibility.rs
bitwalker: Whoops, there used to be a third kind, but it was removed and these discriminants weren't updated

crates/assembly-syntax/src/parse.rs
bitwalker: I've reworked the API here to use `Path` - I think it makes sense to stay consistent with the libstd types here

stdlib/asm/crypto/stark/constants.masm
```diff
@@ -176,284 +176,284 @@ const.ACE_CIRCUIT_PTR=3225420176            # AUXILIARY_ACE_INPUTS_PTR + 12
...
 
-export.aux_trace_com_ptr
+pub proc aux_trace_com_ptr
     push.AUX_TRACE_COM_PTR
 end
 
-export.composition_poly_com_ptr
+pub proc composition_poly_com_ptr
     push.COMPOSITION_POLY_COM_PTR
 end
```
bobbinth, line 329: @Al-Kindi-0 - after this PR is merged, we could probably make a pass through the STARK verifier code to import constants directly.

miden-vm/src/cli/bundle.rs
```diff
@@ -77,7 +77,7 @@ impl BundleCmd {
                     Some(namespace) => namespace.to_string(),
                     None => dir.to_string_lossy().into_owned(),
                 };
-                let library_namespace = namespace.parse::<LibraryNamespace>()?;
+                let library_namespace = LibraryPath::new(&namespace).into_diagnostic()?;
```
bobbinth, line 80: nit: we should probably rename the variable name here into `library_path`.

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -105,62 +112,55 @@ end
...
+If the assembler cannot resolve external symbol references to a known module or item, assembly will fail. You can register modules with the assembler when instantiating it, either in source form, or precompiled form. See the [miden-assembly docs](https://crates.io/crates/miden-assembly) for details. The assembler will use this information to resolve references to imported items during assembly.
 
-If the assembler does not know about the imported modules, assembly will fail. You can register modules with the assembler when instantiating it, either in source form, or precompiled form. See the [miden-assembly docs](https://crates.io/crates/miden-assembly) for details. The assembler will use this information to resolve references to imported procedures during assembly.
-
-#### Re-exporting procedures
-A procedure defined in one module can be re-exported from a different module under the same or a different name. For example:
+#### Re-exporting items
```
bobbinth, line 160: Should we explicitly mention somewhere in this section that we cannot attach attributes to re-exports?

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -170,46 +170,46 @@ In the module shown above, not only is the locally-defined procedure `foo` expor
...
-export.::std::math::u64::mul->mul64
+pub use ::std::math::u64::mul->mul64
 ```
 
-Additionally, you may re-export a procedure using its MAST root, so long as you specify an alias:
+Additionally, you may re-export a procedure using its MAST root, so long as you specify a name for it, as shown below:
 
 ```
-export.0x0000..0000->mul64
+pub use 0x0000..0000->mul64
```
bobbinth, line 184: Question: how does this get resolved? I'm assuming that the compilation will fail if the assembler cannot find a procedure with such root during compilation - right? Either way, probably worth briefly explaining this.

docs/src/user_docs/assembly/code_organization.md
bobbinth, line 238: Is this syntax still supported? I thought we couldn't mix word and non-word constants in the same instruction.

crates/mast-package/src/package/manifest.rs
```diff
@@ -55,7 +68,7 @@ impl PackageManifest {
 
     /// Get the number of procedures exported from this package
     pub fn num_exports(&self) -> usize {
-        self.exports.len()
+        self.exports.values().filter(|export| export.is_procedure()).count()
     }
```
bobbinth, line 70: Should we rename this to something like `num_proc_exports()` to indicate that this returns only the number of procedure exports?

crates/mast-package/src/package/manifest.rs
```diff
@@ -74,19 +87,119 @@ impl PackageManifest {
         digest: &Word,
     ) -> impl Iterator<Item = &PackageExport> + '_ {
```
bobbinth: Similar to the above comment - should this be `get_proc_export_by_digest()`? And I'd probably update the comment to say "all procedure exports".

Also, since this is only for procedures, the return type could probably be `impl Iterator<Item = &ProcedureExport>`.

crates/mast-package/src/package/manifest.rs
```diff
@@ -17,15 +24,21 @@ use crate::Dependency;
 #[cfg_attr(feature = "arbitrary", derive(proptest_derive::Arbitrary))]
 pub struct PackageManifest {
```
bobbinth, line 23: Let's update the doc comments here to mention that we export other things besides procedures from packages too.

crates/assembly/src/tests.rs
bobbinth, line 1: Not for this PR, but I'd love to split this file up into smaller files each containing a logically-related group of tests.

bobbinth (commented):
Looks great! Thank you! I didn't really review changes in the `assembly-syntax` or `assembly/linker` directories in depth, but did go through everything else and left some small comments inline.

I think the next step on this PR is to rebase it form the latest `next`. Once this is done, we should be able to merge - but I also understand that there is ongoing work on refactoring the linker/symbol resolution a bit.

stdlib/asm/crypto/stark/constants.masm
Al-Kindi-0: Agreed!
Created #2355

miden-vm/src/cli/bundle.rs
bitwalker: It is supposed to represent the namespace of the assembled library though, so I think it might be best to keep it either as `library_namespace` or just `namespace`. The parameter name it is passed to is named `namespace` for reference.

docs/src/user_docs/assembly/code_organization.md
bitwalker: It won't be rejected, since it is a MAST root, instead, any use of the `mul64` symbol will be treated as if the aliased MAST root had been used directly, and the assembler presumes that any explicit MAST root reference like this will be resolved at runtime. I can note this in the docs though.

docs/src/user_docs/assembly/code_organization.md
bitwalker: The only thing not permitted is mixing the word literal syntax with other forms. You can freely mix a constant reference (that refers to a word value) and integer literals though, because the assembler expands that into multiple pushes, one for each `.`-delimited item, and we don't know until semantic analysis (at the earliest) whether a constant reference is an integer, word, or something else.

crates/mast-package/src/package/manifest.rs
bitwalker: We could just name it `num_procedures`, but I think it makes more sense to change the behavior here to cover all exports, since that is how it is used in practice.

crates/mast-package/src/package/manifest.rs
bitwalker: I'd change it to `get_procedures_by_digest`, but agreed

stdlib/docs/crypto/stark/mod.md
bitwalker: Good catch, the way the stdlib `build.rs` was gathering procedure docs hadn't been updated, so all re-exported docs got dropped. I've fixed that and restored all the removed docs.

bitwalker:
Only thing left before this can be merged is fixing whatever is broken in the `Arbitrary` impls for package/library exports causing the generated serialization tests to fail. I'll take a look at that first thing tomorrow morning.

PR #2334, opened 2025-11-03
test: add missing instances of Arbitrary for MastNodes

### LINKED ISSUE #2243: test: improve Arbitrary for MastForest to include other block types

huitseeker:
A clean fix of #2243 that does not depend on builder infra (#2259, #2274, #2291, #2301).

FIxes #2243 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 0.20.0 (TBD)
 
+#### Enhancements
+
+#### Changes
+- Added missing implementations of `proptest::Arbitrary` for non-`BasicBlockNode` variants of `MastNode` ([#2334](https://github.com/0xMiden/miden-vm/pull/2334)).
+
 ## 0.19.0 (2025-11-1)
 
 #### Enhancements


core/src/mast/node/split_node.rs
```diff
@@ -251,3 +252,30 @@ impl MastNodeExt for SplitNode {
...
+        use proptest::prelude::*;
+
+        use crate::Felt;
+
+        // Generate two MastNodeId values and digest for the branches
+        (any::<MastNodeId>(), any::<MastNodeId>(), any::<[u64; 4]>())
+            .prop_map(|(true_branch, false_branch, digest_array)| {
+                // Use new_unsafe since we're generating arbitrary nodes
+                // The digest is also arbitrary since we can't compute it without a MastForest
+                let digest = Word::new([Felt::new(digest_array[0]), Felt::new(digest_array[1]), Felt::new(digest_array[2]), Felt::new(digest_array[3])]);
```
adr1anh: A bit more concise

```suggestion
                let digest = Word::from(digest_array.map(Felt::new));
```

PR #2335, opened 2025-11-03
test: add missing instances of Arbitrary for MastNodes

### LINKED ISSUE #2243: test: improve Arbitrary for MastForest to include other block types

huitseeker:
A fix of #2243 that does not depend on builder infra (#2259, #2274, #2291, #2301).

FIxes #2243 .

### DIFF
--- a/core/src/mast/node/basic_block_node/arbitrary.rs
+++ b/core/src/mast/node/basic_block_node/arbitrary.rs
@@ -182,6 +182,28 @@ impl Arbitrary for BasicBlockNode {
 // ---------- Optional: MastForest strategy (behind feature gate) ----------
 
 /// Parameters for generating MastForest instances
+///
+/// # Execution Compatibility
+///
+/// Generated forests will only be executable if certain node types are excluded:
+/// - **Syscalls**: Require matching kernel procedures. Set `max_syscalls = 0` for executable
+///   forests.
+/// - **Externals**: Use random digests that won't match valid procedures. Set `max_externals = 0`
+///   for executable forests.
+/// - **Dyn nodes**: Leave junk on the stack and cannot execute properly. Set `max_dyns = 0` for
+///   executable forests.
+///
+/// The default parameters create executable forests by setting all three of the above to 0.
+///
+/// # Non-executable Forests
+///
+/// If you want to generate forests for testing assembly or parsing without execution:
+/// - Set `max_syscalls > 0` to include syscall nodes
+/// - Set `max_externals > 0` to include external nodes
+/// - Set `max_dyns > 0` to include dynamic nodes
+///
+/// These forests are useful for testing MAST structure and serialization but will fail during
+/// execution.
 #[derive(Clone, Debug)]
 pub struct MastForestParams {
     /// Number of decorators to generate
@@ -196,9 +218,22 @@ pub struct MastForestParams {
     pub max_loops: usize,
     /// Maximum number of call nodes to generate
     pub max_calls: usize,
+    /// Maximum number of syscall nodes to generate
+    ///
+    /// **Warning**: Syscalls require a properly configured kernel with matching procedure hashes.
+    /// Generated syscalls use random procedure digests and will not execute without providing
+    /// a matching kernel. Set to 0 for executable forests.
+    pub max_syscalls: usize,
     /// Maximum number of external nodes to generate
+    ///
+    /// **Warning**: External nodes use random digests that won't correspond to valid procedures.
+    /// Any program with external nodes will fail to execute. Set to 0 for executable forests.
     pub max_externals: usize,
     /// Maximum number of dyn/dyncall nodes to generate
+    ///
+    /// **Warning**: Dyn nodes leave junk on the stack and cannot execute properly.
+    /// These nodes are primarily for testing MAST structure, not execution. Set to 0 for
+    /// executable forests.
     pub max_dyns: usize,
 }
 
@@ -211,8 +246,9 @@ impl Default for MastForestParams {
             max_splits: 1,
             max_loops: 1,
             max_calls: 1,
-            max_externals: 1,
-            max_dyns: 1,
+            max_syscalls: 0,  // Default to 0 for executable forests
+            max_externals: 0, // Default to 0 for executable forests
+            max_dyns: 0,      // Default to 0 for executable forests
         }
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/mast/node/call_node.rs
```diff
@@ -324,3 +325,32 @@ impl MastNodeExt for CallNode {
...
+        use proptest::prelude::*;
+
+        use crate::Felt;
+
+        // Generate callee, digest, and whether it's a syscall
+        (any::<MastNodeId>(), any::<[u64; 4]>(), any::<bool>())
+            .prop_map(|(callee, digest_array, is_syscall)| {
+                // Use new_unsafe since we're generating arbitrary nodes
+                // The digest is also arbitrary since we can't compute it without a MastForest
+                let digest = Word::new([Felt::new(digest_array[0]), Felt::new(digest_array[1]), Felt::new(digest_array[2]), Felt::new(digest_array[3])]);
```
adr1anh: Nit: a bit more concise

```suggestion
                let digest = Word::from(digest_array.map(Felt::new);
```

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +207,189 @@ impl Arbitrary for MastForest {
...
+                        Just((
+                            num_joins,
+                            num_splits,
+                            num_loops,
+                            num_calls,
+                            num_externals,
+                            num_dyns,
+                        )),
+                        // Generate indices for join nodes (need 2 children each)
+                        prop::collection::vec(any::<(usize, usize)>(), num_joins..=num_joins)
```
adr1anh, line 380: `num_joins..=num_joins` is always `num_joins` right?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +207,189 @@ impl Arbitrary for MastForest {
...
+                            } else {
+                                forest.add_call(callee_id).unwrap()
+                            };
+                            all_node_ids.push(call_id);
+                        }
+                    }
+
+                    // Add external nodes
+                    for _ in 0..num_externals {
+                        if let Ok(external_id) = forest.add_external(Word::default()) {
```
adr1anh: Do we want to make the digest arbitrary as well?

adr1anh (approved):
LGTM, just left minor nits

core/src/mast/node/basic_block_node/arbitrary.rs
huitseeker: Since we were generating exactly `num_joins` items, the range `num_joins..=num_joins` was redundant. After my recent changes that made the `max_*` fields explicit, this pattern appears because we're clamping values (e.g., `num_joins = num_joins.min(max_parent_nodes)`), and then generating exactly that many items.

core/src/mast/node/basic_block_node/arbitrary.rs
huitseeker: Good question. The external node digest uses `Word::default()` which is sufficient for now since we mainly care about testing the structure and serialization, not the actual digest values. Making it fully arbitrary would help with assembly testing, though. Also, `Word` doesn't have an `Arbitrary` implementation and lives in a separate crate, so we'd need to generate it from `[u64; 4]` anyway.

huitseeker: I've added randomness for digest as part of the last commit!

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +228,218 @@ impl Arbitrary for MastForest {
...
+                        let node_id = forest.add_node(block).expect("Failed to add basic block");
+                        basic_block_ids.push(node_id);
+                    }
+
+                    // 3) Add control flow nodes in topological order (children already exist)
+                    let mut all_node_ids = basic_block_ids.clone();
+
+                    // Add join nodes
+                    for &(left_idx, right_idx) in &join_pairs {
+                        if left_idx < all_node_ids.len() && right_idx < all_node_ids.len() {
```
plafer, line 478: We expect the children to exist, right? Hence, shouldn't we `assert!` that this condition is true to catch bugs, instead of "silently failing"?

Comment applies everywhere else we do this (e.g. the split nodes below, etc).

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +228,218 @@ impl Arbitrary for MastForest {
...
+
+                    // Add call nodes (mix of regular calls and syscalls)
+                    for (i, &callee_idx) in call_indices.iter().enumerate() {
+                        if callee_idx < all_node_ids.len() {
+                            let callee_id = all_node_ids[callee_idx];
+                            let call_id = if i % 3 == 0 {
+                                // Make every 3rd call a syscall
+                                forest
+                                    .add_syscall(callee_id)
+                                    .unwrap_or_else(|_| forest.add_call(callee_id).unwrap())
```
plafer: Any program with a syscall will [fail to execute](https://github.com/0xMiden/miden-vm/blob/0f83da472687f44ec8d682f6e483ce212ef0bc78/processor/src/fast/call_and_dyn.rs#L55-L57) unless we provide a way to properly build the `Kernel` (i.e. provide all the procedure hashes that are syscall'd).

Haven't thought about how to fix this, but a workaround for now would be to distinguish `CALL` and `SYSCALL` in the `MastForestParams`. Hence, if you want to generate a `MastForest` for the purposes of assembly but never need to execute it, you can add some syscalls in there. But if you want to execute the `MastForest`, then you can set the number of syscalls to 0.

We could also potentially add a `kernel_procedures` field to the `MastForest` - I can see this information being useful in other contexts, but haven't thought about any repercussions yet.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +228,218 @@ impl Arbitrary for MastForest {
...
+                            all_node_ids.push(call_id);
+                        }
+                    }
+
+                    // Add external nodes
+                    for digest in external_digests {
+                        if let Ok(external_id) = forest.add_external(digest) {
+                            all_node_ids.push(external_id);
+                        }
+                    }
```
plafer, line 534: Similarly here, the external digests are just random words, and so any program with them will fail to execute.

Though the number of externals is in `MastForestParams`, so we could simply document that if you intend to execute this `MastForest`, then this parameter must be set to 0.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -207,31 +228,218 @@ impl Arbitrary for MastForest {
...
+
+                    // Add dyn nodes (mix of dyn and dyncall)
+                    for i in 0..num_dyns {
+                        let dyn_id = if i % 2 == 0 {
+                            forest.add_dyn().unwrap()
+                        } else {
+                            forest.add_dyncall().unwrap()
+                        };
+                        all_node_ids.push(dyn_id);
+                    }
```
plafer, line 545: Similarly here, this will fail to execute because there's junk on the stack. This is a bigger lift to make executable, so we can probably just document that for now at least, if you plan to execute the MAST forest, then you shouldn't include any dyn's.

plafer (commented):
Thank you for this!

I left some comments in the context of making `MastForest` that will actually properly execute (which would be awesome to have).

huitseeker:
@plafer I've documented the limitations and made the default value for the invalid nodes be 0 in 2a5e561 I've also opened #2341 as a follow-up.

Issue #2337, opened 2025-11-03
Parallel trace generation: don't assume valid inputs


plafer:
In parallel trace generation, we currently assume that the inputs (e.g. `TraceFragmentContext`s) were correctly built by the `FastProcessor` (i.e. we panic on error).

However, if we move to a model (e.g. with #2252) where the trace fragment contexts are received over the network and could have been built arbitrarily, then we will need to return errors in parallel trace generation instead of panicking.

Himess:
Can i take ? @plafer

plafer:
We're not sure whether or not we want to do this one, so let's hold off for now

bobbinth:
I believe this was closed by https://github.com/0xMiden/miden-vm/pull/2747 and seems like it was a duplicate of https://github.com/0xMiden/miden-vm/issues/2738 (but @plafer, please re-open if not).

PR #2338, opened 2025-11-03
refactor: return errors instead of panicking in parallel trace generation

DeVikingMark:
## Describe your changes

Implements #2337. Replaces panics with error returns in parallel trace generation to support receiving `TraceFragmentContext`s over the network.

- Add `TraceGenerationError` variant
- Update `build_trace()` and helpers to return `Result`
- Replace panics with error returns
- Update call sites

### DIFF
--- a/miden-vm/benches/build_trace.rs
+++ b/miden-vm/benches/build_trace.rs
@@ -83,7 +83,8 @@ fn build_trace(c: &mut Criterion) {
                                 trace_generation_context,
                                 program.hash(),
                                 program.kernel().clone(),
-                            );
+                            )
+                            .unwrap();
                             black_box(trace);
                         },
                         BatchSize::SmallInput,


DeVikingMark:
sorry for ci, wil fix it alittle bit later, hope for the review to understand if Im going to the right direction

plafer:
Thank you for your interest in the project! However as per our [contributing guidelines](https://github.com/0xMiden/miden-vm/blob/next/CONTRIBUTING.md#contribution-quality), you should make sure that the issue is assigned to you before starting to work on it (2nd bullet point).

Whether or not we want to return errors in parallel trace generation is still currently being debated, and would be done as part of a larger refactoring work.

I will try to take a closer look later, but we will probably end up closing this. In the future, I would recommend first commenting on the issue you would like to work on before starting the work, to make sure that it is one worth working on at the moment.

DeVikingMark:
> Thank you for your interest in the project! However as per our [contributing guidelines](https://github.com/0xMiden/miden-vm/blob/next/CONTRIBUTING.md#contribution-quality), you should make sure that the issue is assigned to you before starting to work on it (2nd bullet point).
> 
> Whether or not we want to return errors in parallel trace generation is still currently being debated, and would be done as part of a larger refactoring work.
> 
> I will try to take a closer look later, but we will probably end up closing this. In the future, I would recommend first commenting on the issue you would like to work on before starting the work, to make sure that it is one worth working on at the moment.

Ok, I got it, thanks for clarifying.

PR #2340, opened 2025-11-04
refactor: use builder pattern in parse_advice_inputs

### LINKED ISSUE #2339: Use builder pattern in parse_advice_inputs instead of reassignments
The `parse_advice_inputs` method does multiple reassignments with `mut` instead of using the builder pattern. `AdviceInputs` already has `with_stack_values`, `with_map`, and `with_merkle_store`, but the code doesn't use them properly.

Other parts of the codebase (like in `test_builders.rs`) chain these methods, so this is inconsistent. Should refactor to match that style.

Less mutability, more consistent with the rest of the code.

CreeptoGengar:
Closes #2339


plafer:
Per our contributing guidelines, I don't think we need this change.

Issue #2343, opened 2025-11-05
feat(stdlib, ecdsa, eddsa): introduce support for generating ECDSA/EdDSA signatures


adr1anh:
With the introduction of ECDSA signature verification, we should look into adding handlers to support generating signatures in the Host which can then be verified by the precompile.

Relevant links:
- [example in Aptos](https://github.com/aptos-labs/aptos-core/blob/20120ba4832299db2335f8521f524535ced9c677/consensus/src/pipeline/signing_phase.rs#L43-L49)
- [miden-base](https://github.com/0xMiden/miden-base/pull/2052#discussion_r2491857076)

bobbinth:
Post https://github.com/0xMiden/miden-vm/pull/2344, I think there are a few specific things we need to do:

- For each signature scheme, we should provide a MASM procedure that takes a public key commitment and a message word on the stack, and get the actual public key and signature via the advice provider.
  -  We should probably uniformly name this procedure `verify`, which means that for the ECDSA scheme, the current `verify` would become `verify_prehash`.
- We should provide Rust functions for generating the data that needs to be put into the advice provider (i.e., the public key and the signature).
    - For ECDSA we already have these functions, but we probably should organize them better.
    - For Falcon we have one of the functions (to generate the signature), and the other one lives in `miden-base` ([here](prepare_rpo_falcon512_signature)) - we should probably bring it over.
    - For EdDSA it is a bit more tricky because, AFAIK, there is no ec-recover for EdDSA - and so, we need to get the full public key from somewhere. One option is to redefine `EdDSA` signature to include both signature and the public key (same as we did with Falcon). Another option is to modify the mechanism in `miden-base` of how we get public keys - e.g., maybe the we should add a method to the authenticator to get public key for a given public key commitment.

cc @mmagician and @PhilippGackstatter for thoughts on the last point.

bobbinth:
One other thing that would be nice to do is to add test to make sure that ECDSA and EdDSA public key commitments are computed in the same way both in MASM and in Rust. After https://github.com/0xMiden/crypto/pull/643 I believe that should be the case, but we don't have tests for this.

PhilippGackstatter:
> * One option is to redefine `EdDSA` signature to include both signature and the public key

I think this is a common thing for `EdDSA` signatures and it would also be consistent with how we do it for falcon signatures. I don't have a very strong opinion on this, though.

bobbinth:
[...]

PR #2344, opened 2025-11-05
feat: add ecdsa wrapper for standard signature format

adr1anh:
## Summary

- Adds `verify_ecdsa_k256_keccak` procedure to stdlib for verifying secp256k1 ECDSA signatures using the `miden-crypto` format
- Introduces `word::store_word_u32s_le` helper for converting words to u32 limbs in memory
- Updates `miden-crypto` dependency from 0.18.0 to 0.18.1 to support the new verification flow
- Adds `ecdsa_sign` helper function for generating compatible calldata from secret keys

### DIFF
--- a/stdlib/src/lib.rs
+++ b/stdlib/src/lib.rs
@@ -10,7 +10,7 @@ use miden_assembly::{Library, mast::MastForest, utils::Deserializable};
 use miden_core::{
     EventName, Felt, Word, precompile::PrecompileVerifierRegistry, utils::Serializable,
 };
-use miden_crypto::dsa::ecdsa_k256_keccak::SecretKey;
+use miden_crypto::dsa::ecdsa_k256_keccak;
 use miden_processor::{EventHandler, HostLibrary};
 use miden_utils_sync::LazyLock;
 
@@ -103,24 +103,53 @@ impl Default for StdLibrary {
     }
 }
 
-/// Signs the provided message with the supplied secret key and returns the encoded calldata
-/// expected by `stdlib::crypto::dsa::ecdsa::verify_ecdsa_k256_keccak`.
+// ECDSA SIGNATURE
+// ================================================================================================
+
+/// Signs the provided message with the supplied secret key and encodes this signature and the
+/// associated public key into a vector of field elements in the format expected by
+/// `stdlib::crypto::dsa::ecdsa::secp256k1::verify_ecdsa_k256_keccak` procedure.
+///
+/// See [encode_ecdsa_signature()] for more info.
+pub fn ecdsa_sign(sk: &ecdsa_k256_keccak::SecretKey, msg: Word) -> Vec<Felt> {
+    let pk = sk.public_key();
+    let sig = sk.sign(msg);
+    encode_ecdsa_signature(&pk, &sig)
+}
+
+/// Infers the pubic key from the provided signature and message, and encodes this public key and
+/// signature into a vector of field elements in the format expected by
+/// `stdlib::crypto::dsa::ecdsa::secp256k1::verify_ecdsa_k256_keccak` procedure.
+///
+/// See [encode_ecdsa_signature()] for more info.
+///
+/// # Errors
+/// Returns an error if key recovery from signature and message fails.
+pub fn prepare_ecdsa_signature(
+    msg: Word,
+    sig: &ecdsa_k256_keccak::Signature,
+) -> Result<Vec<Felt>, ecdsa_k256_keccak::PublicKeyError> {
+    let pk = ecdsa_k256_keccak::PublicKey::recover_from(msg, sig)?;
+    Ok(encode_ecdsa_signature(&pk, sig))
+}
+
+/// Encodes the provided public key and signature into a vector of field elements in the format
+/// expected by `stdlib::crypto::dsa::ecdsa::secp256k1::verify_ecdsa_k256_keccak` procedure.
 ///
 /// 1. The compressed secp256k1 public key encoded as 9 packed-u32 felts (33 bytes total).
 /// 2. The ECDSA signature encoded as 17 packed-u32 felts (66 bytes total).
 ///
 /// The two chunks are concatenated as `[PK[9] || SIG[17]]` so they can be streamed straight to
-/// the advice provider before invoking `verify_ecdsa_k256_keccak`.
-pub fn ecdsa_sign(sk: &SecretKey, msg: Word) -> Vec<Felt> {
+/// the advice provider before invoking `secp256k1::verify_ecdsa_k256_keccak`.
+pub fn encode_ecdsa_signature(
+    pk: &ecdsa_k256_keccak::PublicKey,
+    sig: &ecdsa_k256_keccak::Signature,
+) -> Vec<Felt> {
     let mut out = Vec::new();
-    let pk = sk.public_key();
     let pk_bytes = pk.to_bytes();
     out.extend(bytes_to_packed_u32_felts(&pk_bytes));
-
-    let sig = sk.sign(msg);
     let sig_bytes = sig.to_bytes();
     out.extend(bytes_to_packed_u32_felts(&sig_bytes));
-
     out
 }
 


stdlib/asm/word.masm
```diff
@@ -12,6 +12,40 @@ export.reverse
     reversew
 end
 
+# CONVERSION UTILITIES
+#! Writes the components of a word to memory as eight u32 limbs in little-endian order.
+#!
+#! Inputs:  [W3, W2, W1, W0, OUT_PTR, ...]
```
bobbinth: nit: we usually use lowercase letters to denote felts and upper case letters to denote words. So, assuming all these values are felts, we'd write this as:
```
Inputs:  [w3, w2, w1, w0, out_ptr, ...]
```

stdlib/asm/word.masm
```diff
@@ -12,6 +12,40 @@ export.reverse
...
+#! Cycles: 8 * (split + store pair) ~= 176
+export.store_word_u32s_le
+    reversew
+    # => [W0, W1, W2, W3, OUT_PTR, ...]
+
+    u32split
+    # => [W0_hi, W0_lo, W1, W2, W3, OUT_PTR, ...]
+
+    movup.2 u32split
+    # => [W1_hi, W1_lo, W0_hi, W0_lo, W2, W3, OUT_PTR, ...]
```
bobbinth: nit: we could also store the first word to memory here and then drop it from the stack (rather than splitting all elements and then storing all of them after that). May simplify things a little.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
 # PROCEDURES
 # ================================================================================================
 
+#! Prepares calldata for deferred secp256k1 verification.
```
bobbinth: I think this needs to be updated as this procedure also verifies the signature.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+#!
+#! This wrapper mirrors the interface of `rpo_falcon512::verify`: the commitment to the public key
+#! and the message arrive on the operand stack, while the concrete public key and signature bytes
+#! are streamed from the advice stack. The routine stages the calldata in local memory before
+#! invoking the precompile.
+#!
+#! Inputs:
+#!   Operand stack: [PK_RPO, MSG_WORD, ...]
```
bobbinth: nit: I would maybe use `PK_COMMITMENT` instead of `PK_RPO`. In `miden-base` we have `PublicKeyCommitment` type which corresponds to this.

I would also maybe just use `MSG` (writing it in upper case letters already implies that it is a word).

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+#!   - locaddr[20..=27] : keccak256(message) digest (8 felts)
+#!   - locaddr[28..=44] : signature (66 bytes packed as 17 felts)
+#!
+#! The procedure traps if the staged public key does not hash to `PK_RPO`; otherwise it returns
+#! cleanly after emitting the deferred verification request.
+export.verify_bis.45
+    # Load the compressed public key (9 felts) into local memory starting at locaddr.[0..9]
+    adv_push.4 locaddr.0 mem_storew_le dropw
+    adv_push.4 locaddr.4 mem_storew_le dropw
+    adv_push.1 locaddr.8 mem_store
```
bobbinth: It would be more efficient to do this as:
```
padw
adv_loadw loc_storew_le.0
adv_loadw loc_storew_le.4
adv_push.1 loc_store.8
dropw
```
This should save at least 14 cycles.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+    locaddr.20 mem_storew_le dropw
+    locaddr.24 mem_storew_le
+    # => [DIGEST[4..8], ...]
+
+    # store the signature (17 felts) provided via advice at locaddr.[28..45]
+    adv_loadw  locaddr.28 mem_storew_le
+    adv_loadw  locaddr.32 mem_storew_le
+    adv_loadw  locaddr.36 mem_storew_le
+    adv_loadw  locaddr.40 mem_storew_le dropw
+    adv_push.1 locaddr.44 mem_store
```
bobbinth: Similarly, here we can use `loc_storew_le` instead of using `locaddr`.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+    # => [...]
+
+    # Invoke the deferred verification precompile with in-memory calldata
+    locaddr.28
+    locaddr.20
+    locaddr.0
+    # => [pk_ptr, digest_ptr, sig_ptr]
+    exec.verify
+    drop
```
bobbinth: Why do we drop the result here? Should we not assert that the result is `1`?

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+    locaddr.12 movdn.5 exec.word::store_word_u32s_le
+    # [...]
+
+    # Hash the 32-byte message with Keccak256
+    push.32 locaddr.12 exec.keccak256::hash_memory
+    # => [DIGEST[0..8], ...]
+
+    # Store the resulting digest at locaddr.[20..28]
+    locaddr.20 mem_storew_le dropw
+    locaddr.24 mem_storew_le
```
bobbinth: Similar to some other comments, we could use `loc_storew_le` instructions here.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,6 +24,69 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+#!
+#! Local memory layout (element addresses):
+#!   - locaddr[0..=8]   : compressed public key (33 bytes packed as 9 felts)
+#!   - locaddr[12..=19] : message bytes (MSG_WORD written as eight u32 limbs)
+#!   - locaddr[20..=27] : keccak256(message) digest (8 felts)
+#!   - locaddr[28..=44] : signature (66 bytes packed as 17 felts)
+#!
+#! The procedure traps if the staged public key does not hash to `PK_RPO`; otherwise it returns
+#! cleanly after emitting the deferred verification request.
+export.verify_bis.45
```
bobbinth: This is fine for this PR (as we don't want to introduce breaking changes), but for v0.20 release, I'd consider making this the main `verify` procedure and renaming the other one into something like `verify_from_memory` or maybe `verify_internal` if we don't want to expose it.

stdlib/src/lib.rs
```diff
@@ -99,6 +103,29 @@ impl Default for StdLibrary {
...
+/// expected by `stdlib::crypto::dsa::ecdsa::verify_bis`.
+///
+/// The returned felt sequence mirrors the interface of `falcon_sign`:
+///
+/// 1. The compressed secp256k1 public key encoded as 9 packed-u32 felts (33 bytes total).
+/// 2. The ECDSA signature encoded as 17 packed-u32 felts (66 bytes total).
+///
+/// The two chunks are concatenated as `[PK_BYTES || SIG_BYTES]` so they can be streamed straight to
+/// the advice provider before invoking `verify_bis`.
+pub fn ecdsa_sign(sk: &mut SecretKey, msg: Word) -> Vec<Felt> {
```
bobbinth: Does `SecreteKey` need to be mutable here? I think if you run `cargo update` we should get the version of `miden-crypto` that doesn't require this.

Also, we may not always have access to the private key - so, in addition to this function, it would be great to expose a function that converts a public key and a signature into a vector of bytes. Or, even better, if we could convert message and signature into a vector of bytes (maybe using EC recover?) - maybe something like:

```rust
pub fn prepare_ecdsa_signature(pk: &PublicKey, msg: Word) -> Vec<Felt> {
    ...
}
```

For v0.20 release, it would be great to find a better place for these (and for `falcon_sign()`) functions. Exporting them from the library root feels a bit weird.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
adr1anh: I think the existing one should be called `verify_prehash`

stdlib/src/lib.rs
adr1anh: I think for now this is okay, but we should do this more thoroughly in #2343.

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
```diff
@@ -22,10 +24,76 @@ const.SIG_LEN_FELTS=17   # 66.div_ceil(4)
...
+    # => [...]
+
+    # Invoke the deferred verification precompile with in-memory calldata
+    locaddr.28
+    locaddr.20
+    locaddr.0
+    # => [pk_ptr, digest_ptr, sig_ptr]
+    exec.verify
+    # => [result, ...]
+    assert.err="ECDSA signature verification failed"
```
onurinanc, line 88: I'm not sure but I think we shouldn't have the following error here. The reason is that we would like to have an `Unauthorized` as from the TransactionExecutor. If we have, assert.err="ECDSA signature verification failed" message, TransactionExecutor would not be able to receive a `Box<TransactionSummary>` as an output. If we have this error message, then we won't be able to get the message from the executor. cc @bobbinth

stdlib/src/lib.rs
bobbinth: This would actually be super-helpful (specifically the `pub_key + msg` version) because it would plug into `miden-base` pretty easily. Without this, we'd need to modify the existing `Signature` enum to include both signature and public key (the way we do it for Falcon).

stdlib/asm/crypto/dsa/ecdsa/secp256k1.masm
adr1anh: Would it be easier if we returned the Boolean result of the verification?

onurinanc: I believe yes, [falcon implementation](https://github.com/0xMiden/miden-vm/blob/686fea2ac06c7a90a699e4109645d2e363ab2b9e/stdlib/asm/crypto/dsa/rpo_falcon512.masm#L655) returns a bool, too

bobbinth: I believe Falcon signature errors out on invalid signature - so, the way it is now in this PR is consistent with that. We can consider changing this in the future - maybe as a part of https://github.com/0xMiden/miden-vm/issues/2343.

stdlib/src/lib.rs
bobbinth: I added a couple more methods in 026321072dae27a858bc4448e30155e0c619a7d4 to address this.

bobbinth (approved):
Looks good! Thank you!

PR #2345, opened 2025-11-05
opt(processor): pre-allocate main trace buffer in trace generation

### LINKED ISSUE #2160: Parallel trace generation: pre-allocate main trace buffer
Follow-up to #1839.

The current strategy for `CoreTraceFragmentGenerator`s is an artifact of the previous strategy described [here](https://github.com/0xMiden/miden-vm/issues/1558#issuecomment-2572407377), where trace generation would receive trace fragment contexts one by one, rather than all at once. We since then switched to the approach where `FastProcessor::execute_for_trace()` returns _all_ the trace fragment contexts at once, and therefore `build_trace()` has all the information ahead of time.

Currently, each `CoreTraceFragmentGenerator` acts in isolation, and doesn't know anything about other fragments. It writes in its own trace fragment buffer (which later gets copied into the `MainTrace` buffer), and also returns a system & stack row that is to be written into the following fragment (since these are offset by 1).

Now that `build_trace()` knows all the fragment contexts at the start, we can pre-allocate the entire `MainTrace` buffer, because we know its length, as well as where each trace fragment will fall. We can then use `rayon` to have threads write in it directly. Care will need to be taken for the off-by-one issue for stack and system columns - it's not immediately clear to me what's the cleanest way to do it.

This new strategy has the potential to improve performance (by reducing the amount of memory we allocate, as well as remove copies), as well as improve code clarity (specifically by hopefully removing the need to pass stack and system rows between fragments). 

> By off-by-one issue for stack and system rows, I mean that for system and stack columns, the data for clock `i` is written at row `i+1`, and hence falls in the next fragment.

plafer:
Closes #2160

This currently 2x the performance of trace generation, but I would have expected closer to 5x. However, I'll leave the performance analysis + further improvements for the future when we re-prioritize it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - [BREAKING] Standardized hash function naming: renamed `hash_2to1` → `merge` and `hash_1to1` → `hash` across all hash modules (blake3, sha256, keccak256, rpo) ([#2381](https://github.com/0xMiden/miden-vm/pull/2381)).
 - Consolidate debug information into `DebugInfo` struct ([#2366](https://github.com/0xMiden/miden-vm/issues/2366)).
 - Wrapped `hperm` instruction in `rpo::permute` procedure ([#2392](https://github.com/0xMiden/miden-vm/pull/2392)).
+- Pre-allocate main trace buffer in trace generation ([#2345](https://github.com/0xMiden/miden-vm/pull/2345)).
 
 ## 0.19.1 (2025-11-6)
 


processor/src/parallel/mod.rs
```diff
@@ -270,57 +267,77 @@ fn generate_core_trace_fragments(
         ),
     );
 
-    fragments
+    // Run batch inversion on stack's H0 helper column
+    core_trace_columns[STACK_TRACE_OFFSET + H0_COL_IDX] =
+        batch_inversion(&core_trace_columns[STACK_TRACE_OFFSET + H0_COL_IDX]);
```
Al-Kindi-0, line 272: Q: Should we create the issue for investigating per-fragment batch-inversion as we discussed last week?

Issue #2298, opened 2025-10-17
Add `adv.push_mapval_count` instruction


mmagician:
Not for this PR, but we probably get rid of the `get_num_procedures` procedure at the kernel level. Instead in the `miden::account::get_num_procedures` we could get the account code commitment, and then get the number of elements in its pre-image.

However, doing this efficiently would require a new VM instruction - similar to `adv.push_mapvaln` but that pushes only the number of elements in the value for the given key onto the advice stack (right now, `adv.push_mapvaln` would push both the number of elements and the actual pre-image onto the advice stack).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-base/pull/1968#discussion_r2422578312_

adr1anh:
Could this be an event?

bobbinth:
> Could this be an event?

This would use an `emit` instruction under the hood. But maybe you meant that it could be a "library-specific" event? If so, this would work too, but I think this functionality could be generally useful and having it as a "core" instruction would also be consistent with other similar instructions. So, my preference would be to add it at the instruction level.

PR #2349, opened 2025-11-07
Implement `adv.push_mapval_count` instruction

Fumuran:
This PR adds a new event based instruction: `adv.push_mapval_count`.

It works the same way as `adv.push_mapvaln`, the main difference is that it pushes only the number of values in the array which was obtained from the advice map by some key. The resulting array length is pushed to the advice stack.

Operand stack and advice provider transition looks like so:
```
Inputs:
  Operand stack: [KEY, ...]
  Advice stack: [...]
  Advice map: {KEY: [values_array]}

Outputs:
  Operand stack: [KEY, ...]
  Advice stack: [num_values, ...]
  Advice map: {KEY: [values_array]}
```

Closes: #2298.


core/src/sys_events.rs
```diff
@@ -68,6 +68,20 @@ pub enum SystemEvent {
...
+    /// looked up in the advice map using the specified word from the operand stack as the key.
+    ///
+    /// Inputs:
+    ///   Operand stack: [KEY, ...]
+    ///   Advice stack: [...]
+    ///   Advice map: {KEY: values}
+    ///
+    /// Outputs:
+    ///   Operand stack: [KEY, ...]
+    ///   Advice stack: [num_values, ...]
```
adr1anh: Took me a second to get what `num_values` was, and thought it might have been an argument to the instruction. Maybe this is slightly clearer?
```suggestion
    ///   Advice stack: [values.len(), ...]
```

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -295,6 +296,43 @@ fn copy_map_value_to_adv_stack(
...
+///
+/// # Errors
+/// Returns an error if the required key was not found in the key-value map.
+fn copy_map_value_length_to_adv_stack(
+    process: &mut ProcessState,
+    err_ctx: &impl ErrorContext,
+) -> Result<(), ExecutionError> {
+    let key = process.get_stack_word_be(1);
+    let process_clk = process.clk();
+    let advice_provider = process.advice_provider_mut();
```
adr1anh, line 320:
```suggestion
    let advice_provider = process.advice_provider();
```

Fumuran: I'm not sure that we will be able to use the immutable version of the advice provider: we mutate it later [here](https://github.com/0xMiden/miden-vm/blob/520d731abac9ca79e93a94ca1efeff5ea367833b/processor/src/operations/sys_ops/sys_event_handlers.rs#L331) while pushing the length to the advice stack. Or probably I'm missing something.

PR #2350, opened 2025-11-07
fix(masm): align locaddr when proc locals not multiple of 4

### LINKED ISSUE #2348: bug(masm): `locaddr.0` is not aligned when number of proc locals is not a multiple of 4
Example:

```rust
#[test]
fn locaddr_word_alignment_non_multiple_of_four() {
    let source = "
        proc.foo.5
            loc_storew_be.0
        end
        begin
            exec.foo
        end";

    build_test!(source, &[]).execute().unwrap();
}
```

sky-coderay:
## Describe your changes
Fixes #2348

### DIFF
--- a/miden-vm/tests/integration/operations/io_ops/local_ops.rs
+++ b/miden-vm/tests/integration/operations/io_ops/local_ops.rs
@@ -358,7 +358,8 @@ fn locaddr_word_alignment_non_multiple_of_four() {
     // Test for issue #2348: locaddr.0 should be word-aligned even when
     // number of proc locals is not a multiple of 4
     let source = "
-        proc.foo.5
+        @locals(5)
+        proc foo
             loc_storew_be.0
         end
         begin


adr1anh:
This seems fine, and is the same change Claude suggested to me when playing around with it. I'd defer to @plafer who recently updated how we handle frame pointers

sky-coderay:
I’ll wait for further instructions, after which I’ll fix the CI errors, thanks

bobbinth (commented):
Looks good! Thank you for fixing this! Let's make the CI green and I think this would be good to merge.

crates/assembly/src/instruction/mem_ops.rs
```diff
@@ -184,7 +184,11 @@ pub fn local_to_absolute_addr(
         );
     }
 
-    let fmp_offset_of_local = -Felt::from(num_proc_locals - index_of_local);
+    // The frame pointer is always incremented by a word-aligned number of locals to ensure
```
bitwalker, line 187: I'm not sure this is the right place to make this change. @bobbinth correct me if I'm wrong, but the "correct" fix would be to have `fmpupdate`/`fmpadd` actually enforce that the new fmp value is word-aligned, rather than expecting the assembler to do this, since nothing prevents someone from constructing MAST by hand that doesn't do this, so it would need to be enforced by the VM.

This is a reasonable temporary workaround though.

bobbinth: `fmpupdate` and `fmpadd` instructions have been removed recently - and `fmp` now lives in memory (at address $2^{32} - 1$, I believe). So, this is all handled by the assembler now.

bitwalker: Right, but now the updates to `fmp` are performed by `exec`/`dynexec`, which could easily round up to the nearest word-aligned address and enforce this behavior.

bobbinth: I think it was always like this. The number of locals allocated was always a multiple of 4 (and this was handled by the assembler in `fmp_start_frame_sequence()` and `fmp_end_frame_sequence()`). The bug was that when calculating absolute addresses of locals, we were not taking this padding into account.

But maybe I'm not understanding what you are suggesting?

bitwalker: What I'm suggesting is that the guarantee we want to provide, that procedure locals always begin at a word-aligned address, can only really be provided by the code which manages the absolute address stored in `fmp` - that is now handled by the procedure invocation instructions (e.g. `exec`), so I would expect that the handling of `fmp` associated with those instructions would handle this alignment detail, since it is technically a property of the VM (i.e. that upon entry to a procedure, the value of `fmp` is word-aligned).

By having the assembler handle this, the guarantee _could_ be circumvented. I'm not sure you can do anything nasty by doing so, but it feels odd to make the assembler responsible for something like this.

That said, I'll defer to you on which approach is best here, just something that caught my eye

bobbinth: > that is now handled by the procedure invocation instructions (e.g. `exec`),

I'm not sure that's the case. The code associated with manipulating the `fmp` value is a part of the procedure itself (i.e, procedure prologue and epilogue). We need to do this because we don't know how exactly the procedure will be invoked - e.g.,:

- It could be invoked via `exec` and potentially inlined at assembly time.
- It could be invoked via `dynexec`.
- It could be invoked via `call` or `dyncall`.

> I would expect that the handling of `fmp` associated with those instructions would handle this alignment detail, since it is technically a property of the VM (i.e. that upon entry to a procedure, the value of `fmp` is word-aligned).

> By having the assembler handle this, the guarantee _could_ be circumvented. I'm not sure you can do anything nasty by doing so, but it feels odd to make the assembler responsible for something like this.

We can't manipulate the `fmp` value at the VM level, because we don't always know when we start or exit a given procedure (e.g., if a procedure was inlined, the VM has no way to identify it as a procedure in the original source code), but what we can do is detect updates to the `fmp` (because it is in a well-known address) and raise an error if it is being set to a non word-aligned value.

This would be different from this issue though: `fmp` value was already always word-aligned even before this PR. The fix here is for an incorrect local to absolute address conversion.

crates/assembly/src/instruction/mem_ops.rs
```diff
@@ -184,7 +184,11 @@ pub fn local_to_absolute_addr(
...
+    // word-aligned memory access. We need to use the same alignment when calculating the offset.
+    // See fmp_start_frame_sequence() and fmp_end_frame_sequence() in fmp.rs.
+    let aligned_num_locals = num_proc_locals.next_multiple_of(WORD_SIZE as u16);
+    let fmp_offset_of_local = -Felt::from(aligned_num_locals - index_of_local);
```
plafer, line 191:
```suggestion
    let fmp_offset_of_local = Felt::from(index_of_local - aligned_num_locals);
```
Not from this PR, but this is more readable. Otherwise, the outer `-` is easy to miss.

plafer (approved):
LGTM - thank you for fixing! Left one small improvement (not even from this PR).

Last remaining thing here is that we're getting a 403 on the Contribution Quality CI job cc @huitseeker

sky-coderay:
Hey @plafer! Thanks for the suggestion, but it causes overflow - `index_of_local` is usually smaller than `aligned_num_locals` (both u16), so the subtraction panics.

The original version works because it does `aligned_num_locals - index_of_local` first (safe), then negates.

How about this for better readability?

let offset = aligned_num_locals - index_of_local;
let fmp_offset_of_local = -Felt::from(offset);

plafer:
Wah, of course, because they're unsigned ints... Nevermind, please disregard my comment and leave it as it was.

crates/assembly/src/instruction/mem_ops.rs
plafer: As mentioned below, please disregard and leave it as it was

plafer:
@sky-coderay ~there's still a test failing, can you look into it?~

EDIT: I fixed the test

Issue #1983, opened 2025-07-11
Implement `memcopy` procedure in `stdlib`


Fumuran:
We should implement a new `memcopy` procedure in the `std::mem` module. It should copy `Felt`s from the `read_prt` to the `write_ptr`. It should work similar to the `memcopy_words` procedure, but should work directly with the `Felt`s.

 _Originally posted by @bobbinth in [#594](https://github.com/0xMiden/compiler/issues/594#issuecomment-3058508920)_

plafer:
Internally, we should try to use `memcopy_words` for as many felts as possible, and only the last `< 4` felts should be copied one by one. 

@Al-Kindi-0 didn't you already have that logic somewhere in the recursive verifier?

bitwalker:
I suspect the overhead of computing that and handling the branches will be more expensive than just doing the naive felt-by-felt copy for the vast majority of copies I'm expecting to use this with.

In general, I think procedures like this should be low-level primitives, and avoid trying to be overly smart. I was planning to use this in the compiler, but if it is going to do anything other than a straight copy, we're better off having our own intrinsics for this stuff (which was originally the plan, but @bobbinth suggested we implement them in `std` instead). Callers are going to be better positioned to know whether they can call `std::mem::memcopy_words` as an optimization anyway.

plafer:
We could then potentially provide a `memcopy_smart` or whatever, that works as I previously described. I can see this being useful for use cases when the data to be copied is expected to be relatively large, but where you that doesn't necessarily align on word boundaries.

Al-Kindi-0:
> [@Al-Kindi-0](https://github.com/Al-Kindi-0) didn't you already have that logic somewhere in the recursive verifier?

Don't think so, but maybe it is the `memcopy_words` in the stdlib that you are thinking of.


Related to the procedure itself, I would say that if the memory region to copy is relatively small, then as @bitwalker stated, it is better to keep the logic simple as the overhead will be more than the benefit we will get.
However, if the region is very large, then it would be better to use either word or double-word memory load and store ops. One would then have to deal with both ends of the region with respect to word or double-word alignment.

Probably worth having two versions, one which operates on `Felt`-s and another which works on double-words, where the former will be very simple logic-wise and where the latter would be more complicated but better suited for large chunks of memory. The decision of which one to use is then left to the caller (we can provide formulas for cycle count depending on the memory region size to help with this).
 That's my two cents given my limited knowledge of the problems we will typically want to solve.

[...]

PR #2352, opened 2025-11-07
Implement `std::mem::memcopy_felts` procedure

Fumuran:
This PR implements new `memcopy_felts` procedure for the `std::mem` module.

It copies the provided number of `Felt`s from the read pointer to the write pointer.

The stack transition looks like so:
```
Inputs:  [n, read_ptr, write_ptr]
Outputs: []
```

Closes: #1983.


bobbinth:
Not a review, but could we also add a `memcopy` procedure? This procedure would split the input into chunks:

1. Initial chunk (up to 3 elements) to be copied element-by-element (if the input pointer is not word-aligned).
2. Middle chunk - to copy word by-word.
3. Final chunk (up to 3 elements) to be copied element-by-element (if the final pointer is not word-aligned).

Could also be done in another PR.

Fumuran:
>Not a review, but could we also add a `memcopy` procedure?

Makes sense, but I would vote to do so in a separate PR, just to avoid overloading this one. I can create a separate issue for this.

stdlib/asm/mem.masm
```diff
@@ -54,18 +60,64 @@ export.memcopy_words
...
+#! Inputs:  [n, read_ptr, write_ptr]
+#! Outputs: []
+#!
+#! Where:
+#! - n is the number of felts which should be copied.
+#! - read_ptr is the memory pointer where the felts to copy are stored.
+#! - write_ptr is the memory pointer where the felts will be copied.
+#!
+#! Total cycles: 7 + 14 * num_felts
+export.memcopy_felts
```
bobbinth: I would write this using new syntax - i.e., `pub proc memcopy_felts`. This way, once https://github.com/0xMiden/miden-vm/pull/2330, this will still compile.

stdlib/asm/mem.masm
```diff
@@ -54,18 +60,64 @@ export.memcopy_words
...
+        # perform read (2 cycles)
+        dup.1 mem_load
+        # => [value_to_store, -n, read_ptr, write_ptr]
+
+        # perform write (3 cycles)
+        dup.3 mem_store
+        # => [-n, read_ptr, write_ptr]
+
+        # update the counters (6 cycles)
+        add.1 movup.2 add.1 movup.2 add.1 movup.2
```
Al-Kindi-0: nit:
```
# update the pointers and the counter (6 cycles)
movup.2 add.1 movup.2 add.1 movup.2 add.1
```

stdlib/asm/mem.masm
```diff
@@ -54,18 +60,64 @@ pub proc memcopy_words
...
+#! Inputs:  [n, read_ptr, write_ptr]
+#! Outputs: []
+#!
+#! Where:
+#! - n is the number of felts which should be copied.
+#! - read_ptr is the memory pointer where the felts to copy are stored.
+#! - write_ptr is the memory pointer where the felts will be copied.
+#!
+#! Total cycles: 7 + 14 * num_felts
+pub proc memcopy_felts
```
bobbinth: nit: can we rename this to `memcopy_elements`? (sorry - I should have thought about this earlier).

Fumuran: Should I also update all usages of `felts` to be `elements` for this procedure? Like in the docs:
```
#! Copies `n` elements from `read_ptr` to `write_ptr`.
```

bobbinth: Yes, that would be great.

PR #2353, opened 2025-11-08
fix: macOS `sed` issue + small echo tweak

mdqst:
## Describe your changes

fixed `sed_i` so it works on macOS too (BSD `sed` needs that empty `''` arg).
also added a space in the echo message so package names don’t stick to the text.

- Updated `CHANGELOG.md'

### DIFF
--- a/scripts/check-msrv.sh
+++ b/scripts/check-msrv.sh
@@ -28,7 +28,9 @@ sed_i() {
   if sed --version >/dev/null 2>&1; then
     sed -i "$@"
   else
-    sed -i '' "$@"
+    local expr=$1
+    shift
+    sed -i '' "$expr" "$@"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


huitseeker (commented):
Can you clarify? The empty argument to sed is passed here, and this script is tested on OSX.

mdqst:
[huitseeker](https://github.com/huitseeker) sure! on macOS (BSD `sed`), the `-i` flag requires an explicit argument for the backup extension - even if it’s empty.
so `sed -i 's/foo/bar/' file` fails with:

```
sed: 1: "file": invalid command code f
```

while `sed -i '' 's/foo/bar/' file` works fine.

that’s why I added the empty `''` - it keeps the command compatible with both GNU and BSD `sed`.

the space in the `echo` message was just a small cosmetic fix so package names don’t stick to the preceding text.

scripts/check-msrv.sh
```diff
@@ -28,7 +28,9 @@ sed_i() {
   if sed --version >/dev/null 2>&1; then
     sed -i "$@"
   else
-    sed -i '' "$@"
```
huitseeker, line 31: @mdqst the empty argument seems to already be provided.

huitseeker:
Closing for inactivity / lack of clear issue

PR #2360, opened 2025-11-11
refactor: Hoisting PRs follow-up

### LINKED ISSUE #2356: fix: simplify decorator stripping capability
Good remark. The idea is that we want to strip *all* decorator in a `MastForest` recursively, a feature introduced in #2108, and currently served better by `clear()`. 

However, we had introduced `remove_decorators` in various traits and it found its way there to allow a `Linked` `MastNode` to have something sensible to call in this case. This later part doesn't really make sense any more.

_Originally posted by @huitseeker in https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2508761838_
            
TL;DR: decorator stripping was introduced at the `MastForest` level to allow for decorator-free `MastForest` serialization. Because of how `MastNode`s stored decorators prior to #2310/#2323, this percolated recursively to `MastNode` APIs that aim at removing decorators for one specific node. But after these PRs, those APIs pollute the representation, since `MastNode`s are immutable post `MastForest` insertion and they are no longer individually serializable.

This can be seen as a sub-issue of #1580

huitseeker:
A follow-up of #2310 and #2323.

- adds a unit test demonstrating the CSR representation in the comment of `OpToDecoratorIds` and its conversion to COO (raised by @Al-Kindi-0 in #2310 ),
- simplify decorator stripping capabilities, cutting off the recursive descent at the `MastForest` level, preserving the spirit of #2108 (raised by @plafer in #2323), closes #2356.
- introduces, where possible and useful, a debug-only assertion which checks that when we're looking up a `MastNode`' s decorator info by `MastNodeId` from its own APIs, that the ` MastForest`'s `nodes`  vector at the position given by the `MastNodeId`  is indeed `self` [^1],
- makes the processor use direct access to a linked node's op-indexed decorator information, rather than relying on `next_filtered`. Simplify `RawDecoratorInfoIterator`  as a consequence.

[^1]: studying where such an assertion fails immediately puts a spotlight on the two most delicate areas of the assembler

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -203,7 +203,7 @@ impl BasicBlockNode {
                 )
             },
             DecoratorStore::Linked { id } => {
-                // This is usedin MastForestMerger::merge_nodes, which strips the `MastForest` of
+                // This is used in MastForestMerger::merge_nodes, which strips the `MastForest` of
                 // some nodes before remapping decorators, so calling
                 // verify_node_in_forest will not work here.
 
@@ -325,7 +325,7 @@ impl BasicBlockNode {
                     .collect()
             },
             DecoratorStore::Linked { id } => {
-                // This is usedin MastForest::remove_nodes, which strips the `MastForest` of its
+                // This is used in MastForest::remove_nodes, which strips the `MastForest` of its
                 // nodes before remapping decorators, so calling
                 // verify_node_in_forest will not work here.
 


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -207,6 +203,10 @@ impl BasicBlockNode {
                 )
             },
             DecoratorStore::Linked { id } => {
+                // This is usedin MastForestMerger::merge_nodes, which strips the `MastForest` of
```
adr1anh:
```suggestion
                // This is used in MastForestMerger::merge_nodes, which strips the `MastForest` of
```

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -323,6 +325,10 @@ impl BasicBlockNode {
                     .collect()
             },
             DecoratorStore::Linked { id } => {
+                // This is usedin MastForest::remove_nodes, which strips the `MastForest` of its
```
adr1anh:
```suggestion
                // This is used in MastForest::remove_nodes, which strips the `MastForest` of its
```

adr1anh (approved):
LGTM

processor/src/fast/basic_block.rs
bobbinth, line 159: Why did we need to do this in two steps here? Could this not have been just:
```rust
for decorator in current_forest.decorators_for_op(node_id, op_idx_in_block) {
    self.execute_decorator(decorator, host)?;
}
```

bobbinth, line 121: Similar comment as above: could this not have been something like this?:
```rust
for decorator in current_forest.decorators_for_op(node_id, num_ops) {
    self.execute_decorator(decorator, host)?;
}
```

core/src/mast/mod.rs
```diff
@@ -153,11 +153,10 @@ impl MastForest {
...
     /// Removes all decorators from this MAST forest.
     pub fn strip_decorators(&mut self) {
-        for node in self.nodes.iter_mut() {
-            node.remove_decorators();
-        }
+        // Clear all decorator storage
         self.decorators = IndexVec::new();
         self.op_decorator_storage = OpToDecoratorIds::new();
+        self.node_decorator_storage.clear();
     }
```
bobbinth, line 160: Comment for the future: once decorators are stripped, we'll also need to run a "forest compaction" procedure because we may be able to eliminate some nodes which previously differed only by decorators. This may already be captured in one of the issues, but if not, let's note this.

bobbinth (commented):
Looks good! Thank you! I left a few comments for follow-up/future PRs.

PR #2361, opened 2025-11-11
Implement `std::mem::memcopy` procedure

### LINKED ISSUE #2359: Implement smart copy procedure for `std::mem`
Implement `memcopy` procedure which will internally use `memcopy_words` and `memcopy_felts` in order to optimize the copy of the arbitrary unaligned number of values. 

This procedure would split the input into chunks:
1. Initial chunk (up to 3 elements) to be copied element-by-element (if the input pointer is not word-aligned).
2. Middle chunk - to copy word by-word (if the `input_pointer - final_pointer` is multiple of 4, i.e. there will be the moment when both of them are word-aligned).
3. Final chunk (up to 3 elements) to be copied element-by-element (if the final pointer is not word-aligned).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/issues/2352#issuecomment-3505299989_

Fumuran:
This PR implements new `memcopy` procedure for the `std::mem` module.

This procedure copies the provided number of `Felt`s from the read pointer to the write pointer. It uses `memcopy_words` and `memcopy_felts` procedures internally in order to optimize the copy.

The stack transition looks like so:

Inputs:  `[n, read_ptr, write_ptr]`
Outputs: `[]`

Closes: #2359.


docs/src/user_docs/stdlib/mem.md
```diff
@@ -8,8 +8,9 @@ Module `std::mem` contains a set of utility procedures for working with random a
 
 | Procedure                              | Description   |
 | -------------------------------------- | ------------- |
+| `memcopy`                              | Copies `n` felts from `read_ptr` to `write_ptr`.<br/><br/>If possible, this procedure will copy word-aligned words internally using `memcopy_words` procedure, decreasing the total number of cycles required to make a copy.<br/><br/>It is advised to copy small chunks of memory using `memcopy_felts` procedure instead, since it won't have a computation logic overhead.<br/><br/>**Inputs:** `[n, read_ptr, write_ptr]`<br/>**Outputs:** `[]`<br/><br/>Total cycles:<ul><li>Read and write pointers are mutually unaligned: $23 + 14 * num\_felts$</li><li>Read and write pointers are mutually aligned: $66 + 14 * num\_prefix\_felts + 16 * num\_words + 14 * num\_suffix\_felts$</li></ul> |
```
Al-Kindi-0: Q: Could we pinpoint the number of felts where we should prefer one to the other?

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+#! - n is the number of felts which should be copied.
+#! - read_ptr is the memory pointer where the felts to copy are stored.
+#! - write_ptr is the memory pointer where the felts will be copied.
+#!
+#! Panics if:
+#! - read_ptr and write_ptr are not a u32 values.
+#!
+#! Total cycles:
+#! - if read and write pointers are mutually unaligned: 23 + 14 * num_felts
+#! - if read and write pointers are mutually aligned, it's equal to the sum of:
```
Al-Kindi-0: should the first one be`aligned` and the second `unaligned`?

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+#!   - prologue: 14 cycles
+#!   - prefix handling (worst case): 28 + 14 * num_felts_to_next_wa, where num_felts_to_next_wa is 
+#!     the number of felts required to make the read (and write) pointer word-aligned
+#!   - words handling: 15 + 16 * num_words
+#!   - suffix handling: 9 + 14 * num_remaining_felts
+#!     - in total: 66 + 14 * num_prefix_felts + 16 * num_words + 14 * num_suffix_felts
+export.memcopy
+    # check whether the difference between `read_ptr` and `write_ptr` is 4, otherwise these pointers
```
Al-Kindi-0: a multiple of 4 is more precise?

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+    movdn.2 u32assert2.err="memory addresses must be valid u32 values" # 2 cycles
+    # => [read_ptr, write_ptr, n]
+
+    # Compute the remainders of dividing read_ptr and write_ptr by 4. Compare them to get the 
+    # "alignment flag".
+    dup.1 u32mod.4 dup.1 u32mod.4 dup.1 eq # 12 cycles
+    # => [alignment_flag, write_ptr%4, read_ptr, write_ptr, n]
+
+    if.true
+        # if read pointer and write pointer are word-aligned between each other, we will be able to
```
Al-Kindi-0: I think a better terminology here is `mutually word-aligned`, which is defined as `r_ptr - w_ptr = 0 mod 4`

stdlib/asm/mem.masm
Al-Kindi-0: I think I get the intention now, should probably be `mutually word-aligned` as defined in a further-down comment

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+        # use `memcopy_words` in the cycle to optimize the copy
+        # => [write_ptr%4, read_ptr, write_ptr, n]
+
+        ### "prefix" handling #####################################################################
+
+        # if the current read pointer (and write pointer, since now they are in "sync") is not
+        # word-aligned and we need to copy at least the (4 - write_ptr%4) values (so
```
Al-Kindi-0: nit: 
1- "... since inside this branch they are assumed to be mutually word-aligned..."
2- Would it make sense to test if n is large enough early in the procedure so that we can both simplify the logic here and stop early by defaulting to the element copying procedure?

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+        # n >= (4 - write_ptr%4), meaning we have enough values to copy to make pointers
+        # word-aligned), use `memcopy_felts_internal` to copy values until it become word-aligned
+        dup neq.0 # 3 cycles
+        # => [is_not_word_aligned, write_ptr%4, read_ptr, write_ptr, n]
+
+        push.4 movup.2 sub # 4 cycles
+        # => [4 - write_ptr%4, is_not_word_aligned, read_ptr, write_ptr, n] 
+        
+        dup dup.5 u32lte # 7 cycles
+        # => [(4 - write_ptr%4) <= 4, 4 - write_ptr%4, is_not_word_aligned, read_ptr, write_ptr, n]
```
Al-Kindi-0: `(4 - write_ptr%4) <= 4` -> `(4 - write_ptr%4) <= n` ?

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+        else
+            drop movup.2 # 2 cycles
+            # => [n, read_ptr, write_ptr]
+        end
+        # => [n', read_ptr', write_ptr']
+
+        # on that stage we either have word-aligned read and write pointers (if.true case or if 
+        # these pointers were word-aligned initially) or non-word-aligned pointers, but with `n`
+        # less than the number of values required to make pointers word-aligned (e.g. pointer is 
+        # 1001, but n = 2)
```
Al-Kindi-0: nit: I would complete the comment to explain what happens for each case in order to complete the picture.

libcore/tests/mem/mod.rs
```diff
@@ -5,6 +5,171 @@ use miden_utils_testing::{
     build_expected_perm, felt_slice_to_ints,
 };
 
+#[test]
```
Al-Kindi-0, line 8: Should we add some proptests?

Al-Kindi-0 (commented):
Looks good!
Left some comments and questions

stdlib/asm/mem.masm
```diff
@@ -2,6 +2,118 @@ use.std::crypto::hashes::rpo
...
+#! Total cycles:
+#! - if read and write pointers are mutually unaligned: 23 + 14 * num_felts
+#! - if read and write pointers are mutually aligned, it's equal to the sum of:
+#!   - prologue: 14 cycles
+#!   - prefix handling (worst case): 28 + 14 * num_felts_to_next_wa, where num_felts_to_next_wa is 
+#!     the number of felts required to make the read (and write) pointer word-aligned
+#!   - words handling: 15 + 16 * num_words
+#!   - suffix handling: 9 + 14 * num_remaining_felts
+#!     - in total: 66 + 14 * num_prefix_felts + 16 * num_words + 14 * num_suffix_felts
+export.memcopy
```
bobbinth: Let's use the new syntax here (i.e., `pub proc`). And we should merge the latest `next` into this to make sure all tests pass against it (i.e., `export` should no longe be allowed).

Also, I wonder if we can simplify this procedure to not have conditional logic. For example, we could try to compute 3 tuples: `(ptr1, n1)`, `(ptr2, n2)`, `(ptr3, n3)` and then call `memcopy_elements`, `memcopy_words`, and `memcopy_elements` on these unconditionally.

The idea here is that if we start with a word-aligned address, the first tuple would be `(read_ptr, 0)` and the second tuple would be `(read_ptr, n2)` etc.

If it is more efficient, the tuples could be `(start_ptr, end_ptr)` and in that case, for word-aligned addresses, the first tuple would be just `(read_ptr, read_ptr)`.

The main question here is if we can implement computing these tuples efficiently. We will probably need to have a helper procedure for computing "next multiple of 4" - but I think that could be done relatively efficiently.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 #### Enhancements
 
+- Added new `memcopy_felts` procedure for the `std::mem` module. ([#2352](https://github.com/0xMiden/miden-vm/pull/2352)).
```
bobbinth: Should this be `memcopy_elements` instead of `memcopy_felts`?

libcore/tests/mem/mod.rs
Fumuran: I tried to do so, but the main problem is that we can't assert the correctness of the value copied to some new location: in case of the proptest we can't come up with some formula which will represent the value here. But I'm new to the proptest dependency, so maybe I'm missing something.

bobbinth:
After looking at this more, I'm not sure if the extra complexity is worth it. So, I'll close this PR for now. If we do decide that we want it, we can re-open it later.

PR #1938, opened 2025-07-01
Disallow usage of the field modulus as an immediate value

Fumuran:
This small PR disallows the usage of the field modulus as an immediate value. 

This is done by changing the overflow check: now the value is considered invalid if it is greater or equal to the modulus (instead of being just greater).

Closes: #1929.

PR #2362, opened 2025-11-12
refactor: parallel trace generation follow-ups

### LINKED ISSUE #2162: Address all TODOs in parallel trace generation
As a follow-up to #1839.

huitseeker:
Small parallel processing follow-ups of #1938 #2188 and #2264.

This resolves #2162 except for the parallel trace column copy suggestion, which I am not sure about (not CPU bound).

### DIFF
--- a/processor/src/parallel/mod.rs
+++ b/processor/src/parallel/mod.rs
@@ -914,7 +914,7 @@ impl CoreTraceFragmentGenerator {
                 self.add_span_end_trace_row(basic_block_node)?;
             },
             NodeExecutionState::LoopRepeat(node_id) => {
-                finish_loop_node(self, node_id, &initial_mast_forest)?;
+                finish_loop_node(self, node_id, &initial_mast_forest, None)?;
             },
             // TODO(plafer): there's a big overlap between `NodeExecutionPhase::End` and
             // `Continuation::Finish*`. We can probably reconcile.
@@ -983,7 +983,7 @@ impl CoreTraceFragmentGenerator {
                     self.add_end_trace_row(mast_node.digest())?;
                 },
                 Continuation::FinishLoop(node_id) => {
-                    finish_loop_node(self, node_id, &current_forest)?;
+                    finish_loop_node(self, node_id, &current_forest, None)?;
                 },
                 Continuation::FinishCall(node_id) => {
                     let mast_node =
@@ -1134,24 +1134,7 @@ impl CoreTraceFragmentGenerator {
                     self.decrement_size(&mut NoopTracer);
                 }
 
-                while condition == ONE {
-                    self.add_loop_repeat_trace_row(
-                        loop_node,
-                        current_forest,
-                        self.context.state.decoder.current_addr,
-                    )?;
-
-                    self.execute_mast_node(loop_node.body(), current_forest)?;
-
-                    condition = self.get(0);
-                    self.decrement_size(&mut NoopTracer);
-                }
-
-                // 3. Add "end LOOP" row
-                //
-                // Note that we don't confirm that the condition is properly ZERO here, as the
-                // FastProcessor already ran that check.
-                self.add_end_trace_row(loop_node.digest())
+                finish_loop_node(self, node_id, current_forest, Some(condition))
             },
             MastNode::Call(call_node) => {
                 self.update_decoder_state_on_node_start();
@@ -1885,11 +1868,18 @@ fn finish_loop_node(
     processor: &mut CoreTraceFragmentGenerator,
     node_id: MastNodeId,
     mast_forest: &MastForest,
+    condition: Option<Felt>,
 ) -> ControlFlow<()> {
     let loop_node = mast_forest.get_node_by_id(node_id).expect("node should exist").unwrap_loop();
 
-    let mut condition = processor.get(0);
-    processor.decrement_size(&mut NoopTracer);
+    // If no condition is provided, read it from the stack
+    let mut condition = if let Some(cond) = condition {
+        cond
+    } else {
+        let cond = processor.get(0);
+        processor.decrement_size(&mut NoopTracer);
+        cond
+    };
 
     while condition == ONE {
         if let ControlFlow::Break(_) = processor.add_loop_repeat_trace_row(


processor/src/chiplets/mod.rs
```diff
@@ -332,12 +332,28 @@ impl Chiplets {
...
+            s.spawn(move |_| {
+                bitwise.fill_trace(&mut bitwise_fragment);
+            });
+            s.spawn(move |_| {
+                memory.fill_trace(&mut memory_fragment);
+            });
+            s.spawn(move |_| {
+                kernel_rom.fill_trace(&mut kernel_rom_fragment);
+            });
+        });
```
plafer, line 354: FYI this is the old processor that we're about to remove

processor/src/operations/crypto_ops.rs
```diff
@@ -491,6 +491,55 @@ mod tests {
         assert!(process.advice.has_merkle_root(expected_root));
     }
 
+    #[test]
+    fn op_mrupdate_invalid_path_length() {
```
plafer: This is under the old processor; can you move it under the new processor instead (`processor/src/procesor/operations/crypto_ops.rs`)?

processor/src/parallel/mod.rs
```diff
@@ -778,15 +775,24 @@ struct CoreTraceFragment {
...
-            // TODO(plafer): Don't use uninit_vector
-            columns: core::array::from_fn(|_| unsafe { uninit_vector(num_rows) }),
+            columns: core::array::from_fn(|_| {
+                // Create a Vec with uninitialized elements using MaybeUninit
+                let mut vec: Vec<core::mem::MaybeUninit<Felt>> = Vec::with_capacity(num_rows);
+                unsafe { vec.set_len(num_rows) };
+
+                // Safety: The caller (CoreTraceFragmentGenerator) guarantees all elements
+                // are initialized before first read
+                unsafe { core::mem::transmute(vec) }
```
plafer: This is undefined behavior, as discussed in https://github.com/facebook/winterfell/issues/396 - with resolution in [here](https://github.com/facebook/winterfell/issues/396#issuecomment-3146571754) (there's also a very long rabbit hole-y slack thread about it). And we would want `CoreTraceFragment.columns: [Vec<MaybeUninit<Felt>>; CORE_TRACE_WIDTH]`, which will make us rewrite every line that writes to that field.

So let's hold off on that change for now, since it will create a big merge conflict with #2345.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - [BREAKING] Make argument order of `Assembler::compile_and_statically_link_from_dir` consistent with `Assembler::assemble_library_from_dir`
 - [BREAKING] Renamed `Library::get_procedure_root_by_name` to `Library::get_procedure_root_by_path`
 - Streamline MastNode APIs and remove redundant parameters from `execute_op_batch` functions ([#2360](https://github.com/0xMiden/miden-vm/pull/2360)).
+- closed TODOs from parallel trace generation ([#2362](https://github.com/0xMiden/miden-vm/pull/2362).
```
plafer: There are still 13 `TODO(plafer)` left after this - so I would probably just merge this one without a changelog entry

processor/src/processor/mod.rs
```diff
@@ -55,17 +55,43 @@ pub trait Processor: Sized {
...
+    /// 2. the number of quadratic extension field elements read in the `READ` section,
+    /// 3. the number of field elements, one base field element per gate, in the `EVAL` section,
+    ///
+    /// Stack transition:
+    /// [ptr, num_read, num_eval, ...] -> [ptr, num_read, num_eval, ...]
+    ///
+    /// # Note
+    /// This is a default implementation that should be overridden by processors that have
+    /// specific circuit evaluation capabilities. The default implementation returns an error
+    /// since not all processors support circuit evaluation.
```
plafer: Is this true? AFAIU, all processors will need to support this

processor/src/parallel/mod.rs
```diff
@@ -1961,9 +1967,6 @@ impl OperationHelperRegisters for TraceGenerationHelpers {
 // HELPERS
 // ================================================================================================
 
-// TODO(plafer): If we want to keep this strategy, then move the `op_eval_circuit()` method
-// implementation to the `Processor` trait, and have `FastProcessor` and
-// `CoreTraceFragmentGenerator` both use it.
```
plafer, line 1966: The issue that this was pointing to is that `eval_circuit_fast_()` was copy/pasted here (i.e. search for `eval_circuit_fast_()`, and you'll find 2 similar but distinct implementations). I wanted to look into how to properly abstract it away.

Though if too complex, it might be enough to just rename this one - since it's still just the copy/pasted name from the `fast` module.

processor/src/chiplets/mod.rs
huitseeker: Aren't both the old *and* the fast processor calling into `chiplets.into_trace()`?

processor/src/parallel/mod.rs
huitseeker: Addressed in c243e2653ab76a8144aa365a16356bcdd7b77bcb

processor/src/chiplets/mod.rs
plafer: Ah yes good point!

processor/src/parallel/mod.rs
```diff
@@ -1961,12 +1876,51 @@ impl OperationHelperRegisters for TraceGenerationHelpers {
...
-/// Identical to `[chiplets::ace::eval_circuit]` but adapted for use with `[FastProcessor]`.
+/// Executes a loop node by processing its body repeatedly while the condition is true.
+///
+/// This function is shared between `NodeExecutionState::LoopRepeat` and `Continuation::FinishLoop`
+/// to eliminate code duplication. Both cases have identical logic for handling loop execution.
+#[inline(always)]
+fn execute_loop_node(
```
plafer: We should rename this to indicate that this corresponds only to the "finish" phase of the loop execution - maybe `finish_loop_node()` or something.

We can then reuse it as well in `execute_mast_node()` in the `Loop` case (specifically lines 1137-1154)

processor/src/parallel/mod.rs
```diff
@@ -1961,12 +1876,51 @@ impl OperationHelperRegisters for TraceGenerationHelpers {
...
+            loop_node,
+            mast_forest,
+            processor.context.state.decoder.current_addr,
+        ) {
+            return ControlFlow::Break(());
+        }
+
+        if let ControlFlow::Break(_) = processor.execute_mast_node(loop_node.body(), mast_forest) {
+            return ControlFlow::Break(());
+        }
```
plafer, line 1895: Both of these `if let` can be replaced by the `?` operator

processor/src/processor/mod.rs
plafer: We should also remove the default `Err` implementation, since every processor needs to implement this. I'd rather know at compile-time that I'm missing a method implementation than at runtime

processor/src/parallel/mod.rs
huitseeker: `ControlFlow` is not a `Try` type, right?

processor/src/parallel/mod.rs
huitseeker: Though I have done the function renaming, I have not found that this is a straightforward replacement (variable `condition` spans both execution and finishing scopes). Is it worth making `finish_loop_node` take it as argument for generality IYO?

huitseeker: Addressed 51790ac

processor/src/parallel/mod.rs
plafer: Yes [it is](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html#impl-Try-for-ControlFlow%3CB,+C%3E); notice that we use the `?` operator everywhere else in the parallel trace generators, which use `ControlFlow` as a return type.

plafer: I need changes in this PR for #2345, so I'll merge the and fix this there

plafer: Ah, cannot merge due to some commits not being signed

huitseeker: Should be fixed!

PR #2317, opened 2025-10-28
Refactor/remove unused deps and clones

Himess:
### Summary
This PR removes unused dependencies and eliminates redundant clone operations, improving build times and runtime performance.

#### Unused Dependencies Removed
**processor:**
- `logtest` (dev-dependency) - No longer used in tests

#### Redundant Clones Removed (48 total)
Removed unnecessary `.clone()` calls where values are dropped without further use:
- **core**: 22 clones (merger/tests.rs, basic_block_node/tests.rs)
- **assembly**: 4 clones (assembler.rs, mast_forest_builder.rs)
- **stdlib tests**: 9 clones (falcon.rs, fri tests)
- **miden-vm tests**: 6 clones (hasher.rs, exec.rs, crypto_ops.rs)
- **processor**: 6 clones (fast/tests, operations, trace/tests)
- **debug-types**: 1 clone (source_manager.rs)

image

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](../CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #2366, opened 2025-11-13
feat: consolidate decorator information into a single DebugInfo

### LINKED ISSUE #1776: Investigate removing `Decorator` in favor of a new `DebugInfo`

huitseeker:
This PR witnesses that:
- we can make sure that `DebugInfo` is always present in-memory in the `MastForest`, the 1+3+2+2 concerned vectors will just be empty in this case, which obviates the `Option` mentioned in #1776 and greatly simplifies runtime behavior,
- we will just avoid serializing the decorator information in that case (serialization refactor is the subject of my very next PR, and tracked in #1580),
- this allows having the assembler always include decorator information as soon as either of tracing or debug mode are on, along the lines of #1821.

We thus just have to organize the refactor of decorators into a single struct, which this PR does. Fixes #1776.

There's one related TODO item left : separating the `AssemblyOp`s from the usual decorator information, and refactoring the `MastNodeErrorContext` trait, and corresponding `ErrorContextImpl` struct, as asked for by @plafer (and incidentally mentioned by @bobbinth in [this comment of #2323](https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2508446259)). I'd rather have that be the topic of a follow-up issue.

### DIFF
--- a/core/src/mast/debuginfo/mod.rs
+++ b/core/src/mast/debuginfo/mod.rs
@@ -166,7 +166,7 @@ impl DebugInfo {
     }
 
     /// Returns decorator links for a node, including operation indices.
-    pub fn decorator_links_for_node(
+    pub(super) fn decorator_links_for_node(
         &self,
         node_id: MastNodeId,
     ) -> Result<DecoratedLinks<'_>, DecoratorIndexError> {
@@ -195,7 +195,7 @@ impl DebugInfo {
     /// # Note
     /// This method does not validate decorator IDs immediately. Validation occurs during
     /// operations that need to access the actual decorator data (e.g., merging, serialization).
-    pub(super) fn add_node_decorators(
+    pub(super) fn register_node_decorators(
         &mut self,
         node_id: MastNodeId,
         before_enter: &[DecoratorId],
@@ -208,7 +208,7 @@ impl DebugInfo {
     /// Registers operation-indexed decorators for a node.
     ///
     /// This associates already-added decorators with specific operations within a node.
-    pub fn register_op_indexed_decorators(
+    pub(crate) fn register_op_indexed_decorators(
         &mut self,
         node_id: MastNodeId,
         decorators_info: Vec<(usize, DecoratorId)>,
@@ -227,11 +227,6 @@ impl DebugInfo {
     // ERROR CODE METHODS
     // --------------------------------------------------------------------------------------------
 
-    /// Inserts an error code with its message.
-    pub fn insert_error_code(&mut self, code: u64, msg: Arc<str>) {
-        self.error_codes.insert(code, msg);
-    }
-
     /// Returns an error message by code.
     pub fn error_message(&self, code: u64) -> Option<Arc<str>> {
         self.error_codes.get(&code).cloned()
@@ -242,11 +237,9 @@ impl DebugInfo {
         self.error_codes.iter()
     }
 
-    /// Clears all error codes.
-    ///
-    /// This is used when error code information needs to be reset.
-    pub fn clear_error_codes(&mut self) {
-        self.error_codes.clear();
+    /// Inserts an error code with its message.
+    pub fn insert_error_code(&mut self, code: u64, msg: Arc<str>) {
+        self.error_codes.insert(code, msg);
     }
 
     /// Inserts multiple error codes at once.
@@ -257,6 +250,13 @@ impl DebugInfo {
         I: IntoIterator<Item = (u64, Arc<str>)>,
     {
         self.error_codes.extend(error_codes);
+    }
+
+    /// Clears all error codes.
+    ///
+    /// This is used when error code information needs to be reset.
+    pub fn clear_error_codes(&mut self) {
+        self.error_codes.clear();
     }
 
     // TEST HELPERS
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -93,11 +93,6 @@ impl MastForest {
     /// The maximum number of nodes that can be stored in a single MAST forest.
     const MAX_NODES: usize = (1 << 30) - 1;
 
-    /// Adds a decorator to the forest, and returns the associated [`DecoratorId`].
-    pub fn add_decorator(&mut self, decorator: Decorator) -> Result<DecoratorId, MastForestError> {
-        self.debug_info.add_decorator(decorator)
-    }
-
     /// Marks the given [`MastNodeId`] as being the root of a procedure.
     ///
     /// If the specified node is already marked as a root, this will have no effect.
@@ -199,6 +194,7 @@ impl MastForest {
     }
 }
 
+// ------------------------------------------------------------------------------------------------
 /// Helpers
 impl MastForest {
     /// Adds all provided nodes to the internal set of nodes, remapping all [`MastNodeId`]
@@ -271,18 +267,8 @@ fn remove_nodes(
 }
 
 // ------------------------------------------------------------------------------------------------
-
 /// Public accessors
 impl MastForest {
-    /// Returns the [`Decorator`] associated with the provided [`DecoratorId`] if valid, or else
-    /// `None`.
-    ///
-    /// This is the fallible version of indexing (e.g. `mast_forest[decorator_id]`).
-    #[inline]
-    pub fn decorator_by_id(&self, decorator_id: DecoratorId) -> Option<&Decorator> {
-        self.debug_info.decorator(decorator_id)
-    }
-
     /// Returns the [`MastNode`] associated with the provided [`MastNodeId`] if valid, or else
     /// `None`.
     ///
@@ -354,8 +340,30 @@ impl MastForest {
         self.nodes.as_slice()
     }
 
+    pub fn advice_map(&self) -> &AdviceMap {
+        &self.advice_map
+    }
+
+    pub fn advice_map_mut(&mut self) -> &mut AdviceMap {
+        &mut self.advice_map
+    }
+}
+
+// ------------------------------------------------------------------------------------------------
+/// Decorator methods
+impl MastForest {
+    /// Returns a list of all decorators contained in this [MastForest].
     pub fn decorators(&self) -> &[Decorator] {
         self.debug_info.decorators()
+    }
+
+    /// Returns the [`Decorator`] associated with the provided [`DecoratorId`] if valid, or else
+    /// `None`.
+    ///
+    /// This is the fallible version of indexing (e.g. `mast_forest[decorator_id]`).
+    #[inline]
+    pub fn decorator_by_id(&self, decorator_id: DecoratorId) -> Option<&Decorator> {
+        self.debug_info.decorator(decorator_id)
     }
 
     /// Returns decorator indices for a specific operation within a node.
@@ -363,7 +371,7 @@ impl MastForest {
     /// This is the primary accessor for reading decorators from the centralized storage.
     /// Returns a slice of decorator IDs for the given operation.
     #[inline]
-    pub fn decorator_indices_for_op(
+    pub(crate) fn decorator_indices_for_op(
         &self,
         node_id: MastNodeId,
         local_op_idx: usize,
@@ -386,17 +394,6 @@ impl MastForest {
             .map(move |&decorator_id| &self[decorator_id])
     }
 
-    /// Returns decorator links for a node, including operation indices.
-    ///
-    /// This provides a flattened view of all decorators for a node with their operation indices.
-    #[inline]
-    pub fn decorator_links_for_node<'a>(
-        &'a self,
-        node_id: MastNodeId,
-    ) -> Result<DecoratedLinks<'a>, DecoratorIndexError> {
-        self.debug_info.decorator_links_for_node(node_id)
-    }
-
     /// Returns the decorators to be executed before this node is executed.
     #[inline]
     pub fn before_enter_decorators(&self, node_id: MastNodeId) -> &[DecoratorId] {
@@ -409,24 +406,47 @@ impl MastForest {
         self.debug_info.after_exit_decorators(node_id)
     }
 
-    /// Adds decorator IDs for a node to the storage.
-    /// Used when building nodes for efficient decorator access during execution.
+    /// Returns decorator links for a node, including operation indices.
+    ///
+    /// This provides a flattened view of all decorators for a node with their operation indices.
     #[inline]
-    pub(crate) fn add_node_decorators(
+    pub(crate) fn decorator_links_for_node<'a>(
+        &'a self,
+        node_id: MastNodeId,
+    ) -> Result<DecoratedLinks<'a>, DecoratorIndexError> {
+        self.debug_info.decorator_links_for_node(node_id)
+    }
+
+    /// Adds a decorator to the forest, and returns the associated [`DecoratorId`].
+    pub fn add_decorator(&mut self, decorator: Decorator) -> Result<DecoratorId, MastForestError> {
+        self.debug_info.add_decorator(decorator)
+    }
+
+    /// Adds decorator IDs for a node to the storage.
+    ///
+    /// Used when building nodes for efficient decorator access during execution.
+    ///
+    /// # Note
+    /// This method does not validate decorator IDs immediately. Validation occurs during
+    /// operations that need to access the actual decorator data (e.g., merging, serialization).
+    #[inline]
+    pub(crate) fn register_node_decorators(
         &mut self,
         node_id: MastNodeId,
         before_enter: &[DecoratorId],
         after_exit: &[DecoratorId],
     ) {
-        self.debug_info.add_node_decorators(node_id, before_enter, after_exit);
+        self.debug_info.register_node_decorators(node_id, before_enter, after_exit);
     }
+}
 
-    pub fn advice_map(&self) -> &AdviceMap {
-        &self.advice_map
-    }
-
-    pub fn advice_map_mut(&mut self) -> &mut AdviceMap {
-        &mut self.advice_map
+// ------------------------------------------------------------------------------------------------
+/// Error message methods
+impl MastForest {
+    /// Given an error code as a Felt, resolves it to its corresponding error message.
+    pub fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>> {
+        let key = u64::from(code);
+        self.debug_info.error_message(key)
     }
 
     /// Registers an error message in the MAST Forest and returns the corresponding error code as a
@@ -437,13 +457,10 @@ impl MastForest {
         self.debug_info.insert_error_code(code.as_int(), msg);
         code
     }
-
-    /// Given an error code as a Felt, resolves it to its corresponding error message.
-    pub fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>> {
-        let key = u64::from(code);
-        self.debug_info.error_message(key)
-    }
 }
+
+// MAST FOREST INDEXING
+// ------------------------------------------------------------------------------------------------
 
 impl Index<MastNodeId> for MastForest {
     type Output = MastNode;
--- a/core/src/mast/node/split_node.rs
+++ b/core/src/mast/node/split_node.rs
@@ -346,7 +346,7 @@ impl MastForestContributor for SplitNodeBuilder {
         let future_node_id = MastNodeId::new_unchecked(forest.nodes.len() as u32);
 
         // Store node-level decorators in the centralized NodeToDecoratorIds for efficient access
-        forest.add_node_decorators(future_node_id, &before_enter, &after_exit);
+        forest.register_node_decorators(future_node_id, &before_enter, &after_exit);
 
         // Create the node in the forest with Linked variant from the start
         // Move the data directly without intermediate cloning
@@ -460,7 +460,7 @@ impl SplitNodeBuilder {
         let future_node_id = MastNodeId::new_unchecked(forest.nodes.len() as u32);
 
         // Store node-level decorators in the centralized NodeToDecoratorIds for efficient access
-        forest.add_node_decorators(future_node_id, &self.before_enter, &self.after_exit);
+        forest.register_node_decorators(future_node_id, &self.before_enter, &self.after_exit);
 
         // Create the node in the forest with Linked variant from the start
         // Move the data directly without intermediate cloning
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -1319,7 +1319,7 @@ impl MastForestContributor for BasicBlockNodeBuilder {
         // Add node-level decorators to the centralized NodeToDecoratorIds for efficient access
         forest
             .debug_info
-            .add_node_decorators(future_node_id, &before_enter, &after_exit);
+            .register_node_decorators(future_node_id, &before_enter, &after_exit);
 
         // Create the node in the forest with Linked variant from the start
         // Move the data directly without intermediate cloning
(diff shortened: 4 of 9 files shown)


core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,130 @@
...
+
+    /// Efficient access to decorators per operation per node.
+    pub op_decorator_storage: OpToDecoratorIds,
+
+    /// Efficient storage for node-level decorators (before_enter and after_exit).
+    pub node_decorator_storage: NodeToDecoratorIds,
+
+    /// Maps error codes to error messages.
+    pub error_codes: BTreeMap<u64, Arc<str>>,
+}
```
bobbinth, line 77: Would it not be better to make the fields private to hide the internal implementation details of this struct? We could just expose the relevant methods from this struct (e.g., `decorators_for_op()`, `get_decorator_by_id()` etc.).

core/src/mast/node/split_node.rs
```diff
@@ -218,8 +218,10 @@ impl MastNodeExt for SplitNode {
             },
             DecoratorStore::Linked { id } => {
                 // Extract decorators from forest storage when in Linked state
-                let before_enter = forest.node_decorator_storage.get_before_decorators(id).to_vec();
-                let after_exit = forest.node_decorator_storage.get_after_decorators(id).to_vec();
+                let before_enter =
+                    forest.debug_info.node_decorator_storage.get_before_decorators(id).to_vec();
```
bobbinth: Related to the previous comment: ideally, this would be something like `forest.get_before_enter_decorators(id).to_vec()`.

core/src/mast/debuginfo/mod.rs
huitseeker: Yep, cleaned that up and made getters follow [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html?highlight=C-Getter#getter-names-follow-rust-convention-c-getter)

processor/src/fast/call_and_dyn.rs
```diff
@@ -101,6 +101,7 @@ impl FastProcessor {
             current_forest,
         );
 
+        #[cfg(not(feature = "no_err_ctx"))]
         let call_node = current_forest[node_id].unwrap_call();
         let err_ctx = err_ctx!(current_forest, call_node, host);
```
bobbinth: This declares `call_node` variable only if `no_err_ctx` feature is off, right? Would that not cause an issue for line 106? If not, maybe worth adding a small comment as to why.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+    pub fn decorator_mut(&mut self, decorator_id: DecoratorId) -> Option<&mut Decorator> {
+        if decorator_id.to_usize() < self.decorators.len() {
+            Some(&mut self.decorators[decorator_id])
+        } else {
+            None
+        }
+    }
+
+    /// Returns the before-enter decorators for the given node.
+    pub fn before_enter_decorators(&self, node_id: crate::mast::MastNodeId) -> &[DecoratorId] {
```
bobbinth: nit: why not import `MastNodeId` at the top of the file?

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+
+    /// Returns the number of decorators.
+    pub fn num_decorators(&self) -> usize {
+        self.decorators.len()
+    }
+
+    /// Returns the operation decorator storage.
+    pub fn op_decorator_storage(&self) -> &OpToDecoratorIds {
+        &self.op_decorator_storage
+    }
```
bobbinth: This is used only for testing purposes, right? If so, maybe worth gating this behind the `test` attribute.

core/src/mast/mod.rs
```diff
@@ -635,7 +647,7 @@ impl DecoratorId {
         value: u32,
         mast_forest: &MastForest,
     ) -> Result<Self, DeserializationError> {
-        Self::from_u32_bounded(value, mast_forest.decorators.len())
+        Self::from_u32_bounded(value, mast_forest.debug_info.decorators().len())
```
bobbinth: Could this not be:
```rust
Self::from_u32_bounded(value, mast_forest.debug_info.num_decorators())
```

core/src/mast/mod.rs
```diff
@@ -370,48 +357,73 @@ impl MastForest {
...
+        node_id: MastNodeId,
+        before_enter: &[DecoratorId],
+        after_exit: &[DecoratorId],
+    ) {
+        self.debug_info.node_decorator_storage_mut().add_node_decorators(
+            node_id,
+            before_enter,
+            after_exit,
+        );
     }
```
bobbinth: We use this function only in `SplitNode::add_to_forest()` in other places (e.g., `JoinNode::add_to_forest()` we access `debug_info` directly. I think we should make these things consistent:

- We either get rid of this function, or:
- Start using this in all places instead of accessing `debug_info` directly.

Also, the implementation of this function could be just:
```rust
self.debug_info.add_node_decorators(node_id, before_enter, after_exit);
```

core/src/mast/mod.rs
```diff
@@ -231,8 +218,8 @@ impl MastForest {
         // extract decorator information from the nodes by converting them into builders
         let node_builders =
             nodes_to_add.into_iter().map(|node| node.to_builder(self)).collect::<Vec<_>>();
-        self.op_decorator_storage = OpToDecoratorIds::new();
-        self.node_decorator_storage = NodeToDecoratorIds::new();
+        *self.debug_info.op_decorator_storage_mut() = OpToDecoratorIds::new();
+        self.debug_info.clear_node_decorator_storage();
```
bobbinth: Why do we set storage to an empty map and then immediately clear it?

core/src/mast/mod.rs
```diff
@@ -109,7 +95,10 @@ impl MastForest {
 
     /// Adds a decorator to the forest, and returns the associated [`DecoratorId`].
     pub fn add_decorator(&mut self, decorator: Decorator) -> Result<DecoratorId, MastForestError> {
-        self.decorators.push(decorator).map_err(|_| MastForestError::TooManyDecorators)
+        self.debug_info
+            .decorators_mut()
+            .push(decorator)
+            .map_err(|_| MastForestError::TooManyDecorators)
```
bobbinth: I would probably add `DebugInfo::add_decorator()` function to make adding decorators an internal concern of the `DebugInfo`.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+
+    /// Returns the node decorator storage.
+    pub fn node_decorator_storage(&self) -> &NodeToDecoratorIds {
+        &self.node_decorator_storage
+    }
+
+    /// Returns the node decorator storage mutably.
+    pub fn node_decorator_storage_mut(&mut self) -> &mut NodeToDecoratorIds {
+        &mut self.node_decorator_storage
+    }
```
bobbinth: Do we need to have these functions on the public interface? Maybe we could replace them with higher-level functions to hide the internal implementation details.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+        before_enter: &[crate::mast::DecoratorId],
+        after_exit: &[crate::mast::DecoratorId],
+    ) {
+        self.node_decorator_storage
+            .add_node_decorators(node_id, before_enter, after_exit);
+    }
+
+    /// Strips all debug information, removing decorators and error codes.
+    /// This is used for release builds where debug info is not needed.
+    pub fn strip(&mut self) {
```
bobbinth: I think we can probably name this function `clear()`.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+        &self,
+        node_id: crate::mast::MastNodeId,
+    ) -> Result<DecoratedLinks<'_>, DecoratorIndexError> {
+        self.op_decorator_storage.decorator_links_for_node(node_id)
+    }
+
+    /// Clears the node decorator storage.
+    pub fn clear_node_decorator_storage(&mut self) {
+        self.node_decorator_storage.clear();
+    }
```
bobbinth: Do we need this function?

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+
+    /// Returns an iterator over error codes.
+    pub fn error_codes(&self) -> impl Iterator<Item = (&u64, &Arc<str>)> {
+        self.error_codes.iter()
+    }
+
+    /// Clears all error codes.
+    pub fn clear_error_codes(&mut self) {
+        self.error_codes.clear();
+    }
```
bobbinth: Do we need this function? Seems like it is not used (and I'm not sure there is a use case for clearing error codes but not decorators).

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,249 @@
...
+    op_decorator_storage: OpToDecoratorIds,
+
+    /// Efficient storage for node-level decorators (before_enter and after_exit).
+    node_decorator_storage: NodeToDecoratorIds,
+
+    /// Maps error codes to error messages.
+    error_codes: BTreeMap<u64, Arc<str>>,
+}
+
+impl DebugInfo {
```
bobbinth, line 79: Let's organize this `impl` into sections to improve readability. The sections could be constructors, public accessors, decorator methods, error code methods.

core/src/mast/merger/mod.rs
```diff
@@ -164,9 +164,12 @@ impl MastForestMerger {
     }
 
     fn merge_decorators(&mut self, other_forest: &MastForest) -> Result<(), MastForestError> {
-        let mut decorator_id_remapping = DenseIdMap::with_len(other_forest.decorators.len());
+        let mut decorator_id_remapping =
+            DenseIdMap::with_len(other_forest.debug_info.decorators().len());
```
bobbinth: We could use `debug_info.num_decorators()` here instead of `debug_info.decorators.len()` here, right?

core/src/mast/serialization/mod.rs
```diff
@@ -117,7 +117,7 @@ impl Serializable for MastForest {
 
         // decorator & node counts
         target.write_usize(self.nodes.len());
-        target.write_usize(self.decorators.len());
+        target.write_usize(self.debug_info.decorators().len());
```
bobbinth: Similar to other comments - should we use `DebugInfo::num_decorators()` here?

processor/src/fast/basic_block.rs
```diff
@@ -113,11 +113,7 @@ impl FastProcessor {
         // last clock cycle of the BASIC BLOCK ops.
         // For the linked case, check for decorators at an operation index beyond the last operation
         let num_ops = basic_block_node.num_operations() as usize;
-        let decorator_ids = current_forest.decorator_indices_for_op(node_id, num_ops);
-        for &decorator_id in decorator_ids {
-            let decorator = program
-                .get_decorator_by_id(decorator_id)
-                .ok_or(ExecutionError::DecoratorNotFoundInForest { decorator_id })?;
+        for decorator in current_forest.decorators_for_op(node_id, num_ops) {
```
bobbinth, line 116: Here, and in other similar places, we do not yet check whether we are executing in the debug mode or not, right?

processor/src/fast/call_and_dyn.rs
huitseeker: https://github.com/0xMiden/miden-vm/pull/2384

core/src/mast/debuginfo/mod.rs
huitseeker: Do we want to have a clippy config that checks for those details (in this case, the [absolute_paths](https://rust-lang.github.io/rust-clippy/master/index.html?groups=cargo%2Ccomplexity%2Ccorrectness%2Cnursery%2Cpedantic%2Cperf%2Crestriction%2Cstyle%2Csuspicious%2Cdeprecated&levels=allow#absolute_paths) lint), as laid out [here](https://github.com/0xMiden/miden-vm/pull/2317#pullrequestreview-3391325560)?

core/src/mast/mod.rs
huitseeker: Those are different maps. I have a dedicated functions which should clarify in the latest commits.

core/src/mast/debuginfo/mod.rs
huitseeker: Made test-only.

core/src/mast/mod.rs
huitseeker: Used everywhere

core/src/mast/debuginfo/mod.rs
huitseeker: This is mainly used from the `MastForest:strip_decorators()` for the purpose of decorator-less serialization (until we revamp serialization in an upcoming PR). See comment at the top of this module. The name `strip` makes sense to me.

processor/src/fast/basic_block.rs
huitseeker: The assembler discriminates, but for the processors it's another story:
- the old processor respects `ExecutionOption`s,
- the fast processor just has an `in_debug_mode` field (and two constructors, `new_with_advice_inputs` and `new_debug` that hardcode to one alternative or another)

The fast processor's discrimination is in `execute_decorator` and amounts to:
  - Debug decorators: Execute only when in_debug_mode = true (via new_debug())
  - AsmOp decorators: Always ignored - no execution logic 
  - Trace decorators: Always execute - no debug mode check

@plafer and I should striaghten that up in an upcoming PR.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+        self.decorators.is_empty() && self.error_codes.is_empty()
+    }
+
+    /// Strips all debug information, removing decorators and error codes.
+    /// This is used for release builds where debug info is not needed.
+    pub fn strip(&mut self) {
+        self.clear_decorators();
+        self.decorators = IndexVec::new();
+        self.error_codes.clear();
+    }
```
bobbinth, line 127: I still think `clear()` here is a more appropriate name. The "strip" name makes sense in the context of `MastForest` because there we are removing debug info from the forest. But here, we are removing all data from `DebugInfo` - which I think more commonly is called "clear" (or something similar).

Also, why doesn't `clear_decorators()` clear the `self.decorators` field? I would have thought that this would be just:
```rust
self.clear_decorators();
self.error_codes.clear();
```
And `self.decorators = IndexVec::new();` would be done inside `clear_decorators()`.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+            .unwrap_or(&[])
+    }
+
+    /// Returns decorator links for a node, including operation indices.
+    pub fn decorator_links_for_node(
+        &self,
+        node_id: MastNodeId,
+    ) -> Result<DecoratedLinks<'_>, DecoratorIndexError> {
+        self.op_decorator_storage.decorator_links_for_node(node_id)
+    }
```
bobbinth, line 174: Not for this PR, but eventually, I'd love to get rid of this method.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+        node_id: MastNodeId,
+        before_enter: &[DecoratorId],
+        after_exit: &[DecoratorId],
+    ) {
+        self.node_decorator_storage
+            .add_node_decorators(node_id, before_enter, after_exit);
+    }
+
+    /// Adds decorator info for operations within a node.
+    pub fn add_decorator_info_for_node(
```
bobbinth: nit: I think this is used specifically for basic blocks - so, I'd probably rename this to something like `add_basic_block_decorators()`.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+    }
+
+    /// Returns the decorator with the given ID, if it exists (mutable).
+    pub fn decorator_mut(&mut self, decorator_id: DecoratorId) -> Option<&mut Decorator> {
+        if decorator_id.to_usize() < self.decorators.len() {
+            Some(&mut self.decorators[decorator_id])
+        } else {
+            None
+        }
+    }
```
bobbinth: Not for this PR, but eventually, I'd love to get rid of this method so that decorators cannot be mutated after they are added to the forest.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+
+    /// Clears all decorator information while preserving error codes.
+    /// This is used when rebuilding decorator information from nodes.
+    pub fn clear_decorators(&mut self) {
+        self.op_decorator_storage = OpToDecoratorIds::new();
+        self.node_decorator_storage.clear();
+    }
```
bobbinth: As mentioned in one of the above comments: do we not want to clear `self.decorators` here as well?

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+
+    /// Returns an iterator over error codes.
+    pub fn error_codes(&self) -> impl Iterator<Item = (&u64, &Arc<str>)> {
+        self.error_codes.iter()
+    }
+
+    /// Returns a mutable reference to the error codes map.
+    pub fn error_codes_mut(&mut self) -> &mut BTreeMap<u64, Arc<str>> {
+        &mut self.error_codes
+    }
```
bobbinth: Rather exposing this field as mutable, maybe it would be better to bring back `clear_error_codes()`. Then, we could use it in combination with `insert_error_code()` to fulfill the current set of use cases.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - mostly very minor.

Also, I pushed a couple of commits - mostly with code formatting, but also removed one function that wasn't being used.

core/src/mast/debuginfo/mod.rs
bobbinth: Actually, thinking about this a bit more, maybe this should be `register_basic_block_decorators()` because we are not really adding decorators here, but registering already added decorators with a specific basic block.

Similar logic applies to `add_node_decorators()`: maybe it should be `register_node_decorators()`.

core/src/mast/debuginfo/mod.rs
```diff
@@ -1,7 +1,258 @@
...
+    /// Adds node-level decorators (before_enter and after_exit) for the given node.
+    pub fn add_node_decorators(
+        &mut self,
+        node_id: MastNodeId,
+        before_enter: &[DecoratorId],
+        after_exit: &[DecoratorId],
+    ) {
+        self.node_decorator_storage
+            .add_node_decorators(node_id, before_enter, after_exit);
+    }
```
bobbinth: Question: what happens here if any of the provided decorator IDs can't be mapped to a decorator (i.e., an invalid ID is provided)? Should we return an error from here?

core/src/mast/debuginfo/mod.rs
huitseeker: > Also, why doesn't clear_decorators() clear the self.decorators field? 

The method preserves the `decorators` field intentionally because it contains the actual decorator data needed when rebuilding the centralized storage mappings after node restructuring when merging `MastForest`s.

The method is used in `MastForest::remap_and_add_nodes()` with this specific pattern:
```rust
// Extract builders from existing nodes (this moves decorator data into builders)
let nodes_to_add = nodes_to_add.into_iter().map(|node| node.to_builder(self)).collect::<Vec<_>>();

// Clear decorator storage after extracting builders (builders contain decorator data)
self.debug_info.clear();
```

Overall:
1. When `node.to_builder(self)` is called, decorator data is moved from the centralized `DebugInfo` storage into individual node builders. The `decorators` field (IndexVec) still contains the actual decorator data that's being referenced.
2. After clearing the storage structures (`op_decorator_storage` and `node_decorator_storage`), the system rebuilds the decorator mappings by calling `forest.add_node_decorators()` for each node builder. The decorators field provides the data indexed by the rebuilt decorator indexes.
3. The decorators field contains the actual decorator objects that would need to be reconstructed anyway. Clearing it would require reallocation and copying of the same data during the rebuild process.

I chose to rename `clear_decorators()` to `clear_mappings()` and `strip()` to `clear()`.

core/src/mast/debuginfo/mod.rs
huitseeker: We do not, as explained [here](https://github.com/0xMiden/miden-vm/pull/2366#discussion_r2570079739)

core/src/mast/debuginfo/mod.rs
huitseeker: Done

core/src/mast/debuginfo/mod.rs
huitseeker: Made this pub(super) to indicate it's tailor-made to support internal use cases where decorator existence is guaranteed (aka leaving the responsibility to the caller). We could also do runtime checks (costly) or use a [`GhostCell`](https://plv.mpi-sws.org/rustbelt/ghostcell/). Both are rather heavy-duty for this internal method.

core/src/mast/debuginfo/mod.rs
huitseeker: renamed.

core/src/mast/debuginfo/mod.rs
huitseeker: So would I. Made `pub(super)`.

core/src/mast/debuginfo/mod.rs
huitseeker: What's the rationale? Incidentally this is a pure getter.

bobbinth: I meant this for a longer-term goal as we get rid of the "own" vs. "linked" decorator concepts. Though, maybe even sooner than that we can get rid of the `BasicBlock::indexed_decorator_iter()` which I think is the primary user of this method.

bobbinth:
> There's one related TODO item left : separating the `AssemblyOp`s from the usual decorator information, and refactoring the `MastNodeErrorContext` trait, and corresponding `ErrorContextImpl` struct, as asked for by @plafer (and incidentally mentioned by @bobbinth in [this comment of #2323](https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2508446259)). I'd rather have that be the topic of a follow-up issue.

Agreed. I'm going to merge this as is, and let's address the above in follow-ups.

PR #2367, opened 2025-11-13
refactor(processor): Host debug and trace handlers return dynamic errors

adr1anh:
Change host trait methods to return specific error types instead of ExecutionError:
- `on_debug()` returns `Result<(), DebugError>`
- `on_trace()` returns `Result<(), TraceError>`
- `on_assert_failed()` returns `Option<AssertError>`

This allows host implementations to define custom error types while preserving error context through new ExecutionError variants (DebugHandlerError, TraceHandlerError) that capture clock cycle information.

The FailedAssertion variant now includes an optional AssertError field to propagate custom assertion failure context from hosts.


processor/src/host/mod.rs
```diff
@@ -75,23 +75,25 @@ pub trait BaseHost {
         &mut self,
         process: &mut ProcessState,
         options: &DebugOptions,
-    ) -> Result<(), ExecutionError> {
+    ) -> Result<(), DebugError> {
```
plafer, line 78: I think the canonical way to do this would be to make those errors associated types of `BaseHost`:

```rs
pub trait BaseHost {
  type DebugError: Error + Send + Sync + 'static;
  type AssertError: Error + Send + Sync + 'static;
  type TraitError: Error + Send + Sync + 'static;

  ...
}
```

processor/src/errors.rs
```diff
@@ -478,6 +501,29 @@ pub enum AceError {
...
+/// As with `DebugError`, concrete host error types are permitted and boxed at the boundary.
+type TraceError = Box<dyn Error + Send + Sync + 'static>;
+
+/// Boxed assertion context attached to failed assertions.
+///
+/// Hosts may return any type implementing `fmt::Debug + Send + Sync + 'static` from
+/// `BaseHost::on_assert_failed`. The VM then boxes it into this trait object and includes it in
+/// diagnostics. Prefer a concrete, sized type for `BaseHost::AssertInfo` (e.g., `Felt` or `()`),
+/// and let the VM perform the boxing to avoid double-boxing.
+pub(crate) type AssertInfo = Box<dyn fmt::Debug + Send + Sync + 'static>;
```
plafer: I think those are leftover from previous implementation, and now only relevant for the `TestHost`? I would remove this section, as one of the benefits of using the associated types was to not need those types.

stdlib/docs/crypto/hashes/rpo.md
```diff
@@ -10,5 +10,8 @@ Prepares the top of the stack with the hasher initial state.<br /><br />This pro
...
+| hash | Computes RPO hash of a single 256-bit input (1 word = 4 field elements).<br /><br />Input: [A, ...]<br />Output: [B, ...]<br /><br />Where:<br />- A = [a_0, a_1, a_2, a_3] represents 4 field elements (1 word)<br />- B = [b_0, b_1, b_2, b_3] = RPO256(A)<br /> |
+| merge | Merges two 256-bit digests (2 words = 8 field elements) via RPO hash.<br /><br />Input: [B, A, ...]<br />Output: [C, ...]<br /><br />Where:<br />- A = [a_0, a_1, a_2, a_3] represents first digest (1 word)<br />- B = [b_0, b_1, b_2, b_3] represents second digest (1 word)<br />- C = [c_0, c_1, c_2, c_3] = RPO256(A \|\| B)<br /> |
+| permute | Performs RPO permutation on the hasher state.<br /><br />Input: [C, B, A, ...]<br />Output: [C', B', A', ...]<br /><br />Where:<br />- A, B, C are the three words of the hasher state<br />- A', B', C' are the permuted state words<br /><br />Cycles: 1<br /> |
```
plafer: Not sure what these doc changes are (and similarly in other files) - are these meant to be there?

plafer (approved):
LGTM!

processor/src/errors.rs
adr1anh: The Errors are only used for the inner errors of the TraceError or DebugError of the ExecutionError (hence why they're private). The AssertInfo is needed when the processor wraps the AssertInfo returned by the Host. I think it would be slight clearer if we prefix with "Boxed" to make it less ambiguous with the associated types which share the same name.

stdlib/docs/crypto/hashes/rpo.md
adr1anh: I'll double check but I got these after rebuilding with BUILDDOCS. I think next needs to be rebuilt.

processor/src/errors.rs
```diff
@@ -90,8 +110,10 @@ pub enum ExecutionError {
         #[source_code]
         source_file: Option<Arc<SourceFile>>,
         clk: RowIndex,
-        err_code: Felt,
```
bobbinth, line 93: Removing `err_code` would mean that once debug info is stripped from a program, we won't be able to identify errors, right? Basically, right now we have:

1. If the debug info is available, we can print out the full error message.
2. If the debug info is not available, we can print out error code.

If I'm understanding correctly, after this PR, the second option is no longer available. Maybe that's fine, but I just want to make sure we are not doing this by mistake.

processor/src/errors.rs
```diff
@@ -478,6 +500,29 @@ pub enum AceError {
...
+/// As with `DebugError`, concrete host error types are permitted and boxed at the boundary.
+type TraceError = Box<dyn Error + Send + Sync + 'static>;
+
+/// Boxed assertion context attached to failed assertions.
+///
+/// Hosts may return any type implementing `fmt::Debug + Send + Sync + 'static` from
+/// `BaseHost::on_assert_failed`. The VM then boxes it into this trait object and includes it in
+/// diagnostics. Prefer a concrete, sized type for `BaseHost::AssertInfo` (e.g., `Felt` or `()`),
+/// and let the VM perform the boxing to avoid double-boxing.
+pub(crate) type AssertInfo = Box<dyn fmt::Debug + Send + Sync + 'static>;
```
PhilippGackstatter: I would've expected this to be `fmt::Display`. `Debug` is for debug representations and often automatically derived. Error messages should be human-readable. This would also be in line with how `core::error::Error` uses `Display` as the error message instead of `Debug` (https://doc.rust-lang.org/std/error/trait.Error.html#method.description).

processor/src/errors.rs
PhilippGackstatter: I think going all the way with the associated types would require making `ExecutionError` generic over all of those, so callers can access those types directly instead of boxed representations. That seems undesirable because it would introduce a lot of type complexity for not much gain.

Assuming that's correct, I'm not sure if I understand the benefit of having associated types in the first place though if they get wrapped in a boxed, type-erased error anyway. I think simplifying the implementation by removing the associated types and returning `Option<Box<dyn fmt::Display + Send + Sync + 'static>>` from `on_assert_failed` directly would achieve effectively the same? Am I missing something?

bobbinth (approved):
Looks good! Thank you! I left one comment inline.

Eventually, we should make this structure quite a bit simpler. Even in a relatively short term we can:

- Get rid of `DebugHandler::on_assert_failed()` as I don't think there is a need for the host to handle failed assertions in a customizable way - we should simply return a `FailedAssertion` error (previously, we used this to associate error codes with error messages - but now this is handled internally).
- Get rid of `DebugHandler::on_trace()` - this is a legacy method that is currently used only by the compiler's debugger and this will be refactored soon (cc @bitwalker and @huitseeker).
- We may even consider getting rid of `DebugHandler::on_debug()` either because we get rid of these instructions entirely (though people seem to think that these are still useful) or because we'd support only the default behavior for them (i.e., printing to `stdout`).

Let's create an issue for this (unless w already have something along these lines).

processor/src/host/handlers.rs
```diff
@@ -178,26 +178,20 @@ impl Debug for EventHandlerRegistry {
 
 /// Handler for debug and trace operations
 pub trait DebugHandler: Sync {
```
PhilippGackstatter, line 201: Mostly unrelated to the PR, but question: I'm seeing this trait for the first time and it looks like it's not really used in the public processor API. Are Miden VM users expected to use it or would it be beneficial for them? It looks like an internal helper trait to me at first glance, but maybe I'm missing its larger purpose? Basically, as I understand it, it is up to each host to implement `on_debug`, etc. so a concrete host could delegate this to a concrete handler, without needing the trait abstraction.

processor/src/host/mod.rs
```diff
@@ -75,23 +80,17 @@ pub trait BaseHost {
         &mut self,
         process: &mut ProcessState,
         options: &DebugOptions,
-    ) -> Result<(), ExecutionError> {
-        let mut handler = debug::DefaultDebugHandler::default();
-        handler.on_debug(process, options)
-    }
+    ) -> Result<(), Self::DebugError>;
```
PhilippGackstatter: If we end up not having the associated types, I think it'd be nice to retain the default implementations here.

PhilippGackstatter (commented):
Looks good to me!

I think my main question is around the need for the associated types.

processor/src/errors.rs
PhilippGackstatter: Good point. I think to retain some chance of identifying errors in a production environment, having at least the error code would be good.

PhilippGackstatter:
> Get rid of `DebugHandler::on_assert_failed()`

I think there could be some use to it, though not essential to have. One use case I can think of is asset preservation rules in the tx kernel. Currently, if the kernel fails with this error you get the message "total number of assets in the account and all involved notes must stay the same", but it's a tedious process to figure out whether too much or too little of an asset was used, or which one. To make this more helpful, the host could track asset movement internally or extract it from process memory somehow and when it encounters this error, it enriches the failed assertion so you get something like "total number of assets in the account and all involved notes must stay the same: 200 USDC left at the end of the tx".

To be fair, this can be done differently as well, e.g. how we improve [tx fee errors](https://github.com/0xMiden/miden-base/blob/15e787d2c9ee4744be2fedbc5cdd10abf639a5cb/crates/miden-lib/asm/kernels/transaction/lib/epilogue.masm#L320) by emitting an event first and failing execution if the account doesn't have enough. The assertion enrichment route is a nicer because it does not require "polluting" the kernel with events that are only used for error improvement, though it is also less flexible because some context may be gone from the stack by the time the assertion is run. So maybe overall it's fine to remove `on_assert_failed`.

processor/src/errors.rs
adr1anh: Using associate types would be the more "correct" way of defining these error types, and this is what I attempted in the last commit. It does add a lot more complexity which feels slightly unnecessary given that we're boxing them anyway.

Both approaches make sense to me, but happy to revert to the previous type-erased approach.

plafer: Ah yes I see the full picture now... Yes indeed since all we do is box + wrap it in `ExecutionError`, I don't see the point of using associated types either.

Sorry for the suggestion - I'm good with reverting that change.

plafer (approved):
LGTM!

bobbinth (approved):
Looks good! Thank you!

PR #2370, opened 2025-11-15
refactor(assembly): simplify linker and implement link-time const evaluation

bitwalker:
This PR does three things:

1. Unifies the implementation details of symbol resolution shared between semantic analysis and link-time/assembly
2. Reworks the linker implementation to vastly simplify how the linker state is represented, and how symbols are resolved/linked.
3. Implements link-time constant evaluation on top of the new linker implementation, which was blocked on the linker refactoring.

A few last TODOs before this is ready to merge:

- [x] Clean up how linker code is organized and add any missing docs/comments
- [x] Double-check that there are tests covering all the interesting edge cases around symbol resolution.
- [x] Write up a high-level comment/doc on the relationship between the assembler, the linker, semantic analysis, and how symbol resolution is performed in all these phases.


crates/assembly-syntax/src/ast/constants/eval.rs
```diff
@@ -0,0 +1,384 @@
...
+                let (span, op) = op.into_parts();
+                match (lhs, rhs) {
+                    (ConstantExpr::Int(lhs), ConstantExpr::Int(rhs)) => {
+                        let lhs = lhs.into_inner();
+                        let rhs = rhs.into_inner();
+                        let result = match op {
+                            ConstantOp::Add => lhs + rhs,
+                            ConstantOp::Sub => lhs - rhs,
+                            ConstantOp::Mul => lhs * rhs,
+                            ConstantOp::Div | ConstantOp::IntDiv => lhs / rhs,
```
huitseeker: There's a misplaced panic on a division by zero while testing the new link-time constant evaluation feature. Here's what happens:

```rust
#[test]
fn test_constant_division_by_zero() -> TestResult {

    let module = context.parse_module_with_path(
        "div_zero",
        source_file!(&context, r#"
            pub const NUMERATOR = 10
            pub const DENOMINATOR = 0
            pub const BAD_DIV = NUMERATOR / DENOMINATOR
        "#)
    )?;

    let mut assembler = Assembler::new(context.source_manager());
    assembler.compile_and_statically_link(module)?;

    let program_source = source_file!(
        &context,
        r#"
        use div_zero
        begin
            push.div_zero::BAD_DIV
        end"#
    );

    assembler.assemble_program(program_source)
}
```

Right now this test causes a panic in the parser rather than being caught *here* during constant evaluation. The panic happens at `token.rs:284` when doing integer division.

I wonder if it would be helpful to check for division by zero here in the constant evaluation code and provide a nice error message like we do for other issues (cycles, undefined symbols).

crates/assembly-syntax/src/parser/grammar.lalrpop
```diff
@@ -2030,7 +2030,7 @@ MaybeQualifiedConstantPath: Span<PathBuf> = {
             path
```
huitseeker: ## Issue: Documentation Gap for Qualified Constant References

### Clarification: This is NOT a Bug, but a Documentation Gap

After further investigation, this appears to be an **intentional design limitation**, not a bug. The issue is that the documentation doesn't clearly state the limitation.

### What Actually Happens (Intended Design):
- ✅ **Procedures support qualified references**: `exec.::module::procedure` (documented)
- ❌ **Constants do NOT support qualified references**: `push.::module::constant` (intentionally limited, but undocumented)

### Evidence from Implementation:
The code contains explicit design intent:
```rust
/// A constant identifier
///
/// This must refer to a constant definition in the current module.
Constant(Ident),
```

This comment in `/crates/assembly-syntax/src/ast/immediate.rs` shows that constants are **intentionally designed to be local to their module**.

### The Real Problem:
**Documentation Inconsistency** - The docs show procedure qualification working but don't clarify that:
1. Constants can only be referenced locally
2. Qualified constant references are intentionally not supported
3. This creates user expectation gaps

### User Experience Issue:
Developers see `exec.::module::procedure` work and reasonably expect `push.::module::constant` to work too, leading to confusion.

### Test Evidence:
```rust
// This works as documented and intended
exec.::utils::helper

// This fails as intended, but isn't documented as such
push.::std::math::MAX_VALUE     // ❌ "invalid syntax" - intentional limitation
push.math::MAX_VALUE           // ❌ "invalid syntax" - intentional limitation
```

### Recommendation:
**Update documentation** to explicitly state:
> "Constants can only be referenced locally within their module. Qualified constant references (e.g., `push.module::constant`) are not supported, unlike procedure references."

This would prevent developer confusion and clarify the intentional design distinction between procedures and constants.

huitseeker, line 2043: Hmm, do we support qualified constants in the parser implementation?

1. **Grammar**:
   ```rust
   // In grammar.lalrpop
   MaybeQualifiedConstantPath: Span<Arc<Path>> = {
       <l:@L> <is_absolute:MaybeAbsolute> <namespace:(<BareIdent> "::")*> <name:ConstantName> <r:@R>
   ```
   This grammar explicitly supports `module::module::CONSTANT` patterns.

2. **Parser/Instructions**:
   ```rust
   push.::std::math::MAX_VALUE  //  "found a identifier here" - parser rejects
   ```

AFAICT, that rejection is undocumented. Procedures support qualification (`exec.::module::proc`), so maybe constants should too?

Fixing this would either require documenting the limitation, or updating the instruction grammars to accept `MaybeQualifiedConstantPath` instead of just `ConstantName`, similar to how procedures handle `MaybeQualifiedProcedureOrTypePath`.

crates/assembly-syntax/src/ast/constants/eval.rs
```diff
@@ -0,0 +1,384 @@
...
+        }
+    }
+}
+
+/// There are two phases to constant evaluation, one during semantic analysis, and another phase
+/// performed during linking of the final assembly, on any constant expressions that were left
+/// partially or unevaluated during semantic analysis due to external references. This trait is
+/// used to abstract over the environment in which the evaluator runs, so that we can use it in
+/// both phases by simply providing a suitable implementation.
+pub trait ConstEnvironment {
```
huitseeker, line 153: I think the evaluator cannot resolve all symbols from other modules during the linking phase.
This fails (or rather "demonstrates an issue") for me:

```rust
use alloc::string::ToString;
use miden_assembly::testing::{TestContext, source_file};
use miden_assembly::Assembler;

type TestResult = Result<(), miden_assembly::diagnostics::Report>;

#[test]
fn test_cross_module_constant_resolution_failure() -> TestResult {

    // Module A defines and exports a constant
    let module_a = context.parse_module_with_path(
        "cycle::module_a",
        source_file!(&context, r#"
            pub const A_VAL = 10
            pub proc a_proc
                push.A_VAL
            end
        "#)
    )?;

    // Module B imports Module A and defines a constant using it
    let module_b = context.parse_module_with_path(
        "cycle::module_b",
        source_file!(&context, r#"
            use cycle::module_a
            pub const B_VAL = module_a::A_VAL + 5  # <-- Should work but fails
            pub proc b_proc
                push.B_VAL
            end
        "#)
    )?;

    let mut assembler = Assembler::new(context.source_manager());

    // Compile Module A first (this succeeds)
    assembler.compile_and_statically_link(module_a)?;

    // Try to compile Module B (this should fail with cross-module resolution error)
    match assembler.compile_and_statically_link(module_b) {
        Ok(_) => {
            // This would be unexpected
            return Err(miden_assembly::diagnostics::Report::msg(
                "Expected cross-module constant resolution to fail"
            ));
        }
        Err(e) => {
            let error_msg = format!("{}", e);
            // Verify this is the specific cross-module resolution error we're documenting
            if error_msg.contains("undefined symbol reference") && error_msg.contains("module_a::A_VAL") {
                println!("Reproduced the cross-module symbol resolution failure");
                println!("Error: {}", error_msg);
                // This is expected - return Ok to make the test pass
                return Ok(());
            } else {
                println!("Different error than expected: {}", error_msg);
                return Err(e);
            }
        }
    }
}
```

### Error Message:
```
Error: x undefined symbol reference
   ,-[test143:3:31]
 2 |             use cycle::module_a
 3 |             pub const B_VAL = module_a::A_VAL + 5
   :                               ^^^^^^^|^^^^^^^
   :                                      `-- this symbol path could not be resolved
 4 |             pub proc b_proc
   `----
  help: maybe you are missing an import?
```

Should we not be able to resolve recursively to the constant?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Added new `adv.push_mapval_count` instruction ([#2349](https://github.com/0xMiden/miden-vm/pull/2349)).
 - Added new `memcopy_elements` procedure for the `std::mem` module. ([#2352](https://github.com/0xMiden/miden-vm/pull/2352)).
+- Implemented link-time const evaluation ([#2370](https://github.com/0xMiden/miden-vm/pull/2370))
+- Simplified linker implementation and improved consistency of symbol resolution and associated errors ([#2370](https://github.com/0xMiden/miden-vm/pull/2370))
```
huitseeker: We may want to clean that up.

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -0,0 +1,715 @@
...
+            log::debug!(target: "name-resolver::path", "resolving path as '{subpath}' relative to '{ns}'");
+            // Check if the first component of the namespace was previously imported
+            match self.resolve_import(context, Span::new(path.span(), ns))? {
+                SymbolResolution::Exact { gid, path } => {
+                    if subpath.is_empty() {
+                        log::debug!(target: "name-resolver::path", "resolved '{ns}' to imported item '{path}'");
+                        Ok(SymbolResolution::Exact { gid, path })
+                    } else {
+                        log::error!(target: "name-resolver::path", "resolved '{ns}' to imported item '{path}'");
+                        Err(Box::new(SymbolResolutionError::InvalidSubPath {
```
huitseeker, line 189: This is where we get the [issue in my other comment](https://github.com/0xMiden/miden-vm/pull/2370/files#r2570081431).  Instead of immediately throwing `InvalidSubPath`, the resolver could check if the imported item is a module, and if yes, recursively resolve the subpath within that module.

huitseeker (commented):
I left questions inline on surprising behavior.

crates/assembly-syntax/src/ast/constants/eval.rs
bitwalker: Yeah we catch division by zero in some cases (or at least, we used to), but looks like we're missing it in some. Can/should definitely handle that more appropriately, I'll address that here.

crates/assembly-syntax/src/parser/grammar.lalrpop
bitwalker: We don't currently supported qualified constant references as immediate operands, though that limitation is artificial. In practice, what it enforces is that a constant is imported before it is used locally within the module, but that isn't strictly needed - it just mirrors how constants have been used up until now.

I'll document this limitation for now, but we can lift it in a future PR that is non-breaking at the syntax level, just breaking at the AST level (which basically only affects the compiler in practice currently). To lift it, we'd need to modify `Immediate<T>` to use `Arc<Path>` rather than `Ident` for the `Constant` variant; and then refactor all the code that constructs/consumes immediates accordingly.

crates/assembly-syntax/src/ast/constants/eval.rs
bitwalker: Yes, this should be supported, looking to see if this is an issue with `Path` comparisons, or something else.

CHANGELOG.md
bitwalker: Could you clarify what you mean? The two entries refer to the same PR, but are referring to two different changes.

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Yes, though to be more precise, the issue here is that `resolve_import` should return `Exact` only when the path refers to an item, but it is returning `Exact` when it should be returning `Module`. This is almost certainly due to the module-local resolver resolving the path to an import (and thus an exact path), but it cannot determine whether an external path refers to an item, or a module, so it just returns `Exact`. The `resolve_import` implementation isn't handling this by mapping exact paths to `Module` when the path refers to a module not an item.

crates/assembly-syntax/src/ast/constants/eval.rs
bitwalker: Fixed in fe4c8a9. The issue was in an incorrect assumption made by `ConstEvalVisitor` when handling the expression returned by `ConstEnvironment::get` - this occurs when applying symbol rewrites at link-time, specifically to constant items, and is why this issue came up with your test, but not with existing multi-module linker tests.

I've added your test to the suite so that we can regressions along these lines in the future if they occur.

crates/assembly-syntax/src/parser/grammar.lalrpop
bitwalker: Documented in [43fd91f](https://github.com/0xMiden/miden-vm/pull/2370/commits/43fd91fd2775281d6b5cd807f1aa57349e021a26)

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: After further investigation, the issue wasn't actually along these lines, see my follow-up to your other comment for details.

However, I did some checking to determine under what circumstances `resolve_import` might return `SymbolResolution::Exact`, and it will always be when the reference resolves to an item, never a module alias. This is because we recursively resolve aliases, and the module-local resolver always returns `SymbolResolution::External` for imported/aliased items. The link-time resolver will then always resolve `External` to `Module` if the path actually refers to a module rather than an item (in which case either `Exact` or `MastRoot` will be produced).

So TL;DR - `resolve_path`/`resolve_import` is working as expected

crates/assembly-syntax/src/ast/constants/eval.rs
bitwalker: Addressed in 6c334b7

CHANGELOG.md
huitseeker: Either two lines with one single url, or one long line split with a semi column would be relatively easy to interpret for me. Here the url duplication made me think of a typo.

crates/assembly-syntax/src/ast/constants/eval.rs
```diff
@@ -0,0 +1,411 @@
...
+                }
+            },
+            Cont::Eval(ConstantExpr::BinaryOp { span, op, lhs, rhs, .. }) => {
+                continuations.push(Cont::Apply(Span::new(span, op)));
+                continuations.push(Cont::Eval(*lhs));
+                continuations.push(Cont::Eval(*rhs));
+            },
+            Cont::Apply(op) => {
+                let lhs = stack.pop().unwrap();
+                let rhs = stack.pop().unwrap();
```
huitseeker, line 329: Should this not assert on the length of the stack, similarly to the approach taken at  the end of this function?

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -0,0 +1,731 @@
...
+                    path: Span::new(context.span, self.module_path(id).to_path_buf().into()),
+                });
+            }
+        }
+
+        // The path must refer to an item, so resolve the item
+        if path.is_absolute() {
+            self.find(context, path)
+        } else {
+            let (ns, subpath) = path.split_first().unwrap();
```
huitseeker, line 179: I think this panics if the path is empty here, doesn't it?

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -0,0 +1,731 @@
...
+        } else {
+            Cow::Borrowed(context)
+        };
+        let mut resolving = path.map(|p| Arc::<Path>::from(p.to_path_buf()));
+        let mut visited = BTreeSet::default();
+        loop {
+            let current_module_path = self.module_path(current_context.module);
+            log::debug!(target: "name-resolver::find", "resolving {} of {resolving} from {} ({current_module_path})", current_context.display_kind(), &current_context.module);
+
+            let (resolving_symbol, resolving_parent) = resolving.split_last().unwrap();
```
huitseeker, line 410: Could this not also panic on an empty path?

crates/assembly/src/linker/mod.rs
```diff
@@ -986,30 +768,48 @@ impl Linker {
...
+        cache: &mut ResolverCache,
+    ) -> Result<ast::ConstantValue, LinkerError> {
+        let symbol_resolver = SymbolResolver::new(self);
+        let mut resolver = Resolver {
+            resolver: &symbol_resolver,
+            cache,
+            current_module: gid.module,
+        };
+
+        ast::constants::eval::expr(expr, &mut resolver).map(|expr| expr.expect_value())
```
huitseeker, line 834: If the const expression cannot be fully evaluated, this will panic rather than return an error. Is this the behavior we want, when we have the opportunity to signal the user how they messed up and recover?

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -0,0 +1,731 @@
...
+                    }
+
+                    let span = path.span();
+                    let path = module_path.join(subpath);
+                    let context = SymbolResolutionContext {
+                        span: module_path.span(),
+                        module: id,
+                        kind: context.kind,
+                    };
+                    self.resolve_path(&context, Span::new(span, path.as_path()))
```
huitseeker, line 232: Idle thought: could a user-provided program with a lot of components blow the recursion stack? The simple mitigation would be to early reject on deep module paths `mod_1::..::mod_N` for e.g. `N > 100`.

crates/assembly-syntax/src/ast/constants/eval.rs
bitwalker: No, because the stack could be arbitrarily deep at this point, we just expect that there are two operands on the stack here (hence the `unwrap`s). This could be changed to use `expect` instead, but since the only way this can fail is if this function itself is totally broken, i.e. we eval'd operands of a `BinaryOp` and didn't push any value back on the stack for one or both of them, which shouldn't be possible by construction - that feels unnecessary to me.

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Yes, though there will never be a case where the provided path would be empty here, so we would necessarily want to panic if we observed one. We could use `expect` here, but I prefer to use that in cases where it is either unclear what precondition was violated, or where it is reasonable to assume that a panic could be triggered due to user error or invalid inputs.

This code is only called from the linker, and only when we have a path guaranteed to have at least one valid component - if we ever did have an empty path reach this point, it would be clear why the panic was triggered, but not how an empty path reached this part of the code. IMO an `unwrap` makes it clear that we do not expect this operation to ever fail, and if it does, it will require deeper introspection into the linker to understand how that happened.

I don't mind using `expect` instead, I just find it to clutter the code unnecessarily in cases like this, but it doesn't bother me _that_ much.

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Same here - it is effectively impossible for `resolving` to ever be empty. If we ever did manage to observe an empty path here, it wouldn't be clear how that even happened without further investigation.

crates/assembly/src/linker/mod.rs
bitwalker: By the time this is called, we should have caught any issues that would prevent evaluation to a concrete value (i.e. unresolvable symbols, etc.). So if this ever raises a panic, it is a bug in the linker (or perhaps the evaluator), not something a user did incorrectly.

Since `eval::expr` is called both in semantic analysis (where failure to fully evaluate is totally fine) and when obtaining the value of exported constants during assembly (but after linking, so failure to fully evaluate should never happen if the linker is implemented correctly), the API is designed to return `ConstantExpr` rather than `ConstantValue`, but in the linker/assembler, we can safely expect that the returned `ConstantExpr` is always convertible to `ConstantValue`, or we have a bug.

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: No, or at least not due to the number of path components. The recursion is limited in two ways already:

1. We are only resolving/expanding the first component of non-absolute paths, not every component. In order for this expansion step to blow the stack due to recursion, you'd have to have a very long chain of re-exported module paths, e.g. `exec.a::b` resolves `a` to an import like `use lib::a`, and `lib::a` resolves to a re-exported module alias like `pub use lib::a2`, which resolves `lib::a2` to another re-export, and so on. It would be extraordinarily unlikely that someone would have a long enough chain like this within their own code.
2. Re-exported module paths are not preserved in assembled libraries, so you can't end up with a situation where the recursion is extended by the dependency tree - the chain of re-exports has to exist in the sources you provided to the assembler.

You'd have to deliberately be invoking the assembler on a deeply pathological program to even trigger a stack overflow due to the recursion here, and that just doesn't feel like an issue worth addressing explicitly IMO.

bitwalker:
> The PR body has an unchecked TODO: "Write up a high-level comment/doc on the relationship between the assembler, the linker, semantic analysis, and how symbol resolution is performed in all these phases". `docs/src/user_docs/assembly/code_organization.md` does not yet have any mention of e.g. link-time evaluation, and I haven't found an issue tracking the gap, would you mind opening one?

I don't think the documentation I'm talking about here belongs in user-facing docs, personally. The documentation I wanted to provide here is for those working on the assembler itself, to provide the necessary context to be able to reason about the existing code, and make changes without breaking any assumptions that are baked in.

I'm actually not sure where the best place is for that documentation though, @bobbinth what do you think?

libcore/docs/collections/smt.md
```diff
@@ -4,4 +4,4 @@
...
+| peek | Emits the "core::collections::smt::smt_peek" event, which pushes onto the advice stack the<br />value associated with the specified key in a Sparse Merkle Tree defined by the specified root.<br /><br />If no value was previously associated with the specified key, [ZERO; 4] is pushed onto the advice<br />stack.<br /><br />Inputs:<br />Operand stack: [KEY, ROOT]<br />Advice stack:  []<br /><br />Outputs:<br />Operand stack: [KEY, ROOT]<br />Advice stack:  [VALUE]<br /><br />Panics if:<br />- the tree with the specified root does not exits in the VM's advice provider.<br /><br />Cycles: 3 cycles<br /> |
```
bobbinth: This change should go away once the latest `next` is merged into this branch.

libcore/docs/crypto/dsa/ecdsa_k256_keccak.md
```diff
@@ -4,4 +4,4 @@
...
+| verify_impl | Internal implementation of ECDSA signature verification via deferred computation.<br /><br />Emits an event to trigger the precompile handler, reads the verification result from<br />the advice stack, and computes the commitment and tag for tracking deferred verification.<br /><br />This procedure mimics the `ecdsa_secp256k1::PublicKey::verify_prehash()` function from<br />`miden-crypto`, which takes a pre-hashed message that the caller must provide<br />(e.g. obtained using the keccak256 precompile).<br /><br />Input: `[pk_ptr, digest_ptr, sig_ptr, ...]`<br />Output: `[COMM, TAG, result, ...]`<br /><br />Where:<br />- `pk_ptr`: word-aligned memory address containing 33-byte public key<br />- `digest_ptr`: word-aligned memory address containing 32-byte digest<br />- `sig_ptr`: word-aligned memory address containing 66-byte signature<br />- `COMM`: commitment to calldata computed as<br />`rpo256256(rpo256256(rpo256256(pk) \|\| rpo256256(digest)) \|\| rpo256256(sig))`<br />- `TAG`: `[ECDSA_VERIFY_EVENT, result, 0, 0]`<br />- `result`: 1 if signature is valid, 0 if invalid<br /> |
```
bobbinth: Same as the last comment.

libcore/docs/crypto/hashes/rpo256.md
```diff
@@ -1,4 +1,5 @@
-## miden::core::crypto::hashes::rpo
+
+## miden::core::crypto::hashes::rpo256
```
bobbinth, line 2: Same as the last comment.

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -152,11 +156,14 @@ pub fn analyze(
...
 
     // Check unused imports
+    // TODO(pauls): This warning is temporarily disabled until we rework how symbol usage is tracked
+    /*
     for import in module.aliases() {
         if !import.is_used() {
             analyzer.error(SemanticAnalysisError::UnusedImport { span: import.span() });
         }
     }
+     */
```
bobbinth: Is this `TODO` still relevant?

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -179,9 +186,16 @@ fn visit_items(module: &mut Module, analyzer: &mut AnalysisContext) -> Result<()
                 }
 
                 // Evaluate all named immediates to their concrete values
+                log::debug!(target: "const-eval", "visiting procedure {}", procedure.name());
```
bobbinth, line 186: Question (and not for this PR): do we want to keep logging/debugging statements in the code long-term?

crates/assembly/src/tests.rs
```diff
@@ -2589,6 +2779,35 @@ fn module_alias() -> TestResult {
         r#" help: expected "@", or "adv_map", or "begin", or "const", or "enum", or "proc", or "pub", or "type", or "use", or end of file, or doc comment"#
     );
 
+    Ok(())
+}
+
+#[test]
+#[ignore = "disabled until unused import accuracy is improved"]
+fn module_alias_unused_import() -> TestResult {
```
bobbinth, line 2788: Is the "ignore" statement here still relevant? If so, we should probably create an issue for this (unless we already have one).

crates/assembly/src/assembler.rs
```diff
@@ -489,176 +486,37 @@ impl Assembler {
...
-                                SymbolResolution::Local(_) | SymbolResolution::External(_) => {
-                                    unreachable!()
-                                },
-                            }
-                        },
+                let module = &self.linker[module_idx];
+
+                if module.is_mast() {
+                    continue;
+                }
```
bobbinth: Could we add a brief comment about why we are skipping modules that were loaded from MAST directly? Is it because there are no symbols to resolve there?

crates/assembly/src/assembler.rs
```diff
@@ -823,11 +796,21 @@ impl Assembler {
                     // be added to the forest.
```
bobbinth, line 771: Is this `TODO` still relevant after this PR?

crates/assembly/src/assembler.rs
```diff
@@ -823,11 +796,21 @@ impl Assembler {
...
-                    let path = module.path().join(alias.name().as_str()).into();
+                SymbolItem::Alias { alias, resolved } => {
+                    let procedure_gid = resolved.get().expect("resolved alias");
+                    match self.linker[procedure_gid].item() {
+                        SymbolItem::Procedure(_) | SymbolItem::Compiled(ItemInfo::Procedure(_)) => {
+                        },
+                        SymbolItem::Constant(_) | SymbolItem::Type(_) | SymbolItem::Compiled(_) => {
+                            continue;
+                        },
+                        SymbolItem::Alias { .. } => unreachable!(),
```
bobbinth, line 789: Could we add a brief comment about why `unreachable!` is ok here?

crates/assembly/src/assembler.rs
```diff
@@ -866,8 +849,9 @@ impl Assembler {
                     self.linker.register_procedure_root(procedure_gid, proc_mast_root)?;
                     mast_forest_builder.insert_procedure(procedure_gid, procedure)?;
                 },
-                // Type and constant exports are irrelevant at this stage
-                Export::Type(_) | Export::Constant(_) => continue,
+                SymbolItem::Compiled(_) | SymbolItem::Constant(_) | SymbolItem::Type(_) => {
+                    unreachable!()
```
bobbinth: Same comment as above: could we add a brief explanation why `unreachable!` is OK here?

crates/assembly/src/assembler.rs
```diff
@@ -681,6 +539,118 @@ impl Assembler {
         Ok(Library::new(mast_forest.into(), exports)?)
     }
 
+    fn export_symbol(
```
bobbinth, line 521: Could we add a doc comment explaining the purpose of this function?

crates/assembly/src/assembler.rs
```diff
@@ -681,6 +539,118 @@ impl Assembler {
...
+            SymbolItem::Compiled(ItemInfo::Procedure(item)) => {
+                let resolved = match mast_forest_builder.get_procedure(gid) {
+                    Some(proc) => ResolvedProcedure {
+                        node: proc.body_node_id(),
+                        signature: proc.signature(),
+                    },
+                    // We didn't find the procedure in our current
+                    // MAST forest. We still need to
+                    // check if it exists in one of a library
+                    // dependency.
```
bobbinth: nit: comment formatting (i.e., the lines can be longer).

crates/assembly-syntax/src/sema/mod.rs
huitseeker: And do we not want to upgrade them to [tokio-tracing spans](https://github.com/tokio-rs/tracing), which the line @bobbinth is here commenting on is a posterchild use case for?

crates/assembly-syntax/src/sema/mod.rs
bitwalker: Yes, I need to create an issue for it. It's a low priority issue, but worth fixing when we get a chance.

crates/assembly-syntax/src/sema/mod.rs
bitwalker: These are specifically designed to be left in the code - they are zero-cost if logging isn't enabled (or the min log level is higher than the level of the log statement). Since they are particularly useful for debugging the inner workings of the assembler, it's worth keeping them around IMO.

crates/assembly/src/tests.rs
bitwalker: Yes, this is related to the TODO from above, I'll create an issue for tracking this.

crates/assembly-syntax/src/sema/mod.rs
bitwalker: I revisited this just to see how things look now, and I was actually able to re-enable these warnings. It's still not tracking _all_ unused symbols, but unused imports are being tracked again.

crates/assembly/src/assembler.rs
bitwalker: This was actually an artifact of how things used to work, I've removed this check now as it is no longer needed.

crates/assembly/src/assembler.rs
bitwalker: I believe so - this was added by @plafer at one point IIRC, but the changes in this PR wouldn't have an effect on what is documented here.

crates/assembly-syntax/src/sema/mod.rs
bitwalker: @huitseeker I'm not clear on why we would want to use tracing spans for this - we're not looking to instrument this code, but simply provide useful context about how assembly of a module/package is proceeding when troubleshooting things like semantic analysis, const evaluation, and symbol resolution. Taking on a dependency to `tokio`'s tracing infra especially seems egregious to me for this crate.

crates/assembly/src/linker/mod.rs
```diff
@@ -1,206 +1,89 @@
...
+//!    checks that only require minimal module-local context. Initial symbol resolution is performed
+//!    here based on module-local context, as well as constant folding of expressions that can be
+//!    resolved locally. Symbols which refer to external items are unable to be fully processed as
+//!    part of this phase, and is instead left to the linking phase.
+//! 3. _Linking_, the most critical phase of compilation. During this phase, the assembler has the
+//!    full compilation graph available to it, and so this is where inter-module symbol references
+//!    are finally able to be resolved (or not, in which case appropriate errors are raised). This
+//!    is the phase where we catch cyclic references, references to undefined symbols, references to
+//!    non-public symbols from other modules, etc. Once all symbols are linked, the assembler is
+//!    free to compile all of the procedures to MAST, and generate a [crate::Library].
```
bobbinth, line 17: Question (mostly for my education purposes): why not do all symbol resolution in the linker phase (vs. doing local resolution during semantic analysis and final resolution during linking)? If there is a tradeoff here between simplicity and performance (e.g., splitting it into two phases is more performant but also more complicated), I'd probably go with the simpler and less performant code.

crates/assembly/src/assembler.rs
bobbinth: Do we have an issue tracking this?

bobbinth (approved):
Looks good! Thank you! This is still a pretty shallow review from me, but I think we can merge (unless @huitseeker as some other comments).

crates/assembly/src/linker/resolver/symbol_resolver.rs
huitseeker: A panic is indeed not possible in the current call graph of this public method. A great way to signal this call graph is not meant to change to liberally is to make this `pub(crate)`.

crates/assembly/src/linker/mod.rs
bitwalker: It wouldn't actually be meaningfully simpler to only do it during linking, as the implementation wouldn't really change, aside from some superficial differences. The symbol resolution happening during semantic analysis is super basic anyway, mainly verifying procedure invocations, and resolving references to constants for const evaluation (to simplify the AST early as much as possible).

It's likely that in the future, when the parser is reworked, that it might make sense to remove the constant evaluation from semantic analysis entirely, and push it all into the linker. That would probably simplify the const evaluation code a bit, mostly by way of removing the need for some of the abstraction (e.g. `ConstEnviroment`).

PR #2373, opened 2025-11-22
perf(cli): eliminate duplicate proof serialization in ProofFile::write

maradini77:
The `ProofFile::write` method was calling `proof.to_bytes()` twice:
1. Once in the `#[instrument]` attribute to compute the `size` field for tracing
2. Once in the function body to write the proof to disk

This caused unnecessary CPU and memory overhead, especially for large proofs that can be several megabytes in size.

## Solution

Moved the `size` field computation to after the single `proof.to_bytes()` call using `Span::current().record()`. This ensures:
- Proof serialization happens only once
- Tracing information is still captured correctly
- Reduced memory allocation and CPU usage


plafer:
Closing per our contribution guidelines.

PR #2375, opened 2025-11-23
refactor: use hash_iter API for Blake3_256 when hashing concatenated bytes

huitseeker:
Replace` Blake3_256::hash(&vec)` with `Blake3_256.hash_iter(iter)` 

This avoids allocating a contiguous buffer in the hash API and instead uses the new hash_iter method that accepts an iterator of byte slices.

Note: Only refactored locations where bytes were being concatenated from multiple sources into a Vec before hashing. Single-slice hash operations (like hash_string_to_word) were left unchanged as they don't benefit from this optimization.

Updates to `miden-crypto` 0.18.4


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1393,19 +1394,31 @@ impl MastForestContributor for BasicBlockNodeBuilder {
...
         }
 
-        if bytes_to_hash.is_empty() {
+        // Create iterator of slices from all collected data
+        let decorator_bytes_iter = before_enter_bytes
+            .iter()
+            .map(|bytes| bytes.as_slice())
+            .chain(core::iter::once(op_decorator_data.as_slice()))
+            .chain(after_exit_bytes.iter().map(|bytes| bytes.as_slice()))
+            .chain(core::iter::once(assert_data.as_slice()));
```
plafer, line 1405: nit: this can be moved in the `else` branch

plafer (approved):
LGTM

core/src/mast/node/basic_block_node/mod.rs
bobbinth, line 1336: Not for now, but when we get to simplifying how basic blocks are represented, I'd love to find a way to simplify this function (it is not easy to follow now).

bobbinth (approved):
Looks good! Thank you!

PR #2376, opened 2025-11-24
chore: rebuild stdlib

huitseeker:
Those files were moved in #2356 so the CI check (based on `git status --porcelain`) didn't detect they should be absent.


bobbinth (approved):
Looks good! Thank you!

PR #2380, opened 2025-11-25
fix: fix stdlib typo & test

partylikeits1983:
This PR fixes a minor typo in stdlib `u256` library. 

Conversation on Slack: 

https://midengroup.slack.com/archives/C08UNH9UEM8/p1764081887329699

### DIFF
--- a/stdlib/docs/math/u256.md
+++ b/stdlib/docs/math/u256.md
@@ -2,4 +2,4 @@
 ## std::math::u256
 | Procedure | Description |
 | ----------- | ------------- |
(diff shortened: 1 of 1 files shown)


stdlib/asm/math/u256.masm
```diff
@@ -243,7 +243,7 @@ proc mulstep4
     movdn.6
 end
 
-#! Performs addition of two unsigned 256 bit integers discarding the overflow.
+#! Performs multiplication of two unsigned 256 bit integers discarding the overflow.
```
adr1anh, line 246: I think we need to rebuild for the stdlib docs to be updated

partylikeits1983: Updated the docs by running `make doc`, should be updated now.

PR #2382, opened 2025-11-25
Cleanup parallel trace generation

### LINKED ISSUE #2162: Address all TODOs in parallel trace generation

plafer:
Closes #2162

This PR is simply a big cleanup/refactor; it does not add any new functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Pre-allocate main trace buffer in trace generation ([#2345](https://github.com/0xMiden/miden-vm/pull/2345)).
 - Rename the MASM standard library to "miden::core", the crate to `miden-libcore`, and various other MASM module refactors ([#2260](https://github.com/0xMiden/miden-vm/issues/2260))
 - Add a compaction function for achieving maximal sharing out of a `MastForest` with stripped decorators ([#2408](https://github.com/0xMiden/miden-vm/pull/2408)).
+- Refactor and remove tech debt from parallel trace generation ([#2382](https://github.com/0xMiden/miden-vm/pull/2382))
 
 ## 0.19.1 (2025-11-6)
 


huitseeker (commented):
You may want to integrate #2385 while we're at it?

plafer:
> You may want to integrate https://github.com/0xMiden/miden-vm/pull/2385 while we're at it?

I'm not touching any of these files, so we can review that PR independently.

processor/src/parallel/core_trace_fragment/trace_row.rs
```diff
@@ -0,0 +1,604 @@
...
+        // Store the buffer for the next call
+        self.stack_rows = Some(new_stack_rows);
+    }
+}
+
+// HELPERS
+// ===============================================================================================
+
+/// Returns op batch flags for the specified group count.
+fn get_op_batch_flags(num_groups_left: Felt) -> [Felt; 3] {
```
huitseeker, line 588: I noticed this `get_op_batch_flags` function is identical to one in `processor/src/decoder/trace.rs:558`.

Have you considered extracting this shared logic into a common utility function?

processor/src/parallel/core_trace_fragment/trace_row.rs
```diff
@@ -0,0 +1,604 @@
...
+    },
+};
+
+use super::CoreTraceFragmentFiller;
+use crate::{
+    decoder::block_stack::ExecutionContextInfo, parallel::core_trace_fragment::BasicBlockContext,
+    processor::StackInterface,
+};
+
+const HASH_CYCLE_LEN: Felt = Felt::new(miden_air::trace::chiplets::hasher::HASH_CYCLE_LEN as u64);
```
huitseeker: This `HASH_CYCLE_LEN` constant appears in several places across the codebase:

- `processor/src/fast/execution_tracer.rs:37`
- `processor/src/decoder/mod.rs:48`
- `processor/src/parallel/core_trace_fragment/trace_row.rs:28` (this line)

Would it make sense to extract this into a shared constant?

processor/src/parallel/core_trace_fragment/trace_row.rs
```diff
@@ -0,0 +1,604 @@
...
+
+// DECODER ROW
+// ================================================================================================
+
+/// The data necessary to build the decoder part of a trace row.
+#[derive(Debug)]
+struct DecoderRow {
```
huitseeker, line 33: This `DecoderRow` struct and its builders look good! I noticed this pattern of trace row building could potentially be useful across different trace builders (fast, parallel, regular).

Have you considered extracting this into a `TraceRowBuilder` trait? Something like the following, perhaps?

```rust
  /// Types of trace rows that can be built
  pub enum TraceRowType {
      ControlFlow { opcode: u8, hasher_state: (Word, Word), addr: Felt },
      BasicBlockBatch { operation: Operation, op_batch: OpBatch, addr: Felt, group_count: Felt },
      Operation { operation: Operation, op_idx_in_group: usize, user_op_helpers: Option<[Felt; NUM_USER_OP_HELPERS]> },
  }

  pub trait TraceRowBuilder {
      type Context;
      type Error;

      /// Build a trace row and integrate it into the trace
      fn build_trace_row(&mut self, row_type: TraceRowType, context: &mut Self::Context) -> Result<(), Self::Error>;

      /// Get the current row index being built
      fn current_row_index(&self) -> usize;

      /// Increment the clock cycle
      fn increment_clk(&mut self);
  }
```

It might make the pattern reusable and testable across different trace generation contexts.

processor/src/parallel/core_trace_fragment/trace_row.rs
huitseeker: (This is also the general direction of what I was hinting at when suggesting #2385)

processor/src/parallel/core_trace_fragment/mod.rs
```diff
@@ -0,0 +1,1018 @@
...
+mod execution;
+mod trace_row;
+
+// CORE TRACE FRAGMENT
+// ================================================================================================
+
+/// The columns of the main trace fragment. These consist of the system, decoder, and stack columns.
+///
+/// A fragment is a collection of columns of length `fragment_size` or less. Only a
+/// fragment containing a `HALT` operation is allowed to be shorter than
```
adr1anh: Is it "only the last fragment which ends in a HALT operation is allowed to be shorter"?

plafer: Yes, effectively only the last fragment is allowed to be shorter; will rephrase

processor/src/parallel/core_trace_fragment/trace_row.rs
plafer: Normally yes, except that the one in `processor/src/decoder/trace.rs:558` is used by the old processor, which we'll be removing soon. So basically this is the only version that will remain, and will be only used in this file, so I think it makes sense to leave it there

processor/src/parallel/core_trace_fragment/trace_row.rs
plafer: Done; created a `HASH_CYCLE_LEN_FELT` in `processor::utils`

processor/src/parallel/core_trace_fragment/trace_row.rs
plafer: I'm definitely open to exploring that! However I'd like to understand a bit more what you have in mind.

> I noticed this pattern of trace row building could potentially be useful across different trace builders (fast, parallel, regular).

The `CoreTraceFragmentFiller` (i.e. "parallel" trace builder) is currently our only trace builder, given that the fast processor doesn't build a trace, and the "regular" processor (i.e. `Process`) is deprecated/going away. Where else were you seeing this pattern being used?

Also, given that this PR is better seen as a "best effort refactoring", I would move this discussion to a new issue to be addressed in a subsequent PR. What do you think?

processor/src/fast/trace_state.rs
```diff
@@ -270,7 +295,7 @@ pub struct ExecutionReplay {
 #[derive(Debug, Default)]
 pub struct BlockStackReplay {
     /// The parent address - needed for each node start operation (JOIN, SPLIT, etc).
```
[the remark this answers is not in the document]
plafer: Reworded slightly differently

processor/src/parallel/core_trace_fragment/mod.rs
plafer: Rephrased to

```rs
/// A fragment is a collection of columns of length `fragment_size` or less. Only the last fragment
/// is allowed to be shorter than `fragment_size`.
```

processor/src/parallel/core_trace_fragment/trace_row.rs
huitseeker: To me, the creation of `TraceGenerationContext` containing `Vec<CoreTraceFragmentContext>` during execution looks like trace generation of another kind, but yeah, abstraction becomes obsolete when we delete the old processor. Please ignore.

huitseeker (approved):
🙏

PR #2384, opened 2025-11-26
test, ci: use cargo-hack to check all feature combinations

huitseeker:
The workspace has over 60 feature combinations. This introduces a script, make command and CI job that tests them all.

### DIFF
--- a/core/src/mast/serialization/decorator.rs
+++ b/core/src/mast/serialization/decorator.rs
@@ -25,7 +25,7 @@ use crate::{
 #[derive(Debug, PartialEq, Eq)]
 #[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(Arbitrary))]
 #[cfg_attr(
-    all(feature = "arbitrary", test),
+    all(feature = "arbitrary", feature = "std", test),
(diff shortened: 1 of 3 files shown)


core/src/mast/serialization/decorator.rs
```diff
@@ -25,7 +25,7 @@ use crate::{
 #[derive(Debug, PartialEq, Eq)]
 #[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(Arbitrary))]
 #[cfg_attr(
-    all(feature = "arbitrary", test),
+    all(feature = "arbitrary", feature = "std", test),
```
bobbinth, line 28: Should this use `feature = "std"` or `feature = "serde"`? (and same for line 158 below)

I'm basign this purely on the test name - so, may be totally off here.

bobbinth (approved):
Looks good! Thank you! I left just one question inline.

core/src/mast/serialization/decorator.rs
huitseeker: `serde_test(false)`, for this macro, means this doesn't emit the serde tests.

Issue #2378, opened 2025-11-24
Create a `peek` procedure for the `stdlib::collections::smt`


Fumuran:
As it was discussed in the #2222, currently we use `smt_peek` event in the `miden-base` directly, since there is no API in the `stdlib` which could call it. We rely on the fact that the name of this event in the base matches the name in the VM, so it will be handled as a regular `stdlib` event, which is quite unreliable.

We should add a new `peek` procedure for the `stdlib::collections::smt` which essentially will wrap the `smt_peek` event inside. That way we will be able to remove the direct call in the `miden-base` and call this new procedure instead.

This procedure could look like so:
```masm
pub proc peek
    emit.event("stdlib::collections::smt::smt_peek")
end
```

 _Originally posted by @PhilippGackstatter in [#2222](https://github.com/0xMiden/miden-vm/issues/2222#issuecomment-3382469553)_

PR #2387, opened 2025-11-26
Implement `peek` for `std::collections::smt`

Fumuran:
This PR implements new `peek` procedure for the `std::collections::smt` module.

It works similar to the `smt::get` procedure, but instead of pushing the resulting value onto the operand stack, it is pushed to the advice stack, leaving the operand stack unchanged.

Stack transition for this procedure looks like so:

```
Inputs:
  Operand stack: [KEY, ROOT]
  Advice stack:  []

Outputs:
  Operand stack: [KEY, ROOT]
  Advice stack:  [VALUE]
```

Closes: #2378

### DIFF
--- a/stdlib/tests/collections/smt.rs
+++ b/stdlib/tests/collections/smt.rs
@@ -576,12 +576,12 @@ fn test_smt_peek() {
 
     let smt = Smt::with_entries(LEAVES).unwrap();
 
-    // Get all leaves present in tree
+    // Peek all leaves present in tree
     for (key, value) in LEAVES {
         expect_value_from_peek(key, value, &smt);
     }
 
-    // Get an empty leaf
+    // Peek an empty leaf
     expect_value_from_peek(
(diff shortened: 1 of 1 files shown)


stdlib/asm/collections/smt.masm
```diff
@@ -5,6 +5,11 @@ use std::mem
 use std::sys
 use std::word
 
+# Constants
```
bobbinth: nit: let's capitalize this as we usually do.

stdlib/tests/collections/smt.rs
```diff
@@ -548,6 +548,48 @@ fn test_smt_set_remove_in_multi() {
     expect_remove(&smt, K2);
 }
 
+/// Tests `peek` on every key present in the SMT, as well as an empty leaf
+#[test]
+fn test_smt_peek() {
+    fn expect_value_from_get(key: Word, value: Word, smt: &Smt) {
```
bobbinth: nit: should the name here be `expect_value_from_peek`?

Issue #1719, opened 2025-03-22
Create a new `hash_memory_with_state` procedure in `stdlib`


Fumuran:
### Feature description

It should be better form the user experience perspective to create just one new `hash_memory_with_state` interface procedure instead of having `prepare_hasher_state` and `hash_memory_with_state`

See this comment for more information: https://github.com/0xPolygonMiden/miden-base/issues/1237#issuecomment-2745384217

PR #2395, opened 2025-11-27
Implement `hash_elements_with_padding` for `std::crypto::hashes::rpo`

Fumuran:
This PR does a refactoring of the `std::crypto::hashes::rpo` module:
- `prepare_hasher_state` and `hash_memory_with_state` procedures changed to be private
- `hash_elements` procedure was renamed to the `hash_elements_wo_padding` (name is still questionable)
-  new `hash_elements_w_padding` procedure was created.

So essentially public API of the `rpo` module was changed to have only two universal hashing procedures: `hash_elements_wo_padding` (old `hash_elements`) which hashes the memory block without padding the elements, and `hash_elements_w_padding` which does the same thing, but also pads the input values with zeros to the next multiple of 8.

Closes: #1719.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@
 - Consolidate debug information into `DebugInfo` struct ([#2366](https://github.com/0xMiden/miden-vm/issues/2366)).
 - Wrapped `hperm` instruction in `rpo::permute` procedure ([#2392](https://github.com/0xMiden/miden-vm/pull/2392)).
 - `hash_memory_with_state`, `hash_memory_words`, and `hash_memory_double_words` procedures from the `std::crypto::hashes::rpo` module were renamed to the `hash_elements_with_state`, `hash_words`, and `hash_double_words` respectively ([#2395](https://github.com/0xMiden/miden-vm/pull/2395)).
+- [BREAKING] Updgrade `miden-crypto` to 0.19 ([#2399](https://github.com/0xMiden/miden-vm/pull/2399)).
+- Pre-allocate main trace buffer in trace generation ([#2345](https://github.com/0xMiden/miden-vm/pull/2345)).
 
 ## 0.19.1 (2025-11-6)
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1413,13 +1413,14 @@ dependencies = [
 
 [[package]]
 name = "miden-crypto"
-version = "0.18.2"
+version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7eb82051002f9c64878d3b105a7b924de1ee92019231923380cf4ecd7b824f9a"
+checksum = "1162f8648a13e262c6d454ace2357fc1a301c404ca3111a5d2314e0ca286de0c"
 dependencies = [
  "blake3",
  "cc",
  "chacha20poly1305",
+ "curve25519-dalek",
  "ed25519-dalek",
  "flume",
  "glob",
@@ -1433,6 +1434,7 @@ dependencies = [
  "rand_core 0.9.3",
  "rand_hc",
  "serde",
+ "sha2",
  "sha3",
  "subtle",
  "thiserror",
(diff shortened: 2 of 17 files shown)


stdlib/asm/crypto/hashes/rpo.masm
```diff
@@ -117,52 +119,147 @@ end
...
+#! - ptr is the  memory address of the first element to be hashed. This address must be
+#!   word-aligned - i.e., divisible by 4.
+#! - num_elements is the number of elements to be hashed.
+#! - HASH is the resulting hash of the provided memory values.
+#!
+#! Cycles:
+#! - If number of elements divides by 8: 52 cycles + 3 * words
+#! - Else: 185 cycles + 3 * words
+#! where `words` is the number of quads of input values.
+pub proc hash_elements_wo_padding
```
Fumuran: IMO the main question is what name to choose for these procedures. First I wanted to keep `hash_elements_without_padding`, but it seems a bit too long. On the other hand, shortening "without" to "wo" could be confusing. The same works for the `hash_elements_with_padding`.

bobbinth:
Not a review yet, but I'm wondering if this is actually an improvement. The `hash_memory_with_state` procedure (which we should rename into `hash_elements_with_state`) is useful because it allows the user the control the initial state of the sponge. This control could come in hand for enabling custom domain separation or maybe even in encryption settings (cc @Al-Kindi-0 in case that's not useful there). So, I think there is a good argument for keeping `hash_elements_with_state` public.

I would also probably keep `hash_elements` procedure named as is because this is consistent with the naming we use in `miden-crypto`.

So, I think the question is whether we want to introduce a new procedure that does something like `hash_elements_with_padding` (though, I think a better name would be `pad_and_hash_elements`), and I think this depends on how we'd use it. I have actually forgotten how we use these procedures (I'm assuming this was done for some use case have in `miden-base`?)

Al-Kindi-0:
> Not a review yet, but I'm wondering if this is actually an improvement. The `hash_memory_with_state` procedure (which we should rename into `hash_elements_with_state`) is useful because it allows the user the control the initial state of the sponge. This control could come in hand for enabling custom domain separation or maybe even in encryption settings (cc @Al-Kindi-0 in case that's not useful there). So, I think there is a good argument for keeping `hash_elements_with_state` public.

I am not immediately seeing the application to encryption as in that case `hperm` and `crypto_stream` are interleaved and so we can't use generic hashing procedures.

Fumuran:
>So, I think the question is whether we want to introduce a new procedure that does something like `hash_elements_with_padding` (though, I think a better name would be `pad_and_hash_elements`), and I think this depends on how we'd use it.

Right, we need this procedure for the `miden-base`, for the `miden::note::compute_inputs_commitment` to be precise. Though it is the only place where we could potentially use this procedure.

So in that case it feels like the right approach will be to introduce this new procedure, leaving the `hash_elements` and  `prepare_hasher_state` + `hash_elements_with_state` unchanged.

processor/src/stack/overflow.rs
```diff
@@ -266,14 +266,14 @@ impl OverflowTable {
...
-                .iter()
-                .map(OverflowStackEntry::value)
-                .collect();
-
-            self.history.as_mut().unwrap().save_stack_to_history(clk, stack_after_op);
+        let stack_after_op: Vec<Felt> = self
+            .get_current_overflow_stack()
+            .iter()
+            .map(OverflowStackEntry::value)
+            .collect();
```
bobbinth, line 273: There is a discussion about how to properly address this in https://github.com/0xMiden/miden-vm/pull/2400 - so, for now, I'm not sure we should make this change.

stdlib/tests/crypto/rpo.rs
```diff
@@ -393,7 +393,7 @@ fn test_copy_digest() {
 }
 
 #[test]
-fn test_hash_elements() {
+fn test_hash_elements_without_padding() {
```
bobbinth: I think we can revert changes in this file.

stdlib/asm/crypto/hashes/rpo.masm
bobbinth: Let's rename this to `hash_elements_with_state`.

stdlib/asm/crypto/hashes/rpo.masm
```diff
@@ -117,50 +119,50 @@ end
 #! - odd words: 65 cycles + 3 * words
 #! where `words` is the `start_addr - end_addr - 1`
 pub proc hash_memory_words
```
bobbinth: Maybe we should rename this to `hash_words` to make the naming more consistent with `hash_elements`?

stdlib/asm/crypto/hashes/rpo.masm
```diff
@@ -84,24 +86,24 @@ end
 #!
 #! Cycles: 37 + 3 * words, where `words` is the `start_addr - end_addr`
 pub proc hash_memory_double_words
```
bobbinth: Similar to the previous comments, maybe this should be `hash_double_words`?

Issue #2357, opened 2025-11-10
Enable padding the advice stack when pushing advice map entry


PhilippGackstatter:
To pipe an advice map entry to memory and hash it, e.g. using `std::mem::pipe_double_words_to_memory`, the number of elements in the advice map entry must often be a multiple of 8.
This is not always convenient, particularly when inserting advice entries from MASM itself. In the Miden protocol, this is relevant for note inputs (arbitrary user-data in a note), which can be arbitarily sized. So, when a user inserts two note inputs into an advice map entry, it would be convenient to pad it to 8 felts for hashing. Padding them in the advice map entry itself doesn't quite work, as the original number of note inputs should remain accessible when inspecting that entry at a later point, and padding it would make that impossible.

For that reason, it would be convenient to have an instruction like `adv.push_mapvaln`, that pushes an advice map entry onto the advice stack _and_ pads the entry to the next multiple of a parameter `p`.

## Pad Parameter

This parameter could simply work like this:

- If `p == 0`, the behavior is identical to the current behavior.
- If `p != 0`, then the instructions pushes the advice map entry with zero pad elements to the advice stack. The total number of pushed entries is `num_elements_in_entry.next_multiple_of(p)`.
  - So if the advice map entry has 11 elements, and `p = 8`, then 5 zeros plus the original 11 elements are pushed to the stack, such that the zeros are in the deeper part of the stack. This means the advice stack is setup to work correctly with `adv_pipe`.

It probably makes sense to restrict `p` to be within a small range of allowed values, perhaps `0..=8`, or maybe even just one of `0`, `4` and `8`?

## Instruction

Some discussion might be needed for how this would look like exactly at the instruction level, and these are potential options.

## Parameterize existing instructions

One idea is to parametrize the existing instruction, e.g. `adv.push_mapvaln.p` so that it takes an optional parameter `p`.
I like this option because this is effectively just a customization of `adv.push_mapvaln`'s behavior, and an optional parameter seems best to do this.

## Add new instruction

Another approach is to introduce a new instruction `adv.push_padded_mapvaln.p` or `adv.push_mapvaln_padded.p`. This could assume a reasonable default value for `p`, like 8, unless one is explicitly specified. This is a more explicit option.

## adv.push_mapval

[...]

PR #2398, opened 2025-11-28
Implement padding option for `adv.push_mapvaln`

Fumuran:
This PR expands the capabilities of the `adv.push_mapvaln` instruction by adding an option to use padding as an immediate value: `adv.push_mapvaln.padding`.

If the padding value is provided, the list of elements obtained from the advice map got padded with values so that its length will become a next multiple of the provided padding. For example, if the map holds `[1, 2, 3, 4, 5]` list, the advice stack after the `adv.push_mapvaln.8` instruction will look like so: `[5, 1, 2, 3, 4, 5, 0, 0, 0]`.

The final implementation turned out to be a hybrid of the approaches described in the [issue](https://github.com/0xMiden/miden-vm/issues/2357#issuecomment-3587126277). So instead of creating a new instruction, the existing `adv.push_mapvaln` was changed, though only few options for the padding value are supported: 0, 4, and 8. This restriction is dictated by the way we handle system events: we have to have a _constant_ identifier, representing the event. Using this structure, it is hard to implement a variable padding option for the `SystemEvent` enum variant (for the `MapValueToStackN`). Current implementation just adds two new variants to the `SystemEvent`, which represent the `adv.push_mapvaln` instruction with padding 4 and 8.

Closes: #2357


docs/src/user_docs/assembly/io_operations.md
```diff
@@ -56,16 +56,16 @@ System events fall into two categories: (1) events which push new data onto the
...
+| adv.push_mapvaln <br /> adv.push_mapvaln._p_ | [K, ... ]          | [K, ... ]          | Pushes a list of field elements together with the number of elements onto the advice stack (`[n, ele1, ele2, ...]`, where `n` is the number of elements pushed). The list is looked up in the advice map using word $K$ as the key. <br /><br /> If padding _p_ is provided as an immediate value, the list of field elements obtained from the advice map will be padded with zeros, increasing its length to the next multiple of _p_, excluding _num_values_, e.g. `[5, 1, 2, 3, 4, 5, 0, 0, 0]`. _num_values_ in that case will show the initial number if values in the list. <br /> Valid options for the padding value are $0$, $4$, and $8$. Using $0$ explicitly shows that no padding will be added. If no padding provided, it is assumed to be $0$. <br /><br /> Fails if $p \notin \{0, 4, 8\}$ |
```
bobbinth: The phrase "where `n` is the number of elements pushed" is a bit misleading because, if I understood correctly, `n` is the number of elements in the map value which could be different from the number of elements pushed onto the advice stack.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -286,12 +288,39 @@ fn copy_map_value_to_adv_stack(
     process: &mut ProcessState,
     include_len: bool,
     err_ctx: &impl ErrorContext,
+    pad_value: u8,
 ) -> Result<(), ExecutionError> {
```
bobbinth, line 310: nit: I would keep the `err_ctx` parameter as the last one in the parameter list.

Also, maybe I'd rename `pad_value` to `pad_to`.

processor/src/operations/sys_ops/sys_event_handlers.rs
```diff
@@ -286,12 +288,39 @@ fn copy_map_value_to_adv_stack(
...
+    // reverse the map_value to preserve the stack order
+    map_value_vec.reverse();
+
+    // push the padded and reversed map_value and its initial length to the advice stack
+    let advice_provider_mut = process.advice_provider_mut();
+    advice_provider_mut.extend_stack(map_value_vec);
+    if include_len {
+        advice_provider_mut
+            .push_stack(Felt::try_from(map_value_len as u64).expect("value length too big"));
+    }
```
bobbinth: I would probably move most of this code into `AdviceProvider::push_from_map()` and would just add `pad_to` as a parameter for this function.

bobbinth (commented):
Looks good! Thank you! I left a few small comments inline.

crates/assembly-syntax/src/parser/error.rs
```diff
@@ -339,6 +339,13 @@ pub enum ParsingError {
         span: SourceSpan,
         message: String,
     },
+    #[error("invalid padding value for the `adv.push_mapvaln` instruction: {padding}")]
+    #[diagnostic(help("valid padding values are limited to 0, 4, and 8"))]
```
PhilippGackstatter:
```suggestion
    #[diagnostic(help("valid padding values are 0, 4, and 8"))]
```

Nit: Slightly simplified

docs/src/user_docs/assembly/io_operations.md
```diff
@@ -56,16 +56,16 @@ System events fall into two categories: (1) events which push new data onto the
...
+| adv.push_mapvaln <br /> adv.push_mapvaln._p_ | [K, ... ]          | [K, ... ]          | Pushes a list of field elements together with the number of elements onto the advice stack (`[n, ele1, ele2, ...]`, where `n` is the number of elements in the list). The list is looked up in the advice map using word $K$ as the key. <br /><br /> If padding _p_ is provided as an immediate value, the list of field elements obtained from the advice map will be padded with zeros, increasing its length to the next multiple of _p_, excluding _num_values_, e.g. `[5, 1, 2, 3, 4, 5, 0, 0, 0]`. _num_values_ in that case will show the initial number if values in the list. <br /> Valid options for the padding value are $0$, $4$, and $8$. Using $0$ explicitly shows that no padding will be added. If no padding provided, it is assumed to be $0$. <br /><br /> Fails if $p \notin \{0, 4, 8\}$ |
```
PhilippGackstatter:
```suggestion
| adv.push_mapvaln <br /> adv.push_mapvaln._p_ | [K, ... ]          | [K, ... ]          | Pushes a list of field elements together with the number of elements onto the advice stack (`[n, ele1, ele2, ...]`, where `n` is the number of elements in the list). The list is looked up in the advice map using word $K$ as the key. <br /><br /> If padding _p_ is provided as an immediate value, the list of field elements obtained from the advice map will be padded with zeros, increasing its length to the next multiple of _p_, excluding _num_values_, e.g. `[5, 1, 2, 3, 4, 5, 0, 0, 0]`. _num_values_ in that case will be the initial number of values in the list. <br /> Valid options for the padding value are $0$, $4$, and $8$. Using $0$ explicitly shows that no padding will be added. If no padding is provided, it is assumed to be $0$. <br /><br /> Fails if $p \notin \{0, 4, 8\}$ |
```

processor/src/host/advice/mod.rs
```diff
@@ -139,28 +139,50 @@ impl AdviceProvider {
...
+        // get the advice map value
+        let map_value = self.map.get(&key).ok_or(AdviceError::MapKeyNotFound { key })?;
+
+        // if pad_to was provided (not equal 0), resize the map_value vector to the next multiple of
+        // pad_to
+        let mut map_value_vec = map_value.to_vec();
+        if pad_to != 0 {
+            map_value_vec
+                .resize_with(map_value.len().next_multiple_of(pad_to as usize), Felt::default);
+        }
```
PhilippGackstatter, line 177: I like this for its clarity, but since this is used in the processor and somewhat time-critical, I think it might be worth optimizing this a little bit, specifically getting rid of allocating the map entry as a vector (`to_vec`), which could potentially be large. Not sure if it's worth it, others might have a better sense of this.

Would this work?

```rust
let map_value = self.map.get(&key).ok_or(AdviceError::MapKeyNotFound { key })?;

if pad_to != 0 {
    let num_pad_elements = map_value.len().next_multiple_of(pad_to as usize) - map_value.len();
    self.stack.extend(core::iter::repeat_n(Felt::default(), num_pad_elements));
}
self.stack.extend(map_value.iter().rev());

```

Fumuran: Right, makes sense!

processor/src/stack/overflow.rs
```diff
@@ -266,14 +266,14 @@ impl OverflowTable {
...
-            self.history.as_mut().unwrap().save_stack_to_history(clk, stack_after_op);
+        let stack_after_op: Vec<Felt> = self
+            .get_current_overflow_stack()
+            .iter()
+            .map(OverflowStackEntry::value)
+            .collect();
+
+        if let Some(history) = &mut self.history {
+            history.save_stack_to_history(clk, stack_after_op);
         }
```
bobbinth, line 277: I'll merge this as is: while it may not be the optimal approach, I don't think it matters:

- I believe this function is used only in debug mode where we don't care about the performance too much.
- This is a part of the "slow processor" which, hopefully, will go away soon.

If the above is incorrect, we can revert this change later on.

bobbinth (approved):
Looks good! Thank you!

PR #2399, opened 2025-12-01
chore: Upgrade `miden-crypto` to 0.19

adr1anh:


huitseeker (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 - [BREAKING] Standardized hash function naming: renamed `hash_2to1` → `merge` and `hash_1to1` → `hash` across all hash modules (blake3, sha256, keccak256, rpo) ([#2381](https://github.com/0xMiden/miden-vm/pull/2381)).
 - Consolidate debug information into `DebugInfo` struct ([#2366](https://github.com/0xMiden/miden-vm/issues/2366)).
 - Wrapped `hperm` instruction in `rpo::permute` procedure ([#2392](https://github.com/0xMiden/miden-vm/pull/2392)).
+- Updgrade `miden-crypto` to 0.19 ([#2399](https://github.com/0xMiden/miden-vm/pull/2399)).
```
bobbinth: Should this be marked as a breaking change?

core/src/lib.rs
```diff
@@ -84,7 +86,8 @@ pub mod crypto {
     }
 
     pub mod dsa {
-        pub use miden_crypto::dsa::rpo_falcon512;
+        // Upstream renamed module; keep our public API stable
+        pub use miden_crypto::dsa::falcon512_rpo as rpo_falcon512;
```
bobbinth: This may be fine for this PR, but v0.20.0 release will introduce a lot of breaking changes, and making this be a breaking change is probably also fine.

If we don't do it in this PR, we should do a follow-up PR with the rename.

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline.

core/src/lib.rs
adr1anh: I agree and made the change here since it will minimize the number of changes in the next release. However, should we also export the other signature modules?

PR #2400, opened 2025-12-01
chore: fix lint

adr1anh:
This seems to be a bug with clippy since the suggestion triggers a borrow checker error.

Note: we use the new `#[expect(clippy::some_warning)]` syntax which will trigger a warning in the future if this lint is fixed.

### DIFF
--- a/processor/src/stack/overflow.rs
+++ b/processor/src/stack/overflow.rs
@@ -266,6 +266,7 @@ impl OverflowTable {
     /// and `restore_context` functions.
     fn save_stack_to_history(&mut self) {
         let clk = self.clk;
+        #[expect(clippy::unnecessary_unwrap)]
         if self.history.is_some() {
             let stack_after_op: Vec<Felt> = self
                 .get_current_overflow_stack()


adr1anh:
That works too but allocates the vector at every call even if `history == None`. I’m not sure if that's an issue but I'll let @plafer chime in.

plafer:
@huitseeker's solution was applied in #2398; though I agree with @adr1anh that we're allocating even when history is off.

Note though that the `OverflowTable` history should be considered deprecated; it was only used by `Process` for building a debug trace. Hence at this point I'd probably just create an issue to remove the `OverflowTable` history when we remove `Process` (to make sure we don't forget), in which case this issue becomes moot.

plafer:
Closing as discussed offline; included a note to remove the `OverflowTable` history in #2406

huitseeker:
FWIW, the solution I suggested is from #2396 (not #2398) and was not applied to next, since that PR is still open and is generating some discussion — it's going to be a while.
 
I'm in fact OK with merging this PR, because I would like to prioritize having `next` fixed quickly, over optimizing allocation in code that we know is deprecated and will be removed shortly. 

@plafer would appreciate a stamp if this makes sense to you

huitseeker:
Oh, my bad, that was indeed merged 🤦 Don't mind me.

PR #2403, opened 2025-12-02
feat: add PackageKind field to Package struct

### LINKED ISSUE #2402: Store project kind (account, note script, tx script, etc.) in the Miden package header
From https://github.com/0xMiden/miden-base/issues/2111#issuecomment-3589985970

> Consider putting the project kind in the Miden package header along with the name, version, etc. So far I see the following kinds: account, note script, tx script, auth component, generic program.

@bobbinth @bitwalker

Farukest:
Adds a `kind` field to `Package` to indicate the project type (Executable, AccountComponent, NoteScript, TxScript, AuthComponent).

  Closes #2402

  ## Describe your changes
  - Added `PackageKind` enum with 5 variants: Executable, AccountComponent, NoteScript, TxScript, AuthComponent
  - Added `kind` field to `Package` struct
  - Updated serialization format (version 2.0.0 → 3.0.0)
  - Added unit tests for `PackageKind`
  
## Rationale
Adds a `kind` field to `Package` to indicate the project type (Library, Executable, AccountComponent, NoteScript, TransactionScript). This helps consumers understand how to use the package.

## Test plan
- All existing tests pass
- New serde roundtrip test added via `miden_test_serde_macros::serde_test`

  - [x] Updated `CHANGELOG.md`

### DIFF
--- a/crates/assembly/src/tests.rs
+++ b/crates/assembly/src/tests.rs
@@ -14,7 +14,9 @@ use miden_core::{
     },
     utils::{Deserializable, Serializable},
 };
-use miden_mast_package::{MastArtifact, MastForest, Package, PackageExport, PackageManifest};
+use miden_mast_package::{
+    MastArtifact, MastForest, Package, PackageExport, PackageKind, PackageManifest,
+};
 use proptest::{
     prelude::*,
     test_runner::{Config, TestRunner},
@@ -3466,7 +3468,12 @@ prop_compose! {
 
         let manifest = PackageManifest::new(exports).with_dependencies(manifest.dependencies().cloned());
 
-        Package { name, version: None, description: None, mast, manifest, sections: Default::default() }
+        let kind = match &mast {
+            MastArtifact::Executable(_) => PackageKind::Executable,
+            MastArtifact::Library(_) => PackageKind::Library,
+        };
+
+        Package { name, version: None, description: None, kind, mast, manifest, sections: Default::default() }
     }
 }
 
--- a/crates/mast-package/Cargo.toml
+++ b/crates/mast-package/Cargo.toml
@@ -19,7 +19,7 @@ doctest = false
 
 [features]
 default = ["std", "serde"]
-arbitrary = ["dep:proptest-derive", "dep:proptest", "miden-assembly-syntax/arbitrary"]
+arbitrary = ["std", "dep:proptest-derive", "dep:proptest", "miden-assembly-syntax/arbitrary"]
 std = ["miden-assembly-syntax/std", "miden-core/std", "serde/std", "thiserror/std"]
 serde = ["dep:serde", "dep:serde-untagged", "miden-assembly-syntax/serde", "miden-core/serde"]
 
--- a/crates/mast-package/src/lib.rs
+++ b/crates/mast-package/src/lib.rs
@@ -4,12 +4,12 @@
 
 extern crate alloc;
 
+#[cfg(any(test, feature = "std"))]
+extern crate std;
+
 mod artifact;
 mod dependency;
 mod package;
-
-#[cfg(test)]
-extern crate std;
 
 pub use miden_assembly_syntax::{
     Library, PathBuf,
@@ -27,7 +27,8 @@ pub use self::{
         },
     },
     package::{
-        ConstantExport, InvalidSectionIdError, Package, PackageExport, PackageManifest,
-        ProcedureExport, Section, SectionId, TypeExport, Version, VersionError,
+        ConstantExport, InvalidPackageKindError, InvalidSectionIdError, Package, PackageExport,
+        PackageKind, PackageManifest, ProcedureExport, Section, SectionId, TypeExport, Version,
+        VersionError,
     },
 };
--- a/crates/mast-package/src/package/mod.rs
+++ b/crates/mast-package/src/package/mod.rs
@@ -1,3 +1,4 @@
+mod kind;
 mod manifest;
 mod section;
 mod serialization;
@@ -11,6 +12,7 @@ use miden_core::{Program, Word};
 use serde::{Deserialize, Serialize};
 
 pub use self::{
+    kind::{InvalidPackageKindError, PackageKind},
     manifest::{ConstantExport, PackageExport, PackageManifest, ProcedureExport, TypeExport},
     section::{InvalidSectionIdError, Section, SectionId},
 };
@@ -31,6 +33,8 @@ pub struct Package {
     /// An optional description of the package
     #[cfg_attr(feature = "serde", serde(default))]
     pub description: Option<String>,
+    /// The kind of project that produced this package.
+    pub kind: PackageKind,
     /// The MAST artifact ([Program] or [Library]) of the package
     pub mast: MastArtifact,
     /// The package manifest, containing the set of exported procedures and their signatures,
@@ -107,6 +111,7 @@ impl Package {
                 name: self.name.clone(),
                 version: self.version.clone(),
                 description: self.description.clone(),
+                kind: PackageKind::Executable,
                 mast: MastArtifact::Executable(Arc::new(Program::new(
                     library.mast_forest().clone(),
                     node_id,
--- a/crates/mast-package/src/package/serialization.rs
+++ b/crates/mast-package/src/package/serialization.rs
@@ -8,6 +8,7 @@
 //! - `name` (`String`)
 //! - `version` (optional, [`miden_assembly_syntax::Version`] serialized as a `String`)
 //! - `description` (optional, `String`)
+//! - `kind` (`u8`, see [`crate::PackageKind`])
 //!
 //! #### Code
 //! - `mast` (see [`crate::MastArtifact`])
@@ -36,7 +37,7 @@ use miden_core::{
     utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
 };
 
-use super::{ConstantExport, ProcedureExport, TypeExport};
+use super::{ConstantExport, PackageKind, ProcedureExport, TypeExport};
 use crate::{Dependency, MastArtifact, Package, PackageExport, PackageManifest, Section};
 
 // CONSTANTS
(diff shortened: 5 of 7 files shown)


crates/assembly/src/tests.rs
```diff
@@ -3466,7 +3468,7 @@ prop_compose! {
 
         let manifest = PackageManifest::new(exports).with_dependencies(manifest.dependencies().cloned());
 
-        Package { name, version: None, description: None, mast, manifest, sections: Default::default() }
+        Package { name, version: None, description: None, kind: PackageKind::Executable, mast, manifest, sections: Default::default() }
```
bitwalker: The `kind` here should be determined based on whether the `MastArtifact` was an `Executable` or `Library`, i.e. we don't want the kind to be `Executable` for a `Library` package when generating arbitrary packages.

crates/mast-package/src/package/mod.rs
```diff
@@ -16,6 +17,60 @@ pub use self::{
 };
 use crate::MastArtifact;
 
+// PACKAGE KIND
```
bitwalker: Let's move this to a separate source file in the same directory, i.e. `package/kind.rs`

crates/mast-package/src/package/mod.rs
```diff
@@ -16,6 +17,60 @@ pub use self::{
...
+    NoteScript = 2,
+    /// A transaction script package.
+    TxScript = 3,
+    /// An authentication component package.
+    AuthComponent = 4,
+}
+
+impl PackageKind {
+    /// Returns the numeric tag for this package kind.
+    pub const fn tag(&self) -> u8 {
```
bitwalker: Let's remove this function - we only define `tag` when the discriminant is non-trivial to safely obtain (i.e. when the enum has struct/tuple variants).

crates/mast-package/src/package/mod.rs
```diff
@@ -16,6 +17,60 @@ pub use self::{
...
+        *self as u8
+    }
+
+    /// Creates a PackageKind from its numeric tag.
+    pub const fn from_tag(tag: u8) -> Option<Self> {
```
bitwalker: Let's provide this functionality via a `TryFrom<u8>` impl instead.

crates/mast-package/src/package/mod.rs
```diff
@@ -16,6 +17,60 @@ pub use self::{
...
 
+// PACKAGE KIND
+// ================================================================================================
+
+/// The kind of project that produced this package.
+///
+/// This helps consumers of the package understand how to use it (e.g., as an account component,
+/// a note script, a transaction script, etc.).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
bitwalker: For this, I think it would make sense to _not_ derive `Serialize` and `Deserialize`, and instead provide hand-written impls that expect a string representation (corresponding to the `Display` output) when the serializer/deserializer is human-readable, and expect the `u8` representation when the format is _not_ human-readable.

The default `derive` implementation of `Serialize`/`Deserialize` for enums is never ideal, so this is one place where avoiding `derive` of these traits makes sense.

crates/mast-package/src/package/mod.rs
```diff
@@ -16,6 +17,60 @@ pub use self::{
...
+#[repr(u8)]
+pub enum PackageKind {
```
bitwalker: We also need a `Library` kind, for generic libraries of code.

crates/mast-package/src/package/mod.rs
```diff
@@ -127,3 +185,46 @@ impl Package {
         }
     }
 }
+
+// TESTS
+// ================================================================================================
+
+#[cfg(test)]
+mod tests {
```
bitwalker: This can be removed in favor of decorating `PackageKind` with `#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]` and `#[cfg_attr(feature = "arbitrary", derive(proptest_derive::Arbitrary)]`. It tests the same thing, but generates the implementation rather than requiring us to maintain it by hand.

crates/mast-package/src/package/serialization.rs
```diff
@@ -54,7 +55,8 @@ const MAGIC_LIBRARY: &[u8; 4] = b"LIB\0";
 /// The format version.
 ///
 /// If future modifications are made to this format, the version should be incremented by 1.
-const VERSION: [u8; 3] = [2, 0, 0];
+/// Version 3.0.0: Added `kind` field to Package.
```
bitwalker: I'd move this to a `CHANGELOG` entry, or remove it entirely.

crates/mast-package/src/package/mod.rs
bobbinth: Should we also make the enum non-exhaustive? (so that adding more variants later on does not result in breaking changes)

Also, I'm not 100% sure if we need a separate variant for `AuthComponent` - @PhilippGackstatter @mmagician - what do you think?

Overall, I'd order these as follows:
```rust
pub enum PackageKind {
    Library = 0,
    Executable = 1,
    AccountComponent = 2,
    NoteScript = 3,
    TransactionScript = 4,
}
```

PhilippGackstatter: I think even if we don't go all the way with https://github.com/0xMiden/miden-base/issues/1736 and implement `AuthenticationScript`, having an `AccountAuthComponent` does make sense. It should make it easier to enforce in the `AccountBuilder` that exactly one auth component is provided and handle auth components separately from regular ones. So it probably makes sense to already add `PackageKind::AccountAuthComponent`.

bobbinth: My thinking was that we'd be able to relatively easily detect if a component is an auth component or not by scanning through the procedures of the component and checking for the right attribute. We'll probably need to do this anyway because on deserialization we won't know for if sure the component does in fact have an auth procedure. So, the benefit of adding a dedicated kind for this seems a bit questionable.

PhilippGackstatter: That's true. I guess I was thinking on a higher level.
Then it depends on if we end up defining `AccountAuthComponent` as a `Library` or a combination of `Executable` (auth script) + `Library` (auth library), in which case it would be serialized differently from a regular account component. Until that's decided, may be best to hold off on adding that `PackageKind` then.

Farukest:
All done @bitwalker @bobbinth

crates/mast-package/src/package/mod.rs
```diff
@@ -31,6 +33,8 @@ pub struct Package {
     /// An optional description of the package
     #[cfg_attr(feature = "serde", serde(default))]
     pub description: Option<String>,
+    /// The kind of project that produced this package.
+    pub kind: PackageKind,
     /// The MAST artifact ([Program] or [Library]) of the package
     pub mast: MastArtifact,
```
bobbinth, line 39: I guess another question is how to deal with potential inconsistencies between `PackageKind` and `MastArtifact` fields. For example, we don't want to set `PackageKind::Library` but have `MastArtifact::Executable`.

We could make the fields private and then enforce consistency in a constructor - but that may be too much of a change. cc @bitwalker.

crates/mast-package/src/package/mod.rs
```diff
@@ -107,6 +111,7 @@ impl Package {
                 name: self.name.clone(),
                 version: self.version.clone(),
                 description: self.description.clone(),
+                kind: self.kind,
```
bobbinth: Related to the above point, we probably shouldn't just copy over the kind. For example, if we are changing library into an executable, the kind should also become an executable.

In the context of other kinds, I'm not sure this operation even make sense. For example, we can't really convert `AccountComponent` into an executable.

So, I wonder if we should actually provide `make_executable()` function in its current form. Maybe what we actually want is the ability to change the kind of the package where some transitions are possible and others are not. But also, do we actually use this functionality anywhere?

crates/mast-package/src/package/mod.rs
bitwalker: We could probably provide two constructors, one for executables, and one for everything else, along with various builder methods, but not sure how big of a problem it is in practice that these can be mismatched. At worst, we'd raise an error when this is detected at the time of use - in practice, we aren't ever going to produce packages with such a mismatch, so it is of limited utility IMO, but I don't have strong feelings about it.

crates/mast-package/src/package/mod.rs
bitwalker: It's used by `miden-debug` when debugging exports of a library (i.e. you can load a package and specify an arbitrary entrypoint function, and we convert the library package into an executable one where the executable entrypoint is the specified function).

That said, we could probably move the functionality to `miden-debug`, since I believe it is the only consumer of this API currently.

Farukest:
@bitwalker is everything okay ?
as I see; no need any changes for now. right ?

crates/mast-package/Cargo.toml
```diff
@@ -19,7 +19,7 @@ doctest = false
 
 [features]
 default = ["std", "serde"]
-arbitrary = ["dep:proptest-derive", "dep:proptest", "miden-assembly-syntax/arbitrary"]
+arbitrary = ["std", "dep:proptest-derive", "dep:proptest", "miden-assembly-syntax/arbitrary"]
```
bitwalker, line 22: Do you know why this was necessary? Is something generated by the derived impl using libstd?

crates/mast-package/src/lib.rs
```diff
@@ -4,13 +4,13 @@
 
 extern crate alloc;
 
+#[cfg(feature = "std")]
```
bitwalker: This should be `#[cfg(any(test, feature = "std")]`

crates/mast-package/src/package/mod.rs
bitwalker: @Farukest  could you ensure the kind here is `Executable`?

crates/mast-package/Cargo.toml
Farukest: Yes, `proptest` and `proptest-derive` crates require `std` by default. When deriving `Arbitrary` on `PackageKind`, the generated code uses `proptest::strategy::Strategy` which depends on `std`. Without adding `std` to the `arbitrary` feature, compilation fails in `no_std` environments with the `arbitrary` feature enabled.

Farukest:
> @Farukest could you ensure the kind here is Executable?

Done @bitwalker , changed `self.kind` to `PackageKind::Executable`

crates/mast-package/src/lib.rs
Farukest: Done

PR #2408, opened 2025-12-02
feat: Forest compaction

### LINKED ISSUE #2368: feat: set up `MastForest` compaction
Comment for the future: once decorators are stripped, we'll also need to run a "forest compaction" procedure because we may be able to eliminate some nodes which previously differed only by decorators. This may already be captured in one of the issues, but if not, let's note this.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2360#discussion_r2523058880_

huitseeker:
After #2108, we have gained the ability to remove *all* decorators from the nodes of a `MastForest`. As a consequence of this sort of removal, some nodes which would previously have been considered distinct because of having different decorator information (but who would be otherwise identical) can now share the same representation. This begs for a compaction function on `MastForest`s that achieves this sharing, which is what we introduce here.

Fixes #2368.

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -175,18 +175,22 @@ impl MastForest {
     /// forest.strip_decorators();
     ///
     /// // Then compact the forest
-    /// forest.compact().unwrap();
+    /// let root_map = forest.compact();
     ///
     /// // Forest is now compacted with duplicate nodes merged
     /// ```
-    pub fn compact(&mut self) -> Result<MastForestRootMap, MastForestError> {
+    pub fn compact(&mut self) -> MastForestRootMap {
         // Merge with itself to deduplicate nodes
-        let (compacted_forest, root_map) = MastForest::merge([&*self])?;
+        // Note: This cannot fail for a self-merge under normal conditions.
+        // The only possible failures (TooManyNodes, TooManyDecorators) would require the
+        // original forest to be at capacity limits, at which point compaction wouldn't help.
+        let (compacted_forest, root_map) = MastForest::merge([&*self])
+            .expect("Failed to compact MastForest: this should never happen during self-merge");
 
         // Replace current forest with compacted version
         *self = compacted_forest;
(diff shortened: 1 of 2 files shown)


core/src/mast/mod.rs
```diff
@@ -133,11 +133,62 @@ impl MastForest {
...
+    /// ```
+    pub fn compact(&mut self) -> Result<(), MastForestError> {
+        // Merge with itself to deduplicate nodes
+        let (compacted_forest, _root_map) = MastForest::merge([&*self])?;
+
+        // Replace current forest with compacted version
+        *self = compacted_forest;
+
+        Ok(())
+    }
```
plafer: We should return the `root_map` here, in case the user needs to update them somewhere

core/src/mast/mod.rs
```diff
@@ -133,11 +133,62 @@ impl MastForest {
...
+    /// forest.strip_decorators();
+    ///
+    /// // Then compact the forest
+    /// forest.compact().unwrap();
+    ///
+    /// // Forest is now compacted with duplicate nodes merged
+    /// ```
+    pub fn compact(&mut self) -> Result<MastForestRootMap, MastForestError> {
+        // Merge with itself to deduplicate nodes
+        let (compacted_forest, root_map) = MastForest::merge([&*self])?;
```
bobbinth: Under what conditions can merging a forest with itself fail? I'm wondering if we can replace this with `expect()` and avoid returning the result.

bobbinth (approved):
Looks good! Thank you! I left just one small question inline.

core/src/mast/mod.rs
huitseeker: Ah, you're right. `TooManyNodes` or `TooManyDecorators`/`DecoratorIdOverflow` is impossible since merging never grows the number of nodes, `AdviceMapKeyCollisionOnMerge` has nothing to collide with, good catch.

PR #2412, opened 2025-12-04
feat: implement MastForest validation

### LINKED ISSUE #2106: create a `MastForest::validate` for validating structure of MastForest elements
Now that the processor no longer enforces the padding NOOP invariants, we should add a `MastForest::validate()` method that ensures that the `MastForest` is properly built (including but not limited to the padding NOOP invariants). We can then document in the `FastProcessor::execute_*()` methods that it is the responsibility of the caller to ensure that the program's `MastForest` is valid.

Similarly for `MastForest`s that we get from the host when executing an external node, we should offer the option to validate those or not (probably through some `ExecutionOptions` struct we pass to the execute methods).

This should be done as a follow-up PR though.

_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2094#discussion_r2307592884_ (more context in that thread)

huitseeker:
Add validation methods to enforce three core invariants for BasicBlockNode batches:

1. Power-of-two groups: Each batch must have a power-of-two number of groups
2. No immediate endings: No batch should end with an immediate operation
3. ~~Minimal NOOP padding: No batch where second half is all NOOPs and first half is valid~~

This addresses the gap created by PR 2094, where padding NOOPs moved from execution time to assembly time. The validation ensures invariants previously enforced by the processor can now be explicitly checked at any time post assembly (e.g. on deserialized forests).

Changes:
- Extend MastForestError enum with InvalidBatchPadding variant
- Add BasicBlockNode::validate_batch_invariants() method
- Add MastForest::validate() method to validate all nodes
- Add property tests using Arbitrary implementation for stress-testing

Fixes #2106

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -606,8 +606,10 @@ impl MastForest {
 impl MastForest {
     /// Validates that all BasicBlockNodes in this forest satisfy the core invariants:
     /// 1. Power-of-two number of groups in each batch
-    /// 2. No batch ends with an immediate operation
-    /// 3. OpBatch structural consistency (indptr integrity, bounds checking)
+    /// 2. No operation group ends with an operation requiring an immediate value
+    /// 3. The last operation group in a batch cannot contain operations requiring immediate values
+    /// 4. OpBatch structural consistency (num_groups <= BATCH_SIZE, group size <= GROUP_SIZE,
+    ///    indptr integrity, bounds checking)
     ///
     /// This addresses the gap created by PR 2094, where padding NOOPs are now inserted
     /// at assembly time rather than dynamically during execution, and adds comprehensive
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -50,12 +50,13 @@ pub const BATCH_SIZE: usize = 8;
 /// created according to these rules:
 ///
 /// - A basic block contains one or more batches.
-/// - A batch contains exactly 8 groups.
-/// - A group contains exactly 9 operations or 1 immediate value.
+/// - A batch contains up to 8 groups, and the number of groups must be a power of 2.
+/// - A group contains up to 9 operations or 1 immediate value.
+/// - Last operation in a group cannot be an operation that requires an immediate value.
 /// - NOOPs are used to fill a group or batch when necessary.
 /// - An immediate value follows the operation that requires it, using the next available group in
-///   the batch. If there are no batches available in the group, then both the operation and its
-///   immediate are moved to the next batch.
+///   the batch. If there are no groups available in the batch, then both the operation and its
+///   immediate value are moved to the next batch.
 ///
 /// Example: 8 pushes result in two operation batches:
 ///
@@ -423,8 +424,9 @@ impl BasicBlockNode {
 impl BasicBlockNode {
     /// Validates that this BasicBlockNode satisfies the core invariants:
     /// 1. Power-of-two number of groups in each batch
-    /// 2. No batch ends with an immediate operation
-    /// 3. OpBatch structural consistency
+    /// 2. No operation group ends with an operation requiring an immediate value
+    /// 3. The last operation group in a batch cannot contain operations requiring immediate values
+    /// 4. OpBatch structural consistency (num_groups <= BATCH_SIZE, group size <= GROUP_SIZE)
     ///
     /// Returns an error string describing which invariant was violated if validation fails.
     pub fn validate_batch_invariants(&self) -> Result<(), String> {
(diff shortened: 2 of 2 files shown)


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -469,6 +469,92 @@ impl MastNodeErrorContext for BasicBlockNode {
...
+        self.validate_minimal_noop_padding()?;
+
+        Ok(())
+    }
+
+    /// Validates that each batch has a power-of-two number of groups.
+    fn validate_power_of_two_groups(&self) -> Result<(), String> {
+        for (batch_idx, batch) in self.op_batches.iter().enumerate() {
+            let num_groups = batch.num_groups();
+            if !num_groups.is_power_of_two() && num_groups > 0 {
```
plafer:
```suggestion
            if !num_groups.is_power_of_two() {
```

IIRC empty batches (`num_groups == 0`) should be disallowed, right?

core/src/mast/mod.rs
```diff
@@ -740,6 +765,8 @@ pub enum MastForestError {
     AdviceMapKeyCollisionOnMerge(Word),
     #[error("decorator storage error: {0}")]
     DecoratorError(DecoratorIndexError),
+    #[error("invalid batch padding in node {0:?}: {1}")]
```
plafer:
```suggestion
    #[error("invalid batch in basic block node {0:?}: {1}")]
```

Would this be a better error message?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -469,6 +469,92 @@ impl MastNodeErrorContext for BasicBlockNode {
...
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    /// Validates that NOOP padding is minimal.
+    /// i.e. no batch should have its second half as all NOOPs if the first half would be a valid
+    /// batch.
+    fn validate_minimal_noop_padding(&self) -> Result<(), String> {
```
plafer: We allow users to write NOOPs, so technically `noop noop ... noop` should be a valid block. Unless I'm missing something, we shouldn't assume that NOOPs are only padding, and hence we shouldn't run this check?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -469,6 +469,92 @@ impl MastNodeErrorContext for BasicBlockNode {
...
+// ================================================================================================
+
+impl BasicBlockNode {
+    /// Validates that this BasicBlockNode satisfies the three core invariants:
+    /// 1. Power-of-two number of groups in each batch
+    /// 2. No batch ends with an immediate operation
+    /// 3. Minimal NOOP padding (no extraneous NOOPs)
+    ///
+    /// Returns an error string describing which invariant was violated if validation fails.
+    pub fn validate_batch_invariants(&self) -> Result<(), String> {
```
plafer: This indeed recovers the checks that previously were done at runtime by the processor. But I don't think we have any methods that check "full consistency" of nodes (to be run e.g. on deserialization). For example, we should check when deserializing an `OpBatch` that `indptr` is correctly formed (and won't cause us to crash due to out of bounds slice accesses).

Unless this is already captured somewhere in an issue, we should create a new one.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: OK, so the mindset is an adversarial but minimalistic check for a user-provided (probably serialized) payload, got it. That wasn't clear to me when you requested this, I thought it was more of an internal developer check.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: I've added bounds-checks in the last commit!

plafer (approved):
LGTM!

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -417,6 +417,103 @@ impl BasicBlockNode {
...
+                    "Batch {}: {} groups is not power of two",
+                    batch_idx, num_groups
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    /// Validates that no batch ends with an operation that has an immediate value.
+    fn validate_no_immediate_endings(&self) -> Result<(), String> {
```
bobbinth: Is this check correct? I thought that the rule is that no _operation group_ can end with an operation requiring an immediate value. This does imply that no operation batch can end with an operation requiring an immediate value, but it is stricter than that. In fact, I think the rules are:

- No operation group can end with an operation requiring an immediate value.
- The last operation group in a batch cannot contain operations requiring immediate values.

But it is possible that I'm misremembering things.

core/src/mast/node/basic_block_node/mod.rs
bobbinth: Not from this PR, but the second sentence here mixes up batches and groups. It should be: "If there are no groups available in the batch, the both the operation and its immediate value are moved to the next batch."

Also, I think the rules here could be clarified and made more complete. Maybe something like (assuming my last comment is correct):

- A basic block contains one or more batches.
- A batch contains up to 8 groups, and the number of groups must be a power of 2.
- A group contains up to 9 operations or 1 immediate value.
- Last operation in a group cannot be an operation that requires an immediate value.
- NOOPs are used to fill a group or batch when necessary.
- An immediate value follows the operation that requires it, using the next available group in the batch. If there are no groups available in the batch, then both the operation and its immediate are moved to the next batch.

bobbinth (commented):
Thank you! Looks good! Not a full review yet, but I left a couple of comments inline.

core/src/mast/node/basic_block_node/mod.rs
bobbinth: One caveat is that the above rules apply to the padded representation. For example, a batch can contain 3 groups, but then we pad it with the 4th group consisting of a single noop to bring the number of groups to a power of 2.

This goes back to the other discussion: it may be easier (eventually) to serialize the padded representation with additional metadata capturing the number of padded noops in each group. Naively, this would require 24 extra bits ( $8 \codt 3$ ) per batch to encode, which I think may be an acceptable cost - but maybe there are better ways to do this as well.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: 🤦 Yes, this is insufficient. It slipped my mind that no operation *group* rather than batch can finish in in immediate.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: The batches are [already represented in a way that encodes the nature of the padded representation](https://github.com/0xMiden/miden-vm/blob/6ebb3814cd75253125ef78b370139dd1db8148ee/core/src/mast/node/basic_block_node/op_batch.rs#L28-L29). This is one bit per group, which is enough when you also have the operations of the group.

bobbinth: Ah - nice! So, we should be able to encode this info with one byte per batch - or maybe even less. In that case, I do think it is worth it to store padded operation + this extra metadata.

huitseeker: OK, I'll quote you on this in #2448 !

CHANGELOG.md
```diff
@@ -53,6 +54,8 @@
 - [BREAKING] Normalized naming of `verify` procedures of ECDSA precompile ([#2413](https://github.com/0xMiden/miden-vm/issues/2413)).
 - Refactored Blake3_256 fingerprints to allocate less ([#2375](https://github.com/0xMiden/miden-vm/pull/2375)).
 - [BREAKING] Normalized signature encoding methods in the `dsa` module of the core library.
+- [BREAKING] Make the Assembler work in debug mode, remove optionality ([#2396](https://github.com/0xMiden/miden-vm/pull/2396)).
+- [BREAKING] Normalize naming of `verify` procedures of ECDSA precompile ([#2413](https://github.com/0xMiden/miden-vm/issues/2413)).
```
bobbinth: These entries got duplicated somehow. I don't think these changes need to be in this PR.

PR #2418, opened 2025-12-05
fix: Correct YAML indentation under run block in GitHub Actions workflow

huitseeker:
The GitHub Actions runner failed to load .github/workflows/test.yml due to an invalid YAML syntax on line 62:

```
Invalid workflow file: .github/workflows/test.yml#L62
You have an error in your yaml syntax on line 62
```

YAML uses indentation strictly to define structure. In this workflow, the `run:` key specifies a multi-line shell block (`|`), which requires consistent indentation for every line of the script relative to run:.

However, line 62 began with a single space instead of the two spaces used throughout the rest of the file:
```
run: |
   if [ -n "$(git status --porcelain libcore/docs)" ]; then
```
Because of that one-space misalignment, the parser failed to recognize the start of the multi-line literal block, resulting in a syntax error before the workflow could even load.

Now, why is it that maintainers could merge PRs without that workflow running – not to mention succeeding? — that's the real question.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@
 - Add a compaction function for achieving maximal sharing out of a `MastForest` with stripped decorators ([#2408](https://github.com/0xMiden/miden-vm/pull/2408)).
 - Refactor and remove tech debt from parallel trace generation ([#2382](https://github.com/0xMiden/miden-vm/pull/2382))
 - [BREAKING] Added `kind` field to `Package` struct to indicate package type (Executable, AccountComponent, NoteScript, TxScript, AuthComponent) ([#2403](https://github.com/0xMiden/miden-vm/pull/2403)).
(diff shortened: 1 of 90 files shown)


bobbinth:
The signature test failures are because `miden-crypto` encodes ECDSA signatures into 66 bytes but here we assume that the signatures are 65 bytes. That latter is correct, and in fact in the `next` branch of `miden-crypto` we already have the fix to reduce signature size to 65 bytes. So, I'll bring the fix into the `main` branch of `miden-crypto`, make v0.19.2 release, and yank the v0.19.1 release. After that, the test here should pass.

Not sure why the doc tests started to fail now though.

bobbinth:
I've fixed the tests by updating to `miden-crypto` v0.19.2. The doc tests are still failing, but I'll merge as is, and maybe we can figure out the doc test failures in a follow-up.

PR #2420, opened 2025-12-05
perf: run batch inversion concurrently per fragment

### LINKED ISSUE #2405: Run batch inversion concurrently, per fragment
Context: https://github.com/0xMiden/miden-vm/pull/2345#discussion_r2576778297

Farukest:
Description:
## Describe your changes

Run batch inversion on the H0 helper column concurrently using `par_chunks_mut` per fragment, instead of processing the entire column sequentially.

Closes #2405

- [x] Updated `CHANGELOG.md`

### DIFF
--- a/crates/assembly-syntax/src/ast/type.rs
+++ b/crates/assembly-syntax/src/ast/type.rs
@@ -417,7 +417,7 @@ impl PartialEq for PointerType {
 impl core::hash::Hash for PointerType {
     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
         self.pointee.hash(state);
-        self.addrspace.hash(state);
+        self.address_space().hash(state);
     }
 }
 


Farukest:
All done @plafer 
Need your review. Thanks :)

processor/src/parallel/mod.rs
```diff
@@ -249,9 +249,13 @@ fn generate_core_trace_columns(
...
-    // Run batch inversion on stack's H0 helper column
-    core_trace_columns[STACK_TRACE_OFFSET + H0_COL_IDX] =
-        batch_inversion(&core_trace_columns[STACK_TRACE_OFFSET + H0_COL_IDX]);
+    // Run batch inversion on stack's H0 helper column concurrently, per fragment
+    core_trace_columns[STACK_TRACE_OFFSET + H0_COL_IDX]
+        .par_chunks_mut(fragment_size)
+        .for_each(|chunk| {
+            let inverted = batch_inversion(chunk);
+            chunk.copy_from_slice(&inverted);
+        });
```
plafer: While this works, I would do this `batch_inversion()` in `CoreTraceFragmentFiller::fill_fragment()`, right after the call to `self.fill_fragment_impl(execution_state)`.

Basically

```rs
  pub fn fill_fragment(mut self) -> ([Felt; STACK_TRACE_WIDTH], [Felt; SYS_TRACE_WIDTH], usize) {
      let execution_state = self.context.initial_execution_state.clone();
      // Execute fragment generation and always finalize at the end
      let _ = self.fill_fragment_impl(execution_state);

      // Invert H0 column in the stack trace
      let num_rows_built = self.num_rows_built();
      {
          let inverted_h0 = batch_inversion(
              &self.fragment.columns[STACK_TRACE_OFFSET + H0_COL_IDX][0..self.num_rows_built()],
          );
          self.fragment.columns[STACK_TRACE_OFFSET + H0_COL_IDX][0..num_rows_built]
              .copy_from_slice(&inverted_h0);
      }

      let final_stack_rows = self.stack_rows.unwrap_or([ZERO; STACK_TRACE_WIDTH]);
      let final_system_rows = self.system_rows.unwrap_or([ZERO; SYS_TRACE_WIDTH]);
      (final_stack_rows, final_system_rows, num_rows_built)
  }
```

Farukest: Its good.. Done @plafer moved `batch_inversion()` into `fill_fragment()` as suggested.

Farukest:
Hey :)
any update @plafer ?

plafer:
@Farukest can you check why the tests are failing now?

Since all but 2 tests are passing, I wonder if there's an edge case that we're inverting 1 extra/fewer element in certain conditions.

Farukest:
> can you check why the tests are failing now?
> Since all but 2 tests are passing, I wonder if there's an edge case that we're inverting 1 extra/fewer element in certain conditions.

I see. I'm checking now @plafer

Farukest:
> While this works, I would do this batch_inversion() in CoreTraceFragmentFiller::fill_fragment(), right after the call to self.fill_fragment_impl(execution_state).

@plafer I tried to apply your suggestion but moving `batch_inversion()` into `fill_fragment()` causes test failures.

The issue is the execution order:
1. `fill_fragment()` runs and inverts H0 for rows in this fragment
2. `fixup_stack_and_system_rows()` then overwrites the **first row** of each fragment with the previous fragment's buffer
3. This buffer contains **non-inverted** H0 values (from `stack_rows` returned by `fill_fragment`)

So row _65536_ (first row of fragment 2) gets overwritten with a **non inverted** value after being **inverted**.

The fix must run `batch_inversion` **after** `fixup_stack_and_system_rows()` completes. We still get parallelism via `par_chunks_mut()`.

The failing tests showed:
- `stack_h0` at row 65536: slow=17831852600434098177 ( inverted ), parallel=30 ( non inverted )

plafer:
Ah yes that makes sense - thank you for the analysis!

In that case let's revert to what you had initially.

Farukest:
> Ah yes that makes sense - thank you for the analysis!
> In that case let's revert to what you had initially.

Already pushed @plafer . Done ! :)

plafer:
> Already pushed @plafer . Done ! :)

Did you push the intended commits? I still see the incorrect implementation, and tests are still failing

Farukest:
> Did you push the intended commits? I still see the incorrect implementation, and tests are still failing

My mistake. Fixed now.

bobbinth:
@Farukest - thank you for tackling this! I'm curious, did this have any noticeable impact on performance?

plafer:
I checked, and unfortunately it didn't. I figured we might as well merge it, if not only so that we don't have to worry about that being a performance penalty.

Issue #2413, opened 2025-12-04
Re-organize signing functions in stdlib


adr1anh:
Probably not for this PR, but I'd love to re-organize this set of functions as it feels odd to have them all exported from the library root. Also, we should make them somewhat consistent. Specifically:

- For EdDSA we provide `eddsa_sign()` and `encode_eddsa_signature()`.
- For ECDSA we provide `ecdsa_sign()`, `prepare_ecdsa_signature()`, and `encode_ecdsa_signature()`.
- For Falcon we provide just `falcon_sign()` and also use the non-deterministic version.

I think for all 3 we should provide the `*_sign()` and `encode_*_signature()` functions, and we should make Falcon signing deterministic (so that it can be done in `no_std` context as well).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2312#discussion_r2583345343_

bobbinth:
I think this is largely covered by https://github.com/0xMiden/miden-vm/issues/2343#issuecomment-3499422538 - but probably worth making this a separate issue. So, copying the comment here:

- For each signature scheme, we should provide a MASM procedure that takes a public key commitment and a message word on the stack, and get the actual public key and signature via the advice provider.
  -  We should probably uniformly name this procedure `verify`, which means that for the ECDSA scheme, the current `verify` would become `verify_prehash`.
- We should provide Rust functions for generating the data that needs to be put into the advice provider (i.e., the public key and the signature).
    - For ECDSA we already have these functions, but we probably should organize them better.
    - For Falcon we have one of the functions (to generate the signature), and the other one lives in `miden-base` ([here](prepare_rpo_falcon512_signature)) - we should probably bring it over.
    - For EdDSA it is a bit more tricky because, AFAIK, there is no ec-recover for EdDSA - and so, we need to get the full public key from somewhere. One option is to redefine `EdDSA` signature to include both signature and the public key (same as we did with Falcon). Another option is to modify the mechanism in `miden-base` of how we get public keys - e.g., maybe the we should add a method to the authenticator to get public key for a given public key commitment.

Farukest:
Hi @adr1anh @bobbinth ,

I'd like to work on this issue but I've a few questions before :

1. For Falcon, you mentioned the encode function lives in `miden-base`. Is it `prepare_rpo_falcon512_signature` or similar ?
Should it be ported it here ?
[...]

PR #2424, opened 2025-12-05
refactor: normalize naming of ECDSA verify procedures

adr1anh:
Rename ECDSA K256 verification procedures for consistency with EdDSA naming patterns and issue #2413:

- `verify_ecdsa_k256_keccak` → `verify` (high-level wrapper)
- `verify` → `verify_prehash` (manual prehash verification)
- `verify_impl` → `verify_prehash_impl` (internal implementation)

Added docstring note to prehash procedures indicating they are intended for manual signature verification where the caller has already computed the message digest.

### DIFF
--- a/crates/lib/core/tests/crypto/ecdsa_k256_keccak.rs
+++ b/crates/lib/core/tests/crypto/ecdsa_k256_keccak.rs
@@ -10,7 +10,6 @@ use miden_core::{
     precompile::{PrecompileCommitment, PrecompileVerifier},
     utils::{Deserializable, Serializable},
 };
-use miden_crypto::{dsa::ecdsa_k256_keccak::SecretKey, hash::rpo::Rpo256};
 use miden_core_lib::{
     dsa::ecdsa_k256_keccak::sign as ecdsa_sign,
     handlers::{
@@ -18,6 +17,7 @@ use miden_core_lib::{
         ecdsa::{EcdsaPrecompile, EcdsaRequest},
     },
 };
+use miden_crypto::{dsa::ecdsa_k256_keccak::SecretKey, hash::rpo::Rpo256};
 use miden_processor::{AdviceMutation, EventError, EventHandler, ProcessState};
 use rand::{SeedableRng, rngs::StdRng};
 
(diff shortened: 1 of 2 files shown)


docs/src/user_docs/core_lib/crypto/dsa.md
bobbinth, line 1: I think this file is missing a section for the EdDSA signature.

adr1anh:
Should we keep `with_k_unchecked` or just use `prehash` for EdDSA?

bobbinth:
> Should we keep `with_k_unchecked` or just use `prehash` for EdDSA?

I haven't looked at this in detail, but if the concept is the same, I'd go more for consistency (i.e., `verify_prehash`) and we can describe the nuance in the doc comments.

If there are significant differences though (e.g., the interfaces are different), the probably a different name is justified.

docs/src/user_docs/core_lib/crypto/dsa.md
```diff
@@ -8,6 +8,7 @@ Namespace `miden::core::crypto::dsa` contains a set of  digital signature scheme
 
 * `RPO Falcon512`: a variant of the [Falcon](https://falcon-sign.info/) signature scheme.
```
bobbinth, line 9: Not from this PR, but I'd rename this to `Falcon512-RPO`.

adr1anh:
The last two commits can be reverted/changed if necessary
- unify the functions inside a new `dsa.rs` module 
- change `with_unchecked_k_digest` to `prehash` in EdDSA for uniformity

What remains

- [ ] Falcon should use deterministic signing to enable usage in `no-std` context
- [ ] Include public key in EdDSA signature (and ECDSA to avoid having to pass the message as argument to `prepare_signature`)
- [ ] Decide on interface for inputs (secret key as `&[Felt]` or `&SigningKey`)

bobbinth:
> change `with_unchecked_k_digest` to `prehash` in EdDSA for uniformity

Done in 112b9092ba9f7d77c75376ec74a7436bd97eb9bb.

> Falcon should use deterministic signing to enable usage in `no-std` context

Done in 8e76dd9d4b01bac69040dca124ccad1e2ae14272

> Decide on interface for inputs (secret key as `&[Felt]` or `&SigningKey`)

Done in 24258cfacf251420212fc1f197a7f1a709c5b4e1

> Include public key in EdDSA signature (and ECDSA to avoid having to pass the message as argument to `prepare_signature`)

I'm not sure we should do this - so, I didn't do this for now.

PR #2425, opened 2025-12-05
refactor(air): remove Blake3_192 hash function variant

### LINKED ISSUE #2104: Consider removing Blake3-192 variant
This variant is there mainly for proof size reduction and experiments show that the reduction is in the order of $17$% in comparison to Blake3-256 variant.

We should consider removing it as it both makes the codebase uniform and makes the security argument clearer and easier to reason about.

Context [here](https://github.com/0xMiden/miden-vm/pull/2098#pullrequestreview-3161864469)

Himess:
## Summary

Remove the `Blake3_192` variant from the `HashFunction` enum as it provides minimal benefit over `Blake3_256` while adding complexity to the codebase.

## Changes

- Remove `Blake3_192` from `HashFunction` enum in `air/src/proof.rs`
- Update verifier to accept both 96-bit and 128-bit proofs for `Blake3_256`
- Update prover to only use `Blake3_256` for Blake3 proofs
- Change CLI default hasher from `blake3-192` to `blake3`
- Update all tests to use `Blake3_256`
- Add `blake3` as an alias for `blake3-256` in the CLI for convenience

## Breaking Changes

- `HashFunction::Blake3_192` variant is removed
- Default hash function changed from `Blake3_192` to `Blake3_256`
- CLI `--hasher` default changed from `blake3-192` to `blake3`

Closes #2104


bobbinth:
Thank you - but this is not something we'd want to remove just yet. So, I'll close this now - but we may come back to this in the future.

PR #2427, opened 2025-12-05
Rename `miden-libcore` to `miden-core-lib`

### LINKED ISSUE #2421: Crate naming and organization for library crates
We recently renamed `miden-stdlib` crate into `miden-libcore` (which is currently located in the `libcore` directory) - as discussed offline, it may be good to change organization and naming slightly. Specifically:

- We should put the crate into `lib/core` directory. This would open up the possibility to split the crate input multiple crates down the road (e.g., `lib/crypto`).
- We can make the naming of the crate a bit more "structures". Given that this is a "library" crate, I think we could use `-lib` suffix. So, `miden-libcore` would become `miden-core-lib`. There are some variations on this we can consider:
  - We could go with `miden-lib-core` - though, I do think `miden-core-lib` is a bit better.
  - We could use some other suffix - e.g., `pkg`, `sys` or something like that - though, out of these, I think `lib` is a better option.

plafer:
Closes #2421

Moves the `libcore/` directory under `crates/lib/core/` and renames the `miden-libcore` crate to `miden-core-lib`.

Note that the `crates/*` member in the top-level `Cargo.toml` was expanded, since adding the empty `lib/` directory broke that rule (since `lib/` is not a crate).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@
 - `hash_memory_with_state`, `hash_memory_words`, and `hash_memory_double_words` procedures from the `std::crypto::hashes::rpo` module were renamed to the `hash_elements_with_state`, `hash_words`, and `hash_double_words` respectively ([#2395](https://github.com/0xMiden/miden-vm/pull/2395)).
 - [BREAKING] Updgrade `miden-crypto` to 0.19 ([#2399](https://github.com/0xMiden/miden-vm/pull/2399)).
 - Pre-allocate main trace buffer in trace generation ([#2345](https://github.com/0xMiden/miden-vm/pull/2345)).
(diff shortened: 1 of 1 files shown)


adr1anh (approved):
Looks good to me

scripts/check-features.sh
```diff
@@ -9,7 +9,7 @@ echo "Checking all feature combinations with cargo-hack..."
 
 # Set environment variables to treat warnings as errors
 export RUSTFLAGS="-D warnings"
-export MIDEN_BUILD_LIBCORE_DOCS=1
+export MIDEN_BUILD_CORE_LIB_DOCS=1
```
bobbinth: I would maybe rename this here (and in other relevant places) to just `MIDEN_BUILD_LIB_DOCS` to imply that this is for building library docs in general (not just core library docs).

bobbinth (approved):
Looks good! Thank you! I left just one small comment inline.

PR #2430, opened 2025-12-05
refactor: remove MastNodeErrorContext

huitseeker:
This PR 
- hoists `get_assembly_op` to the `MastForest` and 
- correspondingly removes the `MastNodeErrorContext` trait and 
- removes the associated Iterator (`DecoratedOpLinkIterator`) complexity and
- improves test coverage of `get_assemblyop`.

See https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2508446259 for rationale and approach.
As a byproduct, contributes to #1978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 0.21.0 (TBD)
 
+#### Enhancements
+
+#### Changes
+
+- [BREAKING] Move `get_assembly_op` to the `MastForest`, remove trait `MastNodeErrorContext` ([#2430](https://github.com/0xMiden/miden-vm/pull/2430)).
+
 ## 0.20.0 (2025-12-05)
 
 #### Enhancements
--- a/core/src/mast/node/basic_block_node/tests.rs
+++ b/core/src/mast/node/basic_block_node/tests.rs
@@ -489,7 +489,7 @@ fn test_mast_node_error_context_decorators_iterates_all_decorators() {
         .unwrap();
 
     // For basic blocks, we need to combine before_enter, operation-indexed, and after_exit
-    // decorators using the helper method
+    // decorators
     let all_decorators = forest.all_decorators(node_id);
 
     // Should have 3 decorators total: 1 before_enter, 1 during, 1 after_exit


CHANGELOG.md
```diff
@@ -45,6 +51,8 @@
 - [BREAKING] Normalized naming of `verify` procedures of ECDSA precompile ([#2413](https://github.com/0xMiden/miden-vm/issues/2413)).
 - Refactored Blake3_256 fingerprints to allocate less ([#2375](https://github.com/0xMiden/miden-vm/pull/2375)).
 - [BREAKING] Normalized signature encoding methods in the `dsa` module of the core library.
+- [BREAKING] Make the Assembler work in debug mode, remove optionality ([#2396](https://github.com/0xMiden/miden-vm/pull/2396)).
+- [BREAKING] Normalize naming of `verify` procedures of ECDSA precompile ([#2413](https://github.com/0xMiden/miden-vm/issues/2413)).
```
plafer: Those modify the 0.20 changelog; do they belong there still?

core/src/mast/mod.rs
```diff
@@ -493,6 +494,76 @@ impl MastForest {
...
+            // If a target operation index is provided, return the assembly op associated with that
+            // operation.
+            Some(target_op_idx) => {
+                for (op_idx, decorator_id) in decorator_links {
+                    if let Some(Decorator::AsmOp(assembly_op)) = self.decorator_by_id(decorator_id)
+                    {
+                        // when an instruction compiles down to multiple operations, only the first
+                        // operation is associated with the assembly op. We need to check if the
+                        // target operation index falls within the range of operations associated
+                        // with the assembly op.
```
bobbinth, line 545: Not for this PR, but maybe in the future we could make it so that every operation covered by an `AsmOp` maps to this `AsmOp`. Previously, we didn't want to do it because it would have meant duplicating `AsmOp`s for every operation, but now that the structure is separate and ID-based, maybe that's easier.

We should do this only if this simplifies the codebase though.

bobbinth (approved):
Looks great! Thank you! I left a small question for the future inline.

CHANGELOG.md
huitseeker: Ah, yeah, I need to polish that!

core/src/mast/mod.rs
huitseeker: #2446

PR #2422, opened 2025-12-05
chore: Change `#[allow(*)]` to `#[expect(*)]` and remove unused lint bypasses

adr1anh:
Replaces occurrences of `#[allow(*)]` with `#[expect(*)]` which surfaced some lint suppressions which were no longer needed. Motivated by the fix of https://github.com/rust-lang/rust/issues/147648.

PR #1826, opened 2025-05-18
chore(lint): add clippy-extra.sh for optional stricter linting

Kalpazanova53:
## Describe your changes
Closes #1479 

This PR adds an optional script to run stricter Clippy lints across the codebase.

- Introduces `scripts/clippy-extra.sh` to enable additional checks (`pedantic`, `style`, `missing_docs`, etc.)
- The script is intended for contributors who want to proactively catch rare but useful warnings (e.g., unnecessary casts, missing `#[must_use]`, undocumented panics/errors, etc.)
- These checks are not enforced in CI to avoid developer friction, but the script provides an easy opt-in for manual use.

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](./CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #2437, opened 2025-12-07
feat: introduce stricter lint configuration and clippy.toml

huitseeker:
We spend a lot of time on lints and attributes for linting, and manually tuning our own lints for relatively low signal-to-noise stuff (e.g. `clippy::too_many_arguments`).

Examples:
- https://github.com/0xMiden/miden-vm/pull/2422
- https://github.com/0xMiden/miden-vm/pull/2317
- https://github.com/0xMiden/miden-vm/pull/1826

We also inflict nits about this to each other in PR review, e.g. :
- https://github.com/0xMiden/miden-vm/pull/2366#discussion_r2561945992

This PR is about sparing us cycles on this easy-to-mechanize stuff. It introduces bog-standard lint configuration through a .clippy.toml and .cargo/config.toml with a couple of extra-mild lints, and adjusts the code base in function.

<details><summary> Activated Clippy Lints </summary>

Clippy Lints

- **clippy::all**: Enables all default clippy lints that are not specifically enabled by other lints
- **clippy::assertions_on_result_states**: Warns against using assertions on Result types instead of proper error handling
- **clippy::cast_lossless**: Identifies cast conversions between numerical types that could be replaced by safer, more explicit conversion functions like Into or From.
- **clippy::checked_conversions**: Checks for explicit checked conversions that could be made more idiomatic
- **clippy::dbg_macro**: Warns about using the `dbg!` macro in production code
- **clippy::disallowed_methods**: Allows configuration of methods that should not be used
- **clippy::filter_map_next**: Checks for usage of `filter_map().next()` that could be `find()`
- **clippy::flat_map_option**: Checks for `flat_map(Option)` usage that could be more efficiently written
- **clippy::inefficient_to_string**: Warns about inefficient `to_string()` conversions
- **clippy::large_stack_arrays**: Warns about large arrays allocated on the stack that could cause stack overflow
- **clippy::large_types_passed_by_value**: Warns when large types are passed by value instead of by reference
- **clippy::macro_use_imports**: Discourages the use of `macro_use` in favor of more explicit imports
- **clippy::manual_assert**: Suggests using the `assert!` macro instead of manual assertion implementations
- **clippy::manual_ok_or**: Suggests using the `ok_or()` method instead of manual implementations
- **clippy::map_flatten**: Checks for usage of `map().flatten()` that could be replaced with `flat_map()`
- **clippy::map_unwrap_or**: Checks for usage of `map().unwrap_or()` that could be more safely written
- **clippy::match_wild_err_arm**: Warns about wildcard patterns in error handling match arms
- **clippy::needless_borrow**: Detects unnecessary borrows that can be removed
- **clippy::option_option**: Warns about nested `Option<Option<T>>` types that could be simplified
- **clippy::same_functions_in_if_condition**: Detects repeated function calls in if conditions that can be cached
- **clippy::single_match_else**: Suggests using `if let` instead of `match` with a single pattern
- **clippy::trait_duplication_in_bounds**: Detects duplicated trait bounds in where clauses
- **clippy::unnecessary_mut_passed**: Warns when unnecessary mutable references are passed to functions

Rustc Lints

- **nonstandard_style**: Warns about non-standard naming and style conventions
- **rust_2018_idioms**: Enables lint group for Rust 2018 Edition idioms compliance
- **trivial_numeric_casts**: Warns about numeric casts that don't change the value
- **unused_lifetimes**: Warns about unused lifetime parameters
- **unused_qualifications**: Warns about unnecessary type qualifications

Warning Level

- **warnings**: Treats all warnings as errors to prevent problematic code from being accepted

</details>


air/src/constraints/chiplets/bitwise/tests.rs
```diff
@@ -187,14 +187,14 @@ pub fn get_test_frame(
     let output_prev = if cycle_row == 0 {
         ZERO
     } else {
-        Felt::new((result >> previous_shift) as u64)
+        Felt::new(u64::from(result >> previous_shift))
```
adr1anh, line 190: Should we use `Felt::from` here? The advantage of `new` is that it is `const` but otherwise it seems like 
```suggestion
        Felt::from(result >> previous_shift)
```
would be a bit easier to parse

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -650,7 +646,7 @@ impl fmt::Display for BasicBlockNodePrettyPrint<'_> {
 }
 
 enum Mid<'a> {
-    Slice(core::slice::Iter<'a, (usize, DecoratorId)>),
+    Slice(Iter<'a, (usize, DecoratorId)>),
```
adr1anh, line 649: When it comes to using `iter` types which often have the same names, I find the right middle ground is to import the module to annotate the specific type

```suggestion
    Slice(slice::Iter<'a, (usize, DecoratorId)>),
```

core/src/mast/node/external.rs
```diff
@@ -418,12 +409,7 @@ impl proptest::prelude::Arbitrary for ExternalNodeBuilder {
 
         (
             any::<[u64; 4]>().prop_map(|[a, b, c, d]| {
-                miden_crypto::Word::new([
-                    miden_crypto::Felt::new(a),
-                    miden_crypto::Felt::new(b),
-                    miden_crypto::Felt::new(c),
-                    miden_crypto::Felt::new(d),
-                ])
+                Word::new([Felt::new(a), Felt::new(b), Felt::new(c), Felt::new(d)])
```
adr1anh, line 412: I think (?) we have

```suggestion
                Word::from([a,b,c,d])
```

core/src/mast/serialization/info.rs
```diff
@@ -384,7 +384,7 @@ mod tests {
     fn deserialize_large_payloads_fails() {
         // Serialized `CALL` with a 33-bit payload
         let serialized = {
-            let serialized_value = ((CALL as u64) << 60) | (u32::MAX as u64 + 1_u64);
+            let serialized_value = (u64::from(CALL) << 60) | (u64::from(u32::MAX) + 1_u64);
```
adr1anh, line 387:
```suggestion
            let serialized_value = (u64::from(CALL) << 60) | (u64::from(u32::MAX) + 1);
```

crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -45,9 +45,9 @@ pub fn generate_advice_inputs(
 
     // we need to provide the following instance specific data through the operand stack
     let initial_stack = vec![
-        proof.context.options().grinding_factor() as u64,
+        u64::from(proof.context.options().grinding_factor()),
         proof.context.options().num_queries() as u64,
```
adr1anh, line 49: This isn't flagged?

huitseeker:
OK, the goal was to have a PR with a quick resolution, I'll remove the fixes you found controversial.

adr1anh:
> OK, the goal was to have a PR with a quick resolution, I'll remove the fixes you found controversial.

I don't think the `u64::from` is controversial, my comment was aimed towards the fact that we should use `Felt::from` when we're not dealing with a `u64` input, or when we specifically need `const`.

These lints are good otherwise! Maybe we could re-apply them one at a time in separate commits to check if they're all "correct".

Issue #2383, opened 2025-11-25
Add `math::u128` module to the standard library


bobbinth:
We should add a `u128` module to standard library to support 128-bit arithmetic. It should probably largely follow the interfaces of the `u64` module.

I believe @otrho has already implemented a big part of this in the compiler repo - so, hopefully, this would be more like a port of that code to this repo.

otrho:
I've begun a WIP for this in [`otrho/stdlib-u128`](https://github.com/0xMiden/miden-vm/tree/otrho/stdlib-u128).  At this stage it's just the code pulled straight from the compiler, implementing a `wrapping_add`, `wrapping_sub` and a `wrapping_mul`.  I also added basic proptests for these three procs.

So, a couple of questions:
- How are the stdlib docs generated?
- Which other operations do we want/need for `u128`?
    - div and mod?
    - comparisons?
    - bitwise?
    - overflowing and checked versions?

I'm guessing maybe? to the other ops, though comparisons and div maybe not, but probably a yes to the bitwise ops.

Overflowing/checked on `u128` seems a bit much, are people truly counting things with these numbers?

bobbinth:
> Which other operations do we want/need for `u128`?

I think we should try to follow the `u64` module here. So, the procedures would be:

- `wrapping_add`, `overflowing_add`.
- `wrapping_sub`, `overflowing_sub`.
- `wrapping_mul`, `overflowing_mul`.
-  `lt`, `lte`, `gt`, `gte`, `eq`, `neq`, `eqz`, `min`, `max`.
- `div`, `mod`, `divmod`.
- `and`, `or`, `xor`, `shl`, `shr`, `rotl`, `rotr`, `clz`, `ctz`, `clo`, `cto`.

We don't need to implement all of these in a single PR - I'd probably do this over 3 - 4 PRs (e.g., basic arithmetic, division, comparisons, bitwise operations).

If we do decide that some of the above procedures are not needed - we should probably also remove them from the `u64` module.

> How are the stdlib docs generated?

There is some auto-generation happening based on doc comments, but the output may need to be "massaged" to get it into its final form.

partylikeits1983:
`div` is the most anticipated & wanted procedure for u128 :)

bobbinth:
The basic operations have been covered by https://github.com/0xMiden/miden-vm/pull/2617, and I've opened https://github.com/0xMiden/miden-vm/issues/2619 for the remainder.

PR #2438, opened 2025-12-08
Add initial `math::u128` functions for core/math runtime.

otrho:
Edit: ~on top of #2593~
----
Re: #2383.  This is a start at adding `u128` maths support with `add`, `sub` and `mul` in both overflowing and wrapping versions.

The cycle counts are estimates - is there a way to formally get the cycle cost for a proc, or alternatively are there docs somewhere explaining when a `dup` or `movup` will cost 1, 2 or 3 cycles?  I assume it's related to the distance from the top somehow.

The `overflowing_mul` implementation is pretty conservative and could probably be pared down a little.

In particular any time there is an overflow into the 129th bit or higher the value is coverted to a flag/bool and they are all OR'd together at the end (except `add3` can create a '2' overflow, so they're summed and then converted again to a bool).

There are quite likely cases were we know definitely there can either be no overflow, or the overflow is already a bool and doesn't need conversion.  But for now it's passing the tests and works, we can optimise as we go.

### DIFF
--- a/crates/lib/core/asm/math/u128.masm
+++ b/crates/lib/core/asm/math/u128.masm
@@ -144,8 +144,8 @@ pub proc overflowing_mul(b: u128, a: u128) -> (i1, u128)
     # Input: [b_ll, b_ml, b_mh, b_hh, a_ll, a_ml, a_mh, a_hh, ...]
     # Position:  0     1     2     3     4     5     6     7
     #
-    # u32overflowing_mul: [b, a] -> [lo, hi] computes a*b
-    # u32overflowing_madd: [c, b, a] -> [lo, hi] computes a*b+c
+    # u32widening_mul: [b, a] -> [lo, hi] computes a*b
+    # u32widening_madd: [c, b, a] -> [lo, hi] computes a*b+c
 
     # Move a limbs to front for easier access
     movup.7 movup.7 movup.7 movup.7
@@ -155,7 +155,7 @@ pub proc overflowing_mul(b: u128, a: u128) -> (i1, u128)
     # === Column 0: c_ll ===
     # a_ll * b_ll
     dup.4 dup.1                 # [a_ll, b_ll, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh]
-    u32overflowing_mul          # [c_ll, o0, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh]
+    u32widening_mul          # [c_ll, o0, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh]
     movdn.9                     # [o0, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
     # Stack: [o0, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
     # Position: 0   1     2     3     4     5     6     7     8     9
@@ -163,11 +163,11 @@ pub proc overflowing_mul(b: u128, a: u128) -> (i1, u128)
     # === Column 1: c_ml ===
     # a_ml * b_ll + o0: [c, b, a] for madd where c=o0, b=b_ll, a=a_ml
     dup.5 dup.3                 # [a_ml, b_ll, o0, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
-    u32overflowing_madd         # [p1, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
+    u32widening_madd         # [p1, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
 
     # a_ll * b_ml + p1: [c, b, a] where c=p1, b=b_ml, a=a_ll
     dup.7 dup.3                 # [a_ll, b_ml, p1, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
-    u32overflowing_madd         # [c_ml, o1b, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
+    u32widening_madd         # [c_ml, o1b, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll]
     movdn.11                    # [o1b, o1a, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll, c_ml]
     # Sum o1a + o1b, tracking carry for column 2
     u32overflowing_add          # [o1_sum, o1_carry, a_ll, a_ml, a_mh, a_hh, b_ll, b_ml, b_mh, b_hh, c_ll, c_ml]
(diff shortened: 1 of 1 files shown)


crates/lib/core/asm/math/u128.masm
```diff
@@ -0,0 +1,263 @@
...
+    u32overflowing_sub  # [uhh' rhh uhh rmh rml rll ...]
+    movup.2             # [uhh uhh' rhh rmh rml rll ...]
+    or                  # [uhh rhh rmh rml rll ...]
+end
+
+#! Performs subtraction of two unsigned 128 bit integers discarding the underflow.
+#! The input values are assumed to be represented using 32 bit limbs, but this is not checked.
+#! Stack transition is as follows:
+#! [bhh, bmh, bml, bll, ahh, amh, aml, all, ...] -> [chh, cmh, cml, cll, ...], where c = (a - b) % 2^128
+pub proc wrapping_sub(b: u128, a: u128) -> (u32, u128)
```
huitseeker: The return type here says `(u32, u128)` but the procedure only produces 4 stack elements (a `u128`). Looks like a copy-paste from `overflowing_sub`. Should this be `-> u128` to match `wrapping_add`?

crates/lib/core/asm/math/u128.masm
```diff
@@ -0,0 +1,263 @@
...
+#                          o4 allxbml       0
+#                  o5 amlxbml       0       0
+#          f2 amhxbml       0       0       0     amh x2 uses, bml x3 uses
+#                  o7 allxbmh       0       0
+#          f3 amlxbmh       0       0       0     aml x3 uses, bmh x2 uses
+#          f4 allxbhh       0       0       0     all x4 uses, bhh x1 uses
+# -------------------------------------------
+#           f      rhh    rmh     rml     rll
+
+pub proc overflowing_mul(b: u128, a: u128) -> (u64, u128)
```
huitseeker: The return type says `u64` for the overflow flag, but the `u64` module uses `i1` for overflow flags (see `u64::overflowing_mul`). The actual value here is just 0 or 1 from the final `push.0 neq`. Should this be `(u32, u128)` or `(i1, u128)` for consistency?

bitwalker (dismissed):
Aside from @huitseeker's comments, I think this looks good. We'll want to implement division in a subsequent PR, and we may need to make layout changes to both u64 and u128 in a future change, but neither of those need to happen here, so 👍 from me.

crates/lib/core/asm/math/u128.masm
bobbinth: I'm actually not sure the signature in the `u64::overflowing_mul` is correct - or at least it conflicts with the doc comments for the same procedure. Specifically, I think it returns `(u64, u64)` rather than `(i1, u64)`.

If that's correct, could we also implement `u128::overflowing_mul` to look like so?:
```
pub proc overflowing_mul(b: u128, a: u128) -> (u128, u128)
```

otrho: `pub proc overflowing_mul(b: u128, a: u128) -> (u128, u128)` would imply 8 felts being left on the stack IMO, which is obviously not quite right.

bobbinth: I think maybe the naming here may be confusing. What I meant is that semantically, our `u64::overflowing_mul` functions like a widening multiplication (e..g, multiplying two u64 values gives us a 128-bit result). Similarly, would be great if this also behaved like a widening multiplication - i.e., multiplying two u128 values gives us a u256 result (which requires 8 felts to represent).

bitwalker: @bobbinth Those are unusual semantics IMO - `overflowing_mul` in most contexts I've seen (including Rust), treat it as an operation that returns the result as the same type as the operands, but with an additional flag that indicates whether overflow occurred, in which case the result value itself has been wrapped (giving you the overflow from `T::MAX`). 

To get the resulting "widened" value, you'd simply cast the `T::MAX` value and the overflow to the widened type, and add them together. That addition is guaranteed to fit in the widened type, so doesn't need to be checked.

If you want a single function that does the widening multiplication, I think it should be added in addition to `overflowing_mul`, e.g. `carrying_mul` from Rust. And build it on top of the `overflowing_mul` implementation (or perhaps, build `overflowing_mul` in terms of `carrying_mul`, whichever is most efficient).

IMO, people rarely want the widened value when using `overflowing_mul`, instead they want to catch overflow and treat it as an non-panicking error. It is certainly useful on its own.

The compiler currently uses `overflowing_mul` as I described above, and I'd assume most people would assume it works the way I outlined, so if the actual implementation is different, we definitely should fix that.

bobbinth: Yes, agreed that `overflowing_mul` is the right name for the semantics of this procedure. But the issue is that we use `overflowing_mul` to mean `widening_mul` (not `carrying_mul`) in other places (Rust also has [widening_mul](https://doc.rust-lang.org/std/primitive.u64.html#method.widening_mul) though, currently as nightly-only).

So, in a way, this PR introduces an inconsistency because we'll get:

- For `u32` and `u64` `overflowing_mul` will mean widening multiplication.
- For `u128` it'll mean overflowing multiplication.

I think maybe a simple solution here is to rename `u32` and `u64` procedure/operations to `widening_mul` - but that would need to be done in a different PR.

bobbinth: Created https://github.com/0xMiden/miden-vm/issues/2558 for this.

bitwalker: _Moved to [2558](https://github.com/0xMiden/miden-vm/issues/2558#issuecomment-3725299493)_

bobbinth: > 1. How are we actually using the current semantics throughout existing MASM code? I know that there must be a _lot_ of bugs as a result of that, because the compiler for sure treats that overflow value as a boolean, not a u32 value (and I'm just about 100% sure my interpretation was based on the actual Miden docs at some point, though perhaps the docs were fixed and I didn't catch that). Our compiler intrinsics that use `u32overlowing_mul` also appear to all be written with this assumption. We probably need to audit how many uses of `u32overflowing_mul` are making the same incorrect assumption I did, because I assume I'm not alone there.

All uses I've seen assume the "widening" semantics, but of course, I haven't seen them all. In many cases, using the upper bits as a boolean value could be fine - i.e., `0` still means there is no overflow. But yes, if we are checking it to be `1` to indicate that there was overflow, that'll be a bug.

> 2. The behavior of `u32overflowing_mul` is not consistent with `u32overflowing_add` (which has the correct semantics).

For addition, there is no difference between "overflow" and "widening" because even in case of widening you still get either `0` or `1` in the upper bits of the widened value. So, I think I was just trying to keep the naming consistent across addition and multiplication.

> Since the `u64` library is not being currently used outside the compiler (unless that has changed _very_ recently), I think we can just fix the semantics of `u64::overflowing_mul` to be correct, and add a new `u64::widening_mul` function.

Agreed - and that's what I suggested in https://github.com/0xMiden/miden-vm/issues/2558.

bitwalker: > For addition, there is no difference between "overflow" and "widening" because even in case of widening you still get either 0 or 1 in the upper bits of the widened value. So, I think I was just trying to keep the naming consistent across addition and multiplication.

The way the documentation treats it, the output on top of the stack is a boolean (which is effectively true, but misleading), while `u32overflowing_mul` shows that the output on top of the stack, i.e. `[d, c]` is $c←(a⋅b) \mod 2^{32}$ and $d←(a⋅b)/2^{32}$, which is unambiguous.

A quick immediate fix is to update the docs of `u32overflowing_add` to match `u32overflowing_mul`, so that one doesn't read the docs for `u32overflowing_add` and then assume the semantics of `u32overflowing_mul` are the same, just for multiplication. That's likely the mistake I made.

Dominik1999:
Can we still merge this?

bitwalker:
@Dominik1999 Yes, though I should note that it will need to be reworked as part of the effort to make all the integer types that have multi-word layouts (e.g. `u64` and `u128`) use a little-endian limb order (both currently use a big-endian layout). See #2564 for conversation around that.

For now though, I think we can go ahead and merge this, what do you think @bobbinth?

crates/lib/core/asm/math/u128.masm
```diff
@@ -0,0 +1,263 @@
...
+# -------------------------------------------
+#                  rhh    rmh     rml     rll
+
+pub proc wrapping_mul(b: u128, a: u128) -> u128
+                         # [bhh bmh bml bll ahh amh aml all ... ]
+    # all x bll -> rll
+    dup.7 dup.4
+    u32overflowing_mul   # [o0 rll bhh bmh bml bll ahh amh aml all ...]
+
+    # aml x bml + o0 -> rml0
```
huitseeker: Tiny typo: this comment says `aml x bml` but should be `aml x bll` to match the equivalent comment in `overflowing_mul` (line 125). The code itself looks correct.

bobbinth:
Superseded by https://github.com/0xMiden/miden-vm/pull/2617.

PR #2439, opened 2025-12-08
chore: fix PointerType Hash/Eq consistency

KitaroMoha563:
PointerType hashed addrspace directly but compared via address_space(), causing equal values to have different hashes.


bitwalker (changes requested):
The test coverage here is unnecessary IMO, or at least, doesn't feel like the value of them balances out the amount of noise they introduce. If we actually wanted to validate that we aren't making mistakes like this with Rust's various traits, we'd need an enormous number of such tests, and that's just not a valuable use of time as far as I'm concerned.

Otherwise, looks good - if you can remove the test code from this PR, leaving just the patch of the `Hash` impl, I'll merge.

KitaroMoha563:
Hi @bitwalke, removed the test code, pls check ,thanks!

bitwalker (approved):

plafer:
Hey @KitaroMoha563 can you sign your commits? Then we can merge.

KitaroMoha563:
@plafer did it, sir.

PR #2440, opened 2025-12-08
Introduce `FastProcessor::step()`

### LINKED ISSUE #2257: Investigate exposing `FastProcessor::step()`
It would be beneficial for the debugger to have a `FastProcessor::step()` method that steps only one operation/clock cycle at a time, instead of the `FastProcessor::step_continuation()` method that we're planning to add, which steps through an entire continuation at a time.

The ideal way to go about this would be to abstract away the `CoreTraceFragmentGenerator` logic, which effectively already has this functionality (due to every method returning a `ControlFlow`). The trick though is that the `CoreTraceFragmentGenerator`s still traverse the `MastForest` recursively (#2165), and so the ultimate solution would be to use the iterative traversal of the `FastProcessor`, and the `ControlFlow`-based execution of the `CoreTraceFragmentGenerator`s. Ideally, this would be integrated into the `Processor` trait, such that `Processor` would now expose a `step()` rather than an `execute_sync_op()` (i.e. all the `MastForest` traversal logic would now be nicely abstracted away in `Processor`, rather than just operation execution).

Note that for this to work, we cannot slow down the `FastProcessor`, and hence any `ControlFlow` returns would need to be inlined as `ControlFlow::Continue` in "fast execution mode" - however I am not sure if even that would have runtime performance repercussions.

plafer:
Closes #2257

Provides a `FastProcessor::step()` method, which executes 1 clock cycle and returns. As part of this work, the concept of `Continuation` was expanded to be granular at the clock cycle level. That is, a continuation exists to represent the execution remaining after all possible points where the clock is incremented.

The testing strategy is to compare the output of every test run (in `Test::execute()`) between the normal `FastProcessor` execution, and the execution of successive calls to `FastProcessor::step()`.

I believe there's still one shortcoming to be addressed via the debugger API. When `FastProcessor::step()` returns, the debugger one way or another know what operation just executed. For now, the plan is to have the debugger query the `ResumeContext` for the continuation that's about to be executed *before* calling `step()` to figure out the operation that is about to be executed. However, I'll wait for additional feedback on that flow before implementing anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Added new `peek` procedure for the `std::collections::smt` module ([#2387](https://github.com/0xMiden/miden-vm/pull/2387)).
 - Added new `pad_and_hash_elements` procedure to the `std::crypto::hashes::rpo` module ([#2395](https://github.com/0xMiden/miden-vm/pull/2395)).
 - Added padding option for the `adv.push_mapvaln` instruction ([#2398](https://github.com/0xMiden/miden-vm/pull/2398)).
+- Added new `FastProcessor::step()` method that executes a single clock cycle ([#2440](https://github.com/0xMiden/miden-vm/pull/2440))
 
 #### Changes
 


processor/src/fast/step.rs
```diff
@@ -0,0 +1,77 @@
...
+// ===============================================================================================
+
+/// The context required to resume execution of a program from the last point at which it was
+/// stopped.
+#[derive(Debug)]
+pub struct ResumeContext {
+    pub current_forest: Arc<MastForest>,
+    pub continuation_stack: ContinuationStack,
+    pub kernel: Kernel,
+}
```
plafer, line 41: This is currently (slightly) incomplete, in that there's currently no nice API for the debugger to understand what's going to happen in the next clock cycle. The ideal API makes those fields private, and only exposes accessors to the kind of information the debugger cares about.

Specifically, the `ContinuationStack` currently doesn't expose any `peek_next_continuation()` method. But even that wouldn't be enough, since some calls to `step()` will execute more than 1 `Continuation` - due to some continuations not resulting in a clock increment. For example, the `AfterExitDecorators(MastNodeId)` continuation runs the "after exit" decorators for the provided node, and that's it.

One potential solution is to expose a `get_next_continuations() -> Vec<Continuation>`, which return all the continuations that will execute in the next call to `step()`. Is that relevant to the debugger though? Or would it be better to just return the continuation that increments the clock cycle (e.g. we would skip `AfterExitDecorators` and just return `ResumeBasicBlock {node_id, batch_id, op_index}`)?

cc @bitwalker happy to discuss more offline if needed

processor/src/fast/basic_block.rs
```diff
@@ -44,31 +45,70 @@ impl FastProcessor {
...
+            continuation_stack,
+            current_forest,
+            tracer,
+            stopper,
+        )
+        .await
+    }
+
+    /// Executes the give basic block node starting from the RESPAN preceding the specified batch.
+    #[expect(clippy::too_many_arguments)]
```
huitseeker: This should be configured in the `.clippy.toml` file at the root of the repo.

processor/src/continuation_stack.rs
```diff
@@ -58,6 +83,11 @@ impl ContinuationStack {
         Self { stack }
     }
 
+    /// Pushes a continuation onto the continuation stack.
```
huitseeker, line 119: How are debuggers supposed to inspect the continuation stack? I see `pop_continuation()` which mutates state, but I'm not finding a way to peek at what's queued up or check if the stack is empty without consuming entries.

Maybe I'm missing something - is there another API that exposes this? Or would adding methods like `len()`, `is_empty()`, and `peek()` be helpful here?

processor/src/fast/step.rs
```diff
@@ -0,0 +1,77 @@
...
+    fast::FastProcessor,
+};
+
+// RESUME CONTEXT
+// ===============================================================================================
+
+/// The context required to resume execution of a program from the last point at which it was
+/// stopped.
+#[derive(Debug)]
+pub struct ResumeContext {
```
huitseeker: I see you mentioned the debugger API is incomplete. The fields here are public, but figuring out "what operation executes next" from the continuation stack seems tricky. `ResumeBasicBlock` requires looking up the operation in the forest, while other continuations like `AfterExitDecorators` don't map to operations at all.

Would a helper method make sense here? Something like:

```rust
impl ResumeContext {
    pub fn next_operation(&self) -> Option<OperationInfo> {
        // ...
    }
}
```

Just thinking this might make things easier for debugger authors.

processor/src/fast/basic_block.rs
```diff
@@ -79,67 +119,112 @@ impl FastProcessor {
                 );
 
                 // Corresponds to the RESPAN operation added to the trace.
-                self.increment_clk(tracer);
+                self.increment_clk(tracer, stopper).map_break(|_| {
```
huitseeker, line 125: This is subtle and worth documenting. When `step()` stops here during RESPAN, the clock has already incremented (before the stopper check in `increment_clk`). So the continuation needs to be `ResumeBasicBlock` to start executing the batch, not `Respan` which would execute RESPAN again.

This took careful reading to get. A comment explaining the timing would help future maintainers understand why we return `ResumeBasicBlock` here instead of a `Respan` continuation.

processor/src/fast/basic_block.rs
```diff
@@ -165,58 +250,150 @@ impl FastProcessor {
...
+    basic_block_node: &BasicBlockNode,
+    op_batch: &OpBatch,
+    node_id: MastNodeId,
+    batch_index: usize,
+    op_idx_in_batch: usize,
+) -> Continuation {
+    let last_op_idx_in_batch = op_batch.ops().len() - 1;
+    let last_batch_idx_in_block = basic_block_node.num_op_batches() - 1;
+
+    if op_idx_in_batch < last_op_idx_in_batch {
```
huitseeker, line 389: One thing that's not obvious: when this is called from `execute_op_batch` at line 272, `op_idx_in_batch` is the operation that just executed. So `op_idx_in_batch + 1` in the first branch is the next operation to execute, not the current one.

Worth adding a comment about that?

processor/src/continuation_stack.rs
plafer: Please see discussion [here](https://github.com/0xMiden/miden-vm/pull/2440#discussion_r2604027260)

processor/src/fast/basic_block.rs
plafer: Yes, it is indeed subtle. I tried to provide an intuition over the `BreakReason::Stopped` variant ([here](https://github.com/0xMiden/miden-vm/blob/7b15eba9076c13d4b0a5e4db65c19ea1ab7d49cd/processor/src/fast/step.rs#L60-L68)). Is that one helpful? Should it be moved elsewhere?

processor/src/fast/basic_block.rs
plafer: Done

processor/src/fast/basic_block.rs
plafer: Added a bunch of comments, renamed the function and buffed up the docstring

processor/src/fast/basic_block.rs
plafer: Also expanded this RESPAN comment

processor/src/fast/step.rs
plafer: Went ahead and exposed 2 different accessors to `ContinuationStack`: a `peek_next_continuation()` method which gives you the next continuation to execute, as well as a method that gives you an iterator over all continuations to execute in the next clock cycle.

I also made `ResumeContext` fields `pub(crate)` with accessors for all fields, to make sure they aren't mutated outside the processor.

processor/src/fast/basic_block.rs
huitseeker: Thanks!

bitwalker (commented):
Question: I wonder if it would be possible to make `Stopper` the primary way for the debugger to manage breakpoints. As implemented here currently, the caller of `step` doesn't have any way to provide their own `Stopper`, but the `Stopper` itself also doesn't have access to enough information about where/when it's being asked if it should stop execution (e.g. an operation index that we could then use to query decorators/debug info to make a decision about whether there is a relevant breakpoint at that step).

It's not that the debugger can't achieve the same goal by driving the processor step-by-step and extracting this information using the next continuation(s), but it does feel like the thing that should be making the decision to stop or not is the caller of `step`. That said, for it to really feel maximally useful, I think `Stopper` would be better defined as:

```rust
pub trait Stopper {
    type Yield;

    fn should_stop(&self, processor: &FastProcessor, op: OperationIndex) -> ControlFlow<Self::Yield>;
}
```

This would let the `Stopper` communicate with the driver of the processor (i.e. the caller of `step`) using its own "break reason", for example, the set of breakpoints that triggered the stop. It isn't clear to me whether it is actually necessary to have the caller of `step` manage the `ResumeContext`, but if it is, then I'd imagine the signature of `step` would look something like:

```rust
impl FastProcessor {
    pub async fn step<S>(
        &mut self,
        host: &mut impl AsyncHost,
        stopper: &S,
        resume_ctx: ResumeContext,
    ) -> Step<<S as Stopper>::Yield>
    where
        S: Stopper;
}

pub enum Step<T> {
    /// Execution completed successfully
    Finished,
    /// Execution was interrupted by `Stopper::should_stop`
    Stopped {
        /// The context needed to resume execution from the break point
        resume_with: ResumeContext,
        /// The value yielded by `Stopper::should_stop` at the break point
        yielded: T,
    },
    /// Execution was terminated by an error
    Failed(ExecutionError),
}
```

Thoughts?

bitwalker (approved):
Marking this approved, as I believe this provides everything the debugger needs for now. Future improvements to the API as mentioned in my other comment are something we can explore down the road.

plafer:
As discussed offline, let's leave this as an improvement, given that the `step()` function provides the debugger what it needs. One reason is also that exactly what `OperationIndex` needs to look like is still being worked on.

PR #2441, opened 2025-12-08
ci: Automate release

### LINKED ISSUE #2227: Automate release process

huitseeker:
- Automates the release ~~with [release-plz](https://github.com/release-plz/release-plz)~~ `cargo publish --workspace`, with a composite action to ensure the dry-run and publish workflows do not diverge,
- moves the check-msrv workflow to the release workflow, away from the PR workflow,
- lots of this is inspired with how this works on miden-base (see issue below).

> [!IMPORTANT]
> Before using this in production, we'll need to setup [trusted publishing](https://crates.io/docs/trusted-publishing).

Fixes #2227

### DIFF
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -69,14 +69,13 @@ runs:
         chmod +x scripts/check-msrv.sh
         ./scripts/check-msrv.sh
 
-  
     # Dry-run vs real publish
     - name: Dry-run workspace publish
       if: ${{ inputs.mode == 'dry-run' }}
       shell: bash
       run: |
         echo "Running cargo publish --workspace --dry-run"
-        cargo publish --workspace --all-features --dry-run
+        cargo publish --workspace --dry-run
 
     - name: Publish workspace crates
       if: ${{ inputs.mode == 'publish' }}
@@ -85,4 +84,4 @@ runs:
         CARGO_REGISTRY_TOKEN: ${{ env.CARGO_REGISTRY_TOKEN }}
       run: |
         echo "Publishing workspace crates to crates.io"
-        cargo publish --workspace --all-features
+        cargo publish --workspace


bobbinth:
Without having looked at the code: is getting rid of `release-plz` not a good idea? I thought one of the goals was to just run `cargo publish --workspace` - but maybe that doesn't work for some reason?

huitseeker:
@bobbinth
1. I was not too keen on reimplementing the version bumping `cargo publish --workspace` would require, just to save the precisely 4 lines of footprint that release-plz has in this action, and
2. was looking to benefit from uniformity with `miden-base`.
I'm yet to hear things that are broken or unpleasant about `release-plz`. Put @PhilippGackstatter in review for his feedback.

bobbinth:
> I was not too keen on reimplementing the version bumping `cargo publish --workspace` would require, just to save the precisely 4 lines of footprint that release-plz has in this action, and

We usually increment versions manually - for example, the current `next` branch already has version numbers incremented to what needs to be released (i.e., v0.21.0). So, in places when we do use `release-plz` we skip the auto-incrementing over versions.

> was looking to benefit from uniformity with `miden-base`.

We are moving away from `release-plz` in other repos. `miden-client` has already done this, and i believe `miden-faucet` is also doing this. I think `miden-node` and `miden-base` are next, and I believe there is also a plan to do this in the compiler repo.

mmagician:
I think the idea was to move away from external tooling if they don't provide additional value over - and since `cargo publish --workspace` is relatively new, the goal was to abandon `release-plz` in favor of the more "native" solution.

If I'm not mistaken, both tools suffer from the inability to detect circular dev-dependencies, and require at least a certain level of configuration/customization.

huitseeker:
@bobbinth @mmagician Updated to no longer use release-plz. I'd like to suggest moving this to [trusted publishing](https://crates.io/docs/trusted-publishing) over letting the action own a cargo registry token.

scripts/publish-release.sh
bobbinth, line 1: I'd keep this file until we know that the automated process is fully working. If all is good, we can delete it after the next release.

.github/actions/workspace-release/action.yml
```diff
@@ -0,0 +1,88 @@
...
+        cargo publish --workspace --all-features --dry-run
+
+    - name: Publish workspace crates
+      if: ${{ inputs.mode == 'publish' }}
+      shell: bash
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ env.CARGO_REGISTRY_TOKEN }}
+      run: |
+        echo "Publishing workspace crates to crates.io"
+        cargo publish --workspace --all-features
```
bobbinth: Do we need to add `--all-features` here? (i.e., would it publish incorrectly if we don't specify this?).

huitseeker: No, the published assets would be identical, but the build check done as part of `cargo publish` will be stronger (i.e. under `--all-features`). If you think that might be brittle, happy to remove.

plafer: I think our `--all-features` is currently broken (very unfortunately) due to our use of `maybe-async` (whose `async` feature is not additive).

I double-checked that `cargo check --all-features` currently fails on `next`.

huitseeker:
@bobbinth Yes, your config setup looks good to me.

bobbinth:
Seems like the `dry-run` workflow is failing: https://github.com/0xMiden/miden-vm/actions/runs/20114757434 - is this to be expected at this stage?

bobbinth:
And all relevant crates are now set up for trusted publishing.

PR #2447, opened 2025-12-10
feat: add cached commitment to MastForest

### LINKED ISSUE #2346: mast: stable MastForest commitment
One of the blockers for #1949 was that that there is no stable (or fast) way of obtaining the commitment of a MastForest. 

At the time, the only approach we considered was to recompute the sequential commitment to all the roots every time the commitment was requested. This would be inefficient as the processor would be requesting this commitment whenever it would encounter a new MastForest.

While we considered caching this commitment as an additional field, we would've had to recompute this each time the MastForest was mutated. With the recent PRs by @huitseeker which essentially makes the forests immutable, this should now be a lot easier.

huitseeker:
Adds a `commitment()` method that returns a commitment to all roots. The value is cached using `OnceLock` and recomputed when roots or nodes change. Advice map and decorator changes don't affect the cache.

Changes `compact()` to take ownership since it returns a new forest.

Closes #2346

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -8,9 +8,9 @@ use core::{
     fmt,
     ops::{Index, IndexMut},
 };
-use std::sync::OnceLock;
 
 pub use miden_utils_indexing::{IndexVec, IndexedVecError};
+use miden_utils_sync::OnceLockCompat;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -76,7 +76,7 @@ pub struct MastForest {
 
     /// Cached commitment to this MAST forest (commitment to all roots).
     /// This is computed lazily on first access and invalidated on any mutation.
-    commitment_cache: OnceLock<Word>,
+    commitment_cache: OnceLockCompat<Word>,
 }
 
 // ------------------------------------------------------------------------------------------------
@@ -89,7 +89,7 @@ impl MastForest {
             roots: Vec::new(),
             advice_map: AdviceMap::default(),
             debug_info: DebugInfo::new(),
-            commitment_cache: OnceLock::new(),
+            commitment_cache: OnceLockCompat::new(),
         }
     }
 }
@@ -417,9 +417,7 @@ impl MastForest {
     /// transitively includes all of its descendants. Therefore, a commitment to all roots
     /// is a commitment to the entire forest.
     pub fn commitment(&self) -> Word {
-        *self.commitment_cache.get_or_init(|| {
-            self.compute_nodes_commitment(&self.roots)
-        })
+        *self.commitment_cache.get_or_init(|| self.compute_nodes_commitment(&self.roots))
     }
 
     /// Returns the number of nodes in this MAST forest.
--- a/crates/utils-sync/src/lib.rs
+++ b/crates/utils-sync/src/lib.rs
@@ -5,12 +5,14 @@ extern crate alloc;
 #[cfg(any(feature = "std", test))]
 extern crate std;
 
+pub mod once_lock;
 pub mod racy_lock;
 pub mod rw_lock;
 
 #[cfg(feature = "std")]
 pub use std::sync::LazyLock;
 
+pub use once_lock::OnceLockCompat;
 #[cfg(feature = "std")]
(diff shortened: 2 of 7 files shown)


core/src/mast/mod.rs
```diff
@@ -8,6 +7,7 @@ use core::{
     fmt,
     ops::{Index, IndexMut},
 };
+use std::sync::OnceLock;
```
bobbinth: I'm not sure this will work: we use `MastForest` in `mo_std` contexts (e.g., when running the VM in the browser).

core/src/mast/mod.rs
```diff
@@ -72,6 +72,10 @@ pub struct MastForest {
     /// Debug information including decorators and error codes.
     /// Always present (as per issue #1821), but can be empty for stripped builds.
     debug_info: DebugInfo,
+
+    /// Cached commitment to this MAST forest (commitment to all roots).
+    /// This is computed lazily on first access and invalidated on any mutation.
+    commitment_cache: OnceLock<Word>,
```
bobbinth: Could we use the same approach but instead of `OnceLock` use `Option`?

plafer: With `Option` we couldn't use the same approach, because it would force the `commitment()` method to take `&mut self` instead of `&self` (no interior mutability). AFAIK, I don't think this solution works in `no_std` environments. Hence either 

1. we provide this cache in `std` environments only, 
2. we use `Option<Word>`, to be set manually with a `compute_commitment_cache(&mut self)` or something

I think (2) is preferable, as the `MastForestBuilder` could compute the commitment cache once the `MastForest` is built, and only then put it behind a `Arc`.

core/src/mast/mod.rs
huitseeker: Yeah, CI should have reported the failure at https://github.com/0xMiden/miden-vm/actions/runs/20086295256/job/57624390072 hours ago .. 🤦

core/src/mast/mod.rs
huitseeker: I've gotten an alternate implementation working, based on the battle-tested `once_cell::racebox::OnceBox`.

crates/utils-sync/src/lib.rs
```diff
@@ -5,12 +5,14 @@ extern crate alloc;
...
 extern crate std;
 
+pub mod once_lock;
 pub mod racy_lock;
 pub mod rw_lock;
 
 #[cfg(feature = "std")]
 pub use std::sync::LazyLock;
 
+pub use once_lock::OnceLockCompat;
```
plafer, line 15: Similar to what we did with `LazyLock`, we could just call this one `OnceLock`

plafer (approved):
LGTM!

crates/utils-sync/src/lib.rs
huitseeker: I don't think that's wise, considering:
- it has an API much narrower than `OnceLock`,
- it has (due to the `Box`) a very different performance footprint in `no-std` environments.

adr1anh (approved):
Looks good, and this should allow us to implement https://github.com/0xMiden/miden-vm/issues/1949 efficiently.

core/src/mast/mod.rs
```diff
@@ -72,6 +73,10 @@ pub struct MastForest {
     /// Debug information including decorators and error codes.
     /// Always present (as per issue #1821), but can be empty for stripped builds.
     debug_info: DebugInfo,
+
+    /// Cached commitment to this MAST forest (commitment to all roots).
+    /// This is computed lazily on first access and invalidated on any mutation.
+    commitment_cache: OnceLockCompat<Word>,
```
bobbinth, line 79: I think this approach is fine for now, but in the longer term, I'd love for this to be just `commitment: Word`. To do this, we'll need to make `MastForest` largely immutable after construction - i.e., methods like `make_root()` and `remove_node()` would go away. This would require transforming `MastForestBuilder` into a proper builder - i.e., it shouldn't manage `mast_forest` as an internal filed, but would be converted into a `MastForest` once the assembly is complete.

huitseeker: See #2473

Issue #2446, opened 2025-12-10
Investigate creating a 1-1 mapping operation <-> AsmOp


huitseeker:
Not for this PR, but maybe in the future we could make it so that every operation covered by an `AsmOp` maps to this `AsmOp`. Previously, we didn't want to do it because it would have meant duplicating `AsmOp`s for every operation, but now that the structure is separate and ID-based, maybe that's easier.

We should do this only if this simplifies the codebase though.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2430#discussion_r2604723088_

plafer:
> We should do this only if this simplifies the codebase though.

IIUC this results in a slightly simpler `MastForest::get_assembly_op()` method, and nothing else, right? In that case, I can't say that I *really* see the benefit of duplicating the `AsmOp` decorators, but will defer to @huitseeker whether or not this is desirable (since you've been getting your hands dirty with that part of the codebase recently).

bobbinth:
> In that case, I can't say that I _really_ see the benefit of duplicating the `AsmOp` decorators

Without having looked at the code, the idea was not to duplicate decorators (this will probably affect the binary size quite significantly), but to link multiple operations to the same decorator. But again, if this doesn't complicate things too much.

plafer:
> Without having looked at the code, the idea was not to duplicate decorators (this will probably affect the binary size quite significantly), but to link multiple operations to the same decorator.

Hmmm, not sure I follow. IIUC, the current implementation does as you suggest; an assembly operation might compile down to multiple operations, and each of those operations map back to the same `AssemblyOp` decorator (using the `AssemblyOp.num_cycles` field).

For me, "creating a 1-1 mapping operation <-> AsmOp" suggests duplicating the `AssemblyOp` decorators, such that each operation gets its own `AssemblyOp` copy (i.e. the "1-1 part"). This is what's being done in #2455.

bobbinth:
> Hmmm, not sure I follow. IIUC, the current implementation does as you suggest; an assembly operation might compile down to multiple operations, and each of those operations map back to the same `AssemblyOp` decorator (using the `AssemblyOp.num_cycles` field).

[...]

PR #2455, opened 2025-12-11
refactor: create 1-1 mapping between operations and AsmOp decorators

PivasDesant:
Implements 1-1 mapping for operations and AsmOp decorators (#2446)

Before this change, only the first operation was linked to an AsmOp when an instruction compiled to multiple operations. Now all operations are linked.

The implementation adds decorator links for all operations in `set_instruction_cycle_count()`. This also simplifies `get_assembly_op()` since we can check index equality directly instead of doing range checks.

No data duplication since we reuse the same `DecoratorId` for all operations covered by an AsmOp.

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -606,8 +606,7 @@ impl MastForest {
                 // to that AsmOp. We need to find the minimum op_idx for each unique AsmOp decorator
                 // to determine the start of its range, then check if target_op_idx falls within
                 // that range.
-                let mut asmop_ranges: std::collections::HashMap<DecoratorId, (usize, usize)> =
-                    std::collections::HashMap::new();
+                let mut asmop_ranges: BTreeMap<DecoratorId, (usize, usize)> = BTreeMap::new();
 
                 // For Block nodes with target_op_idx, we should only consider operation-indexed
                 // decorators (not before_enter/after_exit) for operation-specific lookups


crates/assembly/src/basic_block_builder.rs
```diff
@@ -161,6 +165,14 @@ impl BasicBlockBuilder<'_> {
         } else {
             assembly_op.set_num_cycles(cycle_count as u8);
 
+            // Create a 1-1 mapping: link the AsmOp to all operations it covers.
+            // The first operation is already linked at op_start, so we add links for
+            // operations from op_start + 1 to op_start + cycle_count - 1.
+            // We insert them in reverse order to maintain the sorted order of the decorators list.
+            for op_idx in ((op_start + 1)..(op_start + cycle_count)).rev() {
+                self.decorators.insert(self.last_asmop_pos + 1, (op_idx, assembly_op_id));
```
huitseeker, line 172: For clarity after this approach was discussed in #2446: This line reuses the same `assembly_op_id` for all operations covered by the AsmOp. That's the correct approach per issue #2446 - we're linking multiple operations to one decorator, not duplicating the decorator itself.

For a 3-cycle instruction starting at operation 5, this creates:
```
decorators = [..., (5, id), (6, id), (7, id), ...]
```
All three entries share the same decorator ID, so no memory duplication of the AsmOp itself.

The reverse insertion order maintains the sorted decorator list without extra sorting cost.

core/src/mast/mod.rs
```diff
@@ -653,7 +650,7 @@ impl MastForest {
 // TEST HELPERS
 // ================================================================================================
 
-#[cfg(test)]
+#[cfg(feature = "testing")]
```
huitseeker: Consider using `#[cfg(any(test, feature = "testing"))]` instead of just `#[cfg(feature = "testing")]`.

The current approach means that unit tests in the `core` crate itself cannot use `all_decorators()` when running `cargo test` without explicitly enabling the `testing` feature.

The suggested alternative is the standard pattern for test helpers that need to be shared across crates while still being available for local unit tests.

huitseeker (approved):
LGTM, thanks for the test!

PivasDesant:
Thanks for the review! So i fixed test naming and also fixed fmt

bobbinth:
@PivasDesant - seems like build/tests are currently failing on this PR. Would you be able to fix this?

core/src/mast/mod.rs
```diff
@@ -602,18 +602,58 @@ impl MastForest {
...
+                let mut asmop_ranges: BTreeMap<DecoratorId, (usize, usize)> = BTreeMap::new();
+
+                // For Block nodes with target_op_idx, we should only consider operation-indexed
+                // decorators (not before_enter/after_exit) for operation-specific lookups
+                match node {
+                    MastNode::Block(_) => {
+                        // For Block nodes, use only operation-indexed decorators
+                        let op_iter = self
+                            .decorator_links_for_node(node_id)
+                            .expect("Block node must have some valid set of decorator links");
```
huitseeker: This `.expect()` will panic if `decorator_links_for_node` fails for a Block node. The previous code was more graceful since it operated on `decorator_links` which was already successfully computed earlier in the function.

Would returning `None` here be safer? Something like:

```rust
let Ok(op_iter) = self.decorator_links_for_node(node_id) else {
    return None;
};
```

Or is the invariant "Block nodes always have decorator links" guaranteed by construction?

core/src/mast/mod.rs
```diff
@@ -602,18 +602,58 @@ impl MastForest {
...
-                        // with the assembly op.
-                        if target_op_idx >= op_idx
-                            && target_op_idx < op_idx + assembly_op.num_cycles() as usize
-                        {
-                            return Some(assembly_op);
+                // With the 1-1 mapping, every operation covered by an AsmOp is explicitly linked
+                // to that AsmOp. We need to find the minimum op_idx for each unique AsmOp decorator
+                // to determine the start of its range, then check if target_op_idx falls within
+                // that range.
+                let mut asmop_ranges: BTreeMap<DecoratorId, (usize, usize)> = BTreeMap::new();
```
huitseeker, line 609: With the 1-1 mapping now in place, could we simplify this function? The current implementation builds a `BTreeMap` and does two passes, but since every operation is now directly linked to its `AsmOp`, we can query by index directly.

Something like:

```rust
pub fn get_assembly_op(
    &self,
    node_id: MastNodeId,
    target_op_idx: Option<usize>,
) -> Option<&AssemblyOp> {
    let node = &self[node_id];

    let find_asmop = |ids: &[DecoratorId]| {
        ids.iter().find_map(|&id| match &self[id] {
            Decorator::AsmOp(asm_op) => Some(asm_op),
            _ => None,
        })
    };

    match target_op_idx {
        Some(idx) => match node {
            MastNode::Block(_) => find_asmop(self.decorators_for_operation(node_id, idx)),
            _ => match idx {
                0 => find_asmop(self.before_enter_decorators(node_id)),
                1 => find_asmop(self.after_exit_decorators(node_id)),
                _ => None,
            },
        },

        None => {
            // Search order: before_enter, then per-op (for blocks), then after_exit
            find_asmop(self.before_enter_decorators(node_id))
                .or_else(|| match node {
                    MastNode::Block(block) => (0..block.num_operations() as usize)
                        .find_map(|i| find_asmop(self.decorators_for_operation(node_id, i))),
                    _ => None,
                })
                .or_else(|| find_asmop(self.after_exit_decorators(node_id)))
        }
    }
}
```

This removes the `Box<dyn Iterator>` and `BTreeMap`, and the common case (Block + target index) becomes a direct lookup.

plafer: I'm still at a loss as to why we're doing this in the first place. Bobbin's [answer to this question](https://github.com/0xMiden/miden-vm/issues/2446#issuecomment-3672045811) was that it would simplify the decorator finding logic - but comparing this with what we currently have, I find them pretty equally complex. IMO, most of the complexity comes from the fact that we have "before enter", "after exit", and sometimes "operations" decorators. The `num_cycles` check is not really the issue - it only applies to "operations" decorators, and is honestly [pretty simple](https://github.com/0xMiden/miden-vm/blob/581ab6f41fa130b21389d37ab4bf24b9ee0c9d3d/core/src/mast/mod.rs#L612-L613).

In contrast, the clear downside of this new approach is increased memory usage.

Am I missing something?

huitseeker: Perhaps it was worth it as an experiment which has shown it may be best to not push this one further? WDYT @bobbinth ? 
Though I would add that the above is a bit complex to avoid allocations, and non-allocative code is always a bit involved.

bobbinth:
I believe this is superseded by https://github.com/0xMiden/miden-vm/pull/2606 - but please feel free to re-open if needed.

Issue #2411, opened 2025-12-04
remove Owned variant of DecoratorStorage 


huitseeker:
Removing `decorator_store` from the nodes is definitely the ultimate goal of this refactoring series. The main obstacle is that we still need the Owned variant for three key scenarios:

1. **Deserialization** - Nodes come back from disk/storage with Owned decorators
2. **Round-trip conversions** - Node → Builder → Node transformations needed for forest operations like merge and rewrite
3. **Testing** - Tests need to create nodes without requiring a full forest

We explored eliminating `Owned` from `add_to_forest()` (see the branch linked in [this above conversation](https://github.com/0xMiden/miden-vm/pull/2323/files#r2487923080)), but that doesn't solve the three cases above.

This is part of a larger refactoring tracked in #1497. That refactoring would eliminate the need for `DecoratorStore` in nodes entirely by storing everything in a centralized arena.

The path forward is likely:
1. Complete this PR series to hoist decorators to the forest (done after this PR),
2. Implement DebugInfo struct (#1776),
3. Break backwards-compatible serialization to split decorator info (#1580).
4. Rewrite the assembler's recursive descent, and refactor to fully commit to decorator info (#1497) which would remove decorator_store from nodes

So yes, it's absolutely the goal, but it's a multi-step process.

_Originally posted by @huitseeker in https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2508543177_

huitseeker:
Update on dependencies:

- #1776 (DebugInfo) - done
- #1580 backend - accomplished in PRs #2470, #2469, #2466 (front-end still TODO)

What remains: rewriting the Assembler to no longer depend on Owned nodes. I don't think this is high-priority at this stage (i.e. I'm comfortable bumping the milestone).

bobbinth:
> What remains: rewriting the Assembler to no longer depend on Owned nodes. I don't think this is high-priority at this stage (i.e. I'm comfortable bumping the milestone).

Agreed. Moving to the next milestone.

The actual refactoring needed for this could look like #2473.

PR #2459, opened 2025-12-11
fix: decorators not copied when calling procedures from statically linked libraries

### LINKED ISSUE #2456: `DecoratorId out of bounds` bug
Attempt to run the `miden-base` using the latest VM version (`v0.20.0`) in some cases resulted in errors during the scripts execution.

It seems like an execution of a program which uses `call` inside results in `DecoratorId out of bounds` error. The simplest example could be created from the `execute_tx_view_script()` [test](https://github.com/0xMiden/miden-base/blob/4b08b9c916bb3a1ecc7e509d51564e9860c1bbc8/crates/miden-testing/src/kernel_tests/tx/test_tx.rs#L623) in the `miden-base`. There we attempt to execute this program:

```text
    begin
        push.1.2
        call.module_1::foo
        exec.sys::truncate_stack
    end
```

If the `call.module_1::foo` is removed, this error will disappear.

It is doubly strange that if the transaction script is slightly modified, this error disappears as well. It's enough to add two drops after push like so:

```
    begin
        push.1.2 drop drop
        call.module_1::foo
        exec.sys::truncate_stack
    end
```

So it feels like adding this extra instructions modifies the way decorators are created during the compilation of this script.

huitseeker:
The issue happens because MastForestBuilder::ensure_external_link copies procedure nodes from the library's forest to the program's forest, but decorators stay behind. Each node keeps its old decorator IDs pointing to decorators that don't exist in the new forest.

The fix follows the same pattern as MastForestMerger::build_with_remapped_children.

Fixes #2456.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.1 (2025-12-14)
+
+- Fix issue where calling procedures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).
+
 ## 0.20.0 (2025-12-05)
 
 #### Enhancements
@@ -13,8 +17,6 @@
 - Added new `peek` procedure for the `std::collections::smt` module ([#2387](https://github.com/0xMiden/miden-vm/pull/2387)).
 - Added new `pad_and_hash_elements` procedure to the `std::crypto::hashes::rpo` module ([#2395](https://github.com/0xMiden/miden-vm/pull/2395)).
 - Added padding option for the `adv.push_mapvaln` instruction ([#2398](https://github.com/0xMiden/miden-vm/pull/2398)).
-- Added new `FastProcessor::step()` method that executes a single clock cycle ([#2440](https://github.com/0xMiden/miden-vm/pull/2440))
-- Fix issue where calling procesures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).
 
 #### Changes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -175,9 +175,9 @@ checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
 name = "base64ct"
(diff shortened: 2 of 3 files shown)


crates/assembly/src/mast_forest_builder.rs
bobbinth, line 1: Not for this PR, but I think the `MastForestBuilder` has become very complex and difficult to reason about (at least without spending significant time understanding all edge cases). I think it would be good to refactor it so that we can reduce the complexity and simultaneously address https://github.com/0xMiden/miden-vm/issues/2411.

One potential approach to this is to stop relying on `MastForest` and `MastNodeId` structs during assembly, and instead, make all references fingerprint-based (where fingerprint similarly to how it is defined currently - i.e., a MAST root + commitment to the related decorators).

So, instead of having `mast_forest` the `MastForestBuilder` would look something like:

```rust
pub struct MastForestBuilder {
    nodes: BTreeMap<MastNodeFingerprint, MastNodeBuilder>,
    ...
}
```
And then all builders would be defined in terms of fingerprints. For example, a `JoinNodeBuilder` could look like so:
```rust
pub struct JoinNodeBuilder {
    children: [MastNodeFingerprint; 2],
    before_enter: Vec<DecoratorId>,
    after_exit: Vec<DecoratorId>,
    digest: Option<Word>,
}
```

We could even maybe store the actual decorators in the node builders rather than IDs - e.g., something like:
```rust
pub struct JoinNodeBuilder {
    children: [MastNodeFingerprint; 2],
    before_enter: Vec<Decorator>,
    after_exit: Vec<Decorator>,
    digest: Option<Word>,
}
```
This way, cloning of MAST node builders becomes trivial and we never need to worry about re-mapping anything.

Then, once the assembly is complete, we can reduce this structure to a `MastForest` and calculate the node IDs at that time.

The obvious downside here is that this will probably make assembly slower, but unless we are slowing things down by something like 10x, I would prefer code simplicity over performance as assembly time is not super important in our context.

The above is just a sketch, and i'm sure there are some nuances that make it more complicated than what I described - but my intuition is that the overall structure will come out quite a bit simpler than what we have now.

huitseeker: #2473

PR #2460, opened 2025-12-11
ci: replace cargo publish dry-run with cargo package

huitseeker:
`cargo publish --workspace --dry-run` fails for workspace crates with
cross-dependencies because it resolves dependencies from crates.io
registry instead of using local workspace members. This causes errors
when local code has changes not yet published to the registry.

Replacing with `cargo package --workspace --no-verify` provides equivalent
validation (manifest correctness, file inclusion checks, packaging) while
properly using local workspace dependencies.

### DIFF
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -59,13 +59,13 @@ runs:
         chmod +x scripts/check-msrv.sh
         ./scripts/check-msrv.sh
 
-    # Dry-run vs real publish
-    - name: Dry-run workspace publish
+    # Package vs real publish
+    - name: Verify workspace packages
       if: ${{ inputs.mode == 'dry-run' }}
       shell: bash
       run: |
-        echo "Running cargo publish --workspace --dry-run"
-        cargo publish --workspace --dry-run
+        echo "Verifying workspace crates can be packaged"
+        cargo package --workspace --no-verify
 
     - name: Publish workspace crates
       if: ${{ inputs.mode == 'publish' }}
--- a/.github/workflows/workspace-dry-run.yml
+++ b/.github/workflows/workspace-dry-run.yml
@@ -1,4 +1,4 @@
-name: Workspace release dry-run
+name: Verify workspace packaging
 
 on:
   push:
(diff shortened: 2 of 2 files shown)


Mirko-von-Leipzig:
Could you link to a `rust` issue for this? I tried finding one and it seems like such an obvious requirement? Unless you mean it fails for circular dependencies?

huitseeker:
@Mirko-von-Leipzig I'm confused as to what issue you are requiring a link for, or what the obvious requirement would be in your mind.

`cargo publish --dry-run` relies on publish dependencies as an intended part of its `--dry-run` behavior, and works well on its own, unless you add the `--workspace` flag *and* cross-crate dependencies in that `workspace`. There, the `--dry-run` fails because it can't find the published dependency when compiling the dependent, whereas the non dry-run would work (by providing that published dependency). All of this is, afaict, intended behavior.

I think what @bobbinth is suggesting here is propagating the behavior of this PR: the `cargo package` job which, as part of CI runs upon merges to `next`, tries to check all crates can be packaged, testing most of the publication steps (though not the dependency on published crates). This gives a good (but of course incomplete) indication that a future `cargo publish --workspace` is likely to be successful.

Mirko-von-Leipzig:
> @Mirko-von-Leipzig I'm confused as to what issue you are requiring a link for, or what the obvious requirement would be in your mind.

I was hoping there was some `cargo` or `rust` issue where this problem was described. To me it seems weird that `cargo publish --dry-run` would choke on anything realistic because that's its entire purpose. I also don't quite understand what _cross-crate_ means -- because of course a workspace has multiple crates which depend on each other? Why else have a workspace really.

Given that there is some weird behavior with `cargo publish` then sure lets propagate, I just wanted to know more :)

As a sort of lead-in, I'm hoping we can get the node (and I guess other repos) to a place where one can do something like:

1. `git tag vX.Y.Z` (and push it)
2. This triggers a release workflow
    - Does a dry-run of all required things. In the node this includes
      - `cargo publish --dry-run`
      - `docker build`
      - linting, testing, etc
      - building debian packages
      - integration tests
      - creates a draft github release
    - Once this succeeds, we then have the option of clicking release which does it for real
    
So having `cargo package` do it _mostly_ correct is a bit of a downgrade

huitseeker:
@Mirko-von-Leipzig you're right to double-click on this, let me investigate a few things about this failure and check it was due to tooling, rather than organic (and due to something specific to our crate).
Relevant links on the cargo side are:
- https://github.com/rust-lang/cargo/pull/15636
- https://github.com/rust-lang/cargo/pull/15711

Issue #2406, opened 2025-12-02
Remove deprecated `Process`


plafer:
As discussed multiple times offline, we should remove the now deprecated `Process` in favor of the `FastProcessor` and `build_trace()` infrastructure.

As mentioned in https://github.com/0xMiden/miden-vm/pull/2400#issuecomment-3602499146, we should also remember to remove the `OverflowTable` history.

Farukest:
Hi @plafer @bobbinth 

I've explored the codebase and understand the scope 
I see it touches 85+ files but I'm happy to take this on. Could you assign it to me?

plafer:
Hey @Farukest, this is a pretty critical one, so we'll take care of it.

PR #2461, opened 2025-12-11
Migrate decoder and operation-specific tests from `Process` to `FastProcessor` (2/3)

plafer:
Work towards #2406

Migrates
- all decoder tests to use `build_trace()` instead of `Process`
- all operation-specific tests are migrated to test the `FastProcessor` instead of `Process`


plafer:
I will merge this PR to reduce the number of in-flight ones; I will gladly address any reviews in a subsequent PR.

PR #2465, opened 2025-12-12
chore: replace hardcoded stack depth with MIN_STACK_DEPTH constant

KitaroMoha563:
Replaces magic number `16` with `MIN_STACK_DEPTH` constant in test helper functions.


bobbinth:
Thank you - but we do not accept PRs w/o associated issues. Also, this is now stale.

Issue #2448, opened 2025-12-10
Simplify MastForest serialization by directly serializing DebugInfo


huitseeker:
## Problem

The MastForest serialization code does unnecessary work. It extracts decorator data from centralized storage, formats it per-node, serializes it, then during deserialization puts it all back into centralized storage.

This round-trip wastes CPU and memory. The wire format should match the in-memory structure.

## Background

The decorator refactoring (#2310, #2323, #2366, #2409) centralized decorator information in the MastForest:

- Nodes added to a forest use `DecoratorStore::Linked` storage (see `core/src/mast/node/basic_block_node/mod.rs:1321`)
- Decorator data lives in `DebugInfo`:
  - `OpToDecoratorIds` - operation-indexed decorators (CSR-based sparse storage)
  - `NodeToDecoratorIds` - before_enter/after_exit decorators

All nodes in a serialized MastForest use `Linked` storage. The `Owned` variant only exists for standalone nodes before forest insertion.

## Current Round-Trip

**Serialization** (`core/src/mast/serialization/mod.rs:128-196`):
```
For each node:
  - Extract decorator info FROM DebugInfo via MastNode::before_enter_decorators(),
    MastNode::after_exit_decorators(), BasicBlockNode::raw_op_indexed_decorators()
  - Format as per-node lists
  - Write to three separate sections
```

**Deserialization** (`core/src/mast/serialization/mod.rs:244-287`):
```
Read three per-node decorator sections
  - Build MastNodeBuilders with Owned-ish storage via
    BasicBlockNodeBuilder::set_decorators() 
  - Call MastNodeBuilder::add_to_forest_relaxed() 
    - Registers decorators back INTO DebugInfo
    - Converts to Linked storage
```

The intermediate `Owned`-ish storage serves little purpose.

## Wire Format Mismatch

Current wire format (per-node):
- Basic block decorator lists section
- Before-enter decorators section
- After-exit decorators section

In-memory structure (centralized):
- Single `OpToDecoratorIds` in DebugInfo
- Single `NodeToDecoratorIds` in DebugInfo

## The Padding Wrinkle

BasicBlockNodes get padded with NOOP operations at creation. Since decorators store indexes into the operation sequence, these indexes need adjustment when padding happens. This is done by `BasicBlockNode::adjust_decorators()` (`core/src/mast/node/basic_block_node/mod.rs:1304`).

The wire format stores **unpadded** operations and decorator indexes. This is important because:
[...]

PR #2466, opened 2025-12-12
feat: serialize `BasicBlock`s in padded representation (1/3)

huitseeker:
As a prelude to #2448, this changes the serialization of `BasicBlock` to reflect the padded contents, to not need to re-batch and pad those blocks again.

The goal of this PR is twofold:
- experiment with and analyze the padded on-disk representation, unlocked by [this comment](https://github.com/0xMiden/miden-vm/pull/2412#discussion_r2611593503) 🎉  — get conversation started about how permissible it is to serialize metadata:
    - One important sub-part of this: in this PR, we're over-consuming bits in the `indptr` representation, not only because each index is represented with one byte (the value is in [0, 72] which fits in 7 bits) but because the groups contain at most 8 ops, so a simple delta-encoding would give us $8 \times 4$ bits $= 32$ bits with an implicit start at 0, instead of $9 \times 8 = 72$. This is not yet implemented, but would cut the ovehead numbers below about in half.
- Groundwork for #2448, which needs to change the over-the-wire representation of decorator info anyway, but has a long paragraph in there on the difficulty of un-padding decorators ("_The Padding Wrinkle_") that disappears in a puff of simplification after this work.

As a consequence of bumping the version number for serialization of `MastForest`, this PR is intended to stay open until we land on the right over-the-wire format. I.e. there's a version bump in there which I don't intend to have in the PRs which will live on top of this. This stack will be three PRs (one PR delta-encoding, one PR #2448)

## Test Data: Miden Standard Library

- 727 total nodes
- 439 basic block nodes (60.4%)
- 284 procedures
- 94,744 operations (with padding)

### Summary

The padded format adds 4.05% size overhead:
- 0.67% from NOOP padding (633 ops)
- 3.38% from batch metadata (33,316 bytes)

Most blocks (92.7%) add ≤34 bytes. 

<details><summary>Size comparison between unpadded and padded serialization formats for MastForest.</summary>

## Size Comparison

| Format | Size | Overhead |
|--------|------|----------|
| Unpadded (`next`) | 838,400 bytes (818.75 KB) | baseline |
| Padded (`serialize-padded-opbatches`) | 872,375 bytes (851.93 KB) | +33,975 bytes (+4.05%) |

The unpadded format cannot guarantee exact OpBatch reconstruction after deserialization.

### NOOP Padding: 633 operations (0.67%)

| Metric | Count |
|--------|-------|
| Total operations | 94,744 |
| NOOP padding | 633 |
| Real operations | 94,111 |

### Batch Metadata: 33,316 bytes (98% of overhead)

| Component | Per Unit | Count | Total |
|-----------|----------|-------|-------|
| Indptr array | 9 bytes | 3,156 batches | 28,404 bytes |
| Padding flags | 1 byte | 3,156 batches | 3,156 bytes |
| Batch count | 4 bytes | 439 blocks | 1,756 bytes |

**Total metadata**: 33,316 bytes (32.54 KB)

**Distribution summary**:
- 92.7% of blocks: ≤34 bytes overhead (1-3 batches)
- 2.3% of blocks: >100 bytes overhead (9+ batches)
- Average: 75.89 bytes per block (skewed by outliers)

Metadata per block: `4 + (num_batches × 10)` bytes

## Wire Format

Each basic block stores:

```
┌─────────────────────────────────────────────────┐
│ Padded Operations (variable)                    │
├─────────────────────────────────────────────────┤
│ Batch Count (u32, 4 bytes)                      │
├─────────────────────────────────────────────────┤
│ Indptr Arrays (9 × u8 per batch)               │
├─────────────────────────────────────────────────┤
│ Padding Flags (1 byte per batch, bit-packed)   │
└─────────────────────────────────────────────────┘
```

**Size**: `ops_size + 4 + (10 × num_batches)` bytes

Script: https://gist.github.com/huitseeker/f957014b15b0ed08a36b7f936079b698
</details>

### DIFF
--- a/crates/lib/core/tests/mast_forest_merge.rs
+++ b/crates/lib/core/tests/mast_forest_merge.rs
@@ -1,4 +1,7 @@
-use miden_processor::MastForest;
+use miden_processor::{
+    MastForest, MastNode, MastNodeExt,
+    utils::{Deserializable, Serializable},
+};
 
 /// Tests that the core library merged with itself produces a forest that has the same procedure
 /// roots.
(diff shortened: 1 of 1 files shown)


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1256,57 +1265,115 @@ fn batch_ops(ops: Vec<Operation>) -> Vec<OpBatch> {
...
 
 // ------------------------------------------------------------------------------------------------
+/// Represents the operation data for a [`BasicBlockNodeBuilder`].
+///
+/// The decorators are bundled with the operation data to maintain the invariant that
+/// decorator indices match the format of the operations:
+/// - `Raw`: decorators have raw (unpadded) indices
+/// - `Batched`: decorators have padded indices
+#[derive(Debug)]
+enum OperationData {
```
bobbinth, line 1272: Question: do we need this "duality" mostly to support the slow processor? That is, is the slow processor the only reason why we need to keep track of the raw (unpadded) indexes?

huitseeker: No, we still support raw indexes for those reasons:
  - Assembly block merging (raw ops & raw decorator indexes)
  - Node creation (raw → padded conversion in initial batching)
  - Fingerprinting (padded → raw conversion for stability, which I will remove, after this stack of PRs)
  
This stack of PRs also removes the use of raw indexes in serialization.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 ## 0.20.1 (2025-12-14)
 
 - Fix issue where calling procedures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).
+- [BREAKING] Change serialization of `BasicBlockNode`s to use padded indices ([#2466](https://github.com/0xMiden/miden-vm/pull/2466/)).
```
plafer: We should move this entry to 0.21.0 (and similar for other 2 stacked PRs)

huitseeker: Done!

plafer:
I think the stack of 3 PRs are good to merge - any reason to hold off?

PR #2468, opened 2025-12-13
ci: reinstate cargo publish --workspace based workflow

huitseeker:
Reverts #2460 and fixes the real issue. Thanks so much @Mirko-von-Leipzig for pushing back [here](https://github.com/0xMiden/miden-vm/pull/2460#issuecomment-3647480360).

## What Happened

PR #2460 replaced `cargo publish --workspace --dry-run` with `cargo package --workspace --no-verify` because we thought cargo was trying to resolve dependencies from crates.io instead of using local workspace members.

That wasn't the problem. The real issue was stale files in `target/package/tmp-registry/` causing cargo to fail with "no hash listed for miden-core" errors.

## The Fix

Clean `target/package` before running cargo publish:

```yaml
- name: Clean packaging directory
  shell: bash
  run: rm -rf target/package
```

## Why This Matters

`cargo publish --dry-run` tests the actual publish flow.


.github/actions/workspace-release/action.yml
```diff
@@ -59,13 +59,20 @@ runs:
...
         ./scripts/check-msrv.sh
 
-    # Package vs real publish
-    - name: Verify workspace packages
+    # Clean packaging directory to avoid stale/corrupted tmp-registry state
+    - name: Clean packaging directory
+      shell: bash
+      run: |
+        echo "Cleaning target/package directory to ensure fresh state"
+        rm -rf target/package
```
Mirko-von-Leipzig, line 67: Interesting; I assume this is an artifact from running `cargo-msrv` checks?

Would `cargo clean` be a more thorough option, or do you think is sufficient and then at least we can leverage some of the msrv compilation cache?

huitseeker: The artefacts created by `cargo msrv` end up in `target/debug`, not `target/package`. Those steps are not related AFAICT.

PR #2470, opened 2025-12-14
feat: direct DebugInfo serialization (3/3)

### LINKED ISSUE #2448: Simplify MastForest serialization by directly serializing DebugInfo

huitseeker:
We were doing wasteful round-trip conversions in MastForest serialization. The old flow extracted decorator data from centralized storage, formatted it per-node for serialization, then reconstructed centralized storage during deserialization.

### Solution

Serialize `DebugInfo` directly in CSR format. Deserialize it first, then use builders to reconstruct nodes that link to the already-populated decorator storage.

Two distinct insertion paths:
- `add_to_forest()`: registers decorators (normal assembly, part of `MastForestContributor`)
- `add_to_forest_relaxed()`: assumes decorators already in `debug_info` (deserialization)

### Prior Work: Padded Serialization

Earlier PRs in this stack switched to padded operation serialization with batch metadata. Operations and decorator indices are written in their internal padded form for exact reconstruction.

This eliminated the "padding wrinkle" - complex conversions between padded and unpadded representations. Decorator indices in memory now match decorator indices on disk, making direct DebugInfo serialization straightforward.

Fixes #2448.

### DIFF
--- a/core/src/mast/debuginfo/decorator_storage.rs
+++ b/core/src/mast/debuginfo/decorator_storage.rs
@@ -327,18 +327,9 @@ impl OpToDecoratorIds {
     pub(super) fn write_into<W: crate::utils::ByteWriter>(&self, target: &mut W) {
         use crate::utils::Serializable;
 
-        // Serialize decorator_ids (DecoratorId serializes as u32)
         self.decorator_ids.write_into(target);
-
-        // Serialize indptr arrays as u32. These are indices into u32-bounded arrays,
-        // so any value > u32::MAX would be invalid.
-        let op_indptr_u32: Vec<u32> =
-            self.op_indptr_for_dec_ids.iter().map(|&idx| idx as u32).collect();
-        op_indptr_u32.write_into(target);
-
-        let node_indptr_u32: Vec<u32> =
-            self.node_indptr_for_op_idx.iter().map(|&idx| idx as u32).collect();
-        node_indptr_u32.write_into(target);
+        self.op_indptr_for_dec_ids.write_into(target);
+        self.node_indptr_for_op_idx.write_into(target);
     }
 
     /// Deserialize OpToDecoratorIds from the source reader.
@@ -348,29 +339,15 @@ impl OpToDecoratorIds {
     ) -> Result<Self, crate::utils::DeserializationError> {
         use crate::utils::{Deserializable, DeserializationError};
 
-        // Deserialize decorator_ids (DecoratorId deserializes from u32)
         let decorator_ids: Vec<DecoratorId> = Deserializable::read_from(source)?;
+        let op_indptr_for_dec_ids: Vec<usize> = Deserializable::read_from(source)?;
+        let node_indptr_for_op_idx: IndexVec<MastNodeId, usize> =
+            Deserializable::read_from(source)?;
 
-        // Deserialize indptr arrays from u32. These are indices into u32-bounded arrays,
-        // so any value > u32::MAX would be invalid.
-        let op_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
-        let op_indptr_for_dec_ids: Vec<usize> =
-            op_indptr_u32.into_iter().map(|idx| idx as usize).collect();
+        let result =
+            Self::from_components(decorator_ids, op_indptr_for_dec_ids, node_indptr_for_op_idx)
+                .map_err(|e| DeserializationError::InvalidValue(e.to_string()))?;
 
-        let node_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
-        let node_indptr_for_op_idx: Vec<usize> =
-            node_indptr_u32.into_iter().map(|idx| idx as usize).collect();
-
-        // Build the structure
-        let result = Self::from_components(
-            decorator_ids,
-            op_indptr_for_dec_ids,
-            IndexVec::try_from(node_indptr_for_op_idx)
-                .map_err(|e| DeserializationError::InvalidValue(e.to_string()))?,
-        )
-        .map_err(|e| DeserializationError::InvalidValue(e.to_string()))?;
-
-        // Validate CSR structure (including decorator ID bounds)
         result.validate_csr(decorator_count).map_err(|e| {
             DeserializationError::InvalidValue(format!("OpToDecoratorIds validation failed: {e}"))
         })?;
--- a/core/src/mast/debuginfo/node_decorator_storage.rs
+++ b/core/src/mast/debuginfo/node_decorator_storage.rs
@@ -321,19 +321,10 @@ impl NodeToDecoratorIds {
     pub(super) fn write_into<W: crate::utils::ByteWriter>(&self, target: &mut W) {
         use crate::utils::Serializable;
 
-        // Serialize decorator vectors (DecoratorId serializes as u32)
         self.before_enter_decorators.write_into(target);
         self.after_exit_decorators.write_into(target);
-
-        // Serialize indptr arrays as u32. These are indices into u32-bounded arrays,
-        // so any value > u32::MAX would be invalid.
-        let before_indptr_u32: Vec<u32> =
-            self.node_indptr_for_before.iter().map(|&idx| idx as u32).collect();
-        before_indptr_u32.write_into(target);
-
-        let after_indptr_u32: Vec<u32> =
-            self.node_indptr_for_after.iter().map(|&idx| idx as u32).collect();
-        after_indptr_u32.write_into(target);
+        self.node_indptr_for_before.write_into(target);
+        self.node_indptr_for_after.write_into(target);
     }
 
     /// Read this CSR structure from a source, validating decorator IDs against decorator_count.
@@ -343,23 +334,12 @@ impl NodeToDecoratorIds {
     ) -> Result<Self, crate::utils::DeserializationError> {
         use crate::utils::Deserializable;
 
-        // Deserialize decorator vectors (DecoratorId deserializes from u32)
         let before_enter_decorators: Vec<DecoratorId> = Deserializable::read_from(source)?;
         let after_exit_decorators: Vec<DecoratorId> = Deserializable::read_from(source)?;
 
-        // Deserialize indptr arrays from u32. These are indices into u32-bounded arrays,
-        // so any value > u32::MAX would be invalid.
-        let before_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
-        let node_indptr_for_before = IndexVec::try_from(
-            before_indptr_u32.into_iter().map(|idx| idx as usize).collect::<Vec<_>>(),
-        )
-        .map_err(|e| crate::utils::DeserializationError::InvalidValue(e.to_string()))?;
-
-        let after_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
-        let node_indptr_for_after = IndexVec::try_from(
-            after_indptr_u32.into_iter().map(|idx| idx as usize).collect::<Vec<_>>(),
-        )
-        .map_err(|e| crate::utils::DeserializationError::InvalidValue(e.to_string()))?;
+        let node_indptr_for_before: IndexVec<MastNodeId, usize> =
+            Deserializable::read_from(source)?;
+        let node_indptr_for_after: IndexVec<MastNodeId, usize> = Deserializable::read_from(source)?;
 
         let result = Self::from_components(
             before_enter_decorators,
@@ -369,7 +349,6 @@ impl NodeToDecoratorIds {
         )
         .map_err(|e| crate::utils::DeserializationError::InvalidValue(e.to_string()))?;
 
-        // Validate CSR structure (including decorator ID bounds)
         result.validate_csr(decorator_count).map_err(|e| {
             crate::utils::DeserializationError::InvalidValue(format!(
                 "NodeToDecoratorIds validation failed: {e}"


core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -140,17 +143,47 @@ impl OpToDecoratorIds {
...
+            if node_indptr_for_op_idx.iter().all(|&ptr| ptr == 0) {
+                return Ok(Self {
+                    decorator_ids,
+                    op_indptr_for_dec_ids,
+                    node_indptr_for_op_idx,
+                });
+            } else {
+                return Err(DecoratorIndexError::InternalStructure);
+            }
+        }
```
bobbinth, line 169: Under what circumstances could this happen? That is, why would there be decorator storage initialized with nodes, but with no decorators?

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -140,17 +143,47 @@ impl OpToDecoratorIds {
     /// - Last value in `node_indptr_for_op_idx` must be <= `op_indptr_for_dec_ids.len() - 1`
     /// - Both `op_indptr_for_dec_ids` and `node_indptr_for_op_idx` must be strictly monotonic (each
     ///   successive value must be >= the previous one)
-    #[cfg(test)]
     pub fn from_components(
```
bobbinth: We should probably update the doc comments for this method (i.e., "validation restrictions" section).

core/src/mast/debuginfo/decorator_storage.rs
bobbinth: Maybe not for this PR, but I'd move the tests into a separate file as this file is already quite big.

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -57,18 +60,18 @@ use crate::mast::{DecoratedOpLink, DecoratorId, MastNodeId};
 pub struct OpToDecoratorIds {
     /// All the decorator IDs per operation per node, in a CSR relationship with
     /// node_indptr_for_op_idx and op_indptr_for_dec_ids
-    decorator_ids: Vec<DecoratorId>,
+    pub(crate) decorator_ids: Vec<DecoratorId>,
```
bobbinth: This and other fields only need to be `pub(super)` right?

Also, why not just implement `Serializable` and `Deserializable` for `OpToDecoratorIds` rather than exposing these fields?

core/src/mast/debuginfo/mod.rs
```diff
@@ -269,6 +296,219 @@ impl DebugInfo {
...
+        let (decorator_data, decorator_infos, string_table) = decorator_data_builder.finalize();
+
+        decorator_data.write_into(target);
+        string_table.write_into(target);
+
+        // Write decorator count and infos
+        target.write_usize(decorator_infos.len());
+        for decorator_info in decorator_infos {
+            decorator_info.write_into(target);
+        }
```
bobbinth: I think this could be just:
```rust
decorator_infos.write_into(target);
```

core/src/mast/debuginfo/mod.rs
```diff
@@ -269,6 +296,219 @@ impl DebugInfo {
...
+        // node_indptr_for_op_idx from Vec<usize>
+        let node_indptr_vec: Vec<usize> = Deserializable::read_from(source)?;
+        let node_indptr_for_op_idx = IndexVec::from_raw(node_indptr_vec);
+
+        let op_decorator_storage = OpToDecoratorIds::from_components(
+            decorator_ids,
+            op_indptr_for_dec_ids,
+            node_indptr_for_op_idx,
+        )
+        .map_err(|e| DeserializationError::InvalidValue(e.to_string()))?;
```
bobbinth: As mentioned in one of the above comments, we could probably move big part of the code form here into `OpToDecoratorIds` and `NodeToDecoratorIds` by implementing `Serializable` and `Deserializable` traits on these structs.

core/src/mast/serialization/mod.rs
```diff
@@ -38,34 +38,24 @@
 //! - before enter decorators (`Vec<(MastNodeId, Vec<DecoratorId>)>`)
 //! - after exit decorators (`Vec<(MastNodeId, Vec<DecoratorId>)>`)
```
bobbinth: Does this description need to be updated?

core/src/mast/serialization/mod.rs
bobbinth: Not for this PR, but we should probably revisit the overall layout. For example, it is not clear to me if we still need to have this section metadata.
And maybe in general we should now group things differently. For example:

```
- MAGIC
- VERSION
- debug section offset (maybe not needed?)
- procedure roots
- node infos
- basic block data
- advice map data
- debug info
  - error codes
  - decorators
  - basic block decorator list
  - node decorator lists (before/after)
```

IIRC, the main reason for having things like "debug decorator offset" upfront is so that we can skip reading the decorator section from the discs when we want to deserialize a MAST forest w/o decorator data. But maybe that's a premature optimization?

core/src/mast/serialization/mod.rs
```diff
@@ -211,7 +152,7 @@ impl Deserializable for MastForest {
 
         // Reading sections metadata
         let node_count = source.read_usize()?;
-        let decorator_count = source.read_usize()?;
+        let _decorator_count = source.read_usize()?;
```
bobbinth: Related to the previous comment. Seems like we don't use this any more - so, maybe we need to update the format accordingly? (in a different PR).

bobbinth (commented):
Looks good! Thank you! I left some questions/comments inline.

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: This occurs when nodes exist in the MastForest but have no decorators attached to any  operations. We've decided to use a dense representation:
- Serialize all indptr entries sequentially [0, 0, 0, 5, 5, 5, 5, 8, ...], including consecutive equal values that represent empty nodes (dense).
- Versus only serialize non-zero ranges as (node_id, start, end) tuples, skipping empty nodes entirely (sparse).

The sparse encoding overhead (storing node IDs + offsets) exceeds savings from skipping empty node entries — and the dense representation is simpler and more predictable. See the script at  https://gist.github.com/huitseeker/7379e2eecffd7020ae577e986057a400 (linked in the PR)

core/src/mast/debuginfo/decorator_storage.rs
plafer, line 1: This file was renamed from `core/src/mast/debuginfo/decorator_storage.rs` to `core/src/mast/debuginfo/decorator_storage/mod.rs`.

Though we use `mod.rs` almost everywhere in the codebase, shouldn't we prefer the `<...>/decorator_storage.rs` version? This was added to Rust so that in our editors we don't have a bunch of `mod.rs` files opened. I started using the new convention in the fast processor module (e.g. `processor/src/fast/basic_block/basic_block.rs`), but we should settle on 1 convention and use it everywhere.

plafer: Also git treated this as a file removal + a new file, so that we don't see the actual diff that happened. Whether or not we want to move to the new convention, keeping it as `decorator_storage.rs` would make reduce the size of the diff and make this PR easier to review.

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -0,0 +1,864 @@
...
+            if window[0] > window[1] {
+                return Err(format!(
+                    "node_indptr_for_op_idx not monotonic: {} > {}",
+                    window[0], window[1]
+                ));
+            }
+        }
+
+        // Node pointers must be valid indices into op_indptr
+        let max_node_ptr = self.op_indptr_for_dec_ids.len().saturating_sub(1);
```
plafer:
```suggestion
        let max_node_ptr = self.op_indptr_for_dec_ids.len() - 1;
```
We checked that `op_indptr_for_dec_ids` is not empty, so it's safe to use `- 1` here. The `saturating_sub(1)` "threw me off" a little, since I thought for a second that we actually cared about the case where `op_indptr_for_dec_ids` is empty.

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: The directory conversion was to extract tests to a separate file (which [@bobbinth also suggested since the file was getting big](https://github.com/0xMiden/miden-vm/pull/2470#discussion_r2617744523)), but it had the side effect of making git treat it as delete+create.

I can revert to the single decorator_storage.rs file and keep tests inline (at the bottom in `mod tests {}`). That preserves the git history/diff and sidesteps the convention question. The file is big but not unmanageable. Does that work for you?

/cc @bobbinth

plafer: > I can revert to the single decorator_storage.rs file and keep tests inline (at the bottom in mod tests {}).

I'm personally ok with tests being split out, but `decorator_storage.rs` being left as is. Basically the diff would show that all tests were "removed" (moved to new file), and then would show what else changed in that file (if anything, which isn't clear at the moment). I shared more details in Slack, but all I'm suggesting amounts to

1. Move + rename `core/src/mast/debuginfo/decorator_storage/mod.rs` -> `core/src/mast/debuginfo/decorator_storage.rs`
2. commit/squash + push

I ended up doing that locally to get a nicer diff.

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -59,6 +62,117 @@ impl NodeToDecoratorIds {
...
+            }
+
+            if *before_slice.last().unwrap() != self.before_enter_decorators.len() {
+                return Err(format!(
+                    "node_indptr_for_before end {} doesn't match before_enter_decorators length {}",
+                    before_slice.last().unwrap(),
+                    self.before_enter_decorators.len()
+                ));
+            }
+        }
```
plafer, line 146: Future PR: this CSR validation logic is copy/pasted for all instances for CSRs that we have. Ideally we would have a `CsrMatrix` type where we would write that once, and build all or concrete CSR types from it. It would look something like

```rs
pub struct CsrMatrix<Data, Idx> {
    data: Vec<Data>,
    indptr: IndexVec<Idx, usize>
}

pub struct NodeToDecoratorIds {
  before_enter_decorators: CsrMatrix<DecoratorId, MastNodeId>,
  after_exit_decorators: CsrMatrix<DecoratorId, MastNodeId>
}
```
Ideally too we're able to reuse that type to define the "two-level CSR" that `OpToDecoratorIds` needs.

core/src/mast/node/external.rs
```diff
@@ -31,6 +31,11 @@ pub struct ExternalNode {
     decorator_store: DecoratorStore,
 }
 
+// CONSTRUCTORS
+// ================================================================================================
+
+impl ExternalNode {}
+
```
plafer: We can remove this

core/src/mast/serialization/mod.rs
```diff
@@ -211,7 +143,7 @@ impl Deserializable for MastForest {
 
         // Reading sections metadata
         let node_count = source.read_usize()?;
-        let decorator_count = source.read_usize()?;
+        let _decorator_count = source.read_usize()?; // Read for wire format compatibility
```
plafer, line 149: since we're bumping the serialization version number, we could just get rid of this?

plafer (approved):
LGTM! `MastForest` serialization is much cleaner now 🙂

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: Done!

core/src/mast/serialization/mod.rs
huitseeker: I would rather keep it, this will be necessary for #2504. Added a code comment.

core/src/mast/node/external.rs
huitseeker: Done

core/src/mast/debuginfo/node_decorator_storage.rs
huitseeker: #2545

core/src/mast/debuginfo/mod.rs
```diff
@@ -334,6 +363,101 @@ impl DebugInfo {
...
+
+        // 5. Read NodeToDecoratorIds CSR (dense representation)
+        let node_decorator_storage = NodeToDecoratorIds::read_from(source, decorators.len())?;
+
+        // 6. Read procedure names (marked serde(skip) but we deserialize manually)
+        let procedure_names_raw: BTreeMap<Word, String> = Deserializable::read_from(source)?;
+        let procedure_names: BTreeMap<LexicographicWord, Arc<str>> = procedure_names_raw
+            .into_iter()
+            .map(|(k, v)| (LexicographicWord::from(k), Arc::from(v.as_str())))
+            .collect();
```
bobbinth: I would maybe add a comment explaining that the presence of these procedures in the MAST forest is validated at the MAST forest level.

core/src/mast/serialization/mod.rs
```diff
@@ -318,24 +200,10 @@ impl Deserializable for MastForest {
...
-                    digest
-                )));
-            }
-        }
-
-        mast_forest.debug_info.extend_procedure_names(procedure_names);
+        // Note: validation of deserialized MastForests (e.g., checking that procedure name
+        // digests correspond to procedure roots) is intentionally not performed here.
+        // Callers should use MastForest::validate() if validation is needed.
+        // See issue #2445 for the plan to address MastForest validation holistically.
```
bobbinth: Rather than referring the reader to the issue, it may be better to just state the we expect this format to be serialized by a trusted source, and so we do not need to validate it on deserialization.

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -57,18 +60,18 @@ use crate::mast::{DecoratedOpLink, DecoratorId, MastNodeId};
...
     /// ```text
     /// decorator_ids[op_indptr_for_dec_ids[node_start + i]..
     ///               op_indptr_for_dec_ids[node_start + i + 1]]
     /// ```
-    op_indptr_for_dec_ids: Vec<usize>,
+    pub(super) op_indptr_for_dec_ids: Vec<usize>,
     /// The decorated operation indices for the n-th node are at
     /// `op_indptr_for_dec_ids[node_indptr_for_op_idx[n]..node_indptr_for_op_idx[n+1]]`
-    node_indptr_for_op_idx: IndexVec<MastNodeId, usize>,
+    pub(super) node_indptr_for_op_idx: IndexVec<MastNodeId, usize>,
```
bobbinth: Why do these need to be `pub(super)`? And AFAICT, fields in `NodeToDecoratorIds` don't need to be `pub(super)` either.

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -216,10 +226,171 @@ impl OpToDecoratorIds {
...
     }
 
+    /// Serialize this OpToDecoratorIds into the target writer.
+    pub(super) fn write_into<W: crate::utils::ByteWriter>(&self, target: &mut W) {
+        use crate::utils::Serializable;
+
+        // Serialize decorator_ids as u32 vector
+        let decorator_ids_u32: Vec<u32> =
+            self.decorator_ids.iter().map(|id| u32::from(*id)).collect();
+        decorator_ids_u32.write_into(target);
```
bobbinth: Why can't we do just:
```rust
self.decorator_ids.write_into(target);
```
Here and skip the extra allocation?

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -216,10 +226,171 @@ impl OpToDecoratorIds {
...
+            .map(|id| {
+                if id as usize >= decorator_count {
+                    return Err(DeserializationError::InvalidValue(format!(
+                        "DecoratorId {} exceeds decorator count {}",
+                        id, decorator_count
+                    )));
+                }
+                Ok(DecoratorId(id))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
bobbinth: If we implement `Deserializable` for `DecoratorId`, this whole thing could be just:
```rust
let decorator_ids: Vec<DecoratorId> = Deserializable::read_from(source)?;
```

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -216,10 +226,171 @@ impl OpToDecoratorIds {
...
+
+        // Deserialize op_indptr_for_dec_ids
+        let op_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
+        let op_indptr_for_dec_ids: Vec<usize> =
+            op_indptr_u32.iter().map(|&idx| idx as usize).collect();
+
+        // Deserialize node_indptr_for_op_idx
+        let node_indptr_u32: Vec<u32> = Deserializable::read_from(source)?;
+        let node_indptr_for_op_idx: Vec<usize> =
+            node_indptr_u32.iter().map(|&idx| idx as usize).collect();
```
bobbinth: Is there a reason why we want to write the pointers as `u32` rather than `usize`? Is it because `usize` uses variable-length encoding?

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -216,10 +226,171 @@ impl OpToDecoratorIds {
...
+
+        // Serialize op_indptr_for_dec_ids
+        let op_indptr_u32: Vec<u32> =
+            self.op_indptr_for_dec_ids.iter().map(|&idx| idx as u32).collect();
+        op_indptr_u32.write_into(target);
+
+        // Serialize node_indptr_for_op_idx
+        let node_indptr_u32: Vec<u32> =
+            self.node_indptr_for_op_idx.iter().map(|&idx| idx as u32).collect();
+        node_indptr_u32.write_into(target);
```
bobbinth: If we are OK with variable-length encoding, serializing these as `usize` will make the code a bit simpler and would skip extra allocations.

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -199,10 +313,162 @@ impl NodeToDecoratorIds {
...
+
+        // before_enter_decorators as Vec<u32>
+        let before_decorators_u32: Vec<u32> =
+            self.before_enter_decorators.iter().map(|id| u32::from(*id)).collect();
+        before_decorators_u32.write_into(target);
+
+        // after_exit_decorators as Vec<u32>
+        let after_decorators_u32: Vec<u32> =
+            self.after_exit_decorators.iter().map(|id| u32::from(*id)).collect();
+        after_decorators_u32.write_into(target);
```
bobbinth: This could be just:
```rust
self.before_enter_decorators.write_into(target);
self.after_exit_decorators.write_into(target);
```

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -199,10 +313,162 @@ impl NodeToDecoratorIds {
...
+
+        // node_indptr_for_before as Vec<usize>
+        let before_indptr_vec: Vec<usize> = self.node_indptr_for_before.as_slice().to_vec();
+        before_indptr_vec.write_into(target);
+
+        // node_indptr_for_after as Vec<usize>
+        let after_indptr_vec: Vec<usize> = self.node_indptr_for_after.as_slice().to_vec();
+        after_indptr_vec.write_into(target);
```
bobbinth: If we implement `Serializable`/`Deserializable` for `IndexVec`, we could simplify and skip the extra allocations here.

core/src/mast/debuginfo/node_decorator_storage.rs
```diff
@@ -199,10 +313,162 @@ impl NodeToDecoratorIds {
...
+            .map(|id| {
+                if id as usize >= decorator_count {
+                    return Err(crate::utils::DeserializationError::InvalidValue(format!(
+                        "DecoratorId {} exceeds decorator count {}",
+                        id, decorator_count
+                    )));
+                }
+                Ok(DecoratorId(id))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
bobbinth: Similar as one of the above comments. We can replace this with 2 lines of code if `DecoratorId` implements `Deserializable`.

bobbinth (approved):
Looks good! Thank you! I left some comments inline. After these are addressed, we should be good to merge.

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: The fields are `pub(super)` because `write_into` and `read_from` methods in this file need to access them for serialization. We could make them private by implementing `Serializable`/`Deserializable` traits directly on `OpToDecoratorIds` instead of having separate methods, but that would require passing `decorator_count` through a different mechanism since `read_from` needs it for validation. The `decorator_count` is helpful for #1580 deserialization, and, I expect, #2504. 

Happy to refactor if you prefer the trait approach.

bobbinth: I removed `pub(super)` locally and everything compiled and tested fine. Am I missing something?

core/src/mast/debuginfo/mod.rs
huitseeker: Done.

core/src/mast/serialization/mod.rs
huitseeker: This is the comment you were asking for above, btw. And this code is definitely going to get modified before the next release. But addressed in 58c49f6de2d516fb05ba4bb7d5d51d8d43000999 anyway.

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: They're invalid if >= u32::max. I'll hide that.

core/src/mast/debuginfo/decorator_storage.rs
```diff
@@ -327,17 +327,15 @@ impl OpToDecoratorIds {
...
-        // Serialize decorator_ids as u32 vector
-        let decorator_ids_u32: Vec<u32> =
-            self.decorator_ids.iter().map(|id| u32::from(*id)).collect();
-        decorator_ids_u32.write_into(target);
+        // Serialize decorator_ids directly (DecoratorId serializes as u32)
+        self.decorator_ids.write_into(target);
 
-        // Serialize op_indptr_for_dec_ids
+        // Serialize indptr arrays as u32 for cross-platform compatibility
+        // (usize varies between 32-bit and 64-bit platforms)
```
bobbinth: Our `usize` serialization is "cross-platform safe" (as long as the actual values stay within 32-bit range, and if not, we have an issue here anyway).

core/src/mast/debuginfo/decorator_storage.rs
bobbinth: The `usize` serialization/deserialization we have will work fine cross-platform as long as the serialized values don't exceed the valid range for `usize` on the target platform (ref code [here](https://github.com/0xMiden/crypto/blob/next/miden-serde-utils/src/byte_reader.rs#L140)).

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: Ah, yes, that's a dev artifact — I used to have `read_from`/`write_into` in another module, directly at the `MastForest`. No longer relevant. Thanks!

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: Yes. I edited the comment which was misleading, this is not primarily about platforms (though that's a valid, if minor concern).

I thought that we had a limit (of `u32::MAX`) on the indices of decorators attached to a given op, and the  indices of ops attached to a given node. Indeed the # of ops per node, and the total # of decorators are bound by `u32::MAX`. I therefore thought there are two equivalent ways of addressing the correctness issue:
- serializing / deserializing as usize, then checking every read value is bound by `u32::MAX`,
- serializing / deserializing as u32.

The later seemed simpler, allocation aside. I now realize that I am not sure any of those are actually bound by `u32::MAX`. We at least clearly can have values in `op_indptr_for_dec_ids` that can exceed `u32::MAX`, and because they are offsets in that array, same goes for the values of `node_indptr_for_op_idx`. See [other comment](https://github.com/0xMiden/miden-vm/pull/2470#discussion_r2667928937).

Thanks for the pushback. Will edit.

core/src/mast/debuginfo/node_decorator_storage.rs
huitseeker: Done.

core/src/mast/debuginfo/decorator_storage.rs
huitseeker: > if not, we have an issue here anyway

I'm just not sure that's true. Assume for simplicity that you have precisely `decorators.len() == u32::MAX / 100` (and that all quantities divide evenly). For each block, do the following:
- name `op_idx` the index of each operation in the block (we enforce `op_idx <= u32::MAX`),
- if `op_idx` is odd, attach all odd-indexed decorators in the `decorators` array to it,
- symmetrically if `op_idx` is even, attach all even-indexed decorators in the `decorators` array to it.

There's no trivial succinct encoding of that map, so `decorator_ids.len()` is `decorators.len() / 2 * num_ops` where `num_ops` is the total number of operations in the `MastForest`. 

Each operation has to index into a `u32::MAX / 200`-long slice into the `decorator_ids` array (which values are all u32s). The 201st such slice has indexes ending at an offset  `> u32::MAX` (the start and end indices of that slice are kept in `op_indptr_for_dec_ids` by definition). Now assume the `MastForest` has in aggregate over 200 operations — no one has flagrantly broken anything documented about the protocol, yet we have values in `op_indptr_for_dec_ids` above the `u32` range.

It takes a bit more work to see the same issue at work in `node_indptr_for_op_idx` but it's the same idea.

core/src/mast/debuginfo/node_decorator_storage.rs
huitseeker: Done

PR #2443, opened 2025-12-09
debuginfo: Add SourceSpan::SYNTHETIC

djolertrk:
Add a sentinel `SourceSpan` that indicates compiler-generated/synthetic code.

This is used to distinguish between:
  - UNKNOWN: debug info missing or failed to parse from DWARF
  - SYNTHETIC: compiler-generated code that doesn't correspond to any user source

PR #2471, opened 2025-12-15
Track variable locations

djolertrk:
Handle `DebugVar` decorator and new `debug_info` section in the final .masp.

### DIFF
--- a/processor/src/fast/operation.rs
+++ b/processor/src/fast/operation.rs
@@ -125,15 +125,6 @@ impl Processor for FastProcessor {
             }
         }
 
-        if self.in_debug_mode() {
-            for &debug_var_id in current_forest.debug_vars_for_operation(node_id, op_idx_in_block) {
-                if let Some(var_info) = current_forest.debug_var(debug_var_id) {
-                    let process = &mut self.state();
-                    host.on_debug_var(process, var_info);
-                }
-            }
-        }
-
         ControlFlow::Continue(())
     }
 
--- a/processor/src/host/handlers.rs
+++ b/processor/src/host/handlers.rs
@@ -8,7 +8,7 @@ use core::{error::Error, fmt, fmt::Debug};
 
 use miden_core::{
     events::{EventId, EventName, SystemEvent},
-    operations::{DebugOptions, DebugVarInfo},
+    operations::DebugOptions,
 };
 
 use crate::{ExecutionError, ProcessorState, advice::AdviceMutation};
@@ -218,7 +218,4 @@ pub trait DebugHandler: Sync {
         );
         Ok(())
     }
-
-    /// This function is invoked when a debug variable annotation is encountered.
-    fn on_debug_var(&mut self, _process: &ProcessorState, _var_info: &DebugVarInfo) {}
 }
--- a/processor/src/host/mod.rs
+++ b/processor/src/host/mod.rs
@@ -7,7 +7,7 @@ use miden_core::{
     crypto::merkle::InnerNodeInfo,
     events::{EventId, EventName},
     mast::MastForest,
-    operations::{DebugOptions, DebugVarInfo},
+    operations::DebugOptions,
     precompile::PrecompileRequest,
 };
 use miden_debug_types::{Location, SourceFile, SourceSpan};
@@ -116,15 +116,6 @@ pub trait Host {
         let mut handler = debug::DefaultDebugHandler::default();
         handler.on_trace(process, trace_id)
     }
-
-    /// Handles a debug variable annotation emitted from the VM.
-    ///
-    /// Unlike [`on_debug`](Self::on_debug), which is triggered by explicit `debug` instructions,
-    /// `on_debug_var` is a parallel annotation layer: there can be N debug var annotations per
-    /// operation, they carry heap-allocated [`DebugVarInfo`] payloads, and they are infallible
-    /// (the host may not reject them). This separation keeps the hot path for `on_debug` lean
-    /// while still allowing the debugger to track source-level variable locations.
-    fn on_debug_var(&mut self, _process: &ProcessorState, _var_info: &DebugVarInfo) {}
 
     /// Returns the [`EventName`] registered for the provided [`EventId`], if any.
     ///
--- a/processor/src/host/default.rs
+++ b/processor/src/host/default.rs
@@ -4,7 +4,7 @@ use miden_core::{
     Word,
     events::{EventId, EventName},
     mast::MastForest,
-    operations::{DebugOptions, DebugVarInfo},
+    operations::DebugOptions,
 };
 use miden_debug_types::{DefaultSourceManager, Location, SourceFile, SourceManager, SourceSpan};
 
@@ -177,10 +177,6 @@ where
 
     fn on_trace(&mut self, process: &ProcessorState, trace_id: u32) -> Result<(), TraceError> {
         self.debug_handler.on_trace(process, trace_id)
-    }
-
-    fn on_debug_var(&mut self, process: &ProcessorState, var_info: &DebugVarInfo) {
-        self.debug_handler.on_debug_var(process, var_info)
     }
 
     fn resolve_event(&self, event_id: EventId) -> Option<&EventName> {


djolertrk:
Goes on top of https://github.com/0xMiden/miden-vm/pull/2443.

djolertrk:
cc @greenhat @bitwalker

djolertrk:
In order to try everything within compiler, I have used the `commit 81cc603c1b9d5e6a55d82daee5fc1016cee33c02 (tag: v0.19.1)` as a base branch... Since the debug variable tracking (https://github.com/0xMiden/compiler/pull/820) depends on this, we may want to implement majority of these things in the compiler.

cc @greenhat @bitwalker

core/src/mast/serialization/decorator.rs
```diff
@@ -123,6 +123,58 @@ impl DecoratorInfo {
...
+                if data_reader.read_bool()? {
+                    debug_var.set_type_id(data_reader.read_u32()?);
+                }
+
+                // Read optional arg_index
+                if data_reader.read_bool()? {
+                    debug_var.set_arg_index(data_reader.read_u32()?);
+                }
+
+                // Read optional source location
```
bitwalker: Would this be redundant with the `Location` of an `AssemblyOp` decorator linked to the same instruction? I would assume so, but if not, I'd be interested to know what the location would be in this case.

If it is the same location the vast majority of the time, then we should only set this one when it differs from the location associated with the instruction itself, otherwise we'll end up bloating the size of the resulting package considerably IMO.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Debug information for tracking a source-level variable.
+///
+/// This decorator provides debuggers with information about where a variable's
+/// value can be found at a particular point in the program execution.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+pub struct DebugVarInfo {
+    /// Variable name as it appears in source code
+    name: String,
```
bitwalker: This should probably be made `Arc<str>` so that we can share references to the same variable name across all uses.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Type information (encoded as type index in debug_info section)
+    type_id: Option<u32>,
+    /// If this is a function parameter, its 1-based index
+    arg_index: Option<u32>,
```
bitwalker: Make this `Option<NonZeroU32` since it is expected to be a 1-based index.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Source file location (file:line:column)
+    location: Option<FileLineCol>,
```
bitwalker, line 34: I don't think this is needed, as mentioned in my other comment - but let me know if this is only intended to be set in cases where the location differs from that of the associated instruction.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,191 @@
...
+/// This enum models the different ways a variable's value might be stored
+/// during program execution, ranging from simple stack positions to complex
+/// expressions.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+pub enum DebugVarLocation {
+    /// Variable is at stack position N (0 = top of stack)
+    Stack(u8),
+    /// Variable is in memory at the given word address
```
bitwalker: *Nit:* the given _element_ address

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,191 @@
...
+    Memory(u32),
+    /// Variable is a constant field element
+    Const(u64),
```
bitwalker: If this is meant to be a `Felt`, lets use that type instead of `u64`

core/src/operations/decorators/mod.rs
```diff
@@ -79,6 +86,7 @@ impl fmt::Display for Decorator {
             },
             Self::Debug(options) => write!(f, "debug({options})"),
             Self::Trace(trace_id) => write!(f, "trace({trace_id})"),
+            Self::DebugVar(debug_var) => write!(f, "debugVar({})", debug_var),
```
bitwalker:
```suggestion
            Self::DebugVar(debug_var) => write!(f, "variable({})", debug_var),
```

The camel-casing here (and with `asmOp`) feel out of place - but setting that aside, I think it'll be more readable display output if we just call it `variable`, since we aren't going to have anything else by that name that it could be conflated with.

crates/assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -273,6 +272,7 @@ pub enum Instruction {
     // ----- debug decorators --------------------------------------------------------------------
     Breakpoint,
     Debug(DebugOptions),
+    DebugVar(miden_core::DebugVarInfo),
```
bitwalker: I don't think this should be added as an instruction - only a decorator, just like `AssemblyOp`.

That said, currently, the AST/assembler doesn't provide a way to attach decorators that don't correspond to an instruction to the AST before assembly.

So - I think the best way to handle this is to support `Instruction` variants like `DebugVar`, but do not pretty-print them like instructions, and do not add support in the grammar for parsing them. The compiler will then have an easy way to emit these decorators, but they don't end up getting treated like normal instructions.

I would suggest we add a method on `Instruction` called `has_textual_representation` which matches on the instruction variant and returns true if it matches any variant other than those like `DebugVar` which are not present in the textual format of MASM. I'm happy to bikeshed the name too, but `has_textual_representation` feels like the most precise way to name it without it unintentionally becoming used for something with different semantics.

crates/assembly-syntax/src/ast/instruction/print.rs
```diff
@@ -281,49 +280,62 @@ impl PrettyPrint for Instruction {
...
+                flatten(const_text("procref") + const_text(".") + text(format!("{module}::{name}")))
+            },
+            Self::ProcRef(InvocationTarget::AbsoluteProcedurePath { name, path }) => {
+                flatten(const_text("procref") + const_text(".") + text(format!("::{path}::{name}")))
             },
 
             // ----- debug decorators -------------------------------------------------------------
             Self::Breakpoint => const_text("breakpoint"),
             Self::Debug(options) => inst_with_imm("debug", options),
+            Self::DebugVar(info) => text(format!("debug.var({})", info)),
```
bitwalker: In this method, I would add the following at the top prior to the `match`:

```rust
if !self.has_textual_representation() {
    return Document::Empty;
}
```

And then for any variant like this (which cannot possibly be matched unless `has_textual_representation` returns true), use `=> unreachable!("expected {self:#?} to have textual representation"),

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,483 @@
...
+// ================================================================================================
+
+/// Source file information.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugFileInfo {
+    /// Path to the source file (index into string table)
+    pub path_idx: u32,
+    /// Optional directory containing the file (index into string table)
+    pub directory_idx: Option<u32>,
```
bitwalker: Isn't this redundant with `path_idx`, i.e. if we have the path of the source file, then that presumably includes any directory information already, no?

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,483 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugVariableInfo {
+    /// Name of the variable (index into string table)
+    pub name_idx: u32,
+    /// Type of the variable (index into type table)
+    pub type_idx: u32,
+    /// If this is a parameter, its 1-based index (0 = not a parameter)
+    pub arg_index: u32,
+    /// Line where the variable is declared
```
bitwalker: I would use `LineNumber`/`ColumnNumber` types from `miden_core::debuginfo` for these, as it is clearer whether the values are 0 or 1-indexed.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,483 @@
...
+    pub line: u32,
+    /// Column where the variable is declared
+    pub column: u32,
+    /// Scope depth (0 = function scope)
+    pub scope_depth: u32,
```
bitwalker, line 563: I'd be interested to understand how one can leverage a non-zero `scope_depth` in practice, i.e. how is it used?

processor/src/fast/mod.rs
```diff
@@ -741,98 +522,55 @@ impl FastProcessor {
...
-                    if let Err(err) = host.on_trace(process, *id) {
-                        return ControlFlow::Break(BreakReason::Err(
-                            ExecutionError::TraceHandlerError { clk, trace_id: *id, err },
-                        ));
-                    }
-                }
+                let process = &mut self.state();
+                host.on_trace(process, *id)?;
+            },
+            Decorator::DebugVar(_debug_var) => {
```
bitwalker: If we keep this behavior, I think I would reword this as:

> DebugVar is strictly a metadata-carrying decorator used by downstream tooling, i.e. it is non-actionable in the processor itself.

That said, IMO we need some way for the debugger to be notified when variable information is found/changes, so that it can update its internal state accordingly, otherwise we will be forced to step every instruction manually to check for this decorator in case that information changes. I'd prefer not to have to do that.

IMO, we should manifest this either as a raw `trace` event, modify `on_debug` to support triggering that callback for this decorator, or add a new callback that is specific to `DebugVar`. Whichever approach we choose, we'd only invoke the callback when tracing is enabled in the processor.

/cc @bobbinth @plafer

processor/src/lib.rs
```diff
@@ -160,87 +181,619 @@ pub struct ChipletsTrace {
...
+                    self.decoder.append_asmop(self.system.clk(), assembly_op.clone());
+                }
+            },
+            Decorator::Trace(id) => {
+                if self.enable_tracing {
+                    let process = &mut self.state();
+                    host.on_trace(process, *id)?;
+                }
+            },
+            Decorator::DebugVar(_debug_var) => {
```
bitwalker: Same as my comment for the `FastProcessor` - but since the slow processor is going away, it isn't as important what we do here.

._target
bitwalker, line 1: Looks like this got accidentally committed

processor/src/lib.rs
djolertrk: This will go away, since I will rebase on top of the latest.

core/src/mast/serialization/decorator.rs
djolertrk: Yes, that is the case -- we will only set this if loc differs from the one attached to the instruction. I will add comments here and there to make it more obvious -- but compiler will also generate it that way.

core/src/operations/decorators/debug_var.rs
djolertrk: Will add comment here to make it clear.

crates/assembly-syntax/src/ast/instruction/mod.rs
djolertrk: I agree!

crates/mast-package/src/debug_info/types.rs
djolertrk: It was intending to kinda match with what other debug info representations do (like DWARF or CodeView), where it tries to keep the base/common root dir shared among other relative paths depending on that root, but since in `miden(c)` we always operate with full paths, we can simplify this.

crates/mast-package/src/debug_info/types.rs
djolertrk: If there is a variable with the same name in a nested/inner block debugger needs to differentiate between that one and the one from outer scope. I will add an example in form of comment here in the code.

djolertrk:
> I think this looks pretty good! See my comments for more specific feedback on things I think we may need to change.
> 
> One thing that I think may be worth adding to `miden_mast_package` as part of this, is a higher-level type or set of types which reify the encoded representation of the debug information into usable form that has been validated and can be directly queried by tools such as the debugger.
> 
> As it stands, what you get from deserializing the section is still a very raw form (i.e. references to other data in the section is not materialized/validated). That is useful as a low-level primitive, but means that any tooling that wants to actually use the information encoded in the package has to build that high-level representation themselves. While in some cases that may be necessary/desirable, I suspect that in practice, a single canonical representation defined in `miden_mast_package` would be sufficient for the uses we have in mind.
> 
> My only other note is that I'd like to see a bit more documentation on what the various bits of information are used for, how to put them together. This would of course be one thing solved by providing a higher-level representation of the debug info, but even then there are some bits of information that are somewhat unclear on what their usage should be (e.g. my comment on `scope_depth`).
> 
> Ultimately, one of the things I'd want the high-level representation to provide, would be a set of API functions that correspond to important debugger queries, without the debugger needing to know too much about the internal encoding of the debug info provided by a `Package`.

Thanks a lot for your comments! I have rebased this on top of the latest next branch, and addressed all comments. My comment regarding API for the debugger -- I have left a comment with a TODO marker, so we can pick it up once compiler side is ready and when we can have hands-on on the consumer/debugger side. Please do note that I have added a couple of documents in the https://github.com/0xMiden/compiler/pull/820 that describe the debug info design at both IRs level and in the final package.

crates/mast-package/src/debug_info/types.rs
bitwalker: If we keep it, I'd make a note that when `directory_idx` is set, then `path_idx` is a relative path, otherwise `path_idx` is expected to be absolute. 

We will probably want to explore splitting paths into those components in the future, so we can further reduce the size of the serialized debug info, but I'd treat that as lower priority for now.

crates/mast-package/src/debug_info/types.rs
bitwalker: I see - in the debugger we only need to care about definitions at the current depth (and above, i.e. lower depth) for purposes of being able to show variable information at the current program point. We don't need to worry about conflicts at the same depth, since only one variable definition at that depth can exist, even if there are multiple scopes with that depth in the current function.

Definitely a good idea to document those semantics here I think, but at least the usefulness is clear to me.

djolertrk: Thanks a lot for the clarification!

crates/mast-package/src/debug_info/types.rs
djolertrk: I have added a TODO marker for this in https://github.com/0xMiden/miden-vm/pull/2471/commits/7bbd98813ca1845669af2937e54e154c1668b48f, so we can introduce these fields later.

processor/src/fast/mod.rs
plafer: I can't speak to whether this decorator is the right approach for the debugger. Though with #2541 we want to move away from representing `AsmOp` (and now `DebugVar` too) as decorators that can be visited needlessly by the processor during execution.

Though for now this is the best we can do, provided that the `DebugVar` decorator is the right approach for the debugger.

Note: is this a feature that would only be accessible for programs written in Rust, or also from MASM? Or in other words, would there be fewer features made available by the debugger for debugging programs written in MASM compared to programs written in Rust?

djolertrk: >Note: is this a feature that would only be accessible for programs written in Rust, or also from MASM? Or in other words, would there be fewer features made available by the debugger for debugging programs written in MASM compared to programs written in Rust?

If there is a way of using a concept of "variable" at MASM level we can create a decorator for it -- since this is not Rust specific, but rather an abstraction. However, it needs to have a name, type..., so debugger can decode its value properly.

bitwalker: @plafer Since MASM doesn't have variables, this decorator wouldn't ever be produced from Miden Assembly sources. The closest we have are procedure locals, but those aren't variables in any practical sense; rather they collectively represent a fixed-size region of automatically-allocated, procedure-scoped memory, so I don't think it would make sense to use `DebugVar` for them - at least not without some MASM syntax for local variables that gives them a name and type, and is then used to derive the size and placement of procedure locals needed for that procedure.

But, as @djolertrk noted, this can be used for any source language with variables, it isn't Rust-specific.

On the topic of debug decorators like this and `AssemblyOp`, I'd be happy to replace them with something that doesn't rely on the decorator mechanism at all, but I'm not actually sure what that would be. All we really need is a way to map a MAST node to the debug information that applies to it, ideally with a way to disambiguate the debug information based on context (e.g. current call stack). AFAIK, that's essentially how decorators work now though (sans the disambiguation bit), so it isn't clear to me what we gain or lose by adding decorators like this. Definitely open to figuring out how best to handle this sort of thing going forward though.

plafer: With #2541 we basically already have a plan to move out any current "decorator" which only "provides metadata" (TLDR they get their own separate CSR matrix). But since that doesn't exist yet, we can just add the `DebugVar` as a decorator for now (as-is in this PR), and we'll pull it out along with `AsmOp` when we address #2541.

djolertrk:
As you can see in the https://github.com/0xMiden/compiler/issues/745, I have back-ported these commits from miden-vm based on "next" to a branch based on "miden-vm v19.0.1" (https://github.com/walnuthq/miden-vm/tree/feature/debug-variable-locations-v0.19.1) so I can test it all together with compiler and debugger, so the latest commits here represent debugger API that is tested with https://github.com/0xMiden/miden-debug/pull/17.

processor/src/fast/mod.rs
plafer: Note: #2541 is being addressed with #2606, so we might have to revisit the strategy in this PR.

djolertrk: Hmm, okay. It may be doable to introduce a `debug_var_storage` and to remove the debug var decorator -- will try this.

djolertrk: I have addressed this and rebased on top of "next".

djolertrk: Rebased again.

djolertrk:
@plafer @bitwalker I am wondering if there is any comments on the latest? Thank you!

bitwalker:
@djolertrk Sorry for the delay! It's been a busy week and haven't been able to get to this review yet, I'll take a look this morning

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+///
+/// String indices in sub-types (e.g., `name_idx` in `DebugFieldInfo`) are relative
+/// to this section's own string table.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugTypesSection {
+    /// Version of the debug types format
+    pub version: u8,
+    /// String table containing type names, field names
+    pub strings: Vec<String>,
```
bitwalker: I'd use `Vec<Arc<str>>` here, because MASM identifiers are always `Arc<str>` internally, so we can reduce the number of allocated strings in memory when we can share them. In all other cases, converting a `String` to an `Arc<str>` requires no reallocation, so it's a win-win.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+    pub fn new() -> Self {
+        Self {
+            version: DEBUG_TYPES_VERSION,
+            strings: Vec::new(),
+            types: Vec::new(),
+        }
+    }
+
+    /// Adds a string to the string table and returns its index.
+    pub fn add_string(&mut self, s: impl Into<String>) -> u32 {
```
bitwalker: I would just expect an `Arc<str>` here, rather than making it generic, since it reduces the likelihood of us unintentionally taking an `Arc<str>`, converting it to a `String` (reallocating in the process), only to convert back to an `Arc<str>`.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+        if let Some(idx) = self.strings.iter().position(|existing| existing == &s) {
+            return idx as u32;
+        }
+        let idx = self.strings.len() as u32;
+        self.strings.push(s);
+        idx
+    }
+
+    /// Gets a string by index.
+    pub fn get_string(&self, idx: u32) -> Option<&str> {
```
bitwalker: Similarly, we may want to return `Arc<str>` here, rather than a `&str`, so that if we want to reuse the allocation we can

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+        self.strings.get(idx as usize).map(String::as_str)
+    }
+
+    /// Adds a type to the type table and returns its index.
+    pub fn add_type(&mut self, ty: DebugTypeInfo) -> u32 {
```
bitwalker: We conventionally strongly type our indices like this, to ensure they don't get misused - I'd recommend we do so here as well.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+        let idx = self.types.len() as u32;
+        self.types.push(ty);
+        idx
+    }
+
+    /// Gets a type by index.
+    pub fn get_type(&self, idx: u32) -> Option<&DebugTypeInfo> {
```
bitwalker: This sort of API is why it's useful to have strongly-typed indices, as we have some guarantee that the index is valid

crates/mast-package/src/debug_info/types.rs
bitwalker: Speaking of which, it might actually be beneficial to internally de-duplicate strings when constructing this type (and other similar types that have an internal vector of strings) - but whether that lives here, or is handled by higher-level tools I don't have a strong opinion on. For example, when we deserialize strings in decorators, the deserializer automatically de-duplicates strings which are deserialized as `Arc<str>`, and since strings stored in decorators (and debug info) necessarily have a _lot_ of duplication, this can be a pretty big win.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+
+/// Primitive type variants supported by the debug info format.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[repr(u8)]
+pub enum DebugPrimitiveType {
+    /// Void type (0 bytes)
+    Void = 0,
+    /// Boolean (1 byte)
+    Bool = 1,
```
bitwalker: *Nit:* It isn't necessary to explicitly provide discriminant values for every variant, only the first one - at least without documenting what to do if this enum needs to be changed (e.g. variants added/removed).

For example, if you remove a variant, that is a backwards incompatible change, so you may as well update all the discriminants anyway. If you add a variant, you may as well require that they be added at the end (which also has the added benefit of being able to easily see when the different variants were added relative to the "base" enum).

I don't have a problem with being explicit, but without further documentation on how evolution of the enum is supposed to proceed, it's probably not going to live up to the implied purpose of backwards-compatibility.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugFileInfo {
+    /// Full path to the source file (index into string table).
+    pub path_idx: u32,
+    /// Optional checksum of the file content for verification.
+    ///
+    /// When present, debuggers can use this to verify that the source file on disk
+    /// matches the version used during compilation.
+    pub checksum: Option<[u8; 32]>,
```
bitwalker: I'd suggest boxing the checksum value, since otherwise it makes the type quite a bit larger even when checksums aren't used (and at least at the moment, I don't think they will be).

crates/mast-package/src/debug_info/types.rs
```diff
@@ -0,0 +1,658 @@
...
+    pub file_idx: u32,
+    /// Line number where the function starts (1-indexed)
+    pub line: LineNumber,
+    /// Column number where the function starts (1-indexed)
+    pub column: ColumnNumber,
+    /// Type of this function (index into type table, optional)
+    pub type_idx: Option<u32>,
+    /// MAST root digest of this function (if known).
+    /// This links the debug info to the compiled code.
+    pub mast_root: Option<[u8; 32]>,
```
bitwalker: I'd use `Word` for this, since that's how we represent MAST roots, and the type is available in this crate (and used for digests already).

processor/src/host/handlers.rs
```diff
@@ -218,4 +224,16 @@ pub trait DebugHandler: Sync {
         );
         Ok(())
     }
+
+    /// This function is invoked when the `DebugVar` decorator is executed.
+    ///
+    /// This provides debuggers with information about source-level variables and
+    /// where their values can be found during execution.
+    fn on_debug_var(
```
bitwalker: I wonder if this should really be fallible - these decorators are supposed to be purely informational, and if there is an error that affects, say the debugger, then we'd probably just ignore the error or panic anyway, rather than returning an error only for the processor to then raise it.

IMO, we should just indicate the implementations should choose one of those two strategies, rather than return an error that the processor must propagate.

Honestly, the same is probably true of `on_trace`, but that's out of scope for this PR.

Thoughts @plafer?

processor/src/host/mod.rs
```diff
@@ -117,6 +117,21 @@ pub trait Host {
...
     }
 
+    /// Handles a debug variable decorator from the VM.
+    ///
+    /// This callback is invoked when the processor encounters a `DebugVar` decorator,
+    /// which provides information about source-level variables and their locations
+    /// during execution. This is useful for debuggers to track variable values.
+    ///
+    /// The default implementation does nothing, as most hosts don't need variable tracking.
+    fn on_debug_var(
```
bitwalker: I'd mark this `#[inline(always)]` so that the whole call gets elided when possible

bitwalker (changes requested):
Looks good to me!

I did have a few notes and a couple small changes that I suggested in comments above, but let me know if you have any questions/objections to any of them, and we can discuss further. None of them are actually blockers to merging this, aside from probably the strongly-typed indices one (since that's a pattern we try to adhere to generally).

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+        // Write value location
+        self.value_location.write_into(target);
+
+        // Write optional type_id
+        if let Some(type_id) = self.type_id {
+            target.write_bool(true);
+            target.write_u32(type_id);
+        } else {
+            target.write_bool(false);
+        }
```
bobbinth: We should be able to simplify this to something like:
```rust
self.type_id.write_into(target);
```

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+
+        // Write optional arg_index
+        if let Some(arg_index) = self.arg_index {
+            target.write_bool(true);
+            target.write_u32(arg_index.get());
+        } else {
+            target.write_bool(false);
+        }
```
bobbinth: Same comment as above.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+}
+
+impl Serializable for DebugVarInfo {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        // Write name as length-prefixed string
+        let name_bytes = self.name.as_bytes();
+        target.write_u16(name_bytes.len() as u16);
+        target.write_bytes(name_bytes);
```
bobbinth: Similar comment as above, I think this could be just:
```rust
self.name.write_into(target);
```

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+            // Write URI as string
+            let uri_str = loc.uri.as_str();
+            target.write_u16(uri_str.len() as u16);
+            target.write_bytes(uri_str.as_bytes());
+            // Write line and column as u32
+            target.write_u32(loc.line.to_u32());
+            target.write_u32(loc.column.to_u32());
+        } else {
+            target.write_bool(false);
+        }
```
bobbinth: Same as above, this could be something like:
```rust
self.location.write_into(target);
```

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+}
+
+impl Deserializable for DebugVarInfo {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
```
bobbinth, line 245: Following the above comments, I think this could can be simplified quite a bit too.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// Type information (encoded as type index in debug_info section)
+    type_id: Option<u32>,
+    /// If this is a function parameter, its 1-based index.
+    arg_index: Option<NonZeroU32>,
+    /// Source file location (file:line:column).
+    /// This should only be set when the location differs from the AssemblyOp decorator
+    /// location associated with the same instruction, to avoid package bloat.
+    location: Option<FileLineCol>,
+    /// Where to find the variable's value at this point
+    value_location: DebugVarLocation,
```
bobbinth, line 36: Question: how do we deal with situations where a variable could be in multiple places? For example, we could have a copy of a variable in memory but also on the stack.

core/src/mast/debuginfo/mod.rs
```diff
@@ -95,6 +98,12 @@ pub struct DebugInfo {
     /// Efficient access to AssemblyOps per operation per node.
     asm_op_storage: OpToAsmOpId,
 
+    /// All debug variable information in the MAST forest.
+    debug_vars: IndexVec<DebugVarId, DebugVarInfo>,
+
+    /// Efficient access to debug variables per operation per node.
+    op_debug_var_storage: OpToDebugVarIds,
```
bobbinth, line 105: Not directly related to this PR, but we now have this pattern repeat 3 times: for decorators, asm ops, and now debug vars. Probably means that we should consider creating a struct that can abstract this away.

processor/src/host/handlers.rs
bobbinth: Question: what's the intended use of this functionality? I'm guessing is that we use it in the debugger to "watch" values of variables defined in higher level languages - right? And then the idea is that every time a new value is assigned to a "watched" variable, the compiler would insert the `DebugVar` decorator and then the host would be able to catch it.

If so, do we actually need a separate callback method for this? An alternative would be to provide this data via the `on_debug()` callback by making `DebugVar` another variant of the `DebugOptions` enum.

crates/assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -288,6 +289,17 @@ impl Instruction {
...
+    /// Returns true if the instruction has a textual representation in MASM syntax.
+    ///
+    /// Some instructions (like `DebugVar`) are only emitted by the compiler as decorators
+    /// and are not present in the textual format of MASM. These instructions cannot be
+    /// parsed from text, and should not be pretty-printed like normal instructions.
+    pub const fn has_textual_representation(&self) -> bool {
+        // DebugVar is a decorator-only instruction emitted by the compiler,
+        // not something that can be written in MASM source code.
+        !matches!(self, Self::DebugVar(_))
+    }
```
bobbinth: Question: why do we need to make `DebugVar` an instruction variant? Could we not treat it in the same way as we treat `AssemblyOp` decorator?

bobbinth (commented):
Not a review yet, but I left some comments inline. Also, where is the best place to read about the motivation behind this. I can kind of guess what it will be used for - but would be good to understand the use case and the overall design approach a bit more.

crates/mast-package/src/debug_info/types.rs
djolertrk: I agree. Addressed.

processor/src/host/handlers.rs
djolertrk: OK. I will try to reflect that in the comment above the function.

processor/src/host/mod.rs
djolertrk: Yeah, I agree.

core/src/operations/decorators/debug_var.rs
djolertrk: Yep, it makes sense.

core/src/operations/decorators/debug_var.rs
djolertrk: Well, if a value lives in multiple places at the same time/program point, debuggers are happy to know at least one (only), since during debugging session the users will just want to print value of a variable `x`, and usually don't care about the detail where it actually lives.

crates/assembly-syntax/src/ast/instruction/mod.rs
djolertrk: Ah, good question. This is basically a leftover from the initial proposal for this. `DebugVar` shouldn't be an `Instruction` variant. It's compiler-generated metadata, not a MASM instruction. We'll remove `Instruction::DebugVar` from `miden-vm` and handle it on the compiler side.

processor/src/host/handlers.rs
djolertrk: >Question: what's the intended use of this functionality? I'm guessing is that we use it in the debugger to "watch" values of variables defined in higher level languages - right? And then the idea is that every time a new value is assigned to a "watched" variable, the compiler would insert the DebugVar decorator and then the host would be able to catch it.

Yes, that is accurate.

>If so, do we actually need a separate callback method for this? An alternative would be to provide this data via the on_debug() callback by making DebugVar another variant of the DebugOptions enum.

Well, `DebugVar` carries more information such as variable name, type, location, and that is why it was implemented like that. For example, we can have N DebugVars attached to the same operation/instruction.
I will try to improve the comment above the function even more.

djolertrk:
@bitwalker @bobbinth -- thanks a lot for your reviews!

> Not a review yet, but I left some comments inline. Also, where is the best place to read about the motivation behind this. I can kind of guess what it will be used for - but would be good to understand the use case and the overall design approach a bit more.

Basically, we want to extend our debugger to be able to find and print variable's values at certain program points. So, in order to achieve that, we need to have information about variables location(s) along with its type, so the debugger can pretty print it accurately. A variable can be moved from one location to another, so we need a way to reflect that also. In DWARF, there is `.debug_loc` section that uses multiple `DW_AT_location` attribute to represent how value was move around through its lifetime.

Here is the GitHub issue dedicated to this feature: https://github.com/0xMiden/compiler/issues/745

processor/src/host/handlers.rs
plafer: Why do we need `Host::on_debug_var()` in the first place? 
This method is not called in this PR, so it's hard for me to double-check my intuition here. In my understanding, we don't need such a method, for the same reason that we don't have a `Host::on_asmop_info()`. Or in other words, I don't see where the processor would ever want to call that method.

djolertrk: Well, debugger can "live" without it actually. So, good point — I was thinking the debugger would need a callback, but it can just query the forest directly, same as AsmOp info. I removed it.

bitwalker: @djolertrk If `on_debug_var` isn't being called from the processor, we should make it do so, otherwise it isn't worth adding at all.

core/src/operations/decorators/debug_var.rs
bitwalker: The value here is always tied to a scope, so the storage for a variable depends on the scope we're in, so even if a variable is in more than one storage location, only one of those is relevant at any given program point.

processor/src/host/handlers.rs
bitwalker: I think we'd be fine either way, but we need to be notified when this information changes during execution, hence the callback. In a world where we can just look up information like this at every instruction ourselves, the callback wouldn't be necessary.

crates/assembly-syntax/src/ast/instruction/mod.rs
bitwalker: No, because the facility used to emit `AssemblyOp` is internal to the assembler, and there is no way for us to easily surface it - representing it as an instruction keeps this simple, and correlates pretty well to how we represent debug intrinsics in the compiler IR.

processor/src/host/handlers.rs
djolertrk: Ah, well, right, I see the point now. I forgotten the fact that debugger still needs to be ported to the fast processor, it does make sense now. I will add that this callback is being invoked from the processor in debug mode only -- will be part of the following change. Thanks @bitwalker!

bitwalker:
@djolertrk I don't have access to your fork, so I can't rebase this PR for you. I switched the base branch to `main`, but the changelog conflicts still need to be resolved.

djolertrk:
> @djolertrk I don't have access to your fork, so I can't rebase this PR for you. I switched the base branch to `main`, but the changelog conflicts still need to be resolved.

Done. Thanks!

bitwalker:
@bobbinth This is ready to merge whenever you get a chance to give it a look

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+//! This information is used by debuggers to map between the Miden VM execution state
+//! and the original source code.
+
+mod serialization;
+mod types;
+
+pub use types::*;
+
+// Re-export serialization traits for use by consumers
+// (the impl blocks are in serialization.rs)
```
bobbinth: I'm not sure I understood this comment. AFAICT, we are not re-exporting serialization traits from here. Is this comment obsolete?

core/src/mast/mod.rs
```diff
@@ -577,6 +577,28 @@ impl MastForest {
...
+    pub fn debug_vars_for_operation(
+        &self,
+        node_id: MastNodeId,
+        local_op_idx: usize,
+    ) -> &[DebugVarId] {
+        self.debug_info.debug_vars_for_operation(node_id, local_op_idx)
+    }
+
+    /// Returns the debug variable with the given ID, if it exists.
+    pub fn debug_var(&self, debug_var_id: DebugVarId) -> Option<&crate::operations::DebugVarInfo> {
```
bobbinth: nit: here and in other places in this file, instead of using absolute path `crate::operations::DebugVarInfo` we usually import items at the top of the file and then here it would be just `DebugVarInfo`.

core/src/mast/debuginfo/debug_var_storage.rs
bobbinth, line 1: I did not review this file in detail, but it feels like there is some code duplication between this and other storages (e.g., `AsmOp` storage). I wonder if there is a good way to avoid this. Let's create an issue for this.

Also, cc @huitseeker for visibility.

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,354 @@
...
+                target.write_u64(felt.as_canonical_u64());
+            },
+            Self::Local(offset) => {
+                target.write_u8(3);
+                target.write_bytes(&offset.to_le_bytes());
+            },
+            Self::Expression(bytes) => {
+                target.write_u8(4);
+                target.write_u16(bytes.len() as u16);
+                target.write_bytes(bytes);
```
bobbinth: Could we not do this as:
```rust
bytes.write_into(target);
```

core/src/operations/decorators/debug_var.rs
```diff
@@ -0,0 +1,354 @@
...
+            3 => {
+                let bytes = source.read_array::<2>()?;
+                Ok(Self::Local(i16::from_le_bytes(bytes)))
+            },
+            4 => {
+                let len = source.read_u16()? as usize;
+                let mut bytes = Vec::with_capacity(len);
+                for _ in 0..len {
+                    bytes.push(source.read_u8()?);
+                }
```
bobbinth: Related to the above comments - we could deserialize this as a vector of bytes directly rather than using a manual loop.

In general, we should try to avoid using the pattern of reading the number of elements and then allocating a vector with this number of elements - as this could lead to a security vulnerability. Instead, we should be using `ByteReader::read_many_iter()` method.

crates/assembly/src/basic_block_builder.rs
```diff
@@ -55,6 +55,9 @@ struct PendingAsmOp {
 pub struct BasicBlockBuilder<'a> {
     ops: Vec<Operation>,
     decorators: DecoratorList,
+    /// Debug variables attached to operations in this block.
+    /// Each entry is (op_index, debug_var_id) similar to decorators.
+    debug_vars: Vec<(usize, DebugVarId)>,
```
bobbinth: nit: could we move this under `asm_ops` (i.e., under line 65 below).

processor/src/host/mod.rs
```diff
@@ -117,6 +117,15 @@ pub trait Host {
...
     }
 
+    /// Handles a debug variable annotation emitted from the VM.
+    ///
+    /// Unlike [`on_debug`](Self::on_debug), which is triggered by explicit `debug` instructions,
+    /// `on_debug_var` is a parallel annotation layer: there can be N debug var annotations per
+    /// operation, they carry heap-allocated [`DebugVarInfo`] payloads, and they are infallible
+    /// (the host may not reject them). This separation keeps the hot path for `on_debug` lean
+    /// while still allowing the debugger to track source-level variable locations.
+    fn on_debug_var(&mut self, _process: &ProcessorState, _var_info: &DebugVarInfo) {}
```
bobbinth: Thinking about this more, it is not clear to me if we need this callback. Could we not do what @bitwalker mentioned in https://github.com/0xMiden/miden-vm/pull/2471#discussion_r2790826734?

I'm assuming the debugger steps through the instructions one-by-one, and if so, it should be able to look up `DebugVarInfo`s for a given operation using `MastForest::debug_vars_for_operation()` method. But maybe that's not how it works?

processor/src/fast/operation.rs
```diff
@@ -125,6 +125,15 @@ impl Processor for FastProcessor {
...
         }
 
+        if self.in_debug_mode() {
+            for &debug_var_id in current_forest.debug_vars_for_operation(node_id, op_idx_in_block) {
+                if let Some(var_info) = current_forest.debug_var(debug_var_id) {
+                    let process = &mut self.state();
+                    host.on_debug_var(process, var_info);
+                }
+            }
+        }
```
bobbinth: As mentioned in the previous comment, I'm hoping we can get rid of the `on_debug_var()` callback, but if not, I think we can move this loop inside the if-statement on line 119 as that one already checks if we are in debug mode.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -0,0 +1,663 @@
...
+            return Err(DeserializationError::InvalidValue(alloc::format!(
+                "unsupported debug_types version: {version}, expected {DEBUG_TYPES_VERSION}"
+            )));
+        }
+
+        let strings_len = source.read_usize()?;
+        let mut strings = alloc::vec::Vec::with_capacity(strings_len);
+        for _ in 0..strings_len {
+            strings.push(read_string(source)?);
+        }
```
bobbinth, line 52: As mentioned in one of the other comments, here and in other places in this file, we should either try to use "native" methods to deserialize vectors or use `ByteReader::read_many_iter()`.

crates/mast-package/src/debug_info/mod.rs
djolertrk: Ah, yes, removing it.

core/src/operations/decorators/debug_var.rs
djolertrk: The `Vec::<u8>::read_from` is actually good fit here.

crates/mast-package/src/debug_info/serialization.rs
djolertrk: Yes, lets use `read_many_iter` here.

core/src/mast/debuginfo/debug_var_storage.rs
djolertrk: https://github.com/0xMiden/miden-vm/issues/2716

processor/src/host/mod.rs
djolertrk: But when using the `FastProcessor`, we still need to be notified in a way that the location has changed, so I think we do need the callback. (cc @bitwalker)

bobbinth: I think the intent was to execute the `FastProcessor` in a mode where we move though execution step-by-step, and so, at each cycle, we should be able to get the debug var info from the current MAST forest. @bitwalker - is that how you were thinking about it?

djolertrk: I talked to @bitwalker, and he confirmed that we can remove the callback. Thanks @bobbinth :)

bitwalker:
@bobbinth This is ready to merge now

PR #2474, opened 2025-12-15
[deubginfo] Add .debug_functions section

djolertrk:
This section contains the names of all procedures (both exported and private) in a package, keyed by their MAST root digest. This allows debuggers to resolve human-readable procedure names during execution.

TODO: Generate this section from asembly and compiler.

### DIFF
--- a/core/src/mast/debuginfo/mod.rs
+++ b/core/src/mast/debuginfo/mod.rs
@@ -132,6 +132,7 @@ impl DebugInfo {
             op_decorator_storage,
             node_decorator_storage: NodeToDecoratorIds::new(),
             error_codes: BTreeMap::new(),
+            procedure_names: BTreeMap::new(),
         }
     }
 


djolertrk:
cc @greenhat @bitwalker

djolertrk:
NOTE: The generation of this section depends on @bitwalker's work on assembler.

crates/mast-package/src/package/sections/debug_functions.rs
```diff
@@ -0,0 +1,168 @@
...
+/// Maps MAST root digests to procedure names for debugging purposes.
+///
+/// This section contains the names of all procedures (both exported and private)
+/// in a package, keyed by their MAST root digest. This allows debuggers to
+/// resolve human-readable procedure names during execution.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugFunctions {
+    /// Map from MAST root digest to procedure name
+    functions: BTreeMap<[u8; 32], String>,
```
bitwalker: Could you elaborate on why you've chosen not to use `Word` for the key here?

bitwalker (commented):
Just one question before I finish up my review, see the comment I left on `debug_functions.rs`

crates/mast-package/src/package/sections/debug_functions.rs
```diff
@@ -0,0 +1,168 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Maps MAST root digests to procedure names for debugging purposes.
+///
+/// This section contains the names of all procedures (both exported and private)
+/// in a package, keyed by their MAST root digest. This allows debuggers to
+/// resolve human-readable procedure names during execution.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugFunctions {
```
bobbinth: Could there be an argument for putting this into the `DebugInfo` struct of the `MastForest`? It already contains all debug-related info about the code it contains and this feels like a natural addition to that - it could look like:
```rust
pub struct DebugInfo {
    /// All decorators in the MAST forest.
    decorators: IndexVec<DecoratorId, Decorator>,

    /// Efficient access to decorators per operation per node.
    op_decorator_storage: OpToDecoratorIds,

    /// Efficient storage for node-level decorators (before_enter and after_exit).
    node_decorator_storage: NodeToDecoratorIds,

    /// Maps error codes to error messages.
    error_codes: BTreeMap<u64, Arc<str>>,

    /// Map from MAST root digest to procedure name
    procedure_names: BTreeMap<Word, String>,            // <- this is the new field
}
```

bitwalker: Either is fine, though I think we'd want to use `Arc<Path>` for the value type, but that's a minor detail. In my mind, the `DebugInfo` struct was very decorator-specific, but since this section is going to be assembler-generated, there isn't any reason why it can't be part of `DebugInfo` instead. We took the approach of defining it as a separate custom section simply because we have other debugging-related metadata that isn't assembler-generated, and will use custom sections, so we were just staying consistent with defining those as custom sections, but its definitely not a requirement.

crates/mast-package/src/package/sections/debug_functions.rs
djolertrk: It can be `Word` actually.

crates/mast-package/src/package/sections/mod.rs
bitwalker, line 1: I'd just remove this for now

core/src/mast/mod.rs
```diff
@@ -650,6 +650,25 @@ impl MastForest {
...
+
+    /// Returns an iterator over all (digest, name) pairs of procedure names.
+    pub fn procedure_names(&self) -> impl Iterator<Item = (Word, &String)> {
+        self.debug_info.procedure_names()
+    }
+
+    /// Inserts a procedure name for the given MAST root digest.
+    pub fn insert_procedure_name(&mut self, digest: Word, name: String) {
+        self.debug_info.insert_procedure_name(digest, name);
+    }
```
bobbinth, line 672: Is this something we'd want the assembler to do during assembly time? If so, should we create an issue to make use of this functionality in the future?

Also, should we check here that the procedure with the specified digest actually exists in the forest? While inserting a procedure with a digest that is not in the forest probably doesn't harm anything, it is likely a result of some bug - and so, at the very least, I think we should put an `assert!` here. But we could also consider returning an error.

core/src/mast/debuginfo/mod.rs
```diff
@@ -74,6 +75,10 @@ pub struct DebugInfo {
 
     /// Maps error codes to error messages.
     error_codes: BTreeMap<u64, Arc<str>>,
+
+    /// Maps MAST root digests to procedure names for debugging purposes.
+    #[cfg_attr(feature = "serde", serde(skip))]
+    procedure_names: BTreeMap<LexicographicWord, String>,
```
bobbinth: nit: AFAICT, we do not modify procedure names after inserting them. Should we then just use `Arc<str>` instead of `String` (as we do for error messages)?

core/src/mast/debuginfo/mod.rs
```diff
@@ -259,6 +267,43 @@ impl DebugInfo {
...
+    // PROCEDURE NAME METHODS
+    // --------------------------------------------------------------------------------------------
+
+    /// Returns the procedure name for the given MAST root digest, if present.
+    pub fn procedure_name(&self, digest: &crate::Word) -> Option<&str> {
+        self.procedure_names.get(&LexicographicWord::from(*digest)).map(|s| s.as_str())
+    }
+
+    /// Returns an iterator over all (digest, name) pairs.
+    pub fn procedure_names(&self) -> impl Iterator<Item = (crate::Word, &String)> {
```
bobbinth: nit: I'd import `Word` at the top of the file so that we don't have to specify the full path here and in a few other places.

core/src/mast/serialization/mod.rs
```diff
@@ -303,6 +314,9 @@ impl Deserializable for MastForest {
             .debug_info
             .extend_error_codes(error_codes.iter().map(|(k, v)| (*k, v.clone())));
 
+        mast_forest.debug_info.clear_procedure_names();
+        mast_forest.debug_info.extend_procedure_names(procedure_names);
```
bobbinth, line 331: I find the need to clear procedure names and then extend them a bit confusing - but I guess we do the same for error codes. Nothing to do in this PR, but maybe it is something we can remedy when we are refactoring MAST serialization (cc @huitseeker).

core/src/mast/serialization/mod.rs
```diff
@@ -220,6 +228,9 @@ impl Deserializable for MastForest {
         let error_codes: BTreeMap<u64, Arc<str>> =
             error_codes.into_iter().map(|(k, v)| (k, Arc::from(v))).collect();
 
+        // Reading procedure names
+        let procedure_names: BTreeMap<crate::Word, String> = Deserializable::read_from(source)?;
```
bobbinth, line 232: Related to one of my previous comments: should we check that all deserialized procedure names map to valid procedure roots in the forest?

core/src/mast/mod.rs
djolertrk: The issue is here https://github.com/0xMiden/miden-vm/issues/2531.

>Also, should we check here that the procedure with the specified digest actually exists in the forest? While inserting a procedure with a digest that is not in the forest probably doesn't harm anything, it is likely a result of some bug - and so, at the very least, I think we should put an assert! here. But we could also consider returning an error.

Yes, it does make sense, at least an assertion should be added. Thanks!

core/src/mast/debuginfo/mod.rs
djolertrk: Yep, works.

core/src/mast/serialization/mod.rs
djolertrk: Yes, it makes sense. Thanks!

core/src/mast/mod.rs
```diff
@@ -650,6 +650,29 @@ impl MastForest {
...
+    pub fn procedure_names(&self) -> impl Iterator<Item = (Word, &Arc<str>)> {
+        self.debug_info.procedure_names()
+    }
+
+    /// Inserts a procedure name for the given MAST root digest.
+    pub fn insert_procedure_name(&mut self, digest: Word, name: Arc<str>) {
+        assert!(
+            self.nodes.iter().any(|node| node.digest() == digest),
+            "attempted to insert procedure name for digest not in forest"
+        );
```
bobbinth, line 670: I think here we'd want to check only against procedure roots - not all nodes. We could use something like `find_procedure_root()` method to check if a given digest is a procedure root.

core/src/mast/serialization/mod.rs
```diff
@@ -303,6 +316,20 @@ impl Deserializable for MastForest {
...
+        mast_forest.debug_info.clear_procedure_names();
+
+        // Validate that all procedure name digests correspond to nodes in the forest
+        for digest in procedure_names.keys() {
+            if !mast_forest.nodes.iter().any(|node| node.digest() == *digest) {
+                return Err(DeserializationError::InvalidValue(format!(
+                    "procedure name references digest not found in forest: {:?}",
+                    digest
+                )));
+            }
```
bobbinth: Same comment as above: iterating through all nodes is going to be pretty expensive because there usually would be much fewer procedures than nodes. So, we should check against procedure roots rather than all nodes.

djolertrk:
Rebased on top of trunk and addressed the comments.

core/src/mast/mod.rs
bitwalker: A key detail here is that these procedure names map to both exported and non-exported procedures in the forest, so as long as the forest provides a way to determine if a given node is a procedure root for a non-exported procedure, then 👍.

bobbinth: Yes, I believe `procedure_roots` contains roots of all procedures (exported and not-exported).

bobbinth (approved):
All looks good! Thank you!

PR #2472, opened 2025-12-15
Migrate backend to Plonky3

Al-Kindi-0:
This PR replaces Winterfell with Plonky3 as the STARK proving backend.

## What Changed

**STARK Backend**: Removed all Winterfell dependencies (`winter-air`, `winter-prover`, `winter-verifier`) and integrated Plonky3. The VM now uses 13 core Plonky3 crates plus 4 custom Miden extensions from `github.com/0xMiden/p3-miden`.

**Hash Functions**: Added support for multiple hash functions with runtime selection. Users can now choose between Blake3, Keccak, RPO256, Poseidon2, and RPX256 for proof generation.

**Constraint System**: Deleted ~7,900 lines of explicit constraint code. Constraints are now expressed using Plonky3's builder pattern via the `MidenAirBuilder` trait from `p3-miden-air`.

**Metal GPU Support**: Removed Metal acceleration code (~1,000 lines) as Plonky3 doesn't support it.

**Dependencies**: Updated `miden-crypto` to a Plonky3-compatible branch (`huitseeker/p3-rebased`).

## Notes

- Supersedes #2258 and #2467 (rebasing was impractical due to divergence)
- Minimum trace length increased to 2048 due to current FRI implementation
- Blake3 uses binary folding (log_folding_factor=1) to avoid verification issues
- Blake3_192 support pending (needs Plonky3 upstream changes)
- base miden-crypto PRs: https://github.com/0xMiden/crypto/pull/720 and https://github.com/0xMiden/crypto/pull/736

PR #2484, opened 2025-12-18
Remove HALT-insertion bug from `build_trace()`

### LINKED ISSUE #2251: Parallel trace generation: remove HALT bug
From #2188's description:

> As identified in #1383, `Process` doesn't always include a `HALT` operation, specifically in the edge case where the last `END` (before the program halting) is right before a power-of-2 boundary, e.g. 63. Then when we insert the random row, this pushes the number of rows to 64, and therefore no padding is inserted (which is the only case where `HALT` is inserted). We mirror that bug into `build_trace()` so that tests pass - there's a comment to remind us to remove that bug once we get rid of `Process`.

Blocked on: removing `Process`

plafer:
Closes #2251 

Now that `Process` is removed, we no longer need to intentionally forget to insert HALT when the trace is exactly a power-of-two length (an old bug in `Process`). We previously added the same bug in `build_trace()` to ensure that both trace generators created the same trace.


bobbinth:
Since #2472 has been merged, this PR needs to be rebased.

Al-Kindi-0 (approved):
LGTM

PR #2493, opened 2025-12-23
perf: deduplicate basic block operation batches in HasherOp replay

### LINKED ISSUE #2250: `HasherOp`: improve handling of basic blocks
When recording the replay data necessary to build the `Hasher` chiplet trace, we need to record the `op_batches: Vec<OpBatch>` and `expected_hash: Digest` necessary to call `Hasher::hash_basic_block()`.

In #2188, we do the naive solution: each time a basic block is entered during execution, clone its operation batches, and store it in the replay. The problem is that if a basic block is entered multiple times, then we will needlessly duplicate its operation batches in the replay. A better representation to only store each `Vec<OpBatch>` could be backed by a `BTreeMap<Digest, Vec<OpBatch>`, where `Digest` is the digest of the batch.

One thing to keep in mind though is that we currently don't parallelize the hasher chiplet trace generation, but that might become necessary if it is/becomes a bottleneck; further benchmarking is needed at the time of writing.

berZKerk:
When recording replay data for Hasher chiplet trace generation, operation batches  for basic blocks were cloned every time a block was entered, causing unnecessary duplication when the same basic block was entered multiple times.

This change introduces a BTreeMap<Word, Vec<OpBatch>> to store each unique set of operation batches only once, keyed by the basic block digest. 
The HasherOp enum now stores only the digest instead of the full batches, reducing memory usage and improving performance for programs with repeated basic block executions.

Fixes #2250

### DIFF
--- a/processor/src/parallel/mod.rs
+++ b/processor/src/parallel/mod.rs
@@ -17,7 +17,7 @@ use miden_air::{
     },
 };
 use miden_core::{
-    Kernel, ONE, Operation, Word, ZERO, field::Field, stack::MIN_STACK_DEPTH, utils::uninit_vector,
+    Kernel, ONE, Operation, Word, ZERO, stack::MIN_STACK_DEPTH, utils::uninit_vector,
 };
 use rayon::prelude::*;
 


processor/src/fast/trace_state.rs
```diff
@@ -1012,8 +1020,36 @@ impl HasherRequestReplay {
     }
 
     /// Records a `Hasher::hash_basic_block()` request.
+    ///
+    /// Deduplicates operation batches by storing them in a map keyed by the basic block digest.
+    /// If the same basic block is entered multiple times, only one copy of the operation batches
+    /// is stored.
     pub fn record_hash_basic_block(&mut self, op_batches: Vec<OpBatch>, expected_hash: Word) {
```
huitseeker, line 1033: Consider adding tests that verify deduplication works when the same basic block is executed multiple times.

The existing tests pass, but none specifically check that `op_batches_map` deduplicates correctly. A simple test could record the same digest three times and verify the map has one entry while `hasher_ops` has three.

Example:

```rust
#[test]
fn test_hasher_request_replay_deduplicates_basic_blocks() {
    let mut replay = HasherRequestReplay::default();
    let digest: Word = [Felt::new(1), Felt::new(2), Felt::new(3), Felt::new(4)];
    let op_batches = vec![OpBatch::new(vec![Operation::Add], Vec::new()).unwrap()];

    replay.record_hash_basic_block(op_batches.clone(), digest);
    replay.record_hash_basic_block(op_batches.clone(), digest);
    replay.record_hash_basic_block(op_batches.clone(), digest);

    assert_eq!(replay.op_batches_map().len(), 1);

    let (hasher_ops, _) = replay.into_parts();
    assert_eq!(hasher_ops.len(), 3);
}
```

processor/src/fast/trace_state.rs
```diff
@@ -1012,8 +1020,36 @@ impl HasherRequestReplay {
...
-        self.hasher_ops.push_back(HasherOp::HashBasicBlock((op_batches, expected_hash)));
+        // Only store the op_batches if we haven't seen this digest before
+        // If the digest already exists, we verify that the op_batches match (they should, since
+        // the digest is computed from the op_batches)
+        match self.op_batches_map.entry(expected_hash) {
+            alloc::collections::btree_map::Entry::Vacant(entry) => {
+                entry.insert(op_batches);
+            },
+            alloc::collections::btree_map::Entry::Occupied(_) => {
```
huitseeker: The comment says "we don't verify to avoid overhead", but a `debug_assert_eq!` would catch bugs during testing without runtime cost in release builds.

Consider:

```rust
alloc::collections::btree_map::Entry::Occupied(entry) => {
    debug_assert_eq!(
        entry.get(),
        &op_batches,
        "Same digest should always map to same op_batches"
    );
},
```

This guards against implementation bugs where different batches produce the same digest.

processor/src/fast/trace_state.rs
huitseeker, line 1: The `impl IntoIterator for HasherRequestReplay` still exists, but seems unused, now. It only returns `hasher_ops` and drops the `op_batches_map`. We should delete it to avoid confusion with `into_parts`.

bobbinth:
Closing this as stale.

PR #2505, opened 2025-12-30
ci: fix cargo publish workflow

huitseeker:
https://github.com/0xMiden/miden-vm/commit/c8e1833ddeb3a037abffb30dc3e01211eb1338bf is failing its `cargo publish --dry-run --workspace` because of a menial time-of-pr versus time-of-merge issue. See commit messages.


Cargo.toml
```diff
@@ -58,7 +58,7 @@ miden-core-lib = { path = "./crates/lib/core", version = "0.21", default-feature
 miden-utils-core-derive = { path = "./crates/utils-core-derive", version = "0.21", default-features = false }
 miden-utils-diagnostics = { path = "./crates/utils-diagnostics", version = "0.21", default-features = false }
 miden-utils-indexing = { path = "./crates/utils-indexing", version = "0.21", default-features = false }
-miden-utils-sync = { path = "./crates/utils-sync", version = "0.21", default-features = false }
+miden-utils-sync = { path = "./crates/utils-sync", version = "0.21" }
```
bobbinth: Is this something that appeared recently? IIRC, we published v0.20 VM with default features turned off for `miden-utils-sync`.

huitseeker: The change establishing std for miden-utils-sync is from the late innings of https://github.com/0xMiden/miden-vm/pull/2447.

bobbinth: I'm still not sure I understand why this is happening. AFAICT, `OnceLockCompat` is exported from `miden-utils-sync` both in `std` and in `no_std` context (in `no_std` we export the wrapper around `OnceBox`). So, even with `default-features = false` it should work - no?

huitseeker: Yep, this is not reproducible for me locally. My hypothesis is that in CI, the dry-run :
  - Packages each crate independently
  - Verifies each package in isolation using a temporary registry
  - This isolated verification context appears to have different feature resolution behavior than normal workspace builds.
  
  Nonetheless, out of an abundance of caution, I'll close this.

Issue #1919, opened 2025-06-26
RFC: Introduce `miden-project.toml` manifest file


bitwalker:
Currently, there is no way to infer the name of a Miden Assembly module, nor is there a way to safely infer the namespace of a project directory containing Miden Assembly modules. In both cases, we currently expect someone to parse modules by providing this information manually. The standard library, for example, is built in its `build.rs` by manually specifying the `std` namespace to `Library::from_dir` - it is neither based on anything in the source tree, nor inferrable from any context available.

This problem extends to _all_ Miden Assembly sources, whether part of a project directory, or standalone modules. As Miden evolves, we now have other complications that are making the situation worse in a variety of ways:

1. It is currently not possible to properly implement an LSP server with anything other than very basic syntax highlighting and file-local symbol indexing. This is because all of the richer features we'd like to implement (go to definition, list references, show docs for imported procedures, etc.) require us to actually parse and perform semantic analysis on the documents in the current workspace. The problem is, we can't actually do that without - _at a minimum_ - knowing the namespace of the workspace. Using the name of a parent directory doesn't even work when you don't know what the actual root directory of the project is - you need some file that you can use as a marker for that purpose. Looking at the standard library, not only is the root directory named `asm`, but even if it was `std`, we still wouldn't know that when looking at say, `std/mem.masm`, that the `std` directory is the root directory for the library.
2. We support dependencies between libraries, and this is currently entirely implicit, i.e. if we encounter a reference to some imported module `foo::bar`, where `foo` is a different namespace than the current one, we have no means by which to figure out where the `foo` library even is. We require someone to provide to `Assembler` the needed libraries so that all such references can be resolved at assembly-time, but that isn't helpful for things such as LSP, where we want to be able to provide information about the referenced module (e.g. completion, docs, type information when it lands, etc.).
3. We now have other metadata about Miden projects that we care about, such as storage layout, and whatever other account component metadata is relevant. We might have other types of configuration that we wish to support on a per-project basis, such as formatting and linter settings, whether to build with debug information, and likely more that I'm not even thinking about at the moment.

For me, it is clear that we are due for implementing a project manifest for Miden Assembly projects. This would provide for a ton of ergonomic improvements across the board, and in particular, enable a large range of LSP functionality that I'm currently blocked on implementing.

## Proposal

I'm proposing a simple project file, named `miden-project.toml`, be implemented using TOML as the underlying text format, shown below in its various permutations.

[...]

PR #2510, opened 2025-12-30
feat: initial implementation of project file format

bitwalker:
This PR implements a new crate, `miden-project`, which implements the data structures for representing Miden projects, along with the AST representation of the `miden-project.toml` file initially sketched out in #1919. Note that the actual syntax of `miden-project.toml` as implemented in this PR differs from that initial sketch in a few ways. See the example project layouts in `crates/project/examples` for the syntax of workspace-based and standalone project files.

I'm still making a few small adjustments, and implementing tests for the parsing/validation bits, but this should be more or less reviewable as-is, at least in terms of the high-level design.

#### Links

* Original RFC for `miden-project.toml` can be found in #1919
* Discussion on dependency specs and versioning can be found [here](https://github.com/0xMiden/compiler/issues/484#issuecomment-3702659109)

#### TODO

- [ ] Finish implementing tests
- [x] Add the necessary `pubgrub` plumbing for package selection during dependency resolution
- [x] Clean up pass (remove unused dependencies, make sure everything is correctly documented, remove unused APIs)

### DIFF
--- a/crates/project/README.md
+++ b/crates/project/README.md
@@ -158,6 +158,26 @@ f = { git = "https://github.com/example/f", rev = "deadbeef" }
 g = { git = "https://github.com/example/g", rev = "deadbeef", version = "~> 0.1.0" }
 ```
 
+#### Linkage
+
+Dependencies are dynamically-linked by default, which means that the assembled
+package will have a runtime dependency on the referenced package, and it is
+assumed that the dependency will have been loaded into the VM at runtime.
+
+You may also specify that a dependency should be statically-linked, like so:
+
+```toml
+[dependencies]
+a = { version = "=0.1.0", linkage = "static" }
+```
+
+When a dependency is statically-linked, it is as if the dependency was defined
+as part of your own package.
+
+NOTE: Kernel dependencies are always statically-linked into executable targets,
+and dynamically-linked for library targets, and it is not possible to change
+this behavior.
+
 #### Semantics
 
 * `a` specifies a semantic version requirement, these would be evaluated against a package registry implementation
@@ -171,18 +191,6 @@ g = { git = "https://github.com/example/g", rev = "deadbeef", version = "~> 0.1.
 In cases where the dependency is resolved to project sources and _not_ an assembled package, the behavior would be to assemble those dependencies first, and then link against them when assembling the current project. This is most useful when linking against packages which are _not_ contracts, or where the contracts are deployed together as a unit.
 
 **NOTE:** Currently there is no canonical package registry, so the resolution of the first two forms described above is dependent on the specific tool that is doing the resolution, namely, how it populates the package index for the resolver provided by this crate.
-
-### Build profiles
-
-TODO
-
-### Custom package metadata
-
-TODO
-
-### Lint configuration
-
-TODO
 
 ## License
 This project is dual-licensed under the [MIT](http://opensource.org/licenses/MIT) and [Apache 2.0](https://opensource.org/license/apache-2-0) licenses.
--- a/crates/project/examples/protocol/userspace/miden-project.toml
+++ b/crates/project/examples/protocol/userspace/miden-project.toml
@@ -8,4 +8,6 @@ path = "mod.masm"
 
 [dependencies]
 miden-utils = { workspace = true, linkage = "dynamic" }
+# NOTE: As described in the miden-project README, this will be linked as a kernel automatically
+# as `miden-tx` defines a kernel target
 miden-tx.workspace = true


crates/project/examples/workspace/note/miden-project.toml
```diff
@@ -0,0 +1,11 @@
+
+[package]
+name = "my-account-note"
```
greenhat: From https://github.com/0xMiden/miden-vm/issues/1919:

>The Rust tooling uses/reproduces much of this information in Cargo.toml - we could move most of that to miden-project.toml instead, or alternatively, simply merge the relevant sections from Cargo.toml into an in-memory miden-project.toml manifest. Much of what I've laid out here is in fact the parts we already define/maintain in Cargo.toml for Rust-based projects today

I wonder what our plan is for Rust projects? 
Currently we store all Miden-related fields in the `[package.metadata.miden]` section of the Cargo.toml. See a [P2ID example](https://github.com/0xMiden/compiler/blob/e8de1aefc40383b3240d8134a1273383760ae5f7/examples/p2id-note/Cargo.toml?plain=1#L18-L30).

Do we want to require `miden-project.toml` and move the Miden fields currently residing in the Cargo.toml there? Multiple fields end up duplicated (name, version, etc.). Plus, the developer would have package dependencies defined in the `miden-project.toml` and the crate dependencies in the Cargo.toml.

I'd suggest to keep storing all Miden fields in Cargo.toml. The `cargo miden build` can embed generated `miden-project.toml` in the compiled Miden package.

bitwalker: I think the issue with storing everything in `Cargo.toml` has a few issues:

1. We will have tooling, and there will be third-party tooling, built around Miden projects, and storing project information in `Cargo.toml` requires those tools to understand not only `miden-project.toml`, but also `Cargo.toml` and however we choose to encode `miden-projet.toml` information inside of it.
2. In the future, other languages that can compile to Miden may not have a TOML configuration file (or at least one that can be piggy-backed on like `Cargo.toml`), in which case they would need to use `miden-project.toml` and would face the same dilemma of duplication between language-specific project config and Miden project config. Being consistent about the boundary between language-specific config and Miden config up front will lead to less confusion IMO.
3. It complicates IDE/LSP integration, since `Cargo.toml` is already used as a language marker for  Rust, it would require parsing `Cargo.toml` to determine if the directory is a Miden/Rust project, or just a Rust project. In some editors it may not be a problem, if extensions are provided an opportunity to do this when a worktree is opened in the editor (e.g. Zed), but I don't believe that is always the case.
4. Even if we were to use `Cargo.toml` for Miden+Rust projects, we still have a situation where we are specifying dependencies in two places in `Cargo.toml` - one for normal crates, and then one for Miden packages under (`[package.metadata.miden.dependencies]`). IMO, it's more likely to confuse users that both sets of dependencies are specified in different places in the same file. If Miden package dependencies are in `miden-project.toml` and pure Rust dependencies are in `Cargo.toml`, the delineation between the two is a lot more clear (and easier to explain).
5. Rust crates which are compiled to Miden will not be published to crates.io - so virtually all of the metadata that would typically go in `Cargo.toml` can be specified in `miden-project.toml` instead, aside from `name` and `version`. GIven that, it isn't clear to me how much information would actually be duplicated, or how much of a pain that would be in practice. Changing the `name` is going to be very rare. Changing `version` is more frequent, so if it is considered painful enough to have to maintain it in two places, we could provide a solution there that allows specifying in `miden-project.toml` that the version is to be inferred from external/language-specific tooling, something like `version = { external = true, language = "rust" } | { external = true, path = "VERSION" }`. TBH, I don't think it is actually a significant enough pain point for that, but 🤷. Aside from `version` though - there just isn't overlap between `Cargo.toml` and `miden-project.toml` to combine them IMO.

I'm not saying I couldn't be persuaded on this point, but it does feel like preemptively addressing an issue that we don't actually know is going to be an issue in practice. I would want to see how much duplication there actually is (given the constraints I outlined above, e.g. defining `description` in `Cargo.toml` doesn't make any sense), and from there determine if it is a significant enough problem to warrant complicating project handling by allowing Rust projects to piggy-back on `Cargo.toml`.

greenhat: Fair points. I'm sold on the project-level `miden-project.toml`. 

The workspace-level `miden-project.toml` still bothers me due to the `members` and `profile` duplication with Cargo.toml. Assuming they have to be maintained by the user.

greenhat:
I like the version-or-digest approach and that we'll store it in the same field. Apart from my question about the workspace level in Rust projects at https://github.com/0xMiden/miden-vm/pull/2510#discussion_r2656191659 everything sounds good.

bitwalker:
This is ready for review, though I expect to add a few more tests still, and write up a README doc that describes the usage and syntax of the project file.

A few design details to note during your review:

* No changes are made to existing crates in this repo for this PR - this is all greenfield code in a new crate
* The majority of this PR's content can be broken up into three categories, which can be reviewed independently to make it easier to focus on the relevant details:
  1. The format/syntax/features of the project file, the data structures that represent its AST, and the data structures that represent the loaded/validated representation which is derived from the AST.
  2. The basic serialization boilerplate, the parsing functions that deserialize the AST from TOML source code, and the code to translate from the AST to the final user-facing data structures.
  3. The dependency resolver and associated implementation details. Note that the resolver does not actually fetch/load packages, it provides only the high-level resolution and version selection based on a simple index of packages, their versions, and the dependency requirements of those versions.
* There are a handful of examples in the crate which are used in the tests, and provide a sense of the syntax and relationship between project file and directory structure/packages.
* `TargetType` is essentially a duplicate of `PackageKind`, and I expect that `PackageKind` will be replaced with `TargetType` in subsequent PRs.
* I expect to make `miden_project::Package` derivable from a `miden_mast_package::Package`, or even embed the manifest directly in the latter, in subsequent PRs. Either way, the goal would be that if you have a `miden_mast_package::Package`, you have access to the version of the manifest used to assemble that package. This can then be used in dependency resolution, etc.
* The dependency resolution algorithm is provided by `pubgrub`, and is set up to largely mimic Cargo's dependency resolution behavior, with some simplifications. Additionally, our usage of it extends its behavior to support version requirements that use content digests in addition to, or in lieu of semantic version constraints. This aspect of the PR is experimental, and likely to change/evolve with further testing, but is an essential component of downstream tooling that will build on this, so some implementation of its functionality was needed.

If you have any questions about specific details, leave a comment here or ping me to discuss, I'm happy to walk through any of this, since there is a lot of context that isn't easy to glean from just the source code or its documentation.

bitwalker:
> I like the version-or-digest approach and that we'll store it in the same field. Apart from my question about the workspace level in Rust projects at [#2510 (comment)](https://github.com/0xMiden/miden-vm/pull/2510#discussion_r2656191659) everything sounds good.

The duplication is hard to avoid at the workspace level. We could derive the `Workspace` for Rust projects from `Cargo.toml`, but that leaves the issue that I mentioned before, which is that this will break expectations for tooling that only cares about Miden projects in general, and not the language-specific details.

Regardless, we need the workspace-level project file for workspace management of Miden Assembly projects anyway, so the issue of duplication with `Cargo.toml` is something specific to the Rust toolchain, rather than an issue to be addressed here, IMO.

Makefile
```diff
@@ -32,7 +32,7 @@ WARNINGS                 := RUSTDOCFLAGS="-D warnings"
 BUILDDOCS                := MIDEN_BUILD_LIB_DOCS=1
 
 # -- feature configuration ------------------------------------------------------------------------
-ALL_FEATURES_BUT_ASYNC   := --features concurrent,executable,metal,testing,internal
+ALL_FEATURES_BUT_ASYNC   := --features concurrent,executable,metal,testing,internal,resolver
```
bobbinth: In `next`, I'm changing this to `ALL_FEATURES` as we con't have an `async` feature any more (https://github.com/0xMiden/miden-vm/pull/2616).

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.3 (TBD)
+
+- Define and implement Miden project file format ([#2510](https://github.com/0xMiden/miden-vm/pull/2510)).
```
bobbinth: Let's rebase this to go into `next`.

crates/project/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+TODO
+
+### Custom package metadata
+
+TODO
+
+### Lint configuration
+
+TODO
```
bobbinth: nit: let's add the "LICENSE" section as we do for other crates.

crates/project/examples/multi_target/miden-project.toml
bobbinth, line 1: It would be good to add some comments explaining the example. My understanding is: this project defines a kernel with two targets:

- The kernel executable, under `kernel/mod.masm`.
- The kernel library under the namespace `userspace`.

If the above is correct, I believe this is similar to how we define the [transaction kernel](https://github.com/0xMiden/miden-base/tree/next/crates/miden-protocol/asm/kernels/transaction) in `miden-base`, where:

- `kernel/mod.masm` is meant to be similar to `main.masm`.
- `userspace/mod.masm` is meant to be similar to `api.masm`.

Currently, we assemble these programmatically using the [build.rs](https://github.com/0xMiden/miden-base/blob/next/crates/miden-protocol/build.rs) script - but once we have the project file (and update the assembler to be able to work with projects), I'm assuming the `build.rs` could be greatly simplified - right? Basically, we'd be able to instantiate the assembler and then just feed the project file to it, and everything else will happen automatically.

A few questions/comments:

- I think `kind = "kernel"` should be attacked to the `lib/mod.masm` target. I assume this is used to assemble the library using `Assembler::assemble_kernel_from_dir()` and this function outputs kernel libraries rather than executables.
  - This would also mean that for `kernel/mod.masm` we'd have `kind = "executable"` - right?
- `Assembler::assemble_kernel_from_dir()` takes two parameters: (1) path to the module - which I think would be `lib/mod.masm`, and (2) path to directory of the library that would get statically linked into the kernel. What would be the directory path here? Or would this be changed somehow?
- Related to the above, when we assemble the kernel library, we don't need to specify the namespace as we automatically get the `$kernel` namespace. So, we probably don't need the `namespace = "userspace"` attribute (and it should probably not be allowed for `kernel` targets).
- How would the module graph look here? Specifically:
  - Since `kernel/mod.masm` is an executable, I'm assuming we can't import anything from there. So, it is unclear why some procedures there are declared as `pub` (that should be an assembly error, right?).
  - I'm also assuming since `userspace/mod.masm` is meant to be a kernel API file, we won't be able to refer to it from `kernel/mod.masm` - right?
  - If the above is correct, what would be the best way to replicate the "shared library" concept which we manually orchestrate in `miden-base`? Specifically, we have the `lib` directory that can be references from both the kernel executable and the kernel library.

crates/project/examples/package/lib/mod.masm
bobbinth, line 1: This file seems to be empty. It may be good to add some dummy procedure here - or a comment explaining why it is intentionally left empty.

crates/project/examples/package/miden-project.toml
bobbinth, line 1: Similar to my other comment, it would be good to add some comments here explaining briefly what this example is.

crates/project/examples/package/miden-project.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "example"
+version = "0.1.0"
+
+[[target]]
+kind = "library"
+path = "lib/mod.masm"
+
+[dependencies]
+my-kernel = { path = "../multi_target/userspace" }
```
bobbinth: Do we not need to identify `my-kernel` dependency as a kernel somehow? Kernel dependencies are different from regular dependencies and there can be only one such dependency per package. Maybe a better syntax here could be:

```toml
[[target]]
kind = "library"
path = "lib/mod.masm"
kernel = { path = "../multi_target" }
```
Also, notice that I've changed the path to just `/multi-target` as I'm assuming we want to point at a package rather than a specific sub-directory of the package.

crates/project/examples/package/miden-project.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[profile.test]
+inherits = "dev"
+network = "local"
+
+[lints.miden]
+unused = "error"
+
+[package.metadata.network.local]
+endpoint = "tcp://127.0.0.1"
```
bobbinth, line 19: This is linked to the `network = "local" definition above, right? But I guess how exactly it is linked would be defined downstream.

crates/project/examples/workspace/account/miden-project.toml
bobbinth, line 1: Similar to the other comments: may be good to add some brief comments here. Also, account components usually come with additional metadata. For example, would be good to illustrate how we'd specify the storage layout of the account here.

crates/project/examples/workspace/account/mod.masm
bobbinth, line 1: May be good to add one or two dummy procedures to this file (or explain why it is intentionally left blank).

crates/project/examples/workspace/account/miden-project.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "my-account"
+version = "0.1.0"
+
+[[target]]
+kind = "account-component"
```
bobbinth: We could specify `namespace` here as well, right? I'd probably do this as we encourage people to define unique namespaces for their components/notes - e.g., `my_project::my_component`

crates/project/examples/workspace/note/mod.masm
bobbinth, line 1: Once we add a dummy procedure to the `my-account` example, we could add an example here to illustrate how the account component would be imported and how a procedure from that component could be invoked.

crates/project/examples/workspace/miden-project.toml
bobbinth, line 1: Similar to other comments: I'd add comments with a brief explanation of this example here.

crates/project/README.md
```diff
@@ -0,0 +1,147 @@
...
+their path derived from the directory structure).
+
+There are other types of targets though, currently the available kinds are:
+
+* `library`, `lib` - produce a package which exports one or more procedures that can be called from other artifacts, but cannot be executed by the Miden VM without additional setup. Libraries have no implicit dependency on any particular kernel.
+* `executable`, `bin`, or `program` - produce a package which has a single entrypoint, and can be executed by the Miden VM directly. Executables have no dependency on any particular kernel. 
+* `account-component` - produce a package which is a valid account component in the Miden protocol, and contains all metadata needed to construct that component. This type is only valid in conjunction with the Miden transaction kernel.
+* `note-script` - produce a package which is a valid note script in the Miden protocol, and exports the necessary metadata and procedures to construct and execute the note. This type is only valid in conjunction with the Miden transaction kernel.
```
bobbinth: Here we say `note-script` but in the examples we use `note`. I think `note` is the right approach.

crates/project/README.md
```diff
@@ -0,0 +1,147 @@
...
+[[target]]
+kind = "lib"       # the type of artifact we're producing
+path = "mod.masm"  # the relative path to the root module
+namespace = "name" # the root namespace of modules parsed for this target
+```
+
+As noted above, we've added a target definition that is equivalent to the default target that is inferred if no targets are explicitly declared: a library, whose root module is expected to be found in `mod.masm`, and whose modules will all belong to the `name` namespace (individual modules will have
+their path derived from the directory structure).
+
+There are other types of targets though, currently the available kinds are:
```
bobbinth: I think we are missing `kernel` from the list below.

CHANGELOG.md
bitwalker: I've rebased it, but kept it against `main`, since ideally we release this in a 0.21 point release. The other PRs which change the assembler will target `next` though

crates/project/examples/multi_target/miden-project.toml
bitwalker: > It would be good to add some comments explaining the example. 

For sure, these were thrown together for simple testing, but I can certainly add some documentation explaining the setup in theory.

> My understanding is: this project defines a kernel with two targets:

Not exactly - or rather, to be more precise, here we're defining a project that can produce two packages:

1. A kernel component, whose root module is located at `kernel/mod.masm` - we're specifying the path because how the project is organized differs from the default single-target layout where the path is assumed to be `mod.masm`.
2. A library component, whose root module is located at `lib/mod.masm`. This library component is intended to be the userspace abstraction over the kernel target. We're specifying a different namespace from the default just to exercise that option here, but by default it would be derived from the package name.

A couple of things to note:

* A project may only define a single kernel target.
* A project may define any number of non-kernel targets

> * `kernel/mod.masm` is meant to be similar to `main.masm`.
> * `userspace/mod.masm` is meant to be similar to `api.masm`.

The canonical expectation by the assembler is that the top-level module is always `mod.masm`. By top-level module, I mean the module which owns the namespace for the library. In practice, if this module is not present, we just treat it as if that module is empty, and parse other modules as children of the top-level namespace, but that's more a convenience than anything else. In practice, we should probably always have a `mod.masm` even if it's empty.

> Currently, we assemble these programmatically using the [build.rs](https://github.com/0xMiden/miden-base/blob/next/crates/miden-protocol/build.rs) script - but once we have the project file (and update the assembler to be able to work with projects), I'm assuming the `build.rs` could be greatly simplified - right? Basically, we'd be able to instantiate the assembler and then just feed the project file to it, and everything else will happen automatically.

Yes, exactly.

> * I think `kind = "kernel"` should be attacked to the `lib/mod.masm` target. I assume this is used to assemble the library using `Assembler::assemble_kernel_from_dir()` and this function outputs kernel libraries rather than executables.

As noted above, the userspace library isn't assembled as part of the kernel, because it exists _separate_ from the kernel - that's why these are two separate targets. This keeps the kernel itself stable, while allowing the userspace to change more frequently.


>   * This would also mean that for `kernel/mod.masm` we'd have `kind = "executable"` - right?

The assembler currently distinguishes between kernels and executables, but maybe that distinction is meaningless, and instead we should just have kernels and libraries, where executables are always kernels - that's effectively the case already.

If we do remove executables as a distinct concept, then we'd have no need to have `executable` targets here.

> * `Assembler::assemble_kernel_from_dir()` takes two parameters: (1) path to the module - which I think would be `lib/mod.masm`, and (2) path to directory of the library that would get statically linked into the kernel. What would be the directory path here? Or would this be changed somehow?

I think we'd just make `assemble_kernel_from_dir` take a single path, the path to the kernel module directory, where `mod.masm` is the kernel module itself, and all other modules are statically linked into the kernel artifact. There's no reason to have them be separate.

> * Related to the above, when we assemble the kernel library, we don't need to specify the namespace as we automatically get the `$kernel` namespace. So, we probably don't need the `namespace = "userspace"` attribute (and it should probably not be allowed for `kernel` targets).

It's not allowed for kernel targets, but to be clear, the library target here is the _userspace_ target, i.e. the library users of the kernel will use to interact with it, not library code consumed internally by the kernel. The latter is implicitly part of the kernel module hierarchy.

> * How would the module graph look here? Specifically:
>   
>   * Since `kernel/mod.masm` is an executable, I'm assuming we can't import anything 

You can import from modules defined as part of the kernel, e.g. if there's a `kernel/a.masm`, you could import that as `use $kernel::a` from the `$kernel` module. Note that those modules are internal to the kernel artifact, and are not visible to users of the kernel.

> So, it is unclear why some procedures there are declared as `pub` (that should be an assembly error, right?).

The assembler does not treat kernels as executables, but as libraries - executables are a separate thing. The kernel interface is defined as the set of procedures exported from the kernel module.

>   * I'm also assuming since `userspace/mod.masm` is meant to be a kernel API file, we won't be able to refer to it from `kernel/mod.masm` - right?

Right, the userspace is a separate target (which presumably depends on the kernel) so its modules will not be available to the kernel itself.

>   * If the above is correct, what would be the best way to replicate the "shared library" concept which we manually orchestrate in `miden-base`? Specifically, we have the `lib` directory that can be references from both the kernel executable and the kernel library.

You'd define a library target that is a dependency of both the kernel and the userspace.

crates/project/examples/package/lib/mod.masm
bitwalker: Sure, it's really just there to make the module organization clear.

crates/project/examples/package/miden-project.toml
bitwalker: I just noticed, but I failed to make the kernel target a dependency of the library target in the `multi-target` project, but that's where the kernel dependency would be derived from.

The correct dependency spec would actually be one of the following:

```
my-kernel = { path = "../multi_target" }

# OR

my-kernel = { path = "../multi_target", targets = ["userspace"] }
```

In both cases, the effect is the same - the kernel and userspace targets are linked against dynamically. The latter selects a specific set of targets to link against, but as mentioned above, dependencies are transitive, so we'd end up linking against the kernel in both anyway.

crates/project/examples/package/miden-project.toml
bitwalker: Right, this is just a sketch of what it might look like, nothing that is actually defined as part of the project file format itself.

crates/project/examples/workspace/account/miden-project.toml
bitwalker: Sure I can throw together some examples of this, though it is essentially undefined as part of the project file format itself, i.e. how metadata is conveyed via the project file is considered a separate concern.

crates/project/examples/workspace/account/miden-project.toml
bitwalker: Yes, you can always specify namespace for any component other than executables and kernels.

bitwalker: I should note that by default the namespace is derived from the project name, and multi-target projects must have unique namespaces for each target or a validation error is raised.

crates/project/README.md
bitwalker: Good catch, thanks!

crates/project/README.md
bitwalker: Indeed, thanks!

crates/project/examples/multi_target/miden-project.toml
bobbinth: I think I understand the intent better now - though, I think it differs a bit from how we set things up in the protocol repo. We have two cases there that would be good to cover with the project file so that we could simplify the `build.rs` files there as much as possible.

The two cases are:

- The transaction kernel (located [here](https://github.com/0xMiden/protocol/tree/next/crates/miden-protocol/asm)).
- The standard smart contracts (located [here](https://github.com/0xMiden/protocol/tree/next/crates/miden-standards/asm/standards)).

The standards library is pretty simple - it is a single library that has two dependencies: (1) Miden core lib and (2) transaction wrapper kernel lib. Both of these are regular dependencies (i.e., non-kernel dependency). I'm thinking that once we move this to a project format we could define this library via a single project file.

The transaction kernel is more complex. It consists of 4 parts:

- The kernel API (defined [here](https://github.com/0xMiden/protocol/blob/next/crates/miden-protocol/asm/kernels/transaction/api.masm)) - i.e., a set of procedures that can be `syscall`-ed by user programs. This library depends only on Miden core lib.
- The protocol library (defined [here](https://github.com/0xMiden/protocol/tree/next/crates/miden-protocol/asm/protocol) - this is a set of procedures that wraps all kernel API procedures. The primary purpose of this library is twofold: (1) to let the user call kernel procedures via the `exec` instruction (i.e., the wrapper procedures handle all necessary padding) and (2) to invoke kernel procedures by index rather than hash. This library depends on the Miden core lib as a regular dependency and on the kernel API library as a kernel dependency.
- Two executable files (defined [here](https://github.com/0xMiden/protocol/blob/next/crates/miden-protocol/asm/kernels/transaction/main.masm) and [here](https://github.com/0xMiden/protocol/blob/next/crates/miden-protocol/asm/kernels/transaction/tx_script_main.masm)) which is basically the "driver" of the transaction kernel. These executables depend on the Miden core lib and some internal modules of the kernel library (but these are used as a regular dependency).
- A set of shared utilities that are used by the kernel API and protocol library modules.

Expressing all these as a single project would probably be quite complicated and may not be worth it. So, would be good to figure out how we can split this up into 2 (or at most 3) projects and then make sure that the project file format covers these. One option is:

- The shared utilities project: this would contain the shared utilities defined as a simple library.
- The transaction kernel project: this would output 2 executables and a single kernel library.
- The protocol project: This would depend on shared utilities and the kernel library.

Also cc @PhilippGackstatter as he may have additional suggestions.

bitwalker: There's a few different ways to structure things, but your general outline sounds perfectly good to me - a few notes:

* The whole thing would be part of a single workspace, so dependencies can be trivially shared between projects in the workspace, including packages within the workspace.
* A single project can cover most of the kernel target pieces (the kernel library, and the two executables)
* The code shared between the kernel, userspace, and smart contracts can be in a separate library, nothing surprising there.
* The userspace library as a separate project makes sense to me - it simplifies things a bit.

I'll get this PR updated tomorrow at some point, and we can do a final pass on what things look like.

PhilippGackstatter: We are also building account components in miden-standards with build.rs. I just described a case (https://github.com/0xMiden/protocol/pull/2527#discussion_r2883102744) where it would be useful to have a separate library to be able to share code between two account components, but that code should not live in `miden::standards` directly. This is currently difficult to do (would require special code in build.rs), but I think if we can easily setup such library projects in the future with this project file, this would be a real help.

Afaict, this should be easily possible with the work done here, so I just wanted to mention that additional use case in miden-protocol as a reference.

One question I have in this context: Does the project file allow specifying if a dependency is linked statically or dynamically, or is this decided at a different level? At first glance, this would seem useful to say "link miden::standards dynamically" but "link multisig library statically".

bitwalker:
Just an update - I'm modifying the project file syntax a bit to restrict it in some ways, as the existing flexibility in the current draft makes for some awkward complexity that can be avoided with some simplification. I should have my updated draft up later tonight.

bitwalker:
I've documented the new approach in `crates/project/README.md`, and will follow up with the changes in implementation tonight/tomorrow morning.

crates/project/src/ast/profile.rs
```diff
@@ -0,0 +1,86 @@
...
+        fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
+        where
+            M: MapAccess<'de>,
+        {
+            let mut profiles = Self::Value::default();
+
+            while let Some((key, mut value)) = access.next_entry::<Span<Arc<str>>, Profile>()? {
+                value.name = key;
+
+                if let Some(prev) =
```
huitseeker, line 69: It looks like new profile entries never get added. In `visit_map`, we replace an existing entry when names match, but when there is no match we don't `push` `value`, so `[profile.*]` items can disappear during deserialization. Should we append in the `else` branch?

crates/project/src/dependencies/resolver/version_set.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    fn intersection(&self, other: &Self) -> Self {
+        if self.digests.is_empty() && other.digests.is_empty() {
+            return Self::from(self.range.intersection(&other.range));
+        }
+
+        let ldigests = BTreeSet::from_iter(self.digests.iter());
+        let rdigests = BTreeSet::from_iter(other.digests.iter());
+        let digests = ldigests.intersection(&rdigests).map(|d| **d).collect();
```
huitseeker: If one side has no digest filter and the other side does, intersecting both digest sets returns empty, which behaves like `Any` later. That can widen or narrow matches incorrectly depending on call path. Should empty digest mean wildcard during intersection?

crates/project/src/dependencies/resolver/version_set.rs
```diff
@@ -0,0 +1,202 @@
...
+        let range = self.range.intersection(&other.range);
+        Self { range, digests }
+    }
+
+    fn contains(&self, v: &Self::V) -> bool {
+        if let Some(digest) = v.digest.as_ref() {
```
huitseeker, line 205: `contains` checks only digest when `v` has one. That skips `self.range.contains(&v.version)`, so a version with a matching digest can pass even when semver is outside the allowed range. Should this require both range and digest checks?

crates/project/src/dependencies.rs
```diff
@@ -0,0 +1,211 @@
...
+        match Self::try_from(spec)? {
+            Self::Path { path: uri, version } => {
+                let workspace_path = workspace
+                    .source_file
+                    .as_ref()
+                    .map(|file| Path::new(file.content().uri().path()));
+                if uri.scheme().is_none_or(|scheme| scheme == "file")
+                    && let Some(workspace_path) = workspace_path.and_then(|p| p.canonicalize().ok())
+                    && let Some(workspace_root) = workspace_path.parent()
+                    && let Ok(resolved_uri) = workspace_root.canonicalize()
```
huitseeker: Is `resolved_uri` computed from `workspace_root.canonicalize()` instead of the dependency path `uri`? 

So `strip_prefix(workspace_root)` always true, so file path dependencies get reclassified as `Workspace` too broadly. 

Did you mean to canonicalize `uri.path()` relative to `workspace_root` and compare that result?

crates/project/src/package.rs
```diff
@@ -0,0 +1,399 @@
...
+        }
+
+        Ok(Box::new(Self {
+            manifest_path,
+            name: package_ast.package.name.clone(),
+            version,
+            description,
+            dependencies,
+            lints: workspace.map(|ws| ws.workspace.config.lints.clone()).unwrap_or_default(),
+            metadata: workspace.map(|ws| ws.workspace.config.lints.clone()).unwrap_or_default(),
```
huitseeker: This looks like a copy/paste bug: `metadata` is initialized from `ws.workspace.config.lints` instead of workspace/package metadata. 

Should this be `ws.workspace.config.metadata`?

crates/project/src/profile.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Returns true if this profile is configured so that we should emit debug information.
+    pub const fn should_emit_debug_info(&self) -> bool {
+        self.debug
+    }
+
+    /// Returns true if this profile is configured so that we should trim file paths in debug
+    /// information to be relative to the current working directory.
+    pub const fn should_trim_paths(&self) -> bool {
+        self.debug
```
huitseeker: `should_trim_paths` returns `self.debug`, not `self.trim_paths`.

crates/project/src/dependencies/version_requirement.rs
```diff
@@ -0,0 +1,402 @@
...
+                    (Unbounded, Unbounded) => (),
+                    (Unbounded, _) => return Ordering::Less,
+                    (_, Unbounded) => return Ordering::Greater,
+                    (Included(l), Included(r)) | (Excluded(l), Excluded(r)) => match l.cmp(&r) {
+                        Ordering::Equal => (),
+                        order => return order,
+                    },
+                    (Included(l), Excluded(r)) => match l.cmp(&r) {
+                        Ordering::Equal | Ordering::Less => return Ordering::Less,
+                        Ordering::Greater => todo!(),
```
huitseeker: Thise `todo!()` branches in `Ord::cmp` look reachable! 😱

crates/project/src/workspace.rs
```diff
@@ -0,0 +1,108 @@
...
+        let mut workspace = Box::new(Workspace {
+            manifest_path: manifest_path.clone(),
+            members: Vec::with_capacity(members.len()),
+        });
+
+        for member in members {
+            let relative_path = Path::new(member.as_str());
+            let manifest_path = workspace
+                .workspace_root()
+                .map(|root| root.join(relative_path).join("miden-project.toml"))
```
huitseeker: A member like `../outside` would resolve outside the workspace root, but I think we want a boundary check here. 

Should we canonicalize the final member manifest path and reject anything not under `workspace_root`?

crates/project/src/dependencies/resolver/provider.rs
```diff
@@ -0,0 +1,665 @@
...
+            PackagePriority::Weak(Reverse(1))
+        }
+    }
+
+    fn choose_version(
+        &self,
+        package: &Self::P,
+        range: &Self::VS,
+    ) -> Result<Option<Self::V>, Self::Err> {
+        std::dbg!(package, range);
```
huitseeker: Do we need this? Internal package/version data printed on stderr in normal runs looks like noise.

crates/project/src/package.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// Load a package from `source`, expected to be a package-level `miden-project.toml` manifest
+    /// which is presumed to be a member of `workspace` for purposes of configuration inheritance.
+    pub fn load_from_workspace(
+        source: Arc<SourceFile>,
+        workspace: &WorkspaceFile,
+    ) -> Result<Box<Self>, Report> {
+        Self::parse(source, Some(workspace))
+    }
+
+    fn parse(
```
huitseeker, line 118: This does a ton of stuff (parse, inherit, normalize, validate) — should it be broken down in small phases with clear i/o that can pass checked invariants from one stage to the next?

crates/project/src/dependencies/resolver/version_set.rs
bitwalker: We don't care about the semantic version range when a digest is specified, as the latter is considered authoritative (i.e. if you request a specific digest of a dependency, then all instances of that dependency in the transitive closure of your dependency graph must be compatible with that exact package, or we will raise a conflict).

crates/project/src/dependencies/version_requirement.rs
bitwalker: Sure, that's why they are `todo!` 😅 - in any case, I'll get those cleaned up before this is merged

crates/project/src/workspace.rs
bitwalker: Sounds good to me 👍

crates/project/src/dependencies/resolver/provider.rs
bitwalker: No, this was unintentionally left in, since I have other work based on this branch

crates/project/src/package.rs
bitwalker: The actual parsing is factored out already, what happens in this function is the meat of actually turning the parsed AST into the normalized and validated `Package` structure. There are a few bits that can be factored out to simplify this function, which I'll do here shortly, but IMO, validation/normalization/inheritance are inherently linked here (i.e. we can't really validate without ensuring we have the complete configuration via inheritance, and we can't inherit without normalizing certain parts of the configuration, such as paths), and I see little benefit in attempting to extricate those into distinct phases, at least with the limited turnaround time I have before this next release. It honestly doesn't matter if this stuff is barely held together with duct tape so long as we have the bones to build on in subsequent releases.

bitwalker:
> It kind of looks like the same concepts are modeled twice, once in AST and once in runtime (`Profile`, `Package`, `Workspace`, `Target`). That creates extra mapping code and makes drift likely across `{ast/,}(profile|workspace)`. Is there a trait-based approach that can keep one canonical model for each concept, and keep parsing types as thin input wrappers only where needed?

This split is intentional, so as to separate the concerns of the human-readable manifest representation (i.e. the AST) from how post-processed and validated manifests are represented.

> There are a few rules spread across `version_requirement.rs`, `resolver/version_set.rs`, and `resolver/provider.rs`. Is there a way to have one source of truth re: those? That would also help with negative tests against those rules, which would help review quite a bit.

I can take another look here and follow up, but certainly if we can share rules around versioning, that would be ideal.

bitwalker:
The failing lint check appears to be due to unrelated code in this PR. I do not know why we are running a lint check against nightly tbh, the errors we get are rarely actionable in the PR which triggers the failing check 🤷

bitwalker:
> There are a few rules spread across version_requirement.rs, resolver/version_set.rs, and resolver/provider.rs. Is there a way to have one source of truth re: those? That would also help with negative tests against those rules, which would help review quite a bit.

I dug into this a bit to see if there were obvious cases of duplication between them, but I'm not finding anything particularly extractable as of yet. Each of those types have subtle semantics that aren't identical, but do have some commonality, just not anything that seems likely to be factorable (that's not to say throwing Claude/Codex at it to see what it comes up with might not find something). If you would like to propose a set of missing tests, I can write those up and at least ensure the existing implementation is sane, and in the process perhaps uncover where such a factorization might be useful or called for.

huitseeker:
> I do not know why we are running a lint check against nightly tbh,

rustfmt option | Tracking issue
-- | --
wrap_comments | https://github.com/rust-lang/rustfmt/issues/3347
imports_granularity | https://github.com/rust-lang/rustfmt/issues/4991
group_imports | https://github.com/rust-lang/rustfmt/issues/5083
condense_wildcard_suffixes | https://github.com/rust-lang/rustfmt/issues/3371
format_code_in_doc_comments | https://github.com/rust-lang/rustfmt/issues/3349
format_macro_matchers | https://github.com/rust-lang/rustfmt/issues/3354
hex_literal_case | https://github.com/rust-lang/rustfmt/issues/5082
match_block_trailing_comma | https://github.com/rust-lang/rustfmt/issues/3380
single_line_if_else_max_width | https://github.com/rust-lang/rustfmt/issues/3357
single_line_let_else_max_width | https://github.com/rust-lang/rustfmt/issues/3357

> The failing lint check appears to be due to unrelated code in this PR
>  the errors we get are rarely actionable in the PR which triggers the failing check

Do you mean the rustfmt check? The second run of `make lint` finishes with an exit code of 0 (i.e. the fixes are idempotent) and the modified files are the following, all introduced in this PR AFAICT.

huitseeker@gringolet.local➜~/tmp/miden-vm-review(bitwalker/project-file✗)» git diff --stat                                                                                   [7:14:41]
 crates/project/src/ast/package.rs  | 6 ++++--
 crates/project/src/dependencies.rs | 3 ++-
 2 files changed, 6 insertions(+), 3 deletions(-)

bitwalker:
> Do you mean the rustfmt check? The second run of `make lint` finishes with an exit code of 0 (i.e. the fixes are idempotent) and the modified files are the following, all introduced in this PR AFAICT.
> 
> [huitseeker@gringolet.local](mailto:huitseeker@gringolet.local)➜~/tmp/miden-vm-review(bitwalker/project-file✗)» git diff --stat [7:14:41] crates/project/src/ast/package.rs | 6 ++++-- crates/project/src/dependencies.rs | 3 ++- 2 files changed, 6 insertions(+), 3 deletions(-)

I cannot reproduce that locally - I run `make lint` locally always before I push. This is the purpose of `rust-toolchain.toml` for anything that isn't an optional check.

EDIT: I should clarify that I cannot reproduce locally using `+nightly`, but since my nightly isn't _literally yesterdays nightly_, that's never guaranteed to succeed.

crates/project/src/dependencies.rs
```diff
@@ -0,0 +1,238 @@
...
+    ) -> Result<Self, InvalidDependencySpecError> {
+        match Self::try_from(spec)? {
+            Self::Path { path: uri, version } => {
+                let workspace_path =
+                    workspace.source_file.as_ref().map(|file| file.content().uri().path());
+                if uri.scheme().is_none_or(|scheme| scheme == "file") &&
+                    let Some(workspace_root) = workspace_path.and_then(|p| p.strip_suffix("miden-project.toml")) &&
+                    uri.path().strip_prefix(workspace_root).is_some() &&
+                    // Make sure the uri is relative to workspace root
+                    (!workspace_root.is_empty() || !uri.path().starts_with('/') || !uri.path().starts_with(".."))
```
huitseeker: I think this guard may mark out-of-workspace paths as `Workspace` in `no_std`. The last check uses `||`, so it passes for most inputs (`!starts_with('/')` is usually true on relative paths, and `!starts_with("..")` is true for absolute paths). 

Would it be safer to require all three conditions with `&&`, or normalize the path before checking membership?

crates/project/src/package.rs
```diff
@@ -0,0 +1,192 @@
...
+        let lib = package_ast.extract_library_target()?;
+        let bins = package_ast.extract_executable_targets();
+
+        Ok(Box::new(Self {
+            manifest_path,
+            name: package_ast.package.name.clone(),
+            version,
+            description,
+            dependencies,
+            lints: workspace.map(|ws| ws.workspace.config.lints.clone()).unwrap_or_default(),
```
huitseeker: It looks like package-local `lints` and `package.metadata` are not merged here, only workspace values are used. Is that intentional? For standalone packages this seems like local config gets dropped.

crates/project/src/dependencies.rs
```diff
@@ -0,0 +1,238 @@
...
+impl TryFrom<Span<&crate::ast::DependencySpec>> for DependencyVersionScheme {
+    type Error = InvalidDependencySpecError;
+
+    fn try_from(ast: Span<&crate::ast::DependencySpec>) -> Result<Self, Self::Error> {
+        if ast.inherits_workspace_version() {
+            return Err(InvalidDependencySpecError::NotAWorkspace { span: ast.span() });
+        }
+
+        if ast.is_host_resolved() {
+            Ok(Self::Registry(ast.version().unwrap().clone()))
```
huitseeker: Could `ast.version().unwrap()` panic on inputs like `foo = {}` (host-resolved dep with no `version`/`digest`)? If so, returning an `InvalidDependencySpecError` here might avoid a hard panic.

crates/project/examples/protocol/kernel/bin/main-alt.masm
huitseeker, line 1: This filename looks incorrect.

crates/project/src/dependencies.rs
bitwalker: This was meant to be:

```rust
(!workspace_root.is_empty() && !(uri.path().starts_with('/') || uri.path().starts_with("..")))
```

In a no-std environment, we don't actually have files anyway, so this is really just attempting to do something reasonable to infer workspace-relative dependencies of a virtual project manifest with respect to a virtual workspace manifest, based on the virtual filenames those manifests carry (if they even have them at all).

I think ideally this bit of code would be `cfg`-d out, but it is retained since it is still possible to parse manifests from strings and not filesystem paths. The only reason to even allow compiling in that configuration is for Wasm targets, but we may be able to require `std` on Wasm targets (via WASI) if we know that satisfies all of our build requirements.

bitwalker:
> That probably justifies different structs, but I think there's some use in having some shared behaviors between those. Here's one example that seemed to make sense to me:
> 
> https://gist.github.com/huitseeker/ae37b40fae6047e264ae3742e9172b25

I'll readily agree that if we can factor out shared behaviors and avoid unnecessary duplication, that's beneficial - but I'm not sure this is the best example of that to demonstrate those benefits - or at least it doesn't seem that way to me. IMO the diff there doesn't seem to meaningfully simplify things in a way that eases maintenance or clarifies readability. All of the original code is still there, but with more boilerplate, and with implementation details spread further afield.

IMO, the place where traits are great for this sort of thing, is when you can reduce boilerplate associated with multiple similar implementations of some behavior, separating the parts that are different from the part(s) that are common. That doesn't really seem to be the case here - but like I said, perhaps it just isn't the best example. I'm certainly open to this sort of thing in general.

> It has the advantage of obviating the issue about merge not inserting new metadata keys, but YMMV.

That's not really a result of the approach though - both versions of the code could exhibit this issue, you've just added the fix for it in your diff.

crates/project/src/ast/package.rs
```diff
@@ -0,0 +1,499 @@
...
+
+        use crate::TargetType;
+
+        let mut bins = Vec::with_capacity(self.bins.len());
+        for target in self.bins.iter() {
+            let span = target.span();
+            let name = target
+                .name
+                .clone()
+                .unwrap_or_else(|| Span::new(target.span(), self.package.name.inner().clone()));
```
huitseeker, line 352: I think there is a mismatch in default bin naming here. This path uses `package.name` when `name` is missing, but validation later uses `$exec` as the default key. 

That can let two unnamed bins pass conflict checks and then collide at extraction time.

crates/project/src/ast.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// caveats:
+    ///
+    /// * If parsing a workspace-level configuration, the workspace members are not checked, so it
+    ///   is up to the caller to iterate over the member paths, and parse/validate their respective
+    ///   configurations.
+    /// * If parsing an individual project configuration which belongs to a workspace, inherited
+    ///   properties from the workspace-level are assumed to exist and be correct. It is up to the
+    ///   caller to compute the concrete property values and validate them at that point.
+    pub fn parse(source: Arc<SourceFile>) -> Result<Self, Report> {
+        if source.as_str().contains("[workspace") {
```
huitseeker: This looks fragile because it only checks `contains("[workspace")`. Could a string value or comment could include that text and change parse mode by accident?

Would it be safer to parse TOML first and detect a real top-level `workspace` table key?

crates/project/src/dependencies.rs
```diff
@@ -0,0 +1,246 @@
...
+                .ok_or(InvalidDependencySpecError::MissingVersion { span: ast.span() })
+        } else if ast.is_git() {
+            let version = match ast.version() {
+                Some(VersionRequirement::Digest(digest)) => {
+                    return Err(InvalidDependencySpecError::GitWithDigest { span: digest.span() });
+                },
+                Some(VersionRequirement::Semantic(v)) => Some(v.clone()),
+                None => None,
+            };
+            let revision = ast
```
huitseeker, line 188: I may be missing intent, but this picks `branch` first when both `branch` and `rev` are set. That silently drops `rev`, which weakens pinning to an exact commit. 

Should invert the priority between branch and rev, or maybe return an error when both are present so the selector stays unambiguous?

crates/project/examples/protocol/utils/miden-project.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "miden-core"
+version = "0.1.0"
+
+[lib]
+namespace = "miden::core"
+path = "mod.masm"
```
bobbinth, line 7: I know this naming is just illustrative, but I'd probably change it to something like `miden-utils` and `miden::utils` to avoid potential confusion with Miden core lib.

Also this made me think: how would we specify the core lib as a dependency for this package? Right now, we are able to "piggy-back" of crates.io and import `miden-corelib` as a Rust dependency. This allows us to both get the `Library` object out of it and also register the required event handles. I guess the approach now would be:

1. We define core lib as a package in this repo, and then we could use git dependencies in this file to import the core lib. A couple of questions on this:
  a. Are git dependencies already supported? (I haven't looked far enough into the code yet)
  b. For git dependencies, would we want to check in assembled packages? Or the source?
  c. How do we specify whether a dependency is imported dynamically or statically? For example, we'd want the core lib to be a dynamic dependency, and local projects to be static dependencies.
2. We'd still need to import `miden-corelib` as a Rust dependency to add the event handlers to the processor during execution.
  a. This would go away in the future once we figure out how to include handler logic into package files - but I think we are some ways off away from this.

crates/project/examples/protocol/kernel/miden-project.toml
```diff
@@ -0,0 +1,18 @@
...
+[[bin]]
+name = "entry"
+path = "bin/main.masm"
+
+[[bin]]
+name = "entry-alt"
+path = "bin/main-alt.masm"
+
+[dependencies]
+miden-core.workspace = true
```
bobbinth: A couple of questions about how dependencies are managed here:

- `miden-core` modules will be statically linked into both the kernel and the executables - right?
- The kernel library would also be statically linked into the executables - right? Would this be as a regular library? Or maybe both regular library and as a kernel?

crates/project/examples/protocol/kernel/miden-project.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "miden-tx"
+version = "1.0.0"
+
+[lib]
+kind = "kernel"
+path = "kernel/mod.masm"
```
bobbinth, line 7: Is the name for the resulting package inferred here from the path? Would we be able to do something like:
```toml
[lib]
kind = "kernel"
name = "tx-kernel"
path = "kernel/mod.masm"
```

crates/project/examples/protocol/userspace/miden-project.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "miden-protocol"
+version = "0.1.0"
+
+[lib]
+namespace = "miden::protocol"
+path = "mod.masm"
+
+[dependencies]
+miden-core.workspace = true
+miden-tx.workspace = true
```
bobbinth, line 13: Here, because `miden-tx` library is defined with `kind = kernel`, we'd automatically infer that this dependency would be set as the kernel for the assembler building `miden-protocol`, right?

And also, I'm assuming if we try to list two kernel libraries as a dependency, we'd get an error - right?

huitseeker:
>> Here's one example that seemed to make sense to me:
>> 
>> https://gist.github.com/huitseeker/ae37b40fae6047e264ae3742e9172b25

> [...] All of the original code is still there, but with more boilerplate, and with implementation details spread further afield.

So, just to explain what I was trying, [merge](https://github.com/0xMiden/miden-vm/pull/2510/changes#diff-342585e41cd98cf9a5475b9c9afb5a2efb48251bc7ab8af3ec189cb3e476def5R114-R137) and [from_ast](https://github.com/0xMiden/miden-vm/pull/2510/changes#diff-342585e41cd98cf9a5475b9c9afb5a2efb48251bc7ab8af3ec189cb3e476def5R79-R107) reimplemented the same override semantics (apply debug override if present, apply trim_paths override if present, merge metadata key-by-key with identical table-merge behavior). Centralizing that logic in one place is a first step, giving `ast::Profile` (which has the same fields) the ability to do that as well is a second. 

>> It has the advantage of obviating the issue about merge not inserting new metadata keys, but YMMV.
>
> That's not really a result of the approach though - both versions of the code could exhibit this issue, you've just added the fix for it in your diff.

 ... in one place.

crates/project/examples/protocol/utils/miden-project.toml
bitwalker: Sure, I can change the name to avoid confusion.

As far as the actual core library goes, there are a few relevant details:

* The stable release of the Miden package would be produced by the build script of the crate, or performed in the release workflow on GitHub. In either case, the resulting Miden package would be appended as an asset to the GitHub release so that it can be founded by `midenup` and other tools that may wish to fetch it from there.
2. For development, one can use either path or git dependencies to refer to assemble the core library from source.
3. Dependencies are presumed to be linked dynamically for now (as they are presumed to be separately published packages). We can trivially support specifying a linkage type for targets though, which would allow opting in to static linking when desired.

For event handling, that ends up being a separate concern from the packaging for now, at least until we switch to a model where the event handlers are compiled to some artifact that gets bundled into the Miden package in a separate section, to be loaded dynamically as-needed. I think the best approach for that is to generate a Wasm binary containing the event handler code, since that ensures event handlers are sandboxed (and avoids a lot of mess that comes with trying to manage dylibs for plugin behavior). Supporting that isn't a big lift though IMO.

crates/project/examples/protocol/kernel/miden-project.toml
bitwalker: > * `miden-core` modules will be statically linked into both the kernel and the executables - right?

No, if we want to support modifying the linkage of dependencies though, that is trivial to do.

> * The kernel library would also be statically linked into the executables - right? Would this be as a regular library? Or maybe both regular library and as a kernel?

Yes, my assembler PR that builds on this always links the library target of a package into any binary targets defined in the same package. In this specific case, the library target is of type `kernel`, so we assemble the kernel first, then link the binary target against that kernel. For non-kernel libraries, the same is true, except the library is treated just like any other statically-linked library.

crates/project/examples/protocol/kernel/miden-project.toml
bitwalker: The library target is always named after the package currently, and only when multiple targets are present (i.e. both a library and binary target, or multiple binary targets) does a name need to be specified for a target, and it is the binary targets that must be explicitly named in that case.

So if you were to assemble this project as a library, you'd get a package named `miden-tx`, containing a kernel library artifact.

If you were to assemble this project via one of the executable targets, you'd get a package named `miden-tx:entry` or `miden-tx:entry-alt`. That naming convention, `package:target` is just one I picked without bikeshedding it much, but we could take any number of approaches here. My main goal was to generally provide a DWIM (do what I mean) interface on the assembler side, i.e. `Assembler::for_project(source_manager, "path/to/miden.project").assemble_target(TargetSelector::Type(TargetType::Kernel))` would build you the kernel library, while `TargetSelector::Name("entry")` would build you the executable target named `entry` in this example. The resulting package would never conflict with the library package if you were to assemble both.

Some of the ergonomics of this are still being polished in that branch though.

crates/project/examples/protocol/userspace/miden-project.toml
bitwalker: Yes on both counts.

crates/project/examples/multi_target/miden-project.toml
bitwalker: @PhilippGackstatter Sorry I missed your question! It doesn't at the moment, but I'm going to add that now, as it is clear we'll need to control linkage in different situations (dynamic-by-default, static opt-in). This would be specified at the dependency level.

crates/project/examples/protocol/kernel/miden-project.toml
bitwalker: Just to follow up, I'm going to add support for specifying dependency linkage, I think that is necessary

crates/project/src/dependencies/resolver/version_set.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn intersection(&self, other: &Self) -> Self {
+        if self.digests.is_empty() && other.digests.is_empty() {
+            return Self::from(self.range.intersection(&other.range));
+        }
+
+        let ldigests = BTreeSet::from_iter(self.digests.iter());
+        let rdigests = BTreeSet::from_iter(other.digests.iter());
+        let digests = ldigests.intersection(&rdigests).map(|d| **d).collect::<SmallVec<[_; _]>>();
```
huitseeker, line 192: I think this intersection logic drops digest constraints when only one side has digests. For example, `digest(D)` intersected with a semantic range ends up with an empty digest set, so the digest pin is lost. 

We should preserve the constrained side's digest filter when the other side is unconstrained, no?

crates/project/src/dependencies/resolver/provider.rs
```diff
@@ -0,0 +1,664 @@
...
+    }
+}
+
+impl From<pubgrub::PubGrubError<PackageResolver<'_>>> for DependencyResolutionError {
+    fn from(error: pubgrub::PubGrubError<PackageResolver<'_>>) -> Self {
+        use pubgrub::PubGrubError;
+
+        match error {
+            PubGrubError::NoSolution(tree) => Self::from(tree),
+            PubGrubError::ErrorRetrievingDependencies { package, version, source: _ } => {
```
huitseeker, line 355: This conversion drops the error detail by binding `source` to `_` and storing `error: String::new()`. Maybe keep `source.to_string()` here so failures explain what actually went wrong?

crates/project/src/dependencies/resolver/index.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// This is essentially a convenience wrapper around [`PackageIndex::insert`] for the common
+    /// case of populating the index with package information loaded from a Miden project.
+    pub fn register(&mut self, package: &Package) {
+        let name = PackageId::from(package.name().into_inner());
+        let version = package.version().into_inner().clone();
+        let version = Version::from(version);
+        let deps = package.dependencies().iter().map(|dep| {
+            let name = PackageId::from(dep.name().clone());
+            let req = match dep.required_version() {
+                None => VersionSet::full(),
```
huitseeker, line 51: This maps dependencies with no explicit version to `VersionSet::full()`. For workspace path dependencies, `required_version()` returns `None`, so those become unconstrained and may resolve to any registry version with the same package name. Is that intended behavior, or should we pin to the workspace version?

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+# * `kernel`
+# * `account-component`
+# * `note`
+# * `tx-script`
+kind = "library"
+# The relative path to the root module, if the project is written in Miden 
+# Assembly. Other languages, such as Rust, will omit this entirely.
+path = "mod.masm" 
+# The root namespace of modules parsed for this target
+namespace = "name"
```
bobbinth, line 100: This is not relevant for `kernel` kind - right? (I think the root namespace there is always `$kernel`).

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+# in this file, also known as the _default target_.
+# 
+# Only one `[lib]` section can be present per-project.
+[lib]
+# The type of artifact we're producing
+# 
+# For `[lib]` the default kind is `library`, but other valid library kinds are:
+# 
+# * `kernel`
+# * `account-component`
```
bobbinth, line 92: As mentioned in some other discussions, `account-component` is a bit verbose, and I'd love to come up with a single word that describes this concept. But I don't have any suggestions on this yet.

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+```toml
+[dependencies]
+# A semantic version constraint
+a = "=0.1.0"
+# A specific package, given by its content digest
+b = "0x......"
+# A path dependency
+c = { path = "../c" }
+d = { path = "../c", version = "~> 0.1.0" }
```
bobbinth, line 154: Could path dependencies point to a `.masp` file - or do they always have to point to a directory? If the latter, how do we resolve which package to get from that directory? (e.g., in case there are multiple `.masp` files in that directory or if there is a `.masp` file and a `miden-project.toml` file in the same directory).

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+# A git dependency
+e = { git = "https://github.com/example/e", branch = "main" }
+f = { git = "https://github.com/example/f", rev = "deadbeef" }
+g = { git = "https://github.com/example/g", rev = "deadbeef", version = "~> 0.1.0" }
```
bobbinth, line 158: Similar question as above but for git dependencies - i.e., how do we fine the package we are interested in given the git path?

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+* `account-component` - produce a package which is a valid account component in the Miden protocol, and contains all metadata needed to construct that component. This type is only valid in conjunction with the Miden transaction kernel.
+* `note` - produce a package which is a valid note script in the Miden protocol, and exports the necessary metadata and procedures to construct and execute the note. This type is only valid in conjunction with the Miden transaction kernel.
+* `tx-script` - produce a package which is a valid transaction script in the Miden protocol, and exports the necessary metadata and procedures to construct and execute the script. This type is only valid in conjunction with the Miden transaction kernel.
+
+As noted earlier, you may define multiple targets in a single Miden project - however you must then request a specific target when assembling the project. Additionally, all targets in a project share the same dependency set.
+
+### Dependency management
```
bobbinth, line 140: It may also be good to explain static vs. dynamic linking for the dependencies.

crates/project/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+TODO
+
+### Custom package metadata
+
+TODO
+
+### Lint configuration
+
+TODO
```
bobbinth: Are these capabilities already implemented in this PR? If not, I'd remove the sections and add them in once we have the capability.

crates/project/examples/protocol/userspace/miden-project.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "miden-protocol"
+version = "0.1.0"
+
+[lib]
+namespace = "miden::protocol"
+path = "mod.masm"
+
+[dependencies]
+miden-utils = { workspace = true, linkage = "dynamic" }
+miden-tx.workspace = true
```
bobbinth, line 13: I would add a comment here explaining that because `miden-tx` is a kernel, this will be automatically linked as a kernel library (i.e., procedures exported from the kernel could be invoked via `syscall` instruction).

crates/project/examples/package/miden-project.toml
```diff
@@ -0,0 +1,19 @@
...
```
partylikeits1983, line 19: It might not be necessary to define the client endpoint in the `miden-project.toml`. From my experience and from what I've seen, its pretty straightforward to instantiate the client with a given endpoint.

crates/project/examples/protocol/miden-project.toml
```diff
@@ -0,0 +1,13 @@
+[workspace]
+members = [
+    "utils",
+    "kernel",
+    "userspace",
```
partylikeits1983, line 5: Say I have a bunch of different note scripts. Would each note script live in its own crate inside of `userspace`? Or could they be in the same directory?

crates/project/src/dependencies/version.rs
```diff
@@ -0,0 +1,149 @@
...
+            write!(f, "{}@{}", &self.version, digest.inner())
+        } else {
+            fmt::Display::fmt(&self.version, f)
+        }
+    }
+}
+
+impl Eq for Version {}
+impl PartialEq for Version {
+    fn eq(&self, other: &Self) -> bool {
```
huitseeker, line 116: I think `PartialEq` here can break transitivity: `1.0.0 == 1.0.0@A` and `1.0.0 == 1.0.0@B`, but `1.0.0@A != 1.0.0@B`. That can make `BTreeMap<Version, ...>` key behavior unstable in the index/resolver path.

crates/project/src/dependencies/version_requirement.rs
```diff
@@ -0,0 +1,93 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::*;
+#[cfg(feature = "serde")]
+use crate::ast::parsing::SetSourceId;
+use crate::{LexicographicWord, Word};
+
+/// Represents a requirement on a specific version (or versions) of a dependency.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
huitseeker, line 16: I think this enum deserialize shape might block the README dependency forms like `a = "=0.1.0"` and `b = "0x..."`. 
With derived enum serde, a plain string usually won't deserialize into `VersionRequirement` unless it's untagged. Would it make sense to use `#[serde(untagged)]` (or a custom deserialize) here?

crates/project/src/package.rs
```diff
@@ -0,0 +1,199 @@
...
+        };
+
+        // Parse the manifest into an AST for further processing
+        let package_ast = ast::ProjectFile::parse(source.clone())?;
+
+        // Extract metadata that can be inherited from the workspace manifest (if present)
+        let version = package_ast.get_or_inherit_version(source.clone(), workspace)?;
+        let description = package_ast.get_or_inherit_description(source.clone(), workspace)?;
+
+        // Compute the set of initial profiles inheritable from the workspace level
```
huitseeker, line 136: Workspace profiles are appended after builtin `dev`/`release`, but later lookup uses first-match search. 

If a workspace defines `dev` or `release`, package inheritance looks like it can still bind to the builtin profile instead of the workspace override. 

We probably want to have workspace profiles replace/merge matching builtin names before package profile resolution?

crates/project/src/workspace.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Represents a Miden project workspace.
+///
+/// Workspaces are comprised of one or more sub-projects that define the member packages of the
+/// workspace.
+#[derive(Debug)]
+pub struct Workspace {
+    /// The file path of the workspace manifest, if applicable.
+    #[cfg(feature = "std")]
+    manifest_path: Option<Box<Path>>,
```
bobbinth, line 17: The manifest path is the absolute path to the workspace's `miden-project.toml` file, right?

crates/project/src/workspace.rs
```diff
@@ -0,0 +1,132 @@
...
+                return Err(ProjectFileError::LoadWorkspaceMemberFailed {
+                    source_file: source.clone(),
+                    span: Label::new(
+                        member.span(),
+                        "workspace members must be located within the workspace root",
+                    ),
+                }
+                .into());
+            }
+            let manifest_path = member_dir.join("miden-project.toml");
```
bobbinth, line 119: nit: would be good to create something like a `MIDEN_PROJECT_FILE` constant use it in places like this instead of string literals.

crates/project/README.md
bitwalker: Correct

crates/project/README.md
bitwalker: We can easily change this in a backwards-compatible way once we've come up with a short form we like, and then remove after a deprecation cycle.

crates/project/README.md
bitwalker: See #2877 for the implementation of this - but to answer your questions:

1. Yes the path can point to a `.masp` file
2. The path can be either a directory (in which case we append `miden-project.toml`) or point to the `miden-project.toml` directly. In these cases we are assembling the referenced package on demand, if it is not already available in the package registry (i.e. already assembled)
3. If the path points to a workspace, not a member of the workspace, then we resolve the member by package name, and raise an error if no such member can be found.

crates/project/README.md
bitwalker: Same rules as `path` dependencies, though #2877 is missing one small tweak I meant to add (that this reminded me of), to allow providing an optional `subpath` key that can be used to specify a subdirectory of the checkout (or a file within the checkout) - whether that is provided or not though, the rules for selecting packages follows that of `path` dependencies.

crates/project/README.md
bitwalker: Good point, we cover this in the assembler, but not here.

crates/project/README.md
bitwalker: They are, but we haven't defined any specific lint configuration keys yet, so we can just remove this section for now

bitwalker:
> One slight question / concern is handling nested directories of masm files, and how to define them in the `miden-project.toml` file.
> 
> For example, in the `miden-agglayer` crate in the `protocol` repository, we have several sub-directories. How would something like this need to be defined in the `miden-project.toml` file?

The directory hierarchy maps to the namespacing hierarchy of the resulting package, so if you wanted it all in a single package, you'd define the following in `protocol/crates/miden-agglayer/asm/miden-project.toml`:

```toml
[package]
name = "miden-agglayer"
version = "0.1.0"

[lib]
path = "mod.masm"
namespace = "miden::agglayer"
```

You don't currently have a `mod.masm`, you need one, but it can just be empty. This module corresponds to the root of the namespace, i.e. `miden::agglayer`.

This maps the directory structure and filenames of the modules contained in the same directory as `mod.masm` to the namespace hierarchy of the Miden Assembly modules, e.g. `protocol/crates/miden-agglayer/asm/agglayer/bridge/bridge_out.masm` would map to `miden::agglayer::bridge::bridge_out`.

It is recommended to add a `mod.masm` in each sub-directory, so that there is a module corresponding to that namespace, but it isn't required.

Now, if you wanted each subdirectory to be it's own sub-project, and to manage `protocol/crates/miden-agglayer/asm` as a workspace, then you'd do basically what you sketched out, i.e. define a workspace manifest at the root, and then in each sub-directory corresponding to a sub-project, define another `miden-project.toml` containing the manifest for that project.


> Related to [this comment](https://github.com/0xMiden/miden-vm/pull/2510/changes#r2945185505), we don't ever instantiate the `Client` in the protocol repo, so it might be a bit strange to have the optionality to set `package.metadata.network.local` in `miden-project.toml` inside the protocol repo. 

To be clear, package metadata is arbitrary key/value data, and that was just an example of the kind of arbitrary metadata that one might want to specify for a project. This sort of metadata is commonly used in Rust projects for third-party tooling. I also wouldn't expect the protocol repo to specify that sort of metadata, but end-user projects which want to configure the defaults for client profiles they use - but it is up to the client to determine whether to use project metadata for this or not.

> Separately, I think the endpoint could be set in a `.env` file. There are other pieces of data that would need to go into a `.env` file in a miden project such as a seed phrase for the authentication component.

A `.env` is more appropriate for secrets, not static configuration like endpoints - but regardless, how project metadata gets used is entirely up to downstream tools, and not the purview of `miden-project` itself.

crates/project/examples/package/miden-project.toml
bitwalker: This is just intended as an example of the sort of metadata that downstream tools can build on, not something we've explicitly planned on AFAIK - the `miden-client` team can determine to what extent, if any, they choose to use project metadata for that sort of thing

crates/project/examples/protocol/miden-project.toml
bitwalker: You could do either, depending on whether you want to build each note as its own package, or whether you want a single package with all the notes, or something in-between. I think it'll take some time to find best practices around this - but I think it is somewhat context-sensitive.

crates/project/src/dependencies/version_requirement.rs
bitwalker: The serialization of this type changed in #2877, so this is no longer relevant

crates/project/src/workspace.rs
bitwalker: Correct

crates/project/src/dependencies/version.rs
bitwalker: I'm going to fix this in #2877 instead, as these types have moved since then

crates/project/src/package.rs
bitwalker: Likewise, I'll address this in #2877

crates/project/src/workspace.rs
bitwalker: I'll address this in #2877

crates/project/README.md
bobbinth: Would be good to mention this in the description somewhere (in the follow-up PR).

crates/project/README.md
bobbinth: Not for this PR, but I wonder if we should flip the default - i.e., make static linking the default.

PR #2524, opened 2025-12-31
fix(core): strip decorators while maintaining valid CSR structure

huitseeker:
Fixes the `MastForest::strip_decorators()` method to properly strip
decorator information while maintaining valid CSR (Compressed Sparse Row)
structure invariants.

The previous implementation called `DebugInfo::clear()` which removed
the entire CSR structure. However, code paths that serialize or access
decorators expect the CSR structure to exist (even if empty), causing
panics when accessing decorator information after stripping.

1. Added `DebugInfo::empty_for_nodes(num_nodes)` method that creates
   an empty but valid CSR structure for a given number of nodes
2. Updated `strip_decorators()` to use this new method instead of `clear()`
3. Added early returns to `OpToDecoratorIds::from_components()` to support
   empty CSR structures (nodes with no decorators)
4. Changed `from_components` visibility from `#[cfg(test)]` to `pub(crate)`
   to allow production use

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [BREAKING] Removed `Process`, `VmStateIterator` and `miden_processor::execute_iter()` ([#2483](https://github.com/0xMiden/miden-vm/pull/2483)).
 - [BREAKING] Removed `miden debug`, `miden analyze` and `miden repl` ([#2483](https://github.com/0xMiden/miden-vm/pull/2483)).
 - [BREAKING] Change backend from winterfell to Plonky3 ([#2472](https://github.com/0xMiden/miden-vm/pull/2472)).
+- Fix issue where calling `strip_decorators()` would panic on further decorator access ([#2524](https://github.com/0xMiden/miden-vm/pull/2524)).
 
 ## 0.20.1 (2025-12-14)
 


core/src/mast/debuginfo/mod.rs
```diff
@@ -109,6 +109,28 @@ impl DebugInfo {
...
 
+    /// Creates an empty [DebugInfo] with valid CSR structures for N nodes.
+    ///
+    /// Used when stripping decorators while keeping CSR structure valid.
+    pub fn empty_for_nodes(num_nodes: usize) -> Self {
+        // For N nodes, CSR needs N+1 zero entries
+        let mut node_indptr_for_op_idx = IndexVec::new();
+        for _ in 0..=num_nodes {
+            let _ = node_indptr_for_op_idx.push(0);
+        }
```
adr1anh, line 120: Should we just initialize the vectors with N+1 zeros, instead of pushing them one by one.

core/src/mast/mod.rs
```diff
@@ -163,24 +163,11 @@ impl MastForest {
...
-    /// let mut forest = MastForest::new();
-    /// // Add decorators and nodes to the forest
-    /// forest.strip_decorators(); // forest is now stripped
-    /// ```
+    /// Useful for performance benchmarking without decorator overhead.
     pub fn strip_decorators(&mut self) {
-        // Clear all debug info (decorators and error codes)
-        self.debug_info.clear();
+        self.debug_info = DebugInfo::empty_for_nodes(self.nodes.len());
     }
```
bobbinth, line 171: Maybe not for this PR, but we should probably rename this into `clear_debug_info()` or something like that because, AFAICT, this is also removing error messages and in the future other things as well (e.g., see https://github.com/0xMiden/miden-vm/pull/2474).

Also, I still don't fully understand why we need to preserve any info in the debug info upon clearing it. The PR description mentions some code paths that rely on this - what are these code paths? Basically, is is something we can change after https://github.com/0xMiden/miden-vm/pull/2470 is merged?

I think the ideal end state would be to have `debug_info` in `MastForest` be an optional field (i.e., `debug_info: Option<DebugInfo>`) - what is preventing us from doing this?

huitseeker:
Am going to propose a better fix in main : #2529

PR #2528, opened 2026-01-04
feat(air): add validation of core_trace_fragment_size

### LINKED ISSUE #2523: Add validation of `core_trace_fragment_size` in `ExecutionOptions::new()`
Not for this PR, but we should also validate `core_trace_fragment_size`.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2472#discussion_r2655231021_

Himess:
Adds input validation for `core_trace_fragment_size` in `ExecutionOptions::new()` and `with_core_trace_fragment_size()`.

**Validation rules:**
- Must be greater than zero
- Must be a power of two

Closes #2523

### DIFF
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -66,6 +66,21 @@ runs:
         echo "Cleaning target/package directory to ensure fresh state"
         rm -rf target/package
 
+    # Clear cargo registry to ensure fresh resolution during verification.
+    # This prevents issues where cached metadata from previous runs
+    # might interfere with workspace dependency feature resolution
+    # during the verification step (related to cargo#14283, cargo#14789).
+    # Specifically, this ensures the temp registry used during workspace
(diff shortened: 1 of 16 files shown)


air/src/options.rs
huitseeker, line 1: The docstring for `ExecutionOptions::new()` should mention that `core_trace_fragment_size` must be greater than zero and a power of two, since these constraints are enforced at runtime.

Something like:
```rust
/// Creates a new instance of [ExecutionOptions] from the specified parameters.
///
/// If the `max_cycles` is `None` the maximum number of cycles will be set to 2^29.
/// 
/// # Errors
/// Returns an error if:
/// - `max_cycles` is outside the valid range
/// - `expected_cycles` exceeds `max_cycles`
/// - `core_trace_fragment_size` is zero or not a power of two
```

Himess: Added the `# Errors` section as requested.

bobbinth (approved):
Looks good! Thank you!

PR #2529, opened 2026-01-04
fix: bypass decorator retrieval in release mode

huitseeker:
This replaces #2524.

This fixes 4 issues in both legacy and fast processor, see individual commits:
1. when `in_debug_mode == false`, decorators should not be _accessed_, never mind executed,
2. when `strip_decorators` is false, the `DebugInfo` provided as replacement should not panic on access,
3. ~~adds a `--strip-decorators` CLI option to the `prove` and `run` commands which strips decorators from the `MastForest` before operation,~~
4. the `err_ctx!` macro should respect `in_debug_mode`, that is it should not perform decorator access in debug mode.

Even without stripping decorators, the combination of 1. and 4. yields a substantive performance improvement:

Baseline (main branch)

  - Total Time: 6906 ms
  - Execution Time: 4015 ms (58.14%)
  - Trace Generation: 4015 ms for 1,048,576 steps (48% padded)

  Optimized (bypass-decorator-retrieval-in-release-mode branch)

  - Total Time: 3533 ms
  - Execution Time: 289 ms (8.17%)
  - Trace Generation: 288 ms for 1,048,576 steps (48% padded)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.20.2 (TBD)
+- Fix issue where decorator access was not bypassed properly in release mode ([#2529](https://github.com/0xMiden/miden-vm/pull/2529)).
+
 ## 0.20.1 (2025-12-14)
 
 - Fix issue where calling procedures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).


miden-vm/src/cli/prove.rs
```diff
@@ -62,6 +62,10 @@ pub struct ProveCmd {
     /// Disable debug instructions (release mode)
     #[arg(long = "release")]
     release: bool,
+
+    /// Strip decorators from the program before proving
+    #[arg(long = "strip-decorators")]
+    strip_decorators: bool,
```
bobbinth: I think it is fine to add this option, but also, I'm not sure using it would be all that meaningful in many cases. For example, for our BLAKE3 benchmark, the actual BLAKE3 code is in the `corelib`, and so, stripping decorators from the program itself wouldn't really make that much of a difference because the `corelib` would still be loaded with decorators.

miden-vm/src/cli/run.rs
```diff
@@ -50,6 +50,10 @@ pub struct RunCmd {
     /// Disable debug instructions (release mode)
     #[arg(short = 'r', long = "release")]
     release: bool,
+
+    /// Strip decorators from the program before running
+    #[arg(long = "strip-decorators")]
+    strip_decorators: bool,
```
bobbinth: Same comment as above.

processor/src/fast/basic_block.rs
```diff
@@ -138,16 +142,22 @@ impl FastProcessor {
         let mut group_idx = 0;
         let mut next_group_idx = 1;
 
+        // Get the node ID once since it doesn't change within the loop
+        let node_id = basic_block
+            .linked_id()
+            .expect("basic block node should be linked when executing operations");
```
bobbinth, line 148: Not for this PR, but getting node ID from the node in this way feels a bit backwards to me. When we apply these changes to `next`, we could probably just pass in `node_id` instead of `back_block` into this function (and in general, we should think how to simplify the number of parameters passed into this function). This will require refactoring how we build error context - but it should be relatively easy (e.g., instead of `node.get_assembly_op(mast_forest, target_op_idx)` we should be able to do something like `mast.get_assembly_op(node_id, target_op_idx)`.

miden-vm/src/cli/prove.rs
huitseeker: Commit ac49cd22ac4bb0a4c218c943261ce730d48cba10 is the least essential of the PR, I'll remove it.

miden-vm/src/cli/run.rs
huitseeker: same answer

processor/src/fast/basic_block.rs
huitseeker: You're right, but a couple of nuances:
- obviously, this is just moving a local variable definition out of a hot loop, not changing the definition (in a PR to main),
- there's upcoming work on removing err_ctx as part of #1978 that may make this obsolete.

core/src/mast/debuginfo/mod.rs
```diff
@@ -109,6 +109,25 @@ impl DebugInfo {
         }
     }
 
+    /// Creates an empty [DebugInfo] with valid CSR structures for N nodes.
+    pub fn empty_for_nodes(num_nodes: usize) -> Self {
+        let mut node_indptr_for_op_idx = IndexVec::new();
+        for _ in 0..=num_nodes {
+            let _ = node_indptr_for_op_idx.push(0);
+        }
```
adr1anh, line 117: Can we initialize the vector with zeros directly, rather than pushing one by one?

huitseeker:
@bobbinth 
Re: confusion between trace generation and execution time, here are sequential runs rebased on next:
```
Baseline (origin/next) - 3 Runs

  | Run     | execute_for_trace | build_trace | Total Time |
  |---------|-------------------|-------------|------------|
  | 1       | 3.58s             | 70.9ms      | 34,991ms   |
  | 2       | 3.57s             | 73.8ms      | 35,100ms   |
  | 3       | 3.56s             | 76.2ms      | 35,001ms   |
  | Average | 3.57s             | 73.6ms      | 35,031ms   |

  Decorator Bypass (decorator-bypass-on-next) - 3 Runs

  | Run     | execute_for_trace | build_trace | Total Time |
  |---------|-------------------|-------------|------------|
  | 1       | 10.7ms            | 86.0ms      | 31,428ms   |
  | 2       | 7.34ms            | 78.9ms      | 31,688ms   |
  | 3       | 11.3ms            | 83.4ms      | 31,857ms   |
  | Average | 9.78ms            | 82.8ms      | 31,658ms   |
  ```

core/src/mast/debuginfo/mod.rs
huitseeker: Not an API that's available on `IndexVec`. Should be fixed, just not on a PR to main. #2532

bobbinth:
> Re: confusion between trace generation and execution time, here are sequential runs rebased on next:
> 
> ```
> Baseline (origin/next) - 3 Runs
> 
>   | Run     | execute_for_trace | build_trace | Total Time |
>   |---------|-------------------|-------------|------------|
>   | 1       | 3.58s             | 70.9ms      | 34,991ms   |
>   | 2       | 3.57s             | 73.8ms      | 35,100ms   |
>   | 3       | 3.56s             | 76.2ms      | 35,001ms   |
>   | Average | 3.57s             | 73.6ms      | 35,031ms   |
> 
>   Decorator Bypass (decorator-bypass-on-next) - 3 Runs
> 
>   | Run     | execute_for_trace | build_trace | Total Time |
>   |---------|-------------------|-------------|------------|
>   | 1       | 10.7ms            | 86.0ms      | 31,428ms   |
>   | 2       | 7.34ms            | 78.9ms      | 31,688ms   |
>   | 3       | 11.3ms            | 83.4ms      | 31,857ms   |
>   | Average | 9.78ms            | 82.8ms      | 31,658ms   |
> ```

Nice! Thank you for running these!

So, it seems like our current execution runs roughly at 100 MHz, right? I think that's pretty nice, and in the future we could probably try to optimize it to get closer to 200 MHz.

Re-trace generation, it should be able take advantage of multi-threading - so, if we run with concurrent feature, I'd expect it to be closer to 20ms range.

PR #2530, opened 2026-01-04
ci: clear cargo registry cache before workspace publish dry-run

huitseeker:
The workspace publish dry-run fails with `unresolved import miden_utils_sync::OnceLockCompat` during verification. This happens because cached registry data from previous CI runs can interfere with cargo's temp registry used for workspace dependency resolution.

During verification, cargo should use the packaged v0.21.0 from the temp registry, but stale cache entries may cause it to resolve against crates.io data instead (where v0.20.1 lacks `OnceLockCompat`).

=> Clear `~/.cargo/registry/{cache,index,src}` before the dry-run to ensure fresh resolution without interference from cached data.

Replaces #2505

### DIFF
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -66,6 +66,21 @@ runs:
         echo "Cleaning target/package directory to ensure fresh state"
         rm -rf target/package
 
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #2533, opened 2026-01-05
feat: Allow constants as repeat iteration count

### LINKED ISSUE #2502: Allow constants as parameters to the `repeat`
### Feature description

Currently `repeat` can take only `u64`, but it will be very convenient to allow using constants as well, like so:

```
const TREE_HEIGHT = 32

repeat.TREE_HEIGHT
    ...
end
```

This feature will increase the readability of the code, making it possible to reduce the number of magic numbers in there.

gejeduck:
Support using constants as iteration counts in `repeat`:

```masm
const TREE_HEIGHT = 32
repeat.TREE_HEIGHT
    ...
end
```

Closes #2502

<details><summary>Original description</summary>

### Summary
Adds support for dynamic iteration counts in the repeat operator.

### Changes
- Updated grammar.lalrpop, op.rs, and assembler.rs
- Added a new assembly test: repeat_dynamic_iteration_count in assembly/src/tests.rs

### Related Issue
#2502

### Issue Description

As part of this change, the iteration count type was updated from u32 to Immediate<U32>.
This works correctly when a literal value is used:

`begin repeat.5 add end end`

However, it fails when the iteration count is provided via a constant:

`const A = 5 begin repeat.A add end end`

In this case, assembly fails with the following error:

`tried to unwrap unresolved constant: 'A'`

The error occurs in assembler.rs when calling Immediate<T>::expect_value.
Other code paths that use the same function (for example in crates/assembly/src/instruction/mod.rs:118) resolve constants correctly, which makes this behavior unclear.

### Question

Could you please clarify why Immediate::expect_value resolves literal values correctly in this context but fails for constants?
Is there an additional resolution step required for constants when used as repeat iteration counts?

</details>

### DIFF
--- a/crates/assembly-syntax/src/parser/grammar.lalrpop
+++ b/crates/assembly-syntax/src/parser/grammar.lalrpop
@@ -948,15 +948,8 @@ While: Op = {
 }
 
 Repeat: Op = {
-  <l:@L> "repeat" <imm:MaybeImm<U32>> <body:Block> "end" <r:@R> =>? {
-    let span = span!(source_id, l, r);
-
-    match imm {
-      Some(imm) => Ok(Op::Repeat { span, count: imm, body }),
-      None => Err(ParseError::User {
-        error: ParsingError::ImmediateOutOfRange { span, range: 1..(u32::MAX as usize) },
-      }),
-    }
+  <l:@L> "repeat" <imm:Imm<U32>> <body:Block> "end" <r:@R> => {
+    Op::Repeat { span: span!(source_id, l, r), count: imm, body }
   }
 }
 
--- a/crates/assembly/src/assembler.rs
+++ b/crates/assembly/src/assembler.rs
@@ -972,10 +972,9 @@ impl Assembler {
                         None,
                         block_builder.mast_forest_builder_mut(),
                     )?;
-                    let value = (*count).expect_value();
 
-                    // I will add validation code for `value` later...
-                    
+                    let iteration_count = (*count).expect_value();
+
                     if let Some(decorator_ids) = block_builder.drain_decorators() {
                         // Attach the decorators before the first instance of the repeated node
                         let first_repeat_builder = block_builder.mast_forest_builder()
@@ -988,11 +987,11 @@ impl Assembler {
                             .ensure_node(first_repeat_builder)?;
 
                         body_node_ids.push(first_repeat_node_id);
-                        for _ in 0..(value - 1) {
+                        for _ in 0..(iteration_count - 1) {
                             body_node_ids.push(repeat_node_id);
                         }
                     } else {
-                        for _ in 0..value {
+                        for _ in 0..iteration_count {
                             body_node_ids.push(repeat_node_id);
                         }
                     }
--- a/crates/assembly/src/tests.rs
+++ b/crates/assembly/src/tests.rs
@@ -162,7 +162,6 @@ fn repeat_dynamic_iteration_count() -> TestResult {
     let source = source_file!(&context, "const A = 5 begin repeat.A add end end");
     let program = context.assemble(source)?;
     insta::assert_snapshot!(program);
-
     Ok(())
 }
 
--- a/crates/assembly-syntax/src/ast/visit.rs
+++ b/crates/assembly-syntax/src/ast/visit.rs
@@ -1012,8 +1012,12 @@ where
             visitor.visit_mut_block(then_blk)?;
             visitor.visit_mut_block(else_blk)
         },
-        Op::While { body, .. } | Op::Repeat { body, .. } => visitor.visit_mut_block(body),
+        Op::While { body, .. } => visitor.visit_mut_block(body),
         Op::Inst(inst) => visitor.visit_mut_inst(inst),
+        Op::Repeat { count,   body, .. } => {
+            visitor.visit_mut_immediate_u32(count)?;
+            visitor.visit_mut_block(body)
+        }
     }
 }
 


Fumuran:
Hey @gejeduck! Thank you for the PR!

I didn't (yet) dive too deeply into our code, but it seems that the constant used in the `repeat` operation is just not get evaluated. I suppose that it should be done somewhere [here](https://github.com/0xMiden/miden-vm/blob/674bd275e778eb90cd3328409647cac50460b382/crates/assembly-syntax/src/ast/visit.rs#L1006), in the `visit_mut_op` function. 

Something like this:
```Rust
Op::Repeat { count, body , ..} => {
    visitor.visit_mut_immediate_u32(count)?;
    visitor.visit_mut_block(body)
}
```

At least this code makes the test pass. But @bitwalker will be able to explain the issue in more detail.

Also, I'm not sure that we should use the `MaybeImm<U32>` in the `grammar.lalrpop`, I think we should use just `Imm<U32>` instead, since we should always expect either a value or a constant name there.

crates/assembly-syntax/src/parser/grammar.lalrpop
```diff
@@ -948,16 +948,16 @@ While: Op = {
...
-    <l:@L> "repeat" "." <count:uint> <body:Block> "end" <r:@R> =>? {
-        let span = span!(source_id, l, r);
-        u32::try_from(count)
-            .map_err(|error| ParseError::User { error: ParsingError::ImmediateOutOfRange { span, range: 1..(u32::MAX as usize) } })
-            .and_then(|count| {
-                NonZeroU32::try_from(count)
-                    .map(|count| Op::Repeat { span, count: count.get(), body })
-                    .map_err(|error| ParseError::User { error: ParsingError::ImmediateOutOfRange { span, range: 1..(u32::MAX as usize) } })
-            })
+  <l:@L> "repeat" <imm:MaybeImm<U32>> <body:Block> "end" <r:@R> =>? {
```
bitwalker: As @Fumuran pointed out, this should be `Imm<U32>` not `MaybeImm<U32>`

crates/assembly/src/assembler.rs
```diff
@@ -972,7 +972,10 @@ impl Assembler {
                         None,
                         block_builder.mast_forest_builder_mut(),
                     )?;
+                    let value = (*count).expect_value();
 
+                    // I will add validation code for `value` later...
```
bitwalker: What validation are you referring to here?

bitwalker (changes requested):
This PR is AI-generated so far as I can tell, including the description. ~It would have been better to ask any clarifying questions in the referenced issue before opening this PR, rather than throwing it over the fence to us~ - but setting that aside:

As @Fumuran mentioned, you need to update the default implementations of the `Visit` and `VisitMut` trait to ensure that when a `Op::Repeat` is visited, that its iteration count immediate is visited using the `visit_(mut_)_immediate_u32` method of the respective trait. This will ensure that the const evaluation passes that run during semantic analysis and link-time visit and rewrite the iteration count now that it might hold a non-literal value.

I've left other comments on other items to address before this can be merged.

crates/assembly/src/assembler.rs
gejeduck: @bitwalker , Thanks for your help and you’re absolutely right. I used an AI tool to reformat the code, and I was planning to add range validation for the value afterward. I’ll review it again and fix the issue shortly.
Thanks again for pointing this out.

gejeduck:
@Fumuran , @bitwalker 
Thanks for your help. I’ve successfully fixed the issue. Could you please check now?

crates/assembly-syntax/src/ast/op.rs
```diff
@@ -29,9 +31,10 @@ pub enum Op {
     /// `for`-style loops where the iteration count is dynamic.
     Repeat {
         span: SourceSpan,
-        count: u32,
+        count: Immediate<u32>,
```
Fumuran, line 34: nit: let's add a brief comment in the doc above that this counter now can be either a value, or a constant name, e.g. `add.1` or `add.CONST`

crates/assembly-syntax/src/ast/op.rs
```diff
@@ -29,9 +31,10 @@ pub enum Op {
...
         body: Block,
     } = 2,
+
```
Fumuran, line 37: nit: let's remove this empty line to preserve style consistency

crates/assembly/src/tests.rs
```diff
@@ -155,6 +155,16 @@ fn repeat_basic_blocks_merged() -> TestResult {
     Ok(())
 }
 
+/// Ensures `repeat` supports dynamic iteration counts provided via constants.
+#[test]
+fn repeat_dynamic_iteration_count() -> TestResult {
```
Fumuran, line 160: Let's add a test which will assert the correct error message in case the provided count value is greater than `u32::MAX`. You can use the `const_conversion_failed_to_u16` test as an example of how the error message could be asserted.

crates/assembly-syntax/src/ast/op.rs
```diff
@@ -2,6 +2,8 @@ use core::fmt;
 
 use miden_debug_types::{SourceSpan, Span, Spanned};
 
+use crate::ast::Immediate;
+
```
Fumuran, line 6: nit: I would import it from `super` in the line below

Fumuran (changes requested):
Looks good, thank you! I left just a few comments inline. Once they are addressed, this PR should be ready to merge. 

Also, let's merge this branch with the latest `next`.

PR #2534, opened 2026-01-05
Merge main into next

huitseeker:
This PR merges `main` into `next`, bringing the decorator bypass optimization from #2529 into the next development branch.

## What's Being Merged

The decorator bypass optimization that was merged to main via #2529, which includes:

1. Gate all decorator retrieval behind `in_debug_mode` checks
2. Prevent expensive decorator traversal when not in debug mode
3. Maintain proper decorator storage structure even when stripped


plafer (approved):
LGTM!

bobbinth:
Ah - I should have merged this w/o squashing. Sorry! Will make a commit into `next` that fixies this.

plafer:
> Ah - I should have merged this w/o squashing. Sorry! Will make a commit into next that fixies this.

~~Given that my PR is stacked on this~~, it's probably fine/easier to let that one as is? If we ever need to revert/bisect, it's still a relatively small commit so should be fine.

PR #2537, opened 2026-01-05
`FastProcessor` now returns an error when cycle count is exceeded

### LINKED ISSUE #2535: Enforce cycle count limit
The `FastProcessor` never enforced a cycle count limit similar to what was done with `Process`. We should add that functionality back.

Related to bullet point in #2517:

> CycleLimitExceeded doesn't seems to be used - are we no longer enforcing upper limits on execution? This is important functionality that is used downstream (e.g., in miden-node).

plafer:
Closes #2535 

As part of this change, we also added a new constructor `FastProcessor::new_with_options()`, which allows users to pass in a `ExecutionOptions`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Added `procedure_names` to `DebugInfo` for storing procedure name mappings by MAST root digest, enabling debuggers to resolve human-readable procedure names during execution (#[2474](https://github.com/0xMiden/miden-vm/pull/2474)).
 
+#### Fixes
+
+- `FastProcessor` now correctly returns an error if the maximum number of cycles was exceeded during execution ([#2537](https://github.com/0xMiden/miden-vm/pull/2537))
+
 #### Changes
 
 - Added `--kernel` flag to CLI commands (`run`, `prove`, `verify`, `debug`) to allow loading custom kernels from `.masm` or `.masp` files ([#2363](https://github.com/0xMiden/miden-vm/pull/2363)).


processor/src/fast/mod.rs
```diff
@@ -132,7 +132,7 @@ pub struct FastProcessor {
     call_stack: Vec<ExecutionContextInfo>,
 
     /// Whether to enable debug statements and tracing.
```
Al-Kindi-0: This should be updated, I believe

plafer: fixed

processor/src/fast/mod.rs
```diff
@@ -147,37 +148,47 @@ impl FastProcessor {
...
     ///
     /// # Panics
     /// - Panics if the length of `stack_inputs` is greater than `MIN_STACK_DEPTH`.
     pub fn new_debug(stack_inputs: &[Felt], advice_inputs: AdviceInputs) -> Self {
-        Self::initialize(stack_inputs, advice_inputs, true)
+        Self::new_with_options(
+            stack_inputs,
+            advice_inputs,
+            ExecutionOptions::default().with_debugging(true).with_tracing(),
+        )
```
plafer, line 179: Note that I enabled tracing with the `new_debug()` to preserve former functionality (mostly in tests).

processor/src/fast/mod.rs
```diff
@@ -147,37 +148,47 @@ impl FastProcessor {
...
     ///
     /// The stack inputs are expected to be stored in reverse order. For example, if `stack_inputs =
     /// [1,2,3]`, then the stack will be initialized as `[3,2,1,0,0,...]`, with `3` being on
     /// top.
-    fn initialize(stack_inputs: &[Felt], advice_inputs: AdviceInputs, in_debug_mode: bool) -> Self {
+    pub fn new_with_options(
+        stack_inputs: &[Felt],
+        advice_inputs: AdviceInputs,
+        options: ExecutionOptions,
+    ) -> Self {
```
bobbinth, line 191: Not for this PR, but we could consider encapsulating stack and advice inputs into a single struct  - e.g., something like `ProgramInputs`.

processor/src/fast/mod.rs
bobbinth: Not for this PR, but would be great to "normalize" `ExecutionOptions` at some point. For example, it is a bit odd that `with_dubbing()` takes a parameter while `with_tracing()` does not.

processor/src/fast/mod.rs
```diff
@@ -230,6 +241,20 @@ impl FastProcessor {
...
+        self.options.enable_debugging()
+    }
+
+    /// Returns true if decorators should be executed.
+    ///
+    /// This corresponds to either being in debug mode (for debug decorators) or having tracing
+    /// enabled (for trace decorators).
+    fn should_execute_decorators(&self) -> bool {
+        self.in_debug_mode() || self.options.enable_tracing()
+    }
```
plafer, line 257: @Al-Kindi-0 I re-requested a review, since I added this method since you approved (which I only noticed was needed when working on #2539). Basically, we should execute decorators not only in debug mode, but also when tracing is enabled.

**This has the unfortunate consequence that the processor will be slowed down when only tracing (but not debug mode) is enabled.**

This is a by-product of the fact that `AsmOp` decorators should not be lumped in with `Debug` and `Trace` decorators, since they do not need to be executed at runtime. Currently, if tracing mode is enabled (or even debug mode just for `debug` decorators), the processor needlessly loops through all `AsmOp` decorators, which really isn't needed - we only need to look them up when an error occurs. `AsmOp` decorators should be stored separately in their own CSR matrix. This is as previously discussed with @huitseeker.

bobbinth, line 257: Should we add `#[inline(always)]` here as well?

Also, not for this PR, but I'm wondering if we should simplify the debug mode - basically, have a single setting that enables both debugging and tracing. I'm not really sure there is a lot of benefit of having them separate.

bobbinth (approved):
Looks good! Thank you! I left just a few small comments inline.

Also, now we are doing a bit more work on every cycle - did this have any noticeable impact on performance?

air/src/errors.rs
```diff
@@ -19,6 +19,4 @@ pub enum ExecutionOptionsError {
     InvalidHashFunction { hash_function: String },
     #[error("core trace fragment size must be greater than 0")]
     CoreTraceFragmentSizeTooSmall,
-    #[error("core trace fragment size {0} must be a power of two")]
-    CoreTraceFragmentSizeNotPowerOfTwo(usize),
```
plafer, line 23: @huitseeker I removed this requirement because some tests require a size that is not a power of 2. AFAIK, this is not required/enforced at runtime, though you claimed that it was [here](https://github.com/0xMiden/miden-vm/pull/2528#discussion_r2659728501). Am I missing something?

processor/src/fast/mod.rs
huitseeker: Opened #2541 to track

processor/src/fast/mod.rs
```diff
@@ -773,18 +797,36 @@ impl FastProcessor {
...
+        stopper: &impl Stopper,
+        continuation: impl FnOnce() -> Option<Continuation>,
+    ) -> ControlFlow<BreakReason> {
         self.clk += 1_u32;
 
         tracer.increment_clk();
 
-        if stopper.should_stop(self) {
-            ControlFlow::Break(())
+        if self.clk >= self.options.max_cycles() as usize {
```
huitseeker, line 825: To answer @bobbinth's question about performance impact: I ran `make exec-info` and a blake3 proving benchmark on both this branch and `next`.

**Results (blake3_1to1 prove):**
- PR branch: total 33.1s, `execute_for_trace` 7.45ms, `build_trace` 78.3ms
- next branch: total 33.1s, `execute_for_trace` 11.1ms, `build_trace` 87.0ms

No regression, execution is actually slightly faster on this branch (likely noise, but definitely not slower).

air/src/errors.rs
huitseeker: Those were the verifications that prompted this comment:

https://github.com/0xMiden/miden-vm/pull/2528/files#diff-dd774facfbbbf4d69b1dcf72c4116145489dd3cb5c63d7466adc861a3c0213f7R184-R188

https://github.com/0xMiden/miden-vm/pull/2528/files#diff-dd774facfbbbf4d69b1dcf72c4116145489dd3cb5c63d7466adc861a3c0213f7R159-R163

My comment was just about having doc that reflects the checks put in at runtime. I didn't catch that the power-of-two check was added speculatively without a concrete runtime requirement. My bad, apologies.

plafer: No worries! Just wanted to make sure that my holiday-induced jello brain didn't miss anything.

processor/src/fast/mod.rs
plafer: > Should we add #[inline(always)] here as well?

Done

> Also, not for this PR, but I'm wondering if we should simplify the debug mode - basically, have a single setting that enables both debugging and tracing. I'm not really sure there is a lot of benefit of having them separate.

Agreed, though let's address this when we know whether better what tracing ends up looking like.

PR #2539, opened 2026-01-05
Only execute trace decorators when tracing is enabled

### LINKED ISSUE #2479: Improve `FastProcessor` debug mode
As part of #2477, the following tests were disabled:

- all tests in `processor/src/tests/debug_mode_decorator_tests.rs`,
- `test_debug_with_debugging()`
- `test_trace_handling()`

This is due to the fact that `in_debug_mode` in the `FastProcessor` currently only controls whether `debug` decorators were executed. As part of this issue, we should

- only execute decorators when `in_debug_mode == true`
- only execute trace handlers when a new `execute_trace_handlers` flag is set

This mirrors the behavior of the flags in `Process`. From an offline discussion, we will probably eventually change what these means. Though for now it's probably best to preserve the former behavior of `Process`.

plafer:
Closes #2479

`trace` decorators are now only executed when `ExecutionOptions::enable_tracing()` is set. The debug mode part was already fixed with #2534. As a result, all tests mentioned in #2479 that were disabled were re-enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 #### Fixes
 
 - `FastProcessor` now correctly returns an error if the maximum number of cycles was exceeded during execution ([#2537](https://github.com/0xMiden/miden-vm/pull/2537))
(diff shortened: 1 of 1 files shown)


huitseeker (approved):
Thank you so much!

Issue #2476, opened 2025-12-16
Re-enable diagnostic tests after `FastProcessor` migration


plafer:
With #2477, the following tests were disabled:

- `test_diagnostic_no_mast_forest_with_procedure()`
- `test_diagnostic_syscall_target_not_in_kernel()`

This comes after migrating them from `Process` to `FastProcessor`, only to realize that the `FastProcessor` didn't output proper diagnostics for both. The `Process`'s solution to use [`add_error_ctx_to_external_error()`](https://github.com/0xMiden/miden-vm/blob/f9fb1be9e03d7db7b0f5e2c9c066ff8c7ae009b4/processor/src/lib.rs#L1017-L1045) doesn't work with the `FastProcessor`, since the `FastProcessor` uses iterative execution instead of recursive execution (and hence can't "catch" the error generated by `execute_external_node()`, and enhance it with an `ErrorContext`). Further investigation is needed to make this work with the `FastProcessor`.

PR #2544, opened 2026-01-06
fix: replace ErrorContext

### LINKED ISSUE #1978: Investigate removing `ErrorContext`

### LINKED ISSUE #2517: Clean up `ExecutionError` enum
... we need to clean up `ExecutionError` enum and this module in general. A few examples:

- `CycleLimitExceeded` doesn't seems to be used - are we no longer enforcing upper limits on execution? This is important functionality that is used downstream (e.g., in `miden-node`).
- `UnexpectedPrecompiles`, `MastNodeNotFoundInForest`, `SmtNodePreImageNotValid`, `SmtNodeNotFound` and probably other variants are not used anywhere. Maybe some are obsolete - and then we should remove them. But it also may be possible that some of them need to be reenabled.
- `InvalidCryptoInput` is confusingly named. I think it should be much more operation-specific (this seems to be used specifically on Merkle root update instruction).
- How we instantiate enum variants is rather inconsistent - sometimes we use constructor functions, sometimes we do this directly. We should normalize the approach.
- Enum variants should be sorted alphabetically.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2472#discussion_r2654512033_

huitseeker:
Closes #1978. Fixes #2517.

This PR removes `ErrorContext` parameter threading from leaf methods in the processor. Previously, `err_ctx` was passed through many layers of function calls just to attach source location when an error occurred. Now, leaf methods return lightweight context-free errors, and source location is added once at execution boundaries.

Credit to @adr1anh for the initial design sketch in #1978 (and an initial branch too).

## Approach

**Before:** Every leaf operation took `err_ctx: &impl ErrorContext` and embedded source location directly in errors.

**After:** Leaf operations return context-free error enums (`OperationError`, `MemoryError`, etc.). Extension traits like `ResultExt::map_exec_err()` add source location at the call site.

```
                        BEFORE                              AFTER

    execute_op ──┬── err_ctx ──► op_div ──► error      execute_op ──► op_div ──► OperationError
                 │                  │                        │                         │
                 ├── err_ctx ──► op_and ──► error            └── .map_exec_err() ◄─────┘
                 │                  │                             (adds context)
                 └── err_ctx ──► op_inv ──► error
```

This reduces `err_ctx` parameter declarations from 106 to 19 (82% reduction). The remaining 19 are trait definitions and boundary functions where context should be added.

## Changes

**New error types:**
- `OperationError` - context-free errors for operations (div-by-zero, assertion failures, type mismatches)
- `IoError` - context-free errors for IO operations (advice + memory)
- `SystemEventError` - context-free errors for system event handlers
- `AceEvalError` - context-free errors for ACE circuit evaluation

**Extension traits for lazy context wrapping:**
- `OperationResultExt::map_exec_err()`
- `MemoryResultExt::map_mem_err()`
- `IoResultExt::map_io_err()`
- `SystemEventResultExt::map_sys_event_err()`
- `AceEvalResultExt::map_ace_eval_err()`

Each error type has its own extension trait because the mapping logic differs:
- `IoError::Execution` unwraps a boxed error rather than wrapping
- Some traits need `op_idx` for precise source location, others don't
- `AdviceResultExt` has a `_no_ctx` variant for context-free cases
- Variant counts differ (SystemEventError has 3, MemoryError has 1)

**Removed unused `ExecutionError` variants** (addresses #2517):
- `MastNodeNotFoundInForest`
- `SmtNodeNotFound`
- `SmtNodePreImageNotValid`
- `UnexpectedPrecompiles`

**Split `ExecutionError::MemoryError`** into two variants:
- `MemoryError { label, source_file, err }` - for user-facing errors with source context
- `MemoryErrorNoCtx(MemoryError)` - for internal operations without meaningful source context

## Review Guide

1. **Start with `processor/src/errors.rs`** - see the new error types and extension traits
2. **Check a leaf operation** like `processor/src/processor/operations/field_ops.rs` - note it returns `OperationError`, not `ExecutionError`
3. **Check a boundary** like `processor/src/processor/operations/mod.rs:execute_sync_op` - note it calls `.map_exec_err(err_ctx, clk)` to add context
4. **Integration tests** show pattern matching still works: `miden-vm/tests/integration/operations/field_ops.rs`

### DIFF
--- a/processor/src/errors.rs
+++ b/processor/src/errors.rs
@@ -522,6 +522,26 @@ impl<T> MapExecErr<T> for Result<T, MemoryError> {
     }
 }
 
+impl<T> MapExecErrWithOpIdx<T> for Result<T, MemoryError> {
+    #[inline(always)]
+    fn map_exec_err_with_op_idx(
+        self,
+        mast_forest: &MastForest,
+        node_id: MastNodeId,
+        host: &impl BaseHost,
+        op_idx: usize,
+    ) -> Result<T, ExecutionError> {
+        match self {
+            Ok(v) => Ok(v),
+            Err(err) => {
+                let (label, source_file) =
+                    get_label_and_source_file(Some(op_idx), mast_forest, node_id, host);
+                Err(ExecutionError::MemoryError { label, source_file, err })
+            },
+        }
+    }
+}
+
 // SystemEventError implementations
 impl<T> MapExecErr<T>
     for Result<T, crate::operations::sys_ops::sys_event_handlers::SystemEventError>
@@ -573,8 +593,8 @@ impl<T> MapExecErrWithOpIdx<T> for Result<T, IoError> {
                 Err(match err {
                     IoError::Advice(err) => ExecutionError::AdviceError { label, source_file, err },
                     IoError::Memory(err) => ExecutionError::MemoryError { label, source_file, err },
-                    // Execution errors are already fully formed, just unwrap
-                    IoError::Execution(err) => *err,
+                    // Execution errors are already fully formed with their own message.
+                    IoError::Execution(boxed_err) => *boxed_err,
                 })
             },
         }
--- a/miden-vm/tests/integration/operations/crypto_ops.rs
+++ b/miden-vm/tests/integration/operations/crypto_ops.rs
@@ -306,7 +306,10 @@ fn crypto_stream_rejects_in_place() {
     let err = test.execute().expect_err("crypto_stream should reject in-place encryption");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::IllegalMemoryAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::IllegalMemoryAccess { .. },
+            ..
+        }
     ));
 }
 
@@ -339,7 +342,10 @@ fn crypto_stream_rejects_partial_overlap() {
         .expect_err("crypto_stream should reject partial overlap (dst within src)");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::IllegalMemoryAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::IllegalMemoryAccess { .. },
+            ..
+        }
     ));
 
     // Test case 2: src starts within dst range (src=1004, dst=1000)
@@ -363,7 +369,10 @@ fn crypto_stream_rejects_partial_overlap() {
         .expect_err("crypto_stream should reject partial overlap (src within dst)");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::IllegalMemoryAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::IllegalMemoryAccess { .. },
+            ..
+        }
     ));
 }
 
@@ -386,7 +395,10 @@ fn crypto_stream_rejects_src_range_overflow() {
     let err = test.execute().expect_err("crypto_stream should reject when src+8 overflows");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::AddressOutOfBounds { .. },
+            ..
+        }
     ));
 }
 
@@ -409,7 +421,10 @@ fn crypto_stream_rejects_dst_range_overflow() {
     let err = test.execute().expect_err("crypto_stream should reject when dst+8 overflows");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::AddressOutOfBounds { .. },
+            ..
+        }
     ));
 }
 
@@ -430,7 +445,10 @@ fn crypto_stream_rejects_unaligned_src() {
     let err = test.execute().expect_err("crypto_stream should reject unaligned src");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::UnalignedWordAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::UnalignedWordAccess { .. },
+            ..
+        }
     ));
 }
 
@@ -451,7 +469,10 @@ fn crypto_stream_rejects_unaligned_dst() {
     let err = test.execute().expect_err("crypto_stream should reject unaligned dst");
     assert!(matches!(
         err,
-        ExecutionError::MemoryErrorNoCtx(MemoryError::UnalignedWordAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::UnalignedWordAccess { .. },
+            ..
+        }
     ));
 }
 
--- a/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_15_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
+++ b/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_15_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
@@ -4,12 +4,24 @@ assertion_line: 135
 expression: fast_stack_outputs
 ---
 Err(
-    MemoryErrorNoCtx(
-        UnalignedWordAccess {
+    MemoryError {
+        label: SourceSpan {
+            source_id: SourceId(
+                4294967295,
+            ),
+            start: ByteIndex(
+                0,
+            ),
+            end: ByteIndex(
+                0,
+            ),
+        },
+        source_file: None,
+        err: UnalignedWordAccess {
             addr: 14,
             ctx: ContextId(
                 0,
             ),
         },
-    ),
+    },
 )
--- a/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_16_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
+++ b/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_16_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
@@ -4,12 +4,24 @@ assertion_line: 135
 expression: fast_stack_outputs
 ---
 Err(
-    MemoryErrorNoCtx(
-        UnalignedWordAccess {
+    MemoryError {
+        label: SourceSpan {
+            source_id: SourceId(
+                4294967295,
+            ),
+            start: ByteIndex(
+                0,
+            ),
+            end: ByteIndex(
+                0,
+            ),
+        },
+        source_file: None,
+        err: UnalignedWordAccess {
             addr: 14,
             ctx: ContextId(
                 0,
             ),
         },
-    ),
+    },
 )
--- a/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_17_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
+++ b/processor/src/fast/tests/snapshots/miden_processor__fast__tests__all_ops__fast::tests::all_ops::test_basic_block::stack_inputs_17_vec__Felt__from_u32_1__Felt__from_u32_2__Felt__from_u32_3__Felt_::operations_68_vec__Operation__HornerExt_.snap
@@ -4,12 +4,24 @@ assertion_line: 135
 expression: fast_stack_outputs
 ---
 Err(
-    MemoryErrorNoCtx(
-        UnalignedWordAccess {
+    MemoryError {
+        label: SourceSpan {
+            source_id: SourceId(
+                4294967295,
+            ),
+            start: ByteIndex(
+                0,
+            ),
+            end: ByteIndex(
+                0,
+            ),
+        },
+        source_file: None,
+        err: UnalignedWordAccess {
             addr: 14,
             ctx: ContextId(
                 0,
             ),
         },
-    ),
+    },
 )
--- a/processor/src/operations/mod.rs
+++ b/processor/src/operations/mod.rs
@@ -1,4 +1,4 @@
-use super::{ExecutionError, Felt};
+use super::Felt;
 
 pub(crate) mod sys_ops;
 pub(crate) mod utils;
--- a/processor/src/operations/utils.rs
+++ b/processor/src/operations/utils.rs
@@ -1,6 +1,6 @@
 use miden_air::trace::RowIndex;
 
-use super::{ExecutionError, Felt};
+use super::Felt;
 use crate::{ContextId, MemoryError, ONE, PrimeField64, ZERO};
 // Note: assert_binary now returns OperationError, imported via crate::OperationError in the
 // function
@@ -16,24 +16,22 @@ pub(crate) fn validate_dual_word_stream_addrs(
     dst_addr: Felt,
     ctx: ContextId,
     clk: RowIndex,
-) -> Result<(), ExecutionError> {
+) -> Result<(), MemoryError> {
     // Convert to u32 and check end-exclusive bounds
     let src_addr_u64 = src_addr.as_canonical_u64();
     let dst_addr_u64 = dst_addr.as_canonical_u64();
 
-    let src_addr_u32 = u32::try_from(src_addr_u64).map_err(|_| {
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { addr: src_addr_u64 })
-    })?;
-    let src_end = src_addr_u32.checked_add(8).ok_or_else(|| {
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { addr: src_addr_u64 })
-    })?;
+    let src_addr_u32 = u32::try_from(src_addr_u64)
+        .map_err(|_| MemoryError::AddressOutOfBounds { addr: src_addr_u64 })?;
+    let src_end = src_addr_u32
+        .checked_add(8)
+        .ok_or(MemoryError::AddressOutOfBounds { addr: src_addr_u64 })?;
 
-    let dst_addr_u32 = u32::try_from(dst_addr_u64).map_err(|_| {
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { addr: dst_addr_u64 })
-    })?;
-    let dst_end = dst_addr_u32.checked_add(8).ok_or_else(|| {
-        ExecutionError::MemoryErrorNoCtx(MemoryError::AddressOutOfBounds { addr: dst_addr_u64 })
-    })?;
+    let dst_addr_u32 = u32::try_from(dst_addr_u64)
+        .map_err(|_| MemoryError::AddressOutOfBounds { addr: dst_addr_u64 })?;
+    let dst_end = dst_addr_u32
+        .checked_add(8)
+        .ok_or(MemoryError::AddressOutOfBounds { addr: dst_addr_u64 })?;
 
     // Check for overlap between [src, src+8) and [dst, dst+8)
     if src_addr_u32 < dst_end && dst_addr_u32 < src_end {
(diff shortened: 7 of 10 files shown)


processor/src/errors.rs
```diff
@@ -508,8 +587,12 @@ macro_rules! err_ctx {
...
 #[macro_export]
 macro_rules! err_ctx {
-    ($mast_forest:expr, $node:expr, $host:expr, $in_debug_mode:expr) => {{ () }};
-    ($mast_forest:expr, $node:expr, $host:expr, $in_debug_mode:expr, $op_idx:expr) => {{ () }};
+    ($mast_forest:expr, $node:expr, $host:expr, $in_debug_mode:expr) => {
+        ()
+    };
+    ($mast_forest:expr, $node:expr, $host:expr, $in_debug_mode:expr, $op_idx:expr) => {
+        ()
+    };
```
plafer: I think we can get rid of the `err_ctx!` macro, the `no_err_ctx` feature and potentially `ErrorContext` altogether. The idea would be to run the `precalc_label_and_source_file()` function inside the `XResultExt::map_x_err()` methods.

For example, I would see `OperationResultExt<T>` being implemented as something like

```rs
impl<T> OperationResultExt<T> for Result<T, OperationError> {
    // The `mast_forest`, `node_id` and `host` are used to call today's
    // `ErrorContextImpl::precalc_label_and_source_file()` internally
    // (converted to a standalone function).
    fn map_exec_err(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        clk: AirRowIndex,
    ) -> Result<T, ExecutionError> {
        self.map_err(|err| {
            let (label, source_file) = calculate_label_and_source_file(mast_forest, node_id, host);
            ExecutionError::OperationError { label, source_file, clk, err }
        })
    }
}
```

The core idea being that since we now delay the label & source file lookup computation with a closure to be run only on error, we don't need any of the `ErrorContext`/`err_ctx!`/`no_err_ctx` machinery anymore.

adr1anh:
I think we can also remove the `clk` since it's quite easy to get wrong, and doesn't provide much info in the failure case. Cc @plafer who suggested this.

huitseeker:
@plafer @adr1anh Thanks for the bold but simplifying suggestions! Have a look at https://github.com/0xMiden/miden-vm/pull/2544/commits/36434b6f7c29c1c612a570d30df5bae87a4fc52c

processor/src/chiplets/memory/errors.rs
```diff
@@ -34,34 +30,9 @@ pub enum MemoryError {
...
-        #[label("tried to access memory address {addr}")]
-        label: SourceSpan,
-        #[source_code]
-        source_file: Option<Arc<SourceFile>>,
-        addr: u32,
-        ctx: ContextId,
-        clk: Felt,
-    },
-    // Note: we need this version as well because to handle advice provider calls, which don't
+    UnalignedWordAccess { addr: u32, ctx: ContextId, clk: Felt },
```
adr1anh: Does it make sense to remove the `clk` here and remove the `UnalignedWordAccessNoClk` variant? cc @plafer

processor/src/chiplets/bitwise/mod.rs
```diff
@@ -124,14 +119,9 @@ impl Bitwise {
...
-    pub fn u32xor(
-        &mut self,
-        a: Felt,
-        b: Felt,
-        err_ctx: &impl ErrorContext,
-    ) -> Result<Felt, ExecutionError> {
-        let a = assert_u32(a, err_ctx)?.as_canonical_u64();
-        let b = assert_u32(b, err_ctx)?.as_canonical_u64();
+    pub fn u32xor(&mut self, a: Felt, b: Felt) -> Result<Felt, OperationError> {
+        let a = assert_u32(a)?.as_canonical_u64();
```
adr1anh, line 123: Not for this PR, but it might make sense to return the resulting `u32` from `assert_u32`.

processor/src/chiplets/kernel_rom/mod.rs
```diff
@@ -74,16 +74,12 @@ impl KernelRom {
...
-        proc_hash: Digest,
-        err_ctx: &impl ErrorContext,
-    ) -> Result<(), ExecutionError> {
+    pub fn access_proc(&mut self, proc_hash: Digest) -> Result<(), OperationError> {
         let proc_hash_bytes: ProcHashBytes = proc_hash.into();
         let access_info = self
             .access_map
             .get_mut(&proc_hash_bytes)
-            .ok_or(ExecutionError::syscall_target_not_in_kernel(proc_hash, err_ctx))?;
+            .ok_or(OperationError::SyscallTargetNotInKernel { proc_root: proc_hash })?;
```
adr1anh: Should we rename `proc_hash` to `proc_root` for consistency (or the other way round)?

processor/src/chiplets/memory/errors.rs
adr1anh: Is `ctx` relevant for the error message?

processor/src/fast/tests/memory.rs
```diff
@@ -32,10 +30,8 @@ fn test_memory_word_access_alignment() {
         let err = FastProcessor::new(&[43_u32.into()])
             .execute_sync(&program, &mut host)
             .unwrap_err();
-        assert_eq!(
-            err.to_string(),
-            "word memory access at address 43 in context 0 is unaligned at clock cycle 1"
-        );
+        // The error is wrapped: "memory error" is the outer message
+        assert_eq!(err.to_string(), "memory error");
```
adr1anh: I think @plafer mentioned the "memory error" is not super useful to users, and we can avoid adding an extra error line by just transparently printing the inner memory error.

processor/src/fast/mod.rs
```diff
@@ -835,35 +832,40 @@ impl FastProcessor {
...
 
         // Merge the advice map of this forest into the advice provider.
         // Note that the map may be merged multiple times if a different procedure from the same
         // forest is called.
         // For now, only compiled libraries contain non-empty advice maps, so for most cases,
         // this call will be cheap.
-        self.advice
-            .extend_map(mast_forest.advice_map())
-            .map_err(|err| ExecutionError::advice_error(err, self.clk, err_ctx))?;
+        self.advice.extend_map(mast_forest.advice_map()).map_advice_err(
```
adr1anh: This might warrant a special error type indicating that this is due to a malformed Mast. Not for this PR though.

processor/src/tests/mod.rs
```diff
@@ -123,14 +125,16 @@ fn test_diagnostic_advice_map_key_not_found_2() {
 
     let build_test = build_test_by_mode!(true, source, &[1, 2]);
     let err = build_test.execute().expect_err("expected error");
+    // Note: The source location points to the first operation in the basic block because
+    // system event errors currently don't have access to the specific operation index.
```
adr1anh, line 128: Is it worth making an issue for this?

processor/src/chiplets/memory/errors.rs
plafer: That would be my preference yes, I don't think seeing "X occurred on clock cycle 5362" ever helped anyone

plafer: Hmmm, my thoughts were that it would help the user distinguish between whether the error occurred in the kernel or user code. But I guess they'll see that from the diagnostics. So yeah, it's probably not that relevant?

adr1anh:
This is great, and thank you for putting this together! There were a few things that I considered a while ago when I tackled this. I think many of these points can be tackled in later PRs though.

#### Stack size

Even with these changes, we end up with a pretty big stack footprint both due to the number of arguments we need to pass to functions, but also due to the growing size of `ExecutionError`. For the former, this could potentially be solved by re-introducing an `ErrorContext` struct that can be passed down as reference. For the latter, we could try to add more `Box`ing of inner types. It's not exactly trivial since `Diagnostic` is not implemented for `Box<DiagnosticError>` directly and can be a bit more awkward.

#### Review use of `help` messages

The current usages of `#[diagnostic(help(...)]` are few and don't add much info to the error. Except for [`AdviceError::MapKeyAlreadyPresent`](https://github.com/0xMiden/miden-vm/blob/cb5d74b08e336e154f911e201e938979e5c5137b/processor/src/host/advice/errors.rs#L16), the other messages could be replaced with a more exhaustive error string. We could even do the same for the latter. 

I think this would lead to simplifications since we can remove `#[diagnostic()]` and `#[diagnostic(transparent)]` attributes on `ExecutionError` variants, while also unifying error messages. Another benefit would be that it makes it easier to `Box` the inner error, helping with the above point.

#### Review actual error variants

I've noticed a couple of errors are only instantiated once or twice.

- `AceError` could probably be simplified to just be a `String` since most variants do not contain any information, and these internal errors are mostly useful for us. I think users would only really care that `"ace circuit evaluation failed"`.
- The same holds for FRI-related errors. 
- We could consolidate NotBinaryValue / NotBinaryValueIf / NotBinaryValueLoop
- DynamicNodeNotFound and NoMastForestWithProcedure are quite similar and I think the difference would be clear from the miette diagnostics (either pointing to a dynexec / dyncall or a procedure). We could make these a single ProcedureNotFound. Conceptually, they're probably more like ExecutionErrors than OperationErrors

We could also introduce a `HostError` for the DuplicateEventHandler / TraceHandlerError / DebugHandlerError / ReservedEventNamespace variants. 

#### Introduce Mast-specific error type

I'm not entirely sure if this makes sense, this is mainly motivated by the fact that a failure to merge a MastForest's AdviceMap returns an AdviceError in [`load_mast_forest`](https://github.com/0xMiden/miden-vm/blob/36434b6f7c29c1c612a570d30df5bae87a4fc52c/processor/src/fast/mod.rs#L835). I'm not sure whether this is actually more useful to users though, and could be tackled at the same time as lazy AdviceMap loading.

processor/src/chiplets/memory/errors.rs
huitseeker: Done!

processor/src/chiplets/kernel_rom/mod.rs
huitseeker: Done

processor/src/fast/tests/memory.rs
huitseeker: Done + commit rationalizing the error wrapping

huitseeker:
Thanks for the thorough review @adr1anh! Really helpful suggestions.

On consolidating `NotBinaryValue` / `NotBinaryValueIf` / `NotBinaryValueLoop`: there's a tradeoff here. The miette diagnostics do point to the exact instruction (`if.true`, `while.true`, etc.), which makes the variant-specific messages somewhat redundant. But the explicit messages ("if statement expected...", "loop condition must be...") provide immediate context without requiring users to parse the source snippet. I'd lean toward keeping them separate for now, but happy to revisit.

  For the other suggestions not yet addressed, I think each warrants its own issue:

  - Stack size reduction (re-introduce `ErrorContext` as a reference struct, more `Box`ing of inner types)
  - Review `#[diagnostic(help(...))]` usage (most could be folded into error messages)
  - Consolidate error variants (`AceError` to `String`, FRI errors, `NotBinaryValue*` variants, `DynamicNodeNotFound`/`NoMastForestWithProcedure`)
  - Introduce `HostError` for handler-related variants
  - Mast-specific error type for `load_mast_forest` failures

  The core scope of this PR was addressing issue #1978: removing eager `ErrorContext computation`. That's done - we've removed the `ErrorContext` trait, `err_ctx!` macro, and `no_err_ctx` feature, replacing them with lazy context resolution via extension traits. I think this is ready for final review and merge, with the above as follow-up work.
  
  
  ----
  **Edit:** opened #2576, #2577, #2578 to track

processor/src/errors.rs
```diff
@@ -261,332 +93,584 @@ pub enum ExecutionError {
...
+    node_id: MastNodeId,
+    host: &impl BaseHost,
+    in_debug_mode: bool,
+) -> (SourceSpan, Option<Arc<SourceFile>>) {
+    // When not in debug mode, skip the expensive decorator traversal entirely.
+    // Decorators (including AsmOp decorators used for error context) should only
+    // be accessed when debugging is enabled.
+    if !in_debug_mode {
+        return (SourceSpan::UNKNOWN, None);
     }
```
plafer: Is this necessary, given that it's only done once and on error right before program exit? Also, I could see us wanting to generate proper diagnostics in production without necessarily running in debug mode. What do you think?

processor/src/errors.rs
```diff
@@ -261,332 +93,584 @@ pub enum ExecutionError {
...
+    /// Maps an `OperationError` to an `ExecutionError` with op index context.
+    fn map_exec_err_with_op_idx(
+        self,
+        mast_forest: &MastForest,
+        node_id: MastNodeId,
+        host: &impl BaseHost,
+        in_debug_mode: bool,
+        op_idx: usize,
+    ) -> Result<T, ExecutionError>;
+}
```
plafer: Any reason we want a new trait for every concrete error type? Couldn't we have 3 traits (one with `map_exec_err()`, one with `map_exec_err_with_op_idx()`, and one with `map_exec_err_no_ctx()`), and implement the appropriate one(s) for `Result<T, OperationError>`, `Result<T, AdviceError>`, etc? Something like (names subject to change)

```rs
// The traits - only showing the 2 main ones for simplicity

pub trait MapExecutionErrorExt<T> {
    /// Maps an error to an `ExecutionError` with the provided context.
    fn map_exec_err(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
    ) -> Result<T, ExecutionError>;
}

pub trait MapExecutionErrorWithOpIdxExt<T> {
    /// Maps an error to an `ExecutionError` with op index context.
    fn map_exec_err_with_op_idx(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
        op_idx: usize,
    ) -> Result<T, ExecutionError>;
}

// OperationError: implement both traits

impl<T> MapExecutionErrorExt<T> for Result<T, OperationError> {
    fn map_exec_err(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
    ) -> Result<T, ExecutionError> {
        // ...
    }
}

impl<T> MapExecutionErrorWithOpIdxExt<T> for Result<T, OperationError> {
    fn map_exec_err_with_op_idx(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
    ) -> Result<T, ExecutionError> {
        // ...
    }
}

// All other error types: only implement MapExecutionErrorExt

impl<T> MapExecutionErrorExt<T> for Result<T, AdviceError> {
    fn map_exec_err(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
    ) -> Result<T, ExecutionError> {
        // ...
    }
}

impl<T> MapExecutionErrorExt<T> for Result<T, MemoryError> {
    fn map_exec_err(
        self,
        mast_forest: &MastForest,
        node_id: MastNodeId,
        host: &impl BaseHost,
        in_debug_mode: bool,
    ) -> Result<T, ExecutionError> {
        // ...
    }
}
```

processor/src/fast/call_and_dyn.rs
```diff
@@ -68,9 +70,8 @@ impl FastProcessor {
             self.caller_hash = callee_hash;
 
             // Initialize the frame pointer in memory for the new context.
-            if let Err(err) = self.memory.write_element(new_ctx, FMP_ADDR, FMP_INIT_VALUE, &err_ctx)
-            {
-                return ControlFlow::Break(BreakReason::Err(ExecutionError::MemoryError(err)));
+            if let Err(err) = self.memory.write_element(new_ctx, FMP_ADDR, FMP_INIT_VALUE) {
+                return ControlFlow::Break(BreakReason::Err(ExecutionError::MemoryErrorNoCtx(err)));
             }
```
plafer: Why do we use `MemoryErrorNoCtx` here? I actually forget, but previously we passed an `ErrorContext`, so wouldn't this failing have given a diagnostic showing the `call.xyz` line as where the error originated? Whereas now we won't get any diagnostic if there's a memory error, right?

Same question applies to other uses of `MemoryErrorNoCtx` in this file.

processor/src/fast/external.rs
```diff
@@ -50,24 +53,38 @@ impl FastProcessor {
 
     /// Analogous to [`Process::resolve_external_node`](crate::Process::resolve_external_node), but
     /// for asynchronous execution.
+    ///
+    /// Note: Since we're resolving an external node, we don't have meaningful error context
+    /// (no "current" forest/node). We pass the external node's parent forest and node_id for
+    /// error context, with debug mode disabled for this resolution step.
```
plafer: Not sure exactly what is meant here, but note that the `FastProcessor` in `next` currently does the "wrong thing" with respect to external node diagnostics (documented in #2476).

i.e. there *is* something we can do with respect to external node diagnostics because we used to do it with `Process` (see [here](https://github.com/0xMiden/miden-vm/blob/f9fb1be9e03d7db7b0f5e2c9c066ff8c7ae009b4/processor/src/lib.rs#L1017-L1045)). That being said, we shouldn't fix this is in this PR, but might be worth adjusting the comment (i.e. we're not outputting the diagnostics we should currently, but this is tracked with #2476)

processor/src/fast/memory.rs
```diff
@@ -96,11 +87,10 @@ impl Memory {
         &mut self,
         ctx: ContextId,
         addr: Felt,
-        clk: RowIndex,
+        _clk: RowIndex,
```
plafer, line 90: remove?

Makefile
```diff
@@ -234,9 +234,9 @@ run-examples: exec ## Runs all masm examples to verify they execute correctly
 # --- benchmarking --------------------------------------------------------------------------------
 
 .PHONY: check-bench
-check-bench: ## Builds all benchmarks (incl. those needing no_err_ctx)
-	cargo check --benches --features internal,no_err_ctx
+check-bench: ## Builds all benchmarks
+	cargo check --benches --features internal
```
plafer, line 238: I'm getting a ~13% performance regression on the `program_execution_fast` (`blake3_1to1`) benchmark, though I didn't investigate further why. Any idea what might be causing this?

processor/src/errors.rs
bitwalker: Agreed, I would think if we're already on the error path, then the expense of extracting maximal context for the error is always worth it, since without it, there is very little to indicate what actually went wrong.

Makefile
huitseeker: Thanks for the report! I reproduced the slowdown. I believe the issue is LLVM was not fully optimizing away closure setup overhead in `map_err` calls. Even when the closure is never invoked (success path), the compiler was generating code for closure environment capture.

I replaced `map_err` closures with explicit `match` patterns in all error extension traits (`processor/src/errors.rs`). This gives LLVM clearer visibility into control flow, and allows it to eliminate error handling overhead on the success path.

**Before:**
```rust
self.map_err(|err| {
    let (label, source_file) = get_label_and_source_file(...);
    ExecutionError::OperationError { label, source_file, err }
})
```

**After:**
```rust
match self {
    Ok(v) => Ok(v),
    Err(err) => {
        let (label, source_file) = get_label_and_source_file(...);
        Err(ExecutionError::OperationError { label, source_file, err })
    },
}
```

I added a couple of other changes (inline annotations, hoisting `in_debug_mode` better).

### Full Benchmark Measurements

Each commit was benchmarked 3 times with 5s warm-up and 10s measurement time.

| Commit | Description | Run 1 | Run 2 | Run 3 | Average | Std Dev | vs Baseline |
|--------|-------------|-------|-------|-------|---------|---------|-------------|
| `origin/next` + `no_err_ctx` | Baseline (feature flag) | 3.052 ms | 3.057 ms | 3.042 ms | **3.050 ms** | ±0.008 ms | - |
| `7c11c959e` | Before fix (with regression) | 3.466 ms | 3.570 ms | 3.420 ms | **3.485 ms** | ±0.076 ms | +14.3% |
| `1ff422bb0` | After fix | 3.054 ms | 3.043 ms | 3.058 ms | **3.052 ms** | ±0.008 ms | +0.07% |
| `17ce096aa` | Tip (with trait consolidation) | 3.104 ms | 3.078 ms | 3.076 ms | **3.086 ms** | ±0.016 ms | +1.2% |

<details><summary> benchmark </summary>

```bash
# Run the fast execution benchmark
cargo bench --bench program_execution_fast --features internal

# To compare against baseline with no_err_ctx feature (if available)
cargo bench --bench program_execution_fast --features internal,no_err_ctx
```

</details>

processor/src/chiplets/memory/errors.rs
huitseeker: Removed!

processor/src/errors.rs
huitseeker: Fixed

processor/src/errors.rs
huitseeker: Thanks! I was very unsure about this initially, because the structs required different arguments at different points in the edition of this PR (see PR description), but turns out this works out nicely enough. See 17ce096aac8867019acb7e6b2b377a8b465beedd

processor/src/fast/memory.rs
huitseeker: It's part of the `MemoryInterface` trait.

processor/src/fast/external.rs
huitseeker: Edited!

processor/src/fast/call_and_dyn.rs
huitseeker: Fixed. This file now uses `map_exec_err()` with full context. Memory errors here will include source diagnostics. `crypto_ops.rs` and `operations/utils.rs` still use `MemoryErrorNoCtx`. These functions don't have access to `MastForest`/`node_id`/`host` parameters.

processor/src/chiplets/bitwise/mod.rs
huitseeker: #2575

processor/src/tests/mod.rs
huitseeker: #2578

processor/src/fast/mod.rs
huitseeker: #2575

plafer:
@huitseeker 3 commits are not signed (so we can't merge)

PR #2469, opened 2025-12-13
feat: delta-encode BasicBlockNode metadata (2/3)

huitseeker:
This PR implements delta encoding for `BasicBlockNode` indptr arrays, building on the padded serialization format from #2466. It reduces batch metadata from 10 bytes to 5 bytes per batch through 4-bit delta encoding.

The delta-encoding mentioned in the previous PR is now implemented:
- **Before**: 9 bytes indptr array (9 × 8 bits = 72 bits)
- **After**: 4 bytes packed deltas (8 × 4 bits = 32 bits, first index implicit)

Clarified the partial validity semantics of OpBatch arrays:
- **`indptr[0..num_groups+1]`** - semantically valid prefix
- **`padding[0..num_groups]`** - semantically valid prefix
- **`groups[0..num_groups]`** - semantically valid prefix
- **Tail entries beyond these prefixes** - undefined garbage for semantic purposes, but must be "safe garbage" for serialization

The entire `indptr` array must be monotonically non-decreasing for delta encoding. `OpBatchAccumulator` fills the tail with the final ops count to maintain this invariant while keeping the tail semantically meaningless.

### Summary

The delta-encoded format adds **2.20% size overhead** compared to unpadded baseline:
- 0.67% from NOOP padding (633 ops)
- 1.53% from batch metadata (17,536 bytes with delta encoding)

**Savings vs padded format: 15,780 bytes (-1.84%)**

Most blocks (92.7%) add ≤22 bytes.

<details><summary>Size comparison across serialization formats for MastForest.</summary>

## Size Comparison

| Format | Size | vs Unpadded | vs Padded |
|--------|------|-------------|-----------|
| Unpadded (`next`) | 825,755 bytes (806.40 KB) | baseline | -3.95% |
| Padded (previous PR) | 859,730 bytes (839.58 KB) | +33,975 bytes (+4.11%) | baseline |
| **Delta-encoded (this PR)** | **843,950 bytes (824.17 KB)** | **+18,195 bytes (+2.20%)** | **-15,780 bytes (-1.84%)** |

### NOOP Padding: 633 operations (0.67%)

| Metric | Count |
|--------|-------|
| Total operations | 94,744 |
| NOOP padding | 633 |
| Real operations | 94,111 |

### Batch Metadata: 17,536 bytes (with delta encoding)

| Component | Per Unit | Count | Total | Previous |
|-----------|----------|-------|-------|----------|
[...]

PR #2549, opened 2026-01-07
feat(core): add stripped MastForest serialization

### LINKED ISSUE #1580: Allow to truncate decorator data when serializing MAST

huitseeker:
On top of #2466 #2469 #2470. Adds support for serializing MastForest without debug information, producing smaller output files for production deployment.

- Add `FLAG_STRIPPED (0x01)` to indicate debug info is omitted
- Add `MastForest::write_stripped()` public method, redirect to a proxy newtype (for type safety).
- Deserializer auto-detects format and creates empty `DebugInfo` when stripped
- Reject unknown flags (reserved bits must be zero)

Together with the above PRs this is stacked on, closes #1580.

Note: this complements rather than replaces the in-memory decorator stripping from #2108, which remains useful for improving execution and proving performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - [BREAKING] Change serialization of `BasicBlockNode`s to use padded indices ([#2466](https://github.com/0xMiden/miden-vm/pull/2466/)).
 - Change padded serialization of `BasicBlockNode`s to use delta-encoded metadata ([#2469](https://github.com/0xMiden/miden-vm/pull/2469/)).
 - Change (de)serialization of `MastForest` to directly (de)serialize DebugInfo ([#2470](https://github.com/0xMiden/miden-vm/pull/2470/)).
+- Add API to serialize the `MastForest` without `DebugInfo` ([#2549](https://github.com/0xMiden/miden-vm/pull/2549)).
 
 ## 0.20.2 (TBD)
 - Fix issue where decorator access was not bypassed properly in release mode ([#2529](https://github.com/0xMiden/miden-vm/pull/2529)).


core/src/mast/serialization/mod.rs
```diff
@@ -83,24 +107,36 @@ const MAGIC: &[u8; 5] = b"MAST\0";
...
+
+        // version
         target.write_bytes(&VERSION);
 
-        // decorator & node counts
+        // node & decorator counts
         target.write_usize(self.nodes.len());
-        // Expected to be used in #2504. Remove if this issue is resolved without using.
-        target.write_usize(self.debug_info.num_decorators());
+        target.write_usize(if stripped { 0 } else { self.debug_info.num_decorators() });
```
bobbinth, line 139: Not for this PR, but I'm still not sure we need to serialize these values.

PR #2559, opened 2026-01-09
Use TryFrom<Vec<T>> impl for IndexVec

### LINKED ISSUE #2532: feat: add From<Vec<T>> for IndexVec<I, T>
Can we initialize the vector with zeros directly, rather than pushing one by one?

_Originally posted by @adr1anh in https://github.com/0xMiden/miden-vm/pull/2529#discussion_r2660941249_

Himess:
Closes #2532


Himess:
Not a problem at all @huitseeker, I've updated the PR as you suggested

huitseeker (approved):

Issue #2509, opened 2025-12-30
Restore AIR constraints using AirScript with Plonky3 backend


Al-Kindi-0:
During the Winterfell → Plonky3 migration (#2472), the AIR constraints in the air/ crate were removed because they were based on Winterfell's Air trait, which is incompatible with Plonky3's AirBuilder trait.

We should be able to restore the AIR constraints for Miden VM using:
  1. AirScript constraints already written in https://github.com/0xMiden/air-script/pull/527
  2. Plonky3 backend added to AirScript

bobbinth:
As a part of this, we should make sure to update the README file for the `miden-air` crate.

Al-Kindi-0:
This done using the direct approach

PR #2562, opened 2026-01-09
Plonky3 MidenVM Air constraints

Leo-Besancon:
## Describe your changes

This draft PR aims to resolve https://github.com/0xMiden/miden-vm/issues/2509. It adds back the AIR constraints for the MidenVM that were removed in #2472, as Plonky3 constraints instead of Winterfell.

For now, I've added the chiplets constraints (note that they have been added in bulk from the air-script generated code).

### DIFF
--- a/air/src/constraints/mod.rs
+++ b/air/src/constraints/mod.rs
@@ -110,7 +110,7 @@ where
 
     // RANGE CHECKER BUS CONSTRAINTS
     #[cfg(feature = "range_constraints")]
-    range::bus::enforce_range_bus_constraints(builder, alpha, &beta_challenges, &aux_current, &aux_next, local, next);
+    range::bus::enforce_range_bus_constraints(builder, local);
 
     // CHIPLETS BUS CONSTRAINTS
     #[cfg(feature = "chiplets_constraints")]
--- a/air/src/constraints/range/bus.rs
+++ b/air/src/constraints/range/bus.rs
@@ -1,7 +1,7 @@
 use core::borrow::Borrow;
 
 use miden_core::field::PrimeCharacteristicRing;
-use miden_crypto::stark::air::MidenAirBuilder;
+use miden_crypto::stark::{air::MidenAirBuilder, matrix::Matrix};
 
 use crate::MainTraceRow;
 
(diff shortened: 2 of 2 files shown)


bobbinth:
As discussed last week, we can narrow down the goal of this PR to be:

1. Implement some subset of constraints that works now (or doesn't require too much debugging).
2. Add a feature to control whether the VM runs with or without constraints enabled.

The idea is to set up the structure such that we can very easily copy-paste improved/updated constraint files in the future and examine their execution.

Leo-Besancon:
Putting this PR as ready for review now that the main pipeline is working. The constraints have been generated from https://github.com/0xMiden/air-script/pull/536

A few notes / todos:
- I should adapt miden-air's README.md
- Run the tests with `--features constraint_eval` to evaluate the miden-vm constraints that are not commented out. For instance:
```shell
cargo test --features constraint_eval --package miden-vm --test miden-cli -- prove_verify::fast_parallel::test_fast_parallel_prove_verify --exact --nocapture

running 1 test
test prove_verify::fast_parallel::test_fast_parallel_prove_verify ... ok
```
- Some tests still fail with the current constraints set, I'll try and debug them quickly. It may be due to a misalignment of the trace (with the system being 6 or 8 columns wide).

air/src/constraints/stack/field_ops/mod.rs
bobbinth, line 1: Why is this file empty? I'm assuming we'll have constraints here eventually, but shouldn't we have a stubbed out function here for evaluating these constraints?

Also relevant for other similar files here.

air/src/unedited_constraints/miden_vm_plonky3.rs
bobbinth, line 1: Why do we need these "unedited_constraints" files? I think we should have only one set of constraints committed. If all constraints are in a single file - that's also fine - but then we probably don't need the empty constraint files I mentioned in the previous comment?

air/src/lib.rs
```diff
@@ -128,48 +142,120 @@ impl Deserializable for PublicInputs {
...
+where
+    A: MidenAir<Felt, EF>,
+    EF: ExtensionField<Felt>,
+    B: AuxTraceBuilder<EF>,
+{
+    inner: Option<A>,
     /// Auxiliary trace builder for generating auxiliary columns.
     aux_builder: Option<B>,
+    phantom: core::marker::PhantomData<EF>,
 }
```
bobbinth: Could we explain what is `inner` and why it is optional?

bobbinth: Could we explain what is `inner` and why it is optional?

air/src/lib.rs
```diff
@@ -128,48 +142,120 @@ impl Deserializable for PublicInputs {
...
+            aux_builder: Some(builder),
+            phantom: core::marker::PhantomData,
+        }
     }
 }
 
-impl<EF, B> MidenAir<Felt, EF> for ProcessorAir<B>
+use p3_matrix::dense::RowMajorMatrix;
+
+use crate::trace::AUX_TRACE_WIDTH;
```
bobbinth: Could we organize this file a bit better so that imports/constants are not scattered?

air/src/lib.rs
```diff
@@ -128,48 +142,120 @@ impl Deserializable for PublicInputs {
...
+        self.inner.as_ref().map(|inner| inner.width()).unwrap_or(TRACE_WIDTH)
     }
 
-    fn aux_width(&self) -> usize {
-        // Return the number of extension field columns
-        // The prover will interpret the returned base field data as EF columns
-        AUX_TRACE_WIDTH
+    fn preprocessed_trace(&self) -> Option<RowMajorMatrix<Felt>> {
+        self.inner.as_ref().map(|inner| inner.preprocessed_trace()).unwrap_or(None)
+    }
```
bobbinth: Do we use preprocessed traces somehow?

air/src/lib.rs
```diff
@@ -128,48 +142,120 @@ impl Deserializable for PublicInputs {
...
+
+    fn num_public_values(&self) -> usize {
+        self.inner.as_ref().map(|inner| inner.num_public_values()).unwrap_or(0) // todo
+    }
+
+    fn periodic_table(&self) -> Vec<Vec<Felt>> {
```
bobbinth, line 244: This function is pretty difficult to read. I think each inner vector is a periodic column, right? If so, I'd probably define these as arrays in a separate file and then here would do `COLUMN_X.to_vec()`.

Separately, it would be nice if the return type here was more structured - e.g., `PeriodicColumnTable`. This would allow us to validate that periods are always powers of two and maybe some other things. I'm assuming this is an API change somewhere upstream - so, for now, may be good to create an issue about this.

verifier/src/lib.rs
```diff
@@ -132,7 +132,10 @@ fn verify_stark(
     let pub_inputs =
         PublicInputs::new(program_info, stack_inputs, stack_outputs, pc_transcript_state);
     let public_values = pub_inputs.to_elements();
-    let air = ProcessorAir::new();
+
+    let inner = miden_air::unedited_constraints::miden_vm_plonky3::MidenVM {};
+    let air = miden_air::ProcessorAir::new(Some(inner));
+    //let air = miden_air::ProcessorAir::new(None::<miden_air::unedited_constraints::miden_vm_plonky3::MidenVM>);
```
bobbinth: I'm not sure about this pattern. Ideally, we keep this as it was (i.e., just `let air = ProcessorAir::new();`) and handle all manipulations internally in the `miden-air` crate.

script.sh
bobbinth, line 1: Would be good to add some comments about what this script does and how to use it. Also, we usually put scripts into the `/scripts` directory.

air/src/lib.rs
```diff
@@ -184,27 +270,52 @@ where
...
+            constraints::range::enforce_range_bus_constraint(builder, local);
+
+            // CHIPLETS CONSTRAINTS
+            constraints::chiplets::enforce_chiplets_transition_constraint(
+                builder,
+                local,
+                next,
+                &periodic_values,
+            );
+            constraints::chiplets::enforce_chiplets_bus_constraint(builder, local);
```
bobbinth: Will we have a separate file for each section? I was imagining this would be something like `constraints::evaluate()` and then internally in the `constraints` module we'd figure out how to organize things (e.g., maybe for hand-written constraints there would be multiple sub-modules, but for generated ones it would be just one file?).

processor/src/trace/mod.rs
```diff
@@ -119,6 +119,17 @@ impl ExecutionTrace {
         public_inputs.to_elements()
     }
 
+    /// Returns the public values for this execution trace.
+    pub fn to_variable_length_public_values(&self) -> Vec<Felt> {
```
bobbinth: Seems like this is identical to `to_public_values()` - why do we need this?

air/src/unedited_constraints/miden_vm_plonky3.rs
Leo-Besancon: The idea was to have:
- in the `unedited_constraints` module the generated constraints from the compilation of the `miden_vm.air` with the Plonky3 backend [here](https://github.com/0xMiden/air-script/pull/536/changes#diff-eddf5d4ad6ae68da5a6121977c9e15843828e2e96e9cc4ec629672681e1aa174), so that we can just copy-paste the code with very minimal changes.
- in the `constraints` module have the constraints in a more readable way (written partly from code generation, partly manually). We could remove them from this PR to avoid code duplication, but it could still be useful to help debug the constraints and ensure the two sets evaluate to the same values. Also, for example the RPO constraints there are straightforward, contrary to the generated code that have the issue I mentionned last week with stack overflow.
- a way to swich from one or the other. Here, it's way the `inner` field below is an option:
  - A None uses the manually written constraints
  - A Some(inner) will delegate to the generated constraints.

But we can certainly simplify the design if we only want one of them.

air/src/lib.rs
Leo-Besancon: The `ProcessorAir` wrapper here has currently two non-critical usages:
- Be an unique interface to switch between the two sets of constraints through the Option (so it is unnecessary if we only keep e.g. the generated code)
- Handle the aux_builder and delegate everything else to the `inner` Air. So it is unecessarty if we instead adapt the air-script codegen to handle them instead. Currently, the `build_aux_trace_with_miden_vm` aux trace builder in air-script does not handle the stateful `AuxTraceBuilder`, we would need to adapt things a bit there, but it's doable.

air/src/lib.rs
Leo-Besancon: No, I think we can keep it commented in the codegen

processor/src/trace/mod.rs
Leo-Besancon: My bad, I did not mean to commit this, I'll remove it. This will be handled in a followup as it needs a bit more design, but basically we will need to handle fixed len public inputs and variable length public inputs separately. Currently they are all returned in `to_public_values()`, which causes some issues when evaluating the constraints.

script.sh
Leo-Besancon: Indeed I think I can remove it and keep it locally for now. I use this to convert from the generated constraints in a more readable form.

air/src/lib.rs
Leo-Besancon: True, I've refactored your first comment in 13203d3da952b3ac4fe0f25f34dae568eebc2fd5. I would like to also put the RPO periodic columns there but they are not padded to len 8 in the exported constants so I can't create a slice from it directly, but it's still cleaner that way.

For your second point, I've created the issue here: https://github.com/0xMiden/p3-miden/issues/14.
I'll need to check when porting the modifications of the upstream PR what would be the impact of this change.

air/src/constraints/stack/field_ops/mod.rs
Leo-Besancon: Indeed, I based the layout on the previous Winterfell constraints, but the stack constraints have not yet been included in the air-script constraints. Removing these files for now, as the final layout may be different.

air/src/constraints/mod.rs
```diff
@@ -6,13 +6,65 @@
...
+    let periodic_values: [_; NUM_PERIODIC_VALUES] =
+        builder.periodic_evals().try_into().expect("Wrong number of periodic values");
+
+    // SYSTEM CONSTRAINTS
+    constraints::enforce_clock_constraint(builder, local, next);
+
+    // STACK CONSTRAINTS
+    //constraints::stack::enforce_stack_boundary_constraints(builder, local);
+    //constraints::stack::enforce_stack_transition_constraint(builder, local, next);
+    //constraints::stack::enforce_stack_bus_constraint(builder, local);
```
Al-Kindi-0: Should we add a comment to explain why these are commented out?
Also worth thinking about a feature-flag-based solution in order to control which constraints are included in evaluation and which ones are not.

air/src/constraints/mod.rs
```diff
@@ -6,13 +6,65 @@
...
 
-use crate::MainTraceRow;
+use crate::{MainTraceRow, NUM_PERIODIC_VALUES};
 
+#[rustfmt::skip]
+pub mod chiplets;
 pub mod range;
 
+/// Enforces all MidenVM constraints
+pub fn enforce_constraints<AB>(builder: &mut AB)
```
Al-Kindi-0: Worth discussing the overall structure of this entry point.
One solution, which I have followed in an upcoming PR, is to use a structure which is VM-component driven, namely put each VM component's constraints in a specific module, with the bus constraints also in the same component module but in a sub-module.

There is also the option of having the above component-based organization but this time we put the main trace constraints in module and put the bus constraints in another module.

Curious about everyone's thoughts on this.

air/src/trace/periodic_columns.rs
```diff
@@ -0,0 +1,59 @@
...
+/// Flag for the first row of each cycle in the periodic column.
+pub const CYCLE_ROW_0: [Felt; 8] = [
+    Felt::new(1),
+    Felt::new(0),
+    Felt::new(0),
+    Felt::new(0),
+    Felt::new(0),
+    Felt::new(0),
+    Felt::new(0),
+    Felt::new(0),
```
Al-Kindi-0: nit: maybe use ZERO and ONE

air/src/unedited_constraints/miden_vm_plonky3.rs
Al-Kindi-0: I would probably keep the `unedited_constraints` very light, especially for this initial PR which should be mainly focused on the organization and future workflow from air-script. This is just a personal suggestion and is up for discussion.

air/src/unedited_constraints/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+//#[rustfmt::skip]
+///// All miden_vm constraints (that have already been written in .air)
+//pub mod miden_vm_plonky3_all;
+//
+//#[allow(clippy::all)]
+//#[allow(unused_imports)]
+//#[allow(unused_variables)]
+//#[rustfmt::skip]
+///// All miden_vm constraints (that have already been written in .air), except 4 RPO related constraints that cause stack-overflow
+//pub mod miden_vm_plonky3_no_crash;
```
Al-Kindi-0: Related to the above point, we could keep this PR light and do some of these in subsequent PRs
This will also allow for the fixes to the air-script constraints, which should be coming soon, to be integrated.

air/src/lib.rs
```diff
@@ -128,48 +142,130 @@ impl Deserializable for PublicInputs {
...
-        // The prover will interpret the returned base field data as EF columns
-        AUX_TRACE_WIDTH
+    fn num_public_values(&self) -> usize {
+        self.inner
+            .as_ref()
+            .map(<MidenVM as p3_miden_air::MidenAir<Felt, EF>>::num_public_values)
+            .unwrap_or(0) // todo
+    }
+
+    fn periodic_table(&self) -> Vec<Vec<Felt>> {
```
Al-Kindi-0, line 244: Would be good to define the periodic columns in their appropriate module and then import/use them here.

Al-Kindi-0 (commented):
Looks good, thanks @Leo-Besancon !
Left some comments which we can discuss.

air/src/constraints/mod.rs
Leo-Besancon: What level of details do you think is useful for the feature flags while still being readable?
For now I'm thinking we can remove the introduced `constraint_eval` feature and replace it with:
- `system_constraints`
- `chiplet_constraints`
- `range_constraints`
- `stack_constraints`
- `decoder_constraints`
- `all_constraints` (that activates all the above flags)
But maybe having the ability to disable/enable bus constraints and specific chiplets could be useful?

air/src/constraints/mod.rs
Leo-Besancon: Thank you for the input, I think I've organized the current code a bit more like your first solution, could you take a look?

air/src/unedited_constraints/mod.rs
Leo-Besancon: > This will also allow for the fixes to the air-script constraints, which should be coming soon, to be integrated.

Great! In the meantime I removed everything that wasn't used, and made a list of things to look / update in a followup. I'll create an issue when I think it's complete.

air/src/lib.rs
Leo-Besancon: Done in the latest commit.

air/src/constraints/mod.rs
Al-Kindi-0: Yes, we could use different flags for different purposes (e.g., human-readable/audited, bus-active, ...).

As for the structure, here is one proposal https://github.com/0xMiden/miden-vm/blob/al-constraints-full/air/src/constraints/mod.rs

air/src/constraints/mod.rs
Al-Kindi-0: Yes, this looks better. We could also add a main vs aux constraints separation as well as mentioned in the above comment.

air/src/unedited_constraints/mod.rs
Al-Kindi-0: Thanks!

air/src/lib.rs
Al-Kindi-0: Thanks!

air/src/trace/chiplets/bitwise.rs
```diff
@@ -55,6 +57,22 @@ pub const PREV_OUTPUT_COL_IDX: usize = B_COL_IDX + 1 + 2 * NUM_DECOMP_BITS;
...
+pub const INV_CYCLE_ROW_7: [Felt; 8] = [ONE, ONE, ONE, ONE, ONE, ONE, ONE, ZERO];
+
+/// The number of periodic columns used in the Bitwise chiplet AIR.
+pub const NUM_BITWISE_PERIODIC_VALUES: usize = 2;
+
+/// Returns the periodic columns used in the Bitwise chiplet AIR.
+pub fn bitwise_periodic_columns() -> Vec<Vec<Felt>> {
+    vec![CYCLE_ROW_0.to_vec(), INV_CYCLE_ROW_7.to_vec()]
+}
+
```
Al-Kindi-0: If these are used only in constraints evaluation, then it might make more sense to move them to the `bitwise` sub-module of the constraints module.

air/src/trace/chiplets/hasher.rs
```diff
@@ -139,3 +140,47 @@ pub const RETURN_STATE_LABEL: u8 = 0b1000 + 1;
 
 /// Index of the auxiliary trace column tracking the state of the sibling table.
 pub const P1_COL_IDX: usize = HASH_KERNEL_VTABLE_AUX_TRACE_OFFSET;
+
+// --- Periodic columns ---------------------------------------------------------------------------
+
```
Al-Kindi-0: Same remark as above

air/src/constraints/mod.rs
Leo-Besancon: Makes sence, this way I can rely on the feature to delegate to the generated code instead of the `inner: Option<MidenVM>` field.

air/src/trace/chiplets/hasher.rs
Leo-Besancon: Moved them (and fixed the RPO round constants as they needed to be transposed)

air/src/constraints/mod.rs
Leo-Besancon: Included this as well in the latest commit!

air/src/constraints/mod.rs
Leo-Besancon: Thanks for the feedback and for the structure. I've begun to include the feature flags, still not 100% happy about them:
- I need some heavy `#[cfg(any(...)]] to match some things, but the alternative includes a lot of feature flags,
- In some case I still need a lot of flags.
I'll try to improve that a bit.

Leo-Besancon:
Note: direct dependencies to p3 have been removed previously, but miden-crypto does not yet expose BusType from p3-miden-air, so I've kept this one dependency.

huitseeker:
Thanks for the contribution @Leo-Besancon – we'll focus on iterating on #2658 (and ideally, transitively on #2659) over this one.

PR #2563, opened 2026-01-09
allow constants in adv_map definition

gejeduck:
### Changes
- Updated `grammer.lalrpop` for `AdviceMap` Form
- Added visit function for `AdviceMapEntry` in `visit.rs`
- Added test function for `adv_map` in `tests.rs`
- Updated `AdviceMapEntry` definition in `advice_map_entry.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Enhancements
 
 - Added `procedure_names` to `DebugInfo` for storing procedure name mappings by MAST root digest, enabling debuggers to resolve human-readable procedure names during execution (#[2474](https://github.com/0xMiden/miden-vm/pull/2474)).
+- Added constants support as an immediate value of the `adv_map` statement ([#2563](https://github.com/0xMiden/miden-vm/pull/2563)).
 
 #### Fixes
 


crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -122,7 +122,9 @@ pub fn analyze(
                 docs.take();
                 analyzer.error(SemanticAnalysisError::UnexpectedEntrypoint { span: body.span() });
             },
-            Form::AdviceMapEntry(entry) => {
+            Form::AdviceMapEntry(mut entry) => {
```
bitwalker, line 125: This isn't the correct place to do const evaluation (i.e. before all of the forms in the module have been initially processed). MASM semantics dictate that the order in which constants are declared does not matter, so long as there are no cycles.

Furthermore, this will break if referencing constants defined in another module, as expansion of those constants is deferred until link-time.

To properly handle this change, the following needs to happen:

1. The `advice_map` field of `Module` should be removed. Instead, we'll maintain a `BTreeSet<Word>` in the analysis context, and use that to check for duplicate advice map keys.
2. `Export` should have a new `AdviceMapEntry` variant added to it.  This will preserve these entries in the AST through to the linker (required for link-time const evaluation).
3. The current `add_advice_map_entry` function needs to be split up, so that the value expansion happens after all forms are processed. The main difference here is that we will still define a constant for the advice map key, and still check for duplicate keys, but rather than updating an advice map for the value, we will simply add the `AdviceMapEntry` to `module.items` similar to the other forms.
4. Now, in `visit_items`, handle the new `Export::AdviceMapEntry` variant similarly to `Export::Constant`, but also, as you've done here, instantiate a `ConstEvalVisitor` and visit the entry.

In the assembler, we will no longer be able to extract the `AdviceMap` structure directly from a `Module`, instead we will need to construct it by visiting all `AdviceMapEntry` items in the module, and adding them to a fresh `AdviceMap`. At the point we are doing this, we will have expected all constants to have been evaluated to concrete values, so we can use `unwrap_value` on the underlying elements of the value vector.

You will necessarily also have to update all places where we pattern match on `Export` variants. For this, you'll have to exercise some judgement on what makes the most sense at each match, but as a general rule:

* If we're looking up a named item in the module, you can ignore the `AdviceMapEntry` variant, as the "name" of the entry is represented by a `Constant` variant in the same module.
* In most other cases, the `AdviceMapEntry` variant can be ignored, with exception of the `VIsit`/`VisitMut` trait implementations, which should be written to properly traverse the AST including this new item variant.
* If there are any places you have questions about, just ask.

crates/assembly/src/tests.rs
```diff
@@ -3028,6 +3028,22 @@ begin adv.has_mapkey assert end"
     Ok(())
 }
 
+#[test]
+fn test_adv_has_map_key_constant() -> TestResult {
```
bitwalker, line 3032: To see the issues I described above, add two new tests:

1. In a copy of this test, move the definition of `CONSTANT` after `A`. This will fail semantic analysis because your implementation will attempt expansion of `CONSTANT` before it has been defined.
2. Create a test which defines `CONSTANT` in a separate module, and then import it into the program module where `A` is defined and used. This will fail semantic analysis because the definition of `CONSTANT` cannot be resolved until link-time.

gejeduck:
@bitwalker , Let me check and fix soon.

crates/assembly-syntax/src/sema/mod.rs
bitwalker: One thing for follow up after this PR: when pretty-printing a `Module`, we currently do not print out advice map entries (but we do print the constants we generate to represent the keys). Not only does this mean pretty-printed modules don't render 1:1 with the original MASM, but they don't even contain all of the items (i.e. the actual advice map entries). This is basically just an artifact of how we're handling advice map entries today though.

After this PR is merged with the changes I described in my other comment - I think we can refactor things to skip generating the keys as separate `Constant` items, and instead resolve references to the keys (as constants) to the corresponding `AdviceMapEntry` item now that we'll have them in the AST. This would make `AdviceMapEntry` items proper items with associated symbols, and bring advice map entries into alignment with other module items.

Issue #2495, opened 2025-12-23
Reorganize processor modules


plafer:
After removing `Process` with #2483, the current module hierarchy should be refactored - some modules are close to empty, names are outdated, etc.

We should wait for the Plonky3 migration to be done though to reduce the amount of merge conflicts.

huitseeker:
@plafer Now we're past the P3 migration, do you want to take this one?

bobbinth:
As discussed offline, let's first merge @Al-Kindi-0's upcoming stack orientation refactoring PR.

plafer:
The high-level view of how I would see the `miden-processor` modules being organized:

- `fast/`: `FastProcessor` and its related concerns (close to what we have today, but cleaned up)
  - e.g. `BlockStack` needs to be moved next to `ExecutionTracer` because that's the only place it's being used
- `trace_generation/`: Everything related to trace generation. That is,
  - `parallel/`: today's parallel trace generation
  - `chiplets/`: the chiplets logic
  - `range/`: the range checker logic
- `processor/`: mostly as-is from today
  - at the same level of `fast/` and `trace_generation/`, since the `Processor` trait it exposes is used by both fast processor and trace generation

Those are the 3 main verticals that were affected by the removal of `Process` and that need  to be cleaned. Most (but not all) of the modules that exist today will end up under one of these (either moved or merged into existing ones); notable exceptions include `host/`, `errors/` and the `Tracer` trait which will remain at the top-level.

bobbinth:
Mostly agree - but what's the difference between `fast` and `processor`?

Also, as a part of this, should probably rename `FastProcessor` into just `Processor` (as we don't have any other processor any more ).

plafer:
> Mostly agree - but what's the difference between fast and processor?

`processor/` contains the `Processor` trait, whereas `fast/` contains the instantiation of the `Processor` trait for the `FastProcessor`. 

Tough as discussed in https://github.com/0xMiden/miden-vm/issues/2591#issuecomment-3799947633, we would rename `FastProcessor` to `Processor` and `Processor` to `AbstractProcessor` (and `fast/` module to `processor/`, etc).

plafer:
Module structure after #2651 LGTM. Only question left is if we want to do the renaming for v0.21? I would suggest:

- `FastProcessor` -> `Processor`
  - `mod fast` -> `mod processor`
- `Processor` (trait) -> `AbstractProcessor`
[...]

PR #2567, opened 2026-01-11
refactor(processor): replace wrapper function with type alias

sashaodessa:
Replaced `invert_column_allow_zeros` wrapper function with a type alias to `batch_inversion_par`, removing redundant code duplication.

### DIFF
--- a/processor/src/utils.rs
+++ b/processor/src/utils.rs
@@ -65,6 +65,11 @@ pub(crate) fn batch_inversion_par(values: &mut [Felt]) {
     );
 }
 
+/// Inverts all values in the provided column in place, leaving zeros unchanged.
+///
+/// This is an alias for [`batch_inversion_par`].
+pub(crate) use self::batch_inversion_par as invert_column_allow_zeros;
+
 /// Montgomery's trick for batch inversion, handling zeros.
 fn batch_inversion_helper(values: &[Felt], result: &mut [Felt]) {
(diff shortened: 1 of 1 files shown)


adr1anh:
It seems cleaner to replace usages of `invert_column_allow_zeros` with `batch_inversion_par`, no?

sashaodessa:
> It seems cleaner to replace usages of `invert_column_allow_zeros` with `batch_inversion_par`, no?

@adr1anh Good point. There’s only one call-site today, so we could replace it with `batch_inversion_par` and drop the alias entirely.
I kept `invert_column_allow_zeros` as an alias mainly for call-site semantics: it makes it explicit that we’re inverting a “column” and that zeros are handled (left unchanged). The alias removes the redundant wrapper body without increasing diff churn.
If you prefer reducing API surface in `utils`, I’m happy to update `parallel/mod.rs` to call `batch_inversion_par` directly and remove the alias.

plafer:
Hey @sashaodessa, thank you for the PR, though I don't think we'll need it. I am planning to remove this module with #2495, and with it I will be reworking those functions.

Please [take a look at our `CONTRIBUTING.md`](https://github.com/0xMiden/miden-vm/blob/next/CONTRIBUTING.md#contribution-quality) for guidance on how to contribute to this repo. Basically, we recommend looking at issues marked `good-first-issue`, and having it be assigned to you before you start working on it.

PR #2547, opened 2026-01-07
Change stack ordering through unified LE convention and sponge state remapping

Al-Kindi-0:
## Describe your changes
Addresses https://github.com/0xMiden/miden-vm/issues/2393

# Little-Endian Convention Migration Guide

This comment describes the breaking changes introduced to adopt consistent little-endian (LE) conventions across the VM. These changes affect instruction semantics, stack layouts, and stdlib procedures.

## Overview

The key principle is **"low limb on top"**: for multi-limb values, the least significant limb is at the lowest stack index (top of stack).

**Before (mixed conventions):**
```
u64 value 0x0001_0002: [hi=1, lo=2, ...]  (hi on top)
```

**After (consistent LE):**
```
u64 value 0x0001_0002: [lo=2, hi=1, ...]  (lo on top)
```

---

#### Output Order Changes

| Instruction | Old Output | New Output |
|------------|------------|------------|
| `u32split` | `[hi, lo, ...]` | `[lo, hi, ...]` |
| `u32overflowing_mul` | `[hi, lo, ...]` | `[lo, hi, ...]` |
| `u32madd` | `[hi, lo, ...]` | `[lo, hi, ...]` |

#### `hperm` (Hash Permutation)

**Old semantics:**
```
Input:  [R1, R0, C, ...] where R0, R1 are rate words, C is capacity
Output: [R1', R0', C', ...]
Digest: elements at indices 4..8
```

**New semantics:**
```
Input:  [R0, R1, C, ...]  where R0, R1 are rate words, C is capacity
        R0 = [r0, r1, r2, r3] with r0 at top
        R1 = [r4, r5, r6, r7]
        C  = [c0, c1, c2, c3]
Output: [R0', R1', C', ...]
Digest: elements at indices 0..4 (RATE0)
```

**Migration:**
- `squeeze_digest` is now equivalent to `swapw.2 dropw dropw`
- Digest is in the first rate word (R0), not the second

#### `hmerge` (2-to-1 Hash)

**Old:** `[B, A, ...]` → `[hash(A, B), ...]`
**New:** `[A, B, ...]` → `[hash(A || B), ...]`

The first word (A) is absorbed first, matching the stack order.

#### `mem_stream` and `adv_pipe`

**Old semantics:**
```
Input:  [C, B, A, addr, ...]
Output: [E, D, A, addr', ...]
```

**New semantics:**
```
Input:  [A, B, C, addr, ...]
Output: [D, E, C, addr', ...]
```

[...]

PR #2579, opened 2026-01-13
fix(processor): remove duplicate stack layout logic in test_op_dup

anim001k:
Removes duplicate logic for building expected stack state in `test_op_dup` by using the existing `build_expected()` helper function instead of manual array reversal.

The test manually constructed the expected stack layout using array reversal (`expected_arr.iter().rev().cloned().collect()`), duplicating the knowledge about how `stack_top()` represents the stack (top at the last index). This same logic is already centralized in the `build_expected()` helper used throughout the test file.


plafer:
We don't accept simple PRs like this (see `CONTRIBUTING.md`). And most of this is changing with #2547.

PR #2580, opened 2026-01-13
fix: use Windows-compatible snapshot file names

### LINKED ISSUE #2561: Windows incompatibility: Snapshot file names contain reserved characters (::)
Snapshot test files in `processor/src/fast/tests/snapshots/` contain `::` characters in their file names, which are reserved characters on Windows.

This prevents Windows users from checking out the repository without using sparse checkout workarounds.

Should we change `::` to `__` in snapshot file names, or is the current behavior acceptable?

Himess:
## Summary

Fixes #2561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - [BREAKING] Removed `SyncHost` and `BaseHost`, and renamed `AsyncHost` to `Host` ([#2595](https://github.com/0xMiden/miden-vm/pull/2595)).
 - [BREAKING] Moved `ExecutionOptions` to `miden-processor`, `ProvingOptions` to `miden-prove`, and `ExecutionProof` to `miden-core` (all out of `miden-air`) ([#2597](https://github.com/0xMiden/miden-vm/pull/2597)).
 - [BREAKING] Removed `on_assert_failed` method from `Host` trait ([#2600](https://github.com/0xMiden/miden-vm/pull/2600)).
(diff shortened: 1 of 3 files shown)


huitseeker (changes requested):
The fix looks good for the fast processor tests, but it looks like `processor/src/parallel/tests.rs` has the same `testname()` pattern at line 707 that wasn't updated. The parallel sna
pshots directory still has files with `::` in their names (e.g. `miden_processor__parallel__tests__parallel::tests::test_trace_generation_at_fragment_boundaries::case_01.snap`).

These should be included in this PR as well.

Himess:
@huitseeker Thanks for review
 The parallel files are included in this commit. GitHub's UI only displays the first 300 of the 1238 changed files. 

You can verify the parallel snapshots are renamed by checking the commit tree:
  `gh api 'repos/0xMiden/miden-vm/contents/processor/src/parallel/snapshots?ref=fix/windows-compatible-snapshot-names' --jq '.[0].name'`
  This shows `miden_processor__parallel__tests__parallel__tests__test_trace... with __ instead of ::.`

plafer:
Hey @Himess I think this is good to go after the conflicts are resolved

plafer:
I just noticed that `jj` (and maybe git?) also fails to parse these files as a valid fileset expression - so definitely a welcome change.

Himess:
Hi @plafer , so sorry for late reply and  force-push.
Resolved conflicts , ready for review.

PR #2583, opened 2026-01-15
feat(test-utils): add MaliciousAdviceHandler for testing MASM validation

### LINKED ISSUE #2288: test stdlib against malicious `AdviceProvider`
### What should be done?

MASM implementation of `sorted_array::find_word` or `smt::set` needs to validate data coming from the advice provider. However, currently there is no way to test if these validations are correct.

It would be useful to be able to inject malicious advice data into MASM code and ensure it behaves as expected.

This would be useful here:
* https://github.com/0xMiden/miden-vm/pull/2248#pullrequestreview-3313018992
* https://github.com/0xMiden/miden-vm/pull/2114/files#diff-fc89d3ec45d98dc9c5a4aedd201a4e2a5dee59b1a623f688808bbe99345b841dR33-R36

### How should it be done?

Allow unit tests to inject arbitrary values when the AdviceProvider is called.

### When is this task done?

MASM asserts validating non-deterministic data can be exercised.

### Additional context

_No response_

Himess:
Closes #2288

### DIFF
--- a/processor/README.md
+++ b/processor/README.md
@@ -2,23 +2,45 @@
 This crate contains an implementation of Miden VM processor. The purpose of the processor is to execute a program and to generate a program execution trace. This trace is then used by Miden VM to generate a proof of correct execution of the program.
 
 ## Usage
-The processor exposes the `execute()` function which takes the following arguments:
+The processor provides multiple APIs depending on your use case:
+
+### High-level API
+The `execute()` function provides a convenient interface that executes a program and generates a complete execution trace:
 
 * `program: &Program` - a reference to a Miden program to be executed.
 * `stack_inputs: StackInputs` - a set of public inputs with which to execute the program.
-* `host: Host` - an instance of a `Host` which can be used to supply non-deterministic inputs to the VM and receive messages from the VM.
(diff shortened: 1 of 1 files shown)


crates/lib/core/tests/collections/sorted_array.rs
```diff
@@ -259,3 +262,182 @@ fn test_misaligned_key_value_find_key_fails() {
...
+
+        begin
+            # Store sorted array in memory
+            push.[8456,415,4922,593] mem_storew_le.100 dropw
+
+            # Setup: [KEY, start_ptr, end_ptr]
+            push.104 push.100 push.[8456,415,4922,593]
+
+            # Inject malicious advice data
+            emit.event(\"{MALICIOUS_ADVICE_EVENT}\")
```
plafer: Trying to remember why we needed this in the first place. Here, instead of emitting an event to populate the advice stack, why don't we just initialize the advice stack at the beginning of the test with the malicious values we want?

Himess: You are %100 right.
Simplified it to just initialize the advice stack at test start.

PR #2585, opened 2026-01-16
refactor(tests): convert crypto ops tests to proptest

### LINKED ISSUE #2573: Convert crypto ops integration tests to proptests
Tests in `miden-vm/tests/integration/operations/crypto_ops.rs` still use `rand_vector::()` and such. We should instead convert them to proptests (if not only for better reproducibility).

Related: https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2676209522

CreeptoGengar:
Convert tests using rand_vector() to proptest for better reproducibility.
This ensures deterministic test execution with seed-based generation.

- Convert hash(), hperm(), and hmerge() tests to use proptest
- Remove unused rand_vector import

Fixes #2573

### DIFF
--- a/miden-vm/tests/integration/operations/crypto_ops.rs
+++ b/miden-vm/tests/integration/operations/crypto_ops.rs
@@ -344,7 +344,10 @@ fn crypto_stream_rejects_in_place() {
     let err = test.execute().expect_err("crypto_stream should reject in-place encryption");
     assert!(matches!(
         err,
-        ExecutionError::MemoryError(MemoryError::IllegalMemoryAccess { .. })
+        ExecutionError::MemoryError {
+            err: MemoryError::IllegalMemoryAccess { .. },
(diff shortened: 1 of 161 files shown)


miden-vm/tests/integration/operations/crypto_ops.rs
```diff
@@ -103,6 +116,31 @@ fn hmerge() {
     assert_eq!(expected_stack_slice, &last_state[4..8]);
 }
 
+proptest! {
+    #[test]
+    fn hmerge_proptest(
```
huitseeker, line 121: minor: there's a `prop_randw::<u64>()` helper in `miden_utils_testing` that generates 4-element vectors, though for 8 values you'd need to combine two of them which is arguably more awkward than this explicit approach. just mentioning it in case you weren't aware of it.

Issue #2572, opened 2026-01-12
Remove `push.a.b.c.d` syntax


plafer:
_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2673105646_

Similar to `adv_push.4` vs `adv_loadw`, I find having both `push.a.b.c.d` and `push.[a,b,c,d]` (that do different things) confusing. 

**I think we should remove the `push.a.b.c.d` syntax to remove the possible confusion.**

Looking at `miden-base` (excluding tests), all but one uses of that syntax is `push.0.0` or `push.0.0.0`. Those would be rewritten as:
- `push.0.0` -> `push.0 push.0`
- `push.0.0.0`: ~~`padw drop` (also resulting in 1 fewer operation)~~
  - **correction:** `padw` expands to 4 `Pad` operations, so you would end up writing `push.0 push.0 push.0`

The last use is `push.0.1`, which you would just expand to `push.0 push.1`. IMO the potential additional confusion in how words should be pushed is not worth the minor syntactic "benefit" that `push.0.0` expands to `push.0 push.0`.

If there's agreement, we should keep this for a subsequent PR.

thoughts? @PhilippGackstatter @Fumuran @partylikeits1983 @mmagician

partylikeits1983:
I do find it helpful to occasionally use `push.0.0.0`, but I think it's more clear to have this instead `push.0. push.0 push.0`. I don't have a strong opinion either way (keeping current syntax or removing it).

bitwalker:
It isn't super clear to me that the dot-delimited syntax is redundant with `push.[a, b, c, d]`, except in the very specific case of `push.a.b.c.d` (arguably) - but those aren't the same thing, i.e. the former is explicitly pushing a word literal, where the latter is sugar for `push.a push.b push.c push.d` - i.e. you actually want `d` on top in that case.

As @partylikeits1983 mentioned, it can be used to push any number of elements sequentially without the visual noise of multiple `push.IMM` stacked together.

~If we do a quick scan to see how frequently uses of the dot-delimited syntax are with numbers of operands other than 4, and we find that it is rare, then I'd be more inclined to remove the syntax. Otherwise, it is low cost sugar to maintain IMO.~

I just noticed that you did a skim to find frequency of use, and it is relatively uncommon. That said, it doesn't appear that the dot-delimited syntax is actually problematic in any way. I'm OK with whatever the majority decides here.

[...]

PR #2586, opened 2026-01-16
docs: document `push.[a,b,c,d]` word literal syntax

### LINKED ISSUE #2556: Document `push.[a,b,c,d]` syntax
The `push.[a,b,c,d]` syntax (which pushes a word on the stack) was never documented.

It would be good to document after #2547, since in that PR the order in which the word is pushed changes to pushing the elements such that `a` sits on top of the stack.

sky-coderay:
Document the `push.[a,b,c,d]` word literal syntax that was previously undocumented.

This syntax pushes a word (4 field elements) onto the stack such that the first element ends up on top.

Closes #2556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,13 +51,17 @@
 - [BREAKING] Convert `ProcessState` to a struct wrapping `FastProcessor`, and rename it to `ProcessorState` ([#2604](https://github.com/0xMiden/miden-vm/pull/2604)).
 - [BREAKING] Cleaned up `StackInputs` and `StackOutputs` API, and use `StackInputs` in `FastProcessor` constructors ([#2605](https://github.com/0xMiden/miden-vm/pull/2605)).
 
-## 0.20.2 (TBD)
+## 0.20.3 (2026-01-27)
 
-- Fix issue where decorator access was not bypassed properly in release mode ([#2529](https://github.com/0xMiden/miden-vm/pull/2529)).
+- Fixed issue where exports of a Library did not have attributes serialized [#2608](https://github.com/0xMiden/miden-vm/issues/2608)
+
+## 0.20.2 (2026-01-05)
+
+- Fixed issue where decorator access was not bypassed properly in release mode ([#2529](https://github.com/0xMiden/miden-vm/pull/2529)).
 
 ## 0.20.1 (2025-12-14)
 
-- Fix issue where calling procedures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).
+- Fixed issue where calling procedures from statically linked libraries did not import their decorators ([#2459](https://github.com/0xMiden/miden-vm/pull/2459)).
 
 ## 0.20.0 (2025-12-05)
 
--- a/crates/assembly-syntax/src/library/mod.rs
+++ b/crates/assembly-syntax/src/library/mod.rs
@@ -605,11 +605,12 @@ impl Deserializable for Library {
                     } else {
                         None
                     };
+                    let attributes = AttributeSet::read_from(source)?;
                     LibraryExport::Procedure(ProcedureExport {
                         node,
                         path: path.clone(),
                         signature,
-                        attributes: Default::default(),
+                        attributes,
                     })
                 },
                 1 => {
@@ -775,7 +776,7 @@ impl Serializable for LibraryExport {
                 node,
                 path: name,
                 signature,
-                attributes: _,
+                attributes,
             }) => {
                 target.write_u8(0);
                 name.write_into(target);
@@ -786,6 +787,7 @@ impl Serializable for LibraryExport {
                 } else {
                     target.write_bool(false);
                 }
+                attributes.write_into(target);
             },
             LibraryExport::Constant(ConstantExport { path: name, value }) => {
                 target.write_u8(1);


docs/src/user_docs/assembly/io_operations.md
```diff
@@ -28,6 +29,27 @@ push.4660.22136.36882.43981
...
+#### Word literal syntax
+
+The `push.[a,b,c,d]` syntax provides a convenient way to push a word (4 field elements) onto the stack. The elements are pushed such that the first element `a` ends up on top of the stack:
+
+```
+push.[1,2,3,4]   # Results in stack: [1, 2, 3, 4, ...]
+                 # where 1 is on top of the stack
+```
+
+This is equivalent to `push.4.3.2.1` but provides a more intuitive syntax when working with words, as the element order in the literal matches the resulting stack order (first element on top).
```
plafer:
```suggestion
This is equivalent to `push.4 push.3 push.2 push.1` but provides a more intuitive syntax when working with words, as the element order in the literal matches the resulting stack order (first element on top).
```
Good idea, but let's compare it to separate `push` statements, since the `push.4.3.2.1` will probably go away with #2572

docs/src/user_docs/assembly/io_operations.md
```diff
@@ -28,6 +29,27 @@ push.4660.22136.36882.43981
...
+
+You can also use slices with word constants to push only a portion of the word:
+
+```
+const WORD = [5,6,7,8]
+
+push.WORD[0]      # is equivalent to push.5
+push.WORD[1..3]   # is equivalent to push.6.7, with 6 on top
```
plafer:
```suggestion
push.WORD[1..3]   # is equivalent to `push.6 push.7`
```

plafer (approved):
Awesome, thank you! This will be good to merge after small suggestions were addressed, and merge conflict is fixed

plafer:
@sky-coderay The commit [05ecf03](https://github.com/0xMiden/miden-vm/pull/2586/commits/05ecf0395ae719bb953e8645ce971d79b9ff4cb8) is not signed (and hence we can't merge) - can you sign it and re-push it?

docs/src/user_docs/assembly/io_operations.md
```diff
@@ -28,6 +29,27 @@ push.4660.22136.36882.43981
...
+This is equivalent to `push.4 push.3 push.2 push.1` but provides a more intuitive syntax when working with words, as the element order in the literal matches the resulting stack order (first element on top).
+
+You can also use slices with word constants to push only a portion of the word:
+
+```
+const WORD = [5,6,7,8]
+
+push.WORD[0]      # is equivalent to push.5
+push.WORD[1..3]   # is equivalent to `push.6 push.7`
+push.WORD[0..4]   # is equivalent to push.[5,6,7,8]
```
bobbinth, line 50: Is `push.WORD[1..3]` document correctly here? Pushing slices should work the same way as pushing words, but just in truncated form. Specifically:

```
const WORD = [5,6,7,8]

push.WORD[1..3]
# => [6, 7, ...]
```
That is, `push.WORD[1..3]` should be equivalent to ``push.7 push.6`. I thought that's how it already worked. If so, we should update the docs. If that's not the case, we should make a PR to fix this because we rely on such semantics quite frequently.

plafer: Ues it is correctly documented - see [analogous comment](https://github.com/0xMiden/miden-vm/blob/ab1aab09c57cac4427ce9cd3f752050b55b1f5db/crates/assembly/src/instruction/env_ops.rs#L63-L64) in the assembler.

Recall that `push.WORD` places `WORD` on the stack as `[5, 6, 7, 8]` (5 on top). So it makes sense that `push.WORD[1..3]` pushes `[6, 7]`, with `6` on top.

You might be thinking in terms of the "old" `push.a.b.c.d` syntax (which is also why I'm in favor of removing it #2572)?

bobbinth: I think the documentation in assembler is correct, but I don't think it is correct here. Specifically, for this example:
```
push.WORD[1..3]
# => [6, 7, ...] # 6 on top
```
But
```
push.6 push.7
# => [7, 6, ...] # 7 on top
```

plafer: Bleh, sorry. You're right.

@sky-coderay can you update the comment to:

```
push.WORD[1..3]   # is equivalent to `push.7 push.6`
```

sky-coderay: Updated, thanks

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -277,7 +290,7 @@ const SAMPLE_WORD = [5,6,7,8]
 const SAMPLE_VALUE = 9
 
 begin
-    push.SAMPLE_WORD[1..3]  # is equivalent to push.6.7
+    push.SAMPLE_WORD[1..3]  # is equivalent to push.6.7, with 6 on top
```
bobbinth: I think the comment here needs to be updated too - i.e., `push.6.7` puts `7` on top AFAIU - so, this should be `push.7.6` - but that's probably confusing - so, i'd do `push.7 push.6`

Issue #2252, opened 2025-10-02
Implement serialization of `TraceGenerationContext`


plafer:
As a follow-up to #2188, we need to implement `Serializable` and `Deserializable` (+ serde's as well) for `TraceGenerationContext`

bobbinth:
One of the questions to answer here is how to serialize the MAST forest information required by the trace generator. Broadly speaking, I think we have 3 options:

- Serialize all MAST forests that were touched during execution. This is probably not a good idea.
- Serialize all procedures that were touched during execution. This could work - though, not sure how easy it is to do.
- Serialize only the MAST nodes that were touched during execution. This will probably require some changes in how the reply structs capture MAST info.

plafer:
Superseded by #2730

PR #2587, opened 2026-01-16
Remove `NodeExecutionState` in favor of `Continuation`

### LINKED ISSUE #2444: Remove `NodeExecutionState` from `CoreTraceFragmentContext`
#2440 expanded the `Continuation` to be able to encode everything that we use the `NodeExecutionState` for in the `CoreTraceFragmentContext`. That is, we should just use the `CoreTraceFragmentContext.continuation` field to encode "what to execute next", as opposed to the `continuation` field *and* the `initial_execution_state`.

This will simplify the parallel trace generation logic, as well as remove the currently `unimplemented!()` continuations (added in #2440).

plafer:
Closes #2444 

This is both a cleanup PR, along with a bug fix that was discovered during the implementation. I merged the bug fix with the cleanup, since removing `NodeExecutionState` was one way to fix the bug, and that was what the initial purpose of the PR was.

## The cleanup

The main goal of this PR is to remove the duplicate logic at the start of `CoreTraceFragmentFiller::fill_fragment_impl()`, which first handled the `NodeExecutionState` associated with the fragment context (e.g. if at a `REPEAT` point, first insert a `REPEAT` trace row, then re-execute the corresponding loop body). However, the `NodeExecutionState` handling logic was very similar to the `Continuation` handling logic - since fundamentally, a `NodeExecutionState` really encodes a continuation. Hence, the cleanup was about using continuations where `NodeExecutionState`s were used. This change also meant that we now properly implement everywhere that was left `unimplemented!()` in `CoreTraceFragmentFiller::fill_fragment_impl()` (leftover from #2440).

## The bug fix

The `NodeExecutionState::End` for loops didn't properly encode the difference between when a loop was entered (in which we need to `decrement_stack_size()`), and when the loop was not entered (in which we *don't* decrement stack size). The previous implementation always decremented the stack size.

We had a test that was supposed to test this (`parallel::tests::test_trace_generation_at_fragment_boundaries::case_08`), but the stack in the test was all `ZERO`s, and hence it didn't catch the fact that we were running an extra `decrement_stack_size()`. The first commit fixes this test by adding a sentinel value, so that the stack is not all 0's - and hence there is now a difference between dropping and not dropping.

As mentioned above, the fix happened naturally by switching over to using `Continuation` to encode the current state of execution at the start of the fragment. We merged `Continuation::FinishLoopUnentered` into `Continuation::FinishLoop` (by parametrizing `FinishLoop` with an extra `was_entered: bool` parameter).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - `FastProcessor` now correctly only executes `trace` decorators when tracing is enabled (with `ExecutionOptions`) ([#2539](https://github.com/0xMiden/miden-vm/pull/2539))
 - Removed the intentional HALT-insertion bug from the parallel trace generation ([#2484](https://github.com/0xMiden/miden-vm/pull/2484))
 - Added missing `as_canonical_u64()` method to `IntValue` in `miden-assembly-syntax`, fixing compilation errors in the generated grammar code ([#2589](https://github.com/0xMiden/miden-vm/pull/2589))
+- Fixed a bug where trace generation would fail if a core trace fragment started on the `END` operation of a loop that was not entered ([#2587](https://github.com/0xMiden/miden-vm/pull/2587)).
 
 #### Changes
 
@@ -38,6 +39,7 @@
 - Removed `ErrorContext` trait and `err_ctx!` macro; error context is now computed lazily by passing raw parameters to error extension traits ([#2544](https://github.com/0xMiden/miden-vm/pull/2544)).
 
 - Use `IndexVec::try_from` instead of pushing elements one by one in `DebugInfo::empty_for_nodes` ([#2559](https://github.com/0xMiden/miden-vm/pull/2559)).
+- [BREAKING] Remove `NodeExecutionState` in favor of `Continuation` ([#2587](https://github.com/0xMiden/miden-vm/pull/2587)).
 
 ## 0.20.2 (TBD)
 - Fix issue where decorator access was not bypassed properly in release mode ([#2529](https://github.com/0xMiden/miden-vm/pull/2529)).


processor/src/parallel/core_trace_fragment/execution.rs
```diff
@@ -68,18 +67,19 @@ impl<'a> CoreTraceFragmentFiller<'a> {
...
+        was_entered: bool,
     ) -> ControlFlow<()> {
         let loop_node =
             mast_forest.get_node_by_id(node_id).expect("node should exist").unwrap_loop();
 
-        // If no condition is provided, read it from the stack
-        let mut condition = if let Some(cond) = condition {
-            cond
-        } else {
+        // If the loop was never entered, we know the condition is ZERO. Otherwise, read it from the
```
huitseeker, line 75: This trusts the `was_entered` flag without checking — even if today it happens to be set correctly. If the flag is wrong (bugs, corruption), we could read the wrong stack slot or skip a drop we should do.

Is it worth adding a check like `if was_entered && self.depth() < MIN_STACK_DEPTH` to catch if the flag says "entered" but there's no condition on the stack to read, for defensive programming?

(If we really are paranoid, we could also check that we're not about to drop a sentinel value or that stack depth is what we expect given loop iteration count)

plafer: You're right that we fully trust the input - though I'm not sure if there's anything way to detect whether or not this flag is properly set. For example, I don't think `if was_entered && self.depth() < MIN_STACK_DEPTH` really helps, since it is always true that `self.depth() >= MIN_STACK_DEPTH` (and if that doesn't work well, then we have much bigger problems). And if it was the case we had a bad bug and `self.depth() < MIN_STACK_DEPTH`, then I don't see how this check would necessarily indicate that the loop was entered?

As we discussed a few times before (notably when discussing whether or not we should do #2252), we currently assume in many other places that the `TraceFragmentContext` is properly constructed (i.e. by our `ExecutionTracer`) - i.e. by us, and without bugs. There might be a better way to do it? I will merge the PR now, but please open an issue to discuss this further if I missed anything, or if you have further ideas on this.

Issue #2473, opened 2025-12-15
Simplify MastForestBuilder with fingerprint-based approach


huitseeker:
`MastForestBuilder` is hard to work with. When copying nodes from statically-linked libraries, we must:
1. Walk the subtree to collect all decorators
2. Build a decorator ID map (source → target)
3. Walk again to copy nodes with remapped IDs

This two-phase approach is error-prone. Bug #2456 happened because decorator IDs weren't remapped.

The root cause: node IDs and decorator IDs are position-dependent. They change when moving nodes between forests.

## Proposed Solution

Stop using `MastForest` during assembly. Use fingerprints instead of IDs.

**Replace:**
```rust
pub struct MastForestBuilder {
    mast_forest: MastForest,
    node_id_by_fingerprint: BTreeMap<MastNodeFingerprint, MastNodeId>,
    statically_linked_mast_remapping: BTreeMap<MastNodeId, MastNodeId>,
    statically_linked_decorator_remapping: BTreeMap<DecoratorId, DecoratorId>,
}
```

**With:**
```rust
pub struct MastForestBuilder {
    nodes: BTreeMap<MastNodeFingerprint, MastNodeBuilder>,
}
```

**Node builders become:**
```rust
pub struct JoinNodeBuilder {
    children: [MastNodeFingerprint; 2],
    before_enter: Vec<DecoratorId>,  // or Vec<Decorator>?
    after_exit: Vec<DecoratorId>,
}
```

Convert to `MastForest` only when assembly finishes.

Fingerprints don't change, so we copy nodes by copying their fingerprints, no remapping needed.
Assembly time isn't critical, so it's acceptable if < 5x slowdown.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2459#discussion_r2617516311_

bobbinth:
cc @bitwalker

huitseeker:
## PRD

### Problem

See [issue #2473](https://github.com/0xMiden/miden-vm/issues/2473) for the problem statement. In
short, `MastForestBuilder` needs an assembly-time identity model that does not depend on positional
IDs from a live `MastForest`.

### Goals

- Replace assembly-time positional identity with stable builder-local references.
- Remove source-to-target node and decorator remapping from static MAST imports.
- Assign final `MastNodeId` and decorator/debug IDs exactly once during final forest construction.
- Keep finalized `MastForest` storage dense and vector-backed for fast iteration.
[...]

PR #2590, opened 2026-01-18
feat: Untrusted deserialization of the MastForest

### LINKED ISSUE #2445: Introduce Trust Boundary in MastForest Deserialization
## Problem

`MastForest` deserialization trusts the bytes it reads. It accepts digests and hashes from the input without checking them. This means malicious or corrupted data can trick the VM into using wrong cryptographic values.

The current code uses `.with_digest()` on all node builders during deserialization ([`MastForestContributor` trait](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/node/mast_forest_contributor.rs#L53-L58)). This skips the normal hash calculation and uses whatever digest was in the serialized bytes.

Example from [`join_node.rs:351-359`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/node/join_node.rs#L351-L359):
```rust
let digest = if let Some(forced_digest) = self.digest {
    forced_digest  // Trusted from serialization - no check!
} else {
    // Normal path: compute from children
    hasher::merge_in_domain(&[left_child_hash, right_child_hash], JoinNode::DOMAIN)
};
```

The serialization code even has an explicit comment about this ([`decorator.rs:50-52`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/serialization/decorator.rs#L50-L52)): *"Note that in this implementation, we trust the encoder"*

## What We Need

Create an untrusted deserialization path that:
- Re-computes all node digests from structure and children
- Resists stack overflows from deeply nested nodes
- Never panics on bad input

The trusted path can stay for internal use where we know the source is good.

## Winterfell Trait Migration

We currently use `Serializable`/`Deserializable` from `winter-utils`. Since we're removing the winterfell dependency, we need to bring these traits into our codebase as part of this work.

## Code Pointers

**Deserialization entry point:**
- [`core/src/mast/serialization/mod.rs:200-308`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/serialization/mod.rs#L200-L308)

**Node builders using forced digests:**
- [`core/src/mast/node/basic_block_node/mod.rs:1438-1441`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/node/basic_block_node/mod.rs#L1438-L1441)
- [`core/src/mast/node/join_node.rs:469-472`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/node/join_node.rs#L469-L472)
- Similar in `split_node.rs`, `loop_node.rs`, `call_node.rs`, `dyn_node.rs`, `external.rs`

**Relaxed validation for deserialization:**
- `add_to_forest_relaxed()` methods in each node type (e.g., [`join_node.rs:485-516`](https://github.com/0xMiden/miden-vm/blob/next/core/src/mast/node/join_node.rs#L485-L516))

## Related PRs

- #2071 - Introduced serde support (delegates to `Serializable`/`Deserializable` anyway, so not useful here)
- #2412 - Added MastForest internal validation (good foundation for this work)

## Open Questions

@bobbinth - You might have a clear view in cache on where should deserialization operate in trusted mode vs untrusted mode? 

Besides, do we agree we should, from an API standpoint, make untrusted the default and keep trusted as an internal optimization?

What resource limits make sense for untrusted deserialization (max depth, max nodes)?

huitseeker:
Edit : ~blocked on https://github.com/0xMiden/crypto/issues/797~ ~`miden-crypto` v0.21.4 (for https://github.com/0xMiden/crypto/pull/798 )~ Ready!

----

Adds `UntrustedMastForest`, a wrapper for safely deserializing `MastForest` from untrusted sources. Also introduces fuzzing infrastructure that has caught one bug.

### Changes

- **`UntrustedMastForest`**: New type that wraps deserialized forests and requires explicit validation before use. Validation recomputes all node hashes and checks topological ordering.
- **Bounds checks**: Reject excessive `node_count` early to prevent allocation bombs. Fix out-of-bounds access in `BasicBlockData` deserialization (found by fuzzing).
- **Fuzzing**: New `miden-core-fuzz` crate with 5 fuzz targets covering deserialization and validation paths.

### Usage

```rust
// Untrusted input (network, user upload, etc.)
let forest = UntrustedMastForest::read_from_bytes(&bytes)?
    .validate()?;

// Trusted input (compiled locally)
let forest = MastForest::read_from_bytes(&bytes)?;
```

### New error variants

- `MastForestError::ForwardReference` - node references a child that appears later in the forest
- `MastForestError::HashMismatch` - recomputed digest doesn't match stored digest

### Fuzzing

```bash
make fuzz-seeds          # generate seed corpus
make fuzz-mast-forest    # fuzz deserialization
make fuzz-mast-validate  # fuzz validation
```

Closes #2445


core/src/mast/mod.rs
```diff
@@ -1091,3 +1232,137 @@ impl<'de> serde::Deserialize<'de> for MastForest {
...
+    /// ```ignore
+    /// // Read from untrusted source
+    /// let untrusted = UntrustedMastForest::read_from_bytes(&bytes)?;
+    ///
+    /// // Validate before use
+    /// let forest = untrusted.validate()?;
+    /// ```
+    pub fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
+        // Use 4× budget to account for allocation overhead during deserialization
+        Self::read_from_bytes_with_budget(bytes, bytes.len().saturating_mul(4))
```
plafer: I don't understand the "4x" part. In my understanding, the budget is the number of bytes that are allowed to be read from the `bytes` buffer. For a legitimate deserialization, can you give an example of a case where more than `bytes.len()` needs to be read from the buffer?

core/src/mast/mod.rs
```diff
@@ -1091,3 +1232,137 @@ impl<'de> serde::Deserialize<'de> for MastForest {
...
+    ///
+    /// // Validate before use
+    /// let forest = untrusted.validate()?;
+    /// ```
+    ///
+    /// # Security
+    ///
+    /// The budget limits:
+    /// - Pre-allocation sizes when deserializing collections (via `max_alloc`)
+    /// - Total bytes consumed during deserialization
```
plafer, line 1355: In practice, we would probably have an upper bound over which we don't deserialize because we consider it an attack (e.g. 1 GB or less), right? And this upper bound would be checked outside of this deserializer - if larger than our max, then we just don't even try deserializing? Just making sure I'm following how this is meant to be used.

core/src/mast/serialization/mod.rs
```diff
@@ -298,6 +308,41 @@ where
...
+    /// After deserialization, callers should use [`super::UntrustedMastForest::validate()`]
+    /// to verify structural integrity and recompute all node hashes before using
+    /// the forest.
+    fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
+        // Use 4× budget to account for allocation overhead during deserialization
+        let mut reader = crate::utils::BudgetedReader::new(
+            crate::utils::SliceReader::new(bytes),
+            bytes.len().saturating_mul(4),
+        );
+        Self::read_from(&mut reader)
```
plafer:
```suggestion
        Self::read_from_bytes_with_budget(bytes, bytes.len().saturating_mul(4))
```

Or `Self::read_from_bytes(bytes)`

core/src/mast/serialization/mod.rs
```diff
@@ -298,6 +308,41 @@ where
...
+    /// Note: This method does not apply budgeting. For untrusted input, prefer using
+    /// [`read_from_bytes`](Self::read_from_bytes) which applies budgeted deserialization.
+    ///
+    /// After deserialization, callers should use [`super::UntrustedMastForest::validate()`]
+    /// to verify structural integrity and recompute all node hashes before using
+    /// the forest.
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let forest = MastForest::read_from(source)?;
+        Ok(super::UntrustedMastForest(forest))
+    }
```
plafer, line 326: From https://github.com/0xMiden/miden-vm/issues/2445#issuecomment-3717985514 I gathered that we want the serialization format to be as compact as possible (presumably by not serializing the digests?) - though here the "untrusted format" uses the same format as the "trusted" format. Is this intended? Or are we going to address this in a future PR?

core/src/mast/mod.rs
huitseeker: Yeah I'm not sure about the number either, and it's mostly a reminder to myself to get my thinking straight on this.

You're right that the budget is mostly a cap on the number of read bytes, and that in that respect, there's no sense in reading more bytes than there are in the serialized data. However we also use this budget as a cap on memory allocation in https://github.com/0xMiden/crypto/pull/757: before allocating, `read_many_iter` checks `requested_count <= remaining_budget / min_serialized_size()`. 

That secondary function could use a bit of breathing room — though doing it through over-sizing the budget is probably not the right way to do this. I'll change this — just need to think about it for a while.

core/src/mast/mod.rs
huitseeker: > we would probably have an upper bound over which we don't deserialize because we consider it an attack (e.g. 1 GB or less), right? 

That would make the most sense, yes, and with limits much lower than 1GB at that. 

But a serialized payload of tiny size can adversarially request a humongous allocation (beyond the OOM limit), so we're trying to think about capping allocations as well.

core/src/mast/serialization/mod.rs
huitseeker: > though here the "untrusted format" uses the same format as the "trusted" format. Is this intended? 

Yes, it is intended, the distinct serialization format is out of scope for this PR. A format w/o hashes is TBD, and requires some design work, e.g. around `ExternalNode`.

core/src/mast/mod.rs
huitseeker: Figured it out: https://github.com/0xMiden/crypto/issues/797

huitseeker: back to 1x!

plafer: Did you push? I still see

```rs
// Use 4× budget to account for allocation overhead during deserialization
Self::read_from_bytes_with_budget(bytes, bytes.len().saturating_mul(4))
```

core/src/mast/mod.rs
```diff
@@ -666,6 +695,108 @@ impl MastForest {
...
+        /// Checks that child_id references a node that appears before node_id in topological order.
+        fn check_no_forward_ref(
+            node_id: MastNodeId,
+            child_id: MastNodeId,
+        ) -> Result<(), MastForestError> {
+            if child_id.0 >= node_id.0 {
+                return Err(MastForestError::ForwardReference(node_id, child_id));
+            }
+            Ok(())
+        }
```
bobbinth, line 724: I'm trying to remember if this is an actual invariant that we enforce in the `MastForest`. I remember we had some discussions about this, and I was advocating for enforcing this invariant - but I don't remember where we've landed.

bobbinth (approved):
Thank you! Looks good! I left one small question inline - but I think we can merge this as is.

This gets us to a good place, but I do think we should have a separate serialization format for un-trusted MAST forest. The main motivation here is to avoid serializing hashes which we know will need to be re-computed anyways. But before going this way, we should probably try to do Q&D benchmark to see how much savings we'd get from such more compact serialization.

core/src/mast/mod.rs
huitseeker: To answer your question about _enforcing_ that invariant (and not the neighbouring question about maintaining that invariant, which is a thing the assembler does at creation & merging): we don't, hence why I add that check here.

Deserialization paths (e.g. `add_to_forest_relaxed` in nodes like `core/src/mast/node/join_node.rs`) intentionally allow forward refs so the forest can be constructed before  validation; the error only surfaces on `validate_node_hashes()`.

bobbinth: Makes sense. I think enforcing this would probably have to wait until something like https://github.com/0xMiden/miden-vm/issues/2473.

PR #2592, opened 2026-01-20
Fix diagnostics when external node resolution fails

### LINKED ISSUE #2476: Re-enable diagnostic tests after `FastProcessor` migration

plafer:
Closes #2476 

Re-enables the 2 tests referenced in #2476, and fixes the expected digests (due to #2547). More specifically,

- `test_diagnostic_syscall_target_not_in_kernel()`: no solution needed, the error context set up in `start_call_node()` already correctly points to the syscall `CallNode`.
- `test_diagnostic_no_mast_forest_with_procedure()`: on error, we walk up one node in the continuation stack to find the caller of the `ExternalNode`, and wse the caller's node ID to build the error context (e.g. to get the diagnostic for the `CallNode`, no the `ExternalNode`). More information can be found in the docstring of the new `maybe_use_caller_error_context()` method.

I didn't add a changelog entry, since this bug was introduced in #2477, which will land in the same release.

### DIFF
--- a/processor/src/fast/external.rs
+++ b/processor/src/fast/external.rs
@@ -5,7 +5,7 @@ use miden_core::mast::{ExternalNode, MastForest, MastNodeExt, MastNodeId};
 
 use crate::{
     AsyncHost, ExecutionError,
-    continuation_stack::ContinuationStack,
+    continuation_stack::{Continuation, ContinuationStack},
     errors::OperationError,
     fast::{BreakReason, FastProcessor, Tracer},
 };
@@ -30,7 +30,12 @@ impl FastProcessor {
             .await
         {
             Ok(result) => result,
-            Err(err) => return ControlFlow::Break(BreakReason::Err(err)),
+            Err(err) => {
+                let maybe_enriched_err =
+                    maybe_use_caller_error_context(err, current_forest, continuation_stack, host);
+
+                return ControlFlow::Break(BreakReason::Err(maybe_enriched_err));
+            },
         };
 
         tracer.record_mast_forest_resolution(resolved_node_id, &new_mast_forest);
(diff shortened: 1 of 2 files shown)


processor/src/fast/external.rs
```diff
@@ -86,3 +88,67 @@ impl FastProcessor {
...
+    }
+
+    // Look for caller context in the continuation stack
+    let Some(top_continuation) = continuation_stack.peek_continuation() else {
+        return original_err;
+    };
+
+    // Extract parent node ID from Call or Dyn continuations
+    let parent_node_id = match top_continuation {
+        Continuation::FinishCall(parent_node_id) | Continuation::FinishDyn(parent_node_id) => {
```
huitseeker: nit: Could add a test that triggers `NoMastForestWithProcedure` via `dyncall` to cover the `FinishDyn` branch here?

huitseeker (approved):
This LGTM.

processor/src/fast/external.rs
plafer: Good catch, this made me realize that we don't use `ExternalNode` with `DynNode`s, since `DynNode`s get the digest to be executed on the stack. `ExternalNode` is only needed for `CallNode`, since it encodes the callee using a `MastNodeId`. Hence, I removed the `FinishDyn` case.

Though this made me double-check whether we should look for other parents (which we didn't do the first time with `Process`). Turns out we were missing a bunch of cases. We weren't catching when this happened in a join, split, or loop. Turns out the assembler doesn't generate `AssemblyOp` decorators for `JoinNode`s (though I still made us look for it in `maybe_use_caller_error_context()`) - I will create an issue for this.

So TLDR, diagnostics should now be better than they were with `Process` because of this - good catch 🙂

plafer: Created #2594

PR #2593, opened 2026-01-21
fix(assembly): u32wrapping_madd and u32wrapping_add3 return lo instead of hi

huitseeker:
Both operations were using `[U32madd/U32add3, Drop]` which dropped the low part and kept the high part (carry). Wrapping semantics require returning the low part (result mod 2^32), so this changes to `[U32madd/U32add3, Swap, Drop]` to match other wrapping operations.

Added tests for u32wrapping_madd and u32wrapping_add3 that would have caught these bugs. Updated documentation to reflect cycle count change from 2 to 3.

### DIFF
--- a/miden-vm/tests/integration/operations/u32_ops/arithmetic_ops.rs
+++ b/miden-vm/tests/integration/operations/u32_ops/arithmetic_ops.rs
@@ -195,26 +195,9 @@ fn u32wrapping_add3() {
     let test = build_op_test!(asm_op, &[b as u64, a, 1]);
     test.expect_stack(&[1]);
 
-    // --- random u32 values with c = 0 -----------------------------------------------------------
-    let a = rand_value::<u32>();
-    let b = rand_value::<u32>();
-    let c = 0_u32;
-    let d = a.wrapping_add(b).wrapping_add(c);
-    let test = build_op_test!(asm_op, &[b as u64, a as u64, c as u64]);
-    test.expect_stack(&[d as u64]);
-
-    // --- random u32 values with c = 1 -----------------------------------------------------------
-    let a = rand_value::<u32>();
-    let b = rand_value::<u32>();
-    let c = 1_u32;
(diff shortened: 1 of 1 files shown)


Al-Kindi-0 (approved):
LGTM, thank you for catching this.
For the sections in the tests using random values, would it make sense to remove them since proptest already covers that?

CHANGELOG.md
```diff
@@ -9,11 +9,12 @@
...
+- `FastProcessor` now correctly returns an error if the maximum number of cycles was exceeded during execution ([#2537](https://github.com/0xMiden/miden-vm/pull/2537)).
+- `FastProcessor` now correctly only executes `trace` decorators when tracing is enabled (with `ExecutionOptions`) ([#2539](https://github.com/0xMiden/miden-vm/pull/2539)).
+- Removed the intentional HALT-insertion bug from the parallel trace generation ([#2484](https://github.com/0xMiden/miden-vm/pull/2484)).
+- Added missing `as_canonical_u64()` method to `IntValue` in `miden-assembly-syntax`, fixing compilation errors in the generated grammar code ([#2589](https://github.com/0xMiden/miden-vm/pull/2589)).
 - Fixed a bug where trace generation would fail if a core trace fragment started on the `END` operation of a loop that was not entered ([#2587](https://github.com/0xMiden/miden-vm/pull/2587)).
+- Fixed `u32wrapping_madd` and `u32wrapping_add3` returning the high part instead of the low part. Both now correctly return `(result mod 2^32)`.
```
plafer: This bug was introduced in #2547 right? Hence from the perspective of the changelog, since this bug was introduced and fixed in the same release, it shouldn't be mentioned here (since users will have never seen it)?

plafer (approved):
LGTM - good catch!

CHANGELOG.md
huitseeker: Fair, will remove

PR #2595, opened 2026-01-21
Rename `AsyncHost` to `Host`, and remove `SyncHost` and `BaseHost`

### LINKED ISSUE #2551: Remove `SyncHost`
Now that `Process` is removed, `SyncHost` is no longer used and should be removed. We should then rename `AsyncHost` to just `Host`.

Suggested in https://github.com/0xMiden/miden-vm/issues/2501#issuecomment-3716683691

plafer:
Closes #2551

### DIFF
--- a/processor/src/host/default.rs
+++ b/processor/src/host/default.rs
@@ -1,14 +1,11 @@
 use alloc::{sync::Arc, vec::Vec};
 
 use miden_core::{DebugOptions, EventId, EventName, Word, mast::MastForest};
-use miden_debug_types::{
-    DefaultSourceManager, Location, SourceFile, SourceManager, SourceManagerSync, SourceSpan,
-};
+use miden_debug_types::{DefaultSourceManager, Location, SourceFile, SourceManager, SourceSpan};
 
 use crate::{
-    AdviceMutation, AsyncHost, BaseHost, DebugError, DebugHandler, EventHandler,
-    EventHandlerRegistry, ExecutionError, MastForestStore, MemMastForestStore, ProcessState,
-    SyncHost, TraceError,
+    AdviceMutation, DebugError, DebugHandler, EventHandler, EventHandlerRegistry, ExecutionError,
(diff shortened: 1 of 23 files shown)


processor/src/host/default.rs
```diff
@@ -119,10 +118,10 @@ where
     }
 }
 
-impl<D, S> BaseHost for DefaultHost<D, S>
+impl<D, S> Host for DefaultHost<D, S>
 where
```
huitseeker, line 120: The bound here changed from `S: SourceManager` to `S: SourceManagerSync`. This makes sense since the unified `Host` trait now includes async methods that need `Send` bounds (via `FutureMaybeSend`).

processor/src/test_utils/test_consistency_host.rs
```diff
@@ -136,9 +136,9 @@ impl Default for TestConsistencyHost {
     }
 }
 
-impl<S> BaseHost for TestConsistencyHost<S>
+impl<S> Host for TestConsistencyHost<S>
```
bobbinth, line 139: Question: what does `TestConsistencyHost` do now? I think originally it was for testing consistency between fast and slow processors, but now that the slow processor is gone, do we still need it?

processor/src/host/default.rs
plafer: Actually I removed it since it wasn't necessary (even when `Send` bound is active), presumably because the `SourceManager` or `Self` don't show up in those futures.

processor/src/test_utils/test_consistency_host.rs
plafer: We use it to check consistency against prior behavior in `test_advice_provider()` (using `insta`), and in other tests to check that trace events occur in the right order. It could maybe be merged with `TestHost`, I'll create an issue for it.

plafer: #2601

PR #2597, opened 2026-01-21
Move `ExecutionOptions`, `ProvingOptions` and `ExecutionProof` out of `miden_air`

### LINKED ISSUE #2482: Move `ExecutionOptions` and `ProvingOptions` out of `miden_air`
As part #2406, `DEFAULT_CORE_TRACE_FRAGMENT_SIZE` had to be duplicated in `miden_air` (from `miden_processor`) because the `ExecutionOptions` are defined in `miden_air` (which doesn't depend on `miden_processor`).

Moving those out would have required moving the `ProvingOptions` out as well. This wasn't done right away to avoid further conflicts with the Plonky3 migration (#2472).

plafer:
Closes #2482

The main changes in this PR are moving out of `miden-air`:
- `ExecutionOptions` into `miden-processor`
- `ProvingOptions` into `miden-prover`
- `ExecutionProof` into `miden-core`
  - since it is used both by `miden-prover` and `miden-verifier`

These were all moved out since they don't have anything to do with AIR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Made `StackInputs` and `StackOutputs` implement `Copy` trait ([#2581](https://github.com/0xMiden/miden-vm/pull/2581)).
 - [BREAKING] Renamed `u32overflowing_mul` to `u32widening_mul`, `u32overflowing_madd` to `u32widening_madd`, and `math::u64::overflowing_mul` to `math::u64::widening_mul` ([#2584](https://github.com/0xMiden/miden-vm/pull/2584)).
 - [BREAKING] Removed `SyncHost` and `BaseHost`, and renamed `AsyncHost` to `Host` ([#2595](https://github.com/0xMiden/miden-vm/pull/2595)).
+- [BREAKING] Moved `ExecutionOptions` to `miden-processor`, `ProvingOptions` to `miden-prove`, and `ExecutionProof` to `miden-core` (all out of `miden-air`) ([#2597](https://github.com/0xMiden/miden-vm/pull/2597)).
 
 ## 0.20.2 (TBD)
 


prover/src/lib.rs
```diff
@@ -7,6 +7,7 @@ extern crate std;
 
 use alloc::string::ToString;
 
+use miden_core::{ExecutionProof, HashFunction};
```
bobbinth: Should this be `pub use`? Previously these were a part of the public interface.

processor/src/execution_options.rs
bobbinth, line 124: Not related to this PR, but I've always found it weird that we have this inconsistency between `with_tracing()` and `with_debugging()`: the former does not take a parameter, but the latter does (probably passing a parameter is better than not).

processor/src/fast/mod.rs
bobbinth, line 159: Not related to this PR (but somewhat related to https://github.com/0xMiden/miden-base/pull/2327#discussion_r2715480736: I think we may want to refactor how we instantiate the fast processor. One option could be to adapt more of a builder pattern - so that we could do something like:

```rust
let processor = FastProcessor::new(stack_inputs)
    .with_advice(advice_inputs)
    .with_options(exec_options);
```

prover/src/lib.rs
plafer: Done

processor/src/execution_options.rs
plafer: Added a commit that adds a `bool` parameter

processor/src/fast/mod.rs
plafer: Created #2602

core/src/errors.rs
```diff
@@ -9,6 +11,18 @@ pub enum InputError {
     InputStackTooBig(usize, usize),
 }
 
+// HASH FUNCTION ERROR
+// ================================================================================================
+
+/// Error type for invalid hash function strings.
+#[derive(Debug, thiserror::Error)]
+pub enum HashFunctionError {
```
adr1anh: Does this need to be an enum?

plafer: Changed it to a struct `InvalidHashFunctionError`

bobbinth (approved):
Looks good! Thank you! (though, not a super-detailed review from me)

Al-Kindi-0 (approved):
LGTM

PR #2598, opened 2026-01-22
chore(processor): audit must-use Felt suppressions

### LINKED ISSUE #2514: Audit must-use Felt suppressions to verify none indicate bugs
During the Plonky3 migration, the new `Felt` type gained a must-use annotation. This caused warnings where we call methods that return `Felt` but don't use the result. We suppressed these with `let _ =`.

**Example locations:**
- `processor/src/stack/tests.rs:200`
- `processor/src/parallel/core_trace_fragment/mod.rs:328`
- `processor/src/chiplets/hasher/tests.rs:141`

These are likely fine (test setup, mutation operations), but we should audit them to make sure none accidentally hide bugs where we should be checking a return value.

We should review each suppression and verify the value really doesn't need checking, then add comments if the intent isn't obvious.

**References:**
- https://github.com/0xMiden/miden-vm/pull/2472#discussion_r2629773691
- https://github.com/0xMiden/miden-vm/pull/2472#discussion_r2629778652

huitseeker:
Audit `let _ = ...` suppressions that were added when the Felt type gained a must-use annotation during the Plonky3 migration.

Changes:
- `assert_binary` returns `Result<(), OperationError>` instead of `Result<Felt, OperationError>` since returned Felt was never used. 
- Add doc comment to `initialize_chiplets` explaining why return values are discarded during replay (only side effects matter for trace gen)
- Add comment to `fill_fragment` explaining why `ControlFlow` is ignored
- Add comments to overflow stack push operations explaining `IndexVec` returns an unused index
- Add comment to hasher Merkle tree tests explaining trace verification

Closes #2514

### DIFF
--- a/processor/src/operations/utils.rs
+++ b/processor/src/operations/utils.rs
@@ -51,10 +51,10 @@ pub(crate) fn validate_dual_word_stream_addrs(
 
 /// Asserts that the given value is a binary value (0 or 1).
 #[inline(always)]
-pub fn assert_binary(value: Felt) -> Result<Felt, crate::OperationError> {
+pub fn assert_binary(value: Felt) -> Result<(), crate::OperationError> {
     if value != ZERO && value != ONE {
         Err(crate::OperationError::NotBinaryValue { value })
     } else {
-        Ok(value)
+        Ok(())
     }
 }
--- a/processor/src/chiplets/hasher/tests.rs
+++ b/processor/src/chiplets/hasher/tests.rs
@@ -84,6 +84,9 @@ fn hasher_permute() {
 
 // MERKLE TREE TESTS
 // ================================================================================================
+//
+// These tests verify trace generation, not computed results. The Merkle roots are validated
+// through `check_merkle_path`.
 
 #[test]
 fn hasher_build_merkle_root() {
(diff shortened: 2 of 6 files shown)


processor/src/operations/utils.rs
```diff
@@ -51,10 +51,10 @@ pub(crate) fn validate_dual_word_stream_addrs(
 
 /// Asserts that the given value is a binary value (0 or 1).
 #[inline(always)]
-pub fn assert_binary(value: Felt) -> Result<Felt, crate::OperationError> {
+pub fn assert_binary(value: Felt) -> Result<(), crate::OperationError> {
```
adr1anh, line 54: This may be a larger change, but I could be useful to return `Option<bool>` instead?

adr1anh (approved):
Left one minor comment that can be addressed later (or here if the change is minimal. Otherwise looks great!

processor/src/operations/utils.rs
plafer: Hmmm, why `Option<bool>` and not just `bool`? Where we return `true` if the value is binary, and `false` otherwise

adr1anh: My thinking was `Result<bool, Err> -> Option<bool>`, but this is just a nit.

Issue #2540, opened 2026-01-05
Remove `on_assert_failed` method from `Host`


bitwalker:
For some background context, see [this comment](https://github.com/0xMiden/miden-vm/issues/2507#issuecomment-3707792340) by @bobbinth and the discussion in that thread.

We no longer really require the `on_assert_failed` method of the `Host` trait, as that functionality can be centralized in the processor, and there is little reason to support it as an extension point. In the past, it was useful to implement this method to render better error messages based on assertion error codes - but now that we can provide error strings to assertions, that is no longer necessary.

Currently, to support this method, the `AssertError` type is an alias of `Box<dyn Error + Send + 'static>`, and due to the fact that it does not also implement the `DIagnostic` trait, assertion failures are unable to print associated source code snippets relevant to the assertion. This makes debugging of assertion failures very difficult from anywhere other than the `miden-debug` executor currently (which takes extra steps to extract source locations when execution fails with an error).

Removing this method will allow us to make `AssertError` a concrete error type that implements `Diagnostic`, thus allowing the processor itself to render assertion errors with their associated source spans (so long as the assertion has an attached `AssembyOp` decorator), and removing this responsibility from downstream users of the processor.

bitwalker:
@djolertrk Do you want to tackle this one? If you comment here I can assign it to you

djolertrk:
Sure :)

plafer:
From a merge conflict perspective, I would probably wait for #2544 to be merged because tackling this one.

plafer:
#2544 is now merged, so @djolertrk feel free to start working on this one

PR #2600, opened 2026-01-22
processor: Remove on_assert_failed

djolertrk:
Resolves https://github.com/0xMiden/miden-vm/issues/2540.

It is a first step through solving the https://github.com/0xMiden/compiler/issues/806 -- enabling source location being printed when `assert!` is triggered.


djolertrk:
cc @bitwalker @greenhat @plafer

processor/src/host/default.rs
```diff
@@ -198,6 +198,7 @@ where
 // ================================================================================================
 
 /// A Host which does nothing.
+#[allow(dead_code)]
 pub struct NoopHost;
```
plafer: The `NoopHost` is still used, so we don't need the `allow(dead_code)`

djolertrk:
> Looks good!
> 
> Left a nit. Please fix the merge conflicts, add a changelog entry, and this should be good to go

Thanks!

PR #2605, opened 2026-01-23
Cleanup `StackInputs` and `StackOutputs`

### LINKED ISSUE #2491: Redesign `StackInputs`
`StackInputs` was introduced with the former `Process`. However, `Process` was removed with #2483, and hence `StackInputs` is only used in the `miden_processor::execute()` method. That is, `FastProcessor` doesn't use it.

We should investigate
- Whether the `FastProcessor` should use `StackInputs` in its API
- Should we flip the orientation of elements that `StackInputs` expects?
- Should we just get rid of `StackInputs`?

plafer:
Closes #2491 

Cleans `StackInputs` and `StackOutputs` API, and use `StackInputs` in `FastProcessor` constructors.


core/src/stack/outputs.rs
bobbinth: It may be good to update this description. Lines 12 and 14 could be phrased better, and reference to `stack` on line 16 seems to be outdated.

plafer: Done

Issue #2545, opened 2026-01-07
feat: define a Generic CSRContainer


huitseeker:
Future PR: this CSR validation logic is copy/pasted for all instances for CSRs that we have. Ideally we would have a `CsrMatrix` type where we would write that once, and build all or concrete CSR types from it. It would look something like

```rs
pub struct CsrMatrix<Data, Idx> {
    data: Vec<Data>,
    indptr: IndexVec<Idx, usize>
}

pub struct NodeToDecoratorIds {
  before_enter_decorators: CsrMatrix<DecoratorId, MastNodeId>,
  after_exit_decorators: CsrMatrix<DecoratorId, MastNodeId>
}
```
Ideally too we're able to reuse that type to define the "two-level CSR" that `OpToDecoratorIds` needs.

_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2470#discussion_r2665837245_

bobbinth:
Closed by https://github.com/0xMiden/miden-vm/pull/2606.

PR #2606, opened 2026-01-25
feat: Separate AsmOp storage from Debug/Trace decorators

### LINKED ISSUE #2541: feat: store AsmOp decorators in their own DebugInfo section
@Al-Kindi-0 I re-requested a review, since I added this method since you approved (which I only noticed was needed when working on #2539). Basically, we should execute decorators not only in debug mode, but also when tracing is enabled.

**This has the unfortunate consequence that the processor will be slowed down when only tracing (but not debug mode) is enabled.**

This is a by-product of the fact that `AsmOp` decorators should not be lumped in with `Debug` and `Trace` decorators, since they do not need to be executed at runtime. Currently, if tracing mode is enabled (or even debug mode just for `debug` decorators), the processor needlessly loops through all `AsmOp` decorators, which really isn't needed - we only need to look them up when an error occurs. `AsmOp` decorators should be stored separately in their own CSR matrix. This is as previously discussed with @huitseeker.

_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2537#discussion_r2662824356_

huitseeker:
~On top of #2590.~ Fixes #2541, #2545

----

Moves `AsmOp` decorators out of the `Decorator` enum into dedicated sparse storage. AsmOps are now stored in a CSR matrix indexed by `(NodeId, OpIdx)` and accessed only when needed for error reporting. This eliminates runtime iteration overhead when tracing or debugging is enabled.

Also adds a generic `CsrMatrix<I, D>` type to replace hand-rolled CSR implementations throughout the codebase.

<details><summary>In Detail</summary>

- **`Decorator` enum**: Removed `AsmOp` variant
- **`CsrMatrix<I, D>`**: New generic type in `miden-utils-indexing` with validation and serialization
- **`OpToAsmOpId`**: New sparse storage mapping `(NodeId, OpIdx) -> AsmOpId`
- **`DebugInfo`**: Added `asm_op_storage` field and `get_assembly_op(node_id, op_idx)` method
- **`OpToDecoratorIds`, `NodeToDecoratorIds`**: Refactored to use `CsrMatrix`

### Out of scope

- applying the `CSRMatrix` generic data structure to the two-level CSR within `OpToDecoratorIds` (one level out of two was converted to using the generic)

</details>

## Breaking changes

MAST format version bumped to `[0, 0, 2]`. Old MAST files cannot be loaded.

### DIFF
--- a/core/src/mast/debuginfo/mod.rs
+++ b/core/src/mast/debuginfo/mod.rs
@@ -1,18 +1,17 @@
 //! Debug information management for MAST forests.
 //!
-//! This module provides the [`DebugInfo`] struct which consolidates all debug-related
-//! information for a MAST forest in a single location. This includes:
+//! This module provides the [`DebugInfo`] struct which consolidates all debug-related information
+//! for a MAST forest in a single location. This includes:
 //!
 //! - All decorators (debug, trace, and assembly operation metadata)
 //! - Operation-indexed decorator mappings for efficient lookup
 //! - Node-level decorator storage (before_enter/after_exit)
 //! - Error code mappings for descriptive error messages
 //!
-//! The debug info is always available at the `MastForest` level (as per issue #1821),
-//! but may be conditionally included during assembly to maintain backward compatibility.
-//! Decorators are only executed when the processor is running in debug mode, allowing
-//! debug information to be available for debugging and error reporting without
-//! impacting performance in production execution.
+//! The debug info is always available at the `MastForest` level (as per issue 1821), but may be
+//! conditionally included during assembly to maintain backward compatibility. Decorators are only
+//! executed when the processor is running in debug mode, allowing debug information to be available
+//! for debugging and error reporting without impacting performance in production execution.
 //!
 //! # Debug Mode Semantics
 //!
@@ -36,9 +35,9 @@
 //!
 //! # Production Builds
 //!
-//! The `DebugInfo` can be stripped for production builds using the [`clear()`](Self::clear)
-//! method, which removes decorators while preserving critical information. This allows
-//! backward compatibility while enabling size optimization for deployment.
+//! The `DebugInfo` can be stripped for production builds using the [`clear()`](Self::clear) method,
+//! which removes decorators while preserving critical information. This allows backward
+//! compatibility while enabling size optimization for deployment.
 
 use alloc::{
     collections::BTreeMap,
@@ -47,13 +46,17 @@ use alloc::{
     vec::Vec,
 };
 
-use miden_utils_indexing::{Idx, IndexVec};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
 use super::{AsmOpId, Decorator, DecoratorId, MastForestError, MastNodeId};
 use crate::{
-    AssemblyOp, LexicographicWord, Word,
+    AssemblyOp, Idx, IndexVec, LexicographicWord, Word,
+    mast::serialization::{
+        StringTable,
+        asm_op::{AsmOpDataBuilder, AsmOpInfo},
+        decorator::{DecoratorDataBuilder, DecoratorInfo},
+    },
     utils::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
 };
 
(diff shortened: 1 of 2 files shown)


processor/src/parallel/snapshots/miden_processor__parallel__tests__parallel::tests::test_trace_generation_at_fragment_boundaries::case_28.snap
plafer, line 1: This file was created, but I don't see a corresponding test case. Should this be removed?

crates/assembly/src/instruction/mod.rs
```diff
@@ -44,36 +44,33 @@ impl Assembler {
...
-                _ => panic!("expected AsmOp decorator"),
-            }
-            decorators.push(asm_op_id);
-        }
-
-        // Compile the instruction, passing the decorators (which may be empty).
+        // For node-creating instructions, finalize the AssemblyOp now (it will have 0 cycles
+        // since no operations have been added yet for this instruction).
+        let pending_node_asm_op =
+            if can_create_node && !matches!(instruction.inner(), Instruction::Exec(_)) {
```
plafer:
```suggestion
            if can_create_node {
```
I ran the tests and confirmed we no longer need this restriction. I had added that condition because some `VmStateIterator`-based tests were failing, but it could very well have a bug in the `VmStateIterator` (no one understood how that thing worked).

core/src/mast/serialization/tests.rs
```diff
@@ -116,8 +116,10 @@ fn confirm_operation_and_decorator_structure() {
         Operation::LogPrecompile => (),
     };
 
+    // Decorator variants - exhaustiveness check for serialization coverage.
+    // Note: AssemblyOp was removed from Decorator enum; AssemblyOps are now stored
+    // separately in DebugInfo's asm_op storage.
```
plafer: Not sure this is useful long term (or at least should be adjusted to not refer to today's current state)

processor/src/parallel/snapshots/miden_processor__parallel__tests__parallel::tests::test_trace_generation_at_fragment_boundaries::case_28.snap
huitseeker: Good catch!

core/src/mast/debuginfo/mod.rs
```diff
@@ -81,6 +87,12 @@ pub struct DebugInfo {
     /// Efficient storage for node-level decorators (before_enter and after_exit).
     node_decorator_storage: NodeToDecoratorIds,
 
+    /// All AssemblyOps in the MAST forest.
+    asm_ops: IndexVec<AsmOpId, AssemblyOp>,
+
+    /// Efficient access to AssemblyOps per operation per node.
+    asm_op_storage: OpToAsmOpId,
```
bobbinth, line 94: Maybe in the future it would make sense to encapsulate these into a single struct. Same could apply to the decorator-specific fields.

core/src/mast/serialization/asm_op.rs
```diff
@@ -0,0 +1,284 @@
...
+// ================================================================================================
+
+/// Fixed-width serialized representation of an [`AssemblyOp`].
+///
+/// Each `AsmOpInfo` stores an offset into the variable-length data blob where the AssemblyOp's
+/// payload is stored. The payload format is:
+///
+/// ```text
+/// num_cycles: u8
+/// should_break: u8 (0 or 1)
```
bobbinth, line 37: Not for this PR, but `should_break` probably won't work any more (and I'm not 100% sure it was working before this PR).

Basically, at some point, this was used to stop execution of the program at a given point by inserting a "breakpoint". Instead of having a dedicated decorator for this, we attached this as a property to the `AssemblyOp` decorator. Now that AsmOps are not evaluated on every cycle, having `should_break` is not really useful here - so, at the very least, we should remove it.

If we do want to bring this functionality back, this would probably have to be its own decorator - similar to the debug and trace decorators. But it is not clear to me if a `breakpoint` decorator is useful in the context of the new processor. cc @plafer.

huitseeker: #2622

Issue #2569, opened 2026-01-12
Unify the API of `u64.masm`, `u128.masm` and `u256.masm`


Al-Kindi-0:
We should probably do a followup to optimize the `u64.masm` module and unify the API with that of `u128.masm` and `u256.masm`

_Originally posted by @Al-Kindi-0 in https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2668849833_

bobbinth:
I'd make this issue a bit broader - to make sure we have consistent treatment of all arithmetic operations in the VM. In my mind, this includes:

- **Native operations** - `u32` and extension field operations. From what I can tell, these are already working as intended. Specifically:
  - For `u32` operations we put the result in the little-endian order (i.e., `[lo, hi]`). The two exceptions are subtraction and division where we output `[borrow, diff]` and `[remainder, quotient]` respectively - which I think is the desired arrangement.
  - For extension field operations, the "base" element is on the top of the stack.
  - When we do comparisons for `[b, a, ...]` we basically do `a op b` - e.g., `a < b`. Subtraction and division work in a similar way (e.g., `a op b`). In both cases, I think this is the desired behavior, but would be good to confirm. cc @bitwalker.
- **Assembly instructions** - I think we are largely there, but there are a few things left to do:
  - I think the docs need to be updated (e.g., we still have "u32overflowing_mul` instead of `u32widening_mul`).
  - For consistency, I would introduce "widening" versions of addition instructions: `U32WideningAdd` and `U32WideningAdd3`.
- **math::u64** module - I think we are largely there. I would maybe introduce the "widening" addition variants to keep things consisent.
- **math::u128** module - this one still needs to be merged and probably updated to be in little-endian order.
- **math::u256** module - this one needs to be refactored to be in little-endian order.

I may have missed something, but I think the above should cover most things.

Al-Kindi-0:
I think this is largely consistent with how I am thinking about it. Will address these in a soon-to-be published PRs.

For arity-two ops, I think the layout `[b, a, ..]` is the right one and is consistent throughout the codebase (will double check this).
We can wait for @bitwalker for clarity on `u128`.

bobbinth:
Closed by #2614.

Al-Kindi-0:
@bobbinth , should we close this now ?

bobbinth:
It is already closed, no?

Al-Kindi-0:
Yes, now it is

Issue #2564, opened 2026-01-09
Revisit `u64.masm` BE format


Al-Kindi-0:
Some of the procedures in this module still have implementations which BE-orientation native.
We should investigate if we can change these to be LE-orientation native as this might yield some improvements (performance/simplicity) given our migration to LE-convention in #2547

_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2673481325_

bitwalker:
Having limbs be ordered most-significant-first has benefits for many common operations on integers (e.g. comparisons, checking for signedness), and makes casts from smaller integer widths to larger ones trivial (as they can be simply zero-extended/padded with `push.0`), or truncated to smaller ones with `drop`.

Additionally, Wasm loads its `i64` type (and the representation that would be `i128`, if that was a native Wasm type) in big-endian limb order on the Wasm operand stack. Since the compiler is the only user of the `u64` MASM library at this time (unless things have changed), making this change to the limb order would pessimize Rust-compiled Miden programs significantly IMO.

While it is perhaps a bit odd to have types whose operand stack layout differs from its heap layout, doing so for integral types has real optimization value, and would be consistent across integer types. For the most part, the only reason the layout of these types would matter, is in the case of `u64`, storing them to the heap as part of a larger word-sized structure using `mem_storew` directly without swapping the limb order first.

However, let's say we were to do this, and make the limb order least-significant-first. This is what would the impact would be to various operations, and what the compiler would need to do to emulate Wasm semantics on the Miden VM.

* Every time a 64-bit is loaded from memory on to the Wasm operand stack, the compiler will need to emit a `swap` to get the limb order into the order expected by Wasm (and thus subsequent instructions). Conversely, it will need to `swap` before writing them back.
* We'd need a new calling convention that is attached to all Wasm-compiled functions, so that we can distinguish between functions that use the Wasm layout on the operand stack, and the Miden layout.
* Every time a 64/128-bit integer is being passed across a procedure boundary with a non-Wasm calling convention, we'd need to swap the limb order. Conversely, any code that is going to call a procedure whose definition uses the Wasm convention, will need to swap the limbs of any integers passed on the operand stack.
* Comparisons will require visiting every limb of the two integer values to determine the result. In the big-endian orientation, one only has to check until the limbs differ to obtain the result.
* The effect on signedness checks is like 1 cycle I think, so pretty minimal.
[...]

PR #2614, opened 2026-01-29
Unify u32, u64 and u128 APIs

Al-Kindi-0:
## Describe your changes

Addresses #2569

Once this is merged, I can also adapt #2438 

- Updated `CHANGELOG.md'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ WARNINGS                 := RUSTDOCFLAGS="-D warnings"
 BUILDDOCS                := MIDEN_BUILD_LIB_DOCS=1
 
 # -- feature configuration ------------------------------------------------------------------------
-ALL_FEATURES_BUT_ASYNC   := --features concurrent,executable,testing,internal
+ALL_FEATURES             := --all-features
 
 # Workspace-wide test features
 WORKSPACE_TEST_FEATURES  := concurrent,testing,executable
@@ -57,12 +57,12 @@ FEATURES_core-lib        :=FEATURES_verifier        :=
 
 .PHONY: clippy
 clippy: ## Runs Clippy with configs
-	cargo +nightly clippy --workspace --all-targets ${ALL_FEATURES_BUT_ASYNC} -- -D warnings
+	cargo +nightly clippy --workspace --all-targets ${ALL_FEATURES} -- -D warnings
 
 
 .PHONY: fix
 fix: ## Runs Fix with configs
-	cargo +nightly fix --allow-staged --allow-dirty --all-targets ${ALL_FEATURES_BUT_ASYNC}
+	cargo +nightly fix --allow-staged --allow-dirty --all-targets ${ALL_FEATURES}
 
 
 .PHONY: format
@@ -82,7 +82,7 @@ lint: format fix clippy ## Runs all linting tasks at once (Clippy, fixing, forma
 
 .PHONY: doc
 doc: ## Generates & checks documentation
-	$(WARNINGS) $(BUILDDOCS) cargo doc ${ALL_FEATURES_BUT_ASYNC} --keep-going --release
+	$(WARNINGS) $(BUILDDOCS) cargo doc ${ALL_FEATURES} --keep-going --release
 
 .PHONY: serve-docs
 serve-docs: ## Serves the docs
@@ -146,7 +146,7 @@ test: ## Run all tests for the workspace
 
 .PHONY: test-docs
 test-docs: ## Run documentation tests (cargo test - nextest doesn't support doctests)
-	$(BUILDDOCS) cargo test --doc $(ALL_FEATURES_BUT_ASYNC)
+	$(BUILDDOCS) cargo test --doc $(ALL_FEATURES)
 
 # -- filtered test runs ---------------------------------------------------------------------------
 
@@ -173,7 +173,7 @@ test-loom: ## Runs all loom-based tests
 
 .PHONY: check
 check: ## Checks all targets and features for errors without code generation
-	$(BUILDDOCS) cargo check --all-targets ${ALL_FEATURES_BUT_ASYNC}
+	$(BUILDDOCS) cargo check --all-targets ${ALL_FEATURES}
 
 .PHONY: check-features
 check-features: ## Checks all feature combinations compile without warnings using cargo-hack
--- a/crates/lib/core/Cargo.toml
+++ b/crates/lib/core/Cargo.toml
@@ -28,8 +28,6 @@ path = "tests/main.rs"
 [features]
 default = ["std"]
 std = ["miden-assembly/std", "miden-utils-sync/std"]
-with-debug-info = []
-legacy-stark-tests = []
 
 [dependencies]
 # Miden dependencies
--- a/crates/lib/core/tests/main.rs
+++ b/crates/lib/core/tests/main.rs
@@ -32,9 +32,9 @@ mod helpers;
 mod mast_forest_merge;
 mod math;
 mod mem;
-#[cfg(feature = "legacy-stark-tests")]
-mod pcs;
-#[cfg(feature = "legacy-stark-tests")]
-mod stark;
 mod sys;
 mod word;
+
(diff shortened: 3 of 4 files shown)


docs/src/design/stack/u32_ops.md
```diff
@@ -59,7 +59,7 @@ $$
 Thus, for a 64-bit value to encode a valid field element, either the lower 32 bits must be all zeros, or the upper 32 bits must not be all ones (which is $2^{32} - 1$).
 
 ## U32SPLIT
-Assume $a$ is the element at the top of the stack. The `U32SPLIT` operation computes $(b,c) \leftarrow a$, where $b$ contains the lower 32 bits of $a$, and $c$ contains the upper 32 bits of $a$. The diagram below illustrates this graphically.
+Assume $a$ is the element at the top of the stack. The `U32SPLIT` operation computes $(b,c) \leftarrow a$, where $b$ contains the lower 32 bits of $a$, and $c$ contains the upper 32 bits of $a$. The output stack is `[b, c, ...]` with $b$ (low limb) on top. The diagram below illustrates this graphically.
 
 ![u32split](../../img/design/stack/u32_operations/U32SPLIT.png)
```
bobbinth, line 64: I think the diagram here needs to be updated (also probably relevant for other diagrams).

crates/lib/core/asm/math/u64.masm
bobbinth: Is the return type correct here? I thought we'd want `(remainder, quotient)` here as this would be more consistent with `u32`.

bobbinth, line 359: Not from this PR, but I wonder why we need to re-write this procedure rather than just use `divmod`? That is, this could be just:
```
pub proc mod(b: u64, a: u64) -> u64
    exec.divmod
    swap drop
end
```
I'm not sure that 2 extra cycles justify this much extra code.

bobbinth, line 347: Same comment as above, and here it'd be just one extra cycle:
```
pub proc div(b: u64, a: u64) -> u64
    exec.divmod
    drop
end
```

crates/lib/core/asm/math/u256.masm
```diff
@@ -1,63 +1,110 @@
...
+#! Reorders the top u256 value between LE and BE limb order. This is an involution.
+#! Input/Output (LE <-> BE):
+#! [x0, x1, x2, x3, x4, x5, x6, x7, ...] <-> [x7, x6, x5, x4, x3, x2, x1, x0, ...]
+proc u256_le_to_be
+    swapw
+    reversew
+    swapw
+    reversew
+    swapw
+end
```
bobbinth, line 12: Could this not be just:
```
proc u256_le_to_be
    reversew
    swapw
    reversew
end
```

crates/lib/core/asm/math/u256.masm
```diff
@@ -66,33 +113,52 @@ pub proc wrapping_sub(rhs: u256, lhs: u256) -> u256
...
     movup.6
     u32overflowing_sub
     movup.2
     add
     movup.5
     movup.5
     movup.2
-    u32overflowing_add
+    u32widening_add
     drop
```
bobbinth: Could we double-check this? It seems a bit odd that we are doing a widening add and then dropping the low bits (in most situations we drop the high bits).

crates/lib/core/asm/math/u256.masm
```diff
@@ -66,33 +113,52 @@ pub proc wrapping_sub(rhs: u256, lhs: u256) -> u256
...
+pub proc overflowing_sub(rhs: u256, lhs: u256) -> (i1, u256)
+    exec.u256_le_to_be_pair
+    exec.sub_with_borrow_be
+    movdn.8
+    exec.u256_le_to_be
+    movup.8
 end
 
 pub proc and(rhs: u256, lhs: u256) -> u256
+    exec.u256_le_to_be_pair
```
bobbinth: For `and` and other bitwise operations, do we actually need to do these rotations? Seems like the endiannes shouldn't matter here (though, I haven't spent too much time thinking it through).

Also nit: let's add a `BITWISE OPERATIONS` or something like that as the header above these procedures.

crates/lib/core/asm/math/u256.masm
```diff
@@ -177,9 +247,11 @@ pub proc xor(rhs: u256, lhs: u256) -> u256
     movup.3
     movup.4
     u32xor
+    exec.u256_le_to_be
 end
 
 pub proc eqz(rhs: u256, lhs: u256) -> i1
+    exec.u256_le_to_be
```
bobbinth: Same comment as for bitwise operations: do we actually need to do these rotations here?

crates/lib/core/asm/math/u256.masm
```diff
@@ -215,8 +288,7 @@ proc mulstep
     movdn.2                    # [c, b, k, a, ...]
     u32widening_madd        # [lo, hi, a, ...]  (computes c*b+k)
     movup.2                    # [a, lo, hi, ...]
-    u32overflowing_add         # [result, carry, hi, ...]  (computes a+lo)
-    swap                       # [carry, result, hi, ...]
+    u32overflowing_add      # [carry, result, hi, ...]  (computes a+lo)
```
bobbinth: nit: comment alignment.

crates/lib/core/asm/math/u256.masm
bobbinth, line 1: Let's create an issue for the future to consider re-writing these procedures to avoid the extra rotations we introduced in this PR. The issue should cover `u64` module as well.

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: Nothing changed here semantically from the original, but will double check it

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: Good point
Fixed now

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: Fixed here and elsewhere

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: fixed

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: Indeed, fixed

crates/lib/core/asm/math/u64.masm
Al-Kindi-0: Indeed, this was a leftover from changing from BE to LE. 
Fixed it now

Al-Kindi-0: fixed

Al-Kindi-0: Indeed, changed now

docs/src/design/stack/u32_ops.md
Al-Kindi-0: Yes, this will part of the PR updating the excalidraws after the recent changes

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: I think https://github.com/0xMiden/miden-vm/issues/2564 captures this, namely that we have to adapt the bodies of some of the procedures to be LE-first.

crates/lib/core/asm/math/u256.masm
Al-Kindi-0: Fixed now, there was a bug here that was introduced when we changed the semantics of `u32overflowing_sub`.
Also added more tests that would have caught this.

Al-Kindi-0:
> Looks god! Thank you! I left some comments inline - most of them a very minor. Updating the diagrams we should address in a different PR.
> 

Yes, we can do the diagrams in a separate PR

> Re `u128` module. I'd say we can probably create another PR that is based on top of this PR and just copy the code from #2438 there.

Yes, working on it now

Issue #2165, opened 2025-09-15
Parallel trace generation: switch to iterative execution


plafer:
The parallel trace generation still uses recursive execution (through a recursive `execute_mast_node()`). We should move to iterative execution, similar to how is done in `FastProcessor::execute_impl()`.

PR #2615, opened 2026-01-29
Abstract away program execution

### LINKED ISSUE #2591: Abstract away `Program` execution logic
Currently, the logic for executing a program in `FastProcessor` and `CoreTraceFragmentFiller` (i.e. parallel trace generation) is duplicated with the exception of operation execution with `Processor::execute_sync_op()`. That is, they do subtlely different things (`FastProcessor` needs to execute decorators, whereas `CoreTraceFragmentFiller` needs to generate the trace), but at an abstract level, they are executing/traversing the program in the exact same way. We should then be able to write that logic only once in `Processor`.

It turns out that we have all the tools we need already in the code base to do that. The general strategy is to start from the `FastProcessor::execute_impl()` implementation (which iteratively traverses the program), and move it to `Processor` by abstracting away the relevant parts. The `Processor` trait will then expose an `execute()` method (with interface similar to `FastProcessor::execute_impl()`) instead of its current `execute_sync_op()`.

Some notable changes:
- `execute_{before,after}_decorators()` will become abstract methods of `Processor`, so that `FastProcessor` can implement them as it does today, and `CoreTraceFragmentFiller` will leave them blank.
- For `CoreTraceFragmentFiller`, the actual trace generation will occur in a new `Tracer` implementation (call it `TraceGenerationTracer`), and more specifically inside the `Tracer::start_clock_cycle()` method.
- `FastProcessor` and `CoreTraceFragmentFiller` have different stopping conditions - this will simply be encoded by the specific `Stopper` trait implementation they use.

This change should both drastically reduce the number of lines of code in the processor, as well as fix #2165 for free. There is also no expected impact on performance.

plafer:
Closes #2591

Abstracts away the processor logic in the new internal function `miden_processor::execution::execute_impl()`. Critically, this function is synchronous, and makes use of the sans-IO pattern (first explored in #2006) to handle things that need to execute asynchronously. As a result, **we see a 1.75x speedup in program execution**, with the processor now running at around 320 MHz (for the `blake3_1to1` benchmark).

Reviewers should pay special attention to the following docstrings, which document the new architecture:
- `miden_processor::execution::execute_impl()`
- `miden_processor::execution::InternalBreakReason`
- `miden_processor::tracer::Tracer` (and the `start_clock_cycle()` method)

In the subsequent PR, we will migrate the parallel trace generation (`CoreTraceFragmentFiller`) to use this new `execute_impl()`, which will result in fewer total lines of processor code, and will make the parallel trace generation execute the program iteratively (#2165).

Additional notes:
- I left some methods where they are intentionally to reduce the diff size; we can clean up the general structure with #2495 
- The cycle count check in `NeverStopper::should_stop()` costs a 25% reduction in performance. Any suggestions to alleviate the hit are welcome 🙂

## blake3_1to1 benchmark results</summary>
**On next branch:**

```
program_execution_fast/blake3_1to1
                        time:   [3.5455 ms 3.5492 ms 3.5531 ms]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
```

**On this branch:**

```
program_execution_fast/blake3_1to1
                        time:   [2.0198 ms 2.0219 ms 2.0241 ms]
                        change: [−43.117% −43.031% −42.947%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Hence, given that the benchmark runs in 647 392 (stack) clock cycles, we go from 181 MHz on `next` to 320 MHz on this branch. Note that before adding the cycle count check in `NeverStopper`, I was getting 1.52ms for this benchmark - or 425 MHz.

### DIFF
--- a/processor/src/execution/dyn.rs
+++ b/processor/src/execution/dyn.rs
@@ -46,8 +46,7 @@ where
 
     let dyn_node = current_forest[current_node_id].unwrap_dyn();
 
-    // Retrieve callee hash from memory, using stack top as the memory
-    // address.
+    // Retrieve callee hash from memory, using stack top as the memory address.
     let callee_hash = {
         let ctx = processor.system().ctx();
         let clk = processor.system().clock();
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -333,6 +333,12 @@ impl FastProcessor {
         &self.memory
     }
 
+    /// Returns a narrowed interface for reading and updating the processor state.
+    #[inline(always)]
+    pub fn state(&mut self) -> ProcessorState<'_> {
+        ProcessorState { processor: self }
+    }
+
     // MUTATORS
     // -------------------------------------------------------------------------------------------
 
@@ -964,17 +970,6 @@ pub struct ExecutionOutput {
     pub advice: AdviceProvider,
     pub memory: Memory,
     pub final_pc_transcript: PrecompileTranscript,
-}
-
-// PROCESSOR STATE
-// ===============================================================================================
-
-/// Processor state accessor.
-impl FastProcessor {
-    #[inline(always)]
-    pub fn state(&mut self) -> ProcessorState<'_> {
-        ProcessorState { processor: self }
-    }
 }
 
 // EXECUTION CONTEXT INFO
--- a/processor/src/execution/basic_block.rs
+++ b/processor/src/execution/basic_block.rs
@@ -250,7 +250,7 @@ where
 }
 
 // HELPERS
-// ----------------------------------------------------------------------------------------------
+// ================================================================================================
 
 /// Executes a single operation batch within a basic block node, starting from the operation
 /// index `start_op_idx`.
@@ -379,7 +379,7 @@ fn get_continuation_after_executing_operation(
 }
 
 // EXPORTS
-// ----------------------------------------------------------------------------------------------
+// ================================================================================================
 
 /// Function to be called after [`InternalBreakReason::Emit`] is handled. See the documentation of
 /// that enum variant for more details.


processor/src/processor.rs
```diff
@@ -81,38 +95,70 @@ pub trait Processor: Sized {
...
+    /// Executes any decorator in a basic block that is to be executed after all operations in the
+    /// block. This only differs from `execute_after_exit_decorators` in that these decorators are
+    /// stored in the basic block node itself.
+    fn execute_end_of_block_decorators(
+        &mut self,
+        basic_block_node: &BasicBlockNode,
+        node_id: MastNodeId,
+        current_forest: &Arc<MastForest>,
+        host: &mut impl Host,
+    ) -> ControlFlow<BreakReason>;
```
bobbinth, line 142: We should try to get rid of this eventually.

processor/src/lib.rs
```diff
@@ -347,3 +358,26 @@ impl<'a> ProcessorState<'a> {
         self.processor.memory.get_memory_state(ctx)
     }
 }
+
+// STOPPER
+// ===============================================================================================
+
+/// A trait for types that determine whether execution should be stopped at a given point.
+pub trait Stopper {
```
bobbinth, line 366: Would be great to provide a bit more info on why we need this and how it is currently used. For example, do we need this only for debug purposes? Or do we use it for other things as well?

processor/src/fast/basic_block.rs
bobbinth, line 24: Not from this PR, but I'm not sure the last sentence here makes sense (or at least it is not clear to me what the "linked case" means).

bobbinth, line 30: Also not from this PR, but why do we need to return `ControlFlow` from here? IIUC, this method either succeeds (continue) or there is an error (break). So, we could probably make it more narrow like:
```rust
fn execute_decorator(
    &mut self,
    decorator: &Decorator,
    host: &mut impl Host,
) -> Result<(), ExecutionError> 
```

processor/src/execution/dyn.rs
```diff
@@ -0,0 +1,201 @@
...
+            mem_addr,
+            processor.system().ctx(),
+            processor.system().clock(),
+        );
+
+        word
+    };
+
+    // Drop the memory address from the stack. This needs to be done before saving the context.
+    processor.stack_mut().decrement_size(tracer);
```
bobbinth, line 76: Not specifically for this PR, but why do we need to pass `tracer` into `decrement_size()` here? Could we not change this line to:
```rust
processor.stack_mut().decrement_size();
tracer.decrement_stack_size();
```

And I think more generally, it is not always obvious what should happen where - e.g., in some cases we delegate updating the tracer to the processor, in other cases we update the tracer directly in these functions. I don't have a specific suggestion on how to improve this - so, just an observation for now.

processor/src/execution/dyn.rs
```diff
@@ -0,0 +1,201 @@
...
+    current_node_id: MastNodeId,
+    current_forest: &mut Arc<MastForest>,
+    continuation_stack: &mut ContinuationStack,
+    host: &mut impl Host,
+    tracer: &mut impl Tracer,
+    stopper: &S,
+) -> ControlFlow<InternalBreakReason>
+where
+    P: Processor,
+    S: Stopper<Processor = P>,
```
bobbinth, line 33: Related to the above comment: we pass a large number of parameters to this and other such functions. This makes it a bit difficult to follow and also feels like some level of abstraction/encapsulation is missing.  

For example, maybe we could introduce something like an `ExecutionContext` struct which would encapsulate some of these parameters (e.g., `current_forest`, `continuation_stack`, `kernel` etc.). I haven't thought through if this actually makes sense (and there could be a good reason why it doesn't) - so, again more of an observation for now.

processor/src/fast/mod.rs
bobbinth, line 449: Related to the above comment: we already have `ResumeContext` which encapsulates continuation stack,  current MAST forest, and kernel. We could probably pass it around instead of the individual parameters (and rename it into something like `ExecutionContext`).

processor/src/processor.rs
plafer: Very much agreed - created #2633

processor/src/execution/dyn.rs
plafer: Agreed, and I think this comment applies to any other `Processor` method (and its various `*Interface` associated types) which take a `Tracer` argument now that we have a "single source of truth" `execute_impl()` implementation.

plafer: Addressed in #2676

PR #2616, opened 2026-01-30
Disable legacy tests

bobbinth:
This PR disables legacy tests (i.e., the recursive verifier tests) as the test files still link to Winterfell modules - but such modules are no longer imported. Also included:

- Removed `legacy-stark-tests` feature.
- Removed `with-debug-info` feature as it wasn't used anywhere.
- Replaced `ALL_FEATURES_BUT_ASYNC` with `ALL_FEATURES` in the Makefile as we no longer have the `async` feature.


Al-Kindi-0 (approved):
LGTM!

PR #2618, opened 2026-01-30
fix(assembly-syntax): properly handle quoted path components

bitwalker:
This is based against `main` because it is a hotfix for known issues in the current release. It is a non-breaking change, so is safe to release.

/cc @greenhat 

---

Prior to this commit, when constructing PathBufs from a string containing quoted components would improperly strip the quotes from those components, which would result in later iteration over components of that path treating that part of the path as unquoted. For example, given the following path string:

`root:root_ns@1.0.0::module::"function::with::quoted::symbol_name"`

Constructing a PathBuf from this string would succeed as expected, but would encode the path internally as if it had been passed as:

`root:root_ns@1.0.0::module::function::with::quoted::symbol_name`

If we later iterate over components of the path, the part that should have been treated as a single component, i.e.
`"function::with::quoted::symbol_name`, would instead be treated as multiple components, e.g. `function`, `with`, `quoted`, and `symbol_name`.

This commit changes how we handle quoted components in the Components iterator, so that the quotes are preserved.

Additionally, a new `PathBuf::push_component` method was added, which allows one to explicitly push a string as a quoted component, ensuring the component is quoted if not already.

A regression test was added that covers the case where this was discovered, and further ensures that constructing a PathBuf from a single string, or building it up via `push_component` produce identical results.

A related issue was also identified and fixed, namely that internally the Components iterator was not consuming trailing/leading path separators after parsing a quoted component, causing unexpected results  when using `Components::as_path`. This has been fixed, making the behavior consistent across both quoted and unquoted identifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.4 (Unreleased)
+
+- Fixed issue with handling of quoted components in `PathBuf` [#2618](https://github.com/0xMiden/miden-vm/pull/2618)
+
 ## 0.20.3 (2026-01-27)
 
 - Fixed issue where exports of a Library did not have attributes serialized [#2608](https://github.com/0xMiden/miden-vm/issues/2608)
--- a/crates/assembly-syntax/src/ast/path/components.rs
+++ b/crates/assembly-syntax/src/ast/path/components.rs
@@ -48,6 +48,16 @@ impl<'a> PathComponent<'a> {
             Self::Normal(id) => id.chars().count(),
         }
     }
+
+    /// Returns true if this path component is a quoted string
+    pub fn is_quoted(&self) -> bool {
+        matches!(self, Self::Normal(component) if component.starts_with('"') && component.ends_with('"'))
+    }
+
+    /// Returns true if this path component requires quoting when displayed/stored as a string
+    pub fn requires_quoting(&self) -> bool {
+        matches!(self, Self::Normal(component) if component.contains("::"))
+    }
 }
 
 impl PartialEq<str> for PathComponent<'_> {
@@ -72,7 +82,13 @@ impl fmt::Display for PathComponent<'_> {
 /// Returns an iterator over the path components represented in the provided source.
 ///
 /// A path consists of at list of components separated by `::` delimiter. A path must contain
-/// at least one component.
+/// at least one component. Path components may be quoted, in which `::` delimiters are ignored.
+/// Quoted path components may also contain characters that are not otherwise valid identifiers in
+/// Miden Assembly.
+///
+/// Note that quoted components may not contain nested quotes - the appearance of a nested quote in
+/// a quoted component will be treated as a closing quote resulting in unexpected behavior or
+/// validation errors as a result.
 ///
 /// # Errors
 ///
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 0.20.3 (2026-01-27)
 
 - Fixed issue where exports of a Library did not have attributes serialized [#2608](https://github.com/0xMiden/miden-vm/issues/2608)
+- Fixed issue with handling of quoted components in `PathBuf` [#2618](https://github.com/0xMiden/miden-vm/pull/2618)
```
bobbinth: v0.20.3 has already been released. So, this entry should go into v0.20.4 section.

crates/assembly-syntax/src/ast/path/path_buf.rs
```diff
@@ -198,6 +198,39 @@ impl PathBuf {
...
+    pub fn push_component<S>(&mut self, component: &S)
+    where
+        S: AsRef<str> + ?Sized,
+    {
+        let component = component.as_ref();
+        match component {
+            "" | "\"\"" => (),
+            "::" if self.inner.is_empty() => {
+                self.inner.push_str("::");
+            },
```
bobbinth, line 215: Is `::` considered to be a valid component?

If I'm understanding correctly, this branch will cover case when the current path is empty, and we are trying to append `::` to it. And so, the resulting path would be just `::` - right?

crates/assembly-syntax/src/ast/path/components.rs
bobbinth: I think the current iteration logic is a bit more complicated than this, right? (otherwise we could have used something as simple as `path.split("::")`). It would be good to describe the iterator behavior in a bit more details.

crates/assembly-syntax/src/ast/path/path_buf.rs
bitwalker: There are two types of components: `Root` and `Normal` - the `Root` component is represented as the string `::` and only ever occurs as the first component in a path, indicating that it is absolute (i.e. it is equivalent to a leading `/` in a file system path). 

So if we're pushing a literal `::` and the path is currently empty, then we treat it as a valid `Root` component, this leads to the following being consistent:

```rust
let p1 = PathBuf::new("::foo::bar::\"baz\"").unwrap();
let mut p2 = PathBuf::default();
for component in p1.components() {
    p2.push_component(component.as_str());
}
assert_eq!(&p1, &p2);
```

Note that `PathBuf` is used for constructing paths from scratch, as well as representing an owned path, so in this case, yes, the resulting path would still return true for `Path::is_empty`, but the underlying buffer would contain `::` - that path would not be valid on its own, but this is why we check certain preconditions (is the path empty, is it valid) when working with paths in general elsewhere in the assembler.

crates/assembly-syntax/src/ast/path/components.rs
bitwalker: Yes, it also performs some validation, and handles quoted components, catching things like empty components and unclosed quotes. I can provide a bit more detail in this doc comment to mention those things as well.

bobbinth: Makes sense. Yeah, the need for quoted identifiers complicates this. I do wonder if there is a way to simplify the iterator code using something like regular expressions - but that's not for this PR.

PR #2623, opened 2026-01-31
feat: tighten up MastForest serialization with varints

huitseeker:
- Add a benchmark that measures the serialized size of the core library MastForest. The benchmark may be of independent interest to measure the impact of the `MastNode` hashes.
- Make MastForest serialization smaller by using compact tags and varint indices/offsets where we store node types, block batch counts, asm op payload offsets, asm op string table indices, asm op locations, decorator data offsets, and procedure roots.
- Keep byte-cost comments in sync with the wire format.

Contributes to #1873

<details><summary>Changes</summary>

- New bench for core lib MastForest serialized size.
- MastNodeType uses a compact tag plus varint payloads.
- Basic block batch count uses a varint.
- AsmOpInfo and asm op payload use varint offsets and indices, including location start/end.
- DecoratorInfo uses a varint offset for decorator data.
- Procedure roots are serialized as a length prefix plus varint u32s.
</details>

## Bench Results
Run: `cargo bench -p miden-core-lib --bench mast_serialization_size -- --nocapture`

Before (8654d7815):
- Full size: 434,065 bytes
- Full time: 2.2229–2.3210 ms
- Full throughput: 178.35–186.23 MiB/s

After (ced394e40):
- Full size: 406,088 bytes
- Full time: 2.1967–2.2708 ms
- Full throughput: 170.55–176.30 MiB/s
- Stripped size: 275,653 bytes
- Stripped time: 640.23–709.10 µs
- Stripped throughput: 370.73–410.61 MiB/s

## Notes
- No backwards compatibility is preserved for these format changes. AIming to piggy-back on the serialization version bump of #2606.

### DIFF
--- a/core/src/mast/serialization/info.rs
+++ b/core/src/mast/serialization/info.rs
@@ -1,6 +1,6 @@
 use alloc::vec::Vec;
 
-use super::{NodeDataOffset, basic_blocks::BasicBlockDataDecoder};
+use super::{NodeDataOffset, basic_blocks::BasicBlockDataDecoder, read_u32_varint};
 use crate::{
     mast::{
         MastForestContributor, MastNode, MastNodeId, Word,
@@ -298,15 +298,6 @@ impl Deserializable for MastNodeType {
     fn min_serialized_size() -> usize {
         1
     }
-}
-
-fn read_u32_varint<R: ByteReader>(source: &mut R) -> Result<u32, DeserializationError> {
-    let value = source.read_usize()?;
-    value.try_into().map_err(|_| {
-        DeserializationError::InvalidValue(format!(
-            "Invalid value: expected to fit in u32, but was {value}"
-        ))
-    })
 }
 
 #[cfg(test)]
--- a/core/src/mast/serialization/info.rs
+++ b/core/src/mast/serialization/info.rs
@@ -13,10 +13,6 @@ use crate::{
 // ================================================================================================
 
 /// Represents a serialized [`MastNode`], with some data inlined in its [`MastNodeType`].
-///
-/// The serialized representation of [`MastNodeInfo`] is guaranteed to be fixed width, so that the
-/// nodes stored in the `nodes` table of the serialized [`MastForest`] can be accessed quickly by
-/// index.
 #[derive(Debug)]
 pub struct MastNodeInfo {
     ty: MastNodeType,
@@ -122,9 +118,9 @@ impl Deserializable for MastNodeInfo {
         Ok(Self { ty, digest })
     }
 
-    /// Returns the minimum serialized size: 8 bytes for MastNodeType + 32 bytes for Word digest.
+    /// Returns the minimum serialized size: 1 byte for MastNodeType + 32 bytes for Word digest.
     fn min_serialized_size() -> usize {
-        40
+        33
     }
 }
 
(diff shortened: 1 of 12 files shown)


plafer (approved):
LGTM!

Throughput on my M4 (this branch):

full: ~200 MB/s
stripped: ~600 MB/s

core/src/mast/serialization/mod.rs
```diff
@@ -199,7 +218,11 @@ impl Deserializable for MastForest {
         let _decorator_count = source.read_usize()?; // Read for wire format compatibility
 
         // Reading procedure roots
-        let roots: Vec<u32> = Deserializable::read_from(source)?;
+        let roots_len = source.read_usize()?;
+        let mut roots = Vec::with_capacity(roots_len);
+        for _ in 0..roots_len {
```
bobbinth, line 223: Question: would this bypass the budgeted reader checks? Basically, if `roots_len` was serialized maliciously to be a very big number, we'd run into issues on line 222, right?

core/src/mast/serialization/info.rs
```diff
@@ -250,65 +252,37 @@ impl MastNodeType {
...
     fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
-        let (discriminant, payload) = {
-            let value = source.read_u64()?;
-
-            // 4 bits
-            let discriminant = (value >> 60) as u8;
-            // 60 bits
-            let payload = value & 0x0f_ff_ff_ff_ff_ff_ff_ff;
-
-            (discriminant, payload)
```
bobbinth, line 286: One of the original goals of the serialization format was to encode each MAST node into exactly 8 bytes so that we could compute an offset for a given node without having to read the full forest from disk. I actually don't think we are relying on this property now - and given our use cases, maybe we won't need to (i.e., we could assume that loading full MAST forests into memory will never get prohibitively expensive).

So, overall, I'm fine with this change - but want to make sure we make we all agree with the implications.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline (only one may need to be addressed before merging).

One more general comment: I'd love to have a more holistic plan for MAST forest serialization. We don't necessarily need to implement it yet, but would be good to know where we are going. Previously, the plan was to have two different formats:

- "fat" - this was supposed to be the trusted serialization where we don't care about the size too much, but care about deserialization/access speed.
- "compact" - this was supposed to be the untrusted serialization (i.e., we need to recompute the hashes and run other checks), and here we'd prioritize size over deserialization speed (and this format would not contain any debug info).

The current format was supposed to be the "fat" one with the "compact" one coming in the future and implemented separately. But recently it feels like we are merging both into a single format. This may be fine, but would be good to have an issue that explicitly outlines where we want to get to (and maybe we can replace some of the existing serialization-related issues with it).

core/src/mast/serialization/info.rs
huitseeker: The current code always deserializes the full forest sequentially - there's no random access, no offset computation, and no seek functionality in the `ByteReader` trait.

The `varint` approach trades random access for size/speed.

huitseeker:
Closing in favor of discussion in https://github.com/0xMiden/miden-vm/issues/2726

PR #2626, opened 2026-01-31
refactor(processor): remove all `_sync` methods for strictly async API

### LINKED ISSUE #2552: Remove `miden_processor::execute_sync()`
As part of a move to make the processor strictly `async`. Users should instead use the async `execute()`, and manage the async runtime themselves.

First suggested in https://github.com/0xMiden/miden-vm/issues/2501#issuecomment-3716683691

Farukest:
## Summary
Update test infrastructure to use async/await pattern directly instead of sync wrappers.

## Changes
- Added async variants to `Test` struct (`execute_async`, `prove_and_verify_async`, `get_last_stack_state_async`)
- Updated all test files to use `#[tokio::test]` with `.await`
- Fixed nested tokio runtime errors ("Cannot start a runtime from within a runtime")
- Removed `#[track_caller]` from async functions (unsupported in stable Rust)

## Files Changed
- `crates/test-utils/src/lib.rs` - Core async method additions
- `processor/src/**/tests/*.rs` - Test updates
- `miden-vm/tests/integration/*.rs` - Integration test updates
- `crates/lib/core/tests/**/*.rs` - Core lib test updates

Closes #2552


crates/test-utils/src/lib.rs
```diff
@@ -384,6 +383,8 @@ impl Test {
         let (program, host) = self.get_program_and_host();
         let mut host = host.with_source_manager(self.source_manager.clone());
 
+        let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
```
huitseeker: The sync test helpers (`execute`, `execute_for_output`, `execute_with_debug_buffer`, `prove_and_verify`, `get_last_stack_state`) now unconditionally create a new tokio runtime. 

This means they will panic if called from within an existing tokio runtime (like inside a `#[tokio::test]`). Users in async contexts should use the new `*_async` variants instead.

This is a behavioral change from the previous implementation which tried to use `Handle::try_current()` to detect nested runtime situations and panic with a clearer message.

Consider adding a doc comment to these sync methods warning about this limitation (or restoring the `Handle::try_current()` guard for better error messages).

Farukest:
All done @huitseeker 👍

huitseeker:
@Farukest Please check CI status.

Farukest:
> @Farukest Please check CI status.

Fixed 👍

plafer:
@Farukest please sign all commits

bobbinth:
Without having looked at the code in much detail yet - do we actually need to make all tests `async`? While the processor is async, we do have the `exec_sync()` function to run in synchronously. Could we use this function for tests?

Farukest:
> @Farukest please sign all commits

All done @plafer :)

plafer:
> Without having looked at the code in much detail yet - do we actually need to make all tests async? While the processor is async, we do have the exec_sync() function to run in synchronously. Could we use this function for tests?

Agreed - it feels better to me to hide the `async`-ness for tests. Looking back at #2552, the scope was originally to just get rid of the top-level `execute_sync()`. Though @huitseeker if you prefer the current approach then I'm also okay with just merging.

huitseeker:
I happen to not have any issue with `#[tokio::test]` and don't see avoiding those as a simplification — not a dig, I just happen to not see the point. If the primary prover API is async, it makes sense to me. I'd even switch the test helpers to be uniformly async too.

huitseeker:
Superseded by #2865.

PR #2627, opened 2026-01-31
fix(assembly): regression in link-time symbol resolution involving quoted path components

bitwalker:
The individual commits describe more of the issue and its resolution. I'd recommend reviewing each commit separately for that reason.

This PR addresses some additional issues I uncovered around paths now that the underlying representation is preserving quoted components:

* Joining paths actually has two different semantics depending on whether we are joining two paths, or joining a symbol to a path. Due to the way `Path::join` was implemented, the semantics for the former were always applied - however this would lead to an issue if you were joining a symbol that contains `::` (as is the case with symbols coming from Rust at times). Because `Path::join` was expecting the input to be `AsRef<Path>`, there wasn't sufficient flexibility to encode via types which set of semantics should be used when joining something to a path. This PR introduces a sealed `Join` trait that implements the correct joining semantics for the set of types we care about. This preserves the ergonomics/expected behavior of `path.join("foo")` and `path.join(Path::new("foo::bar"))`, but will now ensure that something like `path.join("foo::bar")` gets joined as a single quoted component (quoted due to the `::`). That means there is some room for error if one was relying on that to treat the string as a sub-path, but that was not something we were using ourselves anywhere. This is all now fully documented on `Path::join` though.
* If one explicitly quoted something that didn't require it, and then compared to a path where that component hadn't been quoted, the paths would not compare equal due to a lack of canonical form. This can still occur if you convert a `&str` to a `&Path` directly, but now, when constructing paths, they are always canonicalized - one simply has to ensure that they are comparing against canonical paths if they want to avoid this edge case. In the assembler, this is always the case now (so far as I have been able to find).
* We were special-casing `$kernel` and `$main` as path components, automatically treating a path with those as the first component as "absolute" - however this again introduces an issue with canonicalization because we would avoid prepending the root `::` component if `Path::is_absolute` returned true, resulting in `$kernel` and `::$kernel` not being equivalent. This PR removes this special-casing, and only considers those components as special during canonicalization itself (i.e. we will ensure the canonical form is always `::$kernel`/`::$exec`). This simplifies things in a few places as a result.
* We had a few conversion-related trait impls that were not using the correct joining semantics when constructing a PathBuf, this was missed in my last PR, and addressed here.
* I discovered that `serde` deserialization was implemented in a way that would trivially fail if the input could not be borrowed due to reuse of the deserializer for `Path`. I've improved the deserialization logic for Path/PathBuf and related types like `Arc<Path>` to ensure that any string input will suffice. One can still do borrowing deserialization with `Path`, but it is no longer the default.
* I cleaned up the MASM grammar a bit as well, since I found that we weren't properly handling quoted path components there - now we do. This enabled me to remove some rules from the grammar that were unused, and collapse two rules into one.

### DIFF
--- a/crates/assembly-syntax/src/ast/path/join.rs
+++ b/crates/assembly-syntax/src/ast/path/join.rs
@@ -41,14 +41,16 @@ impl Join<Path> for Path {
             return self.to_path_buf();
         }
 
-        if self.is_empty() || other.is_absolute() {
-            return other.to_path_buf();
+        if self.is_empty() {
+            other.to_path_buf()
+        } else if other.is_absolute() || other.is_in_kernel() || other.is_in_exec() {
+            other.to_absolute().into_owned()
+        } else {
+            let mut buf = self.to_path_buf();
+            buf.push(other);
+
+            buf
         }
-
-        let mut buf = self.to_path_buf();
-        buf.push(other);
-
-        buf
     }
 }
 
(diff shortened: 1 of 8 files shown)


bitwalker:
I see the test failure, hold off on review until I've had a chance to resolve that, and I'll ping you

bitwalker:
@bobbinth This is now ready. I was able to confirm that these patches also address the path-related issues in the compiler, unblocking us on that front.

Al-Kindi-0 (approved):
LGTM

crates/assembly-syntax/src/ast/path/join.rs
```diff
@@ -0,0 +1,272 @@
...
+        let p1 = Path::new("foo");
+        let p2 = Path::new("::bar::baz");
+        let joined = Join::join(p1, p2);
+        assert_eq!(joined.as_path(), Path::new("::bar::baz"));
+        let mut components = joined.components();
+        assert_matches!(components.next(), Some(Ok(PathComponent::Root)));
+        assert_matches!(components.next(), Some(Ok(PathComponent::Normal("bar"))));
+        assert_matches!(components.next(), Some(Ok(PathComponent::Normal("baz"))));
+        assert_matches!(components.next(), None);
+    }
```
bobbinth, line 196: Is this an expected behavior that path `p1` (i.e., `foo`) gets silently swallowed here.

bitwalker: Yes, since joining an absolute path to another path (whether absolute or not) is equivalent to overwriting that path. This aligns with how `std::path::Path` works, though we don't actually make use of it ourselves, it just makes the semantics consistent.

PR #2630, opened 2026-02-01
Assorted cleanup

bobbinth:
This PR contains a series of commits with minor assorted cleanups. These include:

- Removed unused dependencies and all direct Plonky3 dependencies (all Plonky3 dependencies are now passed through `miden-crypto`).
- The processor now uses batch inversion from `miden-crypto` (I just copied the code from `miden-processor` to `miden-crypto).
- Removed some no-longer needed functions and traits from `miden-core`.
- Re-organized exports in `miden-core` and `miden-processor`. The main goal here was to avoid exporting too many items at the root and instead group released functionality into sub-modules.
  - This also included moving some code around - e.g., `AdviceInputs` were moved from `miden-processor` to `miden-core`.

This PR touched a lot of files and has a relatively large diff - but it contains very few actual code changes. Most of the diff is related to namespace changes of various components.

Overall, I think that organization of `miden-core` is in a pretty good shape, but there are still a lot of things to do as a part of https://github.com/0xMiden/miden-vm/issues/2495 for the `miden-processor`.

### DIFF
--- a/crates/lib/core/tests/crypto/falcon.rs
+++ b/crates/lib/core/tests/crypto/falcon.rs
@@ -11,10 +11,11 @@ use miden_core::{
 };
 use miden_core_lib::{CoreLibrary, dsa::falcon512_poseidon2};
 use miden_processor::{
-    DefaultHost, ExecutionError, OperationError, ProcessorState, Program, ProgramInfo, StackInputs,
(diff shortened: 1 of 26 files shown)


Al-Kindi-0 (approved):
All looks good to me, thanks for this

PR #2635, opened 2026-02-02
Fix off-by-one error in cycle limit check

### LINKED ISSUE #2632: Programs that execute exactly max_cycles cycles fail with CycleLimitExceeded when they should succeed.
### Packages versions

miden-vm: next

### Bug description

The root cause is in `finalize_clock_cycle_with_continuation`: the clock is incremented before checking the limit. The code does `processor.system_mut().increment_clock()` first, then `stopper.should_stop()` checks `clk >= max_cycles`. This means a program using exactly max_cycles cycles gets rejected because the check happens after the increment.

  The fix is to either check `clk > max_cycles` instead of `clk >= max_cycles`, or check the limit before incrementing the clock.

_No response_

### Relevant log output

```shell

```

Himess:
Closes #2632

### DIFF
--- a/processor/src/fast/tests/mod.rs
+++ b/processor/src/fast/tests/mod.rs
@@ -165,7 +165,7 @@ fn test_cycle_limit_exactly_max_cycles_succeeds() {
     // With 2018 Noop operations, the program uses exactly MIN_TRACE_LEN (2048) cycles.
     // 2018 operations result in 29 operation batches, and this requires executing 28 `RESPAN`
     // operations. So, we get 2018 + 28 = 2046. All of these operations are executed in a single
-    // basic block, and we need 2 more operations for block start (`SPAN`) and block end (`END`).
+    // basic block, and we need 2 more operations for block start (`BEGIN`) and block end (`END`).
     // Before the fix (clk >= max_cycles): this failed because 2048 >= 2048 is true.
     // After the fix (clk > max_cycles): this succeeds because 2048 > 2048 is false.
     const NUM_OPS: usize = 2018;
@@ -173,7 +173,7 @@ fn test_cycle_limit_exactly_max_cycles_succeeds() {
 
     let options = ExecutionOptions::new(
         Some(2048),
-        2048,
+        MIN_TRACE_LEN as u32,
         ExecutionOptions::DEFAULT_CORE_TRACE_FRAGMENT_SIZE,
         false,
         false,


processor/src/fast/tests/mod.rs
```diff
@@ -154,6 +154,40 @@ fn test_cycle_limit_exceeded() {
...
+/// This is a regression test for the off-by-one error where the cycle limit check used `>=`
+/// instead of `>`, causing programs that used exactly `max_cycles` cycles to fail.
+#[test]
+fn test_cycle_limit_exactly_max_cycles_succeeds() {
+    let mut host = DefaultHost::default();
+
+    // With 2018 Noop operations, the program uses exactly MIN_TRACE_LEN (2048) cycles.
+    // Before the fix (clk >= max_cycles): this failed because 2048 >= 2048 is true.
+    // After the fix (clk > max_cycles): this succeeds because 2048 > 2048 is false.
+    const NUM_OPS: usize = 2018;
```
bobbinth, line 171: This is a great explanation! I would maybe also add why 2018 Noops result in 2048 cycles. The math is:

- 2018 operations result in 29 operation batches, and this requires executing 28 `RESPAN` operations. So, we get $2018+28=2046$.
- All of these operations are executed in a single basic block, and we need 2 more operations to for block start (`BEGIN`) and block end (`END`).

processor/src/fast/tests/mod.rs
```diff
@@ -154,6 +154,40 @@ fn test_cycle_limit_exceeded() {
...
+    let program = simple_program_with_ops(vec![Operation::Noop; NUM_OPS]);
+
+    let options = ExecutionOptions::new(
+        Some(MIN_TRACE_LEN as u32),
+        MIN_TRACE_LEN as u32,
+        ExecutionOptions::DEFAULT_CORE_TRACE_FRAGMENT_SIZE,
+        false,
+        false,
+    )
```
bobbinth, line 180: nit: I would probably use `Some(2048)` instead of `Some(MIN_TRACE_LEN as u32)` to make it easier to follow that we are setting `max_cycles` to 2048.

processor/src/fast/tests/mod.rs
Himess: Done, added the explanation for the cycle count calculation.

processor/src/fast/tests/mod.rs
Himess: Fixed using `Some(2048)` now.

Himess:
Thanks for kind answers @bobbinth  ! 
Ready for re-review.

PR #2637, opened 2026-02-02
refactor(assembly): improve link-time symbol resolution internals

bitwalker:
In the course of looking into an issue described
[here](https://github.com/0xMiden/miden-base/pull/2377#issuecomment-3836395355), I discovered that a lot of the symbol resolver complexity was due to handling edge cases at the wrong level. Not only did this result in a lot of excess complexity, but it also failed to account for issues such as the one identified in the linked comment above.

This commit rewrites the core `resolve_path` implementation to work in terms of "expanding" a path into its absolute form, and then resolving the absolute path. This is much more efficient, since we can take a couple of shortcuts based on a few rules, and results in much more intuitive behavior, and a considerably simpler implementation.

Two new tests were added to catch a few edge cases that were not well represented by our existing test suite, and can be used to prevent future regressions as well.

This is a non-breaking change in the public interface of the assembler.

@bobbinth This is targeted against `main` since it fixes issues in the current release without any breaking changes (at least, things that would be caught by our test suite and compilation of the stdlib).

### DIFF
--- a/crates/assembly/src/tests.rs
+++ b/crates/assembly/src/tests.rs
@@ -1593,7 +1593,7 @@ fn link_time_const_evaluation_undefined_symbol() -> TestResult {
         use lib::a::FOO
         begin
             push.FOO
-            exec.a::f
+            exec.lib::a::f
             add
         end"
     );
@@ -4647,6 +4647,140 @@ fn test_kernel_linking_against_its_own_library() -> TestResult {
 }
 
 #[test]
+fn test_syscall_resolution_uses_kernel_module() -> TestResult {
+    let context = TestContext::default();
+
+    let kernel = context.parse_kernel(source_file!(
+        &context,
+        r#"
+        pub proc foo
+            caller
+            drop
+            push.1
+        end
+
+        pub proc bar
+            caller
+            drop
+            push.2
+        end
+        "#
+    ))?;
+
+    let lib = context.parse_module_with_path(
+        "userspace",
+        source_file!(
+            &context,
+            r#"
+            pub proc bar
+                push.0
+            end
+            "#
+        ),
+    )?;
+
+    let source = source_file!(
+        &context,
+        r#"
+        use userspace::bar
+
+        proc foo
+            push.0
+        end
+
+        begin
+            syscall.foo
+            syscall.bar
+        end
+        "#
+    );
+
+    let kernel = Assembler::new(context.source_manager()).assemble_kernel(kernel)?;
+
+    let mut assembler = Assembler::with_kernel(context.source_manager(), kernel);
+    assembler.compile_and_statically_link(lib)?;
+    let program = assembler.assemble_program(source)?;
+
+    let mast = {
+        let entry = program.get_node_by_id(program.entrypoint()).unwrap();
+        format!("{}", entry.to_display(program.mast_forest()))
+    };
+
+    let expected = r#"join
+    join
+        basic_block push(2147483648) push(4294967294) mstore drop noop end
+        asmOp(syscall.::$kernel::foo, 1)
+        syscall.0xe302a447fba1cf164ca808752528232c78b12813da35ec07470c146ed5d3a5d9
+    end
+    asmOp(syscall.::$kernel::bar, 1)
+    syscall.0x2370cdbd7b7c69035d324820a11a4582fd9d42ac74711be2288bc0922fd698bf
+end"#;
+    assert_eq!(mast, expected);
+
+    Ok(())
+}
+
+#[test]
+fn test_syscall_resolution_to_non_kernel_path_is_checked() -> TestResult {
+    let context = TestContext::default();
+
+    let kernel = context.parse_kernel(source_file!(
+        &context,
+        r#"
+        pub proc foo
+            caller
+            drop
+            push.1
+        end
+        "#
+    ))?;
+
+    let lib = context.parse_module_with_path(
+        "userspace",
+        source_file!(
+            &context,
+            r#"
+            pub proc bar
+                push.0
+            end
+            "#
+        ),
+    )?;
+
+    let source = source_file!(
+        &context,
+        r#"
+        begin
+            syscall.userspace::bar
+        end
+        "#
+    );
+
+    let kernel = Assembler::new(context.source_manager()).assemble_kernel(kernel)?;
+    let lib = Assembler::new(context.source_manager()).assemble_library([lib])?;
+
+    let error = Assembler::with_kernel(context.source_manager(), kernel)
+        .with_static_library(lib)?
+        .assemble_program(source)
+        .expect_err("expected diagnostic to be raised, but compilation succeeded");
+
+    assert_diagnostic_lines!(
+        error,
+        "syntax error",
+        "help: see emitted diagnostics for details",
+        "invalid syscall: callee must be resolvable to kernel module",
+        regex!(r#",-\[test[\d]+:3:21\]"#),
+        "2 |         begin",
+        "3 |             syscall.userspace::bar",
+        "  :                     ^^^^^^^^^^^^^^",
+        "4 |         end",
+        "  `----"
+    );
+
+    Ok(())
+}
+
+#[test]
 fn test_linking_imported_symbols_with_duplicate_prefix_components() -> TestResult {
     let context = TestContext::default();
 
@@ -4687,6 +4821,80 @@ fn test_linking_imported_symbols_with_duplicate_prefix_components() -> TestResul
         end
         "#,
     )?;
+
+    Ok(())
+}
+
+#[test]
+#[ignore = "leave disabled until either symbol resolution is rewritten or path semantics are refined"]
+fn test_linking_recursive_expansion() -> TestResult {
+    let context = TestContext::default();
+
+    let a_lib = context.parse_module_with_path(
+        "a",
+        source_file!(
+            &context,
+            r#"
+        pub use b::a
+        pub proc x
+            push.1
+        end
+        "#
+        ),
+    )?;
+
+    let b_lib = context.parse_module_with_path(
+        "b",
+        source_file!(
+            &context,
+            r#"
+        pub use a::a
+        pub proc foo
+            exec.a::x
+        end
+        "#
+        ),
+    )?;
+
+    let assembler = Assembler::new(context.source_manager());
+    let _ = assembler.assemble_library([a_lib, b_lib])?;
+
+    Ok(())
+}
+
+#[test]
+#[ignore = "leave disabled until either symbol resolution is rewritten or path semantics are refined"]
+fn test_linking_recursive_expansion_via_renamed_aliases() -> TestResult {
+    let context = TestContext::default();
+
+    let a_lib = context.parse_module_with_path(
+        "a::a",
+        source_file!(
+            &context,
+            r#"
+        pub use b::a2
+        pub proc x
+            push.1
+        end
+        "#
+        ),
+    )?;
+
+    let b_lib = context.parse_module_with_path(
+        "b",
+        source_file!(
+            &context,
+            r#"
+        pub use a::a->a2
+        pub proc foo
+            exec.a2::x
+        end
+        "#
+        ),
+    )?;
+
+    let assembler = Assembler::new(context.source_manager());
+    let _ = assembler.assemble_library([a_lib, b_lib])?;
 
     Ok(())
 }


bitwalker:
GitHub seems to be choking at the moment, I'll try to get all the jobs re-run as soon as it settles down again

crates/assembly-syntax/src/ast/path/path.rs
```diff
@@ -424,33 +459,58 @@ impl Path {
...
+        let prefix = prefix.to_relative();
+        <Path as StartsWith<Path>>::starts_with_exactly(this, prefix)
     }
 
     #[inline]
     fn starts_with_exactly(&self, prefix: &Path) -> bool {
-        <Self as StartsWith<str>>::starts_with_exactly(self, prefix.as_str())
+        let mut components = self.components();
+        for prefix_component in prefix.components() {
+            let prefix_component = prefix_component.expect("invalid prefix path");
```
bobbinth, line 530: Could we add a safety comment explaining why we expect `prefix_component` to always be valid?

More generally, why do we need to do component validation on iteration rather than on `Path` construction? Seems like it would be easier to reason about some of these things if we could always assume that a `Path` is well-formed.

crates/assembly-syntax/src/ast/path/path.rs
```diff
@@ -424,33 +459,58 @@ impl Path {
...
-                Some(rest) => rest.starts_with(prefix),
-                None => self.inner.starts_with(prefix),
-            }
-        }
+        let this = self.to_relative();
+        <Path as StartsWith<str>>::starts_with_exactly(this, prefix)
     }
 
     #[inline]
     fn starts_with_exactly(&self, prefix: &str) -> bool {
```
bobbinth, line 487: It may be worth adding a comment noting that this compares the `prefix` only against the first component (excluding the root marker). It took me a bit of time to understand this.

crates/assembly-syntax/src/ast/path/path.rs
```diff
@@ -386,6 +399,28 @@ impl Path {
...
 
+    /// Strips `prefix` from `self`, or returns `None` if `self` does not start with `prefix`.
+    ///
+    /// NOTE: Prefixes must be exact, i.e. if you call `path.strip_prefix(prefix)` and `path` is
+    /// relative but `prefix` is absolute, then this will return `None`. The same is true if `path`
+    /// is absolute and `prefix` is relative.
+    pub fn strip_prefix<'a>(&'a self, prefix: &Self) -> Option<&'a Self> {
+        let mut components = self.components();
+        for prefix_component in prefix.components() {
+            let prefix_component = prefix_component.expect("invalid prefix path");
```
bobbinth, line 430: Similar comment as before re adding a note about why we are guaranteed to have a valid `prefix_component` here.

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -153,199 +159,231 @@ impl<'a> SymbolResolver<'a> {
...
+            } else {
+                // A relative path can be expressed in four forms:
+                //
+                // 1. A reference to a symbol in the current module (possibly imported)
+                // 1. A reference to a symbol relative to an imported module, e.g. `push.mod::CONST`
+                // 2. A reference to a symbol relative to an imported module, e.g.
+                //    `push.mod::submod::CONST`
+                // 3. An absolute path expressed relatively, e.g. `push.root::mod::submod::CONST`,
+                //    which should have been expressed as `push.::root::mod::submod::CONST`, but the
+                //    `::` prefix was omitted/forgotten.
```
bobbinth, line 344: Was Line 269 supposed to be deleted? The content seems to be the same as line 270, but it is numbered as `1` (same as line 268).

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -153,199 +159,231 @@ impl<'a> SymbolResolver<'a> {
...
+                //
+                // We first resolve the symbol in the local module to either a local definition, or
+                // an imported symbol.
+                //
+                // If the symbol is locally-defined, the expansion is the join of the current module
+                // path and the symbol name.
+                //
+                // If the symbol is an imported item, then we expand the imported path recursively.
+                break match self
+                    .resolve_local_with_index(context.module, Span::new(span, symbol.as_str()))?
```
huitseeker: When resolving a single-component name in a syscall context, this uses `context.module` instead of the kernel module. 

The old `resolve` function had a branch that rebound `gid.module` to `self.graph.kernel_index` when
`context.in_syscall()` was true. That guard is gone here, so syscall targets can now bind to non-kernel items.

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -153,199 +159,231 @@ impl<'a> SymbolResolver<'a> {
...
+                // module: if this is not the case, then we fall back to attempting to resolve the
+                // path as if it was absolute. If this fails, the path is considered to refer to an
+                // undefined symbol.
+                //
+                // If the path is relative to an import in the current module, then we proceed by
+                // resolving the subpath relative to the import target. This is the recursive part,
+                // and the result of this recursive expansion is what gets returned from this
+                // function.
+                let (imported_symbol, subpath) = path.split_first().expect("multi-component path");
+                if ignored_imports.contains(imported_symbol) {
```
huitseeker: The old `find` function tracked visited full paths to detect cycles like `use.a; use.a::b; use.b::a`. This version only tracks import names in `ignored_imports`, keyed by just the first component. Cycles that involve multiple distinct aliases can fall back to absolute resolution instead of raising a clear recursive alias error.

crates/assembly/src/linker/resolver/symbol_resolver.rs
huitseeker, line 312: The old `resolve` function had a branch that rebound `gid.module` to `kernel_index` when `context.in_syscall()` was true. That guard is gone here, so syscall targets can now bind to non-kernel items. 

For example, I think that if a user module defines `foo` and the kernel also has `foo`, `syscall.foo` might now resolve to the user's `foo` instead of the kernel's `foo`.

crates/assembly/src/linker/resolver/symbol_resolver.rs
huitseeker, line 364: The old `find` function tracked visited full paths like `[a, b, a]` to detect cycles. This only tracks import names like `a`, so multi-alias cycles can slip through. 

IIUC, if module `A` imports `b` and module `B` imports `a`, then resolving `a::b::a::x` would keep expanding instead of raising a recursive alias error. The `ignored_imports` set catches `use lib::lib` style cycles, but not cycles with distinct names.

crates/assembly-syntax/src/ast/path/path.rs
```diff
@@ -424,33 +459,58 @@ impl Path {
...
+                    let is_absolute =
+                        components.next().is_some_and(|c| matches!(c, Ok(PathComponent::Root)));
+                    if !is_absolute {
+                        return false;
+                    }
+                    prefix
+                } else {
+                    prefix
+                };
+                components.next().is_some_and(
```
huitseeker, line 503: Nit: This only checks the first component against the entire prefix string, so `path("foo::bar").starts_with("foo")` returns true but `path("foo::bar").starts_with("foo::bar")` returns false. 

The old behavior used string prefix matching, so `path("foo::bar").starts_with("foo")` would return true as well, but the behavior was different for multi-component prefixes. This might be quite fine if the API is only used with single-component prefixes, but it's a subtle change.

crates/assembly-syntax/src/ast/path/path.rs
bitwalker: Even if we assert in `Path::new` (which just losslessly converts from `str` to `Path`), we'd need two implementations of the components iterator: one used when validating/constructing from a string, and one that assumes the components are valid and just asserts on that assumption being violated. Most of the implementation could actually be shared, we'd probably just have the "assume valid" `Iterator` wrap the current impl and assert on `Err`. That might be better than the current setup.

In any case, I couldn't make those sorts of changes here because they'd be breaking changes, and I wanted to keep this as non-invasive as possible.

crates/assembly-syntax/src/ast/path/path.rs
bitwalker: Good point - we document this for `join` now, but didn't reflect those semantics in the docs here.

crates/assembly-syntax/src/ast/path/path.rs
bitwalker: We're not guaranteed to - but it is part of the inherent contract of these APIs that paths are valid once constructed. If we make the changes discussed in the other comment, then we can actually enforce that, but that sort of change will have to happen in `next`

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Whoops, no, it's just incorrectly numbered

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Good point, this should be enforced by `resolve_invoke_target`, seems like we don't have a regression test for that either, so I'll add one

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: I'll add a regression test for this and make sure we're still properly handling cycles like that - but I don't believe the issue you've described is actually an infinite expansion (at least not without a reproducer), though the behavior would certainly depend on _where_ one is resolving a path like `a::b::a::x` from. But let's assume for a second that it is from `b`, which looks like so:

```
pub use a

pub proc foo
    exec.a::b::a::x
end
```

And `a` looks like:

```
pub use b

pub proc x
    push.1
end
```

So we have a cyclic dependency between `a` and `b`, but one that should be resolvable. Resolving `exec.a::b::a::x` in `b` would proceed as follows:

* The path is relative, so we split off `a` to resolve as an import, which expands to just `a`. We add `a` to `ignored_imports` and proceed to resolve the now partially-expanded path `a::b::a::x`
* The path is relative, so we split off `a` to resolve as an import, but this time it is ignored, instead we attempt to expand the path as absolute, i.e. `::a::b::a::x`
* The path is absolute, so we search for a module with the longest matching prefix, which is `a`, so we attempt to resolve `b::a::x` relative to `a`
* The path is relative, so we split off `b` to resolve as an import, which expands to just `b`, which is added to `ignored_imports`, and we proceed to resolve the now partially-expanded path `b::a::x`
* The path is relative, and again we ignore the import `b`, and convert the path to absolute form and attempt to resolve as `::b::a::x`
* The longest matching prefix of `::b::a::x` is `b`, so we attempt to resolve `a::x` relative to `b`
* We expand the `a` import to `a`, ignore it, and proceed to resolve `::a::x`
* The longest matching prefix of `::a::x` is `a`, so we attempt to resolve `x` relative to `a`
* The relative path `x` is a symbol, so we resolve it to a local definition in `a`

crates/assembly-syntax/src/ast/path/path.rs
bitwalker: Correct - however in our uses we only ever intend to match the leading component when using a string, and the subtleties around matching portions of a path with strings caused bugs when dealing with quoted identifiers.

This change, while technically breaking, actually guarantees the correct semantics for our intended uses of these functions. Essentially, like `join`, we want to ensure that when component structure matters, one has to be explicit about constructing a `Path` for multi-component matching/joining, in order to guarantee consistent behavior when dealing with paths that may have quoted components.

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: I did find a particularly pathological case though which does produce an issue - but I'm not sure if it would have succeeded previously either, in any case, here's the test:

```rust
#[test]
fn test_linking_recursive_expansion() -> TestResult {

    let a_lib = context.parse_module_with_path(
        "a",
        source_file!(
            &context,
            r#"
        pub use b::a
        pub proc x
            push.1
        end
        "#
        ),
    )?;

    let b_lib = context.parse_module_with_path(
        "b",
        source_file!(
            &context,
            r#"
        pub use a::a
        pub proc foo
            exec.a::x
        end
        "#
        ),
    )?;

    let assembler = Assembler::new(context.source_manager());
    let _ = assembler.assemble_library([a_lib, b_lib])?;

    Ok(())
}
```

processor/tests/issue_2456_test.rs
bobbinth, line 1: Question: what's the motivation for moving this file?

crates/assembly/README.md
```diff
@@ -205,8 +204,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .assemble_kernel(kernel)?;
 
     // Instantiate the assembler with multiple options at once
-    let assembler = Assembler::with_kernel(source_manager, kernel_lib)
-        .with_dynamic_library(&CoreLibrary::default())?;
+    let assembler = Assembler::with_kernel(source_manager, kernel_lib);
+    // If you wanted to link against the core library, you'd extend the above
+    // with: `.with_dynamic_library(&miden_core_lib::CoreLibrary::default())?;`
```
bobbinth, line 209: I don't mind this change, but what's the motivation for making it?

processor/tests/issue_2456_test.rs
bitwalker: We had the `miden-assembly` crate depending on `miden-processor` and `miden-core-lib` (both of which depend on `miden-assembly`) just for two integration tests (and also used by the doctests in the `miden-assembly` README).

The problem is that this meant that changing anything in the assembler and then running an assembler test, would first need to build both of those crates (and of particular note, `miden-core-lib`). If there was an issue in the assembler that you were trying to work out, and that issue affected any of the MASM in `miden-core-lib`, you would be unable to run any of the `miden-assembly` tests without getting `miden-core-lib` to build successfully first. That was obviously a problem for me when tinkering with anything involving the linker/symbol resolution.

Now, the two tests that used the processor live in the processor test suite, and this cyclic dependency is now gone. The only downside is that a couple of the doctests in the `miden-assembly` readme can no longer reference `miden-core-lib`, but the need for that is going away soon anyway.

crates/assembly/README.md
bitwalker: See my other comment above.

More generally though, I think the `miden-assembly` README should probably _not_ be referencing stuff from `miden-core-lib` just for the sake of an example. We can illustate the APIs just fine without doing that, and it saves us the need to have a cyclic dependency between `miden-assembly` and `miden-core-lib` which can be a real pain as mentioned in my other comment.

Lastly, referencing `Library` via Rust crates like that is going away anyway, so all of these examples will need to be rewritten regardless.

I left the comment in though just to preserve the usage that was documented there, it just isn't statically checked anymore.

Issue #2256, opened 2025-10-03
Improve benchmarks


plafer:
I see 2 main issues with how we do benchmarks today:

1. Currently our benchmarks reuse the files in the `miden_vm/masm-examples` folder. The problem is that only a few make for good benchmarks, and so we pretty much only use `blake3_1to1.masm` as a reference.
2. The code to walk through the `masm-examples` folder to run the benchmark is copy/pasted (and adapted slightly) for each new benchmark.

To address (1), I no longer think we should reuse `masm-examples` for benchmarking. We should instead have benchmark-specific MASM files, and think carefully about what we want each benchmark to be benchmarking, ensure it runs for a sufficient amount of time (e.g. in the milliseconds rather than microseconds or nanoseconds), etc. To start, we could probably only migrate the `blake3_1to1` benchmark over (effectively preserving the current way we do things).

As for (2), we will no longer be walking `masm-examples`, so all that code should go away. Any new boilerplate code should be extracted into functions in a library somewhere that is accessible to the benchmarks; maybe in `miden-test-utils`?

bobbinth:
I think the ultimate benchmark (or at least the one that we care the most about) is executing transaction kernel. We do have this in `miden-base` (i.e., [here](https://github.com/0xMiden/miden-base/tree/next/bin/bench-transaction)) - but I'm not sure it would be possible to re-create something like this here.

We could still try to get something relatively "representative" if we try to include a mix of operations (e.g., some hashing, memory reads/writes, Merkle path verification etc.).

Al-Kindi-0:
We can probably get the proportions of the mix by analyzing a few representative runs of the transaction kernel. It would have been great to also have the numbers for the batch kernel but I believe we do not have those yet, so maybe we can use the proportions we get for the transaction kernel in order to build representative benchmarks for large traces.

bitwalker:
I think good benchmarking has to start by understanding your bounds, i.e. minimum/typical/maximum theoretical throughput based on representative examples. For example:

* Max throughput: execute a minimal program that does basically nothing (in the context of the protocol, this might be the counter example). 
* Typical throughput: pick a program that does something significantly more expensive, but still very common, e.g. signature verification, hashing, copying data into/out of the advice provider from/to linear memory. Because this represents a typical workload, it gives you a ballpark sense of what your real world throughput could look like.
[...]

PR #2638, opened 2026-02-03
Profile-powered synthetic tx benchmarks (2/2)

huitseeker:
Companion PR of https://github.com/0xMiden/miden-base/pull/2391
----
This branch adds a synthetic transaction kernel benchmark so miden-vm can be tested without pulling in the full protocol, as discussed in issue #2256. It introduces a new benchmark crate with generators, validators, and bench targets that consume the VM profile snapshot exported by miden-base, which is treated as the only real workload benchmark. This keeps miden-vm performance testing representative while staying decoupled from protocol dependencies. 

Workflow across repos:
- Run the transaction kernel bench in miden-base and export the VM profile JSON.
- Copy that profile into `benches/synthetic-tx-kernel/profiles` in miden-vm.
- Run the synthetic benchmarks in miden-vm against that snapshot.


huitseeker:
Closing in favor of #3024

Issue #2636, opened 2026-02-02
Investigate removing all `Tracer` methods except for `{start,finalize}_clock_cycle()`


plafer:
All `Tracer` methods (except for `{start,finalize}_clock_cycle()`) could be removed, and the same functionality implemented in terms of `{start,finalize}_clock_cycle()`.

The current problem with all methods except `{start,finalize}_clock_cycle()` is that their exact semantics are complex and hard to document, specifically because they are called in the middle of the set of mutations that any given operation performs on the processor state. For example, when implementing `Tracer::start_context()`, it is currently unspecified whether the method is called *before* or *after* the processor state was mutated to be in the new context. In the new model, the implementer of `start_clock_cycle()` would detect that the clock cycle executes a `DYNCALL` operation (from the `continuation` argument), and run today's implementation of `Tracer::start_context()`.

This change would make the tracing semantics very simple and clean: at every clock cycle, `start_clock_cycle()` is called before the set of mutations, and `finalize_clock_cycle()` is called after the set of mutations.

PR #2642, opened 2026-02-03
Migrate parallel trace generation to the new `execute_impl()`

### LINKED ISSUE #2165: Parallel trace generation: switch to iterative execution

plafer:
Closes #2165

Follow-up to #2615, where we migrate the parallel trace generation to the new `execute_impl()` function that is generic over the processor (in the process fixing #2165 for free).

At a high level, in the new architecture, parallel trace generator is conceptually split between *execution* (which takes care of reading the `Replay` structures to update the stack, etc), and *trace filling* (which takes care of reading from the processor state before and after each clock cycle to generate the corresponding trace row). Two structs were then introduced: `ReplayProcessor` for execution, and `CoreTraceGenerationTracer` for tracing - which together replace the legacy `CoreTraceFragmentFiller` (which did both functions simultaneously). I strongly believe that this dissociation ultimately will make the processor code much easier to understand, as both concerns are no longer intermingled (though after a few refactorings, see notes below).

The `build_trace()` benchmark was not affected.

### Notes
- This is a first implementation of the new architecture, which ultimately reuses a lot of structs that were previously designed. In a future PR, I would like to better organize things to reflect the new architecture.
  - For example, the `CoreTraceFragmentContext` (and `ExecutionTracer` which is responsible for building it) lump all the state and replays together, which we later have to "untangle" in the trace generation phase (see `split_trace_fragment_context()` function). It would be much cleaner to generate the data in `ExecutionTracer` in a way that it expects to be consumed in `build_trace()`.
  - I've already opened a few issues
- The module structure was left as-is in order to reduce the size of the diff; we can address that with #2495).
- I was forced to add the `Processor` associated type to the `Tracer` trait; this will go away with #2636 
- In this PR, `Processor` was fully removed from the public interface - it is now an implementation detail of the `miden-processor` crate.
  - This allows us to reorganize it as we want in the future without any breaking changes to the crate

### Benchmark results:

`next`:
```
build_trace/blake3_1to1 time:   [92.247 ms 92.450 ms 92.674 ms]
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

current branch:
```
build_trace/blake3_1to1 time:   [94.076 ms 94.209 ms 94.353 ms]
                        change: [+1.6227% +1.9020% +2.1737%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
```
Though I attribute the difference to noise, as I ran the benchmarks multiple times, and in some the new branch was slightly faster, and in others `next` was slightly faster (~1-2% diff).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 - [BREAKING] `Host::on_debug()` and `Host::on_trace()` now take immutable references to `ProcessorState` ([#2639](https://github.com/0xMiden/miden-vm/pull/2639)).
 - [BREAKING] Removed unused `meta` field from `ExecutionTrace` and changed the constructor to take `ProgramInfo` ([#2639](https://github.com/0xMiden/miden-vm/pull/2639)).
 - [BREAKING] Migrate parallel trace generation to use the common `execute_impl()` ([#2642](https://github.com/0xMiden/miden-vm/pull/2642)).
+- [BREAKING] Removed unused `should_break` field from `AssemblyOp` decorator ([#2646](https://github.com/0xMiden/miden-vm/pull/2646)).
 
 ## 0.20.6 (2026-02-04)
 
--- a/core/src/mast/node/basic_block_node/arbitrary.rs
+++ b/core/src/mast/node/basic_block_node/arbitrary.rs
@@ -612,9 +612,8 @@ impl Arbitrary for AssemblyOp {
             prop::collection::vec(any::<char>(), 1..=20)
                 .prop_map(|chars| chars.into_iter().collect()),
             any::<u8>(),
-            any::<bool>(),
         )
-            .prop_map(|(has_location, context_name, op, num_cycles, should_break)| {
+            .prop_map(|(has_location, context_name, op, num_cycles)| {
                 use miden_debug_types::{ByteIndex, Location, Uri};
 
                 let location = if has_location {
@@ -623,7 +622,7 @@ impl Arbitrary for AssemblyOp {
                     None
                 };
 
-                AssemblyOp::new(location, context_name, num_cycles, op, should_break)
+                AssemblyOp::new(location, context_name, num_cycles, op)
             })
             .boxed()
     }
--- a/core/src/mast/serialization/asm_op.rs
+++ b/core/src/mast/serialization/asm_op.rs
@@ -69,7 +69,6 @@ impl AsmOpInfo {
         let mut reader = SliceReader::new(&asm_op_data[self.data_offset as usize..]);
 
         let num_cycles = reader.read_u8()?;
-        let should_break = reader.read_bool()?;
 
         let location = if reader.read_bool()? {
             let uri_idx = reader.read_usize()?;
@@ -87,7 +86,7 @@ impl AsmOpInfo {
         let op_idx = reader.read_usize()?;
         let op = string_table.read_string(op_idx)?;
 
-        Ok(AssemblyOp::new(location, context_name, num_cycles, op, should_break))
+        Ok(AssemblyOp::new(location, context_name, num_cycles, op))
     }
 }
 
@@ -138,9 +137,6 @@ impl AsmOpDataBuilder {
         // Serialize num_cycles (u8)
         self.asm_op_data.write_u8(asm_op.num_cycles());
 
-        // Serialize should_break (bool)
-        self.asm_op_data.write_bool(asm_op.should_break());
-
         // Serialize location
         if let Some(location) = asm_op.location() {
             self.asm_op_data.write_bool(true); // has_location = true
@@ -184,7 +180,7 @@ mod tests {
 
     #[test]
     fn test_asm_op_roundtrip_no_location() {
-        let asm_op = AssemblyOp::new(None, "test_context".to_string(), 5, "add".to_string(), false);
+        let asm_op = AssemblyOp::new(None, "test_context".to_string(), 5, "add".to_string());
 
         let mut builder = AsmOpDataBuilder::new();
         builder.add_asm_op(&asm_op);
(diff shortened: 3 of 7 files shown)


processor/src/lib.rs
```diff
@@ -23,7 +23,7 @@ mod stack;
 mod system;
 
 pub mod fast;
-pub mod processor;
+pub(crate) mod processor;
```
bobbinth, line 26: nit: this could be just `mod processor` - right? (i.e., `pub(crate)` in `lib.rs` is basically the same as `mod`).

processor/src/processor.rs
```diff
@@ -141,26 +132,13 @@ pub trait Processor: Sized {
...
-    // --------------------------------------------------------------------------------------------
-
-    /// Returns the next context ID that would be created given the current state.
-    ///
-    /// Note: This only applies to the context created upon a `CALL` or `DYNCALL` operation;
-    /// specifically the `SYSCALL` operation doesn't apply as it always goes back to the root
-    /// context.
-    fn next_ctx_id(&self) -> ContextId {
-        (self.system().clock() + 1).into()
-    }
```
bobbinth, line 156: Question: why remove this in favor of the `get_next_ctx_id()` function? Having this be a processor method is a bit more convenient since we don't need to import a separate function to use it.

processor/src/processor.rs
```diff
@@ -444,83 +428,3 @@ pub trait HasherInterface {
...
-
-// OPERATION HELPER REGISTERS
-// ================================================================================================
-
-/// Trait for computing helper registers for operations.
-///
-/// Concretely, the fast processor does not compute any helper registers, and thus returns a default
-/// value (all zeros). On the other hand, the trace generation helpers compute the actual helper
-/// registers to be included in the trace.
-pub trait OperationHelperRegisters {
```
bobbinth, line 456: Does removing this trait mean that we now always need to compute helper register values? But I guess the impact is pretty negligible?

processor/src/fast/operation.rs
```diff
@@ -362,123 +358,19 @@ impl StackInterface for FastProcessor {
...
-        DEFAULT_HELPERS
-    }
-}
-
 // HELPERS
 // ================================================================================================
 
 /// Identical to `[chiplets::ace::eval_circuit]` but adapted for use with `[FastProcessor]`.
-pub fn eval_circuit_fast_(
+pub(crate) fn eval_circuit_fast_(
```
bobbinth, line 373: nit (and not from this PR), but why does this function name end with an underscore?

processor/src/trace/parallel/tracer.rs
```diff
@@ -0,0 +1,562 @@
...
+    ) -> Option<ExecutionContextInfo> {
+        if let Continuation::StartNode(node_id) = &continuation {
+            let node = current_forest
+                .get_node_by_id(*node_id)
+                .expect("invalid node ID stored in continuation");
+
+            if let MastNode::Dyn(dyn_node) = node
+                && dyn_node.is_dyncall()
+            {
+                // Capture execution context info after dropping the top stack element.
```
huitseeker: Looking at the DYNCALL handling in `get_execution_context_for_dyncall`, I think there might be an issue with how `overflow_addr` is computed when the stack depth is exactly 16.

Per the execution context docs, DYNCALL drops one element first, then checks if overflow pops are needed. When `stack_depth == MIN_STACK_DEPTH`, the drop leaves us with exactly 16 elements - no overflow pop happens, so `overflow_addr` should be ZERO.

But the current code always peeks `stack_overflow_replay` to get `overflow_addr`, even when no actual overflow pop occurred. The replay queue contains recorded overflow pops from actual overflow operations (when stack depth was > 16). When we peek at this queue without having actually popped from overflow, we get either an old entry from a previous overflow pop or an entry meant for a future overflow pop - either way, it's wrong.

The old code in `ExecutionTracer` seemed to handle this correctly with `overflow_table.last_update_clk_in_current_ctx()`, which returns ZERO when there are no overflow entries in the current context.

A simple fix would be to only peek the replay when `stack_depth - 1 > MIN_STACK_DEPTH`:

```rust
let stack_depth_after_drop = processor.stack.stack_depth() - 1;
let stack_depth = core::cmp::max(stack_depth_after_drop, MIN_STACK_DEPTH) as u32;
let overflow_addr = if stack_depth_after_drop > MIN_STACK_DEPTH {
    match processor.stack_overflow_replay.peek_replay_pop_overflow() {
        Some((_, overflow_addr)) => *overflow_addr,
        None => ZERO,
    }
} else {
    ZERO
};
```

It would be great to add a test case for this to prevent regressions. I started writing one but realized the bug is tricky to catch at the integration level since the incorrect overflow_addr gets buried in decoder trace state that isn't easily asserted against — so I may have missed something. A unit test that directly checks `get_execution_context_for_dyncall` would be more direct, perhaps.

processor/src/trace/parallel/mod.rs
bobbinth, line 123: Not for this PR, but we should probably switch to using `miden_crypto::parallel` module here (which under the hood uses `p3_maybe_rayon`. This would allow us to use parallel iterators and get rid of the direct `rayon` dependency.

processor/src/processor.rs
plafer: My goal is to make the `Processor` trait as simple as possible - that is, a simple description of the processor state. This is where I'm hoping to get to with #2645 (and hopefully we can even remove the `execute_*_decorators()` methods).  

I find that adding a bunch of helper "provided" methods in the trait makes it less clear what we expect an abstract processor to be. This is also why I removed the `execute_sync_op()` provided method in the previous PR.

processor/src/trace/parallel/tracer.rs
plafer: Ah yes good catch!

- failing test: [c7bd4c1](https://github.com/0xMiden/miden-vm/pull/2642/commits/c7bd4c170d7eeea6d656ed4d1ae5a389c1cabb0d)
- fix: [5e98934](https://github.com/0xMiden/miden-vm/pull/2642/commits/5e989349a08af8eb4e318186527453cd40e91f38)

processor/src/processor.rs
plafer: No, we just changed how this information is passed around (now `OperationHelperRegisters` is an enum encoding the same information, one variant per method that we had in the trait)

Issue #2622, opened 2026-01-31
remove `should_break` AsmOp


huitseeker:
Not for this PR, but `should_break` probably won't work any more (and I'm not 100% sure it was working before this PR).

Basically, at some point, this was used to stop execution of the program at a given point by inserting a "breakpoint". Instead of having a dedicated decorator for this, we attached this as a property to the `AssemblyOp` decorator. Now that AsmOps are not evaluated on every cycle, having `should_break` is not really useful here - so, at the very least, we should remove it.

If we do want to bring this functionality back, this would probably have to be its own decorator - similar to the debug and trace decorators. But it is not clear to me if a `breakpoint` decorator is useful in the context of the new processor. cc @plafer.

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2606#discussion_r2749153563_

PR #2646, opened 2026-02-06
Remove `should_break` from `AssemblyOp`

bobbinth:
This PR removes the `should_break` field from `AssemblyOp` decorator. AFAICT, we no longer need this (and I don't think this is used in the compiler).

Closes https://github.com/0xMiden/miden-vm/issues/2622.


bitwalker (approved):
The debugger might have been respecting it, but I don't think it's actually needed by anything there

huitseeker (approved):
🙏 for the cleanup

PR #2650, opened 2026-02-07
feat: add error messages to asserts in the standard library

### LINKED ISSUE #1849: Consider adding error messages to asserts in std library
Consider adding error messages to bare asserts within the std library.

Concretely, I ran into the multiple leaf case in the smt code multiple times already and in this case, the error is a very unhelpful "error code 0" which makes searching for it basically impossible. It would be great to add error messages to _all_ forms of asserts in the std library to provide better errors which saves time during debugging.

One example:

https://github.com/0xMiden/miden-vm/blob/68917e2aca04c332f49cd4adf313afab2c871b8c/stdlib/asm/collections/smt.masm#L133-L135

In my opinion, we should have a CI lint that checks for bare asserts in all our repos for hand-written MASM, because these are low-hanging fruits that improve the DevX. Even better would be to disallow bare asserts in the first place, but that probably requires some more consideration.

huitseeker:
Adds asserts from a handful of generic messages to the asserts in the missing library which did not have any.

The reason I'm keeping it to a handful of messages is they are easy to map, and keep the error table of a manageable size.
If you feel strongly about asserts that should be more explicit, feel free to bikeshed here!

Fixes #1849.


Al-Kindi-0 (approved):
LGTM, thank you!

PhilippGackstatter (approved):
Looks good, thank you!

PR #2651, opened 2026-02-08
Processor module structure cleanup

bobbinth:
This PR continues work on #2495 - though doesn't quite finish it. This includes:

- Removed some duplicate code and un-used functions.
- Moved various trace-related modules under `trace` namespace.
- Cleaned up public exports from `miden-processor` crate.


Al-Kindi-0 (approved):
LGTM!

PR #2658, opened 2026-02-10
Add Plonky3 constraints evaluation pipeline

Leo-Besancon:
## Describe your changes

This PR aims to simplify the existing PR https://github.com/0xMiden/miden-vm/pull/2562. It closes parts of https://github.com/0xMiden/miden-vm/issues/2509. It isolates the structure to evaluate the constraints, provide stubs for constraints implementations, and feature flags to select which constraints to evaluate.

However, the constraints are currently empty (only the existing system and range constraints are kept), and will be added in a followup.


Cargo.toml
```diff
@@ -45,7 +45,7 @@ overflow-checks = true
 
 [workspace.dependencies]
 # Workspace crates
-miden-air = { path = "./air", version = "0.21", default-features = false }
+miden-air = { path = "./air", version = "0.21", default-features = false, features = ["human_readable"] }
```
huitseeker: `human_readable` is enabled by default in workspace-dependencies, but when this feature is active without any constraint features (`system_constraints`, `range_constraints`, etc.), the `eval()` function becomes a no-op that accepts any trace

I'd remove `human_readable` from the default features here, and add a compile-time guard that requires at least one constraint feature when `human_readable` is enabled

air/src/lib.rs
```diff
@@ -185,25 +271,100 @@ where
...
     }
 
+    #[cfg(not(any(
+        feature = "system_constraints",
+        feature = "chiplets_constraints",
+        feature = "range_constraints",
+        feature = "stack_constraints",
+        feature = "decoder_constraints"
+    )))]
+    fn eval<AB: MidenAirBuilder<F = Felt>>(&self, _builder: &mut AB) {}
```
huitseeker, line 249: This no-op `eval()` implementation is selected when `human_readable` is enabled but no constraint features are active.

air/src/lib.rs
```diff
@@ -185,25 +271,100 @@ where
...
+    }
+
+    #[cfg(not(any(
+        feature = "system_constraints",
+        feature = "chiplets_constraints",
+        feature = "range_constraints",
+        feature = "stack_constraints",
+        feature = "decoder_constraints"
+    )))]
+    fn eval<AB: MidenAirBuilder<F = Felt>>(&self, _builder: &mut AB) {}
```
huitseeker, line 311: This is the second no-op `eval()` implementation (for the `human_readable` code path). When no constraint features are enabled, this accepts any trace without verification.

air/src/constraints/mod.rs
```diff
@@ -6,29 +6,143 @@
...
+    let local: &MainTraceRow<AB::Var> = (*local).borrow();
+    let next: &MainTraceRow<AB::Var> = (*next).borrow();
+
+    let periodic_values: [_; NUM_PERIODIC_VALUES] =
+        builder.periodic_evals().try_into().expect("Wrong number of periodic values");
+
+    // FIXME: move these constants to a more appropriate place, and ensure they are consistent
+    // across all constraints modules
+    const MAX_BETA_CHALLENGE_POWER: usize = 15;
+    const AUX_WIDTH: usize = 8;
```
huitseeker: Could you use the actual auxiliary trace width (`AUX_TRACE_WIDTH`)  calculated from the trace layout?

Leo-Besancon: Thanks, using `AUX_TRACE_WIDTH` and `AUX_TRACE_RAND_ELEMENTS - 1` in the latest commit instead of these two constants!

air/src/lib.rs
Leo-Besancon: This should be gated by the impl Block `#[cfg(not(feature = "human_readable"))]`

Cargo.toml
Leo-Besancon: True, I've added the check to catch this. Regarding the number of features, I agree it's a bit too much. Maybe we can:
- only keep `human_readable`, `main_constraints`, `bus_constraints` in this PR
- and potentially add granularity once we do implement constraints (and only for the modules we see a point to toggle them on/off specifically

What do you think?

Leo-Besancon: Simplified the features in the latest commit.

Leo-Besancon:
Closing this PR for now as it will be heavily impacted by the current constraints (with tagging) work. We'll need to revisit the exact needs and take parts of this PR once the constraints are all done.

PR #2659, opened 2026-02-10
initial structure for VM constraints

Al-Kindi-0:
## Describe your changes

As the title says.
The constraints in this PR match the ones already existing on `next` and they are the minimal to ensure we reach the maximal constraints degree on both the main and auxiliary traces.

- Updated `CHANGELOG.md'


air/src/trace/mod.rs
```diff
@@ -121,6 +121,14 @@ pub const CHIPLETS_OFFSET: usize = RANGE_CHECK_TRACE_RANGE.end;
 pub const CHIPLETS_WIDTH: usize = 20;
 pub const CHIPLETS_RANGE: Range<usize> = range(CHIPLETS_OFFSET, CHIPLETS_WIDTH);
 
+/// Shared chiplet selector columns at the start of the chiplets segment.
+pub const CHIPLET_SELECTORS_RANGE: Range<usize> = range(CHIPLETS_OFFSET, 5);
+pub const CHIPLET_S0_COL_IDX: usize = CHIPLET_SELECTORS_RANGE.start;
```
Leo-Besancon, line 126: I would expect these to be in the `chiplets/mod.rs` module as they are part of the chiplets (whereas e.g. `CHIPLETS_WIDTH` above defines the size of the chiplets module in the global layout, so it is in the right place in my opinion).

air/src/constraints/range/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // Powers of 3: {1, 3, 9, 27, 81, 243, 729, 2187}
+    let one_expr = AB::Expr::ONE;
+    let three = AB::Expr::from_u16(3);
+    let nine = AB::Expr::from_u16(9);
+    let twenty_seven = AB::Expr::from_u16(27);
+    let eighty_one = AB::Expr::from_u16(81);
+    let two_forty_three = AB::Expr::from_u16(243);
+    let seven_twenty_nine = AB::Expr::from_u16(729);
+    let two_one_eight_seven = AB::Expr::from_u16(2187);
```
bobbinth, line 84: There is no way to declare these as constants, right?

bobbinth (approved):
Looks good! Thank you!

air/src/constraints/range/mod.rs
Al-Kindi-0: unfortunately not, we could introduce constants for the u16 values but not sure if this is any better

air/src/trace/mod.rs
Al-Kindi-0: Will reconsider in the followup PR

PR #2664, opened 2026-02-10
Optimize hasher op basic block cloning

### LINKED ISSUE #2250: `HasherOp`: improve handling of basic blocks

### LINKED ISSUE #2661: Optimize `HasherOp` to avoid cloning `Vec<OpBatch>` on every basic block entry
### Feature description

`HasherOp::HashBasicBlock` stores a full `Vec<OpBatch>` clone on every basic block entry. In loops this means thousands of deep clones of the same data. The consumer only needs `&[OpBatch]`. Since the data already lives in an immutable `Arc<MastForest>`, we can store `(Arc<MastForest>, MastNodeId)` instead — replacing a deep clone with an Arc increment.
Changes would look like this: 
1) from:
https://github.com/0xMiden/miden-vm/blob/a90c843c9f0e5fa0f53bb531af3e10ee37ed52b0/processor/src/trace/execution_tracer.rs#L475
to this: 
```rust
current_forest.clone(),
mast_node_id,
```
2) from:
https://github.com/0xMiden/miden-vm/blob/a90c843c9f0e5fa0f53bb531af3e10ee37ed52b0/processor/src/trace/parallel/mod.rs#L466-L467
to this:
```rust
HasherOp::HashBasicBlock((forest, node_id, expected_hash)) => {
     let op_batches = forest[node_id].unwrap_basic_block().op_batches();
     let _ = chiplets.hasher.hash_basic_block(op_batches, expected_hash);
```
3) And from:
https://github.com/0xMiden/miden-vm/blob/a90c843c9f0e5fa0f53bb531af3e10ee37ed52b0/processor/src/trace/trace_state.rs#L985
to:
```rust
HashBasicBlock((Arc<MastForest>, MastNodeId, Word)),
```
from
https://github.com/0xMiden/miden-vm/blob/a90c843c9f0e5fa0f53bb531af3e10ee37ed52b0/processor/src/trace/trace_state.rs#L1019-L1020
to this
```rust
    pub fn record_hash_basic_block(
        &mut self,
        forest: Arc<MastForest>,
        node_id: MastNodeId,
        expected_hash: Word,
    ) {
        self.hasher_ops
            .push_back(HasherOp::HashBasicBlock((forest, node_id, expected_hash)));
```

Speed up
Bench results:
Before:

Image

After:

Image

From 1.3x to 2.3x speedup

sashass1315:
## Describe your changes

`HasherOp::HashBasicBlock` now stores `(Arc<MastForest>, MastNodeId, Word)` instead of `(Vec<OpBatch>, Word)`, replacing a deep clone on every basic block entry with a cheap Arc increment.

Fixes #2661
Fixes #2250


plafer (approved):
LGTM, thank you!

I like this solution, as it requires no deep copy at all, unlike my suggestion in #2250 to store the op batches in some `BTreeMap<Digest, Vec<OpBatch>>` (which requires one deep copy per basic block).

The `program_execution_fast` benchmark is unaffected (as expected), while I get a ~15% improvement in the `program_execution_for_trace`:

```
program_execution/blake3_1to1
                        time:   [5.0757 ms 5.0804 ms 5.0856 ms]
                        change: [−16.413% −16.152% −15.923%] (p = 0.00 < 0.05)
                        Performance has improved.
```

huitseeker (approved):
Thanks a lot!

Issue #2634, opened 2026-02-02
Address feedback from 2615


plafer:
Feedback left on #2615 should be addressed

PR #2676, opened 2026-02-11
Remove `Tracer` arguments from `Processor` methods

plafer:
Work towards #2634

Addresses specifically https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2750575610. That is, we no longer pass the `Tracer` to any `Processor` methods. This makes the abstract execution more clear, in that it's no longer up to the `Processor` implementations to correctly call `Tracer` methods.


processor/src/execution/operations/eval_circuit.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    Ok(())
+}
+
+/// Evaluates an arithmetic circuit encoded in memory starting at `ptr`.
+///
+/// This reads `num_vars` quadratic extension field elements from memory (the READ section), then
+/// reads `num_eval` base field element gate encodings (the EVAL section), evaluating each gate
+/// in sequence. Returns the resulting [`CircuitEvaluation`] if the circuit evaluates to zero.
+pub(crate) fn eval_circuit_impl(
```
plafer, line 54: Note that the only reason this is `pub(crate)` is so that it can be used in trace generation tests in the ace chiplets module. When we address #2677, we can make it fully private as it should be.

PR #2680, opened 2026-02-12
Fix performance regression after 2642

plafer:
Fixes a performance regression introduced in #2642 to `program_execution_fast` benchmark (in that PR I had mistakenly only checked for regressions in the `program_execution_for_trace` benchmark).

Today I noticed a 7-8% performance regression on `next` for the `program_execution_fast` benchmark, and `git bisect`ed it down to #2642 (and specifically [5594c5e](https://github.com/0xMiden/miden-vm/pull/2642/commits/5594c5e7fc65c69490ad9128b577c615f889261e) as the offending commit). After doing some digging, I realized that the main thing that changed in the `program_execution_for_trace` loop is that `OperationHelperRegisters` were now being passed out of `execute_op()` and ultimately into `tracer.finalize_clock()` - they were previously simply [being dropped](https://github.com/0xMiden/miden-vm/blob/750b08216d74f96505b090facc2489201598ea47/processor/src/execution/basic_block.rs#L316-L327).

Though I figured that this still shouldn't affect the `program_execution_fast`, since it uses the `NoopTracer` which drops all its inputs in every method, and all our methods are inlined so the optimization passes should be able to optimize the returns away... Except that the newly introduced `get_continuation_after_executing_operation()` was not marked to be inlined. Marking it as `#[inline(always)]` recovered the previous performance. Honestly this is still a bit surprising to me, since `get_continuation_after_executing_operation()` is used in a closure that is being passed to `Stopper::should_stop()` (i.e. *not* `Tracer::finalize_clock_cycle()`) - though it clearly affected the ability to optimize regardless.

## Benchmarks

- `next`
```
program_execution_fast/blake3_1to1
                        time:   [2.2072 ms 2.2119 ms 2.2168 ms]
```

- this branch:
```
program_execution_fast/blake3_1to1
                        time:   [2.0546 ms 2.0600 ms 2.0657 ms]
                        change: [−7.1793% −6.8676% −6.5434%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## Note for the future

#2642 was "squash and merged" into one big commit, which *should* have made this analysis much harder. However, thankfully, I still had the individual commits locally, so I was able to continue the `git bisect` down to a much smaller diff. In the future, IMO for big PRs we should as much as possible create a merge commit of (clean) commits instead of squash-merging.


bitwalker:
> but keeping a clean commit history requires rebasing which is annoying while reviews are happening. 

I'm not convinced keeping a "clean" commit history is a useful objective, rather than requiring PR authors to clean up their commits prior to merging the PR. That's how I prefer to work personally, and it ensures that relevant history is preserved while keeping things relatively clean.

For stacked PRs, there aren't any good options other than giving up on not having cleanup commits and such in the history, because regardless of what you do you need to rebase when stacking PRs. Tools like `jj` make that a lot easier to do though, so it's less of a pain than it used to be.

- For simple PRs, we do what we do now - squash merging so that there is only one simple commit in history.

I'd say this is fine for external contributions, especially those generated by AI tools.

> - For large PRs, the PR author would need to clean up the history after the reviews have been addressed, and then we can merge the cleaned up commits.

💯 

> For the latter, I'd still try to keep the number of commits relatively small - e.g., under a dozen but ideally closer to half a dozen.

I'd say this only matters for the final state that gets merged, during review it can be as messy as the author likes.

PR #2681, opened 2026-02-12
Address comments leftover from 2615

### LINKED ISSUE #2634: Address feedback from 2615

plafer:
Closes #2634 

This PR addresses each remaining comment from #2615, with each commit addressing a comment:

1. Addresses https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2750604005
2. Addresses https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2749848086
3. Addresses https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2750018696
4. Addresses https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2750022987

However, **I would personally revert the last commit**, as converting from `Result` to `ControlFlow` in stable Rust is clunky, and the use of the new method `result_to_control_flow()` decreases readability. I find returning `ControlFlow` to be acceptable given that 
1. the `Processor` trait is an *internal* trait and only we implement all processors, 
2. they are all meant to be used in our `execute_impl()`, which uses `ControlFlow<BreakReason>` to manage control flow,
3. the previous version makes the code a lot more readable

### DIFF
--- a/processor/src/execution/call.rs
+++ b/processor/src/execution/call.rs
@@ -8,7 +8,7 @@ use crate::{
     continuation_stack::Continuation,
     execution::{
         ExecutionState, finalize_clock_cycle, finalize_clock_cycle_with_continuation,
-        get_next_ctx_id, result_to_control_flow,
+        get_next_ctx_id,
     },
     mast::{CallNode, MastForest, MastNodeExt, MastNodeId},
     operation::OperationError,
@@ -41,11 +41,9 @@ where
     );
 
     // Execute decorators that should be executed before entering the node
-    result_to_control_flow(state.processor.execute_before_enter_decorators(
-        current_node_id,
-        current_forest,
-        state.host,
-    ))?;
+    state
+        .processor
+        .execute_before_enter_decorators(current_node_id, current_forest, state.host)?;
 
     state.tracer.start_context();
     state.processor.save_context_and_truncate_stack();
@@ -138,9 +136,7 @@ where
         current_forest,
     )?;
 
-    result_to_control_flow(state.processor.execute_after_exit_decorators(
-        node_id,
-        current_forest,
-        state.host,
-    ))
+    state
+        .processor
+        .execute_after_exit_decorators(node_id, current_forest, state.host)
 }
--- a/processor/src/execution/dyn.rs
+++ b/processor/src/execution/dyn.rs
@@ -8,7 +8,7 @@ use crate::{
     continuation_stack::{Continuation, ContinuationStack},
     execution::{
         ExecutionState, InternalBreakReason, finalize_clock_cycle,
-        finalize_clock_cycle_with_continuation, get_next_ctx_id, result_to_control_flow,
+        finalize_clock_cycle_with_continuation, get_next_ctx_id,
     },
     mast::{MastForest, MastNodeId},
     processor::{MemoryInterface, Processor, StackInterface, SystemInterface},
@@ -39,12 +39,10 @@ where
     );
 
     // Execute decorators that should be executed before entering the node
-    result_to_control_flow(state.processor.execute_before_enter_decorators(
-        current_node_id,
-        current_forest,
-        state.host,
-    ))
-    .map_break(InternalBreakReason::from)?;
+    state
+        .processor
+        .execute_before_enter_decorators(current_node_id, current_forest, state.host)
+        .map_break(InternalBreakReason::from)?;
 
     let dyn_node = current_forest[current_node_id].unwrap_dyn();
 
@@ -215,9 +213,7 @@ where
         current_forest,
     )?;
 
-    result_to_control_flow(state.processor.execute_after_exit_decorators(
-        node_id,
-        current_forest,
-        state.host,
-    ))
+    state
+        .processor
+        .execute_after_exit_decorators(node_id, current_forest, state.host)
 }
--- a/processor/src/execution/join.rs
+++ b/processor/src/execution/join.rs
@@ -4,10 +4,7 @@ use core::ops::ControlFlow;
 use crate::{
     BreakReason, Host, Stopper,
     continuation_stack::Continuation,
-    execution::{
-        ExecutionState, finalize_clock_cycle, finalize_clock_cycle_with_continuation,
-        result_to_control_flow,
-    },
+    execution::{ExecutionState, finalize_clock_cycle, finalize_clock_cycle_with_continuation},
     mast::{JoinNode, MastForest, MastNodeId},
     processor::Processor,
     tracer::Tracer,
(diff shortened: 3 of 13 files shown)


processor/src/execution/mod.rs
```diff
@@ -411,6 +364,17 @@ impl From<BreakReason> for InternalBreakReason {
 // HELPERS
 // ================================================================================================
 
+/// Maps a decorator execution result to a [`ControlFlow<BreakReason>`].
+#[inline(always)]
+pub(super) fn result_to_control_flow(
```
huitseeker: Agree with the note in the PR description that commit d6dc6c283 should be reverted. The `result_to_control_flow()` helper adds an unnecessary conversion layer.

processor/src/execution/mod.rs
```diff
@@ -27,6 +27,23 @@ pub(crate) use external::finish_load_mast_forest_from_external;
...
 
+// EXECUTION STATE
+// ================================================================================================
+
+/// Bundles the common state threaded through all execution functions.
+///
+/// Note: `current_forest` is intentionally excluded from this struct. Including it would prevent
+/// passing node references (obtained from the forest) alongside `&mut ExecutionState` to
+/// functions, since both would borrow the same struct.
+pub(crate) struct ExecutionState<'a, P, H, S, T> {
```
bobbinth, line 38: nit: I'm not sure if "state" is the right suffix here because we are not really describing the "state". I'd probably go with "context" - i.e., `ExecutionContext`.

processor/src/execution/call.rs
```diff
@@ -18,120 +21,126 @@ use crate::{
...
+    result_to_control_flow(state.processor.execute_before_enter_decorators(
+        current_node_id,
+        current_forest,
+        state.host,
+    ))?;
 
-    tracer.start_context();
-    processor.save_context_and_truncate_stack();
+    state.tracer.start_context();
+    state.processor.save_context_and_truncate_stack();
```
bobbinth, line 49: If these calls always happen together, it may make sense to add a simple method on `ExecutionState`. Then here, it could be just:
```rust
state.start_context();
```

processor/src/execution/call.rs
```diff
@@ -18,120 +21,126 @@ use crate::{
...
-        return ControlFlow::Break(BreakReason::Err(e.with_context(current_forest, node_id, host)));
+    if let Err(e) = state.processor.restore_context() {
+        return ControlFlow::Break(BreakReason::Err(e.with_context(
+            current_forest,
+            node_id,
+            state.host,
+        )));
     }
-    tracer.restore_context();
+    state.tracer.restore_context();
```
bobbinth, line 128: Similar to the above comment: if these calls usually happen together, it may be good to add a wrapper for them so that here it could be something like:
```rust
state.restore_context()?;
```

processor/src/execution/join.rs
```diff
@@ -15,70 +18,72 @@ use crate::{
...
     finalize_clock_cycle_with_continuation(
-        processor,
-        tracer,
-        stopper,
+        state.processor,
+        state.tracer,
+        state.stopper,
         || Some(Continuation::AfterExitDecorators(node_id)),
         current_forest,
     )?;
```
bobbinth, line 77: Could `finalize_clock_cycle_with_continuation()` be a method on the `ExecutionState`? Then here it could be invoked like:
```rust
state.finalize_clock_cycle_with_continuation(
    || Some(Continuation::AfterExitDecorators(node_id)),
    current_forest,
)?;
```

processor/src/execution/dyn.rs
```diff
@@ -20,103 +20,104 @@ use crate::{
...
         );
 
         word
     };
 
     // Drop the memory address from the stack. This needs to be done before saving the context.
-    processor.stack_mut().decrement_size();
-    tracer.decrement_stack_size();
+    state.processor.stack_mut().decrement_size();
+    state.tracer.decrement_stack_size();
```
bobbinth, line 77: Similar to the other comments: if these are always called together, may make sense to encapsulate them into something like `ExecutionState::decrement_stack_size()`.

processor/src/execution/mod.rs
```diff
@@ -124,212 +141,148 @@ where
...
-                        host,
-                        tracer,
-                        stopper,
-                    )
-                    .map_break(InternalBreakReason::from)?,
-                    MastNode::Dyn(_) => r#dyn::start_dyn_node(
-                        processor,
+                    MastNode::Join(join_node) => {
+                        join::start_join_node(&mut state, join_node, node_id, current_forest)
+                            .map_break(InternalBreakReason::from)?
```
bobbinth, line 169: Probably not for this PR, but we are frequently passing `(node_id, current_forest)` tuple together. I wonder if there is a newtype we can create around these (though, don't know what it would be called yet).

bobbinth (commented):
Looks good! Thank you! I left some comments inline - most of them are about encapsulating functionality behind methods on the `ExecutionState` struct.

bobbinth:
> **I would personally revert the last commit**, as converting from `Result` to `ControlFlow` in stable Rust is clunky, and the use of the new method `result_to_control_flow()` decreases readability.

No strong preference from my end here. So, if you think it is better - let's revert this commit.

processor/src/execution/mod.rs
plafer: The problem with `ExecutionContext` is that it clashes with the other semantic we already implicitly assigned to an "execution context" through the existing `ExecutionContextInfo`, `ExecutionContextSystemInfo` (needs a refactor), and `ExecutionContextReplay`. That is, we already use "execution context" to refer to the "system execution context": memory context ID and `fn_hash`. It's even [in our docs](https://0xmiden.github.io/miden-vm/user_docs/assembly/execution_contexts.html?highlight=execution%20con#execution-contexts).

Hence `ExecutionState` was the next best name I could find - `processor`, `continuation_stack` and `kernel` encode the (exact) state of the execution.

Switching to `ExecutionContext` here would require renaming the other use of "execution context" (in code + docs) to remove the confusion. Or, we could use a different prefix to `*Context` - `ProcessingContext` or something.

processor/src/execution/call.rs
plafer: I experimented with adding `ExecutionState` methods, and ultimately I think it's best to hold off for now. 

Basically, I wouldn't just want to use `ExecutionState` for start/restore context; but rather I'd want an abstraction that groups the related "processor mutation" calls to the analogous tracer calls. So I tried including all the ones suggested in the comments here, as well as `increment_stack_size()` (since it doesn't make sense for `decrement_stack_size()` to be in there but not `increment_stack_size()`). But that meant that we had to propagate `ExecutionState` down to the "leaf" functions (e.g. `op_add()` which increments the stack size). And it made the `op_add()` signatures (and others) methods quite a bit more complex due to the generic parameters on `ExecutionState` - most of which are not used by these methods).

Basically I'd want to rethink whether this is really the abstraction instead of "blindly" using it everywhere - and this will depend on what the `Tracer` trait ends up looking like in its final form with #2636. So in its current form, `ExecutionState` solves the problem of reducing the number of related parameters in a bunch of methods - but it's still not clear whether it should do more than that.

processor/src/execution/mod.rs
plafer: Explicitly reverted the commit (i.e. 2 commits - add followed by revert) as a form of documentation for this experiment.

PR #2683, opened 2026-02-13
processor: make some modules public for debugger

djolertrk:
Simple PR that enables debugger to use modules regarding fast processor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.21.1 (2026-02-17)
+
+#### Changes
+
+- Re-exported `Continuation` from `miden-processor` to support the external debugger ([#2683](https://github.com/0xMiden/miden-vm/pull/2683)).
+
 ## 0.21.0 (2026-02-14)
 
 #### Major breaking changes
@@ -77,9 +83,6 @@
 - [BREAKING] Removed `breakpoint` instruction from assembly ([#2655](https://github.com/0xMiden/miden-vm/pull/2655)).
 - Removed FRI domain offset from `fri_ext2fold4` operation for Plonky3 compatibility ([#2670](https://github.com/0xMiden/miden-vm/pull/2670)).
 - [BREAKING] Removed `Tracer` arguments from `Processor` methods ([#2676](https://github.com/0xMiden/miden-vm/pull/2676)).
-- Consolidated testing hosts by merging `TestConsistencyHost` into `TestHost` and reusing the unified host in tests ([#2603](https://github.com/0xMiden/miden-vm/pull/2603)).
-- [BREAKING] Remove `Tracer` arguments from `Processor` methods ([#2676](https://github.com/0xMiden/miden-vm/pull/2676)).
-- Re-exported `Continuation` from `miden-processor` to support the external debugger ([#2683](https://github.com/0xMiden/miden-vm/pull/2683)).
 
 ## 0.20.6 (2026-02-04)
 


djolertrk:
cc @bitwalker

processor/src/fast/mod.rs
```diff
@@ -1020,7 +1020,7 @@ struct ExecutionContextInfo {
 // NOOP TRACER
 // ================================================================================================
 
-/// A [Tracer] that does nothing.
+/// A `Tracer` that does nothing.
```
plafer: Any reason for this? A link [`Tracer`] would be better

processor/src/lib.rs
```diff
@@ -12,12 +12,12 @@ use core::{
...
 };
 
-mod continuation_stack;
+pub mod continuation_stack;
 mod debug;
 mod errors;
 mod execution;
 mod execution_options;
-mod fast;
+pub mod fast;
```
plafer: What do you need exposed? Would be better to re-export whatever it is you need (under the `RE-EXPORTS` banner starting at line 41).

djolertrk: I agree. I will add `continuation_stack::Continuation` and `fast::Memory` there, thanks!

processor/src/fast/mod.rs
djolertrk: It (the docs) was failing locally, will revert this.

djolertrk:
As discussed with @bitwalker, targeting this against `main` branch.

processor/src/lib.rs
```diff
@@ -41,9 +41,10 @@ mod tests;
 // RE-EXPORTS
 // ================================================================================================
 
+pub use continuation_stack::Continuation;
 pub use errors::{ExecutionError, MemoryError};
 pub use execution_options::{ExecutionOptions, ExecutionOptionsError};
-pub use fast::{BreakReason, ExecutionOutput, FastProcessor, ResumeContext};
+pub use fast::{BreakReason, ExecutionOutput, FastProcessor, Memory, ResumeContext};
```
bobbinth: Question: why do we need `Memory` to be exported? I'm assuming this is to read memory state of the processor for debugging purposes, right? If so, maybe `ProcessorState` (accessed via `FastProcessor::state()` method) may be a better option? Or is something missing from `ProcessorState`?

bobbinth (approved):
Looks good! Thank you! I left one question inline - but other than that, this looks good.

processor/src/lib.rs
bitwalker: We don't actually need to export this, `ProcessorState` has what we need. @djolertrk see `ProcessorState::get_mem_(state|value|word|addr_range)`. If you remove that export, we can get this merged.

CHANGELOG.md
```diff
@@ -77,6 +77,9 @@
 - [BREAKING] Removed `breakpoint` instruction from assembly ([#2655](https://github.com/0xMiden/miden-vm/pull/2655)).
 - Removed FRI domain offset from `fri_ext2fold4` operation for Plonky3 compatibility ([#2670](https://github.com/0xMiden/miden-vm/pull/2670)).
 - [BREAKING] Removed `Tracer` arguments from `Processor` methods ([#2676](https://github.com/0xMiden/miden-vm/pull/2676)).
+- Consolidated testing hosts by merging `TestConsistencyHost` into `TestHost` and reusing the unified host in tests ([#2603](https://github.com/0xMiden/miden-vm/pull/2603)).
+- [BREAKING] Remove `Tracer` arguments from `Processor` methods ([#2676](https://github.com/0xMiden/miden-vm/pull/2676)).
+- Re-exported `Continuation` from `miden-processor` to support the external debugger ([#2683](https://github.com/0xMiden/miden-vm/pull/2683)).
```
bobbinth: I think lines 80 - 81 should be removed from here, and line 82 moved up to be in v0.21.1 section.

djolertrk: addressed! thanks!

bobbinth (approved):
Looks good! Thank you!

PR #2704, opened 2026-02-15
fix: resolve imported computed constants with non-pub dependencies

### LINKED ISSUE #2696: Imported computed constants fail assembly
If the `memory.masm` module has the following definition:

```
const ACCOUNT_ID_AND_NONCE_OFFSET=4
pub const ACCOUNT_ID_SUFFIX_OFFSET = ACCOUNT_ID_AND_NONCE_OFFSET + 2
```

and then `ACCOUNT_ID_SUFFIX_OFFSET` is imported in another module and used there, assembly fails with:

```
   204 | const ACCOUNT_ID_AND_NONCE_OFFSET=4
   205 | const ACCOUNT_ID_SUFFIX_OFFSET = ACCOUNT_ID_AND_NONCE_OFFSET + 2
       :                                  ^^^^^^^^^^^^|^^^^^^^^^^^^^^
       :                                              `-- this symbol path could not be resolved
```

Even though `ACCOUNT_ID_AND_NONCE_OFFSET` is clearly defined. It seems the problem is that `use` in another module. If the `use` is removed, it assembles fine. So the linker probably doesn't "see" `ACCOUNT_ID_AND_NONCE_OFFSET` in the other module and that's why it fails?

Just wanted to document this in case it is not a known issue. If this is a known and accepted limitation, feel free to close. It's possible to work around this, but it would be nice to have this and if not, it'd be nice to make the error clearer.

avorylli:
Fixes #2696

Imported computed constants that depend on non-public constants from their defining module now resolve correctly during linking. The linker uses the defining module's context when evaluating constant expressions, allowing access to non-pub constants.

- Added test case to verify the fix
- Updated comments to clarify constant resolution logic

### DIFF
--- a/crates/assembly/src/linker/rewrites/module.rs
+++ b/crates/assembly/src/linker/rewrites/module.rs
@@ -263,16 +263,13 @@ impl<'a, 'b: 'a> ConstEnvironment for ModuleRewriter<'a, 'b> {
         };
 
         let constant_module = gid.module;
-        // If this constant is from a different module and the stack is empty (meaning we're
-        // starting to evaluate a constant from another module), push it onto the stack.
-        // This ensures that when dependencies are resolved recursively, they use the correct
-        // module context. The stack will be managed by on_eval_start/on_eval_completed for
-        // proper cleanup.
+        // If this constant is from a different module, push it onto the stack so that
+        // dependencies are resolved in the correct context. We only push if it's not already
+        // at the top of the stack to avoid duplicates, but we allow multiple modules on the
+        // stack for nested constant evaluations.
         if constant_module != self.module_id {
             let mut modules = self.evaluating_constant_modules.borrow_mut();
-            if modules.is_empty() {
-                // Only push if stack is empty - this means we're starting to evaluate
-                // a constant from another module, not resolving a dependency
+            if modules.last() != Some(&constant_module) {
                 modules.push(constant_module);
             }
         }


bitwalker:
After looking at the original issue again, it appears to me that the problem isn't actually due to symbol resolution. The code in that issue should have been fully const-evaluated during semantic analysis, leaving no references to the non-public constant to be resolved at link-time, where context would matter. That leads me to believe that there is an issue with const-eval during semantic analysis that is causing us to leave those constants unevaluated or only partially-evaluated.

I would confirm that first before introducing any other changes to how symbol resolution is performed.

The one other subtlety we'll need to test on top of this is that constants which refer to other private local constants, which in turn reference external constants, are partially-evaluated at least far enough that the reference to the private constant is rewritten to just refer to the external constant. This way, by the time semantic analysis is complete, there are no exported constants which still refer to private constants, making the symbol resolution aspects of this irrelevant.

plafer:
> After looking at the original issue again, it appears to me that the problem isn't actually due to symbol resolution. The code in that issue should have been fully const-evaluated during semantic analysis, leaving no references to the non-public constant to be resolved at link-time, where context would matter. That leads me to believe that there is an issue with const-eval during semantic analysis that is causing us to leave those constants unevaluated or only partially-evaluated.

In an effort to reduce the number of open PRs, I will close this one since it is not currently clear whether this is the right approach. *However*, if after further analysis (by either @bitwalker, @avorylli or whoever else) this turns out to be appropriate, we can just re-open this one.

Issue #2706, opened 2026-02-15
replace unwrap() with error handling for Tokio runtime creation


CreeptoGengar:
Currently, sync wrapper functions use unwrap() when creating Tokio runtimes, which can panic if runtime creation fails (e.g., due to resource constraints).

I can make changes in:
- execute_sync() in processor/src/lib.rs
- step_sync(), execute_by_step_sync(), execute_sync(), execute_for_trace_sync(), execute_sync_mut() in processor/src/fast/mod.rs.

plafer:
The methods work as documented; different design decisions are being explored with #2501.

PR #2707, opened 2026-02-15
fix: handle Tokio runtime creation errors in sync wrappers

CreeptoGengar:
Fixes https://github.com/0xMiden/miden-vm/issues/2706

Replace unwrap() calls with proper error handling when creating Tokio runtimes in sync wrapper functions. Runtime creation can fail due to resource constraints or system limitations, and these errors should be propagated as ExecutionError instead of panicking.

Applied to 6 locations: execute_sync, step_sync, execute_by_step_sync, execute_sync (FastProcessor), execute_for_trace_sync, and execute_sync_mut.


plafer:
See https://github.com/0xMiden/miden-vm/issues/2706#issuecomment-3916182179

PR #2713, opened 2026-02-16
Initial tagging infrastructure

Al-Kindi-0:
## Describe your changes

Matching PR for https://github.com/0xMiden/air-script/pull/539

- Updated `CHANGELOG.md'

### DIFF
--- a/air/src/constraints/tagging/ood_eval.rs
+++ b/air/src/constraints/tagging/ood_eval.rs
@@ -107,10 +107,15 @@ impl OodEvalAirBuilder {
 
     /// Panics if any ID in `0..=CURRENT_MAX_ID` was not recorded.
     pub fn assert_complete(&self) {
-        for (id, entry) in self.used.iter().enumerate() {
-            if entry.is_none() {
-                panic!("missing constraint id {id}");
-            }
+        let missing: Vec<usize> = self
+            .used
+            .iter()
+            .enumerate()
+            .filter_map(|(id, entry)| entry.is_none().then_some(id))
+            .collect();
+
+        if !missing.is_empty() {
+            panic!("missing constraint ids: {missing:?}");
         }
     }
 
--- a/air/src/constraints/tagging/tagged_builder.rs
+++ b/air/src/constraints/tagging/tagged_builder.rs
@@ -34,10 +34,15 @@ impl<AB> TaggedAirBuilder<AB> {
 
     /// Panics if any ID in `0..=CURRENT_MAX_ID` was not recorded.
     pub fn assert_complete(&self) {
-        for (id, entry) in self.used.iter().enumerate() {
-            if entry.is_none() {
-                panic!("missing constraint id {id}");
-            }
+        let missing: Vec<usize> = self
+            .used
+            .iter()
+            .enumerate()
+            .filter_map(|(id, entry)| entry.is_none().then_some(id))
+            .collect();
+
+        if !missing.is_empty() {
+            panic!("missing constraint ids: {missing:?}");
         }
     }
 


air/src/constraints/tagging/ood_eval.rs
```diff
@@ -0,0 +1,238 @@
...
+        let prev_enabled = state::is_enabled();
+        state::set_enabled(true);
+
+        let mut rng = SeededRng::new(seed);
+        let main = RowMajorMatrix::new(
+            (0..crate::trace::TRACE_WIDTH * 2).map(|_| rng.next_felt()).collect(),
+            crate::trace::TRACE_WIDTH,
+        );
+        let preprocessed = RowMajorMatrix::new(Vec::new(), 1);
+        let permutation = RowMajorMatrix::new(
```
adr1anh, line 56: Nit: [RowMajorMatrix::rand](https://github.com/Plonky3/Plonky3/blob/f932d76564eb40c079d9e1a0c1e616966cb8c6d5/matrix/src/dense.rs#L527)

air/src/constraints/tagging/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct TagRecord {
+    /// Stable numeric ID (zero-based).
+    pub id: usize,
+    /// Human-readable namespace for debugging (e.g., "range.main.v.transition").
+    pub namespace: &'static str,
+}
+
+/// Validate ordering, range, and uniqueness of a single tag.
+fn validate_tag(used: &mut [Option<&'static str>], expected: usize, tag: TagRecord) {
```
adr1anh: Nit: this could be a method on TagRecord.

air/src/constraints/tagging/state.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Active tagged block metadata.
+#[derive(Debug)]
+struct TagContext {
+    namespace: &'static str,
+    ids: Vec<usize>,
+    next: usize,
+}
+
+thread_local! {
```
adr1anh, line 19: Should we consider making `TAGGING_ENABLED` a constant rather than local? I think these checks could impact constraint evaluation.

adr1anh: Another option could be to remove the blanket implementation of `TaggingAirBuilderExt` and instead provide an implementation which just forwards to the inner `AirBuilder`.

air/src/constraints/tagging/state.rs
```diff
@@ -0,0 +1,88 @@
...
+    TAGGING_ENABLED.with(|flag| flag.set(enabled));
+}
+
+/// Run `f` under a tagged context with the provided IDs.
+///
+/// This enforces:
+/// - exactly one tagged context at a time (no nesting),
+/// - at least one ID,
+/// - and the number of emitted assertions matches the ID list length.
+pub fn with_tag<R>(ids: Vec<usize>, namespace: &'static str, f: impl FnOnce() -> R) -> R {
```
adr1anh, line 40: Maybe someone more familiar with Rust can chime in, but are we using the right `Fn` bound here?

Al-Kindi-0: Not sure about the question but `FnOnce` here is the least restrictive and is enough for what we need

air/src/constraints/tagging/state.rs
Al-Kindi-0: At the moment I am not too concerned with performance, as long as it is not substantial. 

My main objective is to integrate the full set of constraints, and anything that makes that easier/smoother is in scope.

Al-Kindi-0: Using a wrapper might be better

Al-Kindi-0: I have thought about this and it seems to me that the least invasive solution is the current one. All other solutions either imply more boiler plate code, or many changes outside of the `air` crate.

As for performance considerations, I believe there are no overheads in `release` mode and the only cost is when running the specific tagging parity test.

air/src/constraints/tagging/mod.rs
Al-Kindi-0: done

air/src/constraints/tagging/ood_eval.rs
Al-Kindi-0: We need a a seedable RNG and I went with the simplest solution for now

air/src/constraints/tagging/state.rs
adr1anh: I just saw the conditional compilation flag so I agree!

air/src/constraints/tagging/tagged_builder.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    pub fn records(&self) -> &[TagRecord] {
+        &self.records
+    }
+
+    /// Panics if any ID in `0..=CURRENT_MAX_ID` was not recorded.
+    pub fn assert_complete(&self) {
+        for (id, entry) in self.used.iter().enumerate() {
+            if entry.is_none() {
+                panic!("missing constraint id {id}");
```
Leo-Besancon: Nit: Consider not panicking early and display all the missing ids, may be helpful if we have a lot of constraints.

air/src/constraints/tagging/ood_eval.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    pub fn records(&self) -> &[EvalRecord] {
+        &self.records
+    }
+
+    /// Panics if any ID in `0..=CURRENT_MAX_ID` was not recorded.
+    pub fn assert_complete(&self) {
+        for (id, entry) in self.used.iter().enumerate() {
+            if entry.is_none() {
+                panic!("missing constraint id {id}");
```
Leo-Besancon: same here

air/src/constraints/tagging/tagged_builder.rs
Al-Kindi-0: done

air/src/constraints/tagging/ood_eval.rs
Al-Kindi-0: done

Leo-Besancon (approved):
LGTM!

Soulthym:
Looks good to me too!

PR #2715, opened 2026-02-16
refactor: call Tracer::finalize_clock_cycle() after clock increment

### LINKED ISSUE #2682: Investigate: call `Tracer::finalize_clock_cycle()` after the clock was incremented
Today, `Tracer::finalize_clock_cycle()` is called after all state mutations have been applied for the given clock cycle *except* for the processor clock. This was originally done so that one could use `processor.system().clk()` to refer to the current clock.

However, these semantics create a bit of complexity (in that "everything except 1 thing" has been mutated), and are inconsistent with `Stopper::should_stop()` (which is called after *all* mutations have been applied, including the processor clock.

I think it would be better to make that change, though it still requires a bit more thought. Opinions are welcome.

DeVikingMark:
Call `Tracer::finalize_clock_cycle()` after clock increment for consistency 
with `Stopper::should_stop()`. Updates documentation and fixes clock usage 
in trace generation.

Fixes #2682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 0.22.0 (TBD)
 
+## 0.21.1 (2026-02-17)
+
+#### Changes
+
+- Re-exported `Continuation` from `miden-processor` to support the external debugger ([#2683](https://github.com/0xMiden/miden-vm/pull/2683)).
+
 ## 0.21.0 (2026-02-14)
 
 #### Major breaking changes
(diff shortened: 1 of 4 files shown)


plafer:
This issue is less about the actual change (as done in this PR), but more thinking about if this change is desired.

This issue is assigned to me for that reason. As per our `CONTRIBUTING.md`, you should only open PRs for issues that are assigned to you.

PR #2679, opened 2026-02-12
Update design docs

Al-Kindi-0:
## Describe your changes
Includes also uniformization of chiplets' bus messages, including changes in the docs and the code.

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](../CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #2719, opened 2026-02-18
Remove tests that check for multiple writes in same clock cycle

### LINKED ISSUE #2276: `FastProcessor` memory allows for multiple writes in same clock cycle

plafer:
Closes #2276

Closing this one as a non-issue (and removing the tests related to that functionality). The tests were quite artificial - no one can write a program that executes `op_mstore()` twice in the same clock cycle. Instead, whether the same memory location was written multiple times in the same clock cycle is a property of our operations, and does *not* need to be checked explicitly by the processor every time. That is, we should check each operation individually and ensure that this property is maintained (and add checks if there's a risk that it's not), rather than checking "generally" on each memory write in the processor.

I've only identified the `crypto_stream` operation as having inputs that could result in 2 writes to the same location in the same clock cycle - and in the processor we [already check for this condition](https://github.com/0xMiden/miden-vm/blob/a8a291726270ea70f9befaedb63663680993d210/processor/src/execution/operations/crypto_ops/mod.rs#L617-L628). However, @Al-Kindi-0 is this something that is also handled by our constraints/buses (i.e. if someone uses a malicious prover, will proof verification fail?) - I didn't find the `CryptoStream` operation in our docs (we probably need an issue for that).

### DIFF
--- a/processor/src/execution/operations/io_ops/tests.rs
+++ b/processor/src/execution/operations/io_ops/tests.rs
@@ -384,50 +384,6 @@ fn test_op_pipe() {
     assert_eq!(expected, processor.stack_top());
 }
 
-// CLOCK CYCLE CONFLICT TESTS
-// --------------------------------------------------------------------------------------------
-
-/// Ensures that reading and writing in the same clock cycle results in an error.
-#[test]
-#[ignore = "Re-enable when addressing issue 2276"]
-fn test_read_and_write_in_same_clock_cycle() {
-    let mut processor = FastProcessor::new(StackInputs::default());
-    let mut tracer = NoopTracer;
-
-    assert_eq!(0, processor.memory().num_accessed_words());
-
-    // emulate reading and writing in the same clock cycle (no increment_clk between operations)
-    op_mload(&mut processor, &mut tracer).unwrap();
-    assert!(op_mstore(&mut processor, &mut tracer).is_err());
-}
-
-/// Ensures that writing twice in the same clock cycle results in an error.
-#[test]
-#[ignore = "Re-enable when addressing issue 2276"]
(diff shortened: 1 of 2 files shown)


Al-Kindi-0:
I think docs for `crypto_stream` have been added in #2679.
As for enforcing this via constraints, I think it is done somewhere or at least doable.

huitseeker (approved):
Thanks! LGTM

- I'd love to see a rustdoc in `processor/src/fast/memory.rs` (probably `FastProcessor` module-level doc) documenting the invariant that multiple reads are allowed but multiple writes are not (in the same clock cycle),
- I didn't see the `CryptoStream` doc in #2679 .

plafer:
> I'd love to see a rustdoc in processor/src/fast/memory.rs (probably FastProcessor module-level doc) documenting the invariant that multiple reads are allowed but multiple writes are not (in the same clock cycle),

Documented it as part of the `Memory` struct docstring. The docstring now refers to the memory chiplet constraints as the reason for why these invariants exist (which the interested reader can refer to for full details).

> I didn't see the CryptoStream doc in https://github.com/0xMiden/miden-vm/pull/2679 .

It's [here](https://github.com/0xMiden/miden-vm/blob/d57890dfa8aa8240d34823a6e529dfc9bc1b4821/docs/src/design/stack/crypto_ops.md#cryptostream).

PR #2656, opened 2026-02-09
[DO NOT MERGE] experiment: only use start/finalize_clock_cycle methods in ExecutionTracer

plafer:
Work towards #2636

While conceptually cleaner, this affects the performance considerably of the `program_execution_for_trace` benchmark (~60% worse). Essentially, with this approach, we end up redoing too much work in the start/finalize methods.

Next steps: figure out how to cleanup the `Tracer` API so that it's well defined and conceptually easy to understand - without affecting the performance of `ExecutionTracer`.

## Benchmark results
- `next`

```
program_execution/blake3_1to1
                        time:   [5.9561 ms 5.9679 ms 5.9814 ms]
```

- this branch

```
program_execution/blake3_1to1
                        time:   [9.6478 ms 9.6659 ms 9.6850 ms]
                        change: [+61.500% +61.965% +62.465%] (p = 0.00 < 0.05)
                        Performance has regressd.
```

PR #2720, opened 2026-02-20
Refactor `Tracer` trait

### LINKED ISSUE #2636: Investigate removing all `Tracer` methods except for `{start,finalize}_clock_cycle()`

plafer:
Closes #2636

This PR contains my latest suggestion for what I believe the `Tracer` trait should look like. #2656 showed us that the original suggestion in #2636 to boil down the `Tracer` trait to only 2 methods, despite being *doable*, results in significant performance degradation.

Hence, this PR suggests a new approach: the `start_clock_cycle()` and `finalize_clock_cycle()` are as before, while every other method is considered an "in-cycle" method with the guarantee that at most one in-cycle method will be called between `start_clock_cycle()` and `finalize_clock_cycle()` (i.e. in the middle of a clock cycle). This is to prevent implementations from unknowingly depending on undocumented ordering between in-cycle methods - which we had in `ExecutionTracer`! Specifically, during `dyncall`, the `ExecutionTracer` relied on `Tracer::decrement_stack_size()` (which removed the memory address from the top of the stack, and potentially mutated the overflow table) to be called *before* `restore_context()` (which created a new overflow stack in the overflow table) - despite this relationship between both calls being undocumented. Most of this is documented (and more) in the `Tracer` docstring.

`Tracer::record_mast_forest_resolution()` is an (unfortunate) exception to the above rule.  I found that making it respect the rule would require splitting it into too many methods (ultimately creating noise), and I'm not sure it's necessary (as documented in the `Tracer` docstring). The next best thing would be to get rid of it - suggestions welcome.

## Commit structure

The first 2 commits are rather uncontroversial: they remove the `Tracer::increment_stack_size()`, `Tracer::decrement_stack_size()`, `Tracer::start_context()` and `Tracer::restore_context()` - for which the removal doesn't cause performance degradation, and ultimately drastically reduce the number of `Tracer` calls per clock cycle (in some sense enabling the above "in-cycle method" property).

The last commit refactors the remaining `Tracer` methods to implement the proposed "at most one in-cycle method call per clock cycle" property.

*EDIT*: Added a commit that provides an empty default implementation of all in-cycle `Tracer` methods - since most will probably be unused by most implementations.

## Benchmarks

I noticed a performance degradation in the `build_trace()` benchmark of about 3% (just due to the last commit) - which is small enough to make me wonder whether this is just noise. I will look into it more to see if I can find the reason (which in principle I believe there shouldn't be one).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Consolidated error variants: simplified `AceError` and FRI errors to string-based types, merged `DynamicNodeNotFound`/`NoMastForestWithProcedure` into `ProcedureNotFound`, introduced `HostError` for handler-related variants ([#2675](https://github.com/0xMiden/miden-vm/pull/2675)).
 - [BREAKING] `Processor` and `FastProcessor` decorator execution is now immutable ([#2718](https://github.com/0xMiden/miden-vm/pull/2718)).
+- [BREAKING] `Tracer` API significantly refactored ([#2720](https://github.com/0xMiden/miden-vm/pull/2720)).
 
 #### Fixes
 


processor/src/execution/operations/stack_ops/mod.rs
```diff
@@ -16,31 +16,29 @@ mod tests;
 pub(super) fn op_push<P, T>(
     processor: &mut P,
     element: Felt,
-    tracer: &mut T,
+    _tracer: &mut T,
```
bobbinth: Here, and in other similar places, why do we keep the `tracer` parameter if it is not used any more?

processor/src/execution/operations/sys_ops/mod.rs
```diff
@@ -21,7 +21,7 @@ pub(super) fn op_assert<P, T>(
     processor: &mut P,
     err_code: Felt,
     program: &MastForest,
-    tracer: &mut T,
+    _tracer: &mut T,
```
bobbinth: Same comment as above.

processor/src/execution/operations/stack_ops/mod.rs
plafer: Oops that was an oversight - now fixed

plafer:
> With this PR, we made most operation handlers "tracer-independent". I wonder if there is a good way to push this idea even further somehow to make even more (and maybe even all) operation handlers tracer-independent. I don't have any concrete suggestions though.

This is going back to the "API cleanliness vs performance" tradeoff. If we only had `Tracer::{start_clock_cycle, finalize_clock_cycle}()` as methods, then all operation handlers would be tracer-independent - but performance would take a hit. 

So the next steps would be taking more time to benchmark on a case-by-case basis which operation we can afford to remove the corresponding `Tracer` "in-cycle" method without affecting performance too much. Note that just using the `blake3_1to1` benchmark here doesn't work, since that benchmark doesn't exercise all operations. So this is probably a low priority for a while.

PR #2727, opened 2026-02-23
fix: remove dead imm == 0 checks in u32rotr and prepare_bitwise

milord12345:
## Describe your changes

Both u32rotr and prepare_bitwise in u32_ops.rs match on Some(0) first and emit a Noop. So the next arm Some(imm) never gets imm = 0.

But both still have imm == 0 || in their range checks — this is dead code. The error messages also say 0..=N which is wrong since 0 is handled above as a valid no-op.

Removed the dead imm == 0 checks and fixed the error range to 1..=N.

No behavioral change, just cleanup.

- [x] Repo forked and branch created from next
- [x] Commit messages and codestyle follow [conventions](../CONTRIBUTING.md)
- [x] Self-review completed

### DIFF
--- a/crates/assembly/src/instruction/u32_ops.rs
+++ b/crates/assembly/src/instruction/u32_ops.rs
@@ -334,10 +334,10 @@ pub fn u32rotr(
             span_builder.push_op(Noop);
         },
         Some(imm) => {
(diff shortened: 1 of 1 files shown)


plafer:
Unfortunately, due to the limited review capacity of our team, we will no longer be able to review PRs associated with issues of low priority such as this one.

PR #2731, opened 2026-02-23
Update caller instruction reference

### LINKED ISSUE #2711: `caller` docs include unused parameter
The docs for the `caller` instruction include a `b` parameter that is not otherwise explained:

https://github.com/0xMiden/miden-vm/blob/ab0cd8a9617a5639fff1a590129d5d4ee43f3f45/docs/src/user_docs/assembly/instruction_reference.md?plain=1#L176

Is it actually used or does it need to be removed from here? cc @plafer

plafer:
Closes #2711

I believe the `b` parameter was originally placed as some "delimiter" - it is not actually needed.


Al-Kindi-0 (approved):
LGTM

PR #2736, opened 2026-02-24
Stack shuffling ops constraints

Al-Kindi-0:
## Describe your changes

Paired with https://github.com/0xMiden/air-script/pull/545

- Updated `CHANGELOG.md'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,8 @@
 - [BREAKING] `Tracer` API significantly refactored ([#2720](https://github.com/0xMiden/miden-vm/pull/2720)).
 - Added optional tagging instrumentation for AIR constraints (test-only; enables stable ID tracking and OOD parity checks) ([#2713](https://github.com/0xMiden/miden-vm/pull/2713)).
 - Added general stack transition constraints (shift/no‑shift) ([#2725](https://github.com/0xMiden/miden-vm/pull/2725)).
-- Added optional tagging instrumentation for AIR constraints (test-only; enables stable ID tracking and OOD parity checks) ([#2713](https://github.com/0xMiden/miden-vm/pull/2713)).
 - Added stack overflow table constraints ([#2735](https://github.com/0xMiden/miden-vm/pull/2735)).
-- Added general stack transition constraints (shift/no‑shift) ([#2725](https://github.com/0xMiden/miden-vm/pull/2725)).
-- Added optional tagging instrumentation for AIR constraints (test-only; enables stable ID tracking and OOD parity checks) ([#2713](https://github.com/0xMiden/miden-vm/pull/2713)).
-- Added general stack transition constraints (shift/no‑shift) ([#2725](https://github.com/0xMiden/miden-vm/pull/2725)).
+- Added stack shuffling ops constraints ([#2736](https://github.com/0xMiden/miden-vm/pull/2736)).
 
 #### Fixes
 


air/src/constraints/stack/ops/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+    emit(builder, &mut idx, is_movdn2 * (s2_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn3 * (s3_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn4 * (s4_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn5 * (s5_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn6 * (s6_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn7 * (s7_next.clone() - s0.clone()));
+    emit(builder, &mut idx, is_movdn8 * (s8_next.clone() - s0.clone()));
+
+    // SWAPW
+    emit_list(
```
Al-Kindi-0: Do you find this a bit better readability-wise @adr1anh @Leo-Besancon  ?

Leo-Besancon: I think it's better yes! This is less problematic because it's a private helper, but I think the namings `emit` and `emit_list` are a bit confusing because they hide the usage of `STACK_OPS_BASE_ID` and `STACK_OPS_NAMES`.
Do you think about handling most of the constraints this way? Maybe having a generic helper `emit` that takes the id offset and names as argument, and then implementing something like `emit_stack_ops` that delegates to it would be better?

Al-Kindi-0: Good point!
I have now gone ahead with a generic helper and refactored the current solution. Let me know if it is close to what you were thinking.

Leo-Besancon: Yes this is fine for me, thank you!

air/src/constraints/stack/ops/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+            is_swapdw.clone() * (s14_next.clone() - s6.clone()),
+            is_swapdw * (s15_next.clone() - s7.clone()),
+        ],
+    );
+
+    // CSWAP / CSWAPW: conditional swaps using s0 as the selector.
+    let cswap_c = s0.clone();
+    let cswap_c_inv = AB::Expr::ONE - cswap_c.clone();
+
+    // Binary constraint for the cswap selector (must be 0 or 1).
```
Leo-Besancon, line 374: The comment here is only for one of the constraints

air/src/constraints/stack/ops/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+                * (s0_next.clone()
+                    - (cswap_c.clone() * s2.clone() + cswap_c_inv.clone() * s1.clone())),
+            is_cswap
+                * (s1_next.clone()
+                    - (cswap_c.clone() * s1.clone() + cswap_c_inv.clone() * s2.clone())),
+        ],
+    );
+
+    // Binary constraint for the cswapw selector (same selector as cswap).
+    assert_zeros(
```
Leo-Besancon, line 404: Same here

air/src/constraints/stack/ops/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+            is_cswapw.clone()
+                * (s5_next.clone()
+                    - (cswap_c.clone() * s2.clone() + cswap_c_inv.clone() * s6.clone())),
+            is_cswapw.clone()
+                * (s6_next.clone()
+                    - (cswap_c.clone() * s3.clone() + cswap_c_inv.clone() * s7.clone())),
+            is_cswapw.clone()
+                * (s7_next.clone()
+                    - (cswap_c.clone() * s4.clone() + cswap_c_inv.clone() * s8.clone())),
+            is_cswapw * (cswap_c.clone() * (cswap_c - AB::Expr::ONE)),
```
Leo-Besancon: Nit: I would put this constraint first for consistency with the cswap op, but it's not critical and the tags would need to be updated in the air-script PR if we do it.

air/src/constraints/stack/ops/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        ],
+    );
+
+    // ASSERT: top element must be 1 (shift handled by stack general).
+    assert_zero(builder, &mut idx, is_assert * (s0 - AB::Expr::ONE));
+
+    // CALLER: load fn_hash into the top 4 stack elements.
+    assert_zeros(
+        builder,
```
adr1anh, line 441: Not sure if it would be more or less readable, but we could do

```suggestion
        &mut builder.when(is_caller.clone()),
```

air/src/constraints/stack/ops/mod.rs
Al-Kindi-0: fixed

air/src/constraints/stack/ops/mod.rs
Al-Kindi-0: moved to the beginning

Soulthym:
Looks good!

PR #2737, opened 2026-02-24
fix: propagate `Assembler::warnings_as_errors` in all methods

PhilippGackstatter:
## Describe your changes

Fixes `Assembler::warnings_as_errors` not being propagated in some methods.

- `read_modules_from_dir` basically ignored the assembler settings.
- `impl Parse for Path` did not propagate the assembler flag.

Should be a non-breaking change, and this targets main. Ideally for inclusion in the next patch release.

- Updated `CHANGELOG.md'


crates/assembly-syntax/src/parse.rs
```diff
@@ -397,6 +397,7 @@ impl Parse for &std::path::Path {
             .into_diagnostic()
             .wrap_err("source manager is unable to load file")?;
         let mut parser = Module::parser(options.kind);
+        parser.set_warnings_as_errors(options.warnings_as_errors);
```
bobbinth, line 400: Ideally, this could be something like:
```rust
let mut parser = Module::parser(options);
```

crates/assembly-syntax/src/parser/mod.rs
```diff
@@ -214,6 +215,7 @@ pub fn read_modules_from_dir(
 
         // Parse module at the given path
         let mut parser = ModuleParser::new(ast::ModuleKind::Library);
+        parser.set_warnings_as_errors(warnings_as_errors);
```
bobbinth, line 218: Same comment as above.

PR #2742, opened 2026-02-26
fix: add bounds checks before slicing in MAST serialization

### LINKED ISSUE #2732: missing bounds check before slicing in MAST serialization
### Packages versions

next branch

### Bug description

found this while reading through the serialization code. in decorator.rs, try_into_decorator slices decorator_data at decorator_data_offset without checking bounds first — if the offset is out of range it panics instead of returning DeserializationError. same issue in string_table.rs, read_string validates the table index but then uses the resulting offset to slice into data without any check.

happy to send a fix if I can be assigned.

_No response_

### Relevant log output

```shell

```

johnh4098:
Fixes #2732

`try_into_decorator` in `decorator.rs` slices `decorator_data` at `decorator_data_offset` without checking bounds first — panics if the offset is out of range instead of returning an error.

Same issue in `string_table.rs` — `read_string` uses the offset from the table to slice into `data` without a bounds check.

Both now return `DeserializationError::InvalidValue` if the offset exceeds the data length.


core/src/mast/serialization/decorator.rs
```diff
@@ -46,11 +46,14 @@ impl DecoratorInfo {
         _string_table: &StringTable,
         decorator_data: &[u8],
     ) -> Result<Decorator, DeserializationError> {
-        // This is safe because for decorators that don't use the offset, `0` is used (and hence
-        // will never access an element outside). Note that in this implementation, we trust the
-        // encoder.
-        let mut data_reader =
-            SliceReader::new(&decorator_data[self.decorator_data_offset as usize..]);
+        let offset = self.decorator_data_offset as usize;
```
huitseeker, line 49: Could we add a small negative test that sets `decorator_data_offset` past `decorator_data.len()` and asserts `DeserializationError::InvalidValue`? This new error path seems worth locking in.

core/src/mast/serialization/string_table.rs
```diff
@@ -53,6 +53,12 @@ impl StringTable {
             DeserializationError::InvalidValue(format!("invalid index in strings table: {str_idx}"))
         })?;
 
+        if str_offset > self.data.len() {
```
huitseeker, line 56: Could we add a negative test that uses a `str_offset` beyond `data.len()` and checks for `DeserializationError::InvalidValue`? This would guard the new bounds check.

core/src/mast/serialization/decorator.rs
johnh4098: Added negative tests for both bounds checks in the latest commit.

PR #2743, opened 2026-02-26
fix(assembly-syntax): implement Hash for QualifiedProcedureName

KitaroMoha563:
Implement custom Hash for QualifiedProcedureName to enforce the PartialEq contract, preventing silent bugs when used as a hash map key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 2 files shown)


bitwalker (commented):
Thanks, but this type didn't implement `Hash` previously, so there are no bugs being fixed here.

We don't accept patches like this anymore due to the volume of them we receive, and we're a small team. You must have an issue in the tracker assigned to you first.

PR #2746, opened 2026-02-27
fix(processor): validate expected_cycles after rounding (fixes #2741)

### LINKED ISSUE #2741: expected_cycles validation happens before rounding, should be after
In processor/src/execution_options.rs, expected_cycles is validated against max_cycles before being rounded up via next_power_of_two().max(MIN_TRACE_LEN). This means the check can pass even when the rounded value exceeds max_cycles.

Example: expected_cycles = 65, max_cycles = 100 — check passes, but after rounding expected_cycles becomes 128 > 100 and no error is returned.

The fix is straightforward: move the validation to after the rounding step.

erenyegit:
Made-with: Cursor

- Updated `CHANGELOG.md'

Fixes #2741


plafer:
Yeah it's not a very useful metric, I think we should move away from it

Al-Kindi-0:
Agreed! Number of cells is probably the right metric going forward and we can express it in more user-friendly ways to help users higher up the stack make more sense of it.

bobbinth:
Could we create an issue for this?

plafer:
#2860

PR #2747, opened 2026-02-27
`build_trace()` no longer assumes valid user input

### LINKED ISSUE #2738: trace generation: don't assume valid `TraceGenerationContext`
Currently trace generation (and more specifically the `CoreTraceGenerationTracer`) assumes valid inputs coming from `TraceGenerationContext`. That is, the implicit assumption is that a trusted `FastProcessor` generated the `TraceGenerationContext`, and so for example we will `.expect()` certain values to be present.

However with #2730 this is no longer a valid assumption, and so `build_trace()` should be refactored to return errors instead of panicking on user input.

plafer:
Closes #2738

The main feature of this PR is to make `build_trace()` no longer assume that `TraceGenerationContext` was constructed properly. Hence, all possible panics incurred from using values in `TraceGenerationContext` were removed (or at least that is the intention). This includes no longer assuming "replay" structs were properly populated, as well as not using values coming from `TraceGenerationContext` to directly index into a slice (i.e. we not use `slice::get()` instead).

The main design decision was that `Tracer` methods still don't return errors, and rather are handled internally in the `CoreTraceGenerationTracer`. This means that from an execution standpoint, tracers are not allowed to interfere with execution - which I personally find to be a cleaner model than if they're otherwise allowed to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.22.0 (TBD)
 
+#### Major breaking changes
+- [BREAKING] `build_trace()` no longer assumes valid user input ([#2747](https://github.com/0xMiden/miden-vm/pull/2747)).
+
 #### Changes
 
 - Consolidated error variants: simplified `AceError` and FRI errors to string-based types, merged `DynamicNodeNotFound`/`NoMastForestWithProcedure` into `ProcedureNotFound`, introduced `HostError` for handler-related variants ([#2675](https://github.com/0xMiden/miden-vm/pull/2675)).


processor/src/trace/parallel/mod.rs
```diff
@@ -69,7 +70,7 @@ pub fn build_trace(
     execution_output: ExecutionOutput,
     trace_generation_context: TraceGenerationContext,
     program_info: ProgramInfo,
-) -> ExecutionTrace {
+) -> Result<ExecutionTrace, ExecutionError> {
```
huitseeker, line 73: Could we add a test that feeds a malformed `TraceGenerationContext` (empty replays or bad node id) and asserts `build_trace` returns `Err` instead of panicking? That would lock in the new error paths.

plafer: Done - added a new failing test commit which precedes the existing ones

PR #2752, opened 2026-02-28
Memory chiplet constraints

Al-Kindi-0:
- Updated `CHANGELOG.md'


Leo-Besancon (approved):

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+        builder,
+        &MEMORY_WORD_IDX_TAGS,
+        &mut idx,
+        word_gate * cols.idx1.clone(),
+    );
+}
+
+/// Enforce memory first row initialization constraints.
+///
+/// When entering the memory chiplet, unwritten values must be 0.
```
adr1anh, line 219: I'd make it a tiny bit clearer that this is enforced in the last row of the previous trace. Maybe using `cols_first` instead of `next`

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+}
+
+/// Compute c_i: 1 if v'[i] needs to be constrained (not written to), 0 otherwise.
+///
+/// c_i = is_read' + !is_read' * z_i
+/// where z_i = !is_word' * !f_i (element access and not this element)
+fn compute_c_i<E: PrimeCharacteristicRing>(f_i: E, is_read_next: E, is_word_next: E) -> E {
+    let z_i = (E::ONE - is_word_next) * (E::ONE - f_i);
+    is_read_next.clone() + (E::ONE - is_read_next) * z_i
+}
```
adr1anh: Not super clear. In both places where it is called, we are passing the same inputs. I think we could merge everything into one function which computes the flags `c_0, ..., c_3`? It could be a function on `MemoryColumns`.

This may be a fault in our docs though, but I think the mental overhead comes from the fact that we're using so many negatives which are hard to keep track of. Even defining a temporary flags like `is_write_next` and `is_element_next`.
From what I can tell, the flag indicates whether the next element should be initialized to zero or is being read?

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    let ctx_delta = deltas.ctx_delta.clone();
+    let addr_delta = deltas.addr_delta.clone();
+
+    let gate = flag_memory_active_not_last;
+    let mut idx = 0;
+    tagged_assert_zero_integrity(
+        builder,
+        &MEMORY_DELTA_INV_TAGS,
+        &mut idx,
+        gate.clone() * n0.clone() * (n0.clone() - one.clone()),
```
adr1anh: Mention `n0` is binary and maybe a var `not_n0` since it is recomputed multiple times?

adr1anh: We can also have `gate_n0` for the next constraints.

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    tagged_assert_zero_integrity(
+        builder,
+        &MEMORY_DELTA_INV_TAGS,
+        &mut idx,
+        gate * (one.clone() - n0.clone()) * (one.clone() - n1.clone()) * addr_delta.clone(),
+    );
+}
+
+/// Enforce the combined delta transition constraint.
+fn enforce_delta_transition_constraint<AB>(
```
adr1anh: Maybe inline since its only used once

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    tagged_assert_zero_integrity(
+        builder,
+        &MEMORY_DELTA_TRANSITION_TAGS,
+        &mut idx,
+        flag_memory_active_not_last * (deltas.computed_delta.clone() - deltas.delta_next.clone()),
+    );
+}
+
+/// Enforce f_scw' = !n0 * !n1 for same context/word transitions.
+fn enforce_scw_flag_constraint<AB>(
```
adr1anh: inline as well

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    n1: E,
+    delta_next: E,
+    computed_delta: E,
+}
+
+impl<E> MemoryDeltas<E>
+where
+    E: Clone + Add<Output = E> + Sub<Output = E> + Mul<Output = E>,
+{
+    fn new(cols: &MemoryColumns<E>, cols_next: &MemoryColumns<E>, one: E, two_pow_16: E) -> Self {
```
adr1anh: We can compute `two_pow_16` locally.

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    deltas: &MemoryDeltas<AB::Expr>,
+    one: AB::Expr,
+) where
+    AB: TaggingAirBuilderExt<F = Felt>,
+{
+    // If ctx/word are unchanged and clk_delta = 0, both rows must be reads.
+    // Constraint: f_scw' * (1 - clk_delta * d_inv') * ((1 - is_read) + (1 - is_read')) = 0
+    let clk_no_change = one.clone() - deltas.clk_delta.clone() * cols_next.d_inv.clone();
+    let read_required =
+        (one.clone() - cols.is_read.clone()) + (one.clone() - cols_next.is_read.clone());
```
adr1anh: Should we add computed fields `is_write` and `is_elt` to the MemoryColumns constructor? In any case, an intermediary variable here would make things a bit easier to read.

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+    );
+}
+
+/// Enforce f_scw' = !n0 * !n1 for same context/word transitions.
+fn enforce_scw_flag_constraint<AB>(
+    builder: &mut AB,
+    flag_memory_active_not_last: AB::Expr,
+    cols_next: &MemoryColumns<AB::Expr>,
+    deltas: &MemoryDeltas<AB::Expr>,
+    one: AB::Expr,
```
adr1anh: No need to pass this as argument.

air/src/constraints/chiplets/memory.rs
```diff
@@ -0,0 +1,635 @@
...
+
+    // When v'[i] is not written to:
+    // - if f_scw' = 1: v'[i] = v[i] (copy from previous)
+    // - if f_scw' = 0: v'[i] = 0 (initialize to zero)
+    // Combined: f_scw' * (v'[i] - v[i]) + !f_scw' * v'[i] = 0
+    let constrain_value = |c: AB::Expr, v: AB::Expr, v_next: AB::Expr| {
+        flag_memory_active_not_last.clone()
+            * c
+            * (cols_next.flag_same_ctx_word.clone() * (v_next.clone() - v)
+                + (one.clone() - cols_next.flag_same_ctx_word.clone()) * v_next)
```
adr1anh: do we have an `or` gadget?

air/src/constraints/chiplets/memory.rs
Al-Kindi-0: yes, changed

air/src/constraints/chiplets/memory.rs
Al-Kindi-0: Good point regarding using the positives instead. Have fixed it in this specific place but might be worth creating an issue to do this consistently throughout the codebase.

air/src/constraints/chiplets/memory.rs
Al-Kindi-0: Addressed

air/src/constraints/chiplets/memory.rs
Al-Kindi-0: Don't think so, but might be worth adding that later

PR #2755, opened 2026-02-28
Chiplets bus constraints

Al-Kindi-0:
- Updated `CHANGELOG.md'


air/src/constraints/chiplets/bus/chiplets.rs
```diff
@@ -0,0 +1,1848 @@
...
+        OPCODE_CALL, OPCODE_DYN, OPCODE_DYNCALL, OPCODE_JOIN, OPCODE_LOOP, OPCODE_SPLIT,
+        OPCODE_SYSCALL,
+    },
+};
+use miden_crypto::stark::{air::MidenAirBuilder, matrix::Matrix};
+
+use crate::{
+    Felt, MainTraceRow,
+    constraints::{
+        bus::{MessageEncoder, alphas_from_challenges, indices::B_CHIPLETS},
```
adr1anh, line 48: We should update https://github.com/0xMiden/miden-vm/blob/al-chunk-chiplets-bus/air/src/constraints/bus.rs#L68 to work with only two input challenges from which we generate powers.

air/src/constraints/chiplets/bus/chiplets.rs
```diff
@@ -0,0 +1,1848 @@
...
+    let label: Felt = if is_xor { BITWISE_XOR_LABEL } else { BITWISE_AND_LABEL };
+    let label: AB::Expr = AB::Expr::from(label);
+
+    // Stack values
+    let a: AB::Expr = local.stack[0].clone().into();
+    let b: AB::Expr = local.stack[1].clone().into();
+    let z: AB::Expr = next.stack[0].clone().into();
+
+    let encoder = MessageEncoder::<AB, 4>::from_challenges(alphas);
+    encoder.encode([label, a, b, z])
```
adr1anh, line 373: The `MessageEncoder` doesn't really bring much IMO, we should just be able to have a function which takes a slice of (expanded) challenges and computes the linear combination. Even better, we could just carry around some challenger container with just `alpha, beta` and use Horner which is slightly more efficient in-circuit.

Al-Kindi-0: Agreed, since I think it is better to postpone the powers of `beta` refactoring til later, I have just removed the encoders and simplified things using simple helpers here https://github.com/0xMiden/miden-vm/pull/2764

air/src/constraints/chiplets/bus/chiplets.rs
Al-Kindi-0: I would do this at the end since this will require changes beyond just the constraints.

Al-Kindi-0:
> Looks good overall, but as a followup we should probably clean up the way we handle the aux randomness across the codebase. We're committing to using `[beta, 1, alpha, alpha^2, ....]`. It would be a lot clearer if we made this explicit, for example using a struct with `beta: ExprEF, alphas: &'_ [ExprEF]`.

Fully agree! This should be done once things stabilize since it is a change that will touch many places. We should also update the docs and provide an analysis of the soundness of the multi-set/LogUp argument in our setting.

PR #2757, opened 2026-02-28
Ace codegen

Al-Kindi-0:
## Describe your changes

Adds an `ace-codegen` crate to generate ACE circuits from Miden AIR. It introduces a `RecordingAirBuilder` to capture the AIR constraints, lowers them into a DAG for the verifier's constraints evaluation check, defines READ/EVAL input layouts, and gives tools to encode such circuits.

There is also the `processor_air_chiplet_rows` test which prints the number of rows in the ACE chiplet for evaluating the circuit for `ProcessorAir`. For the currently implemented set of constraints, which contains about 95% of the constraints, it shows that it takes `5712` ACE rows to evaluate the circuit.

There is also an integration test in `crates/lib/core/tests/crypto/circuit_evaluation.rs` that builds the `ProcessorAir` ACE circuit, encodes it, streams it via `adv_pipe`, executes `eval_circuit`, and runs `prove_and_verify`, ensuring the end‑to‑end MASM path works.

- Updated `CHANGELOG.md'

### DIFF
--- a/crates/ace-codegen/src/dag/builder.rs
+++ b/crates/ace-codegen/src/dag/builder.rs
@@ -17,7 +17,7 @@ pub struct DagBuilder<EF> {
 
 impl<EF> DagBuilder<EF>
 where
-    EF: Field + Eq + std::hash::Hash,
+    EF: Field,
 {
     /// Create an empty, hash-consed DAG builder.
     pub fn new() -> Self {
@@ -119,7 +119,7 @@ where
 
 impl<EF> Default for DagBuilder<EF>
 where
-    EF: Field + Eq + std::hash::Hash,
+    EF: Field,
 {
     fn default() -> Self {
         Self::new()
--- a/crates/ace-codegen/src/dag/ir.rs
+++ b/crates/ace-codegen/src/dag/ir.rs
@@ -1,3 +1,6 @@
+use p3_dft::{NaiveDft, TwoAdicSubgroupDft};
+use p3_field::TwoAdicField;
+
 use crate::layout::InputKey;
 
 /// Identifier for a node in the DAG.
@@ -36,10 +39,35 @@ pub enum NodeKind<EF> {
 #[derive(Debug, Clone)]
 pub struct PeriodicColumnData<EF> {
     /// Per-column coefficient vectors (highest-degree first).
-    pub(crate) coeffs: Vec<Vec<EF>>,
+    coeffs: Vec<Vec<EF>>,
 }
 
 impl<EF> PeriodicColumnData<EF> {
+    /// Convert periodic columns (evaluations) into coefficient form for DAG building.
+    ///
+    /// Applies an inverse DFT so the DAG can evaluate them at `z_k` inside the circuit.
+    pub fn from_periodic_table<F>(periodic_table: Vec<Vec<F>>) -> Self
+    where
+        F: TwoAdicField,
+        EF: From<F>,
+    {
+        if periodic_table.is_empty() {
+            return Self { coeffs: Vec::new() };
+        }
+
+        let dft = NaiveDft;
+        let mut coeffs = Vec::with_capacity(periodic_table.len());
+        for col in periodic_table {
+            assert!(!col.is_empty(), "periodic column must not be empty");
+            assert!(col.len().is_power_of_two(), "periodic column length must be a power of two");
+            let values = dft.idft(col);
+            let coeff_row = values.into_iter().map(EF::from).collect();
+            coeffs.push(coeff_row);
+        }
+
+        Self { coeffs }
+    }
+
     /// Number of periodic columns.
     pub fn num_columns(&self) -> usize {
         self.coeffs.len()
@@ -48,6 +76,11 @@ impl<EF> PeriodicColumnData<EF> {
     /// Maximum periodic column length (used to align powers).
     pub fn max_period(&self) -> usize {
         self.coeffs.iter().map(|c| c.len()).max().unwrap_or(0)
+    }
+
+    /// Iterate over the per-column coefficient vectors.
+    pub fn columns(&self) -> &[Vec<EF>] {
+        &self.coeffs
     }
 }
 
--- a/crates/ace-codegen/src/dag/lower.rs
+++ b/crates/ace-codegen/src/dag/lower.rs
@@ -1,14 +1,12 @@
 use std::collections::HashMap;
 
-use p3_dft::{Radix2DitParallel, TwoAdicSubgroupDft};
-use p3_field::{ExtensionField, Field, TwoAdicField};
+use p3_field::{ExtensionField, Field};
 
 use super::{
     builder::DagBuilder,
     ir::{AceDag, NodeId, PeriodicColumnData},
 };
 use crate::{
-    EXT_DEGREE,
     layout::{InputKey, InputLayout},
     quotient::build_quotient_recomposition_dag,
     randomness,
@@ -24,7 +22,7 @@ pub fn lower_expr<F, EF>(
 ) -> NodeId
 where
     F: Field,
-    EF: ExtensionField<F> + Eq + std::hash::Hash,
+    EF: ExtensionField<F>,
 {
     match expr {
         SymExpr::Variable(v) => match v.entry {
@@ -32,7 +30,7 @@ where
             Entry::Aux { offset } => {
                 let index = v.index;
                 let mut acc = builder.constant(EF::ZERO);
-                for coord in 0..EXT_DEGREE {
+                for coord in 0..EF::DIMENSION {
                     let basis =
                         EF::ith_basis_element(coord).expect("basis index within extension degree");
                     let coord_node = builder.input(InputKey::AuxCoord { offset, index, coord });
@@ -104,7 +102,7 @@ pub fn build_verifier_dag<F, EF>(
 ) -> AceDag<EF>
 where
     F: Field,
-    EF: ExtensionField<F> + Eq + std::hash::Hash,
+    EF: ExtensionField<F>,
 {
     let mut builder = DagBuilder::<EF>::new();
     let periodic_nodes = match periodic {
@@ -134,43 +132,15 @@ where
 
     AceDag { nodes: builder.into_nodes(), root }
 }
-
-impl<EF> PeriodicColumnData<EF> {
-    /// Convert periodic columns (evaluations) into coefficient form for DAG building.
-    ///
-    /// Applies an inverse DFT so the DAG can evaluate them at `z_k` inside the circuit.
-    pub fn from_periodic_table<F>(periodic_table: Vec<Vec<F>>) -> Self
-    where
-        F: TwoAdicField + Ord,
-        EF: From<F>,
-    {
-        if periodic_table.is_empty() {
-            return Self { coeffs: Vec::new() };
-        }
-
-        let dft = Radix2DitParallel::<F>::default();
-        let mut coeffs = Vec::with_capacity(periodic_table.len());
-        for col in periodic_table {
-            assert!(!col.is_empty(), "periodic column must not be empty");
-            assert!(col.len().is_power_of_two(), "periodic column length must be a power of two");
-            let values = dft.idft(col);
-            let coeff_row = values.into_iter().map(EF::from).collect();
-            coeffs.push(coeff_row);
-        }
-
-        Self { coeffs }
-    }
-}
-
 fn build_periodic_nodes<EF>(
     builder: &mut DagBuilder<EF>,
     layout: &InputLayout,
     periodic: &PeriodicColumnData<EF>,
 ) -> Vec<NodeId>
 where
-    EF: Field + Eq + std::hash::Hash,
+    EF: Field,
 {
-    if periodic.coeffs.is_empty() {
+    if periodic.num_columns() == 0 {
         return Vec::new();
     }
 
@@ -181,8 +151,8 @@ where
 
     let max_len = periodic.max_period();
     let mut cache = HashMap::<u32, NodeId>::new();
-    let mut nodes = Vec::with_capacity(periodic.coeffs.len());
-    for coeffs in &periodic.coeffs {
+    let mut nodes = Vec::with_capacity(periodic.num_columns());
+    for coeffs in periodic.columns() {
         let col_len = coeffs.len();
         let ratio = max_len / col_len;
         let log_pow_col = ratio.ilog2();
@@ -203,7 +173,7 @@ where
 
 fn horner_eval<EF>(builder: &mut DagBuilder<EF>, point: NodeId, coeffs: &[NodeId]) -> NodeId
 where
-    EF: Field + Eq + std::hash::Hash,
+    EF: Field,
 {
     let mut acc = builder.constant(EF::ZERO);
     for coeff in coeffs.iter().rev() {
--- a/crates/ace-codegen/src/encode.rs
+++ b/crates/ace-codegen/src/encode.rs
@@ -10,7 +10,7 @@
 //! (base-field), then pads to an `adv_pipe` block boundary.
 
 use miden_core::{Felt, Word, crypto::hash::Poseidon2};
-use p3_field::BasedVectorSpace;
+use p3_field::ExtensionField;
 
 use crate::{
     AceError,
@@ -92,7 +92,7 @@ impl EncodedCircuit {
 
 impl<EF> AceCircuit<EF>
 where
-    EF: BasedVectorSpace<Felt> + Copy + Eq + std::hash::Hash,
+    EF: ExtensionField<Felt>,
 {
     /// Encode the circuit into the ACE chiplet format.
     pub fn to_ace(&self) -> Result<EncodedCircuit, AceError> {
--- a/crates/ace-codegen/src/pipeline.rs
+++ b/crates/ace-codegen/src/pipeline.rs
@@ -59,8 +59,8 @@ pub fn build_ace_circuit_for_air<A, F, EF>(
 ) -> Result<AceCircuit<EF>, AceError>
 where
     A: MidenAir<F, EF>,
-    F: TwoAdicField + Ord,
-    EF: ExtensionField<F> + Eq + std::hash::Hash,
+    F: TwoAdicField,
+    EF: ExtensionField<F>,
 {
     let artifacts = build_ace_dag_for_air::<A, F, EF>(air, config)?;
     emit_circuit(&artifacts.dag, artifacts.layout)
@@ -95,8 +95,8 @@ pub(crate) fn build_ace_dag_for_air<A, F, EF>(
 ) -> Result<AceArtifacts<EF>, AceError>
 where
     A: MidenAir<F, EF>,
-    F: TwoAdicField + Ord,
-    EF: ExtensionField<F> + Eq + std::hash::Hash,
+    F: TwoAdicField,
+    EF: ExtensionField<F>,
 {
     let periodic_table = air.periodic_table();
     let counts = input_counts_for_air::<A, F, EF>(air, config, periodic_table.len());
--- a/crates/ace-codegen/src/quotient.rs
+++ b/crates/ace-codegen/src/quotient.rs
@@ -9,7 +9,6 @@ use p3_field::{ExtensionField, Field};
 #[cfg(test)]
 use crate::AceError;
 use crate::{
-    EXT_DEGREE,
     dag::{DagBuilder, NodeId},
     layout::{InputKey, InputLayout},
 };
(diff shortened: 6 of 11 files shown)


adr1anh:
Only did a very high-level review, and it looks really good already. I'm trying to merge the necessary changes to the SymbolicAir to upstream plonky3 so we can reduce the amount of duplicated code we need to carry around. I'll check though if we can do without and if so I'll fast track a review for here.

adr1anh:
I took the liberty of adding a minimal SymbolicExpression/Variable implementation so we can avoid relying on any other air crates from plonky3/uni-stark/p3-miden. It should also make it easier for us to add some optimizations to the graph

crates/ace-codegen/src/dag/ir.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Per-column coefficient vectors (highest-degree first).
+    pub coeffs: Vec<Vec<EF>>,
+}
+
+impl<EF> PeriodicColumnData<EF> {
+    /// Number of periodic columns.
+    pub fn len(&self) -> usize {
+        self.coeffs.len()
+    }
+}
```
adr1anh, line 85: I'd make max period a computed function and rename Len to `num_columns`

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,269 @@
...
+use crate::{
+    AceError, layout::InputKey, quotient::build_quotient_recomposition_dag,
+    randomness::RandomnessPlan,
+};
+
+/// Lower a symbolic expression into DAG nodes using the provided layout.
+pub fn lower_expr<F, EF>(
+    expr: &SymExpr<EF>,
+    builder: &mut DagBuilder<EF>,
+    layout: &crate::layout::InputLayout,
```
adr1anh: Nit : add an import for this

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,269 @@
...
+    periodic_nodes: &[NodeId],
+) -> Result<NodeId, AceError>
+where
+    F: PrimeCharacteristicRing,
+    EF: PrimeCharacteristicRing + BasedVectorSpace<F> + Copy + Eq + Hash,
```
adr1anh: Why not use Field and ExtensionField?

crates/ace-codegen/src/dag/builder.rs
```diff
@@ -0,0 +1,127 @@
...
+/// compact and deterministic.
+#[derive(Debug)]
+pub struct DagBuilder<EF> {
+    nodes: Vec<NodeKind<EF>>,
+    cache: HashMap<NodeKind<EF>, NodeId>,
+}
+
+impl<EF> DagBuilder<EF>
+where
+    EF: PrimeCharacteristicRing + Copy + Eq + std::hash::Hash,
```
adr1anh: Field should be enough

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// Build the verifier-equivalent root expression DAG.
+///
+/// This constructs the folded constraint accumulator, divides by the vanishing
+/// polynomial, recomposes the quotient, and subtracts both sides to yield the
+/// root expression evaluated by the ACE circuit.
+pub fn build_verifier_dag<F, EF>(
+    constraints: &[SymExpr<EF>],
+    layout: &crate::layout::InputLayout,
```
adr1anh: Same as above

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+    periodic: Option<&PeriodicColumnData<EF>>,
+) -> Result<AceDag<EF>, AceError>
+where
+    F: PrimeCharacteristicRing,
+    EF: PrimeCharacteristicRing + BasedVectorSpace<F> + Copy + Eq + Hash,
```
adr1anh: Same as above

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+{
+    if layout.counts.ext_degree != EF::DIMENSION {
+        return Err(AceError::InvalidExtensionDegree {
```
adr1anh: I don't think the parameters are super helpful, this should never happen

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+    let root = builder.sub(folded, quotient);
+
+    Ok(AceDag { nodes: builder.into_nodes(), root })
+}
+
+/// Convert periodic columns into evaluation coefficients for DAG building.
+///
+/// The periodic columns are provided as evaluations; this function applies an
+/// inverse DFT so the DAG can evaluate them at `z_k` inside the circuit.
+pub fn build_periodic_data<F, EF>(
```
adr1anh: This would make more sense as a constructor on PeriodicData

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+        let coeff_row = values.into_iter().map(EF::from).collect();
+        coeffs.push(coeff_row);
+    }
+
+    Ok(PeriodicColumnData { max_len, coeffs })
+}
+
+fn build_periodic_nodes<EF>(
+    builder: &mut DagBuilder<EF>,
+    layout: &crate::layout::InputLayout,
```
adr1anh: Same as earlier

crates/ace-codegen/src/randomness.rs
```diff
@@ -0,0 +1,156 @@
...
+        if num == 0 {
+            return (RandomnessPlan::Direct { num }, None, None);
+        }
+
+        if num_inputs == num {
+            return (RandomnessPlan::Direct { num }, None, None);
+        }
+
+        if num_inputs == 2 {
+            // Stored as [beta, alpha] in memory words.
```
huitseeker: `RandomnessPlan::from_layout` chooses `Direct` whenever `num_randomness_inputs == num_randomness`. Since `input_counts_for_air` hard‑codes `num_randomness_inputs = 2`, any AIR with `num_randomness == 2` will not use the alpha/beta expansion and will map challenges to `InputKey::Randomness`(0/1) instead of `[alpha, 1, beta, ...]`. This conflicts with the docs in layout/mod.rs and this comment.

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+        return Err(AceError::InvalidPeriodicColumn { index: 0, count: periodic.coeffs.len() });
+    }
+
+    let mut cache = HashMap::<usize, NodeId>::new();
+    let mut nodes = Vec::with_capacity(periodic.coeffs.len());
+    for (idx, coeffs) in periodic.coeffs.iter().enumerate() {
+        if coeffs.is_empty() {
+            nodes.push(builder.constant(EF::ZERO));
+            continue;
+        }
```
adr1anh: Isn't this an error?

crates/ace-codegen/src/randomness.rs
```diff
@@ -0,0 +1,156 @@
...
+                alpha_idx,
+                beta_idx,
+            );
+        }
+
+        (RandomnessPlan::Direct { num }, None, None)
+    }
+
+    /// Build the plan from a concrete input layout.
+    pub(crate) fn from_layout(layout: &InputLayout) -> Result<Self, AceError> {
```
huitseeker: This returns `Direct` when `num_randomness_inputs == num_randomness` before checking the alpha/beta path. A regression test for `num_randomness == 2` seems important.

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+        return Err(AceError::InvalidPeriodicColumn { index: 0, count: max_len });
+    }
+
+    let dft = Radix2DitParallel::<F>::default();
+    let mut coeffs = Vec::with_capacity(periodic_table.len());
+    for (idx, col) in periodic_table.into_iter().enumerate() {
+        if col.is_empty() {
+            coeffs.push(Vec::new());
+            continue;
+        }
```
adr1anh: This should be an error

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+        let values = dft.idft(col);
+        let coeff_row = values.into_iter().map(EF::from).collect();
+        coeffs.push(coeff_row);
+    }
+
+    Ok(PeriodicColumnData { max_len, coeffs })
+}
+
+fn build_periodic_nodes<EF>(
```
adr1anh, line 135: The checks can be ignored imo, since we can assume PeriodicColumnData was correctly constructed and validated

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+        let col_len = coeffs.len();
+        let max_len = periodic.max_len;
+        if !max_len.is_multiple_of(col_len) || !max_len.is_power_of_two() {
+            return Err(AceError::InvalidPeriodicColumn { index: idx, count: col_len });
+        }
+        let ratio = max_len / col_len;
+        if !ratio.is_power_of_two() {
+            return Err(AceError::InvalidPeriodicColumn { index: idx, count: col_len });
+        }
+        let log_pow_col = ratio.ilog2() as usize;
```
adr1anh: Log strict usize

crates/ace-codegen/src/dag/lower.rs
```diff
@@ -0,0 +1,271 @@
...
+{
+    let mut acc = builder.constant(EF::ZERO);
+    for coeff in coeffs.iter().rev() {
+        let mul = builder.mul(point, acc);
+        acc = builder.add(*coeff, mul);
+    }
+    acc
+}
+
+fn input_key_for_symbolic<EF>(var: &SymVar<EF>) -> Result<InputKey, AceError> {
```
adr1anh: Can't check in my editor right now, but should this be inlined into the lowering function?

crates/lib/core/tests/crypto/circuit_evaluation.rs
```diff
@@ -81,3 +84,124 @@ fn circuit_evaluation_prove_verify() {
...
+    memory_felts.resize(padded_len, ZERO);
+    let num_adv_pipe = padded_len / 8;
+
+    let mut advice_builder = AdviceStackBuilder::new();
+    advice_builder.push_for_adv_pipe(&memory_felts);
+    let adv_stack = advice_builder.build_vec_u64();
+
+    // Place the circuit in memory at a fixed address.
+    let pointer = 1 << 16;
+    let num_read = encoded.num_vars();
```
huitseeker: Bit confusing: If `eval_circuit` expects `read` rows, should this be `encoded.num_read_rows()`? If vars are correct, can we rename/comment to make the ABI expectation explicit?

crates/ace-codegen/src/encode.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl<EF> AceCircuit<EF>
+where
+    EF: PrimeCharacteristicRing + BasedVectorSpace<Felt> + Copy + Eq + std::hash::Hash,
+{
+    /// Encode the circuit into the ACE chiplet format.
+    pub fn to_ace(&self) -> Result<EncodedCircuit, AceError> {
+        const MAX_NODE_ID: u64 = (1 << 30) - 1;
+
+        if EF::DIMENSION != EXPECTED_EXTENSION_DEGREE {
```
adr1anh: Could be a const assert

huitseeker (commented):
It would be great to have a test for the `num_randomness == 2` edge case where `num_randomness_inputs == 2`, as well as direct‑randomness layouts (`num_randomness_inputs == num_randomness` with inputs supplied directly)

crates/ace-codegen/src/layout/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+//! extension-field elements in point-major order (all openings at `zeta`
+//! followed by all openings at `g * zeta`). Auxiliary trace and quotient
+//! chunk openings are provided as base-field coordinates and merged into
+//! extension elements inside the ACE circuit.
+//!
+//! All offsets and widths in this module are measured in extension-field (EF)
+//! elements. One EF element occupies two base-field elements.
+//!
+//! Besides the AIR constraints themselves, the ACE circuit computes:
+//! - Randomness expansion from `(alpha, beta)` into the full challenge vector `[alpha, 1, beta,
```
adr1anh: Should we make this consistent with the challenges?

crates/ace-codegen/src/layout/plan.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Use `num_randomness` for direct randomness, or `2` for alpha/beta seeds.
+    pub num_randomness_inputs: usize,
+    /// Number of periodic columns.
+    pub num_periodic: usize,
+    /// Number of auxiliary (stark var) inputs reserved.
+    pub num_aux_inputs: usize,
+    /// Number of quotient chunks.
+    pub num_quotient_chunks: usize,
+    /// Extension field degree.
+    pub ext_degree: usize,
```
adr1anh: I think we should just keep ext degree as a constant everywhere. It requires extra checks, and would be a constant if we used Field/ExtensionField (I think)

crates/ace-codegen/src/layout/plan.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub aux_bus_boundary: InputRegion,
+    /// Stark variables (selectors, powers, weights).
+    pub stark_vars: InputRegion,
+}
+
+/// Indexes of canonical verifier scalars inside the stark-vars block.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct StarkVarIndices {
+    /// Width of the base-field portion of the stark-vars block.
+    pub base_width: usize,
```
adr1anh: Why is this necessary?

crates/ace-codegen/src/layout/policy.rs
```diff
@@ -0,0 +1,162 @@
...
+    main: Alignment,
+    aux: Alignment,
+    quotient: Alignment,
+    aux_bus_boundary: Alignment,
+    stark_vars: Alignment,
+    end_align: Option<Alignment>,
+}
+
+impl LayoutPolicy {
+    fn native() -> Self {
```
adr1anh, line 25: Do we need this?

crates/ace-codegen/src/pipeline.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/// Build a verifier-equivalent DAG and layout for the provided AIR.
+///
+/// This is useful when you need the DAG for off-VM checks and want to
+/// assemble inputs separately.
+pub(crate) fn build_ace_dag_for_air<A, F, EF>(
+    air: &A,
+    config: AceConfig,
+) -> Result<AceArtifacts<EF>, AceError>
```
adr1anh, line 95: I feel like we could probably simplify the whole thing to use assertions. The Air is supposed to be correct, and if not then everything else would break.

crates/ace-codegen/src/quotient.rs
```diff
@@ -0,0 +1,238 @@
...
+    AceError,
+    dag::{DagBuilder, NodeId},
+    layout::{InputKey, InputLayout},
+};
+
+/// Evaluate the quotient recomposition at `zeta` using provided inputs.
+#[cfg(test)]
+pub fn eval_quotient<F, EF>(layout: &InputLayout, inputs: &[EF]) -> Result<EF, AceError>
+where
+    F: PrimeCharacteristicRing,
```
adr1anh: Same as above

crates/ace-codegen/src/quotient.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// Build DAG nodes that recombine quotient chunks.
+pub(crate) fn build_quotient_recomposition_dag<F, EF>(
+    builder: &mut DagBuilder<EF>,
+    layout: &InputLayout,
+) -> Result<NodeId, AceError>
+where
+    F: PrimeCharacteristicRing,
+    EF: PrimeCharacteristicRing + BasedVectorSpace<F> + Copy + Eq + Hash,
```
adr1anh: Same as above

crates/ace-codegen/src/randomness.rs
```diff
@@ -0,0 +1,156 @@
...
+    AceError,
+    dag::{DagBuilder, NodeId},
+    layout::{InputCounts, InputKey, InputLayout, InputRegion},
+};
+
+/// Randomness provisioning strategy for the ACE layout.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum RandomnessPlan {
+    /// All randomness values are supplied directly.
+    Direct { num: usize },
```
adr1anh: Do we still need this?

crates/ace-codegen/src/randomness.rs
Al-Kindi-0: Removed the `Direct` variant and simplified things now

crates/lib/core/tests/crypto/circuit_evaluation.rs
Al-Kindi-0: This is indeed confusing and I have fixed it, but we should think about unifying/improving the namings in other places

crates/ace-codegen/src/randomness.rs
Al-Kindi-0: Removed `Direct` and simplified

crates/ace-codegen/src/dag/ir.rs
Al-Kindi-0: done

crates/ace-codegen/src/dag/lower.rs
Al-Kindi-0: simplifed

crates/ace-codegen/src/dag/lower.rs
Al-Kindi-0: simplified

crates/ace-codegen/src/dag/lower.rs
Al-Kindi-0: Changed to using `u32`

crates/ace-codegen/src/dag/lower.rs
Al-Kindi-0: Indeed, fixed

crates/ace-codegen/src/dag/lower.rs
Al-Kindi-0: done

crates/ace-codegen/src/layout/policy.rs
Al-Kindi-0: This simplifies testing when we don't care about masm layout and is very little code overhead, so I think it is justified

crates/ace-codegen/src/layout/mod.rs
Al-Kindi-0: If I understand your comment correctly, I updated the comment and looked at re-using some of the infra for challenges we introduced. It seemed more work than it is worth it for now

crates/ace-codegen/src/quotient.rs
Al-Kindi-0: simplified

crates/ace-codegen/src/pipeline.rs
Al-Kindi-0: Simplified

crates/ace-codegen/src/layout/plan.rs
Al-Kindi-0: Removed

crates/ace-codegen/src/randomness.rs
Al-Kindi-0: Removed `Direct` and then the whole enum now

crates/ace-codegen/src/dag/lower.rs
adr1anh: Nit, I meant using plonky3's log2_strict_usize, but tbh this is okay

crates/ace-codegen/src/layout/mod.rs
adr1anh: I wasn't sure, but don't we use powers of alpha instead of beta?

crates/ace-codegen/src/dag/builder.rs
```diff
@@ -17,7 +17,7 @@ pub struct DagBuilder<EF> {
 
 impl<EF> DagBuilder<EF>
 where
-    EF: PrimeCharacteristicRing + Copy + Eq + std::hash::Hash,
+    EF: Field + Eq + std::hash::Hash,
```
adr1anh: Just `Field` is enough

crates/ace-codegen/src/layout/mod.rs
Al-Kindi-0: we use powers of beta throughout.
Sometimes, we have `[beta, alpha]` but this is just a convenient way for MASM

PR #2761, opened 2026-03-02
fix: pass --max-cycles and --exp-cycles to processor in run command

LeeliRen6:
The run command accepts --max-cycles and --exp-cycles flags but never actually passes them to the processor — it just uses ExecutionOptions::default(). This means users can't constrain execution cycles when using `miden run`, unlike `miden prove` which wires these up correctly.

Fixed by constructing ExecutionOptions from the CLI params and passing them via .with_options(), same as prove.rs already does.

### DIFF
--- a/core/src/mast/debuginfo/debug_var_storage.rs
+++ b/core/src/mast/debuginfo/debug_var_storage.rs
@@ -331,8 +331,7 @@ impl OpToDebugVarIds {
         }
         // Create empty nodes for gaps
         for idx in expected.0..node.0 {
-            self.add_debug_var_info_for_node(MastNodeId::new_unchecked(idx), vec![])
-                .unwrap();
+            self.add_debug_var_info_for_node(MastNodeId::new_unchecked(idx), vec![])?;
         }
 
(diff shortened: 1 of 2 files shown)


plafer (approved):
🙏

plafer:
This is ready to merge, but let's wait until after #2823. Also @bobbinth you will need to do it because the commit is not signed (which is probably fine).

LeeliRen6:
> This is ready to merge, but let's wait until after #2823. Also @bobbinth you will need to do it because the commit is not signed (which is probably fine).

Should I re-sign the commit? or its just fine this time?

plafer:
@LeeliRen6 yes it would simplify things if you could sign the commit

LeeliRen6:
> @LeeliRen6 yes it would simplify things if you could sign the commit

signed

plafer:
> signed

Hmm, your commit still shows as not verified.

Alternatively @bobbinth can you merge this to override the signature requirement?

bobbinth (approved):
Looks good! Thank you!

PR #2764, opened 2026-03-02
Simplify bus messages encoding

Al-Kindi-0:
## Describe your changes
Address review in #2755

- Updated `CHANGELOG.md'

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2002,8 +2002,7 @@ dependencies = [
 [[package]]
 name = "p3-miden-air"
 version = "0.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45a88e6ee9c92ff6c0b64f1ec0d61eda72fb432bda45337d876c46bd43748508"
+source = "git+https://github.com/0xMiden/p3-miden?branch=main#32fba20abd9bc099e0342abeb06e5df46002fbe7"
 dependencies = [
  "p3-air",
  "p3-field",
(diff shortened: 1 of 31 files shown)


adr1anh (commented):
This still feels awkward. Why don't we just pass alpha and beta and generate the powers in a clear way?

I think the requirement around the challenges are
- we always request two challenges from `builder.permutation_randomness()`
- we want to compute the powers only once per Air evaluation (on the prover side, ideally we do it once per trace but this seems complicated).
- we want to allow evaluation of partial messages when we know some coefficients are zero
- we want this to use the same logic for the trace generation and constraint evaluation. 
- we don't want to have to access the heap, hence no `Vec`

This would be a struct that could be `Challenges<'a, ExprEF>{ base: ExprEF, coeffs: &'a [ExprEF] }`. Alternatively, we could use an array for `coeffs` but this means we have to carry the const generic everywhere. The reference would probably be avoidable since we can use the implicit `'_`. The struct could have methods which allow for evaluations of full/partial messages, but more importantly they would be a lot easier to read IMO.

I don't think this PR really reduces the complexity and it would make sense to do a larger change which also changes the trace generation.

Al-Kindi-0:
I think I have mentioned in the original PR that it is not immediate to just use the powers. This requires changes at many levels beyond just the constraints.
The current PR aims at restoring some semblance of being generic over the randomness for now so that we are more faithful to how things are now. During the refactoring to use the alpha and the powers of beta, we can do either the container to encode/fingerprint messages using Horner, or we could use stored powers with const generics. Does this make sense ?

If you don't find value in this PR then I can just close it and do the full refactoring later, for me either way this is a very minimal issue in my opinion.

Al-Kindi-0:
Superseded with more thorough upcoming refactoring

PR #2765, opened 2026-03-02
Refactor MAST forest serialization around fixed-layout hashless sections

### LINKED ISSUE #2504: Implement `Serializable::get_size_hint` for `MastForest`
For the purpose of enforcing maximum size limits of account code and note script's in miden protocol, we need to be able to compute the size of a serialized `MastForest`. Currently, we do this by serializing the mast forest and checking the length, but it would be better if `MastForest` would implement `get_size_hint` (from the `Serializable` trait), to be able to calculate the size without having to allocate a vector and serialize it.

This is not a high priority since there is a workaround, but I wanted to open an issue to document the requirement.

### LINKED ISSUE #2628: Hashless Untrusted MastForest serialization format
## Goal.

Minimize the size of the untrusted serialization format for MastForest by omitting node digests that will be recomputed during validation anyway. This reduces bandwidth & storage costs for scenarios where serialized MAST data comes from untrusted sources.

## Benchmark Results

Measured on the core library's MastForest:

| Format | Size | Savings |
|--------|------|---------|
| Full | 406,309 bytes | - |
| Stripped | 275,653 bytes | 32.1% |
| **Hashless** | **252,901 bytes** | **37.8% vs full, 8.3% vs stripped** |

The hashless format saves 22,752 bytes compared to stripped serialization. This is the improvement we expect from completing this task.

### Current Prototype

Branch: [`huitseeker/hashless-serialization`](https://github.com/0xMiden/miden-vm/tree/huitseeker/hashless-serialization) (on top of #2623)
Commit: `013e9644be9fe750e8ecb1c3b7eeda9ed2604ce0` (or latest on that branch)

What's implemented:
- `FLAG_HASHLESS (0x02)` flag for the format
- `MastNodeInfoHashless` struct that omits digests for all nodes except External
- `MastForest::write_hashless()` method for serialization
- Benchmark comparing hashless vs full vs stripped formats

External nodes keep their digests because they reference procedures not present in the forest - those digests cannot be recomputed from available data.

### Core Code TODOs (3 locations)

1. **`core/src/mast/serialization/mod.rs`** - Deserialization support for hashless format
2. **`core/src/mast/serialization/info.rs`** - `Deserializable` impl for `MastNodeInfoHashless`
3. **`core/src/mast/serialization/tests.rs`** - Round-trip proptests and fuzzing targets

### Further Design Issues

**1. Hashless and stripped flags are coupled**

The code writes `FLAG_HASHLESS | FLAG_STRIPPED` but there's no spec saying these must always be used together. The comment "hashless + stripped - no debug info needed for untrusted input" is an assumption, not a validated rule.

Options:
- (a) Document that hashless always implies stripped and validate during deserialization
- (b) Allow hashless with debug info and separate these concerns

**2. Code duplication in serialization**

`write_into_hashless()` duplicates most of `write_into_with_options()`. Format changes would need updates in multiple places. Consider refactoring to share the common structure.

**3. No verification that External digests are preserved**

The logic in `MastNodeInfoHashless::new()` preserves External node digests, but there are no tests verifying this works.

### LINKED ISSUE #2726: feat: MAST forest deserialization
## MAST Forest Serialization Format

This note specifies the serialization format for MAST forests. It unifies previous explorations of "full", "stripped", and "hashless" formats into a single format with feature flags.

The key insight is that hashless serialization (for untrusted input) and stripped serialization (for production without debug info) are almost identical. Both omit debug info. The only difference is whether node digests are *considered* present/trusted or not. Using feature flags instead of separate formats simplifies the code and makes the relationships explicit.

Fixed-size encoding trades ~6% size increase (see #2623) for major capabilities: direct mmap access without deserialization, constant-time size computation (see #2504), and random node access by index (not currently used).

## Wire Format

The header is 16 bytes:
- MAGIC: [4]u8 = "MAST"
- FLAGS: u8 (STRIPPED=0x01, HASHLESS=0x02)
- VERSION: [3]u8 = [0, 0, 4]
- NODE_SIZE: u16 = 41
- RESERVED: [6]u8

HASHLESS implies STRIPPED.

The core section follows:
- node_count: u32
- decorator_count: u32
- roots_count: u32
- roots: [u32; roots_count]
- basic_block_data_len: u64
- basic_block_data: [u8]
- node_infos: [MastNodeInfo; node_count] (41 bytes each)

Each MastNodeInfo is exactly 41 bytes:
- tag: u8 (node type 0-8)
- data: [u8; 8] (type-specific indices/offsets)
- digest: [u8; 32] (RPO hash, considered 'present'/_trusted_ or absent/_recomputed_ based on HASHLESS flag, but always allocated)

External nodes are a special case. Since they reference procedures outside the forest, their digests cannot be recomputed during validation. The hashless format always includes External digests as the only exception.

## Serialization

Serialization has no branches based on the HASHLESS flag, except for setting that flag during serialization. The serializer writes digests if they exist in the MastForest, regardless of the flag. The HASHLESS flag is a declaration of intent, not a command to modify data.

## Deserialization

Deserialization doesn't have a branch on the HASHLESS flag during reading, with the twist that there are in fact two distinct deserializers, which dictate behavior. The two deserializers have fixed behaviors:

The trusted deserializer (`MastForest::read_from_bytes`) reads digests from the input and uses them directly. It errors if the HASHLESS flag is set, since hashless data cannot be trusted without validation. 

The untrusted deserializer (`UntrustedMastForest::read_from_bytes`) ignores any digests in the input and recomputes them all during validation. It warns (logs) if the HASHLESS flag is _not_ set, since the user intent violates the code path (and may e.g. indicate a user is sending a verbose MAST serialization on-chain, at needless cost).

This design catches mismatches between the user's intent (the flag) and the deserializer's behavior early.

## Default

By default, serialization uses HASHLESS. 

## Impact on Related Work

Issue #2628 (hashless format) becomes a flag rather than a separate format. The deserializer choice determines whether digests are trusted or recomputed.

Issue #2504 (size hint) is fully supported with fixed-size encoding. No iteration needed - just multiply counts by fixed sizes.

Issue #1497 (arena allocation) pairs well with this. Fixed-size nodes allocate in a contiguous array matching the disk layout, enabling zero-copy deserialization. The arena size can be sized easily from the size hint.

## Open Questions

Is 41 bytes the right node size for future extensions? Should we add an optional index section for O(1) node lookup in large forests when using un-stripped mode? How do we evolve the format if we need larger nodes later?

huitseeker:
## Summary

This PR rewrites MAST forest serialization around one fixed structural layout.

The main change is that node structure and node digests are no longer stored in the same section. Each node still has a fixed-width entry, so stripped and hashless payloads keep random access by node index. Internal node hashes now live in their own section. Hashless bytes omit that section. External node digests stay serialized because they cannot be rebuilt from local structure.

The trust boundary is now explicit. On the wire, flags only say what the serializer intended to send. They do not change what the reader trusts.

Trusted `MastForest` deserialization accepts full and stripped payloads with hashes. It rejects hashless payloads. If a caller wants to read hashless bytes, it must use `UntrustedMastForest`.

Untrusted `UntrustedMastForest` deserialization accepts full, stripped, and hashless payloads. It expects hashless input. If extra hashes or `DebugInfo` are sent, it logs an error-level message and ignores that extra data. On validation, it always recomputes all non-external node hashes from structure and uses the rebuilt table after that.

`DebugInfo` keeps one format. It is either present as one trailing section or omitted as a whole. 

This PR also makes a hard format cut-over. The new wire format is version `0.0.3`. Older MAST bytes are rejected.

## Testing & Benchmarking

<details><summary>Details</summary>
Invariants we test:

- one structural wire layout for full, stripped, and hashless payloads
- fixed-width node entries, so node count and node offsets stay cheap to read
- `HASHLESS` implies `STRIPPED`
- hashless bytes omit the general internal-hash section
- external node digests are still serialized
- trusted deserialization rejects hashless before deeper parsing
- untrusted deserialization accepts overspecified payloads but logs that extra wire data was sent
- untrusted validation recomputes non-external hashes once and then serves digest lookups from the rebuilt table
- stripped payloads round-trip without `DebugInfo`
- invalid flags, truncated sections, bad offsets, forward references, malformed batch layouts, and oversized node counts are rejected
- procedure names stay well-defined after hash recomputation, including root-hash corruption and digest-collision cases

Measured on the Miden core library `MastForest`, serialized from `CoreLibrary::default().mast_forest()`.

| Ref | Full | Stripped | Hashless |
| --- | ---: | ---: | ---: |
| `origin/next` (`bd42aeff8`) | 472,034 | 299,972 | n/a |
| `mast-serialization` (`3298fc464`) | 472,066 | 299,972 | 273,732 |

This keeps stripped size flat on the core library. Hashless saves 26,240 bytes vs stripped, or 8.75%, as expected in #2628. It saves 198,334 bytes vs full, or 42.01%.

Measured with the existing criterion benchmark in `miden-vm/benches/deserialize_core_lib.rs`, which times `Library::read_from_bytes(CoreLibrary::SERIALIZED)`.

| Ref | Benchmark | Time |
| --- | --- | ---: |
| `origin/next` (`bd42aeff8`) | `deserialize_core_lib/read_from_bytes` | 3.3250 ms |
| `mast-serialization` (`3298fc464`) | `deserialize_core_lib/read_from_bytes` | 3.1959 ms |

This benchmark measures trusted deserialization of the bundled core library package through `Library::read_from_bytes(CoreLibrary::SERIALIZED)`. On the MAST side, that reaches `MastForest::read_from` and the full trusted materialization path (see #2863). It does not measure hashless bytes, `UntrustedMastForest::validate()`, hash recomputation cost, or stripped/hashless random access.

Measured with a [standalone Criterion benchmark over bare core-library MAST bytes](https://gist.github.com/huitseeker/dd70fc331cd954a4bed265968abd3f57), timing `UntrustedMastForest::read_from_bytes(&bytes)?.validate()`.

| Ref | Mode | Time |
| --- | --- | ---: |
| `origin/next` (`bd42aeff8`) | `full` | 6.4803 ms |
| `origin/next` (`bd42aeff8`) | `stripped` | 4.8023 ms |
| `mast-serialization` (`3298fc464`) | `full` | 7.5294 ms |
| `mast-serialization` (`3298fc464`) | `stripped` | 5.1212 ms |
| `mast-serialization` (`3298fc464`) | `hashless` | 5.5215 ms |

This benchmark measures the untrusted MAST path directly, including validation and hash recomputation, without package-level `Library` overhead. `origin/next` has no hashless mode, so there is no before number for that case. But the numbers there are consistent with #2863.

Edit: Updated `MastForest` deserialization fuzzing to take the new structs into account and ran fuzzing.

</details>

## Issues

Fixes #2726: this addresses the main design point, and improves on the plan laid out in that issue. Hashless bytes no longer carry the internal node hashes, so we no longer burn fixed digest bytes just to keep the slot empty. At the same time, we keep one fixed structural layout, keep random access for stripped and hashless payloads, and keep the strict trusted vs untrusted read paths.

Fixes #2504 (maybe): this addresses the size-accounting goal for *stripped* serialization. `stripped_size_hint()` is exact, and the sectioned layout makes the byte cost of each part explicit. The fixed node-entry section also keeps the format easy to size and scan without full materialization. Hopefully we don't need size accounting for `DebugInfo` (which is gnarly), but for note size limits, this should be enough (/cc @PhilippGackstatter).

Fixes #2628: Hashless payloads now save the internal-hash bytes, which is the cleanest size win available without changing the fixed structural layout. This PR does not add a second compressed wire format, and it does not try to optimize the full or stripped formats beyond this split.

## Web explainer

https://gisthost.github.io/?cc5e99ad466cae75bdfb4b447b5509ac

## Video explainer (<100s)

https://github.com/user-attachments/assets/8e7d4d45-4ecc-4a12-af59-8b2ded533598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 #### Changes
 
+- [BREAKING] Refactored MAST forest serialization around fixed-layout full, stripped, and hashless sections, and bumped the MAST wire format to `0.0.3` ([#2765](https://github.com/0xMiden/miden-vm/pull/2765)).
 - Documented sortedness precondition more prominently for sorted array operations ([#2832](https://github.com/0xMiden/miden-vm/pull/2832)).
 - [BREAKING] Updated the Miden crypto stack to `miden-crypto` and `miden-lifted-stark` v0.24, and switched digest-ordering code to `Word`'s native lexicographic ordering ([#3039](https://github.com/0xMiden/miden-vm/pull/3039)).
 - Borrowed operation slices in basic-block batching helpers to avoid cloning in the fingerprinting path ([#2994](https://github.com/0xMiden/miden-vm/pull/2994)).
@@ -50,7 +51,6 @@
 #### Fixes
 
 - Fixed stale `ReplayProcessor` doc comment links to `ExecutionTracer` after module-structure refactors.
-- [BREAKING] Refactored MAST forest serialization around fixed-layout full, stripped, and hashless sections, and bumped the MAST wire format to `0.0.3` ([#2765](https://github.com/0xMiden/miden-vm/pull/2765)).
 
 ## 0.22.1 (2026-04-07)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@
 #### Fixes
 
 - Fixed stale `ReplayProcessor` doc comment links to `ExecutionTracer` after module-structure refactors.
+- [BREAKING] Refactored MAST forest serialization around fixed-layout full, stripped, and hashless sections, and bumped the MAST wire format to `0.0.3` ([#2765](https://github.com/0xMiden/miden-vm/pull/2765)).
 
 ## 0.22.1 (2026-04-07)
 
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -727,34 +727,6 @@ impl MastForest {
         Ok(())
     }
 
-    fn procedure_name_bindings(&self) -> Vec<(Word, Arc<str>, Option<MastNodeId>)> {
-        self.debug_info
-            .procedure_names()
-            .map(|(digest, name)| (digest, Arc::clone(name), self.find_procedure_root(digest)))
-            .collect()
-    }
-
-    fn remap_procedure_name_bindings(
-        &mut self,
-        bindings: &[(Word, Arc<str>, Option<MastNodeId>)],
-    ) -> Result<(), MastForestError> {
-        self.debug_info.clear_procedure_names();
-        for (digest, name, incoming_root_id) in bindings {
-            let resolved_digest = if let Some(root_id) = incoming_root_id {
-                self.get_node_by_id(*root_id)
-                    .ok_or(MastForestError::NodeIdOverflow(*root_id, self.nodes.len()))?
-                    .digest()
-            } else if self.find_procedure_root(*digest).is_some() {
-                *digest
-            } else {
-                return Err(MastForestError::InvalidProcedureNameDigest(*digest));
-            };
-            self.debug_info.insert_procedure_name(resolved_digest, Arc::clone(name));
-        }
-
-        Ok(())
-    }
-
     /// Validates that all BasicBlockNodes in this forest satisfy the core invariants:
     /// 1. Power-of-two number of groups in each batch
     /// 2. No operation group ends with an operation requiring an immediate value
@@ -770,36 +742,36 @@ impl MastForest {
         self.validate_procedure_name_digests()
     }
 
-    /// Recomputes node hashes in topological order and overwrites node digests.
+    /// Validates that stored node digests match the hashes implied by local structure.
     ///
-    /// For `External` nodes the digest is kept as-is because it is externally provided and cannot
-    /// be reconstructed from local structure alone.
-    fn recompute_node_hashes_in_place(&mut self) -> Result<(), MastForestError> {
-        let procedure_name_bindings = self.procedure_name_bindings();
+    /// For `External` nodes the digest is accepted as-is because it is externally provided and
+    /// cannot be reconstructed from local structure alone.
+    fn validate_node_hashes(&self) -> Result<(), MastForestError> {
         let computed_hashes = self.compute_node_hashes()?;
-        let source_forest = self.clone();
-
-        let mut rebuilt_forest = MastForest::new();
-        rebuilt_forest.debug_info = source_forest.debug_info.clone();
-        rebuilt_forest.advice_map = source_forest.advice_map.clone();
-
-        for (idx, node) in source_forest.nodes.iter().cloned().enumerate() {
-            node.to_builder(&source_forest)
-                .with_digest(computed_hashes[idx])
-                .add_to_forest_relaxed(&mut rebuilt_forest)?;
+        for (node_idx, (node, computed_digest)) in
+            self.nodes.iter().zip(computed_hashes).enumerate()
+        {
+            let expected_digest = node.digest();
+            if expected_digest != computed_digest {
+                return Err(MastForestError::HashMismatch {
+                    node_id: MastNodeId::new_unchecked(node_idx as u32),
+                    expected: expected_digest,
+                    computed: computed_digest,
+                });
+            }
         }
-
-        rebuilt_forest.roots = source_forest.roots.clone();
-        rebuilt_forest.remap_procedure_name_bindings(&procedure_name_bindings)?;
-        rebuilt_forest.commitment_cache = OnceLockCompat::new();
-        *self = rebuilt_forest;
 
         Ok(())
     }
 
     /// Computes node hashes in topological order.
     ///
+    /// The returned vector is aligned with node indices, so `digests[node_id as usize]` is the
+    /// digest of that node.
+    ///
     /// For `External` nodes, the existing digest is returned unchanged.
+    ///
+    /// Returns [`MastForestError::ForwardReference`] if nodes are not in topological order.
     fn compute_node_hashes(&self) -> Result<Vec<Word>, MastForestError> {
         use crate::chiplets::hasher;
 
@@ -1430,10 +1402,10 @@ impl UntrustedMastForest {
     ///
     /// This method performs a complete validation of the deserialized forest:
     ///
-    /// 1. Resolves procedure-name bindings against the incoming root set.
-    /// 2. Recomputes all non-external node hashes and overwrites stored digests.
-    /// 3. Rebinds procedure names to the recomputed root digests.
-    /// 4. Validates structural invariants and topological ordering against the recomputed forest.
+    /// 1. If wire node hashes are present, recomputes all non-external node hashes and requires
+    ///    them to match the serialized digests.
+    /// 2. If the payload is hashless, uses the digests rebuilt during materialization.
+    /// 3. Validates structural invariants, topological ordering, and procedure-name roots.
     ///
     /// # Returns
     ///
@@ -1449,21 +1421,22 @@ impl UntrustedMastForest {
     ///   ([`MastForestError::InvalidProcedureNameDigest`])
     /// - Any node references a child that appears later in the forest
     ///   ([`MastForestError::ForwardReference`])
+    /// - Any non-external wire digest does not match the recomputed digest
+    ///   ([`MastForestError::HashMismatch`])
     /// - Any node's digest cannot be recomputed because structural validation fails first
     ///
     /// Security convention:
-    /// - This validator never trusts serialized digest bytes for non-external nodes.
+    /// - Hashless payloads rebuild non-external digests from structure during materialization.
+    /// - If wire node hashes are present, validation recomputes them and requires them to match.
     /// - External node digests are marshaled as opaque values and are not semantically resolved
     ///   here.
     pub fn validate(self) -> Result<MastForest, MastForestError> {
         let is_hashless = self.layout.is_hashless();
-        let mut forest = self.into_materialized().map_err(MastForestError::Deserialization)?;
+        let forest = self.into_materialized().map_err(MastForestError::Deserialization)?;
 
-        // Step 1: Ensure every non-external digest comes from local structure rather than wire
-        // input. Hashless materialization already rebuilt the digest table, while full/stripped
-        // payloads still need this trust-boundary recomputation here.
+        // Step 1: Validate over-specified wire hashes instead of silently rewriting them.
         if !is_hashless {
-            forest.recompute_node_hashes_in_place()?;
+            forest.validate_node_hashes()?;
         }
 
         // Step 2: Validate the recomputed forest.
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -15,6 +15,9 @@
 //! [`UntrustedMastForest::read_from_bytes`] applies default parsing and validation budgets derived
 //! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_budget`] to tune only the
 //! wire-parsing budget, or [`UntrustedMastForest::read_from_bytes_with_budgets`] to tune both:
+//! the parsing budget limits allocations driven directly by wire counts while reading the payload,
+//! and the validation budget limits later helper allocations needed to materialize and check
+//! stripped or hashless payloads.
 //!
 //! ```ignore
 //! use miden_core::mast::UntrustedMastForest;
@@ -696,6 +699,9 @@ impl MastForest {
     /// 3. The last operation group in a batch cannot contain operations requiring immediate values
     /// 4. OpBatch structural consistency (num_groups <= BATCH_SIZE, group size <= GROUP_SIZE,
     ///    indptr integrity, bounds checking)
+    ///
+    /// This also validates that each stored procedure-name digest resolves to a procedure root in
+    /// the forest.
     ///
     /// This addresses the gap created by PR 2094, where padding NOOPs are now inserted
     /// at assembly time rather than dynamically during execution, and adds comprehensive
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -86,10 +86,7 @@ pub use debuginfo::{
 };
 
 mod serialization;
-pub use serialization::{MastNodeEntry, MastNodeInfo, SerializedMastForest};
-
-mod view;
-pub use view::MastForestView;
+pub use serialization::{MastForestView, MastNodeEntry, MastNodeInfo, SerializedMastForest};
 
 mod merger;
 pub(crate) use merger::MastForestMerger;
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -74,6 +74,9 @@ pub use debuginfo::{
 mod serialization;
 pub use serialization::{MastNodeEntry, MastNodeInfo, SerializedMastForest};
 
+mod view;
+pub use view::MastForestView;
+
 mod merger;
 pub(crate) use merger::MastForestMerger;
 pub use merger::MastForestRootMap;
@@ -517,58 +520,6 @@ impl MastForest {
     /// Returns the exact size of stripped serialization in bytes.
     pub fn stripped_size_hint(&self) -> usize {
         serialization::stripped_size_hint(self)
-    }
-}
-
-/// Read-only view over MAST node metadata.
-///
-/// This trait is implemented by both in-memory [`MastForest`] and serialized
-/// [`SerializedMastForest`] representations, enabling callers to consume a shared random-access
-/// API independent of backing storage.
-pub trait MastForestView {
-    /// Returns the number of nodes in the forest.
-    fn node_count(&self) -> usize;
-
-    /// Returns fixed-width structural metadata for a node at the specified index.
-    fn node_entry_at(&self, index: usize) -> Result<MastNodeEntry, DeserializationError>;
-
-    /// Returns the digest of the node at the specified index.
-    fn node_digest_at(&self, index: usize) -> Result<Word, DeserializationError>;
-
-    /// Returns serialized-equivalent metadata for a node at the specified index.
-    fn node_info_at(&self, index: usize) -> Result<MastNodeInfo, DeserializationError> {
-        Ok(MastNodeInfo::from_entry(
-            self.node_entry_at(index)?,
-            self.node_digest_at(index)?,
-        ))
-    }
-
-    /// Returns the number of procedure roots in the forest.
-    fn procedure_root_count(&self) -> usize;
-
-    /// Returns the procedure root id at the specified index.
-    fn procedure_root_at(&self, index: usize) -> Result<MastNodeId, DeserializationError>;
-
-    /// Returns true when the forest contains no nodes.
-    fn is_empty(&self) -> bool {
-        self.node_count() == 0
-    }
-
-    /// Returns true when `index` is a valid node index.
-    fn has_node(&self, index: usize) -> bool {
-        index < self.node_count()
-    }
-
-    /// Returns all node infos in index order.
-    fn all_node_infos(&self) -> Result<Vec<MastNodeInfo>, DeserializationError> {
-        (0..self.node_count()).map(|index| self.node_info_at(index)).collect()
-    }
-
-    /// Returns all procedure roots in index order.
-    fn procedure_roots(&self) -> Result<Vec<MastNodeId>, DeserializationError> {
-        (0..self.procedure_root_count())
-            .map(|index| self.procedure_root_at(index))
-            .collect()
     }
 }
 
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -12,20 +12,34 @@
 //!     .validate()?;
 //! ```
 //!
-//! For maximum protection against denial-of-service attacks from malicious input, use
-//! [`UntrustedMastForest::read_from_bytes_with_budget`] which limits memory consumption:
+//! [`UntrustedMastForest::read_from_bytes`] applies default parsing and validation budgets derived
+//! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_budget`] to tune only the
+//! wire-parsing budget, or [`UntrustedMastForest::read_from_bytes_with_budgets`] to tune both:
 //!
 //! ```ignore
 //! use miden_core::mast::UntrustedMastForest;
 //!
-//! // Budget limits pre-allocation sizes and total bytes consumed
+//! // Parsing budget only
 //! let forest = UntrustedMastForest::read_from_bytes_with_budget(&bytes, bytes.len())?
+//!     .validate()?;
+//!
+//! // Parsing budget plus explicit validation-allocation budget
+//! let forest = UntrustedMastForest::read_from_bytes_with_budgets(&bytes, bytes.len(), bytes.len() * 7)?
 //!     .validate()?;
 //! ```
 //!
 //! This recomputes all node hashes and checks structural invariants before returning a usable
 //! `MastForest`. Direct deserialization via `MastForest::read_from_bytes` trusts the serialized
 //! hashes and should only be used for data from trusted sources (e.g. compiled locally).
+//!
+//! In practice, the public entry points split into three policies:
+//! - [`MastForest::read_from_bytes`]: trusted full deserialization; rejects hashless payloads and
+//!   trusts serialized non-external digests.
+//! - [`SerializedMastForest::new`]: trusted inspection path; scans only the structural layout
+//!   needed for random access and may accept full, stripped, or hashless payloads.
+//! - [`UntrustedMastForest::read_from_bytes`] and
+//!   [`UntrustedMastForest::read_from_bytes_with_budgets`]: untrusted paths; parse with bounded
+//!   readers and require [`UntrustedMastForest::validate`] before use.
 
 use alloc::{
     collections::{BTreeMap, BTreeSet},
@@ -1346,6 +1360,7 @@ pub struct UntrustedMastForest {
     layout: serialization::ForestLayout,
     advice_map: AdviceMap,
     debug_info: DebugInfo,
+    remaining_allocation_budget: Option<usize>,
 }
 
 impl UntrustedMastForest {
@@ -1398,11 +1413,15 @@ impl UntrustedMastForest {
 
     /// Deserializes an [`UntrustedMastForest`] from bytes.
     ///
-    /// This method uses a [`BudgetedReader`] with a budget equal to the input size to protect
-    /// against denial-of-service attacks from malicious input.
+    /// This method uses a [`BudgetedReader`] plus a bounded validation-allocation budget derived
+    /// from the input size to protect against denial-of-service attacks from malicious input.
+    /// The default validation budget includes room for the retained serialized copy used by the
+    /// deferred-validation path, in addition to stripped/hashless helper allocations. Concretely,
+    /// the default is `bytes.len()` for parsing and `bytes.len() * 7` for validation allocations.
+    /// That `* 7` factor is a coarse convenience bound, not an exact peak-memory formula.
     ///
-    /// For stricter limits, use
-    /// [`read_from_bytes_with_budget`](Self::read_from_bytes_with_budget) with a custom budget.
+    /// For explicit parsing and validation limits, use
+    /// [`read_from_bytes_with_budgets`](Self::read_from_bytes_with_budgets).
     ///
     /// # Example
     ///
@@ -1414,7 +1433,11 @@ impl UntrustedMastForest {
     /// let forest = untrusted.validate()?;
     /// ```
     pub fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
-        Self::read_from_bytes_with_budget(bytes, bytes.len())
+        Self::read_from_bytes_with_budgets(
+            bytes,
+            bytes.len(),
+            serialization::default_untrusted_allocation_budget(bytes.len()),
+        )
     }
 
     /// Deserializes an [`UntrustedMastForest`] from bytes and returns the raw wire flags.
@@ -1422,7 +1445,11 @@ impl UntrustedMastForest {
     /// This enables callers to inspect serializer intent flags (e.g., HASHLESS) without affecting
     /// the untrusted deserialization path.
     pub fn read_from_bytes_with_flags(bytes: &[u8]) -> Result<(Self, u8), DeserializationError> {
-        Self::read_from_bytes_with_budget_and_flags(bytes, bytes.len())
+        Self::read_from_bytes_with_budgets_and_flags(
+            bytes,
+            bytes.len(),
+            serialization::default_untrusted_allocation_budget(bytes.len()),
+        )
     }
 
     /// Deserializes an [`UntrustedMastForest`] from bytes with a byte budget.
@@ -1434,13 +1461,13 @@ impl UntrustedMastForest {
     /// # Arguments
     ///
     /// * `bytes` - The serialized forest bytes
-    /// * `budget` - Maximum bytes to consume during deserialization. Set this to `bytes.len()` for
-    ///   typical use cases, or lower to enforce stricter limits.
+    /// * `budget` - Maximum bytes to consume while parsing the wire payload and pre-sizing
+    ///   wire-driven collections via [`BudgetedReader`]
     ///
     /// # Example
     ///
     /// ```ignore
-    /// // Read from untrusted source with budget equal to input size
+    /// // Read from untrusted source with an explicit parsing budget
     /// let untrusted = UntrustedMastForest::read_from_bytes_with_budget(&bytes, bytes.len())?;
     ///
     /// // Validate before use
@@ -1455,12 +1482,15 @@ impl UntrustedMastForest {
     ///
     /// This prevents attacks where malicious input claims an unrealistic number of elements
     /// (e.g., `len = 2^60`), causing excessive memory allocation before any data is read.
+    ///
+    /// To also cap stripped/hashless validation helper allocations, use
+    /// [`read_from_bytes_with_budgets`](Self::read_from_bytes_with_budgets).
     pub fn read_from_bytes_with_budget(
         bytes: &[u8],
         budget: usize,
     ) -> Result<Self, DeserializationError> {
         let mut reader = BudgetedReader::new(SliceReader::new(bytes), budget);
-        Self::read_from(&mut reader)
+        serialization::read_untrusted_with_flags(&mut reader).map(|(forest, _flags)| forest)
     }
 
     /// Deserializes an [`UntrustedMastForest`] from bytes with a byte budget and returns flags.
@@ -1470,5 +1500,38 @@ impl UntrustedMastForest {
     ) -> Result<(Self, u8), DeserializationError> {
         let mut reader = BudgetedReader::new(SliceReader::new(bytes), budget);
         serialization::read_untrusted_with_flags(&mut reader)
+    }
+
+    /// Deserializes an [`UntrustedMastForest`] from bytes with separate parsing and validation
+    /// budgets.
+    ///
+    /// `parsing_budget` limits wire-driven parsing and collection pre-sizing. `validation_budget`
+    /// additionally caps tracked stripped/hashless helper allocations such as empty debug-info
+    /// scaffolding, digest slot tables, and rebuilt digest tables.
+    pub fn read_from_bytes_with_budgets(
+        bytes: &[u8],
+        parsing_budget: usize,
+        validation_budget: usize,
+    ) -> Result<Self, DeserializationError> {
+        let mut reader = BudgetedReader::new(SliceReader::new(bytes), parsing_budget);
+        serialization::read_untrusted_with_flags_and_allocation_budget(
+            &mut reader,
+            validation_budget,
+        )
+        .map(|(forest, _flags)| forest)
+    }
+
+    /// Deserializes an [`UntrustedMastForest`] from bytes with separate parsing and validation
+    /// budgets and returns flags.
+    pub fn read_from_bytes_with_budgets_and_flags(
+        bytes: &[u8],
+        parsing_budget: usize,
+        validation_budget: usize,
+    ) -> Result<(Self, u8), DeserializationError> {
+        let mut reader = BudgetedReader::new(SliceReader::new(bytes), parsing_budget);
+        serialization::read_untrusted_with_flags_and_allocation_budget(
+            &mut reader,
+            validation_budget,
+        )
     }
 }
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -72,7 +72,7 @@ pub use debuginfo::{
 };
 
 mod serialization;
-pub use serialization::{MastNodeInfo, SerializedMastForest};
+pub use serialization::{MastNodeEntry, MastNodeInfo, SerializedMastForest};
 
 mod merger;
 pub(crate) use merger::MastForestMerger;
@@ -517,8 +517,6 @@ impl MastForest {
     }
 
     /// Returns the exact size of stripped serialization in bytes.
-    ///
-    /// Hashless serialization has the same size as stripped serialization.
     pub fn stripped_size_hint(&self) -> usize {
         serialization::stripped_size_hint(self)
     }
@@ -533,8 +531,19 @@ pub trait MastForestView {
     /// Returns the number of nodes in the forest.
     fn node_count(&self) -> usize;
 
+    /// Returns fixed-width structural metadata for a node at the specified index.
+    fn node_entry_at(&self, index: usize) -> Result<MastNodeEntry, DeserializationError>;
+
+    /// Returns the digest of the node at the specified index.
+    fn node_digest_at(&self, index: usize) -> Result<Word, DeserializationError>;
+
     /// Returns serialized-equivalent metadata for a node at the specified index.
-    fn node_info_at(&self, index: usize) -> Result<MastNodeInfo, DeserializationError>;
+    fn node_info_at(&self, index: usize) -> Result<MastNodeInfo, DeserializationError> {
+        Ok(MastNodeInfo::from_entry(
+            self.node_entry_at(index)?,
+            self.node_digest_at(index)?,
+        ))
+    }
 
     /// Returns the number of procedure roots in the forest.
     fn procedure_root_count(&self) -> usize;
@@ -550,11 +559,6 @@ pub trait MastForestView {
     /// Returns true when `index` is a valid node index.
     fn has_node(&self, index: usize) -> bool {
         index < self.node_count()
-    }
-
-    /// Returns the digest of the node at the specified index.
-    fn node_digest_at(&self, index: usize) -> Result<Word, DeserializationError> {
-        Ok(self.node_info_at(index)?.digest())
     }
 
     /// Returns all node infos in index order.
@@ -701,6 +705,58 @@ impl MastForest {
 // ------------------------------------------------------------------------------------------------
 /// Validation methods
 impl MastForest {
+    fn validate_basic_block_invariants(&self) -> Result<(), MastForestError> {
+        for (node_id_idx, node) in self.nodes.iter().enumerate() {
+            let node_id =
+                MastNodeId::new_unchecked(node_id_idx.try_into().expect("too many nodes"));
+            if let MastNode::Block(basic_block) = node {
+                basic_block.validate_batch_invariants().map_err(|error_msg| {
+                    MastForestError::InvalidBatchPadding(node_id, error_msg)
+                })?;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn validate_procedure_name_digests(&self) -> Result<(), MastForestError> {
+        for (digest, _) in self.debug_info.procedure_names() {
+            if self.find_procedure_root(digest).is_none() {
+                return Err(MastForestError::InvalidProcedureNameDigest(digest));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn procedure_name_bindings(&self) -> Vec<(Word, Arc<str>, Option<MastNodeId>)> {
+        self.debug_info
+            .procedure_names()
+            .map(|(digest, name)| (digest, Arc::clone(name), self.find_procedure_root(digest)))
+            .collect()
+    }
+
+    fn remap_procedure_name_bindings(
+        &mut self,
+        bindings: &[(Word, Arc<str>, Option<MastNodeId>)],
+    ) -> Result<(), MastForestError> {
+        self.debug_info.clear_procedure_names();
+        for (digest, name, incoming_root_id) in bindings {
+            let resolved_digest = if let Some(root_id) = incoming_root_id {
+                self.get_node_by_id(*root_id)
+                    .ok_or(MastForestError::NodeIdOverflow(*root_id, self.nodes.len()))?
+                    .digest()
+            } else if self.find_procedure_root(*digest).is_some() {
+                *digest
+            } else {
+                return Err(MastForestError::InvalidProcedureNameDigest(*digest));
+            };
+            self.debug_info.insert_procedure_name(resolved_digest, Arc::clone(name));
+        }
+
+        Ok(())
+    }
+
     /// Validates that all BasicBlockNodes in this forest satisfy the core invariants:
     /// 1. Power-of-two number of groups in each batch
     /// 2. No operation group ends with an operation requiring an immediate value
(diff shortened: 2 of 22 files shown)


core/src/operations/mod.rs
```diff
@@ -904,6 +904,101 @@ impl Serializable for Operation {
...
+impl Operation {
+    /// Returns the serialized size of this operation in bytes.
+    pub(crate) fn encoded_size(&self) -> usize {
+        let mut size = core::mem::size_of::<u8>();
+        match self {
+            Operation::Assert(err_code)
+            | Operation::MpVerify(err_code)
+            | Operation::U32assert2(err_code) => {
+                size += err_code.get_size_hint();
+            },
```
bobbinth, line 916: If we really wanted to go overboard, we could treat error codes in the same was as we treat decorators as technically, error codes are a part of debug info. But the extra complexity is probably not worth it.

core/src/mast/serialization/info.rs
```diff
@@ -12,19 +12,17 @@ use crate::{
...
-/// index.
-#[derive(Debug)]
-pub struct MastNodeInfo {
+/// This is the random-access portion of the node table. Digests are intentionally modeled
+/// separately so the wire format can move them into dedicated sections.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MastNodeEntry {
     ty: MastNodeType,
-    digest: Word,
 }
```
bobbinth: Do we need both `MastNodeEntry` and `MastNodeType` now? It seems like this is creating an extra level of nesting w/o any significant benefits. We could probably collapse them into a single struct - and I do like the `MastNodeEntry` name better. So, we could have:

```rust
pub enum MastNodeEntry {
    Join {
        left_child_id: u32,
        right_child_id: u32,
    } = JOIN,
    Split {
        if_branch_id: u32,
        else_branch_id: u32,
    } = SPLIT,
    Loop {
        body_id: u32,
    } = LOOP,
    Block {
        // offset of operations in node data
        ops_offset: u32,
    } = BLOCK,
    Call {
        callee_id: u32,
    } = CALL,
    SysCall {
        callee_id: u32,
    } = SYSCALL,
    Dyn = DYN,
    Dyncall = DYNCALL,
    External = EXTERNAL,
}
```

core/src/mast/mod.rs
```diff
@@ -499,8 +500,74 @@ impl MastForest {
...
+        serialization::stripped_size_hint(self)
+    }
+}
+
+/// Read-only view over MAST node metadata.
+///
+/// This trait is implemented by both in-memory [`MastForest`] and serialized
+/// [`SerializedMastForest`] representations, enabling callers to consume a shared random-access
+/// API independent of backing storage.
+pub trait MastForestView {
```
bobbinth: I would maybe move this into a separate file. This file is already pretty huge and difficult to navigate - I'd split it up (but in a different PR).

core/src/mast/mod.rs
```diff
@@ -667,20 +726,80 @@ impl MastForest {
...
-    ///
-    /// This method iterates through all nodes in index order, verifying:
-    /// 1. All child references point to nodes with smaller indices (topological order)
-    /// 2. Each node's recomputed digest matches its stored digest
+    fn procedure_name_bindings(&self) -> Vec<(Word, Arc<str>, Option<MastNodeId>)> {
+        self.debug_info
+            .procedure_names()
+            .map(|(digest, name)| (digest, Arc::clone(name), self.find_procedure_root(digest)))
+            .collect()
+    }
```
bobbinth: nit: I would name this `build_procedure_name_bindings()` as this performs non-trivial amount of work.

As a general rule, unless a method just reads an internal field, I try to use `get_` and `build_` prefixes to signal the amount of work the method performs: `get_` for when the amount of work is pretty small (e.g., reading data from a vector or constructing a simple struct from some internal data) and `build_` when it is significant.

I know we haven't been following this rule consistently - so, no need to do any kind of comprehensive renaming - but I did want to bring it up as that's what the codebase used to follow in the past.

core/src/mast/mod.rs
```diff
@@ -667,20 +726,80 @@ impl MastForest {
...
+
+    fn remap_procedure_name_bindings(
+        &mut self,
+        bindings: &[(Word, Arc<str>, Option<MastNodeId>)],
+    ) -> Result<(), MastForestError> {
```
bobbinth: Let's add some doc comments here to explain what this method does.

Also, I'd move it after `recompute_node_hashes_in_place()` method as these two methods are logically related.

core/src/mast/mod.rs
```diff
@@ -667,20 +726,80 @@ impl MastForest {
...
+        for (idx, node) in source_forest.nodes.iter().cloned().enumerate() {
+            node.to_builder(&source_forest)
+                .with_digest(computed_hashes[idx])
+                .add_to_forest_relaxed(&mut rebuilt_forest)?;
+        }
+
+        rebuilt_forest.roots = source_forest.roots.clone();
+        rebuilt_forest.remap_procedure_name_bindings(&procedure_name_bindings)?;
+        rebuilt_forest.commitment_cache = OnceLockCompat::new();
+        *self = rebuilt_forest;
```
bobbinth: It is not clear to me why we need to do procedure node remappings. My current understanding is:

- Procedure names are a part of debug info. So they don't exist for the "hashless" format, and therefore, no remapping should be needed.
- For the format "with hashes" all the mappings should be already correct (if any of the digests is incorrect, that would be an error) - and so again, no remapping should be necessary.

Am I missing something?

core/src/mast/mod.rs
```diff
@@ -667,20 +726,80 @@ impl MastForest {
...
+        Ok(())
+    }
+
+    /// Computes node hashes in topological order.
     ///
-    /// Returns `MastForestError::HashMismatch` if any node's recomputed digest doesn't
-    /// match its stored digest.
-    fn validate_node_hashes(&self) -> Result<(), MastForestError> {
+    /// For `External` nodes, the existing digest is returned unchanged.
+    fn compute_node_hashes(&self) -> Result<Vec<Word>, MastForestError> {
```
bobbinth, line 744: The order of digests in the return vector is the same as the order of nodes, right? (i.e., we can use `digests[node_id]` would give us the digest of the node with `node_id`). If so, I'd mention this in the doc comment.

I may also make sense to mention that if the nodes are not in topological order, this method will panic (I think).

core/src/mast/serialization/mod.rs
bobbinth: Why do we need to keep decorator count here? Is it useful for anything?

More generally, is this still the best layout? For example, is there still benefit to serializing basic blocks before node data?

core/src/mast/serialization/mod.rs
```diff
@@ -13,8 +25,14 @@
...
-//! (Node info section)
-//! - MAST node infos (`Vec<MastNodeInfo>`)
+//! (Node entries section)
+//! - fixed-width structural node entries (`Vec<MastNodeEntry>`)
+//!
+//! (External digest section)
+//! - digests for `External` nodes only (`Vec<Word>`, ordered by node index)
+//!
+//! (Node hash section - omitted if FLAGS bit 1 is set)
+//! - digests for all non-external nodes (`Vec<Word>`, ordered by node index)
```
bobbinth, line 58: It would be good to provide a bit more info about these sections. My current understanding:

- For each node, we write `MastNodeEntry` which is 8 bytes. These entries are guaranteed to be topologically sorted (i.e., children are written before parents).
- Then, for each external node we write their digests. What is not clear to me is how, given a node entry for an external node, we can look up its digest. For example:
  - We could include the offset into the `MastNodeType::Dyncall` variant (similar to how we do it with the `MastNodeType::Block` variant) - but I don't think that's happening now.
  - We could sort the nodes in the forest so that external nodes are always always come before any other nodes. This would make it trivial to look up the external node digest by `MastNodeId`. But I'm not sure this is happening either.
- Same question for non-external nodes: how do we look up hashes for these nodes for a given `MastNodeEntry`?

Btw, if we do sort the nodes such that external nodes always come first, it would have a couple of advantages:

- External and non-external digests sections become contiguous - i.e., there could be just a single "node digests" section, and for hashless serialization we'd just truncate everything that comes after last external node.
- Looking up node info becomes almost trivial:
  - `MastNodeEntry` would be at `node_entries_offset + mast_node_index`.
  - Node digest would be at `node_digests + mast_node_index`.
- All external node digests would be conveniently located next to each other and we could use this to compute external node hash (which would be useful for quick checks of MAST forest equality - I tagged you in a somewhat related issue in `miden-protocol`).

core/src/mast/serialization/mod.rs
```diff
@@ -182,134 +239,435 @@ impl MastForest {
...
+/// assert_eq!(view.node_count(), forest.nodes().len());
+/// assert!(view.node_info_at(0).is_ok());
+/// ```
+#[derive(Debug)]
+pub struct SerializedMastForest<'a> {
+    bytes: &'a [u8],
+    flags: WireFlags,
+    layout: ForestLayout,
+    resolved: OnceLockCompat<Result<ResolvedSerializedForest<'a>, DeserializationError>>,
+}
```
bobbinth, line 389: Not a comment about this struct specifically, but about the general structure (assuming I understood it correctly) - overall, I like it, but I wonder if we can simplify it somehow by maybe giving up some flexibility and maybe imposing some constraints on the layout of the serialized data (e.g., see previous comment).

For example, maybe this struct could be something like:

```rust
pub struct SerializedMastForest<'a> {
    bytes: &'a [u8],
    flags: WireFlags,
    layout: ForestLayout,
}

impl<'a> SerializedMastForest<'a> {

    /// Returns node entry for the specified MAST node.
    pub fn read_node_entry_at(&self, id: MastNodeId) -> Result<MastNodeEntry, SomeError> {
        self.layout.read_node_entry_at(self.bytes, index)
    }

    /// Returns the digest for the specified MAST node unless the underlying format is hashless,
    /// in which case, returns an error
    pub fn read_node_digest(&self, id: MastNodeId) -> Result<Word, SomeError> {
        ...
    }

    // MastForest materialization
    //---------------------------------------------------------------------------------------------

    /// Materializes the `MastForest`. If the underlying format is hashless, returns an error.
    pub fn deserialize_trusted(&self) -> Result<MastForest, SomeError> {
        ...
    }

    /// Materializes the `MastForest`. Woks for both hashless and non-hashless formats.
    pub fn deserialize_untrusted(&self) -> Result<MastForest, SomeError> {
        ...
    }
}
```
I think this would still give us roughly similar functionality, but may simplify the overall structure quite a bit.

bobbinth (commented):
Looks great! Thank you! I left some comments inline. The main thing I wonder is whether we can simplify the overall structure somehow as it is a bit difficult to follow now.

core/src/operations/mod.rs
huitseeker: I think that’s a reasonable follow-up idea, but I left it out of this PR. Unlike decorators, error codes are still part of the operation encoding path today, so splitting them out would add format complexity for a smaller win. But I'm up for revisiting that as a separate ‘strip more diagnostics’ change.

core/src/mast/mod.rs
huitseeker: Yes, this PR encodes the following design decision:
- Untrusted deserialization will copiously warn (at the error level) on over-specified (read: hash-full, or with-`DebugInfo`) formats, but will not reject them — rather recomputing hashes instead of failing.

This PR had not taken the following design decision — which I think you are hinting at, and would be sufficient for avoiding the remapping:
- if a user passes an over-specified payload (hash-full or with-`DebugInfo`) to untrusted serialization, it *must* contain hashes *equal* to those that will be recomputed to succeed.

That should explain why the code is maximally permissive. I'll make the switch now.

core/src/mast/mod.rs
huitseeker: Done. The doc now says the returned vector is aligned with node indices, and that forward references return `MastForestError::ForwardReference`.

core/src/mast/serialization/mod.rs
huitseeker: Good catch. decorator_count was no longer used by the layout scanner or the readers, so I removed it from the wire format.

Re: basic blocks before node entries, I kept that ordering because, as you know, `Block` entries still store offsets into the basic-block section. Keeping the blob first lets those offsets be final without backpatching and makes it easy to keep node entries fixed-width.

core/src/mast/serialization/mod.rs
huitseeker: I think the API shape you sketched is close to what the code is already doing. SerializedMastForest is the structural view; the extra resolved layer is an internal cache for digest lookup/materialization, especially for hashless payloads where the first digest-backed access needs one-time recomputation. I kept that internal so callers still get a simple API.

core/src/mast/serialization/mod.rs
huitseeker: I added docs for this. The external and non-external digest sections are dense-by-kind, and readers build a compact slot-by-node table from the node-entry section once. That gives O(1) per-node digest lookup without forcing external nodes to be physically reordered.

core/src/mast/mod.rs
huitseeker: I've done a minimal split of the `MastForestView` trait in its own file.

core/src/mast/mod.rs
huitseeker: As discussed elsewhere, this method isn't there any more. For the convention, we should document it because it differs from the more well known Rust APi guideline [C-GETTER](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter). Folks unfamiliar with the code base are unlikely to know what's going on. And FWIW, I find the best way to convey some significant computation is involved is a `#[must_use]` attribute.

core/src/mast/mod.rs
huitseeker: This method should no longer exist.

core/src/mast/mod.rs
huitseeker: Done, see `test_untrusted_forest_rejects_mismatched_wire_root_hash`, `test_untrusted_forest_rejects_invalid_procedure_name_digest_without_remapping`.

core/src/mast/serialization/info.rs
huitseeker: Done!

core/src/mast/serialization/layout.rs
```diff
@@ -0,0 +1,423 @@
...
+    let roots_len_bytes = roots_count
+        .checked_mul(core::mem::size_of::<u32>())
+        .ok_or_else(|| DeserializationError::InvalidValue("roots length overflow".to_string()))?;
+    let _roots_data = source.read_slice(roots_len_bytes)?;
+
+    let basic_block_len = source.read_usize()?;
+    let basic_block_offset = source.offset().checked_sub(body_start).ok_or_else(|| {
+        DeserializationError::InvalidValue("basic-block offset underflow".to_string())
+    })?;
+    let _basic_block_data = source.read_slice(basic_block_len)?;
```
plafer, line 294: Currently, we store the section lengths right before the data itself. Wouldn't it be better to store all lengths as part of the header at the start? In principle, this allows us to reject a serialized MAST forest early (i.e. read the header from the socket, and reject if those values don't make sense)? Of course, an attacker can forge those values, so maybe it's not protecting against anything? Just thinking out loud here

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+fn build_digest_slot_by_node(
+    bytes: &[u8],
+    layout: &ForestLayout,
+) -> Result<Vec<u32>, DeserializationError> {
+    // Digest sections are packed densely by node kind rather than by absolute node index.
+    // This scan records, for each node index, which slot to read from in the corresponding digest
+    // source.
+    let mut slots = Vec::with_capacity(layout.node_count);
```
plafer: `layout.node_count` is untrusted at this point, so this could allocate an arbitrary amount of memory.

Would it make sense for `ForestLayout` to hold trusted values - i.e. all fields have been validated against a (preferably tunable) maximum? If not, then we could also introduce an `UntrustedForestLayout`.

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,331 @@
...
+    bytes: &[u8],
+    layout: &ForestLayout,
+) -> Result<Vec<crate::Word>, DeserializationError> {
+    let basic_block_data_decoder = BasicBlockDataDecoder::new(basic_block_data(
+        bytes,
+        layout.basic_block_offset,
+        layout.basic_block_len,
+    )?);
+
+    let mut digests = Vec::with_capacity(layout.node_count);
```
plafer: Same here, `layout.node_count` is untrusted

core/src/mast/serialization/view.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// This trait is implemented by both in-memory [`super::MastForest`] and serialized
+/// [`super::SerializedMastForest`] representations, enabling callers to consume a shared
+/// random-access API independent of backing storage.
+pub trait MastForestView {
+    /// Returns the number of nodes in the forest.
+    fn node_count(&self) -> usize;
+
+    /// Returns fixed-width structural metadata for a node at the specified index.
+    fn node_entry_at(&self, index: usize) -> Result<MastNodeEntry, DeserializationError>;
```
plafer, line 18: Although this is not a serialization-only type, this method (and others such as `node_info_at()`) have serialization concerns bleed into the API. Hence, I think we should make this explicitly at serialization-only trait (i.e. under `serialization/`).

For example, in #2766 we discuss introducing a `SparseMastForest`, which would imply we need to introduce another `MastForest` trait for the purposes of `execute_impl()` (which would only require a `get_node_by_id()` method I believe).

core/src/mast/serialization/layout.rs
```diff
@@ -0,0 +1,423 @@
...
+pub(super) struct WireFlags(u8);
+
+impl ForestLayout {
+    pub(crate) fn is_hashless(&self) -> bool {
+        self.node_hash_offset.is_none()
+    }
+
+    /// Converts scan-time offsets, which are relative to the post-header payload, into byte
+    /// offsets into the full serialized blob.
+    pub(super) fn resolve(mut self) -> Result<Self, DeserializationError> {
```
plafer: This complicates the mental model, since when I get a `ForestLayout`, I never *know* what the offsets refer to; they might be relative to the start of the buffer, relative to after the header, or relative to byte position `HEADER_LEN * n` (if this method was incorrectly called `n` times). Also, it's probably incorrect to call the "read" methods (such as `read_fixed_section_entry()`) before calling `resolve()` right? 

Is there a reason we need 2 definitions of offset (i.e. one from the start of the buffer, and one from after the header)? Couldn't we have them start from the start of the file always?

plafer (commented):
Looks good so far! Not a full review yet

core/src/mast/serialization/layout.rs
huitseeker: Agreed, that was too easy to misuse. I changed `ForestLayout` so all offsets are absolute from the start of the file and deleted `resolve()`. So there is now a single offset model, and the read helpers are valid immediately on the scanned layout.

core/src/mast/serialization/layout.rs
huitseeker: I left that out of this PR. It would be a more invasive wire-format change, and on the untrusted path those header lengths would still be attacker-controlled, so we’d still need the same section-by-section validation. I think that’s better handled as a separate format-design follow-up if we want earlier streaming rejection semantics.

core/src/mast/serialization/resolved.rs
huitseeker: I documented the policy split more explicitly instead of introducing a second layout type, but made its contract explicit: it represents structurally validated section layout, not ‘trusted contents’. On the untrusted path, the reader determines policy bounds (`SliceReader` for trusted inspection, `BudgetedReader` for untrusted parsing); and later validation-time allocations are budgeted too. Full untrusted validation stays in `UntrustedMastForest`.

core/src/mast/serialization/view.rs
huitseeker: I moved `MastForestView` under `mast::serialization` and kept the top-level re-export unchanged for now. I think that gives us the right layering: this trait stays about serialized-equivalent metadata, while a future execution-facing trait for `SparseMastForest` can stay minimal and not inherit serialization concerns.

core/src/mast/serialization/resolved.rs
huitseeker: Answered [above](https://github.com/0xMiden/miden-vm/pull/2765#discussion_r2979966065)

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,380 @@
...
+    basic_block_len: usize,
+) -> Result<&[u8], DeserializationError> {
+    let end = basic_block_offset.checked_add(basic_block_len).ok_or_else(|| {
+        DeserializationError::InvalidValue("basic-block data overflow".to_string())
+    })?;
+    if end > bytes.len() {
+        return Err(DeserializationError::UnexpectedEOF);
+    }
+    Ok(&bytes[basic_block_offset..end])
+}
```
plafer, line 242: Probably the thing I've found most challenging with this PR is keeping track of the trust boundaries across the different data structures. For example, I was trying to answer the question "when does the `ForestLayout` get validated?", and I think the answer is "a bit everywhere"? For instance, this function is called to create the `node_data` input to `BasicBlockDataDecoder`, and it validates the basic block layout (or at least partially - it makes sure that it lies within bounds). So I can infer from this function that wherever this was called, the layout was not validated yet.

I wonder if we could extend the `UntrustedMastForest` idea fully - that is, create "sanitized" vs "unsanitized" versions of every type we use (`usize`, `Felt`, `Word`, etc). Each function takes the unsanitized version if it's going to be checking boundaries, or the sanitized version if it just wants to assume values are valid. Though I'm not sure if it would be overly verbose and end up complicating things even more.

This would naturally be to be explored in a follow-up PR.

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,380 @@
...
+/// Non-external nodes either read from the internal-hash section or from a rebuilt in-memory hash
+/// table. External nodes always read from the external-digest section.
+#[derive(Debug, Clone)]
+struct ForestDigests {
+    /// Dense slot index for each node within its digest section.
+    ///
+    /// External nodes index into the external-digest section. All other nodes index into the
+    /// general node-hash section or the rebuilt hash table.
+    slot_by_node: Vec<u32>,
+    hash_table: Option<Vec<crate::Word>>,
```
plafer, line 33: I'd document what `None` vs `Some` means - i.e. `None` means the hashes are present in the serialized buffer

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,380 @@
...
+        Ok(Self { slot_by_node, hash_table })
+    }
+
+    fn digest_at(
+        &self,
+        bytes: &[u8],
+        layout: &ForestLayout,
+        index: usize,
+        entry: MastNodeEntry,
+    ) -> Result<crate::Word, DeserializationError> {
```
plafer, line 79: I'd add a docstring, giving an intuition for how the various parameters are used in what condition

core/src/mast/serialization/mod.rs
```diff
@@ -28,28 +68,57 @@
 //! - NodeToDecoratorIds CSR (before_enter and after_exit decorators, dense representation)
 //! - Procedure names map (`BTreeMap<Word, String>`)
 //!
-//! # Stripped Format
+//! In stripped format, the `DebugInfo` section is omitted and readers materialize an empty
+//! `DebugInfo`.
+//!
+//! In hashless format, the internal node-hash section is omitted and `HASHLESS` also implies
+//! `STRIPPED`. External node digests still stay on the wire because they cannot be rebuilt from
+//! local structure.
```
plafer: What is the reason for hashless to imply stripped? Does something break if we try to decode in hashless mode, but not stripped? Otherwise, wouldn't it be simpler to not create a relationship between them?

core/src/mast/serialization/resolved.rs
```diff
@@ -0,0 +1,380 @@
...
+        let digest_slot = self.slot_by_node[index] as usize;
+
+        if matches!(entry, MastNodeEntry::External) {
+            return read_digest_entry(bytes, layout.external_digest_offset, digest_slot);
+        }
+
+        if let Some(hash_table) = &self.hash_table {
+            return Ok(hash_table[index]);
+        }
```
plafer: This method (`ForestDigests::digest_at()`) assumes that `index` is trusted/sanitized right? Which I infer from us using `slot_by_node[index]` instead of `slot_by_node.get(index)`.

This is another example to show that the trust boundaries are hard to follow. To double check this, I ended up moving up the call chain to see from where this is called (the call tree is relatively deep too). One of the call sites is the public `SerializedMastForest::node_info_at()`. The method docstring does not establish the trust boundary for the `index`, and `SerializedMastForest` is also used in the untrusted path - so shouldn't I assume that `index` is *untrusted*?

plafer (commented):
This PR tackles a fundamentally complex problem (i.e. many tangled up concerns), and is a great step forward. However as I mentioned in a few comments, *convincing* myself that all attacker-controlled values are properly sanitized is difficult.

There's a few ways to go about this - either refactor the structure in this PR or merge more or less as-is and open an issue for future improvements.

core/src/mast/serialization/resolved.rs
huitseeker: Done

core/src/mast/serialization/mod.rs
huitseeker: There's no use case for it: there's no use giving someone a debug-laden execution payload across a trust boundary (and if you are billed for it, it's a waste of money).

Nothing fundamental breaks structurally, but no supported consumer treats wire DebugInfo as trusted in hashless mode: MastForest rejects hashless, `SerializedMastForest::new()` accepts it only for structural inspection, and the untrusted path rebuilds what it trusts before use. Allowing ‘hashless but with debug info’ would add another wire mode without changing validation semantics.”

core/src/mast/serialization/resolved.rs
huitseeker: Tightened. The public accessors already bounds-checked before reaching digest resolution, but I removed the unchecked indexing anyway and now return a descriptive `DeserializationError`.

huitseeker:
@plafer It's true that `ForestLayout` gets structural validation up front, while some narrower invariants are still checked at use sites. I tightened the docs in this PR to make the current split explicit: SerializedMastForest::new() is structural inspection only, `UntrustedMastForest` is parsed-but-not-yet-trusted state, and validate() is the hard boundary that produces a trusted `MastForest`. I see where you want to go with encoding sanitized/unsanitized state in types, though I think there are better methods for this, and it feels like a follow-up refactor rather than something to fold into this PR.

core/src/mast/mod.rs
```diff
@@ -1322,32 +1379,43 @@ impl UntrustedMastForest {
...
-        forest.validate()?;
+        // Step 1: Validate over-specified wire hashes instead of silently rewriting them.
+        if !is_hashless {
+            forest.validate_node_hashes()?;
+        }
 
-        // Step 2: Validate topological ordering and recompute hashes
-        forest.validate_node_hashes()?;
+        // Step 2: Validate the recomputed forest.
+        forest.validate()?;
```
plafer, line 1413: In a future PR, I wonder if a desirable state would be: if you are ever able to construct a `MastForest`, then it is valid - which doesn't preclude offering "unchecked method" where maintaining that property is the responsibility of the caller, such as in trusted deserialization paths.

Then, we wouldn't have a (at least public) `validate()` method, since a `MastForest` would be valid by construction.

core/src/mast/serialization/mod.rs
```diff
@@ -182,134 +291,508 @@ impl MastForest {
...
+        }
+        super::DebugInfo::empty_for_nodes(layout.node_count)
+    } else {
+        super::DebugInfo::read_from(&mut recording)?
+    };
+
+    Ok((
+        flags,
+        super::UntrustedMastForest {
+            bytes: recording.into_recorded(),
```
plafer, line 789: If I'm following correctly, the root cause for needing a `TrackingReader` which records the bytes read (ultimately copying everything read in a separate buffer) is because `Deserializable::read_from_bytes(bytes: &[u8])` only gives a reference to the bytes being deserialized, instead of passing ownership.

IMO it would make more sense to change `Deserializable`'s methods to take ownership of the data being deserialized, so that zero-copy deserializers can *actually* be zero-copy. In the case where the caller needs to keep the original buffer around after deserialization, then they can clone it themselves before deserialization (with the same end-effect in that case of having 2 copies of the buffer in memory).

plafer (approved):
LGTM! Great work.

I agree that any "readability improvements" should be left for the future. The fuzzing tests also give a lot of confidence in the current implementation on their own.

Left some potential improvements for the future.

core/src/mast/mod.rs
huitseeker: I agree with that end state. Today I'm not sure we can afford to leave `validate` private yet. 

I checked the current API surface more closely. `MastForest::new()` is fine, and `MastForest::insert_procedure_name()` already enforces the root invariant itself. The two places that can still break the “valid by construction” story are `MastForest::remove_nodes()`, which has a documented orphan-node precondition, and `MastForest::debug_info_mut()`, which exposes `DebugInfo::insert_procedure_name()` / `DebugInfo::extend_procedure_names()` without the forest-level root check.

The cleaner model would be: trusted construction goes through unchecked builder-style APIs, untrusted input goes through `UntrustedMastForest::validate()`, and a constructed `MastForest` is valid by construction. That feels like a good follow-up, but I would keep it out of this PR.

core/src/mast/serialization/mod.rs
huitseeker: I agree there is a good zero-copy follow-up here.

I would keep `Deserializable::read_from()` as the generic reader-based API, and add an extension trait for buffer-backed callers, with an owned entry point like this:

```rust
use minibytes::Bytes;

pub trait DeserializableOwnedExt: Deserializable + Sized {
    fn read_from_owned_bytes(bytes: Bytes) -> Result<Self, DeserializationError> {
        let mut reader = SliceReader::new(bytes.as_ref());
        Self::read_from(&mut reader)
    }
}
```
The owned buffer could be [Bytes from Sapling’s minibytes crate](https://github.com/facebook/sapling/tree/main/eden/scm/lib/minibytes).

Then `UntrustedMastForest` could specialize that path to keep the same `Bytes` buffer for later random-access validation, instead of copying the consumed prefix back out of `TrackingReader`:

```rust
impl UntrustedMastForest {
    pub fn read_from_owned_bytes_with_budgets(
        bytes: Bytes,
        parsing_budget: usize,
        validation_budget: usize,
    ) -> Result<Self, DeserializationError> {
        let mut reader = BudgetedReader::new(SliceReader::new(bytes.as_ref()), parsing_budget);
        let mut tracking = TrackingReader::new(&mut reader);

        let (flags, layout, advice_map, debug_info, remaining_allocation_budget) =
            decode_untrusted_parts(&mut tracking, true, Some(validation_budget))?;

        let consumed_len = tracking.offset();
        log_untrusted_overspecification(flags);

        Ok(Self {
            bytes: bytes.slice(..consumed_len).into(),
            layout,
            advice_map,
            debug_info,
            remaining_allocation_budget,
        })
    }
}
```

That would remove the extra copy on owned-buffer paths without changing the generic `Deserializable` contract. I think that is better as a follow-up than something to fold into this PR.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 #### Fixes
 
 - Fixed stale `ReplayProcessor` doc comment links to `ExecutionTracer` after module-structure refactors.
+- [BREAKING] Refactored MAST forest serialization around fixed-layout full, stripped, and hashless sections, and bumped the MAST wire format to `0.0.3` ([#2765](https://github.com/0xMiden/miden-vm/pull/2765)).
```
bobbinth: This should now move into the v0.23.0 release section.

core/src/mast/mod.rs
bobbinth, line 1329: nit (and not for this PR): given that this file is already pretty big, I'd consider moving `UntrustedMastForest` struct into a separate file (e.g., `src/mast/untrusted.rs`).

core/src/mast/mod.rs
```diff
@@ -12,20 +12,35 @@
 //!     .validate()?;
 //! ```
 //!
-//! For maximum protection against denial-of-service attacks from malicious input, use
-//! [`UntrustedMastForest::read_from_bytes_with_budget`] which limits memory consumption:
+//! [`UntrustedMastForest::read_from_bytes`] applies default parsing and validation budgets derived
+//! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_budget`] to tune only the
+//! wire-parsing budget, or [`UntrustedMastForest::read_from_bytes_with_budgets`] to tune both:
```
bobbinth, line 17: It is not immediately clear what is the parsing vs. validation budgets. Maybe worth adding a sentence or two explaining what these are.

core/src/mast/mod.rs
```diff
@@ -667,20 +689,52 @@ impl MastForest {
...
-    /// This method iterates through all nodes in index order, verifying:
-    /// 1. All child references point to nodes with smaller indices (topological order)
-    /// 2. Each node's recomputed digest matches its stored digest
+    /// This addresses the gap created by PR 2094, where padding NOOPs are now inserted
+    /// at assembly time rather than dynamically during execution, and adds comprehensive
+    /// structural validation to prevent deserialization-time panics.
+    pub fn validate(&self) -> Result<(), MastForestError> {
+        self.validate_basic_block_invariants()?;
+        self.validate_procedure_name_digests()
+    }
```
bobbinth, line 712: nit: the doc comments for this method only mention basic block validation but don't mention procedure name validation. Though, since we are moving debug info out of MAST forest soon - that'll change anyways.

core/src/mast/serialization/info.rs
```diff
@@ -348,6 +286,83 @@ impl MastNodeType {
...
+
+const JOIN: u8 = 0;
+const SPLIT: u8 = 1;
+const LOOP: u8 = 2;
+const BLOCK: u8 = 3;
+const CALL: u8 = 4;
+const SYSCALL: u8 = 5;
+const DYN: u8 = 6;
+const DYNCALL: u8 = 7;
+const EXTERNAL: u8 = 8;
```
bobbinth: nit: could we move these to the top of the file (we usually put such constants into the "CONSTANTS" section).

core/src/mast/serialization/info.rs
```diff
@@ -348,6 +286,83 @@ impl MastNodeType {
...
+
 #[cfg(test)]
 mod tests {
```
bobbinth, line 382: nit: let's add "TEST" section separator above this line.

core/src/mast/serialization/info.rs
```diff
@@ -348,6 +286,83 @@ impl MastNodeType {
...
+        let entry = MastNodeEntry::read_from(source)?;
+        let digest = Word::read_from(source)?;
+        Ok(Self { entry, digest })
+    }
+
+    /// Returns the minimum serialized size: 8 bytes for `MastNodeEntry` + 32 bytes for `Word`.
+    fn min_serialized_size() -> usize {
+        MastNodeEntry::min_serialized_size() + Word::min_serialized_size()
+    }
+}
```
bobbinth, line 376: AFAICT, these implementations on `MastNodeInfo` are only used for testing purposes. If so, would be good to make this more clear (maybe by putting this code behind `#[cfg(test)]`).

Also, more broadly, it seems like `MastNodeInfo` struct is also just a test-only struct (though, I'm less sure about this) - if so, we should mark it (and all related methods) with `#[cfg(test)]`.

core/src/mast/serialization/info.rs
```diff
@@ -348,6 +286,83 @@ impl MastNodeType {
     }
 }
 
+/// Logical node metadata combining fixed-width structure and a digest value.
+///
+/// This is a convenience type for APIs that want both pieces together. The wire format does not
+/// require `MastNodeInfo` to appear as one contiguous fixed-width section.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MastNodeInfo {
```
bobbinth, line 312: nit: let's add a section separator above to indicate that this section is for a `MastNodeInfo`.

core/src/mast/serialization/layout.rs
```diff
@@ -0,0 +1,412 @@
...
+/// [`crate::mast::UntrustedMastForest`].
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct ForestLayout {
+    pub(super) node_count: usize,
+    pub(super) roots_count: usize,
+    pub(super) roots_offset: usize,
+    pub(super) basic_block_offset: usize,
+    pub(super) basic_block_len: usize,
+    pub(super) node_entry_offset: usize,
+    pub(super) node_entry_size: usize,
```
bobbinth: Why do we need `node_entry_size` field here? Isn't it always equal to 8? If we need to make this explicit, we could probably just add something like `MastNodeEntry::SERIALIZED_SIZE` constant.

core/src/mast/serialization/mod.rs
```diff
@@ -1,20 +1,60 @@
...
+//! MastForest.
+//! ```
+//!
+//! The format is:
 //!
 //! (Metadata)
 //! - MAGIC (4 bytes) + FLAGS (1 byte) + VERSION (3 bytes)
 //!
 //! (Counts)
 //! - nodes count (`usize`)
```
bobbinth, line 41: Not for this PR, but I wonder if we should store node count as two parts: `internal_node_count` and `external_node_count`. These, together with the FLAGS, would allow us to immediately compute sizes of node entries, external digests, and node hash sections.

core/src/mast/serialization/mod.rs
```diff
@@ -1,20 +1,60 @@
...
+//! - fixed-width structural node entries (`Vec<MastNodeEntry>`)
+//! - `Block` entries store offsets into the basic-block section above
+//!
+//! (External digest section)
+//! - digests for `External` nodes only (`Vec<Word>`, ordered by node index)
+//! - lookup is dense-by-kind: the Nth external node uses slot N in this section
+//!
+//! (Node hash section - omitted if FLAGS bit 1 is set)
+//! - digests for all non-external nodes (`Vec<Word>`, ordered by node index)
+//! - lookup is also dense-by-kind: the Nth non-external node uses slot N in this section
```
bobbinth, line 59: I still think that making is so that external nodes always precede internal nodes and are lexicographically sorted in the `MastForest` struct itself is a good invariant to have. But implementing this now would be a pain. Once we have something like https://github.com/0xMiden/miden-vm/issues/2473, it should become pretty easy to do though.

core/src/mast/serialization/mod.rs
```diff
@@ -75,30 +148,44 @@ type StringDataOffset = usize;
...
 type StringIndex = usize;
 
+/// Default multiplier for the untrusted validation allocation budget.
+///
+/// The budgeted byte reader limits wire-driven parsing. Hashless and stripped validation also
+/// needs transient per-node allocations for the slot table, empty debug-info scaffolding, and
+/// rebuilt digest table. The generic untrusted path also retains a recorded copy of the consumed
+/// serialized payload for deferred validation, so the default convenience path includes one extra
+/// payload-sized allowance on top of the helper-allocation headroom.
+const DEFAULT_UNTRUSTED_ALLOCATION_BUDGET_MULTIPLIER: usize = 7;
```
bobbinth, line 170: I didn't fully understand this part - e.g., does this mean that if we have 1KB serialized untrusted MAST forest, we need to allocate 7KB of RAM to deserialize it? Or does this mean something else?

core/src/mast/serialization/mod.rs
```diff
@@ -182,134 +292,499 @@ impl MastForest {
...
+    ///
+    /// let view = SerializedMastForest::new(&bytes).unwrap();
+    /// assert!(view.node_info_at(0).is_ok());
+    /// ```
+    pub fn node_info_at(&self, index: usize) -> Result<MastNodeInfo, DeserializationError> {
+        Ok(MastNodeInfo::from_entry(
+            self.node_entry_at(index)?,
+            self.node_digest_at(index)?,
+        ))
+    }
```
bobbinth, line 519: This is used for testing purposes only, right?

core/src/mast/serialization/mod.rs
```diff
@@ -182,134 +292,499 @@ impl MastForest {
...
+) -> Result<[u8; N], DeserializationError> {
+    let slice = read_slice_at(bytes, offset, N)?;
+    let mut result = [0u8; N];
+    result.copy_from_slice(slice);
+    Ok(result)
+}
 
-        Ok(mast_forest)
+#[cfg(test)]
+fn read_slice_at<'a>(
```
bobbinth, line 654: nit: testing and non-testing code is currently interspersed - would be good to move the testing code into a separate section to improve readability.

core/src/mast/serialization/layout.rs
```diff
@@ -0,0 +1,412 @@
...
+    let flags = WireFlags::new(raw_flags)?;
+    if flags.is_hashless() && !allow_hashless {
+        return Err(DeserializationError::InvalidValue(
+            "HASHLESS flag is set; use UntrustedMastForest for untrusted input".to_string(),
+        ));
+    }
+    let layout = scan_layout_sections(source, flags.is_hashless())?;
+
+    Ok((flags, layout))
+}
```
bobbinth, line 176: This would result in multiple passes over the serialized bytes, right? This is fine for this PR, but ideally, we'd both simplify this and make it more efficient. We could do this by front-loading more info into the header. Basically, the header section should be pretty small and by reading this section we should be able to compute expected size of the entire forest (not counting the debug info section which will be moved out soon anyways).

Then, we'd do a single pass over the remaining data (for both trusted and untrusted options) to build the `MastForest`

core/src/mast/serialization/layout.rs
bobbinth, line 1: nit: I'd add some section separators to this file to improve readability.

bobbinth (approved):
Looks good! Thank you! I wasn't able to review all the code in detail - but I did look through most of the non-test code and left some comments inline. There are some minor nits that could be addressed now, and some thoughts for the future.

Overall, I think once we move the debug info out of `MastForest` (which issues covers this?) and have a real builder approach (https://github.com/0xMiden/miden-vm/issues/2473), we should be able to simplify this quite a bit and also make the process more efficient. This will require imposing extra invariants on the `MastForest` struct itself (e.g., external nodes always come first, then they are followed by basic blocks).

We don't need to tackle the above immediately - but would be good to get this done before mainnet so that we don't have to change the MAST forest serialization format shortly after mainnet.

core/src/mast/mod.rs
huitseeker: Added a short note. 
- Parsing budget limits wire-count allocations while reading. 
- Validation budget limits later helper allocations for stripped/hashless checks.

core/src/mast/mod.rs
huitseeker: Updated the docs to mention procedure-name digest validation too

core/src/mast/serialization/info.rs
huitseeker: Kept `MastNodeInfo` public because `MastForestView::node_info_at()` returns it (and random access is an explicit design goal here). Marked the test-only constructor, builder conversion, and serde impls with `#[cfg(test)]`

core/src/mast/serialization/layout.rs
huitseeker: Removed the stored field and added MastNodeEntry::SERIALIZED_SIZE instead.

core/src/mast/serialization/mod.rs
huitseeker: Clarified this in the docs. It is a coarse safety bound, not a claim that 1KB always needs 7KB. Callers can set exact limits with `read_from_bytes_with_budgets()`

core/src/mast/serialization/mod.rs
huitseeker: This stays public because it is part of the random-access view API. The clearly test-only helpers around it were moved behind `#[cfg(test)]`

core/src/mast/mod.rs
huitseeker: #3043

core/src/mast/serialization/mod.rs
huitseeker: #3043

core/src/mast/serialization/mod.rs
huitseeker: I put this in #3043, but I want to note this is non-trivial. The core idea is that our wire representation is as compact as possible, using vector ordering as a way to represent a graph, including for all x-to-node relations, where x can be a node, a decorator, etc .. 

What you're suggesting explicitly breaks that. The main constraint is that external digests cannot simply be sorted independently unless each `External` node entry can identify its sorted digest slot. Otherwise, deserialization cannot know which digest belongs to which external node while preserving the in-memory `MastForest` node order.

To compensate for sorting breaking our implicit slot-positioning properties, we would have to make `MastNodeEntry::External` carry a wire digest slot:
```rust
  External { digest_slot: u32 }
```
This is the smallest way to preserve arbitrary in-memory node order while storing external digests first and sorted on disk. The slot indexes into the sorted external prefix. `try_into_mast_node_builder()` will still ignore structure beyond using the resolved digest.

This is absolutely doable, but this will have consequences on:
- branching (on `External`)
- performance (sorting) and 
- compactness (implicit -> explicit indexation)

If, as I understand it, we're simply trying to compute set differences (see #2980 and its direction), I have doubts that the juice is worth the squeeze.

core/src/mast/serialization/mod.rs
huitseeker: #3043

core/src/mast/serialization/layout.rs
huitseeker: #3043

core/src/mast/serialization/mod.rs
huitseeker: PoC at https://github.com/0xMiden/miden-vm/pull/3055 — the `External` bit is the penultimate commit, responsible for half the size of the PR.

PR #2769, opened 2026-03-03
ci: add one-shot workflow to redirect gh-pages to docs.miden.xyz

### LINKED ISSUE #2660: docs: Outdated references in mdBook documentation (hash function, library name, version)
## Description

The mdBook documentation at https://0xmiden.github.io/miden-vm/ contains several outdated references that don't match the current `next` branch README:

### 1. Hash Function Name
- **Docs say:** "Rescue Prime Optimized"
- **Should be:** "Poseidon2"
- **Location:** `docs/intro/main.md` (Cryptographic operations section)

### 2. Library Name
- **Docs say:** `stdlib`
- **Should be:** `miden-core-lib`
- **Location:** `docs/intro/main.md` (External libraries section)

### 3. Version Number
- **Docs say:** "v0.19"
- **Should be:** Current release version
- **Location:** `docs/intro/main.md` (Status and features section)

## Proposed Fix

I'd be happy to submit a PR to update these references if that would be helpful.

Keinberger:
Closes #2660

The old mdBook site at `0xmiden.github.io/miden-vm` is still serving stale content from Oct 2025. Deleting the `gh-pages` branch and removing the `book.yml` workflow was not enough, GitHub Pages continues serving previously deployed content.

This adds a `workflow_dispatch` workflow that deploys a redirect page via
`actions/deploy-pages` without committing any files to the repo. After merging, the
workflow needs to be triggered once manually from the Actions tab. The Pages source also needs to be set to "GitHub Actions" in repo settings.

Deep links are preserved (e.g. `/miden-vm/intro/main.html` => `docs.miden.xyz/miden-vm/intro/main.html`). The workflow file can be removed after the one-time deployment.


Keinberger:
Hey @huitseeker, that's a fair point about keeping gh-pages available for rustdoc and benchmarks!

I think the redirect and future gh-pages usage aren't mutually exclusive though. The redirect is a one-shot deployment that just replaces the current stale mdBook content. If someone later wants to publish rustdoc or benchmarks to gh-pages (similar to what https://github.com/0xMiden/crypto/pull/810 does), they can deploy new content that overwrites the redirect page.

The immediate problem is that the old mdBook site is still live and serving content from Oct 2025, which is confusing users who find it via search engines (the original report in this issue). I think getting the redirect in place now solves that, and it doesn't block any future gh-pages use cases.

Would that work for you, or do you think we should set up the rustdoc/benchmarks publishing alongside this PR?

huitseeker:
Overall, I plan for both the docs and benches to be accessible in sub-directories of the main URL, e.g. https://0xmiden.github.io/miden-vm/docs/. 

One convincing way of testing whether this redirect works without clobbering the subdirectory access would be to set up a single simple landing page in such a subdirectory (e.g. a static `docs/index.html` saying 'placeholder'). If that works, I'm on board.

Keinberger:
Hey @huitseeker, makes sense!

I updated the PR to:
1. Keep the root `index.html` redirect (handles `0xmiden.github.io/miden-vm/`)
2. Dropped the catch-all `404.html` so subdirectories are completely unaffected
3. Added a placeholder `docs/index.html` so we can verify subdirectory access works after deployment

Without a custom `404.html`, GitHub Pages just shows its default 404 for non-existent paths. That means any content later deployed to `/docs/`, `/benchmarks/`, etc. will be served directly without interference from the redirect

Let me know if that looks good

huitseeker (approved):
Amazing, thanks!

PR #2771, opened 2026-03-03
Operation now only encodes basic block operations

### LINKED ISSUE #2768: Make `Operation` only encode basic block operations
Currently, the `Operation` enum has variant for each possible opcode - notably including `MastNode`-related ones (e.g. `JOIN`, `END`, etc).

We should instead make `Operation` *only* encode operations that can occur inside a `BasicBlockNode`. As it currently stands, it is possible to create a `BasicBlockNode` which contains a `JOIN` opcode, which shouldn't be allowed. As found by the auditors, this can be exploited today by serializing a `BasicBlockNode` which contains e.g. a `JOIN` opcode, which will make the processor crash.

As a bonus, this would cleanup the [`unreachable!()` statements](https://github.com/0xMiden/miden-vm/blob/b99689b8047f8d6315338c36c18fc55ef32f833b/processor/src/execution/operations/mod.rs#L69-L80) that we have in `execute_op()`.

plafer:
Closes #2768 

Cleans up the `Operation` enum, and at the same time addresses issue 44 from the audit, where an attacker can crash the processor by e.g. encoding a `JOIN` operation inside a basic block (which we don't explicitly look for), and then at execution-time would crash the processor [here](https://github.com/0xMiden/miden-vm/blob/b99689b8047f8d6315338c36c18fc55ef32f833b/processor/src/execution/operations/mod.rs#L69). Now, we simply don't represent `Join` (and other control flow operaitons) in the `Operation` enum, where deserialization would fail [in the call to](https://github.com/0xMiden/miden-vm/blob/a8467a94177625b6b55fa2ac74ad19b6d453d76b/core/src/operations/mod.rs#L1001-L1003) `Operation::read_from()`.

Note: I opened this against `next` since adapting this PR to the `audit-fixes-2026` branch required too many fixes.

### DIFF
--- a/air/src/constraints/op_flags/mod.rs
+++ b/air/src/constraints/op_flags/mod.rs
@@ -21,13 +21,7 @@
 
 use core::marker::PhantomData;
 
-use miden_core::{
-    field::PrimeCharacteristicRing,
-    operations::{
-        OPCODE_CALL, OPCODE_DYN, OPCODE_DYNCALL, OPCODE_END, OPCODE_HALT, OPCODE_JOIN, OPCODE_LOOP,
-        OPCODE_REPEAT, OPCODE_RESPAN, OPCODE_SPAN, OPCODE_SPLIT, OPCODE_SYSCALL, Operation,
-    },
-};
+use miden_core::{field::PrimeCharacteristicRing, operations::opcodes};
 
 #[cfg(test)]
 use crate::trace::decoder::NUM_OP_BITS;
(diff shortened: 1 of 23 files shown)


core/src/operations/mod.rs
bobbinth: nit (and maybe it's too much work to be worth it), but maybe we should make this module public and rename it to just `opcodes`, then, we can remove the ``OPCODE_` prefix from internal constants. It could look something like:

```rust
pub mod opcodes {
    pub const NOOP: u8           = 0b0000_0000;
    pub const EQZ: u8            = 0b0000_0001;
    ...
}
```

This makes the constants a bit more encapsulated, and using them would also be a bit simpler:

```rust
pub enum Operation {
    Noop = opcodes::NOOP,
    ...
}
```
And in other places, instead of importing individual constants, we could just import the `opcodes` module.

processor/src/trace/parallel/tracer/trace_row.rs
```diff
@@ -69,13 +72,13 @@ impl DecoderRow {
     ///
     /// This corresponds either to the SPAN or RESPAN operations.
     pub fn new_basic_block_batch(
-        operation: Operation,
+        opcode: u8,
```
bobbinth: I wonder if there is some way to make things a bit more type-safe here. Maybe instead of passing a `u8` we could pass in a boolean indicating whether we are doing a `SPAN` or a `RESPAN`? (e.g., something like `is_respan`).

processor/src/trace/parallel/tracer/trace_row.rs
```diff
@@ -16,7 +16,10 @@ use miden_core::{
     mast::{
         BasicBlockNode, CallNode, JoinNode, LoopNode, MastForest, MastNodeExt, OpBatch, SplitNode,
     },
-    operations::Operation,
+    operations::{
+        OPCODE_CALL, OPCODE_DYN, OPCODE_DYNCALL, OPCODE_END, OPCODE_JOIN, OPCODE_LOOP,
+        OPCODE_REPEAT, OPCODE_RESPAN, OPCODE_SPAN, OPCODE_SPLIT, OPCODE_SYSCALL, Operation,
+    },
```
bobbinth: Related to one of the previous comments: this could be `operations::opcodes`.

air/src/constraints/op_flags/mod.rs
```diff
@@ -635,208 +641,208 @@ where
     op_flag_getters!(degree7_op_flags,
         /// Operation Flag of NOOP operation.
         #[allow(dead_code)]
-        noop => Operation::Noop,
+        noop => Operation::Noop.op_code(),
         /// Operation Flag of EQZ operation.
-        eqz => Operation::Eqz,
+        eqz => Operation::Eqz.op_code(),
```
bobbinth: Maybe all of these could be `opcodes::NOOP`, `opcodes::EQZ`, etc.

bobbinth (approved):
Looks good! Thank you! I left a few comments inline.

core/src/operations/mod.rs
```diff
@@ -135,7 +135,12 @@ mod opcode_constants {
 // OPERATIONS
 // ================================================================================================
 
-/// A set of native VM operations which take exactly one cycle to execute.
+/// The set of native VM basic block operations executable which take exactly one cycle to execute.
```
huitseeker, line 137: Could we add a small negative test that `Operation::read_from()` rejects control-flow opcodes like `OPCODE_SPAN`/`OPCODE_END`? That would lock in the new contract and catch regressions if these ever get reintroduced.

huitseeker (approved):
LGTM, but left a test-only PR targeting this branch here: https://github.com/0xMiden/miden-vm/pull/2774
(the changelog failure isn't relevant)

core/src/operations/mod.rs
plafer: IIUC, this was covered in #2774

core/src/operations/mod.rs
plafer: Done

processor/src/trace/parallel/tracer/trace_row.rs
plafer: Done; introduced an enum (to avoid "boolean blindness")

processor/src/trace/parallel/tracer/trace_row.rs
plafer: Done

air/src/constraints/op_flags/mod.rs
plafer: Done

core/src/operations/mod.rs
huitseeker: yup

huitseeker:
Note: this should be merged into https://github.com/0xMiden/miden-vm/tree/audit-fixes-2026-speculatively-next

PR #2772, opened 2026-03-04
Complete set of constraints

Al-Kindi-0:
## Describe your changes

This contains all the reviewed PRs for the chunked constraints

- Updated `CHANGELOG.md'

### DIFF
--- a/air/src/constraints/chiplets/bus/chiplets.rs
+++ b/air/src/constraints/chiplets/bus/chiplets.rs
@@ -32,14 +32,7 @@
 //! - Air-script: ~/air-script/constraints/chiplets.air
 //! - Processor: processor/src/chiplets/aux_trace/bus/
 
-use miden_core::{
-    FMP_ADDR, FMP_INIT_VALUE,
-    field::PrimeCharacteristicRing,
-    operations::{
-        OPCODE_CALL, OPCODE_DYN, OPCODE_DYNCALL, OPCODE_JOIN, OPCODE_LOOP, OPCODE_SPLIT,
-        OPCODE_SYSCALL,
-    },
-};
+use miden_core::{FMP_ADDR, FMP_INIT_VALUE, field::PrimeCharacteristicRing, operations::opcodes};
 use miden_crypto::stark::{air::MidenAirBuilder, matrix::Matrix};
 
 use crate::{
@@ -1338,11 +1331,11 @@ impl ControlBlockOp {
     /// Returns the opcode value for this control block operation.
     fn opcode(&self) -> u8 {
         match self {
-            ControlBlockOp::Join => OPCODE_JOIN,
-            ControlBlockOp::Split => OPCODE_SPLIT,
-            ControlBlockOp::Loop => OPCODE_LOOP,
-            ControlBlockOp::Call => OPCODE_CALL,
-            ControlBlockOp::Syscall => OPCODE_SYSCALL,
+            ControlBlockOp::Join => opcodes::JOIN,
+            ControlBlockOp::Split => opcodes::SPLIT,
+            ControlBlockOp::Loop => opcodes::LOOP,
+            ControlBlockOp::Call => opcodes::CALL,
+            ControlBlockOp::Syscall => opcodes::SYSCALL,
         }
     }
 }
@@ -1429,7 +1422,7 @@ fn compute_dyn_request<AB: MidenAirBuilder<F = Felt>>(
 ) -> AB::ExprEF {
     // Control block request with zeros for hasher state (callee is dynamic)
     let control_req =
-        compute_control_block_request_zeros::<AB>(local, next, challenges, OPCODE_DYN);
+        compute_control_block_request_zeros::<AB>(local, next, challenges, opcodes::DYN);
 
     // Memory read for callee hash (word read from stack[0] address)
     let callee_hash_req = compute_dyn_callee_hash_read::<AB>(local, challenges);
@@ -1450,7 +1443,7 @@ fn compute_dyncall_request<AB: MidenAirBuilder<F = Felt>>(
 ) -> AB::ExprEF {
     // Control block request with zeros for hasher state (callee is dynamic)
     let control_req =
-        compute_control_block_request_zeros::<AB>(local, next, challenges, OPCODE_DYNCALL);
+        compute_control_block_request_zeros::<AB>(local, next, challenges, opcodes::DYNCALL);
 
     // Memory read for callee hash (word read from stack[0] address)
     let callee_hash_req = compute_dyn_callee_hash_read::<AB>(local, challenges);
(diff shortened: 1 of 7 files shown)


air/src/constraints/public_inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+/// - First row: `stack[i] == stack_inputs[i]` for i in 0..16
+/// - Last row:  `stack[i] == stack_outputs[i]` for i in 0..16
+pub fn enforce_main<AB>(builder: &mut AB, local: &MainTraceRow<AB::Var>)
+where
+    AB: MidenAirBuilder,
+{
+    // Copy public values into local arrays to release the immutable borrow on builder.
+    let pv = builder.public_values();
+    let n = pv.len();
+    debug_assert!(n >= TAIL_LEN, "public values too short: {n} < {TAIL_LEN}");
```
huitseeker: I might be missing something, but `debug_assert!(n >= TAIL_LEN)` is stripped in release. If `public_values()` is shorter, the indexing below will panic. Should we add a real length check and return a structured error instead? This looks like a verifier DoS risk if proofs are untrusted.

air/src/trace/mod.rs
```diff
@@ -183,6 +183,56 @@ pub const ACE_CHIPLET_WIRING_BUS_RANGE: Range<usize> =
 /// Auxiliary trace segment width.
 pub const AUX_TRACE_WIDTH: usize = ACE_CHIPLET_WIRING_BUS_RANGE.end;
 
-/// Number of random elements available to the prover after the commitment to the main trace
-/// segment.
-pub const AUX_TRACE_RAND_ELEMENTS: usize = 16;
+/// Number of random challenges used for auxiliary trace constraints.
+pub const AUX_TRACE_RAND_CHALLENGES: usize = 2;
```
huitseeker, line 187: Now that `AUX_TRACE_RAND_CHALLENGES` is 2, do we also enforce the expected randomness count in proof parsing or the prover/verifier handshake? I worry about silent mismatches if any component still assumes the old 16-element layout.

Cargo.toml
```diff
@@ -95,3 +95,9 @@ getrandom = { version = "0.2", default-features = false, features = ["js"] }
 
 [workspace.lints.rust]
 unexpected_cfgs = { check-cfg = ['cfg(fuzzing)'], level = "warn" }
+
+[patch.crates-io]
+p3-miden-air = { git = "https://github.com/0xMiden/p3-miden", branch = "main" }
```
huitseeker, line 101: Note : this means the present PR will only be mergeable post p3-miden release. This also means that our attempts at masking (through public re-exports) the forked crates that we eventually want to upstream changes to failed (or at least needs to be cleaned up post switch to lifted-stark). /cc @adr1anh

air/src/constraints/tagging/ood_eval.rs
```diff
@@ -59,40 +57,29 @@ impl OodEvalAirBuilder {
...
-            permutation_randomness.push(QuadFelt::ONE);
-        }
-        if crate::trace::AUX_TRACE_RAND_ELEMENTS > 2 {
-            let mut beta_power = beta;
-            for _ in 2..crate::trace::AUX_TRACE_RAND_ELEMENTS {
-                permutation_randomness.push(beta_power);
-                beta_power *= beta;
-            }
-        }
+        // Generate MAX_MESSAGE_WIDTH QuadFelts to keep the RNG state stable (only
```
huitseeker: We now generate `MAX_MESSAGE_WIDTH` randomness here even though only `AUX_TRACE_RAND_CHALLENGES = 2` are used. That keeps fixtures stable, but could it also hide accidental reads past index 1? Should we either generate only 2 in tests or add an explicit guard that constraints only use the first two?

Cargo.toml
Al-Kindi-0: But would this cause any issues if it is merged into `next` ? The idea is this will be just transitory until 2739 is merged pre-release.

huitseeker: Yes: any repo depending on this will not carry the patch transitively.

air/src/constraints/public_inputs.rs
Al-Kindi-0: Fixed, but this should be improved/refactored during the migration with proper and solid handling for public inputs

air/src/trace/mod.rs
Al-Kindi-0: This should improve after the migration as well

air/src/constraints/tagging/ood_eval.rs
Al-Kindi-0: Fixed it now and clarified further the comment

air/src/constraints/public_inputs.rs
```diff
@@ -0,0 +1,72 @@
...
+/// - First row: `stack[i] == stack_inputs[i]` for i in 0..16
+/// - Last row:  `stack[i] == stack_outputs[i]` for i in 0..16
+pub fn enforce_main<AB>(builder: &mut AB, local: &MainTraceRow<AB::Var>)
+where
+    AB: MidenAirBuilder,
+{
+    // Copy public values into local arrays to release the immutable borrow on builder.
+    let pv = builder.public_values();
+    let n = pv.len();
+    assert!(n >= TAIL_LEN, "public values too short: {n} < {TAIL_LEN}");
```
huitseeker, line 43: `assert!(n >= TAIL_LEN, ...)` will panic on malformed proofs. 

If public values can be untrusted, should this be a structured error (or validated earlier) to avoid a verifier DoS?

air/src/constraints/chiplets/bus/wiring.rs
```diff
@@ -0,0 +1,227 @@
...
+        wire_1.clone() * wire_2.clone() * m0.clone() + wire_0.clone() * wire_2.clone() * m1;
+    let eval_terms = wire_1.clone() * wire_2.clone() * m0
+        - wire_0.clone() * wire_2.clone()
+        - wire_0.clone() * wire_1.clone();
+
+    // Gates: non-ACE rows must contribute zero; READ/EVAL are mutually exclusive.
+    let read_gate = ace_flag.clone() * is_read;
+    let eval_gate = ace_flag * is_eval;
+
+    let common_den = wire_0.clone() * wire_1.clone() * wire_2.clone();
```
huitseeker, line 166: The LogUp constraint uses a shared denominator `wire_0 * wire_1 * wire_2`. On READ rows `wire_2` isn’t part of the message; if its encoding can be zero, the constraint becomes vacuous. 

Should READ and EVAL use separate denominators, or constrain `wire_2 != 0` / fixed non-zero on READ?

air/src/constraints/public_inputs.rs
Al-Kindi-0: This will be removed and fixed a bit upper in the stack during the lifted STARK migration work

air/src/constraints/chiplets/bus/wiring.rs
Al-Kindi-0: Yes, there is a completeness loss common to all multi-set/logup arguments, i.e., the terms are non-zero with high probability over the choice of aux randomness

PR #2777, opened 2026-03-04
Harden untrusted deserialization: enforce budgets/depth limits and expand fuzz/tests

huitseeker:
Hardened untrusted deserialization across core formats by enforcing budgets and depth limits, adding bounds checks to prevent panic/abort paths from attacker‑controlled sizes, and expanding fuzzing plus targeted regression tests for library/package, proof, and MAST/debug-info decoding.

### DIFF
--- a/crates/assembly-syntax/src/library/mod.rs
+++ b/crates/assembly-syntax/src/library/mod.rs
@@ -972,6 +972,8 @@ impl Serializable for TypeSerializer<'_> {
 #[derive(Debug)]
 pub struct TypeDeserializer(pub Type);
 
+// Bounds recursive type nesting during deserialization to prevent adversarially deep types from
+// exhausting stack or budgets; 256 is far beyond realistic type depth while keeping parsing safe.
 const MAX_TYPE_NESTING: usize = 256;
 
 impl TypeDeserializer {
--- a/crates/mast-package/src/debug_info/serialization.rs
+++ b/crates/mast-package/src/debug_info/serialization.rs
@@ -25,7 +25,7 @@ impl Serializable for DebugTypesSection {
         // Write string table
         target.write_usize(self.strings.len());
         for s in &self.strings {
-            write_string(target, s);
+            s.as_ref().write_into(target);
         }
 
         // Write type table
@@ -77,7 +77,7 @@ impl Serializable for DebugSourcesSection {
         // Write string table
         target.write_usize(self.strings.len());
         for s in &self.strings {
-            write_string(target, s);
+            s.as_ref().write_into(target);
         }
 
         // Write file table
@@ -129,7 +129,7 @@ impl Serializable for DebugFunctionsSection {
         // Write string table
         target.write_usize(self.strings.len());
         for s in &self.strings {
-            write_string(target, s);
+            s.as_ref().write_into(target);
         }
 
         // Write function table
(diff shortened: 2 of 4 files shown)


crates/mast-package/src/package/serialization.rs
```diff
@@ -139,11 +139,8 @@ impl Deserializable for Package {
 
         // Read custom sections
         let num_sections = source.read_usize()?;
-        let mut sections = Vec::with_capacity(num_sections);
-        for _ in 0..num_sections {
-            let section = Section::read_from(source)?;
-            sections.push(section);
-        }
+        let sections: Vec<Section> =
+            source.read_many_iter(num_sections)?.collect::<Result<_, _>>()?;
```
bobbinth, line 143: Could this not be just:
```rust
let sections  = Vec::<Section>::read_from(source)?;
```

crates/mast-package/src/package/serialization.rs
```diff
@@ -328,10 +325,8 @@ impl Deserializable for PackageManifest {
 
         // Read dependencies
         let deps_len = source.read_usize()?;
-        let mut dependencies = Vec::with_capacity(deps_len);
-        for _ in 0..deps_len {
-            dependencies.push(Dependency::read_from(source)?);
-        }
+        let dependencies: Vec<Dependency> =
+            source.read_many_iter(deps_len)?.collect::<Result<_, _>>()?;
```
bobbinth: Could this not be just:
```rust
let dependencies = Vec::<Dependency>::read_from(source)?;
```

crates/mast-package/src/debug_info/serialization.rs
bobbinth: Would implementing `Serializable` and `Deserializable` on `DebugTypeIdx` simplify this code? For example, it could be just:
```rust
let param_type_indices = Vec::<DebugTypeIdx>::read_from(source)?;
```

bobbinth: I think this method is redundant as `Serializable` is implemented for `str` - and so, we could use "native" serialization.

bobbinth, line 491: Not for this PR, but we should be able to implement `Serializable` and `Deserializable` for `Arc<str>` and then get rid of this method and the complications it causes.

crates/assembly-syntax/src/library/mod.rs
```diff
@@ -945,16 +969,30 @@ impl Serializable for TypeSerializer<'_> {
 /// This is a temporary implementation to allow type information to be serialized with libraries,
 /// but in a future release we'll either rely on the `serde` serialization for these types, or
 /// provide the serialization implementation in midenc-hir-type instead
+#[derive(Debug)]
 pub struct TypeDeserializer(pub Type);
 
-impl Deserializable for TypeDeserializer {
-    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+const MAX_TYPE_NESTING: usize = 256;
```
bobbinth, line 977: Could we add a comment explaining what this is used for and why 256 is a reasonable value?

bobbinth (approved):
Looks good! Thank you! I left some comments inline - mostly nits and some that can be addressed in the future.

crates/mast-package/src/debug_info/serialization.rs
huitseeker: https://github.com/0xMiden/crypto/issues/885

crates/mast-package/src/package/serialization.rs
bobbinth: Seems like this comment wasn't addressed - or does what I proposed not work for some reason?

huitseeker: fix bundled in https://github.com/0xMiden/miden-vm/pull/2802

PR #2779, opened 2026-03-05
Fix push-immediate commitment: reject malformed indptr and validate group bindings

huitseeker:
- Enforce immediate-slot correctness during basic-block decode: reject overlaps and overflow instead of silently dropping.
- Add validation that opcode groups and push immediates are committed to the correct group slots, and that non‑immediate empty groups are zero.
- Add regression tests for malformed indptr and immediate/overlap mismatch cases.

### DIFF
--- a/core/src/mast/serialization/tests.rs
+++ b/core/src/mast/serialization/tests.rs
@@ -1528,6 +1528,9 @@ fn test_untrusted_forest_detects_hash_mismatch() {
     assert_matches!(result, Err(MastForestError::HashMismatch { .. }));
 }
 
+// UNTRUSTED VALIDATION TEST HELPERS
+// --------------------------------------------------------------------------------------------
+
 /// Build a packed operation group from op codes.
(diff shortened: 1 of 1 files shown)


core/src/mast/serialization/basic_blocks.rs
```diff
@@ -238,11 +238,21 @@ impl BasicBlockDataDecoder<'_> {
...
+                                return Err(DeserializationError::InvalidValue(String::from(
+                                    "push immediate exceeds group slots",
+                                )));
+                            }
+                            if indptr[next_group_idx] != indptr[next_group_idx + 1] {
+                                return Err(DeserializationError::InvalidValue(format!(
+                                    "push immediate overlaps operation group at index {}",
+                                    next_group_idx
+                                )));
+                            }
```
plafer: This logic is duplicated from `BasicBlockNode::validate_immediate_commitment()` right? Didn't look in detail, but if this can be done simply, would be good to only write that logic once

PR #2782, opened 2026-03-05
Validate op batch groups to prevent trace panic

huitseeker:
- Enforce the invariant that all non-final op batches are full and only the final batch may be a smaller power‑of‑two, preventing trace generation panics.
- Add validation/tests for untrusted MAST forests to reject non‑full prefix batches and accept full‑prefix + power‑of‑two tail batches.
- Refactor serialization tests to share build_group/make_batch helpers and clean up helper documentation.


CergyK:
Fix looks good

PR #2783, opened 2026-03-05
Tracking migration work for lifted STARK

Al-Kindi-0:
## Describe your changes

This includes migration work for `miden-crypto` and `p3-miden` in order to migrate to lifted STARK backend

- Updated `CHANGELOG.md'

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,138 @@
+[alias]
+# Collection of project wide clippy lints. This is done via an alias because
+# clippy doesn't currently allow for specifiying project-wide lints in a
+# configuration file. This is a similar workaround to the ones presented here:
+# <https://github.com/EmbarkStudios/rust-ecosystem/issues/59>
+#
+# NOTE: Only lints that the codebase currently passes are enabled.
+# Additional lints can be enabled incrementally as the codebase is cleaned up.
+# NOTE: for the `make lint` command to work well, this must contain the same list of lints as the xclippy-fix command below
+xclippy = [
+    "clippy", "--workspace", "--all-targets", "--all-features", "--",
+    "-Wclippy::all",
+    # Lints that currently pass in the codebase:
+    "-Wclippy::await_holding_lock",
+    "-Wclippy::char_lit_as_u8",
+    "-Wclippy::dbg_macro",
+    "-Wclippy::disallowed_methods",
+    "-Wclippy::empty_enums",
+    "-Wclippy::exit",
+    "-Wclippy::filter_map_next",
+    "-Wclippy::flat_map_option",
+    "-Wclippy::float_cmp_const",
+    "-Wclippy::if_let_mutex",
+    "-Wclippy::imprecise_flops",
+    "-Wclippy::invalid_upcast_comparisons",
+    "-Wclippy::large_digit_groups",
+    "-Wclippy::large_stack_arrays",
+    "-Wclippy::large_types_passed_by_value",
+    "-Wclippy::let_unit_value",
+    "-Wclippy::linkedlist",
+    "-Wclippy::lossy_float_literal",
+    "-Wclippy::macro_use_imports",
+    "-Wclippy::manual_ok_or",
+    "-Wclippy::map_flatten",
+    "-Wclippy::mem_forget",
+    "-Wclippy::missing_enforced_import_renames",
+    "-Wclippy::mut_mut",
+    "-Wclippy::mutex_integer",
+    "-Wclippy::needless_borrow",
+    "-Wclippy::path_buf_push_overwrite",
+    "-Wclippy::rc_mutex",
+    "-Wclippy::ref_option_ref",
+    "-Wclippy::same_functions_in_if_condition",
+    "-Wclippy::string_add",
+    "-Wclippy::string_add_assign",
+    "-Wclippy::trait_duplication_in_bounds",
+    "-Wclippy::useless_transmute",
+    "-Wclippy::verbose_file_reads",
+    "-Wclippy::zero_sized_map_values",
+    "-Wfuture_incompatible",
+    "-Wnonstandard_style",
+    "-Wunexpected_cfgs",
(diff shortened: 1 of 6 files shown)


processor/src/trace/parallel/tests.rs
```diff
@@ -381,11 +381,13 @@ fn test_trace_generation_at_fragment_boundaries(
         }
     }
 
-    // Sanity check to ensure that the traces are identical.
-    assert_eq!(format!("{trace_from_fragments:?}"), format!("{trace_from_single_fragment:?}"));
+    // Verify stack outputs match.
```
huitseeker, line 388: This now checks only stack outputs. The old debug equality also covered aux trace columns, merkle store data, program info, and trace_len_summary. Could we compare `DeterministicTrace(&trace_from_fragments)` vs `DeterministicTrace(&trace_from_single_fragment)` (or add explicit checks) so we keep that coverage but stay deterministic?

processor/src/trace/range/mod.rs
```diff
@@ -144,6 +147,10 @@ impl RangeChecker {
         // the "current" row of the main trace but placed into the "next" row of the bus column.)
         write_trace_row(&mut trace, &mut i, 0, (u16::MAX).into());
 
+        // all elements are now initialized
+        let [t0, t1] = trace;
+        let trace = unsafe { [assume_init_vec(t0), assume_init_vec(t1)] };
```
huitseeker, line 154: Could we add a quick guard before `assume_init_vec` to prove `i == target_len`? Right now this path assumes every slot is initialized, and a future caller passing a bad `trace_len` could accidentally turn that into UB.

huitseeker, line 154: I added a small test locally to show the unsafe assumption can read uninitialized memory, and Miri reports UB when this runs.

```rust
#[test]
fn into_trace_with_table_uninitialized_tail_is_ub() {
    let checker = RangeChecker::new();
    let table_len = checker.get_number_range_checker_rows();
    let target_len = (table_len.next_power_of_two()).saturating_mul(2);

    // Pass an inconsistent table length on purpose. This leaves at least one element
    // uninitialized while the function still calls `assume_init_vec`.
    let RangeCheckTrace { trace, aux_builder: _ } =
        checker.into_trace_with_table(table_len + 1, target_len);

    // Force reads across all rows so Miri can report the uninitialized read.
    let _sum: u64 = trace.iter().flatten().map(|v| v.as_canonical_u64()).sum();
}
```

Command I ran:
`cargo +nightly miri test -p miden-processor trace::range::tests::into_trace_with_table_uninitialized_tail_is_ub -- --nocapture`

Miri output includes:
`Undefined Behavior: reading memory ... but memory is uninitialized`

processor/src/trace/parallel/tests.rs
Al-Kindi-0: Added in the latest commit

processor/src/trace/range/mod.rs
Al-Kindi-0: Added

prover/src/lib.rs
```diff
@@ -153,10 +160,48 @@ pub fn prove_sync(
...
+) -> Result<Vec<u8>, ExecutionError>
+where
+    SC: StarkConfig<Felt, QuadFelt>,
+    <SC::Lmcs as Lmcs>::Commitment: Serialize,
+{
+    let log_trace_height = trace.height().ilog2() as u8;
+
+    let mut challenger = config.challenger();
+    challenger.observe_slice(public_values);
+    // TODO: observe log_trace_height in the transcript for Fiat-Shamir binding.
```
huitseeker, line 185: We should link to #2822

bobbinth:
`miden-crypto` v0.23.0 (with `p3-miden` v0.5.0`) release has been published. So, we should probably update this PR to use crates.io versions for relevant dependencies.

Al-Kindi-0:
> `miden-crypto` v0.23.0 (with `p3-miden` v0.5.0`) release has been published. So, we should probably update this PR to use crates.io versions for relevant dependencies.

The `miden-crypto` dependency is now updated

prover/src/lib.rs
adr1anh: There's a link a few lines later

crates/ace-codegen/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "miden-ace-codegen"
+version = "0.21.0"
+edition = "2024"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/0xMiden/miden-vm"
```
bobbinth, line 6: Could we add all the standard fields we usually have for crates (description, contributors, readme etc.)

crates/lib/core/Cargo.toml
```diff
@@ -42,6 +42,7 @@ thiserror.workspace = true
 
 [dev-dependencies]
 criterion.workspace = true
+miden-ace-codegen = { path = "../../ace-codegen" }
```
bobbinth, line 45: Can this not be `miden-ace-codegen.workspace = true`?

processor/src/trace/row_major_adapter.rs
bobbinth, line 1: Do we no longer need the row major adapter? Or did this move somewhere else?

air/src/config.rs
bobbinth, line 1: It would be great to add target security levels for each set of config parameters.

processor/src/trace/mod.rs
```diff
@@ -105,15 +105,19 @@ impl ExecutionTrace {
...
             self.final_pc_transcript.state(),
-        );
-        public_inputs.to_elements()
+        )
+    }
+
+    /// Returns the public values for this execution trace.
+    pub fn to_public_values(&self) -> Vec<Felt> {
+        self.public_inputs().to_elements()
     }
```
bobbinth, line 121: It's not very clear to me that this method adds a lot of value as we can always do `trace.public_inputs().to_elements()`.

PR #2786, opened 2026-03-05
Fix cryptostream bus messages

plafer:
Fixes audit finding 52,

> crypto_stream emits four word-level memory chiplet rows, but [BusColumnBuilder::get_requests_at](https://github.com/0xMiden/miden-vm/blob/089dc6a33cbb77dbbbaa7e424fd7cecd0fe7a9ca/processor/src/trace/chiplets/aux_trace/bus/mod.rs#L74-L145) has no OPCODE_CRYPTOSTREAM case and therefore emits no matching chiplets-bus requests.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -30,6 +30,15 @@ jobs:
           - mast_node_info
           - basic_block_data
           - debug_info
+          - program_deserialize
+          - kernel_deserialize
+          - stack_io_deserialize
+          - advice_inputs_deserialize
+          - operation_deserialize
+          - execution_proof_deserialize
+          - precompile_request_deserialize
+          - library_deserialize
+          - package_deserialize
     timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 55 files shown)


processor/src/trace/tests/chiplets/memory.rs
```diff
@@ -201,6 +201,46 @@ fn b_chip_trace_mem() {
...
+        Felt::new(17),
+    ];
+
+    // Aux trace construction should succeed now that OPCODE_CRYPTOSTREAM emits
+    // the four matching memory chiplet bus requests.
+    let aux_columns = trace.build_aux_trace(&alphas).unwrap();
+    let b_chip = aux_columns.get_column(CHIPLETS_BUS_AUX_TRACE_OFFSET);
+
+    // The chiplets bus should be balanced: final value must be ONE.
+    assert_eq!(*b_chip.last().unwrap(), ONE);
```
huitseeker, line 292: This test only checks that the chiplets bus is balanced. Could we also assert the exact bus requests (addresses, labels, and word ordering) for `crypto_stream`? That would catch swapped words or wrong `src_ptr`/`dst_ptr` while still being a small unit test.

huitseeker (approved):
LGTM, with a suggestion to add a stronger assertion for request contents.

processor/src/trace/tests/chiplets/memory.rs
plafer: Done

adr1anh:
Might be worth targeting either the constraint or migration PR to make sure it's clear that this is fixed in the next release.

plafer:
> Might be worth targeting either the constraint or migration PR to make sure it's clear that this is fixed in the next release.

This statement would apply to basically to all PRs against [audit-fixes-2026-speculatively-next](https://github.com/0xMiden/miden-vm/tree/audit-fixes-2026-speculatively-next), and the agreement with the auditors is that we merge all fixes against this branch. We'll just have to make sure to merge this back in `next` before releasing 0.22.

PR #2789, opened 2026-03-06
Prevent AEAD decrypt size overflows and add coverage for overflow paths

huitseeker:
- Add checked size calculations in the AEAD decrypt handler to avoid overflow and truncation.
- Factor size math into a helper and add unit tests for overflow cases.
- Include a 32-bit–only test to cover usize conversion overflow.


CergyK:
Fix LGTM

PR #2790, opened 2026-03-06
fix: replace unwrap() with ? in recursive calls

### LINKED ISSUE #2785: unwrap() instead of ? in recursive calls in decorator_storage and debug_var_storage
### Packages versions

next branch

### Bug description

    .unwrap();

    .unwrap();

_No response_

### Relevant log output

```shell

```

giwaov:
## Summary

Replace \.unwrap()\ with \?\ in two locations for proper error propagation.

## Changes

- \decorator_storage.rs\: Replace \.unwrap()\ with \?\ in \dd_decorator_info_for_node\ recursive call (line 403)
- \debug_var_storage.rs\: Replace \.unwrap()\ with \?\ in \dd_debug_var_info_for_node\ recursive call (line 335)

Both functions return \Result<(), DecoratorIndexError>\ and use \?\ throughout for error propagation, but these two recursive calls were the only places using \.unwrap()\ instead of propagating errors to the caller.

Closes #2785


bobbinth:
Duplicate of https://github.com/0xMiden/miden-vm/pull/2843.

PR #2791, opened 2026-03-06
feat(air): add optional serde feature

### LINKED ISSUE #2744: `miden-air` unconditionally enables serde on `miden-core`
(Written by claude code)

miden-air v0.21.1 depends on miden-core with features = ["serde"] and default-features = false. This means any crate that transitively depends on miden-air (via miden-processor,
miden-prover, or miden-verifier) always activates serde on miden-core, which cascades to miden-crypto/serde as well.

This is a Cargo feature unification issue — even consumers that don't need serialization pay the compile-time cost of serde derives on all miden-core and miden-crypto types.

Impact observed:

After migrating from Miden VM 0.20 → 0.21, miden-core now appears in the top compilation units in miden-protocol at 12.1s (with features arbitrary, serde, std, testing) — it wasn't in the top 10 previously.
miden-crypto also picked up the serde feature where it previously only had concurrent, std.

Suggested fix:

Make serde an optional feature on miden-air that consumers can opt into, rather than unconditionally requiring it on miden-core:

### miden-air/Cargo.toml
```toml
[features]
serde = ["miden-core/serde"]

[dependencies]
miden-core = { version = "0.21", default-features = false }  # without serde
```

This way, only consumers that actually need serialization pay the cost.

giwaov:
﻿## Summary

Add an opt-in serde feature to miden-air that enables miden-core/serde.

## Problem

As reported in #2744, miden-air was unconditionally enabling serde on miden-core. This causes increased compile times for consumers who don't need serialization.

## Solution

Add an optional serde feature to miden-air/Cargo.toml that enables miden-core/serde. Consumers who need serde can now opt in explicitly.

## Testing

- cargo check -p miden-air passes
- cargo check -p miden-air --features serde passes

Closes #2744

### DIFF
--- a/air/Cargo.toml
+++ b/air/Cargo.toml
@@ -20,6 +20,7 @@ doctest = false
 [features]
 default = ["std"]
 std = ["miden-core/std", "thiserror/std"]
+serde = ["miden-core/serde"]
 concurrent = ["std"]
 testing = []
 


huitseeker:
Not useful, #2744 is fixed modulo some branch management.

PR #2792, opened 2026-03-06
Add recursion guards and depth limits for assembly compilation paths

huitseeker:
- Add cycle detection and explicit depth limits in alias expansion, control-flow compilation, type resolution, and constant folding, returning structured errors on overflow.
- Document limit constants with rustdoc rationale ~and add stacksafe TODOs on recursive functions~.
- Add boundary and over-limit tests for all four guard paths.


crates/assembly-syntax/src/ast/constants/expr.rs
```diff
@@ -127,19 +133,31 @@ impl ConstantExpr {
     /// # Errors
     /// Returns an error if an invalid expression is found while folding, such as division by zero.
     pub fn try_fold(self) -> Result<Self, ParsingError> {
+        self.try_fold_with_depth(0)
+    }
+
+    // TODO: consider the stacksafe crate to guard deep recursion here.
```
plafer: I would prefer not adding any new TODOs to the codebase. These tend to never get resolved (look at the number of 3+ year olds TODOs we already have), and it's a recipe for LLMs to open PRs against it. I'd prefer we open an issue instead (and possibly link to the issue here too if you want).

Applies to other TODOs that were added

huitseeker: Fair. Removed. On that point: https://github.com/0xMiden/miden-vm/pull/2796

PR #2797, opened 2026-03-06
Introduce FastProcessor safe stack access methods

plafer:
Introduces safe stack access methods that will never read outside the stack buffer for use by event handlers.

Fixes audit issue 66:

> A crafted program can deterministically panic the fast processor in release builds by triggering an out-of-bounds stack word read through a system event invoked via emit.

This PR introduces safe access methods (which are slower than the existing ones), only to be used by event handlers (through the `ProcessorState` type), and our internal system event handlers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fixes an possible u64 overflow issue in `op_eval_circuit()` [#2799](https://github.com/0xMiden/miden-vm/pull/2799)
 - Preserved dynexec/dyncall distinction (and digests) when remapping or merging MAST forests ([#2784](https://github.com/0xMiden/miden-vm/pull/2784)).
 - Hardened MASM parsing and constants handling (lexer invalid-token spans, repeat count bounds, constant range checks, field division folding, and `push.WORD[...]` index validation) ([#2803](https://github.com/0xMiden/miden-vm/pull/2803)).
+- Introduced `FastProcessor` safe stack method accesses for event handlers ([#2797](https://github.com/0xMiden/miden-vm/pull/2797)).
 
 ## 0.21.1 (2026-02-24)
 
--- a/processor/src/lib.rs
+++ b/processor/src/lib.rs
@@ -184,7 +184,7 @@ impl<'a> ProcessorState<'a> {
     /// This method can access elements beyond the top 16 positions by using the overflow table.
     #[inline(always)]
     pub fn get_stack_item(&self, pos: usize) -> Felt {
-        self.processor.stack_get(pos)
+        self.processor.stack_get_safe(pos)
     }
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 #### Fixes
 
 - Fixed `Constant::PartialEq` to include `visibility` field in equality comparison, making it consistent with other exportable items (`Procedure`, `TypeAlias`, `EnumType`).
+- Introduced `FastProcessor` safe stack method accesses for event handlers ([#2797](https://github.com/0xMiden/miden-vm/pull/2797)).
```
huitseeker, line 24: This line links to PR #2798, but this change is in #2797. Also, should we add a CHANGELOG for every issue fixed here?

processor/src/lib.rs
```diff
@@ -184,7 +184,7 @@ impl<'a> ProcessorState<'a> {
     /// This method can access elements beyond the top 16 positions by using the overflow table.
     #[inline(always)]
     pub fn get_stack_item(&self, pos: usize) -> Felt {
-        self.processor.stack_get(pos)
+        self.processor.stack_get_safe(pos)
```
huitseeker, line 187: These doc comments still say overflow-table access, but `stack_get_safe`/`stack_get_word_safe` return `ZERO` on out-of-range. Should the docs be updated to match the new semantics?

CHANGELOG.md
plafer: PR number fixed.

> Also, should we add a CHANGELOG for every issue fixed here?

I think so - this will end up making it into 0.22, and it'll be nice to see all the fixes there

processor/src/lib.rs
plafer: > but stack_get_safe/stack_get_word_safe return ZERO on out-of-range.

On out-of-range *in the stack buffer*, which implies that there the `pos` requested is past the end of the overflow table. So the behavior is unchanged, and the docstring is correct. Is this worth documenting explicitly in a comment?

huitseeker: Ah, yes. No, I just needed to read the code comment again.

processor/src/processor.rs
```diff
@@ -149,10 +149,13 @@ pub(crate) trait StackInterface {
     /// index of the returned slice.
     fn top(&self) -> &[Felt];
 
-    /// Returns the element on the stack at index `idx`.
+    /// Returns the element on the stack at index `idx`. `idx` is guaranteed to be less than or
```
adr1anh: Below the comments put the second sentence in a new line. Also, shouldn't we say `idx must be ...`?

plafer: Agreed; fixed

PR #2799, opened 2026-03-06
Fix eval_circuit overflow panic (audit issue 48)

plafer:
Fixes audit issue 48:

> In release builds, eval_circuit can panic due to unchecked u64 overflow in its wire-count check, followed by unchecked narrowing casts that feed an assert! in the ACE evaluator. A malicious program can use crafted stack values to crash a proving worker.

### DIFF
--- a/processor/src/execution/operations/eval_circuit.rs
+++ b/processor/src/execution/operations/eval_circuit.rs
@@ -62,7 +62,7 @@ pub(crate) fn eval_circuit_impl(
     let num_vars = num_vars.as_canonical_u64();
     let num_eval = num_eval.as_canonical_u64();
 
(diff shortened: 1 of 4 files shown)


processor/src/execution/operations/eval_circuit.rs
```diff
@@ -90,7 +90,7 @@ pub(crate) fn eval_circuit_impl(
     }
 
     // Ensure instructions are word-aligned and non-empty
-    let num_read_rows = num_vars as u32 / 2;
+    let num_read_rows = (num_vars / 2) as u32;
```
huitseeker, line 93: This works `MAX_NUM_ACE_WIRES` doesn't permits values above `u32::MAX * 2`.

PR #2802, opened 2026-03-06
Apply code quality fixes and tests

huitseeker:
- simplify package section deserialization via Vec::read_from
- use MastForest register_node_decorators directly in LoopNodeBuilder
- keep non-mergeable basic blocks standalone; add/strengthen test
- fix stack_write_word debug bound and add max-index test
- remove redundant indptr prefix/bounds checks in batch validation
- rename trace buffer fields to *cols for clarity
- update Ext2Mul polynomial comment

### DIFF
--- a/crates/assembly/src/mast_forest_builder.rs
+++ b/crates/assembly/src/mast_forest_builder.rs
@@ -404,7 +404,10 @@ impl MastForestBuilder {
             // check if the block should be merged with other blocks
             if should_merge(
                 self.mast_forest.is_procedure_root(basic_block_id),
-                self.mast_forest[basic_block_id].get_basic_block().unwrap().num_op_batches(),
+                self.mast_forest[basic_block_id]
+                    .get_basic_block()
+                    .expect("merge_basic_blocks: expected BasicBlockNode")
+                    .num_op_batches(),
             ) {
                 // Collect decorators and operations from the block (while still borrowing)
                 // We need owned copies so we can drop the borrow before mutating self
@@ -421,8 +424,6 @@ impl MastForestBuilder {
                     (block_decorators, block_ops)
                 };
                 let ops_offset = operations.len();
-
-                // basic_block_node borrow ends here since we collected owned data above
 
                 // Transfer any pending asm_ops for this block to the merged result
                 self.transfer_asm_ops_for_merge(basic_block_id, ops_offset, &mut merged_asm_ops);
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -331,7 +331,7 @@ impl FastProcessor {
             "Not enough elements on stack to read word starting at index {start_idx}"
         );
 
-        let word_start_idx = self.stack_top_idx - start_idx - 4;
+        let word_start_idx = self.stack_top_idx - start_idx - WORD_SIZE;
         let mut result: [Felt; WORD_SIZE] =
             self.stack[range(word_start_idx, WORD_SIZE)].try_into().unwrap();
         // Reverse so top of stack (idx 0) goes to word[0]


crates/assembly/src/mast_forest_builder.rs
```diff
@@ -401,24 +401,25 @@ impl MastForestBuilder {
...
         for &basic_block_id in contiguous_basic_block_ids {
-            // It is safe to unwrap here, since we already checked that all IDs in
-            // `contiguous_basic_block_ids` are `BasicBlockNode`s
-            let basic_block_node = self.mast_forest[basic_block_id].get_basic_block().unwrap();
-
             // check if the block should be merged with other blocks
             if should_merge(
                 self.mast_forest.is_procedure_root(basic_block_id),
-                basic_block_node.num_op_batches(),
+                self.mast_forest[basic_block_id].get_basic_block().unwrap().num_op_batches(),
```
Al-Kindi-0: nit: would be good to use `expect` with a short message

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -545,20 +545,6 @@ impl BasicBlockNode {
...
-                        "Batch {}: indptr[{}] {} > indptr[{}] {} - array is not monotonic",
-                        batch_idx,
-                        i,
-                        indptr[i],
-                        i + 1,
-                        indptr[i + 1]
-                    ));
-                }
-            }
-
```
Al-Kindi-0, line 561: why is this no longer needed ?

core/src/mast/node/basic_block_node/mod.rs
bobbinth, line 1: I didn't review this file in detail - but it wasn't immediately obvious why the removed code was redundant.

processor/src/fast/mod.rs
```diff
@@ -370,7 +370,7 @@ impl FastProcessor {
     /// `word[0]` goes to stack position start_idx (top), `word[1]` to start_idx+1, etc.
     #[inline(always)]
     pub fn stack_write_word(&mut self, start_idx: usize, word: &Word) {
-        debug_assert!(start_idx < MIN_STACK_DEPTH);
+        debug_assert!(start_idx <= MIN_STACK_DEPTH - WORD_SIZE);
 
         let word_start_idx = self.stack_top_idx - start_idx - 4;
```
bobbinth, line 375: Should we use `WORD_SIZE` instead of `4` here?

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -401,24 +401,25 @@ impl MastForestBuilder {
...
+                    let block_ops: Vec<Operation> = basic_block_node
+                        .op_batches()
+                        .iter()
+                        .flat_map(|b| b.raw_ops().copied())
+                        .collect();
+                    (block_decorators, block_ops)
+                };
                 let ops_offset = operations.len();
 
                 // basic_block_node borrow ends here since we collected owned data above
```
bobbinth: This comment may be obsolete now.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Duplicated a few lines below.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -401,24 +401,25 @@ impl MastForestBuilder {
...
+                        self.mast_forest[basic_block_id].get_basic_block().unwrap();
+                    let block_decorators: Vec<_> =
+                        basic_block_node.raw_decorator_iter(&self.mast_forest).collect();
+                    let block_ops: Vec<Operation> = basic_block_node
+                        .op_batches()
+                        .iter()
+                        .flat_map(|b| b.raw_ops().copied())
+                        .collect();
+                    (block_decorators, block_ops)
+                };
```
bobbinth, line 425: Unless I missed something, it seems like the old and the new code are basically the same. What motivated this change?

core/src/mast/node/basic_block_node/mod.rs
huitseeker: We have out-of-band context for this PR, is it something you managed to use?
TL;DR: The monotonicity check is duplicated, the monotonicity check + final (largest) value check precludes values being out of bounds.

crates/assembly/src/mast_forest_builder.rs
huitseeker: The old code bound basic_block_node for the entire loop iteration. The new code scopes that borrow inside a block, so later calls that require mutable access to self (like `transfer_asm_ops_for_merge`) don’t conflict with the borrow.

CergyK:
LGTM

PR #2803, opened 2026-03-07
Harden MASM parsing and constant evaluation (lexer, repeats, constants, push slices)

huitseeker:
- Fix lexer span handling for invalid tokens after whitespace to avoid panics and return structured errors.
- Enforce repeat count bounds (including decorators) and document the max; add regression tests for zero/too-large counts.
- Enforce constant range checks during folding and evaluation, preventing modulus aliasing and overflows; add tests.
- Implement field division for / in constant folding to match docs; add tests.
- Validate push.WORD[...] slice indices with checked conversions and overflow guards; add single-index and range-index regression tests.

Per-commit review recommended.

### DIFF
--- a/crates/assembly-syntax/src/ast/constants/eval.rs
+++ b/crates/assembly-syntax/src/ast/constants/eval.rs
@@ -6,9 +6,11 @@ use alloc::{sync::Arc, vec::Vec};
 use smallvec::SmallVec;
 
 use crate::{
+    Felt,
     ast::*,
     debuginfo::{SourceFile, SourceSpan, Span, Spanned},
     diagnostics::{Diagnostic, RelatedLabel, miette},
+    parser::IntValue,
 };
 
 /// An error raised during evaluation of a constant expression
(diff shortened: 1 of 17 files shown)


crates/assembly-syntax/src/sema/passes/verify_repeat.rs
```diff
@@ -0,0 +1,38 @@
...
+    analyzer: &'a mut AnalysisContext,
+}
+
+impl<'a> VerifyRepeatCounts<'a> {
+    pub fn new(analyzer: &'a mut AnalysisContext) -> Self {
+        Self { analyzer }
+    }
+}
+
+impl VisitMut for VerifyRepeatCounts<'_> {
```
bitwalker: Since this pass doesn't mutate the AST, you can implement it as `Visit` instead - doesn't really matter too much, but better signals that the implementation is just analysis rather than transformation.

bitwalker (approved):

plafer (approved):

crates/assembly-syntax/src/ast/constants/eval.rs
```diff
@@ -351,12 +353,24 @@ where
...
+                                    return Err(ConstEvalError::DivisionByZero {
+                                        span,
+                                        source_file: env.get_source_file_for(span),
+                                    }
+                                    .into());
+                                }
+                                let lhs = Felt::new(lhs.as_int());
+                                let rhs = Felt::new(rhs.as_int());
+                                IntValue::from(lhs / rhs)
+                            },
```
CergyK, line 373: As discussed, this will be a confusing behavior (`/` as felt division, while `//` is the "usual" division).

huitseeker: #3053

PR #2804, opened 2026-03-08
Harden assembly validation: syscalls, enums, paths, and kernel libraries

huitseeker:
- Reject empty kernel libraries early and avoid installer panics.
- Strengthen path validation to prevent panics on overlong components/total length.
- Fix enum variant handling, discriminant checks, and constant folding overflow.
- Resolve error-message constant aliases without panicking.
- Enforce syscall validation against kernel digests with structured errors.

Per-commit review highly recommended.

### DIFF
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -242,7 +242,7 @@ impl Module {
     pub fn define_enum(&mut self, ty: EnumType) -> Result<(), SemanticAnalysisError> {
         let repr = ty.ty().clone();
 
-        if !repr.is_integer() && repr != ast::types::Type::Felt {
+        if !repr.is_integer() {
             return Err(SemanticAnalysisError::InvalidEnumRepr { span: ty.span() });
         }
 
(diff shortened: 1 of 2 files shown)


crates/assembly-syntax/src/ast/module.rs
```diff
@@ -242,7 +242,7 @@ impl Module {
     pub fn define_enum(&mut self, ty: EnumType) -> Result<(), SemanticAnalysisError> {
         let repr = ty.ty().clone();
 
-        if !repr.is_integer() {
+        if !repr.is_integer() && repr != ast::types::Type::Felt {
```
bitwalker: `is_integer` returns true for `Type::Felt`, so this check is redundant

crates/assembly-syntax/src/ast/type.rs
```diff
@@ -840,7 +816,10 @@ impl EnumType {
         ty: Type,
         variants: impl IntoIterator<Item = Variant>,
     ) -> Self {
-        assert!(ty.is_integer(), "only integer types are allowed in enum type definitions");
+        assert!(
+            ty.is_integer() || ty == Type::Felt,
```
bitwalker: Same here

PR #2805, opened 2026-03-09
Add bounds to allocations in advice map and keccak256/sha512 precompiles

plafer:
Addresses audit issue 36:

> Several host-side handlers derive allocation sizing and loop length from attacker-controlled sizes without explicit caps. In prover-service deployments where the attacker controls the program and advice inputs, adv.insert_mem, adv.push_mapval*, and the keccak256 precompile can be used to force large host-side CPU and memory usage that is not directly metered by VM cycle limits.

### DIFF
--- a/crates/lib/core/src/handlers/keccak256.rs
+++ b/crates/lib/core/src/handlers/keccak256.rs
@@ -23,7 +23,7 @@
 //! A Keccak256 digest (256 bits) is represented as 8 field elements `[h0, ..., h7]`,
 //! each containing a u32 value where `hi = u32::from_le_bytes([b_{4i}, ..., b_{4i+3}])`.
 
-use alloc::{vec, vec::Vec};
+use alloc::{format, vec, vec::Vec};
 use core::array;
 
 use miden_core::{
@@ -67,8 +67,23 @@ impl EventHandler for KeccakPrecompile {
         let ptr = process.get_stack_item(1).as_canonical_u64();
         let len_bytes = process.get_stack_item(2).as_canonical_u64();
 
+        // Enforce the maximum precompile input size to prevent host-side resource exhaustion. Make
+        // sure to compare in u64 to avoid truncation on 32-bit targets.
+        let max = process.execution_options().max_hash_len_bytes();
+        if len_bytes > max as u64 {
+            return Err(format!(
+                "keccak256 input length {len_bytes} bytes exceeds maximum of {max} bytes"
+            )
+            .into());
+        }
+        let len_bytes = usize::try_from(len_bytes).map_err(|_| {
+            EventError::from(format!(
+                "keccak256 input length {len_bytes} exceeds addressable range"
+            ))
+        })?;
+
         // Read input bytes from memory using the shared helper (u32-packed, LE, zero-padded)
-        let input_bytes = read_memory_packed_u32(process, ptr, len_bytes as usize)?;
+        let input_bytes = read_memory_packed_u32(process, ptr, len_bytes)?;
 
         // Build preimage from bytes and compute digest
         let preimage = KeccakPreimage::new(input_bytes);
--- a/crates/lib/core/src/handlers/sha512.rs
+++ b/crates/lib/core/src/handlers/sha512.rs
@@ -5,7 +5,7 @@
 //! record calldata for deferred verification. Verification-time logic recomputes the digest and
 //! commits to both input and output using Poseidon2 hashing.
 
-use alloc::{vec, vec::Vec};
+use alloc::{format, vec, vec::Vec};
 use core::array;
 
 use miden_core::{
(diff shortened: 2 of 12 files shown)


crates/lib/core/src/handlers/keccak256.rs
```diff
@@ -67,6 +67,15 @@ impl EventHandler for KeccakPrecompile {
         let ptr = process.get_stack_item(1).as_canonical_u64();
         let len_bytes = process.get_stack_item(2).as_canonical_u64();
 
+        // Enforce the maximum precompile input size to prevent host-side resource exhaustion.
+        let max = process.execution_options().max_hash_len_bytes();
+        if len_bytes as usize > max {
```
huitseeker: This looks like a width cast before the limit check. On 32-bit targets, `len_bytes as usize` can truncate, so a value above `max_hash_len_bytes` may pass this guard and then flow into `read_memory_packed_u32` with the truncated length. Could we compare in `u64` first (`len_bytes > max as u64`) and then do a checked `usize::try_from(len_bytes)` before reading memory?

crates/lib/core/src/handlers/sha512.rs
```diff
@@ -43,6 +43,15 @@ impl EventHandler for Sha512Precompile {
         let ptr = process.get_stack_item(1).as_canonical_u64();
         let len_bytes = process.get_stack_item(2).as_canonical_u64();
 
+        // Enforce the maximum precompile input size to prevent host-side resource exhaustion.
+        let max = process.execution_options().max_hash_len_bytes();
+        if len_bytes as usize > max {
```
huitseeker: Same pattern here as in `keccak256`: the guard compares `len_bytes as usize` against `max`. On 32-bit this cast can truncate before the check. Would you be open to checking in `u64` first, then using `usize::try_from(len_bytes)` so the conversion cannot silently wrap?

processor/src/execution_options.rs
```diff
@@ -38,6 +45,14 @@ impl ExecutionOptions {
     /// Default fragment size for core trace generation.
     pub const DEFAULT_CORE_TRACE_FRAGMENT_SIZE: usize = 4096; // 2^12
 
+    /// Default maximum number of field elements in a single advice map value inserted via
+    /// `adv.insert_mem`. Set to 2^17 (~1 MB given 8-byte field elements).
+    pub const DEFAULT_MAX_ADV_MAP_VALUE_SIZE: usize = 1 << 17;
+
+    /// Default maximum number of input bytes for a single hash precompile invocation (e.g.
+    /// keccak256, sha512, etc.). Set to 2^20 (1 MB).
+    pub const DEFAULT_MAX_HASH_LEN_BYTES: usize = 1 << 20;
```
huitseeker, line 54: I may be missing a test, but I don’t see boundary coverage for the new caps through execution paths (`max` allowed, `max + 1` rejected) for both `max_hash_len_bytes` and `max_adv_map_value_size`. Could we add one focused test per boundary to lock this behavior in?

crates/lib/core/src/handlers/keccak256.rs
plafer: Fixed

crates/lib/core/src/handlers/sha512.rs
plafer: Fixed

processor/src/execution_options.rs
plafer: Done

processor/src/host/advice/mod.rs
```diff
@@ -120,16 +124,33 @@ impl AdviceProvider {
         Ok([word0, word1])
     }
 
+    /// Checks that pushing `count` elements would not exceed the advice stack size limit.
+    fn check_stack_capacity(&self, count: usize) -> Result<(), AdviceError> {
+        let resulting_size = self.stack.len() + count;
```
huitseeker: Not that this is likely, but could we make this addition overflow-safe with `checked_add`? If `len + count` wraps, the limit check can be bypassed before we return `StackSizeExceeded`.

processor/src/host/advice/mod.rs
```diff
@@ -166,13 +187,19 @@ impl AdviceProvider {
     ) -> Result<(), AdviceError> {
         let values = self.map.get(&key).ok_or(AdviceError::MapKeyNotFound { key })?;
 
+        // Calculate total elements to push including padding and optional length prefix
+        let num_pad_elements = if pad_to != 0 {
+            values.len().next_multiple_of(pad_to as usize) - values.len()
+        } else {
+            0
+        };
+        let total_push = values.len() + num_pad_elements + if include_len { 1 } else { 0 };
```
huitseeker: Could we make `total_push` use `checked_add` as well? The pieces are attacker-influenced sizes, so wrapping here could undercount and skip the stack cap check.

processor/src/host/advice/mod.rs
plafer: Done

processor/src/host/advice/mod.rs
plafer: Done

PR #2808, opened 2026-03-09
Harden arithmetic and crypto boundary checks across u64::shr, ilog2, u32clz, and Falcon mod_12289

huitseeker:
- Tightens boundary/overflow verification logic in four sensitive paths:

  1. u64::shr: fixes behavior for shifts >= 32 by correctly zeroing the high limb in that branch.
  2. ilog2: replaces half-word-style validation with explicit full-width lower/upper bound checks (2^ilog2 <= n < 2^(ilog2+1), with ilog2 == 63 special-case handling).
  3. u32clz: enforces exact zero boundary (n == 0 iff clz == 32) and strengthens mask consistency checks.
  4. Falcon mod_12289: converts previously dropped overflow flags into explicit assertions for quotient/addition overflow safety.

- Adds focused regression/security-style tests for each fix:

  1. u64::shr edge cases at 32, >32, and boundary sweeps.
  2. ilog2 forged advice and boundary-claim rejection/acceptance tests.
  3. u32clz zero/non-zero boundary correctness tests.
  4. Falcon malicious advice/overflow rejection tests.

- Updates cycle-count documentation/comments where instruction costs changed (u64::shr, u32clz, ilog2).

Per-commit review recommended

### DIFF
--- a/crates/lib/core/asm/math/u64.masm
+++ b/crates/lib/core/asm/math/u64.masm
@@ -694,10 +694,7 @@ pub proc shr(n: u32, a: u64) -> u64
     # c_lo = rem1 * multiplier + quot2
 
     # --- STEP 9: Conditional swap to get final result ---
-    movup.2
-    # [!borrow, c_lo, quot1, ...]
-
-    dup movdn.3
+    dup.2
     # [!borrow, c_lo, quot1, !borrow, ...]
 
     cswap
@@ -709,15 +706,10 @@ pub proc shr(n: u32, a: u64) -> u64
     # !borrow=1 keeps c_hi unchanged for n<32; !borrow=0 clears c_hi for n>=32.
     movup.2
     # [!borrow, c_hi, c_lo, ...]
-
-    swap
-    # [c_hi, !borrow, c_lo, ...]
-
     mul
     # [c_hi * !borrow, c_lo, ...]
-
-    # Convert BE output to LE: [c_hi, c_lo] -> [c_lo, c_hi]
     swap
+    # [c_lo, c_hi, ...]
 end
 
 #! Performs left rotation of one unsigned 64-bit integer.


crates/lib/core/tests/crypto/falcon.rs
```diff
@@ -344,6 +344,113 @@ fn test_mod_12289_larger_value() {
...
+    let source = "
+        use miden::core::crypto::dsa::falcon512poseidon2
+        begin
+            exec.falcon512poseidon2::mod_12289
+        end
+    ";
+
+    // Inputs are (a_hi, a_lo). Every case has a non-zero true remainder.
+    let test_cases: [(u64, u64); 5] =
+        [(0, 100_000), (2, 0), (0, 12_290), (1, 1), (0xffff_ffff, 0xffff_fffe)];
```
plafer: nit: we could use `rstest` cases for this - which is cleaner IMO because it doesn't mix the test's business logic with "test management".

But we don't need to change it here; more a comment for future tests.

processor/src/execution/operations/u32_ops/tests.rs
```diff
@@ -307,6 +313,105 @@ fn test_op_u32xor_min_stack() {
...
+    let bad_clz = 32u32;
+
+    assert_ne!(n.leading_zeros(), bad_clz, "sanity: witness must be invalid");
+    assert!(run_verify_clz_gadget(n, bad_clz).is_err());
+}
+
+#[test]
+fn verify_clz_accepts_zero_with_clz_32() {
+    run_verify_clz_gadget(0, 32).expect("gadget should accept valid zero boundary witness");
+}
```
plafer, line 413: I think these should be MASM-based tests, since the ultimate goal is to test the regression with the assembler-generated code for `u32clz`. Instead, we're testing here the hardcoded `run_verify_clz_gadget()`, which might get out of date with what the assembler is doing.

huitseeker: I've added MASM tests, but MASM tests validate observed output under normal system behavior, not
explicit rejection of malicious/wrong CLZ witnesses, so I've kept those.

crates/lib/core/asm/math/u64.masm
```diff
@@ -697,10 +697,24 @@ pub proc shr(n: u32, a: u64) -> u64
...
+    # when n >= 32, c_hi must be zero. multiply c_hi by !borrow to enforce this:
+    # !borrow=1 keeps c_hi unchanged for n<32; !borrow=0 clears c_hi for n>=32.
+    movup.2
+    # [!borrow, c_hi, c_lo, ...]
+
+    swap
+    # [c_hi, !borrow, c_lo, ...]
+
+    mul
+    # [c_hi * !borrow, c_lo, ...]
```
Al-Kindi-0: could we do

```
dup.2
# => [!borrow, c_lo, quot1, !borrow, ...]

cswap
#=> [c_hi_res, c_lo_res, !borrow, ...]

movup.2 mul swap
# => [c_lo, c_hi, ...]
```

?

huitseeker: Thanks!

CergyK:
Fixes look good

PR #2809, opened 2026-03-09
`build_trace()` no longer allocates past a given maximum number of rows

### LINKED ISSUE #2807: Panic in generate_core_trace_columns() when core_trace_contexts is empty
When core_trace_contexts is empty, generate_core_trace_columns() panics instead of returning an error. The function allocates zero-length columns at line 184-185, then unconditionally calls fixup_stack_and_system_rows() at line 228, which writes to core_trace_columns[*][0] at line 320 and panics with index out of bounds.

The error guards at lines 251-256 are unreachable because the panic occurs before them.

The fix is to validate core_trace_contexts.is_empty() at the start of generate_core_trace_columns() and return ExecutionError::Internal("no trace fragments provided in the trace generation context") early.

plafer:
This PR addresses audit finding 61, which discusses how an attacker can force a ~600GB allocation on the prover by sending a program with just `op_mrupdate`, due to incurring however many rows in the Hasher chiplet.

The solution in this PR is against `next` instead of our audit branch, since we deviate slightly from recommendations. Since #2747 (not present in the audit branch), `build_trace()` no longer assumes valid user inputs. Hence, it doesn't make sense to implement the chiplets length check in the `ExecutionTracer` - whatever `TraceGenerationContext` it generates is not trusted by `build_trace()`, and so `build_trace()` will need to redo those checks anyway. Hence, this is what we do in this PR: `build_trace()` checks as its building the chiplets trace that its rows doesn't exceed some maximum.

We also expose a new `build_trace_with_max_len()` which allows the caller to control the maxmimum trace length; `build_trace()` now defaults to $2^{29}$.

The `build_trace` benchmark is maybe 1-2% slower due to the checks, although that's not completely clear since this is roughly in the noise bounds of the benchmark (it's pretty noisy).

EDIT: also Closes #2807.

### DIFF
--- a/processor/src/trace/parallel/mod.rs
+++ b/processor/src/trace/parallel/mod.rs
@@ -40,7 +40,10 @@ use crate::{
 
 pub const CORE_TRACE_WIDTH: usize = SYS_TRACE_WIDTH + DECODER_TRACE_WIDTH + STACK_TRACE_WIDTH;
 
-/// Maximum allowed trace length (2^29 rows).
+/// `build_trace()` uses this as a hard cap on trace rows.
+///
+/// The code checks `core_trace_contexts.len() * fragment_size` before allocation. It checks the
+/// same cap again while replaying chiplet activity. This keeps memory use bounded.
 const MAX_TRACE_LEN: usize = 1 << 29;
 
 pub(crate) mod core_trace_fragment;
@@ -82,7 +85,10 @@ pub fn build_trace(
     )
 }
 
-/// Same as [`build_trace`], but with a configurable maximum trace length.
+/// Same as [`build_trace`], but with a custom hard cap.
+///
+/// When the trace would go over `max_trace_len`, this returns
+/// [`ExecutionError::TraceLenExceeded`].
 pub fn build_trace_with_max_len(
     execution_output: ExecutionOutput,
     trace_generation_context: TraceGenerationContext,
--- a/docs/src/advanced_topics/execution_trace_optimization.md
+++ b/docs/src/advanced_topics/execution_trace_optimization.md
@@ -24,6 +24,8 @@ Note: The maximum gets rounded up to the next power of 2, and the other 2 sets o
 
 In some cases, either the range checker or chiplets could end up requiring more rows than the stack rows, making the true cycle count higher than what the stack-based cycle counter reports.
 
+The runtime also applies a hard trace limit of `2^29` rows in parallel trace building. If replayed core or chiplet rows would pass that limit, execution stops with `TraceLenExceeded` instead of trying to allocate a larger trace.
+
 ## Analyzing trace segments with miden-vm analyze
 
 The `miden-vm analyze` command provides detailed information about trace segment utilization, showing:
--- a/docs/src/usage.md
+++ b/docs/src/usage.md
@@ -109,6 +109,8 @@ You can use the run command with `--debug` parameter to enable debugging with th
 ./target/optimized/miden-vm run [path_to.masm] --debug
 ```
 
+If trace building would exceed the VM trace row limit, `run` returns a trace length error instead of trying to build a larger trace.
+
 ### Inputs
 
 As described [here](https://docs.miden.xyz/miden-vm/overview#inputs-and-outputs) the Miden VM can consume public and secret inputs.
--- a/processor/src/errors.rs
+++ b/processor/src/errors.rs
@@ -58,6 +58,9 @@ pub enum ExecutionError {
     },
     #[error("failed to execute the program for internal reason: {0}")]
     Internal(&'static str),
+    /// This means trace generation would go over the configured row limit.
+    ///
+    /// In parallel trace building, this is used for core-row prechecks and chiplet overflow.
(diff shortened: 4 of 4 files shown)


processor/src/trace/parallel/mod.rs
```diff
@@ -83,6 +101,20 @@ pub fn build_trace(
         fragment_size,
     } = trace_generation_context;
 
+    // Before any trace generation, check that the number of core trace rows doesn't exceed the
+    // maximum trace length. This is a necessary check to avoid OOM panics during trace generation,
+    // which can occur if the execution produces an extremely large number of steps.
+    let total_core_trace_rows = core_trace_contexts.len() * fragment_size;
```
huitseeker: A bit of a technicality, but I think this check can still let us go past the cap by one row. `generate_core_trace_columns()` always appends a HALT row, so when `core_trace_contexts.len() * fragment_size == max_trace_len`, this `>` guard passes but the later `push` can exceed the limit and force a big reallocation.

Could we switch this to a checked bound that includes HALT, like `checked_mul(...).and_then(|n| n.checked_add(1))`, then reject on `>=` overflow/exceed? That would also avoid wraparound in the multiply.

huitseeker: I was able to reproduce the unchecked multiply path with this unit test, which currently fails because `core_trace_contexts.len() * fragment_size` panics before we can return `TraceLenExceeded`.

```rust
#[test]
fn test_build_trace_returns_err_on_core_trace_row_mul_overflow() {
    const MIN_FRAGMENT_SIZE: usize = 1;

    let program = basic_block_program_small();
    let processor = FastProcessor::new_with_options(
        StackInputs::new(DEFAULT_STACK).unwrap(),
        AdviceInputs::default(),
        ExecutionOptions::default()
            .with_core_trace_fragment_size(MIN_FRAGMENT_SIZE)
            .unwrap(),
    );
    let (execution_output, mut trace_generation_context) =
        processor.execute_for_trace_sync(&program, &mut host).unwrap();

    trace_generation_context.fragment_size = usize::MAX;

    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        build_trace_with_max_len(
            execution_output,
            trace_generation_context,
            program.to_info(),
            usize::MAX,
        )
    }));

    assert!(
        matches!(result, Ok(Err(ExecutionError::TraceLenExceeded(_)))),
        "expected TraceLenExceeded on multiplication overflow, got: {result:?}"
    );
}
```

I ran:
`make test-processor test="'test_build_trace_returns_err_on_core_trace_row_mul_overflow'"`
and got:
`thread ... panicked at processor/src/trace/parallel/mod.rs:107:33: attempt to multiply with overflow`.

huitseeker:
@plafer I've left an extension in https://github.com/0xMiden/miden-vm/pull/2810

processor/src/trace/parallel/mod.rs
plafer: Good catch.

> then reject on >= overflow/exceed

Agreed with `checked_mul()`/`checked_add(1)`; but since we added 1, we should reject on `>` right? Basically the `max_trace_len` is allowed, given that we've already accounted for the HALT

plafer: - Implemented the suggestion, but using `>` given that we do a `checked_add(1)` as discussed above.
- Converted the existing test to an `rstest` to check this edge case
  - when core trace len (including HALT) is exactly equal to max trace len, succeed
  - when core trace len (including HALT) is one more than max, fail
- Added a test to make sure we don't panic due to arithmetic overflow when computing `total_core_trace_rows`

If this works for you, feel free to merge

huitseeker: it does

processor/src/trace/parallel/mod.rs
```diff
@@ -70,6 +76,24 @@ pub fn build_trace(
...
+/// Same as [`build_trace`], but with a custom hard cap.
+///
+/// When the trace would go over `max_trace_len`, this returns
+/// [`ExecutionError::TraceLenExceeded`].
+pub fn build_trace_with_max_len(
+    execution_output: ExecutionOutput,
+    trace_generation_context: TraceGenerationContext,
+    program_info: ProgramInfo,
+    max_trace_len: usize,
 ) -> Result<ExecutionTrace, ExecutionError> {
```
bobbinth, line 97: This approach works fine for now, but as we move to multi-table architecture, we'll need to re-work it. 

Basically, what we really care about here is memory consumption rather than the number of trace rows - it just so happens, that in a single-table setup, these are proportional to each other.

In multi-table setup, we'd need to replace this metric with something like "number of trace cells" - though, I think that would be pretty confusing for the end use. A better approach could be just to make this about memory. Specifically, instead of setting `max_trace_len` (or `max_num_trace_cells`), the user would provide the maximum memory they are willing to allocate (e.g., 32 GB), and we'd do the job of converting this "user-friend" metric into some internally tracked metric.

Ideally, we'd have something like `TraceGenerationContext::estimate_prover_memory()` which would give us the amount of memory the prover would need to generate a proof for this context (I believe the context should have all the info about this - but let me know if not).

PR #2810, opened 2026-03-10
Enforce stack-usage contracts and document trace-length cutoff

huitseeker:
Fixes stack contract mismatches in core procedures and adds repro/coverage tests for the affected APIs (`aead::{encrypt,decrypt}`, `keccak256::{merge,hash}`, `sha512::hash_bytes`, and `u256::wrapping_mul`). (audit finding 71)

Also documents the new policy/error semantics in code docs, user docs, and the changelog.

### DIFF
--- a/docs/src/advanced_topics/execution_trace_optimization.md
+++ b/docs/src/advanced_topics/execution_trace_optimization.md
@@ -24,6 +24,8 @@ Note: The maximum gets rounded up to the next power of 2, and the other 2 sets o
 
 In some cases, either the range checker or chiplets could end up requiring more rows than the stack rows, making the true cycle count higher than what the stack-based cycle counter reports.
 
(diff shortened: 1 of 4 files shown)


crates/lib/core/asm/math/u256.masm
```diff
@@ -715,4 +715,7 @@ pub proc wrapping_mul(rhs: u256, lhs: u256) -> u256
     loc_loadw_be.16
     swapw
     exec.u256_le_to_be
+    # drop intermediate x and carry words to match the documented output
+    swapdw dropw dropw
+    swapdw dropw dropw
```
plafer, line 720: We should document this fix in the changelog - but I'll do it in my PR after we merge this

plafer: Actually probably makes more sense to merge my PR first, so we should change the changelog here

PR #2825, opened 2026-03-10
fix(processor): set a bound on continuation stack size

plafer:
Addresses the continuation stack part of audit finding 41:

> The processor has no runtime bound on the size of the heap-allocated [ContinuationStack](https://github.com/0xMiden/miden-vm/blob/089dc6a33cbb77dbbbaa7e424fd7cecd0fe7a9ca/processor/src/continuation_stack.rs#L109-L123) (and related call-context stack growth), which enables two distinct memory exhaustion patterns in attacker-controlled programs:
>
> 1. Dynamic invocation recursion that grows the continuation stack monotonically when recursion does not unwind.
> 2. External loading that grows the continuation stack and can retain many [MastForest](https://github.com/0xMiden/miden-vm/blob/089dc6a33cbb77dbbbaa7e424fd7cecd0fe7a9ca/core/src/mast/mod.rs#L95-L117) instances in deployments that resolve external forests on demand.

Note that we do not address (2), since this is solved in the `Host` implementation not present in this repository (i.e. the `Host` needs to have a limit on how many forests it keeps in memory, after which it either refuses to return new forests, or drops some old forests, or whatever other memory management strategy).

As for (1), we solved this problem by adding a check in the `FastProcessor` stoppers (i.e. `NeverStopper` and `StepStopper`), as we don't want to hardcode the "continuation stack length check" logic in the main execution loop. This is because this only makes sense when running untrusted programs, but we want to also give the ability for a processor that wants to run faster to omit these checks if they e.g. are only running trusted programs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.23.0 (TBD)
+
+#### Fixes
+- Set a bound on `ContinuationStack` size, checked during execution ([#2824](https://github.com/0xMiden/miden-vm/pull/2824)).
+
 ## 0.22.0 (TBD)
 
 #### Changes
--- a/processor/src/continuation_stack.rs
+++ b/processor/src/continuation_stack.rs
@@ -185,6 +185,11 @@ impl ContinuationStack {
     // PUBLIC ACCESSORS
     // --------------------------------------------------------------------------------------------
 
+    /// Returns the number of continuations on the stack.
+    pub fn len(&self) -> usize {
+        self.stack.len()
+    }
+
     /// Peeks at the next continuation to execute without removing it.
     ///
     /// Note that more than one continuation may execute in the same clock cycle. To get all
--- a/processor/src/execution/basic_block.rs
+++ b/processor/src/execution/basic_block.rs
@@ -49,6 +49,7 @@ where
         state.processor,
         state.tracer,
         state.stopper,
+        state.continuation_stack,
         || {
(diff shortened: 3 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.23.0 (TBD)
+
+#### Fixes
+- Set a bound on `ContinuationStack` size, checked during execution ([#2824](https://github.com/0xMiden/miden-vm/pull/2824)).
```
huitseeker, line 6: Ah. Food for thought for my own edits.

bobbinth:
These changes are not in `next` yet, right? Is the plan to merge them after v0.22 release?

huitseeker:
@bobbinth No, we only have the changes from https://github.com/0xMiden/miden-vm/pull/2823

The changes that are in the audit-fix branch but not in next/ are listed at:
https://github.com/0xMiden/miden-vm/compare/next...audit-fixes-2026

The plan for release is to ship whatever we have ready (through a merge), while continuing to work on the audit-fix branch for later, 3rd-party fix-review. We didn't plan for an additional merge of audit-fixes into next, but could do one relatively easily.

bobbinth:
> The changes that are in the audit-fix branch but not in next/ are listed at:
> [next...audit-fixes-2026](https://github.com/0xMiden/miden-vm/compare/next...audit-fixes-2026)

These are now in `next`.

PR #2836, opened 2026-03-11
Fix quote-equivalent path ambiguity in linking and library deserialization

huitseeker:
Quoted and unquoted spellings of the same path were treated as different in some places and the same in others. That could change call targets by link order, overwrite exports silently, or panic on bad quoted leaves.

This change makes deserialized paths canonical, rejects canonical export duplicates, validates procedure export leaves during deserialization, and rejects ambiguous equal-length module prefix matches during resolution.

### DIFF
--- a/crates/assembly-syntax/src/ast/item/resolver/symbol_table.rs
+++ b/crates/assembly-syntax/src/ast/item/resolver/symbol_table.rs
@@ -149,10 +149,13 @@ impl LocalSymbolTable {
                 LocalSymbol::Import { name, .. } => name.clone().into_inner(),
             };
 
-            if let Some(prev) = symbols.insert(name.clone(), id) {
-                panic!(
+            if let Some(prev) = symbols.get(&name).copied() {
+                debug_assert!(
+                    false,
                     "duplicate symbol '{name}' reached local resolver construction (previous={prev:?}, current={id:?})"
                 );
+            } else {
+                symbols.insert(name.clone(), id);
             }
             items.push(symbol);
         }
@@ -503,10 +506,37 @@ mod tests {
         }
     }
 
+    #[cfg(debug_assertions)]
     #[test]
     #[should_panic(expected = "duplicate symbol 'dup' reached local resolver construction")]
     fn local_symbol_table_rejects_duplicate_symbols() {
         let source_manager: Arc<dyn SourceManager> = Arc::new(DefaultSourceManager::default());
         let _table = LocalSymbolTable::new(DuplicateSymbolsForInvariantTest, source_manager);
+    }
+
+    #[test]
+    fn local_symbol_table_duplicate_symbols_have_explicit_behavior() {
+        use std::panic::{AssertUnwindSafe, catch_unwind};
+
+        let source_manager: Arc<dyn SourceManager> = Arc::new(DefaultSourceManager::default());
+        let result = catch_unwind(AssertUnwindSafe(|| {
+            LocalSymbolTable::new(DuplicateSymbolsForInvariantTest, source_manager)
+        }));
+
+        if cfg!(debug_assertions) {
+            assert!(
+                result.is_err(),
+                "debug builds should panic when duplicates reach local resolver construction"
+            );
+        } else {
+            let table = result.expect("release builds should not panic on duplicate symbols");
+            let resolved = table
+                .get(Span::unknown("dup"))
+                .expect("release behavior should keep a deterministic symbol mapping");
+            match resolved {
+                SymbolResolution::Local(id) => assert_eq!(id.into_inner(), ItemIndex::new(0)),
+                other => panic!("expected local symbol resolution, got {other:?}"),
+            }
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 - Hardened syscall target validation to avoid panic paths and reject invalid digests at assembly time ([#2804](https://github.com/0xMiden/miden-vm/pull/2804)).
 - Add bounds to attacker-controlled allocation sizes in advice map and keccak256/sha512 precompiles ([#2805](https://github.com/0xMiden/miden-vm/pull/2805)).
 - Prevented a trace-generation panic by validating op batch groups in basic blocks ([#2782](https://github.com/0xMiden/miden-vm/pull/2782)).
+- Fixed quote-equivalent path ambiguity in library deserialization and linker symbol resolution ([#2836](https://github.com/0xMiden/miden-vm/pull/2836)).
```
huitseeker: Note to self: this will be for 0.23.

plafer:
Note: let's use `audit-fixes-2026` moving forward (I already started doing that with #2825) - probably best to remove the `speculative-next` one as well

CHANGELOG.md
plafer: Now that #2825 is merged, if you rebase against `audit-fixes-2026`, you can move it directly in the new `0.23` section

crates/assembly/src/tests.rs
```diff
@@ -5209,6 +5212,137 @@ fn test_cross_module_quoted_identifier_resolution() -> TestResult {
...
+        r#"
+pub proc add
+    add.1
+end
+"#,
+    );
+
+    let attacker_mod = parse_library_module(
+        source_manager.clone(),
+        r#"::foo::"bar""#,
```
bitwalker: I think this test is not actually covering what we want it to cover. Specifically, it is entirely valid to quote path components that do not need to be quoted, but the assembler should be normalizing and validating these paths before using them anywhere (i.e. only normalized paths should appear in the exports of an assembled library).

So what we actually want to test here is that assembling fails not due to ambiguous paths, but due to conflicting/duplicate paths (i.e. adding the same module namespace twice, which we cannot handle until we start assembling from projects where we can disambiguate based on specific dependency versions).

huitseeker: Right, but shifting the test expectation requires canonicalizing eagerly. I changed things so 
- exported paths are canonicalized before they enter library exports, and 
- removed order-dependent behavior by rejecting duplicate canonical exports instead of silently overwriting, and 
- added coverage for both cross-library and single-library collision cases.

crates/assembly/src/tests.rs
```diff
@@ -5209,6 +5212,160 @@ fn test_cross_module_quoted_identifier_resolution() -> TestResult {
     Ok(())
 }
 
+#[test]
+fn regression_symbol_resolution_duplicate_module_paths_are_rejected_during_linking() {
+    use crate::{Parse, ParseOptions};
+
+    fn parse_library_module(
```
bitwalker: This functionality is already provided by `TestContext` IIRC - but if not, we should add it there so it can be used for other tests

crates/assembly-syntax/src/ast/item/resolver/symbol_table.rs
```diff
@@ -135,6 +140,7 @@ impl LocalSymbolTable {
         S: SymbolTable,
     {
         let mut symbols = BTreeMap::default();
+        let mut duplicate_symbols = BTreeSet::default();
```
bitwalker: When we perform semantic analysis on modules after parsing, we catch duplicate definitions at that point. Are we failing to catch conflicting symbols there? It isn't clear to me how duplicates could still be present by the time we construct a resolver.

crates/assembly/src/tests.rs
huitseeker: Thanks!

crates/assembly-syntax/src/ast/item/resolver/symbol_table.rs
huitseeker: I have added two tests in the latest commit [62553d4](https://github.com/0xMiden/miden-vm/pull/2836/commits/62553d411532ca2ac693ee939f5a8438981dc210) that demonstrate the ingress behavior, passing sema, and hitting the deduplication here. 

I'd be ok with a determination that we maybe dislike the set of symbols against which `define_(type|constant)` resolve[^1], but would that be the case, I suggest we open a new issue to address this (and tack on the sub-task of mopping up the late-deduplication in the symbol table if we so wish), so as to not let the scope of this PR creep (it will receive a review focused on addressing an audit finding, not revamping sema).

[^1]: I'm not quite sure I do just now, given the perf & complexity implications of having full-set "independent" deduplication in all of `define_alias`, `define_procedure`, `define_constant`, and `define_type`.

huitseeker: Actually, never mind, this was easier than I thought, I have added a commit that fixes this!

Issue #2826, opened 2026-03-11
"Trace with id ..." printing is too noisy


greenhat:
I think it originates from https://github.com/search?q=repo%3A0xMiden%2Fmiden-vm+"Trace+with+id"&type=code and I cannot find any way of silencing it when using `Mockchain` from the `miden-client`.

plafer:
Oh wow that does look obnoxious. Will fix today so that it get included in 0.22.

greenhat:
Thank you for the swift response!

PR #2837, opened 2026-03-11
chore(processor): make DebugHandler default method implementations no…

plafer:
Closes https://github.com/0xMiden/miden-vm/issues/2826

Removes prints from `DebugHandler` trait default method implementations, as well as makes `DebugHandler::on_debug()` a no-op. The idea being that the trait doesn't need to guess what implementations need - especially prints which can get obnoxious.


huitseeker (approved):
<3

PR #2838, opened 2026-03-11
fix: reject @locals values that overflow word-alignment

### LINKED ISSUE #2829: @locals(65535) alignment overflow crashes assembler
In `crates/assembly-syntax/src/parser/intel/mod.rs`, the `@locals` directive parses the count as a `u16` and computes the required alignment. When `@locals(65535)` is specified, the alignment calculation overflows:

```rust
let alignment = (count * 8 + 15) / 16; // overflow when count = 65535
```

This causes a panic during assembly. While 65535 locals is unrealistic, the parser should reject this gracefully rather than panic.

Fix: add a bounds check before the alignment calculation, returning a descriptive error for counts that would overflow.

Acelogic:
## Summary

- `@locals(65535)` (and 65533, 65534) causes a panic because `num_locals.next_multiple_of(4)` overflows `u16`
- Adds a bounds check in the parser grammar that rejects `@locals` values above 65532 with a descriptive error
- 65532 is the largest `u16` value that can be safely word-aligned (rounded up to a multiple of 4)

## Details

The overflow occurs in `fmp_start_frame_sequence` and `fmp_end_frame_sequence` (`crates/assembly/src/fmp.rs`) at:
```rust
let locals_frame = Felt::from_u16(num_locals.next_multiple_of(WORD_SIZE as u16));
```

For `num_locals = 65535`, `next_multiple_of(4)` = 65536, which exceeds `u16::MAX`. In debug builds this panics; in release builds it wraps to 0.

The fix validates at parse time using the existing `InvalidLocalsAttr` error variant, so the user gets a clear diagnostic:
```
error: invalid procedure @locals attribute
  --> test:2:5
  |
2 |     @locals(65535)
  |     ^^^^^^^^^^^^^^ number of locals exceeds the maximum of 65532
```

## Test plan

- [x] Added `locals_overflow_rejected` test: verifies `@locals(65535)` produces the expected error
- [x] Added `locals_max_valid_accepted` test: verifies `@locals(65532)` is accepted
- [x] All 86 existing `miden-assembly-syntax` tests pass

Fixes #2829


bitwalker:
You'll need to sign your commits and rebase on `next`, but otherwise should be good! Be sure to read our contributing guidelines for information on what we expect from PRs from external contributors though, it saves everyone a lot of time if little things like commit signing are already sorted.

bobbinth (approved):
Looks good! Thank you!

PR #2839, opened 2026-03-11
Harden basic-block batch invariants and padding decode validation

huitseeker:
- Enforces stronger BasicBlockNode batch invariants for untrusted/decoded data:
- Hardens deserialization in BasicBlockDataDecoder::decode_operations() by rejecting inconsistent padding metadata early.
      
Net effect: helper APIs that operate on raw/padded views no longer hit underflow/panic paths on validated-but-malformed forests.

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -9,7 +9,7 @@ use crate::{
         DecoratedLinksIter, DecoratedOpLink, DecoratorId, DecoratorStore, MastForest,
         MastForestError, MastNode, MastNodeFingerprint, MastNodeId,
     },
-    operations::{DecoratorList, Operation, is_control_flow_opcode},
+    operations::{DecoratorList, Operation},
     prettier::PrettyPrint,
     utils::LookupByIdx,
 };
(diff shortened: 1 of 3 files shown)


core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,24 +442,60 @@ impl BasicBlockNode {
...
+    /// 3. No operation group ends with an operation requiring an immediate value
+    /// 4. The last operation group in a batch cannot contain operations requiring immediate values
+    /// 5. OpBatch structural consistency (num_groups <= BATCH_SIZE, group size <= GROUP_SIZE)
+    /// 6. Immediate values are committed to empty groups and match group contents
+    /// 7. OpBatch padding semantics (no padding on empty groups; padded groups end with NOOP)
     ///
     /// Returns an error string describing which invariant was violated if validation fails.
     pub fn validate_batch_invariants(&self) -> Result<(), String> {
         // Check invariant 1: Power-of-two groups in each batch
         self.validate_power_of_two_groups()?;
```
plafer, line 454: We could bound the number of batches up to some reasonable upper bound, to make sure this doesn't loop for a very long time.

Or, if this is better handled elsewhere (e.g. deserialization), then we could document that we this will not  be checked.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,24 +442,60 @@ impl BasicBlockNode {
...
+            for (op_idx, op) in batch.ops().iter().enumerate() {
+                let opcode = op.op_code();
+                Self::validate_control_flow_opcode(opcode, batch_idx, op_idx)?;
+            }
+        }
+        Ok(())
+    }
+
+    #[inline]
+    fn validate_control_flow_opcode(
```
plafer:
```suggestion
    fn ensure_not_control_flow_opcode(
```
The current name suggests that we are making sure that we have a control flow opcode

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,24 +442,60 @@ impl BasicBlockNode {
...
+        // Check invariant 7: OpBatch padding semantics
+        self.validate_padding_semantics()?;
+
+        Ok(())
+    }
+
+    /// Validates that no batch contains control-flow operations.
+    fn validate_no_control_flow_ops(&self) -> Result<(), String> {
+        for (batch_idx, batch) in self.op_batches.iter().enumerate() {
+            for (op_idx, op) in batch.ops().iter().enumerate() {
```
plafer: Similarly to the number of batches, we could either bound the number of ops be batch, or write in the preconditions that we expect this to already be checked

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -672,6 +708,54 @@ impl BasicBlockNode {
...
+                if !is_padded {
+                    continue;
+                }
+
+                if group_start == group_end {
+                    return Err(format!(
+                        "Batch {}, group {}: empty group cannot be marked as padded",
+                        batch_idx, group_idx
+                    ));
+                }
```
plafer: This is not allowed by `validate_batch_structure()` (lines 601-612) right? If this is meant to be defensive, I would leave a comment that this should already have been rejected

core/src/mast/node/basic_block_node/tests.rs
```diff
@@ -480,6 +509,92 @@ fn test_validate_immediate_commitment_rejects_nonzero_empty_group() {
...
+    let result = block.validate_batch_invariants();
+
+    assert!(result.is_ok());
+}
+
+#[test]
+fn validate_batch_invariants_rejects_malformed_indptr_without_panicking() {
+    let ops = vec![Operation::Add];
+    let mut indptr = [0usize; BATCH_SIZE + 1];
+    indptr[1] = 2;
```
plafer, line 545: Do we have a test somewhere that checks that `validate_batch_structure()` rejects `indptr[i] == indptr[i+1]`? e.g. just passing `indptr = [0usize; BATCH_SIZE + 1];` should fail (related to previous comments)

core/src/mast/serialization/basic_blocks.rs
```diff
@@ -270,6 +270,38 @@ impl BasicBlockDataDecoder<'_> {
...
+                    ))
+                })?;
+
+                if *last_op != Operation::Noop {
+                    return Err(DeserializationError::InvalidValue(format!(
+                        "batch {} group {}: padded group must end with NOOP",
+                        batch_idx, group_idx
+                    )));
+                }
+            }
```
plafer: Instead of copy/pasting the `BasicBlockNode::validate_padding_semantics()`, would it not make sense to instead construct the `OpBatch`, and then call `OpBatch::validate_padding_semantics()` (a new method extracted from the current `BasicBlockNode` impl)? Or even a more general `OpBatch::validate()`, which does all the checks that are now written only in `BasicBlockNode::validate()`

huitseeker: Yep!

core/src/mast/node/basic_block_node/tests.rs
huitseeker: `indptr[i] == indptr[i+1]` indicates an empty group. That's valid for the purpose of storing an immediate push value. `indptr = [0usize; BATCH_SIZE + 1]` should fail, though, at minimum because the batch can't end on an immediate.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: This is better handled elsewhere. The job of deserialization is to ensure you can't deserialize a data structure of size N without sending a number of bytes $O(N)$ with non-trivial constants, and the role of the protocol on top of that is to make sure the size of any received message is bounded.

core/src/mast/node/basic_block_node/tests.rs
plafer: Ah, I misread the check in `validate_batch_structure()` - disregard

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -442,24 +442,60 @@ impl BasicBlockNode {
...
-        self.validate_no_immediate_endings()?;
-
-        // Check invariant 3: OpBatch structural consistency
+        // Check invariant 5: OpBatch structural consistency
+        // This needs to be done early on as it will validate indptr indexes used in later checks.
         self.validate_batch_structure()?;
 
-        // Check invariant 4: Immediate values must be committed to empty groups
+        // Check invariant 2: No control-flow operations in batches
+        self.validate_no_control_flow_ops()?;
```
bobbinth: This is no longer a needed check in the latest `next`, right?

bobbinth (approved):
Looks good! Thank you! I left one comment inline (about the fixes here being a bit out of sync with `next`).

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Ah, yes, that's right, because of #2771 (which should probably have been fixed in the audit branch, because I fear we'll forget to take it into account in review). I added a commit that removes this.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Done with the latest commit

PR #2841, opened 2026-03-12
ci: Opt in to custom clippy lints 

huitseeker:
This introduces locally-runnable `cargo`commands `xclippy` and `xclippy-fix` that opt into additional default-allowed lints, and integrates this in CI so that `make lint` enforces these custom lints.

*No lint is introduced that the code base doesn't already pass*, the PR simply locks into CI what behavior is already correct in our code, preventing regressions.

A list of *commented* lints which do *not* pass is provided with hit counts in the current state, hinting at future arguably useful additions. At this stage, they're just informational.

This is motivated by some of the audit feedback (some of our issues would have been fixed with some of these turned on), and the future lints I have in my sights are those related to numerical behaviors (overflows, casts).

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,4 @@
 [alias]
-xtask = "run --bin xtask --"
-
 # Collection of project wide clippy lints. This is done via an alias because
 # clippy doesn't currently allow for specifiying project-wide lints in a
 # configuration file. This is a similar workaround to the ones presented here:


.cargo/config.toml
```diff
@@ -0,0 +1,140 @@
+[alias]
+xtask = "run --bin xtask --"
```
plafer: this is unused, and doesn't work, right? Running `cargo xtask` fails with "error: no bin target named `xtask` in default-run packages". Looking at the docs from [cargo-xtask](https://github.com/matklad/cargo-xtask), this would expect us to have an `xtask` package in the root

PR #2843, opened 2026-03-12
fix: replace unwrap() with ? in recursive calls in decorator_storage and debug_var_storage

### LINKED ISSUE #2785: unwrap() instead of ? in recursive calls in decorator_storage and debug_var_storage

johnh4098:
Closes #2785

Both `decorator_storage.rs` and `debug_var_storage.rs` use `.unwrap()` on a recursive `Result`-returning call inside a `for` loop. Since the enclosing function already returns `Result`, replacing `.unwrap()` with `?` is the correct and idiomatic fix — errors propagate to the caller instead of panicking.

### DIFF
--- a/crates/assembly-syntax/src/ast/tests.rs
+++ b/crates/assembly-syntax/src/ast/tests.rs
@@ -962,6 +962,37 @@ fn test_ast_parsing_simple_docs() -> Result<(), Report> {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #2846, opened 2026-03-13
refactor: add debug constraint checking and migrate tests from prove_and_verify

adr1anh:
Add `ExecutionTrace::check_constraints()` which validates AIR constraints row-by-row without generating a STARK proof, providing faster feedback and better error diagnostics for tests.

- Add `check_constraints()` to `ExecutionTrace` using the lifted-stark's `debug::check_constraints` with Poseidon2-derived deterministic challenges
- Add `to_row_major_matrix()` to `ExecutionTrace`, replacing the standalone `execution_trace_to_row_major` in the prover's trace_adapter module
- Re-export `AirWitness` and `debug` from `miden-air` so processor does not need a direct `miden-crypto` dependency
- Add `Test::check_constraints()` in test-utils for convenient test usage
- Remove `prover/src/trace_adapter.rs` (dead `aux_trace_to_row_major` and now-inlined `execution_trace_to_row_major`)
- Migrate ~45 test call sites from `prove_and_verify` to `check_constraints`

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -63,7 +63,7 @@ impl ExecutionOptions {
     /// # Errors
     /// Returns an error if:
     /// - `max_cycles` is outside the valid range
-    /// - `expected_cycles` exceeds `max_cycles`
+    /// - after rounding up to the next power of two, `expected_cycles` exceeds `max_cycles`
     /// - `core_trace_fragment_size` is zero
     pub fn new(
         max_cycles: Option<u32>,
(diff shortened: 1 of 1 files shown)


processor/src/trace/mod.rs
```diff
@@ -207,6 +207,54 @@ impl ExecutionTrace {
...
+        let witness = AirWitness::new(&trace_matrix, &public_values, var_len_public_inputs);
+        debug::check_constraints(&ProcessorAir, witness, &aux_builder, &challenges);
+    }
+
+    /// Converts the main trace from column-major to row-major format.
+    ///
+    /// Only includes the first [`TRACE_WIDTH`] columns (excluding padding columns added for
+    /// Poseidon2 rate alignment), which is the width expected by the AIR.
+    // TODO: the padding columns can be removed once we use the lifted-stark's virtual trace
+    // alignment, which pads to the required rate width without materializing extra columns.
```
adr1anh, line 247: Lifted Stark ensures that the proofs are aligned to the hasher's rate, so padding is no longer necessary.

processor/src/trace/mod.rs
```diff
@@ -207,6 +207,54 @@ impl ExecutionTrace {
...
+    pub fn to_row_major_matrix(&self) -> RowMajorMatrix<Felt> {
+        let trace_len = self.get_trace_len();
+        let mut col_major_data = Vec::with_capacity(TRACE_WIDTH * trace_len);
+        for col_idx in 0..TRACE_WIDTH {
+            col_major_data.extend_from_slice(self.main_trace.get_column(col_idx));
+        }
+        let col_major_matrix = RowMajorMatrix::new(col_major_data, trace_len);
+        col_major_matrix.transpose()
+    }
```
bobbinth, line 256: How does this compare in terms of performance to the code we had before?

adr1anh: This was the code we had before, just moved to a more appropriate place. The inner transpose for Matrix is optimized in plonky so I'm not sure there much more we can do. This should go away once we have row-major trace gen.

Issue #2621, opened 2026-01-31
Specify calling conventions/ABIs supported by Miden Assembly


bitwalker:
This corresponds to the `@callconv` attribute, and the relationship between how procedures defined in Miden Assembly get surfaced in higher-level programming languages, and vice versa.

## Background

An ABI (i.e. Application Binary Interface) dictates at a minimum:

* The type system of the abstract machine
* The representation of values in that type system in memory, on the operand stack, and encoded in the advice provider (including alignment and padding rules)

A  calling convention dictates at a minimum:

* How a procedure/function is called (of relevance here, the instruction that must be used, i.e. `syscall` or `call`/`dyncall` or `exec`/`dynexec`)
* The ABI used for specifying arguments and results
* The exact mechanics of how arguments are passed, and results returned, i.e. passed by-value on the operand stack, by-reference using local or heap memory, or non-deterministically using the advice provider.

Combined, these two aspects provide all of the information needed to call a procedure written in one language, from another, so long as both languages understand the calling convention and ABI.

## Current Status

Currently, we've got the following pieces:

* A low-level type system sufficient to define ABI details across multiple languages, with Miden-specific types like `felt`
* A specification of how that type system is laid out in byte-addressable memory, how byte-addressable memory maps to Miden's element-addressable memory
* A convention that dictates that layout of values on the operand stack follows the in-memory layout rules mentioned in the previous bullet point
* An informal limit on the maximum size of a by-value argument/result (i.e. a Miden "word", or four field elements - equivalent to 16 bytes/128 bits), in other words, the maximum size of a single logical value on the operand stack.
* A hard limit on the number of by-value arguments/results when using the `syscall`/`call`/`dyncall` instructions, i.e. whatever is representable using 16 elements on the operand stack. Anything more than this dictates that the advice provider is used to encode the overflow, which is then decoded on the other side.

On top of these things, we need to encode the following information:

* The WebAssembly Canonical ABI - used for all component model exports/imports for languages compiled to Wasm (i.e. Rust). Calling these exports from Miden Assembly requires the MASM caller to adhere to the Wasm Canonical ABI. The type system of the Canonical ABI is a superset of our own type system, but can be represented using the latter. This consists of really two calling conventions, one for Core Wasm functions that are "lifted" into the Canonical ABI, and one for Canonical ABI functions that are "lowered" to the Core Wasm representation.
[...]

PR #2848, opened 2026-03-13
feat(assembly): migrate to latest midenc-hir-type release

bitwalker:
The latest release includes the following changes/improvements:

* Structs can now be named (new) or anonymous (existing behavior)
* Enum types are now representable in the midenc-hir-type system
* Calling conventions were cleaned up:
  - The old `kernel` and `canon-lower` conventions are removed. 
  - The `canon-lift` convention was replaced by `component-model`. This is largely just a rename, but the semantics were clarified too.
  - `CallConv::SystemV` was renamed `CallConv::C`, with its semantics modified to reflect that the C ABI follows the Wasm C ABI, rather than x86 System-V.

This commit integrates those changes and does the following:

* Adds support to the AST for enum types with heterogeneous variants, but does not implement support for parsing such enums yet.
* Modifies the way enums are defined in a Module to export the enum type itself, rather than its discriminant type.
* Modifies type resolution to resolve enum types as an enum type, not the discriminant type.
* Modifies the Visit and VisitMut trait to visit enum discriminant values and payload types. There was actually a bug here where we were not visiting the discriminant values.

---

This change largely addresses #2621, by removing unnecessary `CallConv` variants, and clarifying the semantics of the existing variants. See [here](https://docs.rs/midenc-hir-type/latest/midenc_hir_type/enum.CallConv.html) for details.

Additionally, it provides support for representing the full variety of types (aside from resources) specified by the Wasm Canonical ABI, namely `enum` types. The MASM enum type is a narrower version of enums (i.e. they are C-like enums, where the value of the enum is equivalent to the discriminant type), while the Canonical ABI supports enums with optional payloads associated with each variant. In this PR, we add support to the AST and assembler for both styles of enum, but do not support parsing enums with heterogeneous variants yet. Supporting them isn't that difficult from a type declaration perspective, but the MASM `enum` form currently has a dual role in that it defines both a type and a set of constants corresponding to each variant - we'd need to change the semantics of `enum` so that when it is used as a declaration of a type with variant payloads, we do not define any constants (since the value of any given variant cannot be known statically).

Lastly, this allows struct types to have inherent names, which the parser now infers based on the context (i.e. it will propagate the name `Foo` in `type Foo = struct { .. }` to the struct type on the right-hand side. Currently we don't render that name anywhere, or parse it, i.e. `struct Foo` isn't a thing, but it aligns closer with how struct types are defined in `midenc-hir-type`.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -51,6 +51,6 @@ jobs:
       - name: Run fuzz target (smoke test)
         run: |
           # Build the fuzz target first
-          cargo fuzz build --fuzz-dir miden-core-fuzz ${{ matrix.target }}
+          cargo +nightly fuzz build --fuzz-dir miden-core-fuzz ${{ matrix.target }}
           # Run directly to avoid cargo-fuzz wrapper SIGPIPE issue
           miden-core-fuzz/target/x86_64-unknown-linux-gnu/release/${{ matrix.target }} -max_total_time=60 -runs=10000


crates/assembly/src/linker/resolver/mod.rs
```diff
@@ -160,13 +163,55 @@ impl<'a, 'b: 'a> ast::TypeResolver<LinkerError> for Resolver<'a, 'b> {
     }
 
     fn get_type(
-        &self,
+        &mut self,
         context: SourceSpan,
         gid: GlobalItemIndex,
     ) -> Result<types::Type, LinkerError> {
```
bobbinth, line 169: It feels a bit counter-intuitive that getting a type would need to mutate something. Let's add a brief doc comment explaining why mutations are needed here (and what's getting mutated).

crates/assembly/src/linker/resolver/mod.rs
```diff
@@ -160,13 +163,55 @@ impl<'a, 'b: 'a> ast::TypeResolver<LinkerError> for Resolver<'a, 'b> {
...
         match self.resolver.linker()[gid].item() {
             SymbolItem::Compiled(ItemInfo::Type(info)) => Ok(info.ty.clone()),
-            SymbolItem::Type(ast::TypeDecl::Enum(ty)) => Ok(ty.ty().clone()),
+            SymbolItem::Type(ast::TypeDecl::Enum(ty)) => {
+                let mut variants = SmallVec::<[types::Variant; 4]>::new_const();
```
bobbinth, line 179: Could we add a comment here explaining what the code below does? IIUC, we now iterate over the variants of the enum and perform constant evaluation for each variant. Is this because enum variants can contain expressions now?

crates/assembly-syntax/src/library/mod.rs
```diff
@@ -964,6 +976,36 @@ impl Serializable for TypeSerializer<'_> {
                 target.write_u8(20);
                 FunctionTypeSerializer(ty).write_into(target);
             },
+            Type::Enum(ty) => {
+                target.write_u8(21);
+                target.write_usize(ty.name().len());
+                target.write_bytes(ty.name().as_bytes());
+                TypeSerializer(ty.discriminant()).write_into(target);
+                target.write_usize(ty.variants().len());
```
bobbinth: Having this code here makes it quite fragile. Ideally, the serialization code should live next to the type itself - or we should have extensive tests covering this (a lost of the bugs from the audit were serialization-related).

I think the bigger question is whether we want to use the old Winterfell-style serialization for anything that's not `MastForest` (and maybe a few other types). And if we don't, how can we make sure we protect against adversarially serialize data.

cc @huitseeker.

crates/assembly/src/linker/resolver/mod.rs
bitwalker: Sure, these methods should have always been `mut`, but it previously hadn't been needed, so they were left immutable. Resolution can require the resolver to mutate its own state (in this case, it mutates the resolver cache).

I'll add a note to the trait about this.

crates/assembly/src/linker/resolver/mod.rs
bitwalker: Sure!

Enum variants always carried a discriminant value, but since we were previously never exporting `EnumType` declarations, and thus ignoring the discriminant value associated with `EnumType` in favor of the constants we derived from it during semantic analysis, we did not previously need to evaluate the discriminants during link-time type resolution.

Now, however, we are exporting the actual enum type, and so we need to resolve all of its component parts (here, the discriminant type, and the discriminant values and payload type for each variant).

crates/assembly-syntax/src/library/mod.rs
bitwalker: It is, and I would _love_ to move this serialization handling to `midenc-hir-type` - but I don't want a dependency on `miden-crypto` to do it. IMO, we should move the serialization stuff to a new crate that is standalone and can be maintained separately from `miden-crypto`.

bobbinth: It is in a separate crate now - [miden-serde-utils](https://github.com/0xMiden/crypto/tree/next/miden-serde-utils) - though the crate still lives in the `miden-crypto` repo.

bitwalker: I'll move this code into the type crate then before we merge this

crates/assembly-syntax/src/library/mod.rs
```diff
@@ -948,6 +953,13 @@ impl Serializable for TypeSerializer<'_> {
                 }
                 target.write_u8(ty.len() as u8);
                 for field in ty.fields() {
+                    if let Some(name) = ty.name() {
```
huitseeker: Should this read from `field.name` (or the field-name accessor) instead? Right now a named struct would write the struct name for every field, so field metadata changes after a serialize/deserialize roundtrip.

crates/assembly-syntax/src/library/mod.rs
```diff
@@ -1075,6 +1128,48 @@ impl TypeDeserializer {
             20 => Type::Function(Arc::new(
                 FunctionTypeDeserializer::read_from_with_depth(source, next_depth)?.0,
             )),
+            21 => {
+                use crate::ast::types::{InvalidEnumTypeError, Variant};
+
+                let name = Arc::<str>::from(String::read_from(source)?.into_boxed_str());
+                let discriminant = TypeDeserializer::read_from(source)?.0;
```
huitseeker: Can we keep depth accounting here by calling `read_from_with_depth(source, next_depth)`? Using `TypeDeserializer::read_from(source)` resets depth, and the same pattern appears for `value_ty` below. That looks like a way for nested enum payloads to bypass the recursion limit introduced in #2777 .

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -280,7 +282,12 @@ impl Module {
             // Validate that the discriminant is a valid instance of the `repr` type
             variant.assert_instance_of(&repr)?;
 
-            let Variant { span, docs, name, discriminant } = variant;
+            let Variant { span, docs, name, value_ty, discriminant } = variant;
+
+            assert!(
```
huitseeker: Should this panic instead of giving a normal diagnostic if an enum with payload variants is built through the AST API? This is close to top-level, so I expect it not to be too much work to percolate the failure.

huitseeker (changes requested):
I think there's a typo in serialization, and this breaks the depth-limit of #2777 .

crates/assembly-syntax/src/ast/module.rs
bitwalker: There's no reason we need to panic here at all actually - I've modified this function to skip emitting constants for non-C-like enums, which is the behavior we'll have anyway when we get around to adding syntax support for them.

bitwalker:
I've updated this PR now that the serialization code has been moved into `midenc-hir-type`, addressing @huitseeker's review comments in the process. This should be ready to merge now.

crates/assembly-syntax/Cargo.toml
```diff
@@ -41,7 +40,7 @@ logging = ["dep:env_logger"]
 miden-core.workspace = true
 miden-debug-types.workspace = true
 miden-utils-diagnostics.workspace = true
-midenc-hir-type.workspace = true
+midenc-hir-type = { workspace = true, features = ["serde"] }
```
bobbinth, line 43: Question: previously, the `serde` feature was enabled for `midenc-hir-type` only when `serde` feature for the crate was enabled - but now it seems to be enabled unconditionally - why the change?

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -291,7 +305,7 @@ impl Module {
             })?;
         }
 
-        self.items.push(Export::Type(alias.into()));
+        self.items.push(Export::Type(export.into()));
```
bobbinth, line 308: Was this a bug previously?

bobbinth (approved):
Looks good! Thank you! I left just a couple of questions inline.

crates/assembly-syntax/Cargo.toml
bitwalker: The `serde` feature for `midenc-hir-type` covers all serialization-related dependencies and code, previously we only had `serde`-based serialization implemented there, so we only needed the feature when `serde` was enabled here. Since `miden-serde-utils`-based serialization is not gated behind the `serde` feature here, we must always require the `serde` feature for `midenc-hir-type` for now.

crates/assembly-syntax/src/ast/module.rs
bitwalker: No, previously we were exporting the discriminant type of an enum as the type of that enum - now, we're exporting the full enum type. We need to do this for two reasons:

1. We want the actual enum type to be exported so that bindings generated in, e.g. Rust, can emit an equivalent enum type in that language.
2. The Wasm Canonical ABI supports enum variants with payloads, which means that the type of such enums is _not_ equivalent to the type of the discriminant. It is necessary to export the full enum type so we can determine what the correct type should be in interfaces which use it.

As mentioned elsewhere, we don't yet support expressing variants with payloads in MASM yet, but that's something we can do in a subsequent release.

bobbinth:
I'll merge this as is - and we can address any outstanding issues in follow-ups.

PR #2849, opened 2026-03-14
Harden kernel deserialization invariants and expand serde deserialization fuzz coverage

huitseeker:
Closes a kernel deserialization invariant gap and broadens fuzzing coverage for serde paths.

- Prevents non-canonical kernels (duplicates, non-set representations, oversized serde inputs) from being accepted and later causing invariant-dependent failures.
- Adds serde JSON fuzz targets for all deserialization surfaces that implement `Deserialize`: `Program`, `Kernel`, `MastForest`, `AdviceMap`, `Operation`, `ExecutionProof`, `PrecompileRequest`, `Library`, and `Package`.
- Expands fuzz seed generation (including malformed kernel/program cases),


core/src/program/kernel.rs
```diff
@@ -28,20 +28,7 @@ impl Kernel {
 
     /// Returns a new [Kernel] instantiated with the specified procedure hashes.
     pub fn new(proc_hashes: &[Word]) -> Result<Self, KernelError> {
```
bobbinth, line 37: Would be good to expand doc comments with description of error conditions.

core/src/program/kernel.rs
```diff
@@ -62,6 +49,21 @@ impl Kernel {
     pub fn proc_hashes(&self) -> &[Word] {
         &self.0
     }
+
+    fn from_hashes(mut hashes: Vec<Word>) -> Result<Self, KernelError> {
```
bobbinth: Let's add brief doc comments here.

Also, it may make senes to make this into a public constructor (and if we do so, I'd move this up (e.g., right after line 32).

core/src/program/kernel.rs
huitseeker: Good call, thanks. I added explicit `# Errors` docs on `Kernel::new` covering duplicate hashes and the max-procedure bound.

core/src/program/kernel.rs
huitseeker: I added explicit error-condition docs, made Kernel::from_hashes a public constructor, and moved it right after Kernel::new.”

core/src/mast/serialization/seed_gen.rs
```diff
@@ -93,10 +93,104 @@ fn generate_fuzz_seeds() {
...
+            .unwrap();
+        forest.make_root(block_id);
+
+        let a: Word = [Felt::new(9), Felt::new(10), Felt::new(11), Felt::new(12)].into();
+
+        let mut invalid_program = Vec::new();
+        forest.write_into(&mut invalid_program);
+        invalid_program.write_u8(2);
+        a.write_into(&mut invalid_program);
+        a.write_into(&mut invalid_program);
```
plafer: This assumes the current `Kernel` serialization format - which if it ever (breaking) changes, this will be out of date.

How about we add test-only constructors that don't check the invariants for `Kernel`, and then here actually create a `Program` object and serialize it?

This would apply to everywhere else where with this pattern

miden-core-fuzz/fuzz_targets/kernel_serde_deserialize.rs
plafer, line 1: I ran some fuzzing locally, and the `miden-core-fuzz/target/` directory got created, showing thousands of files to be added to git.

Probably worth updating the `.gitignore` to include this `target/` directory so that we don't have to manually remove it after every time we run the fuzzer?

Makefile
```diff
@@ -257,16 +257,25 @@ fuzz-mast-validate: ## Run fuzzing for UntrustedMastForest validation
 .PHONY: fuzz-all
 fuzz-all: ## Run all fuzz targets (in sequence)
 	-@cargo +nightly fuzz run mast_forest_deserialize --release --fuzz-dir miden-core-fuzz -- -max_total_time=300
```
plafer: It's probably worth adding `fuzz-seeds` as a dependency to make sure it runs before this invocation? It's almost instantaneous, makes sure you actually have the latest seeds (if any at all) to make sure that this long fuzzing session is actually looking for useful inputs.

I'd probably do this for the other fuzzing commands too

core/src/mast/serialization/seed_gen.rs
```diff
@@ -93,10 +93,104 @@ fn generate_fuzz_seeds() {
...
+        duplicate_kernel.write_u8(3);
+        b.write_into(&mut duplicate_kernel);
+        a.write_into(&mut duplicate_kernel);
+        a.write_into(&mut duplicate_kernel);
+        write_seed("kernel_deserialize", "duplicate_kernel.bin", &duplicate_kernel);
+
+        // Serde kernel seeds (JSON payloads) used by kernel_serde_deserialize fuzz target.
+        write_seed("kernel_serde_deserialize", "empty_kernel.json", b"[]");
+        write_seed("kernel_serde_deserialize", "duplicate_kernel.json", b"[[1,2,3,4],[1,2,3,4]]");
+        write_seed("kernel_serde_deserialize", "too_many_hashes.json", br#"[
```
plafer: We already have `serde-json` as a dev dependency - any reason to inline this big array string rather than generating it programmatically and serializing it into json?

huitseeker: replaced the large inline serde JSON seed with programmatic generation!

miden-core-fuzz/fuzz_targets/kernel_serde_deserialize.rs
huitseeker: added miden-core-fuzz/target/ to .gitignore

Makefile
huitseeker: made fuzz entry targets depend on fuzz-seeds

core/src/mast/serialization/seed_gen.rs
huitseeker: switched malformed kernel/program seeds to object-level serialization via a test-only unchecked kernel constructor

PR #2851, opened 2026-03-16
feat(project): implement assembly of projects

bitwalker:
**NOTE:** This is a draft branch implementing the final aspects of project handling needed for current use cases:

* New `Assembler` APIs for configuring an instance of the assembler for a project, and assembling one of its targets to a Miden package
* Tracking of debug information during assembly needed to derive the three debug info sections defined in `miden_mast_package` (debug sources, debug types, debug functions).
* Modifications to the package index/resolver to support:
  * Deriving the appropriate linkage type for a resolved dependency
  * Seeding the package index based on project dependencies referenced via `path`
* Modifications to the linker to stop using `Library` for referencing assembled artifacts, to support linking against assembled packages
* New testing utilities for assembling projects
* Deprecate old `Assembler` APIs that work in terms of `Library`
* `miden-core-lib` was ported to use Miden projects
* Integration test that covers all of the new functionality, including depending on the core lib

This won't be ready to merge until:

- [ ] Implementation has been cleaned up, in particular the details around the package index, registry, and resolver, as these are a bit poorly defined at the moment.
- [ ] Irrelevant implementation details that aren't needed anymore are removed
- [ ] Additional test coverage
- [ ] Documentation


bitwalker:
I'm reworking the approach here, I'll open a new PR when ready.

PR #2855, opened 2026-03-16
ci: enable more clippy lints

huitseeker:
A continuation of #2841. Opts the code base in to 7 clippy lints: 
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#derive_partial_eq_without_eq
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#explicit_deref_methods
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#inefficient_to_string
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#needless_continue
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#rest_pat_in_fully_bound_structs
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#todo
- https://rust-lang.github.io/rust-clippy/stable/index.html?levels=allow#unnested_or_patterns

As well as compiler lints:
- unused_lifetimes
- trivial_numeric_casts

Corrects the few incidences of these lints in the code base.

### DIFF
--- a/crates/assembly-syntax/src/ast/type.rs
+++ b/crates/assembly-syntax/src/ast/type.rs
@@ -24,19 +24,23 @@ const MAX_TYPE_EXPR_NESTING: usize = 256;
 ///
 /// Since type resolution happens in two different contexts during assembly, this abstraction allows
 /// us to share more of the resolution logic in both places.
+///
+/// NOTE: Most methods of this trait take a mutable reference to the resolver, so that the resolver
+/// can mutate its own state as necessary during resolution (e.g. to manage a cache, or other side
+/// table-like data structures).
 pub trait TypeResolver<E> {
     fn source_manager(&self) -> Arc<dyn SourceManager>;
     /// Should be called by consumers of this resolver to convert a [SymbolResolutionError] to the
     /// error type used by the [TypeResolver] implementation.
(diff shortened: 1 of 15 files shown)


crates/assembly-syntax/src/ast/type.rs
```diff
@@ -243,7 +243,7 @@ impl TypeExpr {
 
         while let Some(ty) = worklist.pop() {
             match ty {
-                Self::Primitive(_) => continue,
+                Self::Primitive(_) => {},
```
plafer, line 263: small nit: what's  your take on using `// do nothing` comments to make it explicit that we didn't just forget to implement that block?

huitseeker: I think `{}` conveys that?

plafer: To me it might look like when writing the code you wrote `{}` so that it compiles, but forgot to go back and actually implement the block. But it's a minor thing, I've defaulted to using `// do nothing` but I also never encountered a situation where I doubted that a block was mistakenly left unimplemented... Meh, probably fine as is 🤷‍♂️

PR #2858, opened 2026-03-16
Memoize constant evaluation

huitseeker:
Semantic analysis evaluated constants w/o memoizing referenced constants, allowing attacker-controlled constant graphs (e.g. `C0 = C1 + C1`, `C1 = C2 + C2`, …) to trigger exponential work during parsing/semantic analysis. More specifically `AnalysisContext` always returned `CachedConstantValue::Miss` for defined constants and did not implement `ConstEnvironment::on_eval_completed`, so shared dependencies were re-evaluated repeatedly.

Here, we add constant-value memoization in `AnalysisContext`, update cache via `ConstEnvironment::on_eval_completed`, adds a regression test for shared-subexpression constant chains.


crates/assembly-syntax/src/parser/mod.rs
```diff
@@ -480,4 +480,103 @@ end
...
+
+        let timeout_secs = env::var("MIDEN_CONST_EVAL_TEST_TIMEOUT_SECS")
+            .ok()
+            .and_then(|value| u64::from_str(value.as_str()).ok())
+            .unwrap_or(5);
+        // Leave room for process startup/test harness overhead while keeping a strict budget for
+        // the pathological dependency graph.
+        let status = wait_with_timeout(child, Duration::from_secs(timeout_secs));
+        assert!(status.success(), "child was too slow or failed: {status:?}");
+    }
```
plafer: How was the 5 seconds threshold chosen? I'm wondering if this will lead to a flaky test on some machines (including CI) - or if the 5 seconds is so large that we're not actually testing anything (i.e. previous behavior also took way less than 5 seconds)

huitseeker: Great call. I'll remove the wall-clock timeout/subprocess test and replaced it with a deterministic unit test that checks memoization behavior directly on the crafted shared-subexpression graph.

It should asserts exact cache lookup counts (misses = depth + 1, hits = depth) instead of relying on an arbitrary delay threshold.

PR #2861, opened 2026-03-16
Update stale docs related to caller operation

### LINKED ISSUE #2817: Update stale doc comments about caller and kernel module restrictions
Several doc comments and diagnostic stubs claim restrictions that are not enforced today:

**caller restrictions**

These locations imply kernel-only or syscall-context rules for `caller`:

- `SemanticAnalysisError::CallerInKernel` in `crates/assembly-syntax/src/sema/errors.rs`
- The doc comment on `VerifyInvokeTargets` in `crates/assembly-syntax/src/sema/passes/verify_invoke.rs`
- The doc comment on `SystemInterface::caller_hash` in `processor/src/processor.rs`
- The doc comment on `op_caller` in `processor/src/execution/operations/sys_ops/mod.rs`

However, `caller` is lowered unconditionally by the assembler via `Instruction::Caller` and executed without a syscall-context check by `op_caller`. This matches the user documentation for `caller`.

**Kernel module internal invocation rules**

These docs claim kernel modules may not use `call` or `syscall` internally:

- `ModuleKind::Kernel` in `crates/assembly-syntax/src/ast/module.rs`
- The doc comment on `VerifyInvokeTargets` in `crates/assembly-syntax/src/sema/passes/verify_invoke.rs`
- The comment in `visit_items` in `crates/assembly-syntax/src/sema/mod.rs`

The semantic visitor does not reject kernel modules in `VerifyInvokeTargets::visit_mut_syscall` or `VerifyInvokeTargets::visit_mut_call`. This matches the direction in PR #2296.

Recommendation: update or delete the stale doc comments and diagnostic stubs so they describe current behavior.

plafer:
Closes #2817

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -14,14 +14,17 @@ xclippy = [
     "-Wclippy::await_holding_lock",
     "-Wclippy::char_lit_as_u8",
     "-Wclippy::dbg_macro",
+    "-Wclippy::derive_partial_eq_without_eq",
     "-Wclippy::disallowed_methods",
     "-Wclippy::empty_enums",
     "-Wclippy::exit",
+    "-Wclippy::explicit_deref_methods",
     "-Wclippy::filter_map_next",
     "-Wclippy::flat_map_option",
     "-Wclippy::float_cmp_const",
     "-Wclippy::if_let_mutex",
     "-Wclippy::imprecise_flops",
+    "-Wclippy::inefficient_to_string",
     "-Wclippy::invalid_upcast_comparisons",
     "-Wclippy::large_digit_groups",
     "-Wclippy::large_stack_arrays",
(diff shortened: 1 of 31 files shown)


processor/src/execution/operations/sys_ops/mod.rs
```diff
@@ -46,11 +46,9 @@ where
...
 }
 
-/// Overwrites the top four stack items with the hash of a function which initiated the current
-/// SYSCALL.
-///
-/// # Errors
-/// Returns an error if the VM is not currently executing a SYSCALL block.
+/// Overwrites the top four stack items with the value of the CALLER_HASH register, which is the
+/// hash of the procedure that initiated the most recent SYSCALL, or ZERO if not in a syscall
+/// context.
```
bobbinth, line 51: Question: do we enforce that it is `ZERO` when not in syscall context?

plafer: Yes - boundary constraints enforce that the columns are initialized to `ZERO`, and then the transition constraints only allow a change on {dyn}call/syscall/end. These together enforce that it's non-ZERO only in a syscall context.

PR #2863, opened 2026-03-17
Validate untrusted MAST in library and kernel deserialization

huitseeker:
Library and kernel deserialization now treat artifact forests as untrusted input. `Library::read_from` uses `UntrustedMastForest::read_from(...).validate()` before building the library, and `KernelLibrary::read_from` stays covered through `Library::read_from`.

This blocks spoofed node digests in .masl and kernel artifacts. If node bytes and stored digest do not match, deserialization fails.

Tests now patch a stripped forest digest and confirm both `Library::read_from_bytes` and `KernelLibrary::read_from_bytes` fail with a hash-mismatch validation signal.

### DIFF
--- a/crates/assembly/src/tests.rs
+++ b/crates/assembly/src/tests.rs
@@ -455,7 +455,10 @@ fn locate_node_infos(bytes: &[u8]) -> (usize, usize) {
     (offset, node_count)
 }
 
-fn build_library_bytes_with_spoofed_first_node_digest(lib: &Library, spoof_seed: &str) -> Vec<u8> {
+fn build_library_bytes_with_spoofed_first_node_digest(
+    lib: &Library,
+    spoof_seed: &str,
+) -> (Vec<u8>, Word) {
     use miden_core::serde::{ByteWriter, Serializable};
 
     // Serialize the MastForest in stripped form so the byte layout is minimal and stable.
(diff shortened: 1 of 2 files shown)


crates/assembly/src/tests.rs
```diff
@@ -408,6 +408,132 @@ fn library_serialization() -> Result<(), Report> {
...
+        usize::try_from(value).expect("encoded usize does not fit host usize")
+    } else {
+        let end = (*offset).checked_add(length).expect("offset overflow while reading vint64");
+        let mut encoded = [0u8; 8];
+        encoded[..length].copy_from_slice(&bytes[*offset..end]);
+        *offset = end;
+        let value = u64::from_le_bytes(encoded) >> length;
+        usize::try_from(value).expect("encoded usize does not fit host usize")
+    }
+}
```
plafer, line 435: Any reason to use this instead of `ByteReader::read_usize`?

huitseeker: The helper reads from raw `&[u8]` because this test needs exact byte offsets to patch node-info in place. `ByteReader::read_usize` gives the value, but without an offset-tracking reader it does not directly give consumed-byte position. I kept the local decoder for that.

plafer: Makes sense

huitseeker:
@bobbinth Using `miden-vm/benches/deserialize_core_lib.rs` to measure standard library deserialization time:

  | Case | Commit | Median time |
  | --- | --- | ---: |
  | Before | `7bfba6f54` | `3.2492 ms` |
  | After | current branch | `6.2680 ms` |
  
The change is expected because library deserialization now validates node digests instead of trusting serialized ones.

I added back an `*_unchecked` path.

Issue #2730, opened 2026-02-23
Make `miden_prover::prove()` method sync


plafer:
@Mirko-von-Leipzig has argued that our `miden_prover::prove()` method shouldn't be `async`. Indeed, it fundamentally doesn't need to be: it is currently only `async` because the method currently

1. executes the program (async),
2. builds the trace (sync), and
3. proves the program (sync)

If it took as input the `ExecutionTrace` (i.e. the output of step 2), then it could be made synchronous trivially.

Then, as discussed with @PhilippGackstatter, if `ExecutedTransaction` stored either the `TraceGenerationContext` (output of step 1) or the trace (output of step 2), then AFAIU in the node, this is all we would need to be able to call `prove()`. I believe it's best to store `TraceGenerationContext` in `ExecutedTransaction`, as storing the trace would drastically increase the size of an `ExecutedTransaction` (e.g. ~17MB for a trace with 32k rows). This implies that the node will need to call `miden_processor::build_trace(TraceGenerationContext) -> ExecutionTrace` before being able to call `prove()` - or maybe we offer a convenience `prove(TraceGenerationContext)` method as well, whatever.

The only current blocker to this is `TraceGenerationContext` is currently not serializable (required by `ExecutedTransaction`), first discussed in #2252, due to storing a bunch of `Arc<MastForest>`. I believe the best solution to that is the 3rd option suggested in https://github.com/0xMiden/miden-vm/issues/2252#issuecomment-3478671599, i.e. to store the subset of all nodes of all `MastForest`s that were visited during execution (as part of `ExecutionTracer`). This will require some changes, but nothing too involved.

### Caveat
Building the trace using `build_trace()` assumes that the `TraceGenerationContext` was generated by the `FastProcessor`. Hence, *it would be incorrect to have a public RPC endpoints that accepts `TraceGenerationContext`s that weren't built by us*. @PhilippGackstatter @Mirko-von-Leipzig can you confirm that this would not be the case in this new design - specifically with respect to how the remote prover works?

bobbinth:
> Building the trace using `build_trace()` assumes that the `TraceGenerationContext` was generated by the `FastProcessor`. Hence, _it would be incorrect to have a public RPC endpoints that accepts `TraceGenerationContext`s that weren't built by us_.

What is the consequence of this? That is, if someone submits a maliciously-built `TraceGenerationContexts`, what kind of issues could this cause?

Mirko-von-Leipzig:
[...]

PR #2865, opened 2026-03-17
Make execution and proving sync-first, with runtime-free async compatibility

### LINKED ISSUE #2501: Should the processor expose sync APIs?
### Packages versions

miden-vm: 0.21.0 (upcoming)

## Processor API: Should it be async or sync?

We're hitting test failures in miden-base on miden-vm next (v0.21.0, unreleased) when using the P3 VM processor from async test functions. This raises a question about what the processor API should look like.

The test failure happens at `processor/src/fast/mod.rs:923`:

```rust
pub fn execute_for_trace_sync(...) -> Result<...> {
    let rt = tokio::runtime::Builder::new_current_thread().build().unwrap();
    rt.block_on(self.execute_for_trace(program, host, fragment_size))
}
```

When called from an async test function, it panics:

```
thread 'test_ecdsa_acl' panicked at processor/src/fast/mod.rs:925:12:
Cannot start a runtime from within a runtime.
```

Here's the call chain in miden-base:

```rust
#[tokio::test]
async fn test_ecdsa_acl() -> Result<()> {
    // ... setup ...
    prove_and_verify_transaction(tx)?;  // calls prover.prove()
                                        // which calls processor.execute_for_trace_sync()
                                        // which tries to create a new Tokio runtime
                                        // panic!
}
```

The processor has both async methods like `execute_for_trace()` and sync wrappers like `execute_for_trace_sync()`. The sync wrappers create their own Tokio runtime and call `block_on()`. This works from sync code but breaks from async contexts.

A few options:

**Async only**

Remove the `*_sync()` wrappers. Users who need sync can create their own runtime:

```rust
let rt = tokio::runtime::Runtime::new()?;
rt.block_on(processor.execute_for_trace(...))?;
```

Callers control async execution.

**Detect existing runtime**

Keep sync wrappers but check if we're already in a runtime:

```rust
  match tokio::runtime::Handle::try_current() {
      Ok(_handle) => {
          // We're in a runtime - use block_in_place to avoid blocking the worker thread
          tokio::task::block_in_place(|| {
              // Create a new runtime in this blocking context
              let rt = tokio::runtime::Builder::new_current_thread().build().unwrap();
              rt.block_on(self.execute(program, host))
          })
      },
      Err(_) => {
          // No runtime exists - create one and use it
          let rt = tokio::runtime::Builder::new_current_thread().build().unwrap();
          rt.block_on(self.execute(program, host))
      },
  }
```

Works in both contexts. Calling `block_on()` from within an existing runtime isn't ideal but technically works.

**Spawn blocking**

Use `spawn_blocking()` to move work to another thread when already in a runtime. More complex. Thread overhead.

What makes the most sense? The Winterfell VM was fully synchronous, so sync wrappers were probably added to match that API. Now that P3 uses async internally, maybe it's worth rethinking this.

The same pattern appears in `execute_sync()` and `execute_sync_mut()` in the same file.

huitseeker:
This PR makes execution and proving sync again at the VM core, while keeping async APIs for async users. The processor and prover no longer depend on `tokio::block_on()` to offer sync  entrypoints. Sync execution, trace generation, and proving are native again. Async execution and proving remain available, and async host callbacks are restored for downstream users that still need them.

<details><summary>In detail, this PR does 3 things: </summary>

- it makes `FastProcessor` and `miden_prover` sync-first again, with native `*_sync` paths for execution, trace generation, and proving. 
- it keeps async entrypoints for the same flows, so async callers can still use `execute()`, `prove()`, and async host methods. 
- it restores the async host path in the VM without making the sync core depend on a runtime. 
</details>

### Issues

 - Fixes #2501 : the bug was that sync APIs were only runtime wrappers around async code. That made `execute_for_trace_sync()` panic when called inside an existing runtime. This PR removes that design.
  
- Fixes: #2552, this PR does not remove `miden_processor::execute_sync()`. It resolves the reason that removal was proposed.

- Fixes a good chunk of #2730:  this PR completes the VM part. `miden_prover::prove_sync()` is now native, and the processor exposes native sync execution-for-trace. That means callers can re-execute with a sync host, build the trace, and prove without async overhead. The PR also keeps the split flow visible through the VM surface, so execution-for-trace and trace building can happen as separate steps.

Out of scope for this PR: make `TraceGenerationContext` serializable, or accept untrusted trace-generation inputs.

Supersedes #2626

----

Edit: Since opening this PR (see review), edits kept the async host path intact while restoring a clean `BaseHost` + `SyncHost` split, made `execution_tracer` crate-private, and moved the fast execution entrypoints into
  `execution_api.rs` so the public surface is smaller and easier to follow. We also removed the deprecated `execute_sync_mut` and `execute_for_trace*` wrappers, made `execute_trace_inputs*()` the trace entrypoint, returned `ExecutionOutput` from the top-level `execute()` wrappers, and tightened the trace/proving flow with bundled trace-build inputs and stable advice fingerprints.

### DIFF
--- a/miden-vm/README.md
+++ b/miden-vm/README.md
@@ -36,13 +36,13 @@ Miden crate exposes several functions which can be used to execute programs, gen
 
 ### Executing programs
 
-To execute a program on Miden VM, you can use `execute_sync()`. The async `execute()` variant is
-also available for async callers. `execute_sync()` takes the following arguments:
+To execute a program on Miden VM, you can use `execute()`. The sync `execute_sync()` variant is
+also available for sync callers. These functions take the following arguments:
 
 - `program: &Program` - a reference to a Miden program to be executed.
 - `stack_inputs: StackInputs` - a set of public inputs with which to execute the program.
 - `advice_inputs: AdviceInputs` - the private inputs used to build the advice provider; use `AdviceInputs::default()` when no private inputs are needed.
-- `host: Host` - an instance of a `Host` which can be used to supply non-deterministic inputs to the VM and receive messages from the VM.
+- `host` - an instance of `Host` for `execute()` or `SyncHost` for `execute_sync()`, used to supply non-deterministic inputs to the VM and receive messages from the VM.
 - `options: ExecutionOptions` - a set of options for executing the specified program (e.g., max allowed number of cycles).
 
 The function returns a `Result<ExecutionOutput, ExecutionError>` which will contain the final stack
--- a/processor/src/fast/basic_block/mod.rs
+++ b/processor/src/fast/basic_block/mod.rs
@@ -8,7 +8,9 @@ use miden_core::{
 
 use crate::{
     BaseHost, Host, SyncHost,
+    advice::AdviceMutation,
     errors::{MapExecErrWithOpIdx, advice_error_with_context, event_error_with_context},
+    event::EventError,
     fast::{BreakReason, FastProcessor},
 };
 
@@ -44,7 +46,7 @@ impl FastProcessor {
         node_id: MastNodeId,
         host: &impl BaseHost,
         event_id: EventId,
-        mutations: Result<Vec<crate::advice::AdviceMutation>, crate::event::EventError>,
+        mutations: Result<Vec<AdviceMutation>, EventError>,
     ) -> ControlFlow<BreakReason> {
         let mutations = match mutations {
             Ok(mutations) => mutations,
--- a/processor/src/host/advice/mod.rs
+++ b/processor/src/host/advice/mod.rs
@@ -94,6 +94,7 @@ impl AdviceProvider {
     ///
     /// The fingerprint is insensitive to advice-map insertion order and Merkle-store insertion
     /// order, but it still reflects advice-stack order and precompile-request order.
+    #[must_use]
     pub(crate) fn fingerprint(&self) -> [u8; 32] {
         let stack = self.stack.iter().copied().collect::<Vec<_>>().to_bytes();
         let map = self.map.to_bytes();


processor/src/host/mod.rs
```diff
@@ -93,10 +89,8 @@ pub trait Host {
     /// - Return errors without event names or IDs - the caller will enrich them via
     ///   [`Host::resolve_event()`]
     /// - System events (IDs 0-255) are handled by the VM before calling this method
-    fn on_event(
-        &mut self,
-        process: &ProcessorState<'_>,
-    ) -> impl FutureMaybeSend<Result<Vec<AdviceMutation>, EventError>>;
+    fn on_event(&mut self, process: &ProcessorState<'_>)
+    -> Result<Vec<AdviceMutation>, EventError>;
```
plafer: My understanding is that it is currently a requirement for `Host::get_mast_forest()` and `Host::on_event()` to be async, since the APIs we use in `miden-base` to implement those are async (e.g. [here](https://github.com/0xMiden/protocol/blob/163c54951daadb2b801e7e3b5c004821dc02e4b7/crates/miden-tx/src/executor/exec_host.rs#L517)).

What were the motivations for those changes?

EDIT: How I was picturing this (first discussed [here](https://github.com/0xMiden/miden-vm/issues/2730#issuecomment-3954382932) I believe) was that we would expose an *additional* sync API (based on a new `SyncHost`) - but we'd have to keep the async one as well.

prover/src/lib.rs
bobbinth, line 58: I've been recently thinking that maybe our APIs are incorrectly set up here and we are mixing "areas of responsibilities" because of this. Specifically:

We have a processor that has two jobs (at least at the high level):
- Execute programs - this must be async.
- Build execution traces - this is a sync process.

We have a prover, that currently uses the processor to execute the program and build the trace, and then runs proof generation. Because the prover needs to execute the processor, this function used to be async. But why not change `prove()` function to be something like:

```rust
pub fn prove(
    program_info: ProgramInfo,
    options: ProvingOptions,
) -> Result<ExecutionProof, ExecutionError> {
    ...
}
```

Then, the user would be responsible for executing the program in an async way, and the prover then just takes the result, builds the trace, and generates the proof. This naturally makes the prover sync and I think is a better separation of concerns (and should eliminate many redundancies we have now).

processor/src/host/mod.rs
bobbinth: Yes, these methods need to be async. But as I mentioned in the comment below, I'm not sure why we should try to make the processor sync (or at least the part that executes the program). In my mind:

- Program execution should always be async. We could have a sync wrapper for test purposes, if we want - but that's a "nice-to-have".
- Trace generation and proving should always be sync.

The problem now is that we mix these in ways that makes proving async as well. I think we should have:

- Executing program and generating `TraceGenerationContext` should be async.
- Using `TraceGenerationContext` to build a trace should be sync.
- Generating a proof from `TraceGenerationContext` should be sync.

huitseeker: Yes. I misunderstood #2730 initially, something the last commits fix. This PR makes the split below work at the VM boundary:

```rust 
let (execution_output, ctx) = FastProcessor::new(stack_inputs)
  .execute_for_trace(&program, &mut host)
  .await?;

let trace = miden_processor::trace::build_trace(execution_output, ctx, program.to_info())?;
```

That is the execution model you describe:

  - `execute` + produce `TraceGenerationContext`: async when needed,
  - `build trace`: sync,
  - `prove`: sync.
  
What this PR changes is not the basic idea of that split. What it fixes is that the sync side is native again instead of being built on `block_on(...)` wrappers:

- async execution with async `Host` remains available,
- `FastProcessor::execute_for_trace(...)` remains async,
- `miden_processor::trace::build_trace(...)` remains sync,
- `miden_prover::prove_sync(...)` is native sync again.

One thing still missing is a convenience API for “sync prove from pre-executed data”.  <details><summary>But today that is already possible through lower-level public pieces.</summary>

- `FastProcessor::execute_for_trace(...)` or `execute_for_trace_sync(...)`
- `miden_processor::trace::build_trace(...)`
- `ExecutionTrace::to_row_major_matrix()`
- `ExecutionTrace::public_inputs()`
- `ExecutionTrace::aux_trace_builders()`
- `miden_prover::prove_stark(...)`
- `ExecutionProof::new(...)`
</details>

  So the split path today is effectively:

```rust
let (execution_output, ctx) = FastProcessor::new(stack_inputs)
    .execute_for_trace(&program, &mut host)
    .await?;

let trace = miden_processor::trace::build_trace(execution_output, ctx, program.to_info())?;

let trace_matrix = trace.to_row_major_matrix();
let (public_values, kernel_felts) = trace.public_inputs().to_air_inputs();
let aux_builder = trace.aux_trace_builders();

let proof_bytes = miden_prover::prove_stark(
    &config,
    &trace_matrix,
    &public_values,
    &[&kernel_felts],
    &aux_builder,
)?;
let proof = ExecutionProof::new(proof_bytes, hash_fn, trace.precompile_requests().to_vec());
```

So the missing piece is just the wrapper that makes the final sync proving step ergonomic. A small follow-up could expose that directly, for example:

```rust
pub fn prove_from_trace_sync(
    execution_output: ExecutionOutput,
    ctx: TraceGenerationContext,
    program_info: ProgramInfo,
    options: ProvingOptions,
) -> Result<(StackOutputs, ExecutionProof), ExecutionError>;
```

Edit: That's now added to the last commit.

prover/src/lib.rs
huitseeker: Answered [here](https://github.com/0xMiden/miden-vm/pull/2865#discussion_r2958490886).

processor/src/fast/mod.rs
```diff
@@ -45,6 +45,62 @@ use step::{NeverStopper, StepStopper};
...
+    fn get_mast_forest(&self, _node_digest: &Word) -> Option<Arc<MastForest>> {
+        panic!("HostAdapter::get_mast_forest should not be called directly")
+    }
+
+    fn on_event(
+        &mut self,
+        _process: &ProcessorState<'_>,
+    ) -> Result<Vec<crate::advice::AdviceMutation>, crate::event::EventError> {
+        panic!("HostAdapter::on_event should not be called directly")
+    }
```
plafer: I think bringing back the previous `BaseHost`/`SyncHost`/`AsyncHost` structure would remove the need to `panic!`-implement some methods. Basically the host traits are structured

```rs
// Contains the common methods
trait BaseHost {
  fn get_label_and_source_file();  
  fn on_debug();
  fn on_trace();
  fn resolve_event();
}

trait SyncHost: BaseHost {
  fn on_event();
  fn get_mast_forest();
}

trait AsyncHost: BaseHost {
  fn on_event() -> impl FutureMaybeSend;
  fn get_mast_forest() -> impl FutureMaybeSend;
}
```

Then `execute_impl()` can take a `BaseHost` as argument (which doesn't contain the methods that are not allowed to be called), while `FastProcessor::execute_sync()` takes a `SyncHost`, and `FastProcessor::execute()` takes a `AsyncHost`.

processor/src/host/mod.rs
plafer: Indeed, there's no point in not exposing a sync API (as the latest version of this PR shows). The sans-IO infrastructure (i.e. `execute_impl()`) is sync-first, and allows implementing sync or async wrappers rather trivially. So IMO even if `miden-base` doesn't need it *today*, we should just expose a set of sync *and* async APIs, since it's just as easy for us to do that cleanly.

> Program execution should always be async. We could have a sync wrapper for test purposes, if we want - but that's a "nice-to-have".

What I gathered from @PhilippGackstatter's [comment](https://github.com/0xMiden/miden-vm/issues/2730#issuecomment-3957286650) was that miden-base would actually benefit from having a sync-based execution, for already executed transactions that have all the data they need cached in memory.

processor/src/fast/mod.rs
```diff
@@ -144,6 +200,96 @@ pub struct FastProcessor {
...
+        ExecutionOutput {
+            stack,
+            advice: self.advice,
+            memory: self.memory,
+            final_pc_transcript: self.pc_transcript,
+        }
+    }
+
+    #[inline(always)]
+    fn execution_result_from_flow(
```
plafer, line 153: Would add a small docstring to all the new methods

processor/src/host/mod.rs
PhilippGackstatter: Assuming for proving we do not re-execute but use the trace provided by an `ExecutedTransaction`, the main use case for sync-only execution would be for transaction re-execution in the node (https://github.com/0xMiden/miden-vm/issues/2730#issuecomment-3950341295).

processor/src/fast/mod.rs
huitseeker: Yep, added on the last commit.

processor/src/fast/mod.rs
huitseeker: Added as well

processor/src/fast/basic_block/mod.rs
```diff
@@ -53,39 +108,38 @@ impl FastProcessor {
...
-                )));
-            }
+            );
         }
-        ControlFlow::Continue(())
+
+        let processor_state = self.state();
+        let mutations = host.on_event(&processor_state).await;
+        let host_adapter = HostAdapter::new(host);
+        self.apply_host_event_mutations(current_forest, node_id, &host_adapter, event_id, mutations)
```
plafer: We can get rid of the `HostAdapter` entirely; here we can just pass `host` directly of `apply_host_event_mutations()`. This applies everywhere else where `HostAdapter` is still used

huitseeker: Of course, thanks!

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 - Added `math::u128` division operations ([#2776](https://github.com/0xMiden/miden-vm/pull/2776)).
 - [BREAKING] Migrated to lifted-STARK backend and `miden-crypto` to v0.23 ([#2783](https://github.com/0xMiden/miden-vm/pull/2783)).
 
+- [BREAKING] Migrate to lifted-STARK backend and `miden-crypto` to v0.23 ([#2783](https://github.com/0xMiden/miden-vm/pull/2783)).
+- [BREAKING] Sync execution and proving APIs now require `SyncHost`; async `Host`, `execute`, and `prove` remain available ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
```
bobbinth: I think these need to be moved into v0.23.0 section now.

prover/src/lib.rs
```diff
@@ -64,8 +63,46 @@ pub async fn prove(
...
+#[instrument("prove_trace_sync", skip_all)]
+pub fn prove_from_trace_sync(
+    execution_output: ExecutionOutput,
+    trace_generation_context: TraceGenerationContext,
+    program_info: ProgramInfo,
+    options: ProvingOptions,
+) -> Result<(StackOutputs, ExecutionProof), ExecutionError> {
+    let trace = build_trace(execution_output, trace_generation_context, program_info)?;
+    prove_execution_trace(trace, options)
+}
```
bobbinth, line 94: I think of this one as our main proving method, and I'd potentially rename it into just `prove()` and get rid of the current `prove()` and `prove_sync()`.

But putting this aside, I think inputs into this method contain some redundancies:

1. `ProvingOptions` contain `ExecutionOptions` inside of them, which are not relevant in this context. I'd refactor `ProvingOptions` to remove `ExecutionOptions` from there.
2. `ExecutionOutput` contains `AdviceProvider` and `Memory`, which are not really needed here, and in fact would make this function difficult to use (i.e., one of its goals is to avoid having to serialize advice provider data). I think the only things that we need from there are `StackInputs` and `PrecompileTranscript`. It may make sense to incorporate these into `TraceGenerationContext`.
3. Similarly, `ProgramInfo` could probably be incorporated into `TraceGenerationContext` or maybe some struct that contains `TraceGenerationContext` and the input data.

Basically, I don't think it makes sense to provide all these inputs separately because they are only valid in combination (e.g, it is not like we can provide a different `ProgramInfo` with the same trace generation context).

Overall, I think this method could look like:
```rust
pub fn prove_from_trace_sync(
    options: ProvingOptions,
) -> Result<ExecutionProof, ExecutionError> {
    ...
}
```

Separately, our `ExecutionTrace` seems to contain some extraneous data as well. Specifically, AFAICT, we put the advice provider there mostly for testing purposes and mostly because we are testing execution correctness by building execution traces. This is a remnant of the old structure and many tests can probably be updated to use `ExecutionOutput` instead of `ExecutionTrace`. But this is a separate issue for another PR.

processor/src/trace/mod.rs
```diff
@@ -39,6 +39,7 @@ mod tests;
 // RE-EXPORTS
 // ================================================================================================
 
+pub use execution_tracer::TraceGenerationContext;
```
bobbinth, line 43: Why do we need to export this separately? The whole `execution_tracer` module exported (on line 28 above). If we do want to export this separately, I'd make `execution_tracer` module private.

processor/src/lib.rs
```diff
@@ -116,37 +115,22 @@ pub async fn execute(
...
-#[cfg(not(target_arch = "wasm32"))]
 #[tracing::instrument("execute_program_sync", skip_all)]
 pub fn execute_sync(
     program: &Program,
     stack_inputs: StackInputs,
     advice_inputs: AdviceInputs,
-    host: &mut impl Host,
+    host: &mut impl SyncHost,
     options: ExecutionOptions,
 ) -> Result<ExecutionTrace, ExecutionError> {
```
bobbinth: Not directly related to this PR, but I don't think this function or the `execute()` function currently have the right return values. Specifically, we should be returning `ExecutionOutput` from these methods rather than `ExecutionTrace`. Returning the trace is a remnant of the old structure when the only way to represent execution output was via `ExecutionTrace`.

In general, it is not clear to me if having these methods is all that valuable. Almost everywhere we execute the VM we use the `FastProcessor` struct directly and if we remove trace building from these functions, they would be very thin wrappers over the `FastProcessor` functionality.

processor/src/fast/mod.rs
```diff
@@ -462,6 +568,36 @@ impl FastProcessor {
...
+    ///     miden_processor::trace::build_trace(execution_output, ctx, program.to_info()).unwrap();
+    ///
+    /// assert_eq!(*trace.program_hash(), program.hash());
+    /// ```
+    #[instrument(name = "execute_for_trace_sync", skip_all)]
+    pub fn execute_for_trace_sync(
+        self,
+        program: &Program,
+        host: &mut impl SyncHost,
+    ) -> Result<(ExecutionOutput, TraceGenerationContext), ExecutionError> {
```
bobbinth: As mentioned in the other comment, I think the only thing we should return from here is `TraceGenerationContext` as it should be the only thing required for trace generation. This would require updating the `build_trace()` function - but I think that's the right approach.

Also, lines 569-570 of the doc comment need to be updated.

processor/src/fast/mod.rs
```diff
@@ -470,14 +606,11 @@ impl FastProcessor {
     ) -> Result<(ExecutionOutput, TraceGenerationContext), ExecutionError> {
```
bobbinth: Is the doc comment here correct? This doesn't seem like just a wrapper for the sync method.

processor/src/fast/mod.rs
```diff
@@ -452,6 +548,16 @@ impl FastProcessor {
...
+        self,
+        program: &Program,
+        host: &mut impl SyncHost,
+    ) -> Result<ExecutionOutput, ExecutionError> {
+        self.execute_with_tracer_sync(program, host, &mut NoopTracer)
+    }
+
+    /// Async compatibility wrapper for [`Self::execute_sync`].
+    #[inline(always)]
     pub async fn execute(
```
bobbinth: I think we need to be consistent with `sync`/`async` naming. Currently we use several variations:

- Add `_sync` suffix to sync methods and keep async methods without a suffix  - e.g., `execute()` and `execute_sync()`.
- Add `_async` suffix to async methods and keep sync methods without a suffix - e.g., `load_mast_forest_async()` and `load_mast_forest().
- Put the designator in the middle - e.g., `execute_sync_mut()` and `execute_mut()`.

I think my preference would be to use `_sync` suffix everywhere - but not a strong opinion (i.e., using `_async` suffix or both `_sync` and `_async` suffixes is fine too).

processor/src/fast/mod.rs
bobbinth, line 1: I'm a bit concerned with the size of this file. It is already pretty huge and this PR adds another 200 lines to it. Could we maybe split things out of it. For example, could we put all `_sync` methods into something like `sync_execution.rs` file?

processor/src/trace/mod.rs
huitseeker: Fixed. `execution_tracer` is crate-private now, and `TraceGenerationContext` remains the public re-export.

CHANGELOG.md
huitseeker: Fixed.

processor/src/fast/mod.rs
huitseeker: Fixed. I split the execution APIs into processor/src/fast/execution_api.rs and updated the docs while doing that.

processor/src/fast/mod.rs
huitseeker: Mostly addressed locally. I normalized the obvious outliers and removed `execute_sync_mut()`. If we want one more rename-only pass, I’d keep it small rather than to risk mixing more churn into the design changes.

processor/src/fast/mod.rs
huitseeker: I agree with the direction, but I think that final collapse belongs in a separate API-design pass. This PR now has a single bundled trace/proving input type, which is a step in that direction, but not the end state.

processor/src/fast/mod.rs
huitseeker: This got overtaken by the follow-up cleanup: the deprecated `execute_for_trace*()` wrappers are removed now, and `execute_trace_inputs*()` is the only execution-to-trace entrypoint.

processor/src/lib.rs
huitseeker: Fixed. `miden_processor::execute()` and `execute_sync()` now return `ExecutionOutput`; trace building is explicit via `execute_trace_inputs*()` plus `trace::build_trace()`.

prover/src/lib.rs
huitseeker: Partially addressed locally. I bundled the mutually-dependent proving inputs into `TraceBuildInputs`, so callers no longer thread `ExecutionOutput`, `TraceGenerationContext`, and `ProgramInfo` separately. I did not take the last step to a `TraceGenerationContext`-only proving API in this PR, because that also implies redesigning what lives in `ProvingOptions`, `ExecutionOutput`, and likely `ExecutionTrace`. I’d prefer to keep that as a dedicated follow-up.

processor/src/fast/mod.rs
```diff
@@ -144,6 +138,105 @@ pub struct FastProcessor {
...
+    #[inline(always)]
+    fn trace_result_from_parts(
+        program: &Program,
+        execution_output: ExecutionOutput,
+        tracer: ExecutionTracer,
+    ) -> TraceBuildInputs {
+        let mut trace_generation_context = tracer.into_trace_generation_context();
+        trace_generation_context.bind_execution(program.to_info(), &execution_output);
+        TraceBuildInputs::from_execution(program, execution_output, trace_generation_context)
+    }
```
plafer: nit: this one can go in `execution_api.rs` as well (it's only used there)

processor/src/trace/trace_state.rs
```diff
@@ -769,10 +770,51 @@ impl IntoIterator for BitwiseReplay {
...
+    pub(crate) fn advice_provider_fingerprint(&self) -> &[u8; 32] {
+        &self.advice_provider_fingerprint
+    }
+}
+
 /// Replay data for kernel operations.
 #[derive(Debug, Default)]
 pub struct KernelReplay {
     kernel_proc_accesses: VecDeque<Word>,
+    execution_binding: Option<ExecutedTraceBinding>,
```
plafer: Why are we putting the executed trace binding in the kernel replay? The kernel replay basically records kernel procedure accesses, so I don't immediately see why it makes sense to put it there. I would have expected it to be a standalone field in `TraceGenerationContext` probably?

processor/src/fast/mod.rs
```diff
@@ -144,6 +138,105 @@ pub struct FastProcessor {
...
+                BreakReason::Stopped(_) => {
+                    unreachable!("Execution never stops prematurely with NeverStopper")
+                },
+            },
+        }
+    }
+
+    /// Pairs execution output with the trace inputs captured by the tracer.
+    #[inline(always)]
+    fn trace_result_from_parts(
```
```suggestion
    fn trace_build_inputs_from_parts(
```
Not sure what a "trace result" is - this is probably more descriptive?

processor/src/fast/mod.rs
```diff
@@ -144,6 +138,105 @@ pub struct FastProcessor {
...
+        program: &Program,
+        execution_output: ExecutionOutput,
+        tracer: ExecutionTracer,
+    ) -> TraceBuildInputs {
+        let mut trace_generation_context = tracer.into_trace_generation_context();
+        trace_generation_context.bind_execution(program.to_info(), &execution_output);
```
plafer: We probably never want to call `into_trace_generation_context()` without binding the execution, right? Hence, would it make more sense to have `into_trace_generation_context(ProgramInfo, &ExecutionOutput)` so that we don't forget to call it?

processor/src/fast/mod.rs
```diff
@@ -144,6 +138,105 @@ pub struct FastProcessor {
...
+    }
+
+    /// Converts a step-wise execution result into the next resume context, if execution stopped.
+    #[inline(always)]
+    fn resume_result_from_flow(
+        flow: ControlFlow<BreakReason, StackOutputs>,
+        mut continuation_stack: ContinuationStack,
+        current_forest: Arc<MastForest>,
+        kernel: Kernel,
+    ) -> Result<Option<ResumeContext>, ExecutionError> {
```
plafer: This can go in `execution_api.rs` too (also only used there).

Would `resume_context_from_flow()` be a better name?

processor/src/fast/mod.rs
```diff
@@ -144,6 +138,105 @@ pub struct FastProcessor {
...
+                        kernel,
+                    }))
+                },
+            },
+        }
+    }
+
+    /// Materializes the current stack as public outputs without consuming the processor.
+    #[inline(always)]
+    fn current_stack_outputs(&self) -> StackOutputs {
```
plafer: Can go in `execution_api.rs` too

miden-vm/benches/build_trace.rs
```diff
@@ -109,20 +103,18 @@ fn build_trace(c: &mut Criterion) {
...
+
+                            (host, program.clone(), processor)
+                        },
+                        |(mut host, program, processor)| async move {
+                            let trace_inputs =
+                                processor.execute_trace_inputs(&program, &mut host).await.unwrap();
+                            let trace = trace::build_trace(trace_inputs).unwrap();
                             black_box(trace);
                         },
                         BatchSize::SmallInput,
```
plafer: This is leftover from when we still had `Process` - Unless I'm missing something, it's not benchmarking anything different, and could just be removed

processor/src/host/advice/mod.rs
```diff
@@ -86,6 +90,38 @@ impl AdviceProvider {
         Ok(())
     }
 
+    pub(crate) fn fingerprint(&self) -> [u8; 32] {
```
plafer, line 98: Would add a simple docstring - namely stating that the fingerprint is stable across Merkle store and advice map (probably?) insertion order

processor/src/trace/trace_state.rs
huitseeker: Convenience. It's used as a consistency check that the trace inputs, `ProgramInfo`, and `ExecutionOutput` all came from the same run. It is checked before kernel_replay is used as actual replay input.

A standalone field on `TraceGenerationContext` would be cleaner, but arguably `TraceBuildInputs` is even cleaner, since that is where the binding is really consumed.

processor/src/fast/mod.rs
huitseeker: I agree with the goal but not with pushing execution data into `into_trace_generation_context()`. After the last cleanup, we already make the binding at the `TraceBuildInputs` layer, which feels like the right home for it. `ExecutionTracer::into_trace_generation_context()` can stay pure and only return trace-generation data, while `TraceBuildInputs::from_execution(...)` adds the execution binding. That gives the same “hard to forget” property without mixing execution-output metadata back into `TraceGenerationContext`.

miden-vm/benches/build_trace.rs
huitseeker: Cleaned up

miden-vm/README.md
```diff
@@ -36,14 +36,20 @@ Miden crate exposes several functions which can be used to execute programs, gen
 
 ### Executing programs
 
-To execute a program on Miden VM, you can use `execute()` which takes the following arguments:
+To execute a program on Miden VM, you can use `execute_sync()`. The async `execute()` variant is
+also available for async callers. `execute_sync()` takes the following arguments:
```
bobbinth: nit: it is a bit odd to list `execute_sync()` first and then `execute()` second - I'd probably flip the order.

miden-vm/README.md
```diff
@@ -36,14 +36,20 @@ Miden crate exposes several functions which can be used to execute programs, gen
...
 
 - `program: &Program` - a reference to a Miden program to be executed.
 - `stack_inputs: StackInputs` - a set of public inputs with which to execute the program.
+- `advice_inputs: AdviceInputs` - the private inputs used to build the advice provider; use `AdviceInputs::default()` when no private inputs are needed.
 - `host: Host` - an instance of a `Host` which can be used to supply non-deterministic inputs to the VM and receive messages from the VM.
 - `options: ExecutionOptions` - a set of options for executing the specified program (e.g., max allowed number of cycles).
```
bobbinth, line 46: If we change the order as mentioned in the last comment, this should be fine. But if we keep this about `execute_sync()` then I believe we should use `SyncHost` rather than `Host` for the `host` parameter.

processor/src/fast/basic_block/mod.rs
```diff
@@ -17,6 +17,61 @@ pub use sys_event_handlers::SystemEventError;
...
+    }
+
+    #[inline(always)]
+    fn apply_host_event_mutations(
+        &mut self,
+        current_forest: &MastForest,
+        node_id: MastNodeId,
+        host: &impl BaseHost,
+        event_id: EventId,
+        mutations: Result<Vec<crate::advice::AdviceMutation>, crate::event::EventError>,
```
bobbinth: nit: I'd move imports to the top of the file rather than using absolute paths.

processor/src/trace/mod.rs
```diff
@@ -39,10 +40,106 @@ mod tests;
...
 pub use utils::{ChipletsLengths, TraceLenSummary};
 
+/// Inputs required to build an execution trace from pre-executed data.
+#[derive(Debug)]
+pub struct TraceBuildInputs {
+    execution_output: ExecutionOutput,
+    trace_generation_context: TraceGenerationContext,
+    program_info: ProgramInfo,
+    execution_binding: trace_state::ExecutedTraceBinding,
+}
```
bobbinth, line 55: Let's leave this for a follow-up PR, but this struct needs to be redesigned. A few thoughts:

- We don't need full `ExecutionOutputs` here - we should take the minimal required data from there.
- There is overlap between `ExecuteTraceBinding` and other fields:
  - `ProgramInfo` is present in both `TraceBuildInputs` and in `ExecutedTraceBinding`.
  - `StackOutputs` are present in both `ExecutionOutput` and in `ExecutedTraceBinding`.

Also, I don't really love the name `TraceBuildInputs`. Don't have a great alternative - but maybe something like `ExecutionLog` or `ExecutionRecord` (i.e., something that can be replayed) is better? Would be somewhat parallel with `ExecutionOutputs` and `ExecutionTrace`.

processor/src/trace/trace_state.rs
```diff
@@ -766,6 +767,49 @@ impl IntoIterator for BitwiseReplay {
...
 
+// TRACE BUILD INPUT BINDING
+// ================================================================================================
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ExecutedTraceBinding {
+    program_info: ProgramInfo,
+    stack_outputs: StackOutputs,
+    final_pc_transcript_state: PrecompileTranscriptState,
+    advice_provider_fingerprint: [u8; 32],
```
bobbinth, line 778: Why do we need `advice_provider_fingerprint`? It seems like it would be non-trivial to compute but not sure if it provides some tangible benefit.

processor/src/fast/execution_api.rs
bobbinth, line 1: I didn't review this file in detail - assuming it mostly copies the code from the module root.

processor/src/fast/execution_api.rs
```diff
@@ -0,0 +1,614 @@
...
+    ) -> Result<TraceBuildInputs, ExecutionError> {
+        let mut tracer = ExecutionTracer::new(self.options.core_trace_fragment_size());
+        let execution_output = self.execute_with_tracer_sync(program, host, &mut tracer)?;
+        Ok(Self::trace_build_inputs_from_parts(program, execution_output, tracer))
+    }
+
+    /// Async variant of [`Self::execute_trace_inputs_sync`] for async hosts.
+    #[inline(always)]
+    #[instrument(name = "execute_trace_inputs", skip_all)]
+    pub async fn execute_trace_inputs(
```
bobbinth, line 83: nit: I don't really love the naming here, but similar to my other comment, don't have a great alternative. Maybe `execute_and_record()`? Then it'll look something like:

```rust
pub async fn execute_and_record(
    self,
    program: &Program,
    host: &mut impl Host,
) -> Result<ExecutionRecording, ExecutionError> {
    ...
}
```

processor/src/trace/trace_state.rs
huitseeker: We need it to reject mixed trace inputs when stack outputs and the final PC transcript still match but the advice state does not. The current implementation canonicalizes the advice-map and Merkle-store state before hashing, so it stays stable across insertion order. I agree this should be revisited as part of the follow-up trace-input redesign.

processor/src/fast/execution_api.rs
huitseeker: Yes, it is mostly moved code from `fast/mod.rs` plus the helper renames and trace-input binding cleanup.

processor/src/host/advice/mod.rs
huitseeker: Added. The docstring now says the fingerprint is stable across advice-map and Merkle-store insertion order.

huitseeker:
Checked blake3 bench, aside from faster build times (due to dropping tokio), no change:
```
  - Merge-base 5541082f2
      - make exec-info: 138.30s
      - prove wall-clock: 6.58s
      - CLI reported: Program proved in 6545 ms
  - Branch tip 0d7a77f50
      - make exec-info: 121.88s
      - prove wall-clock: 6.46s
      - CLI reported: Program proved in 6426 ms
```

I also checked the new advice-binding path on examples that are a better fit for it than `blake3_1to1`. The two examples that are more relevant here are:
  - `hashing/unhashing`, which carries a much larger advice stack
  - `merkle_store`, which exercises the Merkle-store part of the advice provider

  Results:
  - `unhashing`: 54 ms at merge-base, 35 ms on this branch
  - `merkle_store`: 42 ms at merge-base, 49 ms on this branch

  These runs are still very short, so a few milliseconds either way are within normal noise. Taken together, they do not show a clear regression from the advice fingerprint change.

processor/src/trace/mod.rs
huitseeker: #2915

PR #2866, opened 2026-03-17
chore(processor): remove `NodeEndFlags` from `NodeEndData`

### LINKED ISSUE #2644: Remove `NodeFlags` from `NodeEndData`
Follow-up for #2642 

These are no longer needed; they can be inferred in `CoreTraceGenerationTracer::finalize_clock_cycle()` from the continuation and processor state.

plafer:
Closes #2644 

The flags were redundant with the information present on the continuation stack. This ended up cleaning the `BlockStack` quite a bit, as we got rid of `BlockType`, and a lot of complex logic around it.

### DIFF
--- a/processor/src/trace/decoder/block_stack.rs
+++ b/processor/src/trace/decoder/block_stack.rs
@@ -1,14 +1,19 @@
 use alloc::vec::Vec;
 
 use miden_air::Felt;
-use miden_core::{ONE, Word, ZERO};
+use miden_core::{Word, ZERO};
 
 use crate::ContextId;
 
 // BLOCK STACK
 // ================================================================================================
 
-/// Keeps track of code blocks which are currently being executed by the VM.
+/// Tracks per-block data needed for trace generation that the [`ContinuationStack`] does not
+/// carry. Specifically, it stores the hasher chiplet addresses (`addr`, `parent_addr`) assigned
+/// during execution, and for CALL/SYSCALL/DYNCALL blocks, the caller's execution context so it
+/// can be restored on END.
+///
+/// [`ContinuationStack`]: crate::continuation_stack::ContinuationStack
 #[derive(Debug, Default, Clone)]
 pub struct BlockStack {
     blocks: Vec<BlockInfo>,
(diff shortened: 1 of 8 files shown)


processor/src/trace/decoder/block_stack.rs
```diff
@@ -24,64 +24,19 @@ impl BlockStack {
...
-                // or the second child
-                BlockType::Join(first_child_executed) => {
-                    (parent.addr, false, !first_child_executed)
-                },
-                _ => (parent.addr, false, false),
-            },
-            // if the block stack is empty, a new block is neither a body of a loop nor the first
-            // child of a JOIN block; also, we set the parent address to ZERO.
-            None => (ZERO, false, false),
+    pub fn push(&mut self, addr: Felt, ctx_info: Option<ExecutionContextInfo>) -> Felt {
```
bobbinth, line 33: The doc comments here need to be updated (e.g., we no longer need the type - maybe good to explain why?)

processor/src/trace/decoder/block_stack.rs
```diff
@@ -24,64 +24,19 @@ impl BlockStack {
...
     pub fn pop(&mut self) -> BlockInfo {
-        let block = self.blocks.pop().expect("block stack is empty");
-        // if the parent block is a JOIN block (i.e., we just finished executing a child of a JOIN
-        // block) and if the first_child_executed hasn't been set to true yet, set it to true
-        if let Some(parent) = self.blocks.last_mut()
-            && let BlockType::Join(first_child_executed) = parent.block_type
-            && !first_child_executed
-        {
-            parent.block_type = BlockType::Join(true);
-        }
```
bobbinth, line 83: How do we determine now if a given block is the first or second child of a `JOIN` block? Or is is no longer relevant?

processor/src/trace/parallel/tracer/mod.rs
```diff
@@ -345,6 +400,13 @@ impl Tracer for CoreTraceGenerationTracer<'_> {
             self.dyn_callee_hash =
                 self.get_dyn_callee_hash(&continuation, processor, current_forest)?;
 
+            // For END operations, determine if the ending node is a body of a LOOP by checking
+            // whether the next clock-incrementing continuation is a FinishLoop.
+            self.is_loop_body = matches!(
```
huitseeker, line 405: Nit: I wonder if it's worth adding a regression test around this `is_loop_body` derivation? It now depends on continuation ordering, so cases like nested loops (where we'd add another continuation-pushing op) or stop/resume boundaries could silently flip END flags.

If one day we add a book-keeping operation that increments the clock (for e.g. profiling, checkpointing) that inserts itself between the inner-finish and the outer-finish of a nested loop, the “next continuation” is no longer `FinishLoop`, so the inferred flag flips to false.

huitseeker (approved):
LGTM! I left a suggestion around the continuation probing.

processor/src/trace/decoder/block_stack.rs
plafer: It was already irrelevant (concretely the `_is_first_child` variable was not used). It was leftover from when `BlockStack` was used with `Process`, and not sure if it was ever used in the new architecture.

processor/src/trace/parallel/tracer/mod.rs
plafer: Not against adding a regression test, but I'm not completely sure how to write it yet.

> It now depends on continuation ordering, so cases like nested loops (where we'd add another continuation-pushing op) or stop/resume boundaries could silently flip END flags.

For nested loops, the corresponding mast for the loop body would need to be `JOIN(old_body, nested_loop)`, such that the JOIN node is now the loop body. And similarly for bookkeeping operations, our `LoopNode` has a single `MastNodeId` child, and so this would force us to `JOIN` whatever bookkeeping operation we'd want to insert. So I think there's currently no way to mess this up in the future.

>  or stop/resume boundaries could silently flip END flags.

Right, a potential future bug is if we truncate the `ContinuationStack` that we pass in to the `CoreTraceFragmentContext`, such that there is no continuation present (and the flag which should be set to `true` is computed to be `false`). However, this test case is already covered by test `test_trace_generation_at_fragment_boundaries()`, cases 7 & 8.

So I *think* that our current test coverage already account for this possible regression?

processor/src/trace/decoder/block_stack.rs
plafer: Updated. I provided some intuition in `BlockStack` docstring too.

processor/src/trace/decoder/block_stack.rs
```diff
@@ -1,14 +1,19 @@
...
 
-/// Keeps track of code blocks which are currently being executed by the VM.
+/// Tracks per-block data needed for trace generation that the [`ContinuationStack`] does not
+/// carry. Specifically, it stores the hasher chiplet addresses (`addr`, `parent_addr`) assigned
+/// during execution, and for CALL/SYSCALL/DYNCALL blocks, the caller's execution context so it
+/// can be restored on END.
+///
+/// [`ContinuationStack`]: crate::continuation_stack::ContinuationStack
 #[derive(Debug, Default, Clone)]
 pub struct BlockStack {
```
bobbinth, line 18: Another question: can this info not be integrated into `ContinuationStack`? Or is that not a good idea for some reason?

That is - could we get rid of `BlockStack` entirely?

processor/src/trace/decoder/block_stack.rs
bobbinth: Makes sense. Another question: the `is_first_child` flag is set in the trace though, we just have another way of determining it - right? Or did we find a way to get rid of it entirely? (I remember it was required at the constraint level - though, don't remember the exact reason now).

processor/src/trace/parallel/tracer/mod.rs
huitseeker: You're right that the book-keeping operations aren't a precise issue. I think the concrete gap may be nested loops. 

> For nested loops, the corresponding mast for the loop body would need to be JOIN(old_body, nested_loop), such that the JOIN node is now the loop body

That's one one specific shape of nested loop: “keep the old loop body and also add a nested loop beside it.” In that case, yes, you would need something like `JOIN(old_body, nested_loop)`, and then the loop body would be the `JOIN`.

I was thinking of a simpler case where the outer loop’s body is directly another loop:
`(loop (loop (block pad drop)))`

I think that is an ordering-sensitive case: when the inner loop and its body are executing, the outer `FinishLoop` is still part of the continuation structure, so you already have multiple loop-related continuations in play.

I'd try something like:
1. Make the outer loop’s body be another loop,
2. Run the inner loop in the “case 9” shape twice,
3. Run the outer loop in the “case 10” shape around those two inner-loop executions.

I can also (try to) write that as a followup?

processor/src/trace/decoder/block_stack.rs
plafer: The goal is to keep `ContinuationStack` as lean as possible, containing just information necessary for the fast path. Anything else is a "tracer-specific" concern - such as the data stored in `BlockStack`, which is only relevant for trace generation concerns (and hence lives in the `ExecutionTracer`).

processor/src/trace/decoder/block_stack.rs
plafer: > Another question: the is_first_child flag is set in the trace though

No, we never write it in the trace. It is rather used in the block hash table, and is always derived from the opcode in the current/next row (e.g. [this](https://github.com/0xMiden/miden-vm/blob/bd42aeff82da2207b023c1f24cb57ef56a31f34a/processor/src/trace/decoder/aux_trace/block_hash_table.rs#L121-L123) is how we infer that value when encountering an `END` operation). So the processor doesn't need to keep track of it for the purposes of trace generation.

processor/src/trace/parallel/tracer/mod.rs
plafer: Yes, please write it as a follow-up, as I still don't see the potential issue (and hence can't write a good test). In the case of `(loop (loop (block pad drop)))`, both in the `block` and the inner `loop` would have a `FinishLoop` continuation after them in the continuation stack, and hence this `is_loop_body` variable would correctly be populated.

Once again, my reasoning is that every loop body node that is entered incurs [this code path](https://github.com/0xMiden/miden-vm/blob/bd42aeff82da2207b023c1f24cb57ef56a31f34a/processor/src/execution/loop.rs#L58-L59), i.e. right after the start continuation, there is a `FinishLoop {was_entered: true}` continuation. The code snippet linked here is part of `ExecutionTracer::start_clock_cycle()`, and is executed when the `loop_node.body()` node is started. So from what I can tell, _unless we make big changes to how loops work_, then the current solution is pretty robust. I also added a WARNING comment in the `start_loop_node()` function, such that if we ever change how this works, then we need to revisit things here.

But if you can think of a way to capture a clear failure case, please open a follow-up issue.

huitseeker: Here it is, LMK if you think this is valuable: https://github.com/0xMiden/miden-vm/pull/2888

PR #2869, opened 2026-03-18
Update the recursive verifier to work with `p3-miden`

### LINKED ISSUE #2822: Fiat-Shamir: ensure `var_len_public_inputs` and `log_trace_height` are observed in the transcript
These parameters, along with the public inputs, need to be properly observed by the challenger before passing it along to the prover/verifier. This is related to the recursive verifier, since this will dictate what layout we chose.

https://github.com/0xMiden/miden-vm/pull/2798#discussion_r2906231483

Al-Kindi-0:
## Describe your changes
Closes #2822 as well.

A few notes:

- The gamma challenge sampling: The gamma (batching) challenge, for batching multiple ACE circuits evaluations, is not sampled correctly in the recursive verifier. A fix is blocked on upstream changes and will be addressed in a follow-up.
- Fiat-Shamir transcript initialization: The FS transcript setup is functional but the API is awkward. Could be cleaned up further.
- Some work have been done in the direction of #2045 , namely there is now a protocol ID version number which should be bumped up if any of the hardcoded parameters (i.e., the ones that the recursive verifier is not parametric over) change. We also added a very light `AcceptableOptions` check like what we had in Winterfell.

Here is the profile of the recursive verifier execution trace when verifying a program with `2^20` cycles:

```

  ┌────────────────────────────┬────────────────┬───────────────┐
  │         Component          │      Rows      │ % of chiplets │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ Core trace (decoder+stack) │ 41,652         │ —             │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ Range checker              │ 5,129          │ —             │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ Chiplets total             │ 273,769        │ 100%          │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ - Hasher                   │ 250,816        │ 91.6%         │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ - Bitwise                  │ 3,104          │ 1.1%          │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ - Memory                   │ 13,758         │ 5.0%          │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ - ACE                      │ 6,090          │ 2.2%          │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ - Kernel ROM               │ 0              │ 0%            │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ Padded trace length        │ 524,288 (2^19) │               │
  ├────────────────────────────┼────────────────┼───────────────┤
  │ Padding                    │ 47%            │               │
  └────────────────────────────┴────────────────┴───────────────┘

```

It is a bummer that the chiplets' trace overflows `2^18` by a few rows causing the trace length to be `2^19` with almost half of the rows being just padding.
Of course lifted STARK + decoupling the core trace from the chiplets' trace will help significantly with this, but we should also be able to reduce the number of rows in the chiplets' trace.

- Updated `CHANGELOG.md'

### DIFF
--- a/air/src/config.rs
+++ b/air/src/config.rs
@@ -9,19 +9,41 @@ use miden_core::{Felt, field::QuadFelt};
 use miden_crypto::{
     field::Field,
     hash::{
-        blake::Blake3Hasher, keccak::Keccak256Hash, poseidon2::Poseidon2Permutation256,
-        rpo::RpoPermutation256, rpx::RpxPermutation256,
+        blake::Blake3Hasher,
+        keccak::{Keccak256Hash, KeccakF, VECTOR_LEN},
+        poseidon2::Poseidon2Permutation256,
+        rpo::RpoPermutation256,
+        rpx::RpxPermutation256,
     },
     stark::{
         GenericStarkConfig,
         challenger::{CanObserve, DuplexChallenger, HashChallenger, SerializingChallenger64},
         dft::Radix2DitParallel,
         fri::PcsParams,
-        hasher::{ChainingHasher, StatefulSponge},
+        hasher::{ChainingHasher, SerializingStatefulSponge, StatefulSponge},
         lmcs::LmcsConfig,
-        symmetric::{CompressionFunctionFromHasher, Permutation, TruncatedPermutation},
+        symmetric::{
+            CompressionFunctionFromHasher, CryptographicPermutation, PaddingFreeSponge,
+            TruncatedPermutation,
+        },
     },
 };
+
+// SHARED TYPES
+// ================================================================================================
+
+/// Miden VM STARK configuration with pre-filled common type parameters.
+///
+/// All Miden configurations use `Felt` as the base field, `QuadFelt` as the extension field,
+/// and `Radix2DitParallel<Felt>` as the DFT. Only the LMCS commitment scheme (`L`) and
+/// Fiat-Shamir challenger (`Ch`) vary by hash function.
+pub type MidenStarkConfig<L, Ch> =
+    GenericStarkConfig<Felt, QuadFelt, L, Radix2DitParallel<Felt>, Ch>;
+
+type PackedFelt = <Felt as Field>::Packing;
+
+/// Number of inputs to the Merkle compression function.
+const COMPRESSION_INPUTS: usize = 2;
 
 // PCS PARAMETERS
 // ================================================================================================
@@ -55,126 +77,6 @@ pub fn pcs_params() -> PcsParams {
     .expect("invalid PCS parameters")
 }
 
-// HASH FUNCTION PARAMETERS
-// ================================================================================================
-
-// Byte-oriented hashes (Blake3, Keccak).
-
-/// Digest size in bytes for byte-oriented hashes.
-const BYTE_DIGEST_SIZE: usize = 32;
-/// Number of inputs to the Merkle compression function.
-const COMPRESSION_INPUTS: usize = 2;
-
-// Algebraic hashes (RPO, Poseidon2, RPX).
-
-/// Sponge state width in field elements.
-const SPONGE_WIDTH: usize = 12;
-/// Sponge rate (absorbable elements per permutation).
-const SPONGE_RATE: usize = 8;
-/// Sponge digest width in field elements.
-const DIGEST_WIDTH: usize = 4;
-
-// SHARED TYPE ALIASES
-// ================================================================================================
-
-type PackedFelt = <Felt as Field>::Packing;
-
-/// Miden VM STARK configuration with pre-filled common type parameters.
-///
-/// All Miden configurations use `Felt` as the base field, `QuadFelt` as the extension field,
-/// and `Radix2DitParallel<Felt>` as the DFT. Only the LMCS commitment scheme (`L`) and
-/// Fiat-Shamir challenger (`Ch`) vary by hash function.
-pub type MidenStarkConfig<L, Ch> =
-    GenericStarkConfig<Felt, QuadFelt, L, Radix2DitParallel<Felt>, Ch>;
-
-/// Byte-oriented LMCS (for Blake3, Keccak).
-type ByteLmcs<H> = LmcsConfig<
-    Felt,
-    u8,
-    ChainingHasher<H>,
-    CompressionFunctionFromHasher<H, COMPRESSION_INPUTS, BYTE_DIGEST_SIZE>,
-    BYTE_DIGEST_SIZE,
-    BYTE_DIGEST_SIZE,
->;
-
-/// Byte-oriented challenger (for Blake3, Keccak).
-type ByteChallenger<H> = SerializingChallenger64<Felt, HashChallenger<u8, H, BYTE_DIGEST_SIZE>>;
-
-/// Algebraic LMCS (for RPO, Poseidon2, RPX).
-type AlgLmcs<P> = LmcsConfig<
-    PackedFelt,
-    PackedFelt,
-    StatefulSponge<P, SPONGE_WIDTH, SPONGE_RATE, DIGEST_WIDTH>,
-    TruncatedPermutation<P, COMPRESSION_INPUTS, DIGEST_WIDTH, SPONGE_WIDTH>,
-    SPONGE_WIDTH,
-    DIGEST_WIDTH,
->;
-
-/// Algebraic duplex challenger (for RPO, Poseidon2, RPX).
-type AlgChallenger<P> = DuplexChallenger<Felt, P, SPONGE_WIDTH, SPONGE_RATE>;
-
-// CONFIGURATION FACTORIES
-// ================================================================================================
-
-/// Creates a Blake3_256-based STARK configuration.
-pub fn blake3_256_config(
-    params: PcsParams,
-) -> MidenStarkConfig<ByteLmcs<Blake3Hasher>, ByteChallenger<Blake3Hasher>> {
-    let lmcs = LmcsConfig::new(
-        ChainingHasher::new(Blake3Hasher),
-        CompressionFunctionFromHasher::new(Blake3Hasher),
-    );
-    let challenger = SerializingChallenger64::from_hasher(vec![], Blake3Hasher);
-    GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
-}
-
-/// Creates a Keccak-based STARK configuration.
-pub fn keccak_config(
-    params: PcsParams,
-) -> MidenStarkConfig<ByteLmcs<Keccak256Hash>, ByteChallenger<Keccak256Hash>> {
-    let hash = Keccak256Hash {};
-    let lmcs = LmcsConfig::new(ChainingHasher::new(hash), CompressionFunctionFromHasher::new(hash));
-    let challenger = SerializingChallenger64::from_hasher(vec![], hash);
-    GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
-}
-
-/// Creates an RPO-based STARK configuration.
-pub fn rpo_config(
-    params: PcsParams,
-) -> MidenStarkConfig<AlgLmcs<RpoPermutation256>, AlgChallenger<RpoPermutation256>> {
-    let perm = RpoPermutation256;
-    let lmcs = LmcsConfig::new(StatefulSponge::new(perm), TruncatedPermutation::new(perm));
-    let challenger = DuplexChallenger::new(perm);
-    GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
-}
-
-/// Creates a Poseidon2-based STARK configuration.
-pub fn poseidon2_config(
-    params: PcsParams,
-) -> MidenStarkConfig<AlgLmcs<Poseidon2Permutation256>, AlgChallenger<Poseidon2Permutation256>> {
-    let perm = Poseidon2Permutation256;
-    let lmcs = LmcsConfig::new(StatefulSponge::new(perm), TruncatedPermutation::new(perm));
-    let challenger = DuplexChallenger::new(perm);
-    GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
-}
-
-/// Creates an RPX-based STARK configuration.
-pub fn rpx_config(
-    params: PcsParams,
-) -> MidenStarkConfig<AlgLmcs<RpxPermutation256>, AlgChallenger<RpxPermutation256>> {
-    let perm = RpxPermutation256;
-    let lmcs = LmcsConfig::new(StatefulSponge::new(perm), TruncatedPermutation::new(perm));
-    let challenger = DuplexChallenger::new(perm);
-    GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
-}
-
-// POSEIDON2 CONFIG TYPE ALIAS
-// ================================================================================================
-
-/// Concrete STARK configuration type for Poseidon2.
-pub type Poseidon2Config =
-    MidenStarkConfig<AlgLmcs<Poseidon2Permutation256>, AlgChallenger<Poseidon2Permutation256>>;
-
 // DOMAIN-SEPARATED FIAT-SHAMIR TRANSCRIPT
 // ================================================================================================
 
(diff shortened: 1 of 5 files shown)


crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -23,211 +50,393 @@ pub struct VerifierData {
...
+
+    // --- advice stack ---
+    let mut advice_stack = Vec::new();
+
+    // 0. Security parameters: [num_queries, query_pow_bits, deep_pow_bits, folding_pow_bits].
+    //    Consumed first by load_security_params in the specific verifier.
+    let num_queries = pcs.tree_indices.len();
+    advice_stack.push(num_queries as u64);
+    let params = config::pcs_params();
+    advice_stack.push(params.query_pow_bits() as u64);
```
huitseeker, line 145: It looks like this uses the number of realized query indices, not the configured query count.

We already noted this (FRI soundness taking into the gap between observed result & protocol setting when sampling w/ replacement), just noting in passing.

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -95,205 +64,368 @@ end
...
+#!
+#! RELATION_DIGEST = hash(PROTOCOL_ID, CIRCUIT_COMMITMENT) is a compile-time constant
+#! computed once per AIR and pushed by the caller (the specific verifier).
+#!
+#! Currently assumes a blowup factor equal to 8.
+#!
+#! Precondition: num_queries, query_pow_bits, deep_pow_bits, and folding_pow_bits must
+#! already be stored in memory before calling this procedure.
+#!
+#! Input: [log(trace_length), rd0, rd1, rd2, rd3, ...]
```
huitseeker, line 166: This looks right for the verifier path. Non-blocking, but would it help to update the low-level direct `init_seed` callers too? A few tests still seem to use the older stack shape, so they can pass without really covering this new contract. A small follow-up or wrapper would keep this seed path exercised.

crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -1,21 +1,74 @@
...
+
+# Acceptable security parameters for Miden VM proof verification.
+# These define the security policy enforced by `assert_acceptable_options`.
+# If the protocol security level changes, update these constants accordingly.
+const ACCEPTABLE_NUM_QUERIES = 27
+const ACCEPTABLE_QUERY_POW_BITS = 0
+const ACCEPTABLE_DEEP_POW_BITS = 0
+const ACCEPTABLE_FOLDING_POW_BITS = 16
+
+# RELATION_DIGEST = hash(PROTOCOL_ID, CIRCUIT_COMMITMENT).
```
huitseeker, line 18: These values now look hand-copied across the Rust circuit build and the MASM verifier. Non-blocking, but would it help to add one regen test that checks `NUM_INPUTS_CIRCUIT`, `NUM_EVAL_GATES_CIRCUIT`, `CIRCUIT_COMMITMENT`, and `RELATION_DIGEST` against `build_batched_ace_circuit()`? That would make future updates much safer.

crates/lib/core/asm/stark/utils.masm
```diff
@@ -19,110 +18,317 @@ end
...
+#! consuming it. This gives gamma == deep_alpha (the next challenge that will be
+#! properly sampled in generate_deep_composition_random_coefficients). Using the
+#! same value for both is unsound.
+#!
+#! TODO: Replace with a proper `sample_ext` call once the Rust-side
+#! StarkTranscript is updated to draw gamma between OOD and DEEP phases.
+#!
+#! Input:  [...]
+#! Output: [gamma_0, gamma_1, ...]
+proc sample_gamma
```
Al-Kindi-0, line 289: Exactly, this is just a workaround.
I am going to create an issue to track this. We can delay making any changes to the backend as we might not need to sample this gamma if our auxiliary trace gets simpler after the migration to the multi-table setting.

crates/lib/core/tests/stark/verifier_recursive/mod.rs
Al-Kindi-0: Good point, we do indeed need the configured and not the potentially de-duplicated number of queries

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: I have a local test which basically does just that, it regenerates the constraints evaluation MASM module. I didn't include it because it was low quality code-wise.
Will try to come up with something

crates/lib/core/tests/stark/verifier_recursive/mod.rs
Al-Kindi-0: Fixed now to use the parameter in `config::pcs_params()`

crates/lib/core/asm/stark/random_coin.masm
Al-Kindi-0: Good catch! I now fixed the tests to the new contract

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Added in https://github.com/0xMiden/miden-vm/pull/2869/commits/487c6cf5c27700d4e49e72465675b96bf41fa4aa

processor/src/execution/operations/fri_ops/mod.rs
```diff
@@ -23,14 +23,19 @@ mod tests;
 /// - Checks that the previous folding was done correctly.
 /// - Shifts the stack to the left to move an item from the overflow table to stack position 15.
 ///
+/// Bit-reversal handling:
+/// - Query values exist on the stack in bit-reversed order.
+/// - p on the stack is a bit-reversed tree index. The instruction internally computes d_seg = p & 3
+///   (bit-reversed coset index) for the consistency check and tau factor. f_pos is provided
+///   separately on the stack and the AIR constrains f_pos = p >> 2 via p = 4*f_pos + d_seg.
```
bobbinth, line 30: So, in this PR, we are using bit-reversed indexes, right? If so, should we first evaluate the impact of doing the extra bit reversal in the prover (to put indexes into the "natural" order)? If the impact is something like making prover 5% slower, but it makes recursive verifier 20% faster, I think we should make this trade-off.

crates/ace-codegen/src/lib.rs
bobbinth, line 1: Are the changes to the `ace-codegen` intertwined with the rest of the recursive verifier? If not, maybe it makes sense to split them out into a stand-alone PR to make reviews easier?

Al-Kindi-0: Indeed, they are, though the changes are not too many in the `ace-codegen` crate.

processor/src/execution/operations/fri_ops/mod.rs
Al-Kindi-0: Agreed

crates/ace-codegen/src/randomness.rs
```diff
@@ -37,16 +37,11 @@ where
     let num = layout.counts.num_randomness;
     assert!(index < num, "challenge index {index} out of range (num={num})");
 
-    if index == 0 {
-        return builder.input(InputKey::AuxRandAlpha);
+    match index {
+        0 => builder.input(InputKey::AuxRandAlpha),
+        1 => builder.input(InputKey::AuxRandBeta),
+        _ => panic!(
```
huitseeker, line 43: This now panics for any challenge index above `1`, but `num_randomness` still comes from the AIR and can be larger. 

Could  a valid AIR can make circuit generation blow up here, and if so should we assert `num_randomness <= 2` earlier?

crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -23,211 +50,395 @@ pub struct VerifierData {
...
+fn infer_widths<F, C>(batch: &BatchProof<F, C>) -> Vec<usize> {
+    batch
+        .openings
+        .values()
+        .next()
+        .map(|opening| opening.rows.iter_rows().map(|row| row.len()).collect())
+        .unwrap_or_default()
+}
+
+fn infer_log_height_from_indices(indices: &[usize]) -> usize {
```
huitseeker: I think this helper is inferring tree height from the biggest queried index. Is that the real tree height `single_proofs` wants?

If all sampled indices happen to land in a low subtree, I think this can build paths that are too short even though the proof is fine.

crates/ace-codegen/src/layout/keys.rs
```diff
@@ -84,18 +83,19 @@ impl InputKeyMapper<'_> {
                 }
             },
             InputKey::AuxBusBoundary(i) => layout.regions.aux_bus_boundary.index(i),
-            InputKey::Z => Some(layout.stark.z),
+            InputKey::VlpiReduction(i) => layout.regions.vlpi_reductions.index(i),
```
Al-Kindi-0: I don't think so,  the MASM layout pre-doubles `num_vlpi` in `pipeline.rs` (`config.num_vlpi_groups * 2`)

crates/ace-codegen/src/randomness.rs
Al-Kindi-0: Agreed, to simplify changed to assert equality with `2`

crates/lib/core/tests/stark/verifier_recursive/mod.rs
Al-Kindi-0: Good point, changed it to use the config's tree height

air/src/config.rs
```diff
@@ -167,3 +167,182 @@ pub fn rpx_config(
     let challenger = DuplexChallenger::new(perm);
     GenericStarkConfig::new(params, lmcs, Radix2DitParallel::default(), challenger)
 }
+
+// POSEIDON2 CONFIG TYPE ALIAS
```
adr1anh: We should create an issue to make this bit cleaner. Ideally we would just need to use the `observe` method on the transcript returned by the config trait. The only thing that seems to be missing here would be the ability to add the initial domain separatation right?

In the mean time, could we just use the same codepath as init_transcript_hash?

processor/src/execution/operations/fri_ops/mod.rs
adr1anh: The PR https://github.com/0xMiden/p3-miden/pull/52 only adds one bit-reversal on the prover side, which I've benchmarked with different sized below. This definitely seems worth it as the savings would be larger for the recursive verifier. 

## Bit-reversal permutation benchmarks (Apple M2 Pro)

Element size: 32 bytes for both variants.

| log₂(n) | n | Goldilocks×4 (32B) | Bytes32 (32B) |
|----------|-----------|-------------------|--------------|
| 16 | 65,536 | 182 µs | 180 µs |
| 17 | 131,072 | 351 µs | 344 µs |
| 18 | 262,144 | 739 µs | 677 µs |
| 19 | 524,288 | 1.52 ms | 1.42 ms |
| 20 | 1,048,576 | 3.38 ms | 3.54 ms |

Both types are 32 bytes so performance is essentially identical — the bit-reversal permutation is purely memory-bound (random swaps over large arrays). ~2× scaling per doubling is expected as the working set outgrows cache.

crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -23,211 +50,389 @@ pub struct VerifierData {
...
+fn build_fixed_len_inputs(pub_inputs: &PublicInputs) -> Vec<u64> {
+    let mut fixed_len = Vec::new();
+    fixed_len
+        .extend_from_slice(&felts_to_u64(pub_inputs.program_info().program_hash().as_elements()));
+    fixed_len.extend_from_slice(&felts_to_u64(pub_inputs.stack_inputs().as_ref()));
+    fixed_len.extend_from_slice(&felts_to_u64(pub_inputs.stack_outputs().as_ref()));
+    fixed_len.extend_from_slice(&felts_to_u64(pub_inputs.pc_transcript_state().as_ref()));
+    fixed_len.resize(fixed_len.len().next_multiple_of(8), 0);
+    fixed_len
+}
```
adr1anh, line 423: Nit: we can create a Vec of Felt and then map.collect by mapping to u64. Compiler probably won't re-allocate

crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -23,211 +50,389 @@ pub struct VerifierData {
...
+
+// HELPERS
 // ================================================================================================
 
-pub fn digest_to_int_vec(digest: &[Word]) -> Vec<u64> {
-    digest
-        .iter()
-        .flat_map(|digest| digest.as_elements().iter().map(|e| e.as_canonical_u64()))
-        .collect()
+fn infer_widths<F, C>(batch: &BatchProof<F, C>) -> Vec<usize> {
```
adr1anh, line 385: Noting that this would be a useful method on the BatchProof.

air/src/config.rs
```diff
@@ -167,3 +167,182 @@ pub fn rpx_config(
...
+/// Returns a `[Felt; SPONGE_WIDTH]` with capacity = PROOF_SEED and rate zeroed.
+fn seed_sponge_state(
+    log_trace_height: u64,
+    permute: impl Fn(&mut [Felt; SPONGE_WIDTH]),
+) -> [Felt; SPONGE_WIDTH] {
+    let mut state = [Felt::ZERO; SPONGE_WIDTH];
+
+    // Phase 1: RELATION_DIGEST (capacity) + PCS_PARAMS (rate) -> INSTANCE_SEED
+    assert_eq!(RELATION_DIGEST.len(), SPONGE_CAPACITY);
+    state[CAPACITY_RANGE].copy_from_slice(&RELATION_DIGEST);
```
adr1anh: Passing this state to the duplex challenger is enough to the use the same observe methods and unify both byte and field based challenger initializations.

crates/ace-codegen/src/pipeline.rs
```diff
@@ -154,15 +135,26 @@ where
     );
 
     let num_randomness = air.num_randomness();
-    assert!(num_randomness > 0, "AIR must declare at least one randomness challenge");
+    assert!(
```
adr1anh, line 138: A comment for above, but the method air.air_layout exists.

adr1anh: For above too, we could match air.log_quotient_degree against `config.num_quotient_chunks`  (and we should probably overwrite the method in the ProcessorAir to avoid re-creating the SymbolicAir).

air/src/config.rs
```diff
@@ -167,3 +167,182 @@ pub fn rpx_config(
...
+/// Sponge capacity (SPONGE_WIDTH - SPONGE_RATE).
+const SPONGE_CAPACITY: usize = SPONGE_WIDTH - SPONGE_RATE;
+
+/// Range of capacity slots within the sponge state array.
+const CAPACITY_RANGE: core::ops::Range<usize> = SPONGE_RATE..SPONGE_WIDTH;
+
+/// Computes the seeded sponge state for sponge-based challengers.
+///
+/// Returns a `[Felt; SPONGE_WIDTH]` with capacity = PROOF_SEED and rate zeroed.
+fn seed_sponge_state(
```
Nashtare: Two random comments that you can feel free to dismiss given my lack of expertise of the codebase:

- out of consistency / ease of review / ease of maintenance regarding the mapping of Rust code and MASM, have we ever considered forcing entry points & higher-level function names to match? For instance here I believe `seed_sponge_state` is equivalent to the MASM `init_seed` label.
- do we have lower level consistency tests enforcing the MASM logic to match the Rust one, apart from global tests?

air/src/config.rs
```diff
@@ -167,3 +167,182 @@ pub fn rpx_config(
...
+/// RELATION_DIGEST = Poseidon2::hash_elements([PROTOCOL_ID, CIRCUIT_COMMITMENT]).
+///
+/// Compile-time constant binding the Fiat-Shamir transcript to the Miden VM AIR.
+/// Must match the constants in `crates/lib/core/asm/sys/vm/mod.masm`.
+pub const RELATION_DIGEST: [Felt; 4] = [
+    Felt::new(9663888320842941557),
+    Felt::new(5569923100392661778),
+    Felt::new(10686243500486164404),
+    Felt::new(9017524969302659247),
+];
```
Nashtare, line 92: nit: is the relation `RELATION_DIGEST = Poseidon2::hash_elements([PROTOCOL_ID, CIRCUIT_COMMITMENT])` enforced in some unit test to trigger CI failure upon mismatch?

air/src/config.rs
```diff
@@ -167,3 +167,182 @@ pub fn rpx_config(
...
+            permutation: RpxPermutation256,
+        }
+    }
+}
+
+/// Helper for bit-oriented `InitTranscript` implementations.
+///
+/// Unlike the sponge-based algebraic challengers (RPO, Poseidon2, RPX) which seed the
+/// capacity directly via `seed_sponge_state`, bit-oriented challengers observe the
+/// RELATION_DIGEST, PCS parameters, and log_trace_height sequentially as a prefix.
```
Nashtare: just for my understanding, why is that?

crates/ace-codegen/src/layout/keys.rs
huitseeker: [We discussed this earlier](https://github.com/0xMiden/miden-vm/pull/2869#discussion_r2973750144).

```rust
  #[test]
  fn masm_layout_vlpi_groups_use_word_stride() {
      let counts = InputCounts {
          width: 1,
          aux_width: 1,
          num_public: 8,
          // Two logical VLPI groups in MASM occupy four EF slots total:
          // [group0, pad0, group1, pad1].
          num_vlpi: 4,
          num_randomness: 2,
          num_periodic: 0,
          num_quotient_chunks: 1,
      };
      let layout = InputLayout::new_masm(counts);

      let vlpi_base = layout.index(InputKey::VlpiReduction(0)).unwrap();
      assert_eq!(layout.index(InputKey::VlpiReduction(0)), Some(vlpi_base));
      assert_eq!(
          layout.index(InputKey::VlpiReduction(1)),
          Some(vlpi_base + 2),
          "MASM VLPI groups should advance by a word-aligned stride"
      );
  }
```

On this branch it fails with left: Some(9) vs right: Some(10), which is the padding-slot bug.
Should this use a layout-specific stride here?

air/src/config.rs
Al-Kindi-0: Good points. One of the reasons of the divergence is related to the Winterfell to Plonky3 change.
As for unit tests, there are some for the most critical procedures but not for all.
Created an issue to track this #2923

air/src/config.rs
Al-Kindi-0: I think `relation_digest_matches_air` is what you are thinking of

air/src/config.rs
Al-Kindi-0: Created #2924
I have cleaned up a bit the handling of the algebraic variants but I would defer anything more till we solve the issue so as to avoid redundant work on the MASM side.

air/src/config.rs
Al-Kindi-0: once we get the initialization method we would be able to unify
Tracked by the issue referenced above

crates/ace-codegen/src/layout/keys.rs
Al-Kindi-0: fixed in the latest commit

air/src/config.rs
adr1anh: Thanks for looking at this. I realized that this would have led to annoying conflicts with #2874. 

I think we can close #2924 with this commit, but I created https://github.com/0xMiden/miden-vm/issues/2926 which includes larger points to improve the way the way we build the ACE and therefore the full commitment.

air/src/config.rs
Al-Kindi-0: Mostly for efficiency of the recursive verifier so that we can reduce at least one call to the permutation online by doing some work offline

PR #2870, opened 2026-03-18
fix(assembly-syntax): remove UB in Meta::from_iter

### LINKED ISSUE #2857: Undefined behaviour in safe attribute metadata conversion
The `FromIterator<MetaItem> for Meta` implementation at `crates/assembly-syntax/src/ast/attribute/meta.rs:41-64` uses `unsafe { core::hint::unreachable_unchecked() }` when the iterator contains mixed `MetaItem` variants. This is undefined behaviour reachable through a safe API.

The function decides the output type based on the first element (`Meta::List` for `Expr`, `Meta::KeyValue` for key-value pairs), then assumes all remaining elements match. If they don't, it hits `unreachable_unchecked()` at lines 51 and 59.

A mixed iterator triggers UB:
```rust
let mixed = vec![MetaItem::from("x"), MetaItem::from(("k", "v"))];
let _ = Meta::from_iter(mixed);
```

The parser only produces homogeneous iterators, so untrusted MASM input can't reach this. But downstream code constructing mixed iterators from external input would hit UB.

Fix by replacing both `unreachable_unchecked()` calls with `unreachable!()` and a short message about mixed variants.

crazywriter1:
## Summary
- Replace `unreachable_unchecked()` with `unreachable!()` in `Meta::from_iter` to avoid UB reachable via a safe API.

Fixes #2857

## Test plan
- cargo test -p miden-assembly-syntax
- cargo clippy -p miden-assembly-syntax --all-targets --all-features -- -D warnings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,20 @@
 # Changelog
 
-## 0.22.0 (TBD)
+## v0.23.0 (TBD)
 
-#### Major breaking changes
+## 0.22.0 (2025-03-18)
 
+#### Enhancements
+
+- Define and implement Miden project file format ([#2510](https://github.com/0xMiden/miden-vm/pull/2510)).
+- Added `math::u128` comparison (`lt`, `lte`, `gt`, `gte`), bitwise (`and`, `or`, `xor`, `not`), and shift (`shl`, `shr`, `rotl`, `rotr`) operations ([#2624](https://github.com/0xMiden/miden-vm/pull/2624)).
 - [BREAKING] `build_trace()` no longer assumes valid user input ([#2747](https://github.com/0xMiden/miden-vm/pull/2747)).
-- [BREAKING] Migrate to lifted-STARK backend and `miden-crypto` to v0.23 ([#2783](https://github.com/0xMiden/miden-vm/pull/2783)).
+- Added `math::u128` division operations ([#2776](https://github.com/0xMiden/miden-vm/pull/2776)).
+- [BREAKING] Migrated to lifted-STARK backend and `miden-crypto` to v0.23 ([#2783](https://github.com/0xMiden/miden-vm/pull/2783)).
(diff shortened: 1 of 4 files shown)


crazywriter1:
CI workflows require approval to run (fork PR). Can a maintainer approve the workflows?

crates/assembly-syntax/src/ast/attribute/meta.rs
```diff
@@ -48,15 +48,19 @@ impl FromIterator<MetaItem> for Meta {
                 core::iter::once(expr)
                     .chain(iter.map(|item| match item {
                         MetaItem::Expr(expr) => expr,
-                        MetaItem::KeyValue(..) => unsafe { core::hint::unreachable_unchecked() },
+                        MetaItem::KeyValue(..) => {
+                            unreachable!("mixed MetaItem variants in iterator: expected Expr")
+                        },
                     }))
```
plafer, line 54: @huitseeker the issue description does mention this as a resolution, but from the premise it seems like this is indeed reachable. Should we not be returning an error here instead?

crazywriter1: Thanks for the review and suggestion. I went with `unreachable!()` because the issue explicitly asked to replace `unreachable_unchecked()` that way to remove UB, and it fixes the immediate problem (no UB from safe API).

Returning an error instead would be nicer for callers (they can handle mixed variants instead of panicking), but `FromIterator` only yields `Self`, not `Result<Self, E>`, so we’d need a different API or a custom collector to surface an error. Happy to switch to an error-returning approach if you and @huitseeker prefer that direction.

bitwalker: The parser does not permit mixed item lists, so in practice this is unreachable, I think the panic here is fine for now

PR #2871, opened 2026-03-18
fix(assembly): correct cycle detection in toposort_caller

### LINKED ISSUE #2827: toposort_caller cycle check is incorrect
In `crates/assembly-syntax/src/sema/passes/toposort.rs:73`, the cycle detection logic for `toposort_caller` checks:

```rust
if visited.contains(&target) {
    return Err(SemanticAnalysisError::CycleInCallGraph { span: node_span });
}
```

This does not distinguish between nodes on the current DFS path (actual cycles) and nodes visited in earlier branches (DAG edges). The check should use a "currently visiting" set that tracks nodes on the recursion stack, not all visited nodes.

As a result, valid call graphs with shared callees may incorrectly report a cycle. The current code works in practice because the visitor is invoked per-procedure, so cross-branch revisits are rare, but the logic is fragile.

Fix: introduce a `visiting: HashSet<NodeId>` that is populated before recursing and cleared after, and check `visiting.contains(&target)` instead of `visited`.

crazywriter1:
## Summary
- Fix cycle detection in `CallGraph::toposort_caller` by using Kahn's algorithm invariant (`output.len() != graph.nodes.len()`) instead of checking edges only for nodes already in the output list. This avoids false positives when edges remain on nodes that were never popped into `output`.

Fixes #2827

## Notes (file location)
The issue description references `crates/assembly-syntax/src/sema/passes/toposort.rs`, but in the current `next` branch the relevant implementation lives in:
- `crates/assembly/src/linker/callgraph.rs`

I found it by searching the repository for `toposort_caller`.

## Test plan
- cargo test -p miden-assembly
- cargo clippy -p miden-assembly --all-targets --all-features -- -D warnings

### DIFF
--- a/crates/assembly/src/linker/callgraph.rs
+++ b/crates/assembly/src/linker/callgraph.rs
@@ -176,6 +176,27 @@ impl CallGraph {
         graph
     }
 
+    /// Computes the set of nodes in this graph which can reach `root`.
+    fn reverse_reachable(&self, root: GlobalItemIndex) -> BTreeSet<GlobalItemIndex> {
+        let mut worklist = VecDeque::from_iter([root]);
+        let mut visited = BTreeSet::default();
+
+        while let Some(gid) = worklist.pop_front() {
+            if !visited.insert(gid) {
+                continue;
+            }
+
+            worklist.extend(
+                self.nodes
+                    .iter()
+                    .filter(|(_, out_edges)| out_edges.contains(&gid))
+                    .map(|(pred, _)| *pred),
+            );
+        }
+
+        visited
+    }
+
     /// Constructs the topological ordering of nodes in the call graph, for which `caller` is an
     /// ancestor.
     ///
@@ -188,16 +209,16 @@ impl CallGraph {
         let mut output = Vec::with_capacity(self.nodes.len());
 
         // Build a subgraph of `self` containing only those nodes reachable from `caller`
-        let mut graph = self.subgraph(caller);
+        let caller_subgraph = self.subgraph(caller);
+        let mut graph = caller_subgraph.clone();
 
-        // Check if any node in the subgraph has an edge back to caller before we erase them
-        let has_edge_to_caller = graph.nodes.values().any(|edges| edges.contains(&caller));
-        let caller_predecessors: BTreeSet<GlobalItemIndex> = graph
+        // Preserve the full set of nodes participating in cycles that close back into `caller`
+        // before we erase those back-edges to seed the traversal from `caller`.
+        let caller_cycle = caller_subgraph
             .nodes
-            .iter()
-            .filter(|(_, edges)| edges.contains(&caller))
-            .map(|(n, _)| *n)
-            .collect();
+            .values()
+            .any(|edges| edges.contains(&caller))
+            .then(|| caller_subgraph.reverse_reachable(caller));
 
         // Remove all predecessor edges to `caller`
         graph.nodes.iter_mut().for_each(|(_pred, out_edges)| {
@@ -218,7 +239,7 @@ impl CallGraph {
             }
         }
 
-        let has_cycle = output.len() != graph.nodes.len() || has_edge_to_caller;
+        let has_cycle = output.len() != graph.nodes.len() || caller_cycle.is_some();
         if has_cycle {
             let mut in_cycle = BTreeSet::default();
             for (n, edges) in graph.nodes.iter() {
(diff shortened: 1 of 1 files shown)


crazywriter1:
CI workflows require approval to run (fork PR). Can a maintainer approve the workflows?

crates/assembly/src/linker/callgraph.rs
```diff
@@ -209,10 +209,7 @@ impl CallGraph {
             }
         }
 
-        let has_cycle = graph
-            .nodes
-            .iter()
-            .any(|(n, out_edges)| output.contains(n) && !out_edges.is_empty());
+        let has_cycle = output.len() != graph.nodes.len();
```
huitseeker: This invariant looks right, but I do not see a `toposort_caller` regression test for the case where a reachable cycle keeps some nodes from ever entering `output`. Could we add one here? The current tests cover the acyclic path and `toposort()`, but not this branch.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
 - [BREAKING] Enum types are now exported from libraries as a `midenc_hir_type::EnumType`, rather than the type of the discriminant. ([#2848](https://github.com/0xMiden/miden-vm/pull/2848))
 
 #### Fixes
+- **toposort_caller**: Fixed cycle detection in assembly call graph ([#2871](https://github.com/0xMiden/miden-vm/pull/2871)).
```
huitseeker, line 41: This looks like it inserted the same entry twice into every historical `#### Fixes` section, not just the current release block. Please fix.

crazywriter1:
- Added regression test for toposort_caller with a reachable cycle
- Fixed CHANGELOG to only have the entry in the current release section
@huitseeker

crates/assembly/src/linker/callgraph.rs
huitseeker: This new invariant fixes the downstream-cycle case, but I think `toposort_caller(A2)` on `callgraph_cycle()` still returns `Ok(...)`. We still drop every edge into `caller` a few lines above, so the backedge `A3 -> A2` gets erased before this check runs. 

Could we add a regression for a cycle that closes back into the root, or keep that edge visible to cycle detection?

crates/assembly/src/linker/callgraph.rs
```diff
@@ -326,6 +323,21 @@ mod tests {
         assert_eq!(err.0.into_iter().collect::<Vec<_>>(), &[A2, A3, B2, B3]);
     }
 
+    #[test]
+    fn callgraph_toposort_caller_with_reachable_cycle() {
```
huitseeker, line 360: Could we make this test cover the root-closing case too, like `callgraph_cycle().toposort_caller(A2).expect_err(...)`? 

That is the case still [missed today that I was commenting on above](https://github.com/0xMiden/miden-vm/pull/2871/changes#r2973733098): `retain(|n| *n != caller)` removes the `A3 -> A2` backedge before `has_cycle` runs. A simple fix would be to record whether the reachable subgraph had any edge into `caller` before that `retain`, then use that flag together with `output.len() != graph.nodes.len()` when deciding if a cycle exists.

crazywriter1:
Thanks for the detailed feedback sir. Added root closing cycle detection: recording edges into `caller` before `retain` and using that flag in `has_cycle`. Also added regression test for `toposort_caller(A2)` on the cycle graph. @huitseeker

CHANGELOG.md
```diff
@@ -37,7 +37,7 @@
 - In `ExecutionTracer`, we no longer record node flags in `CoreTraceFragmentContext` when entering a node (they are redundant) ([#2866](https://github.com/0xMiden/miden-vm/pull/2866))
 
 #### Fixes
-- Fixed C-like enum validation and constant materialization in `define_enum` ([#2887](https://github.com/0xMiden/miden-vm/pull/2887)).
+- **toposort_caller**: Fixed cycle detection in assembly call graph ([#2871](https://github.com/0xMiden/miden-vm/pull/2871)).
```
huitseeker, line 41: I think this line accidentally replaces the `#2887` changelog item instead of adding `#2871` next to it. Since `#2887` is already merged into `next`, this drops that fix from the release notes. Can we keep both entries here?

crazywriter1: It already was like that but it caused a conflict. I'm fixing it now, sorry.

crazywriter1:
Restored the #2887 changelog entry. Both fixes are now listed under the same section. Sorry about that. @huitseeker

crates/assembly/src/linker/callgraph.rs
```diff
@@ -222,6 +228,10 @@ impl CallGraph {
                 in_cycle.insert(*n);
                 in_cycle.extend(edges.as_slice());
             }
+            if has_edge_to_caller {
+                in_cycle.insert(caller);
+                in_cycle.extend(caller_predecessors.iter());
```
huitseeker: This still drops part of the cycle in the root-closing case. For `A2 -> B2 -> B3 -> A3 -> A2`, this branch only rebuilds `{A2, A3}` after the traversal has already erased `B2` and `B3`'s outgoing edges, so the linker error hides two recursive procedures. 

We should derive cycle members from the original reachable subgraph.

bitwalker (approved):

PR #2874, opened 2026-03-18
refactor: use stateful binary sponge for Keccak STARK config

adr1anh:
## Summary

- Switch Keccak LMCS from byte-oriented `ChainingHasher<Keccak256Hash>` to `SerializingStatefulSponge<StatefulSponge<KeccakF, 25, 17, 4>>` with `[Felt; VECTOR_LEN]` packing, matching the pattern used in p3-miden
- Reorganize `air/src/config.rs` so each hash family (Blake3, Keccak, algebraic) has its own self-contained section with constants, types, and config factory
- Extract generic `alg_config()` helper for RPO/Poseidon2/RPX configs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - [BREAKING] `Variant::new` now expects an optional payload type to be specified ([#2848](https://github.com/0xMiden/miden-vm/pull/2848))
 - [BREAKING] Enum types are now exported from libraries as a `midenc_hir_type::EnumType`, rather than the type of the discriminant. ([#2848](https://github.com/0xMiden/miden-vm/pull/2848))
 - In `ExecutionTracer`, we no longer record node flags in `CoreTraceFragmentContext` when entering a node (they are redundant) ([#2866](https://github.com/0xMiden/miden-vm/pull/2866))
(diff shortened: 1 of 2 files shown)


huitseeker (approved):
LGTM. 🙏

Al-Kindi-0:
Would love some more context about the solution and the motivation for this change ?

adr1anh:
There are two versions of Keccak in Plonky3. We were previously using the one which was delegating to one of the dependencies, but this replaces it with a custom hardware accelerated Keccak-f permutation. Using `Packed{Field, Value} = [{F;u64}; KECCAK_LEN]` (where for instance on NEON `KECCAK_LEN=2`, the LMCS can absorb multiple rows in parallel. 

The LMCS is optimized for sponge-based hashes, which is what enables the virtual padding. For non-sponge-based hashes, we use a ChainingStatefulHasher (as we were doing with Keccak) which wraps a CryptographicPermuatation which implements `sponge.absorb(state, rows)` as `hasher.hash(prev_output || rows)`.

For motivation though, I can include a benchmark if necessary, but from previous experiments it gave close to a `KECCAK_LEN`x improvement.

Al-Kindi-0:
Thanks @adr1anh , makes sense now!

Al-Kindi-0 (approved):
LGTM!

PR #2877, opened 2026-03-19
feat: implement project assembly

### LINKED ISSUE #2879: Correctness bugs in miden-project
Verified on `next` at `bd42aeff82da`. Reporting those issues that came up in #2510 for convenience of addressing in #2877.

bitwalker:
This PR is the follow-up to #2510 which actually makes the project file format useful, by adding support to the assembler for assembling a project package from its `miden-project.toml` manifest, to a Miden package (i.e. a `.masp` file).

This PR is broken up into a few commits with relevant changes, but the bulk of the changes are in the first two commits which implement the package registry infrastructure, and the initial project assembly support.

An overview of what's been implemented/changed here:

* A new crate, `miden-package-registry`, whose primary exports are the `PackageRegistry` and `PackageProvider` traits, and all of the aspects relevant to package versioning and selection.
* ~~A new crate, `miden-package-registry-local`, which is comprised of a library component and a new executable called `miden-registry`, which provides local filesystem-based package management for Miden packages, and implements the `PackageRegistry` and `PackageProvider` traits. This integrates with `midenup` to provide package management locally.~~ Split out into #2881 
* A new `ProjectDependencyGraph` type in `miden-project` which handles all of the gritty details around dependency management, i.e. constructing a dependency graph, and resolving references to path/git dependencies locally. This dependency graph is then used by the assembler when actually assembling a project.
* Extensions to the `Assembler` interface for assembling projects via `ProjectAssembler`, a type which provides the extra context needed when assembling projects. The project-related functionality is gated behind the `std` feature, as projects are largely meaningless without filesystem support.
* A new `Assembler::link_package` method which links against a `miden_mast_package::Package` with a provided linkage, this can be used separately from the project assembly features.
* Addition of a `miden-project.toml` to the core library, and modification of the build script to produce both `.masp` and `.masl` outputs from the assembled project.
* An integration test fixture which assembles an executable project with an associated kernel and a dependency on the Miden core library.

~~The only breaking changes are in `miden_mast_package`, namely changes to the `Package` type, removal of the `MastArtifact` and `PackageKind` types (in favor of `miden_mast_package::TargetType`, which is now shared with `miden_project`). The package version was bumped to `4.0.0` as a result of these changes.~~ This was split out into #2886 on which this PR is now based

I'm still doing additional review and testing, but this is basically ready to merge - any additional changes could be made in follow-on PRs.

---
Fixes #2879, #2882, #2883, #2884, #2885

### DIFF
--- a/crates/project/src/lib.rs
+++ b/crates/project/src/lib.rs
@@ -117,13 +117,13 @@ impl Project {
         source_manager: &dyn miden_assembly_syntax::debuginfo::SourceManager,
     ) -> Result<Self, Report> {
         let path = path.as_ref();
-        let (manifest_path, check_first) = if path.is_dir() {
-            (path.join("miden-project.toml").canonicalize().map_err(Report::msg)?, true)
+        let manifest_path = if path.is_dir() {
+            path.join("miden-project.toml").canonicalize().map_err(Report::msg)?
         } else {
-            (path.canonicalize().map_err(Report::msg)?, false)
+            path.canonicalize().map_err(Report::msg)?
         };
 
-        Self::try_load_as_workspace_member(None, &manifest_path, check_first, source_manager)
+        Self::try_load_as_workspace_member(None, &manifest_path, source_manager)
     }
 
     /// Load a project manifest from `path`, expected to be named `name`
@@ -136,26 +136,23 @@ impl Project {
         source_manager: &dyn miden_assembly_syntax::debuginfo::SourceManager,
     ) -> Result<Self, Report> {
         let path = path.as_ref();
-        let (manifest_path, check_first) = if path.is_dir() {
-            (path.join("miden-project.toml").canonicalize().map_err(Report::msg)?, true)
+        let manifest_path = if path.is_dir() {
+            path.join("miden-project.toml").canonicalize().map_err(Report::msg)?
         } else {
-            (path.canonicalize().map_err(Report::msg)?, false)
+            path.canonicalize().map_err(Report::msg)?
         };
 
-        Self::try_load_as_workspace_member(Some(name), &manifest_path, check_first, source_manager)
+        Self::try_load_as_workspace_member(Some(name), &manifest_path, source_manager)
     }
 
     fn try_load_as_workspace_member(
         name: Option<&str>,
         manifest_path: impl AsRef<std::path::Path>,
-        include_initial_path: bool,
         source_manager: &dyn miden_assembly_syntax::debuginfo::SourceManager,
     ) -> Result<Self, Report> {
         use miden_assembly_syntax::debuginfo::SourceManagerExt;
 
         let manifest_path = manifest_path.as_ref();
-
-        let ignore_first_ancestor = !include_initial_path;
         let ancestors = manifest_path
             .parent()
             .ok_or_else(|| {
@@ -164,8 +161,7 @@ impl Project {
                     manifest_path.display()
                 ))
             })?
-            .ancestors()
-            .skip(ignore_first_ancestor as usize);
+            .ancestors();
 
         let initial_package_dir = manifest_path.parent();
         for ancestor in ancestors {
--- a/crates/project/src/tests.rs
+++ b/crates/project/src/tests.rs
@@ -248,6 +248,48 @@ path = "lib.masm"
 }
 
 #[test]
+fn load_project_reference_resolves_workspace_manifest_file_inputs() -> Result<(), Report> {
+    let tempdir = TempDir::new().unwrap();
+    let root = tempdir.path().join("workspace");
+    let dep_dir = root.join("dep");
+    fs::create_dir_all(&dep_dir).unwrap();
+
+    let workspace_manifest = root.join("miden-project.toml");
+    fs::write(
+        &workspace_manifest,
+        r#"[workspace]
+members = ["dep"]
+"#,
+    )
+    .unwrap();
+
+    let dep_manifest = dep_dir.join("miden-project.toml");
+    fs::write(
+        &dep_manifest,
+        r#"[package]
+name = "dep"
+version = "1.2.3"
+
+[lib]
+path = "lib.masm"
+"#,
+    )
+    .unwrap();
+    fs::write(dep_dir.join("lib.masm"), "export.foo\nend\n").unwrap();
+    let dep_manifest = dep_manifest.canonicalize().unwrap();
+
+    let context = TestContext::default();
+    let project =
+        Project::load_project_reference("dep", &workspace_manifest, &context.source_manager)?;
+
+    assert!(project.is_workspace_member());
+    assert_eq!(project.manifest_path(), Some(dep_manifest.as_path()));
+    assert_eq!(format!("{}", project.package().version()), "1.2.3");
+
+    Ok(())
+}
+
+#[test]
 fn workspace_rejects_duplicate_member_package_names() {
     let tempdir = TempDir::new().unwrap();
     let root = tempdir.path().join("workspace");


bitwalker:
@bobbinth This is opened against `main` since I expect to address the leftover issues from #2510 via this PR, and actually be able to use the project functionality as part of the 0.22.x release. I only mention this, because it has been moved to the v0.23.0 milestone, but we will not want to wait until 0.23 to release this functionality, as quite a bit of other tooling we're working on depend on this.

bitwalker:
For reviewers: the primary goal here should be to identify missing test cases, and underspecified semantics, so that we can improve coverage and tighten up the semantics of the project functionality. I would _NOT_ focus on the implementation details for now, unless you have major concerns about something in particular - this doesn't need to be perfect, because we already anticipate that this aspect of the toolchain will need to be tested against real world projects and use cases, and iterated upon based on our findings.

Things that deserve extra attention:

* Ambiguity about the project file format and related semantics
* Missing/poor documentation on projects
* Missing tests for important use cases
* Missing tests to cover edge cases that might come up in practice
* Bugs in version selection/dependency resolution. This has a decent amount of test coverage now, but @huitseeker has been doing a good job of identifying edge cases in previous iterations
* Bugs in how packages are linked, i.e. we expect a package to be linked statically, but it's linked dynamically - or vice versa.
* Mismatches between the behavior we want, and what is currently implemented

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,2002 @@
...
+                resolved_revision: String::read_from(source)?,
+            }),
+            invalid => Err(miden_core::serde::DeserializationError::InvalidValue(format!(
+                "invalid project source provenance tag '{invalid}'"
+            ))),
+        }
+    }
+}
+
+fn load_project_package(
```
huitseeker: A source dependency that is really a workspace member gets reloaded here as a standalone package, so inherited workspace settings are lost (and assembly fails).

crates/assembly/src/linker/mod.rs
```diff
@@ -358,6 +359,32 @@ impl Linker {
...
     }
 
+    /// Add a kernel to the linker after the linker is initially constructed.
+    ///
+    /// This cannot cause any issues with modules already added to the linker (if any), as they
+    /// cannot have directly depended on the kernel, or an error would have been raised.
+    ///
+    /// This will panic if the kernel is empty, or the provided kernel module inf o is not valid for
+    /// a kernel.
+    pub(super) fn link_with_kernel(
```
huitseeker, line 369: I think this assert is flipped, and will panic on being passed any real kernel.

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +66,114 @@ pub struct Package {
...
+            )))
+        } else {
+            Ok(self.unwrap_program())
         }
     }
 
-    /// Creates a new package with [Program] from this [Library] package and the given
-    /// entrypoint (should be a procedure in the library).
+    #[doc(hidden)]
+    pub fn unwrap_program(&self) -> miden_core::program::Program {
```
huitseeker, line 199: Converting an executable `.masp` package back into a `Program` here drops the kernel, so a kernel-backed package turns into a program without its syscall support.

crates/project/src/dependencies.rs
```diff
@@ -52,15 +50,16 @@ impl Dependency {
...
+        let req = match &self.version {
+            DependencyVersionScheme::Registry(version) => return version.clone(),
             DependencyVersionScheme::Workspace { .. } => None,
             DependencyVersionScheme::Path { version, .. } => version.clone(),
             DependencyVersionScheme::Git { version, .. } => {
                 version.as_ref().map(|spanned| VersionRequirement::Semantic(spanned.clone()))
             },
-        }
+        };
+        req.unwrap_or_else(|| VersionRequirement::from(VersionReq::STAR.clone()))
```
huitseeker, line 61: This is the workspace widening bug inherited from #2510 (and mentioned in #2883):`workspace = true` still widens to `*` here, and resolution can still pick a higher registry version instead of the workspace member version.

bitwalker: I'm working on confirming this, but if you have a reproducer it would help. There are existing tests both in `miden-project` and `miden-assembly` covering workspaces (in `project/src/dependency/graph.rs` and `assembly/src/project.rs`; as well as the integration test which assembles the `protocol` fixture), but I'm looking closer to ensure they actually would catch this issue or not

crates/mast-package/src/package/mod.rs
bitwalker: Indeed, I've got a regression test for this, I'll follow up once it is addressed

crates/assembly/src/linker/mod.rs
bitwalker: Yep, not sure how I missed that, it's supposed to be `self.kernel`

crates/assembly/src/project.rs
bitwalker: Indeed, I've added a regression test for this, and will follow up once it's fixed

crates/project/src/dependencies.rs
bitwalker: So the only code path which could exhibit the described behavior was dead code that I had intended to remove (i.e. in `crates/project/src/dependencies/resolver.rs` - but it is not otherwise reproducible in the dependency graph or project assembler (the existing tests confirm this). I've removed that code, so I believe this should no longer be an issue. If you can reproduce with the existing APIs somehow though, let me know.

huitseeker: Shared what I used at the time [in the other thread](https://github.com/0xMiden/miden-vm/issues/2879#issuecomment-4096914933).

bitwalker: Ah, yeah that uses the now-removed methods that were defined on `Package` to register with a package registry.

crates/assembly/src/project.rs
bitwalker: This should now be fixed

crates/mast-package/src/package/mod.rs
bitwalker: This has been addressed by way of a couple changes:

1. Executable packages are now dynamically linked against the kernel they depend on
2. The kernel package is appended to the executable package in a custom section. We use this to recover the kernel when converting an executable package back into a `Program`. If the section is not present, it is presumed that the caller will handle providing the kernel context manually.
3. The kernel context of a package is recorded in the dependencies of the package manifest, and we validate the kernel section based on the manifest (i.e. they must agree).

There is now a test that covers this, sans executing the program. I think the best approach here is to modify the CLI fixtures to use Miden projects, since that will cover both the existing behavior, plus ensuring that an assembled executable package with a kernel is actually executable.

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +66,209 @@ pub struct Package {
...
+                    entrypoint,
+                    kernel_library.kernel().clone(),
+                )),
+                None => Ok(Program::new(mast_forest, entrypoint)),
+            }
+        }
+    }
+
+    #[doc(hidden)]
+    pub fn unwrap_program(&self) -> miden_core::program::Program {
```
huitseeker, line 199: As mentioned in #2866, this still looks panic-prone on malformed executable packages.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,2304 @@
...
+        })?;
+        let expected = match origin {
+            ProjectSourceOrigin::Git { repo, resolved_revision, .. } => {
+                PackageBuildProvenance::Git {
+                    repo: repo.to_string(),
+                    resolved_revision: resolved_revision.to_string(),
+                }
+            },
+            ProjectSourceOrigin::Path | ProjectSourceOrigin::Root => PackageBuildProvenance::Path {
+                source_hash: self.compute_path_source_hash(
```
huitseeker: If a source package `dep` depends on a floating path dep `leaf`, the first build registers a canonical `dep@1.0.0#...`. After changing `leaf` without a semver bump, rebuilding the root still reuses that old `dep` package, because this provenance only hashes `dep`'s own sources and manifest, not the resolved dependency closure.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,2304 @@
...
+        {
+            assert_eq!(package.kind, target.ty);
+            return Ok(Arc::clone(package));
+        }
+
+        let profile = resolve_profile(project.as_ref(), profile_name)?;
+        let mut assembler = self
+            .assembler
+            .clone()
+            .with_emit_debug_info(profile.should_emit_debug_info())
```
huitseeker, line 189: build a source dependency under `dev` first so its canonical package keeps debug info, then rebuild the root under `release`: the release build still loads the dev-built dependency, because the reuse provenance does not record the effective profile or the debug/trim settings.

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +66,209 @@ pub struct Package {
...
-            MastArtifact::Executable(ref prog) => Arc::clone(prog),
-            _ => panic!("expected package to contain a program, but got a library"),
+    fn validate_embedded_kernel_dependency(&self, kernel_package: &Self) -> Result<(), Report> {
+        if !kernel_package.is_kernel() {
+            return Err(Report::msg(format!(
+                "package '{}' embeds '{}', but its kind is '{}'",
+                self.name, kernel_package.name, kernel_package.kind
+            )));
+        }
+
```
huitseeker, line 255: If the `KERNEL` section is missing, this branch quietly builds `Program::new(...)` instead of rejecting the package, so a malformed, stripped, or transitively broken executable package can still claim a kernel dependency in its manifest and be accepted with different runtime semantics.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,2304 @@
...
+            // NOTE: If there are conflicting runtime dependencies on the same package, an error
+            // will be raised. In the future, we may wish to relax this restriction, since such
+            // dependencies are technically satisfiable.
+            if matches!(edge.linkage, Linkage::Dynamic) && !dependency_package.is_kernel() {
+                merge_runtime_dependency(
+                    &mut runtime_dependencies,
+                    package_dependency(dependency_package.as_ref()),
+                )?;
+            }
+            for dependency in dependency_package.manifest.dependencies() {
```
huitseeker: I think this loses the kernel package when the dependency is only propagated transitively.

huitseeker (changes requested):
Review at 913a31d8aca4813465607e82596e4dd5b6ea9bab

crates/mast-package/src/package/mod.rs
bitwalker: By definition it should panic - that's why it's called `unwrap`

crates/mast-package/src/package/mod.rs
bitwalker: That's expected - if the kernel is not attached to the package, it must be provided externally, or it will fail at runtime.

crates/mast-package/src/package/mod.rs
huitseeker: Here's the panic on 913a31d8a: https://gist.github.com/huitseeker/eec0b7b58d041998f4f17f4006fc5312

bitwalker: Ah, right, I see what you mean. Thanks for the repro

crates/mast-package/src/package/mod.rs
huitseeker: Right, though we now fail at “kernel resolution/load time” (better) or at runtime if the caller forgets to attach one and calls a syscall (worse).

bitwalker: Yeah this is an area that we will likely want to tighten up in the future, possibly as part of removing `Program`. In general we currently (so far as I know) just assume that referenced procedure roots will be available at runtime when dynamically linking dependencies, and error out when that isn't the case.

One of my hopes for the package format (especially based on projects with dependency information), was that we could actually have the VM (or whatever is hosting the VM), use the package metadata to validate that everything needed is available before it begins execution. One complication to that though, is that one of the design properties of MAST that @bobbinth expressed to me early on, was the ability to withhold parts of a program that are never going to be executed, and so in those cases, you might have dependencies which aren't needed for a particular execution, but are needed by the package in general - you wouldn't want to raise an error prematurely though. I suspect it might actually be the case that we wouldn't even _have_ a package in those cases, just a partial MAST forest, but I'm not actually sure.

Regardless, since MASM currently supports these sorts of uses (i.e. punting to runtime a dependency on some other package/library/MAST), it's probably not the right context in which to be strict about it, until we have tighter specification around how to communicate to the runtime what is expected by a given artifact that is being executed (and clarification on what component is responsible for that).

crates/mast-package/src/package/mod.rs
bitwalker: This has been addressed in #2886, we now check that the library has the export and that the MAST forest can resolve that to a node, and raise an error if not.

crates/assembly/src/project.rs
bitwalker: This has been fixed and a regression test added

crates/assembly/src/project.rs
bitwalker: Build profiles are now part of the provenance of a package, and a regression test added to cover this

crates/assembly/src/project.rs
bitwalker: Confirmed and fixed, with a regression test to cover it

crates/package-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+            Some(_) => None,
+            None => Some(record),
+        }
+    }
+
+    /// Return the canonical metadata for `package` at the given semantic version.
+    fn get_by_semver(&self, package: &PackageId, version: &SemVer) -> Option<&PackageRecord> {
+        self.available_versions(package)
+            .and_then(|versions| {
+                versions.iter().find(|(candidate, _)| &candidate.version == version)
```
huitseeker: I think the default `get_by_semver` can make canonical selection arbitrary here. 

When multiple artifacts share the same semantic version, this picks the first `BTreeMap` entry, which is really just whichever digest sorts first, not a registry-defined canonical choice. That means semver-only resolution can depend on map ordering instead of policy. 

I'm not sure what the best fix is: should this default either reject ambiguity or require registries with canonical semantics to override it?

crates/package-registry/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /// Returns true if the canonical semantic version of `package` exists in the registry.
+    fn is_semver_available(&self, package: &PackageId, version: &SemVer) -> bool {
+        self.get_by_semver(package, version).is_some()
+    }
+
+    /// Return the metadata for `package` at `version`, if present.
+    fn get_by_version(&self, package: &PackageId, version: &Version) -> Option<&PackageRecord> {
+        let record = self.get_by_semver(package, &version.version)?;
```
huitseeker: I think exact lookups break when a package has more than one artifact for the same semver. `get_by_version` first calls `get_by_semver`, so this only checks the first `1.0.0` record and returns `None` if that digest does not match, even when the exact `1.0.0#...` entry exists later in the map. That also flows into `get_exact_version`, `find_latest(Exact(..))`, dependency validation, and kernel recovery. Would it make sense to look up the full `Version` key directly here instead?

crates/project/src/dependencies/graph.rs
```diff
@@ -0,0 +1,1599 @@
...
+            self.git_cache_root.join(format!("0x{}", DisplayHex::new(&key.as_bytes())));
+        if !checkout_path.exists() {
+            let mut args = vec!["clone"];
+            let mut rev_buffer = String::new();
+            match revision {
+                GitRevision::Branch(name) => {
+                    args.extend_from_slice(&["--branch", name.as_ref()]);
+                },
+                GitRevision::Commit(rev) => {
+                    rev_buffer = format!("--revision={rev}");
```
huitseeker: Nit: this is a pretty recent functionality, from Git 2.49 (March 2025 — so it won't be available on the garden-variety Mac), should we probe this flag is available on the local git version before using it, and emit an error message about supported versions if it isn't?

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,3486 @@
...
+    let bytes = fs::read(path)
+        .map_err(|error| Report::msg(format!("failed to read '{}': {error}", path.display())))?;
+    let package = MastPackage::read_from_bytes(&bytes).map_err(|error| {
+        Report::msg(format!("failed to decode package '{}': {error}", path.display()))
+    })?;
+    Ok(Arc::new(package))
+}
+
+#[cfg(test)]
+mod tests {
```
huitseeker: The size of this file suggests splitting the tests in a different file.

crates/package-registry/src/lib.rs
bitwalker: The registry implementations we have currently reject duplicate semantic versions for a package (there can be only one canonical version for a given semantic version). It allows multiple semantic versions to have the same digest, but not multiple digests with the same semantic version). This is handled when publishing packages though, and you are correct that the API here doesn't enforce that behavior.

I think the best choice would be to document that registry implementations must uphold that canonical rule. Enforcing it in the type system would be tricky though (or at least require another layer of indirection, i.e. have two separate `available_versions` methods, one for querying available digests for a given package, and one for querying available semantic versions for a package). The caller must then choose appropriately.

crates/package-registry/src/lib.rs
bitwalker: I think this is basically an extension of the point you raised in your comment on `get_by_semver`. Since the publish step rejects multiple packages for a given semantic version, these checks as written are safe - but only if the documented semantics of the registry dictate that behavior. I'll mull over whether there is a stronger way to enforce those semantics in the API.

crates/project/src/dependencies/graph.rs
bitwalker: Nice catch, I completely spaced how recent this flag is. I think for now it makes sense to just avoid the implicit optimization here and just use the slower method of cloning the full repo and then checking out the specific revision, with a note to maybe switch to `--revision` at some point in the future - I'd rather not probe for this, since it isn't really worth the complexity IMO.

crates/package-registry/src/lib.rs
bitwalker: Ok, this has been resolved by making the contract around registries tighter, requiring a single canonical package artifact per semantic version. The change is in a new commit on top of this branch

crates/package-registry/src/lib.rs
bitwalker: This was resolved as described [here](https://github.com/0xMiden/miden-vm/pull/2877#discussion_r2977019997)

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,1517 @@
...
+
+    fn load_canonical_package(
+        &self,
+        package_id: &PackageId,
+        version: &miden_project::SemVer,
+    ) -> Result<Option<Arc<MastPackage>>, Report> {
+        let Some(record) = self.store.get_by_semver(package_id, version) else {
+            return Ok(None);
+        };
+        self.store.load_package(package_id, record.version()).map(Some)
```
huitseeker, line 480: Is loading failure something we want to defend against?

Could this end up hard failing when the registry index still has `dep@1.0.0` but the artifact bytes are gone? `get_by_semver` says the package exists, then `load_package` errors here, so a path or git source dep never gets a chance to rebuild from source.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,1517 @@
...
+        }
+
+        let Some(kernel_dependency) = kernel_runtime_dependency(package.as_ref())? else {
+            return Ok(None);
+        };
+
+        let version =
+            PackageVersion::new(kernel_dependency.version.clone(), kernel_dependency.digest);
+        if self.store.get_exact_version(&kernel_dependency.name, &version).is_some() {
+            let kernel_package = self.store.load_package(&kernel_dependency.name, &version)?;
```
huitseeker: This looks like the same stale-store case for embedded kernels. If the exact kernel version is indexed but `load_package` cannot read the artifact, we return the load error here and never try `try_embedded_kernel_package()`.

crates/project/src/dependencies.rs
```diff
@@ -237,9 +256,23 @@ impl DependencyVersionScheme {
...
+                            || uri.path() == format!("{member_path}/miden-project.toml")
+                            || absolutize_path(Path::new(member_path), workspace_root)
+                                .ok()
+                                .is_some_and(|member_dir| {
+                                    resolved_uri == member_dir
+                                        || resolved_uri == member_dir.join("miden-project.toml")
+                                })
+                    });
+                    if is_member {
+                        Ok(Self::Workspace { member: uri.clone() })
```
huitseeker: Could we keep the inherited version requirement here instead of dropping it when a workspace path points at a member? 

In a local repro, `[workspace.dependencies] dep = { path = "dep", version = "=0.1.0" }` plus `dep.workspace = true` still assembles when the member package is actually `0.2.0`, because this becomes `Workspace { member }` and then `required_version()` falls back to `*`. 

This seems to be the bug pattern [I was mentioning earlier](https://github.com/0xMiden/miden-vm/pull/2877#pullrequestreview-3980544469).

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,1517 @@
...
+            },
+            ProjectDependencyNodeProvenance::Registry { selected, .. } => {
+                let package = self.store.load_package(package_id, selected)?;
+                ResolvedPackage {
+                    linked_kernel_package: self.resolve_linked_kernel_package(package.clone())?,
+                    package,
+                }
+            },
+            ProjectDependencyNodeProvenance::Preassembled { path, .. } => {
+                let package = load_package_from_path(path)?;
```
huitseeker: Could we revalidate the selected preassembled artifact here instead of rereading whatever happens to be at `path`? 

The graph already fixed this dependency to a specific `selected` version/digest, but this branch ignores that and loads the current file contents.

crates/project/src/dependencies.rs
bitwalker: I think we could raise an error if this occurs, yes, i.e. we shouldn't allow disagreement of that kind

crates/assembly/src/project.rs
bitwalker: This is now handled explicitly, ensuring we fallback to building from source if the artifact is unreadable

crates/assembly/src/project.rs
bitwalker: This has also been resolved similarly to the rebuild-from-source path when the artifact is unreadable

crates/project/src/dependencies.rs
bitwalker: Done as described - we now raise an error if the version requirement does not agree with the sources

crates/assembly/src/project.rs
bitwalker: Done

crates/project/src/dependencies/graph.rs
```diff
@@ -0,0 +1,1677 @@
...
+    fn resolve_dependency(
+        &self,
+        dependency: &Dependency,
+        parent: &LoadedSourcePackage,
+    ) -> Result<ResolvedDependencyNode, Report> {
+        match dependency.scheme() {
+            DependencyVersionScheme::Registry(requirement) => {
+                let package_id = PackageId::from(dependency.name().clone());
+                let record =
+                    self.registry.find_latest(&package_id, requirement).ok_or_else(|| {
```
huitseeker: I think this might be a followup of [this prior comment](https://github.com/0xmiden/miden-vm/pull/2877#discussion_r2971523254). This resolves registry deps greedily with `find_latest` one edge at a time, even though the PR adds a PubGrub resolver. Two source deps can lock different versions of the same registry package and then trip `merge_runtime_dependency`, even when a shared version exists.

crates/package-registry/README.md
bobbinth, line 1: Let's add a "LICENSE" section to this readme (as we do for other crates).

crates/lib/core/asm/miden-project.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "miden-core"
+version = "0.22.0"
+
+[lib]
+namespace = "miden::core"
+path = "mod.masm"
+
+[profile.release]
+debug = true
+trim_paths = true
```
bobbinth, line 13: May be good to add some comments here explaining:

- Why we enable `debug` for a release build (e.g., we always want the debug info present in the package, and we can easily strip it off when running the VM in release mode).
- What does `trim_paths` do.

crates/assembly/src/lib.rs
```diff
@@ -16,12 +16,14 @@ mod instruction;
...
+#[cfg(feature = "std")]
+mod project;
 
 #[cfg(test)]
 mod mast_forest_merger_tests;
 #[cfg(any(test, feature = "testing"))]
 pub mod testing;
-#[cfg(test)]
+#[cfg(all(test, feature = "std"))]
 mod tests;
```
bobbinth, line 27: This means we now run the underlying test only in `std` - right? What's driving this?

We do use assembler in `no_std` context quite a lot (e.g., for the playground) - so, testing its functionality in `no_std` context is probably important.

crates/assembly/src/linker/mod.rs
```diff
@@ -358,6 +359,33 @@ impl Linker {
         graph
     }
 
+    /// Add a kernel to the linker after the linker is initially constructed.
+    ///
+    /// This cannot cause any issues with modules already added to the linker (if any), as they
+    /// cannot have directly depended on the kernel, or an error would have been raised.
+    ///
+    /// This will panic if the kernel is empty, or the provided kernel module inf o is not valid for
```
bobbinth: typo: "inf o" -> "info".

crates/assembly/src/linker/library.rs
```diff
@@ -1,55 +1,55 @@
...
 /// Represents an assembled module or modules to use when resolving references while linking,
 /// as well as the method by which referenced symbols will be linked into the assembled MAST.
 #[derive(Clone)]
 pub struct LinkLibrary {
-    /// The library to link
-    pub library: Arc<Library>,
+    /// The MAST of this library
+    pub mast: Arc<MastForest>,
+    /// Metadata about the modules and symbols available in the linked forest
+    pub module_infos: Vec<ModuleInfo>,
```
bobbinth, line 17: Not related to this PR (and not something we should change now), but I wonder if we should rename `ModuleInfo` into something like `ModuleMetadata` ("module info" always sounded a bit too vague to me).

crates/assembly/src/linker/library.rs
```diff
@@ -1,55 +1,55 @@
...
-    pub kind: LinkLibraryKind,
+    pub linkage: Linkage,
 }
 
 impl LinkLibrary {
-    /// Dynamically link against `library`
-    pub fn dynamic(library: Arc<Library>) -> Self {
-        Self { library, kind: LinkLibraryKind::Dynamic }
+    /// Construct a [LinkLibrary] from a [miden_mast_package::Package]
+    pub fn from_package(package: Arc<miden_mast_package::Package>) -> Self {
```
bobbinth: nit: i'd convert the absolute import paths to just regular imports at the top of the file.

crates/assembly/src/linker/library.rs
```diff
@@ -1,55 +1,55 @@
...
+            mast,
+            module_infos,
+            linkage: Linkage::Dynamic,
+        }
     }
 
-    /// Statically link `library`
-    pub fn r#static(library: Arc<Library>) -> Self {
-        Self { library, kind: LinkLibraryKind::Static }
+    pub(crate) fn from_library(library: &miden_assembly_syntax::Library) -> Self {
```
bobbinth: Same nit as above.

crates/assembly/src/linker/library.rs
```diff
@@ -1,55 +1,55 @@
...
-    /// required to be provided at runtime, as is the case with dynamically-linked libraries.
-    Static,
+    /// Modify the linkage of this library
+    pub fn with_linkage(mut self, linkage: Linkage) -> Self {
+        self.linkage = linkage;
+        self
+    }
 }
+
+pub use miden_project::Linkage as LinkLibraryKind;
```
bobbinth: nit: let's move this to the top of the file.

crates/package-registry/src/version.rs
```diff
@@ -104,29 +152,13 @@ impl Borrow<SemVer> for Version {
 impl fmt::Display for Version {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if let Some(digest) = self.digest.as_ref() {
-            write!(f, "{}@{}", &self.version, digest.inner())
+            write!(f, "{}#{}", &self.version, digest.inner())
```
bobbinth, line 155: Question: why did we switch from `@` to `#`?

crates/lib/core/build.rs
```diff
@@ -220,17 +220,35 @@ fn main() -> Result<(), Report> {
...
+    // write the masp output
+    package
+        .write_to_file(
+            build_dir
+                .join(ASL_DIR_PATH)
+                .join("core")
+                .with_extension(miden_assembly::package::Package::EXTENSION),
+        )
+        .map_err(|err| io::Error::other(err.to_string()))
+        .into_diagnostic()?;
```
bobbinth: It may be good to encapsulate this into a method of the `Project` struct so that we could do something like:
```rust
package.write_masp_file(build_dir.join(ASL_DIR_PATH).into_diagnostic()?;
```
This would:
- Infer the name of the file from package name - e.g., `miden-core`.
- Use the default package extension - i.e., `.masp`.

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
     },
 }
 
+pub(crate) struct AssemblyProduct {
```
bobbinth: To keep this file from growing much more - let's move this struct (and maybe other related structs) into a separate file.

In general, we should split this file up to make it easier to navigate - but that's for another PR.

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
...
+    // This is unused when the `std` feature is not present
+    #[allow(unused)]
+    kind: TargetType,
```
bobbinth: Maybe I haven't looked far enough yet, but why isn't this relevant in `no_std` context? Does this mean we can't assemble packages unless we are in `std`?

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
...
+    artifact: Arc<Library>,
```
bobbinth: In the future, this will get replaced with a combination of `MastForest` and `Vec<ModuleInfo>` right?

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
...
+        Ok(if let Some(kernel) = self.kernel {
+            Program::with_kernel(self.artifact.mast_forest().clone(), entrypoint, kernel)
+        } else {
+            Program::new(self.artifact.mast_forest().clone(), entrypoint)
+        })
+    }
+
+    pub fn into_kernel_library(self) -> Result<KernelLibrary, Report> {
+        KernelLibrary::try_from(self.artifact).map_err(|error| Report::msg(error.to_string()))
+    }
```
bobbinth: I'm guessing this is a temporary function, right? If so, let's leave a `TODO` comment in here. If not, we should probably do extra checks - e.g., fail if `self.kind` isn't "kernel" or if `self.kernel` isn't `None`.

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
...
+    #[cfg(feature = "std")]
+    pub fn manifest(&self) -> &PackageManifest {
+        &self.manifest
+    }
+
+    pub fn into_artifact(self) -> Arc<Library> {
+        self.artifact
+    }
+
+    pub fn into_program(self) -> Result<Program, Report> {
```
bobbinth: Similar comment as above.

crates/assembly/src/assembler.rs
```diff
@@ -50,6 +53,56 @@ enum AssemblerError {
...
+
+impl AssemblyProduct {
+    fn new(
+        kind: TargetType,
+        artifact: Arc<Library>,
+        kernel: Option<Kernel>,
+        manifest: PackageManifest,
+    ) -> Self {
+        Self { kind, artifact, kernel, manifest }
+    }
```
bobbinth: I'd add assertions here to make sure we don't get an invalid combination - e.g., if `kind = Kernel`, then `kernel` should be `None`.

crates/assembly/src/assembler.rs
```diff
@@ -103,19 +156,38 @@ pub struct Assembler {
...
-    linker: Linker,
+    linker: Box<Linker>,
+    /// The debug information gathered during assembly
+    debug_info: DebugInfoSections,
     /// Whether to treat warning diagnostics as errors
     warnings_as_errors: bool,
+    /// Whether to preserve debug information in the assembled artifact.
+    emit_debug_info: bool,
+    /// Whether to trim source file paths in debug information.
+    trim_paths: bool,
```
bobbinth: We recently removed similar flags from the assembler to simplify things. Specifically, the intent was:

The assembler always assembles programs/libraries/packages with debug info. If the user doesn't need them, they can strip them out from the assembled artifacts.

Is there a reason to change this methodology?

crates/assembly/src/assembler.rs
```diff
@@ -326,6 +408,59 @@ impl Assembler {
...
         Ok(self)
     }
+
+    /// Link against `package` with the specified linkage mode during assembly.
+    pub fn link_package(
+        &mut self,
+        package: Arc<miden_mast_package::Package>,
+        linkage: Linkage,
+    ) -> Result<(), Report> {
+        use miden_mast_package::PackageExport;
```
bobbinth: nit: is there a reason to use an absolute import paths here?

crates/assembly/src/assembler.rs
```diff
@@ -326,6 +408,59 @@ impl Assembler {
...
+                        }
+                    })
+                    .collect::<Vec<_>>();
+                let Some(kernel_module) =
+                    package.mast.module_infos().find(|mi| mi.path().is_kernel_path())
+                else {
+                    return Err(Report::msg(
+                        "invalid kernel package: does not contain kernel module",
+                    ));
+                };
```
bobbinth: We may be able to simplify this a bit:

We could probably add something like `Package::get_kernel_module_info()` that could look something like:
```rust
impl Package {
    pub fn get_kernel_module_info(&self) -> Option<ModuleInfo> {
        ...
    }
}
```
And then, we should be able to get the digests of exported procedures from the module - e.g., as:
```rust
let proc_digests: Vec<Digest> = kernel_module.procedure_digests().collect();
```

crates/assembly/src/assembler.rs
```diff
@@ -1309,3 +1558,239 @@ pub(super) struct ResolvedProcedure {
     pub node: MastNodeId,
     pub signature: Option<Arc<FunctionType>>,
 }
+
+fn register_debug_type(
+    debug_types_section: &mut miden_mast_package::debug_info::DebugTypesSection,
+    declared_name: Option<Arc<str>>,
+    declared_ty: Option<&ast::TypeExpr>,
+    ty: &ast::types::Type,
+) -> Result<miden_mast_package::debug_info::DebugTypeIdx, Report> {
```
bobbinth: Could we move this function to a separate file? Or, could it be implemented as a method on `DebugTypesSection`? e.g., `DebugTypesSection::register_debug_type()`.

Also, would be great to add some doc comments to explaining what it does.

crates/assembly/src/assembler.rs
```diff
@@ -795,6 +1006,44 @@ impl Assembler {
...
+                            file_line_col.column,
+                        )
+                        .with_mast_root(procedure.mast_root());
+                        let func_info = if let Some(type_index) = type_index {
+                            func_info.with_type(type_index)
+                        } else {
+                            func_info
+                        };
+                        self.debug_info.debug_functions_section.add_function(func_info);
+                    }
```
bobbinth: nit: is it possible to encapsulate this in a helper function somehow?

Also, could we add imports at the top of the files so that we don't have to use absolute import paths here?

crates/assembly/src/lib.rs
bitwalker: I think it is going to be impractical to build the assembler without some libstd functionality going forward, which _is_ supported in Wasm (at least the parts that matter for assembly) - but we should have a discussion on what exactly the Wasm users today actually need from the assembler APIs.

Ultimately, the playground will need to support projects, so we'll need to ensure that it has what it needs for that - but my understanding is that the playground does not actually compile+assemble in Wasm (it certainly can't be compiling from Rust there, since the compiler itself doesn't support that configuration currently).

My suspicion is that the assembler itself doesn't actually need to support no-std environments, only the processor - but that is something I'd like more clarity on for sure.

crates/assembly/src/linker/library.rs
bitwalker: `ModuleManifest` or `ModuleDescriptor` might make more sense, as that's essentially what it is (a descriptor of what is in the module, without any of the actual code).

crates/package-registry/src/version.rs
bitwalker: Because we use `package@semver` for display of versioned package identifiers, and `package@semver#digest` as a refinement when the digest is available. That convention seems to be pretty common elsewhere, e.g. Wasm Component Model, Cargo - so it felt like a better choice of separator for this.

crates/assembly/src/assembler.rs
bitwalker: Assembly in general effectively requires `std` - we've just been pretending it doesn't. There is almost no meaningful way to do assembly of anything non-trivial without `std` (where non-trivial here means: has dependencies).

AIUI though, we don't actually have any reason to require no-std support for `miden-assembly` once packages exist (and even now, I believe all the "real" uses of it run in `std`-enabled builds). This is something we should definitely clarify though (i.e. what are the actual build targets we need to support the assembler on), because there isn't really a meaningful way to assemble things without `std` other than a very simple set of modules (punting dependency management up to some other tool, but then _that_ tool is going to require `std`, so we're not actually solving any problems along that path).

crates/assembly/src/assembler.rs
bitwalker: Yes, what lives in `Arc<Library>` now is basically superceded by `Arc<MastForest>` and `PackageManifest` (or its more "raw" equivalent, `Vec<ModuleInfo>`)

crates/assembly/src/assembler.rs
bitwalker: That's already handled by `KernelLibrary::try_from` I believe - but yes, this is basically temporary, at least until we remove `Library`/`KernelLibrary`.

crates/assembly/src/assembler.rs
bitwalker: These aren't externally-visible, user-provided flags - they are set via the build profile of a project. Further, this controls both the debug info in the MAST _and_ the debug sections that the assembler appends to the package - there's little reason to do the work of building them only to make the user strip them manually.

It made sense to simplify the interface back when we expected users to manually set these options via the `Assembler` API, but that's no longer needed.

crates/assembly/src/assembler.rs
bitwalker: To be explicit, as we have both `miden_project::Package` and `miden_mast_package::Package` - this just avoids the possibility of mixing them up/confusing them.

crates/assembly/src/lib.rs
bobbinth: So far, we've tried very hard to make sure that the assembler can run in WASM and currently to do this we use `no_std` as a proxy (i.e., if we don't enable `no_std`, a lot of other functionality gets enabled that can't run in WASM). Dropping support for this will likely cause many breaking changes downstream (e.g., I believe the client is also assembling some scripts "on the fly" etc.).

This, of course, can be changed - but that requires much more thought to ensure we have plans for how to handle it downstream. So, in this PR, we need to make sure we don't break this functionality - i.e., we can still use the assembler in `no_std` and, at the very least, it can still assemble libraries/programs using the old API in `no_std` context.

crates/assembly/src/assembler.rs
bobbinth: I think there are two separate questions which go beyond this PR:

- Do we want to make the assembler WASM-compatible? My current thinking is "yes" as some things downstream rely on it.
- Do we need `no_std` to support WASM? I think the answer here is "no" - but it'll take some work to get us to this point because, as I mentioned in the other comment, we currently conflate `no_std` and WASM - i.e., to compile for WASM we have to turn on `no_std`.

crates/assembly/src/assembler.rs
bobbinth: I'm not too concerned about the extra work here because it is likely to be negligible. Similarly, not very concerned about the package size. But I think it may be good to have a rule that "packages always have debug info" (unless someone went out of their way to strip it out).

My thinking here is that we can very easily control at execution time whether we want to run with or w/o debug info - so, there is very little downside to having debug info virtually always be included in packages.

crates/assembly/src/lib.rs
bitwalker: That is still the case in this PR - we aren't removing compatibility with `no_std` in `miden-assembly`; but the new project assembly functionality _does_ require it.

My point was that:

1. I don't believe we are _actually_ assembling in Wasm anymore. Correct me if I'm wrong, and point me to where we are still doing so. So we aren't actually breaking anyone in this case.
2. Rust can compile to Wasm with libstd, and so long as you don't _use_ anything that requires functionality not provided via the host, it works perfectly fine. Should we actually want to assemble in a Wasm-based application, the actual host functionality we need is minimal (and stuff that can be provided via WASI/shims).

You'll note that the Rust Playground does not support Cargo, and provides no way to define crates. You can compile a single file, and that's it. It can depend on a limited set of crates from crates.io because those crates are bundled into the browser, and it manually handles linking against them. That is essentially what we can provide today with our no-std assembler. Project assembly is equivalent to supporting Cargo in the browser - and that isn't supported today, and we can't support it either.

crates/assembly/src/assembler.rs
bitwalker: > Do we want to make the assembler WASM-compatible? My current thinking is "yes" as some things downstream rely on it.

I would push back on this, or rather, I want to know where we are currently assembling in Wasm, that will continue to need to do so.

The critical point though, is that we _can_ continue to support a limited form of assembly in Wasm, but we _cannot_ support project assembly in general - that would be akin to supporting Cargo in the browser.

> Do we need no_std to support WASM? 

Definitely not.

> I mentioned in the other comment, we currently conflate no_std and WASM

I don't think we really do in `miden-assembly`. Likely the one issue we'd run into is the transitive dependency on `miden-crypto` which compiled with `std` feature might rely on features that would fail to build - but I think we do actually gate things like that with `#[cfg(target_family = "wasm")]` currently.

In any case, I agree that removing `no-std` compatibility from the assembler is something that will take time - but there is no effort to do so in this PR. The tests that run in `std` here cover the same API surface as used in `no-std`, so we don't actually need to enable building the tests in no-std (and we don't run them there anyway).

crates/assembly/src/assembler.rs
bitwalker: I would imagine the major reason why someone would not want to build with debug info, is they don't want to have any of the original source code shipped with the package. By including debug info, you also include the source files, internal symbols, etc. That isn't always desirable.

The idea of having `miden-project.toml` provide this configuration is so that a user doesn't have to write some Rust tool just to strip that stuff from the packages they build.

There isn't any performance advantage of course, and the size difference is negligible - it's really just about the IP and privacy aspects IMO.

bobbinth: Agreed that stripping private info (e.g., local filesystem paths) is critical - but I'd say that including the source is probably a good idea in most cases (at least in our context).

I'm not against brining this functionality back into the assembler - but ideally, it should be a separate discussion and a separate PR.

crates/assembly/src/linker/library.rs
bobbinth: https://github.com/0xMiden/miden-vm/issues/2928

crates/assembly/src/assembler.rs
bobbinth: Created https://github.com/0xMiden/miden-vm/issues/2929 for this.

crates/assembly/src/linker/library.rs
```diff
@@ -1,55 +1,56 @@
...
+    pub fn from_package(package: Arc<Package>) -> Self {
+        let mast = package.mast.mast_forest().clone();
+        let module_infos = package
+            .mast
+            .module_infos()
+            .map(|mut mi| {
+                mi.set_version(package.version.clone());
+                mi
+            })
+            .collect();
```
bobbinth, line 33: I'm not sure I fully understand why we need to update the versions here. Could a version of a module and a version of a package diverge somehow? That is, under what circumstances would a version of a package be different than a version of a module within that package?

bobbinth: Also, AFACT, we are not actually ever using the version field. That is, we can set it here, but there is no way to read the version from `ModuleInfo` (and it doesn't seems to be used anywhere).

Maybe it makes sense to remove the `version` field from the `ModuleInfo` for now?

crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -0,0 +1,297 @@
...
+
+impl DebugInfoSections {
+    pub fn emit_procedure_debug_info(
+        &mut self,
+        source_manager: &dyn SourceManager,
+        span: SourceSpan,
+        path: &Path,
+        signature: Option<Arc<FunctionType>>,
+        procedure: &Procedure,
+    ) -> Result<(), Report> {
```
bobbinth: I moved this function here - but the naming should probably be updated. Maybe it should be `record_procedure_debug_info()`?

crates/assembly/src/assembler/product.rs
```diff
@@ -0,0 +1,58 @@
+use super::*;
+
+pub struct AssemblyProduct {
+    // This is unused when the `std` feature is not present
+    kind: TargetType,
```
bobbinth, line 4: Is the comment here still relevant?

crates/assembly/src/assembler/product.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    #[cfg(feature = "std")]
+    pub fn kind(&self) -> TargetType {
+        self.kind
+    }
+
+    #[cfg(feature = "std")]
+    pub fn manifest(&self) -> &PackageManifest {
+        &self.manifest
+    }
```
bobbinth: Is gating on `std` correct here? It seems like these are used in call contexts?

crates/assembly/src/linker/library.rs
bitwalker: 1. The version isn't known to the `MastForest`, so `version` is always `None` when asking the `MastForest` for its module information (it doesn't store `ModuleInfo`, it derives it on the fly). We're calling `set_version` here to embed that in the `ModuleInfos` we get from `MastForest`
2. The `version` field is going to be needed to disambiguate symbols from multiple versions of the same module that may be present in the linker simultaneously. This adds that metadata so we can start taking it into account, but does not yet do anything with it in the linker

crates/assembly/src/assembler/debuginfo.rs
bitwalker: We use `register_debug_type`, maybe `register_procedure_debug_info` to stay consistent?

crates/assembly/src/assembler/product.rs
bitwalker: Ah damn, I meant to remove that and somehow missed it

crates/assembly/src/assembler/product.rs
bitwalker: They are unused when building no-std, so we get unused symbol warnings. We could use `cfg_attr` though instead, with `expect`, e.g. `#[cfg_attr(not(feature = "std"), expect(unused))]`

crates/assembly/src/linker/library.rs
bobbinth: Would it make sense to introduce `VersionedModuleInfo` instead of adding this optional field? This way, we can enforce stronger type safety. But also, we can remove this from this PR and re-introduce at a point when we actually need it.

bitwalker: The problem is that we can have modules for which we have version information, and modules we don't - so it is truly optional, not something we can enforce via types.

We'll need to take this information into account almost immediately, so I think it makes sense to add the part where we provide the information when its available, before we implement changes to the linker to use it; but we also want to provide this information to downstream users (e.g. the compiler/debugger/LSP) which query a package for module information that they will then use.

We can remove it from this PR of course, but it is definitely something we need, not a "maybe we'll need it, so let's add it" sort of thing.

crates/assembly/src/project/build_provenance.rs
```diff
@@ -0,0 +1,243 @@
...
+};
+use miden_mast_package::{Package as MastPackage, Section, SectionId};
+
+use super::PackageBuildSettings;
+
+// PACKAGE BUILD PROVENANCE
+// ================================================================================================
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) enum PackageBuildProvenance {
```
bobbinth, line 40: Could we add some doc comments here explaining the purpose and uses of this struct?

bitwalker: Certainly, I'll do that here in a sec. In short though, it is used to record the build provenance of a package (including things like profile settings that change the way the package is built). This is then used to detect when a package has changed on disk relative to what has been registered with the package registry - we can reuse the already-assembled package from the registry if the provenance has not changed, rather than re-assembling it; otherwise, we must re-assemble from source. This behavior is specific to source packages (i.e. `path` or `git` dependency schemes).

It is also used when publishing a package to the registry - if that semantic version is already registered, and the build provenance has changed, we return an error, because it cannot be published to the registry without bumping the semantic version (because only a single canonical package per semantic version can exist in the registry). If the provenance is identical, then it's a no-op (i.e. there's nothing to do).

bitwalker: See the updated docs for more in-depth explanation

crates/mast-package/src/package/mod.rs
```diff
@@ -247,7 +279,7 @@ impl Package {
         Ok(())
     }
 
-    fn kernel_runtime_dependency(&self) -> Result<Option<&Dependency>, Report> {
+    pub fn kernel_runtime_dependency(&self) -> Result<Option<&Dependency>, Report> {
```
bobbinth, line 295: I made this function public and removed what seems to have been an exact copy of this function in the assembler crate (I'm assuming the duplication was not intentional).

Would be good to add brief doc comments here.

crates/assembly/src/project/package_ext.rs
```diff
@@ -0,0 +1,284 @@
...
+        target: &Target,
+        current_root: &FsPath,
+    ) -> Result<BTreeSet<PathBuf>, Report>;
+
+    fn resolve_target_source_paths(&self, target: &Target) -> Result<TargetSourcePaths, Report>;
+
+    fn effective_manifest_hash_input(&self) -> Result<String, Report>;
+
+    fn resolve_profile(&self, name: &str) -> Result<&Profile, Report>;
+}
```
bobbinth, line 57: @bitwalker I added this trait to encapsulate what I think is functionality related to packages, but didn't really put too much though into what really belongs here vs. what should be in the project crate.

Also, probably some of the naming should be updated.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,758 @@
...
+            && let Some(package) = cache.get(&cache_key).cloned()
+        {
+            assert_eq!(package.package.kind, target.ty);
+            return Ok(package);
+        }
+
+        let profile = project.resolve_profile(profile_name)?;
+        let mut assembler = self
+            .assembler
+            .clone()
```
bobbinth, line 188: Could cloning the assembler here cause issues later on? Basically, any time we call `assemble_source_package()` we'd start using a "fresh" assembler - i.e., an assembler without any linked libraries. This is fine for public entrypoints - e.g., w/e comes through the `assemble_impl()` but we also call this from `resolve_dependency_package()` - which may assume that we already have linked some packages into the assembler. Though, I'm not 100% sure about this.

crates/assembly/src/assembler/debuginfo.rs
bobbinth: Renamed into `register_procedure_debug_info()`. Also, simplified the signature to be just:
```rust
pub fn register_procedure_debug_info(
    &mut self,
    procedure: &Procedure,
    source_manager: &dyn SourceManager,
) -> Result<(), Report>
```

AFAICT, all other info was already included in the `Procedure` struct - but let me know if I missed some nuance.

crates/assembly/src/project/runtime_dependencies.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::project::ResolvedPackage;
+
+// RUNTIME DEPENDENCIES
+// ================================================================================================
+
+#[derive(Default)]
+pub(super) struct RuntimeDependencies {
+    pub deps: BTreeMap<PackageId, PackageDependency>,
+    pub kernel: Option<Arc<MastPackage>>,
+}
```
bobbinth, line 17: I added this struct to encapsulate handling of runtime dependences. I think this can probably can be simplified further - but we can do it in future PRs.

crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub debug_sources_section: DebugSourcesSection,
+}
+
+impl DebugInfoSections {
+    pub fn register_procedure_debug_info(
+        &mut self,
+        procedure: &Procedure,
+        source_manager: &dyn SourceManager,
+    ) -> Result<(), Report> {
+        if let Ok(file_line_col) = source_manager.file_line_col(*procedure.span()) {
```
huitseeker, line 54: `trim_paths` looks recorded in provenance, but I don’t think it trims any collected debug path yet. `register_procedure_debug_info` writes `file_line_col.uri.path()` as-is, and `apply_debug_options()` says there is nothing to rewrite.

Tiny local repro:

```rust
let leaked_path = package.mast.mast_forest().debug_info().asm_ops().iter()
    .filter_map(|op| op.location())
    .map(|location| location.uri().path())
    .find(|path| path.contains(&host_prefix));
assert!(leaked_path.is_none());
```

That currently finds the tempdir path under a `debug = true, trim-paths = true` profile.

crates/project/src/lib.rs
```diff
@@ -54,6 +55,165 @@ pub type Metadata = Map<Span<Arc<str>>, Span<Value>>;
...
+    ///
+    /// If the given manifest source belongs to a package within a larger workspace, this function
+    /// will attempt to resolve the workspace and extract the package from it.
+    pub fn load_project_reference(
+        name: &str,
+        path: impl AsRef<std::path::Path>,
+        source_manager: &dyn miden_assembly_syntax::debuginfo::SourceManager,
+    ) -> Result<Self, Report> {
+        let path = path.as_ref();
+        let (manifest_path, check_first) = if path.is_dir() {
```
huitseeker: I think this breaks when the dependency path points at a workspace `miden-project.toml` file instead of the directory. For file paths `check_first` is `false`, so `try_load_as_workspace_member` skips the initial directory and never checks that workspace manifest before falling back to `Package::load`.

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,728 @@
...
+
+        // Section: embedded kernel package
+        if target.ty.is_executable()
+            && let Some(kernel_package) = runtime_dependencies.kernel.clone()
+        {
+            sections.push(linked_kernel_package_section(kernel_package.as_ref()));
+        }
+
+        // Section: debug info
+        if self.assembler.emit_debug_info {
```
huitseeker: The new package debug sections look like they come from the wrong assembler instance. This method builds with the local cloned `assembler`, but the sections are read from `self.assembler.debug_info`, so the emitted `DEBUG_*` sections end up invalid.

```rust
let section = package.sections.iter()
    .find(|section| section.id == SectionId::DEBUG_SOURCES)
    .unwrap();
DebugSourcesSection::read_from(&mut SliceReader::new(section.data.as_ref())).unwrap();
```

That panics with `unsupported debug_sources version: 0, expected 1`.

bitwalker: Ah indeed, good catch!

crates/project/src/dependencies/graph.rs
bitwalker: Indeed, I've revisited how the graph is constructed to ensure we use the resolver, with a few new regression tests to cover that behavior.

crates/assembly/src/assembler/debuginfo.rs
bitwalker: That's true - the compiler is set up to use that information, but we weren't handling it for MASM sources. I've implemented functionality to trim source paths now when configured to do so. We may want to refine the behavior in the future though, for now it simply trims paths relative to the current working directory.

crates/project/src/lib.rs
bitwalker: Yes, I refactored this a bit to ensure we use the same behavior everywhere when loading project manifests.

crates/assembly/src/project.rs
bitwalker: AFAIK, we don't assume anything like that when resolving dependencies (the reason we clone the assembler in those cases, is in order to handle assembling the dependency if it is a source dependency). I'll take another look though to make sure I haven't missed any edge cases around this though

bobbinth: Makes sense - though, it may be good then to change the structure a bit to make things a bit more explicit - otherwise, it is not immediately if `self.assembler` is meant for anything more than just holding the initial assembler state (i.e., there is no expectation that `self.assembler` state is going to change over time).
y` 
But also, I think this API is probably going to change anyways as we remove `Library`/`Program` (ideally, we wouldn't need a separate `ProjectAssembler` struct after this) - so, nothing needs to change now.

crates/assembly/src/project/package_ext.rs
```diff
@@ -0,0 +1,323 @@
...
+        let mut excluded = self.excluded_target_roots(target, &root_path)?;
+        excluded.insert(root_path.clone());
+        let support =
+            read_support_module_paths(&root_dir, target.namespace.inner().as_ref(), &excluded)?;
+
+        Ok(TargetSourcePaths { root: root_path, root_dir, support })
+    }
+
+    fn effective_manifest_hash_input(&self) -> Result<String, Report> {
+        let mut manifest = self.to_toml()?;
```
huitseeker, line 216: Provenance here hashes the whole effective manifest, not just the selected target and active profile. So e.g. adding an unrelated `[[bin]]` makes reuse fail with `different source provenance` even when the library bytes stay the same. 

Is there a simple heuristic that would let us hash only the parts that can change the selected build?

crates/assembly/src/project.rs
```diff
@@ -0,0 +1,729 @@
...
+                    manifest_path,
+                    workspace_root.as_deref(),
+                )? {
+                    RegisteredSourcePackage::Loaded(package) => ResolvedPackage {
+                        linked_kernel_package: self
+                            .resolve_linked_kernel_package(package.clone())?,
+                        package,
+                    },
+                    reuse => {
+                        let package = self.assemble_source_package(
```
huitseeker, line 363: This rebuild path always tries to publish the source package after assembly. A source package with a local `.masp` runtime dep fails here with `missing dependency 'predep'`, even though the root build can link that local artifact directly. 
Not sure if we should skip publication when runtime deps are local-only, or instead register those local artifacts first?

crates/assembly/src/project/package_ext.rs
bitwalker: Yeah, it's a bit aggressive currently. I'm not sure if there is a simple heuristic we can use, but there is a refinement we can add to exclude executable target metadata when computing the provenance of library targets.

crates/assembly/src/project.rs
bitwalker: I think the main fix here (which we can do in a follow up PR probably) is to publish pre-assembled artifacts to the store if not already published.

However, we may need to also skip publication for packages which cannot be published to the store for some reason - I'd want to come up with some test cases that hit this after the first fix to see what exactly makes the most sense though.

bitwalker:
> In general, it may be good to create a tracking issue for the changes that we are planning to do and describe the end-state of this work in a bit more detail.

I'll go through all of the follow-up tasks and create issues in the morning, with a tracking issue that provides the overview/end-state we're aiming for. Probably won't have time tonight as it is getting late.

PR #2881, opened 2026-03-19
feat: local package registry implementation

### LINKED ISSUE #2944: Tracking issue for project assembly follow-ups
The initial implementation of the project file and project assembly (see #2510, #2886, #2877 for the bulk of that work) was only the first step in retooling how Miden Assembly projects and packages are defined and assembled. There is not only additional refactoring/clean-up work to be done (e.g. removal of `Library` and `Program`), but there are still various edge cases/bugs to be addressed. This issue is being set up to track that remaining work.

This issue is _NOT_ for discussion on the individual issues. Instead, take that to the individual sub-tasks. However, if there are missing tasks that you feel should be tracked here, feel free to leave a comment so we can discuss whether to track it as part of this issue, or if it should be treated as a separate standalone task.

bitwalker:
This extends #2877 with the local filesystem-based package registry implementation, exposed via the `miden-registry` CLI.


crates/package-registry-local/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+        },
+        Command::List => {
+            for package in registry.list() {
+                println!("{} {}", package.name, package.version);
+            }
+        },
+        Command::Show { package, version } => {
+            let package_id = PackageId::from(package);
+            let version = version.map(|version| version.parse::<Version>()).transpose()?;
+            if let Some(summary) = registry.show(&package_id, version.as_ref()) {
```
huitseeker, line 77: The current `show` path falls through when the package or requested version is missing, so the CLI exits 0 with no output. That makes typos look like success in scripts. Can we return an error here, or at least print a `not found` message and exit non-zero?

crates/package-registry-local/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+    ///
+    /// If no version is specified, the latest version will be returned.
+    ///
+    /// If the package is not indexed, or the specified version cannot be found, this returns `None`
+    pub fn show(&self, package: &PackageId, version: Option<&Version>) -> Option<PackageSummary> {
+        let (version, record) = match version {
+            Some(version) => self
+                .index
+                .get_by_version(package, version)
+                .map(|record| (version.clone(), record))?,
```
huitseeker: Isn't it more reliable to use `record.version().clone()` once the lookup succeeds?

crates/package-registry-local/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+            dependencies.push((dependency_name, requirement));
+        }
+
+        let record = match package.description.clone() {
+            Some(description) => {
+                PackageRecord::new(version.clone(), dependencies).with_description(description)
+            },
+            None => PackageRecord::new(version.clone(), dependencies),
+        };
+        self.register(package.name.clone(), record)?;
```
huitseeker: The registry state here is mutated before either write is durable, and `save()` rewrites `index.toml` in place. 

If the artifact write or index write fails, or two `publish` commands run at once, this process can keep a registered package whose artifact or index was never fully committed, or lose another writer's update. 

Can we stage this with a temp file + rename (or only install the record after persistence succeeds, ideally under a lock, ...) ?

crates/package-registry-local/src/lib.rs
```diff
@@ -0,0 +1,567 @@
...
+            .open(&self.index_path)
+            .map_err(LocalRegistryError::IndexWrite)?;
+        // Acquire an exclusive lock for writing the index file, and return an error if we cannot
+        // obtain one due to any other outstanding lock on the file.
+        //
+        // This will fail if another write is being performed on the same file, or if the index is
+        // currently being loaded by another process.
+        //
+        // See `load` for the non-exclusive lock obtained for reads
+        file.try_lock().map_err(LocalRegistryError::IndexWriteLock)?;
```
huitseeker, line 377: This lock only protects the final write, but the publish flow already loaded and changed `self.index` before it gets here. Two `miden-registry publish` processes can both read the same old index, then the later save writes its stale copy over the earlier package. That can make a successful publish disappear from the index. 

I've made the following fail:
```rust
#[test]
fn publish_preserves_updates_from_other_registry_handles() {
    let tempdir = TempDir::new().unwrap();
    let mut first_registry = load_registry(&tempdir);
    let mut stale_registry = load_registry(&tempdir);

    let first_path = tempdir.path().join("a.masp");
    let second_path = tempdir.path().join("longer-package-name.masp");
    build_package("a", "1.0.0", []).write_to_file(&first_path).unwrap();
    build_package("longer-package-name", "1.0.0", [])
        .write_to_file(&second_path)
        .unwrap();

    first_registry.publish(&first_path).unwrap();
    stale_registry.publish(&second_path).unwrap();

    let reloaded = load_registry(&tempdir);
    assert!(reloaded.show(&PackageId::from("a"), None).is_some());
    assert!(reloaded.show(&PackageId::from("longer-package-name"), None).is_some());
}
```

bitwalker: Indeed, we can address this by checking that the file has not been modified since it was initially loaded when we go to write it - if it has, we reject the write, otherwise we commit it.

huitseeker (commented):
Opened #3093: makes local registry artifacts package-aware while still accepting validated legacy paths, caches preassembled dependencies rather than publishing them (solving #2946), and repairs artifacts only under the index lock so stale or conflicting bytes cannot replace the real package.

crates/package-registry-local/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+                //
+                // This will fail if a handle to the index file has an exclusive lock on it for
+                // writing, see `save` for details.
+                //
+                // Multiple readers can hold this type of lock simultaneously, but a shared lock
+                // cannot be acquired in the presence of an exclusive lock, and vice versa.
+                file.try_lock_shared().map_err(LocalRegistryError::IndexReadLock)?;
+                file.read_to_string(&mut contents).map_err(LocalRegistryError::IndexRead)?;
+            }
+            std::println!("=== index file contents ==\n{contents}\n=== end of file contents ==");
```
huitseeker: This looks like leftover debug output. Since every registry load writes the full index to stdout, `miden-registry list --json` and `show --json` stop being valid JSON once an index exists.

crates/package-registry-local/README.md
```diff
@@ -0,0 +1,69 @@
+# miden-package-registry-local
+
+Local filesystem-backed registry and CLI for publishing and inspecting Miden packages.
+
+The data managed by this registry using the local filesystem is as follows:
+
+- The registry index, stored at `$MIDEN_SYSROOT/etc/registry/index.toml`
+- The artifacts of registered packages, as `$MIDEN_SYSROOT/lib/<digest>.masp`
```
huitseeker: This path looks stale now that artifacts use package-qualified names.

huitseeker (approved):
LGTM, aside from 2x inline fixes which I don't need to re-review.

crates/package-registry-local/src/lib.rs
bitwalker: oops - yep, thanks for catching that

PR #2886, opened 2026-03-20
Update MAST package structure for project assembly

bobbinth:
In an effort to make https://github.com/0xMiden/miden-vm/pull/2877 a bit lighter, this PR splits out changes related to the `mast-package` crate into a separate PR.

The changes here are identical to the changes in https://github.com/0xMiden/miden-vm/pull/2877, except for one case (noted in the comment) - so, hopefully, these should be pretty easy to merge into https://github.com/0xMiden/miden-vm/pull/2877.

### DIFF
--- a/crates/mast-package/src/package/mod.rs
+++ b/crates/mast-package/src/package/mod.rs
@@ -393,3 +393,120 @@ fn arbitrary_library() -> Arc<Library> {
     let value_tree = <Library as Arbitrary>::arbitrary().new_tree(&mut runner).unwrap();
     Arc::new(value_tree.current())
 }
+
+#[cfg(test)]
+mod tests {
+    use alloc::{collections::BTreeMap, sync::Arc, vec, vec::Vec};
+
+    use miden_assembly_syntax::{
+        Library,
+        ast::{Path as AstPath, PathBuf},
+        library::{LibraryExport, ProcedureExport as LibraryProcedureExport},
+    };
+    use miden_core::{
+        mast::{BasicBlockNodeBuilder, MastForest, MastForestContributor, MastNodeId},
+        operations::Operation,
+        serde::Serializable,
+    };
+
+    use super::*;
+    use crate::{Dependency, Version};
+
+    fn build_forest() -> (MastForest, MastNodeId) {
+        let mut forest = MastForest::new();
+        let node_id = BasicBlockNodeBuilder::new(vec![Operation::Add], Vec::new())
+            .add_to_forest(&mut forest)
+            .expect("failed to build basic block");
+        forest.make_root(node_id);
+        (forest, node_id)
+    }
+
+    fn absolute_path(name: &str) -> Arc<AstPath> {
+        let path = PathBuf::new(name).expect("invalid path");
+        let path = path.as_path().to_absolute().into_owned();
+        Arc::from(path.into_boxed_path())
+    }
+
+    fn build_library(export: &str) -> Arc<Library> {
+        let (forest, node_id) = build_forest();
+        let path = absolute_path(export);
+        let export = LibraryProcedureExport::new(node_id, Arc::clone(&path));
+
+        let mut exports = BTreeMap::new();
+        exports.insert(path, LibraryExport::Procedure(export));
+
+        Arc::new(Library::new(Arc::new(forest), exports).expect("failed to build library"))
+    }
+
+    fn build_package(
+        name: &str,
+        kind: TargetType,
+        export: &str,
+        dependencies: impl IntoIterator<Item = Dependency>,
+        sections: Vec<Section>,
+    ) -> Package {
+        let mut package = *Package::from_library(
+            PackageId::from(name),
+            Version::new(1, 0, 0),
+            kind,
+            build_library(export),
+            dependencies,
+        );
+        package.sections = sections;
+        package
+    }
+
+    fn build_kernel_package(name: &str) -> Package {
+        build_package(name, TargetType::Kernel, &format!("{name}::boot"), [], Vec::new())
+    }
+
+    fn kernel_dependency(package: &Package) -> Dependency {
+        Dependency {
+            name: package.name.clone(),
+            kind: TargetType::Kernel,
+            version: package.version.clone(),
+            digest: package.digest(),
+        }
+    }
+
+    #[test]
+    fn embedded_kernel_package_rejects_duplicate_kernel_sections() {
+        let kernel = build_kernel_package("kernel");
+        let kernel_bytes = kernel.to_bytes();
+        let package = build_package(
+            "app",
+            TargetType::Library,
+            "app::entry",
+            vec![kernel_dependency(&kernel)],
+            vec![
+                Section::new(SectionId::KERNEL, kernel_bytes.clone()),
+                Section::new(SectionId::KERNEL, kernel_bytes),
+            ],
+        );
+
+        let error = package
+            .try_embedded_kernel_package()
+            .expect_err("duplicate kernel sections should be rejected");
+
+        assert!(error.to_string().contains("multiple 'kernel' sections"));
+    }
+
+    #[test]
+    fn embedded_kernel_package_rejects_multiple_kernel_runtime_dependencies() {
+        let kernel_a = build_kernel_package("kernel-a");
+        let kernel_b = build_kernel_package("kernel-b");
+        let package = build_package(
+            "app",
+            TargetType::Library,
+            "app::entry",
+            vec![kernel_dependency(&kernel_a), kernel_dependency(&kernel_b)],
+            vec![Section::new(SectionId::KERNEL, kernel_a.to_bytes())],
+        );
+
+        let error = package
+            .try_embedded_kernel_package()
+            .expect_err("multiple kernel runtime dependencies should be rejected");
+
+        assert!(error.to_string().contains("declares multiple kernel runtime dependencies"));
+    }
+}


crates/assembly/src/tests.rs
```diff
@@ -4335,7 +4335,27 @@ fn build_program_example() -> Arc<Program> {
...
+
+    let program = assembler.assemble_program(source).unwrap();
+    let mast_forest = program.mast_forest().clone();
+    let entrypoint = program.entrypoint();
+    let mut exports = BTreeMap::new();
+    let export =
+        LibraryExport::Procedure(ProcedureExport::new(entrypoint, Path::exec_path().into()));
+    exports.insert(Path::exec_path().into(), export);
+
+    Library::new(mast_forest, exports).unwrap().into()
```
bobbinth, line 4362: This should be the only place where there is significant divergence with https://github.com/0xMiden/miden-vm/pull/2877.

crates/mast-package/src/package/serialization.rs
bobbinth, line 20: I believe these comments are now outdated.

crates/mast-package/src/package/target_type.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /// Returns the string representation of this package kind.
+    pub const fn as_str(&self) -> &'static str {
+        match self {
+            Self::Library => "library",
+            Self::Executable => "executable",
+            Self::Kernel => "kernel",
+            Self::AccountComponent => "account-component",
+            Self::NoteScript => "note",
```
huitseeker: This changes the human-readable spelling from `note-script` to `note`, and `FromStr` below no longer accepts `note-script`. Was the rename intended?

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +66,114 @@ pub struct Package {
...
-            _ => panic!("expected package to contain a library, but got an executable"),
+    #[doc(hidden)]
+    pub fn try_into_program(&self) -> Result<miden_core::program::Program, Report> {
+        if !self.is_program() {
+            Err(Report::msg(format!(
+                "cannot convert package of type {} to Executable",
+                self.kind
+            )))
+        } else {
+            Ok(self.unwrap_program())
```
huitseeker: This returns `Result`, but can still panic on malformed executable packages, because `get_export_node_id(::$exec::main)` is typed as infallible. 

With the old `MastArtifact::Executable`, deserialization enforced the invariant. Now an `Executable` package missing `::$exec::main` will crash callers like the CLI (which uses this directly for uses this directly for `.masp` inputs) instead of returning an error. 

A cleaner version would be to add a fallible accessor on `Library`, for example `try_get_export_node_id()` and use it here.

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +66,114 @@ pub struct Package {
...
     }
 
-    /// Creates a new package with [Program] from this [Library] package and the given
-    /// entrypoint (should be a procedure in the library).
+    #[doc(hidden)]
+    pub fn unwrap_program(&self) -> miden_core::program::Program {
+        use miden_assembly_syntax::{Path as MasmPath, ast};
+        use miden_core::program::Program;
+        assert_eq!(self.kind, TargetType::Executable);
+        Program::new(
```
huitseeker: Rebuilding executable packages with `Program::new` here always drops the original kernel. This is reported in #2877 [here](https://github.com/0xMiden/miden-vm/pull/2877#discussion_r2964878700).

crates/mast-package/src/package/serialization.rs
huitseeker: Yes. Besides the nominal change of vocabulary, I think it's worth calling out that I think we were using `MAST ARTIFACT` to paper over the executable-or-library sum type, whereas now the the encoded form is specifically a `Library`.

crates/mast-package/src/package/target_type.rs
bitwalker: Yes, this was requested by @bobbinth elsewhere, though I can't recall which issue/PR specifically

crates/mast-package/src/package/mod.rs
bitwalker: I would prefer to add any such logic directly to `Package` or to `MastForest`, so that we avoid making the removal of `Library` and `Program` any more of a headache than it will already be

crates/mast-package/src/package/serialization.rs
bitwalker: I did update the README doc for `miden-mast-package` on my branch, but not this module doc - I can do that after backporting the `unwrap_program` fix

crates/mast-package/src/package/mod.rs
bitwalker: This is partially resolved here - the assembly side of it is handled in #2877.

crates/mast-package/src/package/serialization.rs
bitwalker: This module doc has had the incorrect details revised. We probably want a more robust document, but I'd wait until we've largely stabilized the format before we go into too much detail with it - we still have another round of changes coming with the removal of `Library`

crates/mast-package/src/package/mod.rs
bitwalker: The new `Package::try_into_program` method provides this, and `unwrap_program` delegates to it. The latter will still panic, but with a better error, and panicking is expected with something called `unwrap`.

bitwalker: An additional check has been added in the `try` method to ensure the underlying MAST has the requisite export present, and returns an `Err` if not

crates/mast-package/src/package/target_type.rs
bobbinth: I think "note" is more appropriate as note packages could contain other things besides the note script. For example:

- Definition of note storage (similar to how we define account storage in account component packages).
- Note constructor procedures (we started adding these in hand-written MASM for standard notes in `miden-protocol`).

But I'd say that if we are switching to "note" nomenclature, we should do this consistently - e.g., the enum variant should also probably be named just `Note`. We could do this in a follow-up PR though.

crates/mast-package/src/package/target_type.rs
```diff
@@ -0,0 +1,182 @@
...
+    Executable = 1,
+    /// A kernel library
+    Kernel = 2,
+    /// An account component
+    AccountComponent = 3,
+    /// A note script
+    NoteScript = 4,
+    /// A transaction script
+    TransactionScript = 5,
+}
```
bobbinth, line 35: Related to one of the above comments (and probably not for this PR), I'd probably rename:

- `NoteScript` to just `Note`.
- `Kernel` to `KernelLibrary`.

crates/mast-package/src/package/section.rs
```diff
@@ -33,6 +33,10 @@ impl SectionId {
     /// `miden-objects::account::AccountComponentMetadata`, i.e. name, descrioption, storage, that
     /// is associated with this package.
     pub const ACCOUNT_COMPONENT_METADATA: Self = Self(Cow::Borrowed("account_component_metadata"));
```
bobbinth, line 35: Not from this PR: but ``miden-objects::account::AccountComponentMetadata` is an obsolete path. Now this is ``miden-protocol::account::AccountComponentMetadata`.

crates/assembly-syntax/src/library/module.rs
```diff
@@ -14,13 +14,21 @@ use crate::{
...
     items: Vec<ItemInfo>,
 }
 
 impl ModuleInfo {
     /// Returns a new [`ModuleInfo`] instantiated library path.
-    pub fn new(path: Arc<Path>) -> Self {
-        Self { path, items: Vec::new() }
+    pub fn new(version: Option<crate::Version>, path: Arc<Path>) -> Self {
+        Self { version, path, items: Vec::new() }
+    }
```
bobbinth: I'd update the doc comment to mention `version` and maybe explain why it is optional.

Also nit: putting an optional parameter first feels a bit odd. I'd maybe write this as:
```rust
pub fn new(path: Arc<Path>, version: Option<crate::Version>) -> Self {
    Self { version, path, items: Vec::new() }
}
```

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +69,218 @@ pub struct Package {
...
+        let main_path = MasmPath::exec_path().join(ast::ProcedureName::MAIN_PROC_NAME);
+        if let Some(digest) = self.mast.get_procedure_root_by_path(&main_path)
+            && let Some(entrypoint) = self.mast.mast_forest().find_procedure_root(digest)
+        {
+            let mast_forest = self.mast.mast_forest().clone();
+            match self.try_embedded_kernel_library()? {
+                Some(kernel_library) => Ok(Program::with_kernel(
+                    mast_forest,
+                    entrypoint,
+                    kernel_library.kernel().clone(),
```
bobbinth, line 155: Not for this PR, but I think what we are calling `Kernel` should probably be renamed as it is just a list of kernel procedure digests. Maybe it should be more like `KernelApi` or `KernelProcedures`?

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +69,218 @@ pub struct Package {
...
+                "expected package '{}' to contain a kernel, but kind was '{}'",
+                self.name, self.kind
+            )));
+        }
+
+        KernelLibrary::try_from(self.mast.clone()).map_err(|error| Report::msg(error.to_string()))
+    }
+
+    #[doc(hidden)]
+    pub fn try_into_program(&self) -> Result<miden_core::program::Program, Report> {
```
bobbinth, line 136: Question: why do we had this and a few other procedures?

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +69,218 @@ pub struct Package {
...
+        };
+        kernel_package.try_into_kernel_library().map(Some)
     }
 
-    /// Unwraps the package's MAST artifact as a [Program] or panics if it is a [Library]
-    pub fn unwrap_program(&self) -> Arc<Program> {
-        match self.mast {
-            MastArtifact::Executable(ref prog) => Arc::clone(prog),
-            _ => panic!("expected package to contain a program, but got a library"),
+    fn embedded_kernel_package(&self) -> Result<Option<Self>, Report> {
```
bobbinth, line 195: May be good to leave a brief doc comment explaining what this method does.

My understanding is that we are trying to read the kernel section of the package and instantiate a kernel library from it. If there is no such section we return `None`, but if the section is present but deserialization of the kernel library fails, we get an error back.

crates/mast-package/src/package/mod.rs
bitwalker: I think it depends on what we do with `KernelLibrary` - if we keep it, I would name that `Kernel`, and just get rid of the `Kernel` type all together (i.e. flatten the digest vector into `KernelLibrary`). If it's possible to just use `Package` for all of this, then we could get rid of `KernelLibrary` too, and have one less type in the mix.

crates/mast-package/src/package/mod.rs
bitwalker: For backwards compatibility with functionality of the current `Assembler` (i.e. extracting a `Library` or `Program` from an assembled package).

crates/mast-package/src/package/mod.rs
bitwalker: Correct, I can add some clarifying docs here though

crates/mast-package/src/package/target_type.rs
bitwalker: Any particular reason why you feel `KernelLibrary` is better here? The `Library` suffix feels a bit redundant. The only real reason to have it, IMO, is if we changed `Executable` to `Kernel`, and `Kernel` to `KernelLibrary`, but tbh, I think the current distinction between `Executable` and `Kernel` is clear enough. What's your take?

bitwalker:
> Also, originally I was thinking't that we'd squash-merge this PR into `main`, and then merge `main` into #2877. But now that #2877 is based on this branch - squash-merged may create complications. @bitwalker - let me know what approach you'd prefer.

It was pretty messy rebasing my branch on this one, and there are two other PRs in the stack, so squash-merging will likely cause some messy rebasing/merging on my end, and I'd prefer to avoid that. If you do just a regular merge of this branch (and I can manually squash my patches on top of your initial commit to keep things contained in a single commit if you prefer), then the `bitwalker/project-assembler` branch will automatically migrate to using `main` as its base branch without any manual rebasing required.

crates/mast-package/src/package/mod.rs
```diff
@@ -48,51 +69,225 @@ pub struct Package {
...
+            && let Some(entrypoint) = self.mast.mast_forest().find_procedure_root(digest)
+        {
+            let mast_forest = self.mast.mast_forest().clone();
+            match self.try_embedded_kernel_library()? {
+                Some(kernel_library) => Ok(Program::with_kernel(
+                    mast_forest,
+                    entrypoint,
+                    kernel_library.kernel().clone(),
+                )),
+                None => Ok(Program::new(mast_forest, entrypoint)),
```
huitseeker, line 157: Happy with @bitwalker's context in [this thread](https://github.com/0xMiden/miden-vm/pull/2877#discussion_r2967835778), noting for my book-keeping.

crates/mast-package/src/package/target_type.rs
bobbinth: I guess it depends on some other names - but right now we overloading the term `Kernel` a lot (e.g., we have the `Kernel` struct which is just a list of digests of kernel procedures). So, I thought that making naming more specific would reduce the ambiguity.

But if we change `Kernel` into something else and always mean "kernel library" when we say "kernel", I think keeping it as is would be fine.

crates/mast-package/src/package/mod.rs
bobbinth: I think we still need to keep a dedicated type for the list of kernel procedures because it is used in places where we may not need (or want) to have the full kernel package (e.g., the verifier needs the list of procedure digests to verify the proofs, but it doesn't really anything else).

I think `KernelApi` or `KernelAbi` are probably the best options I can think of now - e.g.,
```rust
pub struct KernelApi(Vec<Digest>);
```

bobbinth:
> if you do just a regular merge of this branch (and I can manually squash my patches on top of your initial commit to keep things contained in a single commit if you prefer), then the `bitwalker/project-assembler` branch will automatically migrate to using `main` as its base branch without any manual rebasing required.

Do you mean the "Create a merge commit" option? If so, I'm happy to merge - but also, now you should be able to merge as well.

Issue #2862, opened 2026-03-17
C-like enum handling skips validation and constant materialization


huitseeker:
In #2848, commit `8d715cfef` (`fix(assembly): provide better support for exporting non-C-like enum types`), `define_enum` in `crates/assembly-syntax/src/ast/module.rs` added an early return for `ty.is_c_like()`.

The root cause is that this early return runs before the normal enum validation path. Because parsed enums are currently C-like, this bypasses key behavior.

As a result, C-like enum variants are not materialized as constants:

```rust
#[test]
fn c_like_enum_variants_are_defined_as_constants() {
    let context = SyntaxTestContext::default();
    let module = context
        .parse_module(
            r#"
enum Tag : u8 {
    A,
    B,
}
"#,
        )
        .expect("expected C-like enum module to parse");

    let mut constant_names = module.constants().map(|constant| constant.name().as_str());
    assert_eq!(constant_names.next(), Some("A"));
    assert_eq!(constant_names.next(), Some("B"));
    assert_eq!(constant_names.next(), None);
}
```

Expected behavior is constants `A` and `B` are defined.  
Actual behavior is no constants are defined.

The same branch also skips symbol conflict validation, so conflicting names are accepted:

```rust
#[test]
fn c_like_enum_name_conflicts_are_rejected() {
    let context = SyntaxTestContext::default();
    let error = context
        .parse_module(
            r#"
type Tag = u8

enum Tag : u8 {
    A,
}
"#,
        )
        .expect_err("expected enum/type name conflict to be rejected");

    let rendered = format!("{}", PrintDiagnostic::new_without_color(&error));
    assert!(rendered.contains("conflict") || rendered.contains("already defined"));
}
```

Expected behavior is a symbol conflict error for `Tag`.  
Actual behavior is parsing succeeds and both definitions are accepted.

The fix is to keep C-like enums on the existing validation path (including conflict checks and constant definition), and only use the short path for non-C-like enums.

crazywriter1:
I'd like to be assigned to this issue. I have a fix ready in PR #2887.

huitseeker:
Fixed in #2887

PR #2887, opened 2026-03-20
fix(assembly-syntax): keep C-like enum validation/constant materialization in define_enum

crazywriter1:
## Summary
This PR fixes the bug described in #2862 where C-like enums were taking an early return path in `define_enum`, bypassing enum validation and constant materialization.

## What changed
In `crates/assembly-syntax/src/ast/module.rs`, the early-return condition was adjusted so that C-like enums continue through the normal validation flow (including constant definition and symbol conflict checks) instead of being exported without defining variants as constants.

## Test plan
- `cargo test -p miden-assembly-syntax`
- `cargo clippy -p miden-assembly-syntax --all-targets --all-features -- -D warnings`


crazywriter1:
Could you please check? @huitseeker

crazywriter1:
Thanks for the review sir. Added the CHANGELOG entry and commented on #2862 for assignment. @huitseeker

PR #2888, opened 2026-03-20
test(processor): cover nested loop end flags across fragmentation

huitseeker:
Introduces a test to guard against a bug where stop/resume-style fragmentation preserves the trace shape loosely, but loses or mutates `is_loop_body`-related state at `END` rows.


processor/src/trace/parallel/tests.rs
```diff
@@ -434,6 +438,49 @@ fn test_trace_generation_at_fragment_boundaries(
     insta::assert_compact_debug_snapshot!(testname, DeterministicTrace(&trace_from_fragments));
 }
 
+#[test]
+fn test_nested_loop_end_flags_stable_across_fragmentation() {
+    const SMALL_FRAGMENT_SIZE: usize = 7;
```
plafer, line 466: I'd document the program trace clock by clock, so that it's easier to see where this `7` lands, similar to `test_trace_generation_at_fragment_boundaries()`

PR #2893, opened 2026-03-21
fix(assembly): record ProcRef invoke kind correctly in VerifyInvokeTargets

### LINKED ISSUE #2811: procref incorrectly recorded as exec in invocation metadata
In `crates/assembly-syntax/src/sema/passes/verify_invoke.rs:169`, `VerifyInvokeTargets::visit_mut_procref` records `InvokeKind::Exec` instead of `InvokeKind::ProcRef`:

```rust
fn visit_mut_procref(&mut self, target: &mut InvocationTarget) -> ControlFlow<()> {
    self.visit_mut_invoke_target(target)?;
    self.invoked.insert(Invoke::new(InvokeKind::Exec, target.clone())); // ← should be ProcRef
    ControlFlow::Continue(())
}
```

This does not affect linking behavior today because the callgraph edge recording does not branch on `InvokeKind`. However, it misleads tooling or diagnostics that interpret `Procedure::invoked` as an accurate record of instruction kinds.

Fix: change `InvokeKind::Exec` to `InvokeKind::ProcRef` in the insert call.

amathxbt:
Closes #2811

## Summary

`VerifyInvokeTargets::visit_mut_procref` was inserting `InvokeKind::Exec` into the invoked set instead of `InvokeKind::ProcRef`. This means `Procedure::invoked()` produces incorrect metadata for any procedure that uses `procref`, misrepresenting the instruction kind as an `exec`.

## Root cause

In `crates/assembly-syntax/src/sema/passes/verify_invoke.rs`, the `visit_mut_procref` method mistakenly used `InvokeKind::Exec` when recording the invocation:

```rust
// Before
fn visit_mut_procref(&mut self, target: &mut InvocationTarget) -> ControlFlow<()> {
    self.visit_mut_invoke_target(target)?;
    self.invoked.insert(Invoke::new(InvokeKind::Exec, target.clone())); // wrong
    ControlFlow::Continue(())
}
```

## Fix

Changed `InvokeKind::Exec` to `InvokeKind::ProcRef` in the insert call:

```rust
// After
fn visit_mut_procref(&mut self, target: &mut InvocationTarget) -> ControlFlow<()> {
    self.visit_mut_invoke_target(target)?;
    self.invoked.insert(Invoke::new(InvokeKind::ProcRef, target.clone())); // correct
    ControlFlow::Continue(())
}
```

## Notes

- The linking behavior is unaffected today because the call graph edge recording does not branch on `InvokeKind`. However, the incorrect kind misleads tooling and diagnostics that consume `Procedure::invoked` metadata.
- This is a one-line change with no risk of regression.

cc @huitseeker


bobbinth (approved):
Looks good! Thank you!

bitwalker:
I think this should be reverted - the purpose of recording the invoke kind as `exec` is because a `procref` captures a reference to a procedure for _later_ invocation. We must pessimistically assume that the procedure is invoked - the specific means cannot be known, so `exec` is used as it is the most general of the invocation instructions.

The linker largely assumes that `procref` invocation kinds are analogous to "this procedure was named somewhere", but it doesn't carry the full weight of "this procedure was invoked". However, this specific pass records actual `procref` references, i.e. the literal instruction is in the AST - and _that_ translates to being equivalent to "this procedure was invoked".

/cc @huitseeker @bobbinth 

NOTE: We may want to refine the invocation kinds to distinguish procref-derived-from-`procref` from procref-derived-from-symbol-reference - but I largely think that is distinction without a difference in literally all other places in the assembler except here, where we record "invocations". That could change of course, which is why I mention this possibly being worth distinguishing in the future.

bobbinth:
> I think this should be reverted - the purpose of recording the invoke kind as `exec` is because a `procref` captures a reference to a procedure for _later_ invocation. We must pessimistically assume that the procedure is invoked - the specific means cannot be known, so `exec` is used as it is the most general of the invocation instructions.
> 
> The linker largely assumes that `procref` invocation kinds are analogous to "this procedure was named somewhere", but it doesn't carry the full weight of "this procedure was invoked". However, this specific pass records actual `procref` references, i.e. the literal instruction is in the AST - and _that_ translates to being equivalent to "this procedure was invoked".

Yes, I think this makes sense. But I would add a comment explaining why this is done (could be basically what you wrote above) to make sure we don't do this again.

amathxbt:
Thanks for the detailed explanation, @bitwalker that makes total sense now. I misread this as a naming mismatch since we are inside `visit_mut_procref`, but did not account for the fact that `procref` is a deferred invocation and `Exec` is the conservative/pessimistic choice the linker actually needs.

Opened #2908 to revert back to `InvokeKind::Exec` and add a comment explaining the reasoning (essentially what you wrote above), so this does not get changed again by accident. Appreciate the thorough context — and the note about potentially distinguishing `procref`-from-instruction vs `procref`-from-symbol-reference is worth keeping in mind for the future.

cc @bobbinth @huitseeker

PR #2895, opened 2026-03-21
fix(project): VersionSet::contains() must check semver range even when digest is present

### LINKED ISSUE #2884: `VersionSet::contains()` ignores semver when a digest is present
This is pretty minor, but `VersionSet::contains()` only checks the digest when a candidate version has one. It does not also check the semver range. A version can be accepted into the set even when its semver should be rejected.

This repro fails on `next`:

```rust
#[test]
fn version_set_contains_requires_both_digest_and_semver() {
    let digest = Rpo256::hash(b"same-digest");
    let set = VersionSet::from(Version::new("1.0.0".parse().unwrap(), digest));
    let candidate = Version::new("9.9.9".parse().unwrap(), digest);

    assert!(
        !<VersionSet as pubgrub::VersionSet>::contains(&set, &candidate),
        "set {set} unexpectedly contains {candidate}"
    );
}
```

The fix is to require both the digest filter and the semver range when the set has a digest filter. 

As originally reported by @huitseeker

amathxbt:
## Summary

Fixes #2884.

`VersionSet::contains()` short-circuited to check only the content digest when the queried version carried a digest, completely skipping the semver range validation. This meant a version with a known digest could be treated as contained in a set whose semver range explicitly excluded it.

The documented containment semantics (in the struct-level docstring) state that **both** conditions must hold:
1. The semver component of `v` must be covered by `self.range`.
2. If this set specifies a content digest filter (`!self.digests.is_empty()`), then `v` must also carry a matching digest.

## Changes

**`crates/project/src/dependencies/resolver/version_set.rs`**
- Rewrote `contains()` so the semver range is always checked first (early return `false` if it fails).
- Only then applies the optional digest filter: if `self.digests` is empty any semver-matching version is accepted; otherwise the version must carry a digest present in `self.digests`.

## Testing

Unit tests for `VersionSet` containment can be extended to cover the case of a version-with-digest against a set that has a non-overlapping semver range but a matching digest -- previously this returned `true`, now it correctly returns `false`.

## Checklist

- [x] Branched from `next`
- [x] Commits follow semantic commit format
- [x] Fix matches the documented set-containment semantics in the struct docstring
- [x] No emojis in code


huitseeker:
Thanks! But we'll go with the fix that, as mentioned in the issue, is already in #2877. To avoid this sort of double work in the future, has a look at the section in CONTRIBUTING.md, which speaks about issue assignment.

amathxbt:
Perfect work !
If any help needed please tag me so we gonna fix it together I appreciate your works Chad 

@huitseeker

PR #2899, opened 2026-03-23
fix: return structured errors for unsafe assembly resolution paths

huitseeker:
This PR turns four panic paths in assembly and syntax handling into normal errors.

- makes cyclic call graphs fail with typed cycle errors. Self-calls and rootless cycles no longer panic, and linker state is rolled back on failure.

- makes oversized modules fail at a hard item limit. Resolver construction now rejects modules that cannot fit in ItemIndex, instead of building partial state or panicking.

- makes unresolved alias exports fail cleanly during library assembly. Digest exports still work when they resolve to a real procedure, and unknown digests now return an error instead of unwinding.

- makes invoke target checks require procedures. Semantic analysis now rejects local aliases, types, and other non-procedure items as invoke targets, and assembly keeps a second guard so invalid targets return an error instead of panicking.

A review reading commit-by-commit is highly encouraged.

### DIFF
--- a/crates/assembly/src/assembler.rs
+++ b/crates/assembly/src/assembler.rs
@@ -652,11 +652,48 @@ impl Assembler {
             },
 
             SymbolItem::Alias { alias, resolved } => {
-                // All aliases should've been resolved by now
-                let resolved = resolved.get().unwrap_or_else(|| {
-                    panic!("unresolved alias {symbol_path} targeting: {}", alias.target())
-                });
-                return self.export_symbol(resolved, module_kind, symbol_path, mast_forest_builder);
+                if let Some(resolved) = resolved.get() {
+                    return self.export_symbol(
+                        resolved,
+                        module_kind,
+                        symbol_path,
+                        mast_forest_builder,
+                    );
+                }
+
+                let Some(ResolvedProcedure { node, signature }) = self.resolve_target(
+                    InvokeKind::ProcRef,
+                    &alias.target().into(),
+                    gid,
+                    mast_forest_builder,
+                )?
+                else {
+                    return Err(self.unresolved_alias_report("export", &symbol_path, alias));
+                };
+
+                let digest = mast_forest_builder
+                    .get_mast_node(node)
+                    .expect("resolved alias export node must exist")
+                    .digest();
+                let pctx = ProcedureContext::new(
+                    gid,
+                    /* is_program_entrypoint= */ false,
+                    symbol_path.clone(),
+                    Visibility::Public,
+                    signature.clone(),
+                    module_kind.is_kernel(),
+                    self.source_manager.clone(),
+                );
+                let procedure = pctx.into_procedure(digest, node);
+                self.linker.register_procedure_root(gid, digest)?;
+                mast_forest_builder.insert_procedure(gid, procedure)?;
+
+                return Ok(LibraryExport::Procedure(ProcedureExport {
+                    node,
+                    path: symbol_path,
+                    signature: signature.map(Arc::unwrap_or_clone),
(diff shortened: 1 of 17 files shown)


crates/assembly/src/assembler.rs
```diff
@@ -154,6 +154,39 @@ impl Assembler {
...
+            Ok(SymbolResolution::Exact { path, .. })
+            | Ok(SymbolResolution::Module { path, .. })
+            | Ok(SymbolResolution::External(path)) => Some(path.into_inner()),
+            Ok(SymbolResolution::Local(_)) | Ok(SymbolResolution::MastRoot(_)) => None,
+            Err(err) => return Report::new(err),
+        }
+        .or_else(|| match callee {
+            InvocationTarget::Symbol(symbol) => Some(Path::from_ident(symbol).into_owned().into()),
+            InvocationTarget::Path(path) => Some(path.clone().into_inner()),
+            InvocationTarget::MastRoot(_) => None,
```
bitwalker, line 182: We shouldn't produce an error when a MAST root is referenced which we don't have available - we have to assume it is a valid reference and will be provided at runtime via other means.

The only time a MAST root reference can be treated as an error is:

* It is the target of a `syscall` and the MAST root isn't exported from the kernel

Some edge cases that we could consider raising errors on:

* It is the immediate target of a `procref` (i.e. not the target after resolving through aliases/etc. - as it makes no sense to do `procref.0xF...` when you already have the procedure digest on hand.) I suppose we could support this for uniformity in procedure references, but it is still a strange one to allow.
* The MAST root is in the forest, but isn't a procedure root. This one is debatable though, because it could be the case that there is a procedure defined externally (that would be provided at runtime), that happens to collide with a subtree of the forest. Personally I'd probably not error here either.

huitseeker: OK, thanks! ~Will fix.~ Addressed in db9f656

Since you've caught me lacking on late-binding for procedures a few time, is there a place where the runtime design is written down, or should I aim to produce a draft (notably, I've seen only a few entrypoints at a low level: `Program`/`Library`, so I'm guessing this is WIP?) ?

bitwalker: I think unfortunately this is one of those areas where it is a combination of tribal knowledge (primarily @bobbinth, @plafer, and myself), and discussions spread across the repo in a bunch of places. I think we also used to document it in the old Miden docs, but not sure if that is still the case somewhere.

That said, I do think it would be worth our time to produce an up-to-date design document on some of these things that sit at the boundary of what the processor requires/assumes, what the assembler requires/assumes, and what the compiler requires/assumes.

Some things also predate `procref` - like it used to be the case that you could only `dynexec`/`dyncall` a procedure from something you had already assembled, because you could only get the MAST root _after_ assembly. The `procref` instruction solved that issue, so now there isn't _really_ much of a reason to write an actual MAST root in MASM by hand - but there still might be reasons one would do so (for example, a compiler that emits MASM that links against a previously assembled artifact, and does so using MAST roots rather than procedure names). Our compiler doesn't actually do that currently, but that was in large part due to the lack of packages and binding generation (where we could just emit MAST roots rather than procedure names to save the assembler/linker the work).

Anyway, I'm happy to sit down anytime and try to brain dump all the little details like that which I can recall.

bitwalker (approved):

crates/assembly/src/tests.rs
bobbinth, line 1: Not for this PR, but we really should split this file up into smaller files with thematically grouped tests.

bobbinth (approved):
Looks good! Thank you! Though, this is a pretty superficial review from me.

PR #2900, opened 2026-03-23
Remove `adv_push.x`

### LINKED ISSUE #2565: Consider adding convenience instructions for pushing from adv stack
Agree with the discussion above about removing `adv_push.n` instruction. In the future, we could also introduce convenience instructions like:

- `adv_pushw`  - which would expand to `padw adv_loadw`.
- `adv_loaddw` - which would expand to `adv_loadw swapw adv_loadw`.
- `adv_pushdw` - which would expand to `padw adv_loadw padw adv_loadw swapw`

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2675298887_

Al-Kindi-0:
## Describe your changes

Fixes #2565

I haven't added `adv_loaddw` and `adv_pushdw` as I am not sure of their usefulness

- Updated `CHANGELOG.md'


crates/assembly-syntax/src/ast/instruction/print.rs
```diff
@@ -257,7 +257,9 @@ impl PrettyPrint for Instruction {
             Self::MemStream => const_text("mem_stream"),
             Self::AdvPipe => const_text("adv_pipe"),
 
-            Self::AdvPush(value) => inst_with_imm("adv_push", value),
+            Self::AdvPush => const_text("adv_push"),
+            Self::AdvPushPair => const_text("adv_push_pair"),
+            Self::AdvPushW => const_text("adv_pushw"),
             Self::AdvLoadW => const_text("adv_loadw"),
```
PhilippGackstatter, line 262: I think @bobbinth has previously referred to two elements as a half-word (`hw`), which would go nicely with the word (`w` in instructions) and double-word (`dw` in instructions) terminology we already have. 

Should we adopt this instead and have `adv_pushhw` ("push half-word")?

The only downside is it looks a bit off due to the double h, but maybe that's fine in the name of consistency.

crates/lib/core/src/handlers/falcon_div.rs
```diff
@@ -65,9 +65,10 @@ pub fn handle_falcon_div(process: &ProcessorState) -> Result<Vec<AdviceMutation>
     // Assertion from the original code: r_hi should always be zero for Falcon modulus
     assert_eq!(r_hi, ZERO);
 
-    // `mod_12289` consumes the quotient via `adv_push.2` followed by the remainder via
-    // `adv_push.1`. Push the remainder first (so it stays below the quotient) and rely on
-    // `extend_stack_for_adv_push` to take care of the per-word little-endian layout.
+    // `mod_12289` consumes the quotient via `repeat.2 adv_push end` followed by the remainder
+    // via `adv_push`. Push the remainder first (consumed last), then the quotient (consumed
+    // first). `extend_stack` reverses internally, so [q_hi, q_lo] -> advice [q_lo, q_hi]
+    // with q_lo popped first, ending up on top after the two adv_push ops.
```
PhilippGackstatter: If q_lo is popped first, would it not end up deeper in the stack?

crates/assembly/src/instruction/mod.rs
```diff
@@ -382,8 +381,11 @@ impl Assembler {
             Instruction::Caller => env_ops::caller(block_builder),
             Instruction::Clk => block_builder.push_op(Clk),
             Instruction::AdvPipe => block_builder.push_op(Pipe),
-            Instruction::AdvPush(n) => {
-                adv_ops::adv_push(block_builder, proc_ctx, n.expect_value(), n.span())?
+            Instruction::AdvPush => block_builder.push_op(AdvPop),
+            Instruction::AdvPushPair => block_builder.push_op_many(AdvPop, 2),
```
PhilippGackstatter: I think that makes `adv_push_pair` identical to `adv_push adv_push` and `repeat.2 adv_push end`. It looks like the real difference lies in how we construct the advice stack in `AdviceStackBuilder` (specifically `push_for_adv_push_pair`), but I'm not sure this justifies having the `adv_push_pair` instruction at all.

Another approach is to let `adv_push_pair` push in little-endian format based on a little-endian word layout. For example, in the protocol we have asset keys defined as: `[asset_id_suffix, asset_id_prefix, faucet_id_suffix, faucet_id_prefix]`, where suffix can be thought of as `lo` and `prefix` as `hi`.

Then:

```
# We have an asset key in its canonical layout on the advice stack:
# AS => [asset_id_suffix, asset_id_prefix, faucet_id_suffix, faucet_id_prefix]
# OS => []

# push the asset ID in its correct layout: suffix, prefix
adv_push_pair
# AS => [faucet_id_suffix, faucet_id_prefix]
# OS => [asset_id_suffix, asset_id_prefix]

adv_push_pair
# AS => []
# OS => [faucet_id_suffix, faucet_id_prefix, asset_id_suffix, asset_id_prefix]
```

So, based on a little-endian word layout on the advice stack, we can push its half-words in little-endian format onto the operand stack. I think this would be implemented with `AdvPop AdvPop Swap`. I'm honestly not sure if this is useful enough either though - at least I can't think of an instance in the protocol where we'd use this.

PhilippGackstatter (commented):
It looks like `adv_push_pair` is identical to `adv_push adv_push`, and I'm not sure if this was the original intention, but even if, I'm not sure if it's a meaningful enough instruction. Curious what others think.

crates/lib/core/asm/math/u64.masm
```diff
@@ -378,7 +378,7 @@ pub proc divmod(b: u64, a: u64) -> (remainder: u64, quotient: u64)
     # Advice stack now has [q_hi, q_lo, r_hi, r_lo]
     # => [b_lo, b_hi, a_lo, a_hi, ...]
 
-    adv_push.2
+    repeat.2 adv_push end
```
PhilippGackstatter: One more comment: I think `adv_push adv_push` might be easier to read and is less visually busy than the repeat loop, but not a strong opinion.

Al-Kindi-0:
> It looks like `adv_push_pair` is identical to `adv_push adv_push`, and I'm not sure if this was the original intention, but even if, I'm not sure if it's a meaningful enough instruction. Curious what others think.

Agreed, and that is why I didn't include the other ops in the original issue.
From my side, we can remove `adv_push_pair` and make this PR only about removing `adv_push.x`

crates/assembly-syntax/src/ast/instruction/print.rs
bitwalker: I'd have preferred to have us separate instruction suffixes from the opcode mnemonic (e.g. `<opcode>_<suffix>` vs `<opcode><suffix>`), but we're kind of all over the place with that at the moment - but if we're going to add new instructions or make breaking changes to existing ones, it'd be best to move in a consistent direction for sure.

crates/assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -242,7 +242,9 @@ pub enum Instruction {
     MemStream,
     AdvPipe,
 
-    AdvPush(ImmU8),
+    AdvPush,
+    AdvPushPair,
```
bitwalker: I do wonder if `AdvPushPair` (or `AdvPushHw` if we go with @PhilippGackstatter's suggestion) is really worth having at all, vs just forcing `adv_push adv_push`.

I'm not opposed to it, just not sure how frequent it actually occurs in practice to warrant dedicated syntax for it.

crates/assembly-syntax/src/parser/grammar.lalrpop
bitwalker, line 1: Since we have a new lossless parser introduced in #2906 and #2907 - we'll want to probably incorporate these changes there as well - it may actually be preferable to rebase this PR on #2907, since it will help us move along review and merging of that work sooner, and would simplify the task of preserving equivalence between the old and new parsers.

bitwalker (approved):
Looks good to me, setting aside the bikeshedding over the new instruction mnemonics.

I do think my other comment about basing this PR on top of #2907 is strongly worth considering, since we will need to incorporate these changes in that PR anyway, and it will be easier to do by building off that PR, rather than the other way around - but I'm not going to block merging of this over it.

crates/lib/core/src/handlers/falcon_div.rs
Al-Kindi-0: fixed

crates/assembly-syntax/src/ast/instruction/print.rs
Al-Kindi-0: Removed

crates/assembly/src/instruction/mod.rs
Al-Kindi-0: I have just simplified and removed `adv_push_pair`

crates/lib/core/asm/math/u64.masm
Al-Kindi-0: Indeed, did it here and elsewhere

crates/assembly-syntax/src/ast/instruction/mod.rs
Al-Kindi-0: Indeed, I have just gone ahead and removed it as I was on the fence about it to begin with

crates/assembly-syntax/src/parser/grammar.lalrpop
Al-Kindi-0: I had already rebased on next before I saw your review. I also check rebasing on #2907 and it seemed non-trivial, so I would be happy if we can merge it to next, assuming it is not much pain on your side

core/src/advice/stack.rs
```diff
@@ -89,31 +88,20 @@ impl AdviceStackBuilder {
...
-    /// builder.push_for_adv_loadw([w0, w1, w2, w3].into());
-    /// // MASM: padw adv_loadw
+    /// builder.push_word([w0, w1, w2, w3].into());
+    /// // MASM: adv_loadw (or adv_pushw)
     /// // Result: operand stack = [w0, w1, w2, w3, ...] with w0 on top
     /// ```
-    pub fn push_for_adv_loadw(&mut self, word: Word) -> &mut Self {
-        // Push elements without reversal. adv_loadw loads the structural word directly,
-        // so a `reversew` is needed afterward to get canonical order on the operand stack.
+    pub fn push_word(&mut self, word: Word) -> &mut Self {
```
huitseeker, line 104: This looks like a public Rust API rename from `push_for_adv_loadw` to `push_word`. We are fine with a hard break here, but could we still call it out in the changelog or PR text so downstream users do not find it only at compile time?

huitseeker (approved):
LGTM, but do read the inline comment plz.

crates/assembly-syntax/src/parser/grammar.lalrpop
bitwalker: No worries! I can handle it as a follow-up there when I rebase

bobbinth (approved):
Looks good! Thank you!

PR #2901, opened 2026-03-23
processor: expose advice, memory and transcript

djolertrk:
FastProcessor provides `step()` for external stepping loops but no way to extract the final `advice`, `memory`, and `pc_transcript` state afterward, which is needed to construct a valid `ExecutionOutput` via DAP/debugger while debugging transactions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.22.1
+
+#### Enhancements
+
+- Added `FastProcessor::into_parts()` to extract advice provider, memory, and precompile transcript after step-based execution ([#2901](https://github.com/0xMiden/miden-vm/pull/2901)).
+
 ## 0.22.0 (2025-03-18)
 
 #### Enhancements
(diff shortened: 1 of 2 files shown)


huitseeker:
@djolertrk This is about to undergo a bunch of changes in the direction I think you're going for but for different reasons, in #2865 — flagging to avoid duplicate work.

bitwalker:
@huitseeker Is that something planned for the v0.22 cycle? I'm assuming no. Making this function public will allow us to ship the transaction debugger for the v0.14 testnet - as far as I'm aware this is the only missing API  needed for that (this was opened as a draft so I could review it first, but it's ready for review).

djolertrk:
Is there any comment on this? :)

plafer:
@huitseeker if this also makes sense to you, feel free to merge

djolertrk:
hi @huitseeker, do you have any comments on this one? :)

bitwalker:
I'm merging this, due to its triviality and importance to shipping the transaction debugger. I expect that it will get removed/reworked as @huitseeker referred to, but that's totally fine as long as analogous functionality is preserved.

huitseeker:
Sorry for the delayed answer, yes, that's what expect will happen: reworking the triplet used in return position as a bundle.

PR #2903, opened 2026-03-23
fix: reject non-syscall references to exported kernel procedures

### LINKED ISSUE #2902: Linker accepts non-syscall references to exported kernel procedures
Exported kernel procedures are documented as `syscall`-only, but the linker also accepts them as targets of `exec`, `call`, and `procref`.

The root cause is that `SymbolResolver::resolve_invoke_target` only checks whether a resolved procedure belongs to the kernel when the invoke kind is `syscall`. On the other invoke paths, digest and path resolution can still return `SymbolResolution::Exact` for a kernel procedure.

This is a correctness bug in the assembly/linker rules, not a security boundary break. Calling a kernel export outside context 0 only runs that code in the caller's current context, or a new one for `dyncall`. Equivalent code could also live in a normal library.

A small repro is:

```rust
#[test]
fn regression_kernel_exports_are_syscall_only_exec_target_is_rejected() {
    let kernel_src = r#"
pub proc k1
    push.1
end
"#;
    let kernel = Assembler::default()
        .assemble_kernel(kernel_src)
        .expect("kernel assembly must succeed");

    let program_src = r#"
proc user
    exec.::$kernel::k1
end

begin
    call.user
end
"#;

    let res =
        Assembler::with_kernel(context.source_manager(), kernel).assemble_program(program_src);

    assert!(
        res.is_err(),
        "expected `exec.::$kernel::...` to be rejected in executable modules"
    );
}
```

The expected result is that assembly or linking rejects non-`syscall` references to exported kernel procedures.

The fix is to reject kernel procedures as targets of `exec`, `call`, `procref`, `dynexec`, and `dyncall`, and keep exported kernel procedures reachable only through `syscall`.

crazywriter1:
## Summary

Fixes #2902.

Exported kernel procedures were reachable via `exec`, `call`, and `procref` in addition to `syscall`. This PR adds a guard in `Linker::resolve_invoke_target` that rejects any non-`syscall` invocation resolving to a kernel procedure, while still allowing kernel-internal calls (e.g. a kernel proc calling another kernel proc via `exec`).

### Changes

- **`crates/assembly/src/linker/errors.rs`** — Added `KernelProcNotSyscall` error variant.
- **`crates/assembly/src/linker/mod.rs`** — After symbol resolution, check whether the resolved target belongs to the kernel module and the invoke kind is not `syscall`; if so, return the new error. Kernel-internal calls are excluded from the check.
- **`crates/assembly/src/tests.rs`** — Added regression test from the issue (`exec.::$kernel::k1` from a non-kernel module is rejected).
- **`CHANGELOG.md`** — Added entry under v0.23.0.

## Test plan

- [x] `cargo test -p miden-assembly` — 178 passed, 0 failed
- [x] `cargo clippy -p miden-assembly --all-targets --all-features -- -D warnings` — clean
- [x] `cargo fmt` — clean


crates/assembly/src/linker/mod.rs
```diff
@@ -610,7 +610,21 @@ impl Linker {
         target: &InvocationTarget,
     ) -> Result<SymbolResolution, LinkerError> {
         let resolver = SymbolResolver::new(self);
-        resolver.resolve_invoke_target(caller, target)
+        let resolution = resolver.resolve_invoke_target(caller, target)?;
+
+        if let SymbolResolution::Exact { gid, ref path } = resolution
```
huitseeker: I think this still leaves the dynamic path open. This guard only runs for targets that go through `resolve_invoke_target`, but `dynexec` and `dyncall` can still reach an exported kernel proc if the program pushes the proc digest directly instead of using `exec`/`call`/`procref`. I verified that a small `push.<k1_digest>; dynexec` and `push.<k1_digest>; dyncall` program still assembles with an attached kernel. Could we also reject kernel proc digests on the dynamic path so issue `#2902` is fully closed?

crates/assembly/src/tests.rs
```diff
@@ -5644,6 +5644,39 @@ end
     assert_diagnostic!(err, "invalid syscall");
 }
 
+#[test]
+fn regression_kernel_exports_are_syscall_only_exec_target_is_rejected() {
```
huitseeker: I think this regression wants to cover every non-`syscall` route in one place. The current test only checks `exec`, but the issue also called out `call`, `procref`, `dynexec`, and `dyncall`.

Something like this would make the fix complete:

```rust
#[test]
fn regression_kernel_exports_are_syscall_only_for_all_non_syscall_entrypoints() {
    let source_manager = context.source_manager();

    let kernel_src = r#"
pub proc k1
    push.1
end
"#;

    let kernel = Assembler::new(source_manager.clone())
        .assemble_kernel(kernel_src)
        .expect("kernel assembly must succeed");

    let digest = kernel
        .as_ref()
        .get_procedure_root_by_path("::$kernel::k1")
        .expect("kernel export must have a digest");
    let [a, b, c, d] = digest.map(|felt| felt.as_canonical_u64());

    let cases = vec![
        (
            "exec",
            "proc user\n    exec.::$kernel::k1\nend\n\nbegin\n    call.user\nend\n".to_string(),
        ),
        (
            "call",
            "proc user\n    call.::$kernel::k1\nend\n\nbegin\n    call.user\nend\n".to_string(),
        ),
        (
            "procref",
            "proc user\n    procref.::$kernel::k1\n    dropw\nend\n\nbegin\n    call.user\nend\n"
                .to_string(),
        ),
        (
            "dynexec",
            format!("begin\n    push.{d}.{c}.{b}.{a}\n    dynexec\nend\n"),
        ),
        (
            "dyncall",
            format!("begin\n    push.{d}.{c}.{b}.{a}\n    dyncall\nend\n"),
        ),
    ];

    for (kind, program_src) in cases {
        let err = Assembler::with_kernel(source_manager.clone(), kernel.clone())
            .assemble_program(program_src)
            .expect_err("kernel exports should be syscall-only");
        assert_diagnostic!(err, "syscall");
    }
}
```

I checked these sources locally against this branch: `exec`, `call`, and `procref` already fail, while `dynexec` and `dyncall` still assemble today, so this would catch the remaining gap directly.

crazywriter1:
Thanks for the review sir. @huitseeker 

1. **Dynamic path closed**: Added a best-effort static check in the instruction compiler that detects when the last pushed word in a basic block matches a kernel procedure digest before `dynexec`/`dyncall`, and rejects it. (Note: digests computed at runtime from memory or arithmetic cannot be caught at assembly time.)

2. **Comprehensive test**: Replaced the single `exec` test with the suggested test covering all five non-syscall routes (`exec`, `call`, `procref`, `dynexec`, `dyncall`). All cases pass.

crates/assembly/src/linker/mod.rs
bitwalker: There isn't any way for us to in general enforce that a kernel procedure (i.e. an export from the `$kernel` module) cannot be invoked dynamically. I would not complicate symbol resolution with attempting to catch those, and instead push that to the processor to validate.

Just as a trivial example of what I'm talking about:

```
begin
    push.SYSCALL_DIGEST
    mem_storew_le.ADDR
    push.ADDR
    exec.foo
end

proc foo
    dynexec
end
```

Attempting to recognize that any given `dynexec` will dynamically invoke a kernel procedure referenced by the address on the operand stack is unsolvable statically in general.

crazywriter1:
Fixed CI failures: replaced test-only `self.linker()` with a new `pub(crate) is_kernel_proc_digest()` helper on `Assembler`, resolving no-std and doc build errors. Also ran `cargo +nightly fmt`.

crazywriter1:
Pushed the latest changes:

- Removed the assembly-time dynexec/dyncall digest heuristic and related code (`KernelProcDynInvoke`, `ops()`, `is_kernel_proc_digest`) per @bitwalker's feedback.
- Kept the linker guard in `resolve_invoke_target` (`KernelProcNotSyscall`) for static invokes (`exec` / `call` / `procref`).
- Regression test now only covers those static paths.

@bitwalker @huitseeker Ready for another look when you have time.

crates/assembly/src/linker/mod.rs
huitseeker: I think this check is still too broad for digest targets. `get_procedure_index_by_digest()` returns an arbitrary match when more than one proc shares a MAST root, so this can reject a valid non-`syscall` call just because the first match happens to be in the kernel. 

Thinking of @bitwalker 's example in the other comment, I reproduced took a userspace proc that has _the same body) as `::$kernel::k1`: `call.<that digest>`, which now fails even though the userspace proc should keep that digest callable. 

It looks like this guard needs to skip raw digest targets, or only fire when the resolution is unique to the kernel.

huitseeker: That makes sense.

crates/assembly/src/linker/mod.rs
```diff
@@ -610,7 +610,21 @@ impl Linker {
...
+            && caller.kind.is_some()
+            && !caller.in_syscall()
+            && self.kernel_index.is_some_and(|k| k == gid.module && k != caller.module)
```
huitseeker: This only treats the root `$kernel` module as an internal caller. `link_kernel()` marks all AST modules in a kernel build as `ModuleKind::Kernel`, so a helper kernel module should still be able to `exec`/`call`/`procref` an exported root-kernel proc. 

I think the fix here is to make the exemption semantic instead of index-based, e.g. `let caller_is_kernel = self[caller.module].kind().is_kernel();` and then gate this error on `!caller_is_kernel` instead of `k != caller.module`.

crazywriter1: Thanks @huitseeker. I’ll address both points in the next push:

1. **Digest / `MastRoot` targets:** I’ll skip the `KernelProcNotSyscall` guard when the invoke is resolved from a raw digest, so valid userspace calls aren’t rejected when `get_procedure_index_by_digest()` picks an arbitrary match among procedures that share the same MAST root.

2. **Kernel-internal callers:** I’ll replace the root-`kernel_index` vs `caller.module` check with a semantic exemption: only emit the error when `!self[caller.module].kind().is_kernel()`, so helper kernel modules can still `exec`/`call`/`procref` a root-kernel export.

I’ll also resolve the `CHANGELOG.md` merge conflict on this branch and re-run `cargo +nightly fmt` / tests before requesting another review.

crates/assembly/src/tests.rs
```diff
@@ -5644,6 +5644,45 @@ end
     assert_diagnostic!(err, "invalid syscall");
 }
 
+#[test]
+fn regression_kernel_exports_are_syscall_only_for_all_non_syscall_entrypoints() {
```
huitseeker, line 5648: Please also leave a quick note on `#2902` so the issue can be assigned to you cleanly.

huitseeker (approved):
thanks! looks good.

crates/assembly/src/tests.rs
crazywriter1: Done. 👍

PR #2906, opened 2026-03-25
feat: implement new `miden-format` tool, based on a new lossless parser

### LINKED ISSUE #1901: Rewrite MASM parser to produce a concrete syntax tree
There are a few goals we have that the current parser gets in the way of:

1. It does not preserve trivia (whitespace, comments), which is critical for implementing a `miden-format` tool.
2. It performs constant propagation throughout parsing, losing important context needed for an LSP. This too prevents us from implementing a `miden-format` tool, because the resulting output would have all the constants erased
3. It provides poor error recovery at best. This is problematic for an LSP, and particularly integrating into text editors/IDEs, as the user experience is really bad.

To address these issues, the current parser needs to be rewritten to have the following characteristics:

* It produces a concrete syntax tree (CST), which can then be converted to an abstract syntax tree (AST). CSTs preserve all of the textual trivia and tokens, and furthermore, allow parts to be invalid/missing while still being able to continue parsing the rest of the tree. For example, if you start typing a new procedure definition in the middle of a file, e.g. `export.foo|` (where `|` here represents the cursor), the file would fail to parse today because the syntax is invalid on its own. With a CST-based parser however, the parser can easily recover by ignoring the invalid subtree, and continuing with the rest of the file. The mechanics of this are too much to get into here, but suffice to say that this is _the_ way to implement parsers that are designed to provide a smooth text editing/IDE experience.
* Constant propagation needs to be deferred to assembly-time. We don't actually need to do it during parsing, but it simplified some things, so that is what we did. However, by doing so, we've hamstrung our ability to reformat MASM and provide more useful diagnostics related to constants.
* We should move from using a parser generator (LALRPOP) to a handwritten recursive-descent parser. In order to properly implement error recovery over the CST, we need granular control over parsing that we just don't get with parser generators. This is actually a fairly trivial lift, as MASM has a very simple grammar, but the bigger task is ensuring that the new parser actually accepts partial syntax trees (i.e. it doesn't just bail at the first problem, it can parse even in situations like I described earlier, where a definition is only partially-complete).

There are some great tools for building parsers in this style, e.g. [rowan](https://crates.io/crates/rowan). `rust-analyzer` actually uses `rowan` internally (in addition to its `syntax` and `parser` crates to complete the picture).

This is a prerequisite to building a proper MASM formatter and LSP.

bitwalker:
This PR is the first of two (and a third when the legacy LALRPOP parser is removed), which implements a new lossless parser that produces a concrete syntax tree (CST) which preserves trivia and whitespace during parsing, allowing it to be used in interactive editing and formatting use cases.

The latter is prototyped in this PR in addition to the new parser, as a new binary crate `miden-format` which provides `rustfmt`-like functionality for MASM sources.

The [next PR](https://github.com/0xMiden/pull/2907) in this stack is the one which actually provides the new parser as an alternative backend for `miden-assembly-syntax`, deriving the AST from the CST and asserting the equivalence of the legacy and new parsers. It also cleans up the new parser, adds documentation, and addresses a number of small issues found during testing - however I'm opening them in the sequence they were developed for easier review, as it is easier to follow the progression commit-by-commit over the course of the two PRs, due to the size of the parsing code.

The new parser uses recursive descent, rather than a LALR table-driven parser (the one generated by LALRPOP). This has a number of benefits, but provides slightly less assurance that the grammar is unambiguous, as recursive descent parsers can very easily bake in parsing decisions in places where it is technically ambiguous, but the ambiguity is broken by some criteria internal to the parser.

I'll open the next PR in the stack immediately after this one (as they were developed together), and would recommend that we merge that one into this one, and then the combined branch into `next` once we have consensus.

---

## FG's ai-gen intro
https://gisthost.github.io/?e02b89cb340b7cc5b4d25cc157998a8d


bitwalker:
@bobbinth Not sure who to have review this and #2907 - I'd suggest whoever is going to be tackling MASM frontend issues, and so would benefit from a deeper review of the parser internals. Aside from myself, I think @huitseeker has been doing a lot recently on that front - but not sure if there is anyone else interested.

crates/miden-format/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "miden-format"
```
PhilippGackstatter, line 2: Thank you, it's great to have a masm formatter! I tested it in the protocol repo, and noted some thoughts, but may be best to run yourself. Don't want to derail the PR review with this though, so put the feedback in a spoiler.

<details>
<summary>
Formatter Feedback.
</summary>

I tested the formatter on `crates/miden-protocol/asm/kernels/transaction/lib/account.masm` in the protocol repo with:

```
cargo run --bin miden-format -- ../miden-protocol/crates/miden-protocol/asm/kernels/transaction/lib/account.masm
```

Executing this command twice results in a couple of `× syntax error` thrown on the second run.

**Constants**

```diff
-const ERR_ACCOUNT_STORAGE_SLOTS_MUST_BE_SORTED_AND_UNIQUE="slot IDs must be unique and sorted in ascending order"
+const ERR_ACCOUNT_UNKNOWN_STORAGE_SLOT_NAME =
+    "storage slot with the provided name does not exist"
```

```diff
-const ACCOUNT_VAULT_AFTER_REMOVE_ASSET_EVENT=event("miden::protocol::account::vault_after_remove_asset")
+const ACCOUNT_VAULT_AFTER_REMOVE_ASSET_EVENT =
+    event(
+        "miden::protocol::account::vault_after_remove_asset"
+    )
```

It'd be nice if the constants were not reformatted if the line does not exceed 100 characters. For the event, I'd slightly prefer this, but maybe others have other opinions:

```
const ACCOUNT_VAULT_AFTER_REMOVE_ASSET_EVENT = event(
        "miden::protocol::account::vault_after_remove_asset"
    )
```



**Stack Comments**

We currently have the pattern instructions followed by comments without a newline, and a newline after the stack comment, i.e.:

```
# load the old value of the slot
dup.4 exec.get_item_raw
# => [OLD_VALUE, VALUE, slot_ptr]

swapw movup.8
# => [slot_ptr, VALUE, OLD_VALUE]
```

the formatter moves the comments above the instructions:

```
# load the old value of the slot
dup.4 exec.get_item_raw

# => [OLD_VALUE, VALUE, slot_ptr]
swapw movup.8

# => [slot_ptr, VALUE, OLD_VALUE]
```

Ideally, this would be preserved.

**Plain Comments**

This is another odd one:
```diff
-pub proc get_native_storage_slot_type
-    # convert the index into a memory offset
+pub proc get_native_storage_slot_type # convert the index into a memory offset
```

</details>

bitwalker: You may want to run the formatter from #2907, as it incorporates a number of fixes/improvements to the parser I found while porting `miden-assembly-syntax` to use it - I'll re-run your tests myself too though and make sure there are no issues with idempotency.

Definitely agree with the feedback. The tricky aspect with comments is that comments must be anchored to something, and you can see why the things you pointed out with them crop up:

* In the first case, the comment is being considered anchored to the next node in the input, and it is moved so the whitespace between the comment and its anchor is eliminated, while preserving the newline separation between the previous node and the anchor node.
* In the second case, the comment immediately follows the procedure signature, so it is being attached to that, and collapsed to a single line because it fits.

Both of these are fixable though. Comments are always going to be tricky, because unless they are in an "obvious" place, the node they get attached to might be surprising. I expect this is really only an issue with trailing stack comments, specifically because that goes against typical comment conventions (trailing on the same line is one thing, but attaching a comment that is on the next line to the node before it is unusual from a parsing point of view).

I'll fix these in #2907 though, not here.

bitwalker: I've addressed all of these issues, and expanded on the current `miden-format` behavior/choices in its README in #2907. 

The one exception to that is the wrapping of constant values on to the next line, e.g.:

```
const FOO =
    event("foo")
```

Instead of:

```
const FOO = event(
    "foo"
)
```

Going through the diff on formatted files, I'm finding it a lot easier to read the former than the latter, and the latter would require special rules for specific constant expressions (e.g. `event(...)`) that should start on the same line, versus move to the next.

I'll leave that one up for broader discussion amongst the team and those writing MASM - but I think the current behavior might be preferable here personally.

crates/assembly-syntax-cst/src/lexer.rs
```diff
@@ -0,0 +1,388 @@
...
+                };
+                self.advance_while(|ch| ch != '\n' && ch != '\r');
+                self.token(kind, start, self.offset)
+            },
+            _ => unreachable!("unexpected trivia character: {first:?}"),
+        }
+    }
+
+    fn lex_identifier(&mut self, start: usize, first: char) -> Token<'input> {
+        if first == '$' {
```
huitseeker: The CST lexer is tokenizing a wider language than the assembler accepts. I was able to run `miden-format` successfully on `push.$foo`, and `0xZZ` gets split into `0x` plus `ZZ` instead of producing a literal error. The existing lexer only accepts `$kernel` and `$exec` here and validates numeric literals. Should this branch keep those diagnostics so the formatter doesn't accept invalid MASM as valid input?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+
+        Parse {
+            source,
+            green_node: self.builder.finish(),
+            diagnostics: self.diagnostics,
+        }
+    }
+
+    fn parse_source_item(&mut self) {
+        if self.at_kind(SyntaxKind::DocComment) {
```
huitseeker, line 221: The current CST parser accepts any `#!` line as a doc node and never checks whether it still belongs to the next top-level form. e.g. I was able to run `miden-format` successfully on `#! orphan` at EOF and on a docstring separated from the next item by a blank line, while the existing syntax tests still flag those as `unused docstring`. 

Should this path keep emitting that diagnostic so the formatter doesn't start treating invalid modules as valid input?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+        if self.at_name_like() || self.at_keyword_like() {
+            self.bump();
+        } else {
+            self.error_here("expected an import path");
+            self.finish_node();
+            return;
+        }
+
+        loop {
+            self.bump_non_comment_trivia();
```
huitseeker: I think `bump_non_comment_trivia()` will eat the newline after the last path segment before we know whether another `::` follows. 

Can we avoid consuming newlines here until we've confirmed the path really continues?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+        self.parens == 0 && self.brackets == 0 && self.braces == 0
+    }
+
+    fn bump(self, kind: SyntaxKind) -> Self {
+        let mut next = self;
+        match kind {
+            SyntaxKind::LParen => next.parens += 1,
+            SyntaxKind::LBracket => next.brackets += 1,
+            SyntaxKind::LBrace => next.braces += 1,
+            SyntaxKind::RParen => next.parens = next.parens.saturating_sub(1),
```
huitseeker: IIUC, wsing `saturating_sub` here means stray closing delimiters never become syntax errors. Both `const X = )` and `begin\n    foo ]\nend\n` format successfully right now instead of being rejected.

crates/miden-format/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+        return Ok(());
+    }
+
+    for (source, formatted) in formatted_inputs {
+        if source.as_str() == formatted {
+            continue;
+        }
+
+        let path = Path::new(source.uri().path());
```
huitseeker: `collect_inputs` starts from real filesystem paths, but writeback rebuilds the target from `source.uri().path()`. 

This is untested but with `miden_debug_types::Uri`, a Windows path like `C:\foo\bar.masm` is treated as scheme `C` plus path `\foo\bar.masm`, so could this drops the drive and point at the wrong file?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+    }
+
+    fn parse_block(&mut self, terminators: &[&str]) {
+        self.start_node(SyntaxKind::Block);
+        while !self.eof() {
+            if self.at_terminator(terminators) {
+                break;
+            }
+
+            if self.at_kind(SyntaxKind::DocComment) || self.at_regular_trivia() {
```
huitseeker: In blocks we skip `DocComment` the same way as normal trivia. That lets invalid input like:

```masm
pub proc foo
    #! malformed doc
    loc_load.0
end
```

parse and format successfully, even though the existing parser tests reject block-level doc comments. 

Shouldn't this branch emit an error node here instead of accepting the token?

crates/miden-format/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        if mismatches.is_empty() {
+            return Ok(());
+        }
+
+        for path in &mismatches {
+            eprintln!("would reformat {path}");
+        }
+
+        return Err(CliError::CheckFailed(mismatches.iter().map(|uri| uri.to_string()).fold(
```
huitseeker: Would `mismatches.iter().map(ToString::to_string).collect::<Vec<_>>().join("\n")` read a bit simpler?

crates/miden-format/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+        let source = source_manager.load(SourceLanguage::Masm, Uri::from(path.as_path()), source);
+        inputs.push(source);
+        return Ok(inputs);
+    }
+
+    if cli.paths.is_empty() {
+        return Err(CliError::MissingInput);
+    }
+
+    inputs.reserve_exact(cli.paths.len());
```
huitseeker: This looks like `cli.paths.iter().map(|path| source_manager.load_file(path)).collect()`. That would let the iterator carry the `Result<Vec<_>, _>` without the manual `reserve_exact` + `push` loop.

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+        }
+    }
+
+    pub fn span_for_range(&self, range: TextRange) -> SourceSpan {
+        source_span_from_text_range(self.source.id(), range)
+    }
+}
+
+pub fn parse_source_file(source: Arc<SourceFile>) -> Parse {
+    let parser_source = Arc::clone(&source);
```
huitseeker, line 96: A small simplification: could this just be `Parser::new(source.as_ref()).parse(source)`?

huitseeker (commented):
Thanks!

crates/assembly-syntax-cst/src/parser.rs
bitwalker: Just to confirm, was that off 2907, or this branch? I split them up to keep the work streams a bit separate, but planned to merge 2907 into this branch and then merge them together into next, since various issues were fixed in 2907 that aren't reflected here.

That's not to say this issue isn't present in 2907, but wanted to check which branch you were testing from

huitseeker: No, I was testing the present branch 🤔

bitwalker: I'll take a look and see which of those fixes I can slice up and move back to this PR. If you just want to see what the current final state of the formatter is, 2907 is the branch to build from, but if I can migrate the fixes here, then that would simplify review a bit.

crates/assembly-syntax-cst/src/parser.rs
bitwalker: IIRC, the borrow checker complained, but I'll confirm

crates/miden-format/src/formatter.rs
```diff
@@ -0,0 +1,1778 @@
...
+    }
+    rendered.push('\n');
+    rendered.push_str(&format!("{}end", indent_string(indent)));
+    rendered
+}
+
+fn render_procedure(procedure: &Procedure, indent: usize) -> String {
+    let mut lines = Vec::new();
+
+    for attribute in procedure.attributes() {
```
huitseeker: The formatter drops standalone comments between procedure attributes. For example, formatting `@inline\n# keep me\n@locals(1)\npub proc foo\n    nop\nend\n` prints the two attributes and the proc, but loses `# keep me`.

I think the cause is that this loop renders only the attribute nodes and does not look at the layout tokens between them. Could this path keep the inter-attribute comments too, or add a test for that case?

crates/assembly-syntax-cst/src/parser.rs
huitseeker: The path parsing concern is addressed. Current #2907 checks `peek_after_non_comment_trivia()` before consuming trivia in the path continuation loop.

crates/assembly-syntax-cst/src/parser.rs
huitseeker: The exact example in the comment is now fixed: `#! malformed doc` before `loc_load.0` errors on #2907. But the broader issue is not fully gone:` #! malformed doc` before `if.true` still formats successfully.

crates/assembly-syntax-cst/src/parser.rs
huitseeker: Ah, right, because it tries to move source while the parser still borrows from it. never mind.

crates/miden-format/src/formatter.rs
bitwalker: Nice catch! I'll add a test for that and make sure we handle those properly

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+    }
+
+    fn at_name_like(&self) -> bool {
+        matches!(
+            self.current_kind(),
+            Some(SyntaxKind::Ident | SyntaxKind::SpecialIdent | SyntaxKind::QuotedIdent)
+        )
+    }
+
+    fn at_keyword_like(&self) -> bool {
```
huitseeker: Could `at_keyword_like()` be removed here? It only accepts `Ident`, which is already included in `at_name_like()`, so at the numerous callsites (which are nearly all like `self.at_name_like() || self.at_keyword_like()`) this check reads like two cases even though the second branch is covered by the first.

huitseeker (commented):
In general, the formatter mixes a *lot* (comment layout, token spacing, wrapping, and final text rendering) in one large module.

I'm not sure how important its maintainability is here, in light of #2907. But if I had to insist, I'd recommend comment layout be pulled into a small internal shape first, so tests can check where comments attach without asserting the full formatted output.

crates/miden-format/Cargo.toml
PhilippGackstatter: Thanks for addressing the feedback! I ran the formatter from https://github.com/0xMiden/miden-vm/pull/2907 this time around and didn't get syntax errors on repeated runs.

**Max Line Width**

I didn't realize the `MAX_LINE_WIDTH` was set to 80. If set to 100, the formatter produces the results I would've intuitively expected. Maybe we can make this configurable or up it to 100 (this is what we have used so far for as a comment wrapping guideline)?

Otherwise, I'd still slightly prefer if the `event(` started on the same line, just so each line isn't converted to 4 lines (but only 3), e.g. currently we get:

```diff
-const ACCOUNT_VAULT_AFTER_REMOVE_ASSET_EVENT=event("miden::protocol::account::vault_after_remove_asset")
+const ACCOUNT_VAULT_AFTER_REMOVE_ASSET_EVENT =
+    event(
+        "miden::protocol::account::vault_after_remove_asset"
+    )
```

But not super important.

**Comments**

We also have these section dividers:

```
 ### STORAGE COMMITMENT ###################################################

 #! Returns the storage commitment of the active account at the beginning of the transaction.
 #!
```

where the formatter removes the empty line in between - ideally this would be unchanged / left to the authors.

**Structs**

```diff
-pub type AccountId = struct { prefix: felt, suffix: felt }
+pub type AccountId = struct {prefix : felt, suffix : felt}
```

Is this intentional? I find the original more readable due to extra spacing.

There is also an extra space after the field name, which is unusual.

**Procedure Parameters**

```diff
-pub proc tx_prepare_fpi(foreign_account_id: AccountId, foreign_proc_root: word, foreign_procedure_input_15: felt)
-    # validate the provided foreign account ID
+pub proc tx_prepare_fpi(
+    foreign_account_id: AccountId,
+    foreign_proc_root: word,
+    foreign_procedure_input_15: felt
+) # validate the provided foreign account ID
```

The "validate ..." comment should probably stay on an extra line, wdyt? Not sure how easy that is, though.

**Comment Removal**

```diff
-@locals(6) # foreign proc root (4) + foreign account ID (2)
+@locals(6)
```

Here it actually removed a comment.

**Path Imports**

```diff
-use miden::protocol::kernel_proc_offsets::TX_UPDATE_EXPIRATION_BLOCK_DELTA_OFFSET
+use miden::protocol::kernel_proc_offsets
+    ::TX_UPDATE_EXPIRATION_BLOCK_DELTA_OFFSET
```

Is this valid syntax? But it's interesting nonetheless because the line didn't exceed 100 characters before.

**Directories**

Small UX thing, but could we add support for running the formatter on a directory, which formats all MASM files it can find recursively, so we can easily do this to format all files in the `lib` directory?

```
cargo run --bin miden-format -- ../miden-protocol/crates/miden-protocol/asm/kernels/transaction/lib/
```

Other than those small issues, the formatter works well on the tx kernel files. Thank you for writing it!

bitwalker: @PhilippGackstatter Thanks for the report!

I updated 2607 already to handle two of your requests: the default line length, configurable formatter options, recursive formatting, and dropped comments around attributes:

* I agree on max line length of 100, so I've made that adjustment to the default config
* You can configure the formatter via a `miden-format.toml` file, or via the command-line, e.g. `miden-format --config max_line_length=100`. The contents of `miden-format.toml` for that would look like:
```toml
max_line_length=100
```
* Any paths you pass to `miden-format` which are directory paths, will be recursively traversed for `.masm` files, all of which will be formatted. An error will be produced if we hit paths that we cannot access.

As for the other items:

> I'd still slightly prefer if the event( started on the same line, just so each line isn't converted to 4 lines (but only 3)

This is the sort of thing that I'd probably want to make configurable, rather than the default - I know my own preference is to move the value to the next line, indented, so it makes scanning visually/grepping for things easier in my experience. Obviously that's subjective though, since it depends entirely on what people are optimizing for when they make formatting decisions.

> section dividers

Yeah I would want these preserved as well, so I've noted that as something to fix.

> struct field spacing

100% unintended - I prefer padding with spaces as you've described, and would want our formatter to do the same.

> The "validate ..." comment should probably stay on an extra line, wdyt? Not sure how easy that is, though.

Agreed, that's an easy one to fix

> path imports

It's accepted because the parser largely ignores whitespace, but It's not proper formatting behavior IMO, so I'm treating this as a bug.

I think once those are addressed, I'd prefer to address all other improvements in follow-on PRs, so as to avoid this becoming any larger than it already is. We can treat it as experimental until we're able to use it on the core/protocol libs without any undesired formatting behavior.

crates/assembly-syntax-cst/src/parser.rs
bitwalker: I've addressed this and added a test for stray delimiters

crates/miden-format/src/main.rs
bitwalker: I have no idea what the state of our Windows support is, it's probably busted in general, but regardless this is a good question.

`Uri` is intended to represent valid URIs, which do not allow `\` as a component separator. However, we do construct `Uri` from file paths in numerous places, and on Windows this will produce invalid URIs. I've added a helper to `Uri` called `to_path` which takes the raw content of the URI (sans scheme, if present), and feeds it to `PathBuf::from` - this will at least fix the issue pointed out here. However, we have a larger issue of what to do when constructing a `Uri` from paths everywhere else - I'd want to address that in a separate PR, as it's a bigger change, and unrelated to this PR.

crates/miden-format/src/formatter.rs
bitwalker: This has been addressed, with a regression test

crates/assembly-syntax-cst/src/parser.rs
bitwalker: Yep - we can always re-add it later if the distinction matters at some point

bitwalker:
> In general, the formatter mixes a _lot_ (comment layout, token spacing, wrapping, and final text rendering) in one large module.
> 
> I'm not sure how important its maintainability is here, in light of #2907. But if I had to insist, I'd recommend comment layout be pulled into a small internal shape first, so tests can check where comments attach without asserting the full formatted output.

I'll address this in 2907, since it would be a mess to rebase everything. Ultimately I'm planning on us merging 2907 into this branch and then merging both PRs together as a unit into `next`, rather than merging this PR, and then 2907.

crates/assembly-syntax-cst/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+default = ["std"]
+std = ["miden-debug-types/std", "miden-utils-diagnostics/std", "rowan/std", "thiserror/std"]
+
+[dependencies]
+miden-debug-types.workspace = true
+miden-utils-diagnostics.workspace = true
+thiserror.workspace = true
+rowan = { package = "miden-rowan", version = "0.16", default-features = false }
+
+[build-dependencies]
```
huitseeker: This looks like the reason 2906 cargo-shear is red. `build.rs` only reads the `.ungram` file as text and writes it into `OUT_DIR`; it never uses the `ungrammar` crate, so shear reports `unused dependency ungrammar`. Removing this dependency here would let the lower PR pass on its own.

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+
+    fn parse_signature(&mut self) {
+        self.start_node(SyntaxKind::Signature);
+        self.parse_balanced_group(
+            SyntaxKind::LParen,
+            SyntaxKind::RParen,
+            "expected `)` to close procedure parameters",
+        );
+
+        self.bump_non_comment_trivia();
```
huitseeker: This skips newlines after `)` before knowing whether a `->` exists. For a procedure with no result type, a body-leading comment after the signature can get captured inside the signature and then be formatted onto the `proc` header instead of staying in the body.

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+            } else {
+                self.error_here("expected a path segment after `::`");
+                break;
+            }
+        }
+
+        self.finish_node();
+    }
+
+    fn parse_expr_until_line_end(&mut self) {
```
huitseeker, line 429: A comment right after `=` in a multiline `const` or `adv_map` is outside the expression today, so the formatter can treat it as declaration trivia and move it onto the declaration line instead of keeping it with the value.

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+            | SyntaxKind::ColonColon
+            | SyntaxKind::RArrow
+            | SyntaxKind::Plus
+            | SyntaxKind::Minus
+            | SyntaxKind::Star
+            | SyntaxKind::Slash
+            | SyntaxKind::SlashSlash
+            | SyntaxKind::RBracket
+            | SyntaxKind::RParen
+            | SyntaxKind::RBrace
```
plafer, line 1238: What is the reason to have those here? These seem to unintentional get swallowed into instructions, where e.g. in `begin add ) end`, the `)` gets swallowed in the `add` instruction. I'm still wrapping my head around the code, and I see why you'd want e.g. a `Dot` here (for `push.1.2`), but I would expect the parens/brackets to be related to nesting instead (and hence handled elsewhere).

A failing test:

```rs
#[test]
fn does_not_absorb_stray_closing_paren_into_instruction() {
    let parse = parse_text("begin\n    add )\nend\n");

    let instruction = parse
        .syntax()
        .descendants()
        .find(|node| node.kind() == SyntaxKind::Instruction)
        .expect("expected an instruction node");

    let significant: Vec<_> = instruction
        .descendants_with_tokens()
        .filter_map(|element| element.into_token())
        .filter(|token| !token.kind().is_trivia())
        .map(|token| (token.kind(), token.text().to_string()))
        .collect();

    assert_eq!(
        significant,
        vec![(SyntaxKind::Ident, "add".to_string())],
        "stray `)` was absorbed into the `add` instruction: {significant:?}"
    );
}
```

yields

```
assertion `left == right` failed: stray `)` was absorbed into the `add` instruction: [(Ident, "add"), (RParen, ")")]
  left: [(Ident, "add"), (RParen, ")")]
 right: [(Ident, "add")]
```

crates/miden-format/src/main.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        if mismatches.is_empty() {
+            return Ok(());
+        }
+
+        for path in &mismatches {
+            eprintln!("would reformat {path}");
+        }
+
```
plafer: Why do we output this, as the same information gets re-output as part of `CliError::CheckFailed` (also on stderr)? e.g.

```
would reformat /my/file.masm
the following inputs are not formatted:

/my/file.masm
```

partylikeits1983 (commented):
Question, would it make sense to have a test directory for the parser which would contain samples of unformatted and the resulting correctly formatted MASM? Then have tests which show that the parser works correctly?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -0,0 +1,1318 @@
...
+        source_span_from_text_range(self.source.id(), range)
+    }
+}
+
+pub fn parse_source_file(source: Arc<SourceFile>) -> Parse {
+    let parser_source = Arc::clone(&source);
+    Parser::new(parser_source.as_ref()).parse(source)
+}
+
+pub fn parse_text(input: &str) -> Parse {
```
plafer, line 104: Would be good to have high level tests

1. **round trip:** Make sure that `assert_eq!(parse_text(input).syntax().text().to_string(), input);`
2. **idempotency:** `format(format(input)) == format(input)`

for a variety of `input` (perhaps fuzzed). Probably in a subsequent PR?

crates/assembly-syntax-cst/src/lexer.rs
```diff
@@ -0,0 +1,388 @@
...
+                        self.advance_char();
+                    }
+                },
+                '"' => {
+                    self.advance_char();
+                    closed = true;
+                    break;
+                },
+                _ => {
+                    identifier_like &= ch.is_ascii_alphanumeric() || ch.is_ascii_graphic();
```
plafer: `is_ascii_alphanumeric()` is a subset of `is_ascii_graphic()`, so this is equivalent to just `is_ascii_graphic()`.

```suggestion
                    identifier_like &= ch.is_ascii_graphic();
```

crates/miden-format/src/formatter.rs
```diff
@@ -0,0 +1,1778 @@
...
+    Item, Operation, SyntaxKind, SyntaxNode, SyntaxToken,
+    ast::{
+        BeginBlock, Block, IfOp, Import, Instruction, Procedure, RepeatOp, Signature, SourceFile,
+        TypeBody, TypeDecl, WhileOp,
+    },
+    rowan::{NodeOrToken, ast::AstNode},
+};
+
+const INDENT_WIDTH: usize = 4;
+const MAX_LINE_WIDTH: usize = 80;
```
plafer: Future improvement: we should make this configurable

crates/miden-format/src/formatter.rs
```diff
@@ -0,0 +1,1778 @@
...
+    if matches!(next_kind, Equal | RArrow | Colon | Plus | Minus | Star | Slash | SlashSlash) {
+        return true;
+    }
+
+    if next_kind == LBrace {
+        return true;
+    }
+
+    true
+}
```
plafer: We currently default to `true` here - ideally though we would use an explicit `match` such that if we add new token kinds, we get a compilation error here and are reminded to update this function

crates/assembly-syntax-cst/src/parser.rs
bitwalker: It's to recognize things as part of a preceding instruction keyword, such as `push.[1, 2, 3, 4]` or `emit.event("...")` - the punctuation isn't trivia, it belongs to the instruction node.

You are correct that `add )` would get collapsed together in the case where there is a missing matching `(` preceding it - but that is actually fine, and would be caught when lowering to the AST. The benefit in this case would be that we consume this invalid token, and can continue parsing the rest of the source file just fine - rather than having the parser be unable to proceed after hitting `)`.

The key use case for error recovery/resiliency like this is for use in IDE extensions/LSP, where you are going to be getting fed source code as it is being edited. For example, you don't generally want your LSP to stop working the second you type the following in a larger MASM file:

```
...snip..

pub proc incomplete(in: u32
                          ^ cursor is here

pub proc complete(in: u32) -> u32
    ..snip..
end

..snip..
```

Here, you want to recognize that the syntax node representing the procedure definition is incomplete, close out that node incomplete, and proceed to parsing the next syntax node as if the incomplete procedure was actually valid. Thus your CST still has a structure that largely maps to what your AST requires, and errors are contained to an appropriate ancestor node.

It is still possible for the input to be wonky enough that you can't really do much after things go sideways, but that isn't typically the case - most of the time, errors are relatively self-contained to what the user is writing at any given moment.

crates/miden-format/src/main.rs
bitwalker: Whoops, forgot to remove that

crates/miden-format/src/formatter.rs
bitwalker: It is, see #2907

crates/assembly-syntax-cst/src/parser.rs
bitwalker: I'm not sure that's actually an issue? Consider:

```
const FOO = # comment
    42
```

If the comment gets formatted as part of the declaration:

```
# comment
const FOO = 42
```

That's fine? As a reader I'd assume that the comment is for the declaration anyway, as it is on the same line.

The alternatives are are strictly worse IMO - you'd either be unable to reformat this (because the leading comment of the value prevents the value from being placed on the same line as the declaration), or you have to infer that leading trivia should be merged with trailing trivia in specific cases, such as when reformatting this declaration to be on a single line - but that gets messy, e.g. how do you handle this sanely:

```
const FOO = # comment
    42              # other comment
```

I haven't tested to see what the current output is, but I believe it would be:

```
# comment
const FOO = 42 # other comment
```

Which is at least sensible behavior - though it would depend on the connection between the two comments.

bitwalker:
> Question, would it make sense to have a test directory for the parser which would contain samples of unformatted and the resulting correctly formatted MASM? Then have tests which show that the parser works correctly?

@partylikeits1983 My plan is to apply the formatter to the MASM sources in this repo in a separate PR, and once everyone is satisfied with whatever changes that produces, we'd add a test which runs the formatter on all the in-tree sources and asserts that there are no formatting mismatches. That will be a much better corpus, and it will also act as a check on adding MASM sources that are unformatted.

crates/miden-format/src/main.rs
bitwalker: Fixed in #2907

crates/assembly-syntax-cst/src/lexer.rs
bitwalker: Fixed in #2907

crates/miden-format/src/formatter.rs
bitwalker: Addressed in #2907

crates/assembly-syntax-cst/src/parser.rs
bitwalker: I can either add these tests in #2907, or create an issue as a follow-up. What's your preference?

crates/assembly-syntax-cst/src/parser.rs
bitwalker: So this is true when parsing the CST, but the formatter handles this currently, so the issue doesn't actually occur in 2907. I'm going to modify this though so that the CST is more accurate.

crates/assembly-syntax-cst/src/parser.rs
plafer: Makes sense, thanks for the explanation!

crates/assembly-syntax-cst/src/parser.rs
plafer: Up to you - either way, in #2907 is fine as long as you have the capacity/it doesn't hold up merging the PRs.

plafer (approved):
Didn't have time to go through every detail, but overall LGTM (taking into account the fixes that went into #2907)

crates/assembly-syntax-cst/src/parser.rs
bitwalker: Resolved in 2907, which is now merged into this PR

crates/assembly-syntax/Cargo.toml
```diff
@@ -40,6 +41,7 @@ logging = ["dep:env_logger"]
 
 [dependencies]
 # Miden dependencies
+miden-assembly-syntax-cst = { workspace = true, default-features = false }
```
huitseeker, line 44: Does this keep the crate's `no_std` promise? 
`miden-assembly-syntax` is still `#![no_std]` and advertises the `no-std` category, but this new unconditional dependency is not `#![no_std]` and imports `std` in normal code. 

Host `--no-default-features` checks can still pass, but no-std consumers would now pull in a std-only crate.

crates/debug-types/src/lib.rs
```diff
@@ -116,6 +118,17 @@ impl Uri {
             None => path,
         }
     }
+
+    /// Convert this URI to a [std::path::PathBuf], if it represents a file path
+    #[cfg(feature = "std")]
+    pub fn to_path(&self) -> Option<std::path::PathBuf> {
+        let uri = self.0.as_ref();
+        match uri.split_once("//") {
```
huitseeker, line 126: I think this check is off for normal file URIs. `file:///tmp/foo.masm` splits as `("file:", "/tmp/foo.masm")`, so this returns `None`, while `memory:foo.masm` hits the `None` arm and becomes a local path. Should this use `scheme()`/`path()` and only allow no-scheme paths plus `file`?

crates/miden-format/src/formatter.rs
```diff
@@ -0,0 +1,2330 @@
...
+
+    let previous_kind = previous.kind();
+    let next_kind = next.kind();
+
+    if matches!(previous_kind, At | Dot | ColonColon | LAngle | LParen | LBracket | LBrace) {
+        return false;
+    }
+
+    match next_kind {
+        Dot | ColonColon | DotDot | Comma | LAngle | RAngle | LParen | RParen | RBracket
```
huitseeker: Could this keep a space before parenthesized values? Right now `const X = (1)` formats as `const X =(1)` because every `LParen` suppresses the space before it. Maybe this case should only be compact for call-like syntax.

crates/miden-format/src/formatter.rs
```diff
@@ -0,0 +1,2330 @@
...
+    config: &Config,
+) -> Vec<String> {
+    if prefer_multiline_groups {
+        if let Some((open_index, close_index)) =
+            outer_group_indices(tokens, SyntaxKind::LBracket, SyntaxKind::RBracket)
+        {
+            return render_delimited_group(tokens, indent, open_index, close_index, style, config);
+        }
+
+        if let Some((open_index, close_index)) =
```
huitseeker, line 928: I think this skips wrapping for values that start with `(`. With `--config max_line_length=40`, a long `const X = (...)` becomes `const X =` and then one overlong parenthesized line. Should `open_index == 0` still use the delimited wrapping path for value expressions?

crates/miden-format/README.md
```diff
@@ -0,0 +1,272 @@
...
+- put it on the same line as that construct, or
+- put it immediately after that construct with no intervening blank line
+
+If you want a comment to be visually associated with the following construct:
+
+- separate it from the previous construct with a blank line
+
+## Hypothetical Future Configuration
+
+`miden-format` does not currently read a configuration file. If we later introduce something like
```
huitseeker: This looks stale now that the CLI loads `miden-format.toml` from the cwd and also has `--config`. The option table below also says `max_width`/`indent_width`, but the implemented names are `max_line_length`/`indent_size`.

crates/assembly-syntax-cst/src/parser.rs
huitseeker: Fair enough!

crates/debug-types/src/lib.rs
bitwalker: `memory:foo.masm` isn't a valid URI, the scheme must always be followed by `://`

bitwalker:
@huitseeker Give it another round - let me know if you have any final comments

crates/assembly-syntax/src/parser/cst/blocks.rs
```diff
@@ -0,0 +1,289 @@
...
+    } else {
+        return Err(ParsingError::InvalidSyntax {
+            span: context.parse().span_for_node(then_node.syntax()),
+            message: "expected a non-empty `if` block".to_string(),
+        });
+    };
+
+    let else_blk = match else_node {
+        Some(else_node) => {
+            if !else_has_ops {
```
huitseeker, line 104: This rejects `if.true <body> else end`, but the legacy grammar accepted that form by lowering the empty `else` side to a `nop` block. Since this backend is now the default parser, existing MASM with an empty else branch would stop parsing. Could this match the legacy behavior and add a regression test for `if.true add else end`?

crates/debug-types/src/lib.rs
huitseeker: My point exactly:  as written, `file:///tmp/a.masm` has the prefix `file:` here and returns `None`, while `memory:foo.masm` has no `//` and is treated as a local path.

bitwalker: That's intentional - if an explicit scheme prefix (i.e. `<scheme>://`) is not provided, we currently treat the input as having `file://` scheme.

This behavior is (currently) intentional for two reasons: 1. It provides direct compatibility with file paths for our current use cases, while supporting other URI schemes as-needed. 2. We can't meaningfully validate file paths in no-std builds, because we can't know what constitutes a valid file path. So unless an explicit scheme is provided, we assume that the caller is encoding paths in URI-safe form.

I'd ultimately like to delegate this stuff to some external library, which better handles all the details around URIs, provides compatibility with `std::path::Path`, etc. - but that's an issue separate from this PR IMO.

crates/assembly-syntax/README.md
```diff
@@ -2,5 +2,60 @@
 
 This crate provides parsing and semantic analysis of the Miden Assembly language.
 
+## Developer guide
+
+The following sections will guide you through how to make some of the more
```
huitseeker, line 7: This line has trailing whitespace, and `git diff --check origin/next...HEAD` fails here. There is another one in `CHANGELOG.md:127`.

huitseeker (approved):
LGTM!

PR #2907, opened 2026-03-25
feat: switch assembly-syntax parsing to new lossless parser

### LINKED ISSUE #1901: Rewrite MASM parser to produce a concrete syntax tree

bitwalker:
This PR builds on top of #2906, implementing the lowering from the CST to the AST defined in `miden-assembly-syntax`, and asserting the equivalence between the legacy parser and the new one.

Once ready, this should be merged into #2906 and then the combined PR merged into `next`.


crates/assembly-syntax/src/parser/cst/instructions.rs
```diff
@@ -0,0 +1,1850 @@
...
+fn lower_exp_family(
+    context: &mut LoweringContext<'_>,
+    span: SourceSpan,
+    token: &SyntaxToken,
+) -> Result<Option<Vec<ast::Op>>, ParsingError> {
+    if token.kind() == SyntaxKind::Ident
+        && let Some(bits) = token.text().strip_prefix('u')
+        && let Some(bits) = parse_decimal_u64(bits)
+    {
+        let bits = u8::try_from(bits).expect("parsed decimal bit-size should fit in u8");
```
huitseeker: `exp.u256` looks like it will panic here instead of returning the same invalid bit-size diagnostic as `exp.u65`. `parse_decimal_u64("256")` succeeds, then `u8::try_from(bits).expect(...)` unwraps the failed conversion before the `bits < 64` check can reject it.

Could this check stay on the parsed integer first, then convert after the `< 64` branch? That should keep malformed source from crashing the parser.

crates/assembly-syntax/src/parser/cst/instructions.rs
```diff
@@ -0,0 +1,1850 @@
...
+        return Err(ParsingError::ImmediateOutOfRange { span: immediate_span, range: 1..17 });
+    }
+
+    Ok(Some(vec![inst_op(
+        instruction_span,
+        Instruction::AdvPush(Immediate::Value(Span::new(immediate_span, index))),
+    )]))
+}
+
+fn lower_dup(
```
huitseeker: Could the repeated stack-index lowerers below share a helper? They all parse the same immediate shape, range-check it, map an index to an instruction, and wrap it with `inst_op`, so keeping them separate makes these closely related diagnostics easier to drift.

huitseeker (commented):
Thanks, this looks good in general, but there's two areas which could be more effective:

- The instruction lowering module now holds many repeated rules: spelling lookup, immediate parsing, range checks, diagnostics, and AST op mapping. Could those instruction-family rules move behind a smaller helper or table shape, so adding one MASM instruction family does not require editing several separate matches?

- The fragment parser is useful, but its callers still choose how to collect tokens and when to check for trailing tokens. Could fragment lowering take CST nodes directly and own that setup itself? The hope is to keep token collection, span choice, and EOF checks in one place.

- related to this, could we add a short maintainer note for changing MASM instructions? The new CST parser and lowering path make instruction changes touch several places, so a guide for “add an instruction” and “deprecate an instruction” would help keep spelling, parsing, lowering, diagnostics, tests, and formatting in sync.

crates/assembly-syntax/src/parser/cst/instructions.rs
bitwalker: Addressed, along with a test

bitwalker:
> * The instruction lowering module now holds many repeated rules: spelling lookup, immediate parsing, range checks, diagnostics, and AST op mapping. Could those instruction-family rules move behind a smaller helper or table shape, so adding one MASM instruction family does not require editing several separate matches?

I'll take a crack at simplifying things and follow up.

> * The fragment parser is useful, but its callers still choose how to collect tokens and when to check for trailing tokens. Could fragment lowering take CST nodes directly and own that setup itself? The hope is to keep token collection, span choice, and EOF checks in one place.

I think I have an idea how we can incorporate more of that directly into `FragmentParser`, I'll get back to you on my findings.

> * related to this, could we add a short maintainer note for changing MASM instructions? The new CST parser and lowering path make instruction changes touch several places, so a guide for “add an instruction” and “deprecate an instruction” would help keep spelling, parsing, lowering, diagnostics, tests, and formatting in sync.

Certainly!

bitwalker:
@huitseeker I cleaned up some things based on your review:

> * The instruction lowering module now holds many repeated rules: spelling lookup, immediate parsing, range checks, diagnostics, and AST op mapping. Could those instruction-family rules move behind a smaller helper or table shape, so adding one MASM instruction family does not require editing several separate matches?

I've refactored things into a more table-driven approach, which reduces the duplication considerably. There is still some amount of duplication due to keeping parity with the legacy parser in specific ways (namely the implicit folding the legacy parser does). Once we merge this, we can remove the legacy parser and simplify things quite a bit further.

> * The fragment parser is useful, but its callers still choose how to collect tokens and when to check for trailing tokens. Could fragment lowering take CST nodes directly and own that setup itself? The hope is to keep token collection, span choice, and EOF checks in one place.

I've made a couple of changes that removes most of the boilerplate around fragment parsing, and moves some responsibilities to the `AstNode` impl, and boilerplate/fragment-specific details to `FragmentParser` (e.g. EOF checks). All that remains in the fragment parser functions which are free standing, are the specific details of how a fragment should be parsed, which is performed in a callback that takes the `FragmentParser` as an argument - this lets us know when the fragment is finished parsing for the EOF checks.

> * related to this, could we add a short maintainer note for changing MASM instructions? The new CST parser and lowering path make instruction changes touch several places, so a guide for “add an instruction” and “deprecate an instruction” would help keep spelling, parsing, lowering, diagnostics, tests, and formatting in sync.

I've add an initial guide to the `miden-assembly-syntax` README, but we'll likely want to expand on it a bit in various ways (i.e. what the syntax conventions/philosophy is, how to structure various types of forms), but the basics are covered now at least.

crates/miden-format/src/main.rs
```diff
@@ -34,6 +37,10 @@ struct Cli {
     #[arg(long, requires = "stdin")]
     stdin_filepath: Option<PathBuf>,
 
+    /// Set formatter options from the command line
+    #[arg(long)]
+    config: Config,
```
huitseeker: Right now clap treats `config: Config` as required, so `miden-format --stdin` and `miden-format path.masm` fail unless the user passes `--config ...`. 

I hit this locally: help now shows `Usage: miden-format [OPTIONS] --config <CONFIG> [PATH]...`. An `Option<Config>` or a default value would keep the default/file config path usable.

crates/assembly-syntax/src/parser/cst/forms.rs
```diff
@@ -0,0 +1,635 @@
...
+    Ok(forms)
+}
+
+/// Returns `Some(true/false)` when `items[index]` starts a doc-comment group.
+///
+/// The boolean indicates whether the group should become `Form::ModuleDoc` (`true`) or ordinary
+/// item docs (`false`). Non-doc items return `None`.
+fn doc_group_kind(context: &LoweringContext<'_>, items: &[CstItem], index: usize) -> Option<bool> {
+    let item = items.get(index)?;
+    matches!(item, CstItem::Doc(_)).then(|| index == 0 && starts_at_file_beginning(context, item))
```
huitseeker, line 86: The legacy lexer marks `#!` as module docs when it is on the first physical line, so a file that starts with spaces and then `#! docs` still became `Form::ModuleDoc`. Here the doc must start at byte 0, so leading spaces turn it into item docs and can attach it to the next form.

Could this check use the line position instead of byte offset?

crates/assembly-syntax/src/parser/cst/forms.rs
```diff
@@ -0,0 +1,635 @@
...
+fn lower_doc_group(
+    context: &LoweringContext<'_>,
+    items: &[CstItem],
+    is_module_doc: bool,
+) -> ast::Form {
+    let first_span = item_span(context, &items[0]);
+    let last_span = item_span(context, items.last().expect("non-empty doc group"));
+    let span = SourceSpan::new(context.source_file().id(), first_span.start()..last_span.end());
+
+    let mut text = String::new();
```
huitseeker, line 154: The lalrpop path rejected doc text over `u16::MAX` with `DocsTooLarge`, but this CST lowering builds the `String` and returns `Form` directly. That lets oversized docs through the default parser now.

Could we keep the old size check here?

crates/assembly-syntax/src/parser/cst/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+    interned: &mut BTreeSet<Arc<str>>,
+) -> Result<Vec<ast::Form>, Report> {
+    let mut parse = miden_assembly_syntax_cst::parse_source_file(source.clone());
+    let diagnostics = parse.take_diagnostics();
+    if diagnostics.is_empty() {
+        let mut context = LoweringContext::new(parse, interned);
+        lower_source_file(&mut context).map_err(move |err| err.with_source_code(source))
+    } else {
+        Err(Report::from(SyntaxError::from(diagnostics)).with_source_code(source))
+    }
```
plafer, line 77: We currently only output `Severity::Error` diagnostics, so this works. But whenever someone adds a `Severity::Warning` or `Advice` (e.g. a deprecation hint), this will fail. The fix is simply to ensure that there are no diagnostics with severity `Error`.

crates/assembly-syntax/src/parser/cst/instructions.rs
```diff
@@ -0,0 +1,2146 @@
...
+    }
+}
+
+/// Parses a decimal `u8` literal token without accepting non-decimal spellings.
+fn lower_decimal_u8_literal(token: &SyntaxToken) -> Result<Option<u8>, ParsingError> {
+    let Some(value) = lower_decimal_u64_literal(token)? else {
+        return Ok(None);
+    };
+    Ok(u8::try_from(value).ok())
+}
```
plafer, line 2158: For values that don't fit in a `u8`, this returns `Ok(None)`, falling back to the "legacy fallback diagnostic". I'm not completely clear on when we want that mechanism vs specific errors? For example, here I would have expected something like `u8::try_from(value).map_err(|_| ParsingError::ImmediateOutOfRange { ... }).map(Some)`, so that we return `ParsingError::ImmediateOutOfRange` if the literal doesn't fit in a `u8`.

crates/assembly-syntax/src/parser/cst/instructions.rs
```diff
@@ -0,0 +1,2146 @@
...
+            && consumed + used == rest.len()
+        {
+            return Ok(Some(vec![inst_op(
+                instruction_span,
+                Instruction::PushSlice(imm.with_span(imm_span), range),
+            )]));
+        }
+    }
+
+    lower_push_list(context, instruction_span, rest)
```
plafer: `parse_push_slice_range()` returns `Ok(None)` when `parse_decimal_u64()` returns `None` (e.g. we're dealing with `push.X[0xff..0xff]`, where `0xff` fails to parse as `u64`). AFAIK the assumption is that `Ok(None)` should bubble up such that we return a "legacy diagnostic" error. However, we `Ok(None)` falls through to `lower_push_list()` instead.

As a general pattern, instead of using `Ok(None)` to encode "legacy diagnostic error", wouldn't it better to return that legacy error at every site that it is generated directly? IMO, assuming a nice helper, it would make the code more readable/less complex and error prone than the current version. WDYT?

crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -949,10 +1167,36 @@ fn detached_source_file(input: &str) -> Arc<SourceFile> {
...
+    source
+        .as_str()
+        .char_indices()
+        .last()
+        .map(|(offset, _)| {
+            SourceSpan::at(
+                source.id(),
+                u32::try_from(offset).expect("source files larger than 4GiB are not supported"),
+            )
+        })
```
plafer, line 1180: Don't we expect the EOF to be the index 1 past the last (instead of the last index as it is here)? Didn't double-check with the rest of the code, but that's typically my understanding/assumption of what EOF would point to.

crates/assembly-syntax-cst/src/ast.rs
```diff
@@ -271,4 +448,5 @@ fn next_significant_token(node: &SyntaxNode, token: &SyntaxToken) -> Option<Synt
         }
     }
     None
+     */
```
plafer: This is commented-out code - let's remove instead

crates/assembly-syntax/src/parser/cst/mod.rs
bitwalker: This is a more general problem with how we emit diagnostics currently, assembler-wide. I have a plan to tackle this, using the same/similar approach used in the compiler, but it isn't something I want to handle here (i.e. currently, if we swallow diagnostics of any level other than `Error` here, or elsewhere in the assembler, they will never be seen).

The change we need to make more broadly in the assembler, is to have a diagnostics handler be provided as part of the context in which diagnostics can be raised (e.g. as part of the assembler/parser state), and determine top-level `Ok`/`Err` output based on whether any error diagnostics were collected/emitted. This is what the compiler does today, and I think the assembler would benefit from the same approach.

crates/assembly-syntax/src/parser/cst/instructions.rs
bitwalker: There is some code in here to exists solely to keep parity with the old parser, but which can be removed once the legacy parser goes away (in particular, the parts which match the implicit folding behavior of the legacy parser). I can create an issue to revisit the functions in this module after that, to make errors more precise.

Alternatively, we can modify this PR to just remove the legacy parser entirely, and make the changes to remove compatibility with how the legacy parser handled certain cases - but I think it would be best to do that as a follow-on PR personally.

crates/assembly-syntax/src/parser/cst/instructions.rs
bitwalker: I think that should be fine - the source spans would almost certainly be different, so I don't know if it would cause related legacy-compatibility tests to fail, but I can probably just remove those specific tests, since we'd be intentionally diverging in those cases.

crates/assembly-syntax-cst/src/parser.rs
bitwalker: No - if you create a span that is invalid like that (i.e. points to something out of bounds of the input), no span will be rendered at all - where this specifically points to the last valid character in the input, giving you a span that shows that line with the caret pointing to the last character.

I ran into this during testing, and so this is explicitly intended to provide useful reporting output for those diagnostics.

crates/assembly-syntax-cst/src/ast.rs
bitwalker: Ah, indeed, forgot to clean that up

crates/assembly-syntax/src/parser/cst/instructions.rs
plafer: Ah yes that makes sense. And I agree - let's create an issue to revisit this after the legacy parser is removed

crates/assembly-syntax/src/parser/cst/instructions.rs
plafer: I understand the context around compatibility tests now - we can either change that here or wait for the legacy parser to be removed to make larger changes. Up to you

plafer (approved):
Similar to #2906, overall LGTM (modulo resolution of remaining comments)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - Implemented new lossless parser for Miden Assembly sources ([#2906](https://github.com/0xMiden/miden-vm/pull/2906))
 - Created new `miden-format` tool for formatting Miden Assembly sources while preserving comments and certain whitespace choices ([#2906](https://github.com/0xMiden/miden-vm/pull/2906))
++- Switched the default parser backend for Miden Assembly to use the new lossless parser ([#2907](https://github.com/0xMiden/miden-vm/pull/2907))
```
huitseeker: Tiny typo: this starts with `+-`, so it renders as literal text instead of a changelog bullet. Could you drop the extra `+`?

huitseeker (approved):
LGTM! Thanks a lot.

CHANGELOG.md
bitwalker: Well spotted, will do!

PR #2908, opened 2026-03-25
revert(assembly): restore InvokeKind::Exec in visit_mut_procref

amathxbt:
## Summary

Reverts the change introduced in #2893 and adds an explanatory comment.

### What changed
In `visit_mut_procref`, the invoke kind was changed from `InvokeKind::Exec` → `InvokeKind::ProcRef`. This was incorrect.

### Why `InvokeKind::Exec` is correct here
A `procref` instruction captures a procedure reference for *later* invocation. At this point in the pass we cannot know what call kind will ultimately be used — so we must **pessimistically** treat it as an actual invocation. `Exec` is the most general invocation kind, and the linker relies on this signal to correctly track procedure dependencies.

Using `InvokeKind::ProcRef` only communicates "this procedure is named somewhere", which does **not** carry the full weight of "this procedure is invoked". This would cause the linker to under-count invocation dependencies.

An explanatory comment has been added above the line so this does not get accidentally changed again in the future (as suggested by @bobbinth).

### References
- Flagged by @bitwalker in https://github.com/0xMiden/miden-vm/pull/2893#issuecomment-4122254987
- Confirmed by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2893#issuecomment-4122795938

cc @bitwalker @bobbinth @huitseeker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v0.23.0 (TBD)
 
+#### Bug Fixes
+
+- Reverted `InvokeKind::ProcRef` back to `InvokeKind::Exec` in `visit_mut_procref` and added an explanatory comment (#2893).
+
 ## 0.22.0 (2025-03-18)
 
 #### Enhancements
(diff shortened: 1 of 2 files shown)


bitwalker:
You'll need to sign your commits, and then rebase on `next`, but once that's done this can be merged

PR #2909, opened 2026-03-25
Add a package content digest for package identity

huitseeker:
This adds `Package::content_digest()` as the digest for package identity and keeps `Package::digest()` as the MAST digest. The content digest covers the MAST digest, package name, version, kind, manifest, and semantic package sections such as `ACCOUNT_COMPONENT_METADATA`.

The content digest format is versioned as `miden.package.content.v2`. Descriptions and opaque custom sections are still left out. Kernel-section binding is still deferred.

The tests check that the content digest changes when package identity, manifest data, or account-component metadata changes, and stays stable when only descriptions or opaque custom sections change.

### DIFF
--- a/crates/mast-package/src/package/mod.rs
+++ b/crates/mast-package/src/package/mod.rs
@@ -82,7 +82,7 @@ impl Package {
         target: &mut W,
         kernel_digest: Option<&Word>,
     ) {
-        target.write_bytes(b"miden.package.content.v1");
+        target.write_bytes(b"miden.package.content.v2");
         self.digest().write_into(target);
         self.name.write_into(target);
         self.version.as_ref().map(|v| v.to_string()).write_into(target);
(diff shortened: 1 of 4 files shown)


bitwalker (approved):

crates/mast-package/src/package/mod.rs
```diff
@@ -54,6 +64,49 @@ impl Package {
...
+    /// This is distinct from [`Self::digest`], which is only the digest of the underlying MAST
+    /// artifact. The content digest currently binds the MAST digest, package name, semantic
+    /// version, package kind, manifest, and any semantic package sections. Package descriptions
+    /// and opaque custom sections are intentionally excluded for now; kernel-section binding is
+    /// added separately.
+    pub fn content_digest(&self) -> Word {
+        let mut bytes = Vec::new();
+        self.write_content_digest_preimage(&mut bytes, None);
+        Poseidon2::hash(&bytes)
+    }
```
bobbinth, line 78: Not for this PR, but there is an argument to be made that this should be `digest()` and what is now `digest()` should be something like `mast_digest()`.

bobbinth (approved):
Looks good! Thank you!

The changes in this PR are fine, but I think we need to think through the whole commitment approach to MAST forest/package more holistically. Specifically, we have:

Let's create an issue for this (unless we already have one).

PR #2910, opened 2026-03-25
docs(memory): fix delta encoding comments in memory trace

### LINKED ISSUE #2815: Memory trace documentation misstates delta encoding for same-word accesses
The trace layout comment on `Memory` (`processor/src/trace/chiplets/memory/mod.rs:85–92`) defines the delta columns for repeated accesses to the same word and context as `new_clk - old_clk - 1`. Both code paths that compute and consume this delta use `new_clk - old_clk` instead:

- `Memory::append_range_checks`
- `Memory::fill_trace`

The expected rows in memory trace tests also encode `new_clk - old_clk`. Additionally, the comment states that the first row has `d0`, `d1`, and `d_inv` set to zero, but in practice both functions initialise `prev_clk` to `first_clk - 1`, yielding a non-zero first-row delta when `first_clk` is non-zero.

Fix: update the `Memory` trace layout comment to:
- Define the same-word delta as `new_clk - old_clk`
- Update the first-row description to reflect current `prev_clk` initialisation

alimmaster:
## Current behavior

The memory trace documentation in `processor/src/trace/chiplets/memory/mod.rs` states that repeated accesses to the same word/context use a delta of `new_clk - old_clk - 1`.

However, both `append_range_checks()` and `fill_trace()` compute the same-word delta as `new_clk - old_clk`.

The file also says that the first row has `d0`, `d1`, and `d_inv` set to zero, while the implementation initializes `prev_clk` to one less than the first row clock and computes the first-row delta from that value.

## New behavior

The documentation now matches the implementation:

- same-word delta is documented as `new_clk - old_clk`
- the first-row description now reflects the current `prev_clk` initialization
- nearby inline comments were updated to describe the same behavior consistently

## Breaking changes

None.

## Other info

Documentation-only update.
No code behavior was changed.

Fixes #2815


alimmaster:
Closes #2815

huitseeker (approved):
Thanks!

huitseeker:
@alimmaster please add a Changelog line, you can take #2925 as an example

alimmaster:
> @alimmaster please add a Changelog line, you can take #2925 as an example

Added the changelog entry.

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -82,13 +82,14 @@ const INIT_MEM_VALUE: Word = EMPTY_WORD;
...
+///     `old_clk`).
 /// - `d_inv` contains the inverse of the delta between two consecutive context IDs, words, or clock
 ///   cycles computed as described above. It is the field inverse of `(d_1 * 2^16) + d_0`
 /// - `f_scw` is a flag indicating whether the context and the word of the current row are the same
-///   as in the next row.
+///   as in the previous row.
 ///
-/// For the first row of the trace, values in `d0`, `d1`, and `d_inv` are set to zeros.
+/// For the first row of the trace, `prev_clk` is initialized to `first_clk - 1`, so `d0`,
+/// `d1`, and `d_inv` reflect the delta against that value rather than being forced to zero.
```
adr1anh: I find the `old`/`new` a bit confusing when it comes to the constraints, since we often use local/current and next for the two consecutive rows.

I get the this trace is a bit special because we're treating the current row as the previous state and the next one as the current one, but maybe a better naming would be `prev`/`curr` to be a bit more consistent?

Cc @Al-Kindi-0, @plafer

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -248,8 +249,9 @@ impl Memory {
     /// [RangeChecker] chiplet instance, along with their row in the finalized execution trace.
     pub fn append_range_checks(&self, memory_start_row: RowIndex, range: &mut RangeChecker) {
         // set the previous address and clock cycle to the first address and clock cycle of the
```
adr1anh: Nit: should this be "initialize the previous..."

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -248,8 +249,9 @@ impl Memory {
...
-        // trace; we also adjust the clock cycle so that delta value for the first row would end
-        // up being ZERO. if the trace is empty, return without any further processing.
+        // trace; we also adjust the clock cycle to be one less than the first row's clock so the
+        // first-row delta is computed against the immediately preceding clock. if the trace is
```
adr1anh: This means the delta is one right? Isn't that clearer?

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -291,9 +293,10 @@ impl Memory {
     pub fn fill_trace(self, trace: &mut TraceFragment) {
         debug_assert_eq!(self.trace_len(), trace.len(), "inconsistent trace lengths");
 
-        // set the pervious address and clock cycle to the first address and clock cycle of the
-        // trace; we also adjust the clock cycle so that delta value for the first row would end
-        // up being ZERO. if the trace is empty, return without any further processing.
+        // set the previous address and clock cycle to the first address and clock cycle of the
+        // trace; we also adjust the clock cycle to be one less than the first row's clock so the
+        // first-row delta is computed against the immediately preceding clock. if the trace is
+        // empty, return without any further processing.
```
adr1anh: Same comments as above

adr1anh (changes requested):
It took me a while to make sense of the comments, even before the change. I’m mostly familiar with the constraints but it's been a while since I checked the trace generation.

I left some comments with my understanding which I'm not completely sure of. It'd be great if you could actually explain how this works, and propose something clearer overall.

While the changes might be slightly more correct, I don't think they actually help understand what's going on.

huitseeker:
Superseded by #3062

PR #2920, opened 2026-03-26
fix(linker): reject non-syscall invocations of kernel procedures (closes #2902)

### LINKED ISSUE #2902: Linker accepts non-syscall references to exported kernel procedures

amathxbt:
## Summary

Fixes **#2902** — the linker accepted `exec`, `call`, and `procref` targeting exported kernel procedures. Only `syscall` should reach kernel exports.

---

### Root Cause

`SymbolResolver::resolve_invoke_target` only checked whether a resolved procedure belonged to the kernel when `context.in_syscall()` was `true`. All other invoke kinds (`exec`, `call`, `procref`, and digest-based `MastRoot` invocations) fell through to the default `Ok(SymbolResolution::Exact)` arm without any guard.

This means a program like:

```masm
begin
    exec.::::k1   ;; should be rejected — is silently accepted
end
```

assembled successfully and ran kernel code outside the `syscall` context, bypassing the VM's context-switch semantics.

---

### Fix

**1. New diagnostic error variant — `LinkerError::KernelProcNotSyscall`**

```
error: invalid invocation: 'k1' is a kernel procedure and must be called via 
 --> program.masm:2:5
  |
2 |     exec.::::k1
  |     ^^^^^^^^^^^^^^^^^^ invocation occurs here
  |
  = help: use  instead of  to call exported kernel procedures
```

**2. Two focused helper methods on `SymbolResolver`**

```rust
fn is_kernel_proc(&self, gid: GlobalItemIndex) -> bool {
    self.graph.kernel_index.is_some_and(|k| k == gid.module)
}

fn reject_kernel_proc_non_syscall(
    &self, context: &SymbolResolutionContext, callee: Arc<Path>,
) -> LinkerError { ... }
```

**3. Guards in all three resolution arms**

Every `Some(gid) => Ok(SymbolResolution::Exact)` arm that is reached only when `!context.in_syscall()` now calls `is_kernel_proc(gid)` and returns the new error, covering:
- `InvocationTarget::MastRoot` — digest-based invocations
- `InvocationTarget::Symbol` — bare symbol references
- `InvocationTarget::Path` — qualified path references (both Exact and MastRoot sub-arms)

---

### Tests

Three new regression tests in `crates/assembly/src/tests.rs`:

| Test | Expectation |
|---|---|
| `regression_2902_kernel_proc_exec_is_rejected` | `exec.::::k1` → `Err` |
| `regression_2902_kernel_proc_call_is_rejected` | `call.::::k1` → `Err` |
| `regression_2902_kernel_proc_via_syscall_is_accepted` | `syscall.::::k1` → `Ok` |

---

### Note on existing open PR

PR #2903 (by @crazywriter1) tackles the same issue and has received helpful review from @bitwalker and @huitseeker. This PR provides an alternative, complementary implementation using a dedicated error variant and centralised helper methods for better diagnostics. Happy to coordinate and merge whichever approach the maintainers prefer.

---

> @bobbinth @bitwalker @huitseeker @plafer @Fumuran @igamigo — please review. This enforces a documented security invariant that kernel procedures are only reachable via `syscall`.

cc @crazywriter1

### DIFF
--- a/crates/assembly/src/linker/resolver/symbol_resolver.rs
+++ b/crates/assembly/src/linker/resolver/symbol_resolver.rs
@@ -86,6 +86,15 @@ impl<'a> SymbolResolver<'a> {
         self.graph.kernel_index.is_some_and(|k| k == gid.module)
     }
 
+    /// Returns `true` if the *caller* (i.e. `context.module`) is itself part of the kernel module.
+    ///
+    /// Kernel procedures are allowed to call other kernel procedures via `exec`/`call` — the
+    /// `syscall`-only restriction only applies to userland callers (#2902).
+    #[inline]
+    fn is_kernel_caller(&self, context: &SymbolResolutionContext) -> bool {
+        self.graph.kernel_index.is_some_and(|k| k == context.module)
+    }
+
     /// Reject a non-syscall invocation that resolved to a kernel export.
     ///
     /// Kernel procedures must only be reachable through `syscall` so that the VM's
@@ -135,9 +144,10 @@ impl<'a> SymbolResolver<'a> {
                         }
                     },
                     Some(gid) => {
-                        // Non-syscall invocation: reject if the resolved proc is a kernel export.
-                        // Kernel procedures must only be reachable via `syscall` (#2902).
-                        if self.is_kernel_proc(gid) {
+                        // Non-syscall invocation: reject if the resolved proc is a kernel export
+                        // AND the caller is not itself a kernel procedure.
+                        // Kernel procedures may freely call each other via exec/call (#2902).
+                        if self.is_kernel_proc(gid) && !self.is_kernel_caller(context) {
                             let callee = self.item_path(gid);
                             return Err(self.reject_kernel_proc_non_syscall(context, callee));
                         }
@@ -176,8 +186,9 @@ impl<'a> SymbolResolver<'a> {
                         })
                     },
                     SymbolResolution::Exact { gid, path } if !context.in_syscall() => {
-                        // Non-syscall: reject kernel exports (#2902)
-                        if self.is_kernel_proc(gid) {
+                        // Non-syscall: reject kernel exports, unless the caller is also in the
+                        // kernel (kernel procs may freely call each other via exec/call) (#2902).
+                        if self.is_kernel_proc(gid) && !self.is_kernel_caller(&context) {
                             let callee = path.into_inner();
                             return Err(self.reject_kernel_proc_non_syscall(&context, callee));
                         }
@@ -205,9 +216,11 @@ impl<'a> SymbolResolver<'a> {
                         })
                     }
                 },
-                // Non-syscall exact resolution: reject kernel exports (#2902).
-                // Kernel procedures must only be invoked via `syscall`.
-                SymbolResolution::Exact { gid, path } if self.is_kernel_proc(gid) => {
+                // Non-syscall exact resolution: reject kernel exports unless the caller is
+                // itself a kernel procedure (kernel procs may freely call each other) (#2902).
+                SymbolResolution::Exact { gid, path }
+                    if self.is_kernel_proc(gid) && !self.is_kernel_caller(context) =>
+                {
                     let callee = path.into_inner();
                     Err(self.reject_kernel_proc_non_syscall(context, callee))
                 },
@@ -233,8 +246,9 @@ impl<'a> SymbolResolver<'a> {
                             }
                         },
                         Some(gid) => {
-                            // Non-syscall MastRoot resolved to kernel export: reject (#2902)
-                            if self.is_kernel_proc(gid) {
+                            // Non-syscall MastRoot resolved to kernel export: reject unless the
+                            // caller is itself a kernel proc (they may call each other) (#2902).
+                            if self.is_kernel_proc(gid) && !self.is_kernel_caller(context) {
                                 let callee = self.item_path(gid);
                                 return Err(self.reject_kernel_proc_non_syscall(context, callee));
                             }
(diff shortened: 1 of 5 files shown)


amathxbt:
## PR #2920 — Fix Breakdown, CI Debugging & Final Status

This PR fixes issue #2902 — a correctness bug where the assembler linker allowed `exec`, `call`, and `procref` to target exported kernel procedures, bypassing the VM's context-switch semantics.

---

### 🐛 Root Cause (Issue #2902)

Kernel procedures are only valid under `syscall` because the VM enforces a context-switch for kernel calls. Reaching them via `exec`/`call` skips that context switch, violating the security model.

The assembler's `SymbolResolver::resolve_invoke_target` had **no guard** against non-syscall invocations of kernel exports — the issue was reported by .

---

### ✅ Fix Applied

Three changes in `crates/assembly/src/linker/resolver/symbol_resolver.rs`:

1. **New helper**: `is_kernel_proc(gid)` — checks if a resolved `GlobalItemIndex` belongs to the kernel module
2. **New helper**: `reject_kernel_proc_non_syscall(context, callee)` — builds a structured `LinkerError::KernelProcNotSyscall` with diagnostic help message
3. **Guard in all 3 invocation arms** — Symbol, Path, and MastRoot — reject kernel-proc targets unless `context.in_syscall()` is true

New error variant in `crates/assembly/src/linker/errors.rs`:
```
error[KernelProcNotSyscall]: invalid invocation: '::::foo' is a kernel procedure
  and must be called via
  --> src/main.masm:3:5
   |
 3 |     exec.::::foo
   |     ^^^^^^^^^^^^^^^^^^^ invocation occurs here
   |
   = help: use  instead of
```

---

### 🔧 CI Bugs Found & Fixed

| Round | CI Check | Root Cause | Fix |
|-------|----------|------------|-----|
| 1 | `rustfmt` | Long lines in error definition | Reformatted |
| 1 | `E0308 mismatched types` | Passed `context` by value instead of `&context` to `reject_kernel_proc_non_syscall` | Added `&` borrow |
| 1 | `changelog` | Missing entry | Added CHANGELOG entry |
| **2** | **`doc-tests` (README.md lines 139, 175)** | **Critical bug: `linker/mod.rs` dep-graph phase used `kind: None` for all invocations, causing `syscall.foo` to be resolved without the kernel-module context — triggering our new guard for VALID syscall instructions** | **Fixed: pass `kind: Some(invoke.kind)` to preserve the invocation kind** |

The most interesting fix was in **Round 2**. When building the dependency graph (for dead-code elimination), the linker iterated over all invocations with:

```rust
// BEFORE — kind dropped, syscall resolved as if it were exec
let context = SymbolResolutionContext {
    span: invoke.span(),
    module: module_index,
    kind: None,   // 🐛 lost the SysCall kind!
};
```

This caused our kernel-proc guard to fire on completely valid `syscall.foo` instructions, breaking the README doctests. The fix:

```rust
// AFTER — kind preserved
let context = SymbolResolutionContext {
    span: invoke.span(),
    module: module_index,
    kind: Some(invoke.kind),  // ✅ syscall stays syscall
};
```

---

### 🧪 Tests Added (crates/assembly/src/tests.rs)

- `kernel_proc_exec_is_rejected_at_link_time()` — verifies `exec.::::foo` is a linker error
- `kernel_proc_call_is_rejected_at_link_time()` — verifies `call.::::foo` is a linker error
- `kernel_proc_syscall_succeeds()` — verifies `syscall.foo` still works correctly ✅

---

### 📊 Current CI Status
New CI run just started for the latest commit (kind preservation fix). All previous checks were ✅ except doc-tests, which should now pass.

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -184,10 +245,18 @@ impl<'a> SymbolResolver<'a> {
                                 })
                             }
                         },
-                        Some(gid) => Ok(SymbolResolution::Exact {
-                            gid,
-                            path: Span::new(mast_root.span(), self.item_path(gid)),
-                        }),
+                        Some(gid) => {
+                            // Non-syscall MastRoot resolved to kernel export: reject unless the
```
huitseeker, line 249: This is still taking the digest and `MastRoot` path in the linker, which goes the other way from what @bitwalker asked for on `#2903`. The direction there was to avoid trying to enforce dynamic kernel calls in symbol resolution. Given that, we are going to close this in favor of `#2903`.

huitseeker:
Closing in favor of #2903.

PR #2921, opened 2026-03-26
feat(sema): warn on unused private constants in assembler modules (closes #2898)

### LINKED ISSUE #2898: Warn about unused constants
The assembler currently warns about unused imports and unused docstrings, but not about unused constants. 

We use warnings_as_errors in the protocol to make sure we have no unused imports and the same would be nice to have for constants (e.g. no longer used error constants).

Of course, this is not of very high priority.

amathxbt:
## Summary

Implements **#2898**  the assembler already warns about unused imports and unused docstrings; this PR adds an equivalent lint for unused private constant definitions.

---

### Problem

Currently this module assembles without any feedback:

```masm
const STALE_ERROR_CODE = 0xff   # was used once, now dead — silent

export.foo
    push.1
end
```

When `warnings_as_errors` is active (the default in the Miden protocol build via `masm_lints`) dead constants accumulate silently, making it harder to keep the codebase clean.

---

### Implementation

**`crates/assembly-syntax/src/sema/errors.rs`** — new variant:

```rust
#[error("unused constant '{name}'")]
#[diagnostic(
    severity(Warning),
    help("remove it or prefix its name with '_' to suppress this warning")
)]
UnusedConstant { span: SourceSpan, name: String }
```

**`crates/assembly-syntax/src/sema/context.rs`**

- Added `used_constants: BTreeSet<Ident>` field.
- `ConstEnvironment::get()` marks a constant as used on every successful lookup — this covers all usage sites: procedure bodies, repeat counts, immediates, and nested constant expressions.
- Two new read-only accessors: `used_constants()` and `defined_constants()`.

**`crates/assembly-syntax/src/sema/mod.rs`** — new check added after the existing unused-import check:

```rust
// Collect private constants that were never referenced
let used = analyzer.used_constants().cloned().collect::<BTreeSet<_>>();
for (name, span) in analyzer
    .defined_constants()
    .filter(|(n, c)| !c.visibility.is_public() && !used.contains(*n) && !n.starts_with('_'))
{
    analyzer.error(SemanticAnalysisError::UnusedConstant { span, name: name.to_string() });
}
```

---

### Behaviour matrix

| Constant | Warning emitted? |
|---|:---:|
| Private, never referenced | ✅ |
| Private, used in any procedure or constant expr | ❌ |
| Private, name starts with `_` (opt-out) | ❌ |
| `pub` / exported — part of module API | ❌ |

---

### Tests (crates/assembly-syntax/src/sema/tests.rs)

| Test | Scenario |
|---|---|
| `unused_private_constant_emits_warning` | private unused → warning/error |
| `used_private_constant_does_not_warn` | private used in proc → no warning |
| `exported_constant_does_not_warn_even_if_locally_unused` | pub unused locally → no warning |
| `underscore_prefixed_constant_suppresses_warning` | `_RESERVED` → no warning |

---

### Note

This is a **warning** (not a hard error) by default. It is only promoted to a compile error when the caller sets `warnings_as_errors = true`, which is consistent with how `UnusedImport` behaves.

---

> @bobbinth @bitwalker @huitseeker @plafer @Fumuran @igamigo — this was requested in #2898. The change is additive, backward-compatible, and follows the exact pattern of the existing unused-import warning. Happy to adjust the scope or opt-out convention.

### DIFF
--- a/crates/assembly-syntax/src/sema/mod.rs
+++ b/crates/assembly-syntax/src/sema/mod.rs
@@ -185,14 +185,14 @@ pub fn analyze(
             .filter(|(name, cst)| {
                 !cst.visibility.is_public()                    // private only
                     && !used.contains(*name)                   // never referenced
-                    && !name.as_str().starts_with('_')         // not opted-out
+                    && !name.as_str().starts_with('_') // not opted-out
             })
             .map(|(name, cst)| (name.clone(), cst.span))
             .collect();
         for (name, span) in unused {
             analyzer.error(SemanticAnalysisError::UnusedConstant {
                 span,
-                name: name.to_string(),
+                name: alloc::format!("{name}"),
             });
         }
     }
--- a/crates/assembly-syntax/src/sema/mod.rs
+++ b/crates/assembly-syntax/src/sema/mod.rs
@@ -171,6 +171,32 @@ pub fn analyze(
         }
     }
 
+    // Check unused constants (#2898).
+    //
+    // Only private (non-exported) constants are checked: exported constants are part of the
+    // module's public API and callers outside this module may reference them.
+    //
+    // Constants whose names start with `_` are explicitly opted out of this warning,
+    // mirroring the convention used in Rust for intentionally-unused bindings.
+    {
+        let used: alloc::collections::BTreeSet<_> = analyzer.used_constants().cloned().collect();
+        let unused: alloc::vec::Vec<_> = analyzer
+            .defined_constants()
+            .filter(|(name, cst)| {
+                !cst.visibility.is_public()                    // private only
+                    && !used.contains(*name)                   // never referenced
+                    && !name.as_str().starts_with('_')         // not opted-out
+            })
+            .map(|(name, cst)| (name.clone(), cst.span))
+            .collect();
+        for (name, span) in unused {
+            analyzer.error(SemanticAnalysisError::UnusedConstant {
+                span,
+                name: name.to_string(),
+            });
+        }
+    }
+
     analyzer.into_result().map(move |_| module)
 }
 
--- a/crates/assembly-syntax/src/sema/mod.rs
+++ b/crates/assembly-syntax/src/sema/mod.rs
@@ -175,17 +175,13 @@ pub fn analyze(
     //
     // Only private (non-exported) constants are checked: exported constants are part of the
     // module's public API and callers outside this module may reference them.
-    //
-    // Constants whose names start with `_` are explicitly opted out of this warning,
-    // mirroring the convention used in Rust for intentionally-unused bindings.
     {
         let used: alloc::collections::BTreeSet<_> = analyzer.used_constants().cloned().collect();
         let unused: alloc::vec::Vec<_> = analyzer
             .defined_constants()
             .filter(|(name, cst)| {
-                !cst.visibility.is_public()                    // private only
-                    && !used.contains(*name)                   // never referenced
-                    && !name.as_str().starts_with('_') // not opted-out
+                !cst.visibility.is_public() // private only
+                    && !used.contains(*name) // never referenced
             })
             .map(|(name, cst)| (name.clone(), cst.span))
             .collect();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ## 0.22.0 (2025-03-18)
 
 #### Enhancements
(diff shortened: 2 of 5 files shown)


crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -171,6 +171,32 @@ pub fn analyze(
         }
     }
 
+    // Check unused constants (#2898).
+    //
+    // Only private (non-exported) constants are checked: exported constants are part of the
+    // module's public API and callers outside this module may reference them.
+    //
+    // Constants whose names start with `_` are explicitly opted out of this warning,
```
bitwalker: There's no reason to support this - these are constant declarations, not let-bound variables.

bitwalker (changes requested):

crates/assembly-syntax/src/sema/mod.rs
amathxbt: > There's no reason to support this - these are constant declarations, not let-bound variables.

Give me minutes not done yet I'll tag you again once I done fixed with good explanation thank you Chad

amathxbt: Great point @bitwalker — you're absolutely right and I've addressed this in the latest commit.

**Why the  prefix was there originally:**
I borrowed the convention from Rust's compiler () where prefixing with  suppresses the unused-variable lint. The idea was to give developers an escape hatch without having to delete a constant they might want to keep as documentation or future use.

**Why it's wrong here:**
Constants in MASM are *declarations*, not bindings. They don't participate in drop semantics, borrow checking, or any of the lifetime machinery that makes the  convention meaningful in Rust. There's also no precedent in the assembler's existing lints (e.g.  doesn't have a -prefix bypass). Adding it would create an inconsistent, undocumented escape hatch that serves no real purpose.

**What I changed:**
- Removed  from the filter in .
- Updated the corresponding test from  →  to document the correct behaviour.

If developers have a constant they want to keep but not use (e.g. reserved addresses for future ABI compatibility), they can simply export it () to silence the warning — which is the correct signal: it's part of the intentional API surface.

Latest commit: be2d6e1d

amathxbt:
## PR #2921 — Fix Breakdown & CI Status

This PR adds assembler warnings for unused private constants, closing issue #2898.

---

### 🐛 Problem (Issue #2898)

The Miden assembler had no way to tell you when a `const` declaration is defined but never used. Unused constants are dead code and can indicate refactoring mistakes or copy-paste errors. This is similar to Rust's `unused_variable` / `dead_code` lint.

Originally reported by  and confirmed by .

---

### ✅ Feature Implemented

**New warning variant** (`crates/assembly-syntax/src/sema/errors.rs`):
```
warning: unused constant `FOO`
  --> module.masm:1:7
   |
 1 | const FOO = 42;
   |       ^^^ defined here, never used
   |
   = note: remove this constant or use it somewhere in the module
```

**Scope rules** (per 's review feedback):
- ✅ Private constants that are never referenced → **warning**
- ✅ Exported (`pub const`) constants → **no warning** (may be used by downstream consumers)
- ✅ Used constants → **no warning**

**Implementation** (`crates/assembly-syntax/src/sema/mod.rs` + `context.rs`):
1. `AnalysisContext` gains a `used_constants: BTreeSet<Ident>` tracking set
2. `get_constant()` / `simplify_constants()` mark constants as used when they are evaluated
3. After semantic analysis completes, any private constant not in `used_constants` emits `SemanticAnalysisError::UnusedConstant`

---

### 🔧 CI Fixes Applied

| Round | Issue | Fix |
|-------|-------|-----|
| 1 | `rustfmt` alignment | Reformatted help message |
| 1 | `E0599`: `Ident` has no `to_string()` | Changed to `alloc::format!("{}", name)` |
| 1 | `changelog` missing | Added entry under v0.23.0 |
| 2 | ** review**: "no reason to support `_` prefix opt-out — these are constant declarations, not let-bound variables" | **Removed the underscore-prefix opt-out; all unused private constants now warn uniformly** |

---

### 🧪 Tests Added (`crates/assembly-syntax/src/sema/tests.rs`)

- `unused_private_constant_warns()` — private const never used → warning emitted
- `used_constant_does_not_warn()` — used const → no warning
- `exported_constant_does_not_warn()` — `pub const` → no warning regardless of usage

---

### 📊 Current CI Status
All checks are ✅ green except the full test suite still running. No new failures expected.

bobbinth:
Thank you - but I'll close this for now as our review bandwidth is limited and this is not a high priority issue. @amathxbt it may make sense to wait to be assigned to the issue before taking it on.

amathxbt:
> Thank you - but I'll close this for now as our review bandwidth is limited and this is not a high priority issue. @amathxbt it may make sense to wait to be assigned to the issue before taking it on.

I appreciate your feedback bobbin
Thank you see you in the other opened PR

PR #2922, opened 2026-03-27
fix(processor): run after_exit decorators when tracing

KitaroMoha563:
Align after-exit decorator gating with tracing-aware execution, so Trace decorators are emitted when tracing is enabled without debug mode.

PR #2925, opened 2026-03-27
docs(assembly): clarify div vs u32div semantics

### LINKED ISSUE #2911: Strengthen field division warning in assembly docs
The field operations page already notes that "the result of divisions may appear counter-intuitive because divisions are defined via inversions", and the math formula correctly shows `a * b^{-1} mod p`. The `code_organization.md` page also distinguishes `/` (field division) from `//` (integer division) for constant expressions.

However, the warning is easy to miss, and there is no cross-reference between the field operations `div` and the u32 operations `u32div` making it explicit that these are fundamentally different operations. Developers coming from other ecosystems expect `/` to be floor division, and the current documentation relies on them reading the math formula carefully.

I think it would be helpful to make the existing warning more prominent and add a short note like "For integer floor division, use `u32div`" directly on the `div` instruction entry. A cross-reference between the two pages would go a long way.

cwhitak3r:
Closes #2911, make field-division warnings explicit and add cross-references between `div` and `u32div` to avoid integer-division confusion.


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 #### Bug Fixes
 
 - Reverted `InvokeKind::ProcRef` back to `InvokeKind::Exec` in `visit_mut_procref` and added an explanatory comment (#2893).
+- Fixed `FastProcessor` so `after_exit` trace decorators execute when tracing is enabled without debug mode, and added a tracing-only regression test.
```
huitseeker, line 8: This is not the right PR for this actually (#2922), please clean up.

huitseeker:
Closing for inacitivty

Issue #2220, opened 2025-09-26
Constraints: Investigate alternative for hasher sibling table


adr1anh:
The chiplet virtual table is now combined with the chiplet bus, in particular for performing memory requests from the ACE chiplet. At the moment, this table can only be used by the chiplet trace since the hasher chiplet enforces that the running product "resets" (i.e. equals zero) between two operations which update a Merkle tree. 

As a consequence, the virtual table cannot be used by the system and decoder traces since these would affect the boundary check if these traces access the bus within the rows where we are computing a tree update. 

It's unclear to me at the moment what the existing degree of this bus is. If it does indeed have spare capacity, it could be used for example to define the precompile logging opcode discussed in #2147.

The addition of the ACE chiplet required us to add an extra column to the chiplet trace, so we should investigate whether we can use this column to avoid having to enforce a boundary constraint on the auxiliary virtual table column.

adr1anh:
Now that we have an extra column in the chiplet table, we can use one of them to remove the need for $p_1 = 1$ at the start and end of a merkle update request. The solution is to include some "nonce" in all the sibling messages pertaining to the same Merkle tree update, across both `MV` and `MU` blocks. The nonce acts as a domain separator to ensure the siblings of two different updates cannot interfere with each other.

There are two options we can consider.

**Option 1**:
When requesting an operation from the Hasher chiplet, the message for `MV` and `MU` operation includes a new `clk` field, corresponding to the system clock. The two requests `mrupdatenew` and `mrupdateold` include this field, which we can include in the message as the last component (after the state portion). 

In the `MV/MU` blocks, we copy the `clk` to the new column. Then, whenever we have `MV/MVA/MU/MUA`, we copy the clock the next row. This clock is then included in the sibling request/response messages. 

The advantage of this approach is that both `MV` and `MU` blocks are now 

**Option 2**:
We use the `addr` field from the first permutation block, store it in the new column, and copy it until we reach the output row of the update blocks. This avoids having to include a new entry in the `HasherMessage`, but is more brittle as it depends on both the MV and MU block being consecutive (which seems to be guaranteed by the usage of `addr`).

adr1anh:
Another alternative that might be simpler is to interpret the extra column as a counter `ctr`, which we initialize to 0 at the start of every Merkle update request (or even all requests). It is incremented by 1 in each row. We can then derive the address of the request by taking the difference `addr_request = addr-ctr`. This provides a stable and unique identifier for each request. 

[...]

Issue #1614, opened 2025-01-08
Element-addressable memory follow-up


plafer:
This is an issue covering all the things left to do following #1598.

- [ ] Fix the book
  - i.e. fix all the places that refer to word-addressable memory 
- [ ] Add a 32-bit range-check on the memory address being accessed
  - This was previously never implemented
- [ ] Add a 32-bit range-check on the `word_addr` column
  - See context for why this is needed [here](https://github.com/0xPolygonMiden/miden-vm/issues/1064#issuecomment-2549712514) (where `batch` was since then renamed to `word_addr`)

Ideally, we'd only need a single 32-bit range-check, but having both is at least sufficient for correctness.

bobbinth:
Do we have a sketch for the 32-bit range checks? It seems to me that we need these for soundness (and we also need to make sure that the address is a multiple of 4 for word-addressable operations). The naive approach would require 2 more columns in the main trace and 1 more column in the auxiliary trace. This would bring up the cost of element-addressable memory to 3 extra columns in the main trace and 1 extra column in the auxiliary trace, making it somewhat less attractive.

Could we do anything better than the naive approach?

Al-Kindi-0:
Maybe we can avoid the 3 extra columns in the main trace (but not the one in the auxiliary trace) by relying on just range checking the address on the decoder side.
Will open an issue soon to further discuss this.

bobbinth:
I think we can probably discuss this in this issue.

One potential solution which almost works is to add `010` operation prefix as the flag for range check operations on the stack side (similar to how we do this for operations with `100` prefix). Basically, any operation that has `010` or `100` would have 4 range checks available to it. This would require:

1. Making a minor modification to the `EQ` operation (so that it uses a different helper register).
2. Moving `FRIE2F4` operation into a different bucket (e.g., maybe into "high-degree operations").

This would help us with `MLOADW`, `MSTOREW`, and `MSTORE` operations. The unsolved bit is the `MLOAD` operation which has a different prefix because we don't shift stack to the left on `MLOAD`.

Al-Kindi-0:
I would like to pinpoint the problem(s) we are solving. The way I understand it, the main question is:

Do we need to range check **both** the memory address of the op as well as the memory address of the word? Or is range checking only one of the two is enough and if yes, which one?

I think the answer is that we are going to need two u32 range checks, however, we can do both range checks on the side of the decoder. The way to go about it is:

[...]

PR #2856, opened 2026-03-16
refactor(air): remove tagging, apply uniform constraint description, and optimize evaluation

adr1anh:
## Describe your changes

This PR aims to simplify our current constraints. To ease review, each commit can be examined separately to ensure transformations are semantics-preserving.

WIP

## FG's intro

https://gisthost.github.io/?d05c574a2bdbc518ca1bc404f3d02409

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](../CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #2927, opened 2026-03-27
Hasher chiplet redesign

Al-Kindi-0:
This PR bundles several closely related changes in the hasher / chiplets area:

- **Hasher controller/permutation split** -- the hasher trace is split into a compact controller region and a separate permutation segment, enabling permutation deduplication.
- **Packed 16-row Poseidon2 permutation segment** -- the 31-step Poseidon2 schedule is packed from 32 rows down to 16 rows per unique permutation.
- **Sibling table soundness fix** ([#2220](https://github.com/0xMiden/miden-vm/issues/2220)) -- a new `mrupdate_id` column domain-separates sibling-table entries, preventing cross-operation sibling swapping.
- **Memory address range checks** ([#1614](https://github.com/0xMiden/miden-vm/issues/1614)) -- the memory chiplet gets `w0/w1` address-limb columns, with 16-bit range-check lookups routed through the wiring bus.

These changes all touch the chiplets trace layout, bus plumbing, and AIR structure, so landing them together keeps the transition coherent.

### 1. Deduplicate repeated permutations

The old monolithic hasher consumed **32 rows per permutation request**, even if the same input state appeared repeatedly.

With the new design:
- the **controller** records each request as a 2-row `(input, output)` pair,
- the **permutation segment** executes one packed **16-row** cycle per **unique** input state,
- a multiplicity counter records how many controller pairs map to the same cycle.

For `M` requests with `U` unique input states, the rough cost changes from:

- **old:** `32M`
- **new:** `2M + pad_to_16 + 16U`

This is a clear win whenever states repeat (Merkle workloads, identical MAST roots, ...).

### 2. Fix sibling-table soundness

The old sibling-table encoding was vulnerable to cross-operation sibling reuse. Adding `mrupdate_id` domain-separates entries so sibling-table balance is enforced **per MRUPDATE instance**, not globally across unrelated operations.

### 3. Add memory address decomposition checks

The memory chiplet now decomposes word addresses into two 16-bit limbs and proves the decomposition using range-check lookups. This closes an important missing piece in memory soundness while reusing the existing wiring-bus infrastructure.

## Hasher: two-region trace

The hasher trace is split into two contiguous regions:

- **Controller** (`perm_seg = 0`)  
  Compact input/output row pairs, one pair per permutation request.

- **Permutation segment** (`perm_seg = 1`)  
  One packed **16-row Poseidon2 cycle** per unique input state.

A LogUp permutation-link on the shared `V_WIRING` auxiliary column ties controller requests to the corresponding permutation cycles.

## Packed 16-row Poseidon2 schedule

The 31-step Poseidon2 schedule is packed as:

- row 0: `init + ext1`
- rows 1-3: `ext2..ext4`
- rows 4-10: `7 × (3 packed internal rounds)`
- row 11: `int22 + ext5`
- rows 12-14: `ext6..ext8`
- row 15: boundary / final state

Packed internal rows use `s0/s1/s2` as witness columns on permutation rows in order to keep constraints degree bounded. Unused witness slots are explicitly zero-constrained (out of caution) though this could be relaxed.

### Hasher: 16 -> 20

```
s0 s1 s2 | h0..h11 | node_index | mrupdate_id | is_boundary | direction_bit | perm_seg
   3          12          1             1             1              1             1      = 20
```

New / newly significant columns:
- `mrupdate_id` -- domain separator for sibling-table entries
- `is_boundary` -- marks first controller input / last controller output
- `direction_bit` -- propagated Merkle routing bit on controller rows
- `perm_seg` -- explicit controller vs permutation-region flag

### Memory: 15 -> 17

Two new columns:
- `w0`
- `w1`

These decompose the word address into 16-bit limbs. The wiring bus carries the corresponding range-check lookups.

## Constraints

Hasher constraints now total **100**.

<details>
<summary>Constraint group breakdown</summary>

| Group | Count | Purpose |
|-------|------:|---------|
| Selector booleanity | 3 | `s0,s1,s2` binary on controller rows |
| Perm segment | 7 | `perm_seg` confinement, booleanity, monotonicity, cycle alignment |
| Structural | 7 | Confine `is_boundary` / `direction_bit` to valid row types |
| Lifecycle | 2 | Operation lifecycle invariants |
| Controller adjacency | 2 | Input row must be followed by output row |
| Controller pairing | 4 | First-row constraint, output non-adjacency, padding stability |
| Perm witness-shape | 3 | Zero witness slots when unused |
| Perm init+ext | 12 | Row 0 packed transition |
| Perm external | 12 | External-round transitions |
| Perm packed internal | 15 | 3 witness checks + 12 next-state constraints |
| Perm int+ext | 13 | 1 witness check + 12 next-state constraints |
| MRUPDATE ID | 2 | Increment / zero-on-perm rules |
| Sponge capacity | 4 | Preserve capacity across continuations |
| Output index | 1 | Output-row `node_index` rule |
| Merkle index | 4 | Index decomposition / continuity / direction bit |
| Merkle input state | 4 | Zero capacity on Merkle input rows |
| Merkle routing | 5 | Route digest into correct rate half |
| **Total** | **100** | |

</details>

## Trace width impact

| Chiplet | Before | After | Delta |
|---------|-------:|------:|------:|
| Hasher | 16 | 20 | +4 |
| Memory | 15 | 17 | +2 |
| **Net main trace impact** |  |  | **+1** |

The new main trace width is **72**

No new auxiliary columns were added:
- the permutation-link bus shares `V_WIRING`
- memory address range checks also use the existing wiring-bus path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Added `prove_from_trace_sync(...)` for proving from pre-executed trace inputs ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
 - [BREAKING] Reduced the prove-from-trace API to post-execution trace inputs: `TraceBuildInputs` no longer carries full execution output, `prove_from_trace_sync()` takes `TraceProvingInputs`, and `ProvingOptions` no longer include `ExecutionOptions` ([#2948](https://github.com/0xMiden/miden-vm/pull/2948)).
 - Refactor trace generation to row-major format ([#2937](https://github.com/0xMiden/miden-vm/pull/2937)).
+- Redesigned the hasher chiplet to use a controller/permutation split architecture with permutation calls deduplication ([#2927](https://github.com/0xMiden/miden-vm/pull/2927)).
 
 #### Fixes
 


Al-Kindi-0:
To compare against the numbers in #2869 for the recursive verifier (verifying a program executing in 2^20 cycles)

```
  ┌────────────────────────────┬──────────────┬──────────────┬───────────┬──────────┐
  │         Component          │   Old        │   New.       │  Change   │ Savings  │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ Core trace (decoder+stack) │ 41,652       │ 41,516       │ -136      │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ Range checker              │ 5,129        │ 5,217        │ +88       │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ Chiplets total             │ 273,769      │ 118,657      │ -155,112  │ -57%     │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ - Hasher                   │ 250,816      │ 96,256       │ -154,560  │ -62%     │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ - Bitwise                  │ 3,104        │ 3,104        │ 0         │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ - Memory                   │ 13,758       │ 13,406       │ -352      │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ - ACE                      │ 6,090        │ 5,890        │ -200      │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ - Kernel ROM               │ 0            │ 0            │ 0         │          │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ Padded trace length        │524,288 (2^19)│131,072 (2^17)│           │ -4x      │
  ├────────────────────────────┼──────────────┼──────────────┼───────────┼──────────┤
  │ Padding                    │ 47%          │ 9%           │           │          │
  └────────────────────────────┴──────────────┴──────────────┴───────────┴──────────┘
```

This is a 4x improvement in the above case.
(Note that the changes in the decoder+stack number of rows is due to a change in the constraints which affects ACE circuit loading)

air/src/constraints/chiplets/bus/wiring.rs
```diff
@@ -77,107 +92,278 @@ pub fn enforce_wiring_bus_constraint<AB>(
...
+    AB: TaggingAirBuilderExt<F = Felt>,
+{
+    let alpha = &challenges.alpha;
+
+    // Load word-index limbs
+    let w0: AB::Expr = local.chiplets[MEMORY_WORD_ADDR_LO_COL_IDX - CHIPLETS_OFFSET].clone().into();
+    let w1: AB::Expr = local.chiplets[MEMORY_WORD_ADDR_HI_COL_IDX - CHIPLETS_OFFSET].clone().into();
+    let w1_mul4: AB::Expr = w1.clone() * AB::Expr::from_u16(4);
+
+    let den0: AB::ExprEF = alpha.clone() + Into::<AB::ExprEF>::into(w0);
```
huitseeker, line 240: Should this add protocol-level domain separation before `v_wiring` can safely carry ACE wires, raw memory range-check values, and the new hasher perm-link messages together? Right now the memory side uses plain `alpha + w0/w1/4*w1`, ACE uses `encode([clk, ctx, id, ...])`, and the perm-link uses `encode([0|1, h0..h11])` on the same LogUp column. 

If any of those encodings was to alias, could one subsystem cancel another on the shared sum? `#1614` explicitly called out adding an op-label when reusing the wiring bus, and I don't see that namespace implemented here yet.

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
```diff
@@ -0,0 +1,108 @@
...
+        let row: RowIndex = (row_idx as u32).into();
+
+        if !main_trace.is_hash_row(row) {
+            running_sum[row_idx + 1] = running_sum[row_idx];
+            continue;
+        }
+
+        let perm_seg = main_trace.chiplet_perm_seg(row);
+        let hs0 = main_trace.chiplet_selector_1(row);
+        let hs1 = main_trace.chiplet_selector_2(row);
```
Nashtare: nit: the indexing shift is a bit confusing when looking below, maybe renamed to 
```suggestion
        let hs1 = main_trace.chiplet_selector_1(row);
        let hs2 = main_trace.chiplet_selector_2(row);
```
with according updates later in the code would be clearer?

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
```diff
@@ -0,0 +1,108 @@
...
+    row: RowIndex,
+    challenges: &Challenges<E>,
+    label: Felt,
+) -> E {
+    let state = main_trace.chiplet_hasher_state(row);
+    let mut elems = [Felt::ZERO; 1 + STATE_WIDTH];
+    elems[0] = label;
+    elems[1..].copy_from_slice(&state);
+    challenges.encode(elems)
+}
```
Nashtare, line 116: I may be lacking some details here, but related to the soundness issues of having collisions with naive labels above, couldn't we envision reserving one `beta_powers` slot and give each subsystem a tag? 

Say for instance we have
```rust
const SUBSYSTEM_TAG_SLOT: usize = /* whatever first unused beta slot it is */;
const PERM_LINK_TAG: Felt = Felt::new(2);  // ACE = 0, Memory = 1, Perm-link = 2
```

Then this function would effectively become (i'm inlining the `challenges.encode(elems)`):

```suggestion
/// The message includes a domain-separation label and the full 12-element hasher state.
fn encode_perm_link_message<E: ExtensionField<Felt>>(
    main_trace: &MainTrace,
    row: RowIndex,
    challenges: &Challenges<E>,
    label: Felt,
) -> E {
    let mut acc = challenges.alpha
        + challenges.beta_powers[SUBSYSTEM_TAG_SLOT] * PERM_LINK_TAG;
    acc += challenges.beta_powers[0] * label;
    for i in 0..STATE_WIDTH {
        acc += challenges.beta_powers[1 + i] * h_i;
    }
    acc
}
```

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
```diff
@@ -0,0 +1,108 @@
...
+    chiplets::hasher::{HASH_CYCLE_LEN, STATE_WIDTH},
+};
+use miden_core::{Felt, field::ExtensionField};
+
+/// Labels for domain-separating input vs output perm-link messages.
+///
+/// TODO: These naive labels (0 and 1) risk collisions with other messages on the shared
+/// v_wiring column (ACE wiring and memory range checks). Revisit when refactoring the buses.
+const LABEL_IN: Felt = Felt::ZERO;
+const LABEL_OUT: Felt = Felt::ONE;
```
Nashtare, line 23: See my comment below in this file, but I was just wondering if we should not treat this now rather than deferring it?

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
```diff
@@ -0,0 +1,108 @@
...
+            } else if hs0 == Felt::ZERO && hs1 == Felt::ZERO {
+                // Controller output row (RETURN_HASH or RETURN_STATE with s0=0, s1=0): +1/msg_out
+                let msg_out = encode_perm_link_message(main_trace, row, challenges, LABEL_OUT);
+                running_sum[row_idx + 1] = running_sum[row_idx] + msg_out.inverse();
+            } else {
+                running_sum[row_idx + 1] = running_sum[row_idx];
+            }
+        } else {
+            // Permutation segment.
+            // This works because the hasher is always the first chiplet (rows start at 0)
```
Nashtare, line 71: minor: do we have an invariant check that this is actually the case? Just out of precaution

processor/src/trace/chiplets/hasher/mod.rs
```diff
@@ -14,60 +15,96 @@ mod trace;
...
+/// ├────┴───┴───┴────────┴────┴────────────┴─────────┴─────────┴────────┤
 #[derive(Debug, Default)]
 pub struct Hasher {
     trace: HasherTrace,
-    memoized_trace_map: BTreeMap<[u8; 32], (usize, usize)>,
+    /// Maps block digest -> (start_row, end_row) for memoized controller traces.
+    memoized_trace_map: BTreeMap<DigestKey, (usize, usize)>,
+    /// Maps input state -> multiplicity for permutation deduplication.
+    /// During finalize_trace(), one 16-row perm cycle is emitted per entry.
+    perm_request_map: BTreeMap<StateKey, u64>,
```
Nashtare: nit: does ordering matter? I think perf wise a `HashMap` may be more efficient here

Nashtare, line 72: nit: does ordering matter? I think perf wise a HashMap may be more efficient here

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
adr1anh: This is being done uniformly with the constraint refactor PR 😉

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
adr1anh: I’m pretty sure we always have 1 hash, and the chiplet selectors make sure that it comes before any other chiplet

processor/src/trace/chiplets/hasher/mod.rs
adr1anh: I think this is mainly because of no-std, but I don't think the ordering matters.

Nashtare: HashMap can be pulled from alloc / `hashbrown` though (if perf matters here)

docs/src/design/chiplets/index.md
```diff
@@ -23,7 +23,7 @@ The execution trace of the Chiplets module is generated by stacking the executio
 
 Each chiplet is identified within the Chiplets module by one or more chiplet selector columns which cause its constraints to be selectively applied.
 
-The result is an execution trace of 18 trace columns, which allows space for the widest chiplet component (the hash chiplet) and a column to select for it.
+The result is an execution trace of 21 trace columns, which allows space for the widest chiplet component (the hash chiplet, at 20 internal columns) plus the shared chiplet selector prefix.
 
 ![chiplets](../../img/design/chiplets/chiplets.png)
```
bobbinth, line 28: We should also update this image, right?

Al-Kindi-0: This might change in the near future, so I created an issue to track it https://github.com/0xMiden/miden-vm/issues/2967

processor/src/trace/chiplets/hasher/mod.rs
Al-Kindi-0: Exactly, it is a no-std issue. Not sure HashMap works in no-std though. hashbrown is an extra dependency

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
Al-Kindi-0: Does not hurt to have an assertion here. 
And yes, we always have at least one hash.

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
Al-Kindi-0: makes sense, changed

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
Al-Kindi-0: @Nashtare , indeed. This is pretty standard stuff that is better done systematically once the buses are refactored.

processor/src/trace/chiplets/aux_trace/hasher_perm.rs
Al-Kindi-0: Agreed, this is not sound as is but should be easier to fix once the buses are treated uniformly

Al-Kindi-0:
I am merging this so we can conclude #2856 and #2962 and proceed with the multi-table migration.

  I will comment in the referenced PRs with my take on the approaches there, but broadly: we should avoid changes that are not clear wins or do not have major perf improvements.

  There are many opportunities for factoring out and centralizing computations, but it is plausible that once we have automated witness generation for the auxiliary trace, these optimizations could be handled on the backend side. It may therefore make more sense to prioritize changes that improve auditability, readability, and soundness (e.g., domain separation). The same applies to constraints — particularly auxiliary constraints — where simplifications from the unified bus architecture would create easier optimization opportunities on the backend side later.

  In other words, we should prioritize reaching the multi-table migration as soon as possible. The work in the referenced PRs should take this into account.

  Each PR from here through the multi-table milestone should justify its existence with this goal in mind:
  - Performance work is likely lower priority given that things will change soon. If pursued, it should include benchmarks demonstrating the gain — especially if readability is compromised.
  - Readability/auditability work should be discussed and consensus reached, unless the changes are clearly and objectively superior. Changes that add abstraction layers without materially improving understanding for someone familiar with the constraint system do not meet this bar.

PR #2930, opened 2026-03-28
ci: add workflow to verify all PR commits are signed

huitseeker:
Checks each commit in a PR via the GitHub API for signature verification. Fails the check and posts a remediation comment when unsigned commits are found.

### DIFF
--- a/.github/workflows/signed-commits.yml
+++ b/.github/workflows/signed-commits.yml
@@ -1,0 +1,79 @@
+# Verifies that all commits in a PR are signed (GPG or SSH).
+# Posts a comment with remediation steps if unsigned commits are found.
+
+name: signed commits
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  pull_request_target:
+    branches:
+      - main
+      - next
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-signed-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for unsigned commits
+        id: check
+        uses: actions/github-script@v7
+        with:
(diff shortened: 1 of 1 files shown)


bobbinth:
A couple of questions:

- With this change, do we still need the repo-level setting that requires signed commits?
- We discussed at some point whether it makes sense to require external contributors to have signed commits. I think there is a good argument that for external contributors, requiring this doesn't really have much benefit.

huitseeker:
> With this change, do we still need the repo-level setting that requires signed commits?

Yes. This does detection and guidance, not enforcement.

> We discussed at some point whether it makes sense to require external contributors to have signed commits. I think there is a good argument that for external contributors, requiring this doesn't really have much benefit

I'm not aware of that argument — or discussion. Care to refresh my memory?

bitwalker:
> I'm not aware of that argument — or discussion. Care to refresh my memory?

I'm not sure it makes sense to remove the requirement; but the argument AIUI is that unless one of the Miden team members has signed the commit, then what are you getting out of the signature? Not a great deal.

Signatures provide two things: 1. A cryptographic guarantee that the commit cannot be modified without the signer re-signing the commit; 2. Assuming you trust the signer - the ability to trust that the code at that commit has been signed off on by the signer.

IMO, the primary value is really in 1, but alone 1 doesn't mean much, you also need the trust element of 2. In our case, by forcing commits to be signed, the cryptographic guarantee is upheld, and because it is mostly Miden team members signing commits, we largely get the trust benefit as well.

Of course, is anyone actually validating the signatures? Who knows, but at least if someone _wants_ to, they can.

PR #2932, opened 2026-03-28
docs(assembly): document enum variant uniqueness constraint

### LINKED ISSUE #2816: Document that enum variants are module-level constants and must be unique
The documentation describes `enum` as syntactic sugar for a type alias plus a set of const declarations. For example, `enum Level : u8 { ... }` expands into `type Level = u8` and `const DEBUG = 1` style definitions in the same module namespace.

This is enforced by both parsing and semantic analysis:

- The parser requires enum variants to be `ConstantName` tokens via `EnumVariant`
- During semantic analysis, enum variants are pre-registered as constants in the module by the `Form::Enum` handler calling `AnalysisContext::define_constant`
- The module enforces global uniqueness of constant names by scanning existing `Export::Constant` items in `Module::define_constant` and returning `SemanticAnalysisError::SymbolConflict` on duplicates

As a result, defining two enums such as `enum a: u8 { OK }` and `enum b: u8 { OK }` in the same module is rejected, because `OK` is a shared unqualified constant name.

This appears intentional but is not stated explicitly in the documentation.

Recommendation: add an explicit note to the enum documentation stating that enum variants are module-level const names and must be unique within a module.

giwaov:
## Description

This PR adds a note to the **Enum types** section of the assembly documentation (docs/src/user_docs/assembly/code_organization.md) clarifying that:

- Enum variants are expanded into module-level constants during compilation.
- As a result, variant names must be unique across all enums (and constants) within the same module.
- Two enums in the same module cannot define a variant with the same name.

The note includes a concrete example showing the conflict and suggests workarounds (separate modules or distinct variant names).

Closes #2816

### DIFF
--- a/.github/workflows/signed-commits.yml
+++ b/.github/workflows/signed-commits.yml
@@ -1,0 +1,79 @@
+# Verifies that all commits in a PR are signed (GPG or SSH).
+# Posts a comment with remediation steps if unsigned commits are found.
+
+name: signed commits
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  pull_request_target:
+    branches:
+      - main
+      - next
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-signed-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for unsigned commits
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prNumber = context.payload.pull_request.number;
+            const commits = await github.rest.pulls.listCommits({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber,
+              per_page: 250
+            });
+
+            const unsigned = commits.data.filter(c => !c.commit.verification.verified);
+
+            if (unsigned.length === 0) {
+              core.setOutput('unsigned', '');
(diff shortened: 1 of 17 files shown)


docs/src/user_docs/assembly/code_organization.md
```diff
@@ -384,6 +384,15 @@ end
 
 Note that nothing currently validates that a `u8` value passed as an instance of `Level` is actually one of those three constants, that is up to you. Instead, the `enum` syntax is intended to provide more semantic information for readers of the code, and to better express the relationship between the type and the constants in question.
 
+Because enum variants are expanded into module-level constants, variant names must be unique across all enums (and constants) within the same module. For example, the following will produce a compile-time error because `OK` is defined by both enums:
+
+```
+enum Status : u8 { OK, ERR }
+enum Health : u8 { OK, CRIT } # error: duplicate definition of `OK`
```
huitseeker: I think the inline error text here is a little too specific. 

The enum-name collision is real, but the assembler reports the generic `symbol conflict: found duplicate definitions of the same name` diagnostic for this case rather than `duplicate definition of OK`. Could we make this wording less exact so the docs stay aligned with the real output?

huitseeker (approved):
LGTM.

Issue #2292, opened 2025-10-15
Refactor MastForest deserialization to TypeState pattern


huitseeker:
### Feature description

Note: this is about the winterfell `Serializable`/`Deserializable` traits, not about #1962, which should progress on a parallel track.

----
The current MastForest deserialization code uses sequential knowledge of the serialized format and function names to track which part of the data we're reading. This can lead to mistakes where someone calls functions in the wrong order or skips steps accidentally.

The deserialization happens in a specific sequence:
1. Read magic bytes and version
2. Read metadata (node count, decorator count)
3. Read roots
4. Read basic block data
5. Read node infos
6. Read advice map
7. Read error codes
8. Read decorator data
9. Read string table
10. Read decorator infos
11. Read basic block decorators
12. Read before/after decorators

Right now the code relies on developers calling these functions in the right order. If someone calls `read_roots()` before reading the metadata, the code will fail at runtime, not compile time.

We can use the typestate pattern to fix this. The typestate pattern tracks the reading stage using types instead of function names.

Here's how it would work:

```rust
// Define states as empty types
struct Initial;
struct MagicValidated;
struct MetadataRead;
struct RootsRead;
// ... and so on

// The deserializer holds the current state as a type parameter
struct MastForestDeserializer<R, S> {
    reader: R,
    state: PhantomData<S>,
}

// Each state only has methods that make sense for that state
impl<R: ByteReader> MastForestDeserializer<R, Initial> {
    pub fn new(reader: R) -> Self { /* ... */ }

    pub fn read_magic_and_version(self) -> Result<MastForestDeserializer<R, MagicValidated>, DeserializationError> {
        // validate magic and version
        // return new deserializer in MagicValidated state
    }
}

impl<R: ByteReader> MastForestDeserializer<R, MagicValidated> {
    pub fn read_metadata(self) -> Result<MastForestDeserializer<R, MetadataRead>, DeserializationError> {
        // read node count and decorator count
        // return new deserializer in MetadataRead state
    }
}
```

The usage would look like:

```rust
let mast_forest = MastForestDeserializer::new(reader)
    .read_magic_and_version()?
    .read_metadata()?
    .read_roots()?
    .read_basic_block_data()?
    .read_node_infos()?
    .read_advice_map()?
    .read_error_codes()?
    .read_decorator_data()?
    .read_string_table()?
    .read_decorator_infos()?
    .read_basic_block_decorators()?
    .read_before_decorators()?
    .read_after_decorators()?
[...]

PR #2933, opened 2026-03-28
fix(assembly): validate procedure roots during library deserialization

### LINKED ISSUE #2831: Library deserialization accepts uncallable procedure exports
When deserializing a `Library`, the loader accepts procedure exports that are not procedure roots. These exports cannot be invoked via `call` or `exec` because the VM only permits entry at procedure roots.

This is tracked in related issues:
- #2726 — feat: MAST forest deserialization
- #2292 — Refactor MastForest deserialization to TypeState pattern

The deserializer should validate that each exported `MastNodeId` is a procedure root (has no parent in the forest). Non-root nodes should be rejected at load time rather than failing silently at invocation.

Fix: during library deserialization, check that exported procedure node IDs are roots in their respective `MastForest`.

giwaov:
## Description

This PR fixes an issue where \Library::read_from\ accepts procedure exports whose \MastNodeId\ is not a procedure root in the underlying MAST forest. This allows crafting a library with uncallable procedure exports, violating the invariant enforced by \Library::new()\.

This was identified as finding 19 in the audit report.

### Fix
- Added a validation loop to the \Deserializable\ implementation for \Library\ (in \crates/assembly-syntax/src/library/mod.rs\) that checks each procedure export references a valid procedure root in the \MastForest\, matching the existing check in \Library::new()\.

### Test
- Added \library_deserialization_rejects_non_root_export\ test that constructs a tampered serialized library with an export pointing to a non-root node and verifies deserialization rejects it with an appropriate error message.

## Testing

- \cargo test -p miden-assembly library_deserialization_rejects_non_root_export\ passes
- \cargo test -p miden-assembly library_serialization\ still passes (no regression)
- Both \miden-assembly\ and \miden-assembly-syntax\ crates compile cleanly

Closes #2831

### DIFF
--- a/crates/assembly-syntax/src/library/mod.rs
+++ b/crates/assembly-syntax/src/library/mod.rs
@@ -648,6 +648,16 @@ impl Deserializable for Library {
             exports.insert(path, export);
         }
 
+        for export in exports.values() {
+            if let LibraryExport::Procedure(ProcedureExport { node, path, .. }) = export
+                && !mast_forest.is_procedure_root(*node)
+            {
+                return Err(DeserializationError::InvalidValue(format!(
+                    "invalid export: no procedure root for {path} procedure"
+                )));
+            }
+        }
+
         let digest =
             mast_forest.compute_nodes_commitment(exports.values().filter_map(|e| match e {
                 LibraryExport::Procedure(e) => Some(&e.node),
--- a/crates/assembly-syntax/src/library/mod.rs
+++ b/crates/assembly-syntax/src/library/mod.rs
@@ -648,23 +648,8 @@ impl Deserializable for Library {
             exports.insert(path, export);
         }
 
-        for export in exports.values() {
-            if let LibraryExport::Procedure(ProcedureExport { node, path, .. }) = export
-                && !mast_forest.is_procedure_root(*node)
-            {
-                return Err(DeserializationError::InvalidValue(format!(
-                    "invalid export: no procedure root for {path} procedure"
-                )));
-            }
-        }
-
-        let digest =
-            mast_forest.compute_nodes_commitment(exports.values().filter_map(|e| match e {
-                LibraryExport::Procedure(e) => Some(&e.node),
-                LibraryExport::Constant(_) | LibraryExport::Type(_) => None,
-            }));
-
-        Ok(Self { digest, exports, mast_forest })
+        Self::new(mast_forest, exports)
+            .map_err(|err| DeserializationError::InvalidValue(format!("{err}")))
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Fixed debug-only underflow in memory range-check trace generation when the first memory access is at `clk = 0` ([#2976](https://github.com/0xMiden/miden-vm/pull/2976)).
 - Replaced unsound `ptr::read` with safe unbox in panic recovery, removing UB from potential double-drop ([#2934](https://github.com/0xMiden/miden-vm/pull/2934)).
+- Library deserialization now rejects exports whose `MastNodeId` is not a procedure root, closing a silent-failure path ([#2933](https://github.com/0xMiden/miden-vm/pull/2933)).
 - Reverted `InvokeKind::ProcRef` back to `InvokeKind::Exec` in `visit_mut_procref` and added an explanatory comment (#2893).
 - Fixed the release dry-run publish cycle between `miden-air` and `miden-ace-codegen`, and preserved leaf-only DAG imports with explicit snapshots ([#2931](https://github.com/0xMiden/miden-vm/pull/2931)).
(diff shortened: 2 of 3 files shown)


amathxbt:
## Review: PR #2933 — Validate procedure roots during library deserialization

**Overall: 🟡 Correct logic, two CI blockers need fixing.**

Fixes issue #2831.

---

### Core Logic — ✅ Correct

The validation loop added in `crates/assembly-syntax/src/library/mod.rs`:

```rust
for export in exports.values() {
    if let LibraryExport::Procedure(ProcedureExport { node, path, .. }) = export
        && !mast_forest.is_procedure_root(*node)
    {
        return Err(DeserializationError::InvalidValue(format!(
            "invalid export: no procedure root for {path} procedure"
        )));
    }
}
```

**Placement is correct** — happens after all exports are read, before `compute_nodes_commitment`, so a malformed input is rejected before the digest is computed. The use of `if let ... && ...` (let-chains) is fine since the project targets edition 2024 / Rust ≥ 1.90 where let-chains are stable.

---

### CI Failures — 🔴 Must Fix Before Merge

**1. Missing CHANGELOG entry** (job: `changelog`)

The diff contains no `CHANGELOG.md` modification. The CI changelog check fails because there is no entry for this fix. Add an entry under **Bug Fixes** in the current release section, e.g.:

```markdown
- Library deserialization now rejects exports whose `MastNodeId` is not a procedure root, closing a silent-failure path ([#2933](https://github.com/0xMiden/miden-vm/pull/2933)).
```

**2. Nightly rustfmt failures** (job: `rustfmt check nightly on ubuntu-latest`)

Two issues in `crates/assembly/src/tests.rs`:

**(a) Split `use` statements need merging** (reported at line ~412).

Current:
```rust
use miden_core::mast::{BasicBlockNodeBuilder, MastForestContributor};
use miden_core::serde::ByteWriter;
```

nightly rustfmt expects:
```rust
use miden_core::{
    mast::{BasicBlockNodeBuilder, MastForestContributor},
    serde::ByteWriter,
};
```

**(b) `assert!` macro needs reformatting** (reported at line ~455).

Current:
```rust
assert!(result.is_err(), "deserialization should reject exports referencing non-root nodes");
```

nightly rustfmt breaks this onto multiple lines:
```rust
assert!(
    result.is_err(),
    "deserialization should reject exports referencing non-root nodes"
);
```

---

### Test Quality — ✅ Good

The test `library_deserialization_rejects_non_root_export` takes the right approach: it builds a valid library, manually serializes tampered bytes (a forest with a non-root node referenced as an export), and asserts both that deserialization fails **and** that the error message contains `"no procedure root"`. Both assertions are necessary and meaningful.

---

### Informational Note (pre-existing, out of scope)

`Library::read_from` uses the **trusted** `MastForest::read_from` deserializer rather than `UntrustedMastForest` (which recomputes digests). This is a pre-existing concern tracked in #2726 / #2292 and is out of scope here, but worth noting for future hardening.

---

**Summary: fix the CHANGELOG entry and run `cargo +nightly fmt` on the test file, then this is ready.**

huitseeker:
@amathxbt Please cease spam comments.

huitseeker (approved):
🙏

bitwalker (approved):

bitwalker:
@giwaov Can you rebase this and resolve the CHANGELOG conflict? I think we can get this merged once that is done

giwaov:
﻿Rebased onto the latest `next` and force-pushed the branch.

I also reran the regression test locally:
- `cargo test -p miden-assembly library_deserialization_rejects_non_root_export`

That passes now, so this should be ready for another look.

crates/assembly-syntax/src/library/mod.rs
```diff
@@ -648,6 +648,16 @@ impl Deserializable for Library {
             exports.insert(path, export);
         }
 
+        for export in exports.values() {
```
huitseeker: I wonder if this should call `Library::new(mast_forest, exports)` instead of re-checking roots here. The last bug came from deserialization drifting away from the constructor path, and keeping the invariant and digest logic in one place would make the next validation change harder to miss.

giwaov: Updated on the branch now. Library::read_from delegates to Self::new(mast_forest, exports) so the procedure-root invariant and digest computation stay on the constructor path instead of drifting here again.

I also re-ran cargo test -p miden-assembly library_deserialization_rejects_non_root_export locally.

huitseeker: Lol. Good bot

giwaov: Haha, fair enough 😄 Thanks again for the review and for merging this one.

That makes 10 merged PRs on my side now, which I’m really happy about and would love to do more, I also wanted to ask: when Miden eventually goes mainnet, is there likely to be any kind of recognition or reward for contributors, or is it still too early to say? Thank you sir.

PR #2934, opened 2026-03-28
fix(core,assembly): replace unsound ptr::read in panic recovery

### LINKED ISSUE #2814: Latent undefined behaviour in IO panic recovery (ptr::read causes double-drop)

giwaov:
## Description

\Program::write_to_file\ and \Library::write_to_file\ use \catch_unwind\ to convert panics from \ByteWriter\ into \io::Error\. On the downcast success branch, the code uses:

\\\
ust
unsafe { core::ptr::read(&*err) }
\\\

This is unsound: \ptr::read\ performs a bitwise copy of the \std::io::Error\ out of the \Box\, but the original \Box<std::io::Error>\ is then dropped at end of scope, causing a double-drop (undefined behaviour).

## Fix

Replace with \*err\, which safely moves the value out of the \Box\ via \DerefMove\/unbox semantics  no \unsafe\ needed.

## Changes

- \core/src/program/mod.rs\: Replace \unsafe { core::ptr::read(&*err) }\ with \*err\
- \crates/assembly-syntax/src/library/mod.rs\: Same fix

Both changes remove \unsafe\ blocks entirely from these functions.

Closes #2814


huitseeker:
@amathxbt Please cease spam comments.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Rejected non-syscall references to exported kernel procedures in the linker ([#2902](https://github.com/0xMiden/miden-vm/issues/2902)).
 #### Bug Fixes
 
+- Removed unsound `ptr::read` in IO panic recovery that could cause a double-drop ([#2934](https://github.com/0xMiden/miden-vm/pull/2934)).
```
huitseeker: The fix here is smaller than the current wording. This patch removes UB in the `downcast::<std::io::Error>()` arm by replacing `ptr::read` with `*err`, but it does not make normal write failures return `io::Error` today. See [above](https://github.com/0xMiden/miden-vm/pull/2934/changes#r3015035838) for details.

I think the PR title and this changelog line should say that plainly, like `remove UB in panic recovery` or `replace unsound ptr::read in panic recovery`, so readers do not think the full I/O recovery path is fixed.

core/src/program/mod.rs
```diff
@@ -155,13 +155,10 @@ impl Program {
...
         })
-        .map_err(|p| {
-            match p.downcast::<std::io::Error>() {
-                // SAFETY: It is guaranteed to be safe to read Box<std::io::Error>
-                Ok(err) => unsafe { core::ptr::read(&*err) },
-                // Propagate unknown panics
-                Err(err) => std::panic::resume_unwind(err),
-            }
+        .map_err(|p| match p.downcast::<std::io::Error>() {
+            Ok(err) => *err,
```
huitseeker, line 159: The reason this still does not guarantee `io::Error` on normal write failure is that the writer below is not panicking with `io::Error` today. `ByteWriter` for `std::io::Write` uses `write_all(...).expect("write failed")`, so the usual panic payload is the formatted message, not `std::io::Error`, and this branch is skipped. I do not think that wider behavior needs to be fixed in this PR. Reworking that would mean changing how write failures are surfaced, while this patch is already valuable as the smaller safety fix: it removes the UB if the `std::io::Error` branch is ever reached.

huitseeker (commented):
This just needs a small wording change in the PR description & Changelog, otherwise LGTM.

giwaov:
Updated the PR title, commit message, and CHANGELOG wording per your feedback. Now says 'replace unsound ptr::read in panic recovery' to clarify this is the smaller UB fix, not a full I/O recovery path change.

huitseeker (approved):
Thanks! LGTM

PR #2936, opened 2026-03-29
Add kernel support to test builders

lmorett1:
This adds `kernel = ...` support to the test builder macros so kernel-backed tests can be written through the same helper API as other test cases. It removes repeated manual `kernel_source` setup from the syscall and caller tests, keeps the existing execution path unchanged, and makes these tests simpler and less error-prone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Fixes
 - Rejected non-syscall references to exported kernel procedures in the linker ([#2902](https://github.com/0xMiden/miden-vm/issues/2902)).
+- Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
 #### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


miden-vm/tests/integration/flow_control/mod.rs
```diff
@@ -354,12 +341,7 @@ fn call_in_syscall() {
             call.new_ctx
         end";
 
-    let mut test = Test::new(&format!("test{}", line!()), program_source, false);
-    test.kernel_source = Some(test.source_manager.load(
```
Al-Kindi-0, line 358: Would be great if we could avoid the repetitive `kernel= kernel_*`

lmorett1:
> TYVM! Please fix linting in CI, and sign your commits.

 Fixed the lint issue and pushed a signed follow-up commit. Thanks!

miden-vm/tests/integration/flow_control/mod.rs
lmorett1: Nice catch, switched this to .with_kernel(...) to avoid the repeated kernel = ... pattern. Thanks!

Issue #1763, opened 2025-04-15
Investigate row-major trace generation


adr1anh:
### Feature description

We are currently representing a trace in column-major layout (list of vectors of columns) since this is what `winterfell` expects. This is because of the FFTs which act over columns.

This likely has a non-negligible effect on trace generation due to many cache misses. More specifically, traces are constructed by appending a new _row_ for every step of the computation. The memory slots for the elements in a row are not contiguous leading to the CPU having to fetch data from many different addresses.

The [`transpose`](https://docs.rs/transpose/latest/transpose/) provides an optimized implementation for transposing a matrix, using decomposition into smaller blocks to minimize cache-misses. It supports in-place transposition with `max(width, height)` scratch space. Since `winterfell` expects a trace as `Vec<Vec<Felt>>`, we may still need to copy the columns of the transposed matrix to a new vector.

While this may use a bit of extra memory, it should not have too much of an impact since the original row-major trace can be dropped once the proving starts. 

Since this would lead to a major refactor of the VM, we can benchmark this by comparing the time it takes to create a matrix with random data either 
- in row-major order + transposing
- in column-major order. 

Note: We could explore the usefulness of transposition in winterfell too, since row-major order is used both for constraint evaluation and when computing the random linear combination of columns for the DEEP composition polynomial.

### Why is this feature needed?

More efficient trace generation

plafer:
Agreed that this is needed.

In #1642, the claimed improvements of 35-50% were for execution only. The catch is when came time to implement the `Trace` trait required for the prover, I realized that the [`main_segment()` method](https://github.com/facebook/winterfell/blob/6b665a6bde2bb064b2ee33e99b2a285174d1ab97/prover/src/trace/mod.rs#L60) expects the main trace to be represented by `ColMatrix` - which effectively forces us to build the main trace as a `ColMatrix`. I then investigated changing that in winterfell, saw that it was going to be a deep rabbit hole, and opted to kick the can down the road instead. So my current understanding of the problem is that we first need to change the `Trace` API in winterfell (and make all necessary changes to FFTs, etc), and then followed by changes in the VM similar to what was done in #1642.

Related: #1558

Al-Kindi-0:
[...]

PR #2937, opened 2026-03-30
refactor: rework main trace generation to (hybrid) row-major

Nashtare:
## Summary

Rework main-trace generation around a hybrid `MainTrace` design: core and chiplets are now in row-major flat buffers, range checker as two column vectors, with a `MainTrace` transposed view for aux-trace construction so column-oriented reads stay relatively fast.

## Motivation

Column-major trace construction is known to be poorly performing.
See #1763 for full discussion.

## Main changes

- The main trace storage is split by region instead of one big column-major matrix: the system + decoder + stack block is kept as a flat row-major buffer; chiplets are also row-major after construction; the range checker stays as two separate columns (could be changed, see below)

- Chiplets still fill trace columns, then convert to row-major with a fast matrix transpose, and that work is overlapped with other trace work (e.g. core padding) where possible. I looked at making them row-major but that causes poor strided access.

- Auxiliary trace construction no longer does millions of reads through wide strided row-major indexing, now we use the `transposed_view` of the main matrix so that column layout is sequential in memory. (packing aux columns for the prover as before)

## Performance

Output of 
```bash
$ make exec-info
$ MIDEN_LOG=info ./target/optimized/miden-vm prove ./miden-vm/masm-examples/hashing/blake3_1to1/blake3_1to1.masm --release
 ```
 
 averaged over 5 runs, against `next` and this branch:

| Step | `next` | this branch | delta | Note |
|--------|--------|--------|--------|------|
| `build_trace` | 76.6 ms | 94.3 ms | +17.7 ms | _Increase due to chiplet transpose_ |
| `to_row_major_matrix` | 89.2 ms | 18.6 ms | -70.6 ms | _Dominated by allocation now_ |
| `build_aux_trace` | 282 ms | 143 ms | -139 ms | _Dominated by chiplet bus and transpose_ |
| total trace gen. | 447.8 ms | 326.5 ms | -121.3 ms (-27%) |

## Next steps

The design isn't pushed fully, and could deserve a couple follow-up iterations that I haven't done yet, unrelated to the constraint definition cleaning proposed by Adrian that is being helped by the row-major approach:

- Avoid the full aux-path transpose by passing `Parts` into the aux construction if the `AuxTraceBuilders` API can handle it
- Drop the clock scan for last program row if `last_program_row` is threaded into the aux build
- Another potential one: row-major range columns in Parts, chiplet allocation path if it can stay as fast as column fill + transpose, but I doubt it would bring significant improvement, mostly uniformity of the design at best.

- Updated `CHANGELOG.md'

### DIFF
--- a/.github/workflows/signed-commits.yml
+++ b/.github/workflows/signed-commits.yml
@@ -1,0 +1,79 @@
+# Verifies that all commits in a PR are signed (GPG or SSH).
+# Posts a comment with remediation steps if unsigned commits are found.
+
+name: signed commits
+
+permissions:
+  contents: read
+  pull-requests: write
+
+on:
+  pull_request_target:
+    branches:
+      - main
+      - next
+    types: [opened, reopened, synchronize]
+
+jobs:
+  check-signed-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for unsigned commits
+        id: check
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prNumber = context.payload.pull_request.number;
+            const commits = await github.rest.pulls.listCommits({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber,
+              per_page: 250
+            });
+
+            const unsigned = commits.data.filter(c => !c.commit.verification.verified);
+
+            if (unsigned.length === 0) {
+              core.setOutput('unsigned', '');
+              core.info('All commits are signed.');
+              return;
+            }
+
+            const lines = unsigned.map(c =>
+              `- \`${c.sha.slice(0, 8)}\` ${c.commit.message.split('\n')[0]}`
+            );
+
+            core.setOutput('unsigned', lines.join('\n'));
+            core.setOutput('pr_number', String(prNumber));
+            core.setFailed(
+              `Found ${unsigned.length} unsigned commit(s) in this PR. ` +
+              'All commits must be signed (GPG or SSH).'
+            );
+
+      - name: Comment with remediation steps
+        if: failure() && steps.check.outputs.unsigned != ''
+        uses: actions/github-script@v7
+        env:
+          UNSIGNED: ${{ steps.check.outputs.unsigned }}
+          PRNUM: ${{ steps.check.outputs.pr_number }}
+        with:
+          script: |
+            const issue_number = parseInt(process.env.PRNUM, 10);
+            const unsignedList = process.env.UNSIGNED;
+
+            const body = [
+              'This PR contains unsigned commits. All commits must be cryptographically signed (GPG or SSH).',
+              '',
+              'Unsigned commits:',
+              unsignedList,
+              '',
+              'For instructions on setting up commit signing and re-signing existing commits, see:',
+              'https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits'
+            ].join('\n');
+
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number,
+              body
+            });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,14 @@
 
 ## v0.23.0 (TBD)
 
+#### Fixes
+- Rejected non-syscall references to exported kernel procedures in the linker ([#2902](https://github.com/0xMiden/miden-vm/issues/2902)).
 #### Bug Fixes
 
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

File: air/src/trace/main_trace.rs [hunk 1]
```diff
@@ -90,49 +90,398 @@ impl<T> BorrowMut<MainTraceRow<T>> for [T] {
...
+                    } else {
+                        ZERO
+                    }
+                }
+            },
+            TraceStorage::RowMajor(matrix) => {
+                if cfg!(debug_assertions) {
+                    matrix.get(r, col).expect("main trace row/col in bounds")
+                } else {
+                    // SAFETY: indices come from trace iteration and match `matrix` dimensions.
```
huitseeker: That turns any accidental bad `row` or `col` into UB instead of a normal panic, and the same thing happens in the `Transposed` arm below. I don't think that's strictly necessary, see https://github.com/0xMiden/miden-vm/pull/2938

air/Cargo.toml
```diff
@@ -20,7 +20,7 @@ doctest = false
 [features]
 default = ["std"]
 std = ["miden-core/std", "thiserror/std"]
-concurrent = ["std"]
+concurrent = ["std", "rayon"]
```
adr1anh: Still reading through the rest of the PR, but we do re-export `maybe-rayon` through miden-crypto. I think we can use the `par_chunks` from there.

processor/src/trace/chiplets/mod.rs
```diff
@@ -38,7 +38,7 @@ mod tests;
 // ================================================================================================
 
 pub struct ChipletsTrace {
-    pub(crate) trace: [Vec<Felt>; CHIPLETS_WIDTH],
+    pub(crate) trace: Vec<Felt>,
```
adr1anh, line 41: Does it make sense to make this `RowMajorMatrix`? I'd also add a small docstring with the expected dimensions.

processor/src/trace/mod.rs
```diff
@@ -336,20 +335,20 @@ impl ExecutionTrace {
...
-        for col_idx in 0..TRACE_WIDTH {
-            col_major_data.extend_from_slice(self.main_trace.get_column(col_idx));
+        let stored_w = self.main_trace.width();
+        if stored_w == TRACE_WIDTH {
+            return self.main_trace.to_row_major();
         }
-        let col_major_matrix = RowMajorMatrix::new(col_major_data, trace_len);
-        col_major_matrix.transpose()
+
+        assert_eq!(stored_w, PADDED_TRACE_WIDTH);
```
adr1anh, line 350: If nothing breaks too much, we could try removing the padding columns since lifted STARK takes care of virtual zero-padding

processor/src/trace/mod.rs
```diff
@@ -448,33 +459,32 @@ impl<EF: ExtensionField<Felt>> AuxBuilder<Felt, EF> for AuxTraceBuilders {
         // Transpose the row-major main trace into column-major `MainTrace` needed by the
         // auxiliary trace builders. The last program row is the point where the clock
         // (column 0) stops incrementing.
-        let main_trace_col_major = {
+        let main_for_aux = {
             let num_rows = main.height();
-            // Detect last program row from row-major layout using column 0 (clock).
             let last_program_row = (1..num_rows)
                 .find(|&i| {
```
adr1anh: It's definitely more elegant to get the actual last row from the execution, but in the mean time we could try a binary search to find the last active row.

air/Cargo.toml
huitseeker: Yup: https://github.com/0xMiden/crypto/blob/next/miden-crypto/Cargo.toml#L83

processor/src/trace/mod.rs
Nashtare: Would love to do it, but perhaps in a follow up PR?

air/src/trace/main_trace.rs
Nashtare: Merged in [97b78f8](https://github.com/0xMiden/miden-vm/pull/2937/commits/97b78f8ea04fd386ed8051bf23bd4884b5a96205).

air/Cargo.toml
Nashtare: Will remove.

processor/src/trace/mod.rs
Nashtare: Shaving off an extra 5 ms for the aux trace building on average on my machine!

processor/src/trace/chiplets/mod.rs
Nashtare: I think it does, I'll do it in a follow-up PR as it would require also some tweaks in `from_parts` to remove the destructuring there

PR #2938, opened 2026-03-30
perf: make MainTrace::get safe without hidden bounds checks

huitseeker:
Replace the flat Parts backing with typed row storage so `MainTrace::get()` stays fully safe while avoiding compiler-inserted bounds checks on the hot path. Preserve the existing debug layout with borrowing formatters.

(Executable) walkthrough of how this eliminates bounds checks, inspected with [cargo-show-asm](https://github.com/pacak/cargo-show-asm/tree/master):
https://gist.github.com/huitseeker/2b0f6aba4e91c530bc1e3abedfb25e6d

### DIFF
--- a/air/src/trace/main_trace.rs
+++ b/air/src/trace/main_trace.rs
@@ -1,7 +1,10 @@
-use alloc::vec::Vec;
+use alloc::{boxed::Box, vec::Vec};
 use core::{
     borrow::{Borrow, BorrowMut},
+    fmt,
+    mem::ManuallyDrop,
     ops::Range,
+    ptr,
 };
 
 use miden_core::{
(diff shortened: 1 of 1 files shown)


air/src/trace/main_trace.rs
```diff
@@ -241,12 +369,12 @@ impl MainTrace {
                     for i in 0..chunk_rows {
                         let row = start_row + i;
                         let dst = &mut chunk[i * w..(i + 1) * w];
-                        dst[..CORE_WIDTH]
-                            .copy_from_slice(&core_rm[row * CORE_WIDTH..(row + 1) * CORE_WIDTH]);
-                        dst[CORE_WIDTH] = range_checker_cols[0][row];
-                        dst[CORE_WIDTH + 1] = range_checker_cols[1][row];
+                        let row_values = &parts.rows[row];
```
adr1anh, line 372: Just a thought after seeing this pattern in another PR, we could add a helper like

```rust

fn get_n_slices<'a, T, const N: usize>(
    mut data: &'a [T], 
    widths: [usize; N]
) -> [&'a [T]; N] {
    assert!(widths.iter().sum::<usize>() <= data.len());

    widths.map(|w| {
        let (head, tail) = data.split_at(w);
        data = tail;
        head
    })
}

```

Which would help split the row into logical parts

PR #2939, opened 2026-03-30
docs(corelib): document sortedness precondition for sorted array operations

### LINKED ISSUE #2832: Sorted array operations do not validate sortedness
In the standard library, procedures like `std::array::binary_search` and related sorted-array helpers assume the input is sorted but do not validate this precondition. If passed an unsorted array:

1. Binary search may return incorrect results or infinite loop
2. Insertion may produce a still-unsorted array

The documentation states the input must be sorted, so this is expected behavior. However, users may be surprised that invalid input produces silent corruption rather than an error.

Recommendation: either document this more prominently, or add a debug-mode assertion that checks sortedness on entry.

giwaov:
Add prominent warnings to all sorted array procedures (`find_word`, `find_key_value`, `find_half_key_value`, `find_partial_key_value`) clarifying that the input array must be sorted in non-decreasing lexicographic order.

Previously, `find_word` buried the sortedness requirement in a generic crash-conditions list and lacked a dedicated Panics section. The key-value procedures mentioned it under Panics but without emphasis.

## Rationale

Issue #2832 notes that sorted array operations do not clearly communicate the sortedness precondition. Callers passing unsorted input get silently wrong results in the proof (the host event handler validates and returns an error, but this was not documented). This PR makes the contract explicit: sortedness is a bold, caller-ensured precondition on every procedure, with a note that the host event handler enforces it. Panics sections are limited to alignment and range checks only.

## Changes

- Add bold sortedness warning to each procedure's doc comment in the MASM source
- Note that the host event handler validates sortedness and returns an error
- Add a `# Panics` section to `find_word` (matching the style of the other procedures)
- Remove duplicate `Inputs`/`Outputs` block from `find_key_value` doc comment
- Regenerate `sorted_array.md` to match

## Test plan

Docs-only change. CI `check core library docs` job verifies the committed markdown matches the generated output from MASM source. All 15 checks pass on the latest commit.

Fixes #2832


crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -9,14 +9,21 @@ const LOWERBOUND_KEY_VALUE_EVENT = event("miden::core::collections::sorted_array
...
+#! failure. Sortedness is validated by the host-side event handler, but the on-chain verification
+#! only checks ordering locally around the returned pointer.
 #!
 #! Input:  [VALUE, start_ptr, end_ptr]
 #! Output: [is_value_found, value_ptr, start_ptr, end_ptr]
 #!
+#! # Panics
+#!
+#! Panics if:
+#! - words are not sorted in non-decreasing order,
```
huitseeker: The new warning says unsorted input is undefined behavior because proof verification only checks ordering near the returned pointer. Right below that, the new `Panics` list says unsorted input panics. Those read like two different contracts. Can we keep sortedness as a caller-ensured precondition here and limit `Panics` to the alignment and range checks?

huitseeker (changes requested):

crates/lib/core/asm/collections/sorted_array.masm
giwaov: Done . removed the sortedness lines from all Panics sections. Sortedness is now only documented as a caller-ensured precondition in the bold warning, and Panics is limited to the alignment and range checks.

Al-Kindi-0 (commented):
I think that the LOWERBOUND_ARRAY_EVENT event will raise an error if the sorted assumption is not satisfied. Hence, it might make sense to reflect this

giwaov:
Good point. Checked the event handlers in sorted_array.rs and both handle_lowerbound_array and handle_lowerbound_key_value validate sortedness by scanning the entire array, returning SortedArrayError::NotAscendingOrder if any element is smaller than its predecessor. So it is correct to list sortedness under Panics.

Updated: removed the UB language, kept the bold precondition, and added sortedness back to all Panics sections.

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -9,14 +9,18 @@ const LOWERBOUND_KEY_VALUE_EVENT = event("miden::core::collections::sorted_array
...
-#! - `start_ptr <= end_ptr`
+#! **The input array must be sorted in non-decreasing lexicographic order.**
 #!
 #! Input:  [VALUE, start_ptr, end_ptr]
 #! Output: [is_value_found, value_ptr, start_ptr, end_ptr]
 #!
+#! # Panics
+#!
+#! Panics if:
+#! - words are not sorted in non-decreasing order
```
huitseeker: The contract still reads two ways here. The bold warning makes sortedness a caller precondition, but this `Panics` bullet turns it into a guaranteed crash. Proof verification only checks local ordering around the returned pointer.

Could we keep this the same as the earlier thread and treat sortedness as a precondition, not a Panics case? Since this PR is about making the precondition clearer, I think the simplest fix is to remove the words are not sorted... bullet here and in the generated markdown, and leave Panics for the alignment and range checks. If we want to mention the host event error, maybe that fits better in the warning text than in Panics.

huitseeker (commented):
I think this is close — I left an inline comment.

giwaov:
The latest push (de206e3) already addresses this. Sortedness is only in the bold warning as a caller-ensured precondition, with a note that the host event handler validates it and returns an error. The Panics sections are limited to alignment and range checks.

It looks like the review was based on commit 37fdf39, which still had the sortedness bullet in Panics. That was removed in the subsequent force push.

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -176,13 +177,14 @@ end
 #! Half-key means that, out of the keys in the array, only half of the key - the most significant
 #! element (prefix) and the second most significant element (suffix) - need to match.
 #!
+#! **The keys in the array must be sorted in non-decreasing lexicographic order.** The host event handler validates this and will return an error if the keys are not sorted.
```
huitseeker: Nit: I think this is a bit stronger than what `find_half_key_value` checks today. 

In `push_lowerbound_result()`, the `KeySize::Half` path zeroes `word[0]` and `word[1]` before both comparison and the `NotAscendingOrder` check, so an array that is only ordered by the compared half-key still passes. 

Could we reword this to match the current validation? This could be as simple as saying "The half-keys in the array ..."

huitseeker (approved):
Thank you very much!

PR #2941, opened 2026-03-30
docs(corelib): document non-overlap requirement for memory copy and AEAD procedures

### LINKED ISSUE #2835: Core lib procedures can corrupt data when input/output memory ranges overlap
Several core library procedures take memory addresses for input and output buffers. If these ranges overlap, the operations can corrupt data.

For example, hashing procedures that read from one address and write to another assume non-overlapping ranges. Overlapping ranges produce undefined results.

Related to #2390 (Investigate if we should disallow memory operations when (ctx, adr, clk) is the same).

This is a documentation and API design issue rather than a bug — the procedures work correctly for non-overlapping ranges. However, users may not expect this restriction.

Recommendation:
1. Document the non-overlap requirement prominently
2. Consider adding debug-mode assertions that detect overlapping ranges
3. Consider providing "in-place" variants where applicable

giwaov:
Closes #2835

## Summary

Several core library procedures accept separate source and destination memory pointers. If the pointed-to ranges overlap, the forward-copy loops silently corrupt data. This PR adds explicit **must not overlap** warnings to the affected procedures.

### Changes

**`mem.masm`**
- `memcopy_words`: added warning that source `[read_ptr, read_ptr + 4*n)` and destination `[write_ptr, write_ptr + 4*n)` ranges must not overlap.
- `memcopy_elements`: added warning that source `[read_ptr, read_ptr + n)` and destination `[write_ptr, write_ptr + n)` ranges must not overlap.

**`crypto/aead.masm`**
- `encrypt`: strengthened existing 'MUST be different' to 'MUST point to non-overlapping memory ranges'.
- `decrypt`: same strengthening.

**Auto-generated docs** (`mem.md`, `crypto/aead.md`) updated via `MIDEN_BUILD_LIB_DOCS=1 cargo build -p miden-core-lib`.


adr1anh (approved):

Al-Kindi-0 (changes requested):
Looks good, thank you
I would actually do a bit more here and add some assertions inside the procedures to enforce the requirement. This would cost a few cycles but would be even stronger than just a comment.

giwaov:
Added runtime overlap assertions to all four procedures (memcopy_words, memcopy_elements, encrypt, decrypt). Each checks that the source and destination ranges do not overlap using u32gte, and panics with a clear error message if they do. Updated the docs to reflect that overlap now causes a panic instead of silent corruption.

Al-Kindi-0 (commented):
Looks great
I would add a couple of tests to check the unhappy path and make sure that the recent guards do their job

giwaov:
Added unhappy-path tests for all four procedures (memcopy_words, memcopy_elements, encrypt, decrypt). Each test sets up overlapping source/destination ranges and verifies the assertion fires with the expected error message.

All 318 tests pass (318 passed, 0 failed, 1 ignored).

crates/lib/core/tests/mem/mod.rs
```diff
@@ -6,6 +6,48 @@ use miden_utils_testing::{
...
+    // Source [1000, 1000 + 10) = [1000, 1010)
+    // Dest   [1005, 1005 + 10) = [1005, 1015)
+    // These overlap at [1005, 1010).
+    let source = "
+    use miden::core::mem
+
+    begin
+        push.1.2.3.4.1000 mem_storew_be dropw
+        push.5.6.7.8.1004 mem_storew_be dropw
+        push.9.10.11.12.1008 mem_storew_be dropw
```
Al-Kindi-0, line 41: nit: I would use _le variant and push.4.3.2.1 instead here and in other places

Al-Kindi-0 (approved):
LGTM, thank you

PR #2942, opened 2026-03-30
perf(assembly): optimize call graph toposort from O(V*E) to O(V+E)

### LINKED ISSUE #2830: Call graph topological sort scales superlinearly
In `crates/assembly-syntax/src/sema/passes/toposort.rs`, the `toposort` implementation uses nested loops that result in O(n²) or worse complexity for large call graphs. Specifically:

- The visitor pattern traverses each node
- For each node, it may scan the entire graph to find dependencies
- The cycle detection adds another layer of scanning

For projects with many procedures and complex call graphs, this can significantly slow down compilation.

Fix: replace with a standard Kahn's algorithm or DFS-based toposort that maintains O(V + E) complexity. Consider using a more efficient data structure (e.g., `IndexSet` or `FxHashMap`) for tracking visited nodes.

giwaov:
Closes #2830

## Summary

The `CallGraph::toposort()` and `toposort_caller()` implementations used nested loops that resulted in O(V*E) complexity for large call graphs. This PR replaces them with proper O(V+E) Kahn's algorithm.

### Problem

- `num_predecessors()` scanned **all** nodes/edges to count predecessors for a single node, and was called inside the main Kahn's loop for every edge removal  O(E) per call, O(V*E) total.
- `toposort()` cloned the entire graph just to destructively remove edges during the sort.
- Cycle detection used `Vec::contains()` (O(n) per check).
- `reverse_reachable()` scanned all edges per BFS step  O(V*E).

### Fix

- **Pre-compute in-degree map** at the start of each sort. Decrementing an in-degree counter is O(1) vs. O(E) for `num_predecessors()`.
- **Eliminate graph clones**  the in-degree map tracks state instead of destructive edge removal.
- **Use `BTreeSet`** for visited-node lookups in cycle detection  O(log n) vs. O(n).
- **Build reverse adjacency map** in `reverse_reachable()`  single O(V+E) pass instead of O(V*E) scanning.

### Testing

All 9 existing callgraph unit tests pass unchanged. Full `miden-assembly` test suite (180 tests) passes with 0 failures.


crates/assembly/src/linker/callgraph.rs
```diff
@@ -93,64 +93,59 @@ impl CallGraph {
...
             }
         }
 
-        let has_cycle = graph
-            .nodes
-            .iter()
-            .any(|(n, out_edges)| !output.contains(n) || !out_edges.is_empty());
-        if has_cycle {
+        // If not all nodes were visited, the remaining nodes participate in cycles
+        if output.len() != num_nodes {
```
huitseeker, line 142: Could we add a small regression test for the whole-graph cycle case with no initial roots? That was the main shape I wanted pinned down around this new cycle path.

```rust
#[test]
fn callgraph_toposort_whole_graph_cycle_without_roots() {
    let graph = callgraph_cycle_without_roots();
    let err = graph.toposort().expect_err(
        "expected topological sort to fail when every node is blocked behind a cycle",
    );
    assert_eq!(err.0.into_iter().collect::<Vec<_>>(), &[A1, A2, A3]);
}
```

huitseeker (commented):
Overall LGTM, though this will need a rebase.

crates/assembly/src/linker/callgraph.rs
giwaov: Done. added `callgraph_toposort_whole_graph_cycle_without_roots` test and rebased onto `next`.

huitseeker (approved):
🙏

giwaov:
Rebased this on latest `next`.

Checked locally:
- `cargo +1.94 test -p miden-assembly callgraph`
- `git diff --check origin/next...HEAD`

Both pass. The full `miden-assembly` run still hits unrelated Windows temp-path failures in `project::tests`.

PR #2948, opened 2026-03-31
Reduce trace build inputs for prove-from-trace

### LINKED ISSUE #2915: Redesign TraceBuildInputs for trace building and prove-from-trace
`TraceBuildInputs` is the right short-term stopgap from #2865, but it is still larger than the data trace building and prove-from-trace really need.

Right now this bundle repeats data. `ProgramInfo` and `StackOutputs` live both in `TraceBuildInputs` and in its execution binding. `prove_from_trace_sync()` also takes `ProvingOptions`, but those options still carry `ExecutionOptions`, which do not matter once execution is done.

The next step is to define the smallest execution record needed to build a trace and prove from it. That record should carry only the data that trace building and proving consume.

This should:
- remove redundant data from `TraceBuildInputs` and its execution binding
- stop threading full `ExecutionOutput` through the prove-from-trace path when only a smaller subset is needed
- remove `ExecutionOptions` from the prove-from-trace option surface
- make `prove_from_trace_sync()` take the minimal bundled trace/proving input instead of several values that are only valid together
- replace the full advice fingerprint with the smallest replay-specific consistency check that still proves the trace inputs came from the same execution
- clean up some of the APIs made public in #2901 

This follows the main review threads in PR #2865:
- proving API shape: https://github.com/0xMiden/miden-vm/pull/2865#discussion_r2970713359
- trace-generation input shape: https://github.com/0xMiden/miden-vm/pull/2865#discussion_r2970783983
- `TraceBuildInputs` redundancy: https://github.com/0xMiden/miden-vm/pull/2865#discussion_r2993048429
- advice fingerprint cost and purpose: https://github.com/0xMiden/miden-vm/pull/2865#discussion_r2993052919

huitseeker:
`TraceBuildInputs` kept more data than trace building and prove-from-trace need. It also repeated execution state, and `prove_from_trace_sync()` still took execution options after execution was already done.

This PR cuts that path down to the post-execution data that trace building actually reads. It removes the duplicated execution binding, replaces the full advice fingerprint with a cheap digest of deferred precompile requests, and splits proof options from execution options. `prove_from_trace_sync()` now takes a single `TraceProvingInput`s bundle, and the trace-proving helper APIs are tighter.

This also updates the prove-from-trace coverage, including deferred precompile requests across fragment boundaries, and refreshes the affected snapshots on top of the row-major trace base.

Fixes #2915


Nashtare (approved):
Not much to say this looks really clean

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -3,44 +3,61 @@
...
         DefaultHost::default().with_source_manager(Arc::new(DefaultSourceManager::default()));
-
-    // Create proving options with Blake3_256 (96-bit security)
-    let options = ProvingOptions::with_96_bit_security(HashFunction::Blake3_256);
-
-    println!("Proving with Blake3_256...");
-    let (stack_outputs, proof) =
-        prove_sync(&program, stack_inputs, advice_inputs, &mut host, options)
-            .expect("Proving failed");
+    let options = ProvingOptions::with_96_bit_security(hash_fn);
```
adr1anh, line 27: A more drastic change that significantly changes the semantics but improves testing speed would be to use `check_constraints` which asserts that each row satisfies the air constraints. The main disadvantage I would foresee is that we might not catch prover bugs (for example min/max trace height bugs).

A much more useful but heavier test would involve adding an extra step here which would also verify the proof with the recursive verifier to add extra coverage for this load-bearing component.

processor/src/host/advice/mod.rs
```diff
@@ -60,6 +58,7 @@ pub struct AdviceProvider {
 
 impl AdviceProvider {
     #[cfg(test)]
+    #[allow(dead_code)]
```
adr1anh: Could we use? 
```suggestion
    #[expect(dead_code)]
```

processor/src/trace/mod.rs
```diff
@@ -46,10 +46,48 @@ pub use utils::{ChipletsLengths, TraceLenSummary};
...
+    trace_output: TraceBuildOutput,
     trace_generation_context: TraceGenerationContext,
     program_info: ProgramInfo,
-    execution_binding: trace_state::ExecutedTraceBinding,
+}
+
+#[derive(Debug)]
+pub(crate) struct TraceBuildOutput {
+    stack_outputs: StackOutputs,
+    final_pc_transcript: PrecompileTranscript,
```
adr1anh: `precompile_transcript_digest` might be a bit clearer.?

processor/src/trace/mod.rs
```diff
@@ -58,82 +96,65 @@ impl TraceBuildInputs {
...
-        &self.execution_output
+    /// Returns deferred precompile requests generated during execution.
+    pub fn precompile_requests(&self) -> &[PrecompileRequest] {
+        &self.trace_output.precompile_requests
     }
 
-    pub fn trace_generation_context(&self) -> &TraceGenerationContext {
-        &self.trace_generation_context
+    /// Returns the final precompile transcript observed during execution.
+    pub fn final_precompile_transcript(&self) -> &PrecompileTranscript {
```
adr1anh, line 119: I think I do prefer `transcript_digest` overall.

processor/src/trace/mod.rs
```diff
@@ -58,82 +96,65 @@ impl TraceBuildInputs {
...
     }
 
+    /// Returns the program info captured for the execution being replayed.
     pub fn program_info(&self) -> &ProgramInfo {
         &self.program_info
     }
 
+    // Kept for mismatch and edge-case tests that mutate replay inputs directly.
     #[cfg(any(test, feature = "testing"))]
     #[allow(dead_code)]
```
adr1anh: `expect`?

processor/src/trace/mod.rs
```diff
@@ -232,22 +261,19 @@ impl ExecutionTrace {
...
-    /// Intended for proof packaging, where requests are serialized into the proof and no longer
-    /// needed in the trace after consumption.
-    pub fn take_precompile_requests(&mut self) -> Vec<PrecompileRequest> {
-        self.advice.take_precompile_requests()
+        &self.precompile_requests
     }
 
-    /// Returns the final precompile transcript after executing all precompile requests.
+    /// Returns the final precompile transcript observed during execution.
     pub fn final_precompile_transcript(&self) -> PrecompileTranscript {
```
adr1anh, line 273: Naming as suggested above

miden-vm/tests/integration/prove_verify.rs
huitseeker: I've added the recursive step! Please look at `assert_recursive_verify`!

prover/src/lib.rs
```diff
@@ -54,41 +75,42 @@ pub async fn prove(
     stack_inputs: StackInputs,
     advice_inputs: AdviceInputs,
     host: &mut impl Host,
-    options: ProvingOptions,
+    execution_options: ExecutionOptions,
+    proving_options: ProvingOptions,
 ) -> Result<(StackOutputs, ExecutionProof), ExecutionError> {
```
bobbinth, line 80: I would probably get rid of the current implementation of this method and expose a single `prove()` method that would be more like the current `prove_from_trace_sync()`.

prover/src/lib.rs
```diff
@@ -54,41 +75,42 @@ pub async fn prove(
...
 /// This is useful when program execution has already happened elsewhere and only trace building
-/// plus proof generation remain.
+/// plus proof generation remain. The execution settings are already reflected in the supplied
+/// `TraceBuildInputs`, so only proof-generation options remain in this API.
 #[instrument("prove_trace_sync", skip_all)]
 pub fn prove_from_trace_sync(
-    trace_inputs: TraceBuildInputs,
-    options: ProvingOptions,
+    inputs: TraceProvingInputs,
 ) -> Result<(StackOutputs, ExecutionProof), ExecutionError> {
```
bobbinth, line 112: Why do we need to return `StackOutputs` from here? I think returning just the `ExecutionProof` should be sufficient.

processor/src/trace/mod.rs
```diff
@@ -46,10 +46,48 @@ pub use utils::{ChipletsLengths, TraceLenSummary};
...
-    execution_binding: trace_state::ExecutedTraceBinding,
+}
+
+#[derive(Debug)]
+pub(crate) struct TraceBuildOutput {
+    stack_outputs: StackOutputs,
+    final_precompile_transcript: PrecompileTranscript,
+    precompile_requests: Vec<PrecompileRequest>,
+    precompile_requests_digest: [u8; 32],
+}
```
bobbinth, line 60: The naming and high-level APIs still feel a bit off to me. I think we should be able to reduce the number of structs and the number of methods/functions we have for executing/proving.

I'll try to write up an issue about this - but since this far from being a top priority, will probably do this further down the line.

prover/src/lib.rs
huitseeker: If we wanted a proof-only variant, we would make callers keep outputs before handing ownership to the prover, which is viable because [`TraceBuildInputs::stack_outputs()`](https://github.com/0xMiden/miden-vm/blob/5427d52e11b4b73c4a971401ca9a2da5ec8a8b8f/processor/src/trace/mod.rs#L108-L110) already exposes them — but this would downgrade the ergonomics of the whole API and make the correspondence between proving outputs and verification API harder to grasp.

Before:

```rust
pub fn prove_from_trace_sync(
    inputs: TraceProvingInputs,
) -> Result<(StackOutputs, ExecutionProof), ExecutionError>;

let (stack_outputs, proof) = prove_from_trace_sync(TraceProvingInputs::new(
    trace_inputs,
    proving_options,
))?;

verify(program_info, stack_inputs, stack_outputs, proof)?;
```

After:

```rust
pub fn prove_from_trace_sync(
    inputs: TraceProvingInputs,
) -> Result<ExecutionProof, ExecutionError>;

let stack_outputs = *trace_inputs.stack_outputs();
let proof = prove_from_trace_sync(TraceProvingInputs::new(
    trace_inputs,
    proving_options,
))?;

verify(program_info, stack_inputs, stack_outputs, proof)?;
```

Returning only `ExecutionProof` would make the prover API a little smaller, but it would push one extra ownership step onto every caller because the verifier still needs `StackOutputs`.

prover/src/lib.rs
huitseeker: Perhaps what we're looking for is mostly a rename and role cleanup:

```rust
  - prove_from_trace_sync(...) -> prove(...)
  - prove(...) -> execute_and_prove(...)
  - prove_sync(...) -> execute_and_prove_sync(...)
```

I think it makes the real split clear. Execution may be async because the host may be async.  The helper APIs still matter because callers like the CLI often want one call that executes and proves. Dropping the helpers entirely would make the surface smaller, but it would also push processor execution plumbing into simple callers for no real gain.

Issue #2946, opened 2026-03-31
Assembly can fail when building a dependency package from source fails to publish


bitwalker:
_Extracted from [a comment](https://github.com/0xMiden/miden-vm/pull/2877#discussion_r3009639784) by @huitseeker in #2877_.

Currently, the project assembler will publish any artifacts it produces when assembling dependencies from source, to the package registry, so that they are made available to dependents via the registry rather than rebuilding from source every time.

However, this requires that all dependencies of that package are also in the registry. Currently, we do not publish pre-assembled `path` dependencies (i.e. the `path` refers to a `.masp` file) to the registry like other source dependencies, which means that attempting to publish the dependent package can never succeed.

We need to do a couple of things differently here:

1. We need to distinguish between the package registry (which has stricter requirements for publishing) from the package cache used by the assembler.
2. We should never publish to the package _registry_ during assembly, only to the _cache_.
3. All artifacts resolved during assembly should go in the cache, regardless of origin.

It may be the case that the registry implementation and the cache implementation are one and the same (mostly during testing and when using the in-memory registry), but they have separate traits/interfaces, and we should in general not assume that they are implemented by the same type.

PR #2950, opened 2026-03-31
Backport/main post v0.22.0 to next

### LINKED ISSUE #2531: Populate .debug_functions during assembly time
### Feature description

The https://github.com/0xMiden/miden-vm/pull/2474 adds `procedure_name` to `DebugInfo`, but these should be generated by the assembler.

This allows debuggers to resolve human-readable procedure names during execution.

huitseeker:
Merges main back into next.

This also keeps the main behavior for preassembled packages with kernel runtime dependencies. Package resolution keeps the kernel requirement, prefers the stored kernel when present, falls back to a matching embedded kernel when needed, and rejects bad or conflicting embedded kernel metadata. See 88af767e8 individually.

Fixes #2531

### DIFF
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -34,15 +34,8 @@ runs:
     - name: Install Rust toolchain
       uses: dtolnay/rust-toolchain@stable
 
-    - name: Cache cargo registry and git index
-      uses: actions/cache@v4
-      with:
-        path: |
-          ~/.cargo/registry
-          ~/.cargo/git
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-        restore-keys: |
-          ${{ runner.os }}-cargo-
+    - name: Cache Rust build artifacts
+      uses: WarpBuilds/rust-cache@v2
 
     - name: Cleanup large tools for build space
       uses: ./.github/actions/cleanup-runner
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,14 +11,14 @@ permissions:
 
 jobs:
   test:
-    name: test on ubuntu-latest
-    runs-on: ubuntu-latest
+    name: test on warp-ubuntu-latest-x64-4x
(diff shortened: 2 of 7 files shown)


crates/project/src/dependencies/graph.rs
```diff
@@ -441,6 +445,60 @@ impl<'a, R: PackageRegistry + ?Sized> ProjectDependencyGraphBuilder<'a, R> {
                 .map_err(|error| Report::msg(error.to_string()))?;
         }
 
+        for node in graph.nodes.values() {
```
bitwalker: We only embed the kernel into executable packages - and executable packages can't be dependencies. In all other cases, we treat the dependency on a kernel like any other dependency - we can select any kernel version so long as it satisfies the version requirements of all the dependents.

crates/project/src/dependencies/graph.rs
```diff
@@ -441,6 +445,60 @@ impl<'a, R: PackageRegistry + ?Sized> ProjectDependencyGraphBuilder<'a, R> {
...
+                        "embedded kernel package '{}@{}' conflicts with dependency metadata from another preassembled package",
+                        kernel_package.name, kernel_version
+                    )));
+                }
+                continue;
+            }
+
+            let record = PackageRecord::new(kernel_version, kernel_requirements);
+            registry
+                .insert_record(kernel_package.name.clone(), record)
```
bitwalker: This is inserting the embedded kernel into  the registry, but I don't think we should be doing that - the kernel by definition would have needed to be a dependency in the graph, so there should already be a node corresponding to what was declared in the package manifest. If there isn't, something else is wrong.

See #2946 for how I propose we address the issue that I believe you are aiming to address here (though correct me if I'm wrong on that point).

huitseeker:
OK, I scaled this back @bitwalker :
- A preassembled package now keeps its runtime kernel requirement in the normal dependency graph, and assembly fails if that kernel is not actually available.
-   If its pinned metadata no longer matches what the graph selected, we reject it instead of silently using drifted metadata.

bitwalker (approved):

PR #2951, opened 2026-03-31
fix(processor): document and stabilise wrapping-clk semantics in Memory trace builder

### LINKED ISSUE #2828: Release trace generation relies on wrapping subtraction for clk=0 memory access

```rust
```

This works correctly because:

amathxbt:
## Summary

Closes #2828.

`Memory::append_range_checks` and `Memory::fill_trace` both initialise `prev_clk` by subtracting one from the first memory-access clock so that the delta for that row comes out as zero.  The subtractions are intentional wrap-around operations, but:

| Location | Old code | Problem |
|---|---|---|
| `append_range_checks` | `clk.as_canonical_u64() - 1` | Integer underflow / panic in debug builds when `clk == 0` |
| `fill_trace` | `clk - ONE` (Felt arithmetic) | Silent field-element wrap to `P - 1`; not obviously intentional |

Neither subtraction is *incorrect* (the resulting delta arithmetic still produces the right value), but both are surprising to readers and the first one **panics in debug builds** when the first memory access happens at clock cycle 0.

## Fix

* Replace `clk.as_canonical_u64() - 1` with `clk.as_canonical_u64().wrapping_sub(1)` to make the wrap-around explicit and prevent debug panics.
* Add detailed comments to both sites explaining *why* the wrap-around is safe and intentional.
* Update `CHANGELOG.md` under `#### Bug Fixes`.

## Testing

Existing memory-trace tests cover the affected code paths.  No behavioural change; this is a correctness-documentation and debug-build stability fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 #### Bug Fixes
 
 - Reverted `InvokeKind::ProcRef` back to `InvokeKind::Exec` in `visit_mut_procref` and added an explanatory comment (#2893).
+- Documented and stabilised the wrapping-subtraction semantics used to initialise `prev_clk` in `Memory::append_range_checks` and `Memory::fill_trace`; replaced the implicit `- 1` with `wrapping_sub(1)` so the intent is explicit and debug builds no longer overflow ([#2828](https://github.com/0xMiden/miden-vm/issues/2828)).
+
 #### Changes
 
 - Documented that enum variants are module-level constants and must be unique within a module (#2932).
--- a/processor/src/trace/chiplets/memory/mod.rs
+++ b/processor/src/trace/chiplets/memory/mod.rs
@@ -250,8 +250,15 @@ impl Memory {
         // set the previous address and clock cycle to the first address and clock cycle of the
         // trace; we also adjust the clock cycle so that delta value for the first row would end
         // up being ZERO. if the trace is empty, return without any further processing.
+        //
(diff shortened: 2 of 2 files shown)


amathxbt:
This has been a latent footgun for a while. The `- 1` on `clk.as_canonical_u64()` works fine in release builds because the subsequent delta arithmetic wraps right back to a sensible value, but in debug builds the subtraction asserts on overflow and the whole thing falls apart if the first memory access ever lands at clock cycle 0. Switching to `wrapping_sub(1)` makes the code do exactly the same thing in both builds while also making the intent obvious to anyone reading it for the first time.

The comment you added in `append_range_checks` does a great job of explaining the full round-trip: subtracting 1 from 0 gives `u64::MAX`, and then computing `clk - prev_clk` wraps back to 1, which is the right delta for that row. Without that comment a reviewer would almost certainly flag the wrap-around as suspicious.

Worth noting that the `fill_trace` side has a slightly different story since that subtraction happens in Felt field arithmetic, so "clk == ZERO" produces `P - 1` rather than `u64::MAX`. The comment you added there captures that correctly. The two paths are technically using different wrap-around mechanisms, which is an interesting asymmetry, but both are sound for the same structural reason.

One tiny nit: the original comment in `fill_trace` spells "pervious" instead of "previous" and this PR already fixes that in the context block, so that typo cleanup is a nice bonus.

All CI checks are green on this one. Clean fix, good comments, nothing to block.

adr1anh:
These comments just add noise and are overly complex.

PR #2954, opened 2026-03-31
refactor(assembly): remove push.a.b.c.d dot-delimited syntax

### LINKED ISSUE #2572: Remove `push.a.b.c.d` syntax

giwaov:
## Description

Removes the dot-delimited multi-value push syntax (e.g., `push.1.2.3.4`) from the MASM parser, as proposed in #2572.

Users should now use individual push instructions (`push.1 push.2 push.3 push.4`) or word literals (`push.[1,2,3,4]`).

## Changes

**Parser (grammar.lalrpop):**
- Removed the `DotDelimited` generic helper macro (only used by the push rule)
- Replaced the multi-value push rule with a single-value rule: `push.IMM`

**Error handling (error.rs):**
- Removed the `PushOverflow` error variant (no longer reachable since the parser only accepts one value)

**Display (print.rs):**
- Updated `PushFeltList` display to emit space-separated individual pushes (`push.3 push.4 push.8`) instead of dot-delimited (`push.3.4.8`)

**MASM files (26 files):**
- Converted all dot-delimited push patterns to individual push instructions

**Test files (~24 files):**
- Updated all static string literals in Rust tests
- Updated programmatic format-string generators (e.g., `format!("push.{}.{}", a, b)` to `format!("push.{} push.{}", a, b)`)
- Rewrote the `push_many` integration test to use the new syntax

## Testing

- All assembly-syntax tests pass (104)
- All assembly tests pass (180)
- All processor tests pass (1488)
- All core-lib tests pass (326)
- All integration tests pass (350)
- Clippy clean (zero warnings)

Closes #2572

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [BREAKING] Removed the deprecated unbound `TraceBuildInputs::new()` and `TraceBuildInputs::from_program()` constructors; use `execute_trace_inputs_sync()` or `execute_trace_inputs()` instead ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
 - Added `prove_from_trace_sync(...)` for proving from pre-executed trace inputs ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
 - Refactor trace generation to row-major format ([#2937](https://github.com/0xMiden/miden-vm/pull/2937)).
+- [BREAKING] Removed the `push.a.b.c.d` dot-delimited multi-value push syntax; use individual `push` instructions or word literals instead ([#2572](https://github.com/0xMiden/miden-vm/issues/2572)).
 
 ## 0.22.0 (2025-03-18)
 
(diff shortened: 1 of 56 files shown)


Al-Kindi-0:
There is already an open PR for this here https://github.com/0xMiden/miden-vm/pull/2900 , hence I will just close this one now.
Again, please ask and wait for confirmation of an assignment before expending resources, on both your side and ours. Thank you again for your contributions

PR #2955, opened 2026-04-01
fix: improve debug var loc tracking

djolertrk:
This adds var location type to represent `DW_OP_fbreg`, that we were missing in the initial implementation. It is used by both compiler and debugger later to print RUST variables values at certain program point.

Also, fix a bug in registration of variables per node.

Rationale: Blocks with identical operations but different debug variables were being deduplicated, causing crash in compiler and incorrect variable scoping in the debugger

Fixes https://github.com/0xMiden/miden-vm/issues/2961

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -162,6 +162,21 @@ impl BasicBlockNode {
             })
             .collect()
     }
+
+    /// Adjusts padded operation indices back to raw indices for AssemblyOp mappings.
+    pub fn unadjust_asm_op_indices<T: Copy>(
+        asm_ops: Vec<(usize, T)>,
+        op_batches: &[OpBatch],
+    ) -> Vec<(usize, T)> {
+        let pad2raw = PaddedToRawPrefix::new(op_batches);
+        asm_ops
+            .into_iter()
+            .map(|(padded_idx, id)| {
+                let raw = padded_idx - pad2raw[padded_idx];
+                (raw, id)
+            })
+            .collect()
+    }
 }
 
 // ------------------------------------------------------------------------------------------------
--- a/crates/assembly/src/mast_forest_builder.rs
+++ b/crates/assembly/src/mast_forest_builder.rs
@@ -285,9 +285,13 @@ fn serialize_asm_ops_for_node(
 
 /// Looks up and serializes the asm ops registered for a node in an existing forest.
 fn serialize_asm_ops_from_forest_node(forest: &MastForest, node_id: MastNodeId) -> Vec<u8> {
-    let entries = forest.debug_info().asm_ops_for_node(node_id);
+    let mut entries = forest.debug_info().asm_ops_for_node(node_id);
     if entries.is_empty() {
         return Vec::new();
+    }
+
+    if let MastNode::Block(block) = &forest[node_id] {
+        entries = BasicBlockNode::unadjust_asm_op_indices(entries, block.op_batches());
     }
 
     let mut data = Vec::new();
@@ -927,7 +931,10 @@ impl MastForestBuilder {
         self.node_id_by_fingerprint.insert(dedup_fingerprint, new_node_id);
         self.hash_by_node_id.insert(new_node_id, dedup_fingerprint);
 
-        let asm_ops = self.statically_linked_mast.debug_info().asm_ops_for_node(source_node_id);
+        let mut asm_ops = self.statically_linked_mast.debug_info().asm_ops_for_node(source_node_id);
+        if let MastNode::Block(block) = &self.statically_linked_mast[source_node_id] {
+            asm_ops = BasicBlockNode::unadjust_asm_op_indices(asm_ops, block.op_batches());
+        }
         if !asm_ops.is_empty() {
             let mut remapped_asm_ops = Vec::with_capacity(asm_ops.len());
             for (op_idx, asm_op_id) in asm_ops {
@@ -1880,6 +1887,69 @@ mod tests {
         let local_vars = forest.debug_info().debug_vars_for_node(local_block_id);
         assert_eq!(forest.debug_info().debug_var(copied_vars[0].1).unwrap().name(), "x");
         assert_eq!(forest.debug_info().debug_var(local_vars[0].1).unwrap().name(), "y");
+    }
+
+    #[test]
+    fn test_statically_linked_padded_block_dedups_with_equivalent_local_block() {
+        let mut source_builder = MastForestBuilder::new(&[]).unwrap();
+        let ops = vec![
+            Operation::Push(Felt::new(1)),
+            Operation::Drop,
+            Operation::Drop,
+            Operation::Drop,
+            Operation::Drop,
+            Operation::Drop,
+            Operation::Drop,
+            Operation::Push(Felt::new(2)),
+            Operation::Push(Felt::new(3)),
+        ];
+        let asm_op = AssemblyOp::new(None, "padded_ctx".into(), 1, "push.3".into());
+
+        let static_block = source_builder
+            .ensure_block(
+                ops.clone(),
+                Vec::new(),
+                vec![(8, asm_op.clone())],
+                vec![],
+                vec![],
+                vec![],
+            )
+            .unwrap();
+        source_builder.mast_forest.make_root(static_block);
+
+        let (static_forest, _) = source_builder.build();
+        let expected_padded_idx = static_forest.debug_info().asm_ops_for_node(static_block)[0].0;
+
+        let mut builder = MastForestBuilder::new([&static_forest]).unwrap();
+        let copied_block_id =
+            builder.ensure_external_link(static_forest[static_block].digest()).unwrap();
+        let local_block_id = builder
+            .ensure_block(ops, Vec::new(), vec![(8, asm_op)], vec![], vec![], vec![])
+            .unwrap();
+
+        assert_eq!(
+            copied_block_id, local_block_id,
+            "copied padded blocks should dedup with equivalent local blocks",
+        );
+
+        let (forest, remapping) = builder.build();
+        let final_block_id = remapping.get(&copied_block_id).copied().unwrap_or(copied_block_id);
+
+        assert!(
+            forest
+                .debug_info()
+                .asm_op_for_operation(final_block_id, expected_padded_idx - 1)
+                .is_none(),
+            "the asm op must not be attached before its padded operation index",
+        );
+        assert_eq!(
+            forest
+                .debug_info()
+                .asm_op_for_operation(final_block_id, expected_padded_idx)
+                .unwrap()
+                .context_name(),
+            "padded_ctx",
+        );
     }
 
     /// A small procedure root that gets merged into a larger block must keep its own
--- a/core/src/mast/node_builder_utils.rs
+++ b/core/src/mast/node_builder_utils.rs
@@ -74,9 +74,10 @@ where
             MastNodeBuilder::Call(builder)
         },
         MastNode::Block(basic_block_node) => {
-            // For BasicBlockNode, we need to remap op-indexed decorators as well
-            let builder = BasicBlockNodeBuilder::new(
-                basic_block_node.operations().copied().collect(),
+            // Preserve the stored batches so copied blocks fingerprint the same way as raw-built
+            // equivalents, even when padding NOOPs shifted stored metadata indices.
+            let builder = BasicBlockNodeBuilder::from_op_batches(
+                basic_block_node.op_batches().to_vec(),
                 basic_block_node
                     .indexed_decorator_iter(source_forest)
                     .map(|(idx, decorator_id)| {
@@ -84,6 +85,7 @@ where
                         Ok((idx, mapped_decorator))
                     })
                     .collect::<Result<Vec<_>, _>>()?,
+                basic_block_node.digest(),
             )
             .with_before_enter(before_enter_decorators)
             .with_after_exit(after_exit_decorators);


djolertrk:
cc @bitwalker

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -141,6 +149,21 @@ impl MastForestBuilder {
                 .expect("failed to register AssemblyOps - internal ordering error");
         }
 
+        // Register all pending debug variable mappings in sorted node order.
+        // Like AssemblyOps, the CSR structure requires sequential node registration.
+        let mut debug_var_mappings = core::mem::take(&mut self.pending_debug_var_mappings);
+        debug_var_mappings.sort_by_key(|(node_id, _)| *node_id);
+        // Keep only the first registration per node_id (duplicates come from deduplication).
+        let mut seen_debug_var_node_ids = BTreeSet::new();
+        debug_var_mappings.retain(|(node_id, _)| seen_debug_var_node_ids.insert(*node_id));
```
huitseeker: I think this changes semantics for deduplicated blocks, not just the crash behavior. `add_debug_var` treats debug vars as per-occurrence metadata, but this path now keeps only the first debug-var registration for a `MastNodeId`. Since debug vars are not part of the block fingerprint, two identical blocks from different source scopes can legitimately have different live-variable mappings. After this change they no longer crash, but the debugger can show the first block's vars for every deduplicated copy.

djolertrk: @huitseeker thanks for the comment!
Hmm. Yes. A proper fix would require either (a) including debug var info in the block fingerprint to prevent dedup when vars differ, or (b) supporting multiple debug var registrations per node in the CSR. Please let me know what you think.

djolertrk: Or a third/(c) one can be a fix that merges debug vars from all deduplicated copies instead of keeping only the first. In ensure_block, the is_new guard can be removed for debug vars (kept for asm_ops), so all registrations are collected. In build(), mappings for the same MastNodeId would be merged into a single vector rather than discarded. This way the debugger sees variables from all source scopes that map to the deduplicated block.

huitseeker: Yeah, so rephrasing: either include debug-var metadata in the key when debug info is enabled, skip block deduplication when debug vars are present, or store debug-var mappings per block occurrence rather than per deduplicated MastNodeId. In this case, I'd just use debug vars in the fingerprint, I think. Simple, and lands the proper semantics under all branches.

djolertrk: @huitseeker done, thanks!

djolertrk:
@huitseeker please let me know if there is any other comment.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -141,6 +149,20 @@ impl MastForestBuilder {
...
         }
 
+        // Register all pending debug variable mappings in sorted node order.
+        // The CSR structure requires sequential node registration.
+        // Debug vars are included in the block fingerprint, so blocks with different
+        // debug vars are not deduplicated. The dedup here is only a safety measure.
+        let debug_var_mappings =
+            deduplicate_debug_var_mappings(core::mem::take(&mut self.pending_debug_var_mappings));
+
+        for (node_id, debug_vars) in debug_var_mappings {
```
huitseeker, line 183: Even with the cleaner design where debug vars stay external and only contribute to the dedup helper, I think this build path still needs one more step after registration: `remove_nodes()` has to remap `op_debug_var_storage` the same way it already remaps asm-op storage.

So I would separate the two concerns:

- dedup sensitivity: make the fingerprint helper consult external debug-var mappings
- final forest correctness: remap the debug-var CSR when merged nodes are compacted away

That keeps the metadata external, but still fixes the post-compaction lookup bug. We should probably talk about this offline as well.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1443,10 +1445,25 @@ impl BasicBlockNodeBuilder {
...
+            debug_vars: Vec::new(),
             digest: Some(digest),
         }
     }
 
+    /// Sets the debug variable mappings for this basic block.
+    ///
+    /// Debug vars are included in the block fingerprint, so blocks with identical operations
+    /// but different debug variables will not be deduplicated.
+    pub fn with_debug_vars(mut self, debug_vars: Vec<(usize, DebugVarId)>) -> Self {
```
huitseeker: I think we may not want `debug_vars` to live on `BasicBlockNodeBuilder` at all.

The thing we need for dedup is not the serialized `BasicBlockNode`, it is the dedup key that `ensure_node_exists()` uses before the node is inserted. Those are different concerns. One way to keep them separate would be:

1. Keep the actual debug-var storage external, the same way decorators/debug info already live in `debug_info`.
2. In `ensure_block`, compute the dedup fingerprint from the normal block data plus a separate debug-var contribution passed alongside the block, instead of storing that contribution on the builder itself.
3. For already-inserted nodes, when we need to recompute the cached dedup fingerprint after `append_before_enter` / `append_after_exit`, read the debug-var mappings back from `debug_info` and feed them into the same fingerprint helper.

That avoids making `BasicBlockNode::to_builder()` responsible for round-tripping debug vars just so dedup stays correct, and it keeps the storage model aligned with the rest of the debug metadata. The builder would stay about block structure, while debug-var mappings remain external metadata that still participates in dedup.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -599,23 +638,33 @@ impl MastForestBuilder {
     /// entry is `(op_idx, AssemblyOp)` where `op_idx` is the operation index the AssemblyOp
     /// corresponds to.
     ///
-    /// Note: AssemblyOps are only registered for newly created nodes. If a duplicate node already
-    /// exists in the forest (deduplication), the asm_ops are ignored since the existing node
-    /// already has its metadata registered.
+    /// The `debug_vars` parameter contains debug variable metadata for operations in this block.
```
huitseeker, line 1012: If the intent here is only to make debug vars affect dedup, in line with the [above comment](https://github.com/0xMiden/miden-vm/pull/2955/changes/b1a31d259f13e038a15fc8bf1037af7b18010764#r3028114676), I think there is a cleaner shape than storing them on `BasicBlockNodeBuilder`.

A possible split is:

1. Keep `pending_debug_var_mappings` as the source of truth for the per-node CSR registration path.
2. Add a helper that computes the basic-block dedup fingerprint from `(operations, decorators, before/after decorators, assert data, debug-var mappings)`.
3. Call that helper from `ensure_block` using the incoming `debug_vars`, without moving those mappings onto the block builder itself.
4. When a block already exists and later gets rebuilt through `append_before_enter` / `append_after_exit`, recompute its cached dedup fingerprint by reading the node's debug-var mappings back out of `debug_info` and feeding them into the same helper.

That keeps the debug-var metadata external, like the rest of the debug-info storage, while still making dedup sensitive to it. It also avoids having to thread debug vars through `to_builder()` just to preserve fingerprint semantics.

crates/assembly/src/mast_forest_builder.rs
djolertrk: I see, thank you! I agree that this is the approach! Will try to handle it within this PR.

Let me check how to reflect this to debugger and compiler as well.

djolertrk:
@huitseeker I have addressed the comments:

  - debug_vars no longer lives on `BasicBlockNodeBuilder`
  - `append_before_enter`/`append_after_exit` now look up the node's debug vars from `pending_debug_var_mappings` and re-augment the recomputed fingerprint, so we don't lose the debug-var contribution when rebuilding through the `to_builder()`
  - Added `OpToDebugVarIds::remap_nodes()` with the same pattern as the asm-op one, and wired it into  the `MastForest::remove_nodes()`

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -635,6 +725,13 @@ impl MastForestBuilder {
             self.pending_asm_op_mappings.push((node_id, asm_op_mappings));
         }
 
+        // Only register debug variables for newly created nodes.
+        // Blocks with different debug vars have different augmented fingerprints,
+        // so deduplication only occurs when vars are truly identical.
+        if is_new && !debug_vars.is_empty() {
+            self.pending_debug_var_mappings.push((node_id, debug_vars));
```
huitseeker, line 1081: This pushes debug vars onto the original block id, but `merge_basic_blocks()` only carries `pending_asm_op_mappings` into the merged block. If a mergeable block has debug vars, the old node is removed later and `remove_nodes()` drops that debug-var row instead of moving it to the merged node. A small test that merges two blocks with `push_debug_var()` would catch this.

huitseeker (changes requested):
Thanks for the change! I appreciate that on top of the `DebugVarLocation::FrameBase`/`DebugVarInfo`, you're having to fix a real assembler/debug-info bug around deduplicated blocks.

You may want to read what #2958 is doing as a side quest. Our CONTRIBUTING.md mentions some contribs may end up being rolled up in our PRs.

I think this PR is aiming at the right semantics for deduped debug vars, but the current diff appears to miss at least one path: `merge_basic_blocks()` forwards pending asm-op mappings but not pending debug-var mappings in `mast_forest_builder.rs`.

crates/assembly/src/mast_forest_builder.rs
djolertrk: I see. It now handles both asm_ops and debug_vars. Thanks a lot!

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -839,8 +989,15 @@ impl MastForestBuilder {
...
         let mut decorated_builder = self.mast_forest[node_id].clone().to_builder(&self.mast_forest);
         decorated_builder.append_before_enter(decorator_ids);
-        let new_node_fingerprint =
+        let base_fingerprint =
             decorated_builder.fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)?;
+
+        // Re-augment the fingerprint with this node's debug vars so the dedup key
+        // stays consistent with what ensure_block() originally computed.
+        let debug_vars_data =
+            serialize_debug_vars_for_node(&self.pending_debug_var_mappings, node_id);
```
huitseeker: I think this still leaves one dedup path open. `append_before_enter` now re-augments the fingerprint for an existing node, but the `repeat` path in `assembler.rs` still clones a node with `.to_builder(...).with_before_enter(...)` and sends it through generic `ensure_node()`.

For a repeated basic block that carries debug vars, that route still fingerprints only the block/decorator structure, so it can alias a same-op/same-decorator block with different debug vars. 

A small repro is:
```rust
let block = builder.ensure_block(vec![Operation::Add], Vec::new(), vec![], vec![(0, lhs_var)], vec![], vec![])?;
let rebuilt = builder[block].clone().to_builder(builder.mast_forest()).with_before_enter(vec![before_enter]);
let node = builder.ensure_node(rebuilt)?;
```

After `build()`, `debug_vars_for_node(node)` is empty. But the direct path `ensure_block(..., vec![(0, lhs_var)], vec![before_enter], ...)` keeps the debug var.

I think the fix is to make this rebuilt basic-block path use the same debug-var-aware fingerprint and pending-registration flow as `ensure_block`, instead of only handling it in `append_before_enter` / `append_after_exit`. And perhaps adding a regression test that exercises `repeat` with debug vars.

djolertrk: wow, nice catch!!! yes, it was not handled.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,25 +563,69 @@ impl MastForestBuilder {
...
+    /// Removes the debug_vars from pending_debug_var_mappings and adds them to the merged list
+    /// with indices adjusted by the given offset.
+    fn transfer_debug_vars_for_merge(
+        &mut self,
+        source_block_id: MastNodeId,
+        ops_offset: usize,
+        merged_debug_vars: &mut Vec<(usize, miden_core::mast::DebugVarId)>,
+    ) {
+        let (matched, rest): (Vec<_>, Vec<_>) =
+            core::mem::take(&mut self.pending_debug_var_mappings)
```
huitseeker: The dedup key is still too weak when debug vars differ. Two nodes with the same ops but different debug vars must not collapse to one node. That rule applies to basic blocks and to any other node type that can carry debug vars. The fix is to include serialized debug-var data in the dedup key whenever a node has debug vars. One test should cover two basic blocks with the same ops but different vars. Another should cover a non-basic-block node with the same shape and different vars. Both tests should prove that merge keeps them distinct.

huitseeker: This still looks incomplete.

This helper still removes source mappings too early. A block can be merged into a larger block and still stay alive as a root or as a shared node, and in that case the source block still needs its own asm-op and debug-var metadata. The fix is to read the pending mappings for the merged block here, but leave the source mappings in place until the source node is actually removed. 

A test should cover that when a small root block is folded into a larger block, the small root still remains in the forest, and both nodes keep the right metadata.

Another gap is in the merge path. After `merge_roots()`, the merged forest still needs `procedure_names`, `asm_op_storage`, and `op_debug_var_storage` rebuilt from the remapped node ids. The same rule needs to hold for `compact()`. We should test merging two forests with procedure names, asm-op locations, and debug vars keeps that metadata on the final nodes and that compacting such a forest keeps the metadata matched to the compacted nodes.

djolertrk:
> Besides the issue I pointed to inline, there's a few more gaps. Two nodes with the same ops but different debug vars must not collapse to one node. That rule applies to any node type that can carry debug vars. We should include serialized debug-var data in the dedup key whenever a node has debug vars.
> 
> Ideally, I'd see in a test that two basic blocks with the same ops but different vars stay distinct after merge, and that a non-basic-block node with different vars also stays distinct (or better, that a non-basic block node with debug vars is not representable, but that may be much to ask).
> 
> Finally, two things we should want to tackle in a posterior PR (because this isn't strictly limited to debug vars):
> 
> * metadata from an external placeholder root must not override metadata from the concrete root it points to: If the concrete root has asm-op or debug-var metadata, that concrete metadata should win. Procedure names need the same rule, except that a placeholder name should stay as fallback when the concrete root has no name.
> * determinism. If two real roots give different names for the same digest, the result must not depend on merge order. A test should show that merging the same inputs in both orders picks the same final name.

Hmm got it. I will try to come up with more tests for various cases and to see if something else is not caught with this. Thanks a lot for the help here!

djolertrk:
> > Besides the issue I pointed to inline, there's a few more gaps. Two nodes with the same ops but different debug vars must not collapse to one node. That rule applies to any node type that can carry debug vars. We should include serialized debug-var data in the dedup key whenever a node has debug vars.
> > Ideally, I'd see in a test that two basic blocks with the same ops but different vars stay distinct after merge, and that a non-basic-block node with different vars also stays distinct (or better, that a non-basic block node with debug vars is not representable, but that may be much to ask).
> > Finally, two things we should want to tackle in a posterior PR (because this isn't strictly limited to debug vars):
> > 
> > * metadata from an external placeholder root must not override metadata from the concrete root it points to: If the concrete root has asm-op or debug-var metadata, that concrete metadata should win. Procedure names need the same rule, except that a placeholder name should stay as fallback when the concrete root has no name.
> > * determinism. If two real roots give different names for the same digest, the result must not depend on merge order. A test should show that merging the same inputs in both orders picks the same final name.
> 
> Hmm got it. I will try to come up with more tests for various cases and to see if something else is not caught with this. Thanks a lot for the help here!

Thanks a lot once again @huitseeker!

I tried to add more tests and to cover them all in the code.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -184,6 +206,46 @@ fn deduplicate_asm_op_mappings(
         .collect()
 }
 
+/// Serializes debug variable mappings into bytes for fingerprint augmentation.
+fn serialize_debug_vars(debug_vars: &[(usize, miden_core::mast::DebugVarId)]) -> Vec<u8> {
```
huitseeker: I could reproduce an issue locally with a very small test: two blocks with the same `Operation::Add` and the same `DebugVarInfo::new("x", DebugVarLocation::Stack(0))`, but different allocated `DebugVarId`s, do not dedup. My test fails with `left: MastNodeId(0)` and `right: MastNodeId(1)`.

The root cause looks to be that the dedup bytes are built from the allocated ID here:

```rust
for (op_idx, var_id) in debug_vars {
    data.extend_from_slice(&op_idx.to_le_bytes());
    data.extend_from_slice(&var_id.as_u32().to_le_bytes());
}
```

That makes builder-time dedup depend on allocation order instead of the underlying `DebugVarInfo` payload. `merge()` and `compact()` already compare resolved debug-var content, so the builder and merger now disagree about what counts as the same block.

I think the fix is to serialize `(op_idx, DebugVarInfo)` content here instead of `(op_idx, DebugVarId)`, then add a regression test for same payload plus different ID still deduping to one node. We've [discussed this direction before](https://github.com/0xMiden/miden-vm/pull/2955#discussion_r3028114691).

core/src/mast/merger/mod.rs
```diff
@@ -228,9 +232,15 @@ impl MastForestMerger {
             &self.decorator_id_mappings[forest_idx],
         )?;
 
-        let node_fingerprint =
+        let base_fingerprint =
             remapped_builder.fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)?;
 
+        // Augment with the source node's debug vars so same-ops/different-vars
+        // blocks from different forests are not collapsed.
+        let debug_var_data =
```
huitseeker, line 240: I could reproduce a merge-time source-mapping alias here with two forests that have the same block ops and the same debug-var payload, but different `AssemblyOp` contexts. This fails because forest B's mapped root reads back `ctx_a` instead of `ctx_b`.

The gap seems to be that the dedup key is augmented only with debug vars:

```rust
let debug_var_data =
    serialize_debug_var_content_for_node(original_forests[forest_idx], merging_id);
let node_fingerprint = base_fingerprint.augment_with_data(&debug_var_data);
```

but later the asm-op table is copied with a first-wins rule:

```rust
if let alloc::collections::btree_map::Entry::Vacant(e) = asm_entries.entry(new_id) {
    let ops = forest.debug_info.asm_ops_for_node(old_id);
```

So two nodes can collapse even when their source mappings differ, and the merged node keeps only the first forest's diagnostics metadata. I think you either need asm-op content in the dedup key too, or you need to skip dedup when asm-op metadata differs. A regression test with two identical blocks and different `AssemblyOp` contexts would lock this down.

crates/assembly/src/mast_forest_builder.rs
djolertrk: I created some tests locally that I did not include into PR (from compiler level) and it seems that I have not triggered these cases at all properly, since all seemed well handled for me.

Yes, it seems that we should go with `(op_idx, DebugVarId)` serialisation.

core/src/mast/merger/mod.rs
```diff
@@ -282,6 +296,93 @@ impl MastForestMerger {
...
+    /// into the merged forest, remapping all IDs along the way.
+    ///
+    /// Procedure names are merged separately by digest. Per-node asm-op and debug-var
+    /// metadata are remapped by node ID, and when two source nodes map to the same merged
+    /// node (dedup), the first forest's per-node metadata wins.
+    fn merge_debug_metadata(&mut self, forests: &[&MastForest]) -> Result<(), MastForestError> {
+        // Procedure names are keyed by digest, so no node-ID remapping is needed here.
+        // Note: later inserts currently overwrite earlier names for the same digest.
+        for forest in forests.iter() {
+            self.mast_forest.debug_info.extend_procedure_names(
```
huitseeker: This merge can rename a deduped procedure (see #2982). `extend_procedure_names` is keyed only by digest, so the later forest overwrites the earlier alias here.

```rust
#[test]
fn repro_merge_procedure_names_last_name_wins() {
    let mut forest_a = MastForest::new();
    let block_a = block_foo().add_to_forest(&mut forest_a).unwrap();
    forest_a.make_root(block_a);
    let digest = forest_a[block_a].digest();
    forest_a.insert_procedure_name(digest, Arc::from("alias_a"));

    let mut forest_b = MastForest::new();
    let block_b = block_foo().add_to_forest(&mut forest_b).unwrap();
    forest_b.make_root(block_b);
    assert_eq!(forest_b[block_b].digest(), digest);
    forest_b.insert_procedure_name(digest, Arc::from("alias_b"));

    let (merged, root_map) = MastForest::merge([&forest_a, &forest_b]).unwrap();
    let new_a = root_map.map_root(0, &block_a).unwrap();
    let new_b = root_map.map_root(1, &block_b).unwrap();

    assert_eq!(new_a, new_b);
    assert_eq!(merged.procedure_name(&digest), Some("alias_b"));
}
```

First name wins, or keeping all aliases, would avoid the rename.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -566,6 +773,177 @@ impl MastForestBuilder {
...
+    ) -> Result<MastNodeId, Report> {
+        let base_fingerprint = builder
+            .fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)
+            .expect("hash_by_node_id should contain the fingerprints of all children of `node`");
+        let asm_ops_data =
+            serialize_asm_ops_from_forest_node(&self.statically_linked_mast, source_node_id);
+        let debug_vars_data =
+            serialize_debug_vars_from_forest_node(&self.statically_linked_mast, source_node_id);
+        let dedup_fingerprint = base_fingerprint
+            .augment_with_data(&asm_ops_data)
```
huitseeker: This change lets two same-digest roots stay distinct by metadata, but the static-link path still picks the source root by digest only. That makes the second alias copy the first alias's metadata.

```rust
#[test]
fn repro_statically_linked_same_digest_root_uses_first_alias_metadata() {
    let mut source_builder = MastForestBuilder::new(&[]).unwrap();

    let alias_a = source_builder
        .ensure_block(
            vec![Operation::Add],
            Vec::new(),
            vec![(0, AssemblyOp::new(None, "alias_a".into(), 1, "add".into()))],
            vec![],
            vec![],
            vec![],
        )
        .unwrap();
    let alias_b = source_builder
        .ensure_block(
            vec![Operation::Add],
            Vec::new(),
            vec![(0, AssemblyOp::new(None, "alias_b".into(), 1, "add".into()))],
            vec![],
            vec![],
            vec![],
        )
        .unwrap();
    source_builder.mast_forest.make_root(alias_a);
    source_builder.mast_forest.make_root(alias_b);

    let (static_forest, _remapping) = source_builder.build();
    assert_eq!(static_forest[alias_a].digest(), static_forest[alias_b].digest());

    let mut exact_builder = MastForestBuilder::new([&static_forest]).unwrap();
    let exact_alias_b = {
        let node = exact_builder.statically_linked_mast[alias_b].clone();
        let builder = exact_builder.build_with_remapped_ids(alias_b, node).unwrap();
        exact_builder
            .ensure_node_from_statically_linked_source(builder, alias_b)
            .unwrap()
    };
    let (exact_forest, _) = exact_builder.build();

    let mut digest_builder = MastForestBuilder::new([&static_forest]).unwrap();
    let linked_alias_b = digest_builder
        .ensure_external_link(static_forest[alias_b].digest())
        .unwrap();
    let (linked_forest, _) = digest_builder.build();

    assert_eq!(
        exact_forest.debug_info().first_asm_op_for_node(exact_alias_b).unwrap().context_name(),
        "alias_b"
    );
    assert_eq!(
        linked_forest.debug_info().first_asm_op_for_node(linked_alias_b).unwrap().context_name(),
        "alias_b",
        "linking the second same-digest root by digest should preserve that root's metadata",
    );
}
```

The static-link path needs the exact source root here, not just the digest.

core/src/mast/merger/mod.rs
huitseeker: This merge can rename a deduped procedure. `extend_procedure_names` is keyed only by digest, so the later forest overwrites the earlier alias here.

```rust
#[test]
fn repro_merge_procedure_names_last_name_wins() {
    let mut forest_a = MastForest::new();
    let block_a = block_foo().add_to_forest(&mut forest_a).unwrap();
    forest_a.make_root(block_a);
    let digest = forest_a[block_a].digest();
    forest_a.insert_procedure_name(digest, Arc::from("alias_a"));

    let mut forest_b = MastForest::new();
    let block_b = block_foo().add_to_forest(&mut forest_b).unwrap();
    forest_b.make_root(block_b);
    assert_eq!(forest_b[block_b].digest(), digest);
    forest_b.insert_procedure_name(digest, Arc::from("alias_b"));

    let (merged, root_map) = MastForest::merge([&forest_a, &forest_b]).unwrap();
    let new_a = root_map.map_root(0, &block_a).unwrap();
    let new_b = root_map.map_root(1, &block_b).unwrap();

    assert_eq!(new_a, new_b);
    assert_eq!(merged.procedure_name(&digest), Some("alias_b"));
}
```

First name wins, or keeping all aliases, would avoid the rename.

djolertrk: True.

crates/assembly/src/mast_forest_builder.rs
djolertrk: Ah, yes.... This one will be a slightly bigger change I think, let me check what should be done in the assembler linking flow for this (for example we will need source node ID from the linker etc)...

djolertrk: I think that `ProcedureInfo` would need a `source_node_id: Option<MastNodeId>` field, populated during library packaging, and preserved through the linker. That's a cross-crate schema change touching assembly-syntax, mast-package, the linker, and the assembler. Should we postpone that for another PR and fill out TODO here? @huitseeker

huitseeker: Yep, we should probably open an issue for this indeed.

djolertrk: Here it is: https://github.com/0xMiden/miden-vm/issues/2990

djolertrk: And TODO is https://github.com/0xMiden/miden-vm/pull/2955/changes/2a2a12b01d9eed2ca19025f86066273b3ac66423#diff-d41c3716f23b96b0a97303e0ff4eda36797ef21fbdf1f9066b1ba12572c9d63dR1982

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -599,24 +977,52 @@ impl MastForestBuilder {
...
             .with_after_exit(after_exit);
-        let (node_id, is_new) = self.ensure_node_exists(block)?;
+
+        // Compute the base fingerprint from the builder, then augment with external metadata
+        // so the dedup key stays sensitive to source mappings without storing them on the
+        // builder itself.
+        let base_fingerprint = block
+            .fingerprint_for_node(&self.mast_forest, &self.hash_by_node_id)
+            .expect("hash_by_node_id should contain the fingerprints of all children of `node`");
+        let dedup_fingerprint = base_fingerprint
```
huitseeker: Could we gate this on debug info being kept, or compact again after stripping? `apply_debug_options()` clears `mast_forest` debug info when `emit_debug_info` is false, but by then these metadata bytes have already split the dedup key. That means release/stripped builds can keep duplicate nodes that only differed by source metadata which has been discarded.

djolertrk:
Rebased on top of https://github.com/0xMiden/miden-vm/pull/2984 (cc @huitseeker)

crates/assembly/src/mast_forest_builder.rs
djolertrk: We should do it! Nice catch!

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -810,21 +1150,43 @@ impl MastForestBuilder {
...
+                    && self.statically_linked_mast[id].digest() == mast_root
+            })
+            .collect();
+
+        if matching_roots.is_empty() {
+            return self.ensure_node(ExternalNodeBuilder::new(mast_root));
+        }
+
+        let mut first = None;
+        for root_id in matching_roots {
```
huitseeker: Ah, we might have a leftover from the [earlier comment](https://github.com/0xMiden/miden-vm/pull/2955#discussion_r3059508022) — The fix here should have added an exact-root import path, not broadened the digest path. 

`ensure_external_link(mast_root)` had a simple contract: resolve one root for that digest and return that imported root. As we made same-digest aliases distinct by metadata, we changed the digest API to copy every matching root and return the first one, and that can retain extra unreachable alias subtrees in debug-info builds.
  
Specifically, this loop copies every same-digest alias subtree even though the caller only gets `first` back. `build()` only removes merged basic blocks, so the extra aliases stay as unreachable nodes and keep duplicate debug metadata in the final forest. Could we import only the selected alias here?

I could reproduce this with a unit test locally, the setup is: 
- make a static forest with two roots `alias_a` and `alias_b` that have the same digest but different `AssemblyOp` metadata, 
- call `ensure_external_link(digest)`, 
- then mark only the returned `linked` node as a root before `build()`.

In that reproducer the final forest still has `num_nodes() == 2`, `SubtreeIterator::new(&linked, &forest)` only reaches 1 node, and `alias_b_remapped` is still present with its own asm-op metadata even though no root reaches it. So in debug-info builds this loop can retain an unreachable alias subtree, not just return the wrong one.

djolertrk: uh yes yes. it's a leftover.

djolertrk: Done, it basically checks in the test that `num_nodes() == 1` is satisfied.

djolertrk:
@huitseeker @bitwalker can we merge this?

bitwalker:
@huitseeker It looks like the WarpBuild test job is stuck, and I don't see any way to restart/cancel it, or even see the status. Is that something we have to be granted access to?

huitseeker:
@bitwalker @djolertrk The branch should pick up the updated CI with a rebase (it ran the old job on ubuntu-latest).

djolertrk:
> @bitwalker @djolertrk The branch should pick up the updated CI with a rebase (it ran the old job on ubuntu-latest).

Just to add that this is already rebased on top of latest `main`.

huitseeker:
@djolertrk Please run `git rebase --no-ff -S origin/main` and push or give me write rights to the walnuthq branch.

djolertrk:
> @djolertrk Please run `git rebase --no-ff -S origin/main` and push or give me write rights to the walnuthq branch.

done. thanks!

Issue #2929, opened 2026-03-27
Refactor handling of debug information


bobbinth:
We have some inconsistencies in handling of the debug information that gets included with packages/libraries/MAST forests. For example:

- We have debug info sections in both `MastForest` and `Package` and it is not always clear what goes where. It also creates potential for inconsistencies. For example, `MastForest` could have its debug info stripped while `Package` could still have it or vise-versa. This could lead to unexpected behaviors.
- We need to align on when and how we add/remove debug info from compiled artifacts. For example, after [#2877](https://github.com/0xMiden/miden-vm/pull/2877) package-level debug info is controlled via assembler settings, but `MastForest`-level debug info is always included by the assembler and has to be stripped out separately.

There could also be other inconsistencies - but these are the two that come to mind.

The overall goals here, I think, should be:

- Make handling of debug info easy and intuitive for the user - i.e., the user should be able to run a program in debug mode and get helpful error messages etc. w/o the need to tweak too many settings.
- Make sure we don't bundle any private info (e.g., local file paths) into the compiled artifacts.

bobbinth:
I believe we aligned on the path forward here: `DebugInfo` should be moved out of `MastForest` into the `Package`.

PR #2958, opened 2026-04-02
fix(core): preserve asm-op mappings in MastForest::merge

KitaroMoha563:
Preserve and remap AssemblyOp source mappings during MastForest::merge, preventing debug source locations from being lost after forest deduplication and link-time merge.

### DIFF
--- a/core/src/mast/merger/mod.rs
+++ b/core/src/mast/merger/mod.rs
@@ -1,4 +1,11 @@
-use alloc::{collections::BTreeMap, vec::Vec};
+use alloc::{
+    collections::{BTreeMap, btree_map::Entry},
+    string::String,
+    vec::Vec,
+};
+use core::cmp::Ordering;
+
+use miden_debug_types::Location;
 
 use crate::{
     crypto::hash::Blake3Digest,
@@ -7,12 +14,15 @@ use crate::{
         MastNode, MastNodeBuilder, MastNodeFingerprint, MastNodeId, MultiMastForestIteratorItem,
         MultiMastForestNodeIter,
     },
+    operations::AssemblyOp,
     serde::Serializable,
     utils::{DenseIdMap, IndexVec},
 };
 
 #[cfg(test)]
 mod tests;
+
+type AssemblyOpKey = (Option<Location>, String, u8, String);
 
 /// A type that allows merging [`MastForest`]s.
 ///
@@ -27,6 +37,8 @@ pub(crate) struct MastForestMerger {
     node_id_by_hash: BTreeMap<MastNodeFingerprint, MastNodeId>,
     hash_by_node_id: IndexVec<MastNodeId, MastNodeFingerprint>,
     decorators_by_hash: BTreeMap<Blake3Digest<32>, DecoratorId>,
+    asm_op_id_by_value: BTreeMap<AssemblyOpKey, AsmOpId>,
+    asm_op_value_by_id: BTreeMap<AsmOpId, AssemblyOpKey>,
     /// Mappings from old decorator and node ids to their new ids.
     ///
     /// Any decorator in `mast_forest` is present as the target of some mapping in this map.
@@ -35,6 +47,10 @@ pub(crate) struct MastForestMerger {
     ///
     /// Any `MastNodeId` in `mast_forest` is present as the target of some mapping in this map.
     node_id_mappings: Vec<DenseIdMap<MastNodeId, MastNodeId>>,
+    /// AssemblyOp mappings to register after all nodes have been merged.
+    ///
+    /// This is keyed by merged node id and stores `(num_operations, [(op_idx, asm_op_id)])`.
+    pending_asm_op_mappings: BTreeMap<MastNodeId, (usize, Vec<(usize, AsmOpId)>)>,
 }
 
 impl MastForestMerger {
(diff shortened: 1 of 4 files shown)


core/src/mast/merger/mod.rs
```diff
@@ -291,6 +315,99 @@ impl MastForestMerger {
...
+
+        if asm_ops.is_empty() {
+            return Ok(());
+        }
+
+        // We may see the same merged node multiple times due deduplication. Keep the first
+        // mapping we encounter for that node.
+        self.pending_asm_op_mappings
+            .entry(merged_node_id)
+            .or_insert((num_operations, asm_ops));
```
huitseeker: Hmm, `or_insert` still makes the merge order matter when both duplicate nodes already have asm-op mappings. 

I can reproduce it with two identical blocks: forest A had `[(0, mul)]`, forest B had `[(0, mul), (1, add)]`. Merging A then B leaves op 1 resolving to `mul`, while merging B then A resolves op 1 to `add`. 

Could we merge compatible mappings here, or at least prefer the richer mapping? The new test only covers empty vs non-empty, so this case still loses source locations.

huitseeker (changes requested):
Thanks! The underlying bug is real: `MastForest::merge` was losing `AssemblyOp` source mappings during dedup, which breaks debug/source-location fidelity after linking.

However, the new logic still keeps the first mapping for a deduplicated node. That means results still depend on merge order when both duplicates already have asm-op mappings and one is richer than the other. In that case, source info can still be lost.

relates to #2929

core/src/mast/merger/mod.rs
KitaroMoha563: @huitseeker Updated, thanks for your time.

core/src/mast/merger/mod.rs
```diff
@@ -291,6 +320,238 @@ impl MastForestMerger {
...
+                        asm_ops.push((op_idx, merged_asm_op_id));
+                    }
+
+                    previous_asm_op = asm_op_key;
+                }
+
+                (num_operations, asm_ops)
+            },
+            _ => {
+                let Some(asm_op) = source_forest.debug_info.first_asm_op_for_node(source_node_id)
```
huitseeker: This branch drops any extra op-indexed `AssemblyOp` entries on non-block nodes. 

`first_asm_op_for_node()` only keeps the first mapping and then hardcodes `num_operations = 1`, so a source forest that has mappings at `op_idx > 0` on a `Call`/`Loop`/`Split` node will lose them during merge. 

The builder mostly writes `[(0, ...)]` today, but the storage API and `get_assembly_op(node_id, Some(idx))` both support sparse non-block mappings already. 

Can we preserve the full `(op_idx, asm_op)` list here instead of collapsing it to `[(0, first)]`?

core/src/mast/merger/mod.rs
```diff
@@ -291,6 +320,238 @@ impl MastForestMerger {
...
+        let mut merged = Vec::with_capacity(num_operations);
+        for op_idx in 0..num_operations {
+            let merged_asm_op = match (lhs_expanded[op_idx], rhs_expanded[op_idx]) {
+                (Some(lhs_asm_op), Some(rhs_asm_op)) if lhs_asm_op == rhs_asm_op => {
+                    Some(lhs_asm_op)
+                },
+                (Some(lhs_asm_op), Some(rhs_asm_op)) => Some(match preference {
+                    Ordering::Greater => lhs_asm_op,
+                    Ordering::Less => rhs_asm_op,
+                    Ordering::Equal => {
```
huitseeker: This tie case can still build a mapping that never existed in either input forest!

It compares `AsmOpId`s slot by slot, so two equally rich mappings can get mixed after interning. For example, `lhs=[(0, A), (1, B)]` and `rhs=[(0, C), (1, A)]` can collapse to `[(0, A)]` if `A` got the smallest id. That loses the transition at op 1 and points the merged node at a synthetic source map. 

I'd rather pick one whole mapping in the equal-richness case, using a stable value-based key, instead of choosing the smaller id per operation, no?

core/src/mast/merger/mod.rs
```diff
@@ -257,8 +285,11 @@ impl MastForestMerger {
                     returned_id, new_node_id,
                     "hash_by_node_id push() should return the same node IDs as node_id_by_hash"
                 );
+                new_node_id
             },
-        }
+        };
+
+        self.merge_node_asm_ops(original_forests[forest_idx], merging_id, mapped_node_id)?;
```
huitseeker, line 298: The missing case is external replacement. This call keeps asm-op mappings for the normal `Node` path, but an `External` node that is replaced through `ExternalNodeReplacement` never comes through here, so its source mapping is still dropped.

A small repro is an `External(foo)` with `register_asm_ops(external_id, 1, vec![(0, asm_op_id)])` merged with a concrete `Block(foo)`. After merge, `get_assembly_op(mapped_root, None)` comes back `None`.

The root cause is that the replacement branch records the node-id remap but does not merge the replaced node's asm-op data into `mapped_replacement`. The fix is to do that there too, and add a regression test for this path.

huitseeker (changes requested):

core/src/mast/merger/mod.rs
KitaroMoha563: @huitseeker Thanks for your careful review, updated. Could you please take another look?

huitseeker (approved):
Thanks a lot for the contribution!

bitwalker (approved):

PR #2959, opened 2026-04-03
ci: add a daily job to report unused crates

huitseeker:
This adds a daily GitHub Actions job that looks for crates listed in Cargo.toml but not used by the code.

The workspace check skips `miden-vm`, and `miden-vm` is checked in a second step with all features enabled. This avoids false positives from crates that are only used by optional CLI code.

The change also fixes the Cargo metadata used by `cargo-udeps`, restores the missing `miden-project` test macro dependency, and adds the issue template used when the daily check  fails.

### DIFF
--- a/.github/templates/UNUSED_DEPS_ISSUE.md
+++ b/.github/templates/UNUSED_DEPS_ISSUE.md
@@ -1,0 +1,13 @@
+---
+title: Unused dependency check failed
+labels: ci, dependencies
+---
+
+The scheduled unused dependency check reported one or more dependencies that are not used in the
+feature and target combinations covered by `.github/workflows/unused-deps.yml`.
+
+Please review the failing workflow run in the Actions tab and either:
+
+- remove the dependency if it is genuinely unused, or
+- add a documented `cargo-udeps` ignore entry if the dependency is only used by cfg-gated code,
+  generated code, or doctests that `cargo-udeps` cannot see.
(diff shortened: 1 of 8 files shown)


bitwalker:
I'd recommend `cargo-machete` for this, it's basically an improved version of `cargo-udeps`, and what we use in the compiler for this type of lint

huitseeker:
@bitwalker I'm familiar, and I find I prefer `cargo-udeps`, in this context of a daily job where duration does not matter (admittedly , `cargo-machete` is way faster).

- `cargo-machete` is more sensitive for directly-unused `Cargo.toml` entries, because `cargo-udeps` documents that some unused crates are missed when they are "already being used by dependencies of the studied crate" ([cargo-udeps README / Known bugs](https://github.com/est31/cargo-udeps/blob/master/README.md#known-bugs)), and 
- `cargo-machete`'s writeup is explicitly about that transitively-used-dependency blind spot ([cargo-machete writeup](https://bouvier.cc/cargo-machete/)). 
- `cargo-udeps` is more selective / lower-noise, because `cargo-machete` describes itself as "fast (yet imprecise)" and documents false-positive cases like `build.rs`-generated code and renamed crates ([README / intro](https://github.com/bnjbvr/cargo-machete#introduction), [false positives](https://github.com/bnjbvr/cargo-machete#false-positives), [renamed crates](https://github.com/bnjbvr/cargo-machete#renamed-crates)). 

Practical summary: `cargo-udeps` for higher-confidence confirmation.

PR #2960, opened 2026-04-03
ci: move expensive Rust workflows to WarpBuild

huitseeker:
This switches the longest Rust CI workflows over to WarpBuild runners and WarpBuild’s Rust cache.

  Changes :

  - Move test jobs from ubuntu-latest to warp-ubuntu-latest-x64-4x and WarpBuilds/rust-cache@v2 in test
  - Move Workspace release dry-run to warp-ubuntu-latest-x64-4x and WarpBuilds/rust-cache@v2
  - Move workspace-publish to warp-ubuntu-latest-x64-4x as well, since it uses the same shared action

  Scope notes:

  - fuzz.yml is intentionally unchanged
  - The WarpBuild cache change is only applied where the job is also running on a WarpBuild runner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


huitseeker:
@bobbinth We need to tweak our config, the test isn't being picked up 😭 (maybe the bot isn't on the right account)

PR #2963, opened 2026-04-05
fix: replace unwrap() with ? in decorator_storage and debug_var_storage

### LINKED ISSUE #2785: unwrap() instead of ? in recursive calls in decorator_storage and debug_var_storage

bigeez:
## Summary
Fixes #2785

Replaces `unwrap()` calls with proper `?` error propagation in two files:
- `core/src/mast/debuginfo/decorator_storage.rs`
- `core/src/mast/debuginfo/debug_var_storage.rs`

### `decorator_storage.rs`
- `validate_csr`: extracted `last()` results into variables using `ok_or_else(...)? ` instead of calling `unwrap()` twice
- `add_decorator_info_for_node` (recursive): replaced `last().unwrap()` and `next().unwrap()` with `ok_or(DecoratorIndexError::InternalStructure)?`

### `debug_var_storage.rs`
- Same pattern mirrored for `validate_csr` and `add_debug_var_info_for_node`

## Result
- 8 `unwrap()` calls removed across 2 files
- All remaining `unwrap()` calls in these files are inside `#[cfg(test)]` test code only
- `cargo check` passes cleanly

## Why this matters
`unwrap()` in production code panics on unexpected state instead of 
returning a proper error. Using `?` means callers get a `Result` they 
can handle gracefully, which is especially important in recursive 
functions where the call stack is deeper.


bigeez:
## Test Plan
- `cargo check -p miden-core` passes cleanly with no errors
- All remaining `unwrap()` calls in both files are exclusively inside 
  `#[cfg(test)]` blocks and are acceptable in test code
- No behavior changes — only error propagation improved from panic to Result

bigeez:
/quality-review

Note: As an external contributor I don't have permission to self-assign issues. 
I am the author of this fix for #2785.

Test plan:
- `cargo check -p miden-core` passes cleanly
- All remaining `unwrap()` calls in both files are exclusively inside `#[cfg(test)]` blocks
- No behavior changes — only error propagation improved from panic to Result

huitseeker:
Closing in favor of #3026

PR #2964, opened 2026-04-05
fix(mast): merge post-last-op decorators into after_exit (#2633)

### LINKED ISSUE #2633: Investigate removing `Processor::execute_end_of_block_decorators()`
The `DecoratorList` associated with a basic block currently supports the encoding of decorators after the last operation. We should investigate removing this functionality, in favor of using exclusively "after exit" decorators for this purpose.

This would simplify a few things in the processor, such as removing the need for `Processor::execute_end_of_block_decorators()`, as well as `Continuation::AfterExitDecoratorsBasicBlock` (in favor of using `Continuation::AfterExitDecorators`).

Related: https://github.com/0xMiden/miden-vm/pull/2615#discussion_r2749837346

crazywriter1:
## Summary

This change implements the approach discussed in #2633: operation-indexed decorators attached at the sentinel index (after the last padded operation in a basic block) are merged into the node’s `after_exit` list when the MAST is built, instead of being executed via a separate continuation and `execute_end_of_block_decorators`.

Execution after `END` now always goes through `Continuation::AfterExitDecorators`. Legacy forests that still store those decorators only at the op-indexed sentinel are handled in `FastProcessor::execute_after_exit_decorators` by running `decorators_for_op(node_id, num_ops)` before the usual `after_exit` loop, preserving the previous ordering relative to explicit `after_exit` decorators.

## Changes

- **Core:** Added `merge_sentinel_op_decorators_into_after_exit` and call it from `BasicBlockNodeBuilder::build` and `MastForestContributor for BasicBlockNodeBuilder::add_to_forest` so sentinel-index decorators are prepended (in relative order) to `after_exit`; `register_node_decorators` uses the merged list.
- **Processor:** Removed `Continuation::AfterExitDecoratorsBasicBlock` and the `Processor::execute_end_of_block_decorators` trait method (and all implementations, including `ReplayProcessor`). `finish_basic_block` only schedules `AfterExitDecorators`; the `AfterExitDecoratorsBasicBlock` match arm was removed from the execution loop.
- **Fast path:** `execute_after_exit_decorators` runs sentinel op decorators for `MastNode::Block` before iterating `after_exit`, for backward compatibility with older serialized forests.
- **Docs / release notes:** Updated `CHANGELOG.md` under `v0.23.0` with a **[BREAKING]** entry for the removed `Continuation` variant.

## Breaking changes

- **`miden_processor::Continuation`:** The `AfterExitDecoratorsBasicBlock` variant is removed. Downstream code that matches on `Continuation` exhaustively must be updated to drop this arm; behavior for newly built MAST is unchanged from a user perspective because those decorators now live in `after_exit`.

## Testing

- `cargo test -p miden-core -p miden-processor` (and any other tests you ran locally).

Fixes #2633


processor/src/fast/mod.rs
```diff
@@ -534,6 +534,17 @@ impl FastProcessor {
             .get_node_by_id(node_id)
             .expect("internal error: node id {node_id} not found in current forest");
 
+        // Basic blocks may still carry legacy op-indexed decorators at the sentinel index (padded
+        // op count). New MAST merges those into `after_exit` at build time; execute them
+        // here first so order matches historical `execute_end_of_block_decorators` +
+        // `after_exit`.
+        if let MastNode::Block(bb) = node {
```
huitseeker, line 547: This looks like it drops trace-only decorators at the post-last-op sentinel. Before this patch, `execute_end_of_block_decorators()` still ran when tracing was on, even if debug mode was off. Now these decorators go through `execute_after_exit_decorators()`, which returns early on `!in_debug_mode()`, so `Decorator::Trace` at the sentinel never fires in tracing-only mode. Would switching this gate to `should_execute_decorators()` fix it?

huitseeker (changes requested):
Thanks for picking this one up! This mostly looks super good. One issue to fix inline.

crazywriter1:
Addressed in the latest push: `execute_after_exit_decorators` now uses `should_execute_decorators()` for the early return (same as `execute_before_enter_decorators` and the old `execute_end_of_block_decorators`), so trace decorators at the sentinel still run when tracing is enabled without debug mode. Thanks for catching that. @huitseeker

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -33,6 +33,31 @@ mod tests;
 /// Maximum number of operations per group.
 pub const GROUP_SIZE: usize = 9;
 
+/// Op-indexed decorators at padded index `num_padded_ops` run after the last operation; merge them
+/// into `after_exit` (before any existing `after_exit` ids), preserving relative order.
+pub(crate) fn merge_sentinel_op_decorators_into_after_exit(
```
huitseeker, line 38: I think this normalization also needs to happen in `fingerprint_for_node()`. Right now a builder that keeps a decorator at the sentinel op index will hash differently from the node built by `build()` or `add_to_forest()`, because those paths move that decorator into `after_exit` first. That seems to break the `MastForestContributor` invariant that builder fingerprints match the built node, and it can stop equivalent basic blocks from deduping during forest merges.

processor/src/fast/mod.rs
```diff
@@ -523,7 +523,7 @@ impl FastProcessor {
         current_forest: &MastForest,
         host: &mut impl BaseHost,
     ) -> ControlFlow<BreakReason> {
-        if !self.in_debug_mode() {
+        if !self.should_execute_decorators() {
```
huitseeker, line 532: Could we add a regression test for the exact case this changes: tracing on with debugging off? The current decorator tests only cover debug off and debug plus tracing on, so this path still looks unpinned, and the legacy sentinel fallback in `execute_after_exit_decorators()` seems worth covering too.

processor/src/execution/basic_block.rs
```diff
@@ -194,7 +194,7 @@ where
 #[inline(always)]
```
huitseeker, line 193: Now that end-of-block decorators are gone from this path, could we drop this parameter entirely? Keeping an unused `BasicBlockNode` argument here makes the new control flow a little harder to read because it still suggests `finish_basic_block()` needs block-local state.

core/src/mast/node/basic_block_node/mod.rs
huitseeker, line 38: It may also be worth pushing this one step further and making the builder expose one canonical decorator-normalization path that `build()`, `add_to_forest()`, and `fingerprint_for_node()` all share. Right now the same sentinel-to-`after_exit` rule has to stay in sync across multiple code paths, which makes it easy for one of them to drift.

crazywriter1:
Thanks for the thorough review. Addressed the remaining points in the latest push:

- **Fingerprint / invariant:** `fingerprint_for_node` now applies the same sentinel → `after_exit` merge as `build` / `add_to_forest`. Extraction of `prepare_op_batches_and_padded_decorators` keeps those three paths in sync.
- **API cleanup:** Removed the unused `BasicBlockNode` argument from `finish_basic_block` and updated call sites.
- **Tests:** Added tracing-on / debug-off coverage for decorator retrieval and execution, plus a core test that sentinel vs explicit `after_exit` builders fingerprint the same and `build()` to equal nodes. `cargo test -p miden-core -p miden-processor` and `cargo +nightly fmt --all --check` are green.

Let me know if you want anything else tightened up. @huitseeker

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 - [BREAKING] Removed the deprecated unbound `TraceBuildInputs::new()` and `TraceBuildInputs::from_program()` constructors; use `execute_trace_inputs_sync()` or `execute_trace_inputs()` instead ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
 - Added `prove_from_trace_sync(...)` for proving from pre-executed trace inputs ([#2865](https://github.com/0xMiden/miden-vm/pull/2865)).
 - [BREAKING] Reduced the prove-from-trace API to post-execution trace inputs: `TraceBuildInputs` no longer carries full execution output, `prove_from_trace_sync()` takes `TraceProvingInputs`, and `ProvingOptions` no longer include `ExecutionOptions` ([#2948](https://github.com/0xMiden/miden-vm/pull/2948)).
+- [BREAKING] Removed `Continuation::AfterExitDecoratorsBasicBlock`. New MAST merges operation-indexed decorators at the post-last-op sentinel index into `after_exit` at build time; execution uses `AfterExitDecorators` only, with legacy forests still supported ([#2633](https://github.com/0xMiden/miden-vm/issues/2633)).
+- Refactor trace generation to row-major format ([#2937](https://github.com/0xMiden/miden-vm/pull/2937)).
```
huitseeker: This looks like a duplicate of the same `#2937` changelog entry a few lines below. Could this added line be dropped so this PR only adds the `#2633` entry?

huitseeker (commented):
This looks great aside from the duplicate Changelog line. Thanks!

crazywriter1:
Dropped the extra #2937 line next to #2633; changelog now has a single #2937 entry as before. Thanks!

core/src/mast/node/basic_block_node/mod.rs
huitseeker, line 38: This merge helper fixes the builder paths, but `BasicBlockNode::new_owned_with_decorators()` a bit lower still skips the same sentinel -> `after_exit` normalization. That leaves the new invariant incomplete for `test`/`arbitrary`-owned blocks, so round-trips or direct comparisons can still differ from builder-created nodes.

crazywriter1:
Updated `new_owned_with_decorators` to merge sentinel op-decorators into `after_exit` like `build`, so the invariant holds for test/arbitrary-owned blocks. fmt + core/processor tests green. Thanks. @huitseeker

PR #2968, opened 2026-04-06
Fix Falcon remainder validation and u64::rotr overflow handling

huitseeker:
This PR fixes two correctness bugs.

In Falcon `mod_12289`, the advice-supplied remainder is now validated as a `u32` before any u32 operation uses it. This blocks forged non-`u32` remainders from producing the wrong result. Tests cover the forged-remainder path.

In `u64::rotr`, the field arithmetic path is replaced with `u32widening_mul` and `u32widening_madd`. This fixes bad outputs for rotations by `0` and `32` when the input is at or above the Goldilocks modulus. Tests cover both cases.

The `rotr` docs now match in the MASM source, generated core docs, and user docs.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+# Allow actionlint to recognize the org's custom Warp runner label used by the
+# blake3 non-regression workflow.
+self-hosted-runner:
+  labels:
+    - warp-ubuntu-latest-x64-4x
--- a/.github/actions/workspace-release/action.yml
+++ b/.github/actions/workspace-release/action.yml
@@ -43,19 +43,11 @@ runs:
     # Install cargo-msrv for MSRV checks
     - name: Install cargo-msrv
       shell: bash
-      env:
-        RUSTUP_TOOLCHAIN: "1.91"
       run: cargo install cargo-msrv
-
-    - name: Install Rust 1.91 for MSRV check
-      shell: bash
-      run: rustup toolchain install 1.91
 
     # Keep your existing MSRV check
     - name: Check MSRV
       shell: bash
-      env:
-        RUSTUP_TOOLCHAIN: "1.91"
       run: |
         chmod +x scripts/check-msrv.sh
         ./scripts/check-msrv.sh
--- a/.github/workflows/blake3-nonregression.yml
+++ b/.github/workflows/blake3-nonregression.yml
@@ -1,0 +1,372 @@
+name: blake3 1-to-1 non-regression
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      baseline_ref:
+        description: Commit or ref to use as the baseline. Defaults to the current ref's first parent.
+        required: false
+        type: string
+      current_ref:
+        description: Commit or ref to benchmark. Defaults to the workflow ref.
+        required: false
+        type: string
+      regression_threshold_pct:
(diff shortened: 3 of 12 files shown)


crates/lib/core/asm/crypto/dsa/falcon512_poseidon2.masm
```diff
@@ -55,6 +55,9 @@ pub proc mod_12289
     # => [res_hi, res_lo, a_hi, a_lo, ...]
 
     adv_push.1
+    # Advice values are untrusted. U32SUB/U32ADD only constrain their output limbs in the AIR,
+    # so we must validate the remainder before feeding it into u32 arithmetic.
+    u32assert
     dup
```
bobbinth: `u32assert` takes 2 VM cycles because it is a macro instruction that uses `u32assert2` under the hood. But here, we could just use `u32assert2` doing something like:

```
adv_push.1
dup
u32assert2
```
This way, we'll introduce only 1 extra cycle.

crates/lib/core/asm/math/u64.masm
```diff
@@ -826,115 +826,130 @@ end
...
 
     # --- STEP 5: Carry the overflow to the low bits ---
     movup.2
-    # [lo1, hi2, lo2, flag, ...]
+    # [lo1, hi2, lo2, borrow, shift_amt == 0, ...]
 
     add
-    # [lo1 + hi2, lo2, flag, ...]
+    # [lo1 + hi2, lo2, borrow, shift_amt == 0, ...]
     # c_lo = lo1 + hi2 (overflow wraps to low bits)
```
Al-Kindi-0, line 938: worth adding a comment about the safety of using `add` on u32 here

Al-Kindi-0, line 938: worth adding a comment about the safety of using `add` on u32 here

crates/lib/core/tests/math/u64_mod.rs
```diff
@@ -1203,6 +1204,36 @@ fn unchecked_rotr() {
     build_test!(source, &stack![b as u64, a0, a1, 5]).expect_stack(&[c0, c1, 5]);
 }
 
+#[test]
+fn unchecked_rotr_large_value_by_0() {
```
Al-Kindi-0, line 1208: Should we also add tests, especially at the edge points, for `rotl` as well ?

crates/lib/core/asm/math/u64.masm
Al-Kindi-0: Seems worth also add a similar comment about the use of `add` a bit up for `shr` procedure.
Maybe good practice for the future to always argue safety of using field ops on non-field elements

air/src/constraints/stack/stack_arith/tests.rs
```diff
@@ -0,0 +1,185 @@
...
+use super::enforce_main;
+use crate::{
+    MainTraceRow,
+    constraints::op_flags::{ExprDecoderAccess, OpFlags, generate_test_row},
+    trace::{
+        AUX_TRACE_RAND_CHALLENGES, AUX_TRACE_WIDTH, TRACE_WIDTH, decoder::USER_OP_HELPERS_OFFSET,
+    },
+};
+
+struct ConstraintEvalBuilder {
```
adr1anh, line 22: Looks like a nice reusable abstraction, I'll see how it could fit with #3004

PR #2969, opened 2026-04-07
ci: reuse maintenance workflows

huitseeker:
Reuses the shared workflows in https://github.com/0xmiden/.github as opposed to redefining them locally.


bobbinth (approved):
Looks good! Thank you!

PR #2970, opened 2026-04-07
test: reduce proptest cases in VM operation hotspots

huitseeker:
This change cuts the case count to 100 in the three VM operation `proptest!` blocks that dominate runtime: field arithmetic, u32 arithmetic, and stack ops.

The baseline data is concentrated enough that this is the right first cut. These 33 tests used about 464s of CPU time out of about 1200s total proptest CPU time, or roughly 39% of the total. The biggest outliers were sub_proptest at 42s, u32div_proptest at 25s, u32mod_proptest at 23s, and the stack ops block at about 254s total.

The coverage tradeoff here is deliberate and bounded. Each touched test still runs 100 randomized samples per run, and the change is limited to the known hotspots instead of weakening proptests across the workspace.

### DIFF
--- a/miden-vm/tests/integration/operations/field_ops.rs
+++ b/miden-vm/tests/integration/operations/field_ops.rs
@@ -742,6 +742,7 @@ fn gte() {
 // ================================================================================================
 
 proptest! {
+    #![proptest_config(ProptestConfig::with_cases(100))]
     #[test]
     fn add_proptest(a in any::<u64>(), b in any::<u64>()) {
         let asm_op = "add";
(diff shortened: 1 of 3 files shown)


bobbinth (approved):
Looks good! Thank you!

huitseeker:
On an M4 with 16 procs: 
```
next at 3d05a4c8b: 429.98s total, about 7m 10s
tune-proptest-hotspots at 44033c0ba: 245.98s total, about 4m 06s
```
Now how that translates to ubuntu-latest ...

PR #2971, opened 2026-04-07
docs: fix grammar in nondeterministic inputs section of overview

koookxbt:
## Summary

This PR fixes two minor grammar issues in `docs/src/overview.md` in the **Nondeterministic inputs** section:

1. **Subject-verb agreement**: `**Merkle store** which contain structured data` → `**Merkle store** which contains structured data`
   - "Merkle store" is a singular noun and should use "contains" rather than "contain".

2. **Redundant phrasing**: `the VM can either push new elements onto the advice stack, or pop the elements from its top` → `the VM can push new elements onto the advice stack or pop elements from its top`
   - Removed the unnecessary "either...or" construct and simplified "the elements" → "elements" for cleaner prose.

No functional changes — documentation only.

Closes: (no issue — opportunistic doc cleanup)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
 #### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


koookxbt:
Thanks for reviewing! 🙏

These are purely grammatical fixes with no semantic changes to the documentation content:

- The **Merkle store** fix addresses subject-verb agreement ("Merkle store" is singular, so it should use "contains" not "contain").
- The **advice stack** sentence was simplified to remove the redundant "either...or" framing — grammatically valid but unnecessarily wordy for technical documentation.

Happy to adjust anything if preferred!

PR #2972, opened 2026-04-07
docs: improve CLI subcommand descriptions and add logging note in usage.md

koookxbt:
## Summary

This PR improves the CLI documentation in `docs/src/usage.md`:

### Changes

1. **Subcommand descriptions rewritten for clarity**:
   - Switched from the passive "this will execute" style to active present-tense: "executes", "verifies", "compiles" — more consistent with standard CLI documentation.
   - The `prove` description now notes that its output can be used with `verify`, making the relationship between the two subcommands explicit for new users.
   - The `example` description now lists the available examples (`blake3` and `fibonacci`) inline.

2. **Parallelism section**: Added a note that `RAYON_NUM_THREADS=0` causes rayon to automatically choose the thread count based on available CPU cores.

3. **New logging section**: Added a short `#### Enabling logging` section documenting the `MIDEN_LOG` environment variable, which was previously undocumented in the usage guide.

No behavioral changes — documentation only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
 #### Bug Fixes
 
+- Replaced unsound `ptr::read` with safe unbox in panic recovery, removing UB from potential double-drop ([#2934](https://github.com/0xMiden/miden-vm/pull/2934)).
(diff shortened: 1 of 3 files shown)


koookxbt:
Thanks for taking a look! 🙏

Motivation for the key changes:

- **Subcommand style**: Switching from "this will execute" to "executes" matches how most CLIs document their subcommands and is more concise.
- **`prove` → `verify` link**: New users frequently ask how to verify a proof they just generated — making this relationship explicit in the `prove` description should help.
- **`RAYON_NUM_THREADS=0`**: This is a common question and the existing parallelism docs didn't mention the default/automatic behavior.
- **Logging**: The `MIDEN_LOG` env var is used in practice but had no home in the usage guide.

Happy to refine any of this based on feedback!

PR #2976, opened 2026-04-08
Fix memory range-check clk underflow

### LINKED ISSUE #2828: Release trace generation relies on wrapping subtraction for clk=0 memory access

huitseeker:
`Memory::append_range_checks()` used plain unsigned subtraction when the first memory access had `clk = 0`. That path is reachable from a `CALL`, because the frame-pointer write is recorded before the clock moves forward.

This makes the existing wraparound behavior explicit with `wrapping_sub(1)` when seeding `prev_clk`, and `wrapping_sub(prev_clk)` when computing the first clock delta.

The tests add one direct regression for `append_range_checks()` and one full trace-build regression through a `CALL` entrypoint.
Fixes #2828

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
 #### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


Al-Kindi-0 (approved):
LGTM

adr1anh (approved):

PR #2987, opened 2026-04-09
fix: panic on assembler source manager mismatch

### LINKED ISSUE #2986: `Assembler::assemble_library` panics when given a pre-parsed module from a different source manager
Calling `Assembler::assemble_library()` with a pre-parsed `Box<Module>` that was parsed using a *different* `DefaultSourceManager` than the one owned by the assembler causes a panic in `SourceFile::location()`:

```
panicked at crates/debug-types/src/source_file.rs:200:14:
invalid source span: starting byte is out of bounds
```

This affects any downstream user that manually parses a `Module` with their own `DefaultSourceManager` and then passes it to an `Assembler` constructed with a separate source manager (e.g., via `Assembler::default()` or `TransactionKernel::assembler()`).

**We came up with the problem in:** `miden-client` test `prune_preserves_unmodified_storage_slots` ([0xMiden/miden-client#1982](https://github.com/0xMiden/miden-client/pull/1982))

## Root Cause

When two `DefaultSourceManager` instances exist, `SourceId(0)` in manager A refers to a completely different file than `SourceId(0)` in manager B.

Then:

1. **Module parsed with external source manager A.** The parser calls `source_manager.load(...)` which assigns `SourceId(0)` to the user's source file. All `SourceSpan`s in the resulting AST nodes carry this id with byte offsets valid for that file's content.

2. **`Parse for Box<Module>` ignores the assembler's source manager.** When `assemble_library` calls `module.parse_with_options(self.source_manager.clone(), ...)`, the impl at `assembly-syntax/src/parse.rs:137-155` simply returns the module unchanged:

   ```rust
   impl Parse for Box<Module> {
       fn parse_with_options(
           mut self,
           _source_manager: Arc<dyn SourceManager>,  // ← IGNORED
           options: ParseOptions,
       ) -> Result<Box<Module>, Report> {
           // just validates module kind, returns self
           Ok(self)
       }
   }
   ```

3. **Debug info registration resolves spans against the wrong source manager.** During compilation, `register_procedure_debug_info` (at `assembler/debuginfo.rs:54`) calls `source_manager.file_line_col(*procedure.span())` using the *assembler's* source manager B.

4. **`SourceId(0)` resolves to the wrong file.** B's `files.get(SourceId(0))` returns whatever file B has at index 0 (e.g., a kernel module). The byte offsets from the user's longer source file are then applied to B's shorter/different file:

   ```rust
   // source_file.rs:195-201
   pub fn location(&self, span: SourceSpan) -> FileLineCol {
       assert_eq!(span.source_id(), self.id, "mismatched source ids");
       self.content
           .location(ByteIndex(span.into_range().start))
           .expect("invalid source span: starting byte is out of bounds") // ← PANIC
   }
   ```

   The `assert_eq!` passes because both managers happen to assign `SourceId(0)` to their first file (it's an index, not unique). The byte offset check fails because the span's byte range is valid for the original file but not for the unrelated file at the same index in the assembler's manager.

## Reproduction (in miden-vm)

```rust
#[test]
#[should_panic(expected = "invalid source span: starting byte is out of bounds")]
fn assemble_library_panics_with_mismatched_source_manager() {
    let sm_a: Arc<dyn SourceManager> = Arc::new(DefaultSourceManager::default());
    // Populate SourceId(0) in sm_a with a short file
    let _dummy = Module::parser(ModuleKind::Library)
        .parse_str("lib::dummy", "pub proc dummy\n  push.1\nend", sm_a.clone())
        .unwrap();

    let assembler = Assembler::new(sm_a);

    // Parse a longer module with a separate source manager
    let sm_b: Arc<dyn SourceManager> = Arc::new(DefaultSourceManager::default());
    let long_source = "\
# padding line to push the procedure definition past the short file length
# more padding to ensure the byte offset exceeds the boundary
# even more padding for good measure here to be absolutely sure
pub proc bar
  push.1
  push.2
  add
end";
    let module = Module::parser(ModuleKind::Library)
        .parse_str("lib::external", long_source, sm_b)
        .unwrap();

    // Panics: module spans reference SourceId(0) from sm_b,
    // but assembler resolves against sm_a where SourceId(0) is the short file
    let _lib = assembler.assemble_library([module]).unwrap();
}
```

## Proposed Fix

1. **Make`SourceFile::location()` return `Option` instead of panicking.** This removes the `expect()` and lets `file_line_col` return `Err(InvalidBounds)` when a span's byte offset is out of bounds for the resolved file.

2. **`register_procedure_debug_info` distinguishes missing sources from mismatched ones.** 
- If `source_manager.get(source_id)` fails (`InvalidSourceId`), the span is synthetic or from a deserialized module, debug info registration is skipped gracefully. 
- If `get` succeeds but `file_line_col` fails (`InvalidBounds`), the source id resolves to a file but the span doesn't fit it. This indicates a source manager mismatch and is surfaced as a compilation error with a clear message.

juan518munoz:
When a pre-parsed module is assembled using a different source manager than the one it was parsed with, the assembler used to panic during debug info registration because span byte offsets didn't match the resolved source file. 

This PR replaces the panic with proper error handling: missing source ids are skipped silently, while out-of-bounds spans are surfaced as a compilation error with a descriptive message pointing to the source manager mismatch.

Closes #2986

### DIFF
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -5,10 +5,7 @@ use miden_core::{
     advice::AdviceMap,
     serde::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
 };
-use miden_debug_types::{
-    SourceContent, SourceFile, SourceId, SourceLanguage, SourceManager, SourceSpan, Span, Spanned,
-    Uri,
-};
+use miden_debug_types::{SourceFile, SourceManager, SourceSpan, Span, Spanned};
 use miden_utils_diagnostics::Report;
 use smallvec::SmallVec;
 
@@ -129,11 +126,15 @@ pub struct Module {
     pub(crate) items: Vec<Export>,
     /// AdviceMap that this module expects to be loaded in the host before executing.
     pub(crate) advice_map: AdviceMap,
-    /// The source file from which this module was parsed.
+    /// The source file from which this module was parsed, if any.
     ///
-    /// This is used to verify that the source manager used during assembly is the same one that
-    /// was used to parse this module, via pointer identity ([`SourceManager::is_manager_of`]).
-    source_file: Arc<SourceFile>,
+    /// This is `None` for programmatically-constructed modules (e.g. `new_kernel()`,
+    /// `new_executable()`) that have no backing source text.
+    ///
+    /// When present, it is used to verify that the source manager used during assembly is the
+    /// same one that was used to parse this module, via pointer identity
+    /// ([`SourceManager::is_manager_of`]).
+    source_file: Option<Arc<SourceFile>>,
 }
 
 /// Constants
@@ -152,7 +153,11 @@ impl Module {
 impl Module {
     /// Creates a new [Module] with the specified `kind` and fully-qualified path, e.g.
     /// `std::math::u64`.
-    pub fn new(kind: ModuleKind, path: impl AsRef<Path>, source_file: Arc<SourceFile>) -> Self {
+    pub fn new(
+        kind: ModuleKind,
+        path: impl AsRef<Path>,
+        source_file: Option<Arc<SourceFile>>,
+    ) -> Self {
         let path = path.as_ref().to_absolute().into_owned();
         Self {
             span: Default::default(),
@@ -167,21 +172,12 @@ impl Module {
 
     /// An alias for creating the default, but empty, `#kernel` [Module].
     pub fn new_kernel() -> Self {
-        Self::new(ModuleKind::Kernel, Path::kernel_path(), Self::empty_source_file())
+        Self::new(ModuleKind::Kernel, Path::kernel_path(), None)
     }
 
     /// An alias for creating the default, but empty, `$exec` [Module].
     pub fn new_executable() -> Self {
-        Self::new(ModuleKind::Executable, Path::exec_path(), Self::empty_source_file())
-    }
-
-    /// Creates a synthetic, empty [`SourceFile`] for programmatically constructed modules
-    /// that have no backing source text.
-    fn empty_source_file() -> Arc<SourceFile> {
-        Arc::new(SourceFile::from_raw_parts(
-            SourceId::default(),
-            SourceContent::new(SourceLanguage::Masm, Uri::from("synthetic"), String::new()),
-        ))
+        Self::new(ModuleKind::Executable, Path::exec_path(), None)
     }
 
     /// Specifies the source span in the source file in which this module was defined, that covers
@@ -458,9 +454,11 @@ impl Module {
         &self.advice_map
     }
 
-    /// Returns the source file from which this module was parsed.
-    pub fn source_file(&self) -> &Arc<SourceFile> {
-        &self.source_file
+    /// Returns the source file from which this module was parsed, if any.
+    ///
+    /// Returns `None` for programmatically-constructed modules that have no backing source text.
+    pub fn source_file(&self) -> Option<&Arc<SourceFile>> {
+        self.source_file.as_ref()
     }
 
     /// Get an iterator over the constants defined in this module.
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -5,7 +5,10 @@ use miden_core::{
     advice::AdviceMap,
     serde::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
 };
-use miden_debug_types::{SourceFile, SourceManager, SourceSpan, Span, Spanned};
+use miden_debug_types::{
+    SourceContent, SourceFile, SourceId, SourceLanguage, SourceManager, SourceSpan, Span, Spanned,
+    Uri,
+};
 use miden_utils_diagnostics::Report;
 use smallvec::SmallVec;
 
@@ -126,6 +129,11 @@ pub struct Module {
     pub(crate) items: Vec<Export>,
     /// AdviceMap that this module expects to be loaded in the host before executing.
     pub(crate) advice_map: AdviceMap,
+    /// The source file from which this module was parsed.
+    ///
+    /// This is used to verify that the source manager used during assembly is the same one that
+    /// was used to parse this module, via pointer identity ([`SourceManager::is_manager_of`]).
+    source_file: Arc<SourceFile>,
 }
 
 /// Constants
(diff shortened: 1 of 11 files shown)


crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -51,39 +51,55 @@ impl DebugInfoSections {
...
+        // If the source id doesn't exist in this source manager, skip debug info
+        // registration (the span may be synthetic or from a deserialized module).
+        if source_manager.get(span.source_id()).is_err() {
+            return Ok(());
         }
 
+        // The source id exists, so file_line_col should succeed. If it doesn't, the
+        // span's byte offsets are out of bounds for the file at that source id, which
+        // indicates the module was parsed with a different source manager.
+        let file_line_col = source_manager.file_line_col(span).map_err(|err| {
```
huitseeker: Could this still attach debug info to the wrong file when the assembler's manager has the same `SourceId` and that file is long enough? `SourceId` is only local to one manager, so `get` plus `file_line_col` proves the offset fits some file in this manager, but not that the span came from it. In that case the mismatch would still slip through as wrong debug metadata.

huitseeker (changes requested):

huitseeker (changes requested):
Please fix the false-negative.

juan518munoz:
@huitseeker 

Fixing the false negative required a broader change. As `SourceId` is just a `u32` index with no cross-manager uniqueness, so bounds checks alone can't distinguish "right file at index N" from "wrong file at same index."

The proposed fix in 1480deeae4e3 uses `is_manager_of`, which compares `Arc<SourceFile>` pointer identity, making false negatives impossible. To make the original `SourceFile` available at debug info registration time, we have to thread it through. now `Module` stores the `Arc<SourceFile>`, then `LinkModule` carries it through linking, and `register_procedure_debug_info` uses it for the `is_manager_of` check.

crates/assembly/src/assembler/debuginfo.rs
juan518munoz: Yes, you are right, handled in 1480dee

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -153,17 +161,27 @@ impl Module {
...
             items: Default::default(),
             advice_map: Default::default(),
+            source_file,
         }
     }
 
     /// An alias for creating the default, but empty, `#kernel` [Module].
     pub fn new_kernel() -> Self {
-        Self::new(ModuleKind::Kernel, Path::kernel_path())
+        Self::new(ModuleKind::Kernel, Path::kernel_path(), Self::empty_source_file())
```
huitseeker: I think this regresses in-memory AST modules. 

`new_kernel()` and `new_executable()` now attach a synthetic `SourceFile`, but later `register_procedure_debug_info()` treats any file not owned by the assembler's `SourceManager` as an error. 

I reproduced this with a programmatically built module containing one `nop` procedure: `Assembler::new(sm).assemble_library([module])` now returns `source manager mismatch` instead of just skipping debug info. 

Could we represent 'no backing source file' as `None` here?

crates/assembly/src/assembler.rs
```diff
@@ -953,8 +953,11 @@ impl Assembler {
                     // be added to the forest.
 
                     // Record the debug info for this procedure
-                    self.debug_info
-                        .register_procedure_debug_info(&procedure, self.source_manager.as_ref())?;
+                    self.debug_info.register_procedure_debug_info(
+                        &procedure,
+                        self.source_manager.as_ref(),
+                        module_source_file.as_deref(),
```
huitseeker, line 960: The source-manager check lands too late here. 

`compile_procedure()` already uses `self.source_manager` for instruction spans and diagnostics before we reach `register_procedure_debug_info()`, so a mismatched module can still take the wrong source path first. 

I reproduced that with a module parsed in one manager and assembled in another using `loc_loadw_be.2`: formatting the compile error panicked in `SourceFile::location`. We should reject the module before lowering starts.

huitseeker (changes requested):

crates/assembly/src/assembler.rs
juan518munoz: Moved the source-manager check to linking time so it covers this.

713e931

juan518munoz:
Thanks for the re-review @huitseeker, PR is ready to be reviewed again.

I was wondering if it would be preferable to point this to either `main` or other release branch.

huitseeker (approved):
LGTM, thanks for fixing this!

crates/assembly-syntax/src/ast/module.rs
juan518munoz: Forgot to reply here as well, but it's also been handled as of https://github.com/0xMiden/miden-vm/commit/713e931ec55111e3f7284adcae67c6f67c1d26b5

juan518munoz:
Thanks for the review, I've rebased into `main` so this can be later patch-released @huitseeker

EDIT: As a second though, this has breaking changes (public API change), so it may be better suited for `next`.

SantiagoPittella:
If this is the only commit that's going to a new patch version, and a maintainer of this repo agrees, maybe you can bump the crate version here too.

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -126,6 +126,15 @@ pub struct Module {
     pub(crate) items: Vec<Export>,
     /// AdviceMap that this module expects to be loaded in the host before executing.
     pub(crate) advice_map: AdviceMap,
+    /// The source file from which this module was parsed, if any.
+    ///
+    /// This is `None` for programmatically-constructed modules (e.g. `new_kernel()`,
+    /// `new_executable()`) that have no backing source text.
```
huitseeker, line 132: Reflecting @bitwalker 's review on #3007 :
> There are two issues with this:
>
> 1. Programmatically-constructed modules can absolutely have a backing source file - it just means that the source language was not Miden Assembly (or that the AST was literally constructed by hand for a test, but that's not the important case).
> 2. Source spans within a `Module` do _not_ have to come from the same source file, and in fact that is the common case when compiling from Rust source code currently.

huitseeker (commented):
Reflecting @bitwalker 's review on #3007:
> This PR makes some assumptions about how source files are related to modules which are incorrect, so I don't think it makes sense to merge it (or #2987).
> 
> I'm not sure this is even the correct way to handle this in general - if you've somehow gotten your source managers mixed up, you want to know exactly where that happened, and a panic is useful for that. If you instead silently drop the source locations, or try to bubble up an `Err`, you'll lose valuable context about where the wires got crossed.
> 
> IMO, this is really a documentation issue related to how `SourceManager` impls need to be used, both from the caller perspective, and in terms of threading them through internals of the VM or other components.

huitseeker (changes requested):
As above and as reviewed by @bitwalker in #3007.

bitwalker:
Closing this, since the best solution here was to improve the documentation to clarify how `SourceManager` is meant to be used in conjunction with `Assembler`, and what the edge cases are, which has been done and is released.

PR #2988, opened 2026-04-09
fix: merge procedure names with concrete-over-placeholder priority and deterministic tie-breaking

### LINKED ISSUE #2982: DebugInfo Metadata follow-ups
Finally, two things we should want to tackle in a posterior PR (because this isn't strictly limited to debug vars):
-  metadata from an external placeholder root must not override metadata from the concrete root it points to: If the concrete root has asm-op or debug-var metadata, that concrete metadata should win. Procedure names need the same rule, except that a placeholder name should stay as fallback when the concrete root has no name. 
- determinism. If two real roots give different names for the same digest, the result must not depend on merge order. A test should show that merging the same inputs in both orders picks the same final name.

_Originally posted by @huitseeker in https://github.com/0xMiden/miden-vm/pull/2955#pullrequestreview-4062262819_

giwaov:
## Summary

Adds procedure name merging to `MastForest::merge` with two rules from [#2982](https://github.com/0xMiden/miden-vm/issues/2982):

1. **Concrete wins over placeholder**: metadata from an external placeholder root no longer overrides metadata from the concrete root it points to. The concrete name wins; the placeholder name serves as fallback only when the concrete root has no name.
2. **Deterministic tie-breaking**: when two concrete roots provide different names for the same digest, the lexicographically smallest name is chosen so the result does not depend on merge order.

## Rationale

Issue #2982 identified that `MastForest::merge` does not merge procedure names at all, which means debug metadata is silently dropped during forest merging. This causes lost diagnostics information when multiple forests are combined. The fix ensures procedure names survive merging with sensible priority rules: concrete definitions take precedence over placeholders, and deterministic tie-breaking guarantees reproducible results regardless of input order.

## Changes

- **`core/src/mast/merger/mod.rs`**: added `merge_procedure_names()` that collects procedure names from all input forests, classifies each as concrete or placeholder based on whether the source root is external, then picks the winning name per the priority rules above. Called at the end of `merge_inner()`.
- **`core/src/mast/merger/tests.rs`**: three new tests:
  - `mast_forest_merge_concrete_name_wins_over_placeholder` -- concrete name beats placeholder in both merge orders
  - `mast_forest_merge_placeholder_name_as_fallback` -- placeholder name preserved when concrete root has no name
  - `mast_forest_merge_deterministic_procedure_names` -- same result regardless of merge order when two concrete roots disagree

## Test plan

- `cargo test -p miden-core mast_forest_merge_concrete_name_wins_over_placeholder` -- verifies concrete names win over placeholder names in both merge orders
- `cargo test -p miden-core mast_forest_merge_placeholder_name_as_fallback` -- verifies placeholder name is preserved when no concrete name exists
- `cargo test -p miden-core mast_forest_merge_deterministic_procedure_names` -- verifies deterministic output regardless of merge order
- All existing merger tests continue to pass (`cargo test -p miden-core mast_forest_merge`)

Closes #2982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Fixes
 - Rejected non-syscall references to exported kernel procedures in the linker ([#2902](https://github.com/0xMiden/miden-vm/issues/2902)).
 - Reverted the `MainTrace` typed row storage change that caused a large `blake3_1to1` trace-building regression ([#2949](https://github.com/0xMiden/miden-vm/pull/2949)).
+- Fixed `MastForest::merge` to transfer procedure names, prefer concrete root metadata over external placeholders, and choose the lexicographically smallest name when two concrete roots disagree ([#2982](https://github.com/0xMiden/miden-vm/issues/2982)).
 #### Bug Fixes
 
 - Replaced unsound `ptr::read` with safe unbox in panic recovery, removing UB from potential double-drop ([#2934](https://github.com/0xMiden/miden-vm/pull/2934)).
--- a/core/src/mast/merger/mod.rs
+++ b/core/src/mast/merger/mod.rs
@@ -1,6 +1,7 @@
-use alloc::{collections::BTreeMap, vec::Vec};
+use alloc::{collections::BTreeMap, sync::Arc, vec::Vec};
 
 use crate::{
+    LexicographicWord,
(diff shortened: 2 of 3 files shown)


huitseeker:
Thanks! We should re-open when #2982 is ready for work.

PR #2991, opened 2026-04-10
fix(assembly-syntax): canonicalize PathBuf::try_from(String)

KitaroMoha563:
## Describe your changes

- Canonicalize `PathBuf::try_from(String)` by delegating to `PathBuf::new`, so it matches `TryFrom<&str>` and `FromStr`.
- Add a regression test in `path_buf.rs` to ensure `PathBuf::try_from("foo::\"bar\"".to_string())` is equal to `PathBuf::try_from("foo::\"bar\"")` and canonicalizes to `foo::bar`.
- This fixes inconsistent equality/hashing behavior for semantically equivalent paths created from different conversion APIs.

- [x] Updated `CHANGELOG.md'

### DIFF
--- a/crates/assembly-syntax/src/ast/path/path_buf.rs
+++ b/crates/assembly-syntax/src/ast/path/path_buf.rs
@@ -282,8 +282,7 @@ impl TryFrom<String> for PathBuf {
     type Error = PathError;
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        Path::validate(&value)?;
-        Ok(PathBuf { inner: value })
+        PathBuf::new(&value)
     }
 }
 
@@ -572,6 +571,15 @@ mod tests {
             "::\"root_ns:root@1.0.0\"::abi_transform_tx_kernel_get_inputs_4"
         );
         assert_eq!(parent.parent().map(|p| p.as_str()), Some("::\"root_ns:root@1.0.0\""));
+    }
+
+    #[test]
+    fn try_from_string_canonicalizes_like_str() {
+        let from_str = PathBuf::try_from("foo::\"bar\"").unwrap();
+        let from_string = PathBuf::try_from(alloc::string::String::from("foo::\"bar\"")).unwrap();
+
+        assert_eq!(from_string, from_str);
+        assert_eq!(from_string.as_str(), "foo::bar");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/assembly-syntax/src/ast/path/path_buf.rs
```diff
@@ -282,8 +282,7 @@ impl TryFrom<String> for PathBuf {
     type Error = PathError;
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        Path::validate(&value)?;
-        Ok(PathBuf { inner: value })
+        PathBuf::new(&value)
```
bitwalker: This will always reallocate, we only want to reallocate if we need to

bitwalker (changes requested):

crates/assembly-syntax/src/ast/path/path_buf.rs
KitaroMoha563: @bitwalker Updated. Please take another look, thanks for your time.

crates/assembly-syntax/src/ast/path/path_buf.rs
```diff
@@ -278,12 +278,50 @@ impl<'a> TryFrom<&'a str> for PathBuf {
...
+            Err(_) => return false,
+        };
+
+        match component {
+            PathComponent::Root => is_absolute = true,
+            component @ PathComponent::Normal(_) => {
+                let is_quoted = component.is_quoted();
+                let component = component.as_str();
+                let is_special = matches!(component, Path::KERNEL_PATH | Path::EXEC_PATH);
+                let requires_quoting = !is_special && Ident::requires_quoting(component);
```
huitseeker, line 297: This repeats the quoted-vs-unquoted canonical form rule from `push_component`, but it also cannot just call `PathComponent::requires_quoting()` because this check intentionally uses `as_str()` after stripping quotes. 

Could we centralize that predicate or add a short note here? It would make future changes to path quoting less likely to drift between the two paths.

huitseeker (approved):
This LGTM modulo the issue left inline.

crates/assembly-syntax/src/ast/path/path_buf.rs
bitwalker: +1

bitwalker (approved):

PR #2992, opened 2026-04-11
ci: benchmark blake3 1-to-1 for regressions on main

huitseeker:
This adds a `main`-only non-regression benchmark for `blake3_1to1` on `warp-ubuntu-latest-x64-4x`.

The workflow builds and measures two clean revisions: a baseline commit and the pushed commit. Each side runs:

```bash
make exec-info
MIDEN_LOG=info RAYON_NUM_THREADS=8 ./target/optimized/miden-vm prove \
  ./miden-vm/masm-examples/hashing/blake3_1to1/blake3_1to1.masm \
  --release
```

It parses the timing tree from the prove output, compares the main stages and total prove time, uploads the raw logs, writes a job summary, and updates a commit comment on pushes to main. The job fails if the slowdown is above the configured threshold.

The benchmark job runs with read-only contents permission on the Warp runner. The commit comment is posted by a separate follow-up job on ubuntu-latest with contents: write. This keeps the machine that runs the benchmark on the narrowest permission set.

```mermaid
flowchart LR
  A[push to main] --> B[resolve baseline and current refs]
  B --> C[baseline clean build and prove]
  B --> D[current clean build and prove]
  C --> E[parse timings]
  D --> E
  E --> F[summary and artifacts]
  E --> G[commit comment]
  E --> H[fail if slowdown is over threshold]
```

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,5 @@
+# Allow actionlint to recognize the org's custom Warp runner label used by the
+# blake3 non-regression workflow.
+self-hosted-runner:
+  labels:
+    - warp-ubuntu-latest-x64-4x
--- a/.github/workflows/blake3-nonregression.yml
+++ b/.github/workflows/blake3-nonregression.yml
@@ -1,0 +1,318 @@
+name: blake3 1-to-1 non-regression
+
+on:
+  push:
+    branches:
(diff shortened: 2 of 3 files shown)


adr1anh (approved):
This looks great! Looking forward to getting these insights.

PR #2993, opened 2026-04-11
feat(assembly): warn about unused constants

### LINKED ISSUE #2898: Warn about unused constants

giwaov:
## Description

Adds a warning diagnostic for private constants that are declared but never referenced by any procedure or other constant in a module, helping developers catch dead definitions early.

Closes #2898

## Rationale

The assembler already emits warnings for unused imports (`UnusedImport`), but unused constants go silently unnoticed. This change applies the same pattern to constants: during semantic analysis, the compiler tracks which constants are actually referenced (by instructions in procedure bodies or by other constant expressions), and emits an `UnusedConstant` warning for any private constant that is never used. Public constants are excluded because they may be consumed by external modules.

## Changes

- **`crates/assembly-syntax/src/sema/errors.rs`** - Added `UnusedConstant` variant to `SemanticAnalysisError` with `severity(Warning)`
- **`crates/assembly-syntax/src/sema/context.rs`** - Added `used_constants: BTreeSet<Ident>` and `simplifying_constant: Option<Ident>` to `AnalysisContext` for tracking constant usage during evaluation; added `is_constant_used()` and `mark_constant_used()` helper methods
- **`crates/assembly-syntax/src/sema/mod.rs`** - Added unused-constant check loop after the existing unused-import check; marked advice-map-entry constants as used
- **`crates/assembly/src/tests.rs`** - Added three tests: `unused_constant_warning`, `used_constant_no_warning`, `public_constant_no_warning`

## Test plan

- `unused_constant_warning` - verifies the diagnostic output for a private constant that is never referenced
- `used_constant_no_warning` - verifies no warning when a constant is consumed via `push.MY_CONST`
- `public_constant_no_warning` - verifies no warning for `pub const` in a library module
- Existing tests `constant_alphanumeric_expression`, `enum_discriminants_can_reference_constants`, and `test_adv_has_map_key` continue to pass (constants referenced only by other constants, and advice-map-generated constants, are correctly treated as used)

### DIFF
--- a/crates/assembly-syntax/src/sema/context.rs
+++ b/crates/assembly-syntax/src/sema/context.rs
@@ -175,7 +175,7 @@ impl AnalysisContext {
     pub fn resolve_constant_usage(&mut self) {
         let mut worklist: VecDeque<Ident> = self.used_constants.iter().cloned().collect();
         for (name, constant) in &self.constants {
-            if constant.visibility.is_public() {
+            if constant.visibility.is_public() && self.used_constants.insert(name.clone()) {
                 worklist.push_back(name.clone());
             }
         }
--- a/crates/assembly-syntax/src/sema/context.rs
+++ b/crates/assembly-syntax/src/sema/context.rs
@@ -1,6 +1,6 @@
 use alloc::{
     boxed::Box,
-    collections::{BTreeMap, BTreeSet},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     sync::Arc,
     vec::Vec,
 };
@@ -18,8 +18,11 @@ use crate::ast::{
 pub struct AnalysisContext {
     constants: BTreeMap<Ident, Constant>,
     used_constants: BTreeSet<Ident>,
-    /// When set, `get()` will not count a reference to this constant as a use.
-    /// This is used during constant simplification to exclude self-references.
+    /// Edges from a constant to the constants it references.
+    /// Used to compute transitive usage from real roots.
+    constant_deps: BTreeMap<Ident, BTreeSet<Ident>>,
+    /// When set, references are recorded as constant-to-constant edges
+    /// rather than marking the target as directly used.
     simplifying_constant: Option<Ident>,
     imported: BTreeSet<Ident>,
     procedures: BTreeSet<ProcedureName>,
@@ -44,7 +47,11 @@ impl constants::ConstEnvironment for AnalysisContext {
         if let Some(constant) = self.constants.get(name) {
             let is_self_ref = self.simplifying_constant.as_ref() == Some(name);
             if !is_self_ref {
-                self.used_constants.insert(name.clone());
+                if let Some(ref parent) = self.simplifying_constant {
+                    self.constant_deps.entry(parent.clone()).or_default().insert(name.clone());
+                } else {
+                    self.used_constants.insert(name.clone());
+                }
             }
             Ok(Some(CachedConstantValue::Miss(&constant.value)))
         } else if self.imported.contains(name) {
@@ -76,6 +83,7 @@ impl AnalysisContext {
         Self {
             constants: Default::default(),
             used_constants: Default::default(),
+            constant_deps: Default::default(),
             simplifying_constant: None,
             imported: Default::default(),
             procedures: Default::default(),
@@ -120,6 +128,29 @@ impl AnalysisContext {
     /// (e.g. advice map entries) that should not trigger unused constant warnings.
     pub fn mark_constant_used(&mut self, name: &Ident) {
         self.used_constants.insert(name.clone());
+    }
+
+    /// Propagate usage from real roots through constant-to-constant edges.
+    ///
+    /// A constant is considered truly used if it is public, directly referenced
+    /// by a procedure body, or transitively referenced by such a constant.
+    /// This must be called before checking for unused constants.
+    pub fn resolve_constant_usage(&mut self) {
+        let mut worklist: VecDeque<Ident> = self.used_constants.iter().cloned().collect();
+        for (name, constant) in &self.constants {
+            if constant.visibility.is_public() {
+                worklist.push_back(name.clone());
+            }
+        }
+        while let Some(name) = worklist.pop_front() {
+            if let Some(deps) = self.constant_deps.get(&name) {
+                for dep in deps {
+                    if self.used_constants.insert(dep.clone()) {
+                        worklist.push_back(dep.clone());
+                    }
+                }
+            }
+        }
     }
 
     /// Define a new constant `constant`
--- a/crates/assembly-syntax/src/sema/context.rs
+++ b/crates/assembly-syntax/src/sema/context.rs
@@ -21,6 +21,9 @@ pub struct AnalysisContext {
     /// Edges from a constant to the constants it references.
     /// Used to compute transitive usage from real roots.
     constant_deps: BTreeMap<Ident, BTreeSet<Ident>>,
+    /// Edges from a constant to the imports it references.
+    /// Used to avoid marking imports as used when only dead constants reference them.
+    constant_import_refs: BTreeMap<Ident, BTreeSet<alloc::string::String>>,
     /// When set, references are recorded as constant-to-constant edges
     /// rather than marking the target as directly used.
     simplifying_constant: Option<Ident>,
@@ -84,6 +87,7 @@ impl AnalysisContext {
             constants: Default::default(),
             used_constants: Default::default(),
             constant_deps: Default::default(),
+            constant_import_refs: Default::default(),
             simplifying_constant: None,
             imported: Default::default(),
             procedures: Default::default(),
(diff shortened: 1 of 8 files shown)


crates/assembly-syntax/src/sema/context.rs
```diff
@@ -38,6 +42,10 @@ impl constants::ConstEnvironment for AnalysisContext {
     #[inline]
     fn get(&mut self, name: &Ident) -> Result<Option<CachedConstantValue<'_>>, Self::Error> {
         if let Some(constant) = self.constants.get(name) {
+            let is_self_ref = self.simplifying_constant.as_ref() == Some(name);
+            if !is_self_ref {
```
huitseeker, line 52: Because `get()` runs during `simplify_constants()`, this marks `A` as used even in a dead chain like `const A = 1` and `const B = A` where neither constant is referenced from any procedure or exported API. In that case only `B` warns and `A` is missed. 

To really address this, we need to track constant-to-constant edges here and compute reachability from real roots.

huitseeker (changes requested):

crates/assembly-syntax/src/sema/context.rs
giwaov: Good catch - fixed in 5cc929c. Instead of marking constants as used during \get()\, I now track constant-to-constant edges in a separate map and then do a BFS from real roots (procedure references + public constants) to figure out which ones are actually reachable. So \const A = 1; const B = A\ correctly warns on both if neither is referenced.

Added tests for both cases - both unused should warn, and transitive usage through a chain shouldn't.

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -171,6 +171,14 @@ pub fn analyze(
         }
     }
 
+    // Check unused constants
+    analyzer.resolve_constant_usage();
```
huitseeker: I could repro a small mismatch here with a local test. A program shaped like 
````
use lib::a
const DEAD = a::BAR
begin push.1 end
````
 only emits `unused constant` when warnings are promoted to errors, and it skips the matching `unused import` warning. 

The reason seems to be that `visit_mut_constant_ref()` increments `alias.uses` before this liveness pass proves the referencing constant is live, so imports reached only from dead constants stay marked as used.

Comment (giwaov) [reply crates/assembly-syntax/src/sema/mod.rs]: Nice catch - fixed in fda2ca5. The visit_mut_constant_ref path in VerifyInvokeTargets was bumping alias.uses unconditionally, so imports reached only through dead constants slipped past the unused-import check.

Now constant exports record their import references as deferred edges in AnalysisContext instead of incrementing directly. After resolve_constant_usage() runs, only imports from live constants get credited. Moved the unused-import check after constant liveness resolution too.

Added dead_constant_does_not_mask_unused_import to cover the exact scenario you described.

File: crates/assembly-syntax/src/sema/passes/verify_invoke.rs [hunk 3]
```diff
@@ -282,10 +291,16 @@ impl VisitMut for VerifyInvokeTargets<'_> {
     fn visit_mut_constant_ref(&mut self, path: &mut Span<Arc<Path>>) -> ControlFlow<()> {
         if let Some(name) = path.as_ident() {
             self.track_used_alias(&name);
```
huitseeker: The direct-import path still bypasses the new deferral. When a dead constant uses an imported constant alias like `use lib::a::BAR` followed by `const DEAD = BAR`, this branch calls `track_used_alias(&name)` immediately, so the import is treated as used and the expected `unused import` warning never fires.

This wants the same deferral logic for identifier refs that resolve to imports, plus a regression test for direct and renamed constant imports.

crates/assembly-syntax/src/sema/context.rs
```diff
@@ -98,6 +120,76 @@ impl AnalysisContext {
...
+            .entry(constant_name.clone())
+            .or_default()
+            .insert(import_name);
+    }
+
+    /// Increment `alias.uses` only for imports that are referenced by live
+    /// constants. Must be called after `resolve_constant_usage`.
+    pub fn apply_live_constant_import_refs(&self, module: &mut Module) {
+        for (constant_name, import_names) in &self.constant_import_refs {
+            if !self.used_constants.contains(constant_name) {
```
huitseeker, line 157: The liveness check here skips `pub const` roots because `resolve_constant_usage()` only pushes public constants onto the worklist, it does not insert their own names into `used_constants`. 

That means `pub const LIVE = a::BAR` still leaves `use lib::a` looking unused even though the import is needed to evaluate an exported constant.

Marking public constants as live before applying deferred import refs, and adding a regression test for `pub const LIVE = a::BAR`, should close the hole.

huitseeker (changes requested):

giwaov:

1. **Direct-import deferral** — \isit_mut_constant_ref\ now defers \	rack_used_alias\ for identifier refs inside constant bodies (matching the existing deferral for qualified paths). \use lib::a::BAR\ + \const DEAD = BAR\ now correctly reports both unused-import and unused-constant.

2. **Public constants marked live** — \
esolve_constant_usage()\ now inserts public constants into \used_constants\ (not just the worklist), so \pply_live_constant_import_refs\ correctly credits their import references. \pub const LIVE = a::BAR\ no longer triggers a spurious unused-import warning.

Added regression tests for both cases. All 186 assembly unit tests and 104 assembly-syntax tests pass.

crates/assembly-syntax/src/sema/passes/verify_invoke.rs
```diff
@@ -281,11 +290,23 @@ impl VisitMut for VerifyInvokeTargets<'_> {
     }
     fn visit_mut_constant_ref(&mut self, path: &mut Span<Arc<Path>>) -> ControlFlow<()> {
         if let Some(name) = path.as_ident() {
-            self.track_used_alias(&name);
-        } else if let Some((module, _)) = path.split_first()
-            && let Some(alias) = self.module.aliases_mut().find(|a| a.name().as_str() == module)
-        {
-            alias.uses += 1;
+            if let Some(ref const_name) = self.current_constant {
```
huitseeker, line 298: Could we check that `name` is actually an import before recording it here? 

Local constants can reuse an imported alias name: `use lib::a::FOO` plus `const FOO = 1` makes `get()` resolve `FOO` as the local constant, but this branch still records `FOO` as an import ref. If a live constant references that local `FOO`, `apply_live_constant_import_refs()` increments the import use and hides the unused-import warning for the shadowed import.

huitseeker (changes requested):

crates/assembly-syntax/src/sema/passes/verify_invoke.rs
giwaov: Fixed in 3e0983975.

The guard is in two places in `verify_invoke.rs`:

1. **`track_used_alias`** — now calls `analyzer.get_constant(name)` first and returns early if the name resolves to a local constant, so a same-named import is never credited as used.

2. **`visit_mut_constant_ref` (identifier branch)** — the deferred `record_constant_import_ref` call is now also skipped when the identifier resolves to a local constant, so dead-constant→shadowed-import edges don't survive into liveness resolution either.

Added a regression test (`local_constant_shadowing_import_warns_unused_import`) that assembles a module with both `use lib::a::FOO` and `const FOO = 1` and asserts the import is reported unused.

crates/assembly-syntax/src/sema/context.rs
```diff
@@ -38,6 +48,14 @@ impl constants::ConstEnvironment for AnalysisContext {
     #[inline]
     fn get(&mut self, name: &Ident) -> Result<Option<CachedConstantValue<'_>>, Self::Error> {
         if let Some(constant) = self.constants.get(name) {
+            let is_self_ref = self.simplifying_constant.as_ref() == Some(name);
+            if !is_self_ref {
+                if let Some(ref parent) = self.simplifying_constant {
+                    self.constant_deps.entry(parent.clone()).or_default().insert(name.clone());
```
huitseeker: I think this liveness edge only gets recorded for bare names. A same-module qualified path still goes through `get_by_path()` and returns `Ok(None)`, so `pub const B = ::test::lib::A` leaves private `A` looking unused.

giwaov: Fixed in 1b3daf3c.

`get_by_path()` now recognizes same-module qualified constants and routes them through the same liveness tracking as bare local constants. I also added a regression for `pub const B = ::test::lib::A`.

huitseeker (approved):
This LGTM, modulo a quick rebase to fix the CHANGELOG conflicts.

bobbinth:
@giwaov - this PR should probably go into the `next` branch.

giwaov:
Thanks, moved it over to 
ext and rebased on the latest branch.

crates/assembly-syntax/src/sema/context.rs
```diff
@@ -38,9 +49,12 @@ impl constants::ConstEnvironment for AnalysisContext {
     }
     #[inline]
     fn get(&mut self, name: &Ident) -> Result<Option<CachedConstantValue<'_>>, Self::Error> {
-        if let Some(value) = self.cached_constant_values.get(name) {
-            Ok(Some(CachedConstantValue::Hit(value)))
-        } else if let Some(constant) = self.constants.get(name) {
+        if self.constants.contains_key(name) {
+            self.record_constant_ref(name);
+            if let Some(value) = self.cached_constant_values.get(name) {
```
huitseeker, line 56: Could we make sure a cached constant still records its outgoing deps before this early return? I can reproduce a false warning with `const A = B`, `const B = C`, `const C = 1`, then `push.B`: simplifying `A` caches `B`/`C`, so simplifying `B` hits this return and never records `B -> C`. The liveness pass then warns that `C` is unused even though `B` uses it.

huitseeker (changes requested):

giwaov:
Thanks, fixed this. The dependency tracking now follows the constant currently being expanded by the evaluator, so when B is evaluated through A it still records the B -> C edge. I added the A/B/C + push.B regression as well.

Checks run:
- cargo test -p miden-assembly cached_chained_constant_records_transitive_dependency
- cargo test -p miden-assembly unused_constant
- cargo test -p miden-assembly-syntax sema
- rustfmt --check crates/assembly-syntax/src/sema/context.rs crates/assembly/src/tests.rs

huitseeker:
Closing for inactivity

PR #2985, opened 2026-04-09
test(assembly),deny: reduce merge drift from main

huitseeker:
This is merge-prep for the upcoming `main` -> `next` reconciliation (see #2878). Another PR against main (#2984) takes care of the rest.

  Changes:
  - refresh the protocol kernel assembly fixtures to the current `next`-compatible shape
  - remove stale `anstream` / `anstyle-parse` skips from `deny.toml`

  Notes:
  - I intentionally did not port the `main`-only `exec.$kernel::init` fixture lines
  - I also did not port the extra `miden-core` workspace wiring from `main`, because that currently breaks the packaging test on `next`

PR #2984, opened 2026-04-09
fix: preassembled runtime deps on main

huitseeker:
This is merge-prep for the upcoming `main` -> `next` reconciliation (see #2878).

#2950 contained some adjustments of the package PR (originally #2877) for `next`.

The present PR cherry-picks those adjustments to main, and is by consequence expected to remove 5 of the 18 conflicts on the tracking release branch. Another PR against next (#2985) will take care of the rest.

PR #2998, opened 2026-04-13
Merge main into next

huitseeker:
This replaces #2984 and #2985 by a straight merge `main` -> `next` that keeps next where next was already ahead, takes main where there was no semantic drift, and leaves the remaining protocol fixture conflicts in the exact state described by #2984 and #2985.

Cost/benefit:

- no need to review 2 PRs, approach guarantees cleaner merge from the git side,
- but PRs like #2955 will not benefit from the changes back-ported by #2984


adr1anh (approved):
LGMT

PR #3002, opened 2026-04-14
Constraint simplification

huitseeker:
Cleans up the integration test into a proper binary, adds a constraint check to CI + cleans up a bit of formatting.


adr1anh (approved):
TYVM! This is so much nicer than the previous hack

PR #3007, opened 2026-04-14
fix: panic on assembler source manager mismatch (semver-compat)

huitseeker:
This is a version of #2987 that is semver-compatible.

### DIFF
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -128,12 +128,7 @@ pub struct Module {
     pub(crate) advice_map: AdviceMap,
     /// The source file from which this module was parsed, if any.
     ///
-    /// This is `None` for programmatically-constructed modules (e.g. `new_kernel()`,
-    /// `new_executable()`) that have no backing source text.
-    ///
-    /// When present, it is used to verify that the source manager used during assembly is the
-    /// same one that was used to parse this module, via pointer identity
-    /// ([`SourceManager::is_manager_of`]).
+    /// Programmatically-constructed modules do not have a backing source file.
     source_file: Option<Arc<SourceFile>>,
 }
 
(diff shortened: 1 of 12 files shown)


Al-Kindi-0 (approved):
Assuming only the semver change is different from #2987

Al-Kindi-0 (changes requested):
Seems this will need a deeper look by @bitwalker

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -126,6 +126,10 @@ pub struct Module {
     pub(crate) items: Vec<Export>,
     /// AdviceMap that this module expects to be loaded in the host before executing.
     pub(crate) advice_map: AdviceMap,
+    /// The source file from which this module was parsed, if any.
+    ///
+    /// Programmatically-constructed modules do not have a backing source file.
+    source_file: Option<Arc<SourceFile>>,
```
bitwalker, line 132: There are two issues with this:

1. Programmatically-constructed modules can absolutely have a backing source file - it just means that the source language was not Miden Assembly (or that the AST was literally constructed by hand for a test, but that's not the important case).
2. Source spans within a `Module` do _not_ have to come from the same source file, and in fact that is the common case when compiling from Rust source code currently.

bitwalker (changes requested):
This PR makes some assumptions about how source files are related to modules which are incorrect, so I don't think it makes sense to merge it (or #2987).

I'm not sure this is even the correct way to handle this in general - if you've somehow gotten your source managers mixed up, you want to know exactly where that happened, and a panic is useful for that. If you instead silently drop the source locations, or try to bubble up an `Err`, you'll lose valuable context about where the wires got crossed.

IMO, this is really a documentation issue related to how `SourceManager` impls need to be used, both from the caller perspective, and in terms of threading them through internals of the VM or other components.

huitseeker:
@bitwalker OK, let's continue the discussion in #2987.

PR #3008, opened 2026-04-15
fix(core-lib): stop AEAD encrypt from writing padding over source-adjacent memory

huitseeker:
This fixes AEAD encrypt so the final padding block is written at the end of the destination buffer, not into memory after the source plaintext.

The padding path used the source-final pointer instead of the destination-final pointer. That let `encrypt` silently overwrite unrelated memory stored right after the plaintext buffer.

The fix writes the clear padding block to `dst_ptr + num_blocks * 8` before producing the encrypted padding block. The plaintext buffer is left unchanged. Decrypt still writes the clear padding  block at the end of the plaintext output.

The tests now cover both the one-block case and the `num_blocks = 0` case, and confirm that source-side memory is not modified.


crates/lib/core/asm/crypto/aead.masm
```diff
@@ -331,29 +336,36 @@ pub proc decrypt
     movdn.11
     # => [nonce(4), key(4), src_ptr, dst_ptr, num_blocks, tag_ptr, ...]
 
+    # compute the pointer at which clear padding should be written on success
+    dup.10 mul.8
+    dup.10
+    add
+    movdn.11
+    # => [nonce(4), key(4), src_ptr, dst_ptr, num_blocks, padding_ptr, tag_ptr, ...]
```
adr1anh: I would write a hint `padding_ptr = dst_ptr + 8 * num_blocks`

adr1anh (approved):
LGTM

crates/lib/core/asm/crypto/aead.masm
```diff
@@ -185,36 +186,40 @@ pub proc encrypt
...
     mem_storew_le dropw
     # => [dst_ptr_final, rate(8), capacity(4), src_ptr_final, dst_ptr_final, ...]
 
-    # Write second word [0,0,0,0] to dst_ptr_final + 4
     padw
-    # => [0, 0, 0, 0, dst_ptr_final, rate(8), capacity(4), src_ptr_final, dst_ptr_final, ...]
     movup.4 add.4
-    # => [dst_ptr_final+4, 0, 0, 0, 0, rate(8), capacity(4), src_ptr_final, dst_ptr_final, ...]
     mem_storew_le dropw
     # => [rate(8), capacity(4), src_ptr_final, dst_ptr_final, ...]
```
Al-Kindi-0: with the manual substitute for `crypto_stream`, this block is now not needed I believe. I think this is probably a better API, namely the input plaintext shouldn't be touched.

Another related question, should we do the same for the decryption procedure ?

Al-Kindi-0 (approved):
LGTM
One question inline regarding the `decrypt` procedure

crates/lib/core/asm/crypto/aead.masm
huitseeker: Good catch. We no longer need to materialize the clear padding block

PR #3009, opened 2026-04-16
fix: avoid unbound variable errors for NO_CHANGELOG_LABEL and BASE_REF

Ocheretovich:
Rationale: With set -u, if NO_CHANGELOG_LABEL=true and BASE_REF is not set, the script was failing before the early return could exit successfully. Moving the BASE_REF check below the early return fixes this.


scripts/check-changelog.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 set -uo pipefail
 
+: "${BASE_REF:?BASE_REF is not set}"
```
huitseeker, line 4: Could this check move below the `NO_CHANGELOG_LABEL` early return? With `set -u`, `NO_CHANGELOG_LABEL=true` and no `BASE_REF` now fails here before the no-changelog bypass can return success.

huitseeker (changes requested):
BASE_REF is now required before checking NO_CHANGELOG_LABEL, so a run with
NO_CHANGELOG_LABEL=true and no BASE_REF will fail instead of taking the intended successful early exit for “no changelog” PRs.

scripts/check-changelog.sh
Ocheretovich: @huitseeker fixed, please re-review

huitseeker (approved):
Thanks, LGTM!

PR #2962, opened 2026-04-05
All-LogUp auxiliary trace bus constraints

adr1anh:
# All-LogUp Auxiliary Trace

## What this PR does

Every auxiliary-trace bus in the Miden VM used to be described by a bespoke constraint: some were running-product (multiset) checks written out by hand, others were partial LogUp arguments. Each lived in its own per-component file, and the aux-trace generation code was spread across `processor/src/trace/{decoder,range,stack,chiplets}/aux_trace*.rs`, duplicating the algebra row-by-row in imperative Rust.

This PR replaces all of that with a single unified LogUp formulation, described declaratively. Every bus — the block hash table, op-group table, stack overflow, range check, chiplet request bus, sibling table, hasher perm-link, ACE wiring, kernel ROM, and the rest — is now a branch of one descriptive AIR (`MidenLookupAir`) expressed against a small closure-based builder trait. The same description is consumed by the verifier's constraint evaluator, by the prover's aux-trace generator, by the recursive-verifier ACE circuit, and by two debug walkers.

The practical consequences:

- The final layout is **4 extension columns on the main trace** and **3 on the chiplet trace** (seven total). The 4+3 split is deliberate: in a future multi-trace setup, the main-trace column count lands under the next power-of-two boundary without padding.
- **Only one committed boundary value per trace** (two total, with a forced-zero placeholder in the second slot until trace-splitting lands — see `NUM_LOGUP_COMMITTED_FINALS`). Seven column accumulators collapse into one partial-sum column per trace.
- The **kernel ROM chiplet** is now purely lookup-based. Previously it duplicated a row per callsite; now one row per digest carries a multiplicity. For realistic kernels this deletes thousands of chiplet rows.
- The per-component aux-trace Rust code is **gone**, not dual-maintained. Aux-trace generation now walks the same `LookupAir::eval` description the constraint side does.

## How the code is laid out

There are two halves, and it's worth keeping them separate in your head while reading.

**The generic LogUp framework** lives in `air/src/lookup/`. It is deliberately free of any Miden-specific types — the intent is to eventually extract it into its own crate. You'll find the closure-based builder stack (`LookupBuilder` → `LookupColumn` → `LookupGroup` → `LookupBatch`), a `LookupAir` trait modelled on `p3_air::Air`, and three concrete walkers of a `LookupAir` description:

- `ConstraintLookupBuilder` — symbolic evaluation for the verifier and the prover's quotient pass.
[...]

PR #3013, opened 2026-04-17
refactor(air,processor): collapse kernel ROM chiplet to multiplicity-based layout

adr1anh:
Stacked on top of #2962.

## Summary

Under all-LogUp, the permutation-era kernel ROM shape is redundant: soundness follows from multiset equality on the chiplets bus rather than from structural constraints on the trace. This PR collapses the chiplet to **one row per declared kernel procedure**, carrying the SYSCALL multiplicity in column 0.

- **Two bus labels preserved.** INIT anchors every chiplet row to a public-input-declared procedure (without it, a prover could forge a digest row with `m = k` and balance it against `k` forged decoder removes). CALL carries the per-procedure syscall count.
- **Two fractions per active row.** The chiplet now emits `1/(α - INIT(digest))` plus `m/(α - CALL(digest))`. `MAX_INTERACTIONS_PER_ROW` on the chiplet-responses bus bumps from 1 to 2.
- **All three shape constraints deleted:** binary `s_first`, digest contiguity, and the entry-row anchor. `air/src/constraints/chiplets/kernel_rom.rs` is removed entirely, along with the now-unused `ChipletSelectors::kernel_rom` flag.
- **Trace length drops** from `K + Σ num_accesses` to `K` (K = number of declared kernel procedures).

Bus math still balances: `K` chiplet INIT adds vs. `K` public-input INIT removes; `Σ m` chiplet CALL adds vs. `Σ num_syscalls` decoder CALL removes → forces `m = num_accesses` per row.

## Changes

- `air/src/constraints/logup_msg.rs` — add `KernelRomMsg::init` constructor; delete `KernelRomResponseMsg` (no longer need runtime-muxed labels)
- `air/src/constraints/chiplets/columns.rs` — rename `KernelRomCols::s_first` → `multiplicity`
- `air/src/constraints/chiplets/mod.rs` — drop kernel-ROM constraint call site
- `air/src/constraints/chiplets/kernel_rom.rs` — **deleted**
- `air/src/constraints/chiplets/selectors.rs` — remove unused `kernel_rom` flag + intermediate bindings
- `air/src/constraints/lookup/buses/chiplet_responses.rs` — two `g.add`/`g.insert` calls replacing the muxed single add; bump `MAX_INTERACTIONS_PER_ROW`
- `air/src/lib.rs` — update `kernel_proc_message` doc (boundary correction semantics unchanged)
- `processor/src/trace/chiplets/kernel_rom/mod.rs` — rewrite `fill_trace` to one row per proc; `trace_len` derives from `access_map.len()`
- `processor/src/trace/chiplets/kernel_rom/tests.rs` — rewrite for new layout (m=0, m=1, m>1)
- `docs/src/design/chiplets/kernel_rom.md` — rewrite to describe the new layout and two-label rationale

## Test plan

- [x] `make test-air` — 52 passed
- [x] `make test-processor` — 1474 passed (fragment-boundary snapshots for SYSCALL programs regenerated: case_13, case_14)
- [x] `make test-prover` — 2 passed (end-to-end bus-balance validation)
- [x] `cargo clippy -p miden-air -p miden-processor --all-targets` — clean
- [x] `insta` snapshot + `RELATION_DIGEST` regenerated via `cargo run -p miden-core-lib --features constraints-tools --bin regenerate-constraints -- --write`


adr1anh:
One question to answer is whether we want to put the multiplicities in the VLPI, as this removes the chiplet entirely. Unfortunately this would leak the number of calls to each kernel method.

bobbinth:
> One question to answer is whether we want to put the multiplicities in the VLPI, as this removes the chiplet entirely. Unfortunately this would leak the number of calls to each kernel method.

We should keep the number of calls to each kernel procedure private (that was one of the intents old design). Otherwise, it would be very difficult to claim "functional privacy" and in fact, may make inferring types of transactions users do relatively easy.

air/src/constraints/lookup/buses/chiplet_responses.rs
```diff
@@ -390,17 +391,27 @@ pub(in crate::constraints::lookup) fn emit_chiplet_responses<LB>(
                         Deg::NONE,
                     );
 
-                    // Kernel ROM: runtime-muxed s_first → label.
+                    // Kernel ROM: one row per declared procedure, two fractions per active row.
+                    // INIT-labeled add (multiplicity 1) anchors each row to a public-input-declared
+                    // procedure. CALL-labeled add carries the syscall multiplicity from column 0.
+                    let kernel_gate = ctx.chiplet_active.kernel_rom.clone();
+                    let krom_mult: LB::Expr = krom.multiplicity.into();
                     g.add(
```
adr1anh: Change this to remove, so the outer interactions can use 1.

PR #3014, opened 2026-04-17
fix(parallel): align stack word max index bound

mk0walsk:
• ## Describe your changes
                                                                                                                                                                                           
  This PR fixes a boundary-contract mismatch in parallel stack word access.                                                                                                                
                                                                                                                                                                                           
  - Updated `ReplayProcessor::get_word` and `ReplayProcessor::set_word` to allow `start_idx == MIN_STACK_DEPTH - WORD_SIZE` (i.e. max valid index `12`), matching `StackInterface` and fast-path behavior.
  - Added regression test `stack_set_word_allows_max_start_idx` to ensure max-index writes/reads remain valid.                                                                             
  - Related: #2802                                                                                                                                                                         
                                                                                                                                                                                           
  - [x] Updated `CHANGELOG.md`.


processor/src/trace/parallel/processor.rs
```diff
@@ -255,7 +255,7 @@ impl StackInterface for ReplayProcessor {
     }
 
     fn get_word(&self, start_idx: usize) -> Word {
-        debug_assert!(start_idx < MIN_STACK_DEPTH - 4);
+        debug_assert!(start_idx <= MIN_STACK_DEPTH - WORD_SIZE);
```
huitseeker: Would it be clearer to write this as `debug_assert!(start_idx + WORD_SIZE <= MIN_STACK_DEPTH)`? That states the full-word-fit invariant directly, which makes the later offset math easier to check at a glance.

processor/src/trace/parallel/processor.rs
```diff
@@ -274,7 +274,7 @@ impl StackInterface for ReplayProcessor {
     }
 
     fn set_word(&mut self, start_idx: usize, word: &Word) {
-        debug_assert!(start_idx < MIN_STACK_DEPTH - 4);
+        debug_assert!(start_idx <= MIN_STACK_DEPTH - WORD_SIZE);
```
huitseeker: Related to this bound, would it help to switch the next line to `MIN_STACK_DEPTH - start_idx - WORD_SIZE` instead of `- 4`? Using `WORD_SIZE` in both places makes the offset formula much easier to audit.

processor/src/trace/parallel/processor.rs
```diff
@@ -506,3 +506,45 @@ impl Stopper for ReplayStopper {
...
+            MastForestResolutionReplay::default(),
+            1_u32.into(),
+        )
+    }
+
+    #[test]
+    fn stack_set_word_allows_max_start_idx() {
+        let mut processor = build_replay_processor();
+        let start_idx = MIN_STACK_DEPTH - WORD_SIZE;
+        let word = [Felt::new(1), Felt::new(2), Felt::new(3), Felt::new(4)].into();
```
huitseeker: Since this test is about the `WORD_SIZE` boundary, would it make sense to build the sample word from `WORD_SIZE` too instead of a 4-element literal? That keeps the test aligned with the same constant the implementation is checking.

huitseeker (commented):
We could introduce a helper, Something like:
```rust
#[inline(always)]
fn word_start_idx(stack_len: usize, start_idx: usize) -> usize {
    stack_len - start_idx - WORD_SIZE
}
```

Then both replay and fast paths become obviously the same calculation.

processor/src/trace/parallel/processor.rs
mk0walsk: Switched the bound check to 'start_idx + WORD_SIZE <= MIN_STACK_DEPTH' to make the full-word-fit invariant explicit.

processor/src/trace/parallel/processor.rs
mk0walsk: Updated the boundary test word construction to derive from WORD_SIZE instead of a 4-element literal.

processor/src/trace/parallel/processor.rs
mk0walsk: Replaced hardcoded '-4' with WORD_SIZE-based index calculation via a shared helper in replay path.

huitseeker (approved):
TYVM!

bobbinth (approved):
Looks good! Thank you!

bobbinth:
Seems like merging the latest `next` into this branch resulted in CI errors. @mk0walsk, could you resolve them?

PR #3018, opened 2026-04-19
ci: tune warpbuild runners and taiki install pins

huitseeker:
This moves the WarpBuild jobs from warp-ubuntu-latest-x64-4x to warp-ubuntu-latest-x64-8x.

It also renames the visible test job from the full runner label to test on warpbuild, so future runner size changes do not rename the required check.

The zizmor failure came from pinning taiki-e’s moving tool shortcut tags by hash. Those tags move on each release, so old hashes can later look like impostor commits. This now pins a released install-action commit and sets the tool name with tool:.

actionlint.yaml was updated only to allow the new custom runner label. It does not run actionlint on WarpBuild.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -2,4 +2,4 @@
 # blake3 non-regression workflow.
 self-hosted-runner:
   labels:
-    - warp-ubuntu-latest-x64-4x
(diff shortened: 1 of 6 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #3021, opened 2026-04-20
Fix SHA256 word range checks and U32ADD carry constraints

huitseeker:
This PR fixes [TOB-MIDENCORE-4](https://github.com/0xMiden/security-issues/issues/5) and [TOB-MIDENCORE-9](https://github.com/0xMiden/security-issues/issues/8).

The SHA256 fix checks that all message words are valid u32 values before they are used by the SHA256 schedule. This covers hash, merge, and hash_bytes, including words loaded from memory.

The U32ADD fix enforces that the high carry helper limb h3 is zero for U32ADD and U32ADD3. This prevents forged carry outputs such as 65536, including the carry path used by `u64::overflowing_add`.

The verifier artifacts and relation digest constants were updated to match the new AIR. This includes the RELATION_DIGEST constants in Rust and MASM, the AIR snapshot, and the generated recursive verifier constraint program.

Tests were added for rejected SHA256 non-u32 message words, forged U32ADD and U32ADD3 carries, and the `u64::overflowing_add` carry propagation case.


docs/src/design/stack/u32_ops.md
```diff
@@ -121,6 +121,10 @@ $$
 s_1' = h_2 \text{ | degree} = 1
 $$
 
+$$
+h_3 = 0 \text{ | degree} = 1
+$$
+
```
Al-Kindi-0, line 127: Very strange, my memory tells me that we had these constraints in the docs. Maybe in some of the local work I was doing on tagging design docs, but not sure.

Al-Kindi-0 (approved):
LGTM, thank you!

adr1anh (approved):
Looks good. These tests using AirBuilder are great, and we should look at integrating these for other parts of our constraints.

PR #3023, opened 2026-04-20
ci: enable stricter clippy lint set

huitseeker:
Opt xclippy and xclippy-fix into needless_collect, redundant_clone, redundant_closure_for_method_calls, trivially_copy_pass_by_ref, uninlined_format_args, and unused_qualifications.

Apply the required workspace cleanup manually so the stricter check passes without warnings, including generated-parser lint expectations for warnings emitted from LALRPOP output.

This is then enforced by the `make lint` + CI targets.

### DIFF
--- a/crates/assembly-syntax/src/ast/instruction/print.rs
+++ b/crates/assembly-syntax/src/ast/instruction/print.rs
@@ -413,9 +413,9 @@ mod tests {
         assert_eq!("push.3.4.8.9", instruction);
         let instruction = format!(
             "{}",
-            Instruction::Push(Immediate::Value(miden_debug_types::Span::unknown(
-                crate::parser::PushValue::Int(crate::parser::IntValue::U8(3))
-            )))
+            Instruction::Push(Immediate::Value(Span::unknown(crate::parser::PushValue::Int(
+                crate::parser::IntValue::U8(3)
+            ))))
         );
         assert_eq!("push.3", instruction);
 
--- a/crates/assembly-syntax/src/ast/instruction/print.rs
+++ b/crates/assembly-syntax/src/ast/instruction/print.rs
@@ -390,7 +390,11 @@ mod tests {
     use miden_core::crypto::hash::Poseidon2;
     use miden_debug_types::Span;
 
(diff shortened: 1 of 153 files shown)


crates/assembly-syntax/src/ast/instruction/print.rs
```diff
@@ -413,9 +413,9 @@ mod tests {
         assert_eq!("push.3.4.8.9", instruction);
         let instruction = format!(
             "{}",
-            Instruction::Push(Immediate::Value(miden_debug_types::Span::unknown(
-                crate::parser::PushValue::Int(crate::parser::IntValue::U8(3))
-            )))
+            Instruction::Push(Immediate::Value(Span::unknown(crate::parser::PushValue::Int(
+                crate::parser::IntValue::U8(3)
+            ))))
```
Al-Kindi-0: would it be better to bring the parser in scope ? e.g., `crate::parser::IntValue::U8(3)` becomes `IntValue::U8(3)`

crates/assembly-syntax/src/testing/pattern.rs
```diff
@@ -37,11 +37,9 @@ impl Pattern {
...
             panic!(
                 r"expected string was not found in emitted diagnostics:
-expected input to {expected}
-matched against: `{actual}`
-",
-                expected = self,
-                actual = input
+expected input to {self}
+matched against: `{input}`
+"
```
Al-Kindi-0: nit(not related to this PR): would be nice to format these properly

crates/assembly-syntax/src/testing/pattern.rs
```diff
@@ -55,12 +53,10 @@ matched against: `{actual}`
...
                 r"expected string was not found in emitted diagnostics:
-expected input to {expected}
-matched against: `{actual}`
+expected input to {self}
+matched against: `{input}`
 full output: `{context}`
-",
-                expected = self,
-                actual = input
+"
```
Al-Kindi-0: nit: here too

Al-Kindi-0 (approved):
Very nice, thank you!

PR #3024, opened 2026-04-20
Add synthetic-transaction regression bench

Al-Kindi-0:
## Purpose:

Adds a new benchmark crate under `benches/synthetic-tx-kernel/` that reproduces the proving-cost brackets of a real Miden transaction from a small JSON snapshot, without pulling in any protocol-level code. The result is a VM-level regression detector that isolates prover changes from workload changes: when you change the VM, the bench moves; when protocol changes its transaction kernel, miden-vm's bench stays put until a new snapshot is explicitly refreshed.

This is a **simplified variant** of the approach originally prototyped in https://github.com/0xMiden/protocol/pull/2391. The key simplification: instead of maintaining a fused cross-repo benchmark that depends on protocol's transaction machinery at runtime, we row counts and emit an equivalent synthetic MASM program on the `miden-vm` side. The two repos stay loosely coupled via a hand-maintained schema, and `miden-vm`'s bench has no runtime dependency on `protocol`.
The snapshot producer lives in a companion protocol PR (`bin/bench-transaction/src/bin/tx-trace-snapshot.rs`).

 ## How it works

Per bench run:
1. Calibrate each MASM snippet's per-iteration cost against the current VM (once).
2. For each snapshot JSON: solve for per-snippet iteration counts via fixed-point refinement, emit a `begin ... end` MASM program, verify it lands in the target padded-trace brackets (`padded_core_side` and `padded_chiplets`), then run `execute` + `execute_and_prove` Criterion groups.

A bracket miss fails the bench; sub-bracket drift is reported but tolerated because proving cost is driven by padded trace length, not raw row count.

## Validations

 Wall-clock prove times against protocol's real `bench-transaction` prove benches:

 | scenario              | protocol (real) | synthetic (miden-vm) | delta  | % diff |
  | --------------------- | --------------: | -------------------: | -----: | -----: |
  | `consume-single-p2id` |          970 ms |               998 ms | +28 ms |  +2.9% |
  | `consume-two-p2id`    |         1886 ms |              1888 ms |  +2 ms |  +0.1% |


huitseeker (approved):
Thanks a lot, this looks great! I added a commit that improves how the snapshot test catches wrong, missing, or extra committed JSON files.

bobbinth:
This is very cool! Without having looked at the code yet, I'm wondering if we could measure the following aspects:

1. Fastest execution time - without the need to collect any data for trace building (i.e., trying to execute the program as fast as possible).
2. Trace preparation time - i.e., the time it takes to generate `TraceProvingInputs` (so that we could run `prover::prove_from_trace_sync()` as the next step).
3. The full proving time (I believe that's what we measure now).
4. Proof verification time.

Not to hold up this PR - we could also add these as future improvements.

Another thought: I think this functionality is not really tx kernel specific - we could benchmark any external program using this. So, maybe the naming should reflect this - e.g., the name of the crate could be something like `miden-vm-synthetic-bench`. The only place where the kernel-specific naming would come in is for the snapshot JSON files. And I imagine in the future we'd add other snapshots that are not tx kernel-related (e.g., for block/batch kernels).

Al-Kindi-0:
> This is very cool! Without having looked at the code yet, I'm wondering if we could measure the following aspects:
> 
>     1. Fastest execution time - without the need to collect any data for trace building (i.e., trying to execute the program as fast as possible).
> 
>     2. Trace preparation time - i.e., the time it takes to generate `TraceProvingInputs` (so that we could run `prover::prove_from_trace_sync()` as the next step).
> 
>     3. The full proving time (I believe that's what we measure now).
> 
>     4. Proof verification time.
> 
> 
> Not to hold up this PR - we could also add these as future improvements.
> 
> Another thought: I think this functionality is not really tx kernel specific - we could benchmark any external program using this. So, maybe the naming should reflect this - e.g., the name of the crate could be something like `miden-vm-synthetic-bench`. The only place where the kernel-specific naming would come in is for the snapshot JSON files. And I imagine in the future we'd add other snapshots that are not tx kernel-related (e.g., for block/batch kernels).

Agreed!

For your first point, `execute` and `execute_and_prove` are now expand to four axes covering the prover/verifier lifecycle:


| Scenario | exec | trace_prep | prove | verify |
|---|---:|---:|---:|---:|
| `consume single P2ID note` | 3.36 ms | 3.82 ms | 1.02 s | 1.36 ms |
| `consume two P2ID notes` | 4.09 ms | 4.68 ms | 1.95 s | 1.46 ms |
| `create single P2ID note` | 3.21 ms | 3.66 ms | 977 ms | 1.41 ms |

For your second point, the crate was renamed to `miden-vm-synthetic-bench`, and also generalized the loader to anticipate the multi-producer future you mentioned, namely, each `snapshots/*.json` is now treated as a producer file, with scenarios as top-level keys. 
Today's `bench-tx.json` plugs in as one producer; a future `bench-block.json` for example. drops in alongside without consumer-side changes. 
Bench-group naming is `<producer-stem>/<scenario-slug>/<axis>`, e.g. `bench-tx/consume-single-p2id-note/prove`.

Al-Kindi-0:
Updated table after the rebase

| Scenario | real prove | synthetic prove | Δ | % |
  |---|---:|---:|---:|---:|
  | consume single P2ID note | 974 ms | 989 ms | +15 ms | +1.5% |
  | consume two P2ID notes | 1914 ms | 1862 ms | -52 ms | -2.7% |
  | consume CLAIM L1 to Miden | 1883 ms | 1859 ms | -24 ms | -1.3% |
  | consume CLAIM L2 to Miden | 1938 ms | 1846 ms | -92 ms | -4.7% |
  | consume B2AGG bridge-out | 8133 ms | 8264 ms | +131 ms | +1.6% |

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 
 - [BREAKING] Refactored MAST forest serialization around fixed-layout full, stripped, and hashless sections, and bumped the MAST wire format to `0.0.3` ([#2765](https://github.com/0xMiden/miden-vm/pull/2765)).
 - Documented sortedness precondition more prominently for sorted array operations ([#2832](https://github.com/0xMiden/miden-vm/pull/2832)).
+- Added a synthetic transaction-kernel Criterion benchmark for VM-level proving regression detection ([#3024](https://github.com/0xMiden/miden-vm/pull/3024)).
```
bobbinth: nit: I'd probably put this into "Features" section (we don't have one yet, but no reason not to create it).

Also, not related to this PR, but we have "Fixes" and "Bug fixes" sections here - these should probably be merged.

benches/synthetic-bench/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+//! Synthetic benchmark generator for VM-level transaction-load regression tests.
+//!
+//! Given a snapshot captured from a representative transaction in `protocol`, this crate calibrates
+//! a small catalog of MASM snippets against the current VM, solves for iteration counts, emits a
+//! synthetic program, and verifies that the program lands in the target core/chiplets padded
+//! brackets.
```
bobbinth: nit (feel free to ignore): this description is a bit "transaction-specific" - probably fine to keep as is because we don't have any other benchmarks for now.

benches/synthetic-bench/README.md
bobbinth, line 1: nit: I'd add a "LICENSE" section at the end of this file (like we have in other READMEs).

benches/synthetic-bench/README.md
```diff
@@ -0,0 +1,220 @@
...
+3. Run `cargo bench -p miden-vm-synthetic-bench` and verify
+   `=> BRACKET MATCH` for every scenario in the printed verifier
+   tables.
+
+If a kernel change moves a scenario into a different padded bucket,
+the `committed_snapshots_load` test in `src/snapshot.rs` fails with
+the producer/scenario pair and the new bracket -- update
+`COMMITTED_SCENARIO_EXPECTATIONS` accordingly.
+
+## Running
```
bobbinth, line 204: May be good to clarify here (or somewhere else in this file) whether we run the proving/verification benchmarks with Poseidon2 or Blake3 as the hash function.

bobbinth (approved):
Looks great! Thank you! This is a very superficial review from my end (I mostly reviewed just the structure rather than the code itself). I left a few optional nits inline.

> | Scenario | exec | trace_prep | prove | verify |
> |---|---:|---:|---:|---:|
> | `consume single P2ID note` | 3.36 ms | 3.82 ms | 1.02 s | 1.36 ms |
> | `consume two P2ID notes` | 4.09 ms | 4.68 ms | 1.95 s | 1.46 ms |
> | `create single P2ID note` | 3.21 ms | 3.66 ms | 977 ms | 1.41 ms |

I'm a little surprised that `exec` and `trace_prep` steps are that close to each other. I would have expected something like a 10x difference between them.

benches/synthetic-bench/src/lib.rs
Al-Kindi-0: Updated to be more generic and match the producer side

benches/synthetic-bench/README.md
Al-Kindi-0: done

benches/synthetic-bench/README.md
Al-Kindi-0: Done. We are using the default, i.e., `Blake3_256`, but I wonder if we should pick the more realistic Poseidon2

bobbinth: Yeah, I'd probably go with Poseidon2 - but could be done later too.

CHANGELOG.md
Al-Kindi-0: Done for the first point and created #3075 for the second

Al-Kindi-0:
> I'm a little surprised that exec and trace_prep steps are that close to each other. I would have expected something like a 10x difference between them.

Indeed, trace prep might be a bit imprecise. The main difference currently between `exec` and `trace_prep` is the recording work. If we want the building of the execution trace to be included in `trace_prep` then the easiest path would be to make `prove_execution_trace` public upstream.

bobbinth:
> The main difference currently between `exec` and `trace_prep` is the recording work. If we want the building of the execution trace to be included in `trace_prep` then the easiest path would be to make `prove_execution_trace` public upstream.

Ah - makes sense! For some reason I confused `trace_prep` with trace building - but you are right, the recording work shouldn't result in too much overhead.

benches/synthetic-bench/README.md
Al-Kindi-0: Done!
Note that we need to match this on the `protocol` side.

Al-Kindi-0:
Here are the latest numbers with `Poseidon2`

| Scenario | Blake3 prove | Poseidon2 prove | × slower |
  |---|---:|---:|---:|
  | consume single P2ID note | 944 ms | 1.71 s | 1.81× |
  | consume two P2ID notes | 1.82 s | 3.33 s | 1.83× |
  | consume CLAIM L1 to Miden | 1.79 s | 3.21 s | 1.79× |
  | consume CLAIM L2 to Miden | 1.85 s | 3.29 s | 1.78× |
  | consume B2AGG bridge-out | 8.29 s | 14.1 s | 1.70× |

CHANGELOG.md
Al-Kindi-0: Merged the Fixes and Bug fixes as it was localized to only one spot

PR #3027, opened 2026-04-20
Single-accumulator LogUp: fraction vs running-sum columns

adr1anh:
## Summary

- Split aux columns into two roles: **running-sum columns** that accumulate cross-row partial sums (with boundary constraints and committed finals), and **fraction columns** that store per-row `V/U` values directly (`D_i · aux_next_i - N_i = 0`).
- Drops aux trace from 8 to 7 columns and committed finals from 8 to 2, shrinking the Fiat-Shamir transcript and the MASM verifier's `observe_aux_trace` from 3 Poseidon permutations to 1.
- Adds `running_sum_columns()` / `fraction_columns_for()` to the `LookupAir` trait so the accumulator and constraint builder know which columns play which role.

## Test plan

- [ ] `cargo test -p miden-air --lib` (degree audit, symbolic validation, accumulator slow/fast match, encoding equivalence)
- [ ] `cargo test -p miden-processor` (lookup harness, fraction-building, constraint-path oracle match)
- [ ] `cargo test -p miden-air --test bus_degree_inventory` (degree budget check)
- [ ] `make check-constraints` (circuit artifact consistency)
- [ ] Full prove/verify round-trip on a non-trivial program


air/src/constraints/lookup/chiplet_air.rs
```diff
@@ -142,4 +142,15 @@ where
...
+
+    fn running_sum_columns(&self) -> &[usize] {
+        &[0]
+    }
+
+    fn fraction_columns_for(&self, running_sum_col: usize) -> &[usize] {
+        match running_sum_col {
+            0 => &[1, 2],
+            _ => panic!("column {running_sum_col} is not a running-sum column"),
+        }
```
adr1anh: This is not needed

PR #3029, opened 2026-04-20
feat: add action to upload pre-compiled artifacts

### LINKED ISSUE #179: feat: add artifacts for the `miden-vm`

lima-limon-inc:
Closes https://github.com/0xMiden/midenup/issues/179

This PR adds a step to the VM's CI which, after a release is made, adds pre-compiled artifacts to the corresponding Github release page. Currently entailing:
- `miden-vm` CLI binary
- `miden-core.masp` Package
These are then used by `midenup` in order to speed installs up.

Attribution note: This added step was taken from the compiler's CI and adapted for the miden-vm.

Follows up on https://github.com/0xMiden/miden-client/pull/1936

Implementation notes:

As mentioned by @bitwalker in [this comment](https://github.com/0xMiden/miden-vm/pull/3029#discussion_r3119649764), the `.masp` package is being built on the `miden-core-lib` crate's build script. However, `CARGO_TARGET_DIR` is not exposed to build scripts and writing files outside of `OUT_DIR` is [explicitly discouraged by the docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).
There is an ongoing (and even active) conversation on the rust-lang repository regarding saving files outside the `OUT_DIR` which were generated on a build_script (like https://github.com/rust-lang/cargo/issues/2729, https://github.com/rust-lang/cargo/issues/13663, https://github.com/rust-lang/cargo/issues/9661, amongst others). However, since it is currently actively discouraged, I went with the cargo script instead.
As a bonus, `cargo script` stabilization, despite being `nightly` only currently seems to be approaching stabilization soon-ish (see https://github.com/rust-lang/cargo/pull/16569).


.github/workflows/workspace-publish.yml
bitwalker, line 1: There are actually two artifacts that we want to produce from this workflow, as part of publishing a release:

1. The `miden-vm` executable, which this PR does currently
2. The `core.masp` package file, produced by the build script of the `miden-core-lib` crate. We may need a dedicated Makefile task for this though, since there isn't an easy way to locate the build script output directory without capturing the verbose output of `cargo build` in JSON form. We could either modify the build script to output the `.masp` to both `OUT_DIR` and `CARGO_TARGET_DIR/core.masp` (so that we have a predictable location to obtain it from), or write a `cargo` script that links against `miden-core-lib` and writes the embedded package bytes to disk somewhere. Either way, we'll need that artifact as well.

bitwalker (changes requested):

.github/workflows/workspace-publish.yml
lima-limon-inc: Noted! I've both rebased and added `miden-core.masp` with the latest commits. I've also updated the description with the rationale.

scripts/generate-package.rs
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env carg
```
huitseeker: This shebang is missing the `o` in `cargo`. 

`make packages` still works because it calls `cargo +nightly -Zscript`, but running the script directly fails.

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,78 @@ jobs:
...
+            target: x86_64-unknown-linux-gnu
+          - os: ubuntu-latest
+            target: aarch64-apple-darwin
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.release.tag_name }}
```
huitseeker: Could this job be gated on `publish` and verify that the release tag is the same revision being published? 

GitHub starts jobs in parallel by default, and this job builds from `${{ github.event.release.tag_name }}` while `publish` checks out `main`. If publishing fails, or if the tag points at a different commit, the release page can still get assets that `midenup` will treat as matching the crate release.

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,78 @@ jobs:
...
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: miden-vm-${{ matrix.target }}
+      - name: Upload
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          gh release upload ${RELEASE_TAG} miden-vm-${{ matrix.target }}
```
huitseeker: `actionlint` flags this because `RELEASE_TAG` is unquoted. This should usually be a simple semver tag, but quoting it avoids word splitting and the same fix applies to the `miden-core.masp` upload below.

huitseeker (changes requested):
The release job needs to prove the tag is a real stable release from main, because otherwise it can publish binaries that midenup treats as trusted release assets. The guard has to run before publish or upload because release.published accepts any published release.

This requires an explicit guard before upload/publish, for example checking `github.event.release.prerelease == false`, tag name matches `^v?[0-9]+\.[0-9]+\.[0-9]+$`, and/or `git rev-parse "$RELEASE_TAG^{commit}" == git rev-parse origin/main`.

.github/workflows/workspace-publish.yml
lima-limon-inc: Thanks for the `actionlint` tip!
I've addressed it on https://github.com/0xMiden/miden-vm/pull/3029/commits/cd28d9584eef6e3ae27c4f12b5e7780afce89526

scripts/generate-package.rs
lima-limon-inc: Addressed here: https://github.com/0xMiden/miden-vm/pull/3029/commits/5422141d0cd95527c471a20af8cbeb3299cd039b

.github/workflows/workspace-publish.yml
lima-limon-inc: In retrospect, shouldn't `publish` checkout the release tag name as well? When a release is created with a specific tag, we want *that* tag to be bundled as a release, right? Why would we checkout `main`?

Checking `main`out introduces two potential problems if I'm not mistaken:
- Like you mentioned, the `main` branch and the commit corresponding to the released tag might differ
- Doesn't this introduce a sort of race condition? What if there's a commit on the main branch right before the publish action run? That commit would be `cargo publish`ed; yet it was not part of the release.

If I'm not mistaken, the `miden-client` check the tag out instead: https://github.com/0xMiden/miden-client/blob/e663d0d88a9adf04d36b1f72060e0599fa265369/.github/workflows/publish-crates-release.yml#L20-L25 (this got changed on this PR https://github.com/0xMiden/miden-client/pull/1862)

I committed this approach here: https://github.com/0xMiden/miden-vm/pull/3029/commits/ff461ca0551424df7910afb24e3b933b9587c743; but do let me know if you want it reverted.

bitwalker:
#2881 is about to be merged - when you rebase this PR on `next`, can you also add a commit which produces the `miden-registry` executable as an artifact here as well? We'll want that available to `midenup` as well.

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,79 @@ jobs:
...
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          gh release upload "${RELEASE_TAG}" miden-vm-${{ matrix.target }}
+      # Since the miden-core.masp package is triplet independent, we only build it once.
+      - name: Build miden-core.masp
+        if: matrix.os == 'ubuntu-latest' && matrix.target == 'x86_64-unknown-linux-gnu'
+        run: |
+          make packages
```
huitseeker, line 164: This job uses `make packages`, which runs `cargo +nightly -Zscript`, but this workflow only updates the default toolchain. 

On a clean runner without nightly, this step looks like it would fail before `miden-core.masp` is uploaded. 

Should we add an install of nightly here, like the other workflows that run `cargo +nightly`?

.github/workflows/workspace-publish.yml
huitseeker: Yes, I think you’re right.

The release tag should be the source of truth for both jobs. Checking out `main` can publish a commit that is not the tagged release, especially if `main` moves after the release is created.

So `ff461ca` looks right to me. I’d keep `upload-artifacts` behind `needs: [publish]`, and keep the existing `verify-main-head: "true"` check so the tag still has to match `origin/main` before anything is published.

bitwalker:
This PR should be rebased to `main` now, so we can release it as a patch to the v0.23.0 release

.github/workflows/workspace-publish.yml
lima-limon-inc: > Should we add an install of nightly here, like the other workflows that run `cargo +nightly`?

Noted! I've added the explicit install in this commit: https://github.com/0xMiden/miden-vm/pull/3029/commits/b35c2fee95f37b115c4c084c2951384eb3b1a902

> On a clean runner without nightly, this step looks like it would fail before `miden-core.masp` is uploaded.

On a separate note, if I'm not mistaken, running `cargo +nightly` would install the `nightly` if missing. Having said that, I do think installing it separately does make debugging CI errors easier.

lima-limon-inc:
@bitwalker @huitseeker 

Btw, on https://github.com/0xMiden/miden-vm/pull/3029/changes/ec8912b908e672fc39277cdc8cdc7d7484252ce5 I've renamed `miden-core` to `core` as per this comment on `miden-protocol`: https://github.com/0xMiden/protocol/pull/2859#discussion_r3188713902

It can be reverted if needed.

crates/lib/core/asm/miden-project.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "miden-core"
+name = "core"
```
huitseeker: Could we keep the package name as `miden-core` and choose `core.masp` only when writing the release file? `Package::write_masp_file()` uses this name for the file, so this also changes the serialized package id. Anything resolving the package as `miden-core` would stop matching this artifact.

We can discuss doing it for a major release, but not for a PR to main:
https://crates.io/crates/miden-core

CHANGELOG.md
```diff
@@ -69,6 +69,16 @@
 - [BREAKING] Unified all auxiliary-trace buses under a single declarative LogUp `LookupAir` shared by the verifier, prover aux-trace generator, and recursive ACE circuit; reduced committed boundary values to one per trace ([#2962](https://github.com/0xMiden/miden-vm/pull/2962)).
 - Collapsed the kernel ROM chiplet to one row per digest with a LogUp multiplicity, eliminating duplicate-callsite rows ([#2962](https://github.com/0xMiden/miden-vm/pull/2962)).
 - Speed-up AUX range check trace generation by changing divisors to a flat Vec layout ([#2966](https://github.com/0xMiden/miden-vm/pull/2966)).
+- Removed AIR constraint tagging instrumentation, applied a uniform constraint description style across components, and optimized constraint evaluation ([#2856](https://github.com/0xMiden/miden-vm/pull/2856)).
+
+#### Enhancements
+- Added binary artifact compilation to CI to aid `midenup`'s installation spped ([#3029](https://github.com/0xMiden/miden-vm/pull/3029)).
+
+## 0.22.1
```
huitseeker: This creates a second `## 0.22.1` section while the dated `0.22.1` section is still below. Could this entry move into the existing section, and fix `spped`, so the release notes do not split the same version?

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,83 @@ jobs:
...
+
+  # Upload the pre-compiled artifacts (miden-vm CLI and core.masp) on the
+  # GitHub release page.
+  # Used by midenup to speed up installs.
+  upload-artifacts:
+    needs: [publish]
+    name: upload pre-built miden-vm executable artifacts
+    if: ${{ github.repository_owner == '0xMiden' }}
+    permissions:
+      contents: write
```
huitseeker, line 109: Could this job use `environment: release` too? The `publish` job has that guard, but environments are job-scoped. This job gets release write and OIDC permissions, then uploads artifacts that `midenup` will trust, so it should sit behind the same release boundary.

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,83 @@ jobs:
...
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: miden-vm-${{ matrix.target }}
+      - name: Upload
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          gh release upload "${RELEASE_TAG}" miden-vm-${{ matrix.target }}
```
huitseeker: Can we make this upload safe to rerun, for example with `gh release upload --clobber`? If one matrix leg uploads its binary and a later step fails, a rerun will fail here on the existing asset before it can rebuild the rest.

Makefile
```diff
@@ -237,6 +240,15 @@ check-constraints: ## Check core-lib constraint artifacts for drift
 exec-info: ## Builds an executable with log tree enabled
 	cargo build --profile optimized $(FEATURES_LOG_TREE)
 
+.PHONY: exec-dist
+exec-dist: ## Builds the CLI for $(BUILD_TARGET) with --locked (for release artifact uploads)
+# NOTE: the resulting binary is stores in target/<$(BUILD_TARGET)>/optimized/
+	cargo build --profile optimized $(FEATURES_CONCURRENT_EXEC) --target $(BUILD_TARGET) --locked
```
huitseeker: Should this target build just the CLI package, for example with `-p miden-vm --bin miden-vm`? As written, running from the workspace can build more than the release artifact needs, so an unrelated workspace target can block publishing the CLI binary.

huitseeker (changes requested):

CHANGELOG.md
lima-limon-inc: My bad! Fixed and corrected the typo on this commit: https://github.com/0xMiden/miden-vm/pull/3029/commits/d718cf4311540343ef083a107ea917ac8441609f

.github/workflows/workspace-publish.yml
lima-limon-inc: Added the guard here: https://github.com/0xMiden/miden-vm/pull/3029/commits/9a2ab0d587fd1ce9b43ba078a5a7bde8a6988bb7

.github/workflows/workspace-publish.yml
lima-limon-inc: Added `--clobber` on both jobs here: https://github.com/0xMiden/miden-vm/pull/3029/commits/79f5991a7cf912348ea00d5df01cc08a9cbf37b7

Makefile
lima-limon-inc: > Should this target build just the CLI package, for example with -p miden-vm --bin miden-vm? As written, running from the workspace can build more than the release artifact needs, so an unrelated workspace target can block publishing the CLI binary.

I made the change over on this commit: https://github.com/0xMiden/miden-vm/commit/bcfc90ee8ea893e6dd3be7d28af716b0ab12a4b4.

Just as a small note, I originally decided in favor of using a plain `cargo build` command in order to use a single build command for the project. Since the `Makefile` directives are only executed when run, a directive like this one could become "stale", which would only be caught when used. That being said, at this point I'd say it'd be rare for the crate to change names. 

Just thought I'd mention it.

CHANGELOG.md
```diff
@@ -109,7 +114,11 @@
 - [BREAKING] Updated the Miden crypto stack to `miden-crypto` v0.25, and switched SMT leaf hashing to use Poseidon2 domain separation so masm-side leaf digests match `SmtLeaf::hash()` ([#3095](https://github.com/0xMiden/miden-vm/pull/3095)).
 - [BREAKING] Reject post-last operation-indexed decorators in block assembly and serialized MAST forests; use `after_exit` for decorators that run after a block exits ([#3114](https://github.com/0xMiden/miden-vm/pull/3114)).
 - [BREAKING] Removed `Continuation::AfterExitDecoratorsBasicBlock`. New MAST merges operation-indexed decorators at the post-last-op sentinel index into `after_exit` at build time; execution uses `AfterExitDecorators` only, with legacy forests still supported ([#2633](https://github.com/0xMiden/miden-vm/issues/2633)).
+<<<<<<< fabrizioorsi/miden-vm-artifact
```
huitseeker: Looks like this conflict got committed. Can we resolve this hunk and keep both changelog entries without the `<<<<<<<`, `=======`, and `>>>>>>>` lines?

.github/workflows/workspace-publish.yml
```diff
@@ -33,3 +35,84 @@ jobs:
...
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: miden-vm-${{ matrix.target }}
+      - name: Upload
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          gh release upload --clobber "${RELEASE_TAG}" miden-vm-${{ matrix.target }}
```
huitseeker, line 155: Can we shape this around immutable releases instead of uploading with `--clobber` after `release.published`? The best practice is to make the release assets immutable too: create the release as a draft, build and attach all assets, then publish it. That lets us drop `--clobber`; otherwise release assets either stay replaceable or this upload flow will fail once GitHub immutable releases are enabled.

.github/workflows/workspace-publish.yml
huitseeker: Note : this is okay as long as the release tag is the source of truth (note we make those immutable!), but perhaps we can we make the upload job prove it is building the same tag commit that publish just verified? Having publish expose the verified tag SHA and check out that exact SHA here. That keeps the tag-based release flow while removing any ambiguity between the crates and the uploaded assets.

CHANGELOG.md
lima-limon-inc: My bad, rightfully fixed here: https://github.com/0xMiden/miden-vm/pull/3029/commits/39c9be28cab296dc01f40f2efc69644b9b0e358b

The _one_ time I use github's web interface to resolve a conflict and it gets messed up :(.

crates/lib/core/asm/miden-project.toml
lima-limon-inc: Fair point, changed it back here: https://github.com/0xMiden/miden-vm/pull/3029/commits/ec0f92907eb6127552f0c91f1c871c53fb5650fa

.github/workflows/workspace-publish.yml
lima-limon-inc: > The best practice is to make the release assets immutable too: create the release as a draft, build and attach all assets, then publish it. 

Noted. Just to clarify, creating a draft release is beyond the scope of this CI job, right? Are VM releases currently first created as drafts manually?

lima-limon-inc: Did you have something like this in mind: https://github.com/0xMiden/miden-vm/pull/3029/commits/d2d82107100775125561fe34485a003d2969dbb9?

This commit moves artifact generation onto a separate job entirely that's triggered when a release is [created](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=created#release).

.github/workflows/upload-release-artifacts.yml
```diff
@@ -0,0 +1,91 @@
+name: Upload pre-compiled release artifacts
+
+on:
+  release:
+    types: [created]
```
huitseeker: This trigger does not match the draft-release flow described below. GitHub Actions does not run `release` workflows for the `created` activity on draft releases, so creating the required draft release will not start this uploader. Should this be driven by a different trigger, such as the release workflow dispatching it after creating the draft?

.github/workflows/upload-release-artifacts.yml
```diff
@@ -0,0 +1,91 @@
...
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
+        with:
+          subject-path: miden-vm-${{ matrix.target }}
+      - name: Upload
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          gh release upload "${RELEASE_TAG}" miden-vm-${{ matrix.target }}
```
huitseeker: This lost the earlier `--clobber` fix. If one matrix leg uploads this asset and a later step fails, rerunning the workflow will fail on the already-uploaded asset instead of replacing it. Should both upload commands keep `--clobber`?

lima-limon-inc: My mistake. I somehow thought that doing the release in two steps (draft, then publish) removed the need for overwriting uploaded packages. However, just like you said, if a step fails we'll need to re-upload them.

Re-added `--clobber` here: https://github.com/0xMiden/miden-vm/pull/3029/commits/54b1c521233e06a6857eddc02aa203f9ac6ce58e

.github/workflows/upload-release-artifacts.yml
lima-limon-inc: I've been looking into this change and it seems to be quite the rabbithole.

Just like you pointed out, and as per [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release), the `created` type does *not* run the corresponding workflows.

The `release` type you mentioned, [according to the docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release), only runs when "A release was published, or a pre-release was changed to a release". So a a draft would simply be ignored.

There also exists a `published` type that only runs when "A release, pre-release, or draft of a release was published". So this release type is *also* ignores drafts. I went ahead and tested it on this toy project:
- https://github.com/lima-limon-inc/fabriz.io/releases/tag/0.4: This tag was marked as release and the workflow did run, as seen on the uploaded artifacts
- `*1`: I then created a new release as a draft. This time, no workflow was run.

So, it would appear as if there's no way to automatically run a workflow on a created draft. This pain point has already been raised on this discussion: https://github.com/orgs/community/discussions/7118. No official "fix" has been provided yet.

I think the closest thing we have to automatic upload of artifacts for drafts boils down to the `workflow_dispatch` "hack" mentioned in said discussion. I've pushed what it would look like on this commit: 15a09b1b97f1ab14d6a6116d5a1c9e7c433fffde

I've also tried it on that same project and it does indeed work `*2`
The one caveat being that now one must go to `Actions->Stage pre-compiled release artifacts (manual)` instead of the usual `Releases->Create new release workflow`.

Would love to hear your thoughts.

Here are screenshots displaying `*1` and `*2`

.github/workflows/upload-release-artifacts.yml
```diff
@@ -0,0 +1,136 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          RELEASE_TAG: ${{ inputs.tag }}
+          RELEASE_SHA: ${{ inputs.sha }}
+        run: |
+          set -e
+          is_draft=$( { gh release view "${RELEASE_TAG}" --json isDraft | jq -r .isDraft; } 2>&1 )
+          echo "Draft existance query: ${is_draft}"
+          case "${is_draft}" in
+            true)
```
huitseeker: When a draft already exists, this reuses it without checking that the draft points at `RELEASE_SHA`. A rerun with the wrong SHA could upload binaries from one commit onto a draft release for another commit.

Could this validate the draft release target before continuing?

.github/workflows/upload-release-artifacts.yml
```diff
@@ -0,0 +1,136 @@
...
+              ;;
+            false)
+              echo "Release ${RELEASE_TAG} already exists and is published. Refusing to touch it." >&2
+              exit 1
+              ;;
+            "release not found")
+              gh release create "${RELEASE_TAG}" \
+                --draft \
+                --target "${RELEASE_SHA}" \
+                --title "v${RELEASE_TAG}"
```
huitseeker: The input description already asks for a tag like `v0.12.0`, so this title becomes `vv0.12.0`. Using `${RELEASE_TAG}` directly would avoid that.

lima-limon-inc: Nice catch, removed here: https://github.com/0xMiden/miden-vm/pull/3029/commits/0a767b0c7154fef0cdb60f845bf6039a30e7a84f

.github/workflows/upload-release-artifacts.yml
lima-limon-inc: Great point. I've added a validation step here: https://github.com/0xMiden/miden-vm/pull/3029/commits/006ec168bab821b31861c8d0b3f0cd4e91764bb7

.github/workflows/upload-release-artifacts.yml
```diff
@@ -0,0 +1,142 @@
...
+            target: x86_64-unknown-linux-gnu
+          - os: ubuntu-latest
+            target: aarch64-apple-darwin
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.sha }}
```
huitseeker: Could we verify the commit after checkout before building? `inputs.sha` is a free text field, and `actions/checkout` will also accept a branch or tag.

Right now a draft can get assets from that resolved ref, while the later publish job is the only place that checks the release commit against `origin/main`. If the release is published and that later check fails, the assets may already be public for `midenup`.

A small guard after checkout, like fetching `origin/main` and comparing it to `git rev-parse HEAD`, would keep the artifact path under the same rule as crate publishing.

CHANGELOG.md
```diff
@@ -95,6 +95,11 @@
 - [BREAKING] Unified all auxiliary-trace buses under a single declarative LogUp `LookupAir` shared by the verifier, prover aux-trace generator, and recursive ACE circuit; reduced committed boundary values to one per trace ([#2962](https://github.com/0xMiden/miden-vm/pull/2962)).
 - Collapsed the kernel ROM chiplet to one row per digest with a LogUp multiplicity, eliminating duplicate-callsite rows ([#2962](https://github.com/0xMiden/miden-vm/pull/2962)).
 - Speed-up AUX range check trace generation by changing divisors to a flat Vec layout ([#2966](https://github.com/0xMiden/miden-vm/pull/2966)).
+- Removed AIR constraint tagging instrumentation, applied a uniform constraint description style across components, and optimized constraint evaluation ([#2856](https://github.com/0xMiden/miden-vm/pull/2856)).
+
+#### Enhancements
```
huitseeker: Looks like this hunk re-adds entries that are already in this release section: `#2856` is above, and `#2901` is already listed earlier too.

Could we drop these duplicated lines and keep only the new `#3029` artifact entry? That avoids a second `#### Enhancements` block inside the same release notes.

huitseeker (commented):
I've opened https://github.com/lambdaclass/miden-vm/pull/3 to offer an alternate way of doing this keeping the release process draft-first: no public GitHub release exists until the trusted assets are attached and crates have been published. That matches the way midenup will consume these assets as release-trusted outputs.

The main design choice is to keep `.github/workflows/workspace-publish.yml` as the single release workflow because crates.io trusted publishing is tied to that workflow identity. Instead of adding a separate artifact workflow with a free-form sha, this derives the release commit from origin/main, builds assets and crates from that same commit, then publishes the draft release.

We keep `release.published` only as a guarded fallback: it refuses to build assets after a release is already public, and fails with recovery guidance if required assets are missing.

CHANGELOG.md
lima-limon-inc: This has since been removed in https://github.com/0xMiden/miden-vm/commit/41bceaa93b4c9c82325ed20f7483a2f45da3a9f0

.github/workflows/upload-release-artifacts.yml
lima-limon-inc: This has since been removed in https://github.com/0xMiden/miden-vm/commit/41bceaa93b4c9c82325ed20f7483a2f45da3a9f0

lima-limon-inc:
> I've opened [lambdaclass#3](https://github.com/lambdaclass/miden-vm/pull/3) to offer an alternate way of doing this keeping the release process draft-first: no public GitHub release exists until the trusted assets are attached and crates have been published. That matches the way midenup will consume these assets as release-trusted outputs.

Thank you very much for the PR! It's been merged on https://github.com/0xMiden/miden-vm/commit/41bceaa93b4c9c82325ed20f7483a2f45da3a9f0

I've also moved the `prepare-release` logic onto a separate shell script as per https://github.com/lambdaclass/miden-vm/pull/3#discussion_r3277558728.

bitwalker (approved):

huitseeker (approved):
LGTM!

huitseeker:
@lima-limon-inc I'm not allowed to push to the lambdaclass branch, so beware that I can only merge this when it's up-to-date with the latest next.

lima-limon-inc:
> @lima-limon-inc I'm not allowed to push to the lambdaclass branch, so beware that I can only merge this when it's up-to-date with the latest next.

Thanks for the reminder! Changed here: https://github.com/0xMiden/miden-vm/pull/3029/commits/1a643805a64705d5a287d1f2d15fe3443737161c

bitwalker:
@lima-limon-inc Unfortunately, you'll have to merge again - I'm not sure why it's not letting us merge here in the GitHub UI for you. Sorry about that, I didn't manage to catch it in time after your last update

lima-limon-inc:
> @lima-limon-inc Unfortunately, you'll have to merge again - I'm not sure why it's not letting us merge here in the GitHub UI for you. Sorry about that, I didn't manage to catch it in time after your last update

No worries! I re-merged `next` onto my branch. Let me know if it works correctly now 😊 

If needed I can also rebase onto `next` directly

bitwalker:
@lima-limon-inc Incredibly, somehow, I still wasn't able to get it in time - something else got merged again 😩

PR #3038, opened 2026-04-21
ci: stop link checker config errors from opening issues

huitseeker:
lychee-action v2.8.0 runs lychee v0.23, where --exclude-mail is no longer accepted. On scheduled runs we set fail: false, so the action captured that CLI error as a non-zero exit_code, wrote only a stub report, and the next step opened a Link Checker Report issue even though no links had been checked (see #3017 #3020 #3030).

- Remove the obsolete flag because mail links are opt-in now, and require a real lychee markdown summary before opening an issue. 
- If lychee fails *before* producing a report, fail the workflow so configuration errors stay in CI logs, instead of becoming low-signal documentation issues.

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -40,7 +40,6 @@ jobs:
           fail: ${{ steps.flags.outputs.fail_fast }}
           args: |
             --verbose
-            --exclude-mail
             --exclude "localhost"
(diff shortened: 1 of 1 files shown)


bitwalker (approved):

bobbinth (approved):
Looks good! Thank you!

PR #3039, opened 2026-04-21
chore: port miden crypto stack to 0.24

huitseeker:
Update the VM to the 0.24 Miden crypto/STARK crates. This pulls in miden-crypto 0.24 and miden-lifted-stark 0.24, and temporarily patches midenc-hir-type to the PR branch which updates its miden-serde-utils dependency. ~~*The patch is intentionally a proof-of-concept dependency bridge until midenc-hir-type is released with the compatible dependency set.*~~
~~**Please unblock **: https://github.com/0xMiden/midenc-hir-type/pull/7~~

Adapt the Rust API changes from the new stack. Field construction now uses the checked/unchecked constructors required by `miden-field` 0.24, DSA tests and helpers use `SigningKey` in place of the removed `SecretKey export`, MMR `append` sites handle the new `Result` return, and removed crypto re-exports are replaced where needed. `Word` itself gained lexicographic ordering, so use it for that purpose.

Port proof handling to the lifted STARK API. Proof serialization now stores `StarkProof` directly, `AirWitness` is re-exported from `miden-lifted-stark`, verifier/prover code consumes the new `AirInstance` and instance-shape data, and recursive verification derives the trace height from the proof instance instead of carrying the old log-trace-height side channel.

Update the recursive verifier transcript and FRI query flow for the 0.24 protocol. The random coin no longer absorbs the trace height before public inputs, absorbs the new trace-height point at the protocol position, keeps query indices in natural domain order, and converts only at the MASM `fri_ext2fold4` boundary where the native opcode still expects the old row-slot convention.

Refresh generated constraints and protocol constants after the dependency upgrade. This includes the Poseidon2 M4 matrix, relation digest, AIR shape snapshot, regenerated `constraints_eval.masm`, and the changed processor/assembly snapshots caused by the new hash and proof encodings.


air/src/snapshots/miden_air__config__tests__relation_digest_matches_current_air.snap
```diff
@@ -2,6 +2,6 @@
 source: air/src/config.rs
 expression: snapshot
 ---
-num_inputs: 572
-num_eval_gates: 5588
-relation_digest: [7682526984894530630, 13019009422810970716, 16037978224454945133, 1332031755710493241]
+num_inputs: 568
+num_eval_gates: 5540
```
adr1anh, line 6: Interesting... but yay!

crates/lib/core/asm/pcs/fri/frie2f4.masm
```diff
@@ -167,6 +169,26 @@ pub proc verify_query_layer
     mem_loadw_le
     # => [v0, ..., v7, f_pos, p, poe, e0, e1, a0, a1, layer_ptr, rem_ptr, ...]
 
+    # the FRI proof is queried in domain order, but fri_ext2fold4 expects an index whose low
+    # two bits select the bit-reversed position inside the four-element folded row. Build that
```
Al-Kindi-0, line 173: This is fine for now

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -639,20 +650,17 @@ pub proc generate_list_indices
...
-        # so bit_reverse_len_parallel consumes [natural_index, pow2_shift] and leaves
-        # [rev_index, depth, natural_index, idx, query_ptr, ...].
-        dup.5 dup.7 dup.2
-        #=> [natural_index, pow2_shift, depth, natural_index, idx, query_ptr, num_queries, mask, depth, pow2_shift, ...]
-        exec.utils::bit_reverse_len_parallel
-        #=> [rev_index, depth, natural_index, idx, query_ptr, num_queries, mask, depth, pow2_shift, ...]
+        # --- Prepare domain-order query word ---
+        # 0.24 LMCS trees are indexed by domain order, so the Merkle index and the DEEP
+        # evaluation index are both the sampled natural index.
+        dup.5 swap dup movdn.2
```
Al-Kindi-0, line 656: Also fine for now

crates/lib/core/tests/stark/verifier_recursive/mod.rs
```diff
@@ -277,33 +276,24 @@ where
 fn build_merkle_data(
     config: &P2Config,
     stark: &StarkTranscript<Challenge, P2Lmcs>,
-    log_trace_height: usize,
+    _log_trace_height: usize,
```
Al-Kindi-0, line 279: nit: could remove probably now

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -319,31 +318,22 @@ mod recursive_verifier {
     fn build_merkle_data(
         config: &P2Config,
         stark: &StarkTranscript<Challenge, P2Lmcs>,
-        log_trace_height: usize,
+        _log_trace_height: usize,
```
Al-Kindi-0, line 321: nit: same here

Al-Kindi-0 (approved):
LGTM

Some degradation in perf in the recursive verifier due to the change to natural order but should be fixed in a separate PR

crates/lib/core/asm/pcs/fri/frie2f4.masm
adr1anh: Let's make an issue for the Fri fold instruction

adr1anh (approved):
🫡

crates/lib/core/asm/pcs/fri/frie2f4.masm
huitseeker: #3041

PR #3044, opened 2026-04-22
fix(core): keep MAST merge from growing stripped forests

### LINKED ISSUE #3035: `MastForest::compact` increases the MAST size
### What

Calling the `MastForest::compact` on the MASM with stripped debug info (after the `MastForest::clear_debug_info` call) increases the serialized MAST size (using `.to_bytes()`).

### How to reproduce

In the compiler repo in the default (next) branch:
1. Run `cargo make test basic_wallet_and_p2id` test which compiles the `examples/basic-wallet` account code.
2. Pick the compiled package at `target/miden/release/basic_wallet.masp`.
3. Check the size after the `.clear_debug_info()` and `.compact()` calls:
```rust
    let mut note_mast = package.mast.mast_forest().as_ref().clone();
    note_mast.clear_debug_info();
    let stripped_size = note_mast.to_bytes().len();
    let (compacted, _) = note_mast.compact();
    let compacted_size = compacted.to_bytes().len();
```

/cc @bitwalker

huitseeker:
This fixes a case where `MastForest::compact()` could make a stripped forest larger instead of smaller.

The root cause is twofold. First, merge rebuilt basic blocks from the padded operation iterator, so padding ops became "organic" serialized ops. Second, after debug info was cleared, compact still wrote empty node decorator rows for every merged node.

The fix keeps the original basic-block batches during merge and skips empty node decorator rows. The op decorator CSR is still kept as a valid empty structure after debug info is cleared.

Closes #3035.


bitwalker (approved):
LGTM! Ping me for a re-review after this is marked ready (if needed)

greenhat (commented):
LGTM! Thank you for a prompt response!

bitwalker:
@huitseeker Maybe we should apply this to `main` so it can go out in 0.22.2? It looks like it should be a non-breaking change unless I missed something

huitseeker:
@bitwalker Is this a blocker for something? The improvements in serialization size from this fix are in the low single-digit percentages. I agree there is no semver issue, but I am not sure if it's worth the trouble.

bitwalker:
We were observing double-digit reduction in size due to compaction previously on some of our example programs (e.g. 18k -> 13k in one of them), but obviously that can vary significantly. In any case, without compaction, it's a lot easier to hit the account/note code size limits when compiling Rust programs - so it's not a blocker for some feature, but it does directly impact whether Rust components that run up against the limit will be usable or not. Since users won't get stuff in `next` for quite some time (at least a month?), it's probably worth it.

Since we also have at least one other fix queued up in `main`, it probably makes sense to include this if it isn't going to be a major pain to do so. I was thinking the divergence between `main` and `next` in this code wouldn't be too bad since we just did a release, but you'd have a better idea than me whether this would be annoying to backport.

bobbinth:
> We were observing double-digit reduction in size due to compaction previously on some of our example programs (e.g. 18k -> 13k in one of them), but obviously that can vary significantly. In any case, without compaction, it's a lot easier to hit the account/note code size limits when compiling Rust programs - so it's not a blocker for some feature, but it does directly impact whether Rust components that run up against the limit will be usable or not. Since users won't get stuff in `next` for quite some time (at least a month?), it's probably worth it.

I thought most of the reduction in size was due to removing decorators (not compaction) - but if compaction indeed can help reduce the size by something like 30% - I think it should go into `main`. If it is more like around 5% impact, I'd keep it going into `next`.

huitseeker:
You can run the original ignored test from the first PR commit, a3b5f4af0, against the compiler-derived wallet package. It fails exactly as expected:
```
  stripped=29921, compacted=30701
```
So the pre-fix behavior grows the serialized MastForest by 780 bytes after `clear_debug_info().compact()`.

I then measured the same .masp fixture against the current fixed branch:

```
  stripped_size=29921
  compacted_size=29757
  growth_after_compact=-164
```

So the fix avoids 944 bytes versus the buggy compacted output: it removes the +780 byte regression and also
lets compaction produce the intended 164 byte reduction.

Note: this bug fix is an edge case, it is not the case that the bug is “compact makes MAST bigger.” The actual MAST structure still compacts in both versions: 473 nodes down to 470, and write_stripped goes from 29423 to 29269 bytes in both versions. The growth was in serialized debug-info side tables after `clear_debug_info()`: during `compact()`/self-merge, rebuilt nodes had empty `before_enter` and `after_exit decorator` slices, and the old code still registered those empty node-decorator entries. The fix skips registering node decorators when both slices are empty, keeping that sparse storage sparse.

I believe this is an edge-case fix, with a minor performance impact (3%), and should hence go to next.

bobbinth (approved):
Looks good! Thank you!

PR #3047, opened 2026-04-23
test: Speed up repeated assembler tests

huitseeker:
This speeds up repeated assembler tests by caching successful `Test::compile()` results inside the test process. Tests that run the same program with many stack inputs can now reuse the compiled program instead of rebuilding it each time.

The coverage stays the same. Each run still executes the traced path, checks step/resume execution, and builds the trace before checking the final stack.

The word tests now build each test program once and pass different stack inputs through the same full execution path.

This also fixes the miden-core-lib build script watch path so changes under crates/assembly/src correctly trigger rebuilds.


bitwalker (commented):
Do you have a before/after on the difference this makes measured in user time? This looks fine to me, but if the actual difference isn't significant, it probably isn't worth the complexity. I'm presuming it probably does make a significant difference in proptest-based tests, but I don't actually know what the difference would be.

huitseeker:
On a *warmed*`make -j 8 test` (my box has more than 8 physical cores, but the CI machine does have 8)
  | Ref | Run 1 | Run 2 | Avg |
  |---|---:|---:|---:|
  | Before (acfc0ee12) | 41.706s | 41.411s | 41.56s |
  | After (8c20357f0) | 12.350s | 12.741s | 12.55s |

bitwalker (approved):
Nice! Even better than expected. Thanks for capturing those numbers!

PR #3054, opened 2026-04-24
core: fix static linking debug metadata

### LINKED ISSUE #2990: Static linking by digest loses per-alias metadata when multiple roots share the same digest
**Problem**

When a statically linked library contains two procedures with identical operations but different source mappings, they share the same MAST digest. The linker resolves procedure references by digest only, so it always picks the first root — the second procedure silently inherits the first's diagnostics metadata.

This affects debugger accuracy: stepping into the second procedure shows the first procedure's source location.

**Repro**

Two procedures with identical add but compiled from different source contexts (alias_a vs alias_b). After linking the second by digest, its source mapping reads back as alias_a.

**Root cause**

The entire linking chain — from compiled procedure info, through symbol resolution, to the forest builder — only carries a digest (Word). There is no way to specify which exact root to copy when multiple roots match.

**Proposed fix**

Store an optional source node ID in the compiled procedure info, populated during library packaging. When present, use it to copy the exact subtree instead of searching by digest. This touches the AST library types, package format, linker, and assembler.

**Context**

  Identified during review of [#2955](https://github.com/0xMiden/miden-vm/pull/2955#discussion_r3059508022).

djolertrk:
When a statically linked library contains two procedures with identical operations but different source mappings, they share the same MAST digest. The linker resolves procedure references by digest only, so it always picks the first root — the second procedure silently inherits the first's diagnostics metadata. This patch fixes it.

Fixes https://github.com/0xMiden/miden-vm/issues/2990

### DIFF
--- a/crates/assembly-syntax/src/library/module.rs
+++ b/crates/assembly-syntax/src/library/module.rs
@@ -3,6 +3,8 @@ use core::ops::Index;
 
 use midenc_hir_type::FunctionType;
 
+use miden_core::mast::MastNodeId;
+
 use crate::{
     Path, Word,
     ast::{self, AttributeSet, ConstantValue, Ident, ItemIndex, ProcedureName},
@@ -45,9 +47,17 @@ impl ModuleInfo {
         digest: Word,
         signature: Option<Arc<FunctionType>>,
         attributes: AttributeSet,
+        source_root_id: Option<MastNodeId>,
+        source_library_commitment: Option<Word>,
     ) {
-        self.items
-            .push(ItemInfo::Procedure(ProcedureInfo { name, digest, signature, attributes }));
+        self.items.push(ItemInfo::Procedure(ProcedureInfo {
+            name,
+            digest,
+            signature,
+            attributes,
+            source_root_id,
+            source_library_commitment,
+        }));
     }
 
     /// Adds a constant to the module.
@@ -181,6 +191,23 @@ pub struct ProcedureInfo {
     pub digest: Word,
     pub signature: Option<Arc<FunctionType>>,
     pub attributes: AttributeSet,
+    /// The exact procedure root in the source library, if known.
+    ///
+    /// This is needed when multiple exported procedures share the same digest but carry different
+    /// diagnostics metadata.
+    pub source_root_id: Option<MastNodeId>,
+    /// The commitment of the source library forest that `source_root_id` belongs to, if known.
+    pub source_library_commitment: Option<Word>,
+}
+
+impl ProcedureInfo {
+    pub fn source_root_id(&self) -> Option<MastNodeId> {
+        self.source_root_id
+    }
+
+    pub fn source_library_commitment(&self) -> Option<Word> {
+        self.source_library_commitment
+    }
 }
 
 /// Stores the name and value of a constant
--- a/crates/assembly-syntax/src/library/module.rs
+++ b/crates/assembly-syntax/src/library/module.rs
@@ -1,9 +1,8 @@
 use alloc::{sync::Arc, vec::Vec};
 use core::ops::Index;
 
-use midenc_hir_type::FunctionType;
-
 use miden_core::mast::MastNodeId;
+use midenc_hir_type::FunctionType;
 
 use crate::{
     Path, Word,
@@ -42,6 +41,17 @@ impl ModuleInfo {
 
     /// Adds a procedure to the module.
     pub fn add_procedure(
+        &mut self,
+        name: ProcedureName,
+        digest: Word,
+        signature: Option<Arc<FunctionType>>,
+        attributes: AttributeSet,
+    ) {
+        self.add_procedure_with_provenance(name, digest, signature, attributes, None, None);
+    }
+
+    /// Adds a procedure to the module with optional source provenance.
+    pub fn add_procedure_with_provenance(
         &mut self,
         name: ProcedureName,
         digest: Word,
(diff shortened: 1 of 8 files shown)


djolertrk:
cc @huitseeker @bitwalker

crates/assembly/src/assembler.rs
```diff
@@ -1361,6 +1363,8 @@ impl Assembler {
                     kind,
                     target.span(),
                     mast_root.into_inner(),
+                    None,
```
bitwalker, line 1479: We should probably add a note here that the reason we don't have (and do not try to find) any `MastNodeId`/package digest here, is because `SymbolResolution::MastRoot` generally corresponds to a literal digest in the MASM sources, so we don't actually know what source context applies.

In theory we could use heuristics to guess, but I think it is fine for now to omit any context in this case.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -1157,17 +1171,46 @@ impl MastForestBuilder {
...
-    pub fn ensure_external_link(&mut self, mast_root: Word) -> Result<MastNodeId, Report> {
+    /// Adds a node corresponding to the given MAST root, preferring an exact source root when
+    /// provenance from a statically-linked library is available.
+    pub fn ensure_external_link_with_source(
+        &mut self,
+        mast_root: Word,
+        source_library_commitment: Option<Word>,
+        source_root_id: Option<MastNodeId>,
+    ) -> Result<MastNodeId, Report> {
+        if let Some(root_id) =
```
bitwalker, line 1186: I would refactor `find_statically_linked_root` to return `Some(root_id)` if _either_ there is an exact match, or only a match of the MAST root - since there is no difference in how those cases are handled here (and I suspect there would be no difference in any caller).

crates/assembly-syntax/src/library/module.rs
```diff
@@ -177,6 +186,23 @@ pub struct ProcedureInfo {
...
+    ///
+    /// This is needed when multiple exported procedures share the same digest but carry different
+    /// diagnostics metadata.
+    pub source_root_id: Option<MastNodeId>,
+    /// The commitment of the source library forest that `source_root_id` belongs to, if known.
+    pub source_library_commitment: Option<Word>,
+}
+
+impl ProcedureInfo {
+    pub fn source_root_id(&self) -> Option<MastNodeId> {
```
huitseeker, line 219: Could the name-based package path use this exact root too? 

`Package::make_executable` still does `get_procedure_digest_by_name()` and then `find_procedure_root(digest)`, so a library with `alias_a` and `alias_b` as same-digest exports can still make `alias_b` use the first root source mapping when it is turned into an executable. 

Maybe that lookup should return the `ProcedureInfo` (or use `source_root_id`) instead of going back through the digest.

crates/assembly/src/assembler.rs
djolertrk: Yeah I will add a comment here.

crates/assembly/src/mast_forest_builder.rs
djolertrk: Done, thanks for the comment

crates/assembly-syntax/src/library/module.rs
```diff
@@ -41,9 +42,17 @@ impl ModuleInfo {
         digest: Word,
         signature: Option<Arc<FunctionType>>,
         attributes: AttributeSet,
+        source_root_id: Option<MastNodeId>,
```
huitseeker, line 60: Could we keep the old `add_procedure` call shape and add a new helper for provenance? This method is public in `miden-assembly-syntax`, so adding required args is a breaking API change for downstream callers. Callers that do not have provenance could still get `None` inside the old method.

huitseeker (changes requested):
This targets main, and includes a semver-breaking change, see inline.

crates/assembly-syntax/src/library/module.rs
djolertrk: Sure, good point

djolertrk:
> This targets main, and includes a semver-breaking change, see inline.

Yeah.. I addressed that, and I am targeting `next` - no need to target `main`. Thanks!

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -108,8 +114,14 @@ impl MastForestBuilder {
         static_libraries: impl IntoIterator<Item = &'a MastForest>,
     ) -> Result<Self, Report> {
         // All statically-linked libraries are merged into a single MastForest.
-        let forests = static_libraries.into_iter();
-        let (statically_linked_mast, _remapping) = MastForest::merge(forests).into_diagnostic()?;
+        let forests = static_libraries.into_iter().collect::<Vec<_>>();
+        let statically_linked_forest_indices_by_commitment = forests
+            .iter()
+            .enumerate()
+            .map(|(idx, forest)| (forest.commitment(), idx))
```
huitseeker, line 125: We should note this *isn't* a key that changes when debug metadata changes.

`MastForest::commitment()` only hashes procedure root digests, so two forests with the same roots but different debug info share this key. 

This map could point a procedure from one static library at another library's root map, and the copied subtree can still get the wrong diagnostics metadata.  We should at the very least leave a TODO linking to #3067.

djolertrk: Thanks! Added.

huitseeker (approved):
LGTM with known limitation tracked by #3067 

We could also tackle this here by using a source library index, or using a key that commits to the metadata too — but those sound all clunkier than tackling #3067 systematically — @bitwalker 's opinion might vary.

bitwalker:
I agree that it makes sense to tackle #3067 first, and then close this gap in the linker based on the answers we come up with as part of addressing that issue.

bitwalker (approved):

djolertrk:
@bitwalker @huitseeker can we merge this?

huitseeker:
@djolertrk it just needs a rebase — I'd do it for you but you haven't opened maintainer access on the branch you issued the PR from.

djolertrk:
Done :)

Issue #3040, opened 2026-04-22
Separate debug information from MAST entirely


bitwalker:
Currently, we are forced to record certain types of debug information in MAST:

* Source location information for each operation (via `AssemblyOp` decorator)
* Source MASM instruction corresponding to each operation, important for stepping over MASM instructions that correspond to many operations when lowered to MAST (via `AssemblyOp` decorator)
* Source language variable location info (where the variable is currently stored, what its debug type is) for each operation (via `DebugVarInfo` decorator)

Ultimately, none of that has anything to do with MAST at all, and is entirely an artifact of two issues:

1. Historically, we didn't have any way to map external metadata to individual ops in the MAST (i.e. the notion of an "operation index" that uniquely identified a specific operation in a given MAST forest). That has recently changed.
2. The content-addressable nature of MAST means that, barring extra work to prevent it, duplicate instruction sequences (but with unique source locations, etc.) would be collapsed into a single subtree, and it is ambiguous what we do with decorators in that situation. We "resolved" this by means of fingerprinting MAST nodes based on their decorators, and explicitly _not_ de-duplicating subtrees if the fingerprint differed between otherwise-identical subtrees. This has added a great deal of complexity to MAST, but the alternative was a virtually useless debugging experience.

Now that 1 is no longer a blocker, the only remaining blocker to removing debug info entirely from MAST is 2. To resolve this remaining blocker, we need to devise a method to disambiguate multiple sets of debug info/decorators that apply to a single MAST node or operation, during execution. Once we have such a method, we would no longer require node fingerprinting, nor storage of decorators/debug info in MAST itself, and could instead move those concerns to more appropriate places.

I'm following up on some discussion at EuroLLVM this year regarding this very same problem; I heard mention that there is someone at Google working on a solution to this for DWARF, and they apparently already have a design drawn up. I'd like to see if that might be applicable to our use case or not. If so, that might prove to be a better avenue, since it would allow us to more easily map DWARF to our debug info representation once those changes are implemented and propagate through the LLVM ecosystem (e.g. `rustc`).

Aside from that possibility, our next best bet is to disambiguate based on caller context - in other words, the caller always knows unambiguously which function/procedure it was calling, and this can be used to disambiguate between decorator sets applied to MAST nodes corresponding to procedure roots. This allows us to further disambiguate between decorator sets applied to subtrees of that procedure.

Where that gets murky:

[...]

PR #3055, opened 2026-04-25
Clean up MAST forest wire layout and untrusted validation

### LINKED ISSUE #3043: Simplify MAST forest serialization layout
PR #2765 leaves a few layout cleanups for a follow-up. 

### moving UntrustedMastForest

Move `UntrustedMastForest` out of `core/src/mast/mod.rs` into a focused module, such as `core/src/mast/untrusted.rs`. The type marks the untrusted deserialization boundary, so it should not add more weight to the main MAST module. Keep public re-exports stable. Review comment: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122070445

### internal/external counts in header

Consider storing `internal_node_count` and `external_node_count` in the wire header. Today the reader scans node entries to derive these counts. If the header carries them, the reader can compute node-entry, external-digest, and internal-hash section sizes earlier. This can make malformed input cheaper to reject. Review comment: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122423554

### external nodes first/sorted

Consider making external nodes come first in `MastForest` wire data, sorted lexicographically. With that invariant, digest lookup and section layout get simpler. External and internal digests can become one node-digest section, and hashless serialization can omit the internal suffix. This likely belongs with the builder work in #2473 because it changes forest ordering rules. Review comment: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122440986

### multiple passes/header shape

Reduce repeated passes over serialized bytes. A richer header should let the reader compute the expected core payload size early, then read the rest once for both trusted and untrusted paths. This should be considered with the planned move of debug info out of `MastForest`, since debug info may stay variable-sized or move outside the core forest payload. Review comment: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122640545

huitseeker:
Fixes #3043.

This cleans up the MAST forest wire format on top of #2765.

The wire format now stores clear node counts, keeps node IDs stable when reading and writing, and puts external node digests in a sorted prefix before node entries. Hashless data still keeps external digests, but omits internal digests.

Untrusted reads now keep the wire bytes until `validate()` is called. Validation checks structure, child links, rebuilt hashes, external digest slots, and external digest sort order before returning a trusted `MastForest`.

### DIFF
--- a/core/src/mast/serialization/info.rs
+++ b/core/src/mast/serialization/info.rs
@@ -6,6 +6,7 @@ use crate::mast::node::MastNodeExt;
 use crate::{
     mast::{MastForestContributor, MastNode, MastNodeId, Word, node::MastNodeBuilder},
     serde::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
+    utils::Idx,
 };
 
 // CONSTANTS
@@ -70,28 +71,50 @@ impl MastNodeEntry {
 
     /// Constructs a new [`MastNodeEntry`] from a [`MastNode`].
     pub fn new(mast_node: &MastNode, ops_offset: NodeDataOffset) -> Self {
+        Self::new_inner(mast_node, ops_offset, None)
+    }
+
+    /// Constructs a new [`MastNodeEntry`] from a [`MastNode`], remapping child IDs when needed.
+    #[deprecated(note = "use MastNodeEntry::new(); remapped wire ordering has been removed")]
+    pub fn new_with_id_remap(
+        mast_node: &MastNode,
+        ops_offset: NodeDataOffset,
+        id_remap: Option<&[u32]>,
+    ) -> Self {
+        Self::new_inner(mast_node, ops_offset, id_remap)
+    }
+
+    fn new_inner(
+        mast_node: &MastNode,
+        ops_offset: NodeDataOffset,
+        id_remap: Option<&[u32]>,
+    ) -> Self {
         use MastNode::*;
 
         if !matches!(mast_node, &Block(_)) {
             debug_assert_eq!(ops_offset, 0);
         }
 
+        let remap_id = |id: MastNodeId| -> u32 {
+            id_remap.and_then(|remap| remap.get(id.to_usize()).copied()).unwrap_or(id.0)
+        };
+
         match mast_node {
             Block(_) => Self::Block { ops_offset },
             Join(join_node) => Self::Join {
-                left_child_id: join_node.first().0,
-                right_child_id: join_node.second().0,
+                left_child_id: remap_id(join_node.first()),
+                right_child_id: remap_id(join_node.second()),
             },
             Split(split_node) => Self::Split {
-                if_branch_id: split_node.on_true().0,
-                else_branch_id: split_node.on_false().0,
+                if_branch_id: remap_id(split_node.on_true()),
+                else_branch_id: remap_id(split_node.on_false()),
             },
-            Loop(loop_node) => Self::Loop { body_id: loop_node.body().0 },
+            Loop(loop_node) => Self::Loop { body_id: remap_id(loop_node.body()) },
             Call(call_node) => {
                 if call_node.is_syscall() {
-                    Self::SysCall { callee_id: call_node.callee().0 }
+                    Self::SysCall { callee_id: remap_id(call_node.callee()) }
                 } else {
-                    Self::Call { callee_id: call_node.callee().0 }
+                    Self::Call { callee_id: remap_id(call_node.callee()) }
                 }
             },
             Dyn(dyn_node) => {
--- a/core/src/mast/serialization/info.rs
+++ b/core/src/mast/serialization/info.rs
@@ -6,7 +6,6 @@ use crate::mast::node::MastNodeExt;
 use crate::{
     mast::{MastForestContributor, MastNode, MastNodeId, Word, node::MastNodeBuilder},
     serde::{ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable},
-    utils::Idx,
 };
 
 // CONSTANTS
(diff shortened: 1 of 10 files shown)


core/src/mast/serialization/layout.rs
```diff
@@ -281,68 +266,157 @@ fn scan_layout_sections<R: OffsetTrackingReader>(
...
+            source,
+            internal_node_count,
+            crate::Word::min_serialized_size(),
+            "internal node count",
+        )?;
+    }
+
     let roots_count = source.read_usize()?;
-    validate_budgeted_count(source, roots_count, core::mem::size_of::<u32>(), "root count")?;
+    validate_budgeted_count(source, roots_count, size_of::<u32>(), "root count")?;
```
plafer, line 279: `validate_budgeted_count()` checks each section independently, but wouldn't catch if e.g. the allocations of `external_node_count + internal_node_count` exceeds the maximum, right? Ideally our check would account for the sum of all allocations of these sections, not just each one independently.

core/src/mast/serialization/info.rs
```diff
@@ -246,6 +271,114 @@ impl Deserializable for MastNodeEntry {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) struct MastNodeWireEntry {
+    entry: MastNodeEntry,
+    external_digest_slot: Option<u32>,
+}
```
plafer: Docstring would be helpful here to explain how the 2 fields work together (and perhaps a reminder for why the external digest is stored separately)

plafer (approved):
LGTM!

core/src/mast/serialization/info.rs
huitseeker: Addressed in 344b4128ab7bcfff133790514b626068ee0e8d9b — apologies for rewriting history, this was vastly simpler with the upcoming other PR stacked on top.

core/src/mast/serialization/layout.rs
huitseeker: That's true. I have this fixed in a simplification PR on top of this. Will ping here when open.

huitseeker: This is up in https://github.com/0xMiden/miden-vm/pull/3077

core/src/mast/serialization/mod.rs
```diff
@@ -259,40 +267,67 @@ impl MastForest {
...
 
-        // node count
-        target.write_usize(self.nodes.len());
+        // header counts
+        let node_count = self.nodes.len();
+        let external_node_count = self.nodes.iter().filter(|node| node.is_external()).count();
+        let internal_node_count = node_count - external_node_count;
+        target.write_usize(node_count);
+        target.write_usize(internal_node_count);
+        target.write_usize(external_node_count);
```
bobbinth: Question: why do we need to write all 3 values? Couldn't we write just `internal_node_count` and `external_node_count` and then compute `node_count = internal_node_count + external_node_count`?

core/src/mast/serialization/resolved.rs
```diff
@@ -352,6 +348,50 @@ fn recompute_hash_table(
...
+    right: &crate::Word,
+) -> core::cmp::Ordering {
+    serialized_word_bytes(left).cmp(&serialized_word_bytes(right))
+}
+
+fn serialized_word_bytes(word: &crate::Word) -> Vec<u8> {
+    let mut bytes = Vec::with_capacity(crate::Word::min_serialized_size());
+    word.write_into(&mut bytes);
+    bytes
+}
```
bobbinth: Why do we use this custom comparison rather than rely on built-in `Word` `Ord` implementation?

core/src/mast/serialization/mod.rs
```diff
@@ -259,40 +267,67 @@ impl MastForest {
...
+            },
+        );
+
+        let mut external_digest_slot_by_node = Vec::new();
+        external_digest_slot_by_node.resize(node_count, None);
+        for (slot, (_, node_index)) in sorted_external_digests.iter().enumerate() {
+            let slot = u32::try_from(slot)
+                .expect("external digest slot should fit in u32 because MastForest node ids do");
+            external_digest_slot_by_node[*node_index] = Some(slot);
+        }
```
bobbinth: Once we have nodes sorted in the `MastForest` itself, this could should get reduced just to a couple of lines.

But also, if the order of nodes in the `MastForest` is different from the order in the wire format - doesn't this introduce extra complexity? My original thinking was that if the nodes are sorted in the `MastForest` already, it should be very easy to serialize/deserialize them because all `MastNodeId`s remain stable.

bobbinth (commented):
Looks good! Thank you. Not a full review but I left a few comments inline. The main one is about node ordering. On this front, I think this PR is introducing more complexity than is actually needed. I'm not sure we get much benefit from sorting external node digests in the wire format if the `MastForest` itself has different node ordering. So, I think this change is predicated on addressing https://github.com/0xMiden/miden-vm/issues/2473 first.

Overall, it seems like we should bite the bullet and address https://github.com/0xMiden/miden-vm/issues/2473 because w/o having this, we introduce quite a bit of extra complexity (and not just in this PR). My intuition is that with https://github.com/0xMiden/miden-vm/issues/2473 we'll end up with a drastically simpler structure (both for `MastForest` itself as well as sterilization code) and also cut a few thousand lines of code from the codebase.

Another thing we could consider is addressing https://github.com/0xMiden/miden-vm/issues/2473 in conjunction with https://github.com/0xMiden/miden-vm/issues/3040 - it may be natural to tackle both of them together (or one shortly after the other).

huitseeker:
I think we crossed wires here, so let me restate the history and what I think happened.

  In #2765, @bobbinth [suggested that external nodes should eventually precede internal nodes and be sorted in the `MastForest` struct itself](https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122440986).

I pushed back here: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3123654692

The distinction I was trying to make is that sorting external digests only in the wire format, while preserving the current arbitrary `MastForest` node order, breaks the implicit dense-by-kind lookup rule. To make that work, each serialized `External` node has to carry an explicit digest  slot. That adds exactly the slot plumbing, sort-order validation, and extra complexity that this PR now contains.

So I think the miscommunication is that “sort external nodes/digests” was being used for two different designs:

1. A future `MastForest` invariant: external nodes come first and are sorted in the actual in-memory forest. This probably wants #2473 first, and should make serialization simpler.
2. A wire-only sorted external digest section while keeping today’s `MastForest` ordering. This is what #3055 implements, and it is the costly version I was concerned about.

@bobbinth , your latest comments on #3055 seem to confirm that the wire-only version is not worth carrying:
- https://github.com/0xMiden/miden-vm/pull/3055#pullrequestreview-4249450277
- https://github.com/0xMiden/miden-vm/pull/3055#discussion_r3206251096

Given that, I think the clean resolution is to remove the wire-only sorting from this PR. I’ll keep the fixed-layout/hashless/untrusted-validation work, but revert external digests back to node-index / dense-by-kind order, remove the explicit external digest-slot plumbing and sorted-order invariant, and update the changelog accordingly.

core/src/mast/serialization/mod.rs
```diff
@@ -39,6 +39,8 @@
 //!
 //! (Counts)
 //! - nodes count (`usize`)
+//! - internal nodes count (`usize`)
+//! - external nodes count (`usize`)
```
bobbinth, line 43: I think line 41 here should be removed.

core/src/mast/serialization/info.rs
```diff
@@ -70,28 +71,49 @@ impl MastNodeEntry {
...
+    }
+
+    /// Constructs a new [`MastNodeEntry`] from a [`MastNode`], remapping child IDs when needed.
+    pub fn new_with_id_remap(
+        mast_node: &MastNode,
+        ops_offset: NodeDataOffset,
+        id_remap: Option<&[u32]>,
+    ) -> Self {
+        Self::new_inner(mast_node, ops_offset, id_remap)
+    }
```
bobbinth, line 84: I don't think this constructor is used anywhere.

core/src/mast/serialization/layout.rs
```diff
@@ -26,17 +26,16 @@ use crate::{
...
     pub(super) basic_block_len: usize,
     pub(super) node_entry_offset: usize,
     pub(super) external_digest_offset: usize,
-    #[cfg(test)]
-    pub(super) external_digest_count: usize,
     pub(super) node_hash_offset: Option<usize>,
-    #[cfg(test)]
-    pub(super) node_hash_count: usize,
+    pub(super) advice_map_offset: usize,
 }
```
bobbinth, line 39: Most of the offsets here should be computable from other data. It may make sense to convert them to methods (and if we do this, we can make the other fields private and exposed via methods as well).

bobbinth (approved):
Looks good! Thank you! I left some minor comments inline, but I think we can merge as is and address the comments in follow-ups.

huitseeker:
@bobbinth your comments have been integrated in the first commit of #3077.

PR #3057, opened 2026-04-25
docs: clarify mmr get absent leaf behavior

### LINKED ISSUE #3001: Consider documenting `mmr::get` behavior on absent leaf
_From the tx kernel audit report_

The authenticate_note procedure in prologue.masm (in `miden-protocol`) authenticates the MMR leaf associated with the block in which an input note was created. However, the behavior is not specified for the case where the note was created in the future relative to the reference block. In particular, the documentation of the mmr::get procedure does not clarify whether it fails or returns a special value when the requested leaf is not present in the MMR.

Consider documenting the expected behavior for this scenario to improve the readability and
maintainability of the codebase.

giwaov:
This documents the behavior requested in #3001. `mmr::get` only succeeds when the requested position is already present in the MMR (`pos < num_leaves`). If the position is outside the current MMR, the procedure fails rather than returning an empty word or another sentinel value.

Rationale: callers should not have to infer absent-leaf behavior from the implementation, especially for cases like checking a note against a reference block where a future position would otherwise be ambiguous.

I also added the required changelog entry.

Closes #3001.

Testing: `npm run build:dev`


docs/src/user_docs/core_lib/collections.md
```diff
@@ -17,7 +17,7 @@ The following procedures are available to read data from and make updates to a M
 
 | Procedure   | Description   |
 | ----------- | ------------- |
-| get         | Loads the leaf at the absolute position `pos` in the MMR onto the stack. Valid range for `pos` is between $0$ and $2^{32} - 1$ (both inclusive). Inputs: `[pos, mmr_ptr, ...]`. Output: `[N, ...]`. Where `N` is the leaf loaded from the MMR whose memory location starts at `mmr_ptr`. |
+| get         | Loads the leaf at the absolute position `pos` in the MMR onto the stack. Valid range for `pos` is between $0$ and $2^{32} - 1$ (both inclusive), and `pos` must refer to an existing leaf: `pos < num_leaves`, where `num_leaves` is stored at `mmr_ptr[0]`. The procedure fails instead of returning an empty or sentinel word when `pos` is outside the current MMR. Inputs: `[pos, mmr_ptr, ...]`. Output: `[N, ...]`. Where `N` is the leaf loaded from the MMR whose memory location starts at `mmr_ptr`. |
```
bobbinth: We should update documentation in `mmr.masm` file as well.

crates/lib/core/asm/collections/mmr.masm
bobbinth, line 16: Looking at the implementation below, it is actually not immediately clear to me where we perform the `pos < num_leaves` check. If it is implicit in some existing operations, would be good to document it. Otherwise, we could add this check somewhere around like 21 below.

crates/lib/core/asm/collections/mmr.masm
```diff
@@ -6,6 +6,10 @@ use miden::core::math::u64
 #!
 #! This MMR implementation supports only u32 positions.
 #!
+#! The position must refer to an existing leaf: `pos < num_leaves`, where `num_leaves` is stored
+#! at `mmr_ptr[0]`. The procedure fails instead of returning an empty or sentinel word when `pos`
```
huitseeker, line 10: Could this say that callers must ensure `pos < num_leaves` until the procedure enforces it? I don't think it always fails today. This local test passes and shows a one-leaf MMR returning the leaf for `pos = 2`:

```rust
#[test]
fn tmp_review_mmr_get_absent_single_leaf_returns_leaf() -> Result<(), MerkleError> {
    let leaves = &[11];
    let merkle_root = init_merkle_leaves(leaves)[0];

    let mut builder = AdviceStackBuilder::new();
    builder.push_for_adv_loadw(merkle_root);
    let advice_stack = builder.build_vec_u64();
    let source = "
        use miden::core::collections::mmr

        begin
            push.1 push.1000 mem_store
            padw adv_loadw push.1004 mem_storew_le dropw

            push.1000 push.2 exec.mmr::get

            swapw dropw
        end";

    let test = build_test!(source, &[], advice_stack, MerkleStore::new());
    test.expect_stack(&word_to_ints(&merkle_root));

    Ok(())
}
```

So this documents a guarantee the code does not yet provide.

huitseeker (changes requested):
The best we can do with a doc-only PR is to label the behavior as unspecified, because [we're not actually doing the check this PR claims happens](https://github.com/0xMiden/miden-vm/pull/3057#discussion_r3144169579). The best PR we could do is to actually perform the check.

giwaov:
Thanks for catching this. I updated the PR to enforce pos < num_leaves in mmr::get, so the docs now describe behavior the procedure actually guarantees.

huitseeker (approved):
This LGTM, thanks!

PR #3060, opened 2026-04-27
docs: document mmr::get failure behavior for absent leaves

### LINKED ISSUE #3001: Consider documenting `mmr::get` behavior on absent leaf

AlgofootPrint:
Closes #3001

  **Rationale**
  The documentation for `mmr::get` did not specify what happens when
  `pos >= num_leaves` (i.e., the requested leaf is absent from the MMR).
  This ambiguity was flagged in a transaction kernel audit report in the
  context of `authenticate_note`, where notes created in the future
  relative to the reference block would fall into this case.

  Tracing the implementation: when `pos >= num_leaves`, the algorithm
  computes `owner_candidates = 0` and calls `ilog2(0)`, which is
  undefined — causing the procedure to fail. It does not return a special
  value.

  **Changes**
  - Added a `Panics` note to the `#!` docstring in `mmr.masm`
  - Updated the corresponding entry in `docs/collections/mmr.md`

  **Test plan**
  Documentation-only change; no new code paths introduced.


bobbinth:
Closing as duplicate of https://github.com/0xMiden/miden-vm/pull/3057.

PR #3061, opened 2026-04-27
docs(assembly): fix adv_pipe memory address notation in instruction reference

### LINKED ISSUE #3006: Clarify `adv_pipe` documentation
_From the tx kernel audit_

> The documentation of the adv_pipe instruction states: "writes [D,E] to memory at a
and a+1 . a' ← a+2 ". However, this is incorrect, since 2 words (8 field elements) are
written to memory.

I think we should either clarify that we mean word indices here or use `a` and `a+4` which is probably the better choice since the switch to element-addressable memory.

AlgofootPrint:
Closes #3006

  **Rationale**
  The `adv_pipe` entry in `instruction_reference.md` described the memory
  writes as targeting `a` and `a+1` with `a' ← a+2`, which implies
  word-level addressing. Since memory is element-addressable and each word
  is 4 elements, the correct notation is `a` and `a+4` with `a' ← a+8`.

  This was already documented correctly in `io_operations.md` and in the
  design doc (`docs/src/design/stack/io_ops.md`). This change makes
  `instruction_reference.md` consistent with both.

  **Test plan**
  Documentation-only change; no code paths affected.


AlgofootPrint:
/quality-review 
This fixes a factual inaccuracy: the wrong memory addresses (a+1, a' ← a+2) could mislead
developers implementing or auditing code that relies on adv_pipe. The correct element-addressable offsets (a+4, a' ←
a+8) are already used in io_operations.md and the design doc this aligns the instruction reference with both.

bobbinth (approved):
Looks good! Thank you!

PR #3062, opened 2026-04-27
docs(processor): fix memory trace delta encoding comment

### LINKED ISSUE #2815: Memory trace documentation misstates delta encoding for same-word accesses

AlgofootPrint:
Closes #2815

  **Rationale**
  The doc comment for `Memory` in `processor/src/trace/chiplets/memory/mod.rs`
  contained two inaccuracies:

  1. Same-word delta was documented as `new_clk - old_clk - 1`, but both
     `append_range_checks` and `fill_trace` compute `new_clk - old_clk`.

  2. The first-row description stated `d0`, `d1`, and `d_inv` are zero.
     In reality `prev_clk` is initialized to `first_clk - 1`, so the
     first-row delta is always `1` — meaning `d0 = 1`, `d1 = 0`,
     `d_inv = 1`.

  No code changes — comments only.

### DIFF
--- a/processor/src/trace/chiplets/memory/mod.rs
+++ b/processor/src/trace/chiplets/memory/mod.rs
@@ -253,8 +253,8 @@ impl Memory {
     /// [RangeChecker] chiplet instance, along with their row in the finalized execution trace.
     pub fn append_range_checks(&self, memory_start_row: RowIndex, range: &mut RangeChecker) {
         // set the previous address and clock cycle to the first address and clock cycle of the
-        // trace; we also adjust the clock cycle so that delta value for the first row would end
-        // up being ZERO. if the trace is empty, return without any further processing.
+        // trace; we also adjust the clock cycle back by 1 so that the delta for the first row
+        // equals 1. if the trace is empty, return without any further processing.
         let (mut prev_ctx, mut prev_addr, mut prev_clk) = match self.get_first_row_info() {
             Some((ctx, addr, clk)) => (ctx, addr, clk.as_canonical_u64().wrapping_sub(1)),
             None => return,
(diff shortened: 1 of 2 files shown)


processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -92,7 +92,8 @@ const INIT_MEM_VALUE: Word = EMPTY_WORD;
 ///   checks on `w0`, `w1`, and `4 * w1`, these columns prove that memory addresses are valid 32-bit
 ///   values.
 ///
-/// For the first row of the trace, values in `d0`, `d1`, and `d_inv` are set to zeros.
+/// For the first row of the trace, `prev_clk` is initialized to `first_clk - 1`, so the delta is
```
huitseeker, line 95: Could we update the two nearby code comments too? `append_range_checks` and `fill_trace` still say the first-row delta ends up being ZERO, but the new doc correctly says it is `1`. Leaving those stale comments may send readers back to the same wrong idea.

Al-Kindi-0, line 95: `first_clk` may be confusing here, I would add a word or two to explain it

Al-Kindi-0 (approved):
LGTM, modulo a small comment

processor/src/trace/chiplets/memory/mod.rs
AlgofootPrint: Good catch, fixed in the follow-up commit. Also corrected a "pervious" typo in the fill_trace comment while there.

huitseeker (approved):
LGTM

PR #3063, opened 2026-04-27
fix(processor): respect MIN_STACK_DEPTH in DYNCALL stack depth recording

### LINKED ISSUE #2813: ExecutionTracer DYNCALL context capture inconsistent with MIN_STACK_DEPTH drop semantics

AlgofootPrint:
Closes #2813

  **Rationale**
  `ExecutionTracer::record_control_node_start` always computed
  `stack_depth_after_drop = depth - 1` for DYNCALL operations. However,
  when the stack is already at `MIN_STACK_DEPTH`, the drop is a no-op on
  depth — the VM does not go below the minimum. This made the execution
  tracer inconsistent with the parallel tracer (`parallel/tracer/mod.rs`),
  which already handles this case correctly.

  **Change**
  - When `depth > MIN_STACK_DEPTH`: record `depth - 1` (unchanged behavior)
  - When `depth == MIN_STACK_DEPTH`: record `depth` (no drop occurs)

  The `overflow_addr` is unaffected — `last_update_clk_in_current_ctx()`
  already returns `ZERO` when the overflow table is empty, which is the
  correct value for the min-depth case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Fixed the release dry-run publish cycle between `miden-air` and `miden-ace-codegen`, and preserved leaf-only DAG imports with explicit snapshots ([#2931](https://github.com/0xMiden/miden-vm/pull/2931)).
 - Added regression coverage for the exact `max_num_continuations` continuation-stack boundary ([#2995](https://github.com/0xMiden/miden-vm/pull/2995)).
 - Fixed AEAD padding handling so encrypt does not overwrite memory next to the plaintext buffer and decrypt leaves the plaintext output tail untouched ([#3008](https://github.com/0xMiden/miden-vm/pull/3008)).
+- Fixed `DYNCALL` execution tracing to clamp `parent_stack_depth` in `ExecutionContextInfo` to `MIN_STACK_DEPTH`, preventing an underflow when the stack is at minimum depth ([#2813](https://github.com/0xMiden/miden-vm/pull/2813)).
 
 #### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - Fixed the release dry-run publish cycle between `miden-air` and `miden-ace-codegen`, and preserved leaf-only DAG imports with explicit snapshots ([#2931](https://github.com/0xMiden/miden-vm/pull/2931)).
 - Added regression coverage for the exact `max_num_continuations` continuation-stack boundary ([#2995](https://github.com/0xMiden/miden-vm/pull/2995)).
 - Fixed AEAD padding handling so encrypt does not overwrite memory next to the plaintext buffer and decrypt leaves the plaintext output tail untouched ([#3008](https://github.com/0xMiden/miden-vm/pull/3008)).
-- Fixed `DYNCALL` execution tracing to clamp `parent_stack_depth` in `ExecutionContextInfo` to `MIN_STACK_DEPTH`, preventing an underflow when the stack is at minimum depth ([#2813](https://github.com/0xMiden/miden-vm/pull/2813)).
+- Fixed `DYNCALL` execution tracing to clamp `parent_stack_depth` to `MIN_STACK_DEPTH` and to record the correct post-pop `parent_next_overflow_addr` when the caller has multiple overflow entries ([#3063](https://github.com/0xMiden/miden-vm/pull/3063)).
 
 #### Changes
 
--- a/processor/src/trace/execution_tracer.rs
+++ b/processor/src/trace/execution_tracer.rs
@@ -910,9 +910,7 @@ mod tests {
 
     use super::*;
     use crate::{
-        FastProcessor, StackInputs,
-        continuation_stack::ContinuationStack,
-        tracer::Tracer,
+        FastProcessor, StackInputs, continuation_stack::ContinuationStack, tracer::Tracer,
     };
 
     /// Builds a [`MastForest`] containing a single DYNCALL node and returns the forest together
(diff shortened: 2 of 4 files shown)


processor/src/trace/execution_tracer.rs
```diff
@@ -319,15 +319,15 @@ impl ExecutionTracer {
...
-                    let stack_depth_after_drop = processor.stack().depth() - 1;
+                    // Note: the stack depth to record is after the DYNCALL drop. DYNCALL pops
+                    // the top element (the memory address of the callee hash) before entering
+                    // the new context. However, when the stack is already at MIN_STACK_DEPTH,
+                    // the drop is a no-op on depth — the stack stays at MIN_STACK_DEPTH.
+                    let stack_depth_after_drop = if processor.stack().depth() > MIN_STACK_DEPTH as u32 {
+                        processor.stack().depth() - 1
+                    } else {
+                        processor.stack().depth()
+                    };
```
plafer: Let's rewrite this using [`max()`](https://doc.rust-lang.org/std/cmp/fn.max.html)

```rs
let stack_depth_after_drop = max(processor.stack().depth() - 1, MIN_STACK_DEPTH);
```

processor/src/trace/execution_tracer.rs
plafer, line 1: Let's add a unit test too that calls `ExecutionTracer::start_clock_cycle()` directly, and then inspects the `ExecutionContextInfo` in the block added to `ExecutionTracer.block_stack`

AlgofootPrint: Done — updated to use max() as suggested.

processor/src/trace/execution_tracer.rs
AlgofootPrint: This is updated

processor/src/trace/execution_tracer.rs
AlgofootPrint: also done

plafer:
Latest changes look good - can you run `rustfmt`, and add a changelog entry?

AlgofootPrint:
Applied rustfmt and added a changelog entry. Let me know if anything else needs addressing!

processor/src/trace/execution_tracer.rs
```diff
@@ -319,15 +319,11 @@ impl ExecutionTracer {
...
-                    // DYNCALL operation performs a drop, and
-                    // records the stack depth after the drop as the beginning of
-                    // the new context. For more information, look at the docs for how the
-                    // constraints are designed; it's a bit tricky but it works.
-                    let stack_depth_after_drop = processor.stack().depth() - 1;
+                    // Note: the stack depth to record is after the DYNCALL drop. DYNCALL pops
+                    // the top element (the memory address of the callee hash) before entering
+                    // the new context. The depth cannot go below MIN_STACK_DEPTH.
+                    let stack_depth_after_drop =
+                        (processor.stack().depth() - 1).max(MIN_STACK_DEPTH as u32);
```
huitseeker: This still leaves `overflow_addr` as the value before the DYNCALL pop. When the caller has more than one overflow entry, the context needs the address after that pop.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Fixed the release dry-run publish cycle between `miden-air` and `miden-ace-codegen`, and preserved leaf-only DAG imports with explicit snapshots ([#2931](https://github.com/0xMiden/miden-vm/pull/2931)).
 - Added regression coverage for the exact `max_num_continuations` continuation-stack boundary ([#2995](https://github.com/0xMiden/miden-vm/pull/2995)).
 - Fixed AEAD padding handling so encrypt does not overwrite memory next to the plaintext buffer and decrypt leaves the plaintext output tail untouched ([#3008](https://github.com/0xMiden/miden-vm/pull/3008)).
+- Fixed `DYNCALL` execution tracing to clamp `parent_stack_depth` in `ExecutionContextInfo` to `MIN_STACK_DEPTH`, preventing an underflow when the stack is at minimum depth ([#2813](https://github.com/0xMiden/miden-vm/pull/2813)).
```
huitseeker: We link to the PR in Changelog, not the issue.

huitseeker (changes requested):
See my inline comment: this doesn't fix #2813. You can look at PR #2904’s `clk_after_pop_in_current_ctx()` fix and its multiple-overflow-entry regression test to close the gap.

processor/src/trace/execution_tracer.rs
AlgofootPrint: Thanks for the clear pointer to #2904. Added clk_after_pop_in_current_ctx() to OverflowTable (returns the second-to-last entry's clock, ZERO if fewer than two) and updated the DYNCALL branch in record_control_node_start to use it when depth > MIN_STACK_DEPTH. Also added a regression test that asserts
the correct post-pop overflow address when the caller has two overflow entries, and updated the CHANGELOG to link to this PR.

processor/src/trace/execution_tracer.rs
```diff
@@ -318,16 +318,21 @@ impl ExecutionTracer {
...
-                    // DYNCALL operation performs a drop, and
-                    // records the stack depth after the drop as the beginning of
-                    // the new context. For more information, look at the docs for how the
-                    // constraints are designed; it's a bit tricky but it works.
-                    let stack_depth_after_drop = processor.stack().depth() - 1;
+                    // Note: the stack depth to record is after the DYNCALL drop. DYNCALL pops
+                    // the top element (the memory address of the callee hash) before entering
+                    // the new context. The depth cannot go below MIN_STACK_DEPTH.
+                    let depth = processor.stack().depth();
+                    let (stack_depth_after_drop, overflow_addr) =
```
huitseeker: Could you run `cargo +nightly fmt --all` here? The nightly rustfmt check is failing on this new block.

AlgofootPrint: Done, applied cargo +nightly fmt --all   pushed in the latest commit.

huitseeker (approved):
LGTM!

AlgofootPrint:
Appreciate you

huitseeker:
We're going to go with #2904. 
- it tests the actual decoder/main-trace fields, not just ExecutionTracer internals.
- avoids adding the test-only FastProcessor::with_extra_stack_depth helper from #3063.

PR #3064, opened 2026-04-28
docs(assembly): strengthen field division warning and cross-reference u32div

### LINKED ISSUE #2911: Strengthen field division warning in assembly docs

AlgofootPrint:
## Summary

  Closes #2911.

  The existing caution about field division was easy to miss : it was a single
  sentence in a prose paragraph above the operations table, and nothing at the
  point of use (`div` row) flagged the issue.

  ### Changes

  **`field_operations.md`**
  - Replaced the buried sentence with a visible `> **Warning:**` blockquote
    explicitly stating that `div` computes $a \cdot b^{-1} \bmod p$ (not floor
    division) and linking to `u32div` for integer floor division.
  - Added an inline note to the `div` table row so the distinction is visible
    at the point of use.

  **`instruction_reference.md`**
  - Added the same inline note to the `div` row.


plafer (approved):
This is indeed makes it more clear 👍

plafer:
Similarly here, please sign the commit

AlgofootPrint:
Done, commit is now signed.

bobbinth (approved):
Looks good! Thank you!

PR #3071, opened 2026-04-28
Fix u256 wrapping_mul stack preservation

huitseeker:
This fixes `u256::wrapping_mul` so it no longer drops caller-owned stack values below its inputs.

The root cause is an extra cleanup epilogue at the end of the procedure. The product was already on top of the stack, but two `swapdw dropw dropw` rounds removed 16 more values below it.

The fix removes that extra cleanup and adds a regression test with sentinel values below the operands.

Fixes 0xmiden/security-issues#7


Al-Kindi-0 (approved):
LGTM

bobbinth (approved):
Looks good! Thank you!

Issue #2947, opened 2026-03-31
Remove `Library` and `Program` structs


bitwalker:
Now that projects and project assembly have landed, we need to begin to remove the `Library`, `KernelLibrary`, and `Program` structs, and refactor existing uses of those types around `MastForest`/`Package` instead.

* `Program` is a simple wrapper around a `MastForest` and a `Kernel` (which itself is just a vector of procedure roots). This can either become an internal type of the processor, or be removed entirely.
* `KernelLibrary` is a simple wrapper around a `Library`, a `Kernel`, with a `ModuleInfo` corresponding to the `$kernel` module itself. This can be trivially replaced.
* `Library` is a slightly bigger lift. The `Package` struct currently wraps an `Arc<Library>` for convenience, but ultimately, `Library` is just a wrapper around a `MastForest` and an export map (i.e. `Arc<Path>` is the key, `LibraryExport` is the value). The first step would be to migrate all existing `Library` users to `Package`, and then move the functionality implemented on `Library`, into `Package` itself (the `Package` type already has its own export map via `PackageManifest`, so this is really about swapping `Arc<Library>` for `Arc<MastForest>`, and beefing up the serialization code for `Package` to match the quality of `Library` serialization.

Rough implementation plan:

- Beef up the serialization code for `Package` to provide the same level of defense as the current `Library` serialization code. This also means ensuring that the surface area is tested to the same degree as `Library`. This is the main blocker to removing `Library` straight away.
- Migrate all uses of the three structs to use `Package` or `MastForest` directly. If there is still some value in `Program`, then we should make that an internal type of the processor crate (AFAIK that is the only real user), and make it convenient to build from a `Package`/`MastForest`.
- Once migrated, move existing `LIbrary` functionality that isn't already provided by `Package`, to the `Package` struct, and remove `Library`, `KernelLibrary`, and `Program`. As part of this, we'd also clean up the assembler implementation

bobbinth:
As a part of this (or maybe separately), I'd rename `Kernel` into something like `KernelDescriptor`.

bitwalker:
This is resolved via #3106. I don't think we should plan on removing `Program` for the time being, it is still somewhat useful - but if we do decide to remove it, we can treat that as a separate issue.

PR #3073, opened 2026-04-28
Merge audit fixes into next

huitseeker:
This merges `audit-fixes-2026` into `next`, the number of commits being due to squash-merge of #2850 / #2823. This concludes the use of this audit branch.

The merge keeps the fixes from the audit branch while preserving the newer code already on `next`. It also avoids duplicate changelog entries for fixes that had already landed.

Notable points (mostly noted for when I tackle #2947):
- Library deserialization now validates constant and type export names, not only procedure names.
- Library deserialization keeps the existing `Library::new` export checks.
- Duplicate audit changelog entries were removed.
- The continuation stack changelog link now correctly points to `#2825`.


plafer (approved):

bitwalker (approved):
Nothing caught my eye during my pass through this, but I didn't have time for a line-by-line review, so I'm largely relying on the fact that we previously reviewed the changes to the audit branch, and that if all our tests are passing, the merge here doesn't accidentally introduce any breakages.

PR #3074, opened 2026-04-28
Sync user docs with MASM source

### LINKED ISSUE #2694: Core-lib user docs are out of sync with core-lib crate docs
I believe there are currently two places in the Miden VM repo where `miden::core` is documented:
- `crates/lib/core/docs`
- `docs/src/user_docs/core_lib`

The latter ends up in the docs on https://docs.miden.xyz/ (or when using `make serve-docs` on next, which is what I did), but it is out of sync with the other one:

https://github.com/0xMiden/miden-vm/blob/59a9e344fd382801f0e28c02021b203ac6fc324e/docs/src/user_docs/core_lib/math/u64.md?plain=1#L13-L16

The current docs, for instance, show the old big-endian u64 format, and not the new one. This is confusing when migrating. It would be nice to have that updated or de-duplicated with/before the 0.21 release.

Al-Kindi-0:
## Describe your changes

Resolves #2694.

The reported issue (u64 BE/LE in `user_docs`, stale `adv.insert_hqword`, `overflowing_mul` vs `widening_mul`, `u32div` documented like `u32divmod`) turned out to be the visible tip of much wider drift between `docs/src/user_docs/...` and the MASM source. This PR attempts to rectify this as much as possible.

 Worth considering a sustainable fix before this happens again.

- Updated `CHANGELOG.md'


crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -6,7 +6,10 @@
 #! consume an amount of data which is a multiple of the rate (2 words).
 #!
 #! Input: []
-#! Output: [PERM, PERM, PERM, ...]
+#! Output: [R0, R1, C, ...]
```
huitseeker, line 9: This fixes the generated source docs, but `docs/src/user_docs/core_lib/crypto/hashes.md` still says `init_no_padding` outputs `[PERM, PERM, PERM, ...]`. 

Should that row be updated to `[R0, R1, C, ...]` too?

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -256,6 +256,7 @@ Common cryptographic operations, including hashing and Merkle tree manipulations
...
 | `mtree_merge`  | `[L, R, ...]`        | `[M, ...]`       | 16     | Merges Merkle trees with roots `L` (left) and `R` (right) into new tree `M`. Input trees retained.                                                                                                    |
 | `mtree_verify` | `[V, d, i, R, ...]`  | `[V,d,i,R,...]`  | 1      | Verifies Merkle path for node `V` at depth `d`, index `i` for tree `R` (from advice provider). <br /> _Can be parameterized with `err` code (e.g., `mtree_verify.err=123`). Default error code is 0._ |
+| `crypto_stream` | `[rate(8), cap(4), src_ptr, dst_ptr, ...]` | `[ciphertext(8), cap(4), src_ptr+8, dst_ptr+8, ...]` | 1 | Poseidon2-sponge keystream step against memory: loads two words from `src_ptr`, XOR-adds the rate (top 8 stack elements) element-wise to produce ciphertext, writes ciphertext to `dst_ptr`, replaces rate on stack with ciphertext, preserves capacity, increments both pointers by 8. Primitive used by `miden::core::crypto::aead`. |
```
huitseeker: The implementation does field addition here, not XOR. 

Could this say `adds` or `field-adds` instead of `XOR-adds`?

crates/lib/core/asm/crypto/hashes/poseidon2.masm
Al-Kindi-0: yep, fixed

docs/src/user_docs/assembly/instruction_reference.md
Al-Kindi-0: Indeed, corrected

huitseeker (approved):
> Worth considering a sustainable fix before this happens again.

Agreed. We do have [a drift check for those docs that are auto-generated](https://github.com/0xMiden/miden-vm/blob/next/.github/workflows/test.yml#L56-L94), which IIUC isn't the case here. If you have specific ideas on how to make the drift measureable and/or on how to close the gap, please open an issue.

Al-Kindi-0:
> > Worth considering a sustainable fix before this happens again.
> 
> Agreed. We do have [a drift check for those docs that are auto-generated](https://github.com/0xMiden/miden-vm/blob/next/.github/workflows/test.yml#L56-L94), which IIUC isn't the case here. If you have specific ideas on how to make the drift measureable and/or on how to close the gap, please open an issue.


The existing CI check covers only `crates/lib/core/docs/` but It does not cover `docs/src/user_docs/**` and that is were most of the drift was located.

PR #3076, opened 2026-04-29
Fix miden-project release dry-run

huitseeker:
Fixes the workspace-release CI:  `miden-project` uses `toml::Value`, but `toml` only exposes it with its `serde` feature.
This enables `toml/serde` for `miden-project`, so the crate builds in the release dry-run.

Checked with `./scripts/check-msrv.sh` and `cargo publish --workspace --dry-run`.


bobbinth (approved):
Looks good! Thank you!

PR #3077, opened 2026-04-29
Refine MastForest read policies (1/4)

huitseeker:
This PR makes MastForest reads use clearer trust and backing rules. On that later point: `Materialized` is fully-deserialized, memory-copied, `WireBacked` is using the wire bytes, and maximizing random-access.

  | Before | After |
  |---|---|
  | `SerializedMastForest` mixes structural inspection and cache access. | `MastForestWireView` (re-named) is trusted wire-backed cache access. |
  | Hashless bytes enter the structural view. | Hashless bytes enter only through untrusted validation. |
  | Trusted read backing is implicit. | `MastForestReadMode` selects `Materialized` or `WireBacked`. |
  | Full wire-backed debug data has unclear handling. | Full `DebugInfo` is length-walked, warned, and ignored. |
  | Untrusted reads expose many budget constructors. | Untrusted reads use default read, or options-based read. |
  | Parsing and validation budgets share one shape. | Options split wire budget from validation allocation budget. |
  | Header facts live as loose scanner state. | `MastForestHeader` is an internal parsed type. |
  | Advice access is tied to materialized forests. | `AdviceMapView` works for materialized and wire-backed reads. |

Trusted reads models cache access:

- `MastForest::read_from_bytes` materializes a complete trusted forest (full deserialization). 
- `MastForestWireView` gives random access over trusted bytes without materializing the forest, but still requires complete hashes.

Untrusted reads keep validation explicit: 
- there is no wire backing for the Untrusted format (the hash re-computation requires full deserialization),
- The wire budget bounds parsing. The validation budget bounds stripped/hashless helper allocations such as empty debug-info scaffolding, digest slot tables, and rebuilt hash tables.

<details><summary>API shape</summary>

```rust
MastForest::read_from_bytes(bytes)
MastForest::read_view_from_bytes(bytes, MastForestReadMode::Materialized)
MastForest::read_view_from_bytes(bytes, MastForestReadMode::WireBacked)

MastForestWireView::new(bytes)

UntrustedMastForest::read_from_bytes(bytes)
UntrustedMastForest::read_from_bytes_with_options(
    bytes,
    UntrustedMastForestReadOptions::new()
        .with_wire_byte_budget(...)
        .with_validation_allocation_budget(...),
)
```
</details>

<details><summary>Budget accounting</summary>

@plafer noted that per-section checks do not catch the sum of allocations across sections.

This PR keeps early per-section checks, but adds a second budget for validation-time helper allocations. `UntrustedMastForestReadOptions` now separates:

```rust
  .with_wire_byte_budget(...)
  .with_validation_allocation_budget(...)
```

The wire budget bounds parsing and wire-driven pre-sizing. The validation budget is a running budget: each tracked helper allocation subtracts from the remaining total. The parser still checks individual wire sections early, but untrusted validation now also accounts for cumulative helper allocations.
</details>

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -13,21 +13,18 @@
 //! ```
 //!
 //! [`UntrustedMastForest::read_from_bytes`] applies default parsing and validation budgets derived
-//! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_budget`] to tune only the
-//! wire-parsing budget, or [`UntrustedMastForest::read_from_bytes_with_budgets`] to tune both:
-//! the parsing budget limits allocations driven directly by wire counts while reading the payload,
-//! and the validation budget limits later helper allocations needed to materialize and check
-//! stripped or hashless payloads.
+//! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_options`] with
+//! [`UntrustedMastForestReadOptions`] to tune either budget: the wire byte budget limits
+//! allocations driven directly by wire counts while reading the payload, and the validation budget
+//! limits later helper allocations needed to materialize and check stripped or hashless payloads.
 //!
 //! ```ignore
-//! use miden_core::mast::UntrustedMastForest;
+//! use miden_core::mast::{UntrustedMastForest, UntrustedMastForestReadOptions};
 //!
-//! // Parsing budget only
-//! let forest = UntrustedMastForest::read_from_bytes_with_budget(&bytes, bytes.len())?
-//!     .validate()?;
-//!
-//! // Parsing budget plus explicit validation-allocation budget
-//! let forest = UntrustedMastForest::read_from_bytes_with_budgets(&bytes, bytes.len(), bytes.len() * 7)?
+//! let options = UntrustedMastForestReadOptions::new()
+//!     .with_wire_byte_budget(bytes.len())
+//!     .with_validation_allocation_budget(bytes.len() * 7);
+//! let forest = UntrustedMastForest::read_from_bytes_with_options(&bytes, options)?
 //!     .validate()?;
 //! ```
 //!
@@ -38,11 +35,10 @@
 //! In practice, the public entry points split into three policies:
 //! - [`MastForest::read_from_bytes`]: trusted full deserialization; rejects hashless payloads and
 //!   trusts serialized non-external digests.
-//! - [`SerializedMastForest::new`]: structural inspection path for local tooling; scans only the
-//!   layout needed for random access and may accept full, stripped, or hashless payloads, but it is
-//!   not an untrusted-validation entry point.
+//! - [`MastForestWireView::new`]: trusted wire-backed cache access; scans only the layout needed
+//!   for random access and rejects hashless payloads.
 //! - [`UntrustedMastForest::read_from_bytes`] and
-//!   [`UntrustedMastForest::read_from_bytes_with_budgets`]: untrusted paths; parse with bounded
+//!   [`UntrustedMastForest::read_from_bytes_with_options`]: untrusted paths; parse with bounded
 //!   readers and require [`UntrustedMastForest::validate`] before use.
 
 use alloc::{
@@ -91,10 +87,13 @@ pub use debuginfo::{
 };
 
 mod serialization;
-pub use serialization::{MastForestView, MastNodeEntry, MastNodeInfo, SerializedMastForest};
+pub use serialization::{
+    AdviceMapView, AdviceValueView, MastForestReadMode, MastForestReadView, MastForestView,
+    MastForestWireView, MastNodeEntry, MastNodeInfo,
+};
 
 mod untrusted;
-pub use untrusted::UntrustedMastForest;
+pub use untrusted::{UntrustedMastForest, UntrustedMastForestReadOptions};
 
 mod merger;
 pub(crate) use merger::MastForestMerger;
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -14,16 +14,16 @@
 //!
 //! [`UntrustedMastForest::read_from_bytes`] applies default parsing and validation budgets derived
 //! from the input size. Use [`UntrustedMastForest::read_from_bytes_with_options`] with
-//! [`UntrustedMastForestReadOptions`] to tune either budget: the wire byte budget limits
-//! allocations driven directly by wire counts while reading the payload, and the validation budget
-//! limits later helper allocations needed to materialize and check stripped or hashless payloads.
+//! [`UntrustedMastForestReadOptions`] to tune the wire byte budget. This limits allocations driven
+//! directly by wire counts while reading the payload. A separate validation helper budget is
+//! derived from it for later allocations needed to materialize and check stripped or hashless
+//! payloads.
 //!
 //! ```ignore
 //! use miden_core::mast::{UntrustedMastForest, UntrustedMastForestReadOptions};
 //!
 //! let options = UntrustedMastForestReadOptions::new()
-//!     .with_wire_byte_budget(bytes.len())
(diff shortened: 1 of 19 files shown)


core/src/mast/serialization/view.rs
```diff
@@ -1,13 +1,212 @@
...
+#[derive(Debug)]
+pub struct AdviceValueView<'a> {
+    inner: AdviceValueViewInner<'a>,
+}
+
+#[derive(Debug)]
+enum AdviceValueViewInner<'a> {
+    Borrowed(&'a [Felt]),
+    Owned(Vec<Felt>),
+}
```
plafer: This is equivalent to `Cow<'a, [Felt]>`; we should use that standard type instead

core/src/mast/serialization/view.rs
```diff
@@ -1,13 +1,212 @@
...
+    fn as_ref(&self) -> &[Felt] {
+        self.as_slice()
+    }
+}
+
+/// Read-only view over forest advice.
+#[derive(Debug)]
+pub struct AdviceMapView<'a> {
+    inner: AdviceMapViewInner<'a>,
+}
```
plafer, line 58: Would it not make more sense for `AdviceMapView` to be a trait, and use `impl AdviceMapView<'a>` anywhere that `AdviceMapView` is used now? This would naturally make it more extensible, and reduce boilerplate

core/src/mast/serialization/mod.rs
```diff
@@ -390,11 +393,29 @@ fn serialized_size_hint(forest: &MastForest, stripped: bool, hashless: bool) ->
...
+}
+
+/// Read-only trusted MAST forest handle.
+#[derive(Debug)]
+pub enum MastForestReadView<'a> {
+    /// A fully materialized forest.
+    Materialized(MastForest),
+    /// A trusted wire-backed cache view.
+    WireBacked(MastForestWireView<'a>),
+}
```
plafer, line 386: Similarly, do we really need this type, or can we just use `impl MastForestView` (or equivalent) everywhere? Since `MastForestView` is already a "view for reading".

plafer (commented):
Looks good. Mainly have questions about using traits directly rather than creating enums

miden-core-fuzz/fuzz_targets/mast_forest_validate.rs
plafer, line 1: `miden-core-fuzz/fuzz_targets/mast_node_info.rs` now fails to build (due to the renaming of `SerializedMastForest`). Opened #3080 as a result to make sure CI catches this in the future.

core/src/mast/serialization/mod.rs
huitseeker: I think this type is needed if we keep `read_view_from_bytes(bytes, mode)`. The selected backing is a runtime choice, so `impl MastForestView` cannot name both possible concrete return types. The alternatives are this enum, a boxed trait object, or removing the mode-based constructor and asking callers to choose `MastForest::read_from_bytes` or `MastForestWireView::new` directly. I prefer the enum because it is explicit and zero-allocation.

miden-core-fuzz/fuzz_targets/mast_forest_validate.rs
```diff
@@ -13,38 +12,31 @@
...
-        validate_untrusted_with_flags(UntrustedMastForest::read_from_bytes_with_budgets_and_flags(
+        validate_untrusted(UntrustedMastForest::read_from_bytes_with_options(
             data,
-            data.len(),
-            data.len(),
+            explicit_budget_options,
         ));
-        validate_untrusted_with_flags(UntrustedMastForest::read_from_bytes_with_flags(data));
         return;
     };
```
plafer: This block is copy/pasted below. I would instead write it as

```rs
    if let Ok(untrusted) = UntrustedMastForest::read_from_bytes(data) {
        let _ = untrusted.validate();
    };

    validate_untrusted(UntrustedMastForest::read_from_bytes_with_options(
        data,
        small_budget_options,
    ));
    validate_untrusted(UntrustedMastForest::read_from_bytes_with_options(
        data,
        explicit_budget_options,
    ));
```

core/src/mast/serialization/view.rs
huitseeker: You can look at what this looks like: https://github.com/huitseeker/miden-vm/commit/6164d0156aa3243426ec3f01c8e3e1cd8089f854

The trait shape pushes us into RPITIT/object-safety trouble at the `MastForestView` seam:

 > `MastForestView` is not dyn compatible ... because method `advice_map` references an `impl Trait` type in its return type

It also breaks runtime enum dispatch:

> distinct uses of `impl Trait` result in different opaque types

We would have to break out GATs for this, which is IMHO not worth the trouble.

core/src/mast/serialization/view.rs
huitseeker: Right, will fix

miden-core-fuzz/fuzz_targets/mast_forest_validate.rs
huitseeker: Thanks! Will fix here.

core/src/mast/serialization/mod.rs
plafer: Ah yes makes sense, let's leave it as is

plafer (approved):
LGTM!

core/src/mast/mod.rs
```diff
@@ -13,21 +13,18 @@
...
-//! let forest = UntrustedMastForest::read_from_bytes_with_budget(&bytes, bytes.len())?
-//!     .validate()?;
+//! use miden_core::mast::{UntrustedMastForest, UntrustedMastForestReadOptions};
 //!
-//! // Parsing budget plus explicit validation-allocation budget
-//! let forest = UntrustedMastForest::read_from_bytes_with_budgets(&bytes, bytes.len(), bytes.len() * 7)?
+//! let options = UntrustedMastForestReadOptions::new()
+//!     .with_wire_byte_budget(bytes.len())
+//!     .with_validation_allocation_budget(bytes.len() * 7);
+//! let forest = UntrustedMastForest::read_from_bytes_with_options(&bytes, options)?
```
bobbinth, line 27: Question (and this was probably discussed in some earlier PRs and I missed it): why do we need two separate budgets? Or more specifically, would it be possible to infer one from another so that the user only needs to think about one budget?

For example, in the protocol, when we talk about limits on account/script code, we usually mean the wire budget, and the assumption is that in-memory representation, while requiring more space, is proportional to the serialized size (and the multiplier is not huge - e.g., under 10x).

bobbinth (approved):
Not a very detailed review - but looks good! Thank you! I left just one question inline.

core/src/mast/mod.rs
huitseeker: Thanks, yes. I think the API should expose one simple “wire budget” knob unless a caller has a strong reason to do more.

The reason there are two budgets internally is that there are two metered phases. While reading, allocations are bounded by wire facts: the input has at most `bytes.len()` bytes. Later, validation may allocate helper tables that are not present on the wire, such as digest slots or rebuilt hash tables for stripped/hashless input. @plafer caught the need for cumulative accounting here: https://github.com/0xMiden/miden-vm/pull/3055#discussion_r3156689445

That split came from #2765. You also flagged there that the * 7 wording was confusing, since it sounded like “1KB on wire means 7KB RAM”: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122482016. I clarified it as a coarse bound, not a peak-memory formula: https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3123610928

So I agree with simplifying the public shape, but I would keep the two-phase metering internally. The public option can be “wire budget”, and the validation helper budget can be derived from it by a conservative multiplier. That keeps the user model simple without losing the guardrail this split was added for.

PR #3078, opened 2026-04-29
Fix: mast_forest_validate fuzz target inference

huitseeker:
Fixes the `mast_forest_validate` fuzz target build failure from the latest daily fuzz job.

The helper closures used for validating `UntrustedMastForest` results were missing explicit parameter types, which caused Rust to fail inference with `E0282` when compiling the fuzz target. This annotates the closures with the concrete `Result` types returned by the untrusted deserialization helpers.


plafer (approved):
👍

PR #3080, opened 2026-04-29
chore(CI): add fuzz-check job

plafer:
Adds a lightweight CI job that runs `cargo check` on the `miden-core-fuzz` crate; otherwise the current CI doesn't notice compilation errors.


miden-core-fuzz/Cargo.lock
plafer, line 1: This made me notice we didn't have a `Cargo.lock` file committed. Not sure if there are any good reasons not to?

.github/workflows/fuzz-check.yml
plafer, line 1: Just thinking now, we could alternatively expand the `build.yml` workflow to include this job. Thoughts @huitseeker?

huitseeker: Yes, expanding the build.yml job instead will make the triggers much simpler to read.

miden-core-fuzz/Cargo.lock
huitseeker: In the miden-core-fuzz, I think we don't care this much about reproducibility. But not a hill I'd be willing to die on.

plafer: Agreed, though I'm more approaching this from a pragmatic point of view - the file was auto-generated, and it's painful to remove it manually (or selectively not add it to commit). Will add it to gitignore instead.

plafer: Actually I now believe committing the `Cargo.lock` is better for CI caching purposes, since [the hashed `Cargo.lock` is used](https://github.com/marketplace/actions/rust-cache#cache-effectiveness) as part of the cache key, and so fixing it should improve cache hits.

.github/workflows/fuzz-check.yml
plafer: Moved to `build.yml`

.github/workflows/build.yml
```diff
@@ -31,3 +31,21 @@ jobs:
...
+  fuzz-check:
+    name: Check for miden-core-fuzz crate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      - name: Install Rust toolchain
+        run: rustup toolchain install --no-self-update
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
```
plafer: Two notes:
1. Here, I run `rustup toolchain install` instead of `rustup update` (used elsewhere), since AFAIU `toolchain install` will only install the toolchain specified in our `rust-toolchain.toml`, whereas `rustup update` updates stable, nightly, and our toolchain.
2. I run this *before* the `rust-cache` action so that `rust-cache` uses the right cache keys (unlike how it's done e.g. in `test.yml`).

If the above is correct, then I can update the rest of the workflows in a similar way in a separate PR.

.github/workflows/build.yml
```diff
@@ -31,3 +31,21 @@ jobs:
...
+        with: 
+          # We need to explicitly cache the miden-core-fuzz workspace, as only the root workspace is
+          # cached by default.
+          workspaces: |
+            .
+            miden-core-fuzz
+      - name: Check miden-core-fuzz crate
+        run: cd miden-core-fuzz && cargo check
```
huitseeker: Could this use `cargo check --locked` here? Since this PR adds `miden-core-fuzz/Cargo.lock`, plain `cargo check` can update a stale lockfile in CI and still pass, so CI may not catch when the committed lockfile is out of sync.

miden-core-fuzz/Cargo.lock
huitseeker: I think it will reduce cache consumption (something that on WarpBuild is less of a concern) rather than improve cache hits.

.github/workflows/build.yml
```diff
@@ -31,3 +31,21 @@ jobs:
...
+    name: Check for miden-core-fuzz crate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      - name: Install Rust toolchain
+        run: rustup toolchain install --no-self-update
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        with:
```
huitseeker: Nit: trailing space.

.github/workflows/build.yml
plafer: Done

miden-core-fuzz/Cargo.lock
plafer: Switched to using WarpBuild

huitseeker (approved):
Thanks!

PR #3084, opened 2026-04-30
Make all internal `core::math` procedures native little-endian

### LINKED ISSUE #2641: Make internal `core::math` procedures little-endian
We've recently updated external interfaces of `core::math::u64`, `core::math::u128`, and `core::math::u256` procedures to take parameters in the little endian form. However, in some cases we kept internal implementations big endian to simplify the transition. We should revisit these to make sure internal implementations are little endian too. I believe the current state is:

- `core::math::u64` - there are a few procedures which internally use big-endian form.
- `core::math::u128` - I believe the current procedures are already in little-endian, but we should double-check.
- `core::math::u256` - I think most internal procedures there are in big-ending form. So, these will require adjustments.

Al-Kindi-0:
## Describe your changes

Closes #2641 

- Updated `CHANGELOG.md'


huitseeker (approved):
LGTM, please add a CHANGELOG line.

plafer (approved):
👌

crates/lib/core/docs/math/u64.md
```diff
@@ -25,9 +25,9 @@
...
+| rotl | Performs left rotation of one unsigned 64-bit integer.<br />The input value to be rotated is assumed to be represented using 32 bit limbs, but this is not checked.<br />The rotation amount n should be in the range [0, 64), otherwise it will result in an error.<br />Stack transition looks as follows:<br />[n, a_lo, a_hi, ...] -> [c_lo, c_hi, ...], where c = a <<< n (rotate left).<br />This takes 46 cycles.<br /> |
+| rotr | Performs right rotation of one unsigned 64-bit integer.<br />The input value to be rotated is assumed to be represented using 32 bit limbs, but this is not checked.<br />The rotation amount n should be in the range [0, 64), otherwise it will result in an error.<br />Stack transition looks as follows:<br />[n, a_lo, a_hi, ...] -> [c_lo, c_hi, ...], where c = a >>> n (rotate right).<br />This takes 60 cycles.<br /> |
```
bobbinth, line 30: It seems like cycle counts have increased a bit: I would have expected them to go down slightly as we don't need to do the extra stack rotations before/after the body of the procedures. Or were the previous values obsolete?

Al-Kindi-0: The previous values were most probably obsolete or not tight in the first place. The current cycle counts I have more confidence in as they were computed consistently using a small Rust program (i.e., not manually).

There was also one issue that might have caused some mismatch, namely, we changed which of `_be` or `_le` versions of memory operations was native but forgot to update the cycle counts at some point. Though this point is speculative.

In any case, I think we need something more sustainable long term.

PR #3085, opened 2026-04-30
Bound live advice map size

### LINKED ISSUE #2806: Bound the maximum allowed advice map size
#2805 bounds the maximum size of each advice map entry, but this still allows an attacker to create entries of the maximum amount for each key in the advice map. We should then also bound the maximum size of the `AdviceMap`, through a new `ExecutionOptions` field.

The problem today is that there are *many* assumptions that we can build an `AdviceMap` without errors:

- The `From<BTreeMap<_>>` [constructors](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/core/src/advice/map.rs#L119-L129)
- The `FromIterator` of entries [constructor](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/core/src/advice/map.rs#L140-L153)
- The `Extend` [trait implementation](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/core/src/advice/map.rs#L155-L165).

This has downstream effects on `AdviceInputs` constructors, which [assume](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/core/src/advice/mod.rs#L67-L73) we can construct a map without errors, meaning that this will end up being a decently big breaking change.

huitseeker:
This caps the live advice map used during execution.

Advice map entries could grow without a total limit. Some paths also inserted fixed-size values without checking the per-entry value limit.

The fix adds one shared limit for the total number of field elements in live advice map keys and values. `AdviceProvider` enforces both the per-entry value limit and the total element limit for initial advice, host map extensions, and system-event inserts.

`FastProcessor` advice setup is now fallible because loading advice creates bounded live state. Call sites now return the validation error instead of panicking.

Fixes #2806. Fixes 0xmiden/security-issues#13. Fixes 0xmiden/security-issues#17.

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -17,6 +17,8 @@ pub struct ExecutionOptions {
     /// Maximum number of field elements that can be inserted into the advice map in a single
     /// `adv.insert_mem` operation.
     max_adv_map_value_size: usize,
+    /// Maximum total number of field elements allowed in live advice map keys and values.
+    max_adv_map_elements: usize,
     /// Maximum number of input bytes allowed for a single hash precompile invocation.
     max_hash_len_bytes: usize,
     /// Maximum number of continuations allowed on the continuation stack at any point during
@@ -33,6 +35,7 @@ impl Default for ExecutionOptions {
             enable_tracing: false,
             enable_debugging: false,
             max_adv_map_value_size: Self::DEFAULT_MAX_ADV_MAP_VALUE_SIZE,
+            max_adv_map_elements: Self::DEFAULT_MAX_ADV_MAP_ELEMENTS,
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
         }
@@ -50,8 +53,14 @@ impl ExecutionOptions {
     pub const DEFAULT_CORE_TRACE_FRAGMENT_SIZE: usize = 4096; // 2^12
 
     /// Default maximum number of field elements in a single advice map value inserted via
-    /// `adv.insert_mem`. Set to 2^17 (~1 MB given 8-byte field elements).
+    /// execution-time advice map mutations. Set to 2^17 (~1 MB given 8-byte field elements).
     pub const DEFAULT_MAX_ADV_MAP_VALUE_SIZE: usize = 1 << 17;
+
+    /// Default maximum total number of field elements in live advice map keys and values.
+    ///
+    /// Set to 2^20 so the default allows multiple maximum-sized entries while still providing a
+    /// finite host-memory backstop. Each entry contributes 4 key elements plus its value length.
+    pub const DEFAULT_MAX_ADV_MAP_ELEMENTS: usize = 1 << 20;
 
     /// Default maximum number of input bytes for a single hash precompile invocation (e.g.
     /// keccak256, sha512, etc.). Set to 2^20 (1 MB).
@@ -121,6 +130,7 @@ impl ExecutionOptions {
             enable_tracing,
             enable_debugging,
             max_adv_map_value_size: Self::DEFAULT_MAX_ADV_MAP_VALUE_SIZE,
+            max_adv_map_elements: Self::DEFAULT_MAX_ADV_MAP_ELEMENTS,
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
         })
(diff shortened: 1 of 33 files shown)


processor/src/execution_options.rs
```diff
@@ -193,26 +203,37 @@ impl ExecutionOptions {
         self.enable_debugging
     }
 
-    /// Returns the maximum number of field elements allowed in a single advice map value
-    /// inserted via `adv.insert_mem`.
+    /// Returns the maximum number of field elements allowed in a single live advice map value.
     #[inline]
     pub fn max_adv_map_value_size(&self) -> usize {
         self.max_adv_map_value_size
     }
```
plafer, line 210: Now that we properly bound the total number of elements in the entire advice map, do we still need this per-key bound?

processor/src/host/advice/mod.rs
```diff
@@ -334,13 +441,46 @@ impl AdviceProvider {
...
+                },
+            )?;
+            added = added.checked_add(entry_elements).ok_or(
+                AdviceError::AdvMapElementBudgetExceeded {
+                    current: self.map_element_count,
+                    added: usize::MAX,
+                    max: self.max_map_elements,
+                },
+            )?;
+        }
```
plafer, line 476: We currently track the advice map budgets in the `AdviceProvider`; why not directly in the `AdviceMap`? IMO it's cleaner from an encapsulation perspective, as well as allows us to have access to those bound checks if we were to use the `AdviceMap` in some other context.

Then, we also wouldn't need to "soft-duplicate" the merging logic here (where we keep track of how many elements will be added by `self.map.merge()`). If ever that `merge()` logic changes, this counting algorithm becomes outdated.

huitseeker: It's one approach. Morally speaking, we put the global bound in `AdviceProvider` because it is execution policy, not map structure.

`AdviceMap` is a plain core container: it can be built, deserialized, merged, and passed around without knowing `ExecutionOptions`. `AdviceProvider` owns the live execution state and the active limits, so it can enforce the budget at every live insertion point: initial advice, host mutations, and system-event inserts. It also can account atomically for replacements and failed extends.

So `AdviceMap` only exposes `total_element_count()`. The provider decides whether that count is allowed.
  
But we can do it the way you suggest as well.

processor/src/execution_options.rs
huitseeker: Yes, we kept the per-key value bound.

The total element bound prevents unbounded aggregate growth. The per-value bound still preserves the existing invariant that no single advice-map value can be too large. We decided not to remove that because it is a separate, local shape limit and existing public execution option. The model is:

  - `max_adv_map_elements`: total live key + value field elements in the provider’s advice map.
  - `max_adv_map_value_size`: maximum value length for one key.

@Al-Kindi-0 @adr1anh may have opinions on whether we'll have need for this in the future.

processor/src/host/advice/mod.rs
plafer: Yup I see your point - works for me. Can you document basically that this is intentional (i.e. a note reflecting this design choice in `AdviceProvider` and `AdviceMap` docstrings)

processor/src/execution_options.rs
plafer: Yeah my question was basically: is there value in bounding per-key slots? If we think there might be, that's fine by me. Though from my understanding the only "actual" requirement from a security perspective is to ensure no attacker can force too large an allocation - not so much that each value be small in and of itself.

TLDR if it actually makes sense to keep it, that works for me.

plafer (approved):
LGTM!

Al-Kindi-0 (approved):
LGTM

processor/src/execution_options.rs
Al-Kindi-0: It doesn't hurt to have both, provided the bounds are logical. If we had a bound on the number of keys allowed, then a bound on the size of values would be enough. The issue is that both the number of entries (i.e., keys) and the size of each entry (i.e., values) are not bounded, and hence the dual bound makes sense, in my opinion.

processor/src/host/advice/mod.rs
```diff
@@ -48,15 +49,70 @@ const MAX_ADVICE_STACK_SIZE: usize = 1 << 17;
...
 ///    - used to produce a STARK proof using a precompile VM, which can be verified in the epilog of
 ///      the program.
-#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct AdviceProvider {
     stack: VecDeque<Felt>,
     map: AdviceMap,
+    map_element_count: usize,
+    max_map_value_size: usize,
+    max_map_elements: usize,
```
bobbinth, line 58: It feels a bit link an overkill to have 3 settings for the advice map size (I think this was discussed in one of the comments above). Also, don't we need to apply some limit to the stack (and maybe store) components as well to make sure we don't run into memory allocation issues?

Maybe a better approach could be to specify maximum size of the whole advice provider. This could simplify the interface a bit. I'll create an issue for this.

huitseeker: > Also, don't we need to apply some limit to the stack (and maybe store) components as well to make sure we don't run into memory allocation issues?

#3086 

> Maybe a better approach could be to specify maximum size of the whole advice provider.

Valid approach but the tradeoff here is that it's easier on the maintainers, but harder on users to see if they are compliant.

plafer: I think Bobbin was referring to the advice stack though (whereas #3086 addresses the operand stack size)

huitseeker: Oh, ofc, that makes more sense!

bobbinth: After the comment I realized that advice stack is actually already limited - but the limit is hard-coded. I mentioned this in https://github.com/0xMiden/miden-vm/issues/3107.

PR #3086, opened 2026-05-01
Grow FastProcessor stack up to a configured depth limit

### LINKED ISSUE #2818: Introduce mechanism to grow stack buffer size when filled
Currently, the `FastProcessor` [fails execution](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/processor/src/fast/operation.rs#L316) if we reach the "far end" of the internal stack buffer. We should instead grow the internal stack buffer.

Note however that we shouldn't grow the stack buffer indefinitely, as this allows an attacker to provide a program `repeat.VERY_LARGE_NUMBER push.1 end` that exhausts the prover's memory. Hence, we should also provide a new `ExecutionOptions` value to control how much memory the stack buffer is allowed to take, and grow it up until that value. We could also just initialize the stack buffer to that max value, but that's overkill for running most programs who won't need most of that memory.

This issue could be done in conjunction with #2574.

### LINKED ISSUE #2834: Stack size is inconsistently limited in the processor
The processor enforces stack depth limits in different ways:

1. `MIN_STACK_DEPTH` (16) — minimum depth maintained during calls
2. `MAX_STACK_DEPTH` — maximum depth before operations fail

However, some code paths check these limits while others assume they are satisfied. This inconsistency could lead to:
- Operations that should fail succeeding (or vice versa)
- Different behavior between debug and release builds

Related to #2574 (Investigate using VecDeque for stack).

Recommendation: audit all stack depth checks for consistency, or consider unifying on a single bounded-stack abstraction.

huitseeker:
Closes #2818, closes #2834.

`FastProcessor` used a fixed stack buffer as both storage and an implicit stack limit. Other execution paths treated stack depth as a VM rule, so the limit was not enforced in one clear place.

This adds `ExecutionOptions::max_stack_depth` as the operand stack depth limit. The fast stack now grows on demand, but only up to that limit. When the limit is exceeded, execution returns `StackDepthLimitExceeded` instead of an internal stack overflow.

The same limit is passed into trace replay, so fast execution, traced execution, step execution, and replay use the same rule.

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -1,4 +1,5 @@
 use miden_air::trace::MIN_TRACE_LEN;
+use miden_core::program::MIN_STACK_DEPTH;
 
 // EXECUTION OPTIONS
 // ================================================================================================
@@ -24,6 +25,9 @@ pub struct ExecutionOptions {
     /// Maximum number of continuations allowed on the continuation stack at any point during
     /// execution.
     max_num_continuations: usize,
+    /// Maximum number of field elements allowed on the operand stack in the active execution
+    /// context.
+    max_stack_depth: usize,
 }
 
 impl Default for ExecutionOptions {
@@ -38,6 +42,7 @@ impl Default for ExecutionOptions {
             max_adv_map_elements: Self::DEFAULT_MAX_ADV_MAP_ELEMENTS,
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
+            max_stack_depth: Self::DEFAULT_MAX_STACK_DEPTH,
         }
     }
 }
@@ -69,6 +74,12 @@ impl ExecutionOptions {
     /// Default maximum number of continuations allowed on the continuation stack.
     /// Set to 2^16 (65536).
     pub const DEFAULT_MAX_NUM_CONTINUATIONS: usize = 1 << 16;
+
+    /// Default maximum number of field elements allowed on the operand stack.
+    ///
+    /// This preserves the effective stack depth ceiling imposed by the previous fixed
+    /// `FastProcessor` stack buffer.
+    pub const DEFAULT_MAX_STACK_DEPTH: usize = 6615;
 
     // CONSTRUCTOR
     // --------------------------------------------------------------------------------------------
@@ -133,6 +144,7 @@ impl ExecutionOptions {
             max_adv_map_elements: Self::DEFAULT_MAX_ADV_MAP_ELEMENTS,
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
+            max_stack_depth: Self::DEFAULT_MAX_STACK_DEPTH,
         })
     }
 
@@ -246,10 +258,33 @@ impl ExecutionOptions {
         self.max_num_continuations
     }
 
+    /// Returns the maximum number of field elements allowed on the operand stack in the active
+    /// execution context.
+    #[inline]
+    pub fn max_stack_depth(&self) -> usize {
+        self.max_stack_depth
+    }
+
     /// Sets the maximum number of continuations allowed on the continuation stack.
     pub fn with_max_num_continuations(mut self, max_num_continuations: usize) -> Self {
         self.max_num_continuations = max_num_continuations;
         self
+    }
+
+    /// Sets the maximum number of field elements allowed on the operand stack in the active
+    /// execution context.
+    pub fn with_max_stack_depth(
+        mut self,
+        max_stack_depth: usize,
+    ) -> Result<Self, ExecutionOptionsError> {
+        if max_stack_depth < MIN_STACK_DEPTH {
+            return Err(ExecutionOptionsError::MaxStackDepthTooSmall {
+                max_stack_depth,
+                min_stack_depth: MIN_STACK_DEPTH,
+            });
+        }
+        self.max_stack_depth = max_stack_depth;
+        Ok(self)
     }
 }
 
@@ -268,6 +303,11 @@ pub enum ExecutionOptionsError {
     MaxCycleNumTooBig { max_cycles: u32, max_cycles_limit: u32 },
     #[error("core trace fragment size must be greater than 0")]
     CoreTraceFragmentSizeTooSmall,
+    #[error("maximum stack depth {max_stack_depth} must be at least {min_stack_depth}")]
+    MaxStackDepthTooSmall {
+        max_stack_depth: usize,
+        min_stack_depth: usize,
+    },
 }
 
 // TESTS
@@ -325,5 +365,21 @@ mod tests {
         let opts = ExecutionOptions::new(Some(100), 64, 1024, false, false);
         assert!(opts.is_ok());
         assert_eq!(opts.unwrap().expected_cycles(), 64);
+    }
+
+    #[test]
+    fn max_stack_depth_validates_minimum_depth() {
+        let result = ExecutionOptions::default().with_max_stack_depth(MIN_STACK_DEPTH - 1);
+        assert!(matches!(
+            result,
+            Err(ExecutionOptionsError::MaxStackDepthTooSmall {
+                max_stack_depth,
+                min_stack_depth: MIN_STACK_DEPTH,
+            }) if max_stack_depth == MIN_STACK_DEPTH - 1
+        ));
+
+        let result = ExecutionOptions::default().with_max_stack_depth(MIN_STACK_DEPTH);
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap().max_stack_depth(), MIN_STACK_DEPTH);
     }
 }
--- a/processor/src/fast/mod.rs
+++ b/processor/src/fast/mod.rs
@@ -642,7 +642,7 @@ impl FastProcessor {
         }
     }
 
-    fn grow_stack_buffer(&mut self, min_len: usize) {
+    fn grow_stack_buffer(&mut self, requested_min_len: usize) {
         // The maximum allocation is tied to the logical operand stack depth, not to the current
         // buffer position. Using `stack_bot_idx` here would make the allocation ceiling drift when
         // the live stack has moved away from the initial base.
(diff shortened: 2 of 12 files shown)


processor/src/fast/mod.rs
```diff
@@ -587,6 +592,52 @@ impl FastProcessor {
...
+
+    fn grow_stack_buffer(&mut self, min_len: usize) {
+        let max_len = self
+            .stack_bot_idx
+            .saturating_add(self.options.max_stack_depth())
+            .saturating_add(1);
+        let mut new_len = self.stack.len();
+        while new_len < min_len {
+            new_len = new_len.saturating_mul(2);
+        }
```
```suggestion
        while new_len <= min_len {
            new_len = new_len.saturating_mul(2);
        }
```
nit: if `new_len` reach `min_len` we can stop. This would also infinite loop in the (unlikely) case where `min_len == usize::MAX`.

processor/src/fast/mod.rs
```diff
@@ -96,7 +100,7 @@ const INITIAL_STACK_TOP_IDX: usize = 250;
 #[derive(Debug)]
 pub struct FastProcessor {
     /// The stack is stored in reverse order, so that the last element is at the top of the stack.
-    stack: Box<[Felt; STACK_BUFFER_SIZE]>,
+    stack: Box<[Felt]>,
```
plafer, line 110: Out of curiosity, did you try using a `Vec` here? IIRC I had tried it initially and observed a significant drop in performance - though I wonder if you observed the same?

*If* there weren't differences in performance, this would allow us to use `Vec::resize` to grow the stack, which IIUC would use a realloc under the hood, and hence might sometimes successfully grow the buffer without needing to copy the contents over.

processor/src/fast/mod.rs
```diff
@@ -587,6 +592,52 @@ impl FastProcessor {
...
+        if top_idx >= self.stack.len() {
+            self.grow_stack_buffer(top_idx + 1);
+        }
+    }
+
+    fn grow_stack_buffer(&mut self, min_len: usize) {
+        let max_len = self
+            .stack_bot_idx
+            .saturating_add(self.options.max_stack_depth())
+            .saturating_add(1);
```
plafer: We don't explicitly ensure that `min_len <= max_len` - which *might* occur when called from `ensure_stack_capacity_for_top_idx()`? There are many ways to address (debug_assert, return error, explicitly ensure this is the case in callers) - will let you pick whatever you think works best.

Though if we're growing the stack buffer, we could at the same time reset the `stack_bot_idx`. My only "worry' when looking at this is if the attacker is ever able to increase `stack_bot_idx` (not sure if that's possible today but might in the future), then they can force us to allocate basically `usize::MAX` (since we allocate `stack_bot_idx + max_stack_depth`). This would ultimately mean that we merge this `grow_stack_buffer()` functionality with the `reset_stack_in_buffer()` functionality.

processor/src/fast/mod.rs
```diff
@@ -248,8 +252,9 @@ impl FastProcessor {
             // Note: we use `Vec::into_boxed_slice()` here, since `Box::new([T; N])` first allocates
             // the array on the stack, and then moves it to the heap. This might cause a
             // stack overflow on some systems.
-            let mut stack: Box<[Felt; STACK_BUFFER_SIZE]> =
-                vec![ZERO; STACK_BUFFER_SIZE].into_boxed_slice().try_into().unwrap();
+            debug_assert_eq!(ExecutionOptions::DEFAULT_MAX_STACK_DEPTH, DEFAULT_MAX_STACK_DEPTH);
```
plafer: This could be a const assert

processor/src/fast/mod.rs
```diff
@@ -587,6 +592,52 @@ impl FastProcessor {
...
+            .saturating_add(self.options.max_stack_depth())
+            .saturating_add(1);
+        let mut new_len = self.stack.len();
+        while new_len < min_len {
+            new_len = new_len.saturating_mul(2);
+        }
+        new_len = new_len.min(max_len).max(min_len);
+
+        let mut new_stack = vec![ZERO; new_len].into_boxed_slice();
+        new_stack[..self.stack.len()].copy_from_slice(&self.stack);
```
```suggestion
        new_stack[..self.stack_top_idx].copy_from_slice(&self.stack[..self.stack_top_idx]);
```
Basically we only need to copy up to `stack_top_idx` (everything after that is `ZERO`s) - which could make a significant difference when called from `ensure_stack_capacity_for_top_idx()`, where the stack only has 16 elements.

processor/src/fast/mod.rs
```diff
@@ -587,6 +592,52 @@ impl FastProcessor {
...
+    ///
+    /// The operand stack depth limit is the semantic resource bound; the buffer is only an
+    /// implementation detail. We therefore check the logical depth before allocating so a program
+    /// cannot force memory growth beyond `ExecutionOptions::max_stack_depth()`. When storage does
+    /// need to grow, it grows geometrically and remains heap-allocated as a boxed slice. A
+    /// `SmallVec` would put a useful inline buffer inside `FastProcessor`, and preallocating the
+    /// full limit would penalize ordinary programs. This policy is performance-sensitive and should
+    /// be benchmarked against the fixed-buffer baseline.
+    #[inline(always)]
+    fn ensure_stack_capacity_for_push(&mut self) -> Result<(), ExecutionError> {
```
plafer, line 625: This additional code is run on every push - which our `blake3` benchmark doesn't really exercise (and hence we don't see any difference in performance). I wonder though if we have real-world programs that would execute a fair amount a pushes, and measure the performance impact for those? Did you explore that already? Not sure what the latest state of your "benchmarking the transaction kernel + real world transactions" work is, but would be quite useful here

processor/src/fast/mod.rs
huitseeker: Good catch on making this overflow behavior explicit. I think the exact suggested `<=` would over-grow when `new_len == min_len`, but the safety point stands. I’ll change this.

processor/src/fast/mod.rs
huitseeker: I did not benchmark `Vec` against this version. I kept `Box<[Felt]>` to stay close to the old contiguous indexed storage and avoid changing more than the growth behavior. I agree `Vec::resize` may be worth checking, especially if realloc helps. Do you want it in this PR?

processor/src/fast/mod.rs
huitseeker: Agreed. The allocation cap should not depend on a `high stack_bot_idx`. I’ll change growth to recenter the live stack while growing.

processor/src/fast/mod.rs
huitseeker: Agreed. We only need to preserve live stack storage.

processor/src/fast/mod.rs
huitseeker: I have not measured a stack-heavy real workload yet. But give me a minute ...

processor/src/fast/mod.rs
huitseeker: [Results](https://github.com/huitseeker/miden-vm/tree/bench/operand-stack-growth-comparison), using Criterion’s reported time estimate:

  | Scenario | Box | Vec + resize | Change |
  |---|---:|---:|---:|
  | no growth | 48.70 µs | 47.80 µs | -1.87% |
  | grow 1024 | 56.98 µs | 62.13 µs | +12.32% |
  | grow 4096 | 82.36 µs | 84.93 µs | +4.74% |
  | grow 16384 | 183.08 µs | 194.28 µs | +11.83% |

Points against switching to Vec<Felt> + resize.

processor/src/fast/mod.rs
huitseeker: | Axis | Before avg | After avg | Avg change |
  |---|---:|---:|---:|
  | exec | 6.704 ms | 6.707 ms | +0.08% |
  | trace_prep | 7.276 ms | 7.286 ms | +0.33% |
  | prove | 3770.835 ms | 3852.945 ms | +2.36% |
  | verify | 2.921 ms | 2.898 ms | -0.98% |
  
  The benchmark was introduced here in #3024

processor/src/fast/mod.rs
plafer: Yup, checks out with what I had observed previously - although it is still surprising to me. I don't have a good intuition for why `Vec` growth would be worst than a hand-rolled `Box<[T]>` growth...

processor/src/fast/mod.rs
plafer: Awesome.

processor/src/fast/mod.rs
```diff
@@ -587,6 +596,78 @@ impl FastProcessor {
...
+        while new_len < required_len {
+            let next_len = new_len.saturating_mul(2);
+            if next_len <= new_len {
+                new_len = required_len;
+                break;
+            }
+            new_len = next_len.min(max_len);
+        }
+        debug_assert!(new_len <= max_len);
+
```
plafer, line 673: I've stared at this for a while now and find it difficult to convince myself that it's always doing the right thing. How about we do something like

```rs
fn grow_stack_buffer(&mut self, requested_min_len: usize) {
    let max_len = STACK_BUFFER_BASE_IDX
            .saturating_add(self.options.max_stack_depth())
            .saturating_add(1);
    // or whatever else to ensure the min and max make sense
    debug_assert!(max_len >= requested_min_len);

    let live_len = self.stack_size();
    
    // Doubles the current stack length (after recentering), while respecting the minimum
    // and maximum bounds.
    let new_len = {
        // The length we get just by doubling the current stack size
        let target_len = STACK_BUFFER_BASE_IDX.saturating_add(live_len).saturating_add(2).saturating_mul(2);

        // Don't go below the min or above the max
        target_len.max(requested_min_len).min(max_len)
    };
}
```
Might require some nitty refactoring, but overall is much simpler and I believe does what we want

plafer (approved):
LGTM! Modulo a simplification suggestion

processor/src/fast/mod.rs
huitseeker: This is a bit tighter on allocations, which in turn may require more frequent reallocations. But it's totally valid.

processor/src/execution_options.rs
```diff
@@ -24,6 +25,9 @@ pub struct ExecutionOptions {
     /// Maximum number of continuations allowed on the continuation stack at any point during
     /// execution.
     max_num_continuations: usize,
+    /// Maximum number of field elements allowed on the operand stack in the active execution
+    /// context.
+    max_stack_depth: usize,
 }
```
bobbinth, line 31: Not for this PR, but we should clean this up. We may have issues for this already, but maybe worth consolidating them into a single issue. The things we can do:

- Get rid of `expected_cycles`.
- Potentially consolidate `enable_tracing` and `enable_debugging`.
- Consolidate all advice provider-related settings into one.
- Maybe something else?

bobbinth (approved):
Looks good! Thank you! I left one comment for the future inline.

Also, another option was to hard-code maximum stack depth (rather than making it configurable), but since performance implications seem to be negligible, I think it is better to have it be configurable.

PR #3087, opened 2026-05-01
fix(syntax): align Path::MAX_COMPONENT_LENGTH with maximum Path length

bitwalker:
After updating the Rust toolchain the compiler uses, and enabling LTO to shrink the size of the programs we assemble, we immediately hit an error during assembly that is caused due to some Rust symbols to result in paths where the leaf component exceeds `Path::MAX_COMPONENT_LENGTH`, which previously was set to `255`. Fundamentally, we need to bump the limit just due to that issue alone.

However, I chose here to bump the limit to the maximum valid path length (currently `u16::MAX`), but reserve 2 bytes for the absolute path prefix, giving us a new `MAX_COMPONENT_LENGTH` of `u16::MAX - 2`. This ensures that a single-component path can make full use of the maximum valid path length, where before, even though a path could be `u16::MAX` bytes, we were not allowing individual components to use that space if needed. I think this makes the API a bit more consistent, while addressing the issue we hit in the compiler.

This is targeted against `main` so that we can do a point release (i.e. `0.22.3`) containing this fix, as it is a critical blocker for the compiler. This is not a breaking change, we are simply more permissive than we were previously, and can be trivially merged with `next`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.22.3 (2026-05-01)
+
+- Change value of `Path::MAX_COMPONENT_LENGTH` to `u16::MAX - 2` [#3087](https://github.com/0xMiden/miden-vm/pull/3087)
+
 ## 0.22.2 (2026-04-028)
 
 - Improve debug var loc tracking ([#2955](https://github.com/0xMiden/miden-vm/pull/2955)).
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.22.3 (TBD)
```
bobbinth: Let's use today's date (I'll publish this today).

bobbinth (approved):
Looks good! Thank you!

Could you also bump up the version in `Cargo.toml`? This way, I can publish as soon as we merge this.

PR #3088, opened 2026-05-01
fix: fuzzed deserialization bounds

huitseeker:
This fixes two fuzzing failures in byte-slice deserialization.

`trusted read_from_bytes()` paths could read far more logical bytes than the input slice contains. Malformed inputs could use large length fields to drive very large allocation or read work before failing. The fix wraps `MastForest::read_from_bytes()` and `Package::read_from_bytes()` in a BudgetedReader when reading from a byte slice. 

This also changes package manifest export deserialization to use `read_many_iter()`, so it does not pre-allocate from an untrusted length field.

Regression tests cover the fuzz payloads that triggered the failures.


core/src/mast/serialization/tests.rs
```diff
@@ -913,6 +913,25 @@ fn mast_forest_deserialize_invalid_ops_offset_fails() {
     assert_matches!(result, Err(DeserializationError::InvalidValue(_)));
 }
 
+#[test]
+fn mast_forest_read_from_bytes_rejects_fuzzed_overflow_payload() {
```
plafer, line 917: Would add a comment that explains what goes wrong if this test ever fails in the future - e.g. this payload tries to allocate past a reasonable budget

crates/mast-package/src/package/serialization.rs
```diff
@@ -499,6 +511,28 @@ mod tests {
         assert!(matches!(err, DeserializationError::InvalidValue(_)));
     }
 
+    #[test]
+    fn package_read_from_bytes_rejects_fuzzed_oom_payload() {
```
plafer, line 603: Same here

plafer (approved):
❤️ fuzzing

PR #3089, opened 2026-05-02
ci: extend comparative benches to the Synthetic Tx bench

### LINKED ISSUE #2256: Improve benchmarks

huitseeker:
This extends the per-commit comment introduced in #2992 to the benchmark introduced in #3024, and normalizes the headers of the bench so the comments are easily grepped for (BENCHMARK REPORT is the string).

## Example Comment Output

The example below is generated from a real full synthetic benchmark run, with one metric faked to demonstrate the regression rendering:

```md
# BENCHMARK REPORT: synthetic-tx-kernel-nonregression

## Synthetic Transaction Kernel Non-Regression

Status: **REGRESSION**
Threshold: `5.00%`
Baseline: `1764d66ca1c6`
Current: `1234567890ab`
Worst slowdown: `bench-tx/consume-single-p2id-note/prove` moved by `+145.07 ms` (+7.30%)

| Metric | Baseline | Current | Delta | Delta % |
| --- | ---: | ---: | ---: | ---: |
| bench wall | n/a | n/a | n/a | n/a |

| Benchmark | Baseline | Current | Delta | Delta % |
| --- | ---: | ---: | ---: | ---: |
| bench-tx/consume-single-p2id-note/prove | 1,987.33 ms | 2,132.40 ms | +145.07 ms | +7.30% |
| bench-tx/consume-b2agg-note-bridge-out/exec | 23.96 ms | 23.96 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/prove | 16,246.29 ms | 16,246.29 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/trace_prep | 25.76 ms | 25.76 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/verify | 3.27 ms | 3.27 ms | +0.00 ms | +0.00% |
| bench-tx/consume-claim-note-l1-to-miden/prove | 3,874.57 ms | 3,874.57 ms | +0.00 ms | +0.00% |
| bench-tx/consume-claim-note-l2-to-miden/prove | 3,962.15 ms | 3,962.15 ms | +0.00 ms | +0.00% |
| bench-tx/consume-two-p2id-notes/prove | 3,977.85 ms | 3,977.85 ms | +0.00 ms | +0.00% |
```

## Notes

- The synthetic benchmark script intentionally reads Criterion JSON artifacts,
- The table is sorted by slowdown percentage, so the most relevant regression appears first,
- Missing or newly added metric names are reported separately in the comment body.

Closes #2256

### DIFF
--- a/.github/workflows/synthetic-tx-nonregression.yml
+++ b/.github/workflows/synthetic-tx-nonregression.yml
@@ -65,7 +65,7 @@ jobs:
           rustup toolchain install "${toolchain}" --profile minimal --target wasm32-unknown-unknown
 
       - name: Cache Rust dependencies
-        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        uses: WarpBuilds/rust-cache@9d0cc3090d9c87de74ea67617b246e978735b1a1 # v2.9.1
 
       - name: Prepare benchmark directories
         id: dirs
--- a/.github/workflows/synthetic-tx-nonregression.yml
+++ b/.github/workflows/synthetic-tx-nonregression.yml
@@ -1,0 +1,387 @@
+name: synthetic tx kernel non-regression
+
+on:
+  push:
+    branches:
+      - main
+      - next
+  workflow_dispatch:
+    inputs:
+      baseline_ref:
+        description: Commit or ref to use as the baseline. Defaults to the current ref's first parent.
+        required: false
+        type: string
+      current_ref:
+        description: Commit or ref to benchmark. Defaults to the workflow ref.
+        required: false
+        type: string
+      regression_threshold_pct:
+        description: Allowed slowdown before the workflow fails.
+        required: false
+        default: "5"
+        type: string
+      rayon_num_threads:
+        description: RAYON_NUM_THREADS value used for the benchmark.
+        required: false
+        default: "8"
+        type: string
+      scenario_filter:
+        description: Optional SYNTH_SCENARIO filter for the synthetic benchmark.
+        required: false
+        default: ""
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
+    name: Benchmark baseline vs current
+    runs-on: warp-ubuntu-latest-x64-8x
+    timeout-minutes: 240
+    permissions:
+      contents: read
+    outputs:
+      summary: ${{ steps.summary.outputs.summary }}
+      regression: ${{ steps.compare.outputs.regression }}
+      status: ${{ steps.compare.outputs.status }}
+      baseline_sha: ${{ steps.compare.outputs.baseline_sha }}
+      current_sha: ${{ steps.compare.outputs.current_sha }}
+      max_delta_metric: ${{ steps.compare.outputs.max_delta_metric }}
+      max_delta_pct: ${{ steps.compare.outputs.max_delta_pct }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Install Rust toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+          toolchain="$(sed -n 's/^channel = "\(.*\)"$/\1/p' rust-toolchain.toml)"
+          rustup toolchain install "${toolchain}" --profile minimal --target wasm32-unknown-unknown
+
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+
+      - name: Prepare benchmark directories
+        id: dirs
+        shell: bash
+        run: |
+          set -euo pipefail
+          run_root="${RUNNER_TEMP}/synthetic-tx-nonregression-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          artifact_dir="${run_root}/artifacts"
+          worktree_dir="${run_root}/worktrees"
+          mkdir -p "${artifact_dir}" "${worktree_dir}"
+          {
+            echo "run_root=${run_root}"
+            echo "artifact_dir=${artifact_dir}"
+            echo "worktree_dir=${worktree_dir}"
+          } >> "${GITHUB_OUTPUT}"
+
+      - name: Resolve benchmark refs
+        id: refs
+        env:
+          EVENT_NAME: ${{ github.event_name }}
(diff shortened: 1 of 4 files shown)


.github/workflows/synthetic-tx-nonregression.yml
```diff
@@ -0,0 +1,387 @@
...
+
+      - name: Install Rust toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+          toolchain="$(sed -n 's/^channel = "\(.*\)"$/\1/p' rust-toolchain.toml)"
+          rustup toolchain install "${toolchain}" --profile minimal --target wasm32-unknown-unknown
+
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
```
plafer: Q: wouldn't we want to use WarpBuilds' `rust-cache` here, given that we're using their runner?

scripts/synthetic_tx_nonregression.py
```diff
@@ -0,0 +1,422 @@
...
+    if not separator or not scenario:
+        raise ValueError(f"Unexpected synthetic benchmark group directory: {group}")
+    return producer, scenario
+
+
+def collect_criterion_metrics(repo_root: Path, *, estimate: str = "mean") -> dict[str, dict[str, Any]]:
+    criterion_root = repo_root / "target" / "criterion"
+    metrics: dict[str, dict[str, Any]] = {}
+
+    for estimates_path in sorted(criterion_root.glob("bench-tx_*/*/new/estimates.json")):
```
plafer: From Claude,

1. Criterion path glob doesn't match what Criterion actually writes (critical)

`synthetic_bench.rs:191` constructs the group name with a slash:

```rust
let mut group = c.benchmark_group(format!("{producer_stem}/{scenario_slug}"));
```

Criterion preserves `/` in benchmark IDs as a real directory separator, so on disk you get:

```
target/criterion/<producer_stem>/<scenario_slug>/<axis>/new/estimates.json
```

But `synthetic_tx_nonregression.py:102` looks for two-component, underscore-joined groups:

```python
for estimates_path in sorted(criterion_root.glob("bench-tx_*/*/new/estimates.json")):
    axis = estimates_path.parents[1].name
    producer, scenario = split_criterion_group(estimates_path.parents[2].name)
```

and `split_criterion_group` (`:91`) partitions the directory name on `_`. With Criterion's real layout (`bench-tx/consume-single-p2id-note/prove/new/estimates.json`), the glob won't match anything and `collect_criterion_metrics` will raise `ValueError: No Criterion estimates found...` on every real run, failing the workflow at the "Run baseline benchmark" step.

The unit test at `:360` doesn't catch this because it builds the directory itself in the wrong shape (`bench-tx_consume-single-p2id-note/prove/new/...`).

Fix: change the glob to `bench-tx/*/*/new/estimates.json` (or `*/*/*/new/estimates.json` if you want to support more producers), and pull producer/scenario/axis from `parents[3]/[2]/[1]` instead of partitioning. Update the test to mirror the real layout — ideally point it at a real `target/criterion` from a quick local run so the regression doesn't recur.

plafer (commented):
Lightweight review on my end

.github/workflows/synthetic-tx-nonregression.yml
huitseeker: ofc, thanks for catching!

scripts/synthetic_tx_nonregression.py
huitseeker: Thanks, non-trivial one to catch

Al-Kindi-0 (approved):
Looks great, thank you!

What do you think about the following tweak ?

````md
# BENCHMARK REPORT: synthetic-tx-kernel-nonregression

## Synthetic Transaction Kernel Non-Regression

### Run

- Baseline: `1764d66ca1c6`
- Current: `1234567890ab`
- Threshold: `5.00%`
- Bench wall: 145,000.00 ms -> 152,000.00 ms (+7,000.00 ms, +4.83%)

### Result

- Status: **REGRESSION**
- Worst regression: `bench-tx/consume-single-p2id-note/prove` moved by `+145.07 ms` (+7.30%)

### Metrics over threshold

- `bench-tx/consume-single-p2id-note/prove`: +145.07 ms (+7.30%)

### Per-benchmark results (8 of 8)

| Benchmark | Baseline | Current | Delta | Delta % |
| --- | ---: | ---: | ---: | ---: |
| bench-tx/consume-single-p2id-note/prove | 1,987.33 ms | 2,132.40 ms | +145.07 ms | +7.30% |
| bench-tx/consume-b2agg-note-bridge-out/exec | 23.96 ms | 23.96 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/prove | 16,246.29 ms | 16,246.29 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/trace_prep | 25.76 ms | 25.76 ms | +0.00 ms | +0.00% |
| bench-tx/consume-b2agg-note-bridge-out/verify | 3.27 ms | 3.27 ms | +0.00 ms | +0.00% |
| bench-tx/consume-claim-note-l1-to-miden/prove | 3,874.57 ms | 3,874.57 ms | +0.00 ms | +0.00% |
| bench-tx/consume-claim-note-l2-to-miden/prove | 3,962.15 ms | 3,962.15 ms | +0.00 ms | +0.00% |
| bench-tx/consume-two-p2id-notes/prove | 3,977.85 ms | 3,977.85 ms | +0.00 ms | +0.00% |
````

## Potential follow-up (separate PR)

A finer breakdown for offender scenarios via `tracing-forest`, i.e. `prove > commit to main traces / build_aux_trace / evaluate constraints / open / ...`, mirroring the shape `blake3_nonregression.py::KEY_METRICS` already uses. 
Once we have this, we can get rid of the Blake benchmarks, finally.

bobbinth (approved):
A very superficial review from me - but looks good! Thank you!

PR #3091, opened 2026-05-04
Fix assembly privacy checks and build provenance inputs

### LINKED ISSUE #2712: Importing non-`pub` constant works unexpectedly
### Packages versions

Latest `next`: commit `6b6e8b836`.

### Bug description

Importing a non-`pub` constant from a module works, but I expected it to fail since it is not marked `pub`.

```rust
#[test]
fn import_non_pub_constant() -> TestResult {
    let mut context = TestContext::default();
    let module0 = NamedSource::new("test::module", "const ABC = 5");
    let source = source_file!(
        &context,
        "
    use test::module::ABC

    begin
      push.ABC
    end
    "
    );

    context.add_module(module0)?;
    context.assemble(source)?;

    Ok(())
}
```

### Relevant log output

```shell

```

### LINKED ISSUE #2897: Type import in the transaction kernel results in an unexpected error
Attempt to use the imported type inside the `$kernel` results in the `unused import` error. For more details see this comment: https://github.com/0xMiden/protocol/pull/2644#discussion_r2967886275

_Originally posted by @Fumuran in https://github.com/0xMiden/protocol/pull/2644#discussion_r2967886275_

### LINKED ISSUE #2945: Build provenance includes manifest details that don't affect the build
_Extracted from [a comment](https://github.com/0xMiden/miden-vm/pull/2877#discussion_r3009615766) by @huitseeker in #2877._

The computation of build provenance for a source package currently includes the entire effective manifest, not just the selected target, profile, and dependencies. This results in the provenance hash changing even if an unrelated target or profile is modified, e.g. adding a new `[[bin]]` target would be enough to change the provenance for the `[lib]` target. This causes package reuse to fail, even when the artifact would be bit-for-bit identical.

We need to do two things here:

1. Ensure that provenance hashing only considers manifest properties which affect a specific build target and profile.
2. Ensure that we have some mechanism to catch when a manifest property _should_ be included in provenance, but isn't. Ideally, we'd get a compilation error if we haven't specified whether a given property of the manifest either _is_ provenance-affecting, or isn't.

NOTE: This isn't a bug per se - rather we're not reusing package artifacts in these cases, when we should be, and rebuilding them from source instead. However, failing to reuse artifacts in these cases may cause other problems, so I'm treating this as a bug rather than an enhancement/refactoring.

huitseeker:
This closes #2945, closes #2712, and closes #2897.

Some assembly imports were resolved without checking whether the target item was public, and signature-only type uses were skipped by the import visitor. Build reuse also depended on manifest data that should not affect compiled output.

This PR:
- rejects private cross-module imports, including absolute-path aliases
- counts imports used only in procedure signatures
- narrows build provenance to fields that affect output
- marks required core STARK helper procedures public
- updates generated core library docs for those public procedures

Design decisions that could have gone another way:
1. Build provenance means target-specific build identity: does not mean the full effective manifest hash. Only fields that can affect the selected target’s artifact should trigger rebuild/reuse changes.
2. The project model owns provenance classification: Assembly shouldn't hash rendered TOML or infer meaning from serialization. 
3. Unrelated manifest data should not affect artifact reuse: Adding or changing another target should not rebuild a library artifact unless it can affect that library’s package output.
4. Module visibility applies to all item kinds: Private procedures, constants, and types are all private across module boundaries. Constants are not exempt just because they are compile-time values. **This is why this PR includes some masm changes**.
5. Signature-only type references count as import usage: A type used only in a proc argument or result is still semantically used, so its import must be retained / validated.


bitwalker:
> Design decisions that could have gone another way:
> 
> 1. Build provenance means target-specific build identity: does not mean the full effective manifest hash. Only fields that can affect the selected target’s artifact should trigger rebuild/reuse changes.
> 2. The project model owns provenance classification: Assembly shouldn't hash rendered TOML or infer meaning from serialization.
> 3. Unrelated manifest data should not affect artifact reuse: Adding or changing another target should not rebuild a library artifact unless it can affect that library’s package output.
> 4. Module visibility applies to all item kinds: Private procedures, constants, and types are all private across module boundaries. Constants are not exempt just because they are compile-time values. **This is why this PR includes some masm changes**.
> 5. Signature-only type references count as import usage: A type used only in a proc argument or result is still semantically used, so its import must be retained / validated.

All those seem fine to me.

I think the main thing missing here is that we must reject the use of private type declarations in the signature of exported procedures, since it would cause dangling references in the final package in some cases (namely enum types). The type system is structural, not nominal, so most type aliases are transparent (i.e. a private type alias could appear in an exported signature and it wouldn't have an impact on the type signature encoded in the final package) - but for the purpose of binding generation in Rust/other languages, we will probably move towards preserving at least some names in the final artifact, so rejecting that as an edge case will make that doable without causing breakages at some point.

bitwalker (approved):

huitseeker:
@bitwalker Right on, I opened #3092.

miden-vm/tests/integration/cli/cli_test.rs
Al-Kindi-0, line 1: Q: What is the intention of this test ?

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 - Added `Package::content_digest()` to identify package contents without changing the MAST digest, including manifest data and semantic package metadata ([#2909](https://github.com/0xMiden/miden-vm/pull/2909)).
 - Hardened MAST forest and package byte-slice deserialization against fuzzed length fields ([#3088](https://github.com/0xMiden/miden-vm/pull/3088)).
 - [BREAKING] Bounded the live advice map by total field elements during execution; advice-provider setup now returns an error when initial advice exceeds this limit ([#3085](https://github.com/0xMiden/miden-vm/pull/3085)).
+- Fixed project artifact reuse to ignore unrelated manifest fields, rejected private cross-module imports, and kept signature-only type imports live ([#3091](https://github.com/0xMiden/miden-vm/pull/3091)).
```
Al-Kindi-0: Should this be breaking ?

Al-Kindi-0 (approved):
LGTM, modulo a couple of questions

miden-vm/tests/integration/cli/cli_test.rs
huitseeker: The test checks that `bundle --kernel` can still produce a package when the bundled library has no exports. After the new privacy rule, the old fixture was invalid because its kernel called a private proc from `lib_noexports`. The new `kernel_noexports.masm` keeps the same test intent without relying on a private cross-module call.

CHANGELOG.md
huitseeker: Yep, thanks!

PR #3093, opened 2026-05-05
Fix local registry persistence

### LINKED ISSUE #2946: Assembly can fail when building a dependency package from source fails to publish

huitseeker:
Fix local registry artifact storage and cache repair.

Artifacts are now stored with package-qualified paths, while legacy digest-only artifacts still load when they match the requested package.

Preassembled dependency artifacts are cached instead of published. This avoids requiring registry entries for dependency artifacts.

Cache and publish writes now reject stale or conflicting artifacts before replacing files.

Closes #2946.

### DIFF
--- a/crates/assembly/src/project.rs
+++ b/crates/assembly/src/project.rs
@@ -11,7 +11,7 @@ use miden_assembly_syntax::{
 };
 use miden_core::serde::{Deserializable, Serializable};
 use miden_mast_package::{Package as MastPackage, Section, SectionId, TargetType};
-use miden_package_registry::{PackageId, PackageStore, Version as PackageVersion};
+use miden_package_registry::{PackageCache, PackageId, Version as PackageVersion};
 use miden_project::{
     Linkage, Package as ProjectPackage, PreassembledDependencyMetadata, Profile,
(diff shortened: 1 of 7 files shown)


bitwalker (approved):
Nice, thanks for tackling those fixes! LGTM

PR #3045, opened 2026-04-22
feat(hashing): use domain-separated leaf hashing in SMT

gswirski:
Update SMT hashing to follow: https://github.com/0xMiden/miden-vm/issues/2996#issuecomment-4288861592 and https://github.com/0xMiden/crypto/pull/962 implementation.

Closes #2996

PR #3095, opened 2026-05-05
Update miden-vm to crypto 0.25

huitseeker:
Updates miden to miden-crypto v0.25. Due to the breaking changes in SMT hashing, subsumes and includes #3045.


crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -469,6 +511,31 @@ pub proc merge
     hmerge
 end
 
+#! Merges two words (256-bit digests) via Poseidon2 hash with a domain identifier.
+#!
+#! Inputs:  [domain, B, A, ...]
```
Al-Kindi-0: Q: Shouldn't this be `[domain, A, B, ...]` following our LE convention ?

Al-Kindi-0 (approved):
LGTM, modulo a question

plafer (approved):
LGTM

crates/lib/core/asm/crypto/hashes/poseidon2.masm
huitseeker: Indeed, thank you so much for catching this!

PR #3100, opened 2026-05-06
refactor(precompile): linear hash tree transcript

adr1anh:
## Summary

Replace the sponge-with-finalization design of `log_precompile` with a domain-separated linear hash tree:

- `STMNT       = Permute(COMM_0, COMM_1, TAG).rate0` — TAG enters the sponge as initial capacity, acting as a domain separator.
- `state_{i+1} = Permute(state_i, STMNT_i, ZERO).rate0` — 2-to-1 hash fold. The state is itself a complete digest at every step, so `PrecompileTranscript::finalize()` is gone.
- `PrecompileCommitment` is now `(TAG, COMM_0, COMM_1)` — the verifier emits the two halves separately instead of pre-merging them. Drops one `Poseidon2::merge` from each crypto verifier.

### New stack semantics

```
Before: [JUNK, JUNK, STMNT, ...]
After:  [STATE_NEW, IGNORE, IGNORE, ...]
```

The wrapper `sys::log_precompile_request` is now `padw padw + log_precompile + dropw dropw dropw`: the caller leaves a single precomputed `STMNT` word on top, the wrapper pads two zero words below it without any reordering, and the opcode writes its output via the identity lane→slot mapping (matching `HPERM`).

### Touch points

- `core/src/precompile.rs` — struct rebuild, transcript algorithm, `.statement()` accessor; drop `finalize()` and `PrecompileTranscriptDigest`.
- `crates/lib/core/src/handlers/{sha512,keccak256,eddsa_ed25519,ecdsa}.rs` — return `(TAG, COMM_0, COMM_1)`; drop the final 2-to-1 merge.
- `processor/src/execution/operations/crypto_ops/mod.rs` — rebuild `op_log_precompile`; rename `cap_prev` to `state_prev` everywhere.
- `air/src/trace/mod.rs::log_precompile` — new helper/stack ranges for the new layout.
- `air/src/constraints/lookup/{buses/chiplet_requests, buses/block_stack_and_range_logcap, messages, miden_air}.rs` — rename `LogCapacityMsg` → `LogStateMsg`; rebuild bus payloads.
- `crates/lib/core/asm/sys/mod.masm` + crypto MASM files — wrapper rewrite; precompile files now derive `STMNT` via `hperm` with TAG in the initial capacity.
- `verifier/src/lib.rs` — drop `finalize()` call; return the rolling state directly.
- Snapshots for `relation_digest_matches_current_air` and `masm_consistency::case_26` regenerated; design docs updated.

### Follow-up

The hasher chiplet does not currently expose a dedicated 2-to-1 interaction. We use `linear_hash_init` + `return_state` (the same pair HPERM uses), which carries a 12-lane payload on each side even though we only consume `rate0` of the output. A future PR could add `BusId::HasherMergeInit` / pair with `return_hash` to shave 12 elements off the bus payload (and let `STATE_PREV` come from the natural rate0 stack slot), but that requires a new chiplet variant and is left as a separate change.

## Test plan

- [x] `make check` (workspace + tests)
- [x] `make regenerate-constraints` regenerates without diff after rerun
- [x] `make lint` clean
- [x] `make doc` clean
- [x] `make test` passes 3140/3140 (1 leaky pre-existing)
- [x] `make test-fast` passes 3026/3026


huitseeker (approved):
This LGTM!

processor/src/execution/operations/crypto_ops/mod.rs
```diff
@@ -448,63 +446,46 @@ pub(super) fn op_horner_eval_ext<P: Processor, T: Tracer>(
...
-    // Update the processor's precompile sponge capacity
-    processor.set_precompile_transcript_state(cap_next);
-
-    // Write the output to the stack (top 12 elements): [R0, R1, CAP_NEXT, ...].
-    processor.stack_mut().set_word(0, &r0);
-    processor.stack_mut().set_word(4, &r1);
-    processor.stack_mut().set_word(8, &cap_next);
+    let state_new: Word = output_state[Hasher::RATE0_RANGE].try_into().unwrap();
+    let out_rate1: Word = output_state[Hasher::RATE1_RANGE].try_into().unwrap();
+    let out_cap: Word = output_state[Hasher::CAPACITY_RANGE].try_into().unwrap();
```
Nashtare, line 478: nit: minor but I tend to favor `expect()` over `unwrap()`, as it was before the refactor

Nashtare: Same applies in other places as well so perhaps worth a larger refactor if we'd go down that route

Nashtare (approved):
LGTM, thanks!

PR #3101, opened 2026-05-06
perf(processor): replace BTreeMap with HashMap in FastProcessor memory

### LINKED ISSUE #2819: Investigate using a `HashMap` instead of `BTreeMap` for the `FastProcessor`'s memory
We currently [use a `BTreeMap`](https://github.com/0xMiden/miden-vm/blob/f01ceacc1afa65e6d5863cc32308bd791362e260/processor/src/fast/memory.rs#L24) as the backing store for the processor memory. We should investigate whether using a `HashMap` improves the processor's performance on memory-heavy programs.

AlgofootPrint:
## Summary

  Closes #2819.

  The `FastProcessor`'s memory backing store was a `BTreeMap<(ContextId, u32), Word>`. Since memory operations
  (read/write element, read/write word) are pure lookups and inserts with no ordering requirement, `HashMap` is a better
   fit: O(1) average vs O(log n).

  ## Changes

  - Replaced `BTreeMap` with `hashbrown::HashMap` in `processor/src/fast/memory.rs`
  - Added `Hash` to `ContextId` (required for the HashMap key)
  - Added `hashbrown` as a workspace dependency (`default-features = false, features = ["ahash"]`) for `no_std`
  compatibility
  - `get_memory_state` now sorts its output explicitly by address — previously free via BTreeMap, needed to keep debug
  output and test snapshots deterministic

  ## Note on `get_memory_state`

  This method is only called by the debug printer and test utilities, not on any hot path, so the sort has no
  performance impact on normal execution.


AlgofootPrint:
Benchmark results (memory_heavy — 16 384 sequential word writes + reads, measured with Criterion on WSL/Ubuntu):
  
 | Implementation      | Median time |
 |---------------------|-------------|
 | BTreeMap (baseline) | 1.85 ms     |
 | HashMap (this PR)   | 1.83 ms     |

Change: −2.2% (p = 0.50, not statistically significant for sequential access).
 
Why sequential access shows minimal difference: both structures handle sequential addresses cache-efficiently.
BTreeMap traversals stay mostly in cache for in-order access, while HashMap avoids comparison chains but adds hashing. The algorithmic improvement (O(log n) → O(1) amortized) becomes measurable with random-access patterns or much larger address spaces where BTreeMap's tree traversals cause cache misses.
 
Benchmark program added at miden-vm/masm-examples/memory_heavy.masm and is picked up automatically by the program_execution_fast bench suite.

Cargo.toml
```diff
@@ -78,6 +78,7 @@ midenc-hir-type  = { version = "0.6.1", default-features = false }
 
 # Serialization
 bincode = "1.3"
+hashbrown       = { version = "0.15", default-features = false, features = ["alloc", "default-hasher"] }
```
bobbinth: nit: this should go into the "Third-party crates" section below.

But also, in `miden-crypto`, we went with a slightly different approach: we used standard `BTreeMap` in `no_std` context and standard `HashMap` in the `std` context (the standard `HashMap` is just re-export from `hashbrown`, I believe). What are the main benefits of going with this approach?

AlgofootPrint: Good catch on both points. The main benefit of the direct hashbrown dependency is that no_std + alloc builds would also get HashMap instead of BTreeMap but that's an edge case since the processor is almost never used in no_std context in practice. There isn't a strong enough reason to justify the extra dependency. I'll switch to the miden-crypto pattern (BTreeMap in no_std, std::collections::HashMap in std via #[cfg(feature = "std")]) and remove the explicit hashbrown dep entirely.

AlgofootPrint:
All Done @bobbinth switched to the #[cfg(feature = "std")] / #[cfg(not(feature = "std"))] type alias approach and removed the hashbrown dependency entirely. The std build uses std::collections::HashMap and no_std falls back to alloc::collections::BTreeMap, consistent with how miden-crypto does it.

miden-vm/masm-examples/memory_heavy.masm
```diff
@@ -0,0 +1,23 @@
+# Memory-heavy benchmark: 16384 sequential word writes followed by 16384 reads.
```
huitseeker: Nit: Could these new `#` comments start lowercase to match the MASM style? Single-`#` comments are expected to start lowercase. Though this is a rule that's not really followed well

huitseeker: This new top-level example is skipped by `make run-examples`, which only matches `miden-vm/masm-examples/*/*.masm` and `miden-vm/masm-examples/*/*/*.masm`. Could this live under a subdir, or should the Makefile glob include top-level examples too, so CI actually runs it?

AlgofootPrint: Fixed, all single-# comments now start lowercase.

AlgofootPrint: Good catch. Moved it into masm-examples/memory_heavy/memory_heavy.masm so the existing glob picks it up no Makefile change needed.

plafer:
What I'm seeing here is that a memory-heavy benchmark shows a ~1% "improvement" (i.e. within noise levels), and running the synthetic benchmark locally (i.e. p2id note) shows there's no improvement.

I personally wouldn't bother with the complexity and close this PR (though it was good to confirm that there was no improvement!).

AlgofootPrint:
Fair point the data doesn't support the change. Happy to close this. The one thing worth keeping might be the memory_heavy benchmark as a baseline for future memory-related work, but I'll defer to you and @bobbinth on whether that's worth a standalone PR or not.

miden-vm/masm-examples/memory_heavy.masm
bitwalker: Is that really a convention, or I guess more to the point, one we actually want? I'd expect to go the other way if anything (i.e. require comments to be proper english) - all lowercase is bizarre to me

bobbinth:
No strong preferences from my end. On the one hand, the performance delta is not meaningful, but on the other the additional complexity is rather minimal (like 10 lines of pretty trivial code). So, I could go either way.

miden-vm/masm-examples/memory_heavy.masm
AlgofootPrint: I followed @huitseeker's suggestion to go lowercase, but happy to revert to capitalized if that's the preferred direction. Let me know what you'd like and I'll update.

AlgofootPrint:
Given the minimal complexity, I'm happy to keep it open if there's no objection from @plafer. Otherwise I'll close no strong preference either way.

plafer:
I stand by my not seeing the value in this PR - so @huitseeker you can be the tie-breaker.

plafer:
Closing per previous conversation. @huitseeker Please re-open if you deem this necessary.

PR #3102, opened 2026-05-06
Harden program serde entrypoint invariants

vh0rvath:
## Describe your changes
                                                                                                                                                                                                     
  This PR hardens Program serde deserialization so it enforces the same entrypoint invariants as the binary deserialization path, and expands fuzz coverage to exercise post-deserialization Program usage paths.                                                                                                                                                                                       
                                                                                                                                                                                                     
  ### What changed                                                                                                                                                                                   
                                                                                                                                                                                                     
  - Replaced derived Deserialize for Program with a manual implementation in core/src/program/mod.rs.                                                                                                
  - Added invariant checks during serde deserialization:                                                                                                                                             
      - entrypoint must exist in mast_forest.                                                                                                                                                        
      - entrypoint must be a procedure root.                                                                                                                                                         
  - Return structured serde errors instead of allowing invalid Program instances to be created.                                                                                                      
  - Expanded miden-core-fuzz/fuzz_targets/program_serde_deserialize.rs:                                                                                                                              
      - After successful deserialize, call hash() and to_info() on:                                                                                                                                  
          - Program                                                                                                                                                                                  
          - Vec<Program> (bounded iteration)                                                                                                                                                         
          - Option<Program>                                                                                                                                                                          
                                                                                                                                                                                                     
  ### Why                                                                                                                                                                                            
                                                                                                                                                                                                     
  Derived serde deserialization could bypass Program construction invariants and allow malformed objects, which could later panic on normal access paths. This change keeps invariants consistent across deserialization surfaces and improves robustness.                                                                                                                                           
                                                                                                                                                                                                     
  ### Related                                                                                                                                                                                        
                                                                                                                                                                                                     
  - Context/reference: #2849 (https://github.com/0xMiden/miden-vm/pull/2849)                                                                                                                         
                                                                                                                                                                                                     
  ———                                                                                                                                                                                                

  - [ ] Repo forked and branch created from next according to naming convention.                                                                                                                     
  - [ ] Commit messages and codestyle follow ../CONTRIBUTING.md.                                                                                                                                     
  - [ ] Commits are signed (https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).                                                                       
  - [ ] Tests added for new functionality.                                                                                                                                                           
  - [x] Updated CHANGELOG.md.

### DIFF
--- a/core/src/program/mod.rs
+++ b/core/src/program/mod.rs
@@ -27,7 +27,7 @@ pub use stack::{InputError, MIN_STACK_DEPTH, OutputError, StackInputs, StackOutp
 /// execution begins, and a definition of the kernel against which the program must be executed
 /// (the kernel can be an empty kernel).
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
 #[cfg_attr(
     all(feature = "arbitrary", test),
     miden_test_serde_macros::serde_test(binary_serde(true))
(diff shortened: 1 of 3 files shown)


bitwalker:
Just FYI, I'm in the process of removing `Library` and `Program` at the moment - we should probably hold off on this until that is complete.

core/src/program/mod.rs
```diff
@@ -187,6 +187,36 @@ impl Deserializable for Program {
...
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for Program {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let ProgramSerde { mast_forest, entrypoint, kernel } =
+            ProgramSerde::deserialize(deserializer)?;
+
+        if mast_forest.get_node_by_id(entrypoint).is_none() {
```
huitseeker, line 207: Could we add normal serde tests for these two rejection paths too? `Program` already has the `serde_test(binary_serde(true))` macro for valid round trips, and the fuzz target may catch panics, but normal CI still does not pin the malformed cases this change is fixing: an entrypoint outside the forest, and an entrypoint that exists but is not a procedure root.

huitseeker (commented):
Thanks! While we removed Library in #3106, this brings a (narrow) hardening benefit for `core::program::Program`

bitwalker:
@huitseeker We don't serialize programs to disk at all as of the change to remove `Library`, so we don't need the serialization code for `Program` anymore

huitseeker:
As mentioned above, this code for Program is obsolete

Issue #1684, opened 2025-03-05
Investigate switching from `while` loops to `do-while` loops


plafer:
> Does `Operation::Loop` allow for do/while-form loops rather than while/do loops? That would be super useful if so, but my understanding is we only have condition-guarded loops, which is a bit unfortunate as that's the opposite of how loops are canonically represented. In fact, the compiler has to introduce a redundant `if.true` nested within `while.true` to lower its canonicalized loop structure into MASM. 

 _Originally posted by @bitwalker in [#1677](https://github.com/0xPolygonMiden/miden-vm/issues/1677#issuecomment-2696197803)_

To reiterate, the question is whether we want to have our loops run the first iteration unconditionally. The above depends on if we can make the constraints for it work. It would probably require us to change the MASM syntax from `while.true ... end` to `do ... while.true` or something.

bitwalker:
@bobbinth Is this something we could do prior to mainnet? In MASM itself it could be done as a non-breaking change (since the existing loop syntax can be preserved, we can just add a new form for tail-controlled loops), so it really boils down to whether we can/should encode this in MAST as a different node type, or rework the existing one. 

To give you some more detail around this, loops are canonically structured as follows:

* One or more _entering_ blocks, which are predecessor blocks outside the loop, from which control may reach the loop _header_. If there is only a single entering block, and that block only has an edge to the header block, then the entering block in that case is also referred to as the loop _preheader_.
* The _header_ block, which dominates all blocks comprising the loop body.
* A _latch_ block, which is a block within the loop that has an edge to the header block, also called a _backedge_.
* One or more _exiting_ blocks, which are blocks inside the loop with an edge to a block outside the loop. The target of an exiting block is an _exit_ block.

This image visualizes those terms and their relationships:

![loop terminology](https://llvm.org/docs/_images/loop-terminology.svg)

Now, the reason why this is is important, is because in any case where a compiler can prove that a loop executes at least once, it is very common for them to perform _loop rotation_, which converts head-controlled loops (e.g. `while (...) { }`) into tail-controlled loops (e.g. `do { } while(...)`). LLVM in particular does this. Which means that the Miden compiler currently has to "unrotate" these loops. To give a simple example, consider the following loop in Rust:

```rust
fn test(n: NonZeroUsize) {
    for i in 0..n.get() {
        ...
    }
}
```

This gets lowered to something akin to the following C code:

```c
void test(int n) {
    int i = 0;
    do {
        ...
        i++;
[...]

PR #3105, opened 2026-05-06
Introduce `SparseMastForest`

### LINKED ISSUE #2766: Introduce `SparseMastForest`
The current problem with serializing `TraceGenerationContext` (needed for #2730) is that it holds references to all the `Arc<MastForest>` that were visited during program execution. Specifically, any `MastForest` representing a library is likely to not be fully visited - that is, likely not all library procedures are going to be visited during a given program execution.

Hence, a solution to this problem is to build a `SparseMastForest` (or alternatively `MastForestSubset`) in the `ExecutionTracer` of all nodes visited, per `MastForest`. The `TraceGenerationContext` would then have a new `mast_forest_store` field, which stores all the sparse MAST forests encountered during execution.

The `SparseMastForest` would look something like

```rs
pub struct SparseMastForest {
    // Only difference with MastForest
    nodes: BTreeMap<MastNodeId, MastNode>,


    roots: Vec<MastNodeId>,
    advice_map: AdviceMap,
    debug_info: DebugInfo,
    commitment_cache: OnceLockCompat<Word>,
}
```

An important property of the `SparseMastForest` is that it maintains the `MastNodeId`s of the original `MastForest` - so that when we re-execute the program, the existing IDs refer to the right `MastNode`s.

### Serializing `SparseMastForest` using the `MastForest` serialization format

Ideally, we would reuse the `MastForest` serialization format to support sparse MAST forests, rather than maintain 2 separate formats.

I believe the only change needed is adding a `is_dense` flag, which when set corresponds to the current `MastForest` serialization format. However, when not set, then we would have to encode the list of `MastNodeId`s that are present in the forest somewhere in the format; I'm seeing the corresponding `MastNodeInfo`s being encoded as they are today.

When deserializing, it would be an error to deserialize a sparse `MastForest` (i.e. `is_dense` flag not set) into a `MastForest`.

### Execution

Now that we have 2 types representing MAST forests, we would need to make `execute_impl()` generic over some `MastForestForExecution` trait - which I believe only needs to have a `get_node_by_id()` method.

plafer:
Partially closes #2766 (`TraceGenerationContext` serialization not implemented yet)

## Motivation

`TraceGenerationContext` previously held `Arc<MastForest>`s to refer to `MastForest`s, making serialization impractical since most nodes in those forests are not visited, and hence serialized needlessly.

## What changed (high level)

The core line of changes is
- introduce a `SparseMastForest` type (pretty much as described in #2766), 
- In the `ExecutionTracer`, for each `MastForest` accessed during execution, build a `SparseMastForest` containing all the nodes actually accessed during execution,
- Update the execution pipeline to be generic over `MastForest` (where the methods needed during execution are encoded in a new `ExecutableMastForest` trait).

## Tradeoffs

The main tradeoff that we're making here is that we're increasing memory usage for the use case where the trace is built on the same machine that generated the `TraceGenerationContext`, since we're now building `SparseMastForest`s, where just reusing `Arc<MastForest>` would have sufficed (as it did previously). Supporting both modes (using `Arc<MastForest>` for a non-serializable `TraceGenerationContext`, and `SparseMastForest`s for a serializable `TraceGenerationContext`) would require more thought.

### New types in `miden-core`

- **`SparseMastForest`** — a subset of a source `MastForest` containing only the nodes visited during execution. Uses `BTreeMap<MastNodeId, MastNode>` so the original `MastNodeId`s are preserved, allowing the sparse forest to drop into the executor as if it were the dense source.
- **`SparseMastForestBuilder`** — accumulator that records visited `MastNodeId`s and finalizes into a `SparseMastForest`.
- **`ExecutableMastForest`** trait — the API the executor and decorator pipeline need from a forest (`get_node_by_id`, `find_procedure_root`, `decorator_by_id`, `linked_before_enter_decorators`, `linked_after_exit_decorators`, `decorator_indices_for_op`, `advice_map`, `get_assembly_op`, `resolve_error_message`, `Index<MastNodeId>` and `Index<DecoratorId>` as supertraits). Implemented for `MastForest`, `SparseMastForest`, and `Arc<T: ExecutableMastForest>`.
- **`MastForestId`** newtype — opaque handle into a forest store.

### Executor pipeline generic over `ExecutableMastForest`

The whole executor + tracer surface is now generic over `F: ExecutableMastForest`:

- `Continuation<F>`, `ContinuationStack<F>` (only `EnterForest(F)` carries the type)
- `BreakReason<F>`, `InternalBreakReason<F>`
- `Tracer { type Forest: ExecutableMastForest + Clone; ... }`, `Stopper { type Forest; ... }`
- `Processor`'s decorator methods made generic over `F`
- `execute_impl<P, S, T, F>` and every inner exec function
- `MapExecErr`/`MapExecErrWithOpIdx` traits + their impls + free helpers (`with_context`, `advice_error_with_context`, etc.) take a generic forest
- `MastNodeExt::{before_enter, after_exit, verify_node_in_forest}` made generic (derive macro updated)

### `TraceGenerationContext` changes

- `TraceGenerationContext` gains a `mast_forest_store: Vec<Arc<SparseMastForest>>`.
- Fragment-internal `Arc<MastForest>` references are replaced by `MastForestId` indices: `CoreTraceFragmentContext.{continuation: ContinuationStack<MastForestId>, initial_mast_forest_id}`, `MastForestResolutionReplay`, and `HasherOp::HashBasicBlock`.
- `ExecutionTracer` carries per-source-forest `SparseMastForestBuilder`s keyed by forest commitment, populated in `start_clock_cycle`, `record_mast_forest_resolution`, and a new `record_external_node_entered` hook (external nodes don't drive a clock cycle so they need a separate hook). When a node is recorded, its immediate children are also recorded — needed so the trace generator can read child digests on `Join`/`Split`/`Loop`/`Call` rows even when only one child is taken.
- Replay translates `MastForestId` → `Arc<SparseMastForest>` at fragment setup (`split_trace_fragment_context` returns `Result`, validating attacker-controlled ids rather than panic-indexing).

## Validation

Trace generation now uses attacker-controlled `MastForestId`s to access `SparseMastForest`s; added a test to ensure that this properly returns an error for invalid `MastForestId`s.

## Out of scope

- Serialization of `TraceGenerationContext` — this PR makes that tractable but doesn't implement it.
- Trimming `roots`, `advice_map`, and `debug_info` of `SparseMastForest` to only the visited subset.

### DIFF
--- a/core/src/mast/mod.rs
+++ b/core/src/mast/mod.rs
@@ -109,6 +109,9 @@ pub use node_fingerprint::{DecoratorFingerprint, MastNodeFingerprint};
 mod node_builder_utils;
 pub use node_builder_utils::build_node_with_remapped_ids;
 
+mod sparse;
+pub use sparse::{MastForestId, SparseMastForest, SparseMastForestBuilder, VisitKind};
+
 #[cfg(test)]
 mod tests;
 
@@ -997,6 +1000,206 @@ impl IndexMut<DecoratorId> for MastForest {
     #[inline(always)]
     fn index_mut(&mut self, decorator_id: DecoratorId) -> &mut Self::Output {
         self.debug_info.decorator_mut(decorator_id).expect("DecoratorId out of bounds")
+    }
+}
+
+// EXECUTABLE MAST FOREST
+// ================================================================================================
+
+/// A MAST forest that can be used as the source of nodes during program execution.
+///
+/// Implemented by both [`MastForest`] (a dense forest containing all nodes) and
+/// [`SparseMastForest`] (a sparse subset of a forest containing only the nodes visited during
+/// some prior execution). The latter preserves the original [`MastNodeId`]s of its source forest,
+/// which allows it to stand in for the dense forest during re-execution.
+pub trait ExecutableMastForest: Index<DecoratorId, Output = Decorator> {
+    /// Returns the [`MastNode`] associated with the provided [`MastNodeId`] if present, or else
+    /// `None`.
+    fn get_node_by_id(&self, node_id: MastNodeId) -> Option<&MastNode>;
+
+    /// Returns the digest of the node associated with the provided [`MastNodeId`] if present, or
+    /// else `None`.
+    ///
+    /// For dense forests this is equivalent to `get_node_by_id(id).map(|n| n.digest())`. For
+    /// [`SparseMastForest`], it additionally consults the digest-only entries — nodes that were
+    /// referenced (but not entered) during execution and which were therefore stored as digest
+    /// only. Use this method whenever only the digest of a referenced node is needed (e.g. when
+    /// populating the hasher state of a parent's trace row).
+    fn get_digest_by_id(&self, node_id: MastNodeId) -> Option<Word>;
+
+    /// Returns the [`MastNodeId`] of the procedure associated with a given digest, if any.
+    fn find_procedure_root(&self, digest: Word) -> Option<MastNodeId>;
+
+    /// Returns the [`Decorator`] associated with the provided [`DecoratorId`] if present, or else
+    /// `None`.
+    fn decorator_by_id(&self, decorator_id: DecoratorId) -> Option<&Decorator>;
+
+    /// Returns the advice map associated with this forest.
+    fn advice_map(&self) -> &AdviceMap;
+
+    /// Returns the decorators that should be executed before entering the node with the provided
+    /// id, looked up via the forest's debug info (used for nodes whose decorator store is in the
+    /// `Linked` form).
+    fn linked_before_enter_decorators(&self, node_id: MastNodeId) -> &[DecoratorId];
+
+    /// Returns the decorators that should be executed after exiting the node with the provided
+    /// id, looked up via the forest's debug info (used for nodes whose decorator store is in the
+    /// `Linked` form).
+    fn linked_after_exit_decorators(&self, node_id: MastNodeId) -> &[DecoratorId];
+
+    /// Returns decorator indices for a specific operation within a basic block node.
+    fn decorator_indices_for_op(&self, node_id: MastNodeId, local_op_idx: usize) -> &[DecoratorId];
+
+    /// Returns the [`AssemblyOp`] associated with a node (and optionally an operation index inside
+    /// a basic block), if any.
+    ///
+    /// This is used by the error-reporting machinery to produce source locations.
+    fn get_assembly_op(
+        &self,
+        node_id: MastNodeId,
+        target_op_idx: Option<usize>,
+    ) -> Option<&AssemblyOp>;
+
+    /// Resolves an error code to its corresponding error message, if any.
+    fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>>;
+}
+
+impl ExecutableMastForest for MastForest {
+    #[inline(always)]
+    fn get_node_by_id(&self, node_id: MastNodeId) -> Option<&MastNode> {
+        MastForest::get_node_by_id(self, node_id)
+    }
+
+    #[inline(always)]
+    fn get_digest_by_id(&self, node_id: MastNodeId) -> Option<Word> {
+        MastForest::get_node_by_id(self, node_id).map(MastNodeExt::digest)
+    }
+
+    #[inline(always)]
+    fn find_procedure_root(&self, digest: Word) -> Option<MastNodeId> {
+        MastForest::find_procedure_root(self, digest)
+    }
+
+    #[inline(always)]
+    fn decorator_by_id(&self, decorator_id: DecoratorId) -> Option<&Decorator> {
+        MastForest::decorator_by_id(self, decorator_id)
+    }
+
+    #[inline(always)]
+    fn advice_map(&self) -> &AdviceMap {
+        MastForest::advice_map(self)
+    }
+
+    #[inline(always)]
+    fn linked_before_enter_decorators(&self, node_id: MastNodeId) -> &[DecoratorId] {
+        MastForest::before_enter_decorators(self, node_id)
+    }
+
+    #[inline(always)]
+    fn linked_after_exit_decorators(&self, node_id: MastNodeId) -> &[DecoratorId] {
+        MastForest::after_exit_decorators(self, node_id)
+    }
+
+    #[inline(always)]
+    fn decorator_indices_for_op(&self, node_id: MastNodeId, local_op_idx: usize) -> &[DecoratorId] {
+        MastForest::decorator_indices_for_op(self, node_id, local_op_idx)
+    }
+
+    #[inline(always)]
+    fn get_assembly_op(
+        &self,
+        node_id: MastNodeId,
+        target_op_idx: Option<usize>,
+    ) -> Option<&AssemblyOp> {
+        MastForest::get_assembly_op(self, node_id, target_op_idx)
+    }
+
+    #[inline(always)]
+    fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>> {
+        MastForest::resolve_error_message(self, code)
+    }
+}
+
+// Blanket impl: an `Arc<T>` is an `ExecutableMastForest` whenever the underlying `T` is, which
+// allows the executor and tracer plumbing to be generic over a forest type while the live
+// (`Arc<MastForest>`) and replay (`Arc<SparseMastForest>`) paths each pick a concrete instance.
+impl<T> Index<MastNodeId> for Arc<T>
+where
+    T: Index<MastNodeId, Output = MastNode> + ?Sized,
+{
+    type Output = MastNode;
+
+    #[inline(always)]
+    fn index(&self, node_id: MastNodeId) -> &Self::Output {
+        &(**self)[node_id]
+    }
+}
+
+impl<T> Index<DecoratorId> for Arc<T>
+where
+    T: Index<DecoratorId, Output = Decorator> + ?Sized,
+{
+    type Output = Decorator;
+
+    #[inline(always)]
+    fn index(&self, decorator_id: DecoratorId) -> &Self::Output {
+        &(**self)[decorator_id]
+    }
+}
+
+impl<T: ExecutableMastForest + ?Sized> ExecutableMastForest for Arc<T> {
+    #[inline(always)]
+    fn get_node_by_id(&self, node_id: MastNodeId) -> Option<&MastNode> {
+        T::get_node_by_id(self, node_id)
+    }
+
+    #[inline(always)]
+    fn get_digest_by_id(&self, node_id: MastNodeId) -> Option<Word> {
+        T::get_digest_by_id(self, node_id)
+    }
+
+    #[inline(always)]
+    fn find_procedure_root(&self, digest: Word) -> Option<MastNodeId> {
+        T::find_procedure_root(self, digest)
+    }
+
+    #[inline(always)]
+    fn decorator_by_id(&self, decorator_id: DecoratorId) -> Option<&Decorator> {
+        T::decorator_by_id(self, decorator_id)
+    }
+
+    #[inline(always)]
+    fn advice_map(&self) -> &AdviceMap {
+        T::advice_map(self)
+    }
+
+    #[inline(always)]
+    fn linked_before_enter_decorators(&self, node_id: MastNodeId) -> &[DecoratorId] {
+        T::linked_before_enter_decorators(self, node_id)
+    }
+
+    #[inline(always)]
+    fn linked_after_exit_decorators(&self, node_id: MastNodeId) -> &[DecoratorId] {
+        T::linked_after_exit_decorators(self, node_id)
+    }
+
+    #[inline(always)]
+    fn decorator_indices_for_op(&self, node_id: MastNodeId, local_op_idx: usize) -> &[DecoratorId] {
+        T::decorator_indices_for_op(self, node_id, local_op_idx)
+    }
+
+    #[inline(always)]
+    fn get_assembly_op(
+        &self,
+        node_id: MastNodeId,
+        target_op_idx: Option<usize>,
+    ) -> Option<&AssemblyOp> {
+        T::get_assembly_op(self, node_id, target_op_idx)
+    }
+
+    #[inline(always)]
+    fn resolve_error_message(&self, code: Felt) -> Option<Arc<str>> {
+        T::resolve_error_message(self, code)
     }
 }
 
--- a/core/src/mast/sparse.rs
+++ b/core/src/mast/sparse.rs
@@ -1,0 +1,312 @@
+use alloc::{
+    collections::{BTreeMap, BTreeSet},
+    sync::Arc,
+    vec::Vec,
+};
+use core::ops::Index;
+
+use miden_utils_indexing::newtype_id;
+
+use crate::{
+    Felt, Word,
+    advice::AdviceMap,
+    mast::{
+        DebugInfo, Decorator, DecoratorId, ExecutableMastForest, MastForest, MastNode, MastNodeExt,
+        MastNodeId,
+    },
+    operations::AssemblyOp,
+};
+
+// MAST FOREST ID
+// ================================================================================================
+
+// `MastForestId` is an opaque handle to a [`MastForest`] in some forest store such as the
+// `TraceGenerationContext::mast_forest_store`. It is not a content-derived or stable identity for a
+// forest, and must not be compared or reused across stores or trace contexts. It is analogous to
+// `MastNodeId`, which is meaningful only within one forest's node store.
+newtype_id!(MastForestId);
+
+// SPARSE MAST FOREST
+// ================================================================================================
+
+/// A sparse replay view over a single source [`MastForest`]'s [`MastNodeId`] space, retaining only
+/// the nodes visited during execution.
+///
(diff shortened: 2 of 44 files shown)


processor/src/trace/execution_tracer.rs
```diff
@@ -146,21 +173,106 @@ impl ExecutionTracer {
...
             pending_restore_context: false,
             is_eval_circuit_op: false,
         }
     }
 
+    /// Returns the [`MastForestId`] of `forest` in [`Self::mast_forest_builders`], creating a new
+    /// builder for it on first encounter. Forests with equal commitments share the same id.
+    #[inline]
+    fn forest_id(&mut self, forest: &Arc<MastForest>) -> MastForestId {
+        let commitment = forest.commitment();
```
huitseeker: Could this key include the parts copied into `SparseMastForest`, or use forest identity instead? `MastForest::commitment()` only covers the procedure roots; the tests explicitly allow `advice_map` and debug info changes without changing it. If a host returns two forests with the same code but different advice maps or decorators, this reuses the first store entry and later replay can use the wrong data.

processor/src/trace/parallel/processor.rs
```diff
@@ -159,13 +166,14 @@ impl ReplayProcessor {
...
                     // load mast forest from replay
-                    let (root_id, new_forest) =
+                    let (root_id, new_forest_id) =
                         match self.mast_forest_resolution_replay.replay_resolution() {
                             Ok(v) => v,
                             Err(err) => {
                                 return ControlFlow::Break(BreakReason::Err(err));
                             },
                         };
+                    let new_forest = self.mast_forest_store[new_forest_id.to_usize()].clone();
```
huitseeker: This still trusts the `MastForestId` from `mast_forest_resolution_replay`. Since these ids can come from a serialized `TraceGenerationContext`, a bad id here will panic before returning `Err`, unlike `initial_mast_forest_id` above. Can this use the same checked lookup path for both dyn and external resolution replays?

processor/src/trace/execution_tracer.rs
huitseeker: There is also a source-local id issue here. Two forests can have the same root commitment while assigning different `MastNodeId`s to the same nodes. If the second forest reuses the first builder, `record_visit()` records the second forest's ids against the first forest, so finalization can miss nodes or copy the wrong ones. I think the sparse store needs to distinguish forests by their local id layout, not only by root commitment.

processor/src/trace/parallel/tests.rs
```diff
@@ -998,6 +995,45 @@ fn test_build_trace_returns_err_on_bad_node_id_in_hasher_replay() {
...
 
+/// Verifies that `build_trace` returns `Err` (instead of panicking) when a fragment's
+/// `initial_mast_forest_id` references an entry that does not exist in `mast_forest_store`.
+///
+/// `CoreTraceFragmentContext` can come from outside this process (e.g. when serialized and
+/// rebuilt), and so its [`miden_core::mast::MastForestId`]s must be treated as
+/// untrusted/attacker-controlled. `split_trace_fragment_context` validates them rather than
+/// indexing into the store, and the error must propagate up through `build_trace`.
+#[test]
+fn test_build_trace_returns_err_on_invalid_mast_forest_id() {
```
huitseeker: Could this test coverage also tamper with `mast_forest_resolution` for `Dyn` or `External` replay? The initial forest id path is checked, but resolution replay still carries forest ids too, and that is the path most likely to be hit when libraries are involved.

processor/src/trace/execution_tracer.rs
```diff
@@ -146,21 +173,106 @@ impl ExecutionTracer {
...
+    /// digest). Recursive descendants are not added — their inclusion happens naturally as
+    /// execution visits them.
+    #[inline]
+    fn record_visit(&mut self, forest: &Arc<MastForest>, node_id: MastNodeId) {
+        let id = self.forest_id(forest);
+        self.mast_forest_builders[id.to_usize()].record_visit(node_id);
+
+        if let Some(node) = forest.get_node_by_id(node_id) {
+            node.for_each_child(|child_id| {
+                self.mast_forest_builders[id.to_usize()].record_visit(child_id);
```
huitseeker: This records digest-only dependencies as normal visits, so `finalize()` copies the full child node even when replay only needs its digest. That is fine for correctness if the node is never entered, but it leaves size on the table and makes an accidental entry look like a partially-present node instead of a clear pruned/missing-node error.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// A sparse subset of a [`MastForest`] containing only the nodes visited during execution.
+///
+/// Unlike [`MastForest`], which stores nodes contiguously in an `IndexVec`, a [`SparseMastForest`]
+/// uses a `BTreeMap` so that it can preserve the original [`MastNodeId`]s of the source forest
+/// while omitting nodes that were not visited. This allows a [`SparseMastForest`] to be used as a
+/// drop-in replacement for the original [`MastForest`] during re-execution of a program: existing
+/// IDs continue to refer to the correct [`MastNode`]s.
+#[derive(Debug)]
+pub struct SparseMastForest {
```
huitseeker, line 49: Could the subset keep the source forest's original node count? Even if serialization stays out of scope here, that makes later validation much clearer: an ID can be reported as outside the source forest instead of just missing from the sparse map.

huitseeker (commented):
Thanks, this is directionally aligned with the sparse, ID-preserving replay shape from #2766. The main blockers I see are around forest identity and replay validation.

The biggest concern is that `MastForest::commitment()` is being used as the sparse-store key, but replay depends on source-local `MastNodeId`s. Equivalent root commitments do not necessarily mean equivalent local node numbering, and the sparse builder records node IDs against whichever source forest was first stored for that commitment. That can make later replay miss or copy the wrong nodes.

The other issue is consistency of validation. The PR treats `MastForestId`s in reconstructed trace contexts as attacker-controlled and validates the initial forest / continuation path, but `Dyn` and `External` resolution replay still indexes the store directly. That should return an `ExecutionError` too.

The remaining comments are non-blocking follow-ups: keeping original source node count would make sparse validation clearer, digest-only dependencies are currently copied as full nodes, and test coverage should include resolution replay IDs as well as the initial forest ID.

processor/src/trace/execution_tracer.rs
plafer: Right. In [4ce5c6a](https://github.com/0xMiden/miden-vm/pull/3105/commits/4ce5c6a1f4126224e25c50a0fd3f9565b503eb51), switched to using `*const MastForest` as the key in the `forest_ids` map

core/src/mast/sparse.rs
plafer: Done in [9f2fed1](https://github.com/0xMiden/miden-vm/pull/3105/commits/9f2fed1c01479660d9974e7ddb6b6093e3cb33bd)

processor/src/trace/parallel/processor.rs
```diff
@@ -165,13 +172,14 @@ impl ReplayProcessor {
...
```
huitseeker: When `TraceGenerationContext` comes from outside, the resolution replay can carry any `MastForestId`. This still indexes the store directly, so a bad id in a Dyn replay can panic instead of returning `ExecutionError` like `initial_mast_forest_id` now does. Could this go through the same checked lookup as the fragment setup? The External path below has the same shape.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,251 @@
...
+            }
+        }
+
+        SparseMastForest {
+            nodes,
+            num_nodes,
+            roots: source.procedure_roots().to_vec(),
+            advice_map: source.advice_map().clone(),
+            debug_info: source.debug_info().clone(),
+            commitment_cache: source.commitment_cache.clone(),
```
huitseeker: Could this store the source commitment as a plain `Word` instead of cloning the cache? `OnceLockCompat::clone()` is empty in no-std, and it may be empty in std too. Since `roots` keeps every source root but `nodes` only has visited nodes, a later `SparseMastForest::commitment()` can index an unvisited root and panic.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,251 @@
...
+
+// SPARSE MAST FOREST
+// ================================================================================================
+
+/// A sparse subset of a [`MastForest`] containing only the nodes visited during execution.
+///
+/// Unlike [`MastForest`], which stores nodes contiguously in an `IndexVec`, a [`SparseMastForest`]
+/// uses a `BTreeMap` so that it can preserve the original [`MastNodeId`]s of the source forest
+/// while omitting nodes that were not visited. This allows a [`SparseMastForest`] to be used as a
+/// drop-in replacement for the original [`MastForest`] during re-execution of a program: existing
```
huitseeker: Could we name this a little more narrowly? I think this type is a sparse replay view over one source forest's ID space. Saying "drop-in replacement" makes it sound like a second finalized forest shape. Maybe "sparse projection" or "replay view" would make the boundary clearer.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,251 @@
...
+        MastNodeId,
+    },
+    operations::AssemblyOp,
+};
+
+// MAST FOREST ID
+// ================================================================================================
+
+// `MastForestId` is an opaque handle to a [`MastForest`] in some forest store (such as the
+// `mast_forest_store` on a `TraceGenerationContext`). Analogous to [`MastNodeId`], but referring
```
huitseeker: Could this doc call out that `MastForestId` is store-local and positional, not a content identity? The same source forest can have a commitment, but this ID is really an index into the trace context's forest store.

core/src/mast/mod.rs
```diff
@@ -1000,6 +1003,188 @@ impl IndexMut<DecoratorId> for MastForest {
...
+// ================================================================================================
+
+/// A MAST forest that can be used as the source of nodes during program execution.
+///
+/// Implemented by both [`MastForest`] (a dense forest containing all nodes) and
+/// [`SparseMastForest`] (a sparse subset of a forest containing only the nodes visited during
+/// some prior execution). The latter preserves the original [`MastNodeId`]s of its source forest,
+/// which allows it to stand in for the dense forest during re-execution.
+pub trait ExecutableMastForest:
+    Index<MastNodeId, Output = MastNode> + Index<DecoratorId, Output = Decorator>
```
huitseeker: I'm not a fan of making `Index<MastNodeId>` part of this trait.

A sparse replay view can legitimately be missing nodes. Keeping fallible lookup as the trait contract would make that distinction explicit and would also reduce the chance that tampered replay data turns into a panic instead of an `ExecutionError`.

huitseeker (commented):
Mentally speaking (and unsurprisingly) it is easier for me to think about this PR on top of #3139 but YMMV.

processor/src/trace/execution_tracer.rs
plafer: Similar to previous comment, we now use `*const MastForest` for the key

processor/src/trace/parallel/tests.rs
plafer: Added 2 separate tests for each case in [05e6aee](https://github.com/0xMiden/miden-vm/pull/3105/commits/05e6aee22781f087c39c1c018f05d446be071bf3) (as rstest cases).

processor/src/trace/parallel/processor.rs
plafer: Fixed in [05e6aee](https://github.com/0xMiden/miden-vm/pull/3105/commits/05e6aee22781f087c39c1c018f05d446be071bf3)

processor/src/trace/execution_tracer.rs
plafer: Done in [257bc85](https://github.com/0xMiden/miden-vm/pull/3105/commits/257bc850a4aec1cb873d82acc6238d5e6064c6e1). The approach I take there is to store accessed digests in a separate `BTreeSet` in the `SparseMastForestBuilder`, and `record_visit()` now specifies if the access is a "full access", or just the digest. This required adding a `ExecutableMastForest::get_digest_by_id()` method that *must* be used when only the digest is needed.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /// Returns the commitment to this sparse forest, computed from the procedure roots.
+    ///
+    /// The commitment value is derived from the digests of the procedure roots in this forest, and
+    /// is therefore equal to the commitment of the source [`MastForest`] from which this sparse
+    /// forest was built (since [`SparseMastForest`] preserves the source forest's roots).
+    pub fn commitment(&self) -> Word {
+        *self.commitment_cache.get_or_init(|| {
+            let mut digests: Vec<Word> = self.roots.iter().map(|&id| self[id].digest()).collect();
```
huitseeker: `SparseMastForestBuilder` copies all `source.procedure_roots()`, but only stores nodes that were visited.  If a source forest has another exported root that was not executed, this `self[id]` path can panic even though sparse forests are meant to omit unvisited nodes.

Could this avoid indexing every original root?

core/src/mast/sparse.rs
plafer: Good point. Changed it to a `Word` (which stores the result from `source.commitment()`)

core/src/mast/sparse.rs
plafer: Fixed

core/src/mast/sparse.rs
plafer: Can you expand on what you mean, and what you want clarified? How I understand it, this applies to all IDs created from `newtype_id!()` (and also `MastNodeId`). i.e. these IDs are always store-local (e.g. `MastNodeId` is only relevant in the context of its `MastForest` node store, the same way that this `MastForestId` is only relevant in the context of the store that it's in).

How would you rephrase the docstring? It might help me understand where you're trying to go with this

huitseeker (commented):
Shaping up nicely!

core/src/mast/sparse.rs
huitseeker: Yes, that is the model I had in mind too. The only thing I wanted to make harder to misread is that `MastForestId` sits next to forest commitments, so it can look like a stable/content identity for a forest.

Maybe something like:

> `MastForestId` is an opaque index into a particular forest store, such as `TraceGenerationContext::mast_forest_store`. It is not a content-derived or stable identity for a forest, and  must not be compared or reused across stores or trace contexts. It is analogous to `MastNodeId`, which is meaningful only within one forest's node store.

The key bit I care about is “not content-derived or stable”, since that is the distinction from `MastForest::commitment()`.

core/src/mast/sparse.rs
```diff
@@ -0,0 +1,324 @@
...
+        }
+
+        SparseMastForest {
+            nodes,
+            digests,
+            num_nodes,
+            roots: source.procedure_roots().to_vec(),
+            advice_map: source.advice_map().clone(),
+            debug_info: source.debug_info().clone(),
+            commitment_cache: source.commitment_cache.clone(),
```
huitseeker: Another place related to [my last comment](https://github.com/0xMiden/miden-vm/pull/3105#discussion_r3236919444): If `commitment()` is called later, it recomputes through `self.roots`, but roots are copied wholesale and may include nodes that were never visited and are not in `nodes`. Could we store `source.commitment()` directly here instead of cloning the cache?

plafer: Yes - this is implemented but something came up yesterday and I forgot to push. Sorry for the confusion, will push the fix soon

core/src/mast/sparse.rs
plafer: Done in [ca0c021](https://github.com/0xMiden/miden-vm/pull/3105/commits/ca0c021ddd946057f380f2fdd003eeb6346683e4)

core/src/mast/sparse.rs
plafer: [ca0c021](https://github.com/0xMiden/miden-vm/pull/3105/commits/ca0c021ddd946057f380f2fdd003eeb6346683e4)

core/src/mast/sparse.rs
plafer: [ca0c021](https://github.com/0xMiden/miden-vm/pull/3105/commits/ca0c021ddd946057f380f2fdd003eeb6346683e4)

core/src/mast/mod.rs
plafer: Done in [06e2cb4](https://github.com/0xMiden/miden-vm/pull/3105/commits/06e2cb46a95e2a65162b3f4a45a74e4a7ad11255).

This meant using `ExecutableMastForest::get_node_by_id()`, and mapping the `None` case to a `ControlFlow::Break`, for which there is no nice `Option` method available. I tried to alleviate the syntactic eye sore of using `match` statements everywhere for that conversion by introducing `option_map_break_reason()`; open to alternative suggestions though.

core/src/mast/sparse.rs
plafer: Makes sense; done in [8dd5fdb](https://github.com/0xMiden/miden-vm/pull/3105/commits/8dd5fdba65d179989c07b84af85c847b1ad6a432)

huitseeker (approved):
LGTM!

plafer:
Merging this now to avoid merge conflicts with #1684.

Issue #2928, opened 2026-03-27
Rename `ModuleInfo`


bobbinth:
Possible options: `ModuleManifest`, `ModuleDescriptor`, `ModuleMetadata`. My personal preference would be either "descriptor" or "metadata" (with a slight preference towards "descriptor").

This is to be done after https://github.com/0xMiden/miden-vm/pull/2877 is merged.

_Context https://github.com/0xMiden/miden-vm/pull/2877#discussion_r2997319102_

giwaov:
I'd like to take this one. Will rename ModuleInfo to ModuleDescriptor once #2877 is merged.

giwaov:
PR #2989 is up for this. Could a maintainer assign this issue to me? Thanks!

huitseeker:
This is blocked by #2984 and #2985 that are needed to minimize the diff surface between next and main.

bobbinth:
As a part of this, I'd also probably rename the current `Kernel` struct into something like `KernelDescriptor`.

huitseeker:
Closed in #3356

PR #3106, opened 2026-05-06
refactor: remove Library struct

bitwalker:
This PR removes the old `Library` struct, and related APIs, while moving as much of the serialization hardening and Library-specific tests to `Package` as possible.

A few notable changes:

* The various `Assembler` plumbing APIs, such as `assemble_library`, now require a name to be provided - this name is given to the package which is produced. For existing tests, a simple dummy name like `library` or `program` is used, unless something more appropriate was identified.
* The `(with|link)_(dynamic|static)_library` APIs were removed in favor of `with_package`/`link_package` and an explicit linkage.
* `Package` gained some of the convenience functions that `Library` used to provide


crates/mast-package/src/package/mod.rs
```diff
@@ -78,27 +82,78 @@ pub struct Package {
...
-            .with_dependencies(dependencies)
-            .expect("package dependencies should be unique");
+    ) -> Result<Self, ManifestValidationError> {
+        let manifest = PackageManifest::new(exports)?.with_dependencies(dependencies)?;
+
+        // Validate that procedure export node provenance is valid when present
+        for export in manifest.exports() {
+            if let Some(proc) = export.as_procedure()
+                && let Some(node) = proc.node
+                && !mast.is_procedure_root(node)
```
huitseeker: Right now this only checks that the node is a root. Later APIs split on the two values: `module_infos()` registers `digest`, while `get_procedure_node_by_path()` and the package digest use `node`, so a crafted package can make name resolution and executed code disagree.

Could we also reject the case where `proc.node` points at one root but `proc.digest` names a different root?

crates/assembly/src/linker/mod.rs
```diff
@@ -334,29 +337,35 @@ impl Linker {
...
+        }
+
+        log::debug!(target: "linker", "instantiating linker with kernel package {}@{}", &kernel_package.name, &kernel_package.version);
 
         let mut graph = Self::new(source_manager);
-        let kernel_index = graph
-            .link_assembled_module(kernel_module)
-            .expect("failed to add kernel module to the module graph");
+        let mut kernel_index = None;
+        for module_info in kernel_package.module_infos() {
```
huitseeker: Do we still want the old `KernelLibrary` rule that rejected procedure exports outside `$kernel`? `Package::to_kernel()` now just filters `$kernel` exports, and this loop links every module in the kernel package. So a malformed kernel package can carry extra non-kernel modules into the linker instead of being rejected. If that is not intended, kernel package validation should reject non-`$kernel` procedure exports before this point.

crates/assembly/src/linker/mod.rs
```diff
@@ -368,20 +377,33 @@ impl Linker {
...
-        );
-        log::debug!(target: "linker", "modifying linker with kernel {}", kernel_module.path());
-        let kernel_index = self.link_assembled_module(kernel_module)?;
-        self.kernel_index = Some(kernel_index);
+        assert!(self.kernel_package.is_none());
+
+        log::debug!(target: "linker", "modifying linker with kernel package {}@{}", &kernel_package.name, &kernel_package.version);
+
+        let mut kernel_index = None;
+        for module_info in kernel_package.module_infos() {
```
huitseeker, line 376: Could this link the kernel package modules transactionally, or roll back on error? If one of the later `link_assembled_module` calls fails, earlier kernel modules stay in `self.modules` while `kernel`, `kernel_index`, and `kernel_package` remain unset. The AST `link_kernel` path below already rolls back this kind of failed kernel link.

miden-core-fuzz/Cargo.toml
```diff
@@ -177,20 +177,6 @@ test = false
 doc = false
 bench = false
 
-[[bin]]
-name = "library_deserialize"
```
huitseeker, line 181: Could we also remove this from `make fuzz-all`? `miden-core-fuzz/Cargo.toml` drops the `library_deserialize` and `library_serde_deserialize` targets, but the root `Makefile` still runs both, so `make fuzz-all` will fail before it reaches the package targets.

miden-core-fuzz/Cargo.toml
```diff
@@ -177,20 +177,6 @@ test = false
...
-path = "fuzz_targets/library_deserialize.rs"
-test = false
-doc = false
-bench = false
-
-[[bin]]
-name = "library_serde_deserialize"
```
huitseeker, line 188: Do we want the package serde target to exercise the same semantic APIs as `package_semantic_deserialize` after a successful parse? The current `package_serde_deserialize` target only parses JSON and drops the result, so fuzzing covers `to_kernel`, `try_into_program`, embedded kernel parsing, and debug sections only for binary `Package::read_from_bytes` values, not for serde-created `Package` values.

crates/lib/core/src/lib.rs
```diff
@@ -65,7 +66,7 @@ use crate::handlers::{
 ///
 /// # Usage
 ///
-/// The core library is typically used with an [`Assembler`] to enable core library procedures
+/// The core library is typically used with the assembler to enable core library procedures
```
huitseeker, line 70: This usage example still shows the removed `with_dynamic_library` API. It should probably mirror the README example and use `with_package(CoreLibrary::default().package(), Linkage::Dynamic)`, with `Linkage` imported from `miden_assembly`.

crates/mast-package/src/package/manifest.rs
```diff
@@ -278,28 +296,61 @@ pub struct ProcedureExport {
     /// The fully-qualified path of the procedure exported by this package.
     #[cfg_attr(feature = "serde", serde(with = "miden_assembly_syntax::ast::path"))]
     #[cfg_attr(
-        feature = "arbitrary",
+        any(test, feature = "arbitrary"),
         proptest(strategy = "miden_assembly_syntax::arbitrary::path::bare_path_random_length(2)")
     )]
     pub path: Arc<Path>,
+    /// The id of the MAST root node corresponding to this procedure
```
huitseeker, line 308: Could we spell out the invariant between `node` and `digest` here? The old `Library` docs explained why node provenance matters when two exports share a root digest, but `Package` now stores both fields separately. Without stating that a present node must point at this digest, it is easy for callers and tests to treat one as authoritative in one path and the other as authoritative elsewhere.

crates/assembly/README.md
```diff
@@ -126,11 +117,11 @@ or by compiling a kernel module from source, as shown below:
 
 // First, assemble the kernel library
 let kernel_lib = Assembler::new(source_manager.clone())
-    .assemble_kernel("pub proc foo add end")
+    .assemble_kernel("my-kernel", "pub proc foo add end")
     .unwrap();
 
 // Create assembler with the kernel
-let assembler = Assembler::with_kernel(source_manager, kernel_lib);
+let assembler = Assembler::with_kernel(source_manager, kernel_lib.into());
```
huitseeker: This example now leaves `assembler` as a `Result<Assembler, Report>`, unlike the later example which calls `.unwrap()`. Could we either unwrap/propagate here too, or make the snippet return a `Result`? As written, it no longer shows a usable assembler value.

crates/assembly/src/tests.rs
```diff
@@ -5235,46 +4989,9 @@ fn regression_empty_kernel_library_is_rejected() {
...
     let kernel_masm = "pub const FOO = 1\n";
     let err = Assembler::new(source_manager)
-        .assemble_kernel(kernel_masm)
+        .assemble_kernel("kernel", kernel_masm)
         .expect_err("expected empty kernel to be rejected");
-    assert_diagnostic_lines!(err, "library must contain at least one exported procedure");
-}
-
-#[test]
-fn regression_empty_kernel_package_is_rejected_without_panicking() {
```
huitseeker, line 5421: Could we keep an equivalent for this assembler/linker regression? `to_kernel_rejects_empty_kernel_exports` covers the package conversion helper, but this test also proved that `Assembler::link_package` handles an empty kernel package with a structured error instead of panicking or leaving linker state half-mutated.

crates/assembly/src/tests.rs
```diff
@@ -631,12 +522,12 @@ pub proc p
...
 "#;
     let lib = Assembler::default()
-        .assemble_library([lib_src])
+        .assemble_library("lib", [lib_src])
         .expect("library assembly must succeed");
 
     let (bytes, _) =
         build_library_bytes_with_spoofed_first_node_digest(&lib, "spoofed-library-digest");
-    let err = Library::read_from_bytes(&bytes)
+    let err = Package::read_from_bytes(&bytes)
```
huitseeker: Since these now exercise `Package::read_from_bytes`, could we rename this group while touching it? The helper and four test names still say `library`/`kernel_library`, which makes it look like the old `Library` coverage still exists instead of package deserialization coverage.

crates/mast-package/src/package/mod.rs
```diff
@@ -55,18 +61,16 @@ pub struct Package {
...
     /// The project target type which produced this package
     pub kind: TargetType,
-    /// The underlying [Library] of this package
-    ///
-    /// NOTE: This will change to `MastForest` soon. We are currently using `Library` because we
-    /// have not yet fully removed the usage of `Library` throughout the assembler, so it is more
-    /// convenient to use. However, this can change at any time, so you should avoid accessing
-    /// this field directly unless you absolutely need to and can handle potential breakage.
-    pub mast: Arc<Library>,
+    /// The underlying [MastForest] of this package
```
huitseeker, line 80: Since `assembly-syntax::library` is being deleted, could the package docs preserve the useful parts of the old `Library` docs here? In particular: 
- packages/libraries export one or more procedures, 
- exports are keyed by fully-qualified paths, 
- multiple paths can point at the same MAST root, 
- roots may exist with no exported path, and 
- `MastNodeId` provenance matters because two exports with the same digest can still differ by decorators. 

Those details used to make the export map and node-vs-digest behavior much easier to reason about.

huitseeker (changes requested):
Thanks, the direction makes sense, but there are many sharp edges. Request changes for the export node/digest comment, serde bypassing validation, kernel packages accepting non-kernel exports, and kernel-package linking leaving partial state on failure.

crates/mast-package/src/package/mod.rs
```diff
@@ -55,18 +61,16 @@ pub struct Package {
     pub name: PackageId,
     /// An optional semantic version for the package
     pub version: Version,
+    /// The content hash of the exported code of this package, formed by hashing the roots of all
+    /// exports in lexicographical order (by digest, not procedure name)
+    digest: Word,
```
bitwalker, line 75: I'll probably just remove the `serde` deserialization path - it isn't actually used by anything other than tests, and those tests could just as easily use the `Deserializable` impl.

crates/assembly/src/linker/mod.rs
bitwalker: I don't really see any reason to reject that - the actual root kernel module is the only one whose exports end up in the `Kernel`, and any other exports might be useful library functions - but it's certainly easy enough to prevent kernel packages from exporting any symbols other than those from the root kernel module.

@bobbinth Thoughts?

miden-core-fuzz/Cargo.toml
bitwalker: Ah, oops, didn't realize they were also named there, thanks!

crates/assembly/src/tests.rs
bitwalker: It isn't possible to construct a `Package` with empty exports using normal APIs - either by deserializing or via `Package::create` - you'd have to deliberately construct it manually by hand using struct literal syntax in the same code which tries to assemble with it, at which point all bets are off anyway. The only reason we even allow constructing a `Package` using struct literal syntax in the first place is for testing purposes, otherwise we could make it impossible to even do short of `unsafe`.

So I guess my question is - do we want to test how the linker handles an empty kernel package, or do we want to make it impossible to create one in the first place?

crates/assembly/src/tests.rs
bitwalker: Sure - in general though we often call library packages _libraries_ for short, so I don't think this is something we need to change everywhere in general.

crates/mast-package/src/package/mod.rs
bitwalker: Sure thing!

crates/assembly/src/tests.rs
bitwalker: I'm actually mistaken - we can't create packages using struct literal syntax outside of `miden-mast-package` itself, so it's actually impossible to create packages with empty exports currently.

crates/assembly/src/linker/mod.rs
bitwalker: It isn't really possible to do this - unlike the single AST module path, linking a kernel package involves potentially many modules, and each one when linked can modify the linker in various ways. We'd essentially need to clone the whole linker first to ensure that we didn't miss any changes that get made. Even the AST path doesn't get this right (it only clones the things it thinks will get modified).

I would argue that trying to preserve the state of the assembler after an error makes no sense to begin with - nobody is going to construct an assembler, and then keep trying to assemble after an attempt to link a package fails, not only would that be non-sensical (how would you even recover from that error?), it overcomplicates things for little to no gain IMO.

So if we really think we need to handle this, then we should probably make every `Assembler` API that links something, into an API that takes `self: Box<Self>`, so that if an error occurs, there is no way to try and reuse the assembler to recover.

crates/mast-package/src/package/mod.rs
bitwalker: Reviewing the docs, I think all of this is already covered, though not necessarily at the root `Package` struct item docs (and I generally would like to avoid duplicating docs in multiple places, and instead refer folks to other items which provide authoritative docs for some topic):

* The `Package` docs already describe the basic structure and purpose
* `PackageManifest` already describes what items the package contains, including procedures
* I will clarify that the exports are keyed by fully-qualified paths in the manifest, though I'll note that each of the item export types (e.g. `ProcedureExport`) already describe that their path is fully-qualified.
* The `node` field of `ProcedureExport` covers `MastNodeId` vs digest references, provenance, and that roots can exist in the MAST with no corresponding root, as well as the fact that multiple exports can share the same root. This is the appropriate place for this IMO, since it is an implementation detail that is only relevant for procedure exports.

crates/mast-package/src/package/serialization.rs
```diff
@@ -341,18 +478,64 @@ impl Serializable for ProcedureExport {
...
+impl ProcedureExport {
+    pub fn read_from_safe<R: ByteReader>(
+        source: &mut R,
+        mast: &MastForest,
+    ) -> Result<Self, DeserializationError> {
+        use miden_assembly_syntax::ast::types::FunctionType;
+        let path = PathBuf::read_from(source)?.into_boxed_path().into();
+        let node = if source.read_bool()? {
+            let node_id = MastNodeId::from_u32_safe(source.read_u32()?, mast)?;
+            if !mast.is_procedure_root(node_id) {
```
huitseeker, line 490: I think a package can serialize `node = A` where `A` is a valid procedure root, but put `digest = B` for another procedure. `Package::create` rejects that with `is_procedure_root_with_exact_digest`, but this deserialization path only checks `is_procedure_root` and then `recompute_mast_commitment` trusts `Some(node)`. That leaves exports where `get_export_node_id()` and `get_procedure_root_by_path()` disagree about which procedure was exported.

Could this check bind the stored node id to the stored digest too?

miden-vm/src/cli/verify.rs
```diff
@@ -117,14 +117,16 @@ fn load_kernel(kernel_path: &PathBuf) -> Result<Kernel, Report> {
                 Package::read_from_bytes(&bytes).into_diagnostic().wrap_err_with(|| {
                     format!("Failed to deserialize kernel package `{}`", kernel_path.display())
                 })?;
-
-            package.try_into_kernel_library()?
+            Arc::new(package)
```
huitseeker, line 126: Could this keep the package kind check before extracting the kernel? The old `try_into_kernel_library()` path rejected `.masp` files whose package kind was not `Kernel`, but `to_kernel()` only looks for `$kernel` exports. That means `miden-vm verify --kernel some_library.masp` can now accept a non-kernel package if it has forged or accidental `$kernel` exports, while the compile path still rejects the same package through `Assembler::with_kernel()`.

crates/assembly/src/linker/mod.rs
huitseeker: > So if we really think we need to handle this, then we should probably make every Assembler API that links something, into an API that takes self: Box<Self>, so that if an error occurs, there is no way to try and reuse the assembler to recover

FWIW I think that semi-persistence is a nice feature.

huitseeker (commented):
Since `Package` now carries both `ProcedureExport.node` and an independent `ProcedureExport.digest`, should the package fuzz target add an oracle for that new invariant? The old `Library` format only stored the export node and derived the digest from the MAST, so this mismatch state did not exist before. `package_semantic_deserialize` currently calls APIs after decode, but it would not catch a package where the node and digest point at different procedure roots unless we assert that relationship.

crates/assembly/src/linker/mod.rs
```diff
@@ -334,29 +337,35 @@ impl Linker {
...
-            .expect("failed to add kernel module to the module graph");
+        let mut kernel_index = None;
+        for module_info in kernel_package.module_infos() {
+            let is_kernel_module = module_info.path().is_kernel_path();
+            let module_index = graph.link_assembled_module(module_info)?;
+            if is_kernel_module {
+                kernel_index = Some(module_index);
+            }
+        }
+        debug_assert!(kernel_index.is_some());
```
plafer: Probably best to `assert!()` here? Otherwise in release builds this will just silently fail (and supposedly cause errors further down the line anyways)

crates/mast-package/src/package/mod.rs
```diff
@@ -44,61 +50,117 @@ use crate::{Dependency, Version};
...
+        self.digest = digest;
+        Ok(())
+    }
+
+    /// Produces a new library with the existing [`MastForest`] and where all key/values in the
+    /// provided advice map are added to the internal advice map.
+    pub fn with_advice_map(mut self, advice_map: AdviceMap) -> Self {
+        self.extend_advice_map(advice_map);
+        self
+    }
```
plafer, line 158: nit: What's the value of `with_advice_map()`? For it to be useful, I would have expected it to be `&self` (and actually creating a copy of `AdviceMap`) - but now it's basically the same as `extend_advice_map()`?

crates/mast-package/src/package/mod.rs
```diff
@@ -402,68 +556,48 @@ impl Package {
...
+                    Some(node_id),
+                    digest,
+                    None,
+                ))],
+                self.manifest.dependencies.clone(),
+            )
+            .map_err(Report::msg)?;
+
+            package.description = self.description.clone();
+            package.sections = self.sections.clone();
```
plafer, line 588: Q: does it always make sense to keep all sections? Or should we be filtering out only the ones that we know make sense?

crates/assembly/src/assembler.rs
```diff
@@ -500,24 +426,29 @@ impl Assembler {
         self,
         dir: impl AsRef<std::path::Path>,
         namespace: impl AsRef<Path>,
-    ) -> Result<Arc<Library>, Report> {
+    ) -> Result<Box<Package>, Report> {
         use miden_assembly_syntax::parser;
 
         let dir = dir.as_ref();
         let namespace = namespace.as_ref();
+        let name = namespace.as_str().replace("::", "-");
```
plafer, line 434: Would it be a good idea to add a `name: Option<String>` argument to `assemble_library_from_dir()`? This forces the user to have the library name be derived from the path - a decent default, but I would expect being able to opt out of that and provide my own name.

crates/assembly/src/linker/mod.rs
```diff
@@ -334,29 +337,35 @@ impl Linker {
...
-    ) -> Self {
-        assert!(!kernel.is_empty());
-        assert!(
-            kernel_module.path().is_kernel_path(),
-            "invalid root kernel module path: {}",
-            kernel_module.path()
-        );
-        log::debug!(target: "linker", "instantiating linker with kernel {}", kernel_module.path());
+        kernel_package: Arc<MastPackage>,
+    ) -> Result<Self, Report> {
```
plafer, line 343: nit: This method is very similar to `link_with_kernel()` - is there a good reason to keep them separate? Or at the very least, extract their common implementation into a helper

crates/mast-package/src/package/mod.rs
bitwalker: The `with_*` family of functions are builders, and are only used during initial construction of a `Package`

crates/mast-package/src/package/mod.rs
bitwalker: I'm not really sure what the right answer is - but IMO, this function is designed to simply make the package appear as executable type - it doesn't (and probably shouldn't) do anything more than that.

crates/assembly/src/assembler.rs
bitwalker: I would rather get rid of this function entirely to be honest - this use case should be using Miden projects going forward, and the lack of flexibility around naming is a helpful nudge in that direction while this function still exists.

miden-vm/src/cli/verify.rs
```diff
@@ -135,7 +137,7 @@ fn load_kernel(kernel_path: &PathBuf) -> Result<Kernel, Report> {
     };
 
     // Extract kernel from kernel library
-    Ok(kernel_lib.kernel().clone())
+    kernel_pkg.to_kernel()
```
huitseeker, line 146: Could this keep the old kind check for `.masp` kernels? `to_kernel()` only looks for `$kernel` exports, so a package with `kind = Library` but forged `$kernel::*` exports would now be accepted by `miden verify --kernel`. 

The previous `try_into_kernel_library()` path rejected non-kernel packages before extracting the kernel.

crates/mast-package/src/package/serialization.rs
```diff
@@ -327,6 +458,12 @@ impl Deserializable for PackageExport {
 impl Serializable for ProcedureExport {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
         self.path.write_into(target);
+        if let Some(node_id) = self.node {
```
huitseeker, line 461: This changes the package wire layout, but the package `VERSION` is still `[4, 0, 0]`. 

Technically, we should bump the package format version.

crates/assembly/src/linker/mod.rs
```diff
@@ -368,20 +358,33 @@ impl Linker {
...
-        );
-        log::debug!(target: "linker", "modifying linker with kernel {}", kernel_module.path());
-        let kernel_index = self.link_assembled_module(kernel_module)?;
-        self.kernel_index = Some(kernel_index);
+        assert!(self.kernel_package.is_none());
+
+        log::debug!(target: "linker", "modifying linker with kernel package {}@{}", &kernel_package.name, &kernel_package.version);
+
+        let mut kernel_index = None;
+        for module_info in kernel_package.module_infos() {
```
huitseeker, line 376: A consequence of [the latest fix](https://github.com/0xMiden/miden-vm/pull/3106#discussion_r3228618112): `to_kernel()` only keeps `$kernel` exports, so a malformed package with `kind = Kernel` can still bring extra public non-kernel modules into the linker.

Could this rely on a stricter kernel-package validation before registering every `module_info()` here?

bitwalker: I don't believe that's considered malformed - there is no restriction on a kernel package providing non-kernel library functionality. The linker will disallow calling any such functionality via `syscall` already.

crates/mast-package/src/package/manifest.rs
```diff
@@ -62,38 +76,20 @@ impl PackageManifest {
...
+                has_procedures = true;
+            }
+            normalize_export(&mut export)?;
             manifest.add_export(export)?;
         }
 
-        Ok(manifest)
-    }
+        if !has_procedures {
+            return Err(ManifestValidationError::NoProcedures);
```
huitseeker, line 89: This makes standalone manifest deserialization reject values that `PackageManifest` can still create and serialize, like `PackageManifest::default()` or a dependency-only manifest.

Could the "must export a procedure" rule live at the `Package` validation layer instead? That keeps `PackageManifest` round-trippable on its own.

crates/mast-package/src/package/arbitrary.rs
```diff
@@ -0,0 +1,98 @@
...
+                    exports,
+                    dependencies.clone(),
+                )
+                .expect("invalid package")
+            })
+            .boxed()
+    }
+
+    type Strategy = proptest::prelude::BoxedStrategy<Self>;
+}
```
huitseeker, line 107: This can panic during arbitrary value generation when the generated exports collide after path normalization. Could this use `prop_filter_map` or build a de-duplicated export set before calling `Package::create`?

crates/mast-package/src/package/manifest.rs
bitwalker: I went with removing the `Default` impl - package manifests already can't be created with no exports, so that closes the last gap AFAIK that would allow round-tripping to fail

bitwalker:
@bobbinth This is ready to merge once you've done your review

miden-core-fuzz/Cargo.toml
huitseeker, line 181: The GitHub fuzz workflow still lists `library_deserialize` in its target matrix and generated-corpus list, but this PR deletes that fuzz target. Locally, `cargo +nightly fuzz run library_deserialize --fuzz-dir miden-core-fuzz -- -runs=0` now fails with `no bin target named library_deserialize`.

Could this also remove the stale workflow entries so the scheduled/manual fuzz job does not fail on that target?

crates/assembly/README.md
```diff
@@ -26,18 +26,22 @@ use miden_assembly::Assembler;
...
 
+let source_manager = Arc::new(DefaultSourceManager::default());
+
 // Instantiate a default, empty assembler
-let assembler = Assembler::new(Arc::new(DefaultSourceManager::default()));
+let assembler = Assembler::new(source_manager.clone());
 
-// Emit a program which pushes values 3 and 5 onto the stack and adds them
-let program1 = assembler.assemble_program("begin push.3 push.5 add end")
+// Emit an executable package, named `prg` which pushes values 3 and 5 onto the
```
huitseeker: Nit:: `git diff --check origin/next...HEAD` fails on this line and the similar note just below because both have trailing spaces. Could we strip those before merge?

crates/assembly/src/project.rs
```diff
@@ -225,29 +231,31 @@ where
             None => self.load_target_sources(project.as_ref(), target)?,
         };
 
-        let product = match target.ty {
-            TargetType::Executable => assembler.assemble_executable_modules(root, support)?,
+        let mut product = match target.ty {
+            TargetType::Executable => {
+                if let Some(kernel_package) = runtime_dependencies.kernel.clone() {
+                    assembler.link_package(kernel_package, Linkage::Dynamic)?;
```
huitseeker: Could this skip kernels that were already linked in the dependency loop?

crates/mast-package/src/package/mod.rs
```diff
@@ -44,61 +50,117 @@ use crate::{Dependency, Version};
...
 #[derive(Debug, Clone, Eq, PartialEq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Package {
     /// Name of the package
     pub name: PackageId,
     /// An optional semantic version for the package
     pub version: Version,
+    /// The content hash of the exported code of this package, formed by hashing the roots of all
+    /// exports in lexicographical order (by digest, not procedure name)
+    digest: Word,
```
huitseeker, line 75: Could we either make `mast` private ? `digest` is now cached, but `mast` is still public, so a caller can replace it and `Package::digest()` keeps returning the old commitment.


```rust
#[test]
fn replacing_public_mast_leaves_cached_digest_stale() {
    let mut package = build_package("first", TargetType::Library, "first::entry", [], vec![]);
    let replacement = build_package_with_mul_root();

    package.mast = replacement.mast.clone();

    assert_eq!(package.digest(), replacement.digest());
}
```

huitseeker (approved):
Super, thanks a lot!

crates/assembly/src/assembler.rs
bobbinth, line 1: nit (and not for this PR): I'd move this file into `src/assembler/mod.rs` to be consistent with how we do it for other modules (and same goes for `src/project.rs` -> `src/project/mod.rs`).

crates/assembly/src/assembler.rs
```diff
@@ -47,11 +51,11 @@ pub(crate) const MAX_CONTROL_FLOW_NESTING: usize = 256;
...
 /// The [Assembler] produces a _Merkelized Abstract Syntax Tree (MAST)_ from Miden Assembly sources,
-/// as an artifact of one of three types:
+/// as a [`Package`] artifact. In general, packages come in three primary varieties:
 ///
-/// * A kernel library (see [`KernelLibrary`])
-/// * A library (see [`Library`])
-/// * A program (see [`Program`])
+/// * A kernel library (i.e. [`TargetType::Kernel`])
+/// * A program (see [`TargetType::Executable`])
+/// * A library (all other target types)
```
bobbinth, line 58: nit (and not related this pr): I'd rename `TargetType` into `PackageKind` or something similar.

crates/assembly/src/assembler.rs
```diff
@@ -436,15 +360,16 @@ impl Assembler {
...
     ///
     /// # Errors
     ///
     /// Returns an error if parsing or compilation of the specified modules fails.
     pub fn assemble_library(
         self,
+        name: impl Into<PackageId>,
         modules: impl IntoIterator<Item = impl Parse>,
-    ) -> Result<Arc<Library>, Report> {
+    ) -> Result<Box<Package>, Report> {
```
bobbinth, line 372: Question: why do we return `Box< Package>` from here rather than just `Package`? It would seem like returning just `Package` would be less opinionated here (and I'm not sure we need to be more opinionated in this case).

This also applies to the other `assemble_` methods in this module.

crates/assembly/src/assembler.rs
```diff
@@ -292,94 +298,14 @@ impl Assembler {
...
-        self.linker.link_library(library).map_err(Report::from)
-    }
-
-    /// Statically link against `library` during assembly.
-    ///
-    /// See [`Self::link_static_library`]
-    pub fn with_static_library(mut self, library: impl AsRef<Library>) -> Result<Self, Report> {
-        self.link_static_library(library)?;
+    /// Link against `package` with the specified linkage mode during assembly.
+    pub fn with_package(mut self, package: Arc<Package>, linkage: Linkage) -> Result<Self, Report> {
```
bobbinth, line 302: Would be good to add a description of error conditions. For example, I'm assuming this will fail if we try to link the same package twice - once with static linking and once with dynamic linking - but it would be good to describe this.

crates/assembly/src/assembler.rs
```diff
@@ -138,15 +142,17 @@ impl Assembler {
...
-    /// Start building an [`Assembler`] with a kernel defined by the provided [KernelLibrary].
-    pub fn with_kernel(source_manager: Arc<dyn SourceManager>, kernel_lib: KernelLibrary) -> Self {
-        let (kernel, kernel_module, _) = kernel_lib.into_parts();
-        let linker = Box::new(Linker::with_kernel(source_manager.clone(), kernel, kernel_module));
-        Self {
+    /// Start building an [`Assembler`] with a kernel defined by the provided kernel package.
+    pub fn with_kernel(
+        source_manager: Arc<dyn SourceManager>,
+        kernel: Arc<Package>,
+    ) -> Result<Self, Report> {
```
bobbinth, line 149: Should we change this into a more "builder-like" method - e.g.:

```rust
pub fn with_kernel(mut self, kernel: Arc<Package>) -> Result<Self, Report> {
   ...
}
```

This would be more consistent with other similar methods (e.g., `with_package()`).

crates/assembly/src/assembler.rs
```diff
@@ -500,24 +426,29 @@ impl Assembler {
...
     ///
     /// # Errors
     ///
     /// Returns an error if parsing or compilation of the specified modules fails.
-    pub fn assemble_kernel(self, module: impl Parse) -> Result<KernelLibrary, Report> {
+    pub fn assemble_kernel(
+        self,
+        name: impl Into<PackageId>,
+        module: impl Parse,
+    ) -> Result<Box<Package>, Report> {
```
bobbinth, line 451: Maybe not for this PR, but it's worth expanding the comments here to describe the "shape" of the returned package as well as error conditions.

For example, IIRC, the returned package will have one top-level namespace (`$kernel`) and all procedures from the provided `module` will be exported from this namespace. Also, this package must be linked into the assembler using `with_kernel()` method (i.e., trying to use `with_package()` with it, will fail).

crates/assembly/src/assembler.rs
```diff
@@ -662,6 +594,7 @@ impl Assembler {
                     // We didn't find the procedure in our current MAST forest. We still need to
                     // check if it exists in one of a library dependency.
                     None => {
+                        log::trace!(target: "assembler::export_symbol", "no procedure found in forest");
```
bobbinth, line 597: I know we do this in other places, but I'm still not sure we should be keeping `log` statements in library code.

crates/assembly/README.md
```diff
@@ -26,18 +26,22 @@ use miden_assembly::Assembler;
...
-// Note: assemble_program() takes ownership of the assembler, so create a new one for the next program
-let assembler2 = Assembler::new(Arc::new(DefaultSourceManager::default()));
+// Note: assemble_program() consumes the assembler, so create a new one for the
+// next program
+let assembler2 = Assembler::new(source_manager.clone());
 
-// Emit a program from some source code on disk (requires the `std` feature)
-let program2 = assembler2.assemble_program(Path::new("../../miden-vm/masm-examples/fib/fib.masm"))
+// Emit a program from some source file on disk (requires the `std` feature)
+let _ = assembler2.assemble_program("prg", Path::new("../../miden-vm/masm-examples/fib/fib.masm"))
```
bobbinth, line 44: nit: why not use `program2` name for the returned program here as before?

crates/assembly/src/assembler.rs
bobbinth: Agreed - let's create an issue for removing these.

crates/assembly/README.md
```diff
@@ -49,43 +53,39 @@ let program2 = assembler2.assemble_program(Path::new("../../miden-vm/masm-exampl
...
-# use miden_assembly::Assembler;
+# use miden_assembly::{Assembler, Linkage};
 # use miden_assembly_syntax::debuginfo::DefaultSourceManager;
 # use miden_core_lib::CoreLibrary;
 # use std::sync::Arc;
 #
-let assembler = Assembler::new(Arc::new(DefaultSourceManager::default()))
-    .with_dynamic_library(&CoreLibrary::default())
+# // Create a source manager
+# let source_manager = Arc::new(DefaultSourceManager::default());
```
bobbinth, line 82: Not for this PR, but it would be good to add a section about the source manager at some point - e.g., to explain its intended usage and maybe provide an example of how to use it correctly.

crates/assembly/src/testing.rs
```diff
@@ -192,8 +193,8 @@ impl TestContext {
 
     /// Add the modules of `library` to the [Assembler] constructed by this context.
     #[track_caller]
-    pub fn add_library(&mut self, library: impl AsRef<Library>) -> Result<(), Report> {
-        self.assembler.link_dynamic_library(library)
+    pub fn add_library(&mut self, package: Arc<miden_mast_package::Package>) -> Result<(), Report> {
+        self.assembler.link_package(package, miden_project::Linkage::Dynamic)
     }
```
bobbinth, line 198: nit: should this be `add_package()`?

processor/src/host/default.rs
```diff
@@ -227,6 +227,15 @@ pub struct HostLibrary {
...
 }
 
+impl From<Arc<miden_mast_package::Package>> for HostLibrary {
+    fn from(package: Arc<miden_mast_package::Package>) -> Self {
+        Self {
+            mast_forest: package.mast_forest().clone(),
+            handlers: vec![],
+        }
+    }
+}
```
bobbinth, line 237: I wonder if `Package` can now replace the concept of `HostLibrary`. This will require a deeper refactoring because `EventHandler` type needs to move either into core or package crates - but if we want the handlers to be eventually distributed with packages, we'd need to do this anyways. Let's create an issue to explore this.

miden-vm/src/cli/utils.rs
```diff
@@ -68,25 +69,28 @@ pub fn get_masm_program(
...
+            .link_package(CoreLibrary::default().package(), Linkage::Dynamic)
             .wrap_err("Failed to load stdlib")?;
 
         // Link user libraries
-        for library in &libraries.libraries {
-            assembler.link_dynamic_library(library).wrap_err("Failed to load libraries")?;
+        for library in libraries.libraries.iter().cloned() {
+            assembler
+                .link_package(library, Linkage::Dynamic)
+                .wrap_err("Failed to load libraries")?;
```
bobbinth, line 83: I know this was old behavior, but should we link user libraries statically? Or maybe better yet, maybe we should have a CLI parameter that specifies linkage for user libraries. Let's create an issue for this (unless we already have one).

crates/assembly/src/assembler/product.rs
```diff
@@ -1,68 +1,97 @@
...
 
 pub struct AssemblyProduct {
-    kind: TargetType,
-    artifact: Arc<Library>,
-    kernel: Option<Kernel>,
-    manifest: PackageManifest,
+    package: Box<Package>,
+    kernel_package: Option<Arc<Package>>,
     debug_info: Option<DebugInfoSections>,
 }
```
bobbinth, line 9: Could we add some documentation to this struct? Otherwise, it is not immediately obvious what different fields here mean (e.g., why is `debug_info` separate from `package` and why `kernel_package` is optional).

crates/mast-package/src/package/manifest.rs
```diff
@@ -24,16 +21,21 @@ use crate::{Dependency, PackageId};
 
 /// The manifest of a package, containing the set of package dependencies (libraries or packages)
 /// and exported items (procedures, constants, types), if known.
-#[derive(Debug, Default, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "arbitrary", derive(proptest_derive::Arbitrary))]
+///
+/// Exports declared in the package manifest are keyed by their fully-qualified path.
+///
+/// Dependencies must each specify a unique package identifier, i.e. it is not allowed to have
+/// multiple dependencies on the same package identifier, even if they are different versions.
```
bobbinth, line 28: Question: not being able to depend on multiple versions of the same package - is that a fundamental limitation, or maybe something we'd like to relax in the future?

crates/mast-package/src/package/manifest.rs
```diff
@@ -376,3 +445,54 @@ impl fmt::Debug for TypeExport {
             .finish()
     }
 }
+
+fn normalize_export(export: &mut PackageExport) -> Result<(), ManifestValidationError> {
```
bobbinth, line 449: This feels like it should be a method on `PackageExport`. Is there a reason to have it as a free function here?

Regardless, it would be great to add a doc comment explaining what normalization actually does.

crates/mast-package/src/package/manifest.rs
```diff
@@ -278,28 +301,74 @@ pub struct ProcedureExport {
...
+    ///
+    /// NOTE: While one might get the impression that `MastNodeId` is a unique identifier for each
+    /// procedure that gets assembled to the MAST, that isn't actually true. If multiple nodes have
+    /// the same digest _and_ the same decorators, they will have the same fingerprint, and thus
+    /// will be collapsed into a single node in the MAST and have the same `MastNodeId`.
+    ///
+    /// If this field contains `None`, the digest is used to resolve a MAST node.
+    #[cfg_attr(any(test, feature = "arbitrary"), proptest(value = "None"))]
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub node: Option<MastNodeId>,
```
bobbinth, line 329: I'm not sure I understood why `node` could be `None`. I thought every procedure export should have a corresponding node in the underlying MAST forest, including re-exported procedures which would have an `External` node in the forest. Did this behavior change somehow?

Separately, and not for this PR, I'm wondering if we should make all fields in this struct private and use accessor methods to read them. This is mostly to make sure we always get a valid `ProcedureExport` since this is a publicly-facing type.

crates/mast-package/src/package/mod.rs
bobbinth, line 75: Not for this PR, but having a mix of public and private fields here feels a bit inconsistent - I'd probably make them all private (and add relevant accessor methods).

crates/mast-package/src/package/mod.rs
```diff
@@ -44,61 +50,117 @@ use crate::{Dependency, Version};
...
-    #[cfg_attr(feature = "serde", serde(default))]
     pub sections: Vec<Section>,
 }
 
 /// Construction
 impl Package {
-    /// Construct a [Package] from a [Library] by providing the necessary metadata.
-    pub fn from_library(
+    /// Construct a [Package] from its essential component parts
+    pub fn create(
```
bobbinth, line 93: nit: should this be just `new()`?

crates/mast-package/src/package/mod.rs
```diff
@@ -107,9 +169,16 @@ impl Package {
     /// The file extension given to serialized packages
     pub const EXTENSION: &str = "masp";
 
+    /// Returns a reference to the MAST contained in this package
+    #[inline]
+    pub fn mast_forest(&self) -> &Arc<MastForest> {
```
bobbinth, line 174: nit: sometimes we call such methods `.mast()` and `.mast_forest()` - I think we should pick a name and use it consistently. I have a slight preference towards `.mast()`.

crates/mast-package/src/package/mod.rs
```diff
@@ -172,12 +241,134 @@ impl Package {
...
+        self.module_infos()
             .find(|mi| mi.path().is_kernel_path())
             .ok_or_else(|| Report::msg("invalid kernel package: does not contain kernel module"))
     }
 
+    /// If this package depends on a kernel, this method extracts the [Dependency] corresponding to
+    /// it.
+    ///
+    /// Returns `Err` if the dependency metadata for this package contains multiple kernels.
+    pub fn kernel_runtime_dependency(&self) -> Result<Option<&Dependency>, Report> {
```
bobbinth, line 253: Could we check this error condition at construction time? Basically, once a package is constructed, it would be good to enforce (as much as we can) that all dependencies are consistent/correct.

crates/mast-package/src/package/mod.rs
```diff
@@ -487,46 +621,38 @@ impl Package {
...
-    ) -> Box<Self> {
-        let library = arbitrary_library();
-
-        Self::from_library(name, version, kind, library, dependencies)
-    }
-}
+    #[cfg(feature = "std")]
+    pub fn deserialize_from_file(
+        path: impl AsRef<std::path::Path>,
+    ) -> Result<Self, miden_core::serde::DeserializationError> {
```
bobbinth, line 628: nit: we usually name such methods `read_from_file()`.

crates/mast-package/src/package/mod.rs
```diff
@@ -172,12 +241,134 @@ impl Package {
...
+    ///
+    /// This allows debuggers to resolve human-readable procedure names during execution.
+    pub fn procedure_name(&self, digest: &Word) -> Option<&str> {
+        self.mast.procedure_name(digest)
+    }
+
+    /// Returns an iterator over all (digest, name) pairs of procedure names.
+    pub fn procedure_names(&self) -> impl Iterator<Item = (Word, &Arc<str>)> {
+        self.mast.procedure_names()
+    }
```
bobbinth, line 281: Maybe be good to clarify what "name" here means. Is it a fully-qualified path? Or just the last leg of the path?

crates/mast-package/src/package/mod.rs
```diff
@@ -172,12 +241,134 @@ impl Package {
...
+    pub fn get_procedure_node_by_path(&self, path: impl AsRef<Path>) -> Option<MastNodeId> {
+        let path = path.as_ref().to_absolute();
+        self.manifest
+            .get_export(path.as_ref())
+            .and_then(PackageExport::as_procedure)
+            .and_then(|export| export.node.or_else(|| self.mast.find_procedure_root(export.digest)))
+    }
+
+    /// Returns an iterator over the module infos of the library.
+    pub fn module_infos(&self) -> impl Iterator<Item = ModuleInfo> {
```
bobbinth, line 328: Not for this PR, but I think we should rename `ModuleInfo` to something like `ModuleDescriptor` or `ModuleManifest`. I think we discussed it in some thread before - but not sure if we have an issue for this. May be a good thing to add to the list of follow-ups to this PR.

bobbinth (approved):
Looks good! Thank you! I left a bunch of comments inline - most of them are non-blocking and can be addressed in follow-up pRs.

crates/mast-package/src/package/mod.rs
bitwalker: I believe we already have an open issue for this

crates/mast-package/src/package/mod.rs
bitwalker: I was attempting to make the change here as minimal and mechanical as possible - so I kept the original name from 'Library`. I don't mind changing it here, but maybe some of those charges should go in a follow up PR?

crates/mast-package/src/package/mod.rs
bitwalker: I kept the `Library` naming, but I'm fine with changing it to `mast`

crates/mast-package/src/package/mod.rs
bitwalker: Mostly trying to avoid making unnecessary breaking changes to the `Package` API as part of this PR

crates/mast-package/src/package/manifest.rs
bitwalker: It's not fundamental, and only relevant for runtime-provided dependencies, so it complicates things there.

It also depends on the linker including version information in resolution.

miden-vm/src/cli/utils.rs
bitwalker: I don't think we should make linkage decisions like that at this level.

> have a CLI parameter that specifies linkage for user libraries. 

We have this in `midenc` already (and in `miden build` as a result. It's also exposed in the project file format.

processor/src/host/default.rs
bitwalker: Definitely

crates/assembly/src/testing.rs
bitwalker: I consider them interchangeable in this context - the package has to be a library package here

crates/assembly/README.md
bitwalker: I recently did this, it's covered in the `Assembler` docs now - unless you meant something more specific/detailed.

crates/mast-package/src/package/mod.rs
bobbinth: Yeah, this, and many other comments can be left for follow-up PRs.

crates/assembly/src/assembler.rs
bitwalker: It's a pragmatic choice - it is regularly needed for debugging assembly. We could probably add an adapter layer to globally control whether log statements like this actually reach an installed logger, but it just hasn't really been an issue so far. Eventually, I expect the compiler to be the only direct consumer of these APIs, once the switch to packages is pervasive.

crates/assembly/README.md
bobbinth: I meant specifically having a short section in this README file (could probably mostly copy-paste of docs form elsewhere).

crates/assembly/src/assembler.rs
bitwalker: The `Package` struct is far too large to be passing around by value - in practice is always going to be moved onto the heap anyway as an `Arc`, and converting from `Box` to `Arc` is free. The only reason we don't go direct to `Arc` is because we often want to append custom sections to the package after it is assembled.

bobbinth: Right, but the caller could just as easily convert it into `Arc`, `Box` or w/e else they need - and they could also pass by reference using just `&Package`. My point was that we could just return `Package` and leave it up to the caller to decide what they want to do with it.

bitwalker: > Right, but the caller could just as easily convert it into Arc, Box or w/e else they need - 

My point is that the callers are not only limited, but necessarily need to convert to `Arc` anyway - virtually all Package-oriented APIs want the `Arc`. We're not restricting callers from moving the `Package` to the stack if they really want it there for some reason, but that's an exception to the vast majority of uses (and I can't think of any reasons why one would want it unboxed anyway) - instead we're optimizing for the uses of these APIs that actually exist in practice.

This struct is too large to leave unboxed (and `clippy` will complain about it if it's in pedantic mode)

crates/mast-package/src/package/mod.rs
huitseeker: #2928

PR #3109, opened 2026-05-06
Bind MMR to num_leaves

huitseeker:
This fixes ~two~ one core library issue~s~.

- MMR pack/unpack commitments now include num_leaves, so the same padded peaks cannot be reused with a different leaf count.
 - ~~AEAD decrypt now checks that the advised plaintext re-encrypts to the exact input ciphertext and tag. A bad event handler can no longer swap in plaintext that only matches the tag.~~ (moved to #3147)
 
 Fixes 0xmiden/security-issues#18


core/src/lib.rs
```diff
@@ -85,13 +85,371 @@ pub mod serde {
...
+            use super::{BTreeMap, BTreeSet, Felt, InnerNodeInfo, MerklePath, Vec, Word, ZERO};
+            use crate::serde::{
+                ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
+            };
+
+            /// A fully materialized Merkle Mountain Range.
+            #[derive(Debug, Clone, Default)]
+            pub struct Mmr(miden_crypto::merkle::mmr::Mmr);
+
+            impl Mmr {
```
krushimir, line 118: nit/readability: maybe worth adding a small comment here about why we wrap the `miden-crypto` MMR types? I assume this is temporary until the matching `miden-crypto` commitment change is available here. 

Right now it's pretty clear in the context of this PR, but later people might wonder. A breadcrumb would be helpful.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 - [BREAKING] Bounded the live advice map by total field elements during execution; advice-provider setup now returns an error when initial advice exceeds this limit ([#3085](https://github.com/0xMiden/miden-vm/pull/3085)).
 - Rejected empty kernel packages before linking so malformed dependency metadata returns a structured package error instead of reaching the linker's non-empty-kernel assertion ([#3082](https://github.com/0xMiden/miden-vm/pull/3082)).
 - [BREAKING] Fixed project artifact reuse to ignore unrelated manifest fields, rejected private cross-module imports, and kept signature-only type imports live ([#3091](https://github.com/0xMiden/miden-vm/pull/3091)).
+- Bound MMR commitments to the leaf count and made AEAD decrypt verify the input ciphertext as well as the tag ([#3109](https://github.com/0xMiden/miden-vm/pull/3109)).
```
krushimir: Should this be marked as `[BREAKING]`? This seems similar to other changelog entries that mark hash/serialization/state compatibility changes as breaking.

crates/lib/core/tests/collections/mmr.rs
```diff
@@ -369,6 +370,84 @@ fn test_mmr_unpack_invalid_hash() {
     assert!(test.execute().is_err());
 }
 
+#[test]
+fn poc_mmr_unpack_peak_hash_does_not_bind_num_leaves() -> Result<(), MerkleError> {
```
krushimir: nit: could we rename this test to avoid the `poc_` prefix and reflect the fixed behavior, e.g. `fn test_mmr_unpack_rejects_forged_num_leaves()`? Since this is now a regression test for the fix, `poc_` makes it read a bit like the vulnerability is still present.

krushimir (approved):
LGTM! Left some non-blocking comments.

Al-Kindi-0 (commented):
Looks great!
The AEAD fix could probably be improved but shouldn't block merging this. The other question regarding MMR peaks I left in the accompanying `crypto` PR is what should be discussed

huitseeker:
@Al-Kindi-0 Agreed that the AEAD path could be made cleaner. I see this PR as the minimal VM-side hardening: the host still supplies plaintext through advice, but the MASM procedure now re-encrypts it and compares every ciphertext block, the encrypted padding block, and the tag against the caller-provided input. That closes the substitution issue without changing the event-handler/API shape. I’d prefer to track a cleaner/lower-cycle AEAD design separately unless there’s a specific change you want before merging.

Al-Kindi-0:
> @Al-Kindi-0 Agreed that the AEAD path could be made cleaner. I see this PR as the minimal VM-side hardening: the host still supplies plaintext through advice, but the MASM procedure now re-encrypts it and compares every ciphertext block, the encrypted padding block, and the tag against the caller-provided input. That closes the substitution issue without changing the event-handler/API shape. I’d prefer to track a cleaner/lower-cycle AEAD design separately unless there’s a specific change you want before merging.

Agreed, all good on my end on the AEAD front

huitseeker:
I’m going to close this without merging so the VM does not land a commitment format that crypto is about to replace.

The PR fixes the real issue: `pack`/`unpack` must bind `num_leaves`, and the current prefix format is a valid minimal security fix for 0xMiden/security-issues#18. The crypto-side discussion has moved the final release format toward a typed, domain-separated, versioned commitment API instead of:

```text
H([num_leaves, 0, 0, 0] || padded_peaks)
```

Since the VM core library and `miden-crypto` must compute exactly the same MMR commitment, I’d rather close this PR and reopen the VM-side patch once [the crypto issue](https://github.com/0xMiden/miden-vm/issues/3530) defines the canonical trait/API and domain value. The regression coverage from this PR remains the right coverage to preserve.

PR #3114, opened 2026-05-07
Deprecate post-last op decorators

### LINKED ISSUE #2633: Investigate removing `Processor::execute_end_of_block_decorators()`

huitseeker:
This closes #2964, and partially closes #2633. The post-last op decorators are not rejected outright yet, so as to not break e.g. the compiler, but they are logged as deprecated, and converted to after-exit decorators.

- cherry-picks the relevant PR #2964 commits with original authorship preserved
- ~normalizes legacy post-last op decorators into `after_exit`~ rejects post-last op decorators
- ~logs a deprecation warning that post-last op decorators will be rejected in a future release~
- rejects out-of-bounds decorator indices beyond the post-last sentinel
- validates linked/serialized decorator storage so malformed forests cannot retain post-last op decorators
- avoids false deprecation warnings from internal basic-block merging

### DIFF
--- a/core/src/mast/node/basic_block_node/mod.rs
+++ b/core/src/mast/node/basic_block_node/mod.rs
@@ -139,6 +139,7 @@ impl BasicBlockNode {
         if operations.is_empty() {
             return Err(MastForestError::EmptyBasicBlock);
         }
+        validate_decorator_indices_within_ops(operations.len(), &decorators)?;
 
         // Validate decorators list (only in debug mode).
         #[cfg(debug_assertions)]
@@ -1239,6 +1240,20 @@ pub(crate) fn validate_decorators(operations_len: usize, decorators: &DecoratorL
     }
 }
 
+fn validate_decorator_indices_within_ops(
+    operations_len: usize,
+    decorators: &DecoratorList,
+) -> Result<(), MastForestError> {
+    if let Some(&(operation_idx, _)) = decorators.iter().find(|(idx, _)| *idx >= operations_len) {
+        return Err(MastForestError::DecoratorOpIndexOutOfBounds {
+            operation_idx,
+            num_operations: operations_len,
+        });
+    }
+
+    Ok(())
+}
+
 /// Raw-indexed prefix: how many paddings strictly before raw index r
 ///
 /// This struct provides O(1) lookup for converting raw operation indices to padded indices.
@@ -1492,6 +1507,7 @@ impl BasicBlockNodeBuilder {
                 if operations.is_empty() {
                     return Err(MastForestError::EmptyBasicBlock);
                 }
+                validate_decorator_indices_within_ops(operations.len(), &decorators)?;
 
                 // Validate decorators list (only in debug mode).
                 #[cfg(debug_assertions)]
@@ -1610,6 +1626,7 @@ impl MastForestContributor for BasicBlockNodeBuilder {
                 if operations.is_empty() {
                     return Err(MastForestError::EmptyBasicBlock);
                 }
+                validate_decorator_indices_within_ops(operations.len(), &decorators)?;
 
                 // Validate decorators list (only in debug mode).
                 #[cfg(debug_assertions)]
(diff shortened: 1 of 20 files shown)


bitwalker:
I think as long as the assembler places trailing decorators in the `after_exit` set for the containing block (rather than attaching them to the last "real" op in the block), then this poses no problem for the compiler.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -33,6 +33,31 @@ mod tests;
...
+            debug_assert!(
+                idx < num_padded_ops,
+                "decorator op index {idx} exceeds padded op count {num_padded_ops}"
+            );
+            rest.push((idx, id));
+        }
+    }
+    let after_merged: Vec<DecoratorId> = sentinel.into_iter().chain(after_exit).collect();
+    (rest, after_merged)
+}
```
plafer: Still wrapping my head around when this is called with relation to everything else, given that it affects the fingerprint computation. If I understand Paul's [comment](https://github.com/0xMiden/miden-vm/pull/3114#issuecomment-4399302778), then we should be able to outright reject a `DecoratorList` with decorators in the "sentinel" position - and hence skip this merging phase completely (i.e. only validate that no decorators in the sentinel position exists). This would reduce the complexity quite a bit, (at least for me coming back to this part of the code base and needing to piece everything back together).

huitseeker:
@bitwalker @plafer Given @bitwalker 's [comment](https://github.com/0xMiden/miden-vm/pull/3114#issuecomment-4399302778), I have added a test for the trailing behavior in 920cc1f8f7806eba9f3bd1057a6e1c039f07112d and taken the more extensive variant of this PR, which rejects post-last op decorators. Will clean up the merge behavior next.

processor/src/fast/mod.rs
```diff
@@ -556,6 +556,17 @@ impl FastProcessor {
...
+        // Basic blocks may still carry legacy op-indexed decorators at the sentinel index (padded
+        // op count). New MAST merges those into `after_exit` at build time; execute them
+        // here first so order matches historical `execute_end_of_block_decorators` +
+        // `after_exit`.
+        if let MastNode::Block(bb) = node {
+            let num_ops = bb.num_operations() as usize;
+            for decorator in current_forest.decorators_for_op(node_id, num_ops) {
+                self.execute_decorator(decorator, host)?;
+            }
+        }
```
plafer: I think it's fair to assume no one will be using MAST forests that were compiled with previous versions at this point? And hence we could remove this

plafer (approved):
LGTM! It's very nice to finally have this cleaned up.

processor/src/execution/basic_block.rs
```diff
@@ -194,7 +194,7 @@ where
 #[inline(always)]
 pub(super) fn finish_basic_block<P, H, S, T>(
     state: &mut ExecutionState<'_, P, H, S, T>,
-    basic_block_node: &BasicBlockNode,
+    _basic_block_node: &BasicBlockNode,
     node_id: MastNodeId,
     current_forest: &Arc<MastForest>,
 ) -> ControlFlow<BreakReason>
```
bobbinth, line 200: Do we still need the `_basic_block_node` argument here?

bobbinth (approved):
Not a very in-depth review - but looks good! Thank you!

processor/src/execution/basic_block.rs
huitseeker: Integrated in the first commit of #3077

PR #3096, opened 2026-05-05
feat(verifier): outer-LogUp boundary in MASM + VLPI restructure

adr1anh:
## Summary

Implements the outer-LogUp boundary correction in the MASM recursive verifier and
restructures the public-input surface that surrounds it. The boundary identity
`c_total + aux_sum == 0` is now folded entirely on the operand stack from the
caller-facing reduced-inputs block; the non-fixed-length pieces of the public
input (program hash, transcript state, kernel-procedure digests) move out of the
FLPI region into a typed VLPI window that both the Rust prover and MASM verifier
absorb identically into the FS transcript.

Originally drafted as several smaller pieces; ships as one PR because the AIR
LogUp wiring, public-input layout, prover/verifier signatures, and MASM streaming
loop are mutually load-bearing.

### Public input layout

- `program_hash` and `transcript_state` move from FLPI to VLPI. FLPI shrinks to
  stack i/o (32 felts).
- New 16-felt caller-facing reduced-inputs block at `reduced_inputs_ptr`:
  `[kernel_H | zero_pad | program_digest | transcript_state]`. The zero pad
  keeps the FS rate-aligned across two `mem_stream + permute` chunks.
- New `kernel_H = Poseidon2::hash_elements(...)` summary over all kernel-
  procedure digests. Per-digest absorption is `[d0, d1, d2, d3, 0, 0, 0, 0]`
  — one permute per digest, 4 advice felts per digest (canonical, no pad in the
  tape).

### Typed observation, simplified signatures

- `PublicInputs::observe<C: CanObserve<Felt>>(challenger)` mirrors the MASM
  verifier's FS schedule exactly: kernel_H + zero pad (8) → program_digest +
  transcript_state (8) → stack i/o (32). Six rate-aligned permutes total.
- `prove_stark` / `verify_stark_proof` simplified to
  `(config, pub_inputs, trace_or_proof)`; they derive `(public_values, vlpi)`
  internally via `pub_inputs.to_air_inputs()`.
- The slice-shaped `observe_var_len_public_inputs` plumbing is gone.

### MASM kernel-digest streaming

- Per-iteration body: `padw padw + adv_loadw + dupw + horner4_base + hperm`.
  No `adv_pipe`, `mem_stream`, `horner_eval_base`, or `tmp1` scratch. The full
  sponge state lives on the stack across iterations; capacity carries between
  iters, R0/R1 are rebuilt from advice each iter, and the final R0 is squeezed
  after the loop.
[...]

PR #3115, opened 2026-05-07
feat: multi-AIR support

### LINKED ISSUE #3032: Tracking Issue: Split main and chiplets trace
This issue tracks issues related to transitioning Miden VM to a multi-trace AIR (main + chiplets)

Nashtare:
## Summary                                                                                                                                                                                                                                           

Add multi-AIR support.
closes #3032 

- Splits `ProcessorAir` into `CoreAir` and `ChipletsAir`, and switches the prover/verifier to `prove_multi`/`verify_multi`.                                                                               
- Each AIR commits its own power-of-two trace height, so the prover stops padding both halves to the tallest one.
- MASM recursive verifier migrated

## Performance

Ran 2 versions of basic programs (`fibonacci` and `hperm`) at different sizes on my M4 pro.

| Program | `next` | this branch | Δ time | `next` proof | this branch proof | Δ size |                                                                                                                                                                   
  | --- | ---: | ---: | ---: | ---: | ---: | ---: |      
  | fibonacci_150 | 20.3 ms | 24.5 ms | +21% | 49.3 KB | 51.8 KB | +5.1% |                                                                                                                                                                          
  | fibonacci_1000 | 69.3 ms | 60.3 ms | -13% | 70.4 KB | 72.3 KB | +2.8% |                                                                                                                                                                         
  | hperm_x100 | 39.2 ms | 32.8 ms | -16% | 61.3 KB | 63.9 KB | +4.3% |                                                                                                                                                                             
  | hperm_x500 | 205.7 ms | 128.9 ms | -37% | 83.5 KB | 85.9 KB | +3.0% |
  
When running 
```bash
make exec-info;
MIDEN_LOG=info ./target/optimized/miden-vm prove ./miden-vm/masm-examples/hashing/blake3_1to1/blake3_1to1.masm --release --hasher poseidon2
```

I'm getting ~10.5% performance improvement on my M4 pro.
  
                                                                                                                                                                                                
 ## Test plan                                              

  - [x] cargo test -p miden-air --release — all symbolic AIR validations pass for CoreAir and ChipletsAir independently.                                                                                                                                
  - [x] cargo test -p miden-vm --release prove_verify — full Rust prove → verify cycle (14/14), including new fixtures:
    - [x] test_poseidon2_prove_verify_rust_only — multi-AIR Rust round-trip.                                                                                                                                                                            
    - [x] test_equal_heights_recursive — both AIRs at MIN_TRACE_LEN; exercises the (log_height, caller_index) tie-break in the MASM air_order reconstruction.                                                                                           
    - [x] test_hash_heavy_divergent_heights — chip_height > core_height; exercises the OOD shuffle and per-AIR-height boundary handling on the SLICED Core trace.                                                                                       
  - [x] cargo test -p miden-core-lib --test core-lib --release — recursive MASM verifier passes on every fixture (350/350), including all multi-AIR regression programs.                                                                                
  - [x] RELATION_DIGEST snapshot regenerated; matches between Rust (air/src/config.rs) and MASM (crates/lib/core/asm/sys/vm/mod.masm).                                                                                                                  
  - [x] Synthetic benches confirm prover speed-up on hash-heavy workloads where chip_height > core_height and a small regression-free path on Fibonacci-shape workloads where core_height > chip_height.


air/src/constraints/lookup/chiplet_air.rs
```diff
@@ -136,14 +162,35 @@ where
...
-        emit_hash_kernel_table::<LB>(builder, &ctx);
-        emit_v_wiring::<LB>(builder, &ctx);
+        emit_chiplet_lookup_columns(builder, local_chiplet, next_chiplet, clk_plus_one);
     }
 }
+
+/// Emit the three chiplet-trace LogUp columns (responses, hash-kernel virtual table,
+/// wiring) using a caller-supplied responder address.
+///
+/// Shared between [`ChipletLookupAir`]'s [`LookupAir`] impl (`clk_plus_one` from the Core
```
huitseeker: This looks stale now: `ChipletLookupAir::eval` above also passes `local_chiplet.chip_clk`, so both callers source the responder address from `chip_clk`. The older `system.clk + 1` wording also appears in the `ChipletBusContext` doc.

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -201,32 +299,62 @@ mod recursive_verifier {
         let mut challenger = config.challenger();
         config::observe_protocol_params(&mut challenger);
         challenger.observe_slice(&public_values);
-        let var_len_public_inputs: &[&[Felt]] = &[&kernel_felts];
-        config::observe_var_len_public_inputs(&mut challenger, var_len_public_inputs, &[WORD_SIZE]);
+        let chiplets_var_len: &[&[Felt]] = &[&kernel_felts];
+        config::observe_var_len_public_inputs(&mut challenger, chiplets_var_len, &[WORD_SIZE]);
+
+        let proof: miden_crypto::stark::proof::StarkProof<Felt, Challenge, P2Config> =
```
huitseeker: This block is now very close to `crates/lib/core/tests/stark/verifier_recursive/mod.rs` too: deserialize proof, seed the challenger, replay the transcript, recover per-AIR heights, pack advice, and build Merkle data. Could the integration test reuse that helper, or could the common advice builder move into a small test-support module?

verifier/src/lib.rs
```diff
@@ -211,18 +214,85 @@ where
...
+        return Err(StarkVerificationError::NonCanonicalAirOrder {
+            expected: vec![0, 1],
+            actual: proof_air_order.to_vec(),
+        });
+    }
+
+    let mut caller_heights = [0usize; 2];
+    let mut seen = [false; 2];
+    for (&caller_idx, &log_h) in proof_air_order.iter().zip(log_trace_heights) {
+        let Some(seen_slot) = seen.get_mut(caller_idx as usize) else {
```
huitseeker: Small tidy: this function repeats the same `NonCanonicalAirOrder { expected: vec![0, 1], actual: proof_air_order.to_vec() }` construction several times. A local helper/closure like `let non_canonical = || ...` would make the actual validation easier to scan. Since `proof_shapes` is parsed here, it could also get `proof_air_order` from `proof_shapes.air_order()` instead of taking it separately.

huitseeker (approved):
Overall, LGTM with some nits inline

CHANGELOG.md
```diff
@@ -69,6 +69,16 @@
...
+- Made all internal `core::math` procedures natively little-endian ([#3084](https://github.com/0xMiden/miden-vm/pull/3084)).
+- [BREAKING] Updated the Miden crypto stack to `miden-crypto` 0.25, and switched SMT leaf hashing to use Poseidon2 domain separation so masm-side leaf digests match `SmtLeaf::hash()` ([#3095](https://github.com/0xMiden/miden-vm/pull/3095)).
+- [BREAKING] Split the execution AIR into Core + Chiplets AIRs
+- Follow-up refactoring + couple perf improvements on trace generation ([#2953](https://github.com/0xMiden/miden-vm/pull/2953)).
+
+## 0.22.1
+
+#### Enhancements
+
+- Added `FastProcessor::into_parts()` to extract advice provider, memory, and precompile transcript after step-based execution ([#2901](https://github.com/0xMiden/miden-vm/pull/2901)).
```
huitseeker: This should probably be rebased/cleaned up.

Nashtare: Should be good now!

air/src/constraints/columns.rs
```diff
@@ -36,15 +36,22 @@ use crate::trace::{CHIPLETS_WIDTH, TRACE_WIDTH};
...
     pub system: SystemCols<T>,
     pub decoder: DecoderCols<T>,
     pub stack: StackCols<T>,
     pub range: RangeCols<T>,
-    pub(crate) chiplets: [T; CHIPLETS_WIDTH - 1],
+    pub(crate) chiplets: [T; CHIPLETS_WIDTH - 2],
     /// Permutation segment selector: consumed by `build_chiplet_selectors`.
     pub s_perm: T,
+    /// Chiplet-trace row counter: starts at 1 on the first row, increments by 1 each row.
+    pub chip_clk: T,
```
adr1anh: These should be probably be in the chiplets array.

air/src/constraints/columns.rs
```diff
@@ -99,26 +106,213 @@ impl<T> MainCols<T> {
...
+///
+/// `chip_clk` is the chiplet-trace row counter — see the field doc on `MainCols::chip_clk`
+/// for full semantics.
+#[repr(C)]
+pub struct ChipletCols<T> {
+    pub(crate) chiplets: [T; CHIPLETS_WIDTH - 2],
+    /// Permutation segment selector: consumed by `build_chiplet_selectors`.
+    pub s_perm: T,
+    /// Chiplet-trace row counter: starts at 1 on the first row, increments by 1 each row.
+    pub chip_clk: T,
```
adr1anh, line 80: Same as above. For a more general refactor, I'd move them to the front of the array (unfortunately having to modify all the column indices), rename `s_perm -> s_00` and `s0 -> s01`.

air/src/constraints/lookup/chiplet_air.rs
```diff
@@ -61,30 +64,49 @@ pub(crate) trait ChipletLookupBuilder: LookupBuilder<F = Felt> {
...
     /// Build the shared chiplet-trace context for one `eval` call.
-    pub fn new(builder: &LB, local: &'a MainCols<LB::Var>, next: &'a MainCols<LB::Var>) -> Self {
+    ///
+    /// Takes per-row `&ChipletCols` views and an externally-supplied `clk_plus_one` value
+    /// (the chiplet-trace `chip_clk` column).
+    pub fn new(
+        builder: &LB,
+        local: &'a ChipletCols<LB::Var>,
+        next: &'a ChipletCols<LB::Var>,
+        clk_plus_one: LB::Expr,
```
adr1anh: Why isn't it this in the ChipletsCols?

air/src/constraints/stack/stack_arith/tests.rs
```diff
@@ -128,7 +128,7 @@ fn set_u32_helpers(row: &mut MainCols<Felt>, lo: u32, hi: u32) {
 fn eval_stack_arith(local: &MainCols<Felt>, next: &MainCols<Felt>) -> Vec<QuadFelt> {
     let mut builder = ConstraintEvalBuilder::new();
     let op_flags = OpFlags::new(&local.decoder, &local.stack, &next.decoder);
```
adr1anh: I think the opflags are computed by the parent, so this could be passed as argument.

air/src/lib.rs
adr1anh: This doesn't seem correct. With two traces, we have two boundary aux values which must sum to zero with the external verifier inputs.

air/src/lib.rs
```diff
@@ -549,3 +558,518 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+        1
+    }
+
+    fn num_var_len_public_inputs(&self) -> usize {
+        // Kernel digests boundary-cancel against the kernel-rom bus, which lives on
+        // `CHIPLET_COLUMN_SHAPE[0]` — owned by ChipletsAir.
+        1
+    }
+
+    fn reduced_aux_values(
```
adr1anh: The aux values reduction is going to change in the next lifted stark. Morally, it's a global check rather than a per-Air one. It's probably simpler to put the entire check in the ProcessorAir.

air/src/lib.rs
```diff
@@ -549,3 +558,518 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+    }
+
+    fn num_var_len_public_inputs(&self) -> usize {
+        match self {
+            Self::Core(a) => <CoreAir as LiftedAir<Felt, EF>>::num_var_len_public_inputs(a),
+            Self::Chiplets(a) => <ChipletsAir as LiftedAir<Felt, EF>>::num_var_len_public_inputs(a),
+        }
+    }
+
+    fn reduced_aux_values(
```
adr1anh, line 576: As mentioned above, we should put all the logic here (making sure its only run once though), since well put the reduction in a separate trait. Happy to discuss this in more detail async.

air/src/constraints/chiplets/mod.rs
```diff
@@ -22,23 +23,26 @@ pub mod selectors;
...
     selectors: &ChipletSelectors<AB::Expr>,
 ) where
     AB: MidenAirBuilder,
 {
     // Selector constraints (including hasher internal selectors) are enforced in
     // build_chiplet_selectors (called from lib.rs).
 
+    // Chiplet-trace row counter (boundary + transition).
+    chip_clk::enforce_chip_clk_constraints(builder, local, next);
+
```
adr1anh: I would inline this to avoid too much indirection

air/src/constraints/lookup/miden_air.rs
```diff
@@ -33,30 +33,41 @@ pub(crate) const MIDEN_COLUMN_SHAPE: [usize; 7] = [
...
-///   through a reducer that sums `Σ multiplicity / encode(msg)`).
+/// Thin wrapper around [`emit_core_boundary`] + [`emit_chiplets_boundary`]; used by
+/// `ProcessorAir::reduced_aux_values` and the debug walker. Per-AIR `reduced_aux_values`
+/// impls call only their half (see the per-bus partition in the helpers below).
 ///
 /// See `program_hash_message`, `transcript_messages`, and `kernel_proc_message` in
 /// `air/src/lib.rs` for the canonical formulas this mirrors.
 pub(crate) fn emit_miden_boundary<B: BoundaryBuilder>(boundary: &mut B) {
+    emit_core_boundary(boundary);
+    emit_chiplets_boundary(boundary);
```
adr1anh: let's inline everything here like before.

air/src/constraints/columns.rs
```diff
@@ -36,15 +36,22 @@ use crate::trace::{CHIPLETS_WIDTH, TRACE_WIDTH};
 ///
 /// The `s_perm` column is separated from the chiplets array because it is consumed
 /// exclusively by the chiplet selector system, not by any chiplet's constraint code.
+///
+/// `chip_clk` is the chiplet-trace row counter (value `row_index + 1`); it serves as the
+/// chiplet-side responder address for the hasher LogUp bus. Required for the multi-AIR
+/// split — eliminates the cross-trace `system.clk + 1` read that appeared in the legacy
+/// monolithic chiplet bus emitter.
```
adr1anh: This seems too specific for a docstring comment.

air/src/constraints/columns.rs
```diff
@@ -36,15 +36,22 @@ use crate::trace::{CHIPLETS_WIDTH, TRACE_WIDTH};
...
 #[repr(C)]
 pub struct MainCols<T> {
```
adr1anh: Is there a reason why we still need this struct at all?

air/src/trace/main_trace.rs
```diff
@@ -100,7 +100,19 @@ enum TraceStorage {
         core_rm: Vec<Felt>,
         chiplets_rm: Vec<Felt>,
         range_checker_cols: [Vec<Felt>; 2],
+        /// Unified storage height: `max(core_rows, chiplets_rows)`. Used for the
+        /// `to_row_major` view that callers like `check_constraints` and the
```
adr1anh: Shouldn't `check_constraints` work on the two traces separately?

air/src/trace/main_trace.rs
```diff
@@ -100,7 +100,19 @@ enum TraceStorage {
...
+        /// in `to_core_chiplets_matrices`. `core_rows ≤ num_rows`; the rows
+        /// beyond `core_rows` in `core_rm` and `range_checker_cols` are
+        /// zero-padded.
+        core_rows: usize,
+        /// Per-AIR height for the Chiplets matrix in
+        /// `to_core_chiplets_matrices`. `chiplets_rows ≤ num_rows`; the rows
+        /// beyond `chiplets_rows` in `chiplets_rm` are zero-padded.
+        chiplets_rows: usize,
     },
     RowMajor(RowMajorMatrix<Felt>),
```
adr1anh: I don't think we need this anymore either.

air/src/constraints/lookup/buses/chiplet_responses.rs
```diff
@@ -126,8 +126,9 @@ pub(in crate::constraints::lookup) fn emit_chiplet_responses<LB>(
     // --- Emit everything into a single LogUp column ---
 
     // All hasher response variants encode their row at `clk + 1` (so they cancel against
-    // the matching request at `clk`).
-    let clk_plus_one: LB::Expr = local.system.clk.into() + LB::Expr::ONE;
+    // the matching request at `clk`). Sourced from the context — see
+    // `ChipletBusContext::clk_plus_one` for the cross-trace bookkeeping.
+    let clk_plus_one: LB::Expr = ctx.clk_plus_one.clone();
```
adr1anh: Get this from the local row

air/src/constraints/lookup/chiplet_air.rs
```diff
@@ -61,30 +64,49 @@ pub(crate) trait ChipletLookupBuilder: LookupBuilder<F = Felt> {
...
-    pub next: &'a MainCols<LB::Var>,
+    /// Typed view of the current row (Chiplets half of the trace).
+    pub local: &'a ChipletCols<LB::Var>,
+    /// Typed view of the next row (Chiplets half of the trace).
+    pub next: &'a ChipletCols<LB::Var>,
     /// Per-chiplet `is_active` flags, computed from `local`'s selector columns via the
     /// builder-provided hook.
     pub chiplet_active: ChipletActiveFlags<LB::Expr>,
+    /// Hasher response address, sourced from the chiplet-trace `chip_clk` column.
+    pub clk_plus_one: LB::Expr,
```
adr1anh: remove

air/src/constraints/lookup/chiplet_air.rs
```diff
@@ -61,30 +64,49 @@ pub(crate) trait ChipletLookupBuilder: LookupBuilder<F = Felt> {
...
 
 impl<'a, LB> ChipletBusContext<'a, LB>
 where
     LB: ChipletLookupBuilder,
 {
     /// Build the shared chiplet-trace context for one `eval` call.
-    pub fn new(builder: &LB, local: &'a MainCols<LB::Var>, next: &'a MainCols<LB::Var>) -> Self {
+    ///
+    /// Takes per-row `&ChipletCols` views and an externally-supplied `clk_plus_one` value
+    /// (the chiplet-trace `chip_clk` column).
```
adr1anh: Let's make sure to remove all the explicit mentions of clk_plus_one as these add a lot of overhead to the comments

air/src/constraints/lookup/miden_air.rs
```diff
@@ -33,30 +33,41 @@ pub(crate) const MIDEN_COLUMN_SHAPE: [usize; 7] = [
...
-/// Only col 0 is a real committed final; slot 1 is a placeholder forced to zero, kept for
-/// forward-compatibility with the MASM recursive verifier (which absorbs 2 boundary
-/// values). All paths that emit or consume the pair must preserve the zero in slot 1.
-///
-/// TODO(#3032): reduce to 1 once trace splitting lands and each sub-trace has its own
-/// accumulator.
+/// `ProcessorAir` (the legacy aggregator) has a single real LogUp accumulator; slot 1 is
+/// forced to zero so its layout matches the multi-AIR proof shape (Core final at slot 0,
+/// Chiplets final at slot 1). All paths that emit or consume the pair must preserve the
+/// zero in slot 1 for `ProcessorAir`.
```
adr1anh: Is it necessary to mention the previous behavior?

air/src/constraints/columns.rs
```diff
@@ -99,26 +106,213 @@ impl<T> MainCols<T> {
...
 }
 
 impl<T> Borrow<MainCols<T>> for [T] {
     fn borrow(&self) -> &MainCols<T> {
-        debug_assert_eq!(self.len(), TRACE_WIDTH);
-        let (prefix, shorts, suffix) = unsafe { self.align_to::<MainCols<T>>() };
-        debug_assert!(prefix.is_empty() && suffix.is_empty() && shorts.len() == 1);
+        // The prover stores rows at `PADDED_TRACE_WIDTH` (next multiple of 8) for
+        // `adv_pipe` rate-8 hashing. `MainCols` itself models only `TRACE_WIDTH` columns;
+        // any trailing padding is ignored.
```
adr1anh: I thought we had removed PADDED_TRACE_WIDTH.

air/src/trace/main_trace.rs
```diff
@@ -134,23 +146,40 @@ impl MainTrace {
...
+    ///
+    /// `core_rows` and `chiplets_rows` are the per-AIR padded heights returned by
+    /// `to_core_chiplets_matrices`. Both must be powers of two, both ≤ `num_rows`.
+    /// The data arrays are sized to `num_rows`; rows beyond each per-AIR height are
+    /// expected to be zero-padded.
     pub fn from_parts(
         core_rm: Vec<Felt>,
         chiplets_rm: Vec<Felt>,
         range_checker_cols: [Vec<Felt>; 2],
         num_rows: usize,
```
adr1anh: We can derive this from `max(core_rows, chiplets_rows)`

air/src/trace/main_trace.rs
```diff
@@ -275,6 +308,98 @@ impl MainTrace {
         }
     }
 
+    /// Splits the trace into the per-AIR `(Core, Chiplets)` matrix pair used by the multi-AIR
+    /// proving path.
+    ///
+    /// - **Core matrix** is `NUM_CORE_COLS = 51` wide and `core_rows` tall (the leading system +
+    ///   decoder + stack + range columns).
+    /// - **Chiplets matrix** is `CHIPLETS_WIDTH` wide and `chiplets_rows` tall (chiplet data +
+    ///   `s_perm` + `chip_clk`).
```
adr1anh: No need to explain this in the comments, as they should really be merged in the chiplets section

air/src/config.rs
```diff
@@ -109,6 +109,15 @@ pub fn observe_protocol_params(challenger: &mut impl CanObserve<Felt>) {
     challenger.observe(Felt::ZERO);
 }
 
+/// Absorbs the multi-AIR `air_order` permutation into the challenger.
+/// Required by `miden-lifted-stark`: its `observe_heights` binds per-AIR heights but
+/// not the proof_order permutation, so the caller must bind it.
+pub fn observe_air_order<C: CanObserve<Felt>>(challenger: &mut C, air_order: &[u32]) {
+    for &caller_idx in air_order {
+        challenger.observe(Felt::new_unchecked(caller_idx as u64));
```
adr1anh: I think we should align this to the rate (i.e. pad with zeros)

adr1anh (changes requested):
Not a full review yet (I haven't checked the MASM), but I think there is still some opportunity to simplify this PR. It seems like we still have a lot of code which would be necessary to support a single trace regime with both sub-traces concatenated. For instance, `MainCols` doesn't seem like something we would still need to use. 

This PR is also highlighting a lot of short-commings with the existing Lifted STARK API, and awkwardness around handling the auxiliary inputs. I'm taking note of them, but would be happy to discuss this further so we can try to evolve it to better match this actual use case. A non-exhaustive list of my own observations:
- For a multi-trace instance, we should only have a single list of "public inputs", i.e. the list of values passed to the Air. There isn't a strong reason why different Airs need their own list. 
- In a similar way, we should make `aux_public_inputs` a single list of values (or list of list, since each trace produces different aux values). These are only given to the external verifier. 
- Combining both of the above together, this means that we should probably separate the public inputs verification into a single trait, making it more explicit that this is a global check rather than a per-Air one.

I think this is all pointing towards a need of a `PublicInputs` trait, which would
- own the air/aux public inputs (each `&[F]`)
- provide a method for observing itself into the challenger
- describe how to evaluate the boundary check, when given the challenges and the prover-supplied aux values.

I also wanted to link to https://github.com/0xMiden/miden-vm/pull/3096, which might be interesting to integrate in this PR instead. The current ACE handling of the VLPI is quite hack-y, and there are clear benefits of handling the logUp boundary reduction in the recursive verifier directly. 

Happy to go into more detail async!

air/src/lib.rs
Nashtare: The ProcessorAir logic is kinda legacy now, and could be part of a larger refactoring / cleanup (that I could have allegedly done here although I felt the changes were already quite massive) but the global identity you describe is already implemented in the multi-air path where are aggregate the two `ReducedAuxValues.sum`

Although it is a bit fragile and your other comment about a global PublicInput trait would address

air/src/lib.rs
Nashtare: Agreed, although for the reason above I'd rather not patch the ProcessorAir but instead remove it, add the PublicInputs trait you mentioned, and address this

air/src/trace/main_trace.rs
Nashtare: Yes, this one runs against ProcessorAir (see comments above) but I'll migrate that to a multi-air helper

air/src/lib.rs
adr1anh: How big a refactor are we talking? I’m a bit weary of having unused code lingering, especially if it duplicates "complicated" stuff.

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+    }
+}
+
+// CHIPLETS AIR
+// ================================================================================================
+
+/// Chiplets-trace AIR for the multi-AIR proving path.
+///
+/// Owns the 22-column chiplet section (chiplets data + s_perm + chip_clk) and the three
+/// chiplet-trace LogUp accumulator columns. Counterpart to [`CoreAir`].
```
adr1anh: A bit too specific about the struct's fields

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+        let next: &ChipletCols<_> = main.next_slice().borrow();
+
+        constraints::lookup::chiplet_air::emit_chiplet_lookup_columns(builder, local, next);
+    }
+
+    fn eval_boundary<B>(&self, boundary: &mut B)
+    where
+        B: BoundaryBuilder<F = LB::F, EF = LB::EF>,
+    {
+        constraints::lookup::miden_air::emit_chiplets_boundary(boundary);
```
adr1anh, line 505: I would move this to the CoreAir, mentioning with a TODO that we do the check there until we have the new interfaces.

air/src/lib.rs
adr1anh: > Agreed, although for the reason above I'd rather not patch the ProcessorAir but instead remove it, add the PublicInputs trait you mentioned, and address this

I think I meant MidenAir, since that is the "global" air. Consolidating the variable length public check in there is much closer to what it will look like with the upcoming `Instance` trait.

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+            Self::Core(a) => BaseAir::<Felt>::width(a),
+            Self::Chiplets(a) => BaseAir::<Felt>::width(a),
+        }
+    }
+
+    fn num_public_values(&self) -> usize {
+        match self {
+            Self::Core(a) => BaseAir::<Felt>::num_public_values(a),
+            Self::Chiplets(a) => BaseAir::<Felt>::num_public_values(a),
+        }
```
adr1anh: With the `Instance` abstraction, all traces share the same "Air public inputs" which in our case is basically just the stack IO (we still include the program hash and precompile state because we used to do bus boundary checks in the Air's running sum/product in the last/first row). I'd want to make this point a bit clearer already, in that both airs must have the same number of public inputs.

Could we for now add an assert that the number of public inputs match the constant?

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+
+/// Homogeneous wrapper that lets [`CoreAir`] and [`ChipletsAir`] share a single trait-object
+/// type for `prove_multi`/`verify_multi`. Upstream's `prove_multi<F, EF, A, B, SC>` takes
+/// `&[(&A, AirWitness, &B)]` — both `A` (the AIR) and `B` (the aux builder) must be the same
+/// type across all instances, so we dispatch through this enum.
+#[derive(Copy, Clone, Debug)]
+pub enum MidenAir {
+    Core(CoreAir),
+    Chiplets(ChipletsAir),
+}
```
adr1anh, line 520: I’m wondering if we really need Core and Chiplets Airs to be Airs themselves. As we've seen, some of the Air methods are awkward to implement on a sub trace, and could be consolidated into a singular MidenAir method. I'll point out which of the methods below we could do this for. 

We could then remove the LiftedAir implementation on the sub traces, and dispatch from the MidenAir to the struct (not trait) methods which are conceptually different.

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+
+impl<EF: ExtensionField<Felt>> LiftedAir<Felt, EF> for MidenAir {
+    fn periodic_columns(&self) -> Vec<Vec<Felt>> {
+        match self {
+            Self::Core(a) => <CoreAir as LiftedAir<Felt, EF>>::periodic_columns(a),
+            Self::Chiplets(a) => <ChipletsAir as LiftedAir<Felt, EF>>::periodic_columns(a),
+        }
+    }
+
+    fn num_randomness(&self) -> usize {
```
adr1anh, line 548: This one is global, and could be removed from the sub trace Airs

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+    }
+
+    fn num_aux_values(&self) -> usize {
+        match self {
+            Self::Core(a) => <CoreAir as LiftedAir<Felt, EF>>::num_aux_values(a),
+            Self::Chiplets(a) => <ChipletsAir as LiftedAir<Felt, EF>>::num_aux_values(a),
+        }
+    }
+
+    fn num_var_len_public_inputs(&self) -> usize {
```
adr1anh, line 565: This one is global, and could be removed from the sub trace Airs

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+        match self {
+            Self::Core(a) => <CoreAir as LiftedAir<Felt, EF>>::num_var_len_public_inputs(a),
+            Self::Chiplets(a) => <ChipletsAir as LiftedAir<Felt, EF>>::num_var_len_public_inputs(a),
+        }
+    }
+
+    fn reduced_aux_values(
```
adr1anh, line 576: This one is global, and could be removed from the sub trace Airs

air/src/lib.rs
```diff
@@ -549,3 +558,508 @@ impl<'a, EF: ExtensionField<Felt>> BoundaryBuilder for ReduceBoundaryBuilder<'a,
...
+            Self::Chiplets(a) => <ChipletsAir as LiftedAir<Felt, EF>>::log_quotient_degree(a),
+        }
+    }
+}
+
+impl<EF> AuxBuilder<Felt, EF> for MidenAir
+where
+    EF: ExtensionField<Felt>,
+{
+    fn build_aux_trace(
```
adr1anh, line 668: Consolidating both paths into this method would remove quite a few lines of code (ie removing the other AuxBuilder impls)

adr1anh (commented):
I’ve left a few more comments around the Air crate. Overall, I'll push for removing the now-legacy ProcessorAir in this PR. It'll reduce the diff and make the multi trace path clearer.

Second, I think at this point we would gain in clarity by **_not_** implementing LiftedAir for CoreAir and ChipletsAir, and instead only doing for MidenAir. The trait implementations would be relegated to struct implementations, and we would remove any that don't make sense (ie public inputs/variable length public inputs stuff). This should bring us a lot closer to the upcoming `Instance`.

air/src/lib.rs
Nashtare: It's about 800L of removed code, I've added it on top of additional cleanups.

air/src/constraints/columns.rs
Nashtare: This is going to be quite noisy, I'd rather hold off and do it on a follow-up PR if you don't mind

Al-Kindi-0:
@Nashtare , what is the status so far ? I was thinking of adding a few commits on top with some tweaks to the MASM verifier but I think it might be better to get this one in asap, wdyt ?

Nashtare:
@Al-Kindi-0 Agree this is quite cumbersome as blocking a couple parallel works building on top. I think @adr1anh wanted to give the MASM a proper review before considering this done and doing the remaining refactoring pieces + follow-up iterations on top, but I may have misunderstood?

air/src/constraints/columns.rs
adr1anh: Can we create an issue for this?

air/src/constraints/lookup/miden_air.rs
```diff
@@ -1,62 +1,33 @@
...
+/// accumulators on Core columns:
+/// - `BlockHashTable` lives on `MAIN_COLUMN_SHAPE[1]` (block_hash + op_group merged column).
+/// - `LogPrecompileTranscript` lives on `MAIN_COLUMN_SHAPE[0]` (block_stack + range + log-cap
+///   merged column).
 ///
-/// See `program_hash_message`, `transcript_messages`, and `kernel_proc_message` in
-/// `air/src/lib.rs` for the canonical formulas this mirrors.
-pub(crate) fn emit_miden_boundary<B: BoundaryBuilder>(boundary: &mut B) {
+/// Both fractions therefore belong to `CoreAir::reduced_aux_values` post-split.
+pub(crate) fn emit_core_boundary<B: BoundaryBuilder>(boundary: &mut B) {
```
adr1anh, line 30: Is it possible to consolidate them into a single `emit_boundary`?

air/src/lib.rs
```diff
@@ -249,84 +248,228 @@ pub const LOGUP_AUX_TRACE_WIDTH: usize = 7;
...
-/// independently computes expected boundary messages from variable length public values and checks
-/// them against the final column values.
-#[derive(Copy, Clone, Debug, Default)]
-pub struct ProcessorAir;
+/// Denominators are `α + Σ βⁱ · field_i` with random `α, β`; on any legitimate proof they
+/// are non-zero with overwhelming probability. A malformed/adversarial proof can still
+/// drive a denominator to zero, so the reducer captures the first failure and surfaces it
+/// to `reduced_aux_values`, which bubbles a [`ReductionError`] to the verifier rather than
+/// panicking.
+struct ReduceBoundaryBuilder<'a, EF: ExtensionField<Felt>> {
```
adr1anh: I'd move this lower down where it was before to reduce the diff

air/src/trace/main_trace.rs
```diff
@@ -92,22 +92,23 @@ impl<T> BorrowMut<MainTraceRow<T>> for [T] {
...
-    },
-    RowMajor(RowMajorMatrix<Felt>),
-    Transposed {
-        matrix: RowMajorMatrix<Felt>,
-        num_cols: usize,
-        num_rows: usize,
-    },
+struct TraceStorage {
+    core_rm: Vec<Felt>,
+    chiplets_rm: Vec<Felt>,
```
adr1anh: This could/should be a RowMajorMatrix right?

air/src/constraints/columns.rs
Nashtare: https://github.com/0xMiden/miden-vm/issues/3165

Al-Kindi-0:
> @Al-Kindi-0 Agree this is quite cumbersome as blocking a couple parallel works building on top. I think @adr1anh wanted to give the MASM a proper review before considering this done and doing the remaining refactoring pieces + follow-up iterations on top, but I may have misunderstood?

Agreed, I think it is better, if possible, to do any big remaining improvements/fixes post merge. I would focus on the non-MASM part to speed things up as well

crates/ace-codegen/src/layout/keys.rs
```diff
@@ -37,6 +45,16 @@ pub enum InputKey {
...
+    /// Per-AIR lifted selectors for Core at `z^{r_core}` (`r_core = n_max / n_core`).
+    /// Equal to the canonical `IsFirst`/`IsLast`/`IsTransition` when Core is at log_max.
+    /// Only present in `is_multi_air = true`.
+    IsFirstCore,
+    IsLastCore,
+    IsTransitionCore,
+    /// Per-AIR lifted selectors for Chiplets at `z^{r_chip}`. Mirror of `*Core`.
+    IsFirstChip,
+    IsLastChip,
+    IsTransitionChip,
```
adr1anh, line 57: As a follow up, we should investigate making this generic over any number of traces, and keep the ace codegen independent of the VM's air

crates/lib/core/asm/stark/utils.masm
```diff
@@ -269,6 +279,106 @@ pub proc set_up_auxiliary_inputs_ace
...
+        dup.1 dup.1 ext2mul
+        movup.2 sub.1
+        dup push.0 neq
+    end
+    drop
+    # => [z_lift_0, z_lift_1, out_offset, ...]
+
+    # vanishing = z^N - 1 (loaded from slot 1 = offset +2/+3).
+    exec.constants::auxiliary_ace_inputs_ptr add.2 mem_load
+    exec.constants::auxiliary_ace_inputs_ptr add.3 mem_load
```
adr1anh: I'd need to double check, but is this better than just reading the word, and dropping the half we don't need. This pattern appears often. Maybe a helper could be nice too (load_quad?)

adr1anh (commented):
Mostly some small nits. I checked over the masm and it seems good. Thanks a lot for dealing with all my comments, LGTM!

adr1anh (approved):
Mostly some small nits. I checked over the masm and it seems good. Thanks a lot for dealing with all my comments, LGTM!

crates/ace-codegen/src/layout/keys.rs
Nashtare: https://github.com/0xMiden/miden-vm/issues/3166

air/src/constraints/lookup/miden_air.rs
Nashtare: I'm not sure to see the advantage of it. Boundary emission is a "per-trace" concern, and there would be kind of awkward back references if these were put into MidenAir impl and dispatched via the Enum variants, or where you thinking of something else? Feel free to push your idea directly, or we can merge as is (I've addressed your other comments) and we can discuss it further. This will most likely get refactored away with the `Instance` abstraction anyway

air/src/trace/main_trace.rs
```diff
@@ -116,171 +118,195 @@ pub struct MainTrace {
...
-        }
-    }
-
     /// Builds from `build_trace` outputs: core and chiplets row-major, range checker column
     /// vectors.
+    ///
+    /// `core_rows` and `chiplets_rows` are the per-AIR padded heights returned by
+    /// `to_core_chiplets_matrices`. Both must be powers of two. The data arrays are sized to
+    /// `max(core_rows, chiplets_rows)`; rows beyond each per-AIR height are expected to be
+    /// zero-padded.
```
adr1anh, line 135: Why do we need to pad? We also should be able to derive the heights purely from the vectors lengths

air/src/trace/main_trace.rs
```diff
@@ -116,171 +118,195 @@ pub struct MainTrace {
...
-            },
+        let TraceStorage {
+            core_rm, chiplets_rm, range_checker_cols, ..
+        } = &self.storage;
+
+        assert!(r < core_rm.height(), "main trace row index in bounds");
+        assert!(col < TRACE_WIDTH, "main trace column index in bounds");
+
+        if col < CORE_WIDTH {
+            core_rm.values[r * CORE_WIDTH + col]
```
adr1anh: This could use matrix.get

air/src/trace/main_trace.rs
```diff
@@ -116,171 +118,195 @@ pub struct MainTrace {
...
     pub fn width(&self) -> usize {
-        match &self.storage {
-            TraceStorage::Parts { .. } => TRACE_WIDTH,
-            TraceStorage::RowMajor(matrix) => matrix.width(),
-            TraceStorage::Transposed { num_cols, .. } => *num_cols,
-        }
+        TRACE_WIDTH
     }
 
     /// Row-major matrix of this trace.
```
adr1anh, line 197: Should this mention that it concatenates both matrices and pads with empty rows so they are all the same height?

air/src/trace/main_trace.rs
```diff
@@ -116,171 +118,195 @@ pub struct MainTrace {
...
-        match &self.storage {
-            TraceStorage::Parts { num_rows, .. } => *num_rows,
-            TraceStorage::RowMajor(matrix) => matrix.height(),
-            TraceStorage::Transposed { num_rows, .. } => *num_rows,
+    /// Splits the trace into the per-AIR `(Core, Chiplets)` matrix pair used by the multi-AIR
+    /// proving path.
+    ///
+    /// Each matrix is sliced to its own per-AIR padded height: hash-heavy programs grow the
+    /// chiplets trace while core stays small (and vice versa).
+    pub fn to_core_chiplets_matrices(&self) -> (RowMajorMatrix<Felt>, RowMajorMatrix<Felt>) {
```
adr1anh, line 248: Should this not consume self, and reuse the row major buffers?

air/src/trace/main_trace.rs
adr1anh, line 1: We should open a sub-issue on the row major refactor, to use Matrix methods instead of relying on `.values` and manual row slicing.

adr1anh (commented):
Just some more small nits.

air/src/trace/main_trace.rs
Al-Kindi-0: We do have at least one open issue for the row major form refactor to which we can add the above remark

air/src/constraints/lookup/miden_air.rs
adr1anh: We can tackle this later as part of the migration to the next crypto. I think this will be clearer with the new API. The idea is that boundary checks are global, as they interact with the interactions from all of the traces. Even if some checks morally belong to one specific Air (e.g. kernel) we don't gain any expressivity by doing so, instead of doing them all as a bunch.

air/src/trace/main_trace.rs
Nashtare: It's not needed per say, but I'd rather do it on a separate follow-up PR because it is more widespread than the other nitpicky comments and would probably deserve a proper review

adr1anh (approved):
LGTM

PR #3120, opened 2026-05-08
ci: keep synthetic tx nonregression bounded

huitseeker:
The synthetic tx workflow was taking about two hours because it ran the full synthetic benchmark matrix twice, including the prove axis. The largest generated transaction spends roughly 48 seconds per prove iteration, and Criterion was collecting 30 samples, so one axis alone could consume about 24 minutes per side.

Make the CI path explicit and bounded: default it to exec, trace_prep, and verify; expose bench axes and Criterion timing knobs; lower the workflow timeout; and keep the full prove benchmark available through `workflow_dispatch` by setting `bench_axes=all` or `prove`.

Also add a minimum absolute slowdown gate so sub-millisecond timing noise does not fail next, and publish the benchmark report on an associated PR when one exists before falling back to a commit comment.

### DIFF
--- a/.github/workflows/synthetic-tx-nonregression.yml
+++ b/.github/workflows/synthetic-tx-nonregression.yml
@@ -20,10 +20,35 @@ on:
         required: false
         default: "5"
         type: string
+      regression_min_delta_ms:
+        description: Minimum absolute slowdown before the workflow fails.
+        required: false
+        default: "1"
+        type: string
       rayon_num_threads:
         description: RAYON_NUM_THREADS value used for the benchmark.
         required: false
         default: "8"
+        type: string
+      sample_size:
+        description: Criterion sample size.
+        required: false
+        default: "10"
+        type: string
+      measurement_time_secs:
+        description: Criterion measurement time per benchmark.
+        required: false
+        default: "10"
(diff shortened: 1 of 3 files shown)


bobbinth (approved):
Not an in-depth review - but looks good! Thank you!

Al-Kindi-0 (approved):
LGTM
Though 10 samples had some variance and hence the 30 samples choice.
Could be interesting to make the faster/smallest benchmark (e.g., the simple transaction bench) have 30 or 50 samples, while the slower/largest benchmarks can use 10 samples.

bobbinth:
> Though 10 samples had some variance and hence the 30 samples choice.
> Could be interesting to make the faster/smallest benchmark (e.g., the simple transaction bench) have 30 or 50 samples, while the slower/largest benchmarks can use 10 samples.

Yes, that could work very well. I'll merge this as is for now to unblock `next`, and we can experiment more in follow-up PRs.

huitseeker:
Screenshot 2026-05-08 at 3 39 20 PM
@Al-Kindi-0 @bobbinth we do have the ability to set parameters when running this manually through workflow-dispatch

PR #3121, opened 2026-05-09
fix(processor): pass borrow to record_u32_range_checks in op_u32sub

### LINKED ISSUE #3122: bug(processor): op_u32sub passes ZERO instead of borrow to record_u32_range_checks
### Summary

In `processor/src/execution/operations/u32_ops/mod.rs`, `op_u32sub` calls `record_u32_range_checks` with `ZERO` as the second argument instead of the computed `borrow` flag.

### Root Cause

Every other u32 arithmetic operation passes its auxiliary output value (carry, hi, borrow) as the second argument to `record_u32_range_checks`. `op_u32sub` deviated from this pattern by hard-coding `ZERO`:

```rust
// Buggy
tracer.record_u32_range_checks(processor.system().clock(), diff, ZERO);
```

This means the borrow auxiliary is never included in the range-check trace column for subtraction, leaving it unconstrained by that mechanism and making the subtraction trace inconsistent with the constraint pattern shared by all other u32 ops.

| Operation | arg 1 | arg 2 (should be auxiliary output) |
|-----------|-------|-------------------------------------|
| `op_u32add` | `sum` | `carry` |
| `op_u32mul` | `lo` | `hi` |
| `op_u32split` | `lo` | `hi` |
| `op_u32sub` | `diff` | ~~`ZERO`~~ → should be `borrow` |

### Expected Behaviour

`borrow` (already computed as `Felt::from(a < b)`) should be passed as the second argument.

```rust
// Correct
tracer.record_u32_range_checks(processor.system().clock(), diff, borrow);
```

### Affected File

`processor/src/execution/operations/u32_ops/mod.rs`

amathxbt:
## Summary

`op_u32sub` passed `ZERO` as the second argument to `record_u32_range_checks` instead of the `borrow` flag, making the u32 subtract trace inconsistent with every other u32 arithmetic operation.

Fixes #3122

## Rationale

Every u32 arithmetic op in the processor passes its auxiliary output as the second argument to `record_u32_range_checks` so the range-check column covers both output values. `op_u32sub` was the only operation that deviated from this pattern — it hard-coded `ZERO` instead of `borrow`, meaning the borrow auxiliary was never recorded in the range-check trace column. This leaves the borrow value unconstrained by that mechanism and creates an inconsistency in the trace layout vs. the constraint pattern expected by the AIR.

## Root Cause

```rust
// Before (buggy)
tracer.record_u32_range_checks(processor.system().clock(), diff, ZERO);
```

Compare with the pattern used by every other u32 op:

| Operation | First arg | Second arg |
|-----------|-----------|------------|
| `op_u32add` | `sum` | `carry` |
| `op_u32mul` | `lo` | `hi` |
| `op_u32split` | `lo` | `hi` |
| `op_u32sub` | `diff` | ~~`ZERO`~~ **`borrow`** |

## Fix

```rust
// After (fixed)
tracer.record_u32_range_checks(processor.system().clock(), diff, borrow);
```

`borrow` is already computed in `op_u32sub` as `Felt::from(a < b)` before this call, so the change is a single-argument substitution with no other code changes needed.

## Test Plan

1. Run the existing u32 operations test suite: `cargo test -p miden-processor -- u32`
2. Run the full processor test suite: `cargo test -p miden-processor`
3. Run all workspace tests: `cargo test --workspace`

The fix is a one-argument substitution; existing tests that exercise `u32sub` with a borrow (i.e. `a < b`) will now correctly include the borrow in the range-check trace, matching the AIR constraint expectation.

## CHANGELOG

Added `## v0.24.0 (TBD)` section with a `#### Fixes` entry.

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -51,7 +51,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Open issue on failure (only if fail-fast is false and not a fork)
-        if: steps.lychee.outputs.exit_code != 0 && steps.flags.outputs.fail_fast != 'true' && !github.event.repository.fork
+        if: ${{ steps.lychee.outputs.exit_code != 0 && steps.flags.outputs.fail_fast != 'true' && !github.event.repository.fork }}
         uses: peter-evans/create-issue-from-file@v5
         with:
           title: Link Checker Report
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -8,6 +8,7 @@ on:
         required: false
         default: false
         type: boolean
+  workflow_dispatch: {}
   schedule:
     # runs once a day at 01:00 UTC (adjust as needed)
     - cron: '0 1 * * *'
@@ -28,7 +29,7 @@ jobs:
           if [ "${{ github.event_name }}" = "schedule" ]; then
             echo "fail_fast=false" >> $GITHUB_OUTPUT
           else
-            echo "fail_fast=${{ inputs.fail-fast }}" >> $GITHUB_OUTPUT
+            echo "fail_fast=${{ inputs.fail-fast || 'false' }}" >> $GITHUB_OUTPUT
           fi
 
       - name: Link Checker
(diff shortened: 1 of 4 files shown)


amathxbt:
My PAL 
**1. Linked issue**
Created and linked issue #3122 (filed in this repo) see "Fixes #3122" in the updated PR description above.

**2. Rationale**
Added a **Rationale** section to the PR body explaining *why* the change is needed: every other u32 op passes its auxiliary output (carry/hi/borrow) as the second argument to `record_u32_range_checks` so both output values are covered by the range-check column. `op_u32sub` was the sole exception — it hard-coded `ZERO` — meaning the borrow was never recorded in the trace, leaving it unconstrained by that mechanism and creating an inconsistency vs. the AIR constraint pattern.

**3. Test plan**
Added a **Test Plan** section: run `cargo test -p miden-processor -- u32` (existing u32 suite) and `cargo test --workspace`. The fix is a single-argument substitution (`ZERO` → `borrow`); no new test is required because the borrow value (`0` or `1`) trivially satisfies `[0, 2^32)` either way — the correctness issue is about trace consistency, not a wrong computed result.

The PR description has been updated with all three additions.

adr1anh:
See CONTRIBUTING.md, the issue **must be assigned to you**, or your PR will get flagged.

plafer:
Closing per https://github.com/0xMiden/miden-vm/issues/3122#issuecomment-4425073794

PR #3127, opened 2026-05-10
ci: comment synthetic tx bench results on commits

huitseeker:
Makes better sense wrt the tirgger, fixes an error where the workflow doesn't have write access, and aligns with the blake3 benchmark.


bobbinth (approved):
Looks good! Thank you!

bitwalker (approved):

PR #3128, opened 2026-05-11
fix(fri_ops): remove dead domain-segment guard and replace `panic!` with `unreachable!`

### LINKED ISSUE #3132: fri_ops: dead domain-segment guard and panic! wildcard arms are unreachable by construction
## Problem

`op_fri_ext2fold4` computes `domain_segment` via `p & 3`, which guarantees a value in
`[0, 3]` by construction. The subsequent guard `if domain_segment > 3 { return Err(...) }`
is therefore **dead code** that can never fire.

The three private helper functions `bit_reverse_segment`, `get_tau_factor`, and
`get_domain_segment_flags` each carry a `_ => panic!(...)` wildcard arm. These panics are
unreachable given the bitmask invariant, but the intent is not documented. Any future
refactoring that removes `& 3` would silently re-activate the panics.

## Expected Behaviour

- Replace the dead guard with a comment explaining the bitmask invariant.
- Replace the `panic!` wildcard arms with `unreachable!` to make the contract explicit.

## Affected File

`processor/src/execution/operations/fri_ops/mod.rs`

koookxbt:
Fixes #3132

## Summary

`op_fri_ext2fold4` derives `domain_segment` via the bitmask `p & 3`, which guarantees a
value in `[0, 3]`. The subsequent guard `if domain_segment > 3 { return Err(...) }` is
therefore dead code that can never fire.

The three private helpers `bit_reverse_segment`, `get_tau_factor`, and
`get_domain_segment_flags` each carry a `_ => panic!(...)` wildcard arm. Because the
bitmask invariant is not visible inside the helpers, future refactoring (e.g. removing
the `& 3` mask) could silently re-activate those panics.

## Changes

- Remove the unreachable `if domain_segment > 3` guard and replace it with a comment
  explaining the bitmask invariant.
- Change the three `panic!` wildcard arms to `unreachable!` so the contract is explicit.

## Rationale

A `panic!` in a wildcard arm reads as "this is a reachable error path". `unreachable!`
communicates "this branch cannot be reached by construction". The distinction matters for
future contributors and for analysis tools. The dead guard is also misleading — it suggests
the invariant is not already enforced by the bitmask, when in fact it is.

## Test Plan

The existing FRI test suite in `fri_ops/tests.rs` covers `op_fri_ext2fold4` with valid
domain-segment values (0–3). No test exercises the removed dead guard because it was
unreachable by construction; the tests continue to pass unchanged. The `unreachable!` arms
remain untested for the same reason — they cannot be triggered through the public API.

### DIFF
--- a/processor/src/execution/operations/fri_ops/mod.rs
+++ b/processor/src/execution/operations/fri_ops/mod.rs
@@ -59,7 +59,7 @@ where
     // p is the bit-reversed tree index; compute f_pos and d_seg from it
     let p = processor.stack().get(9).as_canonical_u64();
     let domain_segment = p & 3;
-    let folded_pos = Felt::new(p >> 2);
+    let folded_pos = Felt::new_unchecked(p >> 2);
     // the power of the domain generator which can be used to determine current domain value x
     let poe = processor.stack().get(10);
     if poe.is_zero() {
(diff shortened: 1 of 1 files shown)


adr1anh:
See CONTRIBUTING.md, the issue must be assigned to you first.

PR #3137, opened 2026-05-12
fix(dsa): remove dead Option return type from falcon512_poseidon2::sign

Olexandr88:
The falcon512_poseidon2::sign function claimed it could return None if the secret key was malformed, but this was never possible  the underlying SecretKey::sign() is infallible. Removed the misleading Option wrapper and updated all callers accordingly.

## Rationale
SecretKey::sign() in miden-crypto is infallible — it returns Signature directly, not Result or Option. The Option<Vec<Felt>> return type was 
dead code and the documentation was incorrect.

## Test plan
- cargo test -p miden-core-lib - all 351 tests pass


huitseeker:
There is a lint called `clippy::unnecessary_wraps` that can take care of that and a pair of commands 'xclippy`/`xclippy-fix` lints configured in this repo that can make sure this is systematically checked, not just in Falcon. If you care about this issue, please fix it as a PR offering a maintainable, mechanized fix of all instances of this problem.

PR #3138, opened 2026-05-12
ci: Harden GitHub Actions security checks

huitseeker:
GitHub Actions workflows execute code from external repositories, so mutable action refs like @v4 or @main are a supply-chain risk: the same workflow file can run different code later if the tag moves. Enforcement stays centered on zizmor, which now fails CI when actions are not pinned to immutable commits.

Run `zizmor` and `actionlint` when workflow policy files change. The `pull_request` trigger stays path-filtered to `.github` and policy files so ordinary Rust/docs PRs do not pay for an Actions-specific check, while every PR that edits workflow behavior gets validated before merge.

Also make token scope explicit. A top-level `permissions: {}` block disables GitHub's default token grants, and jobs opt back into only the permissions they need: read-only checkout for normal jobs, Pages/OIDC for docs deploys, issues write for scheduled link reports, and contents write only for benchmark commit comments.

### DIFF
--- a/.github/workflows/blake3-nonregression.yml
+++ b/.github/workflows/blake3-nonregression.yml
@@ -26,6 +26,8 @@ on:
         default: "8"
         type: string
 
+permissions: {}
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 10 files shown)


bobbinth (approved):
Looks good! Thank you!

plafer (approved):

PR #3140, opened 2026-05-13
feat(deferred): content-addressed deferred-DAG with built-in SystemEvent dispatch

adr1anh:
## Summary

Stacks on top of #3100 (precompile linear-hash-tree refactor). Adds a content-addressed deferred-computation DAG to the VM and promotes its three driving events to first-class `SystemEvent` variants with dedicated MASM keywords.

The branch is two layers:

- **Layers 1–7** (`bdc8fa5b..05982c972`): the original deferred-DAG infrastructure — content-addressed data model, processor-side state + transaction, type-handler trait/registry, `Field0` handler, host wiring, witness extraction via reachability BFS, and end-to-end tests.
- **Layers 8–13** (`117882a0f..3e2c9bf87`): the cleanup pass that promotes the three events from a host-registry intercept to native VM system events, and adds MASM sugar.

## Architecture (after the cleanup)

- `SystemEvent` gains three variants: `DeferredRegisterLeaf`, `DeferredRegisterOp`, `DeferredAssertEq` (`core/src/events/sys_events.rs`). Dispatch is const-lookup, same fast path as `MerkleNodeToStack`, `MapValueToStack`, etc.
- State ownership splits by mutation pattern, mirroring the precompile design:
  - `DeferredState` (accumulated DAG nodes + assertions) → on `AdviceProvider`, alongside `pc_requests`. Flows out via `ExecutionOutput.advice`.
  - `TypeHandlerRegistry` (read-only config) → on `FastProcessor`, set via `with_deferred_registry` builder, like `ExecutionOptions`.
- New system-event handlers in `processor/src/fast/basic_block/sys_event_handlers.rs` use a `(DeferredState, &TypeHandlerRegistry)` disjoint-borrow helper on `FastProcessor` to mutate state and consult the registry in one call.
- `DefaultHost` no longer carries deferred state or the `on_event` intercept; it's a plain host again.
- MASM keywords `deferred.register_leaf` / `deferred.register_op` / `deferred.assert_eq` lower to `push.<event_id>; emit; drop` via the existing `SystemEvent` lowering machinery. A new `DeferredEventNode` AST type keeps the `deferred.` prefix distinct from the `adv.` family.
- The deferred events return nothing to MASM (no advice push), so the interface is async-shaped — a future Host implementation could process DAG bookkeeping off-thread without touching the dispatch path. No async machinery is added here.

## Test plan

- [x] `cargo test -p miden-processor --test deferred_e2e` — three e2e cases (full register/eval/assert path, mismatch failure, legacy `EventHandler` interop) drive the new SystemEvent dispatch via the sugared keywords.
- [x] `cargo test -p miden-core --lib events::sys_events` — exhaustive consistency check on the LOOKUP table.
- [x] `cargo test -p miden-processor --lib` — full processor suite (1561 tests) for regressions.
- [x] `cargo test -p miden-assembly test_deferred_keywords_assemble` — smoke test that the new sugar parses and assembles.
- [x] `cargo build --workspace` and `cargo clippy --workspace --all-targets` clean (preexisting `kernel_rom` warning unrelated).
- [ ] Manual MASM smoke once a real type handler is plumbed end-to-end (out of scope for this PR).


crates/assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -247,6 +247,8 @@ pub enum Instruction {
     AdvLoadW,
 
     SysEvent(SystemEventNode),
+    DeferredRegister,
+    DeferredEvaluate,
```
adr1anh: Should these be system event nodes

adr1anh: Maybe this should be adv.register_deferred and adv.evaluate_deferred

core/src/deferred/field0.rs
```diff
@@ -0,0 +1,281 @@
...
+/// explicitly.
+///
+/// **Tag layout (Field0-specific, opaque to the processor):**
+/// - `[1, 0, op, 0]` where `op` selects the operation:
+///   - `0` — canonical leaf (payload is 8 u32-limbs)
+///   - `1` — `add` op node (payload is `lhs_digest || rhs_digest`)
+///   - `2` — `mul` op node (payload is `lhs_digest || rhs_digest`)
+///   - `3` — equality-assertion marker (not storable as a node body)
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Field0Handler;
```
adr1anh, line 27: It would be interesting to put this in the core library as a test. Instead of in the processor. Also a good test to see if our framework has the right public API

processor/src/deferred/schema.rs
```diff
@@ -0,0 +1,142 @@
...
+
+// NODE TYPE
+// ================================================================================================
+
+/// Classification produced by [`Schema::is_valid`] at register time.
+///
+/// The processor routes the registered node based on this verdict — into the DAG node map for
+/// expressions, or into the assertion list (with transcript fold + verification) for assertions.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NodeType {
```
adr1anh: Maybe this also contains a chunk node

adr1anh (commented):
Some follow-ups
- move state and witness and other types, and traits into core
- add support for chunk nodes
- rewrite existing precompiles using this new framework
- once those pass, remove old precompiles using implementations
- remove old framework for precompiles (ie list precompile data in the advice provider)
- change api of log precompile wrapper to take a digest of an assertion.

Id want to double check that when we evaluate a node, we also intern the result. 


We also will want to write an issue with a proposed spec for the tag format. First version is pure Keccak and existing fake ECDSA. Next one includes support for finite field and group operations. Ideally, the precompile vm crate would define its own trait for categorizing the different type of operations, so that we can combine them arbitrarily into one single schema. 

As a more involved follow-up, we should also think about how we can approach nested deferred states. The goal is to support the mode where a program verifies multiple proofs, each returning a precompile digest. The verifier then registers these assertions and logs them. This probably requires a special type of transcript node. We would populate the advice inputs with a map of deferred witnesses, and when registering a transcript node, we load the deferred witness into the running deferred state. We should make sure the empty transcript is the zero word. However, it would be amazing if we could commit to a schema, and use that as a tag for the transcript node produced by that schema. This would make it super easy to version a particular set of precompiles. Going even further, we could bundle together
- an implementation of a schema
- the AirBuilder constraints
- a trace witness generator
- the set of MASM procedures which wrap the precompiles
- a recursive verifier procedure for verifying a proof of that schema.

adr1anh:
## Self-review triage

Posting this so reviewers don't have to mentally re-bucket the ~30 inline self-review comments. Each comment falls into one of four buckets; after this comment lands I'll delete the inline comments in buckets A and C-defer (they'd otherwise pull attention away from the live discussion).

### A — already addressed by recent commits (will be deleted)

Twelve inline comments are resolved by the most recent refactor pass and will be deleted shortly:

- **`b2ad1fc59` (node-first API + collapse modules)** resolves the bulk: `core/src/deferred/{node,payload,hash,mod}.rs` collapsed into a single `core/src/deferred.rs`; `Node::digest()` is now a method; `processor/src/deferred/events.rs` removed; handlers re-export trimmed to `pub use field0::Field0Handler;`; `Instruction::DeferredRegister`/`DeferredEvaluate` flattened out of the wrapper enum; `crates/assembly/src/tests.rs` uses `padw padw padw`.
- **`9a2579b86` (node-first state API)** makes `DeferredState::register`/`evaluate` take `Node` rather than raw `(Tag, Payload)`.
- **`b951cb07a` (Schema collapse)** reduces `Schema` to `is_valid` + `eval`, which subsumes the "better name than `decode_node`" comment.

### B — landing in this PR (inline comments stay)

Six small, mechanical changes, each pinned by its inline comment:

- **B1.** Flip the MASM operand-stack layout for deferred sys-events to `[PAYLOAD_LO, PAYLOAD_HI, TAG]` so MASM callers can feed `hperm` directly without re-orienting. (`sys_events.rs:308`)
- **B2.** Allow `DeferredEvaluate` on assertion nodes — `Schema::eval` already returns the node itself on success, so the advice-stack contract is uniform. (`sys_events.rs:329`)
- **B3.** Intern intermediary expression nodes touched during `DeferredState::evaluate`, and document on both `DeferredWitness` and `DeferredState::evaluate` *why* (the witness binds the whole reduction proof, not just the canonical answer). Implementation: `Schema::eval` takes `&mut DeferredState` and inserts each visited node. (cover-review item, no inline)
- **B4.** Extract the deferred handlers from `sys_event_handlers.rs` into a sibling `processor/src/fast/basic_block/deferred_handlers.rs`. (`sys_event_handlers.rs:552`)
- **B5.** Drop the fully-qualified `crate::deferred::DeferredState` path; `use` it at the top of `fast/mod.rs`. (`fast/mod.rs:361`)
- **B6.** `#[derive(Default)]` on `DeferredWitness`/`DeferredState` in place of the manual impls. (`core/src/deferred.rs:102`)

### C-now — promoted into this PR (inline comments stay)

Originally tagged "follow-up" but lifted into this PR because they define the right shape of the framework. Better that external reviewers see it as intended:

- **C1. Move `DeferredState`, `Schema`, `NodeType`, `NoopSchema`, `SchemaError`, `Field0Handler` into `miden-core`.** The framework is incoherent while the data model lives in `core` but the interpreter trait lives in `processor` — a user implementing `Schema` should not need to depend on `processor`. The three `processor/src/deferred/*` "put this in core/src" inline notes stay until this lands, then go away with the move.
- **C5. Field0 becomes a `miden-core` integration test.** Field0 is the canonical demonstration that the public API in core is shaped correctly; putting it under `core/tests/` means breaking the API breaks this test — exactly the canary we want. Today it hides inside `processor` and escapes that signal.
- **C6. Rename to `adv.register_deferred` / `adv.evaluate_deferred`.** Deferred dispatch is advice-shaped (no operand-stack return, advice-stack push), so it belongs in the `adv.*` family alongside `adv.push_*` / `adv.insert_*`. Makes the intent obvious to a MASM reader.

### C-defer — future PR (inline comments will be deleted)

Issues will be filed and linked here once opened:

- **`Schema` as a generic vs `&mut dyn Schema`** on `DeferredState` and `FastProcessor`. Choice is cleaner to make once `Schema` lives in core (post-C1) and the ripple through every processor construction site can be measured.
- **Chunk nodes + `adv.register_deferred_chunk`** for word-aligned bulk data. Explicitly a follow-up: new node variant, new advice-map shape, new event handler.
- **`DeferredWitness::new` validating-vs-unchecked split.** Validating constructor needs a `Schema`; cleaner alongside C1 and once verifier-side ergonomics are known.
- **`extract_witness(&self)` → `into_witness(self)`.** State ships via `&self` on `AdviceProvider`; self-consuming would force a `mem::take` dance — not worth the churn before witness sizes are measured.
- **Stack-read helpers** in the new `deferred_handlers.rs` (post-B4). The helper aggregates 3 stack reads into one `Node`; keeping it makes the new module self-contained. Resolved as a side-effect of B4 — no separate work.

Cover-review follow-ups (separate issues, not blocking this PR):

- Port existing precompiles to the deferred framework.
- Delete the old precompile machinery once ported.
- Change `log_precompile` to take a digest of an assertion.
- Spec issue for tag format (Keccak + ECDSA → ff/curve ops → schema-as-tag).
- Nested deferred states (multi-proof verifier, transcript node committing to a schema).

adr1anh:
Closed in favor of https://github.com/0xMiden/miden-vm/pull/3170

PR #3141, opened 2026-05-13
processor: add unified advice provider element budget

### LINKED ISSUE #3107: Consider imposing a single size limit on the entire advice provider
Post https://github.com/0xMiden/miden-vm/pull/3085, we are imposing various size limits on the advice map. These limits are configurable. We also have a hard-coded limit on the advice stack, and no limit on the Merkle store (though, it is not easy to programmatically push a lot of data into the store).

It may be better to have a single setting that bounds the size of the entire advice provider. The user would specify this before running the program, and sizes of advice stack, advice map, and Merkle store would all be counted against this single limit.

If we could also express this limit in bytes, that may be even better.

mk0walsk:
## Describe your changes                                                                                                                                                                       
                                                                                                                                                                                                 
  This PR adds an optional unified size budget for the entire advice provider and enforces it consistently in `miden-processor`.                                                                 
                                                                                                                                                                                                 
  - Added `ExecutionOptions::max_adv_provider_elements` (default: `usize::MAX`) so existing behavior is preserved unless the caller opts in.                                                     
  - Added `AdviceError::AdvProviderElementBudgetExceeded` for explicit and diagnosable budget failures.                                                                                          
  - Enforced the unified budget across advice stack growth, advice map growth, and Merkle store growth in `AdviceProvider`.                                                                      
  - Applied checks during initial `AdviceInputs` loading, runtime mutations, and `set_options` reconfiguration.                                                                                  
  - Kept Merkle-store checks atomic under bounded mode, and kept a fast path under unbounded mode (`usize::MAX`) to avoid unnecessary cloning overhead.                                          
  - Added targeted tests for stack/map/Merkle growth and for rejecting option tightening below current usage.                                                                                    
                                                                                                                                                                                                 
  Rationale: issue #3107 asks for a single limit across the whole advice provider; this change closes that gap and makes host-side resource boundaries explicit and configurable.                
                                                                                                                                                                                                 
  Fixes #3107.                                                                                                                                                                                   
                                                                                                                                                                                                 
                                                                                                                                                                                                 
  - [ ] Updated `CHANGELOG.md`.


processor/src/execution_options.rs
```diff
@@ -20,6 +20,11 @@ pub struct ExecutionOptions {
     max_adv_map_value_size: usize,
     /// Maximum total number of field elements allowed in live advice map keys and values.
     max_adv_map_elements: usize,
+    /// Maximum total number of field elements allowed across the entire advice provider state.
```
huitseeker: Could this either include `pc_requests` in the budget or avoid saying "entire advice provider state"? `AdviceProvider` stores precompile requests too, and `extend_precompile_requests()` can still grow that state without checking this new limit.

Also @adr1anh how do we want to approach bounding precompile calls?

mk0walsk: Thanks for catching this—I’ve updated the wording to state that max_adv_provider_elements currently covers only advice stack/map/Merkle-store elements (not pc_requests) and added the corresponding changelog entry in this PR.

adr1anh: > Could this either include `pc_requests` in the budget or avoid saying "entire advice provider state"? `AdviceProvider` stores precompile requests too, and `extend_precompile_requests()` can still grow that state without checking this new limit.
> 
> Also @adr1anh how do we want to approach bounding precompile calls?

In https://github.com/0xMiden/miden-vm/pull/3140, it seems like we'll be promoting precompile requests into their own structure, replacing the existing Vec. Therefore, I don't think it's worth bounding the current one.

processor/src/host/advice/mod.rs
```diff
@@ -590,11 +727,20 @@ impl AdviceProvider {
...
         I: IntoIterator<Item = InnerNodeInfo>,
     {
-        self.store.extend(iter);
+        if self.max_provider_elements == usize::MAX {
+            self.store.extend(iter);
+            return Ok(());
+        }
+
+        let mut store = self.store.clone();
+        store.extend(iter);
```
huitseeker: This checks the provider budget only after the cloned store has already absorbed the whole iterator. With a finite budget, a very large `ExtendMerkleStore` can still allocate far past the limit before this returns an error.

Could this path check incrementally, or otherwise bound the incoming node count before extending the clone?

processor/src/host/advice/mod.rs
```diff
@@ -56,6 +59,7 @@ pub struct AdviceProvider {
     map_element_count: usize,
     max_map_value_size: usize,
     max_map_elements: usize,
+    max_provider_elements: usize,
```
plafer, line 62: The issue called for a single bound for advice provider size; adding yet another bound makes this quite complex. IMO we should once and for all agree on how we want to bound advice provider data, and only then implement that change.

mk0walsk: Good point. I changed the semantics so that when `max_adv_provider_elements` is set (i.e. not `usize::MAX`), it is now the single aggregate budget across stack/map/Merkle-store growth, and  `max_adv_map_elements` is no longer enforced in that mode.                                                                                                                                                                                                                                                                                                                                      
  `max_adv_map_elements` remains active only in legacy mode (when provider budget is unbounded), so existing behavior is preserved unless unified mode is explicitly enabled.                                                                                                                                                                                                                   
  I also added tests for both insertion-time and `set_options()` behavior in unified mode, and updated the docs/changelog accordingly.

processor/src/host/advice/mod.rs
mk0walsk: Good catch. I updated `extend_merkle_store()` to enforce the provider budget incrementally while applying nodes to the cloned store, instead of checking only after a full `extend`. This prevents the bounded path from absorbing arbitrarily large iterators before failing.                                                                                                                                                                                                                                                                                                         
  I also added tests to verify:                                                                                                                                                                  
  1) budget-failure keeps the Merkle store unchanged (atomic behavior), and                                                                                                                      
  2) extending with already-present nodes still succeeds at provider cap.

processor/src/host/advice/mod.rs
plafer: From my understanding, @bobbinth is pushing for a single advice provider value, while @huitseeker is pushing for more fine-grained bounds - claims I make based on past PR comments. IMO we should pick one or the other - I'm personally fine with either. Thoughts?

bobbinth: What's the reasoning having more fine-grained limits? We just want to make sure that VM code cannot crush the machine by trying to push too much data into the advice provider, right? If so, setting the overall limit on the size of the advice provider should be sufficient. And I think this should be easier for the user too (e.g., the VM can consume at most 16 MB for the advice provider data).

huitseeker:
Thanks for pushing this forward. I’m going to close this PR without merging because the crypto meeting decision changed the implementation shape.

I left the decision on #3107 here: https://github.com/0xMiden/miden-vm/issues/3107#issuecomment-4513143124

The short version is that we want a byte-size advice-provider limit plus a distinct deferred/precompile budget, so the current element-count implementation should not land as-is.

processor/src/host/advice/mod.rs
huitseeker: Yep, we are going for such a simpler limit, with some nuance on dealing with precompiles: https://github.com/0xMiden/miden-vm/issues/3107#issuecomment-4513143124

PR #3142, opened 2026-05-13
Simplify MastForestBuilder with builder-local refs

### LINKED ISSUE #2473: Simplify MastForestBuilder with fingerprint-based approach

huitseeker:
Fixes #2473. On top of #3208.

`MastForestBuilder` used a live `MastForest` while assembling code. That made assembly depend on position-based `MastNodeId` values. Static linking then had to copy nodes and metadata, then remap those IDs. That was hard to reason about and caused bugs like #2456.

This changes assembly to use two identities (not three):

- `MastNodeKey` (a `Word` alias, replaced `MastNodeFingerprint`): content key for deduplication.
- `MastNodeRef`: builder-local handle to a pending node.
- `MastNodeId`: final dense position in a completed `MastForest`.

The decorator abstraction — `DecoratorRef`, `DecoratorId`, `PendingDecoratorRefs`, `FinalDecoratorIds` — was eliminated entirely. Assembly-operation and debug-variable metadata are stored directly in the pending node record, keyed by `AsmOpRef` and `DebugVarRef`.

The old `core/src/mast/decorators/` directory was renamed to `debug_metadata/` and the decorator wire slots in the serialization format were removed (version `0.0.1` → `0.0.4` progressive stripping).

The builder now owns pending node, asm-op, and debug-var records in direct fields. Static imports are converted into builder-local refs immediately. Final `MastNodeIds` and metadata IDs are assigned once, during finalization, after liveness and ordering are known.

Finalized `MastForests` stay dense and vector-backed, and are treated as immutable after creation. Construction and merge paths now build forest parts first, then create the forest in one step.

Coverage includes builder acceptance tests, finalization invariants, public immutability checks, debug/source metadata preservation, and generated MAST compilation benchmarks.

## Code/Test split

Computed from the current PR merge base (`46692cdeb7`) to the current PR tip (`f334a556ec`).

| Category | Added | Deleted | Net |
|---|---:|---:|---:|
| Non-test | 3,330 | 942 | +2,388 |
| Test | 668 | 246 | +422 |
| Bench | 83 | 7 | +76 |
| Total | 4,081 | 1,195 | +2,886 |

<details><summary>Benchmark summary: no regression</summary>

The generated MAST compilation benchmark harness is introduced by the first commit in this PR (`e9d1419d4c`). That commit no longer builds the benchmark file exactly as-is after later in-branch assembler API changes, so the baseline below uses `e9d1419d4c` with only the final benchmark source file restored. Both sides therefore run the same benchmark cases.

Command:

```sh
cargo bench --profile optimized -p miden-core-lib --bench compilation -- --sample-size 10 --measurement-time 3 --warm-up-time 1
```

Baseline: `e9d1419d4c` with current `crates/lib/core/benches/compilation.rs`
After: `f334a556ec` (current PR tip)

| benchmark | baseline | after | change |
|---|---:|---:|---:|
| compile_core_lib/all | 96.588 ms | 74.231 ms | 23.1% faster |
| compile_generated_programs/deep_control_flow/16 | 10.640 ms | 8.1638 ms | 23.3% faster |
| compile_generated_programs/repeated_subtrees/96 | 8.0598 ms | 7.9031 ms | 1.9% faster |
| compile_generated_programs/error_code_heavy/160 | 537.83 us | 525.78 us | 2.2% faster |

</details>

## Presentation 🚨 *Updated!*

https://gisthost.github.io/?449386f99d0c6d4055d943801f7e6ad1


crates/assembly/src/mast_forest_builder.rs
```diff
@@ -264,157 +1258,162 @@ fn append_serialized_asm_op(data: &mut Vec<u8>, op_idx: usize, asm_op: &Assembly
...
-) -> Vec<u8> {
-    for (id, vars) in pending {
-        if *id == node_id {
-            return serialize_debug_vars(forest, vars);
+impl FinalForestLayout {
+    fn plan(
+        procedure_root_refs: Vec<MastNodeRef>,
+        removable_node_refs: BTreeSet<MastNodeRef>,
+        nodes: &IndexVec<MastNodeRef, PendingMastNode>,
+    ) -> Self {
```
plafer: The `MastForestBuilder` keeps track of all nodes that should be removed, and here we further filter out false positives. Wouldn't it be simpler/more robust to not keep track of anything in the `MastForestBuilder`, and then compute the `removable_node_refs` here as every `node` in `nodes` that is not reachable from the roots?

While writing this, I'm realizing that the downside of my suggestion is that we won't catch nodes that are only accessed dynamically (which I believe the currently implemented approach would work). If this makes sense, I would keep the current implementation, and document *why* we are doing that way (specifically w.r.t. dynamic calls).

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -33,6 +39,504 @@ const PROCEDURE_INLINING_THRESHOLD: usize = 32;
...
+    FinalizeForest {
+        #[source]
+        source: MastForestError,
+    },
+    #[error("failed to register decorators for node {node_id:?}: {source_msg}")]
+    RegisterDecorators { node_id: MastNodeId, source_msg: String },
+}
+
+#[derive(Clone, Debug)]
+struct PendingMastNode {
```
plafer: Would add a docstring to all "intermediate" types in this file, even if just a Claude-generated one

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -33,6 +39,504 @@ const PROCEDURE_INLINING_THRESHOLD: usize = 32;
...
+    fn new() -> Self {
+        Self {
+            debug_info: DebugInfo::new(),
+            nodes: IndexVec::new(),
+            node_id_by_ref: BTreeMap::new(),
+            final_decorator_id_by_ref: BTreeMap::new(),
+        }
+    }
+
+    fn register_live_decorators(
```
plafer: The call order of these methods matters (e.g. this one must be called before `materialize_live_nodes()`). Given that this is just a "local" type, it would be overkill to use a type state pattern; so I'd just document this ordering properly. Optionally, we could also have some internal state that tracks which method should be called next (and errors out if called out of order) - but that might be overkill too

core/src/mast/mod.rs
```diff
@@ -142,18 +139,98 @@ pub struct MastForest {
...
+            }
+        }
+
+        Ok(Self {
+            nodes: parts.nodes,
+            roots,
+            advice_map: parts.advice_map,
+            debug_info: parts.debug_info,
+            commitment_cache: OnceLockCompat::new(),
+        })
```
plafer, line 217: This is a trusted path, so technically we shouldn't need to do *any* validation, right? This module's docstring only mentions that `MastForest::read_from_bytes()` trusts the serialized hash (without any mention of structural validation). I would make it clear here and there exactly what we mean by "trusted"

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -1385,19 +2205,307 @@ impl MastForestBuilder {
...
+
+            record_test_root(&mut builder, *node_refs.last().unwrap());
+
+            let (forest, remapping) = builder.build().unwrap().into_parts();
+            assert_finalization_invariants(&forest, &remapping);
+        }
+    }
+
+    #[test]
+    fn deterministic_stress_builds_deep_repeated_metadata_forest() {
```
plafer: What does this test verify? I would add this in a docstring (not clear to me from the name, and it's a pretty big test)

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -1715,49 +2855,208 @@ mod tests {
...
+
+        let (forest, remapping) = builder.build().unwrap().into_parts();
+        let final_root_id = remapping[&root_block_ref];
+
+        assert!(forest.is_procedure_root(final_root_id));
+        assert_eq!(forest[final_root_id].digest(), root_digest);
+    }
+
+    #[test]
+    fn test_build_orders_external_nodes_before_non_external_nodes() {
```
plafer, line 1129: Let's document this requirement in the `build()` docstring

plafer (commented):
Looks quite good! Loving that `MastForest` is finally immutable.

core/src/mast/mod.rs
huitseeker: I clarified the meaning of trusted here: the trusted path accepts serialized node hashes as authoritative and does not recompute them, but it still performs cheap bounds checks needed for safe indexing. 

So ‘trusted’ is about content commitments, not accepting structurally impossible indexes (which can happen accidentally, e.g. data package truncation).

crates/assembly/src/mast_forest_builder.rs
huitseeker: documented the required order on `MastForestFinalizer`.

crates/assembly/src/mast_forest_builder.rs
huitseeker: added a doc comment explaining what it verifies.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Yes, dynamically accessed nodes is the reason I kept the candidate-based pruning model.

The important distinction is what the reachability walk starts from. If we walk from every procedure root in the final forest, that should be equivalent for liveness. But if we walk only from the entrypoint or the static call graph, it is not enough: `dynexec`/`dyncall` nodes do not have child edges to their runtime targets. At runtime, the processor resolves the target digest by looking it up among procedure roots via `MastForest::find_procedure_root`.

The current policy matches the old shape: the builder only marks refs made obsolete by local rewrites, such as basic-block merging or decorator cloning. Finalization then removes only those candidates, after filtering out anything that is still a procedure root or still referenced as a child. That keeps pruning conservative while still deleting the temporary rewrite artifacts.

I've edited the comment on `nodes_to_remove`.

core/src/mast/mod.rs
```diff
@@ -145,18 +144,101 @@ pub struct MastForest {
     commitment_cache: OnceLockCompat<Word>,
```
bobbinth: Can this not be just `commitment: Word` now? One of the goals of this refactoring was to make it so that the MAST forest commitment does not change after construction - or is this coming in a future PR?

core/src/mast/mod.rs
```diff
@@ -145,18 +144,101 @@ pub struct MastForest {
...
     pub fn new() -> Self {
-        Self {
+        Self::from_parts(MastForestParts {
             nodes: IndexVec::new(),
             roots: Vec::new(),
             advice_map: AdviceMap::default(),
             debug_info: DebugInfo::new(),
+        })
+        .expect("empty MastForest parts should be valid")
+    }
```
bobbinth, line 156: Do we still need this? It is not clear to me that empty MAST forests should be valid. I think a valid MAST forest should contain at least one exported procedure.

core/src/mast/mod.rs
```diff
@@ -145,18 +144,101 @@ pub struct MastForest {
...
+    pub fn from_raw_parts(
+        nodes: IndexVec<MastNodeId, MastNode>,
+        roots: Vec<MastNodeId>,
+        advice_map: AdviceMap,
+        debug_info: DebugInfo,
+    ) -> Result<Self, MastForestError> {
+        Self::from_parts(MastForestParts { nodes, roots, advice_map, debug_info })
+    }
+
+    pub(crate) fn from_parts(parts: MastForestParts) -> Result<Self, MastForestError> {
```
bobbinth, line 170: Would be great to add doc comments to these constructors as it is not immediately clear that this constructor recomputes all node hashes.

core/src/mast/mod.rs
```diff
@@ -145,18 +144,101 @@ pub struct MastForest {
...
+        let forest = Self {
+            nodes: parts.nodes,
+            roots: parts.roots,
+            advice_map: parts.advice_map,
+            debug_info: parts.debug_info,
             commitment_cache: OnceLockCompat::new(),
+        };
+
+        forest.validate()?;
+        forest.validate_node_hashes()?;
```
bobbinth, line 192: Shouldn't most of validation logic above be a part of `MastForest::validate()`?

In general, it is not super clear what is being validated where. For example, `MastForest::validate()` runs `validate_procedure_name_digests()` which validates data in `DebugInfo` and this is not covered by `DebugInfo::validate()`.

We should probably think of a way to make the validation code more "intuitive" - but that's for another PR (once the general structure stabilizes a bit more).

core/src/mast/mod.rs
```diff
@@ -197,54 +274,37 @@ impl MastForest {
         }
     }
```
bobbinth, line 259: I think maybe this procedure is the reason why we kept `commitment_cache` rather than changing it to just `commitment`? If so, would recomputing the commitment be prohibitively expensive here (my understanding is that this is used only in tests, and should affect test runtimes only).

Also, is there a longer-term path to removing this method entirely?

crates/assembly/src/mast_forest_builder.rs
bobbinth, line 1: This file is way too big. Could we split it up into several related modules (e.g., under the `src/mast_forest_builder` directory)?

bobbinth (commented):
Thank you! Not a full review yet - but I left some comments inline. I'm still in the process of trying to understand the overall approach.

core/src/mast/mod.rs
huitseeker: Done. `MastForest` now stores `commitment: Word` directly.

core/src/mast/mod.rs
huitseeker: I kept `MastForest::new()` for now because it is still used as a legacy scratch constructor in tests and helper code, and this PR is already narrowing the production construction paths around finalized parts. I added [a doc comment saying this is legacy builder-style/test support](https://github.com/0xMiden/miden-vm/blob/cd6e353642c5b648ddbafe2ad81bd37601179d5f/core/src/mast/mod.rs#L159-L163) rather than the normal final assembly path.

core/src/mast/mod.rs
huitseeker: Done. I documented the construction policy split.

core/src/mast/mod.rs
huitseeker: The constructor has to perform some checks before it can safely assemble a MastForest, because validate() works on an assembled forest and may index by root/node ids. I added a local comment to make that boundary explicit: bounds/link checks stay before construction; cross-field semantic checks stay on the constructed forest.

Merging the two leads to grifting risk.

core/src/mast/mod.rs
huitseeker: `into_stripped()` no longer needs a cache. It only removes debug metadata, so it leaves nodes, roots, node digests, and the forest commitment unchanged.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Tthe current branch has already moved the builder toward that: refs/ids, pending records, final layout/finalizer, static import/copying, builder API, and tests are now split.

I would leave deeper file movement as follow-up cleanup rather than mixing more mechanical module churn into this review. 

The behavioral changes in this PR are already large enough that keeping the remaining split separate should make review easier.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -103,73 +589,338 @@ impl MastForestBuilder {
...
 
     /// The builder keeps node identity independent from debug metadata.
     pub fn set_emit_debug_info(&mut self, _emit: bool) {}
 
-    /// Returns a reference to the underlying [`MastForest`].
-    pub fn mast_forest(&self) -> &MastForest {
-        &self.mast_forest
+    fn maybe_augment(&self, fp: MastNodeFingerprint, _data: &[u8]) -> MastNodeFingerprint {
+        fp
     }
```
plafer: Do we still need this? It's also not clear what "augment" means here given that the body is a noop

core/src/mast/mod.rs
```diff
@@ -140,31 +138,139 @@ pub struct MastForest {
...
+        let node_count = parts.nodes.len();
+        let mut roots = Vec::with_capacity(parts.roots.len());
+        for root_id in parts.roots {
+            if root_id.to_usize() >= node_count {
+                return Err(MastForestError::NodeIdOverflow(root_id, node_count));
+            }
+            if !roots.contains(&root_id) {
+                roots.push(root_id);
+            }
+        }
```
plafer: This silently skips duplicates, whereas `from_parts()` doesn't. We should probably be consistent between both?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1476,7 +1493,40 @@ impl BasicBlockNodeBuilder {
...
+    /// [`Self::from_op_batches`] so the digest is validated against the operation batches.
+    pub fn from_op_batches_preserving_digest(
+        op_batches: Vec<OpBatch>,
+        decorators: DecoratorList,
+        digest: Word,
+    ) -> Self {
+        Self::from_op_batches_unchecked(op_batches, decorators, digest)
+    }
+
+    pub(in crate::mast) fn from_op_batches_unchecked(
```
plafer: Let's add a docstring here too, specifying exactly what is "unchecked"

processor/src/execution/operations/crypto_ops/tests.rs
```diff
@@ -804,7 +804,12 @@ fn test_op_mrupdate_merge_subtree() {
 // HELPER FUNCTIONS
 // --------------------------------------------------------------------------------------------
 
+/// Creates a canonical Felt from an arbitrary u64 value.
+fn felt(value: u64) -> Felt {
+    Felt::new_unchecked(value % Felt::ORDER)
+}
```
bobbinth, line 810: Maybe we should add a constructor like this on the `Felt` type in `miden-crypto` - e.g., `Felt::new_reduced()` or something like that.

core/Cargo.toml
bobbinth: GH flags this as "unused optional dependency"?

core/src/mast/node/basic_block_node/mod.rs
bobbinth, line 85: Not from this PR, but I think these separators are misplaced here (the code below is not related to serialization).

bobbinth, line 782: nit: this file is pretty large and it would be great to split it up. For example, `BasicBlockNodeBuilder` and related code could go into `src/mast/node/basic_block_node/builder.rs`.

crates/assembly/src/basic_block_builder.rs
```diff
@@ -56,11 +59,11 @@ pub struct BasicBlockBuilder<'a> {
...
-    /// Finalized AssemblyOps with their operation indices (op_idx, AssemblyOp).
-    asm_ops: Vec<(usize, AssemblyOp)>,
+    /// Finalized AssemblyOps with their operation indices (op_idx, asm_op_ref).
+    asm_ops: Vec<(usize, AsmOpRef)>,
     /// Debug variables attached to operations in this block.
-    /// Each entry is (op_index, debug_var_id).
-    debug_vars: Vec<(usize, DebugVarId)>,
+    /// Each entry is (op_index, debug_var_ref).
+    debug_vars: Vec<(usize, DebugVarRef)>,
     mast_forest_builder: &'a mut MastForestBuilder,
```
bobbinth, line 67: Question: why does the `BasicBlockBuilder` need to have a backlink to the `MastForestBuilder`? Could we not collect all the relevant info locally in the `BasicBlockBuilder` and then on something like `BasicBlockBuilder::build()` output the node + all relevant debug info that would then be added to the MAST forest.

I may be wrong, but trying to add data to the MAST forest incrementally through the `BasicBlockBuilder` introduces more complexity than needed.

core/src/mast/mod.rs
```diff
@@ -150,16 +154,53 @@ impl MastForest {
...
+    /// Builds a [`MastForest`] from completed parts.
+    pub(crate) fn from_parts(parts: MastForestParts) -> Result<Self, MastForestError> {
+        if parts.nodes.len() > Self::MAX_NODES {
+            return Err(MastForestError::TooManyNodes);
+        }
+
+        let node_count = parts.nodes.len();
+        for &root_id in &parts.roots {
+            if root_id.to_usize() >= node_count {
+                return Err(MastForestError::NodeIdOverflow(root_id, node_count));
```
bobbinth, line 178: nit: something like `InvalidRootNodeId` could be a more descriptive error here.

core/src/mast/mod.rs
bobbinth, line 272: This and other mutators are still here because we need them for the merger? Ideally, we'd get rid of all such mutators - but that's probably for another PR?

core/src/mast/debuginfo/mod.rs
```diff
@@ -272,6 +272,106 @@ impl DebugInfo {
...
+    pub(super) fn remap_debug_var_storage(&mut self, remapping: &BTreeMap<MastNodeId, MastNodeId>) {
+        self.op_debug_var_storage = self.op_debug_var_storage.remap_nodes(remapping);
+    }
+
+    /// Drops debug metadata entries that are no longer referenced by any retained node, and remaps
+    /// node-indexed storage to the compacted metadata IDs.
+    pub(super) fn compact_node_metadata(&mut self) {
+        self.compact_asm_ops();
+        self.compact_debug_vars();
+    }
```
bobbinth, line 289: Once we get rid of mutators on the `MastForest` - this will also go away, right?

core/src/mast/node/mast_forest_contributor.rs
```diff
@@ -62,6 +102,23 @@ impl MastNodeBuilder {
         }
     }
 
+    /// Build a node whose child IDs have already been remapped into final forest positions.
+    ///
+    /// This path is used by finalizers that materialize nodes in topological order before a
+    /// complete [`MastForest`] exists. Control-node builders must carry their already-computed
+    /// digest on this path.
+    pub fn build_linked(self, _node_id: MastNodeId) -> Result<MastNode, MastForestError> {
```
bobbinth: Question: why do we take `_node_id` as a parameter here (doesn't seem like it is being used)?

Also, it is not immediately clear to me why we need this method. Could this be avoid with a bit extra logic in the node builders themselves? For example, in the `JoinNode`, if the digest is set, we'll use it, and if not, we'll look up the children in the provided MAST forest.

core/src/mast/serialization/resolved.rs
```diff
@@ -156,9 +156,12 @@ impl<'a> ResolvedSerializedForest<'a> {
             })?;
         }
 
+        let mut roots = Vec::with_capacity(self.procedure_root_count());
         for index in 0..self.procedure_root_count() {
-            mast_forest.make_root(self.procedure_root_at(index)?);
+            roots.push(self.procedure_root_at(index)?);
         }
+        mast_forest.roots = roots;
+        mast_forest.commitment = mast_forest.compute_nodes_commitment(&mast_forest.roots);
```
bobbinth: This will change in one of the future PRs, right? Basically, computing a commitment to the MAST forest should be done at forest construction only.

core/src/mast/mod.rs
```diff
@@ -207,39 +247,47 @@ impl MastForest {
             return BTreeMap::new();
         }
```
plafer, line 275: This and `make_root()` were removed in a previous iteration of this PR, right? I remember there being no `&mut self` methods anymore. Did we bring them back intentionally somehow, or is this a merge conflict resolution mishap?

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -93,151 +109,350 @@ impl MastForestBuilder {
...
-
-    for node in mast_forest.nodes() {
-        node.for_each_child(|child_id| {
-            if nodes_to_remove.contains(&child_id) {
-                nodes_to_remove.remove(&child_id);
-            }
-        });
+    for (raw_op_idx, op) in op_batches.iter().flat_map(OpBatch::raw_ops).enumerate() {
+        if matches!(op, Operation::Assert(_) | Operation::U32assert2(_) | Operation::MpVerify(_)) {
+            data.extend_from_slice(&raw_op_idx.to_le_bytes());
```
plafer: `usize::to_le_bytes()` returns 4 bytes on u32 architectures and 8 bytes on 64-bit - we should cast to `u64` before calling `to_le_bytes()`.

crates/assembly/src/mast_forest_builder/node_identity_policy.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Finalization plan that decides which builder-local records become final forest nodes.
+pub(super) struct FinalForestLayout {
+    pub(super) procedure_root_refs: Vec<MastNodeRef>,
+    pub(super) live_node_refs: Vec<MastNodeRef>,
+}
+
+impl FinalForestLayout {
+    pub(super) fn plan(
+        procedure_root_refs: Vec<MastNodeRef>,
+        _removable_node_refs: BTreeSet<MastNodeRef>,
```
plafer: remove?

crates/assembly/src/mast_forest_builder.rs
huitseeker: Thanks for the catch: both basic-block error-code fingerprint helpers now cast `raw_op_idx` to `u64` before serializing.

core/src/mast/mod.rs
huitseeker: `MastForest::make_root()` and `MastForest::remove_nodes()` are now only compiled for tests / `testing` / `arbitrary` support builds — as they were in a prior iteration of this PR. The merger still needs root marking internally, so the production path now uses a private `mark_root()` helper rather than exposing the public mutator.

crates/assembly/src/mast_forest_builder/node_identity_policy.rs
huitseeker: Yep. The old merge-candidate parameter and builder-side `merged_basic_block_refs` set are gone.

core/src/mast/mod.rs
huitseeker: This direction is now covered by follow-up PR #3221.

That PR splits debug info out of `MastForest`, which makes the validation boundary more natural: `MastForest::validate()` can focus on executable forest structure, while package/source debug metadata validation moves to the debug-info/package layer.

plafer (approved):
LGTM!

bobbinth (approved):
Looks good! Thank you! Not a real review - but I left some comments inline. These are from a few days ago - so, may be dated/irrelevant.

Since this already has @plafer approval, I'm OK with merging this as is, and I can review the final state when I do the review of https://github.com/0xMiden/miden-vm/pull/3221.

core/src/mast/debuginfo/mod.rs
huitseeker: Yes. Once `MastForest` no longer owns debug metadata, these forest-local helpers should go away or move.

For this PR, they are already test-only because they only support `remove_nodes()`.

core/src/mast/mod.rs
huitseeker: I left this alone for now because `NodeIdOverflow` is used consistently for node-id bounds failures across builders and deserialization. If we want better diagnostics, I would do that as a small follow-up error-surface cleanup rather than mix it into #3142.

core/src/mast/mod.rs
huitseeker: At current #3142 tip, the mutable APIs under discussion are already limited:

- `make_root()` is only for tests/arbitrary support.
- `remove_nodes()` is only for tests.
- the debug metadata remap/compact helpers used by `remove_nodes()` are only for tests.

The production merger path uses internal construction/root marking rather than exposing those public mutators. 

#3221 will make the next step cleaner because debug info moves out of MastForest, so fewer debug-specific mutation/remap paths need to exist on the forest at all.

core/Cargo.toml
huitseeker: Good catch, the arbitrary support now uses manual `Arbitrary` impls, so the `proptest-derive` optional dependency was no longer needed there.

crates/assembly/src/basic_block_builder.rs
huitseeker: The backlink is not ideal.

It exists because `BasicBlockBuilder` does more than collect VM operations. While lowering instructions, it also needs to allocate builder-owned refs:

- `set_instruction_cycle_count()` creates `AssemblyOps` and stores `AsmOpRefs`.
- `push_debug_var()` stores `DebugVarRefs`.
- `register_error()` stores assertion and error messages.
- `procref` resolves a procedure body ref to its root digest.

So the backlink keeps those arenas owned by `MastForestBuilder`, while `BasicBlockBuilder` stores only refs.

The cleaner shape is for `BasicBlockBuilder` to return a built block plus pending metadata. `MastForestBuilder` would then intern that metadata when it adds the block.

I would do this after #3221, because #3221 already changes where debug info lives.

core/src/mast/node/mast_forest_contributor.rs
huitseeker: Good point. The unused node-id argument is gone.

`build_linked()` is the path for nodes whose children are already final `MastNodeIds` and whose digest is already known. The finalizer needs that because it builds nodes before there is a complete `MastForest` to pass into the normal `build()` path.

Here, a forced digest means the digest stored on the builder with `with_digest()`. Normal builders compute the digest from operations or child nodes. `build_linked()` preserves the known digest instead.

I would keep this separate from `build()`. A single `build()` method that sometimes reads from the forest and sometimes ignores it would be less clear.

core/src/mast/serialization/resolved.rs
huitseeker: At the time of your comment, `ResolvedSerializedForest::materialize()` still built an empty forest, filled nodes, assigned roots, then patched commitment.

That is fixed now. `materialize()` now collects nodes and roots, then calls `MastForest::from_trusted_deserialization_parts()`. That constructor computes the commitment once, from the final nodes and roots.

It does not recompute node digests on the trusted path (trusted deserialization keeps the serialized node digests).

That is separate from untrusted deserialization. `UntrustedMastForest::validate()`recomputes non-external node digests from structure, or rebuilds them for hashless input, before returning a usable `MastForest`.

PR #3143, opened 2026-05-14
fix(clippy): enable and fix clippy::unnecessary_wraps across workspace

Olexandr88:
Enables the clippy::unnecessary_wraps lint in .cargo/config.toml for both xclippy and xclippy-fix commands, and fixes all instances across the workspace using cargo +nightly xclippy-fix plus manual fixes for cases where callers needed to be updated.

Closes the concern raised in #3137

### DIFF
--- a/crates/assembly/src/instruction/crypto_ops.rs
+++ b/crates/assembly/src/instruction/crypto_ops.rs
@@ -275,5 +275,4 @@ fn update_mtree(block_builder: &mut BasicBlockBuilder) {
 
     // stack: [V_old, R_new, ...] (25 cycles)
     block_builder.push_ops(ops);
-
 }
--- a/crates/assembly/src/instruction/crypto_ops.rs
+++ b/crates/assembly/src/instruction/crypto_ops.rs
@@ -135,7 +135,7 @@ pub(super) fn mtree_set(block_builder: &mut BasicBlockBuilder) -> Result<(), Rep
     // stack: [d, i, R_old, V_new, ...]
 
     // stack: [V_old, R_new, ...] (30 cycles)
-    update_mtree(block_builder)
+    Ok(update_mtree(block_builder))
 }
 
 /// Creates a new Merkle tree in the advice provider by combining trees with the specified roots.
@@ -203,7 +203,7 @@ fn read_mtree_node(block_builder: &mut BasicBlockBuilder) {
 /// and perform the mutation on the copied tree.
 ///
 /// This operation takes 30 VM cycles.
-fn update_mtree(block_builder: &mut BasicBlockBuilder) -> Result<(), Report> {
+fn update_mtree(block_builder: &mut BasicBlockBuilder) {
     // stack: [d, i, R_old, V_new, ...]
     // output: [R_new, R_old, V_new, V_old, ...]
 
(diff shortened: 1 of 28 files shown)


crates/assembly/src/project.rs
```diff
@@ -557,42 +557,42 @@ where
         &self,
         package_id: &PackageId,
         selected: &PackageVersion,
-    ) -> Result<bool, Report> {
+    ) -> bool  {
```
huitseeker: `cargo fmt --check` currently fails on this branch. This line is one example, but rustfmt also wants a few blank lines and function wrapping cleaned up in nearby changed files.

crates/assembly/src/instruction/debug.rs
```diff
@@ -10,7 +10,7 @@ use crate::{ProcedureContext, ast::DebugOptions, diagnostics::Report};
 pub fn compile_options(
     options: &DebugOptions,
     proc_ctx: &ProcedureContext,
-) -> Result<miden_core::operations::DebugOptions, Report> {
+) -> miden_core::operations::DebugOptions {
```
huitseeker, line 13: Since this no longer returns `Result`, `diagnostics::Report` at the top of this file is now unused. Please drop that import too.

huitseeker (changes requested):
This will need a CHANGELOG line with `[BREAKING]`, see [CONTRIBUTING.md](https://github.com/0xMiden/miden-vm/blob/next/CONTRIBUTING.md)

crates/assembly/src/instruction/debug.rs
Olexandr88: done

crates/assembly/src/project.rs
Olexandr88: done

crates/assembly/src/instruction/crypto_ops.rs
```diff
@@ -135,7 +135,7 @@ pub(super) fn mtree_set(block_builder: &mut BasicBlockBuilder) -> Result<(), Rep
     // stack: [d, i, R_old, V_new, ...]
 
     // stack: [V_old, R_new, ...] (30 cycles)
-    update_mtree(block_builder)
+    Ok(update_mtree(block_builder))
```
huitseeker: This still wraps a unit-returning helper inside `Ok(...)` IIUC? Since this PR is about removing unnecessary wrappers, this would read cleaner as:

```rust
update_mtree(block_builder);
Ok(())
```

crates/assembly/src/instruction/procedures.rs
```diff
@@ -113,14 +113,10 @@ impl Assembler {
                 .digest()
         };
 
-        self.procref_mast_root(mast_root, block_builder)
+        Ok(self.procref_mast_root(mast_root, block_builder))
```
huitseeker: I think this keeps the same unit-wrapping pattern. Since `procref_mast_root` cannot fail now, this would be clearer as a plain call followed by `Ok(())`.

huitseeker (approved):
Thanks a lot!

bobbinth (approved):
Looks good! Thank you!

PR #3145, opened 2026-05-14
Handle Empty Control Flow

jschnelder:
## Describe your changes                                                                                                                                                                       
  - Allowed parser-generated empty control-flow forms to parse successfully: `if.<cond> end`, `if.<cond> else end`, `while.true end`, and `repeat.<N> end`.                                      
  - Added a semantic validation pass (`VerifyEmptyControlFlow`) that warns on empty `while.true` and empty `repeat` bodies.                                                                      
  - Kept behavior safe and explicit: with `warnings_as_errors`, empty loop bodies are rejected unless `nop` is provided intentionally.                                                           
  - Updated syntax/sema/assembler tests to cover both acceptance and rejection boundaries.                                                                                                       
  - Validation evidence:                                                                                                                                                                         
    - `cargo test -p miden-assembly-syntax empty_` → passed                                                                                                                                      
    - `cargo test -p miden-assembly empty_` → passed                                                                                                                                             
                                                                                                                                                                                                 
  - [x] Updated `CHANGELOG.md`.

### DIFF
--- a/crates/assembly-syntax/src/ast/op.rs
+++ b/crates/assembly-syntax/src/ast/op.rs
@@ -16,9 +16,9 @@ pub enum Op {
     /// Can be either `if`..`end`, or `if`..`else`..`end`.
     If {
         span: SourceSpan,
-        /// This block is always present and non-empty
+        /// This block is always present, but may be empty.
         then_blk: Block,
-        /// This block will be empty if no `else` branch was present
+        /// This block is empty when no `else` branch was present.
         else_blk: Block,
     } = 0,
     /// Represents a condition-controlled loop
--- a/crates/assembly-syntax/src/parser/cst/blocks.rs
+++ b/crates/assembly-syntax/src/parser/cst/blocks.rs
@@ -73,8 +73,15 @@ fn lower_operation(
 
 /// Lowers an `if.true` / `if.false` operation.
 ///
-/// This preserves legacy semantics for empty branches: a missing `then` branch is only tolerated
-/// when an `else` branch exists, in which case the empty side is lowered as an explicit `nop`.
+/// The CST parser can represent empty branches directly, so lowering accepts these forms:
+///
+/// * `if.<cond> end`
+/// * `if.<cond> else end`
+/// * `if.<cond> else <ops> end`
+///
+/// When no source operations are present in either branch, both branches lower to empty blocks.
+/// For non-empty then-branches with omitted/empty else-branches, we preserve legacy behavior and
+/// synthesize a `nop` else-branch.
 fn lower_if_op(context: &mut LoweringContext<'_>, op: &CstIfOp) -> Result<ast::Op, ParsingError> {
     let span = context.parse().span_for_node(op.syntax());
     let cond = parse_if_condition(context, op)?;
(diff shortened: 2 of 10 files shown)


bitwalker:
You'll need to rebase this on top of the new CST-based parser that was just merged. You don't need to modify the legacy LALRPOP grammar, just the CST lowering path

crates/assembly-syntax/src/parser/grammar.lalrpop
bitwalker, line 1: This parser is no longer used, and you didn't update the lowering code for the CST parser to match what was described here. Since none of the tests are actually exercising these parser grammar changes, I would just remove them.

It's fine to keep this grammar in sync with the CST-based parsing path, but it's going to be removed in the next week or so, so it's the new parser that's actually important.

bitwalker (changes requested):

huitseeker:
@jschnelder Please rebase on the newer parser, now merged.

jschnelder:
> @jschnelder Please rebase on the newer parser, now merged.

@huitseeker  Thanks for the review 

  I removed the `grammar.lalrpop` changes and moved the empty control-flow handling to the CST path instead (`parser/cst/blocks.rs`), so the behavior now lives in the parser that is actually used.

  What changed:
  - CST lowering now accepts:
    - `if.<cond> end`
    - `if.<cond> else end`
    - `while.true end`
    - `repeat.<N> end`
  - Kept semantic safety by warning on empty `while` / `repeat` via `VerifyEmptyControlFlow` (and these are rejected when `warnings_as_errors = true`).
  - Updated parser/sema/assembly tests accordingly.
  - Rebased on latest `next` and force-pushed.

  Validation:
  - `cargo +1.95.0-x86_64-pc-windows-msvc test -p miden-assembly-syntax empty_`
  - `cargo +1.95.0-x86_64-pc-windows-msvc test -p miden-assembly empty_`
  - `cargo +1.95.0-x86_64-pc-windows-msvc test -p miden-assembly invalid_repeat`

crates/assembly-syntax/src/parser/cst/blocks.rs
```diff
@@ -84,30 +91,20 @@ fn lower_if_op(context: &mut LoweringContext<'_>, op: &CstIfOp) -> Result<ast::O
...
     let then_blk = if then_has_ops {
         lower_block(context, &then_node)?
-    } else if else_node.is_some() {
-        nop_block(span)
     } else {
-        return Err(ParsingError::InvalidSyntax {
-            span: context.parse().span_for_node(then_node.syntax()),
-            message: "expected a non-empty `if` block".to_string(),
-        });
+        empty_block(context.parse().span_for_node(then_node.syntax()))
```
huitseeker, line 99: Could you also update the `Op::If` AST comment? This path now builds an empty `then_blk`, but `ast/op.rs` still says the then block is always present and non-empty.

The assembler currently turns an empty compiled body into a `Noop`, so I do not think this breaks execution. The risk is that future visitors may trust the stale invariant and skip the empty-branch case.

huitseeker (commented):
This LGTM aside for the CI failure and the inline nit.

crates/assembly-syntax/src/parser/cst/blocks.rs
jschnelder: Thanks, good catch.                                                                                                                                                                            
                                                                                                                                                                                                 
  I updated the `Op::If` AST invariant comment to match the current CST lowering behavior:                                                                                                       
  - `then_blk` is always present, but may be empty                                                                                                                                               
  - `else_blk` is empty when no `else` branch is present

crates/assembly-syntax/src/ast/op.rs
```diff
@@ -16,9 +16,9 @@ pub enum Op {
     /// Can be either `if`..`end`, or `if`..`else`..`end`.
     If {
         span: SourceSpan,
-        /// This block is always present and non-empty
+        /// This block is always present, but may be empty.
         then_blk: Block,
-        /// This block will be empty if no `else` branch was present
+        /// This block is empty when no `else` branch was present.
```
huitseeker, line 21: Nit: this invariant still is not true for ordinary `if.true <ops> end`. The CST lowering keeps the legacy representation and synthesizes a `nop` else block when the `then` branch has operations and no `else` branch was written, so this comment should either mention that exception or the lowering should stop inserting the synthetic `nop`.

huitseeker:
Closing for inactivity

PR #3147, opened 2026-05-14
Verify AEAD ciphertext during decrypt

huitseeker:
This extracts the AEAD decrypt hardening from #3109.

AEAD decrypt uses host-provided advice for the plaintext. This PR makes the MASM decrypt procedure re-encrypt that advised plaintext and verify that it reproduces the caller-provided ciphertext blocks, encrypted padding block, and authentication tag.

This prevents a bad event handler/prover from swapping in plaintext that only passes tag verification.

 Fixes 0xmiden/security-issues#19


Al-Kindi-0 (approved):
LGTM

Nashtare (approved):
LGTM

Issue #3144, opened 2026-05-14
miden-assembly 0.22.2 silently changes MAST procedure digests for unchanged source


WiktorStarczewski:
## Summary

`miden-assembly` 0.22.2 changes the public procedure digests of compiled libraries even when the MASM source, Rust source, and all other build inputs are bit-identical. This silently breaks compatibility with anything built against 0.22.1.

Concretely: the `BASIC_FUNGIBLE_FAUCET_LIBRARY` compiled by `miden-standards` exports two procedures. One of those procedure digests changes between miden-assembly 0.22.1 and 0.22.2 with no change to any source input.

Downstream consequence (Miden Wallet user report we traced back to this): a faucet account deployed via a Rust crate that resolved `miden-assembly = "0.22"` to 0.22.3 cannot be recognized as a standard `BasicFungibleFaucet` by a wallet whose bundled SDK was built against 0.22.1, because the wallet's `BasicFungibleFaucetComponent::from_account()` call ultimately runs `procedure_digests().all(|d| procedures_set.contains(d))` and one digest doesn't match. The wallet throws `MissingBasicFungibleFaucetInterface` and the user's notes silently disappear from the UI.

## Affected versions

- **Regression introduced in:** `miden-assembly` **0.22.2** (published 2026-04-29).
- Confirmed identical (still broken) in **0.22.3**.
- Last known-good: **0.22.1**.

Bisected by running the reproducer below against each of v0.22.1, v0.22.2, v0.22.3.

## Reproducer

Self-contained — paste into a clean shell. Clones a fresh copy of `miden-vm`, creates three worktrees (one per tag), then builds a tiny consumer crate that prints `BasicFungibleFaucet`'s procedure digests from `miden-standards 0.14.5`. The only thing varying between runs is the `[patch.crates-io]` block; source files, miden-standards version, and Rust toolchain are held constant.

Prerequisites: `git`, a stable Rust toolchain new enough for edition 2024.

```bash
# Fresh clone + three pinned worktrees.
WORK=$(mktemp -d)
cd "$WORK"
git clone --quiet https://github.com/0xMiden/miden-vm.git
cd miden-vm
git worktree add ../miden-vm-0221 v0.22.1
git worktree add ../miden-vm-0222 v0.22.2
git worktree add ../miden-vm-0223 v0.22.3
cd ..

# Tiny consumer crate that prints BFF procedure digests.
mkdir -p test-crate/src
cd test-crate

[...]

PR #3149, opened 2026-05-14
ci: guard miden-standards component roots

huitseeker:
Adds a non-blocking release gate for public `miden-standards` account component procedure roots.

For patch releases, the workspace release action now compares the component root set from the current checkout against the previous patch tag, for the same `miden-standards` crate version. This runs in the shared `workspace-release` action, so it covers both workspace dry-run and publish workflows.

This catches regressions like #3144, where unchanged public component MASM produced different procedure roots across a patch release.

### DIFF
--- a/scripts/check-masm-export-digests.rs
+++ b/scripts/check-masm-export-digests.rs
@@ -9,10 +9,11 @@ miden-assembly-syntax-current = { package = "miden-assembly-syntax", path = "../
 miden-mast-package-current = { package = "miden-mast-package", path = "../crates/mast-package" }
 miden-package-registry-current = { package = "miden-package-registry", path = "../crates/package-registry", features = ["resolver"] }
 
-miden-assembly-previous = { package = "miden-assembly", git = "https://github.com/0xMiden/miden-vm", tag = "v0.22.4" }
-miden-assembly-syntax-previous = { package = "miden-assembly-syntax", git = "https://github.com/0xMiden/miden-vm", tag = "v0.22.4" }
-miden-mast-package-previous = { package = "miden-mast-package", git = "https://github.com/0xMiden/miden-vm", tag = "v0.22.4" }
-miden-package-registry-previous = { package = "miden-package-registry", git = "https://github.com/0xMiden/miden-vm", tag = "v0.22.4", features = ["resolver"] }
+# The release wrapper rewrites these tags to the latest release tag on main.
+miden-assembly-previous = { package = "miden-assembly", git = "https://github.com/0xMiden/miden-vm", tag = "v0.23.0" }
+miden-assembly-syntax-previous = { package = "miden-assembly-syntax", git = "https://github.com/0xMiden/miden-vm", tag = "v0.23.0" }
+miden-mast-package-previous = { package = "miden-mast-package", git = "https://github.com/0xMiden/miden-vm", tag = "v0.23.0" }
+miden-package-registry-previous = { package = "miden-package-registry", git = "https://github.com/0xMiden/miden-vm", tag = "v0.23.0", features = ["resolver"] }
 ---
 
 use std::{
--- a/scripts/check-masm-export-digests.rs
+++ b/scripts/check-masm-export-digests.rs
@@ -1,0 +1,209 @@
+#!/usr/bin/env -S cargo +nightly -Zscript
+---cargo
+[package]
+edition = "2024"
+
+[dependencies]
+miden-assembly-current = { package = "miden-assembly", path = "../crates/assembly" }
+miden-assembly-syntax-current = { package = "miden-assembly-syntax", path = "../crates/assembly-syntax" }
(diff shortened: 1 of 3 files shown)


bitwalker (changes requested):
IMO, this sort of thing is an ideal use case for `cargo +nightly -Zscript` - you can write the whole script in Rust, with dependencies, and make this a _lot_ more trivial:

How I'd do this:

* Define a `cargo` script which declares a dependency on the previous/current VM releases, but aliases them so that both can be used simultaneously.
* Have the script take as input a path to a MASM project to assemble with both versions of the assembler, and assemble them to packages
* Take the union of the procedure exports of both packages produced (based on export name), then compare the digests. Report any mismatches (or just bail on the first mismatch)

Not only will that be a lot more readable, it will ensure that we are only failing this job when the same exported procedure fails to assemble to the same digest, and we don't have to hardcode any information about the package being checked. For example, we could also run this same check against the transaction kernel, or the core library.

huitseeker:
@bitwalker I've developed the rust script version here:
https://github.com/0xMiden/miden-vm/compare/main...huitseeker:issue-3144-rust-script?expand=1

This version demonstrates the complication pretty clearly: the previous tag is hardcoded in the embedded manifest, both crate stacks have to be imported under aliases, extraction code is duplicated because the previous/current `Package` and `PackageExport` types are distinct, and the registry cache type had to be chosen to work across versions.

bitwalker:
> @bitwalker I've developed the rust script version here: https://github.com/0xMiden/miden-vm/compare/main...huitseeker:issue-3144-rust-script?expand=1
> 
> This version demonstrates the complication pretty clearly: the previous tag is hardcoded in the embedded manifest, both crate stacks have to be imported under aliases, extraction code is duplicated because the previous/current `Package` and `PackageExport` types are distinct, and the registry cache type had to be chosen to work across versions.

Thanks!

I think that's a lot nicer from a readability/maintainability perspective. The duplication is expected since types between those versions are mutually incompatible, but it's really not that bad here (and could be factored out with `macro_rules!` easily enough, since the code between patch versions should be identical). The alternative in `awk`/shell is much harder to follow, and that's coming from someone comfortable with both.

If nobody else feels strongly about it, then I'll defer, but if it was up to me I'd go with the Rust script.

huitseeker:
@bitwalker Reset to the rust-script version!

scripts/check-masm-export-digests.rs
```diff
@@ -0,0 +1,209 @@
...
+                println!("{name} {}", current_export.describe());
+            },
+            (Some(previous_export), Some(current_export)) => {
+                if previous_export.digest != current_export.digest {
+                    eprintln!(
+                        "::error::export digest changed for {name}: previous={}, current={}",
+                        previous_export.digest, current_export.digest,
+                    );
+                }
+                if previous_export.signature != current_export.signature {
```
bitwalker: I would suggest we only treat this as a problem when the previous version had a signature. If we _add_ a signature where one was previously missing, that's non-breaking (or at least, we should assume it is).

scripts/check-masm-export-digests.rs
```diff
@@ -0,0 +1,209 @@
...
+                    eprintln!(
+                        "::error::export signature changed for {name}: previous={}, current={}",
+                        previous_export.signature, current_export.signature,
+                    );
+                }
+                if previous_export.attributes != current_export.attributes {
```
bitwalker: I think we should probably only treat this as a problem when ABI-specific attributes change (i.e. `@callconv`, `@auth_script` I think are the only ones that are special). I think adding them, like signatures, can be considered non-breaking - but removing/changing them should be errors.Since attributes in general are supposed to be discardable/optional, we probably shouldn't raise an error unless the attribute(s) is special in some way.

scripts/check-masm-export-digests.rs
```diff
@@ -0,0 +1,209 @@
...
+                            .unwrap_or_else(|| "None".to_string()),
+                        attributes: procedure
+                            .attributes
+                            .iter()
+                            .map(ToString::to_string)
+                            .collect::<Vec<_>>()
+                            .join(" "),
+                    },
+                )),
+                PackageExport::Constant(_) | PackageExport::Type(_) => None,
```
bitwalker: If we're going to raise an error about signatures changing, then we should probably treat type changes as errors, as those types may appear in the exported signatures.

I'd also consider changes to previously-defined constants as a breaking change, but since they are purely data, and are meant to abstract over the constant value, I can see an argument for those not mattering as much.

bitwalker (changes requested):
Nice! This is super handy. I have a couple of small adjustments I think we should make, but overall this looks great!

bitwalker (approved):

bobbinth (approved):
Not a deep review - but looks good! Thank you!

The PR title threw me off a bit - I think what we are actually testing here is the root stability of the `core` library (rather than `standards` library).

Issue #2961, opened 2026-04-03
Debug vars lost during block deduplication and merging


djolertrk:
### Packages versions

miden-vm v22.0

### Bug description

Blocks with identical operations but different debug variables were being deduplicated, causing crash in compiler and incorrect variable scoping in the debugger. We need to handle it the same way we handle `asm_ops`.

### Relevant log output

```shell

```

djolertrk:
cc @huitseeker (please note that I do not have access to assign this to myself :) )

bobbinth:
Is this issue still relevant? If yes, it it actually a sub-issue of https://github.com/0xMiden/miden-vm/issues/3040?

djolertrk:
> Is this issue still relevant? If yes, it it actually a sub-issue of [#3040](https://github.com/0xMiden/miden-vm/issues/3040)?

Hi @bobbinth. This is solved by https://github.com/0xMiden/miden-vm/pull/2955. We should close this. Thanks!

bobbinth:
Closed by #2955.

huitseeker:
Reopened by #3155, 3157, #3182

PR #3151, opened 2026-05-14
Restore metadata-neutral MAST shape for procedure roots

### LINKED ISSUE #3144: miden-assembly 0.22.2 silently changes MAST procedure digests for unchanged source

huitseeker:
This reverts the metadata-sensitive MAST shape changes introduced around #2955/#3019 and the later static-link provenance follow-up.

The root issue is not that procedure hashing directly includes debug info. It should not, and it does not need to. The issue is that debug/decorator metadata became part of node identity during assembly and merge. That changed when blocks could be deduplicated or merged. The resulting MAST shape changed, and the  shape is what the procedure root hashes.

In short:

- fingerprinting may depend on debug/decorator metadata
- procedure hashing must not depend on debug metadata
- but if fingerprinting changes dedup/merge decisions, it changes MAST shape
- changed MAST shape changes procedure roots

<details><summary>#2955 made debug metadata affect whether equal executable blocks reuse the same MastNodeId</summary>

- same ops + different metadata
- before #2955: same node id
- after #2955: different node ids

It changes the answers seen by merge/removal logic :
before:
    inner block reuses exported procedure-root node id
    -> merge sees "this is a procedure root"
    -> large block is not merged/inlined

  after:
    inner block gets separate non-root node id
    -> merge sees "this is not a procedure root"
    -> block can be merged/inlined into neighbors
</details>

This PR restores the old rule: asm/debug metadata does not decide MAST node identity. When structurally equal nodes dedup, the first metadata wins.

The static-link exact-root debug metadata path is also rolled back. Same-digest debug provenance remains a follow-up under #2990. This PR accepts that hard cutoff rather than keeping a compatibility path.

Reopens #2961 on the 0.22 line. Back-merging this to next would re-open #2961 and #2990. Closes #3144.


bitwalker:
We can't merge this to main right? We already released 0.23, so this fix for 0.22 has to be made manually in a separate branch based of the 0.22.3 tag AIUI

huitseeker:
@bitwalker Except I think we need to patch 0.23 as well, to maintain the invariant that shape (and therefore roots) should not depend on debug info.

huitseeker:
Closing in favor of #3155. Will reopen here when we prepare the back-merge into the v0.23.0 line.

PR #3154, opened 2026-05-15
Enforce Public Types in Exported Signatures

### LINKED ISSUE #3092: Reject private type declarations in exported procedure signatures
Exported procedure signatures can currently expose private local type declarations.

For example:

```masm
enum PrivateEnum {
    A = 0
}

pub proc f(value: PrivateEnum)
    nop
end
```

This is accepted because the type reference is local to the same module. The existing private-symbol check only rejects private references across module boundaries.

The problem is that a `pub proc` signature becomes package surface. If the signature contains a private type, the final package can contain a public procedure signature that refers to a type that is not exported. This can leave dangling type references for consumers, especially for enum types.

The expected behavior is:

- exported procedure signatures may only reference public/exported named types
- this should apply even when the type is declared in the same module
- nested signature types should be checked recursively
- enum payload types exposed through public signatures should not introduce private dangling references

This is related to #2944.

cmende:
## Describe your changes                                                                                                                                                                       
                                                                                                                                                                                                 
  This PR closes a privacy gap in semantic checks for exported procedure signatures: `pub proc` signatures could still reference private local types (including transitive alias chains and  absolute paths like `::module::PrivateType`).                                                                                                                                                  
  It now enforces that all types reachable from exported signatures are public, adds a dedicated diagnostic (`PrivateTypeInExportedSignature`), and includes regression tests in both `assembly-syntax` and `assembly`.                                                                                                                                                                        
  This is needed to prevent leaking private module internals through public APIs and aligns with the privacy boundary work tracked in #3092 (following #3091).                                   
                                                                                                                                                                                                 
  - [x] Updated `CHANGELOG.md`.


cmende:
Addressed. I traced the failing check to `rustfmt check nightly on ubuntu-latest` only. The issue was formatting drift in `crates/assembly-syntax/src/sema/mod.rs` (no logic changes). Fixed in commit `db3957350` and pushed to `exported-signature-privacy.

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -187,6 +189,107 @@ pub fn analyze(
...
+                SymbolResolution::Local(item) => Some(item.into_inner()),
+                SymbolResolution::External(path)
+                    if path.parent().is_some_and(|parent| parent == module.path()) =>
+                {
+                    let Some(local_name) = path.last() else {
+                        return;
+                    };
+                    module.index_of(|item| item.name().as_str() == local_name)
+                },
+                SymbolResolution::External(_)
```
huitseeker, line 251: This leaves a cross-module gap in the recursive check. If module A has `type Private = felt` and `pub type PublicAlias = Private`, then module B can expose `pub proc f(x: A::PublicAlias)` and this branch stops before checking what the public alias expands to. That still leaks a private type dependency through an exported signature, just through another module.

huitseeker (changes requested):

crates/assembly/src/assembler.rs
```diff
@@ -806,14 +814,154 @@ impl Assembler {
...
+                if !symbol.visibility().is_public() {
+                    continue;
+                }
+
+                let signature = match symbol.item() {
+                    SymbolItem::Procedure(proc) => {
+                        let proc = proc.borrow();
+                        proc.signature().cloned()
+                    },
+                    SymbolItem::Alias { .. }
```
huitseeker: Right now aliases are skipped here, but `export_symbol()` later turns a `SymbolItem::Alias` into a `ProcedureExport` with the target signature.

So you can bypass verification with something like this:
```rust
type PrivateType = felt

proc hidden(value: PrivateType)
    nop
end

pub use ::cycle::module_a::hidden -> exposed
```

A temporary test expecting `assemble_library([module])` to fail instead got a library export for `::cycle::module_a::exposed` with `signature: Some(FunctionType { params: [Felt], ... })`. So the private type is erased after resolution, and the new visibility check never sees it.

Could this check include public aliases that export procedures too?

huitseeker (changes requested):

cmende:
@huitseeker 
Great catch — you’re right, public procedure aliases weren’t being validated before.

I’ve updated the exported-signature visibility checks to also cover public aliases that resolve to procedures. We now resolve the target procedure and validate its declared signature recursively (including nested type dependencies) in the defining module context, the same way we already do for direct public procedure exports.

Also added a regression test:
`exported_procedure_alias_signature_with_private_local_type_is_rejected`

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -187,6 +189,107 @@ pub fn analyze(
...
+            verify_exported_signature_type_expr(module, analyzer, &ty.elem, visiting_types);
+        },
+        TypeExpr::Struct(ty) => {
+            for field in ty.fields.iter() {
+                verify_exported_signature_type_expr(module, analyzer, &field.ty, visiting_types);
+            }
+        },
+        TypeExpr::Ref(path) => {
+            let resolution = match module.resolve_path(path.as_deref(), analyzer.source_manager()) {
+                Ok(resolution) => resolution,
```
huitseeker, line 230: The check at line 230 compares `path.parent() == module.path()` for same-module absolute paths. If a module was parsed with a relative path like `wallet::memory` but references `::wallet::memory::PrivateType` (with `::` prefix), the comparison fails because `::wallet::memory` != `wallet::memory`. Consider normalizing both sides before comparing.

crates/assembly/src/assembler.rs
```diff
@@ -804,14 +812,189 @@ impl Assembler {
...
+                                let proc = proc.borrow();
+                                proc.signature().cloned()
+                            },
+                            SymbolItem::Alias { .. }
+                            | SymbolItem::Constant(_)
+                            | SymbolItem::Type(_)
+                            | SymbolItem::Compiled(_) => None,
+                        };
+                        (target_gid.module, signature)
+                    },
```
huitseeker, line 799: `assemble_executable_modules` does not call `verify_exported_signature_type_visibility` before compilation. A program with a `pub proc` entrypoint referencing a private type would bypass the check entirely, since only `assemble_library_product` and `assemble_kernel_module` trigger the verification.

bitwalker:
I think this PR will either be made a whole lot more trivial by #3220, because visibility is actually modeled properly in general there, and the only thing we probably need to layer on to what is in that PR, is some additional checks to make sure that an exported type doesn't reference private types/type aliases.

cmende:
@bitwalker That makes sense. I’ll wait for #3220 merged and rebase on it.

huitseeker:
@cmende Please rebase :)

huitseeker:
Closing in favor of #3273

PR #3156, opened 2026-05-15
fix(project): do not require bin targets to specify path

bitwalker:
Rust-based projects use virtual targets (i.e. no source paths given), but the parsing of project files was requiring executable targets to specify source paths in an effort to catch ambiguous/overlapping targets. This only applies when paths are present though, as when virtual targets are used, it is up to the assembling party to provide sources for the target they are requesting be built.

This PR simply removes this check, since the lack of a source path implies the target is virtual, and thus the lack of a path is intentional.


huitseeker (approved):
LGTM, thanks

PR #3158, opened 2026-05-16
feat(core): executable `MastForest` generation

### LINKED ISSUE #2341: Support executable MastForest generation with syscalls, externals, and dyn nodes
### Feature description

The MastForest Arbitrary implementation generates non-executable forests when including certain node types:

1. **Syscalls**: Generated with random procedure digests - fail to execute without matching kernel procedures
2. **External nodes**: Use random digests that don't correspond to valid procedures
3. **Dynamic nodes**: Leave junk on the stack and cannot execute properly

Currently users must set these to 0 for executable forests, limiting execution testing using this strategy.

See https://github.com/0xMiden/miden-vm/pull/2335#discussion_r2486834656
https://github.com/0xMiden/miden-vm/pull/2335#discussion_r2486842056
https://github.com/0xMiden/miden-vm/pull/2335#discussion_r2486848817

### 1. Add kernel_procedures field to MastForestParams

```rust
pub struct MastForestParams {
    // ... existing fields
    /// Optional procedure hashes for syscall generation
    pub kernel_procedures: Option<Vec<Word>>,
}
```

When generating syscalls, randomly select from provided hashes. Could also generate a matching kernel containing the procedure hashes.

### 2. Support executable external nodes

Add `external_procedures: Option<Vec<Word>>` field or make external nodes reference existing basic blocks within the same forest.

### 3. Consider kernel generation support

Add `generate_kernel: bool` field to generate both forest AND compatible kernel, returning `(MastForest, Kernel)`.

surajk-m:
Closes #2341

Rework the `MastForest` `Arbitrary` impl so that the default sample is structurally executable: every external resolves to a procedure root in the same forest, every syscall callee is in the paired kernel, and the external call graph is acyclic. Dyn nodes are suppressed in `Executable` mode (program synthesis problem, deferred to a future `MASM level generator`). A `StructureOnly` mode preserves the previous permissive behaviour for serialization and merge tests that need broad structural coverage.

The strategy now also returns a paired `Kernel`, exposed through `forest_kernel_strategy(...)`, so callers can construct a `Program` against the generated forest without registering external procedure hashes themselves.


core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -748,3 +1321,518 @@ impl Arbitrary for Kernel {
...
+                ..Default::default()
+            })
+        ) {
+            // Build adjacency: for each external, the source procedure root that contains it
+            // points to the target procedure root whose digest the external matches.
+            let mut edges: Vec<(MastNodeId, MastNodeId)> = Vec::new();
+            for (idx, node) in forest.nodes().iter().enumerate() {
+                let MastNode::External(ext) = node else { continue };
+                let id = MastNodeId::new_unchecked(idx as u32);
+                let Some(src_root) = containing_root(&forest, id) else { continue };
```
plafer: Do we guarantee that generated forests are pruned (i.e. all nodes are reachable by a procedure root)? If not, then it's likely that many nodes won't be reachable, in this case this test won't add anything to `edges` and will succeed.

Haven't looked at the rest of the diff yet, but at this point, it seems like a good idea to enforce all `MastForest`s are pruned, and here we can `.expect()` nodes to belong to a root.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -187,29 +187,83 @@ impl Arbitrary for BasicBlockNode {
...
+/// Parameters for generating `MastForest` instances via proptest.
+///
+/// # Defaults
+///
+/// `MastForestParams::default()` selects [`GenerationMode::Executable`] and sets
+/// every `max_*` to a small non-zero number except the three that must be opted into
+/// explicitly: `max_syscalls`, `max_externals`, and `max_dyns` all default to `0`. This keeps
+/// the default sample distribution backward-compatible with the pre-feature generator.
+/// Callers that want to exercise syscalls, externals, or dyn nodes should set the
+/// corresponding cap to a non-zero value (and, for dyns, switch to `StructureOnly`).
```
plafer: This is not necessary. We previously defaulted those to 0 as a workaround for some issue we were having. From what I understand, we should still default to 0 dyn nodes (as we can't guarantee that generated forests with `dyn` are executable), but we can lift the restriction on `max_syscalls` and `max_externals`

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -187,29 +187,83 @@ impl Arbitrary for BasicBlockNode {
...
+///
+/// # Future work — MASM-level dyn generation
+///
+/// Dyn nodes are intentionally unsupported in `Executable` mode because the generator
+/// cannot synthesise valid operand-stack preconditions: each dyn/dyncall reads its callee
+/// digest off the operand stack at runtime, and producing a semantically valid stack state
+/// requires program-synthesis-level reasoning. The intended extension is an assembly-level
+/// (MASM) generator that can emit instruction sequences which push a valid callee digest
+/// before each dyn/dyncall; until such a generator exists, callers who need executable dyn
+/// coverage should rely on hand-written fixtures or opt into `StructureOnly` mode.
```
plafer, line 319: Let's create an issue for this

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -224,354 +278,873 @@ pub struct MastForestParams {
...
+    // no syscalls fired and no inclusion bits were set.
+    if hashes.is_empty()
+        && let Some(&first) = root_ids.first()
+    {
+        hashes.push(forest.get_node_by_id(first).expect("root id is valid").digest());
+    }
+
+    if hashes.len() > Kernel::MAX_NUM_PROCEDURES {
+        hashes.truncate(Kernel::MAX_NUM_PROCEDURES);
+    }
```
plafer, line 1135: This violates invariant 4, where some syscalls won't find their procedure in the kernel.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -224,354 +278,873 @@ pub struct MastForestParams {
...
-    /// executable forests.
+    /// Defaults to `0`. In [`GenerationMode::Executable`] this field is treated as
+    /// `0` and no dyn nodes are emitted (see the *Future work* note on the struct doc). In
+    /// [`GenerationMode::StructureOnly`], a mix of `DynNode::new_dyn` and
+    /// `DynNode::new_dyncall` instances may be emitted up to this bound; such forests are
+    /// intended for structural tests and are not guaranteed to execute.
     pub max_dyns: usize,
+    /// Controls whether the generator emits forests with the closure invariants enforced
+    /// ([`GenerationMode::Executable`]) or legacy structure-only forests
+    /// ([`GenerationMode::StructureOnly`]).
```
plafer: I wouldn't call structure-only forests "legacy", as they can still be useful

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -187,29 +187,83 @@ impl Arbitrary for BasicBlockNode {
...
+/// `MastForestParams::default()` selects [`GenerationMode::Executable`] and sets
+/// every `max_*` to a small non-zero number except the three that must be opted into
+/// explicitly: `max_syscalls`, `max_externals`, and `max_dyns` all default to `0`. This keeps
+/// the default sample distribution backward-compatible with the pre-feature generator.
+/// Callers that want to exercise syscalls, externals, or dyn nodes should set the
+/// corresponding cap to a non-zero value (and, for dyns, switch to `StructureOnly`).
+///
+/// # Structural closure invariants in `Executable` mode
+///
+/// When `mode` is [`GenerationMode::Executable`], the generator enforces the
```
plafer, line 269: We currently don't ensure that the generated `MastForest` successfully executes on *any* stack input right? I think we should make sure that this is the case, otherwise this "executable mode" is not as useful. That is, the main use case would be to write proptests such as "give me a random executable mast forest, and ensure that the processor doesn't crash".

I believe that all that is left here would be to never include any operation that can fail in a basic block, such as u32 operations and memory operations (as they require inputs to be in the `u32` range). This PR would lay the groundwork, and in a future PR we can look at generating stack values as well such that we may introduce those operations in executable MAST forests too.

plafer (dismissed):
Thank you for taking this one on!

Perhaps the biggest requested change is the last one about adding an invariant to what an "executable mast forest" is.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -224,354 +278,873 @@ pub struct MastForestParams {
...
+    let max_loops = params.max_loops;
+    let max_calls = params.max_calls;
+    let max_syscalls = params.max_syscalls;
+    let max_externals = params.max_externals;
+    let max_dyns = params.max_dyns;
+    let mode = params.mode;
+
+    // Stage 1: basic blocks + decorators (independent of everything else).
+    (
+        prop::collection::vec(any_with::<BasicBlockNode>(bb_params), 1..=blocks_end),
```
huitseeker: This drops the lower bound from `MastForestParams::blocks`. A caller that asks for `128..=128` now samples `1..=128`, so fixed-size tests and the updated benchmarks no longer get the requested forest size. Can this use `params.blocks.clone()` here instead of rebuilding the range from `1`?

core/src/mast/node/basic_block_node/arbitrary.rs
huitseeker: TO give more context: syscalls are emitted before this point and each one may insert its callee digest into `kernel_pool`; if the pool grows past `Kernel::MAX_NUM_PROCEDURES`, truncating here can drop a digest that an already-created syscall targets. It seems safer to cap syscall emission/insertion before adding the syscall, or reserve remaining capacity before applying `kernel_inclusion`.

huitseeker (changes requested):
Thanks for tackling this one!

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Hey @plafer, Thanks for the review. Generated forests are now pruned in both modes (`mark_unreachable_nodes_as_roots` promotes unreachable nodes to roots), and the DAG test `.expect()`s the containing/target roots instead of skipping. New property tests cover this.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done! `max_syscalls` and `max_externals` now default to `1`, `max_dyns` stays at `0` since we can't guarantee dyn forests are executable.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Thanks @huitseeker! The truncation is gone, the kernel is now capped at emission: once `kernel_pool` hits `MAX_NUM_PROCEDURES`, syscalls can only target roots already in the kernel (same as the frozen path), so no digest an emitted syscall depends on is ever dropped. `finalise_kernel` then just fills the remaining capacity from `kernel_inclusion`.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Agreed. dropped "legacy".

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Added as invariant 5. Executable mode now draws basic block operations only from `is_infallible_op` - excluding `Inv`, all `u32` ops, memory/advice ops, binary-operand ops, `Caller`, `Emit`, and the crypto/STARK helpers (anything whose processor impl has an error path). `executable_blocks_contain_only_infallible_ops` verifies it. As you noted, control flow condition values still come from the stack, so full run to completion on arbitrary inputs is the follow-up, I've documented that gap under "Remaining executability gap".

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done. The strategy now uses the caller's full blocks range (only clamping a degenerate 0 bound up to 1, since the build phases need at least one block). 128..=128 samples exactly 128 again. Added `blocks_range_lower_bound_is_respected` to cover it.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -205,351 +318,512 @@ impl Default for MastForestParams {
...
+        .boxed()
+}
+
+impl Arbitrary for MastForest {
+    type Parameters = MastForestParams;
+    type Strategy = BoxedStrategy<Self>;
+
+    /// Delegates to [`forest_kernel_strategy`] and drops the kernel.
+    fn arbitrary_with(params: Self::Parameters) -> Self::Strategy {
+        forest_kernel_strategy(params).prop_map(|(forest, _)| forest).boxed()
```
huitseeker, line 769: `Executable` defaults to `max_syscalls = 1`, but this drops the only `KernelDescriptor` that makes those syscalls valid. A caller using the standard `any_with::<MastForest>` API can therefore get a forest that fails when passed to `Program::new`, despite the mode's execution promise. The processor property only covers `forest_kernel_strategy`, where the descriptor is kept. Could this `Arbitrary` path suppress syscalls, or could the executable API return a type that keeps the forest and kernel together?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -205,351 +318,512 @@ impl Default for MastForestParams {
...
+
+/// Builds the kernel from the syscall targets, then from roots selected by `inclusion` while
+/// room remains. Falls back to the first root, so the kernel is never empty.
+fn generated_kernel(
+    forest: &DenseMastForestBuilder,
+    roots: &[MastNodeId],
+    mut hashes: Vec<Word>,
+    inclusion: &[bool],
+) -> KernelDescriptor {
+    for (&root, &included) in roots.iter().zip(inclusion) {
```
huitseeker: nit: Could the generated kernel contain only emitted syscall targets and stay empty when there are no syscalls? `kernel_inclusion` and this fallback add unrelated roots even though an empty `KernelDescriptor` is valid. Removing both would make the output easier to explain and delete some seed plumbing.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -205,351 +318,512 @@ impl Default for MastForestParams {
...
+                    0..=caps.max_splits,
+                    0..=caps.max_loops,
+                    0..=caps.max_calls,
+                    0..=caps.max_syscalls,
+                    0..=caps.max_externals,
+                    0..=if executable { 0 } else { caps.max_dyns },
+                ),
+            )
+        })
+        .prop_flat_map(
```
huitseeker: nit: Could these node choices use `prop::collection::vec(..., 0..=cap)` directly and let the collection strategy own the count? Keeping one `prop_flat_map` only where a later vector truly needs an exact size would make the shrinking model shorter and easier to follow.

huitseeker (changes requested):
Thanks for taking this on, and sorry for the late review!

The main approach looks right.  I left one blocking line comment about the forest-only Arbitrary implementation dropping the paired kernel. Once that API contract is resolved, I think the PR is in good shape. 

I also left two optional simplification comments where some kernel and strategy plumbing could be removed.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done! the `Arbitrary` path now suppresses syscalls in `Executable` mode and is covered by a processor proptest running under `Program:new`, `forest_kernel_strategy` remains the way to get syscalls with their kernel.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done! node vectors use vec(..., 0..=cap) directly.

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -18,183 +19,301 @@ use crate::{
...
-        Self { max_ops_len: 8 }
+        Self { max_ops_len: 8, executable: false }
     }
 }
 
-// ---------- Arbitrary for BasicBlockNode ----------
+/// Strategy for the operations of a basic block described by `params`.
+pub fn block_ops_strategy(params: &BasicBlockNodeParams) -> BoxedStrategy<Vec<Operation>> {
+    if params.executable {
+        prop::collection::vec(op_infallible_strategy(), 1..=params.max_ops_len)
```
huitseeker, line 237: `block_ops_strategy` can return more than `max_ops_len` operations because `balance_ops` appends drops after sampling up to the limit. For example, `max_ops_len = 1` can yield `[Push(_), Drop]`. Could the final block stay within this documented maximum, with a regression property for the bound?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -205,351 +324,471 @@ impl Default for MastForestParams {
...
+        prop::collection::vec(any::<(usize, usize)>(), 0..=params.max_externals)
+            .prop_map(|picks| {
+                picks
+                    .into_iter()
+                    .map(|(root, sibling)| ExternalPick::Local { root, sibling })
+                    .collect()
+            })
+            .boxed()
+    } else {
+        prop::collection::vec(any::<[u32; 4]>(), 0..=params.max_externals)
```
huitseeker, line 380: nit: `ExternalPick::Random` and `word_strategy` now restrict every digest limb to 32 bits, so the arbitrary serialization and merge paths no longer cover field values with high bits set. Could these strategies keep full field coverage by using `Arbitrary for Word`?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -18,183 +19,301 @@ use crate::{
...
+/// Strategy for sequences of non-control flow operations.
 pub fn op_non_control_sequence_strategy(
     max_length: usize,
 ) -> impl Strategy<Value = Vec<Operation>> {
     prop::collection::vec(op_non_control_strategy(), 1..=max_length)
 }
 
-// ---------- Parameters ----------
+/// Returns true if `op` cannot fail during execution, whatever the VM state.
+pub fn is_infallible_op(op: &Operation) -> bool {
```
huitseeker, line 173: nit: `is_infallible_op` has no caller, and only the parent module calls `block_ops_strategy`. Could we remove this wrapper and make `block_ops_strategy` `pub(super)` so this feature does not add unused public API?

huitseeker: nit: Could we remove this public helper? Nothing in the workspace calls `is_infallible_op`, and the generator already uses `stack_delta` directly. This seems to add API surface without serving the implementation.

huitseeker: nit: Could we remove this public helper? Nothing in the workspace calls `is_infallible_op`, and the generator already uses `stack_delta` directly. This seems to add API surface without serving the implementation.

huitseeker: nit: Could we remove this public helper? Nothing in the workspace calls `is_infallible_op`, and the generator already uses `stack_delta` directly. This seems to add API surface without serving the implementation.

huitseeker (commented):
Thanks, the `max_ops_len` fix is the main blocking issue, otherwise this is in a good shape

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done, both use `any::<Word>()` now.

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done, `balance_ops` now reserves room for its closing drops, so blocks stay within `max_ops_len`. Added a property over `max_ops_len` in 1..=12 asserting the bound and balance.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@
 
 - Preserved the public `ParsingError` enum layout while adding protocol ABI attribute checks ([#3812](https://github.com/0xMiden/miden-vm/pull/3812)).
 - Fixed issue where parsing of pointer types dropped address space information ([#3790](https://github.com/0xMiden/miden-vm/pull/3790)).
+#### Changes
+
+- Reworked the `MastForest` `Arbitrary` impl. `GenerationMode::Executable` (the default) yields forests whose procedure roots run to completion on any operand stack: infallible, stack-balanced basic blocks, binary split and loop conditions, externals that resolve to local roots, and syscalls into a paired `KernelDescriptor` exposed through `forest_kernel_strategy`. `GenerationMode::StructureOnly` keeps the permissive behavior ([#3158](https://github.com/0xMiden/miden-vm/pull/3158)).
```
huitseeker: This entry now sits under the released `v0.32.1` section. Could you move it under `v0.34.0 (Unreleased)` and remove the duplicate `Changes` heading here?

core/src/mast/node/basic_block_node/arbitrary.rs
```diff
@@ -18,183 +19,299 @@ use crate::{
...
+    Operation::HPerm,
+];
+
+/// Operations that cannot fail whatever the stack, advice provider, or memory hold, paired with
+/// the change each one makes to the stack depth.
+///
+/// Left out: `Inv` (traps on zero), `And`, `Or`, `Not`, `CSwap` and `CSwapW` (binary operands),
+/// the `U32*` family (u32-range operands), memory and advice operations, `Caller`, `Emit`,
+/// `Assert`, and the crypto/STARK helpers.
+const INFALLIBLE_OPS: &[(Operation, i8)] = &[
```
huitseeker: nit: `INFALLIBLE_OPS` repeats stack deltas already defined by `Operation::increments_stack_size()` and `Operation::decrements_stack_size()`. Could the table only name safe operations and derive each delta from those methods, so the stack effect has one source of truth?

huitseeker (approved):
This essentially needs a Changelog edit, otherwise lgtm!

core/src/mast/node/basic_block_node/arbitrary.rs
surajk-m: Done, `INFALLIBLE_OPS` only names the safe operations now and `stack_delta` reads the effect from `increments_stack_size` / `decrements_stack_size`.

huitseeker (approved):
LGTM! Thanks so much

PR #3160, opened 2026-05-17
processor: enforce continuation stack limit in execution loop

TharescdqTuA:
## Describe your changes

Follow-up to #2825.

Move the continuation stack size check out of the fast processor stoppers and into the shared execution path. `ContinuationStack` now carries the configured limit, so the same stack snapshots are checked during fast execution, step resume, and replay.

This keeps `max_num_continuations` as an execution option while preventing new stoppers or replay-only paths from bypassing the bound.

- [ ] Updated `CHANGELOG.md`


processor/src/execution/mod.rs
```diff
@@ -150,7 +150,14 @@ where
...
 
-    while let Some(continuation) = state.continuation_stack.pop_continuation() {
+    loop {
+        check_continuation_stack_size(state.continuation_stack)
+            .map_break(InternalBreakReason::from)?;
+
+        let Some(continuation) = state.continuation_stack.pop_continuation() else {
+            break;
+        };
+
```
plafer: We check the stack size twice per continuation; I would leave the `while let ...` as is, and check the continuation stack size at the end of the loop iteration before `should_stop()` (essentially preserving the previous behavior).

Then we also no longer need to put this in a separate `check_continuation_stack_size()` function.

processor/src/continuation_stack.rs
```diff
@@ -96,21 +100,32 @@ impl Continuation {
 /// This allows the processor to execute a program iteratively in a loop rather than recursively
 /// traversing the nodes. It also allows the processor to pass the state of execution to another
 /// processor for further processing, which is useful for parallel execution of MAST forests.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub struct ContinuationStack {
     stack: Vec<Continuation>,
+    max_len: usize,
```
plafer: nit: Let's put the "check max length" behavior in the processor directly. So we no longer need this field, and we need a new `ContinuationStack::len()` method - which we can use to check the size directly in the `execute_impl()` continuation loop.

This is a similar design decision as we've made with `AdviceProvider` vs `AdviceMap` size (i.e. the advice map doesn't know its size bounds, the advice provider does).

plafer (changes requested):
Thank you, I like the change! A few small things to change, and then this is good to go!

processor/src/continuation_stack.rs
```diff
@@ -96,21 +100,32 @@ impl Continuation {
...
+}
+
+impl Default for ContinuationStack {
+    fn default() -> Self {
+        Self {
+            stack: Vec::default(),
+            max_len: usize::MAX,
```
huitseeker: The `Default` impl sets `max_len = usize::MAX`, which effectively disables the stack limit. If any code path accidentally uses `ContinuationStack::default()` instead of `ContinuationStack::new()`, stack overflow protection is lost. I think it's worth setting up a constant (e.g. 1024) as the default limit instead.

TharescdqTuA:
@huitseeker @plafer Done.

processor/src/fast/execution_api.rs
```diff
@@ -230,6 +243,11 @@ impl FastProcessor {
                     if let Some(continuation) = maybe_continuation {
                         continuation_stack.push_continuation(continuation);
                     }
+                    if continuation_stack.len() > max_num_continuations {
+                        return Err(ExecutionError::Internal(
+                            "continuation stack size exceeded the allowed maximum",
+                        ));
+                    }
```
plafer, line 250: I might be missing an edge case, but I don't think we need the `max_num_continuations` in `resume_context_from_flow()`? Even when `step()`ing through a program, the continuation stack length check will be executed in `finalize_clock_cycle_with_continuation_and_op_helpers()`. The downside of the current approach is that the check is copy/pasted in 2 separate places, which increases the complexity slightly, and IMO we don't really care about off by one here

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@
 
 #### Fixes
 
+- Preserved `AssemblyOp` source mappings when merging `MastForest`s, preventing source-location loss after node deduplication.
```
huitseeker, line 22: Looks like this line was pulled in by accident. The same `AssemblyOp` entry is already present below with its `#2958` link, and this PR only needs the continuation-stack changelog line.

huitseeker (commented):
Another thing once you rebase: `ReplayProcessor::new()` hard-codes `ExecutionOptions::DEFAULT_MAX_NUM_CONTINUATIONS` instead of threading the configured execution option through trace replay.

huitseeker:
Closing for inactivity

PR #3161, opened 2026-05-17
Add CodSpeed benchmark workflows

huitseeker:
Introduce two push-triggered GitHub Actions workflows that publish the existing non-regression benchmark workloads to [CodSpeed](https://codspeed.io/) on push (*as a trial: we'll consider paying for more runtime and running this on PRs if it goes well*). One workflow runs the Blake3 1-to-1 proving benchmark from the miden-vm package, and the other runs the synthetic transaction benchmark from miden-vm-synthetic-bench. Both workflows install the repository Rust toolchain, cache Rust dependencies, install cargo-codspeed, build the dedicated benchmark target, and then hand execution to CodSpeedHQ/action in simulation mode.

Add native CodSpeed benchmark binaries instead of trying to run the existing Criterion harnesses directly. The CodSpeed harnesses measure exactly the benchmarked section with CodSpeed's start/end hooks, which keeps the expensive proving path from being **repeated** by a Criterion sampling loop. The synthetic CodSpeed target runs all four axes, `exec`, `trace_prep`, **`prove`**, and `verify`, so CodSpeed covers the actual prover work that the old fast synthetic CI default intentionally skipped.

Preserve the existing push-comment non-regression workflows. The legacy Blake3 workflow is untouched, and the legacy synthetic workflow still uses the original synthetic_bench Criterion target and its existing default axes. The only change to that target is an efficiency improvement: when `prove` and `verify` are selected together, `verify` reuses the proof produced by the `prove` axis instead of generating another proof solely for setup.

Add the codspeed crate as a dev-dependency for the two packages that own the new native harnesses, and update Cargo.lock accordingly. The workflows pin third-party actions and cargo-codspeed versions, keep checkout credentials disabled, and grant only read contents plus OIDC permissions for CodSpeed upload


.github/workflows/codspeed.yml
Al-Kindi-0, line 1: Seems like we have to separate workflows, this one and the synth-bench one, right ? It [seems](https://codspeed.io/docs/integrations/ci/github-actions/configuration#running-benchmarks-in-parallel)  like this could be problematic for CodSpeed

.github/workflows/codspeed-blake3.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          set -euo pipefail
+          cargo install cargo-codspeed --version 4.6.0 --locked
+
+      - name: Build benchmark
+        shell: bash
+        run: |
+          set -euo pipefail
+          cargo codspeed build --profile optimized -p miden-vm --bench blake3_1to1_codspeed
+
```
Al-Kindi-0: What about concurrency ?

benches/synthetic-bench/benches/synthetic_codspeed.rs
Al-Kindi-0, line 1: Might be good later, once the trial works out, to remove redundnacy between this and the `synthetic_bench.rs` module

Al-Kindi-0 (approved):
LGTM, modulo a few remarks/questions

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+
+      - name: Install Rust toolchain
+        shell: bash
+        run: |
+          set -euo pipefail
+          toolchain="$(sed -n 's/^channel = "\(.*\)"$/\1/p' rust-toolchain.toml)"
+          rustup toolchain install "${toolchain}" --profile minimal --target wasm32-unknown-unknown
```
plafer, line 37: From their [docs](https://codspeed.io/docs/benchmarks/rust/criterion#running-the-benchmarks-in-your-ci), they use the [`moonrepo/setup-rust`](https://github.com/moonrepo/setup-rust), which seems to do all the Rust-related stuff (i.e. toolchain install, setup cache, etc). Probably worth using when we don't explicitly care about using the `WarpBuilds` cache?

.github/workflows/codspeed-blake3.yml
```diff
@@ -0,0 +1,55 @@
...
+  contents: read
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  codspeed:
+    name: Run Blake3 CodSpeed benchmark
```
plafer: I don't see this job running in this PR's CI (or maybe I'm blind?). Is this expected?

plafer (commented):
Still ramping up on Codspeed, but looks very cool!

.github/workflows/codspeed.yml
huitseeker: Not required. The CodSpeed example uses moonrepo/setup-rust, but their actual requirement is cargo-codspeed plus CodSpeedHQ/action.

With our security posture, less dependencies is better, and the WarpBuild setup routes our cache to the right hardware, which moonrepo/setup-rust is unlikely to (because it's developed orthogonally from Warpbuild).

.github/workflows/codspeed-blake3.yml
huitseeker: Yes, expected for now. CodSpeed recommends pull_request for PR reports, but we only have limited free runtime, so this PR restricts CodSpeed to pushes to main/next and manual dispatch. If the trial is useful, we can revisit PR-triggered runs later.

benches/synthetic-bench/benches/synthetic_codspeed.rs
huitseeker: Agreed. I kept the native CodSpeed harness separate for this trial because it avoids changing the legacy Criterion path more than necessary. If CodSpeed sticks, a follow-up refactor to share the scenario planning/setup code would make sense.

.github/workflows/codspeed-blake3.yml
huitseeker: If this is about GitHub workflow concurrency, the workflow has a concurrency group. If this is about prover thread count, you’re right: synthetic has RAYON_NUM_THREADS=8, but Blake3 should mirror the legacy non-regression benchmark. I’ll add that while merging the workflows.

.github/workflows/codspeed-blake3.yml
plafer: Makes sense. Did you trigger a manual run yet so we can see the results?

.github/workflows/codspeed.yml
huitseeker: Good catch!

.github/workflows/codspeed-blake3.yml
huitseeker: As you can see, that's not available on a push trigger, you only see the codspeed remnants because this used to trigger on PR in an old since-replaced version.

plafer (approved):
LGTM - curious to see how things work out with Codspeed.

PR #3167, opened 2026-05-18
`u256` core-lib parity

### LINKED ISSUE #1697: Add `div` procedure to `math/u256.masm`
### Feature description

For DeFi applications that require computing prices and ratios, it is useful to be able to divide large fixed point integers. 

The current `u64` math library works well for this, but it would be useful to be able to have the same procedures in the `u256` library as in the `u64` library. 

For fixed point arithmetic with integers with at least $$8$$ decimal points it is easy to overflow/underflow when using 64 bit integers.

The Pragma team is working on an price oracle and they have also expressed desire for having a `div` function in the `u256` library as they have to calculate the average and median of several large fixed point integers. 

Precision for fixed point arithmetic.

Al-Kindi-0:
## Describe your changes

A followup to #2569 which aims at make all of the math modules have the same public API.
Adds `u256` procedures that bring the core-lib u256 API to rough parity with the u64 one. There is also a small proc-order cleanup.

- Updated `CHANGELOG.md'


crates/lib/core/src/handlers/u256_div.rs
```diff
@@ -0,0 +1,166 @@
+//! U256_DIV system event handler for the Miden VM.
+//!
+//! This handler implements the U256_DIV operation that pushes the result of [u256] division
```
huitseeker: The docs job is failing here because `[u256]` is treated as a Rust link, but there is no item named `u256` in scope. Escaping it as `\[u256\]` or using backticks should fix the rustdoc error.

Al-Kindi-0: fixed

crates/lib/core/asm/math/u256.masm
```diff
@@ -450,3 +618,866 @@ pub proc wrapping_mul(b: u256, a: u256) -> u256
...
+    padw loc_loadw_le.8              # [c0..c7,  c8..c15, ...]
+end
+
+# ===== BIT COUNTING ==============================================================================
+
+#! Counts the number of leading zeros of an unsigned 256 bit integer.
+#! Stack transition looks as follows:
+#! [a0, a1, a2, a3, a4, a5, a6, a7, ...] -> [count, ...], where count is the number of leading
+#! zero bits in a (0..=256).
+pub proc clz(a: u256) -> u8
```
huitseeker: Could these return types be widened? The docs say this can return `0..=256`, and `clz(0)` does return `256`, so `u8` cannot represent the full result. Same question applies to `ctz`, `clo`, and `cto`.

Al-Kindi-0: fixed in https://github.com/0xMiden/miden-vm/pull/3167/commits/87f6b6ea24984492bd23c79b0a15610b46238abb

huitseeker (approved):
Lovely, tyvm!

PR #3168, opened 2026-05-18
fix(CI): convert clippy warnings into errors

plafer:
Running `make clippy` outputs many warnings, which the CI doesn't consider a failure because they are not errors.

This PR converts all warnings to errors, and fixes all clippy errors.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -58,6 +58,8 @@ jobs:
   clippy:
     name: clippy nightly on ubuntu-latest
     runs-on: ubuntu-latest
+    env:
+      RUSTFLAGS: -D warnings
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
--- a/crates/assembly-syntax/src/sema/mod.rs
+++ b/crates/assembly-syntax/src/sema/mod.rs
@@ -132,7 +132,7 @@ pub fn analyze(
                 analyzer.error(SemanticAnalysisError::UnexpectedEntrypoint { span: body.span() });
             },
             Form::AdviceMapEntry(entry) => {
-                add_advice_map_entry(&mut module, entry.with_docs(docs.take()), &mut analyzer)?;
+                add_advice_map_entry(&mut module, entry.with_docs(docs.take()), &mut analyzer);
             },
         }
     }
@@ -175,7 +175,7 @@ pub fn analyze(
     analyzer.has_failed()?;
 
     // Run item checks
-    visit_items(&mut module, &mut analyzer)?;
+    visit_items(&mut module, &mut analyzer);
 
     // Check unused imports
     for import in module.aliases() {
@@ -192,7 +192,7 @@ pub fn analyze(
 ///
 /// When this function returns, all local analysis is complete, and all that remains is construction
 /// of a module graph and global program analysis to perform any remaining transformations.
-fn visit_items(module: &mut Module, analyzer: &mut AnalysisContext) -> Result<(), SyntaxError> {
+fn visit_items(module: &mut Module, analyzer: &mut AnalysisContext) {
     let is_kernel = module.is_kernel();
     let locals = BTreeMap::from_iter(
         module
@@ -306,8 +306,6 @@ fn visit_items(module: &mut Module, analyzer: &mut AnalysisContext) -> Result<()
             alias.uses = 1;
         }
     }
-
-    Ok(())
 }
 
 fn define_alias(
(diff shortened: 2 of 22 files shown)


.github/workflows/lint.yml
```diff
@@ -58,6 +58,8 @@ jobs:
   clippy:
     name: clippy nightly on ubuntu-latest
     runs-on: ubuntu-latest
+    env:
```
bitwalker, line 61: I don't think we should be treating nightly clippy warnings as errors - we may not be able to fix them in stable. I would only do this when running clippy on the stable toolchain (i.e. don't allow PRs to be merged when clippy has issues with the code on our current toolchain)

plafer: `make clippy` runs [our own `xclippy` alias](https://github.com/0xMiden/miden-vm/blob/70ba76fe719b3373fe275ff3144d68bf7a4ce9db/.cargo/config.toml#L10) where we are clear on exactly what we look for.

> we may not be able to fix them in stable

Hence AFAIU this shouldn't be a concern, right?

bitwalker (approved):

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -364,11 +362,7 @@ fn define_procedure(
 /// key.
 ///
 /// Returns `Err` if the symbol is already defined
-fn add_advice_map_entry(
-    module: &mut Module,
-    entry: AdviceMapEntry,
-    context: &mut AnalysisContext,
-) -> Result<(), SyntaxError> {
+fn add_advice_map_entry(module: &mut Module, entry: AdviceMapEntry, context: &mut AnalysisContext) {
```
huitseeker, line 363: Could this doc comment be updated while changing the signature? It still says this returns `Err`, but the function now records the duplicate-symbol error in `context` and returns `()`.

crates/assembly/src/tests.rs
```diff
@@ -1,3 +1,6 @@
+// Test helpers share a `TestResult` return type so `?` can be used uniformly.
+#![allow(clippy::unnecessary_wraps)]
```
huitseeker: Could this allow be narrowed to the few helpers or tests that need the shared `TestResult` shape? As written, it turns off `unnecessary_wraps` for this whole large test file, so new cases can grow under the new `-D warnings` policy without being noticed.

(also, these were meant to be fixed in #3143)

.github/workflows/lint.yml
```diff
@@ -58,6 +58,8 @@ jobs:
...
+      RUSTFLAGS: -D warnings
```
huitseeker, line 62: Could this deny setting live in the repo clippy aliases instead of only in the workflow env? For example, putting `-Dwarnings` after `--` in `.cargo/config.toml` for `xclippy` makes `cargo +nightly xclippy`, `make clippy`, and CI enforce the same policy. It also keeps CI on the repo's custom `xclippy` lint set instead of a plain clippy shape.

I checked the fix path in a detached worktree too: after adding `-Dwarnings` to `xclippy-fix`, `cargo +nightly xclippy-fix` completed successfully and left no source changes, only my test edit to `.cargo/config.toml`. So keeping `xclippy` and `xclippy-fix` aligned looks workable from that local run.

huitseeker (commented):
Did you figure out what broke the config by which CI barfs when clippy lints are not passing? That was obviously the intent of the setup!

.github/workflows/lint.yml
plafer: > Could this deny setting live in the repo clippy aliases instead of only in the workflow env?

My thinking was to leave them as warnings for local runs, since this is typically what you want - run clippy interactively, and see any complaints as warnings. I then saw the "convert them to errors" mostly relevant to the CI (so that the job actually fails when there are warnings), and hence only included it here.

I can be convinced otherwise though - given the above, do you still think having errors locally as well is better? Happy to make the change.

> I checked the fix path in a detached worktree too: after adding -Dwarnings to xclippy-fix, cargo +nightly xclippy-fix completed successfully 

Ah! I tried `xclippy-fix` but it didn't actually fix anything - so that was due to it not seeing any *errors*? I'll play with it locally too.

plafer:
> Did you figure out what broke the config by which CI barfs when clippy lints are not passing?

The cause is that warnings won't trigger a CI failure - errors will (hence the `-Dwarnings`)

crates/assembly/src/tests.rs
plafer: Removed, and fixed all tests

crates/assembly-syntax/src/sema/mod.rs
plafer: Done

plafer:
Merging now to avoid other PRs creating new warnings. I'll address any remaining issue separately in a follow-up PR.

PR #3170, opened 2026-05-19
feat(deferred): add content-addressed deferred-DAG framework (1/2)

adr1anh:
This is the first framework PR for the deferred-computation / Precompile VM work.

It introduces the substrate we need before migrating production precompiles: a content-addressed deferred DAG, processor-owned deferred state, MASM-facing system events, a canonical wire format, and reference precompiles for tests.

Follow-up work, starting from the `precompiles/crate` stack, will wire this into the proof model, migrate the current precompiles, and eventually remove the old request-list precompile framework.

## Why this exists

Today, precompile work is represented as a flat list of requests. Each request is committed independently and folded into a transcript.

That model is too coarse for the Precompile VM we are moving toward. The work we want to prove is naturally graph-shaped: committed values, larger data payloads, operations, predicates, and shared intermediate results. A flat transcript loses that structure and forces unnecessary recomputation and hashing.

This PR introduces the deferred DAG as the shared language for that future system. It lets the VM, host, and future Precompile VM talk about deferred work as a graph of content-addressed nodes, while keeping the production precompile migration out of this PR.

### Deferred state

Adds `miden_core::deferred`, the core framework for deferred computation.

At a high level, the framework provides:

- content-addressed deferred nodes;
- structured tags for domain separation;
- committed data payloads split into hash-rate-sized chunks;
- value nodes for small committed values such as digests, field elements, or 256-bit values;
- join nodes for operations, predicates, and graph edges;
- framework-owned `TRUE` and statement/root machinery;
- precompile-owned semantics through `Precompile` and `PrecompileRegistry`;
- eager registration and evaluation;
- evaluation memos;
- deferred-state budgeting through `ExecutionOptions`;
- canonical wire serialization and rehydration.

A key property is that node identity is domain-separated: the digest commits to both the node’s tag and its payload. This lets different precompiles safely define their own node families without colliding with each other or with framework-owned nodes.

The processor now owns a `DeferredState` during execution and returns it in `ExecutionOutput.deferred_state`.

That gives tests and experimental tooling a concrete place to inspect what a MASM program registered and evaluated.

The processor surface is intentionally small in this branch: execution owns and returns a deferred state, execution options bound its size, and testing/experimental builds can install a deferred precompile registry. Production registry installation is left for the follow-up integration work.

### MASM system events

Adds three MASM-facing deferred system events:

- `adv.register_deferred`
- `adv.register_deferred_data`
- `adv.evaluate_deferred`

These are the minimal execution interface needed to test the framework from MASM.

They let a MASM program:

- register a deferred node from stack data;
- register a memory-backed deferred node;
- ask the processor’s deferred registry to evaluate a registered node;
- consume the canonical result from the advice stack.

The important boundary here is that MASM can now create deferred computations and interact with the processor’s deferred state, without yet making those computations part of the proof.

The new E2E tests focus on this boundary. They check that MASM programs can drive the new system events and that the expected nodes and evaluations appear in `ExecutionOutput.deferred_state`.

### Wire format

Adds `DeferredStateWire`, the canonical transport format for deferred state.

The wire format serializes the root-reachable closure of the deferred DAG in deterministic order. It is designed for the future proof/witness path, where a verifier will rehydrate deferred state from untrusted data and check it against a committed root.

This PR adds the wire format and validation rules now, but does not yet make deferred roots proof-bound.

Today, MASM-created deferred nodes are registered and evaluated, but they are not automatically logged into the deferred root. Because the wire format serializes root-reachable nodes, the processor E2E tests inspect the live `DeferredState` directly rather than treating the wire/root as a proof artifact.

This PR also adds fuzz coverage for deferred wire deserialization.

### Reference precompiles

Adds reference/test precompiles for exercising the framework:

- `Uint`
- `Group`
- `Hash`
- `Sig`

These are fixtures for testing the shape of the framework. They are not the final production precompile library.

They are useful because they exercise the kinds of relationships the Precompile VM will need: values, larger committed data, operations over child nodes, predicates that reduce to `TRUE`, and helper nodes produced during evaluation.

### Precompile transcript domain cleanup

This PR also updates the existing `log_precompile` request-list transcript fold from the old zero-capacity merge to a fixed domain-separated Poseidon2 fold using `PRECOMPILE_TRANSCRIPT_DOMAIN = [1, 0, 0, 0]`.

This changes the existing precompile transcript path and bumps the corresponding AIR relation digest. The change is intentional and anticipates the upcoming migration, but it is still the existing request-list precompile path. It is not deferred-root proof wiring yet.

## What this PR does not do

This PR does **not** make deferred computation part of `ExecutionProof`.

It also does not:

- migrate production precompiles;
- remove the legacy request-list witness;
- expose the final Precompile VM proof interface;
- make MASM-created deferred nodes proof-bound;
- replace the existing precompile APIs with the final user-facing deferred precompile surface.

Those pieces belong to follow-up work, starting with the `precompiles/crate` stack and the proof-wire integration.

## How to review this

The main question for this PR is whether the substrate is right.

In particular:

- does the deferred DAG model give the Precompile VM the structure it needs?
- are tags and node digests domain-separated in the right way?
- can MASM register and evaluate the nodes needed for future precompile wrappers?
- is the processor/deferred-state interface small enough for this staging PR?
- does `DeferredState` capture the information needed for the proof-wiring step?
- is the wire format strict enough for future untrusted proof/witness input?
- are the reference precompiles useful examples for the production migration?

The proof-model cutover is intentionally left for the next stack of PRs.


Al-Kindi-0:
Thanks a lot for closing the draft PRs and opening this new cleaned-up version. This made the architecture make more sense to me, and I can see the value of having a generic schema layer if the goal is to support many deferred-computation domains behind one common graph/evaluation interface, assuming I am not missing anything major.
Having an additive PR first and a separate deletion PR is much easier to reason about.

My remaining concern is about the dependency that might be inferred between two goals:

1. Removing the legacy precompile framework.
2. Introducing a generic schema-based deferred computation substrate.

I agree with the first goal. I’m not yet convinced the second needs to be a prerequisite for it. In other words, even with the cleaner PR split, this still seems to make the deletion phase depend on first accepting the generic `Schema` / `PrecompileSchema` framework. I don’t think that dependency is fundamental given the current state of things and our immediate goals.

My concern is mainly about timing and scope. I’m not convinced this level of generality is worth taking on at this stage. The code surface, and therefore the review surface, becomes large enough that it is hard for reviewers to keep the core idea in focus. That makes it harder for others, myself included, to understand the invariants, give meaningful feedback, and build confidence in the design.

My current view is that the `Schema` / `Precompile` abstraction is probably more appropriate once we have several downstream consumers or many precompile families that genuinely need a shared plugin-style interface. Right now, the main consumer is still us, and the first few useful cases are concrete and known. In that setting, I think a smaller closed set of typed operations would be easier to reason about and easier to land incrementally.

The broader thing I’ve been thinking about is that small, concrete, reviewable steps may get us to the right design faster over a reasonable time horizon than a large feature-complete solution upfront. A smaller slice gives more people a chance to engage with the actual relevant invariants during review, and lets the abstractions grow from pressure over time as real use cases accumulate, rather than being introduced all at once.

Of course, we are under pressure to hit the mainnet milestone, and I don’t want to ignore that. But I think that pressure has to be balanced against reviewability, shared understanding, and the risk of locking in a broad abstraction before we have enough concrete cases proving that we need it.
One practical benefit of smaller incremental steps is that they make it easier for more people to engage meaningfully. That matters not only for design quality, but also for throughput: once more people understand the invariants and have skin in the game, it becomes easier to split off subtasks, parallelize implementation, and sustain momentum.

Concretely, I’d prefer starting with the smallest deferred-precompile path needed for the immediate migration, and only introducing a `Schema`-like layer once we see enough repeated structure across domains to justify it.
  
Curious what others think, since there are probably multiple views on this. @bobbinth @plafer @Nashtare

adr1anh:
_this will be crafted into an issue later_

This RFC is relevant to one deferred-DAG cleanup we should probably do in `miden-vm`: make deferred `Tag` opaque and consider giving transcript `AND` its own domain.

## Context

The deferred-DAG code currently exposes `Tag` internals:

```rust
pub struct Tag {
    pub id: Felt,
    pub args: [Felt; 3],
}
```

and reuses the all-zero `Tag::TRUE` for transcript `AND` nodes:

```rust
pub const TRUE: Tag = Tag { id: ZERO, args: [ZERO; 3] };

pub fn and(lhs: Digest, rhs: Digest) -> Self {
    Self::join(Tag::TRUE, lhs, rhs)
}
```

So `AND(prev, stmt)` is currently hashed with all-zero Poseidon2 capacity and is equivalent to `Poseidon2::merge(prev, stmt)`.

Under this RFC, though, protocol-visible Poseidon2 uses should have typed, registered, nonzero capacity tags, and all-zero capacity is reserved for Merkle inner-node compression. A deferred transcript `AND` node is not a Merkle inner node, so it probably should not keep using all-zero capacity long-term.

## Proposal

Do this in two steps.

### 1. Make `Tag` opaque

Hide the fields while preserving the canonical 4-felt layout:

```rust
pub struct Tag {
    word: [Felt; 4],
}
```

Expose only intentional APIs:

```rust
impl Tag {
    pub const TRUE: Tag = Tag::from_word([ZERO; 4]);

    pub const fn new(id: Felt, args: [Felt; 3]) -> Self;
    pub const fn as_word(&self) -> [Felt; 4];
    pub const fn from_word(word: [Felt; 4]) -> Self;

    pub(crate) const fn id(&self) -> Felt;
    pub(crate) const fn args(&self) -> [Felt; 3];
}
```

This avoids external assumptions about the Rust representation while keeping the wire/hash format explicit.

### 2. Consider a distinct nonzero `AND` tag

Introduce a framework-owned `Tag::AND` and use it for transcript chaining:

```rust
pub fn and(lhs: Digest, rhs: Digest) -> Self {
    Self::join(Tag::AND, lhs, rhs)
}
```

Given this RFC, the preferred direction may be a registered VM capacity domain, e.g.:

```text
name     = DEFERRED_TRANSCRIPT_AND
capacity = [0, DEFERRED_TRANSCRIPT_AND_V1, 0, 0]
rate     = [prev_root, stmt_digest]
```

rather than deriving an ID by hashing a string like deferred precompile IDs do today.

## Impact

Making `Tag` opaque should be non-breaking at the protocol level.

Changing `AND` to nonzero capacity is protocol-breaking. It would affect:

- `Node::and`
- `DeferredState::rehydrate`
- reserved-tag checks in the deferred precompile registry
- `PrecompileTranscript::record`
- `log_precompile`
- AIR constraints for `LOGPRECOMPILE`
- MASM docs/tests
- relation digest / protocol constants

In particular, `AND(prev, stmt)` would no longer equal `Poseidon2::merge(prev, stmt)` unless we add a typed-capacity merge helper.

## Suggested acceptance criteria

### Opaque `Tag`

- [ ] `Tag` fields are private.
- [ ] `as_word()` / `from_word()` remain the explicit protocol boundary.
- [ ] No digest, wire-format, or transcript-root changes.

### Nonzero `AND`

- [ ] Decide whether `AND` should use an RFC-registered VM selector.
- [ ] Add `Tag::AND` / `DEFERRED_TRANSCRIPT_AND_V1`.
- [ ] `Node::and` uses `Tag::AND`, not `Tag::TRUE`.
- [ ] `rehydrate` validates chain nodes against `Tag::AND`.
- [ ] `log_precompile`, AIR, and `PrecompileTranscript::record` use the same capacity.
- [ ] Tests cover that `AND(TRUE, TRUE)` no longer aliases `Node::TRUE`.

huitseeker (commented):
I think this lgtm

huitseeker (approved):

adr1anh:
Heads up: I force-pushed a fairly large reshaping of this PR.

The important change is that this PR is no longer purely additive. The earlier version left the legacy `LOGPRECOMPILE` fold as the old zero-capacity `Poseidon2::merge(prev, stmt)` path and planned to keep AIR / legacy precompile transcript behavior unchanged.

The current version introduces the framework-owned `Tag::AND = [1, 0, 0, 0]` and moves the transcript/conjunction fold onto that domain. That affects `LOGPRECOMPILE`, `PrecompileTranscript::record`, `sys::log_precompile_request`, `sys::log_node_digest`, the AIR chiplet-request constraints, and the recursive verifier relation digest constants.

What did not change: this is still 1/2. Production precompile migration and the proof-model cutover remain in #3172.

processor/src/fast/mod.rs
```diff
@@ -226,6 +231,11 @@ impl FastProcessor {
     /// [`Self::with_advice`] or use [`Self::new_with_options`].
     pub fn with_options(mut self, options: ExecutionOptions) -> Result<Self, AdviceError> {
         self.advice.set_options(&options)?;
+        self.deferred_state = DeferredState::new(
```
huitseeker, line 245: Could `with_options()` preserve the existing deferred registry here? Right now it rebuilds `DeferredState` with `PrecompileRegistry::new()`, so this order loses installed precompiles:

```rust
FastProcessor::new(StackInputs::default())
    .with_deferred_precompiles(uint_precompiles())?
    .with_options(ExecutionOptions::default())?;
```

I tested that by registering a `Uint` deferred node after the options update; execution fails with `DeferredError { err: InvalidNode }`. That makes the builder order matter in a way the new `with_deferred_precompiles()` API does not signal.

huitseeker (changes requested):

core/src/deferred/wire.rs
```diff
@@ -0,0 +1,498 @@
...
+                return Err(IntegrityError::InvalidStructure);
+            }
+        }
+
+        state.root = root;
+
+        // `to_wire` emits the deterministic root-reachable closure. Equality makes the accepted
+        // format strict: root-last, canonical DFS order, and no dangling or duplicate
+        // entries.
+        if state.to_wire()? != *self {
```
huitseeker, line 104: I'd avoid recursion here.

A valid left-deep `AND` chain can pass the decoder and budget checks, then `to_wire()` recurses once per node through `visit_state_digest()`. With the default deferred-state budget, that can still be tens of thousands of frames, so adversarial proof data can stack-overflow the verifier instead of returning an `IntegrityError`. An iterative canonical-order check would keep this as a normal reject path.

adr1anh:
Closing in favor of #3222. The deferred framework work from this PR is now integrated into #3222, which targets next and should be reviewed as the combined deferred-precompile migration stack.

PR #3171, opened 2026-05-19
perf: full row-major trace generation

Nashtare:
## Describe your changes

Completes the VM execution trace transition to row-major with no transposition after #2937 (the range checker is folded into the core buffer's two trailing columns in place) and moves trace matrices instead of copying.

Also includes a couple of the follow-up cleanups from #3115.

Performances on `blake3_1to1` example:

  - `to_core_chiplets_matrices`: ~25 ms → ~0.2 µs
  - Chiplets trace path: ~60 ms → ~32 ms
  - Overall trace generation: ~125 ms → ~80 ms (~36% faster)

- Updated `CHANGELOG.md'


CHANGELOG.md
```diff
@@ -117,7 +119,6 @@
 - Made all internal `core::math` procedures natively little-endian ([#3084](https://github.com/0xMiden/miden-vm/pull/3084)).
 - [BREAKING] Updated the Miden crypto stack to `miden-crypto` v0.25, and switched SMT leaf hashing to use Poseidon2 domain separation so masm-side leaf digests match `SmtLeaf::hash()` ([#3095](https://github.com/0xMiden/miden-vm/pull/3095)).
 - [BREAKING] Reject post-last operation-indexed decorators in block assembly and serialized MAST forests; use `after_exit` for decorators that run after a block exits ([#3114](https://github.com/0xMiden/miden-vm/pull/3114)).
-- [BREAKING] Split the execution AIR into Core + Chiplets AIRs ([#3115](https://github.com/0xMiden/miden-vm/pull/3115)).
```
Nashtare, line 150: Got misplaced in one merge commit I believe.

air/src/trace/main_trace.rs
```diff
@@ -118,73 +107,80 @@ pub struct MainTrace {
...
-    /// Builds from `build_trace` outputs: core and chiplets row-major, range checker column
-    /// vectors.
+    /// Builds from the two row-major `build_trace` buffers: the Core matrix (system +
+    /// decoder + stack plus the two trailing range-checker columns) and the Chiplets matrix.
     ///
-    /// `core_rows` and `chiplets_rows` are the per-AIR padded heights returned by
-    /// `to_core_chiplets_matrices`. Both must be powers of two. The data arrays are sized to
-    /// `max(core_rows, chiplets_rows)`; rows beyond each per-AIR height are expected to be
-    /// zero-padded.
+    /// Heights are derived from the buffer lengths and must each be a power of two.
```
adr1anh: Not strictly necessary

air/src/trace/main_trace.rs
```diff
@@ -118,73 +107,80 @@ pub struct MainTrace {
...
-            if nc < RANGE_CHECK_TRACE_WIDTH {
-                range_checker_cols[nc][r]
+        let TraceStorage { core_rm, chiplets_rm, .. } = &self.storage;
+        if col < CORE_STORAGE_WIDTH {
+            let core_h = core_rm.height();
+            if r < core_h {
+                core_rm.get(r, col).expect("Accessed element is in bounds")
+            } else if col == CLK_COL_IDX {
+                Felt::from_u32(r as u32)
             } else {
```
adr1anh: Is this branch necessary?

air/src/trace/main_trace.rs
```diff
@@ -217,12 +208,9 @@ impl MainTrace {
...
-                dst[..CORE_WIDTH]
-                    .copy_from_slice(&core_rm.values[row * CORE_WIDTH..(row + 1) * CORE_WIDTH]);
-                dst[CORE_WIDTH] = range_checker_cols[0][row];
-                dst[CORE_WIDTH + 1] = range_checker_cols[1][row];
-                dst[CORE_WIDTH + 2..CORE_WIDTH + 2 + cw]
-                    .copy_from_slice(&chiplets_rm.values[row * cw..(row + 1) * cw]);
+                for (col, cell) in dst.iter_mut().enumerate() {
+                    *cell = self.get(RowIndex::from(row as u32), col);
+                }
             }
```
adr1anh: Nit: get the row as slice and copy the slice

air/src/trace/main_trace.rs
```diff
@@ -324,36 +256,16 @@ impl MainTrace {
...
-        row[..CORE_WIDTH]
-            .copy_from_slice(&core_rm.values[row_idx * CORE_WIDTH..(row_idx + 1) * CORE_WIDTH]);
-        row[CORE_WIDTH] = range_checker_cols[0][row_idx];
-        row[CORE_WIDTH + 1] = range_checker_cols[1][row_idx];
-        row[CORE_WIDTH + 2..CORE_WIDTH + 2 + CHIPLETS_WIDTH].copy_from_slice(
-            &chiplets_rm.values[row_idx * CHIPLETS_WIDTH..(row_idx + 1) * CHIPLETS_WIDTH],
-        );
+        for (col, cell) in row[..w].iter_mut().enumerate() {
+            *cell = self.get(RowIndex::from(row_idx as u32), col);
+        }
```
adr1anh: Same as above

air/src/trace/main_trace.rs
```diff
@@ -324,36 +256,16 @@ impl MainTrace {
...
-            if nc < RANGE_CHECK_TRACE_WIDTH {
-                range_checker_cols[nc].clone()
-            } else {
-                let cc = nc - RANGE_CHECK_TRACE_WIDTH;
-                (0..h).map(|r| chiplets_rm.values[r * CHIPLETS_WIDTH + cc]).collect()
-            }
-        }
+        (0..self.num_rows())
+            .map(|r| self.get(RowIndex::from(r as u32), col_idx))
+            .collect()
```
adr1anh: This might be worth parallelizing. As follow-up though, do we still need to get the columns?

processor/src/trace/chiplets/bitwise/mod.rs
adr1anh, line 1: The bitwise chiplet also needs to be converted to row major order

processor/src/trace/chiplets/hasher/tests.rs
```diff
@@ -536,10 +537,12 @@ fn hash_memoization_basic_blocks_check() {
 /// Builds the full hasher trace (controller + perm segment).
 fn build_trace(hasher: Hasher) -> Vec<Vec<Felt>> {
```
adr1anh, line 544: Same as above for bitwise

adr1anh (changes requested):
Not yet a full review, but I don't think this really resolves #3115. For example, the chiplets are still in column major order. This is probably going to be a much larger refactor, but the resulting resolution would remove most of the index constants, and instead rely on borrowing slices as the relevant column structs, à-la Plonky3.

processor/src/trace/utils.rs
```diff
@@ -72,37 +105,17 @@ impl<'a> TraceFragment<'a> {
     /// Updates a single cell in this fragment with provided value.
     #[inline(always)]
     pub fn set(&mut self, row_idx: RowIndex, col_idx: usize, value: Felt) {
-        self.data[col_idx][row_idx] = value;
+        let row = row_idx.as_usize();
+        self.band[row * self.stride + self.col_start + col_idx] = value;
```
huitseeker: Should we keep a real bounds check here? The old column-backed fragment would panic if `col_idx` was outside this fragment. With the row-major offset, a bad local column can now write into the next selector or chiplet column if `col_start + col_idx` still lands inside the row. Same concern for `fill_column`; an `assert!(col_idx < self.num_cols)` would keep this fail-fast in release builds.

air/src/trace/main_trace.rs
```diff
@@ -118,73 +107,80 @@ pub struct MainTrace {
...
-                    .get(r, nc - RANGE_CHECK_TRACE_WIDTH)
-                    .expect("Accessed element is in bounds")
+                core_rm.get(core_h - 1, col).expect("Accessed element is in bounds")
             }
+        } else if r < chiplets_rm.height() {
+            chiplets_rm
+                .get(r, col - CORE_STORAGE_WIDTH)
+                .expect("Accessed element is in bounds")
+        } else {
+            ZERO
```
huitseeker: Do we still need to either synthesize chiplet padding rows here or add a test that says this unified view is only a projection? Before this change, the chiplet side was padded to the unified height, including selectors and `chip_clk`. Now rows after `chiplets_rm.height()` are all zero, so `to_row_major_matrix()` can show a tail that does not satisfy the chiplet padding or `chip_clk` rules when Core is taller than Chiplets.

huitseeker (commented):
Holy moly Batman! 🚀 perf!

Nashtare:
> Not yet a full review, but I don't think this really resolves #3115. For example, the chiplets are still in column major order. This is probably going to be a much larger refactor, but the resulting resolution would remove most of the index constants, and instead rely on borrowing slices as the relevant column structs, à-la Plonky3.

Ah yeah I didn't include hasher / ACE. Let me put it back to draft mode while I iterate on it then will reopen with the full changes.

processor/src/trace/chiplets/ace/trace.rs
adr1anh, line 1: I think we can go further in the ACE and change the way we build the circuit evaluation to be natively row-major.

processor/src/trace/chiplets/bitwise/mod.rs
```diff
@@ -50,20 +50,18 @@ const INIT_TRACE_CAPACITY: usize = 128;
 ///   contains the full result of the bitwise operation.
 #[derive(Debug)]
 pub struct Bitwise {
-    trace: [Vec<Felt>; TRACE_WIDTH],
+    /// Row-major trace buffer: `TRACE_WIDTH` contiguous cells per row.
+    trace: Vec<Felt>,
```
adr1anh: Why not use RowMajorMatrix?

Nashtare:
@adr1anh should be good for review now. I also added insta snapshots.

air/src/constraints/chiplets/columns.rs
```diff
@@ -45,6 +45,7 @@ pub fn borrow_chiplet<T, S>(slice: &[T]) -> &S {
 /// | w0, w1, w2   | h0..h3              | h4..h7   | h8..h11     | m  --  --  --   |
 /// ```
 #[repr(C)]
+#[derive(Debug)]
```
adr1anh: We could also derive Clone. Copy might be too much since the struct is large.

air/src/trace/main_trace.rs
```diff
@@ -118,73 +107,74 @@ pub struct MainTrace {
...
+const _: () = assert!(super::CHIPLETS_OFFSET == CORE_STORAGE_WIDTH);
 
 impl MainTrace {
-    /// Builds from `build_trace` outputs: core and chiplets row-major, range checker column
-    /// vectors.
-    ///
-    /// `core_rows` and `chiplets_rows` are the per-AIR padded heights returned by
-    /// `to_core_chiplets_matrices`. Both must be powers of two. The data arrays are sized to
-    /// `max(core_rows, chiplets_rows)`; rows beyond each per-AIR height are expected to be
-    /// zero-padded.
```
adr1anh, line 135: This is the function which takes pre-padded traces right? Or is it possible to store non-power-of-two matrices, and then return the proper one? Basically, the invariants - and even the purpose - of MainTrace are a bit unclear.

From my understanding, we want the following
- The fast processor generates an execution trace, which is a structured log of all the operations the VM needs to prove
- From this, we generate a witness trace, the RowMajorMatrix (for core and chiplets) passed to the prover. Passed this point, the processor doesn't need to inspect the row major matrices so it shouldn't need to read anything from the witness trace, since it can get the information from its execution trace. I would make it the responsibility of the processor to correctly pad the trace to the next power of two. 

We should figure out if all of these accessors are really necessary on this struct. I think a lot of them could be implemented on MainCols instead. These could be accessed by having a method on the main trace which returns `&MainCols` for a given row.

The main purpose of this struct seems to mainly serve as a way of constructing a combined trace which pads the core or chiplets trace to the size of the other. 

This is obviously a larger refactor which we can address as a new sub-issue to #1763.

processor/src/trace/chiplets/ace/trace.rs
adr1anh, line 56: Refactor CircuitEvaluation to row-major

The current struct was designed around traces being generated in column major order. The more natural approach is probably to have fields `read: Vec<ReadRow>` and `write: Vec<WriteRow>` containing the wire values for each row, along with a map from each wire to it's multiplicities. From this, we could output a stream of AceRow.

processor/src/trace/utils.rs
```diff
@@ -12,44 +12,77 @@ use crate::{operation::Operation, utils::ToElements};
...
 
 // TRACE FRAGMENT
 // ================================================================================================
 
-/// TODO: add docs
+/// A writable, row-major view over one chiplet's region of the chiplets trace.
+///
+/// A chiplet occupies a contiguous band of rows and a contiguous band of columns
+/// `[col_start, col_start + num_cols)`, so chiplet-local column `c` maps to global column
+/// `col_start + c`. Writes land directly in that band of the shared row-major buffer.
```
adr1anh: Is this specific to the Chiplet trace? If so, maybe we could make the name of this struct match

adr1anh (commented):
These are some pending comments I had written before your new followup. I think the main point we should clarify is the role of main trace.

air/src/trace/main_trace.rs
```diff
@@ -118,73 +107,74 @@ pub struct MainTrace {
...
-// TODO: Could be tailored more efficiently?
-#[cfg(feature = "concurrent")]
-const ROW_MAJOR_CHUNK_SIZE: usize = 512;
+/// Physical row width of `core_rm`: the full per-AIR Core matrix width (`NUM_CORE_COLS`),
+/// i.e. system+decoder+stack columns plus the two trailing range-checker columns.
+const CORE_STORAGE_WIDTH: usize = crate::constraints::columns::NUM_CORE_COLS;
+const _: () = assert!(CORE_STORAGE_WIDTH == RANGE_CHECK_TRACE_OFFSET + RANGE_CHECK_TRACE_WIDTH);
+// The unified row layout is `[core | range | chiplets]`, so the chiplets section starts
+// exactly at the end of the core matrix.
+const _: () = assert!(super::CHIPLETS_OFFSET == CORE_STORAGE_WIDTH);
```
huitseeker: CI is failing here with `-D warnings`: this can be `CHIPLETS_OFFSET` directly, and the two `TraceStorage { core_rm, chiplets_rm, .. }` bindings below need the trailing `..` removed. After that `make clippy` should clear the current failure.

processor/src/trace/range/mod.rs
```diff
@@ -87,53 +74,55 @@ impl RangeChecker {
...
+    /// padding rows (`[0, num_padding_rows)`, all zero) are left untouched.
+    pub fn write_range_into_core(
+        self,
+        core_data: &mut [Felt],
+        stride: usize,
+        m_off: usize,
+        v_off: usize,
+        range_table_len: usize,
+        core_height: usize,
+        main_trace_len: usize,
```
huitseeker: Since the production caller now passes `core_height` for both height arguments, can `main_trace_len` and the extension loop below go away? 

Keeping this writer scoped to the Core height would make the API easier to read; unified projection padding can stay in `MainTrace`.

processor/src/trace/chiplets/ace/trace.rs
Nashtare: I've included it in #3177 addressing the comment there

Nashtare: Addressed in #3177:

```rust
/// One row of the ACE chiplet trace in `EVAL` mode: a single arithmetic gate `(id_0, v_0)` with
/// two inputs `(id_1, v_1)` (left) and `(id_2, v_2)` (right), the instruction pointer that
/// produced it, and the gate's `eval_op` selector.
#[derive(Debug, Clone, Copy)]
struct EvalNode {
    ptr: Felt,
    eval_op: Felt,
    id_0: Felt,
    v_0: QuadFelt,
    id_1: Felt,
    v_1: QuadFelt,
    id_2: Felt,
    v_2: QuadFelt,
}

/// Contains the variable and evaluation nodes resulting from the evaluation of a circuit.
/// The output value is checked to be equal to 0.
///
/// The set of nodes is used to fill the ACE chiplet trace.
#[derive(Debug, Clone)]
pub struct CircuitEvaluation {
    ctx: ContextId,
    clk: RowIndex,
    wire_bus: WireBus,
    read_nodes: Vec<ReadNode>,
    eval_nodes: Vec<EvalNode>,
}
```

air/src/trace/main_trace.rs
Nashtare: Part of it has already been addressed in #3177. What remains would be:
- deletion of migrated per-cell accessors (does not make much sense to keep such thin wrappers IMO)
- lift some higher-level boolean queries (`is_left_shift`, etc) onto `CoreCols` / `ChipletCols` themselves
- resolve the padding-ownership question

But I'd agree with you that this would deserve its own issue.

Al-Kindi-0:
@Nashtare , should I wait for #3177 to be merged before reviewing this one ?

Nashtare:
I initially split both for ease of review (#3177 being mostly mechanic refactor), but I could combine them if that's preferred. This would also reduce misdirections like Adrian's comment above that actually got already addressed in the follow-up PR.

Al-Kindi-0:
Whatever you think is best, thanks!

air/src/constraints/snapshots/miden_air__constraints__columns__tests__ace_col_map_layout.snap
adr1anh, line 1: these are great!

processor/src/trace/chiplets/ace/trace.rs
```diff
@@ -82,154 +88,132 @@ impl CircuitEvaluation {
...
-        columns[V_0_1_IDX][offset..offset + num_rows].copy_from_slice(&self.col_wire_out.v_1);
-
-        // Fill wire 1 columns for all rows
-        columns[ID_1_IDX][offset..offset + num_rows].copy_from_slice(&self.col_wire_left.id);
-        columns[V_1_0_IDX][offset..offset + num_rows].copy_from_slice(&self.col_wire_left.v_0);
-        columns[V_1_1_IDX][offset..offset + num_rows].copy_from_slice(&self.col_wire_left.v_1);
+    /// Writes this circuit evaluation's rows into the row-major buffer `out`
+    /// (`ACE_CHIPLET_NUM_COLS` contiguous cells per row), starting at row `offset`. `out`
+    /// is assumed zero-initialized, so columns that are zero on a row are left untouched.
+    pub fn fill(&self, offset: usize, out: &mut [Felt]) {
```
adr1anh, line 157: Nit: at the start we can do `let (out_rows, _) = out.as_chunks_mut(W)`

processor/src/trace/chiplets/bitwise/mod.rs
```diff
@@ -50,20 +52,20 @@ const INIT_TRACE_CAPACITY: usize = 128;
 ///   contains the full result of the bitwise operation.
 #[derive(Debug)]
 pub struct Bitwise {
-    trace: [Vec<Felt>; TRACE_WIDTH],
+    trace: RowMajorMatrix<Felt>,
```
adr1anh: Nit: We could store the data as `Vec<BitwiseOp>` with `BitwiseOp { op: Op, a: u32, b: u32 }`. Very minor, but it keeps the Bitwise struct small, and focuses the trace generation code to fill_trace

processor/src/trace/chiplets/hasher/mod.rs
adr1anh, line 1: The same comment as above on BitWise applies here, ie, not needing to hold the data in RowMajorMatrix for it to be copied in the chiplets' RowMajorMatrix later. But we should investigate whether it's worth it in terms of clarity and or speed/efficiency. Generating this portion is probably one of the heaviest per cell, so there may be some more we could do.

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -315,9 +316,10 @@ impl Memory {
             None => return,
         };
 
-        // iterate through addresses in ascending order, and write trace row for each memory access
-        // into the trace. we expect the trace to be 17 columns wide.
-        let mut row: RowIndex = 0.into();
+        let num_rows = self.trace_len();
+        let mut buffer = vec![ZERO; num_rows * MEMORY_TRACE_WIDTH];
```
adr1anh, line 320: Nit: as chunks

processor/src/trace/chiplets/mod.rs
```diff
@@ -237,115 +239,76 @@ impl Chiplets {
...
+        let set_col_ones = |trace: &mut [Felt], col: usize, rows: core::ops::Range<usize>| {
+            for r in rows {
+                trace[r * W + col] = ONE;
             }
+        };
+        set_col_ones(trace, 0, 0..hasher_ctrl_len); // s_ctrl: hasher controller rows
+        set_col_ones(trace, 1, memory_start..trace_len);
+        set_col_ones(trace, 2, ace_start..trace_len);
+        set_col_ones(trace, 3, kernel_rom_start..trace_len);
+        set_col_ones(trace, 4, padding_start..trace_len);
```
adr1anh: I think we'd get better memory access patterns if we set the selectors row by row, rather than by column, along with the clock. (and we could also parallelize).

adr1anh: Looking a bit lower though, the whole trace generation pipeline would benefit from fusing the selector setting with the row-filling.

processor/src/trace/chiplets/bitwise/mod.rs
```diff
@@ -83,98 +113,79 @@ impl Bitwise {
...
-            ];
-            cols.b_bits = [
-                Felt::new_unchecked(b & 1),
-                Felt::new_unchecked((b >> 1) & 1),
-                Felt::new_unchecked((b >> 2) & 1),
-                Felt::new_unchecked((b >> 3) & 1),
-            ];
-            cols.prev_output = Felt::new_unchecked(prev_output);
-            cols.output = Felt::new_unchecked(output);
+        let mut values = vec![ZERO; self.trace_len() * TRACE_WIDTH];
```
adr1anh: Nit: we can remove the allocation. In the loop, we can allocate an array of size `TRACE_WIDTH*OP_CYCLE_LEN` and then write it to the trace in each iteration

processor/src/trace/chiplets/hasher/trace.rs
```diff
@@ -198,82 +139,229 @@ impl HasherTrace {
...
     // --------------------------------------------------------------------------------------------
 
-    /// Fills the provided trace fragment with trace data from this hasher trace instance.
+    /// Fills the provided trace fragment by materializing the op log row by row.
     pub fn fill_trace(self, trace: &mut ChipletTraceFragment) {
         debug_assert_eq!(self.trace_len(), trace.len(), "inconsistent trace lengths");
         debug_assert_eq!(TRACE_WIDTH, trace.width(), "inconsistent trace widths");
 
-        trace.copy_rows_from(&self.trace.values);
+        let mut values = vec![ZERO; self.row_count * TRACE_WIDTH];
```
adr1anh: Same comment as above

air/src/trace/main_trace.rs
```diff
@@ -584,126 +395,149 @@ impl MainTrace {
...
 
     // CHIPLETS COLUMNS
     // --------------------------------------------------------------------------------------------
 
     /// Returns chiplet column number 0 at row i.
+    ///
+    /// # Panics
+    /// Panics if `i` is past the chiplets-AIR height. Callers iterating the unified trace
+    /// must guard via [`Self::chiplets_height`] or filter via [`Self::is_hash_row`] /
+    /// [`Self::is_bitwise_row`] / [`Self::is_memory_row`] / [`Self::is_ace_row`], which
```
huitseeker, line 410: Should the remaining chiplet predicates get the same `chiplets_height()` guard as these helpers? In the split-trace model `num_rows()` can be larger than `chiplets_height()`, and `is_kernel_row()`, `f_mv()`, and `f_mu()` still read chiplet columns first. A caller filtering unified rows with one of them would panic past the chiplets matrix instead of returning false.

air/src/trace/main_trace.rs
```diff
@@ -485,24 +298,19 @@ impl MainTrace {
...
+        bits[0]
+            + bits[1] * Felt::from_u64(2)
+            + bits[2] * Felt::from_u64(4)
+            + bits[3] * Felt::from_u64(8)
+            + bits[4] * Felt::from_u64(16)
+            + bits[5] * Felt::from_u64(32)
+            + bits[6] * Felt::from_u64(64)
     }
 
     /// Returns an iterator of [`RowIndex`] values over the row indices of this trace.
```
Al-Kindi-0: I would seriously consider removing this method

processor/src/trace/tests/decoder.rs
Al-Kindi-0, line 1: There are are a number of places where `row_iter` is used in this module, I would try to replace them with something more solid

processor/src/trace/parallel/tests.rs
Al-Kindi-0, line 1: I would replace the usage of `num_rows` method here as well

processor/src/trace/utils.rs
```diff
@@ -58,51 +112,55 @@ impl<'a> TraceFragment<'a> {
...
+            row[self.col_start..self.col_start + self.num_cols]
+                .copy_from_slice(&src[r * self.num_cols..(r + 1) * self.num_cols]);
+            if write_chip_clk {
+                row[clk_col] = Felt::from_u32((self.row_offset + dst_row + 1) as u32);
+            }
         }
-
-        let num_rows = data[0].len();
-        Self { data, num_rows }
     }
```
Al-Kindi-0, line 164: `TraceLenSummary` and its related code needs a bit of a re-work given the migration away from the monolithic setting

Al-Kindi-0 (commented):
Looks good, thanks a lot!

Minor inline comments, but I think we should avoid APIs that expose a unified row model. `max(core_height, chiplets_height)` is useful for limits/logging, but row iteration and row access should be table-local. So I would prefer removing or gating `row_iter`, replacing remaining call sites with explicit Core/Chiplets heights, and renaming/reworking the trace length summary to avoid monolithic “main trace” terminology.

processor/src/trace/utils.rs
Nashtare: I've reworked part of it (while also addressing the comments above), but deeper `TraceLenSummary` restructuring (per-AIR shape vs. one max-trace-len) would rather belong with the `MainTrace` cleanup tracked under #1763 I believe?

huitseeker (approved):
This is awesome, thanks a lot!

Al-Kindi-0:
@huitseeker @adr1anh should we just merge this one and get it out of the way asap ?

adr1anh (approved):

PR #3172, opened 2026-05-19
feat(deferred): cut over precompiles to the deferred-DAG proof model (2/2)

adr1anh:
This is **2/2** of the deferred-computation rework, stacked on #3170.

#3170 adds the deferred-DAG framework and aligns the existing transcript fold with the framework’s `AND` domain. This PR performs the cutover: proofs now carry the deferred DAG witness, the production precompiles move onto the deferred framework, and the legacy precompile-request witness path is removed.

## Scope update since the earlier version

This branch changed shape after the first review pass.

The earlier version exposed a public raw `verify_stark(...)` entry point, used `precompile_schema()` terminology, and installed precompile semantics directly on `FastProcessor`.

That is no longer the design in this PR.

The current version uses:

- `PrecompileRegistry` terminology instead of “schema”;
- host/library-owned registry installation instead of `FastProcessor`-owned precompiles;
- `verify_with_precompiles(...)` as the public registry-aware verifier entry point;
- a private raw STARK verifier path;
- a hardened canonical deferred wire format.

Older review comments mentioning `precompile_schema()`, public `verify_stark(...)`, or `FastProcessor::with_precompiles(...)` refer to the previous shape of the branch.

### Proofs carry the deferred DAG witness

`ExecutionProof` now carries a `DeferredStateWire` instead of the old precompile request list.

Verification treats that wire as untrusted input. Before accepting the deferred commitment, the verifier rehydrates the wire under an installed `PrecompileRegistry`, validates the graph, replays the transcript, and reduces all reachable logged statements.

The resulting deferred root is then used as the STARK public input.

This replaces the old model where the proof carried a list of `PrecompileRequest`s and the verifier replayed legacy per-precompile request handlers.

### Verifier API split

The public verifier surface is now split by whether precompile semantics are needed:

- `miden_verifier::verify(...)` preserves the precompile-free API by using an empty registry.
- `miden_verifier::verify_with_precompiles(...)` verifies proofs whose deferred wire requires an installed `PrecompileRegistry`.
- `miden_vm::verify(...)` installs the production `CoreLibrary` registry.

Raw `verify_stark(...)` is private in this PR. A future Precompile VM / externally-proven deferred commitment path is intentionally not exposed here.

### Registry installation is host/library-owned

Production precompile semantics are exposed by `CoreLibrary::precompiles()` and installed through the normal host/library path.

`HostLibrary` carries precompile registries, and `DefaultHost::load_library()` merges them into the host. Execution, trace construction, proving, and verification then all operate against the same installed registry.

This avoids having a separate processor-specific precompile installation path.

### Production precompiles move onto the deferred framework

The production precompiles now implement the deferred `Precompile` trait under `crates/lib/core/src/precompiles/`:

- `keccak256`
- `sha512`
- `ecdsa_k256_keccak`
- `eddsa_ed25519`

Their MASM wrappers keep the public procedure names and stack contracts where possible, but internally use the deferred machinery:

- `sys::register_expr`
- `sys::register_chunk`
- raw `adv.evaluate_deferred`
- `sys::log_node_digest`

There is intentionally no safe generic `sys::evaluate` helper. `adv.evaluate_deferred` returns an unbound host hint, so each precompile wrapper must bind the result in-circuit and log a precompile-specific predicate that rehydration will re-check.

### The deferred wire is canonical

The deferred witness format is hardened into a canonical, sectioned wire representation.

Rehydration validates the graph, checks semantic reductions, reconstructs the accepted state, and rejects equivalent-but-non-canonical encodings. This gives the verifier one canonical deferred commitment for the supplied witness.

## Removed legacy path

This PR removes the old precompile-request framework, including:

- `core::precompile::PrecompileRequest`
- `PrecompileCommitment`
- `PrecompileVerifier`
- `PrecompileVerifierRegistry`
- `PrecompileTranscriptState`
- legacy per-precompile handlers under `crates/lib/core/src/handlers/`
- `sys::log_precompile_request`
- legacy precompile-request fuzz targets
- old request-list proof witness plumbing

Existing proofs that depend on the old precompile request list are not compatible with this branch.

## Out of scope

This PR does **not** expose a public raw STARK verification API for externally supplied deferred commitments.

It also does **not** make the future Precompile VM prove deferred DAGs directly. This PR makes the production proof model carry and verify the deferred DAG witness; proving that deferred work in a separate Precompile VM is follow-up work.

## Reviewer notes

The highest-value review areas are:

1. `core/src/deferred/*`  
   Canonical wire/state semantics, graph validation, reduction, and rehydration integrity.

2. `core/src/proof.rs`, `prover/src/lib.rs`, `verifier/src/lib.rs`  
   Proof witness shape, deferred-root derivation, and verifier API boundaries.

3. `processor/src/host/*`, `processor/src/fast/*`, and library-loading paths  
   Registry propagation from loaded libraries into execution, trace construction, proving, and verification.

4. `crates/lib/core/src/precompiles/*` and the MASM wrappers  
   Per-precompile tags, canonical reductions, host-hint binding, and compatibility with existing procedure contracts.

5. AIR / recursive verifier artifacts  
   Deferred-root public input threading and consistency with the #3170 transcript-domain update.

## Testing / coverage

This branch includes or updates coverage for:

- migrated production precompile wrappers;
- deferred wire round-trips;
- canonical / non-canonical deferred witness handling;
- tampered witness rejection;
- bounded deferred-state growth;
- end-to-end prove/verify with a non-empty deferred root;
- fuzz coverage for deferred wire deserialization.

Use GitHub checks as the source of truth for the current CI state.


miden-core-fuzz/Cargo.toml
adr1anh: Find a way to re-instate

miden-vm/src/cli/run.rs
```diff
@@ -153,7 +153,8 @@ fn run_masp_program(params: &RunCmd) -> Result<(ExecutionTrace, [u8; 32]), Repor
     .map_err(|err| Report::msg(format!("{err}")))?;
 
     let processor = FastProcessor::new_with_options(stack_inputs, advice_inputs, exec_options)
-        .map_err(|err| Report::msg(format!("{err}")))?;
+        .map_err(|err| Report::msg(format!("{err}")))?
+        .with_precompiles(std::sync::Arc::new(CoreLibrary::default().precompile_schema()));
```
huitseeker, line 157: This same schema install looks needed in the `prove` CLI path too. 

`miden-vm/src/cli/prove.rs` still calls `miden_prover::prove_sync`, which builds a `FastProcessor` with the empty default `PrecompileRegistry`. After this migration, programs using the core precompile wrappers need `CoreLibrary::precompile_schema()` installed before execution, so `miden run` works here but `miden prove` can fail on the same supported program.

`prove` should mirror this path by constructing the processor with `.with_precompiles(Arc::new(CoreLibrary::default().precompile_schema()))`, executing trace inputs, and then calling `prove_from_trace_sync`

crates/lib/core/asm/sys/mod.masm
```diff
@@ -125,7 +104,7 @@ pub proc register_chunk
     # => [NODE_DIGEST, ...]
 end
 
-#! Folds a node digest into the rolling precompile-transcript via the `log_precompile` opcode.
+#! Folds a node digest into the rolling deferred commitment via the `log_precompile` opcode.
```
huitseeker: This looks like it needs the generated core docs update in this PR too.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -39,158 +36,96 @@ const SIG_LEN_FELTS = 17   # 65.div_ceil(4)
...
+#!   locaddr[9..12]   zero pad (alignment to felt 12)
+#!   locaddr[12..20]  keccak256(message) digest (8 felts)
+#!   locaddr[20..37]  sig (65 bytes packed as 17 felts)
+#!   locaddr[37..40]  zero pad (alignment to felt 40)
+#!   locaddr[40..48]  msg bytes scratch (8 felts, transient — only used during keccak hash)
 #!
-#! The procedure traps if:
-#! - The public key does not hash to `PK_COMM` (invalid commitment)
-#! - The signature verification fails
+#! Local memory is implicitly zero-initialised at procedure entry, so the two pad regions stay
```
huitseeker, line 52: Can we explicitly clear `locaddr[9..12]` and `locaddr[37..40]` before `verify_prehash`? 

The registered 40-felt chunk includes those padding slots, and the Rust precompile rejects non-zero padding. 

If the locals contain stale values, valid signatures can fail depending on prior frame contents.

processor/src/fast/basic_block/deferred_handlers.rs
```diff
@@ -135,10 +143,24 @@ pub(super) fn handle_deferred_register_chunk(
...
+        .ok_or(MemoryError::AddressOutOfBounds { addr: u64::MAX })?;
     if end > u32::MAX as u64 {
         return Err(MemoryError::AddressOutOfBounds { addr: end }.into());
     }
+    let max_deferred_elements = processor.options.max_deferred_elements();
+    let chunk_elements = (n as usize)
+        .checked_mul(8)
+        .and_then(|payload_elements| payload_elements.checked_add(4))
+        .unwrap_or(usize::MAX);
+    let projected = processor
```
huitseeker, line 157: This precheck charges the full chunk size before knowing whether the chunk node is already interned. 

Under a tight `max_deferred_elements` limit, a repeated `register_chunk` can fail here ... even though `DeferredState::register` would deduplicate it and not grow the state.

core/src/deferred/wire.rs
```diff
@@ -593,6 +594,22 @@ impl Deserializable for WireNode {
         let rhs = source.read_u32()?;
         Ok(Self { tag, lhs, rhs })
     }
+
+    fn min_serialized_size() -> usize {
+        Tag::min_serialized_size() + 2 * core::mem::size_of::<u32>()
```
huitseeker, line 599: Clippy is failing here with `unused_qualifications` because `size_of` is already in scope.

processor/src/trace/parallel/snapshots/miden_processor__trace__parallel__tests__trace__parallel__tests__test_trace_generation_at_fragment_boundaries__case_01.snap
```diff
@@ -2,4 +2,4 @@
...
..., 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535]], core_rows: 64, chiplets_rows: 128 }, last_program_row: RowIndex(13) }, program_info: ProgramInfo { program_hash: Word([10033257784419832427, 14686140193024196894, 15436783799654791584, 11131623826151749048]), kernel: Kernel([]) }, stack_outputs: StackOutputs { elements: [0, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }, deferred_state: DeferredState { nodes: {}, root: Word([0, 0, 0, 0]), evals: {Word([0, 0, 0, 0]): Node { tag: Tag { id: 0, args: [0, 0, 0] }, payload: Expression([0, 0, 0, 0, 0, 0, 0, 0]) }}, num_elements: 0 }, trace_len_summary: TraceLenSummary { main_trace_len: 14, range_trace_len: 39, chiplets_trace_len: ChipletsLengths { hash_chiplet_len: 96, bitwise_chiplet_len: 0, memory_chiplet_len: 0, ace_chiplet_len: 0, kernel_rom_len: 0 }, padded_trace_len: Some(128) } }
```
huitseeker, line 5: CI still reports all 27 `test_trace_generation_at_fragment_boundaries` snapshot cases failing. These snapshot files were updated in this PR, but the generated output in the test job still differs from what is committed.

huitseeker (changes requested):
At a minimum, we should fix the snapshots.

adr1anh:
Heads up: I force-pushed a fairly large reshaping of this PR after the earlier review pass.

The old version exposed a public raw `verify_stark(...)`, used `precompile_schema()` terminology, and installed precompile semantics directly on `FastProcessor`.

The current version no longer does that. The public verifier API is now `verify(...)` for the empty-registry path and `verify_with_precompiles(...)` for proofs that need a `PrecompileRegistry`; raw `verify_stark(...)` is private. Production precompiles are exposed as `CoreLibrary::precompiles()` and installed through `HostLibrary` / `DefaultHost::load_library()`, so execution, trace building, proving, and verification all use the same host-owned registry.

This force-push also hardens the deferred witness into a canonical wire format and rejects non-canonical or invalid deferred witnesses during rehydration.

Comments mentioning `precompile_schema()`, public `verify_stark(...)`, or `FastProcessor::with_precompiles(...)` are about the previous shape of the branch.

PR #3174, opened 2026-05-19
Remove owned decorator storage from MAST nodes (3/4)

### LINKED ISSUE #2411: remove Owned variant of DecoratorStorage

huitseeker:
This closes #2411 by making MAST nodes keep decorator links instead of owned decorator data.

Before this change, nodes could carry their own decorator payloads. That made cloned, merged, and imported forests harder to reason about because decorator data could live in more than one place. The forest now owns decorators once, and nodes refer to them through LinkedDecoratorStore.

This keeps MastForest as the source of truth for decorators, removes standalone owned-node construction, and shrinks the node constructors around the linked-store model.

The public behavior is unchanged: decorators still attach to nodes and survive construction, merge, serialization, and execution. The internal model is simpler: node structure points to forest-owned decorator records instead of copying them.

Closes #2411.

### DIFF
--- a/core/src/mast/node/join_node.rs
+++ b/core/src/mast/node/join_node.rs
@@ -263,7 +263,8 @@ impl MastNodeExt for JoinNode {
         let id = self.decorator_store.linked_id();
         // Verify that this node is the one stored at the given ID in the forest
         let self_ptr = self as *const Self;
-        let forest_node = forest.get_node_by_id(id).expect("linked node id must be present in forest");
+        let forest_node =
+            forest.get_node_by_id(id).expect("linked node id must be present in forest");
         let forest_node_ptr = match forest_node {
             MastNode::Join(join_node) => join_node as *const JoinNode as *const (),
             _ => panic!("Node type mismatch at {id:?}"),
--- a/core/src/mast/node/join_node.rs
+++ b/core/src/mast/node/join_node.rs
@@ -36,11 +36,6 @@ pub struct JoinNode {
 }
 
 impl JoinNode {
-    pub(super) fn into_linked_decorator_store(mut self, node_id: MastNodeId) -> Self {
-        self.decorator_store = DecoratorStore::Linked { id: node_id };
-        self
-    }
-
     pub(crate) fn linked_decorator_store_id(&self) -> Option<MastNodeId> {
         self.decorator_store.linked_id()
     }
@@ -254,28 +249,12 @@ impl MastNodeExt for JoinNode {
     type Builder = JoinNodeBuilder;
 
     fn to_builder(self, forest: &MastForest) -> Self::Builder {
-        // Extract decorators from decorator_store if in Owned state
-        match self.decorator_store {
-            DecoratorStore::Owned { before_enter, after_exit, .. } => {
-                let mut builder = JoinNodeBuilder::new(self.children);
-                builder = builder
-                    .with_before_enter(before_enter)
-                    .with_after_exit(after_exit)
-                    .with_digest(self.digest);
-                builder
-            },
-            DecoratorStore::Linked { id } => {
-                // Extract decorators from forest storage when in Linked state
-                let before_enter = forest.before_enter_decorators(id).to_vec();
-                let after_exit = forest.after_exit_decorators(id).to_vec();
-                let mut builder = JoinNodeBuilder::new(self.children);
-                builder = builder
-                    .with_before_enter(before_enter)
-                    .with_after_exit(after_exit)
-                    .with_digest(self.digest);
-                builder
-            },
-        }
+        let (before_enter, after_exit) = self.decorator_store.into_node_level_decorators(forest);
+
+        JoinNodeBuilder::new(self.children)
+            .with_before_enter(before_enter)
+            .with_after_exit(after_exit)
+            .with_digest(self.digest)
     }
 
     #[cfg(debug_assertions)]
@@ -388,6 +367,30 @@ impl JoinNodeBuilder {
                 self.after_exit,
             ),
         })
+    }
+
+    pub(in crate::mast) fn build_linked_with_decorators(
+        self,
+        node_id: MastNodeId,
+    ) -> Result<(JoinNode, Vec<DecoratorId>, Vec<DecoratorId>), MastForestError> {
+        let Self {
+            children,
+            before_enter,
+            after_exit,
+            digest,
+        } = self;
+        let digest =
+            NodeBuilderLifecycle::new(&before_enter, &after_exit, digest).forced_digest()?;
+
+        Ok((
+            JoinNode {
+                children,
+                digest,
+                decorator_store: DecoratorStore::Linked { id: node_id },
+            },
+            before_enter,
+            after_exit,
+        ))
     }
 }
 
--- a/core/src/mast/node/join_node.rs
+++ b/core/src/mast/node/join_node.rs
@@ -13,7 +13,7 @@ use crate::mast::MastNode;
 use crate::{
     Felt, Word,
     mast::{
-        DecoratorId, DecoratorStore, ExecutableMastForest, MastForest, MastForestError,
+        DecoratorId, ExecutableMastForest, LinkedDecoratorStore, MastForest, MastForestError,
         MastNodeFingerprint, MastNodeId, digest,
     },
     operations::opcodes,
@@ -31,11 +31,11 @@ use crate::{
 pub struct JoinNode {
     children: [MastNodeId; 2],
     digest: Word,
-    decorator_store: DecoratorStore,
+    decorator_store: LinkedDecoratorStore,
 }
 
 impl JoinNode {
-    pub(crate) fn linked_decorator_store_id(&self) -> Option<MastNodeId> {
+    pub(crate) fn linked_decorator_store_id(&self) -> MastNodeId {
         self.decorator_store.linked_id()
     }
 }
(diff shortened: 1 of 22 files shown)


core/src/mast/node/join_node.rs
```diff
@@ -28,20 +28,14 @@ use crate::{
 /// first child first and the second child second.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
 pub struct JoinNode {
     children: [MastNodeId; 2],
     digest: Word,
-    decorator_store: DecoratorStore,
+    decorator_store: LinkedDecoratorStore,
 }
```
bobbinth, line 35: Question: why do we still need a "decorator store" associated with MAST nodes? Can we remove the functionality where a node is "aware" of its own decorators? If not, which code relies on these lookups? 

Ideally, all decorator lookups should go through `MastForest` (in the future `Package`) or `MastForestBuilder`.

huitseeker: This is not a store in any sense, if you look at the data structure, it's a `MastNodeId`. It helps manage the IOC needed to access the `DebugInfo` relative to a node — and pretty much all the code that accesses any sort of `DebugInfo` in any context does it using the `MastNode` as a "key".

Fixing that anti-pattern is the topic of #3040 .

bobbinth: Don't we access debug info via the `MastForest` itself? - i.e., via:

- `MastForest::decorators_for_op()`.
- `MastForest::before_enter_decorators()`.
- `MastForest::after_exit_decorators()`.

huitseeker: That is what I'm saying. This PR removes node-owned decorator data, but it does not remove every path where a node supplies the key used to query decorators.

As I said, the actual decorator storage is forest-owned. [`LinkedDecoratorStore`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/decorator_store.rs#L14) stores only a `MastNodeId`, and its read methods delegate back to the forest:
  [`before_enter`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/decorator_store.rs#L26) calls `forest.linked_before_enter_decorators(self.id)`, and [`after_exit`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/decorator_store.rs#L34) calls `forest.linked_after_exit_decorators(self.id)` The forest then reads from `DebugInfo`, e.g.
  [`MastForest::before_enter_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/mod.rs#L588), [`MastForest::after_exit_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/mod.rs#L594), and
  [`DebugInfo::decorators_for_operation`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/debuginfo/mod.rs#L239).

  So there are two different questions here:

  1. Nodes do not own decorators. The decorator table and node/op mappings are in forest debug info.
  2. Do some APIs still query decorators through a node? Yes. The node still carries the `MastNodeId` link used to select the right forest row.

For before/after decorators, execution asks the node while passing the forest: [`node.before_enter(current_forest)`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/mod.rs#L546) and
  [`node.after_exit(current_forest)`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/mod.rs#L574). Those calls go through `MastNodeExt`, then through `LinkedDecoratorStore`, then back to the forest via
  [`ExecutableMastForest::linked_before_enter_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/mod.rs#L994) and [`linked_after_exit_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/mod.rs#L999).

For basic-block op decorators, the processor is forest-first after it has a node id: [`basic_block.linked_id()`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/execution/basic_block.rs#L254) gives the row key, and [`current_forest.decorator_indices_for_op(node_id, op_idx_in_block)`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/operation.rs#L130) reads the forest-owned mapping.

The basic-block iterators have the same node-routed lookup shape. [`indexed_decorator_iter`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L195) gets `self.decorators.linked_id()` and then calls `forest.decorator_links_for_node(id)`.
  [`raw_decorator_iter`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L237) does the same, plus forest lookups for `before_enter` and `after_exit`.
[`raw_op_indexed_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L289) also gets the linked id and queries `forest.decorator_links_for_node(id)`. The private pretty-print iterator follows the same pattern through
  [`OperationOrDecoratorIterator::new_with_forest`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L1007) and
  [`next_decorator_if_due`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L1022).

So the accurate summary is: this PR removes owned decorator storage from nodes, but it does not fully remove node-routed decorator access. A stricter follow-up could hide `LinkedDecoratorStore`, make the node id explicit at these call sites, and route all decorator reads through `MastForest`/`ExecutableMastForest` APIs directly. Since these calls only exist in the debugging branch of execution, #3040 looks like the right place to do it.

bobbinth: > For before/after decorators, execution asks the node while passing the forest: [`node.before_enter(current_forest)`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/mod.rs#L546) and
> [`node.after_exit(current_forest)`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/mod.rs#L574).

These could be easily fixed - right? In fact, it seems like we are doing extra work to first get the node by ID (e.g., [here](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/processor/src/fast/mod.rs#L542-L544)) and then use the node to route decorator retrieval through it.

> The basic-block iterators have the same node-routed lookup shape. [`indexed_decorator_iter`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L195) gets `self.decorators.linked_id()` and then calls `forest.decorator_links_for_node(id)`.
> [`raw_decorator_iter`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L237) does the same, plus forest lookups for `before_enter` and `after_exit`.
> [`raw_op_indexed_decorators`](https://github.com/0xMiden/miden-vm/blob/908f8911d2ea3c96084d9fb87bb492ff920ad3b0/core/src/mast/node/basic_block_node/mod.rs#L289) also gets the linked id and queries `forest.decorator_links_for_node(id)`.

AFAICT, these are used only in tests and even there, we should be able to "fix" them pretty easily. Basically, wherever we have access to a node, we have access to its ID, and so, instead of doing something like `node.get_decorators(mast_forest)` should be able to do `mast_forest.get_decorators_for_node(node.id)` (function names are illustrative).

Doing this would be possible after in this PR (right after owned decorators are removed), and should remove a bunch of code that we no longer need.

huitseeker: `raw_decorator_iter` and `raw_op_indexed_decorators` appear removable or gateable from a public API perspective; `indexed_decorator_iter` is still production infrastructure; the pretty-print iterator is production-reachable.

> Basically, wherever we have access to a node, we have access to its ID, and so, instead of doing something like node.get_decorators(mast_forest) should be able to do mast_forest.get_decorators_for_node(node.id) (function names are illustrative).

Yes, [that's what I was saying](https://github.com/0xMiden/miden-vm/pull/3174#discussion_r3299569337):
 
> > A stricter follow-up could hide LinkedDecoratorStore, make the node id explicit at these call sites, and route all decorator reads through MastForest/ExecutableMastForest APIs directly

I agree that adapting or removing these is possible.

huitseeker:
Closing in favour of #3208.

PR #3175, opened 2026-05-20
fix(utils-core-derive): correct derive macro panic message

sashaphmn:
The `derive_mast_forest_contributor` macro returned a stale copy-paste panic message referencing `EnumThispatch`, which does not correspond to any derive macro in this crate.

This updates the diagnostic to correctly reference`MastForestContributor`, improving compiler output clarity when the derive is used on a non-enum type.


huitseeker (approved):
Thanks!

sashaphmn:
Hey @huitseeker could you please add the no changelog

sashaphmn:
I added a changelog entry, but since it was marked as no changelog, should I delete it?

PR #3176, opened 2026-05-20
In-VM `secp256k1` ECDSA verification with GLV + 4-way MSM

Al-Kindi-0:
Given a secp256k1 public key $Q$, a message digest $e$, and a signature $(r, s)$, ECDSA verification returns $1$ if and only if

$$
R = u_1 \cdot G + u_2 \cdot Q,
\quad
u_1 = e \cdot s^{-1} \bmod n,
\quad
u_2 = r \cdot s^{-1} \bmod n,
$$

and $R_x \bmod n = r$. Here $G$ is the curve generator, $n$ is the order of the scalar field, and the point arithmetic happens in the base field $\mathbb{F}_p$.

The verifier also parses calldata, range-checks $r$ and $s$, decompresses $Q$ when the public key is compressed, derives $u_1, u_2$, reduces $e \bmod n$, and validates the host hints used along the way. Those steps matter, but the expensive piece is the double scalar multiplication $u_1 \cdot G + u_2 \cdot Q$. The design therefore reduces the size of that MSM and makes the field arithmetic behind the point operations cheaper in the VM.

## Idea 1: GLV to halve the scalar size

secp256k1 has a special structure: there is a scalar $\lambda$ and a base-field constant $\beta$ such that for any point $P = (x_P, y_P)$,

$$
[\lambda] P = (\beta \cdot x_P \bmod p,\; y_P).
$$

Applying $\phi$ is cheap compared with a scalar multiplication: it is one base-field multiplication on the $x$ coordinate. Call this map $\phi(P) = (\beta \cdot x_P \bmod p,\; y_P)$.

The GLV decomposition uses this fact: every 256-bit scalar $k$ can be written as

$$
k \equiv k_a + k_b \cdot \lambda \pmod{n}
\qquad \text{with} \qquad |k_a|,\, |k_b| < 2^{128}.
$$

So $[k] P = [k_a] P + [k_b] \phi(P)$ — a 128-bit scalar mult on $P$ plus a 128-bit scalar mult on $\phi(P)$, instead of a single 256-bit scalar mult. **The scalars are halved in length** by replacing one scalar multiplication on $P$ with a two-term multi-scalar multiplication over $P$ and its endomorphism image $\phi(P)$.

The split halves $k_a, k_b$ can be negative. The VM stores each as a non-negative magnitude plus a one-bit sign, and absorbs the sign by conditionally negating the relevant base point (negation on secp256k1 is $-P = (x_P, -y_P)$, which is cheap). So when an entry of the form "$[k_b] \phi(P)$" appears below, the VM is actually computing $[|k_b|] (s_{k_b} \cdot \phi(P))$ where $s_{k_b} \in \lbrace +1, -1 \rbrace$ is the sign bit.

The host computes the $(k_a, k_b)$ split off-VM. In-VM, the verifier checks the defining equation $k_a + k_b \cdot \lambda \equiv k \pmod{n}$ and verifies that the two magnitudes fit in 128 bits.

We apply GLV to both scalars $u_1$ and $u_2$. The original double scalar mult becomes a **four-way 128-bit multi-scalar multiplication (MSM)**:

$$
R = k_a^{(1)} \cdot G + k_b^{(1)} \cdot \phi(G) + k_a^{(2)} \cdot Q + k_b^{(2)} \cdot \phi(Q).
$$

## Idea 2: Shamir's trick — fuse the four scalar mults

A textbook 4-way MSM does four scalar mults separately and sums the results. Shamir's trick fuses them into a single 128-iteration loop (the scalars are now 128 bits long) that processes **one bit of each scalar at every step in parallel**:

$$
\text{accum} \leftarrow [2] \text{accum} + T[\,b_i^{(1)}\,b_i^{(2)}\,b_i^{(3)}\,b_i^{(4)}\,]
$$

where $b_i^{(j)}$ is the $i$-th bit of the $j$-th scalar, and $T$ is a small precomputed table holding the 15 non-empty subset-sums of $\lbrace G, \phi(G), Q, \phi(Q) \rbrace$. The 4-bit mask selects which of those points to add this round.

Result: **128 doublings and ~120 additions total**. Four independent 128-bit double-and-add scalar multiplications would use $4 \times 128 = 512$ doublings and about $4 \times 64 = 256$ additions in expectation, plus the final sums. The dominant win is sharing the doublings; additions also drop because one subset-sum add handles all nonzero bits in a round.

## Idea 3: SZ + Horner — fast modular multiplication

Every point operation (`point_double`, `point_add`) is a handful of field multiplications in $\mathbb{F}_p$, where $p$ is a 256-bit prime. Doing modular multiplication directly in-VM is expensive: a $256 \times 256$-bit multiply produces a 512-bit intermediate that then has to be reduced modulo $p$. A single ECDSA verification uses well over a thousand modular multiplications, so the per-modmul cost dominates the field-arithmetic cost.

We use a **hint-and-check** pattern that turns modular multiplication into a Schwartz-Zippel-style polynomial-identity check evaluated at a single random point. The MASM procedures for this are generated by the small `miden-sz-codegen` crate for both the base and scalar fields of secp256k1.

**Limb encoding.** Each 256-bit operand is encoded as a degree-15 polynomial with 16-bit coefficients:

$$
a(x) = \sum_{i=0}^{15} a_i\, x^i, \qquad a_i \in [0, 2^{16}).
$$

At $x = W := 2^{16}$ the polynomial evaluates back to the integer $a$. The reason we pick **16-bit limbs and not 32-bit** for the identity is so that the coefficients of the polynomial product $a(x) \cdot b(x)$ stay small. These coefficients are convolutions $\sum_{j+k = i} a_j b_k$ of up to 16 limb pairs each, so they stay bounded by $16 \cdot 2^{32} = 2^{36}$ and fit comfortably in the Miden Goldilocks field ($|\mathbb{F}_M| \approx 2^{64}$). With 32-bit limbs the convolution would reach $\sim 2^{67}$ and wrap in the VM field.

**The naive identity, and why it isn't a polynomial identity.** We want to check $a \cdot b = q \cdot p + c$ over the integers, where an honest witness has $c = a \cdot b \bmod p$ and $q = \lfloor a b / p \rfloor$. As an *integer* relation, evaluating at $x = W$ gives $a(W)\cdot b(W) - q(W)\cdot p(W) - c(W) = 0$ — but as a *polynomial* identity, $a(x)\cdot b(x) - q(x)\cdot p(x) - c(x)$ is NOT the zero polynomial. Its coefficients only sum away when $x = W$ specifically. Schwartz-Zippel needs a true polynomial identity to have anything to say.

**The carry polynomial.** If the integer relation holds at $x = W$, then $a(x)\cdot b(x) - q(x)\cdot p(x) - c(x)$ vanishes at $W$. Equivalently, there exists a **carry polynomial** $e(x)$ for which

$$
a(x)\cdot b(x) - q(x)\cdot p(x) - c(x) = (W - x) \cdot e(x).
$$

This is the usual radix carry relation written as a polynomial identity. Let

$$
f(x) = a(x)\cdot b(x) - q(x)\cdot p(x) - c(x)
     = \sum_i f_i x^i,\qquad e(x)=\sum_i e_i x^i.
$$

Comparing coefficients in $f(x)=(W-x)e(x)$ gives

$$
f_i = W e_i - e_{i-1},\qquad e_{-1}=0,
$$

or, equivalently,

$$
f_i + e_{i-1} = W e_i.
$$

So $e_i$ is exactly the outgoing carry after processing coefficient $i$: the raw residual coefficient $f_i$ plus the incoming carry $e_{i-1}$ must be a multiple of the radix $W$.

The implementation stores 32 carry coefficients for alignment. For 16-limb modmul the real carry polynomial has degree at most 29, so the top two coefficients are zero. In the signed representation below, the VM checks this by asserting

$$
e_{pos}[30] = e_{neg}[30] = e_{pos}[31] = e_{neg}[31] = 0.
$$

For an honest witness, the host computes $e$ off-VM and supplies it as advice.

**Signed carries.** The coefficients of $e(x)$ can be negative (limb subtractions can borrow). Field elements aren't signed, so the host splits $e = e_{pos} - e_{neg}$, both with non-negative u32 coefficients. The identity the verifier actually checks is

$$
a(x)\cdot b(x) - q(x)\cdot p(x) - c(x) - (W - x)\bigl(e_{pos}(x) - e_{neg}(x)\bigr) = 0,
$$

No minimality condition is imposed on this split. For example, $(e_{pos,i}, e_{neg,i})$ and $(e_{pos,i}+t, e_{neg,i}+t)$ represent the same signed carry coefficient. This redundancy is harmless for the algebraic check: the identity depends only on the difference $\delta_i = e_{pos,i} - e_{neg,i}$. What matters is that both halves are individually u32-bounded, which bounds every $\delta_i$, and that the top two coefficients of both signed halves are zero.

**The check.** The MASM verifier evaluates this five-term identity at a Fiat–Shamir challenge point $\alpha$ in the quadratic extension of the Miden base field. The host supplies $\alpha$ through advice for efficiency, but the VM independently recomputes it and asserts equality, so the host cannot choose any arbitrary challenge.

The online transcript is:

$$
a_{\mathrm{u32}} \,\Vert\, b_{\mathrm{u32}} \,\Vert\, q_{15..0} \,\Vert\, c_{15..0} \,\Vert\, e_{pos,31..0} \,\Vert\, e_{neg,31..0}.
$$

The modulus polynomial $p$ is part of the public statement. For efficiency, it is placed first in the Fiat-Shamir statement and its Poseidon2 effect is precomputed once and then hardcoded. The online sponge starts from the resulting modulus-specific capacity state and then absorbs the instance-specific data above. Equivalently, the online transcript is pinned to the statement, including the modulus, without absorbing the modulus coefficients on every call.

The streamed witness coefficients are consumed in the same order used by the identity check. Each `adv_pipe` chunk writes coefficients to local memory, checks that they are u32 values, and absorbs them into the Poseidon2 transcript. The input operands $a,b$ live on the operand stack, so the verifier first stores them as u32 limbs, absorbs those limbs with `mem_stream`, and then splits them into u16 coefficients for Horner evaluation.

**Range checks and integer lifting.** A polynomial identity holding over the Goldilocks field $\mathbb{F}_M$ does not automatically imply the same identity over the integers: coefficients could differ by a multiple of the VM field modulus and still vanish in $\mathbb{F}_M$. To rule that out, the verifier also enforces:

- **u32 range checks** on every advice-streamed coefficient of $q, c, e_{pos}, e_{neg}$ via `u32assertw` (each felt is checked as it is streamed in). The verifier intentionally allows a redundant radix representation for $q$ and $c$: their coefficients need not be canonical u16 digits. For example, $c_0 = W, c_1 = 0$ and $c_0 = 0, c_1 = 1$ encode the same value at $x = W$; their difference is the carry move $W-x$. These are different polynomials away from $W$ and may require different carry witnesses, but uniqueness of the witness is not required.

What the verifier needs is that the encoded integers $Q = q(W)$ and $C = c(W)$ satisfy $a \cdot b = Q \cdot p + C$, and that $C$ is the canonical residue. The polynomial identity proves the integer equation: evaluating it at $x = W$ kills the carry term $(W - x)e(x)$. The final canonicality check recombines $c$ into eight u32 limbs and enforces $C < p$. The quotient witness $q$ needs no separate coefficient-canonicality check: given $a,b < p$ and $C < p$, the integer equation forces the represented integer $Q = q(W)$ to equal $(a \cdot b - C) / p = \lfloor a b / p \rfloor < p$. The coefficient vector for $q$ itself need not be the canonical base-$W$ expansion of $Q$.

The u32 bound is the range needed to lift the checked identity from the VM field back to an integer identity. In fact, write a coefficient of the checked polynomial as

  $$
  h_k = A_k - B_k - c_k - W\delta_k + \delta_{k-1},
  \qquad
  \delta_i = e_{pos,i} - e_{neg,i},
  $$

where $A_k$ is the $k$-th coefficient of $a(x)b(x)$ and $B_k$ is the $k$-th coefficient of $q(x)p(x)$. Under the enforced bounds, conservatively:

  $$
  |A_k| < 16\cdot 2^{32} = 2^{36},\quad
  |B_k| < 16\cdot 2^{48} = 2^{52},\quad
  |c_k| < 2^{32},\quad
  |W\delta_k| < 2^{48},\quad
  |\delta_{k-1}| < 2^{32}.
  $$

Thus $|h_k| < 2^{53}$ for every coefficient, comfortably below the Goldilocks modulus $2^{64} - 2^{32} + 1$. Therefore a checked equality in $\mathbb{F}_M$ cannot be hiding a nonzero multiple of the VM field modulus. This is the important safety condition for redundant representations: allowing noncanonical coefficients is fine only because their ranges are bounded tightly enough that no field wraparound can masquerade as an integer carry.
- **Top carry coefficients pinned to zero**: the VM asserts $e_{pos}[30] = e_{neg}[30] = e_{pos}[31] = e_{neg}[31] = 0$. The carry witness is stored in 32 slots for alignment; for this 16-limb modmul shape, only coefficients 0 through 29 can be nonzero.
- **Canonical reduction check** $c < p$: the verifier recombines adjacent base-$W$ coefficients of $c$ into eight u32 limbs, asserts those recombined limbs are valid u32 values, and checks the resulting u256 integer is less than $p$. Without this bound, the host could hint $c' = c + k \cdot p$ for some $k$ and adjust $q$ accordingly.

The surrounding field/scalar routines are responsible for the reduced-input precondition ($a, b < p$). The modmul proc itself enforces the canonical output bound $c < p$, but it does not separately prove that both inputs are reduced.

**Soundness model.** For any fixed malformed witness $(a, b, q, c, e_{pos}, e_{neg})$ whose polynomial identity is non-zero, a uniformly random $\alpha \in \mathbb{F}_M^2$ accepts with probability at most roughly $\text{degree}/|\mathbb{F}_M^2|$, about $2^{-123}$ for these parameters. The deployed verifier is rendered non-interactive using Fiat-Shamir, so the usual random-oracle assumption is part of the soundness model.

Putting the pieces together, under the reduced-input precondition: if the verifier accepts, then except with Schwartz-Zippel/Fiat-Shamir soundness error, the checked polynomial identity is valid. Evaluating that identity at $x = W$ kills the carry term. The range checks then lift the field identity to the integer equation $a \cdot b = q(W)\cdot p + c(W)$, and the canonical reduction $c(W) \in [0, p)$ gives $c(W) = a \cdot b \bmod p$ and $q(W) = \lfloor a b / p \rfloor$. The coefficient vector for $q$ itself need not be canonical.

## Idea 4: Hint-and-check for everything else

The same hint-and-check pattern handles the remaining building blocks that would otherwise be costly:

- **Modular inverse** ($s^{-1}$ for the $u_1, u_2$ derivation): host hints $b = a^{-1} \bmod p$; in-VM checks $a \cdot b \equiv 1 \pmod p$ via one modmul.
- **Point decompression** (recovering $Y$ from $(X, \text{parity})$ when a public key is given in compressed form): host hints $Y$; in-VM checks $Y^2 \equiv X^3 + 7 \pmod p$ and the parity bit matches. This costs three base-field modmuls (`X^2`, `X^3`, `Y^2`) plus cheap additions and comparisons.

## Cost

| Operation | Cycles | Measurement notes |
|---|---:|---|
| `verify_prehash_native` (full ECDSA verify) | **1,716,505** | typical valid sig |
| `point_double(G)` | 7,459 | |
| `point_add(G + 2G)` | 5,925 | |
| `modmul_k1_base` (one call) | ~1,138 | generated SZ verifier |

The ignored comparison bench `modmul_k1_base_hint_checker_comparison_cycles` records deterministic hint-and-check alternatives on the same operands:

| Checker | Cycles | Relative to generated SZ |
|---|---:|---:|
| Generic deterministic checker (computes $a \cdot b$ and $q \cdot p$ directly) | 4,026 | 3.54× slower |
| k1 pseudo-Mersenne deterministic checker | 2,696 | 2.37× slower |
| Generated SZ checker | 1,138 | 1.00× |

The full verifier cost is dominated by point operations inside the MSM. Further large reductions require either fewer point operations (for example a different MSM recoding) or a faster sound modular-multiplication verifier.

### Update on commit `940e023f2`

| Operation | Cycles | Measurement notes |
|---|---:|---|
| `verify_prehash_native` (full ECDSA verify) | **1,401,938** | typical valid sig |
| `point_double(G)` | 6,036 | |
| `point_add(G + 2G)` | 4,894 | |
| `modmul_k1_base` (one call) | ~918 | generated SZ verifier |

| Checker | Cycles | Relative to generated SZ |
|---|---:|---:|
| Generic deterministic checker (computes $a \cdot b$ and $q \cdot p$ directly) | 4,026 | 4.39× slower |
| k1 pseudo-Mersenne deterministic checker | 2,696 | 2.94× slower |
| Generated SZ checker | 918 | 1.00× |

### Update on commit `ea179d9bb`

  | Operation | Cycles | Measurement notes |
  |---|---:|---|
  | `verify_prehash_native` (full ECDSA verify) | **1,323,553** | typical valid sig |
  | `modmul_k1_base` (one call) | ~853 | generated SZ verifier |

  | Checker | Cycles | Relative to generated SZ |
  |---|---:|---:|
  | Generic deterministic checker (computes $a \cdot b$ and $q \cdot p$ directly) | 4,026 | 4.72× slower |
  | k1 pseudo-Mersenne deterministic checker | 2,696 | 3.16× slower |
  | Generated SZ checker | 853 | 1.00× |


crates/lib/core/Cargo.toml
```diff
@@ -41,11 +41,13 @@ miden-air = { workspace = true, optional = true }
 miden-ace-codegen = { workspace = true, optional = true }
 
 # External dependencies
+num = { version = "0.4", default-features = false, features = ["alloc"] }
```
huitseeker: CI is failing workspace inheritance because `num` is now declared twice in this crate. Could we move it to `[workspace.dependencies]` and use `num.workspace = true` here and in dev-deps?

crates/sz-codegen/src/emit.rs
```diff
@@ -0,0 +1,790 @@
...
+        rel.name
+    );
+}
+
+/// Poseidon2 sponge state after absorbing the modulus (16 u16 limbs, natural low-to-high
+/// order, two rate-8 chunks) from a zero-initialized state.
+///
+/// Only the capacity lanes (`state[8..12]`) are load-bearing; the first online absorb
+/// overwrites the rate. The MASM verifier therefore only embeds the capacity (see
+/// [`emit_initial_sponge_state`]).
```
huitseeker: The doc job is failing here because this public rustdoc link points at private `emit_initial_sponge_state`. Could this be plain code text, or link to a public item instead?

crates/lib/core/src/handlers/u256_modmul.rs
```diff
@@ -0,0 +1,391 @@
...
+pub(crate) const NUM_CARRY_COEFFS: usize = 32;
+/// Limb base for the u16-limb Schwartz-Zippel modmul family.
+const W: i64 = 1 << 16;
+
+/// Output of [`compute_modmul_witness`]: everything the SZ verifier needs to check
+/// `a * b = q * p + c`.
+///
+/// Fields:
+/// - `q_u16`, `c_u16`: quotient and remainder of the division, as u16 limbs.
+/// - `e_pos`, `e_neg`: non-negative halves of the signed carry polynomial (verified carry
```
huitseeker: The rustfmt job is failing on this doc block. Running `cargo +nightly fmt --all` rewrites this wrapping.

crates/lib/core/src/handlers/glv_split_k1.rs
```diff
@@ -0,0 +1,301 @@
...
+//!   k_a = k - c1·a1 - c2·a2
+//!   k_b = -(c1·b1 + c2·b2)
+//!
+//! For secp256k1's specific basis these formulas yield magnitudes `< 2^128`
+//! (`tests::split_magnitudes_strictly_below_2_pow_128` is a 256-sample regression guard).
+//!
+//! Soundness lives entirely on the verifier side: a handler bug or a malicious prover
+//! supplying a different witness can only ever cause the MASM verifier to reject.
+//!
+//! References: Gallant, Lambert, Vanstone, "Faster Point Multiplication on Elliptic Curves
```
huitseeker, line 25: Could we extend this reference block with the FakeGLV / https://ia.cr/2025/933 source for the proof-system-oriented GLV/hint-check methodology? Paper refs are cheap durable docs here, especially around the split formulas and why this shape is safe to verify in-VM.

crates/lib/core/asm/math/k1_point.masm
```diff
@@ -0,0 +1,1145 @@
...
+    loc_loadw_le.88        loc_load.116 add.8   mem_storew_le
+    loc_loadw_le.92        loc_load.116 add.12  mem_storew_le
+    loc_loadw_le.96        loc_load.116 add.16  mem_storew_le  dropw
+end
+
+# ===== 4-WAY MULTI-SCALAR MULTIPLICATION (MSM(4, 128)) ==========================================
+
+#! WARNING -- NOT A GENERIC MSM. SPECIALIZED FOR THE secp256k1 GLV VERIFIER.
+#!
+#! Computes `R = k_1·P_1 + k_2·P_2 + k_3·P_3 + k_4·P_4` via 4-way Shamir's trick at 128
```
huitseeker: Could we add a source citation for this Shamir/Straus step? e.g. https://ia.cr/2003/257. Future readers will recognize this as simultaneous scalar multiplication if the doc comment points to a standard ECC reference, and the older Straus vector-addition-chain paper is the scholarly root.

huitseeker (commented):
Overall LGTM!

crates/lib/core/src/handlers/glv_split_k1.rs
huitseeker: brain fart, this is ECDSA so you don't use FakeGLV, you're using the real OG GLV!

crates/lib/core/asm/math/k1_point.masm
Al-Kindi-0: Added a Straus/Shamir simultaneous-MSM reference in `msm_4_128_glv_k1`'s docstring (HMV Algorithm 3.48), rather than ePrint 2003/257 which is about trading inversions for multiplications and isn't quite related.

crates/lib/core/src/handlers/u256_modmul.rs
Al-Kindi-0: fixed

crates/sz-codegen/src/emit.rs
Al-Kindi-0: fixed

crates/lib/core/Cargo.toml
Al-Kindi-0: fixed

adr1anh:
I wonder if we can avoid the `e_pos/e_neg` split and instead represent the carry polynomial directly as a single centered signed polynomial.

Concretely, let the checked identity be

$a(x)b(x) - q(x)m(x) - c(x) = (W - x)e(x)$,

where each coefficient of `e` is represented as a field element but range-constrained to

$-2^{31} \le e_i < 2^{31}$.

This can be checked by asserting

$e_i + 2^{31} < 2^{32}$

after field addition. Positive coefficients then live in `[0, 2^31)`, and negative coefficients live near the top of the Goldilocks field, so the shifted value lands in `[0, 2^32)` exactly for the intended signed range.

This should also make the integer-lifting bound slightly stronger: instead of bounding
$\delta_i = e_{pos,i} - e_{neg,i}$ by roughly $2^{32}$, we get $|e_i| < 2^{31}$, hence $|W e_i| < 2^{47}$ rather than $2^{48}$.

The main implementation caveat is that the transcript/Horner evaluation should still use the actual signed field element `e_i`, not the shifted value. So the shifted value would need to be used only for the range check, or alternatively we could stream shifted coefficients and subtract the known offset polynomial in the identity.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -194,3 +198,453 @@ proc verify_prehash_impl
...
+#! `dest_pt_addr` (X, Y, is_infinity = 0) and pushes flag = 1. On any check failure the flag
+#! is 0 and the point at `dest_pt_addr` is whatever the math happened to write (the caller
+#! must check the flag before using the point).
+#!
+#! Stack: [dest_pt_addr, x_addr, parity_addr, ...]  ->  [flag, ...]
+#!
+#! Caller layout:
+#!   x_addr      : 8 u32 LE integer limbs (word-aligned read; 2 word loads).
+#!   parity_addr : 1 felt holding 0 or 1.
+#!   dest_pt_addr: 20 felts of point storage (5 words).
```
adr1anh: Nit: maybe denoting the output as `X_0, X_1, Y_0, Y_1, [is_infinity, 0, 0, 0]` makes the layout a bit clearer, while being explicit about the last 3 field elements getting clobbered.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -194,3 +198,453 @@ proc verify_prehash_impl
...
+
+    # Load parity from parity_addr to mem[16].
+    loc_load.55  mem_load  loc_store.16
+
+    # Check X < p. If X is non-canonical, substitute X = 1 so the f_k1::mul calls below do
+    # not trap on quotient overflow. The flag is already 0 in that case so the proc returns
+    # 0 regardless of what the math computes.
+    push.0xffffffff dup dup dup
+    push.0xffffffff dup
+    push.0xfffffffe.0xfffffc2f
```
adr1anh: Nit: It's only used twice, but we could make a procedure pushing the modulus. (we could also dup 6x)

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh, line 1: As a general comment (though I'm not exactly sure of our usual conventions), what would you think of using constants for the local storage pointers, allowing for example `loc_load.FLAG_PTR` avoiding direct indexing.

crates/lib/core/asm/math/f_k1.masm
```diff
@@ -0,0 +1,143 @@
...
+#!
+#! Returns 0 when a = 0; otherwise p - a (which is in (0, p)).
+pub proc neg(a: u256) -> u256
+    # eqz consumes one copy of a; keep another for the p - a branch.
+    dupw.1 dupw.1                                                # [a, a, ...]
+    exec.::miden::core::math::u256::eqz                          # [is_zero, a, ...]
+
+    if.true
+        # a = 0: drop original a, push zero word as the result.
+        dropw dropw padw padw
```
adr1anh: I Think this is redundant

crates/lib/core/asm/math/k1_point.masm
adr1anh, line 1: Nit: I think using these helpers could help with readability 
```
# ===== MEMORY HELPERS ===========================================================================

#! Copies a point-sized region (5 words / 20 felts) from `src_addr` to `dst_addr`.
#!
#! Stack: [dst_addr, src_addr, ...]  ->  [...]
proc copy_point
    repeat.5
        padw  dup.5  mem_loadw_le
        dup.4  mem_storew_le  dropw
        add.4  swap  add.4  swap
    end
    drop drop
end

#! Writes the canonical point-at-infinity representation to `out_addr`:
#! X = 0, Y = 0, is_infinity = 1, reserved = 0.
#!
#! Stack: [out_addr, ...]  ->  [...]
proc write_identity
    padw  dup.4         mem_storew_le dropw
    padw  dup.4 add.4   mem_storew_le dropw
    padw  dup.4 add.8   mem_storew_le dropw
    padw  dup.4 add.12  mem_storew_le dropw
    push.0.0.0.1  dup.4 add.16  mem_storew_le dropw
    drop
end
```

Al-Kindi-0:
Nice remark! I applied your idea but ended up taking the “stream shifted coefficients and subtract the known offset polynomial” variant.

Concretely, the verifier now uses a single carry witness `e_shifted` instead of `e_pos/e_neg`, with `e_shifted_i = e_i + 2^31` and checks

$a(\alpha)\cdot b(\alpha) - q(\alpha)\cdot m(\alpha) - c(\alpha) = (W - \alpha) * (e_{shifted}(\alpha) - offset(\alpha))$

where `offset = [2^31; 32]`.

The adaptation was to avoid doing the per-coefficient shift inside the VM. The host sends already-shifted u32 coefficients, so the verifier can keep using the existing cheap`u32assertw` path.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -194,3 +198,453 @@ proc verify_prehash_impl
...
+    mul                                                  # [r_valid, ...]
+    dup  loc_load.128  mul  loc_store.128                # flag &= r_valid; r_valid stays
+    not
+    if.true
+        # r = 1 fallback prevents q-overflow in modmul below; flag is already 0.
+        padw  loc_storew_le.36  dropw
+        push.0.0.0.1  loc_storew_le.32  dropw
+    end
+
+    # ----- Validate s in (0, n_k1); substitute s = 1 if not (also keeps inv(s) from trapping) -----
```
adr1anh: I think we should reject high-s signatures here.

Right now the native verifier checks only `s != 0 && s < n`. But secp256k1 ECDSA signatures are malleable: if `(r, s)` verifies, then `(r, n - s)` also verifies.

The deferred/host path goes through `miden-crypto`/`k256`, which rejects high-s signatures. So the native path can accept a signature that the host path rejects.

I think the right fix is to add a low-s check, i.e. `s <= n / 2`, before computing `s^-1`, and fold failure into the validity flag like the existing invalid-s fallback.

A good regression test would mutate a valid signature by replacing `s` with `n - s` and assert that `verify_prehash_native` returns `0`.

crates/lib/core/asm/math/f_k1.masm
```diff
@@ -0,0 +1,143 @@
...
+#! Multiplicative inverse: c = a^(-1) mod p, with a in (0, p).
+#!
+#! Hint-and-verify: emits `U256_INV_K1_BASE_EVENT` to ask the host for `inv = a^(-1) mod p`,
+#! then checks the result with one `modmul_k1_base` call (asserting `a * inv == 1 mod p`).
+#! The MASM proc traps with "f_k1::inv: hint not a multiplicative inverse" if the host
+#! returns a non-inverse, including when called with `a = 0`.
+#!
+#! Stack transition:
+#! [a0..a7, ...]  ->  [c0..c7, ...]
+pub proc inv(a: u256) -> u256
```
adr1anh, line 134: I think the hinted inverse should also check that the returned inverse is canonical.

Currently `inv` checks only that `a * inv == 1 mod p`, and `k1_scalar::inv` does the same modulo `n`. That proves the hint is congruent to the inverse, but not that the returned value is actually `< p` / `< n`.

For example, a malicious advice provider could return `p + 1` as `f_k1::inv(1)`, and the multiplication check would still pass. The proc would then return a non-canonical field element, even though the module contract says field/scalar values are canonical.

This probably does not show up with the honest Rust handler, but the proof should not rely on the advice provider being honest. Other hinted boundary values, like the `Y` value in point decompression, are range-checked before use.

I think the fix is to assert `inv < p` / `inv < n` before returning from `f_k1::inv` and `k1_scalar::inv`.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -194,3 +198,453 @@ proc verify_prehash_impl
...
+
+#! Verifies an ECDSA secp256k1 signature in-VM. Same calldata format as `verify_prehash`:
+#! pk is 33 BE bytes, digest is 32 BE bytes, sig is 65 BE bytes (r || s || v), all packed
+#! as u32-LE felts.
+#!
+#! Stack: [pk_ptr, digest_ptr, sig_ptr, ...]  ->  [result, ...]
+#!
+#! Returns 1 if the signature is valid, 0 otherwise. No trap on failure: every check
+#! contributes to a multiplicative validity flag, and the math is run with safe-fallback
+#! values whenever an input would otherwise trap a sub-proc.
```
adr1anh: I think the “No trap on failure” comment should mention the calldata precondition.

The parser uses `u32shr`, `u32and`, and `u32lt` directly on calldata felts. So if a calldata felt is not actually a valid `u32`, the native verifier can still trap before returning `0`.

That seems fine if the intended contract is that the caller/range constraints ensure the calldata is valid packed-u32 data before calling this proc. But I think the comment should say that explicitly, e.g. “No trap on cryptographic verification failure, assuming calldata felts are valid packed u32 values.”

That makes it clear this proc is not trying to validate arbitrary field elements by itself.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
Al-Kindi-0: Done

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
Al-Kindi-0: Done

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
Al-Kindi-0: Done

crates/lib/core/asm/math/f_k1.masm
Al-Kindi-0: Agreed, left over after an optimization pass

crates/lib/core/asm/math/k1_point.masm
Al-Kindi-0: Did the second one and skipped the first one as there was a perf hit

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
Al-Kindi-0: Good catch! Added the check now

crates/lib/core/asm/math/f_k1.masm
Al-Kindi-0: Good point!
I have split inv into canonical + fast variants in both f_k1 and k1_scalar:
- `inv` wraps `inv_unsafe` with a final `c < p` / `c < n` assert. We use this when the inverse is exposed to anything other than an immediate `modmul`.
- `inv_unsafe` (fast) returns a u32-valid u256 congruent to `a^-1` but not bounded by the modulus. Safe to use when the result is consumed by `modmul_k1_*` AND the other operand to that `modmul` is canonical (< p / < n). The SZ modmul's soundness only requires one canonical operand. All current `inv_unsafe` callsites satisfy that condition.
- The canonical wrapper adds ~75 cycles per call (~19k cycles on `verify_prehash_native`, ~1.45%). Routing the ~250 inv calls per verify through `inv_unsafe`helps avoiding the overhead due to the check.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
Al-Kindi-0: Agreed, and the docs are updated to be more explicit

huitseeker:
This is a substantial cryptographic implementation adding in-VM secp256k1 ECDSA verification to the Miden core library...

PR #3177, opened 2026-05-20
refactor: drop trace column access via constants

Nashtare:
## Describe your changes

Replace trace column accesses via constants with typed `Cols<T>` row access

I kept related constants in tests as we perform column-major access there. But these could be refactored too

- Updated `CHANGELOG.md'


processor/src/trace/parallel/tracer/trace_row.rs
```diff
@@ -481,23 +485,22 @@ impl<'a> CoreTraceGenerationTracer<'a> {
         stack: &StackState,
         decoder_row: DecoderRow,
     ) {
-        let mut row = [ZERO; CORE_TRACE_WIDTH];
+        let mut row_data = [ZERO; CORE_STORAGE_WIDTH];
```
adr1anh: Should we borrow this as a CoreCols, thereby removing the ranges?

processor/src/trace/chiplets/memory/mod.rs
```diff
@@ -369,31 +372,30 @@ impl Memory {
...
                     let (delta_hi, delta_lo) = split_element_u32_into_u16(delta);
-                    trace.set(row, D0_COL_IDX, delta_lo);
-                    trace.set(row, D1_COL_IDX, delta_hi);
-                    trace.set(row, D_INV_COL_IDX, delta.try_inverse().unwrap_or(ZERO));
-
-                    if prev_ctx == ctx && prev_addr == felt_addr {
-                        trace.set(row, FLAG_SAME_CONTEXT_AND_WORD, ONE);
+                    cols.d0 = delta_lo;
+                    cols.d1 = delta_hi;
+                    cols.d_inv = delta.try_inverse().unwrap_or(ZERO);
```
adr1anh: Followup issue: Batch invert memory deltas

processor/src/trace/chiplets/ace/trace.rs
adr1anh, line 1: While we're at it, it would be nice to change CircuitEvaluation to "row major" as well. Something like 
```rust
CircuitEvaluation {
  read_nodes: Vec<ReadNode>, // v_L and v_R
  eval_nodes: Vec<EvalNode>, // (id_L, v_L), (id_R, v_R), v_O
  multiplicities: Vec<u32> // multiplicities[i] = m_i
}
```

crates/test-utils/src/lib.rs
```diff
@@ -871,74 +871,86 @@ pub fn push_inputs(inputs: &[u64]) -> String {
 /// Helper function to get column name for debugging
 pub fn get_column_name(col_idx: usize) -> String {
```
adr1anh, line 936: We could create a `CoreCols<& str>` I think

adr1anh (approved):
This is really nice! I still want to go through the adjacent code to see if there are more things we can tackle, but it's looking great!

processor/src/trace/chiplets/memory/mod.rs
Nashtare: I've included it here

air/src/constraints/columns.rs
```diff
@@ -33,7 +33,7 @@ use crate::trace::{CHIPLETS_WIDTH, TRACE_WIDTH};
 /// `[T; NUM_CORE_COLS]` slice or the prefix of a `[T; TRACE_WIDTH]` row via
 /// `Borrow<CoreCols<T>>`.
 #[repr(C)]
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone, Copy, Default)]
```
adr1anh: I'll push back on `Copy`, since these structs can get quite large.

air/src/constraints/stack/columns.rs
```diff
@@ -2,10 +2,10 @@ use miden_core::program::MIN_STACK_DEPTH;
 
 /// Stack columns in the main execution trace (19 columns).
 #[repr(C)]
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct StackCols<T> {
     /// Top 16 stack elements s0-s15.
-    pub(crate) top: [T; MIN_STACK_DEPTH],
+    pub top: [T; MIN_STACK_DEPTH],
```
adr1anh, line 8: Perhaps a follow up. We should add `StackCols.{get, get_word, get_range}`, maybe mutable variants too. This would remove a bunch of boiler plate that appears in trace gen and the constraints.

air/src/constraints/chiplets/columns.rs
```diff
@@ -21,6 +21,13 @@ pub fn borrow_chiplet<T, S>(slice: &[T]) -> &S {
     &cols[0]
 }
 
+/// Zero-copy mutable cast from a slice to a `#[repr(C)]` chiplet column struct.
+pub fn borrow_chiplet_mut<T, S>(slice: &mut [T]) -> &mut S {
```
Al-Kindi-0: Should this be private ?

More generally, I wouldn't mind a bit of duplication with the following pattern

```rust
impl<T> Borrow<ChipletsCols<T>> for [T] {...}
```

as it is safe for internal use as well. Wdyt ?

air/src/constraints/op_flags/mod.rs
```diff
@@ -1054,32 +1054,31 @@ fn accumulate_depth_deltas<const N: usize, E: PrimeCharacteristicRing>(
...
+        row.decoder.extra[1] = bit_6 * bit_5;
+    }
 
     // Safety: `CoreCols` is `#[repr(C)]` with the same layout as `[Felt; NUM_CORE_COLS]`.
-    unsafe {
-        core::mem::transmute::<[Felt; NUM_CORE_COLS], crate::constraints::columns::CoreCols<Felt>>(
-            row,
-        )
-    }
+    unsafe { core::mem::transmute::<[Felt; NUM_CORE_COLS], CoreCols<Felt>>(row_data) }
```
Al-Kindi-0: Ideally this would be just
```rust
let row: &CoreCols<Felt> = row_data.as_slice().borrow();
*row
```
right ?

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,43 @@ impl MainTrace {
         TRACE_WIDTH
     }
 
+    /// Returns a typed view of the Core-AIR row at index `i`.
```
Al-Kindi-0: nit: Core trace ?

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,43 @@ impl MainTrace {
...
+    /// # Panics
+    /// Panics if `i` is past the Core height.
+    #[inline]
+    pub fn core_row(&self, i: RowIndex) -> &CoreCols<Felt> {
+        let r = i.as_usize();
+        let row = &self.storage.core_rm.values[r * NUM_CORE_COLS..(r + 1) * NUM_CORE_COLS];
+        row.borrow()
+    }
+
+    /// Returns a typed view of the Core-AIR row at index `i`, clamping past the Core height
```
Al-Kindi-0: same nit here and elsewhere

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,43 @@ impl MainTrace {
...
+    /// to the last Core row. Mirrors the last-row-replicate projection of [`Self::get`] for
+    /// indices past the Core height; **does not** apply the `CLK = r` bump.
+    #[inline]
+    pub fn core_row_projected(&self, i: RowIndex) -> &CoreCols<Felt> {
```
Al-Kindi-0: The docs are terse to read in my opinion, here and elsewhere, in contrast to the general tone of the repo, at least as I remember it.

What about

```rust
  /// Returns the stored core trace row for `i`, or the final core row if `i` is
  /// past the core trace height.
  ///
  /// This is only for APIs that render the split traces as one logical main
  /// trace. The returned row is not modified; callers must handle `clk` separately.
  fn core_row_or_last(&self, i: RowIndex) -> &CoreCols<Felt> {
```

Note that the name has changed, which also seems to me to be better intention-wise.

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,43 @@ impl MainTrace {
...
+        let r = i.as_usize();
+        let core_h = self.storage.core_rm.height();
+        let row_idx = r.min(core_h - 1);
+        let row =
+            &self.storage.core_rm.values[row_idx * NUM_CORE_COLS..(row_idx + 1) * NUM_CORE_COLS];
+        row.borrow()
+    }
+
+    /// Returns a typed view of the chiplets-AIR row at index `i`.
```
Al-Kindi-0: same nit here

air/src/trace/main_trace.rs
Al-Kindi-0: Should we check that `i` is in range ?

air/src/trace/main_trace.rs
```diff
@@ -271,20 +274,25 @@ impl MainTrace {
 
     /// Returns the value of the clk column at row i.
     pub fn clk(&self, i: RowIndex) -> Felt {
-        self.get(i, CLK_COL_IDX)
+        let r = i.as_usize();
```
Al-Kindi-0, line 299: So this and `get` have different behavior ? I wonder if we can have a consistency rule for this across similar situations. At least it would be good to document this

Al-Kindi-0 (commented):
Looks great!

Overall, probably in a follow-up PR, I think we should remove or gate the `MainTrace` methods that still encode the legacy monolithic trace view. In the split-table model, `Core` and `Chiplets` have independent heights, so there should generally be no need to read or synthesize values past a table's own height. If debug/test tooling wants to display both tables together, it can do that explicitly with optional row parts rather than projecting missing rows into values. This would simplify the accessors and make the multi-table semantics clearer.

air/src/constraints/chiplets/columns.rs
Nashtare: I'll go with a macro to reduce the duplication

air/src/trace/main_trace.rs
Nashtare: I've clarified it

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,60 @@ impl MainTrace {
...
+    /// Returns the stored core trace row at index `i`.
+    ///
+    /// # Panics
+    /// Panics if `i` is past the core trace height — callers needing access at unified-trace
+    /// row indices past the core height should use [`Self::core_row_or_last`] instead, which
+    /// applies the documented last-row-replicate projection.
+    #[inline]
+    pub fn core_row(&self, i: RowIndex) -> &CoreCols<Felt> {
+        let r = i.as_usize();
+        let row = &self.storage.core_rm.values[r * NUM_CORE_COLS..(r + 1) * NUM_CORE_COLS];
```
adr1anh: nit: `core_rm.row(r)`

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,60 @@ impl MainTrace {
...
+    /// Panics if `i` is past the unified trace height — i.e. past both the core and chiplets
+    /// heights. Callers iterating over `0..num_rows()` are within bounds by construction.
+    #[inline]
+    pub fn core_row_or_last(&self, i: RowIndex) -> &CoreCols<Felt> {
+        let r = i.as_usize();
+        assert!(r < self.num_rows(), "row index past the unified trace height");
+        let core_h = self.storage.core_rm.height();
+        let row_idx = r.min(core_h - 1);
+        let row =
+            &self.storage.core_rm.values[row_idx * NUM_CORE_COLS..(row_idx + 1) * NUM_CORE_COLS];
```
adr1anh: same nit

air/src/trace/main_trace.rs
```diff
@@ -184,6 +149,60 @@ impl MainTrace {
...
+    ///
+    /// # Panics
+    /// Panics if `i` is past the chiplets trace height. Unlike [`Self::core_row_or_last`],
+    /// this method does not project past the per-AIR height; callers iterating the unified
+    /// trace must guard against rows where the chiplets trace is shorter than the core.
+    #[inline]
+    pub fn chiplet_cols(&self, i: RowIndex) -> &ChipletCols<Felt> {
+        let r = i.as_usize();
+        let row =
+            &self.storage.chiplets_rm.values[r * NUM_CHIPLETS_COLS..(r + 1) * NUM_CHIPLETS_COLS];
```
adr1anh: same nit

air/src/trace/main_trace.rs
```diff
@@ -203,7 +222,8 @@ impl MainTrace {
         } else {
             let last = core_h - 1;
             dst[..NCC].copy_from_slice(&core_rm.values[last * NCC..(last + 1) * NCC]);
```
adr1anh: same

processor/src/trace/parallel/mod.rs
```diff
@@ -615,45 +614,50 @@ fn pad_core_row_major(core_trace_data: &mut Vec<Felt>, main_trace_len: usize) {
...
+    let (last_hasher_first_half, last_stack): ([Felt; 4], [Felt; STACK_TRACE_WIDTH]) = {
+        let last: &CoreCols<Felt> = core_trace_data[last_row_start..last_row_start + w].borrow();
+        let hs = &last.decoder.hasher_state;
+        let last_hasher: [Felt; 4] = [hs[0], hs[1], hs[2], hs[3]];
+        let mut last_stack = [ZERO; STACK_TRACE_WIDTH];
+        last_stack[..16].copy_from_slice(&last.stack.top);
+        last_stack[16] = last.stack.b0;
+        last_stack[17] = last.stack.b1;
+        last_stack[18] = last.stack.h0;
+        (last_hasher, last_stack)
```
adr1anh: Can we remove the manual indexing here?

processor/src/trace/parallel/mod.rs
```diff
@@ -377,46 +375,47 @@ fn fixup_stack_and_system_rows(
 fn push_halt_opcode_row(
     core_trace_data: &mut Vec<Felt>,
```
adr1anh, line 387: Can this be a row major matrix?

air/src/trace/main_trace.rs
Nashtare: core_rm.row(r) yields an iterator by value not a slice so I don't think that works?

adr1anh (commented):
Looks really good! My main nit that would be nice to address in this PR instead of as a followup is to review all the places where we manually get a row slice using the stride. 

There are several ways to do this
- Use RowMajorMatrix with get_row (more invasive)
- Use [`slice::as_chunks`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks) when the stride is a constant
- Maybe adding a helper or extension trait on a slice to get the slice at a given row

air/src/trace/main_trace.rs
adr1anh: `row_slice` should work

processor/src/trace/parallel/mod.rs
Nashtare: I don't think we'd want to though: `pad_core_row_major` needs `resize` and `push_halt_opcode_row` needs `extend_from_slice`, both of which would need to access the inner `.values` if we were to go with `RowMajorMatrix` here, so I can't see a clear advantage over a raw `Vec`

air/src/trace/main_trace.rs
Nashtare: Also `row_slice(r)` returns `Option<impl Deref<Target = [T]>>` which makes the borrow chain somewhat awkward too I'd say, compared to the direct access

huitseeker (commented):
LGTM!

huitseeker (approved):

air/src/trace/main_trace.rs
```diff
@@ -481,47 +491,57 @@ impl MainTrace {
...
     // CHIPLETS COLUMNS
     // --------------------------------------------------------------------------------------------
 
+    /// Returns `ZERO` past the chiplets-AIR height to match the unified-projection
+    /// semantics of `MainTrace::get` — `is_*_row` accessors iterate the full
+    /// `num_rows()` and rely on chiplet columns reading `ZERO` outside the chiplets matrix.
+    #[inline]
+    fn chiplet_col_or_zero(&self, i: RowIndex, col: usize) -> Felt {
+        if i.as_usize() >= self.storage.chiplets_rm.height() {
+            return ZERO;
```
huitseeker, line 510: I think this still makes rows past the chiplets height look like bitwise rows:

```rust
#[test]
fn projected_rows_past_chiplets_height_are_not_bitwise_rows() {
    let trace = build_trace_from_ops(vec![Operation::Noop; 128], &[]);
    let main = trace.main_trace();
    let (_core_matrix, chiplets_matrix) = main.to_core_chiplets_matrices();

    for row in chiplets_matrix.height()..main.num_rows() {
        let idx = RowIndex::from(row);
        assert!(!main.is_bitwise_row(idx), "projected row {row} classified as bitwise");
        let _ = main.chiplet_bitwise_a(idx);
    }
}
```

We can address that in #3171

air/src/trace/main_trace.rs
Nashtare: I hadn't seen your "_`row_slice` also work_" answer when I typed this, but let me back up my point. The reason I call this approach somewhat awkward is that we would turn 

```rust
        let row = &self.storage.core_rm.values[r * NUM_CORE_COLS..(r + 1) * NUM_CORE_COLS];
        row.borrow()
```

into

```rust
        let row = self.storage.core_rm.row_slice(i.as_usize()).expect("row in bounds");
        let cols: &CoreCols<Felt> = Borrow::<CoreCols<Felt>>::borrow(&*row);
        unsafe { &*(cols as *const CoreCols<Felt>) }
```

Now, if that is still preferred, then I'll change all 3 affected sites but I'm not sure the trade is worth it.

adr1anh: Ah I see now, thanks for showing the detail. That doesn't seem nearly as nice.

adr1anh: How about using https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks though? This takes advantage of the stride being constant.

Nashtare: Yeah this works neatly, thanks!

PR #3178, opened 2026-05-20
Fix MASM tooling, source paths, and package edge cases

### LINKED ISSUE #3238: `miden-format` removes intentional newlines
`miden-format` removes intentional newlines, e.g.:

```diff
# PROCEDURES
# =================================================================================================
-
# DELTA COMPUTATION
# -------------------------------------------------------------------------------------------------
-
#! Computes the commitment to the native account's delta.
```

It would be nice if this could be kept as-is.

huitseeker:
This PR fixes several small correctness gaps in the MASM-to-MAST assembly path. Per-commit review is advised.

- `miden-format` (#2906) now rewrites files atomically and reports stdout failures in `--stdin` mode. A failed file write no longer risks truncating the source file; formatter-as-filter output treats `BrokenPipe` as normal pipeline termination but reports other write/flush failures.
- Source URI handling now uses a restricted form. Schemes must use `://`; local paths and local `file://` URIs still work; unsupported or non-local URIs do not become paths.
- Package loading now handles a valid non-member package inside a broken workspace and enforces dependency-name expectations for project references. A path dependency that resolves to a package with a different declared name is rejected at the loader boundary.
- Local registry behavior is stricter and safer. It pins the miden-registry CLI name, keeps quiet invalid-version probes quiet, requires at least one version for availability, and writes index/artifact files atomically.
- MASM test helpers now report errors instead of panicking in linked-input compile paths, check diagnostic line counts exactly, reject generated memory addresses that overflow, and assert stack/memory expectations from a single execution.
- Assembly coverage now includes userspace package assembly, ~duplicate debug-decorator MAST nodes~ (edit: removed following #3182), and unknown CST `$...` special identifiers.
- Fixes #3238.

### DIFF
--- a/crates/package-registry-local/src/main.rs
+++ b/crates/package-registry-local/src/main.rs
@@ -5,13 +5,7 @@ use miden_package_registry::{PackageId, PackageRegistry, Version};
 use miden_package_registry_local::LocalPackageRegistry;
 
 #[derive(Debug, Parser)]
-#[command(
-    name = "miden-registry",
-    version,
-    about,
-    rename_all = "kebab-case",
-    arg_required_else_help(true)
-)]
+#[command(version, about, rename_all = "kebab-case", arg_required_else_help(true))]
 struct Cli {
     #[command(subcommand)]
     command: Command,
@@ -139,8 +133,6 @@ fn run(
 
 #[cfg(test)]
 mod tests {
-    use clap::CommandFactory;
-
     use super::*;
 
     fn empty_test_registry() -> (tempfile::TempDir, LocalPackageRegistry) {
@@ -152,18 +144,6 @@ mod tests {
         .unwrap();
 
         (tempdir, registry)
-    }
-
-    #[test]
-    fn help_uses_public_binary_name() {
-        let mut command = Cli::command();
-        assert_eq!(command.get_name(), "miden-registry");
-
-        let help = command.render_help().to_string();
-        assert!(
-            help.contains("Usage: miden-registry"),
-            "help should advertise the public binary name:\n{help}",
-        );
     }
 
     #[test]
--- a/crates/package-registry-local/src/main.rs
+++ b/crates/package-registry-local/src/main.rs
@@ -49,6 +49,13 @@ fn main() -> Result<ExitCode, Box<dyn std::error::Error>> {
     let cli = Cli::parse();
     let mut registry = LocalPackageRegistry::load_from_env()?;
 
+    run(cli, &mut registry)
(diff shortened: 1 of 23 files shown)


bitwalker:
I think it would best to base this PR on #3106. It'll encourage us to get that merged ASAP, and avoid the merge conflicts

crates/package-registry-local/src/main.rs
bitwalker, line 1: The premise of this commit is flawed - this tool will not be invoked directly as `miden-registry` (it _can_ be, but that's not the primary entry), users will actually be invoking it via `miden registry`, or some other subcommand. We don't want to hardcode the binary name here. I'm also not convinced this even needs to be tested - `clap` derives the binary name from the crate name, and if we ever changed the crate name, a whole bunch of other stuff would have to change too.

bitwalker: Sorry, GitHub UI for adding review comments to a commit is dumb - I was referring to the commit titled "Pin local registry CLI name".

bitwalker (commented):
Overall looks fine to me, other than the two things I noted (basing this on the `Library` removal branch - or at least the parts that touch `Assembler` APIs; and the `miden-registry` CLI name thing).

crates/package-registry-local/src/main.rs
huitseeker: OK! I removed the hardcoded `miden-registry` command name and dropped the help-name test. The remaining CLI tests only cover the quiet invalid-version behavior.

huitseeker (commented):
I’d prefer to keep this PR based on `next` for now rather than rebasing onto #3106. I agree #3106 is the direction of travel, but this PR is already reviewable as incremental hardening on the current branch. I checked the interaction beforehand:

- The package/project fixes still matter, ofc.
- The miden-format, source URI, CST parser, diagnostics, and MASM test-helper fixes are outside #3106’s Library removal path.
- The debug decorator dedup test still matters. #3106 changes package/library structure, but it does not replace the processor-side regression around duplicate MAST nodes with distinct debug metadata.
- The local registry and package availability fixes are still useful because #3106 does not remove registry/package resolution semantics.
- Conflicts are expected around assembler/linker/package API names and call sites, not around the core behavior being fixed. 

I’ll keep the overlap small and handle any final conflicts after #3106 lands.

bobbinth:
Not a review yet, but I'd probably merge this after https://github.com/0xMiden/miden-vm/pull/3106 is merged.

bitwalker (approved):

PR #3179, opened 2026-05-20
Reduce CodSpeed benchmark workload

huitseeker:
Keeps the CodSpeed benchmarks inside the existing single workflow, but make the simulated workloads small enough for the free monthly runtime budget.

The last 500 push-triggered workflow rows on origin/next represent 66 unique next pushes over 22.61 days, or about 88.85 pushes per month. With a 600 minute monthly CodSpeed budget, that leaves about 6.75 minutes per push.

The previous Blake3 CodSpeed harness proved 100 Blake3 iterations under CPU simulation and was still running after more than four hours. Make the Blake3 iteration count environment-configurable and run one iteration in CI. Also restrict the synthetic CodSpeed run to the smallest transaction snapshot while keeping all axes, including prove and verify, so proving remains covered without running every transaction proof on each push.


Al-Kindi-0 (approved):
LGTM
One thing though, if we are on a budget, then we can just keep the synth benchmarks and remove the Blake one

PR #3155, opened 2026-05-15
Restore metadata-neutral MAST shape on v0.22

huitseeker:
This is the v0.22 release-line version of #3151.

The root issue was not that procedure hashing directly includes debug info. It does not. The issue is that #2955 made debug/decorator metadata affect MAST node identity. That changed dedup and merge decisions. Those decisions changed MAST shape, and MAST shape changes procedure roots. See also #3151.

This PR reverts #2955 on top of v0.22.3 and prepares v0.22.4.

This restores the public `miden-standards` account component procedure roots from v0.22.1.

This intentionally reopens #2961 on the v0.22 line. The 0.22.4 patch favors stable public procedure roots over the debug metadata precision fixed by #2955.

PR #3182, opened 2026-05-21
Bring metadata-neutral MAST shape fix to next

huitseeker:
This brings the root-stability fix from main into next, following #3157 and #3155.

The root issue was that debug/decorator metadata could affect MAST node identity. That could change dedup and merge decisions, which could change MAST shape. Since MAST shape affects procedure roots, unchanged source could get new procedure roots.

This keeps debug and asm metadata out of node identity on next. Structurally equal nodes dedup again, and the first metadata wins.

This keeps the next branch on its current development-line versions while carrying the procedure-root fix forward.


plafer (approved):
👍

Issue #3169, opened 2026-05-18
Add event-based `core::debug` module to replace `debug.*` decorators


partylikeits1983:
## Summary

Introduce a `core::debug` module that provides print-style ("println") debugging for MASM (operand stack, memory, advice stack, and advice map), built on the existing `emit` **event** mechanism instead of MAST **decorators**.

This is a **prerequisite for #3069** (`Remove debug.* ops/decorators`) and, transitively, for #3040 (removing debug info from MAST). It must land before #3069: it gives users a non-decorator replacement for `debug.stack` / `debug.mem` / `debug.adv_stack` before those decorator-based instructions are removed.

## Motivation

From the discussion in #3069:

- `Decorator::Debug` is one of only two `Decorator` variants and a significant source of MAST complexity, since decorators force decorator-based MAST-node fingerprinting (see #3040).
- @bobbinth proposed a `core` `debug` module; @bitwalker endorsed "replacing them with procedures in `core` that provide equivalent functionality ... using `trace`/`emit` events that hosts can register callbacks for."
- Print-style debugging is genuinely relied upon by engineers writing MASM and must not regress when decorators are removed.

The `emit` instruction, `Operation::Emit`, and `host.on_event()` path already exists and carries **no** decorator/MAST cost. Moving debugging onto it lets #3069 delete `Decorator::Debug` cleanly without removing the capability.

### New `core::debug` module

Add a module to the core library (e.g. `crates/lib/core/asm/debug/mod.masm`) exposing procedures, each of which `emit`s a distinct well-known event:

```
use.core::debug

begin
    # ...
    exec.debug::debug_stack          # prints top 16 operand-stack elements
    # ...
    push.100 push.0
    exec.debug::debug_mem            # [start, end], prints mem range start..end, consumes args
    # ...
    exec.debug::debug_adv_stack_all  # prints the advice stack
    # ...
    push.KEY                         # WORD key
    exec.debug::debug_adv_map        # looks up key, prints the Vec<Felt>, consumes key
end
```

Proposed procedures (names mirror the existing `debug.*` instructions to keep migration mechanical):

| Procedure | Stack input | Behavior |
|---|---|---|
| `debug::debug_stack` | `[]` | print the top 16 operand-stack elements |
| `debug::debug_mem` | `[start, end]` | print memory in range `start..end`; consumes both args |
| `debug::debug_mem_all` | `[]` | print the full memory of the current context |
[...]

Issue #3069, opened 2026-04-28
Remove `debug.*` ops/decorators


bitwalker:
> @bobbinth I'd like to close this as not planned, and move any debugging-related features to the [`miden-debug`](https://github.com/0xMiden/miden-debug) or [`miden-lsp`](https://github.com/0xMiden/miden-lsp) repos instead.
> 
> Now that `miden-debug` is in sync with the VM, I think we'd be better off adding functionality there, or to the LSP server (so that we can surface things in IDE extensions, such as a pane to view advice map entries), rather than baking things like this into MAST/the processor. 

 _Originally posted by @bitwalker in [#2241](https://github.com/0xMiden/miden-vm/issues/2241#issuecomment-4307032950)_

---

This issue is meant to track the removal of the various `debug.*` operations (really decorators) in MASM.

The rationale is:

* Decorators as a concept add quite a bit of complication to MAST and the processor, and currently only exist for three things: `trace` (which really only the compiler uses), the various `debug.*` helpers, and instruction-level debug information (i.e. `AssemblyOp` and `DebugVarInfo`)
* We're working on replacing the use of `trace` and instruction-level debug info with metadata stored in the package, rather than MAST. This means we'd be carrying all the complexity of decorators just for the `debug.*` helpers.
* Up until now, the `debug.*` helpers have been critical for debugging, because no other debugging facilities really existed, or were usable in all cases (e.g. the protocol library). That is no longer the case. Not only is `miden-debug` in sync with the VM now, but it also supports debugging of transactions executed by the MockChain or `miden-client`. We've also implemented DAP support for our IDE extensions (VSCode and Zed), so interactive debugging in your editor is possible. The debugger provides much more functionality for examining the state of a program than the `debug.*` helpers at this point.

However, we're not quite to the point where we can remove decorators (see #3040), but a solution for the last remaining blocker is being worked on that will enable that. By removing the `debug.*` helpers in advance, we can push people towards `miden-debug` and the IDE extensions, not only improving those tools by increasing their use, but clearing the way for removing decorators without breaking existing MASM code.

If you actively use the `debug.*` helpers, I'd like to hear from you in this thread what pain points you experience in trying to use the debugger. I should note that there are three ways to use the debugger: 
  1. As a Rust library, either as a high-level wrapper around executing packages, or as the execution engine for the transaction kernel
  2. As a TUI-based interactive debugger (via `miden debug` or `miden-debug` directly)
[...]

PR #3185, opened 2026-05-22
Add event-based `miden::core::debug` print module

partylikeits1983:
## Summary

Implements #3169: a `miden::core::debug` MASM module providing print-style debugging of the operand stack, memory, advice stack, and advice map, built on the `emit` **event** mechanism instead of MAST **decorators**. This is a prerequisite for #3069 (removing the `debug.*` ops/decorators) and, transitively, #3040. Because events carry no MAST/decorator cost, moving debugging onto `emit` lets those decorators be removed without losing the capability.

The module exposes six procedures — `print_stack`, `print_mem`, `print_mem_all`, `print_adv_stack`, `print_adv_stack_all`, and `print_adv_map` — each a thin wrapper that emits a distinct well-known event. A single `DebugPrinter` host handler is registered for all six events via `CoreLibrary::handlers()`, so any host that loads the core library gets them by default. Unlike the `--debug`-gated decorators, these are explicit `exec` calls and print whenever invoked, regardless of debug mode. The handlers reuse the VM's existing tree-style debug formatting, now exposed from `miden_processor` as `write_stack`/`write_interval`/`format_value` and shared with the decorator path.

## Naming

Procedure names follow @bobbinth's `print_` proposal in #3169 rather than the `debug_*` names in the issue body. `print_adv_map` is the keyed lookup (equivalent to `print_adv_map_entry` from the thread); there is no separate whole-map dump. The `_all` variants are kept for parity with the existing `DebugOptions::MemAll` / full-advice-stack behaviors, though they could be dropped in favor of passing a large `end` value.


crates/lib/core/src/lib.rs
```diff
@@ -143,7 +144,13 @@ impl CoreLibrary {
             (LOWERBOUND_ARRAY_EVENT_NAME, Arc::new(handle_lowerbound_array)),
             (LOWERBOUND_KEY_VALUE_EVENT_NAME, Arc::new(handle_lowerbound_key_value)),
             (AEAD_DECRYPT_EVENT_NAME, Arc::new(handle_aead_decrypt)),
-        ]
+        ];
+
+        // Print-style debugging handlers backing the `miden::core::debug` module. These print VM
+        // state to stdout whenever the corresponding `print_*` procedure is executed.
+        handlers.extend(debug_handlers());
```
huitseeker: Right now `miden-core-lib` defaults to `std`, but its feature only enables `miden-assembly/std` and `miden-utils-sync/std`. 

The default debug handler uses `StdoutWriter`, and that writer only calls `std::print!` when `miden-processor` itself has `std` enabled, so these `print_*` handlers can be registered and still silently write nothing unless another crate happens to enable that feature. A small default-handler test through `CoreLibrary::handlers()` would catch this too.

huitseeker (commented):
This is great and very close, I've opened #3193 to fix a few issues.

crates/lib/core/asm/debug.masm
```diff
@@ -0,0 +1,103 @@
...
+
+# CONSTANTS
+# ================================================================================================
+
+const PRINT_STACK_EVENT = event("miden::core::debug::print_stack")
+const PRINT_MEM_EVENT = event("miden::core::debug::print_mem")
+const PRINT_MEM_ALL_EVENT = event("miden::core::debug::print_mem_all")
+const PRINT_ADV_STACK_EVENT = event("miden::core::debug::print_adv_stack")
+const PRINT_ADV_STACK_ALL_EVENT = event("miden::core::debug::print_adv_stack_all")
+const PRINT_ADV_MAP_EVENT = event("miden::core::debug::print_adv_map")
```
bobbinth: I would remove `PRINT_MEM_ALL_EVENT` and `PRINT_ADV_STACK_ALL_EVENT` (setting a large "to" value should give us the same functionality).

But I would split `PRINT_ADV_MAP_EVENT` into two:

- `PRINT_ADV_MAP_EVENT` - to print the entire advice map.
- `PRINT_ADV_MAP_ITEM_EVENT` - to print just a specific item.

crates/lib/core/asm/debug.masm
```diff
@@ -0,0 +1,103 @@
...
+
+#! Prints the full memory of the current context.
+#!
+#! Inputs:  [...]
+#! Outputs: [...]
+#!
+#! Cycles: 3
+pub proc print_mem_all
+    emit.PRINT_MEM_ALL_EVENT
+end
```
bobbinth, line 64: Related to the above comment, this could be:

```
pub proc print_mem_all
    push.MAX_FELT_VALUE push.0
    emit.PRINT_MEM_EVENT
    drop drop
end
```

crates/lib/core/src/lib.rs
bitwalker: I'm not sure we should register these by default - the debugger for example would want to handle these differently in order to redirect the output to an appropriate place (e.g. in the TUI, printing anything directly to stdout will break the rendering of the TUI; while in the IDE extensions, we would want to direct the output to the debug log, and so on).

bitwalker (dismissed):
Overall, this looks good, the only issue I have is with the handlers, as noted in my review.

huitseeker:
- Removed default registration of `miden::core::debug` stdout print handlers from `CoreLibrary::handlers()`.
- Documented that debug print handlers must be registered explicitly via `handlers::debug::debug_handlers()` or host-specific handlers.
- Updated tests so default core handlers are asserted not to include debug printer events.
- Fixed the out-of-bounds `print_mem `regression test to explicitly register debug handlers and assert `MemoryError::AddressOutOfBounds`.

crates/lib/core/src/lib.rs
huitseeker: Changed the default.

crates/lib/core/asm/debug.masm
huitseeker: I applied this for `print_adv_stack_all`: it now emits `PRINT_ADV_STACK_EVENT` with `[0, MAX_FELT_VALUE)`, while the old all-event name is only kept host-side as a legacy alias. For memory, I kept `PRINT_MEM_ALL_EVENT` distinct because the same sentinel approach collides with user-provided `print_mem` arguments and bypasses range validation. 

For advice map, I split the full-map and keyed-item paths, but kept `print_adv_map`/`PRINT_ADV_MAP_EVENT` as the keyed legacy alias and added `print_adv_map_all` plus `print_adv_map_item` for the new explicit behavior.

crates/lib/core/asm/debug.masm
huitseeker: It makes a normal `print_mem` call with range `[0, MAX_FELT_VALUE)` indistinguishable from print_mem_all at the handler. That means an out-of-bounds `print_mem` range would silently print all memory instead of returning the existing address validation error. I restored a distinct `PRINT_MEM_ALL_EVENT` for memory and added a regression test for that case.

bobbinth: Is that an issue though? We only print out "initialized" memory, right? If so, the behavior is fine in both cases.

Or is the issue that `MAX_FELT_VALUE` is not a valid memory address? If so, we could probably just used `U32_MAX_VALUE` (or w/e is the largest valid memory address).

crates/lib/core/src/lib.rs
bobbinth: I would actually keep the default as is (i.e., printing to stdout) but allow to override/change this behavior. The main motivation is that users who are most likely to use this functionality need quick&dirty debugging functionality and having to do extra configuration (especially if this configuration requires some code changes in unrelated places) would be too difficult for them.

But, the environment like the debugger could configure the handlers as needed with custom writers etc.

Regarding configuring the handlers: how would this look like right now? For example, would the user need to intercept the `DefaultHost::load_library()` somehow to be able to configure the handlers?

I think a good approach may be to configure a "writer" for the host, and then the host would pass the writer into each handler in case the handler needs to print something. This would effectively change the handler signature to something like:

```rust
pub trait EventHandler: Send + Sync + 'static {
    fn on_event<W: fmt::Write + Send + Sync = StdoutWriter>(
        &self,
        process: &ProcessorState,
        writer: W,
    ) -> Result<Vec<AdviceMutation>, EventError>;
}
```

Thought, maybe there is a better way.

huitseeker (approved):
LGTM!

crates/lib/core/src/lib.rs
huitseeker: The default should stay simple: `CoreLibrary::handlers` can keep registering stdout debug handlers, so `exec.debug::print_*` works without extra host setup.

Custom tools can still replace those handlers. Today that means calling `DefaultHost::replace_handler` for each debug event after loading the core library, or building a `HostLibrary` with custom handlers. That is workable (though arguably not very ergonomic).

A larger host-level writer API sounds useful, but I would keep it as a follow-up. A generic `on_event<W: fmt::Write>` would not fit the current `Arc<dyn EventHandler>` setup because generic methods are not object-safe.

processor/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub use execution_options::{ExecutionOptions, ExecutionOptionsError};
 pub use fast::{BreakReason, ExecutionOutput, FastProcessor, ResumeContext};
 pub use host::{
     BaseHost, FutureMaybeSend, Host, MastForestStore, MemMastForestStore, SyncHost,
-    debug::DefaultDebugHandler,
+    debug::{DefaultDebugHandler, StdoutWriter, format_value, write_interval, write_stack},
```
bobbinth, line 54: These are exported from here because we want to share them between the debug procedures and debug decorators, right? I'm assuming once debug decorators go away, the procedures will move into the `core` library - yes?

crates/lib/core/asm/debug.masm
```diff
@@ -0,0 +1,138 @@
...
+# ================================================================================================
+
+const PRINT_STACK_EVENT = event("miden::core::debug::print_stack")
+const PRINT_MEM_EVENT = event("miden::core::debug::print_mem")
+const PRINT_MEM_ALL_EVENT = event("miden::core::debug::print_mem_all")
+const PRINT_ADV_STACK_EVENT = event("miden::core::debug::print_adv_stack")
+const PRINT_ADV_MAP_EVENT = event("miden::core::debug::print_adv_map")
+const PRINT_ADV_MAP_ALL_EVENT = event("miden::core::debug::print_adv_map_all")
+const PRINT_ADV_MAP_ITEM_EVENT = event("miden::core::debug::print_adv_map_item")
+const MAX_FELT_VALUE = 18446744069414584320
```
bobbinth, line 22: How is `PRINT_ADV_MAP_ITEM_EVENT` different from `PRINT_ADV_MAP_EVENT`? I think we just need the `PRINT_ADV_MAP_ITEM_EVENT`.

Also, I still don't think we need the `PRINT_MEM_ALL_EVENT`. If the problem is what gets printed out, we can change the semantics of how we print out memory (i.e., not print out un-initialized memory).

crates/lib/core/asm/debug.masm
```diff
@@ -0,0 +1,138 @@
...
+#!
+#! Where:
+#! - KEY is the word used as the advice map key.
+#!
+#! Cycles: 7
+pub proc print_adv_map_item
+    emit.PRINT_ADV_MAP_ITEM_EVENT
+    # => [KEY, ...]
+    dropw
+end
```
bobbinth: Related to the previous comment - these procedures seem to be identical.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -0,0 +1,288 @@
...
+pub const PRINT_MEM_ALL_EVENT_NAME: EventName = EventName::new("miden::core::debug::print_mem_all");
+/// Prints the advice stack in the range `[start, end)`.
+pub const PRINT_ADV_STACK_EVENT_NAME: EventName =
+    EventName::new("miden::core::debug::print_adv_stack");
+/// Prints the entire advice stack.
+///
+/// This legacy event name is kept as a host-side alias for already-assembled programs. The MASM
+/// `print_adv_stack_all` procedure emits [`PRINT_ADV_STACK_EVENT_NAME`] with an unbounded range.
+pub const PRINT_ADV_STACK_ALL_EVENT_NAME: EventName =
+    EventName::new("miden::core::debug::print_adv_stack_all");
```
bobbinth: Why do we need to keep the "legacy" event here? We are introducing the new functionality here - there is no "legacy" yet - or am I missing something?

crates/lib/core/src/handlers/debug.rs
```diff
@@ -0,0 +1,288 @@
...
+/// The default set prints operand-stack and memory state to stdout. Advice-stack and advice-map
+/// printers are excluded because they may expose witness data.
+pub fn default_debug_handlers() -> Vec<(EventName, Arc<dyn EventHandler>)> {
+    let printer: Arc<dyn EventHandler> = Arc::new(DebugPrinter::default());
+    vec![
+        (PRINT_STACK_EVENT_NAME, printer.clone()),
+        (PRINT_MEM_EVENT_NAME, printer.clone()),
+        (PRINT_MEM_ALL_EVENT_NAME, printer),
+    ]
+}
```
bobbinth, line 63: It is not clear to me that we need these:

- In a security-sensitive context, we shouldn't have any printing - so, we'd use the `noop_debug_handlers().
- These are debug-oriented instruction and people shouldn't leave debug instructions in production code. In the development context, we don't really care because the user running the program is doing so for debug purposes.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -0,0 +1,288 @@
...
+        (PRINT_ADV_MAP_EVENT_NAME, handler.clone()),
+        (PRINT_ADV_MAP_ALL_EVENT_NAME, handler.clone()),
+        (PRINT_ADV_MAP_ITEM_EVENT_NAME, handler),
+    ]
+}
+
+/// Returns the `(EventName, handler)` pairs that back the full `miden::core::debug` module.
+///
+/// All events share a single [`DebugPrinter`] instance writing to stdout.
+pub fn debug_handlers() -> Vec<(EventName, Arc<dyn EventHandler>)> {
```
bobbinth, line 84: I think these should be the default handlers.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but in the interest of time, we can address them in a follow-up too. The main ones are:

- Remove un-needed events and extra code.
- Move printing procedures for stack to the `core` library crate.

All of these can be done as a part of https://github.com/0xMiden/miden-vm/issues/3069.

crates/lib/core/asm/debug.masm
```diff
@@ -0,0 +1,138 @@
...
+const PRINT_ADV_MAP_ALL_EVENT = event("miden::core::debug::print_adv_map_all")
+const PRINT_ADV_MAP_ITEM_EVENT = event("miden::core::debug::print_adv_map_item")
+const MAX_FELT_VALUE = 18446744069414584320
+
+# PROCEDURES
+# ================================================================================================
+
+#! Prints the entire operand stack.
+#!
+#! This is a drop-in replacement for the `debug.stack` decorator.
```
adr1anh, line 29: Is it necessary to mention this if we are going to remove the debug decorators in this same release?

crates/lib/core/src/handlers/debug.rs
huitseeker: You are right. I removed the extra `PRINT_ADV_STACK_ALL_EVENT_NAME`.

crates/lib/core/src/handlers/debug.rs
huitseeker: I kept `noop_debug_handlers()` as the simple way to make programs with debug events run without output.

crates/lib/core/src/handlers/debug.rs
huitseeker: Default handlers now cover stack and memory printing to stdout.

processor/src/lib.rs
huitseeker: Yes. These helpers are shared while `debug.*` decorators still exist. Once decorators are removed under #3069, the printing helpers can move to `miden-core-lib`.

crates/lib/core/asm/debug.masm
huitseeker: I removed `PRINT_ADV_MAP_EVENT` and the `print_adv_map` alias. The keyed path is now only `print_adv_map_item`.

I kept `PRINT_MEM_ALL_EVENT`. `print_mem` validates an explicit caller range; `print_mem_all` prints initialized memory. Keeping separate events avoids special sentinel rules in  the handler.

crates/lib/core/asm/debug.masm
huitseeker: Removed the duplicate `print_adv_map` procedure. The keyed lookup procedure is now only `print_adv_map_item`.

crates/lib/core/asm/debug.masm
huitseeker: The issue is semantic, not only the max value. `print_mem` means “print this caller range,” so invalid ranges should still fail. `print_mem_all` means “print initialized memory.” A separate event keeps that distinction clear.

crates/lib/core/asm/debug.masm
huitseeker: In general, we're not good at producing migration guides — I actually like having breadcrumbs. It's more of a comment for my upcoming PR which actually removes debug decorators.

PR #3186, opened 2026-05-22
Fix CodSpeed walltime Criterion integration

huitseeker:
The CodSpeed job currently runs the synthetic transaction bench but uploads no walltime results.

The cause is that walltime builds do not set cfg(codspeed), so our manual backend switch selects plain Criterion instead of codspeed-criterion-compat. The bench finishes, but CodSpeed has no result files to process.

This changes the synthetic bench to use codspeed-criterion-compat as the criterion crate, matching CodSpeed’s Rust docs. The compat crate now chooses the right backend for normal Criterion runs, CodSpeed walltime runs, and CodSpeed analysis runs. The old cfg(codspeed) switch and lint allowance are removed.

See #3183, #3179, #3161

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1946,7 +1946,6 @@ name = "miden-vm-synthetic-bench"
 version = "0.24.0"
 dependencies = [
  "codspeed-criterion-compat",
- "criterion",
  "miden-processor",
  "miden-vm",
  "serde",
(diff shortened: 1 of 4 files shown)


plafer (approved):

Al-Kindi-0 (approved):
LGTM

PR #3187, opened 2026-05-22
Change semantics of `LoopNode` to do-while semantics

plafer:
Partially addresses #1684

## Summary

Converts the `LoopNode` to **do-while semantics**: the body is entered unconditionally and the trailing condition is read only at REPEAT/END. The head-controlled entry check that `while.true` always implied is now performed by a `SplitNode` that the assembler inserts around the `LoopNode`.

This is the first subset of [#1684](https://github.com/0xMiden/miden-vm/issues/1684) — only the LOOP-side changes described in section 1 of the issue comment.

## Changes

- **Processor / continuation stack** — `Continuation::FinishLoop` no longer carries a `was_entered` bit; `start_loop_node` enters the body unconditionally, `finish_loop_node` always drops the trailing condition on END.
- **Assembler** — `Op::While { body }` now compiles to `Split { on_true: Loop { body }, on_false: noop }`. The `while.true` asm_op is attached to both nodes so source-mapped diagnostics from inside the body still resolve.
- **No MASM syntactic change** — `while.true` is still the user-facing keyword and parses identically. The MAST printer for `LoopNode` now reads `loop` instead of `while.true` to reflect the underlying semantics.
- **ExecutionTracer / parallel tracer** — `StartNode` for `MastNode::Loop` no longer decrements the overflow table (LOOP reads no stack input).
- **`NodeFlags::loop_entered` → `NodeFlags::is_loop`** — the `h5` decoder column is unchanged, but its meaning shifts from "the LOOP body executed at least once" to "this END is closing a LOOP node" (in which case the `END` node must drop an element from the stack).
- **AIR constraints**
  - LOOP dropped from the SPLIT/LOOP s0-binarity check (LOOP no longer reads s0).
  - LOOP moved out of `left_shift_depth1` / `left_shift_scalar` and into `no_shift_depth0`.
  - `f_loop_body = loop_op * s0 + repeat` simplified to `loop_op + repeat` (degree drops from 6 to 5).
  - `RELATION_DIGEST` regenerated; `num_eval_gates` 5540 → 5532.
- **Docs**: updated the docs

## Error message regression

When a non-binary value sits at the top of the stack at `while.true` entry, the binarity error now originates from the wrapping SPLIT, so the message reads "if statement expected a binary value..." instead of "loop condition must be a binary value...". The source pointer still lands on the `while.true` token. Iteration-time binarity (REPEAT/END) is unchanged. See `processor::tests::test_diagnostic_not_binary_value_loop_node`. However, I assume this will be addressed naturally by subsequent PRs (e.g. whether we remove `while.true` loops, or whete


air/src/constraints/op_flags/mod.rs
```diff
@@ -338,6 +338,8 @@ where
             op5(opcodes::SPAN),
             // +JOIN         — control flow: begins join block
             op5(opcodes::JOIN),
+            // +LOOP         — control flow: do-while LOOP reads no stack input
+            op5(opcodes::LOOP),
```
huitseeker, line 342: This stack effect change looks like it also needs the block-stack lookup path to stop using `s0` as the loop flag. `air/src/constraints/lookup/buses/block_stack_and_range_logcap.rs` still emits the `LOOP` push with `is_loop = s0`, but after this change `LOOP` no longer reads `s0`.

Take a wrapped `Split -> Loop` program on stack `[1, 42]`: the SPLIT consumes the entry condition, so the LOOP row has `s0 = 42`, while the matching loop END uses `is_loop = 1`. A local one-off decoder-bus test that expects the LOOP push to use `ONE` fails with the expected `BlockStackMsg` missing. 

Should the `LOOP` block-stack push use `ONE` here instead of `s0`?

plafer: Ahh yes good catch! Fixed in [25b8211](https://github.com/0xMiden/miden-vm/pull/3187/commits/25b82116a983223b2793e6c43341121aee8cadff) (including a regression test)

huitseeker (approved):
This LGTM. Beware, the LoopNode struct doc comment reads:

 ```rust
   /// A Loop node defines condition-controlled iterative execution. When the VM encounters a Loop
   /// node, it will keep executing the body of the loop as long as the top of the stack is `1``.
 ```

 The first sentence describes old while-entry semantics ("as long as the top of the stack is 1"), but under do-while the entry is unconditional — the condition is only checked at REPEAT/END, not at entry. The second sentence is still correct for the iteration behavior.

air/src/constraints/op_flags/mod.rs
```diff
@@ -932,6 +930,7 @@ impl<E: PrimeCharacteristicRing> OpFlags<E> {
         /// Operation Flag of SPLIT operation.
         split => opcodes::SPLIT,
         /// Operation Flag of LOOP operation.
+        #[expect(dead_code)]
         loop_op => opcodes::LOOP,
```
bobbinth, line 934: Why do we need to keep it here?

docs/src/design/stack/op_constraints.md
```diff
@@ -201,7 +201,7 @@ This group contains operations which require constraints with degree up to $3$.
 | `LOGPRECOMPILE` |     $94$     |   `101_1110`    | [Crypto ops](./crypto_ops.md#log_precompile) |     $5$     |
 | `<unused>`    |     $95$     |   `101_1111`    |                                        |     $5$     |
 
-Note that the `SPLIT` and `LOOP` operations are grouped together under the common prefix `101010`, and thus can have a common flag of degree $4$ (using $e_0$ for degree reduction). This is important because both of these operations shift the stack to the left.
+Note that the `SPLIT` and `LOOP` operations share the common prefix `101010` and can be detected together with a flag of degree $4$ (using $e_0$ for degree reduction). Only `SPLIT` shifts the stack to the left, however: `LOOP` is do-while and reads no stack input — see [LOOP block decoding](../decoder/index.md#loop-block-decoding).
```
bobbinth, line 204: Since we cannot use the `101010` common prefix any more to detect left-shift operations, does this cause any issues? That is, does degree of some constraint need to increase because just `SPLIT` by itself now requires degree 5 instead of 4?

One other thought: it may be possible to move `LOOP` to the currently unused slot `101_1111`, and move `REPEAT` from the very-high degree group in place of `LOOP` (so that `SPLIT` and `REPEAT` share a common prefix). At least from left-shift detection standpoint, this should allow using degree 4 for both.

If the above works, then we can free up space for one instruction in the very-high degree group.

processor/src/tests/mod.rs
```diff
@@ -1081,16 +1081,20 @@ fn test_diagnostic_not_binary_value_loop_node() {
...
     let err = build_test.execute().expect_err("expected error");
+    // The entry-check error originates from the SPLIT that the assembler wraps around the LOOP
+    // when desugaring `while.true`, so the message reads "if statement". The source pointer is
+    // still the `while.true` token because the SPLIT carries that asm_op. The iteration-check
+    // (REPEAT/END) still produces a `NotBinaryValueLoop` error — see masm_errors_consistency
+    // case_2 for that path.
     assert_diagnostic_lines!(
         err,
-        "  x loop condition must be a binary value, but got 2",
+        "  x if statement expected a binary value on top of the stack, but got 2",
```
bobbinth, line 1091: nit: maybe the message here would read slightly better if it said "conditional statement expected...". This should apply equally good to `if.true` and `while.true`.

PR #3188, opened 2026-05-23
refactor(utils-sync): fix OnceLockCompat API contract

Lubov66:
Found and fixed a semantic incompatibility in OnceLockCompat<T>, same type name, different runtime behavior depending on feature flags.Found and fixed a semantic incompatibility in OnceLockCompat<T>, same type name, different runtime behavior depending on feature flags. On no_std, take() always returned None instead of the value, and Clone silently reset the cache instead of copying it. Replaced the std type alias with a unified wrapper, renamed take() to reset(), and removed the misleading Clone impl.


crates/assembly/src/tests.rs
```diff
@@ -1595,6 +1595,7 @@ fn test_push_word_slice_invalid() {
         push.SAMPLE_WORD[6..3]
     end
     "
+        .to_string()
```
huitseeker: Small nit: `source_file!` already calls `$source.to_string()` in both macro arms, so this creates a `String` and then converts it again inside the macro. Passing the literal directly should keep the clippy fix while avoiding the extra allocation.

huitseeker (commented):
Please add a line to the Changelog marked as Breaking, and linking to this PR!

crates/assembly/src/tests.rs
Lubov66: Done, passed the literal directly without .to_string()

core/src/mast/mod.rs
```diff
@@ -218,6 +218,20 @@ impl MastForest {
...
+/// Clone implementation
+impl Clone for MastForest {
+    fn clone(&self) -> Self {
+        Self {
+            nodes: self.nodes.clone(),
+            roots: self.roots.clone(),
+            advice_map: self.advice_map.clone(),
+            debug_info: self.debug_info.clone(),
+            // commitment_cache is derived data — reset on clone, will be recomputed on next access
+            commitment_cache: OnceLockCompat::new(),
```
huitseeker: Could this be rebased onto the eager `commitment` field from `next`? On this head, `cargo check -p miden-core` fails because `MastForest` no longer has `commitment_cache`:

```text
error[E0560]: struct `MastForest` has no field named `commitment_cache`
   --> core/src/mast/mod.rs:231:13
    |
231 |             commitment_cache: OnceLockCompat::new(),
    |             ^^^^^^^^^^^^^^^^ `MastForest` does not have this field
    |
    = note: available fields are: `commitment`
```

The reset calls below seem to have the same stale assumption; they probably need to update `self.commitment` again instead of invalidating a cache.

Lubov66: Done, updated to use the eager commitment field

core/src/mast/mod.rs
```diff
@@ -255,7 +268,12 @@ impl MastForest {
     ///   clearly doesn't belong to this MAST forest).
     #[cfg(any(test, feature = "arbitrary"))]
     pub fn make_root(&mut self, new_root_id: MastNodeId) {
-        self.mark_root(new_root_id);
+        assert!(new_root_id.to_usize() < self.nodes.len());
```
huitseeker: Could this keep calling `mark_root(new_root_id)`? This now repeats the same invariant logic as the private helper just above, so the next change to root marking would need to update both paths.

huitseeker (commented):
Please reduce the duplication between `mark_root` and its alias `make_root`, then this LGTM!

core/src/mast/mod.rs
```diff
@@ -218,6 +218,19 @@ impl MastForest {
     }
 }
 
+// ------------------------------------------------------------------------------------------------
+/// Clone implementation
+impl Clone for MastForest {
```
huitseeker: Since `commitment` is now stored directly, this custom `Clone` looks the same as the derived one: clone the owned fields and copy/clone the `Word`. Could we put `Clone` back in the derive and drop this impl?

Lubov66: Done, switched back to #[derive(Clone)]

crates/utils-sync/src/once_lock.rs
```diff
@@ -1,81 +1,105 @@
...
+/// | no_std  | `once_cell::race::OnceBox<T>` |
+///
+/// # Why not `Clone`?
+///
+/// Under `std`, `OnceLock::clone` copies the initialized value.
+/// Under no_std, `OnceBox` does not support value extraction, so a "clone"
+/// would silently produce an empty instance — a violation of the `Clone`
+/// contract. `Clone` is therefore not implemented on either configuration.
+/// If you need a fresh instance, call `OnceLockCompat::new()` explicitly.
 pub struct OnceLockCompat<T> {
```
huitseeker, line 22: Could we add a small direct test for this wrapper? The PR changes the public contract here: `take()` becomes `reset()`, `Clone` goes away, and `std` no longer uses a type alias. `cargo test -p miden-utils-sync` only runs `racy_lock` tests right now, so the cross-feature behavior is mostly covered by compile checks rather than assertions.

huitseeker (approved):
Left one inline nit, but LGTM

crates/utils-sync/src/once_lock.rs
Lubov66: Added direct OnceLockCompat tests covering initialization, reset, and reinitialization after reset

plafer:
@Lubov66 please sign all commits

plafer (approved):
👍

Lubov66:
All commits are now signed

PR #3193, opened 2026-05-25
Complete core debug handler coverage

huitseeker:
- Enabled `miden-processor/std` from `miden-core-lib/std`, so default debug handlers actually print instead of silently no-oping.
- Added production-path coverage using `DefaultHost::with_library(&CoreLibrary::default())`.
- Added direct ranged `print_adv_stack` coverage.
- Hardened advice-stack range conversion against 32-bit usize wrapping.
- Fixed `write_stack(..., Some(0), ...)` underflow and added coverage.
- Let make lint apply the rustfmt wrap that CI was failing on.

### DIFF
--- a/crates/lib/core/Cargo.toml
+++ b/crates/lib/core/Cargo.toml
@@ -27,7 +27,7 @@ path = "tests/main.rs"
 
 [features]
 default = ["std"]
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

PR #3195, opened 2026-05-26
fix(fuzz): remove leading '-' from fuzz recipe lines 

huitseeker:
Individual targets (fuzz-mast-forest, fuzz-mast-validate, fuzz-mast-node-info, fuzz-mast-forest-wire-view): removed the leading '-' so that cargo-fuzz failures are properly propagated to make's exit code.

fuzz-all: replaced per-line '-' suppression with a shell-based failure collection pattern. Each fuzz command sets FAILED=1 on error, and the recipe exits with  at the end. This lets all targets run even if some fail, while still reporting overall failure.


bobbinth (approved):
Looks good! Thank you!

PR #3197, opened 2026-05-26
Remove unused core APIs

huitseeker:
Public APIs in the VM core that are not used, and not used downstream either (checked compiler, debug, node, client, protocol).


core/src/lib.rs
```diff
@@ -116,21 +116,6 @@ pub mod crypto {
...
-    {
-        let mut doc = Document::Empty;
-        for (i, item) in items.into_iter().enumerate() {
-            if i > 0 {
-                doc += const_text(", ");
-            }
-            doc += item.render();
-        }
-        doc
-    }
```
plafer, line 133: This module is now empty, we can just remove it

plafer (approved):
👍

core/src/lib.rs
huitseeker: The re-exports are still present! `PrettyPrint::render` is called across `core`/ and `crates/assembly-syntax/`.

bobbinth (approved):
Looks good! Thank you!

PR #3201, opened 2026-05-27
Remove debug decorators after core::debug

### LINKED ISSUE #3069: Remove `debug.*` ops/decorators

huitseeker:
PR #3185 added `miden::core::debug` event procedures. This PR removes the old `debug.*` decorator path that those procedures replace.

The removed MASM instructions are `debug.stack`, `debug.mem`, `debug.local`, and `debug.adv_stack`. They no longer parse, and the assembler no longer emits `Decorator::Debug`.

`Decorator::Debug` is removed from core. Old serialized debug decorator values are rejected by the MAST wire format. Trace decorators remain supported.

The processor no longer has a debug-decorator host callback. Old debug-mode decorator tests and examples are removed or moved to `core::debug` and trace-decorator coverage.

The CLI `run` and `prove `commands no longer expose --release as a debug-mode switch. `core::debug` calls are ordinary events, so that flag no longer had useful behavior there. `bundle --release` is unchanged.

Closes #3069.


processor/src/execution_options.rs
```diff
@@ -171,10 +171,8 @@ impl ExecutionOptions {
     /// Enables execution of programs in debug mode when the `enable_debugging` flag is set to true;
     /// otherwise, debug mode is disabled.
     ///
-    /// In debug mode the VM does the following:
-    /// - Executes `debug` instructions (these are ignored in regular mode).
-    /// - Records additional info about program execution (e.g., keeps track of stack state at every
-    ///   cycle of the VM) which enables stepping through the program forward and backward.
+    /// This flag is kept for backwards compatibility but no longer affects decorator execution
+    /// since the `Decorator::Debug` variant has been removed.
     pub fn with_debugging(mut self, enable_debugging: bool) -> Self {
```
bobbinth, line 176: Let's remember to remove this in subsequent PRs.

processor/src/host/debug.rs
```diff
@@ -51,39 +51,15 @@ impl<W: fmt::Write + Sync> DefaultDebugHandler<W> {
...
-            },
-        }
-        .map_err(DebugError::from)
+    fn on_trace(&mut self, _process: &ProcessorState, _trace_id: u32) -> Result<(), TraceError> {
+        Ok(())
     }
 }
 
+#[allow(dead_code)]
 impl<W: fmt::Write + Sync> DefaultDebugHandler<W> {
```
bobbinth: Any reason for keeping these around?

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -324,29 +324,27 @@ High-level constructs for controlling the execution flow.
...
 | `log_precompile`   | `[_, STMNT, _, ...]` | `[STATE_NEW, OUT_RATE1, OUT_CAP, ...]` | 1      | Folds a precomputed precompile statement into the rolling transcript. Reads the per-call statement word `STMNT` from `stack[4..8]` (the HPERM rate1 slots), applies the Poseidon2 permutation `Permute(STATE_PREV, STMNT, ZERO)` (with `STATE_PREV` supplied non-deterministically via helper registers), and writes the permutation output back via the identity lane→slot mapping (matching HPERM): `STATE_NEW` (= output `rate0`) at `stack[0..4]`, then `rate1` and `capacity` halves at `stack[4..8]` and `stack[8..12]`. Callers normally drop all three words immediately. See "Precompile flow" for initialization details. |
 
 ## Debugging Operations
 
-Instructions for inspecting VM state during execution. These do not affect VM state or program hash and are only active when the assembler is in debug mode.
+Procedures for inspecting VM state during execution. These are ordinary `core::debug` procedure calls that emit events, so adding them changes the program being executed. Procedures with stack inputs also change VM state by consuming those inputs.
```
bobbinth: Maybe we should add a comment that it is recommended not to include calls to debug procedures in production code (though, maybe that's obvious?).

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -324,29 +324,27 @@ High-level constructs for controlling the execution flow.
...
-  - `debug.local.I.M`: Prints local memory from index `I` to `M` (inclusive, `M >= I`, `0 <= I, M < 65536`).
-- **Cycles:** 0 (does not consume VM cycles).
+- **Procedures:**
+  - `print_stack`: Prints the entire operand stack. Inputs: `[...]`. Outputs: `[...]`. Cycles: 3.
+  - `print_mem`: Prints memory in the range `[start, end)` of the current context. Inputs: `[start, end, ...]`. Outputs: `[...]`. Cycles: 5.
+  - `print_mem_all`: Prints the full memory of the current context. Inputs: `[...]`. Outputs: `[...]`. Cycles: 3.
+  - `print_adv_stack`: Prints the advice stack in the range `[start, end)`. Inputs: `[start, end, ...]`. Outputs: `[...]`. Cycles: 5.
+  - `print_adv_stack_all`: Prints the full advice stack. Inputs: `[...]`. Outputs: `[...]`. Cycles: 7.
+  - `print_adv_map_all`: Prints the full advice map. Inputs: `[...]`. Outputs: `[...]`. Cycles: 3.
+  - `print_adv_map_item`: Looks up a WORD key in the advice map and prints the associated list of field elements. Inputs: `[KEY, ...]`. Outputs: `[...]`. Cycles: 7.
```
bobbinth, line 346: One thing I realized is that we no longer have a way to easily print out procedure locals. Printing out all locals may be difficult (mostly because non sure how to get the end-range), but having a procedure that is equivalent to `debug.local.I` should be pretty easy and I think we should add it (could be in a different PR).

bobbinth (approved):
Looks good! Thank you! I left a few comments inline - some of the can be left for future PRs.

core/src/operations/decorators/mod.rs
```diff
@@ -31,17 +28,13 @@ use crate::mast::{DecoratedOpLink, DecoratorFingerprint};
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
 pub enum Decorator {
-    /// Prints out information about the state of the VM based on the specified options. This
-    /// decorator is executed only in debug mode.
-    Debug(DebugOptions),
```
plafer, line 36: The `core/src/mast/debuginfo.mod.rs` module-level docstring is now stale, e.g.

- "All decorators (debug, trace, and assembly operation metadata)" line 6
- "Debug decorators are not executed" line 24
- etc

processor/src/execution_options.rs
plafer: Why not remove it in this PR? Or at least deprecate it (though I think we should just remove)

crates/lib/core/asm/debug.masm
```diff
@@ -26,7 +25,7 @@ const MAX_FELT_VALUE = 18446744069414584320
 
 #! Prints the entire operand stack.
 #!
-#! This is a drop-in replacement for the `debug.stack` decorator.
+#! Prints the entire operand stack.
```
plafer: Duplicate comment

processor/src/host/mod.rs
```diff
@@ -77,16 +76,6 @@ pub trait BaseHost {
...
 
-    /// Handles the debug request from the VM.
-    fn on_debug(
-        &mut self,
-        process: &ProcessorState,
-        options: &DebugOptions,
-    ) -> Result<(), DebugError> {
-        let mut handler = debug::DefaultDebugHandler::default();
-        handler.on_debug(process, options)
-    }
```
plafer, line 88: Should update the `BaseHost` docstring, for (3.), removing "debug events"

processor/src/execution_options.rs
huitseeker: Agreed. This PR leaves `enable_debugging` as a compatibility shim. I’ll remove the remaining API in a follow-up.

processor/src/host/debug.rs
huitseeker: The private `DefaultDebugHandler` print helpers are dead now. I’ll remove them and keep only the shared formatting helpers used by `core::debug`.

processor/src/host/handlers.rs
```diff
@@ -193,22 +184,11 @@ impl Debug for EventHandlerRegistry {
     }
 }
 
-// DEBUG HANDLER
+// TRACE HANDLER
 // ================================================================================================
 
-/// Handler for debug and trace operations
+/// Handler for trace operations
 pub trait DebugHandler: Sync {
```
```suggestion
pub trait TraceHandler: Sync {
```

Should also rename `DefaultHost::with_debug_handler()` accordingly

docs/src/user_docs/assembly/instruction_reference.md
huitseeker: I’ll add a short note that these procedures should be removed from production programs.

docs/src/user_docs/assembly/instruction_reference.md
huitseeker: Agreed. A `core::debug::print_local` procedure is the right replacement for `debug.local.I`. I’d keep that as a follow-up PR so this one stays focused on removing the old decorator path.

crates/assembly-syntax/src/parser/cst/instructions.rs
```diff
@@ -1878,6 +1749,7 @@ fn lower_error_msg(
 }
 
 /// Lowers a token that must represent a `u8` immediate or constant reference.
+#[allow(dead_code)]
 fn lower_u8_immediate(
```
plafer: Remove?

docs/src/user_docs/assembly/instruction_reference.md
huitseeker: #3203

plafer (approved):
LGTM

processor/src/execution_options.rs
huitseeker: I’d keep that as a follow-up because removing it changes `ExecutionOptions::new`, `FastProcessor::with_debugging`, and existing test setup beyond the decorator removal.

core/src/operations/decorators/mod.rs
huitseeker: I’ll update the module docs to describe trace decorators and debug metadata only.

processor/src/host/mod.rs
huitseeker: I’ll update it to refer only to trace events.

processor/src/host/handlers.rs
huitseeker: Agreed. The handler is trace-only now. I’ll rename the trait and the `DefaultHost` methods accordingly.

crates/assembly-syntax/src/parser/cst/instructions.rs
huitseeker: Yes. It was only kept after removing `debug.*`; I’ll delete it.

Issue #3198, opened 2026-05-26
release: immutable release procedure


lima-limon-inc:
If I'm not mistaken, the [0.23.2 release](https://github.com/0xMiden/miden-vm/releases/tag/v0.23.2) was not released with this new method, hence the lack of artifacts.

On https://github.com/0xMiden/miden-vm/pull/3029#discussion_r3228912559, @huitseeker mentioned that VM releases will eventually be made immutable. Thusly, the github workflow added on https://github.com/0xMiden/miden-vm/pull/3029 reflects this by only uploading the pre-built binaries on the release's *draft* page, not the "actual" release.

Because of a limitation on Github Action triggers, releases cannot be made using the standard "Releases" tab present on the right hand side of https://github.com/0xMiden/miden-vm. Instead, a Github action that creates the draft release needs to be used, which can be found on https://github.com/0xMiden/miden-vm/actions page.

Sadly, it's my understanding that as long as [Github lacks the ability to run an action on draft creation](https://github.com/orgs/community/discussions/7118), we cannot have immutable releases released via the standard "Release" page.

bobbinth:
Ah! I can make a v0.23.3 release and follow the procedure. IIUC, the correct process is:

- Prepare the `main` branch for the release (e.g., update changelog, update crate versions etc.) - but do not create release/tag.
- Then, in the "Actions" page, run the "Publish workspace to crates.io" workflow, and there, specify the new tag for the release (e.g., `v0.23.3`).

Is this correct?

lima-limon-inc:
> Ah! I can make a v0.23.3 release and follow the procedure. 

Thank you very much 😊 

> IIUC, the correct process is:
>     * Prepare the `main` branch for the release (e.g., update changelog, update crate versions etc.) - but do not create release/tag.
> 
>     * Then, in the "Actions" page, run the "Publish workspace to crates.io" workflow, and there, specify the new tag for the release (e.g., `v0.23.3`).
> 
> 
> Is this correct?

Indeed it should!
It's a bit of a shame that we have to go through this "alternative" path.

bobbinth:
I've updated the `main` branch to v0.23.3 - but when I try to run "Publish workspace to crates.io" from `main`, I see the following:

<img width="318" height="283" alt="Image" src="https://github.com/user-attachments/assets/1981d4c3-c51c-489b-83c2-74bd8c0931b8" />

Should we merge anything else into `main` to make it work?

[...]

PR #3204, opened 2026-05-27
feat: `workflow-publish` forward port

lima-limon-inc:
Related https://github.com/0xMiden/miden-vm/issues/3198

This PR forward ports the newly modified `workflow-publish.yml` to enable the creation of releases from the main branch.

### DIFF
--- a/scripts/generate-package.rs
+++ b/scripts/generate-package.rs
@@ -1,0 +1,38 @@
+#!/usr/bin/env -S cargo +nightly -Zscript
+
+---
+[dependencies]
+miden-assembly = { path = "../crates/assembly" }
+miden-package-registry = { path = "../crates/package-registry" }
+---
+
+use std::env;
+
+use miden_assembly::{Assembler, ProjectTargetSelector, Report, diagnostics::IntoDiagnostic};
+use miden_package_registry::InMemoryPackageRegistry;
+
+fn main() -> Result<(), Report> {
+    // We obtain the workspace root like so becaue CARGO_TARGET_DIR is not set
+    // for cargo scripts.
+    // This does mean that this script is only intended to be run from the
+    // project's rootk.
+    let workspace_root = env::current_dir().expect("could not read PWD");
+    let core_lib_project_dir = workspace_root.join("crates/lib/core/asm");
+
+    let target_dir = workspace_root.join("target");
+    let packages_dir = target_dir.join("packages");
+    std::fs::create_dir_all(&packages_dir).unwrap_or_else(|_| {
+        panic!("could not create packages/ directory in {}", packages_dir.display())
+    });
+
+    let assembler = Assembler::default();
+    let mut registry = InMemoryPackageRegistry::default();
+    let mut project_assembler = assembler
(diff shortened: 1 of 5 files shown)


scripts/generate-package.rs
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env -S cargo +nightly -Zscript
+
+---
```
huitseeker, line 3: The cargo script manifest starts with `---` instead of `---cargo`. The Makefile invokes this script via `cargo +nightly -Zscript`, which requires the `---cargo` header to locate the `[dependencies]` section. Without it, the script will fail immediately with a parse error before producing the `core.masp` artifact.

Compare with `scripts/check-masm-export-digests.rs` which uses `---cargo` on line 2 and correctly parses as a cargo script.

scripts/prepare-release.sh
```diff
@@ -0,0 +1,65 @@
...
+    # Canonical release path: operators provide only the tag. The commit is
+    # derived from origin/main so release assets, crates, and the GitHub
+    # release all point at the same protected revision.
+    release_sha="${main_sha}"
+
+    if git rev-parse --verify --quiet "refs/tags/${RELEASE_TAG}^{commit}" >/dev/null; then
+        tag_sha="$(git rev-parse "refs/tags/${RELEASE_TAG}^{commit}")"
+        if [[ "${tag_sha}" != "${release_sha}" ]]; then
+            echo "::error::Existing tag ${RELEASE_TAG} points at ${tag_sha}, expected ${release_sha}."
+            echo "::error::Tags are immutable under the release ruleset. Do not move or delete ${RELEASE_TAG}; choose a new tag after main points at the intended release commit."
```
huitseeker, line 21: `RELEASE_TAG` is interpolated directly into git/shell commands without validation. While `workflow_dispatch` requires repo write access, defense-in-depth suggests adding a semver check before the tag is used:

```bash
if ! [[ "${RELEASE_TAG}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
    echo "::error::Tag must match semver format vMAJOR.MINOR.PATCH"
    exit 1
fi
```

This would catch typos or malformed tags early rather than produce cryptic git errors later.

huitseeker (changes requested):
Is the lack of `---cargo` in the rust-script a thing allowed by nightly?

scripts/generate-package.rs
huitseeker: My bad, this does not need to be the case, except for highlighting

scripts/prepare-release.sh
huitseeker: This is a nit

huitseeker (approved):

bobbinth:
[v0.23.3](https://github.com/0xMiden/miden-vm/releases/tag/v0.23.3) release has been published!

scripts/generate-package.rs
lima-limon-inc: > My bad, this does not need to be the case, except for highlighting

For reference, my IDE (GNU Emacs 30.2) also struggles to properly highlight the `cargo` scripts. I think this will probably be the case until they are `stable`ized

PR #3205, opened 2026-05-27
feat(corelib): add core::debug::print_mem_addr

### LINKED ISSUE #3203: Create a core::debug::print_local
One thing I realized is that we no longer have a way to easily print out procedure locals. Printing out all locals may be difficult (mostly because non sure how to get the end-range), but having a procedure that is equivalent to `debug.local.I` should be pretty easy and I think we should add it (could be in a different PR).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/3201#discussion_r3312038578_
            
Follows #3185 #3201

partylikeits1983:
## Summary

Closes #3203.

Adds `miden::core::debug::print_mem_addr`, which prints the contents of a single memory cell given its address.

The issue originally asked for a `print_local` procedure, but a debug procedure can't resolve a caller's local index (it has no context about the caller's frame). The caller already has `locaddr.X` to turn a local index into an absolute address — what was missing was a clean way to print a *single* cell, since `print_mem` only takes a range `[start, end)`. So this adds `print_mem_addr`, and locals are covered transparently via:

```
locaddr.0 exec.debug::print_mem_addr
```

## Details

- **MASM** (`asm/debug.masm`): new `print_mem_addr` proc emitting a dedicated `miden::core::debug::print_mem_addr` event, consuming its single `addr` argument.
- **Handler** (`src/handlers/debug.rs`): new `PRINT_MEM_ADDR_EVENT_NAME`, registered in the default/noop/full handler sets (it's a non-sensitive memory printer, so it's in the default set). The dispatch arm validates the address via `get_mem_addr_range(1, 1)` (rejecting out-of-bounds the same way `print_mem` does) and prints the cell value — or an "uninitialized" message for an untouched address.
- **Tests** (`tests/debug.rs`): capture test, uninitialized-cell test, stack-effect test, and the default-registration assertion.
- Regenerated `docs/debug.md` and added a CHANGELOG entry.

Follows #3185 and #3201.

## Testing

- `cargo test -p miden-core-lib --test debug` — 19 passed (3 new).
- `cargo clippy -p miden-core-lib --all-targets` — clean.
- `cargo +nightly fmt` — clean.


partylikeits1983:
Superseded by #3207 — branch renamed to `ajl-print-mem-addr`, commit re-signed, and rebased onto latest `next`.

PR #3207, opened 2026-05-27
feat(corelib): add `core::debug::print_mem_addr`

### LINKED ISSUE #3203: Create a core::debug::print_local

partylikeits1983:
## Summary

Closes #3203.

Adds `miden::core::debug::print_mem_addr`, which prints a single memory cell given its address. Combine with `locaddr` to print a procedure local:

```
locaddr.0 exec.debug::print_mem_addr
```

This covers the `print_local` request from #3203: a debug procedure can't resolve a caller's local index, but `locaddr` already turns a local into an absolute address — what was missing was a way to print one cell (`print_mem` only takes a `[start, end)` range).

## Changes

- `asm/debug.masm`: new `print_mem_addr` proc emitting a dedicated event.
- `src/handlers/debug.rs`: handler registered in the default/noop/full sets; validates the address like `print_mem` and prints the cell (or "uninitialized").
- Tests, regenerated `docs/debug.md`, and a CHANGELOG entry.

Follows #3185 and #3201. Supersedes #3205 (branch renamed to `ajl-print-mem-addr`).


crates/lib/core/asm/debug.masm
```diff
@@ -50,6 +51,22 @@ pub proc print_mem
...
+#!
+#! Where:
+#! - addr is the address of the memory cell to print.
+#!
+#! Cycles: 4
+pub proc print_mem_addr
+    emit.PRINT_MEM_ADDR_EVENT
+    # => [addr, ...]
+    drop
+end
```
bobbinth, line 76: Could this not be just:
```
pub proc print_mem_addr
    dup add.1 swap
    emit.PRINT_MEM_EVENT
    drop drop
end
```

This way, we won't need to introduce another envent.

crates/lib/core/asm/debug.masm
bobbinth, line 23: As mentioned in the above comment, I don't think `PRINT_MEM_ADDR_EVENT` is needed.

I also think we should get rid of the `PRINT_MEM_ALL_EVENT` (we should be able to use `PRINT_MEM_EVENT` in the `print_mem_all` procedure).

Lastly, we can probably rename `PRINT_ADV_MAP_ALL_EVENT` into just `PRINT_ADV_MAP_EVENT`.

bobbinth (commented):
Looks good! Thank you! I left a couple of comments inline.

crates/lib/core/asm/debug.masm
partylikeits1983: Done — `print_mem_addr` now reuses `PRINT_MEM_EVENT` via `dup add.1 swap`; the dedicated event was dropped.

crates/lib/core/asm/debug.masm
partylikeits1983: Done — removed `PRINT_MEM_ADDR_EVENT` and `PRINT_MEM_ALL_EVENT` (both now reuse `PRINT_MEM_EVENT`), and renamed `PRINT_ADV_MAP_ALL_EVENT` → `PRINT_ADV_MAP_EVENT`.

bobbinth (approved):
Looks good! Thank you!

crates/lib/core/asm/debug.masm
```diff
@@ -50,14 +49,34 @@ pub proc print_mem
...
+#!
+#! Inputs:  [addr, ...]
+#! Outputs: [...]
+#!
+#! Where:
+#! - addr is the address of the memory cell to print.
+#!
+#! Cycles: 6
+pub proc print_mem_addr
+    dup add.1 swap
```
huitseeker, line 72: Nit: This misses the last valid memory address. If `addr == u32::MAX`, `add.1` makes the exclusive end larger than `u32::MAX`, and `get_mem_addr_range` rejects that end before the handler can print the cell. The old dedicated handler avoided this edge case.

crates/lib/core/asm/debug.masm
```diff
@@ -50,14 +49,34 @@ pub proc print_mem
...
 #! Prints the full memory of the current context.
 #!
 #! Inputs:  [...]
 #! Outputs: [...]
 #!
-#! Cycles: 3
+#! Cycles: 7
 pub proc print_mem_all
-    emit.PRINT_MEM_ALL_EVENT
+    push.4294967295 push.0
```
huitseeker: Nit: This range is `[0, u32::MAX)`, so it excludes an initialized cell at address `u32::MAX`. The old full-memory event printed the whole initialized memory map, including that last address. We may need to keep a distinct full-memory path or special-case this boundary.

huitseeker (commented):
LGTM, apart for the edge case — how would we want to address? 🤔

crates/lib/core/asm/debug.masm
partylikeits1983: Great catch! At `u32::MAX` the `add.1` pushes the exclusive end to 2^32, and `get_mem_addr_range` `processor/src/lib.rs:235-237` errors on anything past `u32::MAX`.

The fix is to change `get_mem_addr_range` to allow an end up to 2^32 and return `Range<u64>` then `print_mem_addr` will work as-is.

And having `print_mem_all` push 2^32 instead of `u32::MAX` so it picks up the last cell fixes `print_mem_all`

partylikeits1983:
thank you @huitseeker for catching [this](https://github.com/0xMiden/miden-vm/pull/3207#discussion_r3328957855)!

Went ahead and implemented [this](https://github.com/0xMiden/miden-vm/pull/3207#discussion_r3329041701). Fixes not being able to print the memory cell at `u32::MAX`.

`get_mem_addr_range` now allows the exclusive end up to 2^32 and returns `Range<u64>` so `print_mem_addr` works at `u32::MAX` with no MASM change. 

`print_mem_all` pushes 2^32 instead of `u32::MAX` to pick up that last
cell. Added regression tests for both.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -201,11 +181,19 @@ fn slice_range(slice: &[Felt], start: usize, end: usize) -> &[Felt] {
...
+/// than the entire address space.
+///
+/// The range is half-open `[start, end)` with `end` up to `2^32`, hence `Range<u64>`, so that the
+/// cell at `u32::MAX` can be addressed.
 fn write_mem_range<W: fmt::Write>(
     w: &mut W,
     process: &ProcessorState,
-    range: core::ops::Range<u32>,
+    range: core::ops::Range<u64>,
 ) -> fmt::Result {
```
bobbinth, line 255: I'm not too sure about this change. Technically, the range of valid addresses here should be limited to `u32` values. I think the issue is that we are using `Range` which is exclusive of the last value. Maybe a better solution would be to use `RangeInclusive`? This way, we properly constrain the valid addresspace.

bitwalker: I would change the type of range to a new type parameter `R`, bounded as `R: core::ops::RangeBounds<u32>`. That allows range syntax with both inclusive/exclusive ranges to be used

crates/lib/core/src/handlers/debug.rs
```diff
@@ -201,43 +181,78 @@ fn slice_range(slice: &[Felt], start: usize, end: usize) -> &[Felt] {
...
 
-/// Prints memory in the range `[start, end)` of the current context.
+/// Reads a half-open `[start, end)` memory range off the operand stack (positions `start_idx`,
+/// `end_idx`) and converts it to an inclusive `start..=end-1` range of valid `u32` addresses.
+///
+/// Unlike [`ProcessorState::get_mem_addr_range`], the exclusive end may be up to `2^32` so the
+/// cell at `u32::MAX` can be printed (the range `[addr, addr+1)` for `addr == u32::MAX`, or the
+/// full range `[0, 2^32)`). The inclusive representation keeps both bounds within `u32`, so no
+/// address value ever exceeds `u32::MAX`.
+fn read_mem_print_range(
```
huitseeker, line 218: The next branch had a 1024-cell limit on print ranges (enforced in the handler). That limit and the associated test (`print_mem_rejects_oversized_range`) are gone in this PR. Was removing the limit intentional? The old test would catch a caller accidentally passing a huge range.

partylikeits1983: Wasn't intentional. It got dropped along with the 1024-cell limit when I consolidated the events. Brought both back: the limit now exempts that full range, and `print_mem_rejects_oversized_range` is passing again.

crates/lib/core/src/handlers/debug.rs
partylikeits1983: Implemented @bitwalker's suggestion of a new type parameter `R: RangeBounds<u32>` on `write_mem_range`, so the address space stays constrained to `u32` while still accepting both inclusive and exclusive ranges.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -201,43 +204,92 @@ fn slice_range(slice: &[Felt], start: usize, end: usize) -> &[Felt] {
...
+where
+    W: fmt::Write,
+    R: RangeBounds<u32>,
+{
     let (ctx, clk) = (process.ctx(), process.clock());
-    writeln!(w, "Memory state before step {clk} for context {ctx}:")?;
+    let Some((start, end)) = resolve_addr_bounds(&range) else {
+        return writeln!(w, "Memory state before step {clk} for context {ctx}: range is empty.");
+    };
     let items: Vec<_> = process
```
huitseeker, line 277: Could we keep explicit `print_mem` enumerating every address in the requested range? With:

```masm
push.42 push.100 mem_store
push.110 push.100
exec.debug::print_mem
```

this prints only `0x64..0x67`, even though the header says `[100, 109]`. On `next`, `write_mem_range` iterated the requested range and `write_interval` showed missing cells as `EMPTY`, so addresses 104..109 did not disappear.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -149,24 +148,28 @@ impl<W: fmt::Write + Send + Sync + 'static> EventHandler for DebugPrinter<W> {
...
-            let len = range.end - range.start;
-            if len > MAX_PRINT_MEM_RANGE {
-                return Err(format!(
-                    "print_mem range length {len} exceeds maximum of {MAX_PRINT_MEM_RANGE}"
-                )
-                .into());
+            let range = read_mem_print_range(process, 1, 2)?;
+            // Guard against an accidentally huge explicit range; the full `[0, 2^32)` range used by
+            // `print_mem_all` is exempt.
+            if let Some((first, last)) = resolve_addr_bounds(&range)
```
huitseeker: Could this exemption be limited to `print_mem_all`? A program can call the shared event path directly:

```masm
push.4294967296 push.0
exec.debug::print_mem
```

That succeeds and prints the full `[0, u32::MAX]` range, so ordinary `print_mem` can skip the documented 1024-address cap. A separate full-memory event or a marker for `print_mem_all` would keep the cap true for user-supplied ranges.

crates/lib/core/src/handlers/debug.rs
partylikeits1983: Fixed in 78f56b2. print_mem enumerates every address in the requested range again and shows uninitialized cells as EMPTY, same as on next. Added a test for your exact repro.

crates/lib/core/src/handlers/debug.rs
partylikeits1983: Good point. Brought back a dedicated print_mem_all event in 78f56b2, so the 1024 address cap now applies to every print_mem range with no exemption. Added a test that the full range is rejected. The shared event turned out to lose real semantics here: print_mem enumerates a capped range while print_mem_all lists only initialized cells.

crates/lib/core/src/handlers/debug.rs
```diff
@@ -149,13 +157,16 @@ impl<W: fmt::Write + Send + Sync + 'static> EventHandler for DebugPrinter<W> {
...
             write_stack(w, operand_stack, None, "Stack", process.clock())?;
         } else if id == PRINT_MEM_EVENT_NAME.to_event_id() {
-            let range = process.get_mem_addr_range(1, 2)?;
-            let len = range.end - range.start;
-            if len > MAX_PRINT_MEM_RANGE {
-                return Err(format!(
-                    "print_mem range length {len} exceeds maximum of {MAX_PRINT_MEM_RANGE}"
-                )
-                .into());
+            let range = read_mem_print_range(process, 1, 2)?;
```
huitseeker: Nit: Could this helper return concrete bounds after it parses the stack values? `read_mem_print_range` already turns the half-open stack range into either an empty range or inclusive `u32` bounds, so returning something like `Option<(u32, u32)>` would let the cap check and `write_mem_range` share the same value. 

That keeps the `u32::MAX` case, but avoids resolving `RangeBounds` twice and encoding empty as included/excluded bounds on the same address.

huitseeker (commented):
This LGTM, but I think there's a possible simpler concrete bounds shape instead of resolving RangeBounds twice.

huitseeker (approved):

crates/lib/core/src/handlers/debug.rs
partylikeits1983: Good idea, will look into this

partylikeits1983: thank you for this suggestion, I updated `read_mem_print_range` to now return `Option<(u32, u32)>` for inclusive bounds and `None` for an empty range.

PR #3208, opened 2026-05-27
Remove trace decorators and decorator scaffolding

### LINKED ISSUE #2411: remove Owned variant of DecoratorStorage

### LINKED ISSUE #3192: Remove `trace` decorator
Maintaining the current set of decorators adds quite a bit of complexity in both the assembler (`MastForestBuilder` and related structs) and in the processor, and makes the PRs like #3142 more complicated than they need to be. There are currently two types of decorators: `Debug` and `Trace` - we should remove both.

Once #3069 is addressed, we'll have removed the `Debug` decorators. For this, we need to merge #3185 first.

The next step, is to remove `Trace` decorators - and in fact, we can do this in parallel with #3185. AFAIK, removing the `Trace` decorator should be pretty straight forward - there are only two places where we use them downstream:

1. In the complier for `println`-like functionality. This can easily be migrated to using `Event` operations instead with no real change in functionality. cc @mooori and @bitwalker.
2. In the debugger for keeping track of call stack. This too we should be able to migrate to using `Event` operations instead - though, there would be a slight change in functionality as `Event` would remain a part of the MAST. cc @bitwalker in case I'm off here.

For the second point, I think this would be a temporary change because once we have proper source location tracking implemented (resulting from addressing https://github.com/0xMiden/miden-vm/issues/3040), the debugger will use that functionality rather than rely on `Trace`/`Event`-based approach. And overall, using `Event` operations is pretty cheap and shouldn't cause any performance issues in the interim.

huitseeker:
Closes #3192. Closes #2411.

Removes `trace` decorators and the remaining decorator infrastructure after `debug.*` decorators were removed.

MAST now keeps debug metadata as `AssemblyOp`, `DebugVarInfo`, error codes, and procedure names. These records are not executable.

This removes:

- `trace.*` syntax and parser support
- `Decorator` and decorator IDs, lists, and stores
- node before/after hooks and operation-indexed decorator plumbing
- processor decorator continuations, hooks, host callbacks, CLI flags, and tests
- decorator remapping and fingerprinting in builders, merging, sparse forests, packages, and serialization
- unreleased decorator wire slots from MAST format `[0, 0, 4]`

Compatibility cutoff: old payloads with debug or trace decorators are not supported. Since `[0, 0, 4]` has not shipped, this updates its format notes instead of bumping the version again.

Follow-up: debug/println-style work from #3192 should use event-based plumbing, not decorators.


core/src/mast/debuginfo/mod.rs
```diff
@@ -3,37 +3,18 @@
...
-//! - Node-level decorator storage (before_enter/after_exit)
+//! - Assembly operation source mappings
+//! - Debug variable locations
 //! - Error code mappings for descriptive error messages
+//! - Procedure names
 //!
-//! The debug info is always available at the `MastForest` level (as per issue 1821), but may be
-//! conditionally included during assembly to maintain backward compatibility. Trace decorators are
-//! only executed when tracing is enabled. Assembly operation metadata remains available for
+//! Executable decorators have been removed. Assembly operation metadata remains available for
```
bobbinth: Do we need this statement?

core/src/mast/debuginfo/debug_var_storage.rs
```diff
@@ -15,19 +14,34 @@ use proptest::prelude::*;
...
+    /// Node index is invalid (either out of sequence or already added).
+    #[error("invalid node index {0:?}")]
+    NodeIndex(MastNodeId),
+    /// Operation index is invalid for the node.
+    #[error("invalid operation index {operation} for node {node:?}")]
+    OperationIndex { node: MastNodeId, operation: usize },
+    /// Internal CSR structure is invalid.
+    #[error("internal CSR structure is invalid")]
+    InternalStructure,
+}
```
bobbinth: Should this struct be one module higher (i.e., in `src/mast/debuginfo/mod.rs`)?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -98,61 +95,21 @@ impl BasicBlockNode {
...
-    ///
-    /// Returns an error if:
-    /// - `operations` vector is empty.
+    /// Returns a new [`BasicBlockNode`] instantiated with the specified operations.
     #[cfg(any(test, feature = "arbitrary"))]
-    pub(crate) fn new_owned_with_decorators(
-        operations: Vec<Operation>,
-        decorators: DecoratorList,
-    ) -> Result<Self, MastForestError> {
+    pub(crate) fn new_owned(operations: Vec<Operation>) -> Result<Self, MastForestError> {
```
bobbinth: Should this be just `new()`?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -377,81 +184,16 @@ impl BasicBlockNode {
...
     /// - Operations (exact equality)
-    /// - Before-enter decorators (by ID)
-    /// - After-exit decorators (by ID)
-    /// - Operation-indexed decorators (by iterating and comparing their contents)
-    ///
-    /// Unlike the derived PartialEq, this method works correctly with both owned and linked
-    /// decorator storage by accessing the actual decorator data from the forest when needed.
     #[cfg(test)]
-    pub fn semantic_eq(&self, other: &BasicBlockNode, forest: &MastForest) -> bool {
+    pub fn semantic_eq(&self, other: &BasicBlockNode, _forest: &MastForest) -> bool {
```
bobbinth: Should we remove the `_forest` parameter from here?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -789,60 +507,23 @@ impl MastNodeExt for BasicBlockNode {
...
-        BasicBlockNodeBuilder::from_op_batches(self.op_batches, padded_decorators, self.digest)
-            .with_before_enter(before_enter)
-            .with_after_exit(after_exit)
+    fn to_builder(self, _forest: &MastForest) -> Self::Builder {
+        // Use from_op_batches to avoid re-batching existing operation batches.
+        BasicBlockNodeBuilder::from_op_batches(self.op_batches, self.digest)
     }
 
     #[cfg(debug_assertions)]
     fn verify_node_in_forest<F>(&self, forest: &F)
```
bobbinth, line 506: Doesn't seem like this method is doing anything any more - is this right?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -789,60 +507,23 @@ impl MastNodeExt for BasicBlockNode {
...
-        }
+        let _ = forest;
     }
 }
 
 struct BasicBlockNodePrettyPrint<'a> {
     block_node: &'a BasicBlockNode,
-    mast_forest: &'a MastForest,
+    _mast_forest: &'a MastForest,
 }
```
bobbinth: We probably don't need this struct any more as we should be able to implement `PrettyPrint` directly on `BasicBlockNode`.

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1601,130 +870,62 @@ impl BasicBlockNodeBuilder {
...
 
         Ok(node_id)
     }
 
     fn fingerprint_for_node(
         &self,
-        forest: &MastForest,
+        _forest: &MastForest,
         _hash_by_node_id: &impl LookupByIdx<MastNodeId, MastNodeFingerprint>,
     ) -> Result<MastNodeFingerprint, MastForestError> {
```
bobbinth: The concept of `MastNodeFingerprint` now seems to be the same as node hash. If so, two questions:

- Does this cause any issues for `AsmOp` and `DebugVar` assignments (i.e., mapping them to nodes). I'm assuming not because it doesn't seem like these were covered by the fingerprint before.
- Could we get rid of `MastNodeFingerprint` entirely? If yes, probably best to do in a follow-up PR.

core/src/mast/node/dyn_node.rs
```diff
@@ -81,84 +75,29 @@ impl DynNode {
...
-    }
-
-    fn multi_line_pre_decorators(&self) -> Document {
-        self.concatenate_decorators(self.node.before_enter(self.mast_forest), Document::Empty, nl())
-    }
-
-    fn multi_line_post_decorators(&self) -> Document {
-        self.concatenate_decorators(self.node.after_exit(self.mast_forest), nl(), Document::Empty)
-    }
+    _mast_forest: &'a MastForest,
```
bobbinth: It seems like we may not need this struct any more.

core/src/mast/node/dyn_node.rs
```diff
@@ -495,46 +316,23 @@ impl proptest::prelude::Arbitrary for DynNodeBuilder {
...
-impl Default for DynNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct DynNodeBuilderParams {}
```
bobbinth: Do we still need this struct?

core/src/mast/node/external.rs
```diff
@@ -34,90 +31,34 @@ use crate::{
...
-    }
-
-    fn multi_line_pre_decorators(&self) -> Document {
-        self.concatenate_decorators(self.node.before_enter(self.mast_forest), Document::Empty, nl())
-    }
-
-    fn multi_line_post_decorators(&self) -> Document {
-        self.concatenate_decorators(self.node.after_exit(self.mast_forest), nl(), Document::Empty)
-    }
+    _mast_forest: &'a MastForest,
```
bobbinth: Similar as some of the above comments - we may not need this struct any more (probably applies to other node types as well).

crates/assembly/src/snapshots/miden_assembly__tests__decorators_basic_block.snap
```diff
@@ -8,11 +8,11 @@ begin
...
         drop
-        trace(0)
+        debug(stack.0)
         add
-        trace(1)
+        debug(stack.1)
         mul
         noop
-        trace(2)
+        debug(stack.2)
```
bobbinth: Is this right? I thought `debug` instructions are also gone?

crates/assembly/src/snapshots/miden_assembly__tests__decorators_call.snap
```diff
@@ -10,10 +10,10 @@ begin
             mstore
             drop
             noop
-            trace(0)
-            trace(1)
+            debug(stack.0)
+            debug(stack.1)
```
bobbinth: Same comment as above.

crates/assembly/src/snapshots/miden_assembly__tests__decorators_dyn-2.snap
```diff
@@ -10,13 +10,13 @@ begin
...
-            trace(0)
-            trace(1)
-            trace(2)
-            trace(3)
-            trace(4)
+            debug(stack.0)
+            debug(stack.1)
+            debug(stack.2)
+            debug(stack.3)
+            debug(stack.4)
```
bobbinth: Same comment as above (and probably relevant for other places too).

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -1577,80 +926,6 @@ mod tests {
 
     /// Non-block nodes with different AssemblyOps alias during assembly.
     #[test]
-    fn test_non_block_nodes_dedup_different_asm_ops() {
```
bobbinth, line 1580: Should this test stay? Seems like it was testing `AssmblyOps` rather than decorators.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -1283,223 +873,6 @@ mod tests {
...
-        assert_eq!(cloned_vars[0].1, var_id);
-
-        let original_vars = forest.debug_info().debug_vars_for_node(block_id);
-        assert_eq!(original_vars.len(), 1, "original should keep its debug vars");
-        assert_eq!(original_vars[0].1, var_id);
-    }
-
-    /// Same-ops blocks with different debug vars still alias. The first node's metadata wins.
-    #[test]
-    fn test_ensure_node_preserving_debug_vars_dedups_different_debug_vars() {
```
bobbinth, line 1439: Same comment as above - should this test stay?

crates/project/examples/protocol/kernel/kernel/io.masm
bobbinth, line 1: Should we remove this file? Any reason for keeping it?

processor/src/fast/mod.rs
```diff
@@ -247,21 +236,12 @@ impl FastProcessor {
 
     /// Enables or disables debugging mode.
     ///
-    /// This is kept for backwards compatibility but no longer affects decorator execution
-    /// since the `Decorator::Debug` variant has been removed.
+    /// This is kept for backwards compatibility but no longer affects runtime execution.
     pub fn with_debugging(mut self, enabled: bool) -> Self {
```
bobbinth: What does `debug` mode enable now?

processor/src/trace/parallel/mod.rs
```diff
@@ -785,15 +785,13 @@ fn translate_snapshot_continuation_stack(
             },
             Continuation::FinishCall(id) => Continuation::FinishCall(id),
             Continuation::FinishDyn(id) => Continuation::FinishDyn(id),
-            Continuation::FinishExternal(id) => Continuation::FinishExternal(id),
```
bobbinth, line 788: Why `FinishExternal` is no longer needed here?

processor/src/execution_options.rs
```diff
@@ -162,17 +158,10 @@ impl ExecutionOptions {
...
-        self
-    }
-
     /// Enables execution of programs in debug mode when the `enable_debugging` flag is set to true;
     /// otherwise, debug mode is disabled.
     ///
-    /// This flag is kept for backwards compatibility but no longer affects decorator execution
-    /// since the `Decorator::Debug` variant has been removed.
+    /// This flag is kept for backwards compatibility but no longer affects runtime execution.
     pub fn with_debugging(mut self, enable_debugging: bool) -> Self {
```
bobbinth: We should remove this in a follow-up PR.

bobbinth (commented):
Not a super deep review - but looks great! Thank you!

I left some comments inline - these mostly fall into the following categories:

- There is quite a bit of additional clean-up that I think we can do - but given the size of this PR, I think we should defer this to a follow-up PR.
- `trace` statements seem to have been replaced with `debug` statements in some printed-out MAST files. I wasn't sure why that could happen.
- It seems like some tests may have been removed too early (e.g., the ones that test `AsmOp` or `DebugVar` deduplication).

crates/assembly/src/mast_forest_builder.rs
huitseeker: The exact old `test_ensure_node_preserving_debug_vars_dedups_different_debug_vars` test depended on the removed decorator clone path, so I did not restore it verbatim. I added replacement coverage in `test_ensure_block_dedups_different_debug_vars`, which keeps the same metadata identity invariant: debug vars do not prevent block dedup, and the first node's metadata wins.

core/src/mast/debuginfo/debug_var_storage.rs
huitseeker: I moved `DebugInfoIndexError` up to `mast::debuginfo`

core/src/mast/node/basic_block_node/mod.rs
huitseeker: Yes. I renamed the test/arbitrary constructor to `new()`.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: removed the unused `forest` parameter and updated the call sites.

crates/assembly/src/snapshots/miden_assembly__tests__decorators_basic_block.snap
huitseeker: These were stale orphaned snapshots from the old decorator tests; `debug.*` is gone too. I deleted the `decorators_*` snapshots rather than keeping snapshots that imply a `trace` to `debug` rewrite.

crates/assembly/src/snapshots/miden_assembly__tests__decorators_call.snap
huitseeker: Same: removed

crates/assembly/src/snapshots/miden_assembly__tests__decorators_dyn-2.snap
huitseeker: same: removed

crates/assembly/src/mast_forest_builder.rs
huitseeker: I restored the non-block-node `AssemblyOp` dedup test without decorator dependencies.

crates/project/examples/protocol/kernel/kernel/io.masm
huitseeker: I removed the no-op protocol `io.masm` fixture/example syscall and reduced the example syscall table accordingly.

core/src/mast/node/dyn_node.rs
huitseeker: I removed the empty `DynNodeBuilderParams` struct and use `()` as the `Arbitrary::Parameters` type now.

processor/src/trace/parallel/mod.rs
huitseeker: This is a consequence of removing decorator execution. `FinishExternal` existed only to run `after_exit` decorators after an external node finished; the actual forest switch/return path is still represented by `EnterForest(old_forest)` plus `StartNode(resolved_node_id)`. Once decorators are gone, keeping a finish continuation for `External` would be a no-op variant with no VM semantics.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: For basic blocks there are no child IDs to validate against the forest after decorator removal, so the implementation is intentionally empty while the trait hook remains for child-bearing node types.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: I collapsed `MastNodeFingerprint` into the node digest, removed the metadata-root wrapper, and added coverage documenting that assertion error codes are metadata and do not affect MAST node identity.

processor/src/fast/mod.rs
huitseeker: I removed the remaining `ExecutionOptions` / `FastProcessor` debugging API

processor/src/execution_options.rs
huitseeker: I removed `ExecutionOptions::with_debugging`, `enable_debugging`, `FastProcessor::with_debugging`, and the extra constructor parameter, then updated the call sites.

core/src/mast/node/external.rs
huitseeker: removed the now-redundant `ExternalNode` pretty-print wrapper and made the node implement `PrettyPrint` / `Display` directly.

core/src/mast/node/dyn_node.rs
huitseeker: removed the now-redundant `DynNode` pretty-print wrapper and made the node implement `PrettyPrint` / `Display` directly.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: cleaned up the wrappers that had become pure passthroughs after decorator removal. The wrappers that still carry `MastForest` context for child lookup remain because they still do real work.

core/src/mast/debuginfo/mod.rs
huitseeker: Removed

crates/assembly/tests/fixtures/protocol/kernel/kernel/io.masm
bitwalker, line 1: I'm not really sure I understand why this fixture was changed - the `println` procedure here is illustrative, and doesn't conflict with anything, or use the `trace` infra being removed.

crates/mast-package/src/package/manifest.rs
```diff
@@ -311,17 +311,15 @@ pub struct ProcedureExport {
     /// export corresponds to, when multiple exports may have the same digest (conversely, some
     /// procedure roots in the MAST may not be associated with any exports).
     ///
-    /// Provenance is important because debug info/decorators may differ between two logically
-    /// distinct procedures, even if they ultimately compile to the same MAST. The MAST will use
-    /// decorators to "fingerprint" nodes that have the same digest, but differ in their
-    /// decorators. This fingerprint is then used to keep all distinct nodes in the forest. The
-    /// only way to guarantee that you will get the precise MAST node that corresponds to the
-    /// specific procedure you've named, is to use the MAST node, rather than the digest.
+    /// Provenance is important because source/debug metadata may differ between two logically
```
bitwalker: You've removed the explanation of fingerprinting here, but then leave mention of it later in the comment. I would simply modify the fingerprinting explanation that was here to remove mention of decorators - the rest of it remains accurate AFAIK

crates/project/examples/protocol/kernel/kernel/io.masm
bitwalker, line 1: Same note here about this example as I had on the equivalent test fixture

crates/test-utils/src/lib.rs
```diff
@@ -580,45 +562,11 @@ impl Test {
...
 
         processor.execute_sync(&program, &mut host).map(|output| (output, host))
     }
 
-    /// Compiles the test's source to a Program and executes it with the tests inputs. Returns
-    /// the [`StackOutputs`] and a [`String`] containing all trace output.
-    ///
-    /// If the execution fails, the output is printed `stderr`.
-    #[cfg(not(target_family = "wasm"))]
-    pub fn execute_with_trace_buffer(&self) -> Result<(StackOutputs, String), ExecutionError> {
```
bitwalker, line 594: We probably still need this, but oriented towards capturing event output instead no? If all uses of `trace` need to use `emit`, I would assume we would still have some tests that need equivalent capturing functionality.

If not, and we don't need it for the new `core::debug` procedures, then disregard

huitseeker: I think we can leave this removed. `execute_with_trace_buffer` was specifically tied to the old trace-handler output path, which no longer exists  after removing trace decorators. Event output is handler-specific rather than a VM-global output stream.

For the new `miden::core::debug` procedures, the capture coverage now lives in `crates/lib/core/tests/debug.rs`, where the tests register `DebugPrinter` with an in-memory writer. `Test` still exposes `with_event_handler(s)` for tests that need equivalent custom event capture.

crates/mast-package/src/package/manifest.rs
huitseeker: rewritten

crates/project/examples/protocol/kernel/kernel/io.masm
huitseeker: reinstated

crates/assembly/tests/fixtures/protocol/kernel/kernel/io.masm
huitseeker: reinstated

processor/src/fast/tests/advice_provider.rs
```diff
@@ -7,7 +7,7 @@ use crate::test_utils::TestHost;
 fn test_advice_provider() {
     let kernel_source = "
         pub proc foo
-            push.2323 mem_store.100 trace.11
+            push.2323 mem_store.100
```
plafer: The value of this test was using trace events to check that the memory state (and other things in the "processor snapshots") were the same at all trace events. In its current form though, the test no longer checks that.

*Ideally* we replace the `trace` calls with whatever the replacement (a std assembler procedure that maps to an event I assume?) to generate the same snapshot, without removing the snapshot file. Would this easily be doable in this PR (I haven't looked at it all at this point)?

plafer (commented):
No more decorators; am I dreaming? The -17k diff shows just how much complexity is being removed here 👍.

So far, reviewed the processor code, which LGTM

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -640,9 +573,7 @@ impl MastForestBuilder {
 
     /// Adds a node to the forest, and returns the [`MastNodeId`] associated with it.
     ///
-    /// Note that only one copy of nodes that have the same MAST root and decorators is added to the
-    /// MAST forest; two nodes that have the same MAST root and decorators will have the same
-    /// [`MastNodeId`].
+    /// Note that only one copy of nodes with the same fingerprint is added to the MAST forest.
```
plafer: nit: fingerprints no longer exist

crates/project/examples/protocol/userspace/mod.masm
```diff
@@ -2,8 +2,3 @@ pub proc panic
     push.1
     syscall.system_call
 end
-
-pub proc println(message: ptr<u8, addrspace(byte)>)
-    push.2
-    syscall.system_call
-end
```
bobbinth, line 9: Why do we need to remove this?

crates/assembly/tests/fixtures/protocol/userspace/mod.masm
```diff
@@ -2,8 +2,3 @@ pub proc panic
     push.1
     syscall.system_call
 end
-
-pub proc println(message: ptr<u8, addrspace(byte)>)
-    push.2
-    syscall.system_call
-end
```
bobbinth, line 9: Same question as in the last comment: why did we need to remove this?

core/src/mast/node/basic_block_node/mod.rs
```diff
@@ -1601,194 +853,65 @@ impl BasicBlockNodeBuilder {
...
-    ) -> Result<MastNodeFingerprint, MastForestError> {
-        // For BasicBlockNode, we need to implement custom logic because BasicBlock has special
-        // decorator handling with operation indices that other nodes don't have
-
-        // Process based on operation data type
-        let (op_batches, digest, raw_decorators) = match &self.operation_data {
-            OperationData::Raw { operations, decorators } => {
+        _forest: &MastForest,
+        _hash_by_node_id: &impl LookupByIdx<MastNodeId, Word>,
+    ) -> Result<Word, MastForestError> {
```
bobbinth: We may be able to get rid of the `hash_by_node_id` parameter. Upon a quick look, I didn't see it used anywhere (here or in other nodes).

processor/src/fast/tests/advice_provider.rs
huitseeker: Fixed in `54a68802df`. I restored the checkpoint semantics with stack-neutral events: the MASM now uses `push.<id> emit drop` at the old checkpoint sites, and `TestHost` records `ProcessorStateSnapshot`s keyed by event id.

One detail: the host observes the synthetic event id on top of the stack during `emit`, so the snapshot helper skips that item when reconstructing the old processor snapshot shape. The restored `event_checkpoints` insta snapshot now covers stack, memory, advice stack, and advice map state at those event checkpoints

crates/assembly/src/mast_forest_builder.rs
huitseeker: I updated the stale comment and local builder names to use digest terminology. The only remaining `fingerprint` spelling in this file is the existing trait method call, which I left alone because renaming that broader API is outside this small review fix.

core/src/mast/node/loop_node.rs
```diff
@@ -449,38 +281,12 @@ impl proptest::prelude::Arbitrary for LoopNodeBuilder {
...
-impl Default for LoopNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct LoopNodeBuilderParams {}
```
bobbinth: Is this still needed?

core/src/mast/node/split_node.rs
```diff
@@ -475,38 +323,12 @@ impl proptest::prelude::Arbitrary for SplitNodeBuilder {
...
-impl Default for SplitNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct SplitNodeBuilderParams {}
```
bobbinth: Is this still needed?

core/src/mast/node/call_node.rs
```diff
@@ -552,45 +376,20 @@ impl proptest::prelude::Arbitrary for CallNodeBuilder {
...
-impl Default for CallNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct CallNodeBuilderParams {}
```
bobbinth: Is this still needed?

core/src/mast/node/join_node.rs
```diff
@@ -515,38 +351,12 @@ impl proptest::prelude::Arbitrary for JoinNodeBuilder {
...
-impl Default for JoinNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct JoinNodeBuilderParams {}
```
bobbinth: Is this still needed?

core/src/mast/node/external.rs
```diff
@@ -398,45 +224,22 @@ impl proptest::prelude::Arbitrary for ExternalNodeBuilder {
...
-impl Default for ExternalNodeBuilderParams {
-    fn default() -> Self {
-        Self {
-            max_decorators: 4,
-            max_decorator_id_u32: 10,
-        }
-    }
-}
+#[derive(Clone, Debug, Default)]
+pub struct ExternalNodeBuilderParams {}
```
bobbinth: Is this still needed?

bobbinth (approved):
Looks good! Thank you! I left some minor comments inline, but these can be addressed in a follow-up PR.

Also, I didn't review some components in depth (e.g., `MastForestBuilder`) as I know these will be heavily refactored in subsequent PRs.

crates/project/examples/protocol/userspace/mod.masm
huitseeker: Yep, overly broad, predated our change where we emit `sys.println` via `emit.PRINTLN` and drops the message pointer, instead of using trace/decorator machinery.

core/src/mast/node/basic_block_node/mod.rs
huitseeker: done

crates/project/examples/protocol/kernel/kernel/io.masm
huitseeker: Reinstated to show emit

crates/project/examples/protocol/userspace/mod.masm
huitseeker: reinstated

crates/assembly/tests/fixtures/protocol/userspace/mod.masm
huitseeker: reinstated

core/src/mast/node/mast_forest_contributor.rs
```diff
@@ -19,39 +18,17 @@ pub trait MastForestContributor {
     /// This method computes the fingerprint for a node directly from the builder data
     /// without first constructing a MastNode, providing the same result as the
     /// traditional fingerprint computation approach.
-    fn fingerprint_for_node(
-        &self,
-        forest: &MastForest,
-        hash_by_node_id: &impl LookupByIdx<MastNodeId, MastNodeFingerprint>,
-    ) -> Result<MastNodeFingerprint, MastForestError>;
+    fn fingerprint_for_node(&self, forest: &MastForest) -> Result<Word, MastForestError>;
```
plafer, line 21: I don't think we still need the fingerprint terminology/distinction anymore, right? i.e. we can use commitment/digest instead

core/src/mast/node/mast_forest_contributor.rs
```diff
@@ -130,229 +106,28 @@ impl proptest::prelude::Arbitrary for MastNodeBuilder {
...
-        assert_ne!(
-            fp_u32assert2, fp_mpverify,
-            "Basic blocks with U32assert2 vs MpVerify should have different fingerprints"
-        );
-    }
-
     #[test]
-    fn basic_block_fingerprint_different_assert_values_no_decorators() {
+    fn basic_block_fingerprint_ignores_assert_error_codes() {
         let forest = MastForest::new();
```
plafer, line 117: This test asserts that assert codes don't influence the fingerprint (i.e. the MAST root). Aren't assert codes stored in a group, such that they *do* influence the MAST root/fingerprint? Is the plan to move the assert code out of the operation and into debug info?

plafer (approved):
LGTM!

Left question/suggestion, which can all be addressed in subsequent PRs/issues

core/src/mast/node/mast_forest_contributor.rs
huitseeker: Answered [above](https://github.com/0xMiden/miden-vm/pull/3208#discussion_r3320949929): will deal with it later.

core/src/mast/node/mast_forest_contributor.rs
huitseeker: Good question. The assert code is not currently stored as an op-group immediate for MAST hashing. `OpBatchAccumulator::push_op()` only packs `op.op_code()` into the operation group, and `Operation::imm_value()` returns `Some(_)` only for `Push`, not for `Assert`, `U32assert2`, or `MpVerify`.

So this test is asserting the current invariant: changing the error code changes the serialized/basic-block operation payload used for diagnostics, but not the MAST root / dedup identity. The error-code message map remains in `DebugInfo`, and this PR is not moving the numeric assert code out of `Operation`.

That said, the test message saying “debug metadata” is too loose because the numeric code is still carried by the operation. We should tighten the wording to say that assert error codes are diagnostic payload and should not affect MAST node identity.

bitwalker (approved):
LGTM

PR #3209, opened 2026-05-27
fix(ci): drop stale blake3 benchmark release flag

huitseeker:
The Blake3 non-regression workflow was still invoking:

```sh
miden-vm prove ... --release
```

`prove --release` was removed when debug decorators were removed (in favor of `--trace`), so the benchmark jobs now fail at CLI argument parsing before collecting measurements.

This updates `scripts/blake3_nonregression.py` to call the current prove CLI shape.


bobbinth (approved):
Looks good! Thank you!

PR #3211, opened 2026-05-28
feat(syntax): support parsing of inline MASM blocks

bitwalker:
This is to enable support for languages that want to parse MASM source code embedded in the source of that language. Such blocks are expected  to be structured as if the MASM was in the body of a procedure, i.e.  no top-level items.

This has no effect on parsing or syntax, it is simply exposing some additional low-level parser functionality.

### DIFF
--- a/crates/assembly-syntax-cst/src/parser.rs
+++ b/crates/assembly-syntax-cst/src/parser.rs
@@ -105,6 +105,31 @@ pub fn parse_text(input: &str) -> Parse {
     parse_source_file(detached_source_file(input))
 }
 
+/// Parses a inline MASM from a subset of a source-managed file into a lossless CST.
+///
+/// Content of an inline MASM block is parsed like the body block of a procedure - it is not
+/// supported to define top-level items in an inline MASM block
+pub fn parse_inline_masm(source: Arc<SourceFile>, bounds: Option<SourceSpan>) -> Parse {
+    let parser_source = Arc::clone(&source);
+    if let Some(bounds) = bounds {
+        Parser::new_bounded(parser_source.as_ref(), bounds).parse_inline_masm(source)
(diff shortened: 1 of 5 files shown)


crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -105,6 +105,31 @@ pub fn parse_text(input: &str) -> Parse {
     parse_source_file(detached_source_file(input))
 }
 
+/// Parses a inline MASM from a subset of a source-managed file into a lossless CST.
+///
+/// Content of an inline MASM block is parsed like the body block of a procedure - it is not
+/// supported to define top-level items in an inline MASM block
+pub fn parse_inline_masm(source: Arc<SourceFile>, bounds: Option<SourceSpan>) -> Parse {
```
bobbinth, line 112: Question: for the inline MASM, how would we define what the input/output stack states are? That is, how would it read parameters from and return something to the surrounding context?

bobbinth (approved):
Not a very deep review - but looks good! Thank you! I left one question inline.

crates/assembly-syntax-cst/src/parser.rs
bitwalker: That's up to the context in which this is used - in essence it's the caller's responsibility to reason about it. 

For the use case I have in my language, the inputs/outputs are declared as part of the inline MASM syntax, e.g.:

```
let mut out: u32;
let input = 42u32;

let result: u32 = unsafe {
    masm(input, inout out) {
        dup.1
        swap
        mem_store
    }
};
```

This shows both types of output from inline MASM: returned on the operand stack, and written to local memory. The compiler knows what to expect from the MASM and ensures it adheres to the stack discipline required.

PR #3215, opened 2026-05-28
Remove uneeded clippy exceptions

huitseeker:
1. **Removing annotations that no longer fire** (the lint no longer applies after upstream refactors or after clippy itself changed its heuristics).
2. **Refactoring the code** so clippy is satisfied without a suppression
3. **Bumping clippy configuration thresholds** (`too_many_lines: 200`, `too_many_arguments: 11`, `type_complexity: …`) where the lint genuinely flags legitimate code that should not be split.


crates/package-registry-local/src/lib.rs
```diff
@@ -171,6 +171,8 @@ impl LocalPackageRegistry {
         let index_checksum: [u8; 32];
         let index = if index_path.exists() {
             let mut contents = String::with_capacity(4 * 1024);
+            // Lock acquisition requires holding the file handle, so we can't use
+            // fs::read_to_string here (it would open a separate handle without the lock).
             #[allow(clippy::verbose_file_reads)]
```
Mirko-von-Leipzig: Might be better to use `expect` with a `reason = ""`

Mirko-von-Leipzig (approved):
I would recommend banning `allow` in favor of `expect`

crates/package-registry-local/src/lib.rs
huitseeker: Great idea, done

crates/assembly/src/project.rs
```diff
@@ -668,8 +672,7 @@ enum RegisteredSourcePackage {
 
 struct LoadedTargetSources {
     root: Box<Module>,
-    #[allow(clippy::vec_box)]
```
bitwalker, line 671: This is exactly the sort of place where this suppression makes sense to keep rather than refactoring - we don't _use_ modules from this vector, we're just passing them along after loading. There is zero benefit in reallocating these just to satisfy `clippy` here IMO

huitseeker: Ah, yes, indeed, replaced by an expect.

bitwalker (approved):

PR #3216, opened 2026-05-28
feat(assembly): extensible project assembly via source providers

bitwalker:
This PR introduces the ability to extend project assembly to arbitrary source languages (e.g. Rust), while leaving a lot of the core plumbing to the assembler, so that the assembler can request assembly of a project target in source form on-demand, rather than needing to pre-assemble all non-MASM source projects in advance.

As a result of this extensibility, we no longer need to allow the `path` value in `miden-project.toml` targets to be elided, and so this PR requires that it be specified from now on. Now, the path can refer to any file type (so long as it has a file extension), and the appropriate source provider for that file type will be used to compile to Miden Assembly AST for use by the assembler.

The default project assembler is instantiated with a source provider for MASM projects, and new APIs have been added to construct the assembler with additional/alternative source providers.


crates/assembly/src/project/providers.rs
bobbinth, line 1: nit: we should probably move this under `src/project/providers/mod.rs` as this is the convention we use for most of the codebase.

crates/assembly/src/project/dependency_graph.rs
```diff
@@ -152,10 +155,15 @@ impl DependencyGraph {
         origin: &ProjectSourceOrigin,
         manifest_path: &FsPath,
         _workspace_root: Option<&FsPath>,
```
bobbinth: Not from this PR, but why do we pass `_workspace_root` here if we are not using it? Seems like this is an internal function, and if we need to add this parameter at a later date, we can do it then.

crates/assembly/src/project/dependency_graph.rs
```diff
@@ -284,4 +325,52 @@ impl DependencyGraph {
             },
         }
     }
+
+    fn compute_path_source_hash(
+        &self,
+        context: &TargetAssemblyContext<'_>,
+        source_provider: &SourceProvider,
+    ) -> Result<Word, Report> {
```
bobbinth, line 320: Could we add a brief do comment explaining what this function does? My understanding is that we compute a unique identifier for a project identified by `context` based on paths to all relevant files in the project - i.e., we concatenate the root source file path with all supporting file paths and then hash the resulting string.

crates/assembly/src/project.rs
```diff
@@ -91,17 +124,94 @@ pub struct ProjectSourceInputs {
...
+        Ok(Self {
+            path: path.into_boxed_path(),
+            content: content.into_boxed_str(),
+        })
+    }
+}
+
+pub struct SourceProvider {
+    registered: BTreeMap<&'static str, Box<dyn ProjectSourceProvider>>,
+}
```
bobbinth: nit: maybe a better name for this struct could be `SourceProviderRegistry`?

Also, would be good to add some brief doc comments to this struct and put it into a separate section. Or maybe a better place for this would be in `src/project/providers/registry.rs`.

crates/assembly/src/project.rs
```diff
@@ -44,6 +47,20 @@ impl Assembler {
...
     ) -> Result<ProjectAssembler<'a, S>, Report>
+    where
+        S: PackageCache + ?Sized,
+    {
+        let masm_provider = Box::new(MasmSourceProvider) as Box<_>;
+        self.for_project_at_path_with_providers(manifest_path, store, [masm_provider])
+    }
+
+    /// Get a [ProjectAssembler] configured for the project whose manifest is at `manifest_path`.
+    pub fn for_project_at_path_with_providers<'a, S>(
```
bobbinth, line 58: Not for this PR, but I wonder if we should refactor the surface API of the assembler to be package-first. Basically, would be nice to have something like `Assembler::assemble_package()` (or something similar) rather than having to do it in multiple steps - first, creating the `PackageAssembler`, then potentially modifying the returned package assembler, and then finally using it to assemble a package.

crates/assembly/src/project.rs
```diff
@@ -91,17 +124,94 @@ pub struct ProjectSourceInputs {
...
+        self
+    }
+
+    #[inline]
+    pub fn get_provider(&self, file_type: &str) -> Option<&dyn ProjectSourceProvider> {
+        self.registered.get(file_type).map(AsRef::as_ref)
+    }
+}
+
 pub struct ProjectAssembler<'a, S: PackageCache + ?Sized> {
```
bobbinth, line 195: One general comment (and probably not for this PR), we should add better doc comments to the `ProjectAssembler` and related structs. This should probably be done once we finalize the overall API of the assembler as mentioned in the above comment.

crates/assembly/src/project/providers/masm.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        if path.extension() == Some(AsRef::<std::ffi::OsStr>::as_ref(Module::FILE_EXTENSION)) {
+            paths.push(path);
+        }
+    }
+
+    Ok(())
+}
+
+fn excluded_target_roots(context: &TargetAssemblyContext<'_>) -> BTreeSet<PathBuf> {
```
bobbinth, line 146: Could we add some doc comments to this function? It is not super clear to me what it is trying to do (seems like we are collecting all executable module paths unless the path is for the root for the project?

Also, is this logic only relevant for MASM provider? Or should this function be a method on the `TargetAssemblyContext` struct?

crates/assembly/src/project/providers/masm.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+    fn provide_source_provenance(
+        &self,
+        context: &TargetAssemblyContext<'_>,
+    ) -> Result<ProjectSourceProvenanceInputs, Report> {
+        resolve_target_source_paths(context)
+    }
+}
+
+fn load_target_sources(context: &TargetAssemblyContext<'_>) -> Result<ProjectSourceInputs, Report> {
```
bobbinth: nit: the body of this function could probably be just in the `provide_sources()` implementation above. AFAICT, there isn't a need for this extra hop.

crates/assembly/src/project/providers/masm.rs
```diff
@@ -0,0 +1,171 @@
...
+        context: &TargetAssemblyContext<'_>,
+    ) -> Result<ProjectSourceInputs, Report> {
+        load_target_sources(context)
+    }
+    fn provide_source_provenance(
+        &self,
+        context: &TargetAssemblyContext<'_>,
+    ) -> Result<ProjectSourceProvenanceInputs, Report> {
+        resolve_target_source_paths(context)
+    }
```
bobbinth, line 49: Question: it seems like we call `provide_source_provenance()` as one of the first things in the `provide_sources()` function. Would it make sense to instead include the result of `provide_source_provenance()` into `ProjectSourceInputs` and get rid of `provide_source_provenance()` method? Something like:

```rust
pub struct ProjectSourceInputs {
    pub root: ModuleFile,
    pub support: Vec<ModuleFile>,
}

/// `ModuleWithProvenance` is just a placeholder name - something else could be a
/// better choice
pub struct ModuleWithProvenance {
    pub module: Box<Module>,
    pub path: Box<std::path::Path>,
    pub source: Box<str>,
}
```

Or would this complicate things rather than make them simpler?

bobbinth (commented):
Looks good! Thank you! Not a very detailed review, but I left some comments inline.

Also, one higher-level comment (which I also mentioned inline): I wonder if we should try to make the assembler interfaces "more intuitive" in the follow-up PRs. Right now, they seem like in a transitionary state between the old assembler and the new, package-first assembler.

crates/assembly/src/project/providers.rs
bitwalker: I seem to recall we were going to move towards the recommended style with this - obviously we still have a lot of `mod.rs` around, but new code has been largely using this new style for some time.

I'm fine with moving it, but we should probably clarify whether we're all on the same page with what style to use going forward, since we've discussed this a few times in the past and I no longer know where we're at with it. I know a common complaint from others with `mod.rs` was that it can get murky in some editors what file you are looking at if you have a bunch open, whereas on the flip side it allows all modules in a namespace to be in a single directory. One caveat of that point though, is the file system structure doesn't match the module structure as closely. I don't really care one way or the other on the file system part - I mostly navigate by module/file name anyway, so the `mod.rs` is only slightly more annoying there, but it's minor in my case.

crates/assembly/src/project.rs
bitwalker: We can do this, so long as we can require building the assembler crate with `std` - which we should do anyway IMO. The only caveat has been the web client, but I'm unclear on whether it will continue to need to assemble things in the presence of packages, or if it even does today. If it does need to assemble things itself, we should have that team try to use the `wasm32-wasip1` or `-wasip2` targets, and build with `std`. It's unclear to me why the web client would need the assembler, but I don't know much about how it is implemented. It _is_ possible to target Wasm with `std` enabled, but some APIs may fail at runtime without WASI support - and that could be fine for their use case.

crates/assembly/src/project/providers/masm.rs
bitwalker: This will become irrelevant with Path Semantics 2.0, which I'm just about done implementing. Currently, this is needed to ensure we don't pull in root modules of other targets of the same project, since we just blindly gather all `.masm` files in the same directory/sub-directories. The new syntax/semantics around paths and modules makes the module structure explicit in code (like Rust), so this kind of filtering goes away.

crates/assembly/src/project/providers/masm.rs
bitwalker: We can't do that because `provide_source_provenance` is called while computing the dependency graph, and we don't want to be parsing MASM/compiling Rust code during dependency graph construction when all we need is the source provenance inputs.

crates/assembly/src/project/providers/masm.rs
bitwalker: I've added a note to this function, both to explain its purpose and indicate that it should be removed after #3220 is merged

crates/assembly/src/project.rs
```diff
@@ -228,11 +338,27 @@ where
...
+
+        // Section: build provenance
+        //
+        // This is produced before actual assembly, while we still have the sources on hand
+        if let Some(provenance) = self.dependency_graph.build_source_provenance(
+            &package_id,
+            project.as_ref(),
+            target,
+            profile_name,
+            &self.source_provider,
```
huitseeker, line 331: This seems to change the `assemble_with_sources()` contract. When `sources` is `Some`, we normalize those modules above, but this provenance call still asks the registered file provider to hash the target path on disk. That means caller-provided modules can either fail because the manifest path is missing/unsupported, or get a `PROJECT_SOURCE_PROVENANCE` section for files that were not actually assembled.

The existing `PackageBuildProvenance` docs still say caller-provided module builds intentionally omit this section. 

Should we keep the old `has_provided_sources` skip here, or hash the provided modules directly?

crates/assembly/src/project/providers/masm.rs
```diff
@@ -0,0 +1,173 @@
...
+impl ProjectSourceProvider for MasmSourceProvider {
+    fn file_type(&self) -> &'static str {
+        "masm"
+    }
+
+    fn provide_sources(
+        &self,
+        context: &TargetAssemblyContext<'_>,
+    ) -> Result<ProjectSourceInputs, Report> {
+        let ProjectSourceProvenanceInputs { root, support } = resolve_target_source_paths(context)?;
```
huitseeker, line 16: Nit: this path seems worth separating from provenance loading. `provide_sources()` calls `resolve_target_source_paths()`, which builds `SourceFileProvenance` and reads every MASM file into memory, then `parse_module_file()` reads the same files again through `ModuleParser::parse_file()`.

Could this split path discovery from content loading? 

That would make the normal assembly path simpler and keep provenance hashing from influencing how source loading works.

crates/assembly/src/project/dependency_graph.rs
```diff
@@ -169,12 +171,40 @@ impl DependencyGraph {
...
+                    )));
+                };
+                let root_path = project_root.join(&target_path);
+                let root_path = root_path.canonicalize().map_err(|error| {
+                    Report::msg(format!(
+                        "failed to resolve target source '{}': {error}",
+                        root_path.display()
+                    ))
+                })?;
+                let context = TargetAssemblyContext {
```
huitseeker: Nit: this context setup duplicates the manifest/root/target-path resolution in `ProjectAssembler::load_target_sources()`. The two paths are already slightly different because this provenance path hard-codes `warnings_as_errors: false`, while source loading uses the assembler flag.

Could we share the target-context construction helper?

huitseeker (commented):
Mostly LGTM, one question on `assemble_with_sources`.

crates/assembly/src/project/providers/masm.rs
bitwalker: After #3220, we will have to tolerate some redundancy here, because the only way to know the set of source files that actually contribute to the AST is to parse them - i.e. the current behavior where we just assume all the source files in some directory belong to the target (minus a few obvious paths we can definitively exclude) is going away in favor of explicit module structure.

For Rust, things are somewhat simpler because we are required to set things up in a one-crate-per-component configuration right now, so we can cheat a bit there - but in principle we would be required to do the same kind of thing as MASM if we allowed multiple package artifacts to be derived from a single crate.

crates/assembly/src/project/dependency_graph.rs
bitwalker: It's hardcoded in this case because we don't actually care about parse/compile warnings when just asking for source provenance.

That said, we can probably factor out the common details here at least

crates/assembly/src/project.rs
bitwalker: After thinking on this a bit, I think it probably makes sense to remove the ability to assemble with pre-parsed modules entirely now - the actual use case for it is no longer relevant after this. It was originally motivated specifically for Rust, where we needed to compile Rust to MASM in advance of assembly - that goes away with source providers.

Purely manual assembly can still be performed using the existing `Assembler` APIs if needed. We're aiming to remove most direct uses of the `miden-assembly` crate anyway (other than possibly the one exception of invoking the project assembly path from a build script); and outside of tests of the assembler itself, there is basically no need for assembly of virtual projects.

bitwalker: I've removed `assemble_with_sources` - we'll need to merge #3220 before the next release so the compiler has a migration path, but that's the plan anyway.

I did add some additional checks on the inputs returned from the source provider, to make sure that the root module kind and namespace match what was declared for the target, but the "normalization" that was implicitly applied before has also been removed - better to return an error than silently change what was returned from the source provider.

crates/assembly/src/project/dependency_graph.rs
bitwalker: I've factored this out as `TargetAssemblyContext::new` - we still suppress `warnings_as_errors` when computing source provenance for the dependency graph, but we don't have the duplication around context initialization anymore.

crates/assembly/src/project.rs
```diff
@@ -44,6 +47,20 @@ impl Assembler {
...
+        let masm_provider = Box::new(MasmSourceProvider) as Box<_>;
+        self.for_project_at_path_with_providers(manifest_path, store, [masm_provider])
+    }
+
+    /// Get a [ProjectAssembler] configured for the project whose manifest is at `manifest_path`.
+    pub fn for_project_at_path_with_providers<'a, S>(
+        self,
+        manifest_path: impl AsRef<FsPath>,
+        store: &'a mut S,
+        providers: impl IntoIterator<Item = Box<dyn ProjectSourceProvider>>,
```
huitseeker, line 62: The new provider constructors expose `Box<dyn ProjectSourceProvider>`, but the provider API is not reachable from outside `miden-assembly`.

Could we re-export the provider trait, context, and provenance input types from `lib.rs`? Without that, external Rust providers cannot implement this new API.

bitwalker: Good catch, definitely need those to be available externally

crates/assembly/src/lib.rs
```diff
@@ -45,7 +45,10 @@ pub use miden_mast_package as package;
 #[doc(hidden)]
 pub use self::linker::LinkerError;
 #[cfg(feature = "std")]
-pub use self::project::{ProjectAssembler, ProjectSourceInputs, ProjectTargetSelector};
+pub use self::project::{
+    MasmSourceProvider, ProjectAssembler, ProjectSourceInputs, ProjectSourceProvenanceInputs,
```
huitseeker, line 49: Looks like one provider type is still missing from this export set. A downstream provider has to return `ProjectSourceProvenanceInputs`, but its public fields use `SourceFileProvenance`, and that type is still only under the private `project` module.

 Could we re-export `SourceFileProvenance` here too?

huitseeker (commented):
Just one missing export

huitseeker (approved):
LGTM, thanks!

bobbinth (approved):
Looks good! Thank you!

PR #3217, opened 2026-05-29
feat(assembly): remove legacy parsing backend

bitwalker:
This PR follows up on the introduction of a new parser for MASM in #2906 and #2907, by removing the legacy LALRPOP backend and related cruft. It also removes some of the hacks used to preserve certain legacy parser behaviors like constant folding, which would result in the AST being different than what was actually written by the user.


huitseeker (approved):
Yay! Love the red diff!

PR #3218, opened 2026-05-30
perf: O(log n) name conflict checks in Module via BTreeMap index

### LINKED ISSUE #3212: O(n²) item insertion in Module's define_* methods
### Name-map dedup for O(n²) item insertion in module.rs

**Where:** `crates/assembly-syntax/src/ast/module.rs` — `Module` struct and all `define_*` methods (`define_constant`, `define_type`, `define_enum`, `define_procedure`, `define_alias`).

**The problem:** Every `define_*` method scans the whole `Vec<Export>` to check for name conflicts. On a module with thousands of exports — for example a big auto-generated wasm file — each definition does a full linear scan. That's O(n²) insertion. The `too_many_items` regression test takes 80 seconds because of this.

**The fix:** Add a `BTreeMap<String, usize>` to `Module` that maps each export name to its index in `items`. The conflict check becomes a `name_map.get(name)` — O(log n) instead of O(n). The map is kept in sync by updating it inside `push_export()` right after pushing onto `items`.

Five methods changed. Each lost a `.iter().find(...)` and gained a `.name_map.get(...)` using the stored index for the error span lookup.

**The result:** `too_many_items` drops from 80 seconds to 1.5 seconds. The other `define_*` methods see the same savings whenever a module has many items.

oksanaphmn:
Fixes #3212

Name conflict detection in Module previously scanned all exports, resulting in O(n²) insertion time for large modules. This change adds a BTreeMap-based name index, reducing conflict checks to O(log n).

A new Module::take_items() helper keeps the index synchronized when module contents are consumed, and existing consumers in sema and linker were updated to use it.

This significantly improves performance for modules containing large numbers of items

### DIFF
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -549,10 +549,10 @@ impl Module {
     pub fn items(&self) -> &[Export] {
         &self.items
     }
-
-    /// Get a mutable reference to the storage for items defined in this module
-    pub fn items_mut(&mut self) -> &mut Vec<Export> {
-        &mut self.items
+    /// Returns a mutable iterator over the items in this module.
+    /// Note: does not expose `Vec` directly to preserve the `name_map` invariant.
+    pub fn items_mut(&mut self) -> impl Iterator<Item = &mut Export> {
+        self.items.iter_mut()
     }
 
     /// Returns items exported from this module.
--- a/crates/assembly-syntax/src/ast/module.rs
+++ b/crates/assembly-syntax/src/ast/module.rs
@@ -155,6 +155,8 @@ pub struct Module {
     /// Maps export name to its position in `items`, for O(log n) conflict checks.
     /// Must be kept in sync with `items` via `push_export`.
     name_map: BTreeMap<String, usize>,
+    /// Whether mutable item access may have changed item names without updating `name_map`.
+    name_map_dirty: bool,
     /// AdviceMap that this module expects to be loaded in the host before executing.
     pub(crate) advice_map: AdviceMap,
 }
@@ -184,6 +186,7 @@ impl Module {
             kind,
             items: Default::default(),
             name_map: BTreeMap::new(),
+            name_map_dirty: false,
             advice_map: Default::default(),
         }
     }
@@ -238,6 +241,7 @@ impl Module {
 
     pub(crate) fn push_export(&mut self, item: Export) -> Result<(), SemanticAnalysisError> {
         self.ensure_item_capacity(item.span())?;
+        self.ensure_name_map_current();
         let idx = self.items.len();
         let prev = self.name_map.insert(item.name().to_string(), idx);
         debug_assert!(prev.is_none(), "duplicate export inserted via push_export: {}", item.name());
@@ -245,15 +249,29 @@ impl Module {
         Ok(())
     }
 
+    fn ensure_name_map_current(&mut self) {
+        if !self.name_map_dirty {
+            return;
+        }
+
+        self.name_map.clear();
+        self.name_map.extend(
+            self.items.iter().enumerate().map(|(idx, item)| (item.name().to_string(), idx)),
+        );
+        self.name_map_dirty = false;
+    }
+
     /// Takes all items from this module, clearing the name index.
     /// Used by the linker to consume module contents.
     pub fn take_items(&mut self) -> Vec<Export> {
         self.name_map.clear();
+        self.name_map_dirty = false;
         core::mem::take(&mut self.items)
     }
 
     /// Defines a constant, raising an error if the constant conflicts with a previous definition
     pub fn define_constant(&mut self, constant: Constant) -> Result<(), SemanticAnalysisError> {
+        self.ensure_name_map_current();
         if let Some(&idx) = self.name_map.get(constant.name.as_str()) {
             return Err(SemanticAnalysisError::SymbolConflict {
                 span: constant.span,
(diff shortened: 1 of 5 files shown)


crates/assembly-syntax/src/ast/module.rs
```diff
@@ -146,6 +152,9 @@ pub struct Module {
     kind: ModuleKind,
     /// The items (defined or re-exported) in the module body.
     pub(crate) items: Vec<Export>,
+    /// Maps export name to its position in `items`, for O(log n) conflict checks.
+    /// Must be kept in sync with `items` via `push_export`.
```
huitseeker, line 156: Could this index also cover callers that use `items_mut()`? That method is still public and returns `&mut Vec<Export>`, so a caller can add an item without updating `name_map`. After that, the next `define_*` call only checks the stale map and can miss a duplicate.

I could repro it with this test:

```rust
#[test]
fn items_mut_can_bypass_module_name_index() {
    let mut module = Module::new(ModuleKind::Library, Path::new("mod"));
    module.items_mut().push(Export::Type(type_alias_with_name("dup").into()));

    let result = module.define_constant(constant_with_name("dup"));

    assert!(
        matches!(result, Err(SemanticAnalysisError::SymbolConflict { .. })),
        "expected SymbolConflict after direct item insertion, got {result:?}"
    );
}
```

It fails with `got Ok(())`. Maybe `items_mut()` needs to go away (ideally), return a narrower API, or mark the index dirty so it is rebuilt before conflict checks.

huitseeker (commented):
Thanks for tackling this!

crates/assembly-syntax/src/ast/module.rs
oksanaphmn: You're right, even though the current callers only iterate over items, items_mut() could still be used to modify the underlying Vec and bypass the name_map invariant.

I've pushed a follow-up change that makes items_mut() return an iterator instead of &mut Vec<Export>. Existing callers continue to work as before, but it's no longer possible to push or remove exports directly and leave name_map out of sync.

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -540,10 +549,10 @@ impl Module {
     pub fn items(&self) -> &[Export] {
         &self.items
     }
-
-    /// Get a mutable reference to the storage for items defined in this module
-    pub fn items_mut(&mut self) -> &mut Vec<Export> {
-        &mut self.items
+    /// Returns a mutable iterator over the items in this module.
+    /// Note: does not expose `Vec` directly to preserve the `name_map` invariant.
+    pub fn items_mut(&mut self) -> impl Iterator<Item = &mut Export> {
```
huitseeker, line 587: This still seems to leave one way to make `name_map` stale: a caller can rename an item through the returned `&mut Export`. For example, constants and type aliases expose `name` as a public field, so the next `define_*` call checks the old map and misses the duplicate.

I could repro it with this test:

```rust
#[test]
fn items_mut_rename_can_bypass_module_name_index() {
    let mut module = Module::new(ModuleKind::Library, Path::new("mod"));
    module.define_constant(constant_with_name("old")).unwrap();

    for item in module.items_mut() {
        let Export::Constant(constant) = item else {
            continue;
        };
        constant.name = ident_with_name("dup");
    }

    let result = module.define_type(type_alias_with_name("dup"));

    assert!(matches!(result, Err(SemanticAnalysisError::SymbolConflict { .. })));
}
```

It fails with `got Ok(())`. Maybe this should return a narrower iterator, rebuild the index after mutable visits, or avoid exposing mutable access to fields that decide the export name.

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -228,32 +238,40 @@ impl Module {
 
     pub(crate) fn push_export(&mut self, item: Export) -> Result<(), SemanticAnalysisError> {
         self.ensure_item_capacity(item.span())?;
+        let idx = self.items.len();
+        let prev = self.name_map.insert(item.name().to_string(), idx);
+        debug_assert!(prev.is_none(), "duplicate export inserted via push_export: {}", item.name());
```
huitseeker, line 247: This assertion can be reached from normal input now. A C-like enum variant is inserted as a constant before the enum type itself is pushed, so this panics in debug builds:

```rust
#[test]
fn enum_variant_matching_enum_name_reports_symbol_conflict() {
    let context = SyntaxTestContext::default();
    let source = "\
enum DUP : u8 {
    DUP,
}
";
    let error = context
        .parse_module(source)
        .expect_err("expected symbol conflict when enum variant matches enum name");
    assert_symbol_conflict(&error, "DUP");
}
```

It fails at this `debug_assert!` with `duplicate export inserted via push_export: DUP`. Maybe `push_export` should reject the duplicate without mutating `name_map`, or `define_enum` should check variant names against the enum alias before inserting the variant constants.

oksanaphmn: Fixed in the latest commit. define_enum now checks variant names against the enum name before inserting variant constants, so enum DUP : u8 { DUP } now reports a SymbolConflict instead of hitting the assertion. Added a regression test as well

crates/assembly-syntax/src/ast/module.rs
oksanaphmn: You're right, renaming an item after insertion can leave name_map stale because some name fields are still public.

I couldn't find any existing code paths that do this today, but the invariant isn't fully enforced. Fixing it properly would likely require tightening the API around those fields, which is probably better handled as a follow-up change.

huitseeker (approved):
Thanks!

PR #3220, opened 2026-05-31
feat(assembly): module/path semantics 2.0

### LINKED ISSUE #1890: Revisit `ParseOptions` and its uses to see if it can be removed
### What should be done?

The `ParseOptions` struct is used by the `Parse` trait to allow callers of `compile_and_link_statically` to override certain parser options. It was initially created quite some time ago, and the way the `Assembler` in particular is used has changed quite a bit. We should re-evaluate whether the remaining uses of non-default `ParseOptions` fit a simpler pattern that would allow us to remove `ParseOptions ` entirely, replace it with something else, or simplify it to remove no longer used functionality.

The `Parse` trait itself is used by the `Assembler` API to allow passing a variety of types that can be parsed (or converted) into a `Module`, such as files, file paths, strings of all types, `Arc<SourceFile>`, and even already-parsed `Module`s - to the various `Assembler` APIs such as `Assembler::compile_and_statically_link`. This provides a significant ergonomic win downstream of the assembler that is worth having the trait, compared to requiring callers to provide a `Box<Module>` themselves (although that _is_ an option, it would result in a lot of duplicated boilerplate downstream).

A couple of notes:

* The `path` field is used to allow providing a `LibraryPath` for the parsed `Module`, when one cannot be inferred from the `Parse` impl itself, or when the inferred `LibraryPath` would be incorrect or invalid. For example, the impl for `&Path` and `PathBuf` will infer a `LibraryPath` from the file system path (parent directory + base name of source file), while the impl for `&str` (and things that deref to `&str`) have nothing to infer a `LibraryPath` from at all. In the latter case, we need to be able to provide the path; and in the former case, the file system path may not match the desired `LibraryPath` (for example, kernel modules are required to be in an "unnameable" namespace that cannot be inferred).
* The `kind` field is used to allow overriding the kind of `Module` being parsed (i.e. whether it is an executable, kernel, or library module). This is related to `path` for obvious reasons, in that if you are overriding the path, you likely also need to override the `kind` if you are trying to compile something that _isn't_ an executable module (the only kind which can be inferred from syntax alone, due to the `begin..end` block defining a program entrypoint.
* The `warnings_as_errors` field is used to allow overriding how the parser handles warnings, by the assembler when it parsed a module, using the assembler-wide `warnings_as_errors` value.

### How should it be done?

What we need to evaluate is where we are setting these, and whether those use cases can be converted to either manually parsing into a `Module` (I'd suggest we only do this in the `assembly` crate itself), or replaced with something simpler.

This may require some refactoring of defaults in the parser/assembler, but we should be careful about how we approach those changes if we determine they are a possible solution.

### When is this task done?

When we've either simplified/removed `ParseOptions`, or determined the juice is not worth the squeeze.

### Additional context

_No response_

### LINKED ISSUE #2643: Refine semantics of imports/paths expressed in Miden Assembly
Recently, there's been a handful of issues related to pathing and symbol resolution in the assembler, see the description for each of these for a taste:

* #2637 
* #2330 
* #2627

We're mostly playing whack-a-mole here though, because the semantics of paths expressed in Miden Assembly are so loose, that it is extremely complicated to resolve an arbitrary path at link-time.

Here is a list of the things that contribute to the complexity of the current symbol resolution algorithm. To preface the list, I'm going to reference an example program that illustrates some of the complexities:

```
# lib/a/mod.masm
use lib::b::b2

pub proc a1
    exec.b2
end

# lib/b/mod.masm
pub use lib::a

pub proc b1
    exec.a::a1
end

pub proc b2
    push.1
end

# main.masm

use lib::b
begin
    exec.b::b1
end
```

1. Instructions may reference items (constants, procedures) using paths that may be either unqualified (a single identifier, resolved in the module's scope), partially-qualified (e.g. `exec.parent::child`, where `parent` may or may not have been imported into scope, these are always resolved from the containing module's scope), fully-qualified (e.g. `exec.::parent::child`, in which case the path is resolved from the global namespace).
2. The target of an import (i.e. `use`) may refer to (alias) any path, either module or item. At link-time, we use the context at each usage of the alias to drive resolution (e.g. `exec.imported_symbol` dictates that `imported_symbol` resolves to a procedure item through the corresponding alias. However, we can't know anything about the aliased path until we actually try to resolve it at link-time.
3. Imports may build off other imports in the same module, e.g. `use foo::bar` brings `bar` into scope in the module, and so `use bar::baz` later in the module is considered perfectly valid - we resolve all imports that are not fully-qualified, by first checking if the leading component of the path refers to an import in the current module (ignoring itself), and if so, resolving the rest of the imported path against that other import.
4. Imports may be re-exported, e.g. `pub use lib::a` in module `::lib::b` of the example code above, means that the path `::lib::b::a::a1` is a valid path which must be rewritten using recursive expansion to `::lib::a::a1`. In other words, any path may require this recursive expansion step to reach the concrete path of the item it references.
5. Paths can be relative (to the context in which the path is being resolved) or absolute (resolved in the global namespace). However, as a consequence of the previous point, even absolute paths may contain components that refer to re-exported items (i.e. there isn't any such thing as "absolute + fully-expanded"), which means that we must use heuristics to handle resolution of absolute paths that we can't match to a module/item at first look. For example, in the example program above, a path `::lib::b::a::b2` would _appear_ to refer to an item `b2` in the module `::lib::b::a` - but there is no such module (we only have `::lib::a`, `::lib::b`, and `::$exec`). The path _is_ resolvable though, but it requires us to:
    * First find the longest common prefix of the path which matches up to a known module - in this case that is `::lib::b` - and then resolve the remainder of the path, `a::b2` in the context of that module. 
    * Since `a::b2` is a relative path, we start by trying to identify if `a` is an imported symbol, which it is, expanding to `lib::a`, to that we join the remaining subpath, giving us `lib::a::b2`
    * Since `lib::a::b2` is relative, we again try to resolve relative to imports in the module, but there is no imported symbol `lib`. We fallback to treating the path as absolute in this case, i.e. `::lib::a::b2`.
    * We look for the longest common prefix of `::lib::a::b2`, which is `::lib::a`, and proceed to resolve `b2` relative to it
    * The symbol `b2` was imported from `lib::b::b2`, so we proceed with resolving that path
    * Since `lib::b::b2` is relative, we try to resolve relative to imports, but there is no imported symbol `lib`, so we fallback to treating it as absolute, i.e. `::lib::b::b2`
    * We look for the longest common prefix of `::lib::b::b2`, which is `::lib::b`, and proceed to resolve `b2` relative to it
    * The symbol `b2` refers to a concrete item, we're done
6. Lastly, imports can be cyclical (on purpose or otherwise), so we need to be able to detect cycles in resolution. This is possible, but currently unimplemented. Note that in the example program above, `::lib::a` and `::lib::b` refer to each other, but not in a cyclical fashion (i.e. resolution can succeed without entering an infinite loop).

Fundamentally, the issue comes from the level of flexibility permitted here, and some of it we really don't need, and removing it would simplify some things considerably:

1. We don't need to allow imports to build off other imports - this would allow us to further impose the next restriction.
2. We should dictate that imports are always assumed to be absolute paths, even if specified using relative path syntax (e.g. `use foo::bar` is equivalent to `use ::foo::bar`).

There are also some legitimately useful functionality that likely is not worth the weight:

1. By allowing re-exports to refer to modules and not just items, you gain a fair amount of flexibility in how you organize your code and refer to items within a library and externally - you can even support reorganizing your module hierarchy in backwards-compatible ways by simply re-exporting the new module in place of the old (e.g. `pub use ::mylib::new->old`). On the other hand, these ergonomic benefits come at an enormous complexity cost - namely it requires the recursive expansion technique described above. In short, we lose the ability to look at an absolute path and know that it must refer to a concrete item, or that at most, the item it points to is a re-exported alias to another concrete item.
2. We currently allow instructions to reference relative paths with arbitrarily many components, i.e. referring to nested modules of an imported module, e.g. `exec.imported_module::submodule::symbol`. This adds complexity when dealing with invoke targets. Instead, if we dictated that instructions which reference a path, must do so either as a local symbol (e.g. `exec.foo`), an import-qualified item path (e.g. `exec.imported::foo`), or an absolute item path (e.g. `exec.::lib::submodule::foo`), we'd be able to trivially determine how to handle such paths during symbol resolution - in conjunction with specifying that imports are always aliasing absolute paths.

My latest PR that fixes some known issues with symbol resolution itself has known issues, and while I was able to address all of them with a larger refactoring, that refactoring was gross enough that I didn't even submit it for review. I then attempted a second from-scratch implementation, and quickly ended up in a similar place. Long story short, the existing semantics in Miden Assembly are impossible to sanely write a maintainable symbol resolver for IMO, and rather than continue beating my head against the wall on it, I'd rather revisit our priors.

This is something we should probably try and land in v0.22

### LINKED ISSUE #2833: Fallback to absolute path in assembly accepts convoluted paths
When resolving imports, the assembler falls back to treating the path as absolute if relative resolution fails. This can accept confusing paths like:

```
use foo::bar::baz  # if foo::bar::baz fails, tries ::foo::bar::baz
```

Related to #2643 (Refine semantics of imports/paths expressed in Miden Assembly).

The fallback behavior makes it harder to diagnose import errors because typos may silently resolve to unexpected absolute paths instead of failing clearly.

Recommendation: either remove the absolute-path fallback or make it require explicit `::` prefix to indicate absolute paths, improving error clarity.

bitwalker:
This PR is the follow-up/implementation of the propsed changes in #2643, what I've been calling "Path Semantics 2.0". It imposes a stricter, more explicit set of syntax and semantics around a set of tightly inter-related concepts in Miden Assembly:

* How namespaces are assigned to modules/packages
* How module structure is defined
* How the module surface exported from a Miden package is determined
* What constitutes a legal symbol reference/symbol path
* What constitutes a legal import or re-export

The new syntax and semantics are documented both in the user documentation under `docs`, as well as in the `miden-assembly` readme, but I summarize/compare the old and new approaches below briefly.

NOTE: The review breaks changes up across multiple commits, in the order that things were introduced/migrated - I would suggest looking through the changes commit-by-commit to understand how things changed; the commits which contain the test changes (and thus the bulk of the line count) are separated from other changes, with few exceptions.

## Semantics v1

* Namespaces were implicit (derived from filesystem directory names) or explicitly provided to the assembler out-of-band. This made it difficult to know how to refer to submodules of the same library. Miden projects (i.e. `miden-project.toml`) improved this for projects, but non-`miden-project.toml`-controlled packages still had the old ambiguity.
* The set of modules that belonged to a specific namespace was also implicit - any `.masm` in the directory tree would be parsed and included.
* The module surface of assembled libraries/packages was derived entirely from the visibility of items in the module tree, unless you explicitly provided a subset of modules to the linker as the public interface. This made it difficult to control the visibility of items exported from the package unless you used the lower-level APIs available from the assembler.
* There was no way to express the set of packages that a package depended on in the absence of `miden-project.toml`. If you referred to a namespace in your MASM code, it was left up to the linker to try and resolve it from the full module graph.
* Essentially any path/symbol was legal, you could import modules/items and export/re-export modules/items. Imports could refer to other imports in scope, to an almost unbounded extent. Imports could be relative to the current namespace or the global namespace - determining which was based on what was in scope. This could easily result in incorrect symbol resolution. It also required recursive path expansion, and due to the loose semantics, multiple fallback paths would be attempted to try and make the reference valid - this was not only error-prone/unstable, but expensive.
* Symbol resolution did not properly model item visibility, which resulted in bugs such as being able to reference private constants of a module from another one.

In short - the v1 semantics were horribly weak and implicit behavior was rampant.

## Semantics v2

* A Miden Assembly namespace is always rooted in a single top-level module that determines the module structure of that namespace by explicitly declaring its submodules and their visibility using `mod <name>`/`pub mod <name>`. This top-level module is called the "root" module of the namespace.
* A Miden project (via `miden-project.toml`) declares one or more targets, each of which specifies its namespace (via the `namespace` config, or inferred from the package name), as well as the path to the root module for that target using the `path` config. The set of modules included in that target is derived by parsing that root module, and then walking the module tree from the root. The filesystem structure is derived from the module structure, the exact same as Rust.
* In the absence of a `miden-project.toml` (which we should probably eventually make a requirement), the namespace of a package must be declared in the root module using the new `namespace <path>` declaration syntax. This is only supported in the root module, and if declared in the root module of a `miden-project.toml`-controlled target, must agree with what is declared in `miden-project.toml`. Similarly, if the modules in that package need to refer to external modules (i.e. defined in another package), those dependencies must be declared using the new `extern package "<package-id>"` declaration syntax. Again, this is only valid in the root module, and in the presence of a `miden-project.toml` all of the declared `extern package`s must be in the dependency list of the `miden-project.toml`.
* All descendants of a module are able to reference that module, as well as all other descendants of that module, i.e. modules under the same common parent have "internal" visibility. This is _not_ true of module _items_ though, such as procedures/constants/types - those must obey the normal visibility rules. So even if two modules `a` and `b` have the same common parent, `a` can only reference `pub` exports of `b` and vice versa.
* Only direct submodules and module items can be exported from a module - it is not legal to re-export modules.
* You may import both modules and items. Imports are always resolved in the global namespace, i.e. `use foo::bar` is equivalent to `use ::foo::bar`. As a consequence of this, imports cannot refer to other imports in the same module, and cannot be self-aliasing (i.e. `use path::to::self` is rejected). The one exception to this rule is imports which are submodule-relative - these must use the `self::` prefix (i.e. `use self::submodule::item`) to indicate to the resolver that the path resolves via a submodule rather than via the global namespace.

These semantics are much clearer and comprehensible. The symbol resolution implementation is now _drastically_ simpler and more efficient. We properly model visibility and module structure in the assembler and the package manifest. It is now much easier to reason about MASM that is not controlled by a `miden-project.toml`, and there is less ambiguity in how modules are included into a project target.

Additionally, these changes made it possible to finally remove `ParseOptions` (which closes #1890), and makes the `Parse` trait more useful on its own.

Most of the line count here is updating tests, the actual implementation is much smaller and simpler now.


crates/assembly/src/linker/mod.rs
```diff
@@ -558,6 +613,27 @@ impl Linker {
 
                 let module_index = ModuleIndex::new(module_index);
 
+                for import in module.imports() {
+                    let alias = import.alias();
+                    let context = SymbolResolutionContext {
+                        span: alias.target().span(),
+                        module: module_index,
+                        kind: None,
```
huitseeker: This looks like it lets a library turn a kernel proc into a normal `exec` target. I can reproduce it with the public API:

```rust
let lib = Assembler::with_kernel(source_manager.clone(), Arc::clone(&kernel_pkg))?
    .assemble_library("lib", "namespace lib\n\npub use $kernel::k1->k1", std::iter::empty::<&str>())?;

Assembler::with_kernel(source_manager, kernel_pkg)?
    .with_package(Arc::from(lib), Linkage::Dynamic)?
    .assemble_program("program", "namespace $exec\n\nuse lib::k1\n\nbegin\n    exec.k1\nend")?;
```

That final `exec.k1` succeeds. IIUC, imports are pre-resolved with `kind: None`, then the public import is exported with the resolved kernel gid, and a later consumer sees the callee as `lib::k1` instead of a kernel export. 

Should we reject public re-exports of kernel procs or preserve enough kernel provenance so the syscall-only check still fires?

crates/mast-package/src/package/manifest.rs
```diff
@@ -74,14 +130,27 @@ impl PackageManifest {
...
         };
         let mut has_procedures = false;
         for mut export in exports {
-            if export.is_procedure() {
+            normalize_export(&mut export)?;
+            if let Some(proc) = export.as_procedure() {
                 has_procedures = true;
+                // The presence of `begin` in any exported module is automatically made the
+                // entrypoint for that package.
+                if proc.path.last().is_some_and(|name| name == ast::ProcedureName::MAIN_PROC_NAME) {
```
huitseeker, line 144: Could we restrict this implicit entrypoint inference to the canonical `$exec::$main` path?

The source-level case I first tried does not reproduce: a library with `pub proc main` assembles successfully. But the manifest/package-level behavior does reproduce for a non-executable export whose final component is the internal `$main` name:

```rust
let non_exec_main_path: Arc<AstPath> =
    Arc::from(AstPath::new("foo").join(ProcedureName::MAIN_PROC_NAME).into_boxed_path());
let non_exec_main_export = PackageExport::Procedure(ProcedureExport::new(
    non_exec_main_path,
    None,
    Word::default(),
    None,
));

println!(
    "manifest with foo::$main export entrypoint: {:?}",
    PackageManifest::new([non_exec_main_export.clone()]).as_ref().map(|manifest| manifest.entrypoint())
);

let package_with_non_exec_main = Package::create(
    PackageId::from("foo"),
    Version::new(1, 0, 0),
    TargetType::Library,
    Arc::new(MastForest::new()),
    [non_exec_main_export],
    [],
);
```

The observed output is:

```text
manifest with foo::$main export entrypoint: Ok(Some("foo::$main"))
library package with foo::$main export: Err(found an executable entrypoint in a package declared with non-executable type)
```

So a library/package export named `foo::$main` is treated as an executable entrypoint even though it is not `$exec::$main`.

crates/assembly-syntax/src/parser/mod.rs
```diff
@@ -156,155 +170,198 @@ fn parse_forms_internal(
...
+    let mut modules = Vec::new();
+
+    let mut parser = ModuleParser::new(kind);
+    parser.set_warnings_as_errors(warnings_as_errors);
+    let root_ast = parser.parse_file(namespace.as_deref(), &root, source_manager.clone())?;
+
+    let namespace = Arc::<Path>::from(root_ast.path().to_path_buf().into_boxed_path());
+    let submodules = root_ast.submodules().to_vec();
+    seen.insert(namespace.clone());
+    walk_module_tree(
```
huitseeker, line 239: Could we align project source loading with this declaration-driven traversal, or document the intentional difference? The new standalone root loader ignores `.masm` files that are not reached through `mod` declarations, but project assembly still recursively discovers every `.masm` under the target root directory and then rejects undeclared ones.

I reproduced the mismatch with this throwaway project:

```rust
fs::write(
    project_dir.join("miden-project.toml"),
    r#"[package]
name = "discovery"
version = "1.0.0"

[lib]
path = "lib.masm"
"#,
)?;
fs::write(
    project_dir.join("lib.masm"),
    r#"pub proc entry
    push.1
end
"#,
)?;
fs::write(
    project_dir.join("unused.masm"),
    r#"pub proc orphan
    push.2
end
"#,
)?;

let standalone = Assembler::default()
    .assemble_library_from_root(project_dir.join("lib.masm"), Some(&Path::new("discovery")));

let mut project_assembler = Assembler::default()
    .for_project_at_path(project_dir.join("miden-project.toml"), &mut registry)?;
let project = project_assembler.assemble(ProjectTargetSelector::Library, "dev");
```

Observed output:

```text
standalone root assembly with undeclared unused.masm: Ok
project assembly with undeclared unused.masm: Err(module '::discovery::unused' is not declared by its parent module '::discovery' as `mod unused` or `pub mod unused`)
```

That gives the two public assembly paths different semantics for the same source tree.

huitseeker (commented):
Thanks! 

I do like moving off the “filesystem shape implies module graph” mindset. I like the explicit source/module surface semantics. 

But as a reviewer, it's a lot of moving parts at once — I find it hard to keep track of the invariants that govern:

  - source namespace
  - package target namespace
  - exported items
  - declared module surfaces
  - filesystem-discovered support modules
  - executable entrypoint metadata
  - kernel syscall-only linkage rules

... see for example the inline comments on entrypoint inference or the one on how root assembly is declaration-driven, but project assembly still recursively scans files. 

It would be tremendously useful to have a set of ADRs or perhaps just an up-to-date doc[^1] on some strict allowed initial shapes, initially quite restrictive (and then we can open them up incrementally later):

- target namespace == source namespace
-  executable entrypoint == `$exec::$main` or explicit valid export
- module graph == declared mod tree
- package manifest == complete closed-world surface
- dependencies == manifest contract

[^1]: note how core/README.md still references `Program`/`ProgramInfo`. The `try_into_program()`/`unwrap_program()` shims remain in `mast-package`. Worse: `core/src/program.rs` no longer exists — so the README now references a deleted file.

bitwalker:
> But as a reviewer, it's a lot of moving parts at once — I find it hard to keep track of the invariants that govern:
> 
> * source namespace
> * package target namespace
> * exported items
> * declared module surfaces
> ~~* filesystem-discovered support modules~~

I did extend the `miden-assembly` README to cover the semantics of these, as well as the user documentation, but I'm happy to extend either as desired to make things more explicit.

To help build the intuition here, I break things down along the following lines:

1. Two types of assembly: project and manual
2. Two manners in which sources can be provided: MASM source files, or MASM AST produced by some frontend. In practice, the latter route is only needed for compiling other languages to MASM, and for testing purposes.

When assembling manually, the assembling party has a lot more flexibility in how they assemble a target artifact, and the assembler itself requires them to provide information and structure that would otherwise be obtained from `miden-project.toml`:

* Target namespace and type
* Which module is the target namespace root, and which are supporting modules. 
* The public interface of the package
* What external packages are dependencies, and how to link them

MASM now provides a way to declare the target namespace with the `namespace` declaration, and external package dependencies with `extern package` declarations. Target type can be inferred to be executable or library based on the presence of `begin`. The explicit module structure (via `mod` declarations) now tells us not only the expected shape of a namespace, but the visibility of items in that namespace, both internally and externally.

This allows assembly of even without a project, while only needing to tell the assembler two things:

* The type of artifact you want (if you want something different than what is inferred from the root module)
* The path to the root module. **NOTE:** We no longer traverse the filesystem to locate all `.masm` sources and feed them all to the assembler (i.e. "discovery") - instead, the set of filesystem paths visited is the root module path, and the set of filesystem paths that correspond to its `mod` declarations (and those of its descendants), so it is explicit what gets pulled in as source inputs to assembly.

**NOTE:** Manual assembly is expected to be a niche use-case. In most cases outside of the compiler, direct use of the `Assembler` API is likely to only be present in test code or build scripts, with most usage leveraging project assembly, not the low-level manual assembly approach. However, the syntax is still useful for handwritten MASM test code, and allows details like namespacing to live in the source itself, rather than somewhere else.

Project assembly is far more straightforward - we ge the information we need for assembly straight from `miden-project.toml`, and from the module structure derived for a specific target.

For `.masm` source projects, both `namespace` and `extern package` are expected to be elided in favor of specifying those details in `miden-project.toml`. I've allowed both declarations to be specified, so long as they agree with what is in `miden-project.toml` - but that isn't strictly necessary, it just doesn't make sense to be pedantic about it unless they diverge.

For non-MASM source projects, those declarations aren't relevant.

> * executable entrypoint metadata

This is straightforward IMO:

* An executable `Package` knows its entrypoint (whatever its named)
* MASM syntax uses `begin` to define an executable module and its entrypoint procedure. That procedure is given a special name `$main` that is unnameable in MASM, but is otherwise not special in any meaningful way.
* Prior to explicit namespacing, we always assigned the executable module the `$exec` namespace. We no longer need to do this, or even support `$exec` as a valid namespace. The only reason to keep it, would be as a fallback in the very narrow case where you want to allow assembly of a MASM program from a source file that doesn't have a `namespace` declaration _and_ you haven't explicitly provided a namespace to the assembler. IMO we can just require an explicit namespace always and get rid of `$exec` entirely.
* When assembling an executable target, we expect to get AST that contains the `$main` procedure. This is the only way in which `$main` remains special today, but it is essentially just a convention - we could use other means to identify the entrypoint other than name, we just don't yet.
* Post-assembly, you can convert a `Package` into an executable one by specifying which export of the package should act as the entrypoint - at this point the name of the entrypoint procedure is irrelevant.

> * kernel syscall-only linkage rules

Nothing has changed here AFAIK - syscalls must still target exports from the root kernel module.

> ...root assembly is declaration-driven, but project assembly still recursively scans files.

I should have noted in the PR description that I'm expecting #3216 to land before this does, and was waiting to rebase on those changes before modifying how project assembly obtains sources, since those changes would directly conflict with each other.

> It would be tremendously useful to have a set of ADRs or perhaps just an up-to-date doc[1](#user-content-fn-1-b766731a22087c37a3e115b1d2df8420) on some strict allowed initial shapes, initially quite restrictive (and then we can open them up incrementally later):
> 
> * target namespace == source namespace
> * executable entrypoint == `$exec::$main` or explicit valid export
> * module graph == declared mod tree
> * package manifest == complete closed-world surface
> * dependencies == manifest contract
> 

The restrictions you've listed here are already the case for project assembly. For manual assembly as well for the most part - to the extent that there is additional flexibility, it is in terms of what modules may participate in the public interface of the package, and that's fine IMO (we wouldn't gain anything by making it more restrictive).

> 1. note how core/README.md still references `Program`/`ProgramInfo`. The `try_into_program()`/`unwrap_program()` shims remain in `mast-package`. Worse: `core/src/program.rs` no longer exists — so the README now references a deleted file.

I'm not sure who is generally responsible for keeping the `core` crate README up to date, but `Program` still exists, so I'm not sure when/how that fell out of sync, I can certainly update it as part of this PR though. While we've made `try_into_program`/`unwrap_program` hidden in preparation for deprecating/removing `Program`, we may actually keep it (it's still useful in some fashion), so I don't think we should remove references to it just yet (but we should update them).

crates/assembly-syntax/src/parser/mod.rs
bitwalker: Sorry about the confusion here - I'm waiting to update project assembly source handling after #3216 lands and I've rebased this PR on it - I didn't want to deal with conflicts between the two branches, as that PR changes source handling considerably. I should've called that out in the description, but forgot to note it there when drafting.

crates/mast-package/src/package/manifest.rs
bitwalker: That's expected - `foo::$main` is absolutely allowed as entrypoint in a package of executable type; but as you observed, such a procedure is rejected in a non-executable package. This is intentional, because in MASM, the `$main` symbol is special, and can only be produced from a module containing a `begin` block - but that module can have any namespace it wants, the rule is that there can be only one entrypoint in the program. I will take another pass through though and make sure we are enforcing that properly at assembly-time.

For manually-constructed AST, one _can_ give that name to any procedure, even in a non-executable module, but it will be rejected when constructing the package - which is what I'd expect us to do with invalid uses of special identifiers.

We do document the rules for valid identifiers, and the expectation from manually-constructed AST is that it adheres to the invariants that parsed AST must adhere to, so I'd expect the output you see here.

crates/assembly/src/assembler.rs
```diff
@@ -725,25 +758,24 @@ impl Assembler {
...
     ) -> Result<Box<Package>, Report> {
-        let options = ParseOptions {
-            kind: ModuleKind::Executable,
-            warnings_as_errors: self.warnings_as_errors,
-            path: Some(Path::exec_path().into()),
-        };
-
-        let program = source.parse_with_options(self.source_manager.clone(), options)?;
-        assert!(program.is_executable());
+        let program = source.parse(self.warnings_as_errors, self.source_manager.clone())?;
```
bitwalker, line 770: Ah, yes, that's a bug - the `$exec` namespace should only be used when `namespace` is elided and an explicit namespace was not given to the assembler.

crates/mast-package/src/package/serialization.rs
```diff
@@ -329,26 +368,48 @@ impl<'de> serde::Deserialize<'de> for PackageManifest {
...
                                 return Err(serde::de::Error::duplicate_field("dependencies"));
                             }
                             dependencies = Some(map.next_value::<Vec<Dependency>>()?);
                         },
+                        Field::Entrypoint => {
+                            if entrypoint.is_some() {
+                                return Err(serde::de::Error::duplicate_field("entrypoint"));
+                            }
+                            entrypoint =
+                                Some(map.next_value::<PathBuf>().map(Arc::<ast::Path>::from)?);
```
bitwalker: Yep, good catch - not sure why that didn't get caught by the tests, I'll have to take a closer look

crates/assembly/src/linker/mod.rs
bitwalker: I think we should probably just reject re-exports of kernel procedures like this, since there is no valid reason to allow it. I'll follow up on this.

crates/assembly/src/assembler.rs
huitseeker: I think this one is still live on `3898c1f` too. Calling 
````rust
Assembler::default().assemble_program("program", "namespace foo\n\nbegin\n    push.1\nend\n")
````
succeeds and prints:

```text
entrypoint=Some("::$exec::$main")
export=::$exec::$main
```

So an executable declared under `foo` is still accepted, but the package surface is published as `$exec::$main`.

crates/mast-package/src/package/serialization.rs
huitseeker: I think this accepted issue is still live on `3898c1f`. A small path-dependency repro that builds a non-executable `PackageManifest`, serializes it with `serde_json`, and reads it back prints:

```text
{"exports":[...],"modules":[],"dependencies":[],"entrypoint":null}
serde_json decode failed: invalid type: null, expected a valid Path/PathBuf at line 1 column 231
```

It looks like the map visitor still reads `entrypoint` as `PathBuf`, while `Serialize` writes the field as an `Option`.

huitseeker (commented):
Thanks a lot for the expansion of docs/README framing around manual vs project assembly, namespace declarations, extern package, module surfaces, and `$main`/`$exec` semantics. The added kernel syscall-only tests, among others, also look great.

crates/assembly-syntax/src/parser/mod.rs
```diff
@@ -156,155 +170,198 @@ fn parse_forms_internal(
...
+    let mut seen = BTreeSet::<Arc<Path>>::new();
+    let mut modules = Vec::new();
+
+    let mut parser = ModuleParser::new(kind);
+    parser.set_warnings_as_errors(warnings_as_errors);
+    let root_ast = parser.parse_file(namespace.as_deref(), &root, source_manager.clone())?;
+
+    let namespace = Arc::<Path>::from(root_ast.path().to_path_buf().into_boxed_path());
+    let submodules = root_ast.submodules().to_vec();
+    seen.insert(namespace.clone());
```
plafer, line 238: `seen` is never inserted after this point, so anything except the root namespace will never be "seen", right? Disregard if this is what you were referring to with https://github.com/0xMiden/miden-vm/pull/3220/changes#r3364590762

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -68,13 +84,56 @@ pub fn analyze(
...
+            },
+            Form::Namespace(ns) if !namespace_allowed => {
+                analyzer.error(SemanticAnalysisError::MisplacedNamespaceDeclaration {
+                    span: ns.span(),
+                });
+            },
+            Form::Namespace(ns) => {
+                if let Some(unused) = docs.as_ref() {
+                    analyzer.error(SemanticAnalysisError::UnusedDocstring { span: unused.span() });
+                }
```
plafer: nit: should we `docs.take()` here so that the docstring doesn't leak into the next item?

crates/assembly/src/linker/namespaces.rs
```diff
@@ -0,0 +1,1324 @@
+// This module is being introduced ahead of the resolver rewrite. Some lookup APIs and definition
```
plafer: How do we make sure that we don't forget this here forever? Are there plans to remove it in a subsequent PR? Otherwise we could either put it on the specific dead code items, or create an issue?

plafer (approved):
LGTM! Though a lightweight review on my end

benches/synthetic-bench/src/calibrator.rs
```diff
@@ -157,16 +157,18 @@ mod tests {
     fn measures_trivial_program() {
         // `measure_program()` already cross-checks our derived totals against the processor's
         // authoritative values; this test just smoke-checks basic measurement.
-        let shape = measure_program("begin push.1 drop end").expect("measure");
+        let shape = measure_program("namespace $exec\n\nbegin push.1 drop end").expect("measure");
```
bobbinth: In this context `$exec` is the only valid namespace, right? Or can we specify something else here?

If it is the only valid namespace, could we omit specifying it? Or would this complicate things?

crates/lib/core/asm/stark/mod.masm
```diff
@@ -1,4 +1,10 @@
-use miden::core::stark::verifier
+pub mod constants
+pub mod deep_queries
+pub mod ood_frames
+pub mod public_inputs
+pub mod random_coin
+pub mod utils
+pub mod verifier
```
bobbinth, line 7: It is not clear to me that any of these need to be public. But I guess we make them public for testing purposes? cc @Al-Kindi-0

crates/lib/core/asm/mod.masm
```diff
@@ -1,3 +1,13 @@
 #! This is the top-level module of the miden::core namespace.
 #!
 #! It is currently empty as we do not export any procedures from here.
```
bobbinth: I would change the phrasing here slightly as this module is no longer empty.

crates/lib/core/docs/collections.md
bobbinth, line 1: Not for this PR, but we should probably update the doc generation logic to add content to files like this. Maybe these should list public modules and skip the empty procedure table.

Let's add it to the list of follow-ups for after this PR.

crates/lib/core/docs/mod.md
```diff
@@ -1,5 +1,3 @@
-This is the top-level module of the miden::core namespace.<br /><br />It is currently empty as we do not export any procedures from here.<br />
```
bobbinth, line 1: Do we no longer support outputting module-level comments?

docs/src/user_docs/assembly/code_organization.md
bobbinth, line 64: Not related to this PR, but the phrase "A procedure can only call itself using dynamic invocation" is somewhat confusing, and I don't think it is actually correct (a procedure can call any other procedure using dynamic invocation).

bobbinth, line 94: Also not related to this PR, but I think we have more built-in attributes now. For example, `@callconv` attribute.

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -146,27 +146,53 @@ A program cannot contain any exported procedures.
...
-2. We can import the `bar` namespace into scope with `use foo::bar`, which allows us to invoke the procedure using a relative path, i.e. `exec.bar::baz`. This approach only requires you to specify the fully-qualified path once to bring a namespace into scope, which reduces visual noise, while retaining a visual cue as to where a symbol reference is defined.
-3. We can import the `baz` item directly into scope with `use foo::bar::baz`, allowing us to treat it like a locally-defined item, e.g. `exec.baz`
+1. We can use the absolute/fully-qualified path to invoke the procedure, i.e. `exec.::foo::bar::baz`. Absolute code paths must be prefixed with `::`.
+2. We can import the `bar` module into scope with `use foo::bar`, which allows us to invoke the procedure using an import-qualified path, i.e. `exec.bar::baz`. Import declarations are resolved from the global module tree, so `foo::bar` does not need a leading `::`.
+3. You can refer to an item which is a descendant of the current module by using submodule-relative paths, indicated by the `self::` prefix, e.g. `use self::child::item`, where `child` is a submodule of the current module.
```
bobbinth: nit: this is the only item in the list where we use "you" rather than "we".

Also, my understanding is that this is mostly intended for bringing children of child modules into scope and "flattening" the module hierarchy - right? If so, I'd maybe change the example to be `use self::child::submodule` or something like that.

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -146,27 +146,53 @@ A program cannot contain any exported procedures.
...
+4. We can import the `baz` item directly into scope with `use foo::bar::baz`, allowing us to treat it like a locally-defined item, e.g. `exec.baz`.
```
bobbinth: Do we need this option? I thought we discussed that importing items would not be supported. Unless supporting this is close to trivial - I'd remove it (e.g., removing a couple hundred lines of code because of this is probably worth it in my mind).

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -182,15 +208,22 @@ begin
...
 #### Re-exporting items
-Items imported into a module, can also be simultaneously re-exported from that module, using the name bound to the symbol when it was imported. For example:
+Items can be imported and simultaneously re-exported from a module with `pub use`. For example:
 ```
-use miden::core::math::u64
-
-pub use u64::add
-pub use u64::mul->mul64
+pub use miden::core::math::u64::add
+pub use miden::core::math::u64::mul->mul64
```
bobbinth: Does this also make `add` and `mul64` available in the current module? For example, would we be able to do `exec.mul64` in the `foo` procedure below?

Overall, my understanding is as follows:

- `use` can import both module and items from modules (procedures, constants etc.). Though, I think the main usage of this would be to import modules (and as I mentioned in one of the above comments, we could even make this purely for module imports).
  - `use self::` can also import both modules and procedures.
- `pub use` can export only items, and I'm assuming that these items also become imported in the current module (i.e., the answer to my question above is "yes").

I guess it is a bit confusing that `use` and `pub use` look pretty similar but the intended targets are different: mostly modules for `use` and only items for `pub use`.

If we do disallow item imports for `use` we could probably change it to something like `use mod` and this would be a more clear syntax, I think:

```
# now we can do something like exec.u64::add
use mod miden::core::math::u64

# This exports `add` and maybe brings it into local scope?
pub use miden::core::math::u64::add
```

docs/src/user_docs/assembly/code_organization.md
```diff
@@ -146,27 +146,53 @@ A program cannot contain any exported procedures.
...
+end
+```
+
+The declaration in the above example tells the assembler that `api` and `internal` are child modules of the current module. Submodules are available by name inside the declaring module's scope as if they had been imported, as demonstrated by `exec.api::foo`.
+
+The visibility on the submodule declaration controls who may resolve paths through that submodule. A private `mod internal` child is visible from the module that declared it and from that module's descendants. Code outside that internal scope must go through public submodule edges, declared with `pub mod`. This module visibility is separate from item visibility: a submodule, regardless of its own visibility, always exposes only the procedures, constants, and types that are themselves public or re-exported.
+
+If a linked module is a child of another module, the parent must declare it. Missing declarations and declarations without a corresponding child module are reported as assembly errors.
+
 #### Importing modules
```
bobbinth, line 168: Do we not need to mention `extern package` syntax somewhere in this section?

miden-vm/tests/integration/cli/data/main.masm
```diff
@@ -1,5 +1,3 @@
-use lib::lib
-
 begin
-    exec.lib::lib_proc
+    exec.::lib::lib_proc
 end
```
bobbinth, line 3: Isn't `lib` here an external dependency? If so, don't we need to use `extern package` to be able to refer to it?

processor/tests/issue_2456_test.rs
bobbinth: As mentioned in one of the above comment, why do we need to specify the namespace for programs? Wouldn't it always be `$exec`? Or do we allow programs to have sub-modules somehow?

crates/debug-types/src/source_manager.rs
```diff
@@ -152,6 +152,17 @@ pub trait SourceManager: Debug {
         let content = SourceContent::new(lang, name.clone(), content);
         self.load_from_raw_parts(name, content)
     }
+    /// Load the given `content` into this [SourceManager] with a name generated from the hash of
+    /// its contents
+    fn load_anonymous(&self, lang: SourceLanguage, content: String) -> Arc<SourceFile> {
+        use alloc::format;
+
+        use miden_crypto::hash::sha2::Sha256;
+        let digest = Sha256::hash(content.as_bytes());
```
bobbinth, line 161: Question: why did we pick SHA256 here? Usually, we use BLAKE3 for hashing strings.

crates/test-utils/src/lib.rs
```diff
@@ -73,6 +73,29 @@ mod test_builders;
...
+pub fn executable_source(source: impl AsRef<str>) -> String {
+    let source = source.as_ref();
+    if source.trim_start().starts_with("namespace ") {
+        source.to_string()
+    } else if source.starts_with('\n') {
+        format!("namespace $exec{source}")
+    } else {
+        format!("namespace $exec\n{source}")
+    }
+}
```
bobbinth: Related to some of the above comments - if we don't require programs to declare the `$exec` namespace, this function can probably go away.

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -150,13 +144,32 @@ pub struct Module {
...
     kind: ModuleKind,
+    /// The namespace given by a `namespace` declaration, if present in this module
+    ///
+    /// The `namespace` declaration is optional when the module being parsed is part of a project
+    /// controlled by a `miden-project.toml` manifest. If present at the same time as a controlling
+    /// manifest, the namespace must agree with what is specified in `miden-project.toml`.
+    ///
+    /// The `namespace` declaration is only permitted in the root module of a project - if present
+    /// in any submodules, an error will be raised.
+    pub(crate) namespace_decl: Option<Span<Arc<Path>>>,
```
bobbinth, line 161: If present, `namespace_decl` should be a prefix of `path`, and otherwise, it would be an error - right? If so, I'd mention this in the comment.

crates/assembly-syntax/src/ast/module.rs
```diff
@@ -231,6 +245,14 @@ impl Module {
         self.span = span;
     }
 
+    /// Registers that `namespace` was explicitly declared for this module via `namespace <ns>`
+    ///
+    /// This will also override the module path to ensure they agree
+    pub(crate) fn set_declared_namespace(&mut self, namespace: Span<Arc<Path>>) {
+        self.path = namespace.to_path_buf();
+        self.namespace_decl = Some(namespace);
+    }
```
bobbinth, line 270: Related to the above comment: it seems like if the `namespace_decl` is set, `path` must contain exactly the same value (which I think makes sense).

If so, could we just use `path` to carry this info and instead of `namespace_decl` use something like `is_root_module` boolean field? This would remove the need to keep `namespace_decl` and `path` synchronized.

crates/assembly/src/assembler.rs
```diff
@@ -276,69 +247,78 @@ impl Assembler {
...
+    /// - ~/masm/core/math/README.md  -> Ignored
     #[cfg(feature = "std")]
-    pub fn compile_and_statically_link_from_dir(
+    pub fn compile_and_statically_link_from_root(
         &mut self,
-        dir: impl AsRef<std::path::Path>,
-        namespace: impl AsRef<Path>,
+        root: impl AsRef<std::path::Path>,
+        namespace: Option<&Path>,
     ) -> Result<(), Report> {
```
bobbinth, line 311: Probably not for this PR, but I wonder if we need this method at all. If we are assembling a project from the filesystem, we should probably require that there is a project file. But I guess it doesn't hurt to keep this for now for backward compatibility.

crates/assembly/src/assembler.rs
```diff
@@ -412,75 +392,46 @@ impl Assembler {
     pub fn assemble_library(
         self,
         name: impl Into<PackageId>,
-        modules: impl IntoIterator<Item = impl Parse>,
+        root: impl Parse,
+        support: impl IntoIterator<Item = impl Parse>,
     ) -> Result<Box<Package>, Report> {
```
bobbinth: Could we update doc comments here to reflect the current signature of this method?

crates/assembly/src/assembler.rs
```diff
@@ -412,75 +392,46 @@ impl Assembler {
...
+    /// discovered and linked from `root`.
     #[cfg(feature = "std")]
-    pub fn assemble_library_from_dir(
+    pub fn assemble_library_from_root(
         self,
-        dir: impl AsRef<std::path::Path>,
-        namespace: impl AsRef<Path>,
+        root: impl AsRef<std::path::Path>,
+        namespace: Option<&Path>,
     ) -> Result<Box<Package>, Report> {
```
bobbinth, line 417: Similar to one of the above comments: I think these `_from_root()` methods could probably go away. If we are building from a filesystem, we should be using the project file. Let's add this to the list of follow-ups.

crates/assembly/src/mast_forest_merger_tests.rs
```diff
@@ -17,15 +18,19 @@ fn merge_programs(
     program_b: &str,
 ) -> Result<(MastForest, MastForest, MastForest, MastForestRootMap), Report> {
     let context = TestContext::new();
-    let module = context.parse_module_with_path("lib::mod", program_a)?;
+    let module = context.parse_module(program_a)?;
 
     let lib_a = Assembler::new(context.source_manager())
-        .assemble_library("lib-a", [module])
+        .assemble_library("lib-a", module, None::<Box<ast::Module>>)
         .map(Arc::from)?;
```
bobbinth, line 25: Is there a way to make this API more ergonomic somehow? Basically, simplify the `None::<Box<ast::Module>>` part.

crates/assembly-syntax/src/parser/mod.rs
```diff
@@ -67,60 +67,74 @@ impl ModuleParser {
     /// Parse a [ast::Module] from `source`, and give it the provided `path`.
     pub fn parse(
         &mut self,
-        path: impl AsRef<Path>,
+        path: Option<&Path>,
         source: Arc<SourceFile>,
         source_manager: Arc<dyn SourceManager>,
     ) -> Result<Box<ast::Module>, Report> {
```
bobbinth, line 81: Could we explain in the comments what happens `path` is not provided? Do we infer the path somehow?

crates/assembly/README.md
```diff
@@ -108,16 +143,26 @@ You can provide a kernel in one of two ways: a precompiled kernel package,
...
+# use miden_assembly::{
+#     Assembler, Path,
+#     ast::{Module, ModuleKind},
+#     debuginfo::{DefaultSourceManager, SourceManager},
+# };
 # use std::sync::Arc;
 #
 # // Create a source manager
-# let source_manager = Arc::new(DefaultSourceManager::default());
+# let source_manager: Arc<dyn SourceManager> = Arc::new(DefaultSourceManager::default());
```
bobbinth, line 152: nit (and not from this PR): I don't think the `#` directive is respected in Markdown files. If so, we can probably remove it from this and other examples.

bobbinth (commented):
Looks good! Thank you! Not a full review yet, but I think I've gone through 2/3 of changes and left some comments inline. The main ones are:

- I think the documentation for how to use `extern package` is missing from user-facing docs.
- I wonder if `assemble_program()` should assume that the namespace is `$exec` (which I think is the only valid namespace for a program). This way, we can avoid adding `namespace $exec` to many tests.
- It feels like the distinctions between `use` and `pub use` are a bit too subtle and maybe it is worth, allowing imports of only modules and changing `use` to `use mod` to make it explicit that we are importing a module (and make it more clear that `pub use` is not applicable to modules).

crates/assembly/src/linker/namespaces.rs
bitwalker: I'll reword this module doc now that the resolver rewrite is complete

benches/synthetic-bench/src/calibrator.rs
bitwalker: Yes it can be omitted in these cases

crates/lib/core/asm/stark/mod.masm
bitwalker: If they don't need to be, we can certainly make them private - I made them public because our previous library visibility semantics left these modules public (one of the benefits of the more explicit structure going forward is that we can actually control visibility much more precisely).

crates/lib/core/asm/mod.masm
bitwalker: Agreed

crates/lib/core/docs/collections.md
bitwalker: Agreed

crates/lib/core/docs/mod.md
bitwalker: We do, I'd have to double check why it's not showing up here

docs/src/user_docs/assembly/code_organization.md
bitwalker: Thanks for noticing the grammar issue there - I'll fix that up.

You can import items or submodules this way, so the example here was intended to show how you'd bring an item defined in a submodule into scope. Of course the same thing can be used to bring a descendant submodule into scope as well.

What you can't do is re-export submodules - that rule is the same regardless of where the imported module is defined.

docs/src/user_docs/assembly/code_organization.md
bitwalker: It is trivial to support - essentially zero complexity added relative to only allowing imports of modules. It also makes `use` vs `pub use` consistent with regard to items (the only restriction is `pub use` of modules - you can `use` them, but not `pub use` them).

It is the removal of the ability to re-export module aliases that results in the biggest simplification to the resolution semantics. The explicit module hierarchy and the specification that `use` is always relative to the global namespace unless prefixed with `self::`  is the next most impactful change. There is very little to be gained from further restrictions, since those rules already make resolution very simple.

crates/lib/core/asm/stark/mod.masm
Al-Kindi-0: Probably testing was the main driver, but is that not a problem any longer ? cc @bitwalker

docs/src/user_docs/assembly/code_organization.md
bitwalker: > Does this also make add and mul64 available in the current module? For example, would we be able to do exec.mul64 in the foo procedure below?

Yes.

> `use` can import both module and items from modules (procedures, constants etc.). 

Correct

> Though, I think the main usage of this would be to import modules (and as I mentioned in one of the above comments, we could even make this purely for module imports).

I think the next most common use would be importing constants and types into scope. Importing procedures directly is likely to be rare (outside of re-exports) - but supporting it is zero extra effort.

> `pub use` can export only items, and I'm assuming that these items also become imported in the current module

Yes. `pub use` is an import statement, but the visibility of the imported symbol is public, in effect "re-exporting" the imported symbol.

> I guess it is a bit confusing that `use` and `pub use` look pretty similar but the intended targets are different: mostly modules for use and only items for pub use.

I don't think most people see `use` as only for modules, I know I don't. I agree that it is unfortunate that `pub use` cannot target modules as well - but we cannot support that regardless without bringing back a lot of complexity to symbol resolution. It is _more_ confusing for `pub use` to be allowed for items, but `use` to support only modules, or modules and items.

This was one of the reasons why I proposed the syntax I did in the original proposal to make what is imported distinct from where it is imported from, i.e. `use {foo, bar} from baz::qux`. This would allow us to disambiguate that `foo` and `bar` are exported symbols from `baz::qux`. The same thing in our current syntax (which is largely preserved in this PR) would be `use baz::qux::foo`, which is ambiguous - the path could refer to a module `baz::qux::foo`, an item `foo` in a module `baz::qux`, or a re-exported module `foo` in `baz::qux` (which requires us to then recursively expand `foo` in the context of `baz::qux`). The restrictions on re-exporting modules via `pub use` in this PR are thus intended to remove the recursive expansion requirement, as we can guarantee that an import statement _must_ refer to either an exact module path, or an item exported from an exact module path.

If we were to use my originally proposed syntax, then we could support both items and modules with `use` and `pub use` equally.

> If we do disallow item imports for `use` we could probably change it to something like `use mod` and this would be a more clear syntax, I think:

I'm not sure what this gains us - the biggest issue is re-exported module aliases which introduce ambiguity that cannot be disambiguated with the current syntax, nor with the restriction you've suggested.

The additional restriction I've imposed in this PR that imports cannot refer to other imports does remove some of the worst edge cases around resolution - but it is the ambiguity described above (and the need for recursive expansion) that introduces the bulk of the complexity to symbol resolution.

crates/assembly/src/assembler.rs
bitwalker: These are still used internally by the project assembler, but yes, we probably wouldn't need to export them, except for testing.

crates/assembly/src/mast_forest_merger_tests.rs
bitwalker: Possibly - I'd have to check if we really need to support `Parse` here, or if we can just require an iterator of `Box<ast::Module>`, in which case the disambiguation with turbofish syntax wouldn't be needed.

crates/assembly-syntax/src/parser/mod.rs
bitwalker: It is inferred from  the `namespace` declaration (if present), with a fallback in case the module contains `begin .. end`, in which case we'll infer the module name as `$exec` - otherwise we produce an error indicating that the namespace was not declared or explicitly provided.

I can expand the docs here.

docs/src/user_docs/assembly/code_organization.md
bitwalker: I intentionally omitted it from user docs, because all users should need to know about is the project assembly path - they have zero reason to be using anything else. Instead, it is covered in the `miden-assembly` README file for now.

We can certainly add it to the user docs, but since the use case for `namespace`/`extern package` is in places where Miden projects are not being used, and we're trying to consolidate around MIden projects, I felt it was better to keep that syntax less visible for now.

miden-vm/tests/integration/cli/data/main.masm
bitwalker: The purpose of `extern package` is to inform the linker that the given package is expected/referenced, and to raise an error if it is not. If you omit the `extern package` then you must manually ensure that all external dependencies are linked, otherwise you'll get an undefined symbol error.

We could be stricter about this once we make the assembler require `std` - but for now `extern package` is basically a hint to the linker.

processor/tests/issue_2456_test.rs
bitwalker: We don't - it can be inferred in these cases, but I had made this change before supporting inference in the case of `$exec`.

crates/debug-types/src/source_manager.rs
bitwalker: No specific reason other than it is a common hash algorithm for purposes like this. The specific hash algorithm doesn't really matter so long as it is fast and relatively collision-resistant, that's good enough. If we prefer BLAKE3 for this, that's an easy change.

crates/test-utils/src/lib.rs
bitwalker: Yes, it can be removed

crates/assembly-syntax/src/ast/module.rs
bitwalker: No, if present `namespace_decl` and `path` must be the same, and that's validated during parsing/semantic analysis. I can mention that here though.

crates/assembly-syntax/src/ast/module.rs
bitwalker: The AST needs to represent when the `namespace` declaration was present, versus when it was inferred/explicitly set - it must be tracked distinctly from `path`, as `path` can be set to an explicit value separate from the `namespace` declaration, and we catch this during semantic analysis.

This method here is called _after_ we've validated that a `namespace` declaration doesn't conflict - but it sets both values, because we initially instantiate the module with an empty `path` (if an explicit path wasn't provided to the parser), so this just ensures that both values match, regardless of whether `path` is the empty path, or an explicitly-provided path that happens to already be equal to the declared namespace.

crates/assembly/src/assembler.rs
bitwalker: Yes, with project assembly, this function doesn't have much use. IIRC, it isn't used by the project assembler either, so it solely exists for backwards-compatibility for the moment.

crates/lib/core/asm/stark/mod.masm
bitwalker: If we need to construct test programs that call functions in these modules directly, then yes, they'll need to be public for now.

crates/mast-package/src/package/serialization.rs
bitwalker: I hadn't patched this yet

crates/assembly-syntax/src/parser/mod.rs
bitwalker: You are correct - fixed in 93f4e67f

crates/assembly-syntax/src/ast/module.rs
bobbinth: Should we add an `assert!` here then? Basically, if the `path` is set and it is not equal to the incoming `namespace` - that should be an error - right?

docs/src/user_docs/assembly/code_organization.md
bobbinth: I think the current semantics of `pub use` are fine - i.e., we don't need to change anything about it.

My point was mostly about the `use` semantics. As you've mentioned, it is somewhat ambiguous in terms of what we are trying to import - i.e., an item or a module. I think there are two broad ways to approach this:

1. Restrict `use` to only import modules.
2. Change the syntax for importing modules vs. items.

If restricting `use` to only work with modules doesn't reduce complexity, probably doesn't make sense to pursue this option.

Regarding changing the syntax, we also have a couple of options. What you suggested:

```
use {foo, bar} from baz::qux
```
If we go this route, then we should also probably do the same for `pub use` - i..e.,

```
pub use {foo, bar} from baz::qux
```


Or maybe something like this:
```
# imports an item and is consistent with `pub use`
use foo::bar::baz

# imports the full module and looks different from use/pub use
use mod foo::bar
```

In choosing between these options, I'd go with w/e reduces the complexity of the code the most because from feature standpoint, I think they are basically the same.

bitwalker: Yes, if we go with the syntax I proposed, the syntax of the `use` form would be the same regardless of whether or not it is `pub`, same as the current syntax, so that the intuition around `pub` is that it changes visibility of the imported items only. The full syntax would then be two forms:

1. Import items with `use {item,..} from module` syntax, `pub` makes imported symbols public (i.e. re-exported)
2. Import modules with plain `use module`. `pub` on this form would be rejected for now, but we could support it in the future easily enough with some changes to the way we extract information from the CST before lowering to AST and performing semantic analysis. We'll need to make such changes to better support LSP queries/refactorings anyway, but due to how the MASM frontend currently works, I don't think it's worth it to allow re-exporting modules yet, as it is awkward to reason about imports/re-exports prior to linking currently, so it just isn't worth it until that situation changes.

That syntax makes it statically detectable whether an import refers to an item or a module, which allows us to better verify whether a symbol reference is valid or not even without computing the full namespace graph.

We could use `use mod` for 2 above, but that's a bit ambiguous grammar-wise (we want all top-level forms to start with a distinct keyword), and wouldn't actually be necessary (i.e. the extra keyword doesn't disambiguate anything that isn't already unambiguous).

crates/assembly-syntax/src/ast/module.rs
bitwalker: Well, no - semantic analysis already handles catching mismatches, but `path` may be an empty path when this is called (in cases where no explicit path was provided to the parser), so we still want to set both `path` and `namespace_decl` here. The only time that `path` would be non-empty when this is called, would be if an explicit namespace was given to the parser _and_ there is a `namespace` declaration in the module source _and_ they are the same path (in which case setting `path` is purely redundant).

We could add an assert that checks that the path is one of those two options though

docs/src/user_docs/assembly/code_organization.md
bobbinth: Let's go then with:

```
# import single item from module
use {foo} from module

# import multiple items from module
use {foo, bar} from module

# import and re-export a single item from module
pub use {foo} from module

# import and re-export multiple items from module
pub use {foo, bar} from module

# import the module
use module

# pub use module, not allowed
```

My main consideration here is extra complexity. I'm fine with doing the simpler approach if it reduces complexity, but if not, the above works perfectly well.

miden-vm/tests/integration/cli/data/main.masm
bobbinth: I'm not sure I fully understood this part.

> If you omit the `extern package` then you must manually ensure that all external dependencies are linked, otherwise you'll get an undefined symbol error.

Does "manually ensure" here mean that we need to link the external dependencies using the assembler's Rust API? If so, don't we need to do it in all cases? That is, if we do include `extern package` we still need to manually link all the external modules - or would the assembler (even if in the future) try to find them somehow?

> We could be stricter about this once we make the assembler require `std`

How would `std` impact things? I'm assuming that if we are using filesystem-based assembly (i.e., with `std`) we would always have a project file to define external dependencies. And my understanding was that `extern package` would be useful only when we don't use filesystem-based assembly (e.g., in tests).

Basically, I thought we have two distinct "worlds":

- Assembling from filesystem and then we must have a project file. And in this case, we don't need `extern package`.
- Assembling manually, and here we don't have a project file and need to use `extern package`. But, IIUC, you are saying that in this case, we don't really need `extern package`?

crates/debug-types/src/source_manager.rs
bobbinth: I'd switch to BLAKE3 here for consistency as I don't think we use SHA256 for such purposes anywhere.

crates/assembly/src/assembler.rs
bobbinth: Let's add this to the list of assembler cleanups that we'd want to do in a follow-up PR.

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -68,13 +83,56 @@ pub fn analyze(
...
                     analyzer.error(SemanticAnalysisError::UnusedDocstring { span: unused.span() });
                 }
+                namespace_allowed = false;
+            },
+            Form::Namespace(ns) if !namespace_allowed => {
+                analyzer.error(SemanticAnalysisError::MisplacedNamespaceDeclaration {
+                    span: ns.span(),
+                });
+            },
+            Form::Namespace(ns) => {
```
huitseeker, line 93: Could we set `namespace_allowed = false` after accepting the first namespace declaration?

I can get a silent overwrite with:

```masm
namespace first
namespace second

pub proc helper
    push.1
end
```

That parses as `::second` instead of reporting the second `namespace` as misplaced.

crates/assembly/README.md
```diff
@@ -58,11 +61,41 @@ will want to either use projects (see `miden-project` for more info), or factor
...
+use miden::core::math::u64
+
+pub proc add64
+    exec.u64::wrapping_add
+end
+```
+
+The `namespace` declaration gives the module its fully-qualified path. It is optional when the caller or project manifest already provides the module path, but if both are present they must agree. It must appear before any item, import, submodule, or external package declaration.
+
+The `extern package "<package-id>"` declaration records an external package dependency for a standalone root module. It is optional when the module is assembled as part of a project, because the manifest supplies the dependency set. When reading a module tree from a root source file, `namespace` and `extern package` declarations belong in that root source file.
```
huitseeker: This reads stronger than the code today. I can parse and assemble a standalone root with:

```masm
namespace depuser
extern package "miden-core@0.1.0"

pub proc foo
    push.1
end
```

but the resulting package manifest still has zero dependencies. I only see `extern_packages` recorded on the AST module; it does not seem to feed `PackageManifest::dependencies()` or get checked against project dependencies yet.

Should this paragraph say the declaration is parsed but not wired into package metadata yet?

miden-vm/src/cli/bundle.rs
```diff
@@ -44,46 +44,44 @@ impl BundleCmd {
...
                     return Err(Report::msg("`kernel` must be a file"));
                 };
                 assembler.link_package(CoreLibrary::default().package(), Linkage::Dynamic)?;
-                let library =
-                    assembler.assemble_kernel_from_dir(namespace, kernel, Some(&self.dir))?;
+                let namespace = match self.namespace.as_deref() {
+                    Some(ns) => ns,
+                    None => ast::Path::KERNEL_PATH,
+                };
+                let library = assembler.assemble_kernel_from_root(namespace, kernel)?;
```
huitseeker: In the `--kernel` branch, `self.root` is validated above but does not appear to be used. The command docs say this builds a kernel library from `kernel` using the library of `root` as the kernel namespace, but the implementation only passes `kernel` to `assemble_kernel_from_root`.

Can we either use `root` in kernel mode ?

bitwalker: That option is defunct now that submodules are explicit (the purpose it served was to assemble a kernel with its support modules excluded from the interface of the kernel - which can now be expressed using submodule declarations directly).

I'll make sure it's removed.

crates/assembly/README.md
bitwalker: Yes, as noted in my earlier comments, `extern package` is currently just a hint to the linker - but this PR does not take full advantage of it yet (there are other changes needed before we can do so).

I can add a note some of this is pending future work.

docs/src/user_docs/assembly/code_organization.md
bitwalker: This has been implemented now. I'm pretty happy with the simplifications that resulted from it. It also helped me identify an area where I had not properly handled resolution around re-exported items. The change here landed in a single commit titled `syntax: distinguish item vs module import syntax` if you want to see what changed in the assembler as a result of this.

crates/assembly-syntax/src/ast/ident.rs
```diff
@@ -215,7 +215,18 @@ impl AsRef<str> for Ident {
 
 impl fmt::Display for Ident {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Display::fmt(&self.name, f)
+        if Self::requires_quoting(&self.name) {
+            write!(f, "\"{}\"", &self.name)
```
huitseeker: Nit: `Ident::new("a\"b")` is valid today, but this now prints `"a"b"`, and using that as MASM source like `proc "a"b"` fails with a syntax error. I'd escape the identifier before adding quotes. The string/value printers use `escape_debug()` for this kind of case, so this probably needs the same treatment before the quotes are added.

crates/assembly/src/assembler.rs
```diff
@@ -519,19 +461,25 @@ impl Assembler {
...
+        let namespace = Some(Path::KERNEL.into());
+        let (root, support) = miden_assembly_syntax::parser::read_modules_from_root(
+            sys_module_path,
+            namespace,
+            Some(ModuleKind::Kernel),
+            self.source_manager.clone(),
+            self.warnings_as_errors,
+        )?;
+
+        self.linker.link_modules(support)?;
```
huitseeker: In a root kernel with `pub mod sys` and a public proc in `sys.masm`. `assemble_kernel_from_root` loads the support module here, but `link_kernel` still returns only the root module, so a child-only kernel fails with `package must contain at least one exported procedure`; if I add a root proc, the package exports only `::$kernel::root` and drops `::$kernel::sys::helper`.

The `.masm` kernel paths in `run`/`prove` and `verify` also parse just the root file and fail earlier with `undefined module '::$kernel::sys'` on the same case.

Could this include the public submodules in the kernel package surface?

bitwalker: I would expect the empty kernel with a non-empty submodule to still fail assembly (since the kernel would define no syscalls), but the latter issue is unexpected. Agreed that we probably need to be returning the full kernel package surface now that visibility is modeled properly (and we diagnose re-exporting procedures from the kernel module as not allowed).

crates/assembly/src/linker/resolver/symbol_resolver.rs
```diff
@@ -196,6 +331,7 @@ impl<'a> SymbolResolver<'a> {
                 resolution => Ok(resolution),
             },
         }?;
+        let resolution = self.ensure_procedure_target(context, resolution)?;
```
huitseeker, line 334: I think this still misses compiled kernel submodules. I can build a kernel package with:

```masm
pub mod io
```

and `::$kernel::io` exporting `pub proc k`, then assemble a userspace program with:

```masm
begin
    exec.::$kernel::io::k
end
```

That succeeds today, but it should fail like the root-kernel `exec.::$kernel::k1` case. The guard below only treats `kernel_index` or `ModuleKind::Kernel` as kernel code; submodules loaded back from the kernel package are MAST modules under `$kernel::*`, so they look like normal library modules.

bitwalker: I disagree - why should that fail? Submodules of the kernel don't define syscalls, only the procedures exported from the kernel root are syscalls - or at least that's how we've defined it up until now. The kernel carrying library functions that can be exec'd is reasonable IMO

huitseeker: OK, I was conflating “exported from a kernel package” with “is a syscall”, fair. I'd love a few tests / spectacularly-visible doc to help mentally keep track of:
 
- `exec.::$kernel::k1` fails: userspace code must not compile if it tries to call a root kernel export with exec.
- `syscall.::$kernel::k1` works: userspace code should compile, and the call should lower to a valid syscall to the root kernel export.
- `exec.::$kernel::io::k` may work: if io is a public kernel submodule, userspace code may compile when calling that support function with exec.
- `syscall.::$kernel::io::k` fails: userspace code should not compile if it tries to treat a kernel submodule export as a syscall, because only root `$kernel` exports define syscalls

huitseeker (approved):
LGTM, thanks a lot!

crates/assembly/src/linker/resolver/symbol_resolver.rs
bitwalker: Agreed, now that we actually support exporting library functions from the kernel, we need documentation for this.

I have added/verified that we have tests that assert that re-exporting procedures from the kernel library is rejected; and that we can assemble kernels with public submodules that are accessible with `exec`. I'll confirm that we also have tests covering `exec` vs `syscall` behaviors outlined above, and add documentation around these constraints in the user docs.

Issue #2980, opened 2026-04-08
Track mainnet MastForest capabilities: sparse representations and subset checks


huitseeker:
## Background

The roadmap puts sparse MastForest and subset checks in scope for mainnet, while explicitly keeping full set reconciliation out of scope.

## Existing Sub-Issues

- https://github.com/0xMiden/miden-vm/issues/2766

## Missing Child Issues That Should Likely Be Opened Under This Umbrella

- [ ] Create sub-issue `Implement MastForest subset-check API`

## Scope

- Track the sparse-MastForest capability already represented by existing work
- Define subset-check semantics for MastForest
- Decide where the API belongs
- Add the minimum implementation needed for mainnet
- Document non-goals, especially full reconciliation / diff workflows

bobbinth:
Closing as I think this is now done - but if not, @huitseeker, please re-open.

PR #3221, opened 2026-06-02
Split debug info out of MastForest

### LINKED ISSUE #2716: Deduplicate code in debug var, `AsmOp` and other storages
Check: https://github.com/0xMiden/miden-vm/pull/2471#discussion_r2815878167

Basically we need to try to reduce code duplication as much as possible in the storages recently added.

cc @huitseeker @bobbinth @bitwalker

### LINKED ISSUE #2929: Refactor handling of debug information

### LINKED ISSUE #2961: Debug vars lost during block deduplication and merging

### LINKED ISSUE #2982: DebugInfo Metadata follow-ups

### LINKED ISSUE #2990: Static linking by digest loses per-alias metadata when multiple roots share the same digest

### LINKED ISSUE #3040: Separate debug information from MAST entirely

huitseeker:
This PR makes `MastForest` an execution-only data structure. Debug metadata now lives in package debug sections. Package debug info is the trusted read and write path for source metadata.

The source debug graph records source MAST nodes separately from execution MAST nodes. This lets diagnostics and debug variables tell apart two source occurrences that compile to the same executable node.

Fixes:

- Closes #2716: removes duplicated debug metadata finalization paths.
- Closes #2929: package debug info has one owner and one trust boundary.
- Closes #3040: production debug info is no longer owned by `MastForest`.
- Closes #2961: debug vars survive MAST node dedup without source occurrence ambiguity.
- Closes #2990: static imports keep the right source metadata for same-digest aliases.
- Closes #2982: debug metadata merges are covered by regression tests.
- Closes 0xMiden/security-issues#16: normal package file reads validate embedded MAST node digests before use.

Main changes:

- Remove legacy `core::mast::DebugInfo` and make `MastForest::merge` execution-only.
- Add package debug sections for the source graph, source map, error messages, and the other package debug tables.
- Key assembly op, debug variable, and inline-call metadata by source node.
- Preserve debug on trusted package reads, discard it on untrusted reads, and validate embedded MAST forests on normal package reads.
- Restore source-rich processor diagnostics and add regression tests for debugless execution, source occurrence disambiguation, static import aliases, embedded kernel validation, and diagnostic recovery.

<details><summary>Diff split</summary>

Compared:

```text
origin/next...HEAD
```

| Category | Added | Deleted | Net |
|---|---:|---:|---:|
| Non-test | 8643 | 5332 | +3311 |
| Test | 1954 | 1566 | +388 |
| Bench | 0 | 0 | 0 |
| Total | 10597 | 6898 | +3699 |

The previous diff table was measured around `349bd0d571` (`docs: clarify unchecked package reader trust boundary`). Since then, the branch commits add `1583` lines, delete `222` lines, and add `1361` net lines.

The visible PR diff grew from the old table by `1497` added lines, `134` deleted lines, and `1363` net lines.

</details>

Web intro/prez: https://gisthost.github.io/?ea3e248b3b47023962082e99218b5382


processor/src/execution/operations/crypto_ops/mod.rs
```diff
@@ -96,7 +96,7 @@ pub(super) fn op_hperm<P: Processor, T: Tracer>(
 pub(super) fn op_mpverify<P: Processor, T: Tracer, F>(
     processor: &mut P,
     err_code: Felt,
-    program: &F,
+    _program: &F,
```
plafer: let's remove. Applies to other operation files as well (e.g. same with `op_assert()`)

processor/src/execution/join.rs
```diff
@@ -34,9 +34,24 @@ where
...
-    state.continuation_stack.push_start_node(join_node.second());
-    state.continuation_stack.push_start_node(join_node.first());
+    let first_source_node = match state.child_source_node(0) {
+        Ok(source_node) => source_node,
+        Err(err) => return ControlFlow::Break(BreakReason::Err(err)),
+    };
+    let second_source_node = match state.child_source_node(1) {
+        Ok(source_node) => source_node,
+        Err(err) => return ControlFlow::Break(BreakReason::Err(err)),
+    };
```
plafer, line 44: We now get the children ids from the debug info (based on how `child_source_node()` is implemented); so a program without debug info fails to execute, right? Successful execution must not be dependent on whether debug info is present or not. This applies to all other node types

plafer (changes requested):
Unless I missed something, I think successful execution is now dependent on debug info being present.

processor/src/execution/join.rs
huitseeker: The code you're commenting on asks for source children (arguably confusing naming — I'm open to suggestions). This does not require debug info. The return type is:

```rust
Result<Option<DebugSourceMastNodeId>, ExecutionError>
```

When debug info is absent, `child_source_node()` returns `Ok(None)`:

```rust
let Some(source_debug_info) = self.source_debug_info else {
    return Ok(None);
};
let Some(current_source_node) = self.current_source_node else {
    return Ok(None);
};
```

The executable children still come from the join node:

```rust
push_start_node_with_source(join_node.second(), second_source_node);
push_start_node_with_source(join_node.first(), first_source_node);
```

The first argument is the executable child. The second argument is an optional source sidecar. `None` means there is no source sidecar.

`split`, `loop`, and `call` use the same shape. They choose executable children from the executable node:

- `split_node.on_true()` / `split_node.on_false()`
- `loop_node.body()`
- `call_node.callee()`


`child_source_node()` can fail only when debug-aware execution is active and the trusted debug graph is malformed.

huitseeker (commented):
@plafer There are two paths into the processor.

Normal execution creates a plain continuation stack and passes no package debug info:

```rust
execute_sync(...)
    -> execute_with_tracer_sync(...)
    -> ContinuationStack::new(program)
    -> execute_impl(..., None)
```

Debug-aware execution opts in to package debug info:

```rust
execute_with_package_debug_info_sync(...)
    -> source_aware_continuation_stack(...)
    -> execute_impl(..., Some(package_debug_info))
```

Both paths reach the same node dispatcher. The difference is only whether source sidecars are present.

```mermaid
flowchart TD
    normal_api["execute_sync<br/>execute_with_tracer_sync"]
    normal_stack["ContinuationStack::new(program)<br/>source_nodes = None"]
    normal_loop["execute_impl(..., None)"]

    debug_api["execute_with_package_debug_info_sync"]
    debug_stack_builder["source_aware_continuation_stack(...)"]
    root_lookup{"unique source root?"}
    debug_stack["ContinuationStack::new_with_source_node(...)<br/>source_nodes = Some(...)"]
    debug_loop["execute_impl(..., Some(package_debug_info))"]

    dispatch["execution::execute_impl<br/>dispatch by MastNode kind"]
    child_nodes["join / split / loop / call"]

    normal_api --> normal_stack --> normal_loop --> dispatch
    debug_api --> debug_stack_builder --> root_lookup
    root_lookup -- "no" --> normal_stack
    root_lookup -- "yes" --> debug_stack --> debug_loop --> dispatch
    dispatch --> child_nodes
```

The child-bearing nodes call `child_source_node()`, then push executable children with optional source sidecars. The executable child is not taken from debug info.

```mermaid
flowchart TD
    start["join / split / loop / call start"]
    exec_child["executable child from MastNode"]
    source_lookup["state.child_source_node(child_index)"]
    has_debug{"debug info and current source node?"}
    no_source["Ok(None)"]
    debug_lookup["PackageDebugInfo::child_source_node(...)"]
    source_ok["Ok(Some(source_child)) or Ok(None)"]
    source_err["ExecutionError::Internal"]
    push_plain["push_start_node_with_source(exec_child, None)"]
    push_debug["push_start_node_with_source(exec_child, Some(source_child))"]

    start --> exec_child
    start --> source_lookup --> has_debug
    has_debug -- "no" --> no_source --> push_plain
    has_debug -- "yes" --> debug_lookup
    debug_lookup -- "valid or absent" --> source_ok --> push_debug
    debug_lookup -- "malformed" --> source_err
    exec_child --> push_plain
    exec_child --> push_debug
```

With that said, I've added `untrusted_debug_stripped_child_bearing_package_executes_without_debug_info` to isolate this case that was perhaps less clear in the code.

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
```diff
@@ -0,0 +1,150 @@
...
+
+impl Idx for SourceMastNodeId {}
+
+/// Finalized source/debug occurrence for a reduced execution node.
+#[derive(Clone, Debug)]
+pub(crate) struct SourceMastNode {
+    exec_node: MastNodeId,
+    children: Vec<SourceMastNodeId>,
+    op_start: usize,
+    op_end: usize,
```
plafer, line 24: It's not clear to me what these are (at least on a first read). Can we explain in a docstring?

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
```diff
@@ -0,0 +1,150 @@
...
+use miden_core::{
+    mast::MastNodeId,
+    operations::{AssemblyOp, DebugVarInfo},
+    utils::{Idx, IndexVec},
+};
+
+/// Final dense ID for a source/debug occurrence of a MAST node.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(transparent)]
+pub(crate) struct SourceMastNodeId(u32);
```
plafer: All new id types in this PR don't use the macro we have; any reason for it? We should either use that macro everywhere, or get rid of it IMO.

Also, is `SourceMastNodeId`/`DebugSourceMastNodeId` the best names? I would probably keep the "MAST node ID" terminology for the actual `MastForest`, and use something else for the debug graph; all the variations on `MastNodeId` are a bit confusing. How about something along the lines of `DebugGraphId`?

processor/src/execution/dyn.rs
```diff
@@ -103,14 +103,22 @@ where
...
 
     // if the callee is not in the program's MAST forest, then we need to break to allow the
     // implementing processor to fetch it (possibly asynchronously in an external library in the
     // host).
     match current_forest.find_procedure_root(callee_hash) {
         Some(callee_id) => {
-            state.continuation_stack.push_start_node(callee_id);
+            let source_node = match state.source_debug_info {
+                Some(source_debug_info) => {
+                    source_debug_info.unique_source_root_for_exec_node(callee_id).unwrap_or(None)
```
plafer: Why do we silently drop the error with `unwrap_or(None)` here? Similarly for `finish_load_mast_forest_from_external()`

processor/src/fast/execution_api.rs
```diff
@@ -214,6 +414,41 @@ impl FastProcessor {
...
+        program: &Program,
+        package_debug_info: &PackageDebugInfo,
+    ) -> Result<ContinuationStack<Arc<MastForest>>, ExecutionError> {
+        let Some(source_node) = package_debug_info
+            .unique_source_root_for_exec_node(program.entrypoint())
+            .map_err(|_| {
+                ExecutionError::Internal(
+                    "package debug source graph has ambiguous or malformed entrypoint roots",
+                )
+            })?
```
plafer, line 603: Thinking out loud here - do we actually want to fail execution when debug info is malformed? Another solution could be to remember that it's malformed, and if an error occurs, output some "debug info was malformed" message. Though that might result in complexity we don't want.

processor/src/execution/external.rs
```diff
@@ -84,11 +89,24 @@ where
 
     tracer.record_mast_forest_resolution(resolved_node_id_new_forest, &new_mast_forest);
 
+    let source_node =
+        match (source_debug_info, old_forest.get_node_by_id(resolved_node_id_new_forest)) {
+            (Some(source_debug_info), Some(resolved_node_old_forest))
+                if !resolved_node_old_forest.is_external()
+                    && resolved_node_old_forest.digest() == external_node_old_forest.digest() =>
```
plafer: Looking up an ID from the new forest `resolved_node_id_new_forest` into the old forest `old_forest` is confusing, and I'm still struggling to wrap my head around why this works. Can we simplify this expression, and add a comment explaining what we're doing?

crates/assembly/src/mast_forest_builder/finalizer.rs
```diff
@@ -147,61 +165,15 @@ impl MastForestFinalizer {
...
-        mut self,
+        self,
         procedure_root_refs: &[MastNodeRef],
+        procedure_source_root_refs: &[SourceMastNodeRef],
+        source_nodes: &IndexVec<SourceMastNodeRef, PendingSourceMastNode>,
+        asm_op_by_ref: &IndexVec<AsmOpRef, AssemblyOp>,
+        debug_vars: &IndexVec<DebugVarRef, DebugVarInfo>,
         advice_map: AdviceMap,
-        error_codes: BTreeMap<u64, Arc<str>>,
+        _error_codes: BTreeMap<u64, Arc<str>>,
```
plafer: Remove

processor/src/tests/mod.rs
```diff
@@ -386,13 +411,10 @@ fn test_diagnostic_failed_assertion() {
...
-        regex!(r#",-\[test[\d]+:4:13\]"#),
-        " 3 |             push.1.2",
-        " 4 |             assertz.err=\"some error message\"",
-        "   :             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^",
-        " 5 |             push.3.4",
-        "   `----",
+        format!(
+            "  x assertion failed with error code: {}",
+            error_code_from_msg("some error message")
+        ),
```
plafer: This is a regression in output; we used to get the nice error message, and we now only get the error code. Is this intended? Applies to other tests in this file

crates/assembly/src/assembler.rs
```diff
@@ -920,28 +960,31 @@ impl Assembler {
...
 
     fn apply_debug_options(
         &self,
         mast_forest: miden_core::mast::MastForest,
     ) -> miden_core::mast::MastForest {
-        if !self.emit_debug_info {
-            return mast_forest.without_debug_info();
-        }
+        mast_forest
+    }
```
plafer: Remove?

core/src/mast/serialization/mod.rs
```diff
@@ -251,20 +232,18 @@ impl Serializable for MastForest {
...
     /// Internal serialization with options.
     ///
-    /// When `stripped` is true, the DebugInfo section is omitted and the FLAGS byte
-    /// has bit 0 set.
+    /// Current writers always omit the legacy DebugInfo section and set the STRIPPED flag.
     fn write_into_with_options<W: ByteWriter>(
         &self,
         target: &mut W,
-        stripped: bool,
+        _stripped: bool,
```
plafer: Remove `_stripped` parameter, instead of commenting that it's ignored?

processor/src/execution/mod.rs
```diff
@@ -148,9 +188,14 @@ where
...
         stopper,
+        source_debug_info,
+        current_source_node: None,
     };
 
-    while let Some(continuation) = state.continuation_stack.pop_continuation() {
+    while let Some((continuation, source_node)) =
+        state.continuation_stack.pop_continuation_with_source()
+    {
+        state.current_source_node = source_node;
```
plafer: On my machine, I'm getting a ~10% performance reduction to the `program_execution_fast/blake3_1to1` benchmark (a bit noisy, but roughly from 2.90ms up to 3.1-3.2ms).

Since the benchmark strips away the debug info, we can attribute the regression to the extra code that gets executed even when debug info is stripped away (still investigating a more precise cause). The strategy we took previously with the `Tracer` and `Stopper` is to use generics to avoid injecting any code that is not taken (e.g. using the `NoopTracer` makes all `tracer` calls vanish at compile time). I wonder if we can reuse this idea with debug info, e.g. either with a `SourceTracking`/`NoSourceTracking` marker, or a `const DEBUG_INFO: bool` generic instead of passing `source_debug_info = None`?

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
huitseeker: ~Yep, my bad, I removed unused operation arguments where the argument was not part of a required trait or generic hook. Cases where the signature is required by the shared hook shape are left in place.~

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
huitseeker: I used newtype_id! where it fits. I extended the macro so builder-local IDs can use it with restricted visibility, and converted the builder-local ID wrappers to it.

I also renamed the debug graph node IDs away from *`MastNode`* naming. The package debug graph now uses `DebugSourceNodeId `/ `DebugSourceNode`, and the builder uses `SourceNodeId` / `SourceNodeRef` / `PendingSourceNode`.

The serialized public package ID stays custom because it has explicit binary codec and serde behavior. I added a doc comment that calls that out.

processor/src/execution/dyn.rs
huitseeker: `Ok(None)` still means there is no source sidecar.  I fixed things so `Err(_)` from the trusted package debug graph no longer gets swallowed. 

Dynamic calls now fail closed with an internal execution error when the debug graph is ambiguous or malformed.

processor/src/fast/execution_api.rs
huitseeker: I agree with your "failing closed" suggesiton here. Normal execution without package debug info is unaffected. A caller that passes package debug info has opted into trusted debug metadata. If that metadata is malformed or ambiguous, continuing would risk reporting the wrong source context.

processor/src/execution/external.rs
huitseeker: I removed the cross-forest lookup. The old code indexed the old forest with a node ID from the newly loaded forest, which was hard to justify.

The external target still executes from the newly loaded forest. The source sidecar is now looked up only through the package debug graph for the resolved target root. A missing mapping is allowed. An ambiguous or malformed mapping fails closed (I added a test for that).

processor/src/tests/mod.rs
huitseeker: good catch, mostly a matcher regression — will address.

huitseeker: Thanks for the push, this is now fixed. At the same time, 
- `CryptoError::Operation` now applies package debug info when source context is present.
- `mtree_verify.err` has execution-level coverage for actual MerklePathVerificationFailed.
- Package debug info can now restore error messages even when source-node/span lookup is unavailable.

processor/src/execution/mod.rs
huitseeker: I can reproduce a regression, but nowhere near that magnitude on my machine (I get a regression of ~1.5%), see the repro details at https://gist.github.com/huitseeker/864d1f85bbd61646e3d784a74e407de7

huitseeker: It's even less conclusive on my x86_64 machine, sleipnir (which is less noisy than my Mac as it's a headless machine where nothing else runs):

On sleipnir I tried both before/after this branch, as well as creating a pure debug-less dispatch, an executor that does not take as argument nor has a single LoC related to debug:

  
| Case | Commit | Run 1 mean | Run 2 mean |
| --- | --- | ---: | ---: |
| merge-base with `origin/next` | `2a5dfc0d6` | `3.4926 ms` | `3.4502 ms` |
| branch tip of issue-3040-split-debuginfo  | `a2d89b699` | `3.2020 ms` | `3.2020 ms` |
| direct pure dispatcher | [`7ded1746c`](https://github.com/0xMiden/miden-vm/commit/7ded1746c660ef03d2e1b67c1b23c26beda92951) | `3.0101 ms` | `3.0045 ms` |
  
  So:

Putting DebugInfo in the package did not make this benchmark slower than origin/next on sleipnir. The original local +1% slowdown was likely machine noise or local-state noise.

core/tests/ui/mast_forest_immutable/add_debug_var.stderr
bobbinth, line 1: Why do we need this file and the associated test? Trying to call a non-existent method on the `MastForest` should be a compile error - right? Or am I missing something?

core/tests/ui/mast_forest_immutable/debug_info_mut.stderr
bobbinth, line 1: Same question as above.

core/src/mast/node/mod.rs
```diff
@@ -66,7 +66,7 @@ pub trait MastNodeExt {
     #[cfg(debug_assertions)]
     fn verify_node_in_forest<F>(&self, forest: &F)
     where
-        F: ExecutableMastForest + ?Sized;
+        F: crate::mast::ExecutableMastForest + ?Sized;
```
bobbinth: Is this method actually needed? Unless I missed something, it seems like all implementations of this method have empty bodies.

core/src/mast/mod.rs
bobbinth, line 362: Seems like all methods in this `impl` block are relevant only for tests. It would be a bit cleaner to put the whole block behind the `#[cfg(test)]` gate (and change the name of the section to "Test Helpers").

bobbinth, line 254: nit: I would move this method into the "Test Helpers" section mentioned in the above comment.

core/src/mast/mod.rs
```diff
@@ -330,19 +286,16 @@ impl MastForest {
...
     pub fn compact(self) -> (MastForest, MastForestRootMap) {
         // Merge with itself to deduplicate nodes
         // Note: This cannot fail for a self-merge under normal conditions.
-        // The only possible failures (TooManyNodes, TooManyDebugInfoEntries) would require the
-        // original forest to be at capacity limits, at which point compaction wouldn't help.
+        // The only possible failure (TooManyNodes) would require the original forest to be at a
+        // capacity limit, at which point compaction wouldn't help.
         MastForest::merge([&self])
             .expect("Failed to compact MastForest: this should never happen during self-merge")
     }
```
bobbinth, line 303: Is this method still meaningful? Given that debug info is gone, what else could cause the forest to be "uncompacted"?

core/src/mast/mod.rs
bobbinth, line 222: This method is still here because we need it for the merger, right? If so, let's add a comment explaining this.

bobbinth, line 595: Is this still relevant? IIRC, stripped was for stripping out debug info. Since now debug info is gone, how is this different from non-stripped options?

Or is this just a naming issue? (e.g., should this be just `write()`?)

bobbinth, line 686: This is needed because we don't enforce this invariant (i.e., no forward refs) at construction time yet, right?

core/src/mast/merger/mod.rs
bobbinth, line 49: Question (mostly because I forgot how this works): when we merge forests, some nodes may become inlined, right? Are we already handling this?

Basically, we could have a situation when forest `A` has an external node pointing at a procedure in `B`. Once these two become merged, depending on the size of this procedure, we may be able to inline the procedure at the callsites of `A`.

bobbinth (commented):
Looks good! Thank you! Not a full review (so far, I mostly looked at the `miden-core` crate) - but I left some small comments inline.

processor/src/execution/operations/crypto_ops/mod.rs
huitseeker: Done in the cleanup pass.

core/src/mast/mod.rs
huitseeker: It's a bit more than a naming issue, and basically a 'TODO:serialization'

Since debug info moved out of MastForest, the MastForest-level STRIPPED flag should go away. The presence or absence of package debug sections belongs in package serialization. 

I had planned keep this PR focused and handle that as a follow-up serialization cleanup PR (unless you want it folded into this branch).

processor/src/execution/mod.rs
plafer: Great! As discussed offline, the blake3_1to1 benchmark is just noisy.

processor/src/execution/external.rs
plafer: I still see `old_forest.get_node_by_id(resolved_node_id_new_forest)` - was the change pushed? If intended, I would add a comment explaining what we're doing here, and why this oddity is expected.

huitseeker: Small correction to my earlier answer: the current local fix does not blindly look up a loaded-forest node ID in caller debug info. `source_debug_info`belongs to the caller package, while r`esolved_node_id_new_forest` belongs to the loaded forest. 

The code now consults caller debug info only when that same ID also exists in the caller forest as a non-external node with the external target digest. Otherwise the loaded forest executes with no source sidecar. I added `loaded_external_forest_does_not_use_caller_source_roots` for that case (and kept fail-closed behavior for ambiguous same-package external roots).

core/tests/ui/mast_forest_immutable/add_debug_var.stderr
huitseeker: Yep, removed

core/tests/ui/mast_forest_immutable/debug_info_mut.stderr
huitseeker: removed

core/src/mast/node/mod.rs
huitseeker: I removed it from `MastNodeExt`, removed the generated dispatch from the derive macro, and deleted the empty implementations.

core/src/mast/mod.rs
huitseeker: Yes. It is kept private so normal callers cannot mutate roots directly, but the merger needs it when rebuilding the root set after remapping nodes into the merged forest. I added a short comment there.

processor/src/execution/mod.rs
```diff
@@ -41,6 +44,42 @@ pub(crate) struct ExecutionState<'a, P, H, S, T, F> {
     pub host: &'a mut H,
     pub tracer: &'a mut T,
     pub stopper: &'a S,
+    pub source_debug_info: Option<&'a PackageDebugInfo>,
+    pub current_source_node: Option<DebugSourceMastNodeId>,
```
plafer: nit: elsewhere we use `*node` to refer to actual nodes (e.g. `LoopNode`), and `*node_id` to refer to the id (see e.g. `start_loop_node()`, with `loop_node` and `current_node_id` variables). I would stick to this convention with debug nodes too (e.g. rename this variable `current_source_node_id`, `child_source_node()` -> `child_source_node_id()`, etc). WDYT?

core/src/mast/mod.rs
huitseeker: I narrowed the test-only helpers, but I did not put the whole block behind `#[cfg(test)]` because `make_root()` is also used by arbitrary support. That is why it is gated as `#[cfg(any(test, feature = "arbitrary"))]`. I can rename the section to make that split clearer.

huitseeker: Normal construction keeps children before parents, but serialized bytes can encode a bad node order. `compute_node_hashes()` is the validation path that detects and rejects that during untrusted deserialization. `test_untrusted_forest_detects_forward_reference` covers this by reordering a valid serialized forest and checking that validation returns `ForwardReference`.

huitseeker: Same constraint as above: the pure test helpers are test-only, but e.g. `make_root()` also supports arbitrary generation. I can move or rename the section so it does not look like normal production mutator surface.

core/src/mast/merger/mod.rs
huitseeker: The merger handles external replacement, but not as a separate size-based inlining pass. `MultiMastForestNodeIter` can yield `ExternalNodeReplacement;` the merger then maps the old external node ID to the replacement node already merged into the new forest. 

Later, parents are rebuilt with remapped children and deduplicated by fingerprint. 

So call sites can stop pointing at an external node after merge, but there is no extra procedure-size-based inlining decision here yet.

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
plafer: Similarly here, this is still unchanged. Is this expected?

processor/src/execution/external.rs
plafer: > The code now consults caller debug info only when that same ID also exists in the caller forest as a non-external node with the external target digest.

This is the part I don't understand. Realistically, when would we have an `ExternalNode` point to another node in the same MAST forest?

core/src/mast/mod.rs
huitseeker: `compact()` is still meaningful without debug info. The assembly builder interns pending nodes by key as it builds, so assembly-produced forests should already be compact for the normal duplicate cases. But `compact()` is still meaningful at the raw `MastForest` layer because duplicates can appear as a result of basic block merging for instance. There's also a `MastForestContributor::add_to_forest` but its only caller is the merger for now (and we should keep that being so by converting this from a trait method to a private helper in a follow-up PR). Plus the existing compaction tests exercise that lower-level path by adding duplicate blocks directly and then self-merging the forest.

So I would keep `compact()` for raw MastForest normalization and tests, but it's no longer load-bearing, and removing it is no longer a debug-info cleanup pass.

processor/src/execution/mod.rs
huitseeker: Yes, done in the latest commit.

plafer (commented):
At this point I think this LGTM once the outstanding comments have been resolved 👍

crates/assembly/src/mast_forest_builder/source_debug_graph.rs
huitseeker: My bad, I answered on the wrong comment (but I did fix the underlying issue): current code does now have field docs on lines 17 and 96 of source_debug_graph:

```rust
  /// Final execution MAST node represented by this source occurrence.
  exec_node: MastNodeId,
  /// Source/debug children in the same order as the execution node children they describe.
  children: Vec<SourceNodeId>,
  /// Inclusive operation start within the final execution node.
  op_start: usize,
  /// Exclusive operation end within the final execution node.
  op_end: usize,
  /// Assembly operation metadata attached to operation offsets in this occurrence.
  asm_ops: Vec<(usize, AssemblyOp)>,
  /// Debug variable metadata attached to operation offsets in this occurrence.
  debug_vars: Vec<(usize, DebugVarInfo)>,
  }
  ...
  /// Source/debug occurrence graph produced alongside a reduced execution MAST forest.
  #[derive(Clone, Debug, Default)]
  pub(crate) struct SourceDebugGraph {
      /// Dense source/debug occurrence nodes keyed by [`SourceNodeId`].
```

The line you're commenting on is the `op_end` field itself, not the comment line which has been inserted above, so you should not expect GitHub's PR comment display to mark this as outdated, and look at the whole-file diff instead which will show those comments.

processor/src/execution/external.rs
huitseeker: OK, my bad, it hardly makes sense. Blasted the whole lot in ac0cdf619fa1ffb174bebdab0f4ab2b27ba662f8

plafer (approved):
LGTM!

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +332,721 @@ impl DebugFunctionsSection {
...
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        Ok(Self(source.read_u32()?))
+    }
+
+    fn min_serialized_size() -> usize {
+        4
+    }
+}
+
+/// A source/debug MAST occurrence that points at a reduced execution node.
```
bitwalker: I think some elaboration on what a "reduced execution node" means is warranted here, since this is likely to be of especially critical importance in the implementation of debuggers and other related tooling. I'd be explicit both in terms of what an execution node is (presumably a MAST node) and what reduction refers to, so that it is possible to map this back to source language debug info in a way that is predictable (i.e. how does a `DebugSourceNode` correspond to debug info I've emitted as, or attached to, MASM AST nodes; and what happens in the event that two procedures share the same MAST root, but have differing debug info).

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +332,721 @@ impl DebugFunctionsSection {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugSourceAsmOp {
+    /// Source/debug occurrence that owns this operation row.
+    pub source_node: DebugSourceNodeId,
+    /// Operation index local to the reduced execution node.
+    pub op_idx: u32,
+    /// Optional source location for the assembly operation.
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub location: Option<Location>,
```
bitwalker, line 1199: I would suggest we attempt to collapse `Location` references, so that we only ever serialize a unique `Location` once - otherwise there will be an enormous amount of redundant duplication (a single location may be referenced by several MASM AST nodes, which in turn may correspond to a large number of MAST ops).

For MASM sources, such duplication is really only a problem for instructions which produce many MAST ops; but for higher-level source languages, you might have a single high-level expression that produces many MASM ops, so bloat is exacerbated significantly.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +332,721 @@ impl DebugFunctionsSection {
...
+    /// Assembly context name.
+    pub context_name: String,
```
bitwalker, line 1201: We currently avoid carrying strings like this in these types - instead they get interned in the strings table of the containing custom section - that way we avoid serializing the same data many times over, and it allows deserializing into reference-counted string allocations rather than duplicate heap allocations for the same string (and only if you actually care about the data).

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +332,721 @@ impl DebugFunctionsSection {
     }
 }
 
+// PACKAGE DEBUG INFO
+// ================================================================================================
+
+/// Trusted package-owned debug information decoded from well-known debug sections.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
bitwalker: I don't think we need `serde` serializers for these (and in fact, it makes it very difficult to serialize/deserialize these types efficiently, i.e. by interning strings). I'd stick to our `miden-serde-utils` serializers for this data.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +332,721 @@ impl DebugFunctionsSection {
...
+// ================================================================================================
+
+/// Assertion error message keyed by its runtime error code.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugErrorMessage {
+    /// Runtime error code emitted by the assembled assertion operation.
+    pub err_code: u64,
+    /// Human-readable assertion error message from source.
+    pub message: Arc<str>,
```
bitwalker, line 1472: This should be a reference into an interned strings table in the containing section IMO

crates/mast-package/src/debug_info/types.rs
```diff
@@ -562,124 +1293,6 @@ impl DebugFunctionInfo {
...
-// DEBUG VARIABLE INFO
-// ================================================================================================
-
-/// Debug information for a local variable or parameter.
-///
-/// This struct captures the source-level information about a variable, enabling
-/// debuggers to display variable names, types, and locations to users.
-#[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub struct DebugVariableInfo {
```
bitwalker, line 586: Why is this and `DebugInlinedCallInfo` being removed?

crates/mast-package/src/package/mod.rs
```diff
@@ -41,7 +44,52 @@ pub use self::{
     section::{InvalidSectionIdError, Section, SectionId},
     target_type::{InvalidTargetTypeError, TargetType},
 };
-use crate::{Dependency, Version};
+use crate::{
+    Dependency, Version,
+    debug_info::{DebugSourceNodeId, PackageDebugInfo},
+};
+
+/// Errors raised while stripping package-owned debug information.
```
bitwalker: I'd maybe move these error types to a separate module, since this module is getting pretty big already

bitwalker (changes requested):
Overall looks good! I found a couple of things I had questions on, or that I felt we should probably address here - see my other comments. I'm wrapping up review of the other areas of the PR, and will follow up with any further comments if I have any additional findings/questions.

core/src/mast/serialization/mod.rs
```diff
@@ -1066,35 +865,22 @@ fn decode_from_reader<R: ByteReader>(
...
-                })?,
-                "empty debug-info scaffolding",
-            )?;
-        }
-        super::DebugInfo::empty_for_nodes(layout.node_count)
-    } else {
-        super::DebugInfo::read_from(&mut recording)?
-    };
-
+    if !flags.is_stripped() {
```
bitwalker, line 881: IIUC, this means that `read_from_bytes`/`read_view_from_bytes(Materialized)` and untrusted materialization could accept arbitrary trailing bytes that are not accepted by `MastForestWireView::new`. I'm not sure if that's actually a problem in practice - but it could be depending on what follows MAST in the input stream.

crates/assembly/src/mast_forest_builder/static_import.rs
bitwalker: It looks like ambiguity around source provenance can change linkage semantics.  An ambiguous exact lookup returns None here, before the digest-only static fallback below. `ensure_external_link_with_source_ref` will then emit an `External` node on line 209. IIUC, that means debug/source ambiguity can result in a statically linked procedure being effectively dynamically linked. Perhaps this should copy the digest-matched execution node and drop source metadata?

processor/src/execution/dyn.rs
```diff
@@ -51,24 +51,145 @@ where
...
+        Continuation::FinishDyn(current_node_id),
+        state.current_source_node_id(),
+    );
+
+    // if the callee is not in the program's MAST forest, then we need to break to allow the
+    // implementing processor to fetch it (possibly asynchronously in an external library in the
+    // host).
+    match current_forest.find_procedure_root(callee_hash) {
+        Some(callee_id) => {
+            let source_node_id = match state.source_debug_info {
```
bitwalker: It looks like the presence of package debug info can change execution results. Specifically, the source-aware dyn path here turns multiple source roots for the same callee into `ExecutionError::Internal`, i.e. execution fails where it would have otherwise succeeded. The `package_source_debug_execution_rejects_ambiguous_local_dyn_root` test asserts this, but I would expect that the presence of debug info should _not_ affect the outcome of execution. Ambiguity should probably just degrade the debug info/source context, rather than change execution behavior.

processor/src/fast/execution_api.rs
```diff
@@ -48,6 +64,23 @@ impl FastProcessor {
         self.execute_with_tracer(program, host, &mut NoopTracer).await
     }
 
+    /// Async variant of [`Self::execute_with_package_debug_info_sync`].
+    #[inline(always)]
+    pub async fn execute_with_package_debug_info(
```
bitwalker, line 102: I think these new debug execution APIs lose source node disambiguation provided by the package manifest. They take only a `Program` + `PackageDebugInfo`, and then choose a source node using `unique_source_root_for_exec_node(program.entrypoint())` in `source_aware_continuation_stack`, where the package manifest actually preserves the correct source node. So if the entrypoint has the same digest as some other procedure, where the `Package` would be able to disambiguate debug info, this will fail to do so.

crates/mast-package/src/package/mod.rs
```diff
@@ -624,9 +902,14 @@ impl Package {
     }
 
     #[cfg(feature = "std")]
+    /// Reads a trusted local package file.
+    ///
+    /// This preserves package-owned debug sections and should be used only for files/cache entries
+    /// controlled by the same trusted build or execution system. Use [`Self::read_from_bytes`] for
+    /// bytes received across a trust boundary.
     pub fn deserialize_from_file(
```
bitwalker: This function is often called on user-controlled bytes in the `miden-vm` CLI (and others, such as the compiler) - but this is using the trusted deserialization path now, when it should probably be using the untrusted path, since it explicitly documents that callers should not be using it on untrusted bytes.

crates/mast-package/src/package/mod.rs
```diff
@@ -366,6 +466,185 @@ impl Package {
...
+        row_kind: &'static str,
+    ) -> Result<(), PackageDebugInfoError> {
+        let Some(source_node) = source_graph.nodes().get(source_node_id.as_u32() as usize) else {
+            return Err(PackageDebugInfoError::InvalidReference {
+                message: format!(
+                    "{row_kind} row references missing source node {source_node_id:?}"
+                ),
+            });
+        };
+        if op_idx < source_node.op_start || op_idx >= source_node.op_end {
```
bitwalker, line 1001: This checks source-map rows against the claimed range, but not `op_start <= op_end` or `op_end <= BasicBlockNode::num_operations()` - we probably need to sanity check those as well right?

crates/mast-package/src/package/mod.rs
```diff
@@ -366,6 +466,185 @@ impl Package {
...
+        if let Some(source_map) = debug_info.source_map.as_ref()
+            && source_graph.is_none()
+            && !source_map.is_empty()
+        {
+            return Err(PackageDebugInfoError::InvalidReference {
+                message: "debug source map is present without a debug source graph".to_string(),
+            });
+        }
+
+        let Some(source_graph) = source_graph else {
```
bitwalker, line 620: We may want to document that this function only really validates the source graph and not much else, since if there is no source graph present, nothing else really gets checked (e.g. malformed table indices in the debug info sections other than the source graph).

bitwalker (changes requested):
I'm finished with my initial review, and uncovered a few other issues (though you'll have to confirm if they are indeed issues, or if I misunderstood some details).

crates/mast-package/src/debug_info/types.rs
huitseeker: `DebugSourceNode` now says what `exec_node` means after MAST reduction. Here, reduction means the existing MAST normalization and deduplication path: equal source bodies may share one executable MAST node. The source graph keeps the source occurrence identity in `DebugSourceNodeId`, so two source procedures can share one executable node without becoming the same source location.

crates/mast-package/src/debug_info/types.rs
huitseeker: Done. `DebugSourceMapSection` stores each `Location` once, and assembly-op rows store a location index.

crates/mast-package/src/debug_info/types.rs
huitseeker: Done. `DebugSourceMapSection` now has a string table for row context names and operation text. Rows store string indexes.

crates/mast-package/src/debug_info/types.rs
huitseeker: Done. These package debug records now use the explicit Miden serializer. They no longer derive serde.

crates/mast-package/src/debug_info/types.rs
huitseeker: This — now — matches the current layout. `DebugFunctionsSection` owns the string table, and `DebugFunctionInfo` stores `name_idx` and `linkage_name_idx`.

crates/mast-package/src/package/mod.rs
huitseeker: Done. Package debug errors now live outside the main package module.

crates/mast-package/src/debug_info/types.rs
huitseeker: Local variables now live in the source map as `DebugSourceVar`, keyed by source node and operation index. Inline calls now live there too as `DebugSourceInlineCall`, keyed the same way. Each inline-call row stores the callee function and source-file position. There is no producer yet, but the row storage and query accessors are in place. Serialization and validation cover the new rows too.

core/src/mast/serialization/mod.rs
huitseeker: Done. The materialized reader and wire-backed reader both reject trailing bytes after a stripped MAST payload.

crates/assembly/src/mast_forest_builder/static_import.rs
huitseeker: Yep, I've had to carve cleaner policies depending on the exact issue you raise. Here in source selection, you're right: two procedures in different files may reduce to the same executable node. That is valid. When the linker has source provenance, it must select the called source occurrence and carry that `DebugSourceNodeId` into the package source graph. An end-to-end test now covers two identical procedures in different files; the package keeps the called file's metadata and source-aware execution succeeds.

processor/src/execution/dyn.rs
huitseeker: This one takes the other side of the cleaner 'debug execution failure' boundary: for a dynamic local call, the runtime may only have the callee digest. If multiple source roots match that executable node, the processor cannot safely pick one. In that case it executes without a callee source sidecar instead of attaching the wrong metadata. If a selected source node is present and its child graph is malformed, execution still fails closed.

crates/mast-package/src/package/mod.rs
huitseeker: `Package::deserialize_from_file` now uses the untrusted package reader. That closes the package-file path for 0xMiden/security-issues#16: a package file with spoofed kernel MAST node digests is rejected before execution. The same is true for normal `Package::read_from_bytes` and embedded kernel packages decoded through `try_embedded_kernel_package`. 

`Package::read_from_bytes_unchecked` still skips MAST validation and must stay limited to bytes already validated inside the same trust domain.

processor/src/fast/execution_api.rs
huitseeker: Done. The processor has an entrypoint API that accepts the manifest-selected `DebugSourceNodeId`. Callers with package manifest context do not need to infer the source root from only the executable root.

crates/mast-package/src/package/mod.rs
huitseeker: Fixed, nice catch. Source-map rows now check the source-node range and the executable block operation count.

crates/mast-package/src/package/mod.rs
huitseeker: Done. Package debug validation now checks debug types, sources, functions, source graph roots and nodes, source-map rows, inline-call rows, and the source-map/source-graph relationship.

crates/mast-package/src/package/serialization.rs
```diff
@@ -53,7 +78,7 @@ const MAGIC_PACKAGE: &[u8; 5] = b"MASP\0";
 /// The format version.
 ///
 /// If future modifications are made to this format, the version should be incremented by 1.
-const VERSION: [u8; 3] = [6, 0, 0];
+const VERSION: [u8; 3] = [5, 0, 0];
```
bitwalker: This looks like it might be a merge conflict - `next` has `VERSION` at `6`, whereas this is reverting it to `5`

processor/src/execution/external.rs
```diff
@@ -88,7 +87,11 @@ where
     continuation_stack.push_enter_forest(old_forest.clone());
 
     // Push the root node of the external MAST forest onto the continuation stack.
-    continuation_stack.push_start_node(resolved_node_id_new_forest);
+    //
+    // Caller package debug info describes the forest that contained the `External` node, not the
```
bitwalker: This seems like it basically makes debug info useless for any call into dynamically linked packages, because we don't load the debug sidecar for the external package.

I also noticed that the `From<Arc<Package>>` conversion for `HostLibrary` effectively discards the package debug info for the loaded package, so we'll have no debug info for code in the core library or protocol libraries (since those are almost always present as host libraries). I think we should probably try and preserve package debug info, and use caller context to select/disambiguate debug info for external nodes where possible (that last detail could be done in a follow-up PR, but I think it will be crucial for debug info to be fully useful).

processor/src/execution/dyn.rs
bitwalker: Does this need to use `push_with_source_node_id`?

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -743,6 +958,10 @@ impl MastForestBuilder {
                 };
                 let ops_offset = operations.len();
 
+                if let Some(source_ref) = self.latest_source_ref_by_node_ref.get(&basic_block_ref) {
```
bitwalker: This records merged source occurrences from `latest_source_ref_by_node_ref`, but if the same deduplicated `MastNodeRef` appears twice in one merge window, both merged ranges can use the latest source occurrence rather than the distinct occurrence for each original block. I'm not sure if that's intentional or not, but wanted to call it out.

crates/mast-package/src/package/mod.rs
bitwalker, line 1237: This copies the sections into the new package, but doesn't copy the `debug_sections_trusted` flag, so previously trusted sections become untrusted

bitwalker (changes requested):
Looking good! Just a couple questions/issues that came up during this pass, but once those are resolved we're probably good to go

crates/mast-package/src/package/serialization.rs
huitseeker: Indeed! thanks this was a bad conflict result. The package binary format version is back to 6.0.0, matching next, and the mast-package README now says the same version.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Fixed locally. The merge pass now selects source refs by occurrence order inside the merge window, instead of reading `latest_source_ref_by_node_ref`. I added a regression test with two identical source blocks that dedup to one executable block, then merge into one block while keeping separate source ranges.

crates/mast-package/src/package/mod.rs
huitseeker: My bad, fixed and tested.

processor/src/execution/external.rs
huitseeker: Fixed in the latest commit. `HostLibrary::from(Arc<Package>)` now keeps trusted package debug info instead of dropping it. The host stores loaded code as `LoadedMastForest`, which carries both the `Arc<MastForest>` and the package-owned debug info for that same forest.
External and dynamic loads now select the loaded root source node from the loaded package debug graph. While executing the loaded forest, the processor switches to  that package debug info; when it returns, it restores the caller debug info. Stripped or untrusted loaded packages still execute, just without loaded-package source labels.

processor/src/execution/dyn.rs
huitseeker: Yes, and this is now fixed. `finish_load_mast_forest_from_dyn_start()` takes the selected loaded-package `DebugSourceNodeId` and pushes the loaded root with:
`push_with_source_node_id(Continuation::StartNode(root_id), new_source_node_id)`

If the loaded package has no usable root source node, this is `None` and execution continues without a source sidecar. If it has trusted debug info, the loaded root gets the loaded package source sidecar, not the caller’s.

processor/src/fast/execution_api.rs
```diff
@@ -472,64 +979,126 @@ impl FastProcessor {
...
-        Ok((root_id, mast_forest))
+        Ok((root_id, mast_forest, loaded_package_debug_info, loaded_source_node_id))
+    }
+
+    fn loaded_package_source_context(
+        loaded_mast_forest: &LoadedMastForest,
+        root_id: MastNodeId,
+        source_aware_execution: bool,
+    ) -> Result<(Option<Arc<PackageDebugInfo>>, Option<DebugSourceNodeId>), ExecutionError> {
+        if !source_aware_execution {
```
bitwalker, line 1090: This will fail to load package source context in cases where the caller package had no source context, but the callee package does. The lack of caller source context does mean that we don't have as much information to resolve ambiguities, but in cases where the callee is unambiguous, there is no reason to skip loading the source context for it.

This also affects execution started from APIs like `execute_sync` vs `execute_with_package_debug_info_sync` as well, though perhaps in that case it is intentional that we would never load any debug info for any packages. However, if that's the intention, I think we probably can't rely on inferring whether execution is source-aware or not based on the presence of source context in the current package, and instead use a global flag for it. That way we don't unintentionally switch into non-source-aware execution mode if some loaded package is missing source context.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +325,866 @@ impl DebugFunctionsSection {
...
+            for row in source_map.debug_vars() {
+                let source_node = source_id_map.get(&row.source_node).copied().ok_or(
+                    PackageDebugInfoMergeError::MissingSourceNodeMapping {
+                        forest_index,
+                        source_node: row.source_node,
+                    },
+                )?;
+                debug_vars.push(DebugSourceVar::new(source_node, row.op_idx, row.var.clone()));
+            }
+            // Inline-call rows refer to function and source-file tables. This source-only merge
```
bitwalker: Could you elaborate on this a bit? Specifically what is meant by higher-level package composition? It appears that this results in silently dropping inline call metadata. Is the inline call data being handled somewhere else, or are we just dropping it?

crates/mast-package/src/debug_info/types.rs
```diff
@@ -279,6 +325,866 @@ impl DebugFunctionsSection {
...
+    /// Merges package-owned source/debug metadata after a [`miden_core::mast::MastForest`] merge.
+    ///
+    /// [`miden_core::mast::MastForest::merge`] remains execution-only. This helper applies the
+    /// returned node mappings to package source/debug sections so callers can merge
+    /// `(MastForest, PackageDebugInfo)` pairs without reattaching debug metadata to the forest.
+    ///
+    /// This merges only `debug_source_graph` and `debug_source_map`; type, source-file, and
+    /// function sections have their own string/index tables and are left for higher-level package
+    /// composition.
+    pub fn merge_source_debug<'a>(
```
bitwalker, line 464: I can't find any callers of this function outside of tests - is this meant to be used in a subsequent PR?

bitwalker (commented):
A couple remaining bugs/questions, one of which I mention below since I can't attach a comment to code in a file that isn't modified, the rest are in comments:

* The CLI is bypassing the new `HostLibrary::from(Package)` conversion, resulting in execution not getting any debug info. See `cli/run.rs:191`.

processor/src/fast/execution_api.rs
huitseeker: Yep, good catch. Fixed by making the package-debug entrypoints source-aware from entry. `execute_sync` stays source-unaware. A new test covers a caller with no selected source root and an unambiguous loaded package source span.

crates/mast-package/src/debug_info/types.rs
huitseeker: The reviewed merge dropped inline calls 🤦  Fixed by merging the type, source-file, and function tables, then remapping inline-call rows into the merged package debug info. The test covers two source nodes that share one executable node.

crates/mast-package/src/debug_info/types.rs
huitseeker: We are keeping `PackageDebugInfo::merge_source_debug` as the debug side of `MastForest::merge`, which avoids deprecating functionality, and paves the way for the set operations we want as part of https://github.com/0xMiden/miden-vm/issues/2980.

huitseeker:
@bitwalker You were right, the cli's `run` loaded only the MastForest, so trusted package debug info was dropped before execution. Fixed by loading each `Arc<Package>` into the host through `HostLibrary::from(Arc<Package>)`. Normal file reads still use untrusted package deserialization.

core/src/mast/mod.rs
huitseeker: Filed https://github.com/0xMiden/miden-vm/issues/3263 for the cleanup.

processor/src/fast/execution_api.rs
```diff
@@ -472,64 +977,126 @@ impl FastProcessor {
...
+                },
+                _ => crate::errors::procedure_not_found_with_context(node_digest),
+            }
         })?;
+        let mast_forest = loaded_mast_forest.mast_forest().clone();
 
         let root_id = mast_forest.find_procedure_root(node_digest).ok_or_else(|| {
             Err::<(), _>(OperationError::MalformedMastForestInHost { root_digest: node_digest })
-                .map_exec_err(current_forest, node_id, host)
+                .map_exec_err()
```
bitwalker: I think this might drop source context for the caller. 

When the host returns a `LoadedMastForest` that does not actually contain the requested digest, both sync/async loaders create this `MalformedMastForestInHost` via `map_exec_err`. The external-call error helper recognizes `OperationError` errors, but rebuilds them with `err.clone().with_context()` rather than package source context (see `external.rs:69`). So source-aware execution preserves the caller span for “host has no package”, but still loses it for “host returned the wrong/malformed package”. 

I’d add a regression test with a host returning a forest that lacks the requested root and route that case through `with_package_source_context` when `package_debug_info` and `source_node_id` are available.

bitwalker (approved):
I found one last bug, but I'm marking this approved under the assumption it'll be fixed before merge (or after if that's easier), so that I'm not holding things up.

Really nice work here!

processor/src/fast/execution_api.rs
huitseeker: Thanks, fixed in the latest commit

PR #3222, opened 2026-06-03
Add miden-precompiles crate with hash and signature precompiles

adr1anh:
This is the combined deferred-precompile migration PR.

It includes the deferred framework work that was previously split out as #3170, plus the concrete `miden-precompiles` crate and the migration of the existing crypto precompile proof path onto deferred state. #3170 can now be closed in favor of this PR.

The useful way to review this PR is **commit-by-commit**, not as one 20k-line diff. In particular, the generated math MASM should not be reviewed line-by-line; review the Rust semantics, templates, and tests instead.

## Why this exists

The legacy precompile path models precompile work as a flat request list. A MASM program logs requests, the host records them, and the proof commits to a transcript of those requests.

That representation is too coarse for the direction we need: proof-bound precompile work is naturally graph-shaped. Hashes, committed byte chunks, uint/field values, curve points, signature equations, predicates, and shared intermediate results all need stable identities and explicit dependencies.

This PR replaces that legacy request-list model with a content-addressed deferred DAG:

- MASM registers deferred nodes and logs statements about them.
- The processor owns a `DeferredState` during execution.
- `log_deferred` constrains deferred-root transitions in the AIR.
- Proofs carry a canonical `DeferredStateWire` witness instead of legacy precompile requests.
- The verifier rehydrates that wire under the official precompile registry and uses the resulting root as a public input.
- Concrete crypto precompile semantics live in the new `miden-precompiles` crate, not as ad-hoc handlers in the core library.

The main design question for reviewers is not “is every generated MASM line beautiful?” but:

> Are the new deferred-state, proof, and precompile boundaries sound, reviewable, and staged in the right places?

## High-level changes

- Domain-separates the legacy/deferred transcript fold with `Tag::AND = [1, 0, 0, 0]` instead of zero-capacity `Poseidon2::merge`.
- Adds `miden_core::deferred`: opaque tags, content-addressed nodes, precompile registry, eager evaluation, canonical root tracking, and canonical wire serialization/rehydration.
- Adds deferred advice events and `log_deferred` execution/AIR semantics.
- Replaces `ExecutionProof`’s legacy precompile request witness with `DeferredStateWire`.
- Adds the `miden-precompiles` crate and `miden::precompiles` MASM package.
- Implements official deferred precompiles for:
  - Keccak256 and SHA512 hash assertions;
  - fixed uint/field domains (`u256`, secp256k1/r1 fields, Ed25519 fields);
  - fixed curve operations for secp256k1, secp256r1, and Ed25519;
  - MASM DSA wrappers over those primitives.
- Removes legacy core crypto precompile handlers and reimplements core hash / ECDSA K256 Keccak wrappers through the official precompiles.
- Installs the official precompile registry by default in high-level VM/proving/testing entry points.

#### `25f729085` — `precompile: domain-separate legacy log_precompile transcript`

Review this first because it defines the transcript/root domain used by the later cutover.

What to check:

- `PrecompileTranscript`, processor execution, AIR constraints, MASM docs, and recursive verifier constants all agree on the same fold domain: `[1, 0, 0, 0]`.
- The stack/helper layout is unchanged except for the domain-separated fold.
- AIR relation digest and recursive verifier constants were updated together.
- No remaining legacy transcript path still uses the old all-zero capacity fold.

This is intentionally protocol-affecting.

#### `cc1e01f50` — `core: introduce deferred DAG state and canonical wire format`

This is the core framework from #3170.

Start with:

- `core/src/deferred/node.rs`
- `core/src/deferred/state.rs`
- `core/src/deferred/wire.rs`
- `core/src/deferred/precompile_registry.rs`

What to check:

- `Tag` is opaque and framework-reserved tags cannot be confused with precompile-owned tags.
- Node identity commits to both tag and payload.
- `Node::TRUE` / `TRUE_DIGEST` represent the empty valid state.
- `Node::and` uses the framework `AND` domain, not a Merkle-style zero-capacity merge.
- `DeferredState` registration is eager: shape, child closure, precompile semantics, canonical result, and budget accounting happen at registration/evaluation time.
- Predicates succeed by reducing to `TRUE`; non-true predicates must not be silently logged.
- `DeferredStateWire` is canonical, topological, root-last, duplicate-free, and checked under the installed registry.
- Wire rehydration from untrusted proof data is bounded and rejects malformed/noncanonical data.

Prior reviewer context: #3170 raised scope concerns about introducing a generic registry/DAG substrate before the concrete migration. In this combined PR, the substrate is no longer abstract-only: later commits immediately use it to migrate the concrete crypto precompile proof path. Reviewers should still focus on the substrate invariants above before moving to concrete precompile code.

#### `fb89725e0` — `precompiles: add crate scaffold and embedded MASM package`

This commit is scaffolding only.

Review:

- workspace membership and dependency shape;
- `precompiles/Cargo.toml`;
- build-time MASM packaging;
- the `miden::precompiles` namespace;
- `PrecompilesLibrary` and empty initial `registry()`.

Do not review concrete hash/math/DSA semantics here; those arrive in later commits.

#### `ce0833fe1` — `processor,air: add deferred advice events and log_deferred semantics`

This is the main execution/AIR commit.

What to check:

- `FastProcessor` now owns a `DeferredState`.
- `adv.register_deferred`, `adv.register_deferred_data`, and `adv.evaluate_deferred*` expose the minimal MASM-facing deferred interface.
- `log_deferred` reads a statement digest, folds it with the previous deferred root under `Tag::AND`, and verifies the same transition against `DeferredState`.
- AIR lookup messages bind the processor transition to the hasher chiplet and deferred-root bus.
- Public inputs now include the final deferred root.
- `with_options()` / registry installation do not accidentally discard an installed deferred registry or accumulated state.

Important boundary: advice events return hints. They are not proof-bound unless MASM code logs deferred assertions tying those hints back to circuit-visible commitments.

#### `da7fa8a85` — `proofs: replace precompile request transcript with deferred state wire`

This is the proof cutover.

What to check:

- `ExecutionProof` carries `DeferredStateWire` instead of legacy `PrecompileRequest`s.
- The prover serializes the execution `DeferredState` into canonical wire form.
- The verifier rehydrates that wire under the official registry and uses the resulting root as the deferred-root public input.
- Legacy request-list fuzz targets are replaced by deferred-wire fuzz targets.
- Recursive verifier/public-input plumbing stays consistent with the new 40-felt public input layout.

This commit intentionally does not support old proof formats or arbitrary user-provided verifier registries in the default high-level verification path.

#### `38fc8493a` — `precompiles: implement deferred hash precompiles`

Adds Keccak256 and SHA512 deferred hash assertions.

What to check:

- Hash advice is treated as untrusted: the wrapper obtains a digest, then logs a deferred assertion binding the preimage chunks to the expected digest.
- `n_bytes` determines the exact preimage length; trailing padding must be zero.
- Digest widths are enforced: 8 u32 felts for Keccak256, 16 for SHA512.
- Runtime hash event handlers still respect `ExecutionOptions::max_hash_len_bytes()` before reading/hashing memory.
- The verifier-side deferred precompile evaluation is registry/state-budgeted rather than driven by the processor’s runtime hash advice path.

#### `96fe8bf26` — `precompiles: implement uint, field, and curve precompiles with generated MASM`

This is the largest code-adding commit. Please do **not** line-review the generated MASM files.

Review manually:

- `precompiles/src/math/uint/*`
- `precompiles/src/math/curve/*`
- `precompiles/src/codegen/masm.rs`
- `precompiles/src/codegen/templates/*.tpl`
- integration tests under `precompiles/tests/integration/*`

Generated outputs to skim only for structure/API shape:

- `precompiles/asm/math/u256.masm`
- `precompiles/asm/math/*_base.masm`
- `precompiles/asm/math/*_scalar.masm`
- `precompiles/asm/math/{secp256k1,secp256r1,ed25519}.masm`

What to check:

- Fixed domains only; this is not a runtime-pluggable bigint/curve library.
- u256/field values are little-endian `u32[8]` limbs.
- Field values are canonical modulo the configured modulus.
- Mixed-domain uint operations are rejected.
- Curve `VALUE` nodes validate affine coordinates and identity encoding.
- Curve operations reject wrong-curve / wrong-scalar-domain inputs.
- MSM pair counts and scalar/point domains are enforced.
- Generated MASM drift is checked by the generator/CI, not by manual review of thousands of generated lines.

Prior review comments about speculative codegen hooks, duplicate const parsers, dynamic registries, and cycle-count sources of truth appear to have been simplified in the current version. One open organizational question remains: curve wrappers currently live under `miden::precompiles::math`; reviewers may still prefer a future `crypto::curves` namespace.

#### `72d6efe27` — `precompiles: add DSA MASM wrappers over hash and curve primitives`

Adds official MASM DSA wrappers over the deferred primitives.

Review:

- `precompiles/asm/crypto/dsa/ecdsa_secp256k1.masm`
- `precompiles/asm/crypto/dsa/eddsa_ed25519.masm`
- corresponding integration tests.

What to check:

- Encoding and endianness assumptions are explicit and correct.
- Host/advice outputs are bound by deferred assertions before being trusted.
- ECDSA prehash verification treats `z` as caller-supplied; message binding belongs in the higher-level Keccak wrapper.
- The core K256/Keccak wrapper binds `z = Keccak256(message)` and checks the public-key commitment.
- EdDSA challenge construction binds `R`, `A`, and message as intended.
- Scalar canonicality, zero checks, identity/low-order policy, and final signature equations are correct.
- Stack comments match the actual MASM stack effects.

This is the most important commit for cryptographic review. The PR should not overclaim here: the wrappers are designed to make the algebra and message binding visible in MASM, but reviewers should still check those details carefully.

#### `e0fb18fe9` — `core-lib: aggregate precompiles package and remove legacy handlers`

This changes the library composition and removes old core host handlers.

What to check:

- `CoreLibrary` embeds/aggregates the `miden-precompiles` package.
- Legacy core handlers for hash/DSA are removed.
- Core tests and docs are adjusted to the new handler/precompile split.
- The core library no longer owns concrete precompile semantics; it delegates to the official precompiles package.

#### `fbed5ecd9` — `core-lib: implement legacy hash wrappers via official precompiles`

Restores the core hash namespaces by delegating to official precompile wrappers.

What to check:

- Existing MASM-facing hash procedure shapes remain usable.
- Return values and memory contracts match prior expectations.
- The proof machinery underneath now goes through deferred hash assertions.

#### `981df70e0` — `core-lib: implement legacy ECDSA K256 Keccak wrapper via official precompiles`

Migrates the old core ECDSA K256/Keccak entry point onto the official secp256k1 prehash verifier.

What to check:

- Public-key commitment and message hashing are still bound.
- The native-advice ABI change is documented honestly.
- The wrapper should be described as a namespace/message-format migration layer, not as proof-format compatibility with old precompile requests.

#### `e83bdbe23` — `api: load official precompiles by default`

Completes the default install story.

Why this matters: `miden-precompiles` depends on processor event-handler types, so the processor crate cannot simply depend on `miden-precompiles` and install it internally everywhere.

What to check:

- High-level VM/prover/testing/CLI paths install `miden_precompiles::registry()` by default.
- Raw `FastProcessor` remains low-level and explicit: direct users still install deferred precompiles themselves if they bypass the high-level APIs.
- Test utilities steer users toward the new default processor construction path where appropriate.

This addresses the earlier review concern that users could load `PrecompilesLibrary` but not install the matching registry outside tests.

## Key invariants across the whole PR

- The deferred root is a public input and is constrained by AIR.
- `log_deferred`’s constrained Poseidon2 transition and `DeferredState`’s host-side transition must be the same transition.
- The prover and verifier use the same official precompile registry semantics.
- `DeferredStateWire` must be canonical and safe to decode from untrusted proof data.
- Advice values are never proof by themselves; MASM wrappers must log deferred assertions that bind advice to committed data.
- Generated math MASM is derived from reviewed Rust descriptors/templates and guarded by generation checks.
- The old precompile proof witness is gone; this PR is not proof-format compatible with legacy precompile-request proofs.

## Known reviewer focus / follow-ups

These are worth keeping in mind during review:

- #3170’s framework work is integrated here; that PR should be closed to avoid split review state.
- The wire canonicality/rehydration path should be reviewed with adversarial deep graphs in mind. In particular, make sure malformed or very deep wire fails as an ordinary error rather than risking verifier stack exhaustion.
- The crypto wrappers need focused review for message binding, canonicality, identity/low-order behavior, scalar reduction, and stack discipline.
- The public namespace for curve wrappers may still deserve a decision (`math::*` today vs a possible future `crypto::curves::*`).
- Raw processor users still need explicit registry installation; high-level VM/prover/testing entry points install the official registry by default.

## Useful validation commands

Targeted commands for this stack:

```sh
cargo test -p miden-precompiles
make check-precompile-masm
cargo test -p miden-core deferred
cargo test -p miden-processor deferred_e2e
```

Broader validation should include the normal workspace/test/recursive-verifier CI jobs because this PR changes the proof public-input layout and relation digest constants.


precompiles/asm/crypto/dsa/eddsa_ed25519.masm
```diff
@@ -0,0 +1,54 @@
...
+# ================================================================================================
+
+#! Verifies an Ed25519 / SHA-512 EdDSA signature with a precomputed nonce digest via the
+#! deferred-DAG framework.
+#!
+#! The caller provides a single pointer to a contiguous 40-felt buffer with the precompile layout:
+#!
+#!   bytes[0..32]    pk
+#!   bytes[32..96]   k_digest = SHA-512(R || A || message)
+#!   bytes[96..160]  sig
```
Nashtare: I think we actually do need to recompute `k_digest` in -circuit, otherwise we could have forgeries? i.e. if `k` is arbitrary we can find (R,s) sig satisfying the verifier for any fixed public key

precompiles/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -0,0 +1,56 @@
...
+#! Verifies a secp256k1 / Keccak256 ECDSA signature over a pre-hashed message via the deferred-DAG
+#! framework.
+#!
+#! The caller provides a single pointer to a contiguous 40-felt buffer packing
+#! `pk || digest || sig` (130 bytes), zero-padded to the chunk boundary:
+#!
+#!   bytes[0..33]    pk
+#!   bytes[33..65]   digest (32 bytes)
+#!   bytes[65..130]  sig (65 bytes)
+#!   bytes[130..160] zero pad (30 bytes)
```
Nashtare: I don't see `digest` being binded to the signed message, I think this may lead to forgeries too

precompiles/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -0,0 +1,56 @@
...
+# asserts these values.
+
+const PRECOMPILE_ID = 5216793680992356579
+const VERIFY_TAG_ID = 0
+
+# PROCEDURES
+# ================================================================================================
+
+#! Verifies a secp256k1 / Keccak256 ECDSA signature over a pre-hashed message via the deferred-DAG
+#! framework.
```
Nashtare: I think the assumption we have that the deferred root is canonical per statement may not be exactly accurate here, as (unlike miden-crypto does) we don't check for canonicity / small-order here

precompiles/src/math/weierstrass/mod.rs
```diff
@@ -0,0 +1,1395 @@
...
+    /// Builds a unary scalar-multiplication expression node.
+    pub fn mul128_node(curve: WeierstrassCurve, point: Digest, scalar: [u32; 4]) -> Node {
+        let params = Word::new(scalar.map(Felt::from_u32));
+        Node::unary(Self::tag(Self::MUL128_OP_ID, curve), point, params)
+            .expect("Weierstrass mul128 tag is precompile-owned")
+    }
+
+    fn extend_init_nodes_with_point(
+        nodes: &mut Vec<Node>,
+        curve: WeierstrassCurve,
```
huitseeker: This is the line Clippy is failing on in CI. `WeierstrassCurve` is 520 bytes, so with `-D warnings` this needs to pass `curve` by reference or otherwise avoid the by-value copy.

precompiles/src/hash/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+}
+
+/// Evaluates a `preimage` data node: unpack chunks to bytes (stripping zero-pad down to the
+/// `n_bytes` carried in `args[1]`), hash, and emit the canonical `digest` node.
+fn evaluate_preimage<H: HashFunction>(
+    args: [Felt; 3],
+    payload: &Payload,
+) -> Result<Node, PrecompileError> {
+    let n_bytes = u32::try_from(args[1].as_canonical_u64())
```
huitseeker: This path seems to skip the `max_hash_len_bytes` limit that the existing core Keccak/SHA event handlers enforce before hashing. A caller with a lower hash limit can still register a deferred preimage as long as it fits the deferred-state budget, so the new wrappers can do more host hashing than `ExecutionOptions` allows.

precompiles/src/codegen/masm.rs
```diff
@@ -0,0 +1,948 @@
...
+    } else {
+        Err(format!("invalid uint constant name `{name}`"))
+    }
+}
+
+fn digest_const_name(name: &str) -> String {
+    format!("{}_DIGEST", name.to_ascii_uppercase())
+}
+
+fn append_extra_sections(
```
huitseeker: This generic `extra_sections` hook looks a bit speculative right now: every caller passes `&[]`, and the actual special cases already have named placeholders (`FIELD_INVERSION_*`, `EXTRA_CONSTANT_*`). Given the size of this PR, I would cut `extra_sections`, `append_extra_sections`, `render_extra_section`, `apply_cycle_count_replacements`, and their unit tests for now. 

If a future field really needs another custom block, adding one explicit placeholder will be easier to review than carrying an unused extension point.

precompiles/src/codegen/masm.rs
```diff
@@ -0,0 +1,948 @@
...
+const UINT_TEMPLATE_PATH: &str = "precompiles/src/codegen/templates/uint.masm.tpl";
+const UINT_TEMPLATE: &str = include_str!("templates/uint.masm.tpl");
+const UINT_FIELD_INV_CONSTANTS_TEMPLATE: &str =
+    include_str!("templates/uint_field_inv_constants.masm.tpl");
+const UINT_FIELD_INV_TEMPLATE: &str = include_str!("templates/uint_field_inv.masm.tpl");
+const WEIERSTRASS_TEMPLATE_PATH: &str = "precompiles/src/codegen/templates/weierstrass.masm.tpl";
+const WEIERSTRASS_TEMPLATE: &str = include_str!("templates/weierstrass.masm.tpl");
+const REGENERATE_COMMAND: &str = "make regenerate-precompile-masm";
+
+pub const SECP256K1_WEIERSTRASS_CYCLES: &[(&str, u64)] = &[
```
huitseeker: These cycle-count tables are another source of truth, and the test suite already recomputes the values from the assembled MAST. Can the generator derive the `*_CYCLES` replacements from the assembled generated MASM instead? That would remove the manual tables and make the checked-in comments a generated artifact rather than data that has to be updated in lockstep.

precompiles/src/math/uint/precompile.rs
```diff
@@ -0,0 +1,458 @@
...
+        assert_digest_consts(source, "TWO_DIGEST", UintElement::<M>::two().to_node().digest());
+
+        for constant in M::EXTRA_INIT_CONSTANTS {
+            let name = digest_const_name(constant.name);
+            let digest = UintElement::<M>::new(constant.value).unwrap().to_node().digest();
+            assert_digest_consts(source, &name, digest);
+        }
+    }
+
+    fn masm_const(source: &str, name: &str) -> u64 {
```
huitseeker: This MASM const parser is duplicated here and again in the Weierstrass tests, while the hash module already has the same helper for `const NAME = VALUE` parsing. Please move the shared test helper once and reuse it instead of growing three local parsers in this new crate.

precompiles/src/math/uint/precompile.rs
```diff
@@ -0,0 +1,458 @@
...
+                Self::Eq { lhs, rhs }
+            },
+        })
+    }
+}
+
+/// Deferred precompile for 256-bit modular arithmetic.
+#[derive(Clone, Debug)]
+pub struct UintPrecompile {
+    moduli: BTreeMap<Felt, UintModulus>,
```
huitseeker: This `BTreeMap`/builder shape feels heavier than the current API needs. The supported moduli are fixed by this crate and `Default` installs all of them, so a static supported-moduli table + lookup by id should cover the same behavior with less dynamic registration surface. New fields would just append to the table.

precompiles/src/math/weierstrass/mod.rs
```diff
@@ -0,0 +1,1395 @@
...
+        }
+        scalar[i] = limb as u32;
+    }
+    Ok(scalar)
+}
+
+/// Deferred precompile for generic short-Weierstrass point operations.
+#[derive(Clone, Debug)]
+pub struct WeierstrassPrecompile {
+    curves: BTreeMap<Felt, WeierstrassCurve>,
```
huitseeker: Same simplification as the uint precompile: this dynamic curve registry/builder is a lot of API for two built-in curves. Unless custom runtime curves are an intended public API, a static list plus lookup by id would be smaller and easier to audit.

precompiles/src/math/weierstrass/mod.rs
```diff
@@ -0,0 +1,1395 @@
...
+        self
+    }
+
+    /// Returns the installed curve for `id`.
+    pub fn curve(&self, id: Felt) -> Option<WeierstrassCurve> {
+        self.curves.get(&id).copied()
+    }
+
+    /// Returns the installed curve for the statically known spec `C`.
+    pub fn curve_for<C: WeierstrassCurveSpec>(&self) -> Option<WeierstrassCurve> {
```
huitseeker: This helper appears unused. If the registry stays, I would still delete `curve_for<C>()` until there is a caller.

precompiles/src/codegen/masm.rs
adr1anh: Agreed, I'll remove it to make things simpler

precompiles/src/hash/mod.rs
adr1anh: The full migration branch replaces the maximum hash length with a more general bound on the total size of the deferred state.

precompiles/src/codegen/masm.rs
adr1anh: I'll remove the cycle count from the code gen, and focus on the number of cycles for the full primitive.

huitseeker (commented):
This overall looks good now, do we want to move this to a real rebased review @adr1anh ? /cc @Al-Kindi-0

adr1anh:
I'll rebase #3170 onto origin/next, then merge #3170 onto here (#3222). Then, I'll squash #3224 into #3222, and close that PR.

precompiles/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl AsRef<Package> for PrecompilesLibrary {
+    fn as_ref(&self) -> &Package {
+        &self.0
+    }
+}
+
+impl From<&PrecompilesLibrary> for HostLibrary {
```
huitseeker: Can we make the precompile registry installable through the normal processor API before exposing this as a `HostLibrary`? 

Loading `PrecompilesLibrary` gives users the MASM package and `field_inv` handler, but these wrappers call `adv.register_deferred*`, and the processor still starts with an _empty_ deferred registry. 

The only hook I found for installing `registry()` is `FastProcessor::with_deferred_precompiles`, but that method is gated behind `#[cfg(any(test, feature = "testing"))]`. I also checked this with a small downstream crate using normal `miden-processor` features, and it fails to compile with `no method named with_deferred_precompiles`. 

So outside tests, a caller can load this library but cannot make the deferred wrappers execute successfully.

precompiles/asm/math/ed25519.masm
bobbinth, line 1: Putting this (and other similar files) into the `math` namespace feels a bit off to me. Maybe a better place would be `asm/crypto/curves`?

processor/src/fast/operation.rs
```diff
@@ -65,6 +65,23 @@ impl Processor for FastProcessor {
...
+    fn log_deferred_statement(
+        &mut self,
+        statement_digest: Digest,
+        expected_new_root: Word,
+    ) -> Result<(), OperationError> {
+        self.deferred_state
+            .log_verified_statement(statement_digest, expected_new_root)
+            .map(|_| ())
+            .map_err(OperationError::from)
+    }
```
adr1anh: Should move to System

processor/src/fast/mod.rs
```diff
@@ -316,11 +307,10 @@ impl FastProcessor {
...
 
-    /// Installs deferred precompiles for tests and experimental end-to-end coverage.
+    /// Installs deferred precompiles into this processor's deferred state.
     ///
-    /// This is intentionally gated behind `testing`. Production registry installation is not part
-    /// of this branch; the default processor uses an empty deferred registry.
-    #[cfg(any(test, feature = "testing"))]
+    /// The default processor uses an empty deferred registry. Call this when executing programs
+    /// that link against concrete deferred precompile packages such as `miden-precompiles`.
     pub fn with_deferred_precompiles(
```
adr1anh: `with_precompiles`

verifier/src/lib.rs
```diff
@@ -56,70 +54,25 @@ pub use exports::*;
...
-
-    verify_stark(
-        program_info,
-        stack_inputs,
-        stack_outputs,
-        pc_transcript_state,
-        hash_fn,
-        proof_bytes,
+    let state = DeferredState::from_wire(
+        Arc::new(miden_precompiles::registry()),
```
huitseeker, line 96: This hard-codes verification to `miden_precompiles::registry()`, but execution can still build a proof with another deferred registry through `FastProcessor::with_deferred_precompiles` or by passing prebuilt trace inputs into the prover.

Can we keep a verifier entrypoint that accepts a `PrecompileRegistry`, with `verify()` using the built-in registry as the default? Otherwise valid proofs made with a matching custom registry have no verifier path.

crates/lib/core/asm/crypto/dsa/eddsa_ed25519.masm
```diff
@@ -50,114 +37,6 @@ const SIG_LEN_FELTS = 16      # 64.div_ceil(4)
...
-    swapw
-    # => [COMM_PK_MSG, COMM_SIG, result]
-    exec.poseidon2::merge
-    # => [COMM, result]
-
-    # Create TAG = [EDDSA_VERIFY_EVENT, result, 0, 0] on stack
-    # Stack positions 4-7 should have [event_id, result, 0, 0] for get_stack_word(4)
-    push.0.0 dup.6 push.EDDSA_VERIFY_EVENT swapw
-    # => [COMM, TAG, result]
+    push.0 assert.err="eddsa_ed25519::verify implementation pending"
```
huitseeker: This keeps the old `miden::core` Ed25519 export, but the old implementation has been replaced by an unconditional trap while the new working path lives under `miden::precompiles`. Do we want this compatibility wrapper in this PR? If yes, it should probably delegate now; if not, I would rather delete the export/docs until the wrapper is real.

precompiles/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+#[derive(Clone)]
+pub struct PrecompilesLibrary(Arc<Package>);
+
+impl PrecompilesLibrary {
+    /// Serialized representation of the `miden-precompiles` package.
+    pub const SERIALIZED: &'static [u8] =
+        include_bytes!(concat!(env!("OUT_DIR"), "/assets/miden-precompiles.masp"));
+
+    /// Returns a reference to the [`MastForest`] underlying the precompiles library.
+    pub fn mast_forest(&self) -> &Arc<MastForest> {
```
huitseeker: This API surface looks a bit wider than needed. `package()` already gives callers the package, and from there they can reach the MAST forest if they really need it.

Can we drop `mast_forest()` and the `AsRef<Package>` impl for now? That leaves one clear way to get the wrapped package instead of three overlapping access paths.

docs/src/design/deferred/index.md
```diff
@@ -62,40 +58,48 @@ content yields an identical digest, so equal subterms are shared automatically (
   `Tag::AND = [1, 0, 0, 0]` tags semantic conjunction nodes. No precompile may claim either id.
   Deferred statement accumulation uses the same semantic `AND` constructor as a restricted
   right-spined chain.
-- A **payload** is the node's body, in one of three shapes:
+- A **payload** is the node's body, in one of four shapes:
```
huitseeker, line 61: This section still says payloads have four shapes, but the implementation and the precompile design page both treat `PairList` as a first-class payload shape.

Can we document `PairList` here too, including its child ordering, canonical wire encoding, arity checks, and budget behavior? Otherwise this page gives implementers an incomplete model of what the verifier accepts.

precompiles/src/lib.rs
adr1anh: Unfortunately, the processors cannot depend on `miden-precompiles` since the latter implements event handlers which depend on the processor. The approach taken in a later commit is to make all public user-facing executing/proving/testing methods install the registry. Moreover, the PrecompileLibrary is folded into the CoreLibrary so that the handlers and precompile wrappers always get included by default.

processor/src/fast/mod.rs
adr1anh: Maybe doc hidden

miden-core-fuzz/Cargo.toml
```diff
@@ -163,20 +163,6 @@ test = false
...
-doc = false
-bench = false
-
-[[bin]]
-name = "precompile_request_serde_deserialize"
-path = "fuzz_targets/precompile_request_serde_deserialize.rs"
-test = false
-doc = false
-bench = false
-
```
adr1anh, line 179: We may want to re-introduce something similar for deferred wire state

miden-core-fuzz/README.md
```diff
@@ -137,6 +125,7 @@ cargo +nightly fuzz run deferred_state_wire_deserialize --fuzz-dir miden-core-fu
 cargo +nightly fuzz run deferred_state_wire_serde_deserialize --fuzz-dir miden-core-fuzz
 ```
 
+
```
adr1anh: Remove

miden-vm/src/lib.rs
```diff
@@ -29,6 +29,46 @@ pub use miden_verifier::VerificationError;
 #[cfg(feature = "internal")]
 pub mod internal;
 
+/// Executes and proves a Miden program with the built-in precompile registry.
+pub async fn prove(
```
adr1anh: I'm not sure we need these

precompiles/src/hash/handlers.rs
```diff
@@ -0,0 +1,187 @@
...
+struct HashDigestHandler<H>(PhantomData<H>);
+
+impl<H> Default for HashDigestHandler<H> {
+    fn default() -> Self {
+        Self(PhantomData)
+    }
+}
+
+impl<H: HashFunction> EventHandler for HashDigestHandler<H> {
+    fn on_event(&self, process: &ProcessorState<'_>) -> Result<Vec<AdviceMutation>, EventError> {
```
adr1anh: Given that we now require a handler for hashes, maybe we should go back and restore the max_hash_length that we removed in a previous commit

precompiles/src/hash/handlers.rs
```diff
@@ -0,0 +1,187 @@
...
+            });
+        }
+    }
+
+    out.truncate(len_bytes);
+    Ok(out)
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum HashDigestEventError {
```
adr1anh: Why not use thiserror

crates/lib/core/precompiles/asm/hashes/keccak256.masm
adr1anh, line 1: Rather than use a common interface with SHA512, we might want to make Keccak return the hash on the stack since it is only 8 slots, and we could probably save some cycles

precompiles/asm/crypto/dsa/ecdsa_secp256k1.masm
```diff
@@ -0,0 +1,398 @@
...
+#!
+#! Assumptions/deferred checks:
+#! - pointers are double-word aligned and backed by u32-range felts;
+#! - Q is a canonical secp256k1 curve VALUE digest built from the public-key limbs;
+#! - SIG_R, SIG_S, and Z are canonical k1 scalar VALUE digests built from memory limbs;
+#! - raw z < 2^256 < 2n, so one subtraction is enough after loading the prehash;
+#! - SIG_R != 0 is checked here, and k1_scalar::inv traps/binds correctness for SIG_S != 0;
+#! - U1/U2 are passed as scalar digests to the curve scalar multiplication wrappers;
+#! - scalar/curve equality constraints are logged through deferred precompile machinery.
+@locals(0)
```
adr1anh: Remove all `@locals(0)`

crates/lib/core/precompiles/asm/crypto/dsa/ecdsa_secp256k1.masm
```diff
@@ -0,0 +1,398 @@
...
+#! Memory layout uses math-native little-endian u32 limb encoding:
+#! - pubkey_ptr[0..8]  = qx, pubkey_ptr[8..16] = qy
+#! - digest_ptr[0..8]  = raw 256-bit prehash z
+#! - sig_ptr[0..8]     = SIG_R, sig_ptr[8..16] = SIG_S
+#!
+#! Assumptions/deferred checks:
+#! - pointers are double-word aligned and backed by u32-range felts;
+#! - Q is a canonical secp256k1 curve VALUE digest built from the public-key limbs;
+#! - SIG_R, SIG_S, and Z are canonical k1 scalar VALUE digests built from memory limbs;
+#! - raw z < 2^256 < 2n, so one subtraction is enough after loading the prehash;
```
adr1anh: Is this still correct?

crates/lib/core/precompiles/asm/crypto/dsa/ecdsa_secp256k1.masm
```diff
@@ -0,0 +1,398 @@
...
+
+    swapw
+    # => [U1_DIGEST, Q_DIGEST, U2_DIGEST, ...]
+    movupw.2
+    # => [U2_DIGEST, U1_DIGEST, Q_DIGEST, ...]
+    swapw
+    # => [U1_DIGEST, U2_DIGEST, Q_DIGEST, ...]
+
+    exec.secp256k1::msm2_generator
+    # => [MSM_POINT_DIGEST, ...]
```
adr1anh: Should we change the ABI of this function or the inner one so that we have fewer stack suffling ops

crates/lib/core/precompiles/asm/crypto/dsa/ecdsa_secp256k1.masm
```diff
@@ -0,0 +1,398 @@
...
+#!
+#! Preconditions for callers:
+#! - VALUE_U32[8] is already u32-constrained;
+#! - VALUE < 2n, so one subtraction is sufficient.
+#!
+#! The modulus is staged once into this proc's locals and passed by pointer to loop-shaped helpers.
+#! The branch condition is the final borrow of VALUE - n, computed on a duplicate of VALUE so the
+#! original remains available for the conditional subtraction.
+@locals(8)
+proc reduce_u256_mod_k1_scalar_unchecked
```
adr1anh: We should actually be able to do this using the uint/field method which uses precomputed powers of 2^128

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -1,7 +1,8 @@
-use miden::core::crypto::hashes::keccak256
-use miden::core::crypto::hashes::poseidon2
-use miden::core::sys
-use miden::core::word
+#! Compatibility skeleton for the legacy `::miden::core::crypto::dsa::ecdsa_k256_keccak` API.
+#!
+#! This module restores namespace/API stability for callers that import the old core crypto path.
+#! The procedures below intentionally trap instead of returning fake verification results. They
+#! will delegate to the corresponding `::miden::precompiles::*` implementations later.
```
adr1anh: This seems unnecessary to mention "compatibility" ?

crates/test-utils/src/lib.rs
```diff
@@ -85,6 +86,17 @@ pub fn module_source(path: impl AsRef<Path>, source: impl ToString) -> String {
     }
 }
 
+#[cfg(not(target_family = "wasm"))]
+fn new_vm_default_processor(
```
adr1anh, line 91: Document, saying this should be used instead of new, maybe mentioning that this can be revisited when precompiles does not depend on processor

precompiles/src/dsa/mod.rs
```diff
@@ -0,0 +1,16 @@
+//! Digital-signature MASM wrappers built on the native math/hash precompiles.
```
huitseeker: `cargo shear` flags this as an empty file because it only has module docs. Can this text move to `precompiles/src/lib.rs`, or can this module grow a real item so the new crate passes shear?

verifier/Cargo.toml
```diff
@@ -19,13 +19,14 @@ doctest = false
...
 default = ["std"]
-std = ["miden-air/std", "miden-core/std", "thiserror/std"]
+std = ["miden-air/std", "miden-core/std", "miden-precompiles/std", "thiserror/std"]
 
 [dependencies]
 # Miden dependencies
 miden-air.workspace = true
 miden-core.workspace = true
 miden-crypto.workspace = true
+miden-precompiles.workspace = true
```
huitseeker, line 29: `cargo shear` also fails on this crate because `verifier/Cargo.toml:35` still lists `tracing`, but the verifier code no longer uses it. The failing line is not new in the diff, so I am anchoring this on the nearby dependency change.

crates/lib/core/Cargo.toml
```diff
@@ -66,6 +72,7 @@ fs-err = { version = "3.1" }
 miden-assembly = { workspace = true, features = ["std"] }
 miden-assembly-syntax = { workspace = true, features = ["std"] }
 miden-package-registry = { workspace = true, features = ["resolver"] }
+miden-precompiles = { workspace = true, features = ["std"] }
```
huitseeker: `cargo shear` is still failing in this crate after the test reshuffle: `miden-prover` and `miden-verifier` are now unused in `crates/lib/core/Cargo.toml` at lines 61 and 63. Those lines are not new in the diff, so I am anchoring this on the nearby changed dependency block.

processor/src/trace/mod.rs
```diff
@@ -106,14 +83,9 @@ impl TraceBuildInputs {
...
-    pub fn precompile_requests(&self) -> &[PrecompileRequest] {
-        &self.trace_output.precompile_requests
-    }
-
-    /// Returns the final precompile transcript observed during execution.
-    pub fn final_precompile_transcript(&self) -> &PrecompileTranscript {
-        &self.trace_output.final_precompile_transcript
+    /// Returns the final deferred state captured for the execution being replayed.
+    pub fn deferred_state(&self) -> &DeferredState {
+        &self.trace_output.deferred_state
```
huitseeker, line 88: CI is failing just below this changed block. `RUSTFLAGS=-D warnings cargo check --all-targets --manifest-path processor/Cargo.toml --no-default-features` reports `TraceBuildInputs::{into_parts, from_parts}` as dead code. Since this patch changed the trace-output surface from precompile requests to deferred state, can the stale test-only helpers be moved next to the tests that still need them, or get an `expect(dead_code)` that covers this target?

precompiles/src/math/curve/mod.rs
```diff
@@ -0,0 +1,1058 @@
+//! Fixed curve deferred precompile backed by fixed uint coordinate domains.
```
huitseeker: Can we split this part out of the hash/signature precompile review? The current PR also carries the deferred DAG substrate, proof/public-input changes, uint and curve arithmetic families, generated MASM, docs, fuzz, and workflow churn. 

Having just the precompile crate and compatibility migration would be much easier to reason about separately.

precompiles/asm/miden-project.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "miden-precompiles"
+version = "0.24.0"
```
huitseeker: This looks like old package metadata carried forward after the workspace moved to `0.25.0`. Can the new `miden-precompiles` MASM package and the `miden-core` dependency on it use the current version, or is there a reason standalone MASM package resolution should stay on `0.24`?

core/src/proof.rs
```diff
@@ -22,29 +22,30 @@ use crate::{
...
+/// deferred-state wire opening needed to rehydrate the deferred DAG under a verifier-supplied
+/// precompile registry. However, the proof does not contain public inputs needed to verify the
+/// STARK proof.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct ExecutionProof {
     pub proof: Vec<u8>,
     pub hash_fn: HashFunction,
-    pub pc_requests: Vec<PrecompileRequest>,
+    pub deferred_state: DeferredStateWire,
```
huitseeker, line 34: This is the proof-format cutover: `ExecutionProof` now carries deferred wire instead of legacy precompile requests. That seems bigger than the precompile crate review itself. Do you want this to stay in this PR, or should the proof/public-input migration be split so it gets reviewed as a protocol boundary change?

precompiles/src/codegen/masm.rs
```diff
@@ -0,0 +1,468 @@
...
+
+/// Runs write (`--write`) or staleness-check (`--check`) mode for generated precompile MASM.
+pub fn run(mode: Mode) -> Result<(), String> {
+    match mode {
+        Mode::Check => check(),
+        Mode::Write => write(),
+    }
+}
+
+/// Checks checked-in generated MASM files against the fixed codegen descriptors.
```
huitseeker: Can the generated MASM/codegen part be split from the semantic precompile review? Reviewing the descriptors/templates is useful, but this PR also adds thousands of generated MASM lines. If we use `build.rs` to generate those on build, this would make the actual logic easier to review.

I might look into this myself.

huitseeker: https://github.com/0xMiden/miden-vm/pull/3304

adr1anh: #3304 is a lot nicer, thanks!

verifier/src/lib.rs
```diff
@@ -56,70 +54,25 @@ pub use exports::*;
...
+        &deferred_wire,
+        usize::MAX,
```
huitseeker: This drops the verifier-side budget for the deferred wire. Execution uses `max_deferred_elements` with a default cap of `1 << 20`, but `verify()` rebuilds the wire with `usize::MAX`.

Can we use a finite verifier cap, or expose a verifier entrypoint that takes one?

precompiles/src/lib.rs
adr1anh: Addressed: `PrecompilesLibrary` now exposes only `package()` as the public package accessor; `mast_forest()` and `AsRef<Package>` were removed. Resolving.

docs/src/design/deferred/index.md
adr1anh: Addressed: the deferred design docs now document `PairList` payloads, including pair ordering, canonical wire encoding, arity checks, and budget behavior. Resolving.

precompiles/asm/miden-project.toml
adr1anh: Addressed: bumped the `miden-precompiles` MASM package metadata to `0.25.0`. Resolving.

precompiles/asm/crypto/dsa/eddsa_ed25519.masm
adr1anh: This EdDSA wrapper no longer exists in the current PR, so the verifier path this comment was reviewing has been removed rather than fixed here. Resolving as obsolete.

precompiles/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh: Addressed in the current wrapper: the legacy `verify` path computes `Keccak256(MSG)` and passes that prehash into `ecdsa_secp256k1::assert_verify_prehash`; the raw verifier entrypoint is explicitly prehash-based. Resolving this stale thread.

precompiles/src/math/weierstrass/mod.rs
adr1anh: Stale after the curve/precompile refactor; this code path no longer exists and current clippy CI is passing. Resolving.

precompiles/src/hash/mod.rs
adr1anh: Addressed: the hash event handler now checks `process.execution_options().max_hash_len_bytes()` before reading/hashing memory. Resolving.

precompiles/src/codegen/masm.rs
adr1anh: Addressed: the generic `extra_sections` hook and related helpers/tests were removed. Resolving.

precompiles/src/codegen/masm.rs
adr1anh: Addressed: generated MASM codegen no longer carries manual cycle-count replacement tables; cycle baselines live in tests. Resolving.

precompiles/src/math/uint/precompile.rs
adr1anh: Stale after the codegen/test reshuffle; the duplicated MASM const parser is no longer present. Resolving.

precompiles/src/math/uint/precompile.rs
adr1anh: Addressed: the uint precompile now uses fixed `UintDomain` dispatch rather than a dynamic registry/builder API. Resolving.

precompiles/src/math/weierstrass/mod.rs
adr1anh: Addressed: the curve precompile now uses fixed `CurveId` dispatch for supported curves instead of the old dynamic registry/builder shape. Resolving.

precompiles/src/math/weierstrass/mod.rs
adr1anh: Addressed by the curve refactor; `curve_for<C>()` is gone. Resolving.

precompiles/src/lib.rs
adr1anh: Addressed: `FastProcessor::with_deferred_precompiles` is now a public processor hook, and the high-level VM/prover/CLI/test entry points install `miden_precompiles::registry()` by default. Resolving.

crates/lib/core/asm/crypto/dsa/eddsa_ed25519.masm
adr1anh: Addressed by removal: the old `miden::core` Ed25519 compatibility export is no longer present in the current tree, so there is no trapping compatibility wrapper left in this PR. Resolving.

processor/src/trace/mod.rs
adr1anh: Addressed. The stale helpers are gone/guarded, and `RUSTFLAGS="-D warnings" cargo check --all-targets --manifest-path processor/Cargo.toml --no-default-features` passes. Resolving.

precompiles/src/codegen/masm.rs
adr1anh: Addressed by the build/codegen change: math MASM is generated during the precompiles build, so the checked-in package no longer carries the generated math MASM files. Resolving.

precompiles/src/math/curve/mod.rs
```diff
@@ -0,0 +1,1052 @@
...
+            assert_eq!((x, y), expected);
+        }
+    }
+
+    #[test]
+    fn ed25519_affine_identity_canonicalizes() {
+        const ONE_LIMBS: Limbs = [1, 0, 0, 0, 0, 0, 0, 0];
+
+        let curve = CurveId::Ed25519;
+        assert_eq!(curve.point_from_affine(ZERO_LIMBS, ONE_LIMBS).unwrap(), CurvePoint::Identity);
```
huitseeker: This fails on the current head. `Ed25519::point_from_affine(0, 1)` now goes through the short-Weierstrass equation and returns `InvalidPayload`, so `cargo test -p miden-precompiles ed25519_affine_identity_canonicalizes -- --nocapture` fails here. 

I also thought we were removing Ed25519 support?

crates/lib/core/asm/miden-project.toml
```diff
@@ -1,11 +1,14 @@
...
 name = "miden-core"
-version = "0.22.0"
+version = "0.24.0"
 
 [lib]
 namespace = "miden::core"
 path = "mod.masm"
 
+[dependencies]
+miden-precompiles = { version = "0.24", linkage = "dynamic" }
```
huitseeker: One half of the version bump still looks stale. `precompiles/asm/miden-project.toml` now declares `miden-precompiles` as `0.25.0`, but this dependency still asks package resolution for `miden-precompiles` `0.24`. This should use the current `0.25` version too

verifier/src/lib.rs
adr1anh: Addressed in `a77b74f60`.

`verify()` now rehydrates the deferred wire with the shared finite default deferred-state budget instead of `usize::MAX`. I also added `verify_with_max_deferred_elements(...)` for proofs produced with a non-default execution budget. I avoided a generic options type since this is currently a single verifier knob.

crates/lib/core/asm/miden-project.toml
adr1anh: Addressed in `52998b71e`.

Updated `crates/lib/core/asm/miden-project.toml` to depend on `miden-precompiles` `0.25`, matching `precompiles/asm/miden-project.toml`.

precompiles/src/math/curve/mod.rs
adr1anh: Addressed in `e018b5af2`.

EdDSA support remains removed, but I kept the third fixed curve for precompile VM coverage. To make the model explicit, I renamed the curve surface from `Ed25519`/`ed25519` to `Ed25519Sw`/`ed25519_sw`, including the generated MASM wrapper name. The stale Edwards affine identity assumption `(0, 1)` is gone; the test now checks identity behavior in the short-Weierstrass representation.

`cargo test -p miden-precompiles` passes.

precompiles/src/math/uint/arithmetic.rs
Nashtare, line 1: I don't think we're using this file even in the follow-up PR, so effectively dead code. Worth removing you reckon?

core/src/deferred/wire.rs
```diff
@@ -0,0 +1,611 @@
...
+                    .payload()
+                    .as_data()
+                    .map_err(|_| IntegrityError::InvalidStructure)?
+                    .to_vec(),
+            },
+            NodeType::Join => {
+                let (lhs, rhs) =
+                    node.payload().as_join().map_err(|_| IntegrityError::InvalidStructure)?;
+                self.visit_state_digest(state, lhs)?;
+                self.visit_state_digest(state, rhs)?;
```
Nashtare: Are we bounding recursion depth somewhere? This worries me a little, seems similar to an issue we had in the P3 recursion in the past we're we'd get verifier DOS. AFAICT recursion is bounded by tree depth only (we're only deduping shared subgraphs).

I think what may be the cleanest fix is to make `visit_state_digest` iterative

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -1,196 +1,306 @@
...
+    # => [PK_COMM, MSG, ...]
 
-    adv_loadw loc_storew_le.4 dropw
-    adv_push loc_store.8
+    locaddr.0 mem_load u32and.255
+    # => [PK8, PK_COMM, MSG, ...]
 
-    # Compute the expected commitment and ensure it matches the provided PK_COMM
-    push.PK_LEN_FELTS locaddr.0 exec.poseidon2::hash_elements
+    # Assert Poseidon2::hash_elements(PK[9]) == PK_COMM.
```
adr1anh: The upstream `miden-crypto` ECDSA commitment format has changed: `PublicKey::to_commitment()` now commits directly to native affine coordinates as `Poseidon2::hash_elements(QX[8] || QY[8])`, rather than to the compressed SEC1 `PK[9]` encoding. So this wrapper should no longer rebuild the compressed SEC1 key in the verify hot path just to bind `PK_COMM`. We can simplify this by absorbing the two `adv_pipe` chunks for `QX` and `QY` directly into Poseidon2 while storing them to the precompile ABI buffer, then compare the squeezed digest to `PK_COMM`. I pushed an implementation of this approach on `precompiles/ecdsa-pk`. It also keeps SEC1 compression available separately as a public core helper, `::miden::core::crypto::dsa::ecdsa_k256_keccak::compress_public_key`, for callers that still need the packed compressed format.

huitseeker:
Note to self: as this introduces new crates, this will need integrating with the tooling of #3280 which is about to be merged.

verifier/src/lib.rs
adr1anh: This is now intentionally fixed in the other direction: the public VM/prover/verifier path does not support custom deferred precompile registries.

`FastProcessor::new_with_options()` initializes `DeferredState` with `miden_precompiles::registry()`, and the verifier rehydrates the proof-carried deferred wire with the same built-in registry. The old `with_deferred_precompiles` hook is gone, so proofs using non-standard deferred semantics are not part of the supported public API.

`PrecompileRegistry` remains generic at the lower deferred-state layer, but verification is deliberately constrained to the built-in Miden precompile registry so a proof only verifies if its deferred wire rehydrates under the official semantics.

adr1anh:
Tooling integration for the new crates is addressed by c5312efc3 (`ci: include precompile crates in release list`).

precompiles/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh: Addressed in the current native-coordinate verifier path.

The wrapper binds the advice-provided public key coordinates to `PK_COMM`, and the raw verifier then loads `Q` through the secp256k1 curve precompile and `SIG_R`/`SIG_S` through the k1 scalar precompile. Those paths validate the curve point and scalar encodings before the deferred equality constraints are accepted. `SIG_R` is checked nonzero here, and `SIG_S` is checked through scalar inversion.

So the current verifier is not relying only on the old deferred-root assumption; the relevant curve/scalar values are validated by the precompile-backed path.

miden-core-fuzz/Cargo.toml
adr1anh: Leaving this as a follow-up rather than blocking this PR.

The existing targets cover deferred-wire deserialization, but a useful semantic fuzz target needs more thought: ideally it would generate valid and partially-valid deferred states/wires, possibly starting from states built with the default precompile registry. We should do a focused pass, potentially with an agent, to understand the existing fuzz target patterns before adding that.

crates/lib/core/precompiles/asm/hashes/keccak256.masm
adr1anh: This is fine as-is for this PR.

The public `miden::core` Keccak API should stay stable and easy to use. The lower-level `miden::precompiles` module is intentionally separate so callers can tell they are using lower-level precompile plumbing, where a more specialized/optimized ABI is acceptable. We can revisit the low-level Keccak ABI as a focused cycle optimization later.

precompiles/src/math/curve/mod.rs
adr1anh: Discussed async. We are trying to reduce/split the broadening where practical, but the deferred substrate, proof binding, and precompile wrappers are coupled enough that this PR still carries the full cutover. Resolving this stale scope thread.

core/src/proof.rs
adr1anh: Discussed async. This PR is intentionally doing the full proof/public-input cutover: `ExecutionProof` carries the deferred wire so verification can rehydrate it and bind the resulting deferred root. We should call this out in the PR/changelog as a protocol-boundary change.

precompiles/src/math/uint/arithmetic.rs
adr1anh: Keeping this: it is used by the uint spec now. `precompiles/src/math/uint/mod.rs` declares `mod arithmetic`, and `spec.rs` imports the arithmetic helpers for canonical checks, add/sub/mul, Barrett reduction, and inversion.

core/src/deferred/wire.rs
adr1anh: Addressed: `visit_state_digest` is iterative now. It uses an explicit `pending: Vec<WireEncodeStep>` work stack rather than recursive traversal.

precompiles/src/dsa/mod.rs
adr1anh: Addressed by refactor/removal: `precompiles/src/dsa/mod.rs` no longer exists, so there is no docs-only empty module for `cargo shear` to flag.

verifier/Cargo.toml
adr1anh: Addressed: `verifier/Cargo.toml` no longer lists `tracing`.

crates/lib/core/Cargo.toml
adr1anh: Addressed: `crates/lib/core/Cargo.toml` no longer lists `miden-prover` or `miden-verifier` in this dependency block.

precompiles/src/math/curve/mod.rs
adr1anh: This is obsolete after the later cleanup: the Ed25519/EdDSA surface has been removed from this PR. We should make sure the PR changelog calls out the breaking removal of EdDSA, along with SHA-512 being removed for now and reintroduced once it is supported by the precompiles prover.

release-packages.txt
```diff
@@ -19,8 +19,10 @@ miden-package-registry
 miden-project
 miden-assembly
 miden-processor
+miden-precompiles
```
huitseeker: This needs to move above `miden-processor`. `processor/Cargo.toml` now depends on `miden-precompiles`, and the publish path reads `release-packages.txt` in order, so a real publish can try `miden-processor` before its new crate exists at the target version.

huitseeker (commented):
@adr1anh We've moved to infra that allows us to release packages at various versions, so we need to keep a file that contains the packages in topological order.

adr1anh:
Addressed in the prover-stack branch via f9616f73b: `release-packages.txt` now keeps the publish list in topological dependency order and places the precompile crates before their dependents: `miden-precompiles -> miden-precompiles-prover -> miden-processor -> miden-prover`.

huitseeker (approved):
This looks really good to me now, thanks @adr1anh ! This will still need a little bit of rebase agility (#3252 for instance is on its way to next, obviating the release-packages.txt changes that I'm sorry to have bugged you about), but overall it's good to go!

core/src/deferred/node.rs
```diff
@@ -0,0 +1,796 @@
...
+
+    /// Shape for a data node with a non-zero semantic chunk count.
+    pub fn data_chunks(n: u32) -> Option<Self> {
+        (n != 0).then_some(Self::Data)
+    }
+
+    /// Shape for a pair-list node with a non-zero semantic pair count.
+    pub fn pair_list(n: u32) -> Option<Self> {
+        (n != 0).then_some(Self::PairList)
+    }
```
Al-Kindi-0: As written, `n` does nothing here, its only use is a non-zero check, and zero-length payloads are already unconstructible (`Payload::try_data` / `try_pair_list_chunks` reject empty collections at construction). Meanwhile the `NodeType` doc says shapes intentionally don't carry arity.

precompiles/src/math/curve/mod.rs
```diff
@@ -0,0 +1,1207 @@
...
+                let y = UintPrecompile::limbs_from_value_node(y_node, curve.base_domain())?;
+                curve.point_from_affine(x, y)
+            },
+        }
+    }
+
+    /// Trusted decoder for a payload of an already-canonical curve `VALUE` node.
+    ///
+    /// Canonical curve `VALUE` nodes are produced by this precompile, so release builds rely on the
+    /// prior checked boundary instead of revalidating curve membership here. Structural checks
```
Al-Kindi-0, line 726: Could we add a regression test attesting to that failure at registration of an off-curve value ?

core/src/events/sys_events.rs
```diff
@@ -292,6 +292,112 @@ pub enum SystemEvent {
...
+    /// (`[2, 0, 0, 0]`) is framework-owned opaque data; malformed id-2 tags are rejected during tag
+    /// decode. The installed registry decodes `TAG` via
+    /// [`crate::deferred::DeferredState::decode`]; `TRUE` is not accepted by this event. Tags that
+    /// semantically require more data chunks or pairs are rejected during precompile-specific
+    /// evaluation. Registration is performed by [`crate::deferred::DeferredState::register`], so
+    /// semantic failures surface immediately.
+    ///
+    /// This event does not push advice and does not return the node digest. Assembly code that
+    /// later uses that digest must compute it in-circuit from the same `TAG` and payload.
+    ///
```
Al-Kindi-0: nit: "in-circuit" seems off here, better probably "in-VM"

core/src/events/sys_events.rs
```diff
@@ -292,6 +292,112 @@ pub enum SystemEvent {
...
+    ///   Operand stack: unchanged
+    ///   Advice stack:  canonical payload only, word-ordered for `adv_pushw` LIFO consumption
+    DeferredEvaluatePayload,
+
+    /// Registers and eagerly evaluates a memory-backed deferred node.
+    ///
+    /// `TAG` is one word (4 field elements). The installed registry decodes `TAG` to determine the
+    /// memory-backed payload shape. The stack-supplied `n_chunks` determines the memory range and
+    /// is shared with any in-circuit digest loop. Data tags read exactly `n_chunks` 8-felt
+    /// [`crate::deferred::DataChunk`] values from memory starting at `ptr`; exact
```
Al-Kindi-0: I would write things here, and in other places, in a more readable and less terse style, for example:

The host reads `n_chunks` chunks from memory at `ptr`; nothing in this event constrains that read. A program that relies on the registered node must recompute the node digest itself in MASM — a Poseidon2 linear hash over the same memory range (`n_chunks` absorptions with the tag word as the initial capacity, as `register_mem` does). If the event's range and the hashed range disagree, the registered node is not bound to anything in the trace.

crates/lib/core/codegen/src/templates/curve.masm.tpl
```diff
@@ -0,0 +1,387 @@
...
+
+#! Registers an MSM PairList staged in memory.
+#! Input:  [ptr, n, ...]
+#! Output: [MSM_POINT_DIGEST, ...]
+#! Memory layout: pair i at ptr + 8*i is `[POINT_DIGEST, SCALAR_DIGEST]`.
+pub proc msm_mem
+    push.0
+    push.0
+    push.MSM_OP_ID
+    push.PRECOMPILE_ID
```
Al-Kindi-0: Q: why not `push.MSM_TAG` ?

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -1,198 +1,279 @@
...
+#!   - loc[16..24] : Keccak256 message/digest scratch buffer.
+#!
+#! No result word is pushed. The procedure traps on malformed native limbs, invalid/off-curve public
+#! keys, invalid signature scalars, or failed signature verification.
 #!
-#! The procedure traps if:
-#! - The public key does not hash to `PK_COMM` (invalid commitment)
-#! - The signature verification fails
-@locals(52)
+#! Low-s note: the signature only needs to satisfy the ECDSA equation with canonical, nonzero scalar
```
Al-Kindi-0: Q: Don't we reject high-s cases in `miden-crypto` ? If yes, shouldn't we do the same here ? Or is this postponed for a followup work ?

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -1,198 +1,279 @@
...
+    # alongside the coordinate payload, so direct comparison to the registered identity VALUE digest
+    # is sound and avoids a second curve evaluation.
+    exec.secp256k1::push_identity
+    # => [IDENTITY_VALUE_DIGEST, VERIFY_POINT_VALUE_DIGEST, X_OR_TRUE_DIGEST, Y_OR_TRUE_DIGEST, SIG_R_SCALAR_VALUE_DIGEST, ...]
+    eqw
+    # => [is_identity, IDENTITY_VALUE_DIGEST, VERIFY_POINT_VALUE_DIGEST, X_OR_TRUE_DIGEST, Y_OR_TRUE_DIGEST, SIG_R_SCALAR_VALUE_DIGEST, ...]
+    assertz.err="invalid ECDSA verification point: identity"
+    # => [IDENTITY_VALUE_DIGEST, VERIFY_POINT_VALUE_DIGEST, X_DIGEST, Y_DIGEST, SIG_R_SCALAR_VALUE_DIGEST, ...]
+    dropw dropw
+    # => [X_DIGEST, Y_DIGEST, SIG_R_SCALAR_VALUE_DIGEST, ...]
```
Al-Kindi-0: Here, and in other places, the following might save a few cycles:

```
#! Asserts two words are unequal, consuming both.
#!
#! Input:  [I, V, ...]  where I = [i0, i1, i2, i3], V = [v0, v1, v2, v3]
#! Output: [...]
#!
#! Traps iff all four lanes are equal (I == V).
proc assert_neqw
    # => [i0, i1, i2, i3, v0, v1, v2, v3, ...]
    movup.4 eq
    # => [b0, i1, i2, i3, v1, v2, v3, ...]          b0 = (v0 == i0)
    movup.4 movup.2 eq and
    # => [b01, i2, i3, v2, v3, ...]                 b01 = b0 & (v1 == i1)
    movup.3 movup.2 eq and
    # => [b012, i3, v3, ...]                        b012 = b01 & (v2 == i2)
    movup.2 movup.2 eq and
    # => [b0123, ...]                               b0123 = 1 iff I == V
    assertz.err="words are equal"
    # => [...]
end
```

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -1,198 +1,279 @@
...
+    swapw dropw
+    # => [X_DIGEST, SIG_R_SCALAR_VALUE_DIGEST, ...]
+
+    exec.k1_base::eval
```
Al-Kindi-0: The current state of my understanding is that `eval` logs an `eq(EXPR, VALUE)` when the host claims an answer. But called on a digest that already names a (claimed) canonical VALUE node, the logged statement is `eq(X, X)`, i.e., redundant.

This is what happens here, the verify point is already pinned to the MSM by the `eq` that `secp256k1::eval` logged a few lines above, and canonicity is hereditary in the sense that child closure forces the coordinate nodes to be registered first, and registration validates VALUE encodings and hence they're canonical. So materializing a coordinate isn't a new claim, it's just an opening of an existing commitment — and openings are checkable by the Miden VM itself, nothing needs to be deferred. Yet the second `eval` pays a deferred fold, an EQ+AND registration, wire entries, and a verifier re-evaluation for a tautology.

This suggests introducing a value-side wrapper (e.g., `materialize_value`) that advises the payload, re-hashes it with the VALUE tag, and does `assert_eqw` against the input digest.

Al-Kindi-0 (commented):
Did an initial pass and I couldn't find any major issues. Left a few comments which could improve things.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh: This isn't actually required, due to the way we are validating signatures. The verify function loads them from the advice, they are untrusted data. The low-S is unnecessary since a cheating prover could always flip the sign of the value to make the check pass. 

The only case where it is needed is if we are validating a specific Ethereum  signature, whose byte representation is bound to a specific context. Our interface doesn't currently support this though, and I’m not sure we actually need it.

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -383,261 +371,33 @@ mod fast_parallel {
     }
 
     #[test]
-    fn test_prove_from_trace_sync_preserves_precompile_requests() {
-        let LoggedPrecompileProofFixture {
```
Nashtare, line 387: all those tests were removed (including `log_precompile_request_procedure()` in `sys` module as well, to be replaced with proving "begin log_deferred end", and I believe we only run `execute_sync` for more complex ones involving Keccak / uint / etc..? Maybe would be worth including full e2e proof generation

Nashtare: Also two other comments, maybe worth as follow-up strenghtening pass:
- I don't think we're ever testing the MASM recursive verifier with a non-zero deferred root?
- I don't see tampering tests (to make sure we'd catch integrity issues after mutating the deferred wire)

verifier/src/lib.rs
Nashtare: > verification is deliberately constrained to the built-in Miden precompile registry so a proof only verifies if its deferred wire rehydrates under the official semantics.

This may lead to a slight DOS issue for the verifier though I believe? We may want in-depth protection, i.e. maybe something like wire byte-cap analogous to `MAX_STARK_PROOF_BYTES`, otherwise the current may induce couple minutes of verifier CPU time before being caught by the STARK machinery

Nashtare (approved):
I had a couple pending comments addressed in the last commit, so overall LGTM.
My main concern is about lacking extensive negative / adversarial tests but this could be bundled in the follow-up PRs.

CHANGELOG.md
```diff
@@ -35,6 +36,15 @@
...
+- [BREAKING] Removed the `miden::core::crypto::dsa::eddsa_ed25519` MASM module, Rust handler, docs, and tests. EdDSA support is temporarily removed from core-lib and will be reintroduced once it is supported by the precompiles prover.
+- [BREAKING] Removed the `miden::core::crypto::hashes::sha512` MASM module, Rust handler, docs, and tests. SHA-512 support is temporarily removed from core-lib and will be reintroduced once it is supported by the precompiles prover.
+- [BREAKING] Changed the `miden::core::crypto::dsa::ecdsa_k256_keccak` advice/signature ABI to `QX[8] || QY[8] || SIG_R[8] || SIG_S[8]` as little-endian u32 field elements. Existing 65-byte signature advice must be re-encoded as `(r, s)` limbs without a recovery byte.
+- [BREAKING] Removed the public `miden::core::crypto::dsa::ecdsa_k256_keccak::verify_prehash` and raw `miden::precompiles::crypto::dsa::ecdsa_secp256k1::assert_verify_prehash` ECDSA prehash verifier entrypoints. ECDSA K256 Keccak verification is now exposed only through the high-level `verify` procedure, whose implementation inlines the verifier, loads signature scalars directly from advice, and avoids the raw prehash memory ABI.
```
Nashtare: nit: should be moved to v0.26

core/src/deferred/node.rs
adr1anh: Resolved in `0bc438b55`: removed the `value`, `data_chunks(n)`, and `pair_list(n)` helpers. `NodeType` now carries shape only; constructors enforce nonempty payloads and precompiles validate semantic lengths.

precompiles/src/math/curve/mod.rs
adr1anh: Added `off_curve_affine_value_fails_at_registration`, which directly asserts that `state.register(point)` rejects the off-curve affine value.

core/src/events/sys_events.rs
adr1anh: Updated this to “inside the VM” and removed the corresponding “in-circuit” terminology from the related core documentation.

core/src/events/sys_events.rs
adr1anh: Reworked this section and the related deferred-event docs to spell out the selected range, lack of an AIR binding, the `register_mem` hash construction, and the consequence of hashing a different sequence.

crates/lib/core/codegen/src/templates/curve.masm.tpl
adr1anh: Changed `msm_mem` to use `push.MSM_TAG`.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh: Replaced the preserving `eqw` plus cleanup with the existing consuming `word::eq` helper here and at the related equality call sites, followed by `assertz` where inequality is required.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
adr1anh: Implemented this as `open_value`: it advises the payload, re-hashes it with the domain’s `VALUE_TAG`, and asserts digest equality inside the VM. The ECDSA path now uses it for `X_DIGEST`; tests confirm it neither advances the deferred root nor adds wire entries.

miden-vm/tests/integration/prove_verify.rs
adr1anh: Agreed. I’ve captured the remaining full-proof and deferred-wire tampering coverage in #3378, with the broader recursive-verifier work also tracked in #3250.

verifier/src/lib.rs
adr1anh: Agreed. I expanded #3378 to explicitly cover enforcing the deferred-wire budget before decoder preallocation, traversal, hashing, or payload cloning, including proofs supplied directly as an in-memory `ExecutionProof`.

CHANGELOG.md
adr1anh: Good catch. Moved the #3222 entries from `v0.25.0` to the unreleased `v0.26.0` Changes section in `950488812`.

PR #3223, opened 2026-06-03
Add Keccak-256 and SHA-512 precompiles

adr1anh:
## Summary

Ports Keccak-256 and SHA-512 deferred precompiles into `miden-precompiles`.

Keccak and SHA-512 are handled together because they share the same MASM wrapper shape, Rust precompile structure, byte/chunk codec logic, and package tests.

## Changes

- Adds hash precompile registry entries for Keccak-256 and SHA-512.
- Adds MASM wrappers under `miden::precompiles::crypto::hashes`.
- Adds shared hash codec/precompile logic.
- Adds package export and end-to-end execution coverage.

## Design Notes

- Hash precompiles live outside the core library namespace.
- Programs link the MASM package via `PrecompilesLibrary`.
- Processor execution explicitly installs the Rust deferred registry.

## Tests

- `cargo test -p miden-precompiles`


adr1anh:
Absorbed into #3222; precompiles/crate now carries the hash and signature precompile payload.

PR #3226, opened 2026-06-03
chore(processor): Add an execution option to bound the memory size

plafer:
Fixes a security vulnerability where the processor's memory was still unbounded.

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -26,6 +26,9 @@ pub struct ExecutionOptions {
     /// Maximum number of field elements allowed on the operand stack in the active execution
     /// context.
     max_stack_depth: usize,
+    /// Maximum number of field elements allowed in the processor's memory at any point during
+    /// execution, rounded up to the nearest multiple of 4.
+    max_memory_elements: usize,
 }
 
 impl Default for ExecutionOptions {
@@ -39,6 +42,7 @@ impl Default for ExecutionOptions {
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
             max_stack_depth: Self::DEFAULT_MAX_STACK_DEPTH,
+            max_memory_elements: Self::DEFAULT_MAX_MEMORY_ELEMENTS,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


processor/src/fast/mod.rs
```diff
@@ -264,7 +264,7 @@ impl FastProcessor {
             clk: 0_u32.into(),
             ctx: 0_u32.into(),
             caller_hash: EMPTY_WORD,
-            memory: Memory::new(),
+            memory: Memory::new(options.max_memory_elements().div_ceil(WORD_SIZE)),
```
huitseeker: Could this also update the memory limit in `with_options()`? This path builds `Memory` from `options.max_memory_elements()`, but `FastProcessor::new(...).with_options(...)` keeps the `Memory` created by `new()`, so the requested cap is ignored.

I reproduced it with this shape:

```rust
let options = ExecutionOptions::default().with_max_memory_elements(3 * WORD_SIZE);
let processor = FastProcessor::new(StackInputs::default()).with_options(options).unwrap();
let err = processor.execute_sync(&program, &mut host).unwrap_err();
```

The test returned `Ok` instead of `MemoryElementLimitExceeded`, and the debug output still showed the default `max_entries: 67108864`.

processor/src/fast/memory.rs
```diff
@@ -19,15 +19,32 @@ use crate::{ContextId, MemoryAddress, MemoryError, processor::MemoryInterface};
...
 }
 
 impl Memory {
-    /// Creates a new memory instance.
-    pub fn new() -> Self {
-        Self::default()
+    /// Creates a new memory instance allowing at most `max_entries` word entries.
+    ///
+    /// To derive `max_entries` from an element-addressable limit, divide it by [`WORD_SIZE`].
+    pub fn new(max_entries: usize) -> Self {
```
huitseeker: Could `Memory::new` keep taking the element limit and convert to word entries internally? The new execution option is named in elements, but this constructor now takes word entries, so callers have to remember to do `div_ceil(WORD_SIZE)` themselves.

Keeping that unit conversion inside `Memory` would simplify the call sites and make it harder to pass the wrong limit.

huitseeker (changes requested):
I think we're missing cap enforcement on one `Memory` builder, see inline.

processor/src/fast/mod.rs
plafer: Good catch; fixed

processor/src/fast/memory.rs
plafer: Done

processor/src/execution_options.rs
```diff
@@ -227,6 +241,13 @@ impl ExecutionOptions {
         self.max_stack_depth
     }
 
+    /// Returns the maximum number of field elements allowed in the processor's memory (rounded to
```
huitseeker: Nit: Could this doc say it returns the raw configured value instead? `with_max_memory_elements(5)` stores `5`, and `max_memory_elements()` returns `5`; only `Memory::new` / `set_max_elements` rounds that up to whole words.

As written, callers may expect this accessor to report the effective rounded cap.

huitseeker (approved):
This LGTM!

processor/src/execution_options.rs
```diff
@@ -137,6 +150,7 @@ impl ExecutionOptions {
             max_hash_len_bytes: Self::DEFAULT_MAX_HASH_LEN_BYTES,
             max_num_continuations: Self::DEFAULT_MAX_NUM_CONTINUATIONS,
             max_stack_depth: Self::DEFAULT_MAX_STACK_DEPTH,
+            max_memory_elements: Self::DEFAULT_MAX_MEMORY_ELEMENTS,
```
bobbinth, line 153: Probably not for this PR, but I think we should collapse most of these into a single setting. I don't think the user particularly cares whether their RAM gets exhausted because of stack, advice map, advice stack, or memory components. So, having a single memory allocation limit for the entire processor would be better (unless it causes some complications in the implementation).

plafer: Created #3233

processor/src/execution_options.rs
plafer: Good catch, fixed

PR #3227, opened 2026-06-04
fix(debug): encode enum and u256 debug types

djolertrk:
In assembler, we did not support `U256` and `Enums` in `.debug_types`. This patch adds that.

This will need handling in debugger.

### DIFF
--- a/crates/assembly/src/assembler/debuginfo.rs
+++ b/crates/assembly/src/assembler/debuginfo.rs
@@ -9,7 +9,7 @@ use std::{
 use miden_assembly_syntax::debuginfo::{FileLineCol, Location, Uri};
 use miden_mast_package::debug_info::{
     DebugFieldInfo, DebugFunctionsSection, DebugPrimitiveType, DebugSourcesSection, DebugTypeIdx,
-    DebugTypeInfo, DebugTypesSection,
+    DebugTypeInfo, DebugTypesSection, DebugVariantInfo,
 };
 
 use crate::{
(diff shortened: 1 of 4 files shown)


djolertrk:
cc @bitwalker

crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -325,13 +328,30 @@ fn register_debug_type(
...
+                .iter()
+                .zip(enum_ty.discriminant_values())
+                .map(|(variant, discriminant)| {
+                    let name_idx = debug_types_section.add_string(variant.name.clone());
+                    let type_idx = variant
+                        .value
+                        .as_ref()
+                        .map(|ty| register_debug_type(debug_types_section, None, None, ty))
+                        .transpose()?;
+                    Ok(DebugVariantInfo { name_idx, type_idx, discriminant })
```
huitseeker: Should we also carry the variant payload offset here? `EnumType` computes per-variant offsets with `variant_offsets()`, and those offsets are not always the discriminant size. A `U8` discriminant with a `U64` payload reports `size=12` and payload `offset=4`, but this record only keeps the payload type and discriminant. A debugger would not know where to read the payload for aligned variants.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -181,6 +181,7 @@ const TYPE_TAG_ARRAY: u8 = 2;
 const TYPE_TAG_STRUCT: u8 = 3;
 const TYPE_TAG_FUNCTION: u8 = 4;
 const TYPE_TAG_UNKNOWN: u8 = 5;
+const TYPE_TAG_ENUM: u8 = 6;
```
huitseeker, line 184: Should we bump `DEBUG_TYPES_VERSION` with this new tag? This writes `Enum` as type tag `6` while the section still says version `1`. An older version-1 reader accepts the section version and then rejects the record as `invalid type tag: 6` instead of stopping at an unsupported version. The new `U256` primitive tag `16` has the same issue.

djolertrk: I was thinking about bumping it during the development, but this metadata is still effectively unused by any consumer: actually, the current consumer of the `.debug_types` is `objtool dump debug-info` only, and `debugger` and `miden-client` consumption is being added now (we are adding pretty-printing of the variables values with ABI types).

crates/assembly/src/assembler/debuginfo.rs
djolertrk: Yes, that’s a good catch! I am adding it.

crates/mast-package/src/debug_info/serialization.rs
huitseeker: Fair!

huitseeker (approved):
TYVM!

bitwalker (approved):
LGTM!

Issue #993, opened 2023-07-04
Change AdviceMap key in JSON inputs to use [u64; 4]


frisitano:
To define entries in the advice map using a JSON input file we currently use a hexadecimal string to represent the `Word` key.  The `Felt`s in the word are encoded as little endian u64 and then the hex outputs combined to produce the final hex string.  Instead we should migrate to specifying the keys as [u64; 4] in JSON such that the keys can be easily interpreted from the JSON.

bobbinth:
What is the difficulty of dealing with hex encoding? Is it because there is no straight-forward way to convert hex strings to/from words? If so, maybe a better solution would be provide utility functions to do something like `word_to_hex()` and `word_from_hex()`?

frisitano:
> What is the difficulty of dealing with hex encoding? Is it because there is no straight-forward way to convert hex strings to/from words? If so, maybe a better solution would be provide utility functions to do something like `word_to_hex()` and `word_from_hex()`?

Yes this is part of the problem.  If we want to provide the utility methods then they should be provided via the CLI. I think this will help quite a bit.

However, there is also a formatting difference between how keys are referenced in the JSON input file and assembly code.  If we could achieve parity between the two then it may yield superior UX. E.g.

`inputs.json`:
```json
    "advice_map": {
        "0.0.0.1": [5, 6, 7, 8],
        "0.0.0.2": [9, 10, 11, 12],
        "0.0.0.3": [13, 14, 15, 16]
    },
```

`code.masm`:
```
...
push.0.0.0.1 adv.push_mapval 
```

In my opinion it's easier to cross reference between masm and `inputs.json` making UX better.

bobbinth:
> However, there is also a formatting difference between how keys are referenced in the JSON input file and assembly code. If we could achieve parity between the two then it may yield superior UX.

Agreed that we should have parity between the two - but I think we can do this with hex strings as well. Specifically, [push](https://0xpolygonmiden.github.io/miden-vm/user_docs/assembly/io_operations.html#constant-inputs) instruction supports immediates formatted as hex strings. For example, the following should be equivalent:
```
push.0x1234.0xabcd
push.0x0000000000001234000000000000abcd
```
And I think using hex strings in both places is more intuitive as people usually think of hashes as strings.

frisitano:
[...]

PR #3228, opened 2026-06-04
refactor: bump `miden-crypto` to v0.26.0

### LINKED ISSUE #3225: Upgrade miden-lifted-stark to v0.26.0
The VM should upgrade `miden-crypto` and `miden-lifted-stark` to v0.26.0. 
Blocked on https://github.com/0xMiden/midenc-hir-type/pull/14 (and midenc-hir-type v0.7.0)

This is separate from the rand dependency update. `miden-lifted-stark` v0.26.0 removed several v0.25 APIs used by the VM.

Needed changes:

- Use `MultiAir`, `Statement`, and `ProverStatement` instead of the old `(AIR, AirWitness, AuxBuilder)` model.
- Move external input logic from `LiftedAir::num_var_len_public_inputs` and `LiftedAir::reduced_aux_values` to `MultiAir::eval_external`.
- Build aux traces with `LiftedAir::build_aux_trace` instead of `AuxBuilder`.
- Remove use of deleted trace-order types such as `AirInstance` and `InstanceShapes`.
- Import proof code from the new modules: `pcs` instead of `fri`, and `proof` instead of `transcript`.
- Use `StarkProofData` for the serialized proof, `StarkProof` for the parsed proof view, and `StarkProof::from_data` for parsing.
- Use the v0.26.0 `ProverInstance` and `VerifierInstance` APIs.
- Check the recursive verifier against the new proof and transcript layout.

This unblocks the rand upgrade in #2875 without mixing the lifted-STARK API migration into that PR.

Nashtare:
## Summary

Migrates the workspace from `miden-crypto` / `miden-lifted-stark` **0.25 → 0.26**. The change touches the AIR definitions, the prover/verifier crates, the processor's debug constraint checker, the `rand`-based crypto tests, and the hand-written MASM recursive STARK verifier.

Both the native Rust prove→verify path **and** the in-VM recursive verifier are fully migrated and passing.

closes #3225

## Dependency changes

| Crate | From | To | Why |
|---|---|---|---|
| `miden-crypto` | 0.25 | 0.26 | target bump |
| `miden-lifted-stark` | 0.25 | 0.26 | pulled by `miden-crypto`; must move together |
| `midenc-hir-type` | 0.6.1 | 0.7.0 | 0.6.1 pinned `miden-serde-utils 0.25`; the crypto bump pulls `0.26`, leaving two incompatible copies that broke `mast-package` serde. 0.7.0 uses `0.26`. |
| `rand` / `rand_chacha` (dev) | 0.9 | 0.10 | 0.26's AEAD uses `rand 0.10`'s `Rng`/`TryRng` traits |

### AIR (`crates/lib/core`, `air`)
0.26 replaced the per-AIR witness/aux-builder model with a `MultiAir` trait:
- Dropped the `AuxBuilder` impl; folded `build_aux_trace` into `LiftedAir` (now also receives `air_inputs`/`aux_inputs`).
- Removed `num_var_len_public_inputs` / `reduced_aux_values` / `log_quotient_degree` from `LiftedAir`. The constraint degree is now derived symbolically (`constraint_degree() -> ConstraintDegrees`); the old hand-set `log_quotient_degree = 3` override is gone (the symbolic value is identical).
- Added **`MidenMultiAir`** implementing `MultiAir`, whose `eval_external` carries the cross-AIR LogUp boundary reduction (the kernel-procedure digests now flow as `aux_inputs`), plus a custom `observe` that reproduces the existing public-input + kernel-digest Fiat-Shamir absorption so the MASM verifier's `process_public_inputs` is unchanged.

### Prover / Verifier
- Switched to `Statement` / `ProverStatement` / `ProverInstance` / `VerifierInstance`.
- Proof wire type `StarkProof` → `StarkProofData`; the structured parse view is now `StarkProof` (parsed via `StarkProof::from_data`).
- Removed the manual `validate_canonical_air_order` check — 0.26 derives the multi-AIR ordering deterministically from the per-AIR trace heights carried on the proof, and binds the heights into Fiat-Shamir, so a tampered ordering is rejected via transcript mismatch. The obsolete `rejects_non_canonical_air_order` test (which tampered with the now-removed `InstanceShapes` wire field) was removed.

### Processor
- The debug `ExecutionTrace::check_constraints` now builds a `ProverStatement` and uses the new `debug::check_constraints(prover_statement, challenger)` API.

### `rand` 0.10 (crypto tests)
- `rand`/`rand_chacha` dev-deps bumped to 0.10. API rename: the old `RngCore` (low-level) → `Rng`, and the old high-level `Rng` sampling methods (`random`, `random_range`) → the new `RngExt` trait. Seeded outputs are stable (72/72 crypto tests pass).

### MASM recursive verifier (`crates/lib/core/asm/stark`, `…/sys/vm`)
0.26 changed the Fiat-Shamir transcript layout: `Statement::observe` + the per-AIR trace-height observation are now framework-internal, and `log_trace_heights` replaces the `InstanceShapes`/`air_order` wire format.
- Replaced `observe_trace_height_and_discard_instance_challenge` with `observe_instance_shape` (buffers `[instance_count, log_core, log_chip]` without permuting) + a new `reseed_main_after_shape` (buffers the main-trace commitment onto those elements and permutes once — matching the DuplexChallenger's single duplex over the 7 buffered elements before the first post-commitment sample).
- The Rust bridge (`recursive_verifier.rs`) now parses via `StarkProof::from_data` and reads `pcs_proof` / `deep_proof` / `fri_proof`.

#### Root cause of the recursive verifier failure: LDE coset shift

After the transcript work, the recursive verifier still failed at the constraint/quotient identity (`circuit does not evaluate to zero`, and a FRI degree-respecting-projection inconsistency). Root cause, in `miden-lifted-stark` PR #993 (the `Coset` domain refactor): the LDE coset shift changed from the field generator `g = 7` to

```
lde_coset.shift() = F::GENERATOR.exp_power_of_2(F::TWO_ADICITY − log_lde) = 7^(2^(32 − log_lde))
```

The MASM verifier hardcoded shift = 7, which silently broke **two** places:
1. the quotient reconstruction node `s0 = shift^N` (→ wrong `Q(z)` → ACE circuit ≠ 0), and
2. the DEEP-query x-coordinates `x = offset · ω_lde^i` (→ FRI inconsistency).

Fix (MASM only):
- `utils.masm`: quotient `s0 = 7^(2^29)` (`= shift^N = g^(2^(TWO_ADICITY − log_blowup))`, trace-height-independent) via repeated squaring, replacing `7^N`.
- `constants.masm`: `DOMAIN_OFFSET` / `DOMAIN_OFFSET_INV` consts → memory pointers, with getters/setters.
- `random_coin.masm` (`init_seed`): compute the per-proof offset `7^(2^(32 − log_lde))` and its inverse by squaring `g`, `32 − log_lde` times after deriving `log(lde_size)`.

## Testing

| Suite | Result |
|---|---|
| `prove_verify` (native + recursive + precompiles + fast-parallel, all hash fns) | 12/12 ✅ |
| `core-lib` `stark::*` (recursive small/large/kernel-even/odd/single + VLPI) | 34/34 ✅ |
| `core-lib` crypto (rand 0.10) | 72/72 ✅ |
| `air` / `ace-codegen` / `verifier` / `processor` lib tests | 21 / 69 / 2 / 1516 ✅ |
| `relation_digest_matches_current_air` (AIR circuit unchanged) | ✅ |

The AIR constraint circuit / `RELATION_DIGEST` is unchanged by this migration — only the surrounding trait structure and the verifier's domain math changed.

## Notes

- The recursive-verifier root cause was diagnosed by temporarily vendoring `miden-lifted-stark` via `[patch.crates-io]` with instrumentation in `verifier::verify` / `reconstruct_quotient`, plus Rust cross-checks in `ace_read_check.rs`. All of that scaffolding has been reverted; the fix stands alone on the published 0.26 crate.
- `s0` uses `repeat.29` where `29 = TWO_ADICITY(32) − LOG_BLOWUP(3)`, consistent with the existing MASM convention of hardcoding the (fixed) blowup (e.g. `repeat.3`).

- Updated `CHANGELOG.md'


prover/src/lib.rs
```diff
@@ -191,38 +190,29 @@ where
...
-        (&chiplets_air, chiplets_witness, &chiplets_air),
-    ];
+    // `air_inputs` are the fixed public values; `aux_inputs` are the kernel-procedure
+    // digests (the only variable-length public input today). The lifted prover absorbs
+    // both into Fiat-Shamir internally, along with the per-AIR trace heights.
+    let statement =
+        Statement::new(MidenMultiAir::new(), public_values.to_vec(), kernel_felts.to_vec())
+            .map_err(|e| ExecutionError::ProvingError(e.to_string()))?;
+    let prover_statement =
+        ProverStatement::new(statement, vec![core_trace.clone(), chiplets_trace.clone()])
```
huitseeker: Could `prove_stark` take these matrices by value now? The caller just built them with `trace.into_core_chiplets_matrices()`, so this line copies both full trace buffers only to satisfy the new `ProverStatement::new(statement, vec![...])` ownership.

An owned helper or owned `prove_stark` parameters would let this move the matrices instead of cloning the prover's hot-path trace data.

huitseeker (commented):
This looks great overall, but for the dubious clone of trace buffers, see inline.

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -93,153 +93,159 @@ pub proc process_public_inputs
...
+    end
     drop
-    # => [...]
 
-    # 3) Load the auxiliary randomness (alpha and beta) from advice.
-    #    Stored as [beta0, beta1, alpha0, alpha1] since `horner_eval_ext` requires memory
-    #    word-alignment.
+    # MAX_AUX_INPUTS, then len(kernel_felts) = 4 * num_kernel_proc_digests.
+    push.MAX_AUX_INPUTS exec.random_coin::observe_felt
+    exec.constants::tmp3 mem_load mul.4 exec.random_coin::observe_felt
```
huitseeker, line 218: Could this assert the same max as the Rust statement before absorbing the length? 

`MidenMultiAir::max_aux_inputs()` is `Kernel::MAX_NUM_PROCEDURES * WORD_SIZE`, so Rust rejects more than 255 kernel digests. This MASM path still accepts much more: [the existing `#[case(1000)]` in `variable_length_public_inputs`](https://github.com/0xMiden/miden-vm/blob/328071990a6de3487c3189f08c7d8ad545c9d408/crates/lib/core/tests/stark/mod.rs#L195-L291) passes locally with `cargo test -p miden-core-lib variable_length_public_inputs --tests`.

Without the same bound here, the recursive verifier can absorb a statement length that the Rust prover/verifier refuses.

huitseeker (approved):
Looks awesome! One thing to fix in the auxiliary inputs bound. Please mark as "closes #3225" or manually close when you merge.

crates/lib/core/asm/sys/vm/public_inputs.masm
Nashtare: Good catch, I've fixed it and added an additional test.

air/src/constraints/lookup/miden_air.rs
```diff
@@ -26,7 +26,7 @@ pub const NUM_LOGUP_COMMITTED_FINALS: usize = 2;
 /// - `LogPrecompileTranscript` lives on `MAIN_COLUMN_SHAPE[0]` (block_stack + range + log-cap
 ///   merged column).
 ///
-/// Both fractions therefore belong to `CoreAir::reduced_aux_values` post-split.
+/// Both fractions therefore belong to `CoreAir::boundary_correction` post-split.
```
Al-Kindi-0: "post-split" is a history/process comment and not a product comment. Should probably be removed

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -6,8 +6,16 @@
...
+#! rate[output_len - 1].
+#!
+#! Two absorption paths are provided:
+#!   - `observe_felt`/`flush_buffer`: a faithful buffered duplex matching p3's `DuplexChallenger`
+#!     (append to `rate[input_len]`, permute only once the rate fills, flush a partial buffer
+#!     lazily). Used for the public inputs, instance shape, and main-trace commitment, whose
+#!     element stream is not rate-aligned.
+#!   - `reseed_direct`/`reseed_with_felt`/etc.: rate-aligned absorptions that write directly to
+#!     rate memory and permute immediately. Valid only when the input buffer is empty (the stream
+#!     is known to start and end on a rate boundary).
```
Al-Kindi-0, line 18: This is fine for now, but mimicking the Rust implementation adds a substantial overhead

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -136,6 +144,62 @@ pub proc sample_bits
...
+#!
+#! Mirrors p3's `DuplexChallenger::observe`: the element is appended to the rate input buffer at
+#! `rate[input_len]`, the output buffer is invalidated, and the sponge is permuted only once the
+#! buffer fills (`input_len == 8`). A partial buffer is flushed lazily by `flush_buffer` (or by the
+#! next permutation), so the permutation boundaries follow the absorbed-element count, not any
+#! per-section alignment.
+#!
+#! Input:  [felt, ...]
+#! Output: [...]
+pub proc observe_felt
```
Al-Kindi-0, line 157: same comment here and elsewhere

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -243,6 +307,25 @@ pub proc init_seed
...
+    dup push.32 swap sub
+    #=> [32 - log_lde, log(lde_size), trace_length, ...]
+    push.7 swap
+    #=> [count, offset = 7, log(lde_size), trace_length, ...]
+    dup push.0 neq
+    while.true
+        swap dup mul swap
+        sub.1
+        dup push.0 neq
+    end
```
Al-Kindi-0, line 322: nit: could probably be a bit faster by putting the count deeper

crates/lib/core/asm/stark/verifier.masm
```diff
@@ -125,7 +123,7 @@ pub proc verify
     exec.utils::swap_air_regions_if_chip_first
 
     #==============================================================================================
-    #       VI) Draw gamma for reduced_aux_values batching and evaluate constraints
+    #       VI) Draw gamma for the LogUp boundary (eval_external) batching and evaluate constraints
```
Al-Kindi-0: nit: remove (eval_external)

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -93,153 +93,165 @@ pub proc process_public_inputs
...
+    exec.constants::variable_length_public_inputs_address_ptr mem_load
+    exec.constants::tmp3 mem_load
+    swap
+    # => [var_len_ptr, num_iter, ...]
+    dup.1 push.0 neq
+    while.true
+        dup add.7 mem_load exec.random_coin::observe_felt
+        dup add.6 mem_load exec.random_coin::observe_felt
+        dup add.5 mem_load exec.random_coin::observe_felt
+        dup add.4 mem_load exec.random_coin::observe_felt
```
Al-Kindi-0, line 230: same remark regarding efficiency of the random coin

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -280,86 +363,37 @@ pub proc init_seed
...
+#! duplex at the first post-commitment sample (auxiliary randomness). On exit: output_len=8.
+#!
+#! Input: [c0, c1, c2, c3, ...]
+#! Output: [...]
+pub proc reseed_main_after_shape
+    exec.observe_felt
+    exec.observe_felt
+    exec.observe_felt
+    exec.observe_felt
+    exec.flush_buffer
```
Al-Kindi-0, line 396: same remark

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -280,86 +363,37 @@ pub proc init_seed
...
-    # => [...]
-
-    exec.load_random_coin_state
-    # => [R1, R2, C, ...]
-    exec.poseidon2::permute
-    # => [R1', R2', C', ...]
-    exec.store_random_coin_state
-    # => [...]
+pub proc observe_instance_shape
+    push.2 exec.observe_felt
```
Al-Kindi-0, line 378: I think this is redundant by virtue of the next two lines, but it doesn't hurt

Al-Kindi-0 (commented):
Did an initial pass and things look good.
Left a few comments, not all need to be addressed in this PR

crates/lib/core/asm/stark/random_coin.masm
Nashtare: My feel with crypto-v0.26 was that we'd want to align both rather than having this custom `observe()` logic here. I could revert it to how it was before though. Do you also prefer that @adr1anh?

Al-Kindi-0: What I mean was that the MASM implementation could avoid some overhead, like it always did, without changing the Rust implementation.
We can just keep as is and then fix latter as part of a cleanup/optimization pass over the recursive verifier

Al-Kindi-0 (approved):
LGTM
We can merge as is to avoid blocking other work (cc @huitseeker )

huitseeker:
@Al-Kindi-0 thanks, I opened #3239 for follow-up

PR #3230, opened 2026-06-05
Clean up processor error handling

### LINKED ISSUE #2575: Minor error handling cleanups
Small follow-ups from #2544.

- [x] `assert_u32` should return `u32` directly instead of `Felt` (`chiplets/bitwise/mod.rs`) - [comment](https://github.com/0xMiden/miden-vm/pull/2544#discussion_r2671949499)
- [ ] Review `#[diagnostic(help(...))]` usage - most could fold into the error message itself - [comment](https://github.com/0xMiden/miden-vm/pull/2544#issuecomment-3724332474)
- [ ] Consider dedicated error type for malformed MAST in `load_mast_forest` (`fast/mod.rs`) - or defer to lazy AdviceMap loading - [comment](https://github.com/0xMiden/miden-vm/pull/2544#discussion_r2671979023)
- [ ] Consider consolidating `NotBinaryValue` / `NotBinaryValueIf` / `NotBinaryValueLoop` (tradeoff: explicit messages vs miette diagnostics redundancy) - [comment](https://github.com/0xMiden/miden-vm/pull/2544#issuecomment-3724332474)

huitseeker:
This cleans up the remaining error handling follow-ups from #2575.

- The diagnostic help item is fixed by moving the useful help text into the main error messages.

- The malformed MAST item is fixed by giving the `load_mast_forest` error path a named helper and adding diagnostic coverage for the missing-root case.

- The binary-value item is fixed by replacing `NotBinaryValue`, `NotBinaryValueIf`, and `NotBinaryValueLoop` with one `NotBinaryValue` error that carries its context.

Closes #2575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - [BREAKING] `ProjectAssembler::assemble_with_sources` has been removed - projects require assembly from the filesystem going forward ([#3216](https://github.com/0xMiden/miden-vm/pull/3216))
 - [BREAKING] `miden-vm bundle` now treats the `--kernel` option as a flag; when set, it expects the file path given to `bundle` to be the path to the root module of the kernel, and the support library for the kernel is derived from explicit submodule declarations in that module.
 - [BREAKING] Moved debug info ownership out of `MastForest` and into package debug sections, adding source-node debug metadata that preserves distinct source occurrences after MAST node deduplication ([#3221](https://github.com/0xMiden/miden-vm/pull/3221)).
+- Cleaned up processor error handling for diagnostics, malformed MAST loading, and binary-value checks ([#3230](https://github.com/0xMiden/miden-vm/pull/3230)).
 
 #### Fixes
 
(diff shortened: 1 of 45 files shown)


bitwalker:
> The diagnostic help item is fixed by moving the useful help text into the main error messages.

Could you elaborate on this more? It looks to me like it just moves the text around, rather than fixing the output - personally I find it a lot more readable to have them split up into error message and separate tips/help/suggestion text parts. The thing you generally have to be careful about is putting important information in _label_ text, because those won't render without source spans, so they should only provide useful contextual information that can be dropped if there aren't spans - but I don't think that applies in this case, since this just moved the `help` text, which is always emitted.

plafer (approved):
LGTM

huitseeker:
@bitwalker the idea from https://github.com/0xMiden/miden-vm/pull/2544#issuecomment-3724332474 is that they're marginally not a significant benefit for the occurrences changed here, and removing them keeps the error self-contained, which helps boxing the inner error.

adr1anh (approved):
Nice, TYVM!

PR #3232, opened 2026-06-05
feat(assembler): add do-while loop syntax

### LINKED ISSUE #1684: Investigate switching from `while` loops to `do-while` loops

plafer:
Closes #1684 

Adds the `do <block> while <cond> end` syntax that maps nicely to the recent tail-controlled `LoopNode`.

The part I'm less sure about is and to look out for is the changes to CST parser; the main challenge was distinguishing the `while` in `while.true` and the new `do ... while ... end` (called "bare while" in the code).


crates/assembly-syntax-cst/src/parser.rs
```diff
@@ -1113,7 +1156,27 @@ impl<'input> Parser<'input> {
...
+            }
+        })
+    }
+
+    /// Returns `true` if the current token is a `while` keyword that is *not* followed by a `.`
+    /// suffix (i.e. the `while` that closes a `do`..`while` body, not a nested `while.true`).
+    fn at_bare_while(&self) -> bool {
+        self.at_keyword("while")
+            && !matches!(
+                self.next_relevant_top_level_token(self.pos + 1)
```
bitwalker: I think this should just examine literally the next token - if there is anything in between the `while` and the `.`, treat it like the "bare" `while` was found - that will make something like `while   .true` a syntax error at the `.true`. This will permit such spellings to be matched as a `while.true` opcode, and I think we should probably be stricter about when we match `while.true` in this setting.

bitwalker (changes requested):
Looks good! Just one change I think we should make, but otherwise I think you covered all the necessary bases on the parser/formatter end

crates/assembly-syntax-cst/src/parser.rs
plafer: Done

bitwalker (approved):

huitseeker (approved):
This looks great!

Al-Kindi-0 (approved):
Very nice!

PR #3236, opened 2026-06-06
Fail closed on missing CLI input files

huitseeker:
`run` and `prove` treated a missing inferred `.inputs` file as empty public inputs. This could create a valid run or proof for the wrong public statement.

This changes input loading so the inferred `.inputs` file must exist, just like an explicit `--input` file. Existing no-input CLI fixtures now include an explicit empty `.inputs` file.

Fixes 0xmiden/security-issues#24


adr1anh (approved):
Looks good

Al-Kindi-0 (approved):
LGTM

PR #3237, opened 2026-06-06
fix: Reject empty FRI query regions

huitseeker:
`frie2f4::verify` accepts an empty query region when the query pointer equals the first FRI layer pointer. In that case, the verifier can skip all query checks.

This rejects empty query regions before verifier dispatch by requiring `query_ptr < fri_layer_ptr`.

Closes https://github.com/0xMiden/security-issues/issues/27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 #### Fixes
 
+- Rejected empty query regions in the standalone FRI verifier ([#3237](https://github.com/0xMiden/miden-vm/pull/3237)).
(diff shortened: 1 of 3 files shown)


Nashtare (approved):
LGTM

Al-Kindi-0 (approved):
Though we do check that the number of FRI queries is above a certain threshold, it doesn't hurt to add the content of the current PR.

LGTM

huitseeker:
@Al-Kindi-0 to clarify : `stark::verify`/`sys::vm::verify_proof` do check query count policy, while direct `frie2f4::verify` only enforces non-empty query data. Happy to shift to documenting if that's what we want to do (though `frie2f4::verify` remains public)

PR #3240, opened 2026-06-09
fix: Pin initial system context in the AIR

huitseeker:
the AIR only pinned the first-row clock. The first-row context and function hash were only transition-constrained, so a prover could start from a forged caller hash.

This pins row 0 to the root context and an empty function hash.

The recursive verifier artifacts and relation digest are updated for the new AIR.

FIxes 0xmiden/security-issues#20


Al-Kindi-0 (approved):
LGTM, tyvm!

plafer (approved):
LGTM!

PR #3241, opened 2026-06-09
chore: gate proptest rand dependency

### LINKED ISSUE #2875: Update `rand` to `v0.10`
Currently we use two different versions of the `rand` crate: `0.9` for the VM itself and `0.10` for the `miden-field`, which is part of the `miden-crypto`.

This crates some collisions, which for now are just skipped (see [deny.toml file](https://github.com/0xMiden/miden-vm/blob/88c98f49537f89a737ef95ac5a989d2b4bd42c4b/deny.toml#L54)).

We should update the VM's `rand` (and, hence, `rand-core`) dependency to consistently use `0.10` version.

huitseeker:
This fixes #2875 by keeping proptest, and its rand 0.9 dependency, behind arbitrary/testing. Most of the heavy lifting done prior to that in #3228 by @Nashtare 

Deterministic tests still run without arbitrary. Targeted test presets for VM and core-lib now enable testing, so their property and diagnostic tests still run.


plafer (approved):
LGTM

PR #3243, opened 2026-06-10
perf(verifier): absorb public inputs in blocks in the recursive verifier

Nashtare:
## Describe your changes

Absorb the PI stream in 8-element blocks instead of per-felt buffered observes. Yields ~33–48% fewer cycles on the absorb , ~2-4% fewer cycles on full recursive verify.

- Updated `CHANGELOG.md'


crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -190,48 +192,101 @@ proc load_kernel_digests
...
+        dup.2 add.7 mem_load dup.2       mem_store
+        dup.2 add.6 mem_load dup.2 add.1 mem_store
+        dup.2 add.5 mem_load dup.2 add.2 mem_store
+        dup.2 add.4 mem_load dup.2 add.3 mem_store
+        swap add.4 swap
+        movup.2 add.8 movdn.2
+        sub.1
+        dup push.0 neq
+    end
+    drop drop drop
```
Al-Kindi-0: Not sure I follow what is going on here, but this seems to be inefficient

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -190,48 +192,101 @@ proc load_kernel_digests
...
+        dropw dropw
+        # => [C, ptr]
+        dup.4 add.4 padw movup.4 mem_loadw_le
+        # => [R2, C, ptr]
+        dup.8 padw movup.4 mem_loadw_le
+        # => [R1, R2, C, ptr]
+        exec.poseidon2::permute
+        movup.12 add.8 movdn.12
+        exec.constants::tmp4 mem_load sub.1 dup exec.constants::tmp4 mem_store push.0 neq
+    end
```
Al-Kindi-0, line 265: Same remark here

Al-Kindi-0 (commented):
Looks good
Still opportunities for hand tuning which will be done in the future

crates/lib/core/asm/sys/vm/public_inputs.masm
Nashtare: Actually the loop is not needed anymore, let me clean that up

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -167,71 +169,118 @@ proc load_kernel_digests
...
+        dropw
+        # => [d3, d2, d1, d0, Y, ptr+8, dst, remaining, ...]
+        # Write the digest in natural order to the absorb scratch: d_i at dst + i. Element
+        # stores, since the scratch stream offset (43 + 4d) is not word-aligned.
+        movup.9 dup add.4 movdn.10
+        # => [dst, d3, d2, d1, d0, Y, ptr+8, dst+4, remaining, ...]
+        swap dup.1 add.3 mem_store
+        swap dup.1 add.2 mem_store
+        swap dup.1 add.1 mem_store
+        mem_store
```
Al-Kindi-0, line 199: we could probably just use `reversew` and `mem_storew` here

Al-Kindi-0 (approved):
LGTM

crates/lib/core/asm/sys/vm/public_inputs.masm
Nashtare: Wouldn't we have issues with the misalignment here though?

crates/lib/core/asm/stark/random_coin.masm
```diff
@@ -146,26 +146,25 @@ end
...
 #! Absorb a single base field element into the transcript via the buffered duplex.
 #!
-#! Mirrors p3's `DuplexChallenger::observe`: the element is appended to the rate input buffer at
-#! `rate[input_len]`, the output buffer is invalidated, and the sponge is permuted only once the
-#! buffer fills (`input_len == 8`). A partial buffer is flushed lazily by `flush_buffer` (or by the
-#! next permutation), so the permutation boundaries follow the absorbed-element count, not any
-#! per-section alignment.
+#! Mirrors p3's `DuplexChallenger::observe`: appends the element at `rate[input_len]` and permutes
+#! only once the buffer fills (`input_len == 8`); a partial buffer is flushed lazily by
+#! `flush_buffer`. `output_len` is not invalidated per element, so callers must `flush_buffer`
```
huitseeker: Nit: Could we more clearly keep `observe_felt` from looking like the safe p3 `observe` in docs here, or keep clearing `output_len`?

p3 clears its output buffer on every observe, and `sample` duplexes when input is pending. With this version, a caller that samples, then observes one felt, then samples again without an explicit `flush_buffer` can read stale rate output because `output_len` is still nonzero. 

The current verifier path flushes before sampling, but this proc is public and still reads like the general observe primitive.

huitseeker (approved):
LGTM, though I'd like to make the `observe_felt` footgun a bit more clear

crates/lib/core/asm/sys/vm/public_inputs.masm
Al-Kindi-0: And why the nonalignment ? It is also fine to keep as is and we can come back to it later on

Nashtare: Because the Rust side removed the padding with miden-crypto v0.26

PR #3244, opened 2026-06-10
fix: Constrain LOGPRECOMPILE tail stack slots

huitseeker:
LOGPRECOMPILE replaces the top 12 stack slots with [R0, R1, CAP_NEXT] and leaves stack[12..15] unchanged. The AIR only bound the hasher inputs and outputs, so those four tail slots could change across the row.

  This adds preservation constraints for stack[12..15] during LOGPRECOMPILE, updates the AIR digest and recursive verifier artifacts, and adds a focused constraint test for forged tail slots.
  
  Closes 0xmiden/security-issues#21


adr1anh:
Nice, thanks for fixing this.

FYI I pushed an alternate version here:

https://github.com/0xMiden/miden-vm/compare/next...adr1anh/fix_log_precompile_constraints

Relevant bit:

https://github.com/0xMiden/miden-vm/blob/adr1anh/fix_log_precompile_constraints/air/src/constraints/op_flags/mod.rs#L387-L395

Your fix adds the `s_next[12..16] = s[12..16]` constraints directly under `log_precompile`, which is correct.

My version gets the same constraints via the existing generic stack-transition machinery: `LOGPRECOMPILE` is classified like `HPERM` at depth 12, since the hasher lookup already constrains `stack_next[0..12]`, and only `s12..s15` need normal no-shift preservation.

So adding `LOGPRECOMPILE` to `no_shift_depth12` makes `no_shift_at(12..16) = 1`, and `stack/general.rs` reduces to the same equalities:

```text
log_precompile * (s_next[i] - s[i]) = 0, i = 12..15
```

Test is here:

https://github.com/0xMiden/miden-vm/blob/adr1anh/fix_log_precompile_constraints/air/src/constraints/op_flags/tests.rs#L417-L453

I think this is slightly cleaner since it fixes the stack-impact classification rather than adding a one-off crypto constraint.

Al-Kindi-0:
Basically as @adr1anh mentioned, we should just make sure that the op flag is part of the combined no-shift stack flag at depth 12

Al-Kindi-0 (approved):
LGTM

PR #3245, opened 2026-06-10
fix: Bind memory word address limbs

huitseeker:
The memory AIR range-checks w0 and w1, but did not bind them to word_addr. A forged trace could use a non-
  canonical word_addr while the checked limbs stayed valid.

  This adds the missing relation:

  word_addr = 4 * (w0 + 2^16 * w1)

  It also updates the recursive verifier artifacts and adds a focused AIR test for malformed address limbs.

fixes 0xmiden/security-issues#33


Al-Kindi-0 (approved):
LGTM

Nashtare (approved):
LGTM thanks!

PR #3247, opened 2026-06-11
fix: Constrain EVALCIRCUIT stack slots

huitseeker:
EVALCIRCUIT reads stack[0..2] as ACE init inputs and leaves the visible stack unchanged. The AIR only bound the ACE init lookup, so stack slots could change across the row.

This classifies EVALCIRCUIT as a no-shift stack operation, updates the AIR digest and recursive verifier artifacts, and adds a focused op-flag test.

FIxes 0xmiden/security-issues#22


adr1anh (approved):
LGTM once rebased

air/src/constraints/op_flags/tests.rs
```diff
@@ -398,6 +398,33 @@ fn composite_swap_flags() {
     assert_eq!(op_flags.left_shift(), ZERO);
 }
 
+/// Tests composite flags for EVALCIRCUIT (ACE init reads positions 0..3 without changing the
+/// visible stack).
+#[test]
+fn composite_evalcircuit_flags() {
```
Nashtare, line 404: This only checks flag values, perhaps it'd be worth adding a regression test encoding the exploit too?

Nashtare: actually can't this test be folded into `composite_no_shift_flags`? Just realized they're pretty much identical

Nashtare (approved):
LGTM, minor comments inline.

This makes me think though that we may want to strengthen this area, as we've fixed 2 similar issues recently. In other VM projects I worked on we always had some exhaustive, table-driven test over all possible valid opcodes x positions asserting each combination so that any change / new opcode would be triggered immediately. Possibly done as a follow-up though, non-blocking for this

Al-Kindi-0:
@huitseeker , I have fixed some other missing constraints and subsumed this PR while doing so in https://github.com/0xMiden/miden-vm/pull/3249 
We can merge this and then I can rebase, or we can just go for #3249 , both options are fine by me

huitseeker:
Closing in favor of #3249

PR #3248, opened 2026-06-11
fix: Bound FRI preprocess advice counts

huitseeker:
`frie2f4::preprocess` trusts advice-provided counts before writing FRI data into fixed workspace regions.

This adds bounds for query, layer, and remainder counts before the writes happen. It also keeps zero FRI layers valid, since small proofs can skip  folding layers.

Fixes https://github.com/0xMiden/security-issues/issues/34


Al-Kindi-0:
I really think we should remove the `preprocess` procedure and merge its function as part of a test or couple of tests. This procedure's only purpose is for testing and hence it should live near tests. This will imply then that questions regarding soundness for `preprocess` will become irrelevant.

huitseeker:
@al-kindi-0 Good idea. The test-only FRI preprocess helper is no longer exported by corelib; so tests now define it locally when they need to prepare verifier workspace

crates/lib/core/tests/stark_asserts/mod.rs
```diff
@@ -142,3 +142,219 @@ fn check_pow_invalid_has_message() {
     let test = build_test!(source, &[], advice_stack);
     expect_assert_error_message!(test);
 }
+
+const FRI_PREPROCESS_SOURCE: &str = "
```
Al-Kindi-0: Do we still need this and the tests using it in the current module ? 
I think it should be used only in the `fri/mod.rs` 

Basically once `preprocess` is a test helper we don't need the tests added in this module

Al-Kindi-0 (approved):
LGTM
Left a comment that should reduce the diff of this PR

adr1anh (approved):
LGTM

PR #3252, opened 2026-06-12
fix: Bound AEAD decrypt advice output

huitseeker:
AEAD decrypt reads and decrypts the full ciphertext before the advice stack size limit rejects the plaintext output.
This rejects decrypt requests whose plaintext would exceed the advice stack limit before reading ciphertext or running host-side decryption. The limit now uses the  processor’s advice stack cap directly.
Fixes 0xMiden/security-issues#29.


Al-Kindi-0 (approved):
LGTM

Nashtare (approved):

PR #3253, opened 2026-06-15
fix(assembly): preserve semantic type expressions

djolertrk:


djolertrk:
cc @bitwalker

djolertrk:
This is addressing problem from: https://github.com/0xMiden/compiler/pull/1151#discussion_r3400025658

crates/assembly-syntax/src/ast/type.rs
```diff
@@ -382,26 +382,34 @@ impl From<Type> for TypeExpr {
...
                         span: SourceSpan::UNKNOWN,
                         name,
                         ty: ft.ty.clone().into(),
                     }
-                }),
-            )),
+                });
+                Self::Struct(
+                    StructType::new(name, fields)
+                        .with_repr(Span::unknown(t.repr()))
```
huitseeker, line 402: Could we add a render round trip for this case? Once this keeps the non-default repr, `StructType::render` prints it before the keyword, like `@bigendian struct { ... }`, but the parser only accepts it after `struct`, like `struct @bigendian { ... }`. A concrete big-endian/aligned/packed/transparent struct can now convert back to a `TypeExpr` that renders as assembly syntax we cannot parse. Moving the repr after `struct` in the renderer, or accepting the prefix form in the parser, would close that loop.

crates/assembly-syntax/src/ast/type.rs
```diff
@@ -1263,4 +1271,38 @@ mod tests {
...
+            let expr = TypeExpr::from(ty.clone());
+            let TypeExpr::Primitive(actual) = expr else {
+                panic!("expected primitive type expression for {ty}, got {expr:?}");
+            };
+            assert_eq!(actual.into_inner(), ty);
+        }
+    }
+
+    #[test]
+    fn type_expr_from_type_preserves_struct_metadata() {
```
huitseeker, line 1353: Could we make this cover a syntax round trip too? This starts from a concrete `Type::Struct` whose fields already have names, but the assembly path still drops those names before this conversion. In `TypeExpr::Struct::resolve_type`, we push only `field_ty` into `StructType::from_parts`, so a parsed `struct { prefix: felt, suffix: felt }` becomes fields with no names and converts back as `field0`/`field1`. Passing `(field.name.clone().into_inner(), field_ty)` there would make this preservation work for types that came from assembly syntax too.

crates/assembly-syntax/src/ast/type.rs
djolertrk: Yes, thanks!

crates/assembly-syntax/src/ast/type.rs
djolertrk: Sure.

crates/assembly-syntax/src/parser/cst/fragments.rs
```diff
@@ -805,17 +806,20 @@ impl<'a, 'b> FragmentParser<'a, 'b> {
...
 
         let rparen =
             self.expect_kind(SyntaxKind::RParen, "expected `)` to close `align(...)` annotation")?;
-        Ok(Span::new(
-            join_spans(self.token_span(&at), self.token_span(&rparen)),
-            TypeRepr::align(value),
-        ))
+        let repr = match name_text {
+            "align" => TypeRepr::align(value),
+            "packed" => TypeRepr::packed(value),
```
huitseeker, line 819: Could we keep `@packed(N)` on the same power-of-two rule as the type repr docs? This path accepts `@packed(3)` today because `parse_positive_alignment` only checks `1..=65535`, but `TypeRepr::Packed` says the alignment must be a power of two.

djolertrk: hmm yes. I am adding tests for it as well.

huitseeker (approved):
Works for me!

bitwalker:
@djolertrk Can you rebase this?

bitwalker (approved):
This can be merged once rebased

PR #3254, opened 2026-06-15
fix: Bind EdDSA verification to message

huitseeker:
`eddsa_ed25519::verify_prehash` let callers supply the EdDSA challenge digest directly. That bypassed the normal message binding step.

This removes the public prehash verifier (where prehash means this specific API, not that of RFC 8032's Ed25519ph). EdDSA verification now passes the public key, message, and signature to the precompile, and the host verifies the signature against the message.

I checked this is not used downstream. We can always re-open another API once the precompile VM stabilizes.

Fixes https://github.com/0xMiden/security-issues/issues/23


adr1anh (approved):
The API looks a lot better this way, thanks! I'll note that this precompile is getting re-written with the new infrastructure, and will adapt the new version to match this safer interface. This is a stop gap until we have the Air constraints for Edwards curves and SHA512

Nashtare (approved):

PR #3255, opened 2026-06-15
fix: Reject duplicate SMT multi-leaf keys

huitseeker:
SMT multi-leaf preimages could authenticate more than one value for the same key. `smt::get` returned the first match, and `smt::set`could delete only one copy.

This rejects multi-leaf preimages whose keys are not strictly sorted and unique before lookup or update logic runs.

Closes 0xMiden/security-issues#31.


crates/lib/core/asm/collections/smt.masm
```diff
@@ -288,8 +288,14 @@ proc get_multi_leaf_value
...
 
+    locaddr.0
+    # => [start_ptr, end_ptr, K, R]
+
+    exec.assert_unique_multi_leaf_keys
+    # => [start_ptr, end_ptr, K, R]
+
     # (4 cycles)
-    movdn.4 locaddr.0 movdn.4
+    drop movdn.4 locaddr.0 movdn.4
```
plafer, line 305: nit: comment should now be `(5 cycles)`

plafer (approved):
LGTM

Al-Kindi-0:
I think there are two separate points worth discussing here.

First, what is the intended contract of `smt::get` and `smt::set`? If these procedures may be called with a root/advice pair supplied from outside the canonical SMT APIs, then I agree that the multi-leaf preimage needs to be validated.

If, on the other hand, these procedures are intended to operate only on roots produced by canonical SMT construction/update logic, then the extra check is redundant and we should document that precondition clearly.

My thinking and mental model has always been with second understanding but this is a good opportunity to double check that.

Second, if we do need to support untrusted roots/preimages, I wonder if we can combine validation with lookup/update. The current approach first scans the full multi-leaf to assert strict ordering/uniqueness, then calls `sorted_array::find_key_value` to find the key or insertion position. Since validation already requires looking at every key, we could instead use one SMT-specific linear scan that:

- asserts that adjacent keys are strictly increasing;
- records whether the target key was found;
- records the matching pointer or insertion position;
- continues to the end so that the entire leaf is validated.

Then `smt::get` can use the result to load the value or return `ZERO`, and `smt::set` can use the same result to replace, remove, or insert. This would preserve the security property while avoiding a separate validation pass followed by another lookup helper.

crates/lib/core/asm/collections/smt.masm
```diff
@@ -666,6 +706,15 @@ proc set_multi_leaf
     push.LEAF_DOMAIN exec.mem::pipe_double_words_preimage_to_memory_with_domain
     # => [end_ptr; NV, K, V, leaf_size, R]
 
+    locaddr.0
+    # => [start_ptr, end_ptr; NV, K, V, leaf_size, R]
+
+    exec.assert_unique_multi_leaf_keys
+    # => [start_ptr, end_ptr; NV, K, V, leaf_size, R]
```
bobbinth: This is a very expensive check. My guess is that for any leaf with $n$ key-value pairs, it'll add ~150n cycles.

bobbinth (approved):
Looks good! Thank you! The changes itself here are correct, but I'm with @Al-Kindi-0 here somewhat: we are trying to protect against a case where there is a critical bug somewhere else which allowed the SMT to be mutated incorrectly in the first place.

While this is a good "defense-in-depth" approach, it does come with non-negligible cost. Specifically, for any multi-value leaf, reading the values becomes much more expensive to the point where having 1K values in a leaf (which I believe is the current hard limit), will make a transaction potentially prohibitive.

So, overall, while this introduces a good extra check, it does open up a potential griefing attack vector where trying to get a few dozen entries into the same leaf makes reading this leaf significantly more expensive then before.

Now, in the protocol, we always randomize the keys - and so, getting even a few entries to end up in the same leaf is going to be pretty difficult - so, maybe that's OK.

crates/lib/core/asm/collections/smt.masm
huitseeker: If we check untrusted multi-leaf preimages, we have to scan the whole leaf, even when the key is found early. Otherwise, a duplicate key later in the same leaf could pass.

The last commit only removes the extra lookup pass. It does not remove the `O(leaf size)` check. Given randomized protocol keys, the worst case should be hard to force, but the cost is still part of the tradeoff.

huitseeker:
@Al-Kindi-0 ok, I agree with that contract split. I changed the lookup to use one SMT scan, which checks strict key order, records the match or insertion point, and keeps scanning to the end. That removes the separate validation pass before lookup.

crates/lib/core/asm/collections/smt.masm
```diff
@@ -666,14 +751,17 @@ proc set_multi_leaf
     push.LEAF_DOMAIN exec.mem::pipe_double_words_preimage_to_memory_with_domain
     # => [end_ptr; NV, K, V, leaf_size, R]
 
+    locaddr.0 drop
+    # => [end_ptr; NV, K, V, leaf_size, R]
```
bobbinth: This is a no-op, right (we push something onto the stack and then drop it immediately)? Why is this needed?

crates/lib/core/asm/collections/smt.masm
```diff
@@ -327,6 +325,93 @@ proc get_multi_leaf_value
...
 
+#! Finds a key in a multi-leaf preimage and validates that all keys are unique and sorted.
+#!
+#! Inputs:  [K, start_ptr, end_ptr, ...]
+#! Outputs: [is_key_found, key_ptr, start_ptr, end_ptr, ...]
+#!
+#! Panics if:
+#! - any adjacent key pair is not strictly increasing.
+@locals(10)
+proc find_unique_multi_leaf_key
```
bobbinth: It is not really clear to me if this procedure is more efficient than the previous approach. But I think it can be optimized quite a bit to remove extra `word::lt` calls and the number of conditional statements.

crates/lib/core/asm/collections/smt.masm
```diff
@@ -327,6 +325,93 @@ proc get_multi_leaf_value
...
+            loc_load.6 sub.8 padw movup.4 mem_loadw_le
+            # => [PREV_KEY, start_ptr, end_ptr, ...]
+
+            loc_load.6 padw movup.4 mem_loadw_le
+            # => [NEXT_KEY, PREV_KEY, start_ptr, end_ptr, ...]
+
+            exec.word::lt
+            assert.err="invalid multi-leaf preimage: keys must be unique and sorted"
+            # => [start_ptr, end_ptr, ...]
+        end
```
bobbinth: It would be great to add comments throughout this procedure to explain what we are doing and why. For example, the intent of procedure locals is not clear (seems like some of them are duplicated?)

crates/lib/core/asm/collections/smt.masm
huitseeker: leftover stack setup from the earlier validation call. 🤦

crates/lib/core/asm/collections/smt.masm
```diff
@@ -327,6 +325,93 @@ proc get_multi_leaf_value
...
+        # => [K, LEAF_KEY, start_ptr, end_ptr, ...]
+
+        exec.word::eq
+        # => [is_key_found, start_ptr, end_ptr, ...]
+
+        if.true
+            loc_load.6 loc_store.7
+            push.1 loc_store.8
+        end
+        # => [start_ptr, end_ptr, ...]
```
bobbinth: Why not bring the `is_key_found` check to the top? Basically, we can read the `is_key_found` from memory, and if the key has already been found, no need to do `exec.word::eq` or really anything else. The logic could look like:

```
while.true
    # do the ordering checks

   # read the is_key_found from memory
   # => [is_key_found, ...]
 
   if.false
        # do the exec.word::eq comparison
        # => [key_found, ...]

        if.true
            # save 1 to memory as is_key_found
            # save current key reference to memory
        end
   end

end
```

There are ways to optimize it even more by keeping more variables on the stack, but that would require deeper analysis.

crates/lib/core/asm/collections/smt.masm
```diff
@@ -327,6 +332,102 @@ proc get_multi_leaf_value
...
+                dupw padw loc_loadw_le.0 swapw
+                # => [LEAF_KEY, K, LEAF_KEY, start_ptr, end_ptr, ...]
+
+                exec.word::lt
+                # => [K < LEAF_KEY, LEAF_KEY, start_ptr, end_ptr, ...]
+
+                if.true
+                    loc_load.5 loc_store.6
+                    push.1 loc_store.8
+                end
```
bobbinth, line 498: This branch is needed because we need to keep track of the potential insert position for the `set` procedure, right?

I would add a comment explaining this. But also, this introduces non-negligible overhead, which is "dead weight" for the `get` procedure - right? Maybe we should consider having two versions of this procedure - one for `get` and the other one for `set`?

crates/lib/core/asm/collections/smt.masm
huitseeker: Done. The get helper stops doing equality checks once the key is found, but keeps scanning to validate the rest of the leaf order.

crates/lib/core/asm/collections/smt.masm
huitseeker: Done. `get` now uses a separate helper that only tracks the matching key. `set` keeps the helper that tracks the first valid insertion position, since it needs that when the key is absent.

Both helpers still validate that the multi-leaf keys are strictly sorted and unique.

crates/lib/core/asm/collections/smt.masm
huitseeker: Done. I removed the unused local, added a locals layout comment, and split the helper names so the get and
set paths are easier to follow.

Al-Kindi-0 (approved):
LGTM now

Issue #3251, opened 2026-06-12
refactor: replace hardcoded AIR reordering with a committed permutation tree


Nashtare:
##  Context

The recursive STARK verifier supports a multi-AIR statement (currently Core + Chiplets). The ACE constraint circuit is generated once, in a fixed caller order ([Core, Chiplets]), but the proof has trace data in proof order, with AIRs sorted by ascending trace height, which isn't robust and does not support well incoming new AIRs without deeper code refactoring. The way we currently address this is by reordering the trace data in MASM so that the proof-order data matches the caller-order in the circuit. 

The main problems it causes:
- hardcoded to 2 AIRs (Core and Chiplets), with not easy generalization to N AIRs (we are going to support more soon with the Precompile VM)
- it is relatively flaky and harder to audit
- it does not scale well with the number of AIRs if we continue doing this "combinatorial" trick in the circuit when re-ordering

## Solution

Initially suggested by @Al-Kindi-0, one better way is to instead commit to every valid AIR ordering up front and let the verifier open the one that matches the proof. Basically, that means having a Merkle Tree whose leaves cover all possible combination of supported AIRs, representing the corresponding ACE circuit generated in those exact orders. We'd be then seeding the transcript with the root of that Merkle tree (among already existing other seeds), and at verification time, we'd derive the proof-order AIR indices by opening the matching leaf and run the corresponding circuit.

This approach is fundamentally better because it is:

- generic in the number of AIRs
- conceptually cleaner and harder to mess up

Al-Kindi-0:
Thanks @Nashtare !
The above is already part of some ongoing work I am doing at the moment and should land pretty soon.
Unless this is urgent I can take, but otherwise feel free to assign it to yourself or someone else.

PR #3256, opened 2026-06-16
feat(verifier): assert the outer-LogUp boundary in MASM; restructure kernel public inputs

### LINKED ISSUE #3033: Consider moving all logUp outer check logic to MASM
With the the logUp abstraction, and assuming a multi-trace setup, we end up with one auxiliary value per trace representing the sum of all the interactions in that trace (by using only the first column as running sum).

This means that the "outer" logup checks need to sum
- 2x auxiliary boundary values
- 2x `log_precompile` init and final fractions (
- Nx `kernel_rom` init fractions
- 1x `program_hash` fraction

All of the fractions' messages are single words, so it seems feasible to accumulate these in a loop without too much extra logic (mainly memory layout).

Pros:
- Remove outer check logic from ACE
- Fewer ACE hashes and evaluations
- More uniform MASM
- Only one place where we have to define outer logup interactions.

The main reason why not to do this is if it costs too many MASM instructions or if the code becomes more complicated than it is currently.

### LINKED ISSUE #3166: refactor: make the ace codegen independent of the VM's air
```rust
    /// Per-AIR lifted selectors for Core at `z^{r_core}` (`r_core = n_max / n_core`).
    /// Equal to the canonical `IsFirst`/`IsLast`/`IsTransition` when Core is at log_max.
    /// Only present in `is_multi_air = true`.
    IsFirstCore,
    IsLastCore,
    IsTransitionCore,
    /// Per-AIR lifted selectors for Chiplets at `z^{r_chip}`. Mirror of `*Core`.
    IsFirstChip,
    IsLastChip,
    IsTransitionChip,
```

> As a follow up, we should investigate making this generic over any number of traces, and keep the ace codegen independent of the VM's air

_Originally posted by @adr1anh in https://github.com/0xMiden/miden-vm/pull/3115#discussion_r3258159201_

### LINKED ISSUE #3246: Define a canonical kernel commitment
# Define a canonical kernel commitment

## Summary

Miden VM does not currently have a single canonical commitment to a `Kernel`.

Today, a kernel is represented by its canonical list of procedure digests, and proofs bind to that kernel by observing the raw digest list as variable-length statement data. This issue proposes defining a canonical `kernel_commitment` and using that commitment as the stable identifier for a kernel.

The proposed commitment is `Poseidon2::hash_elements(flatten(kernel.proc_hashes()))`, where `kernel.proc_hashes()` is the canonical ordered list already maintained by `Kernel`.

## Motivation

A kernel commitment is useful outside the proving stack.

Downstream users and tooling may need to:

- identify which kernel a program was run with;
- compare a program’s kernel against a blessed or well-known kernel;
- check that expected kernel procedures are present;
- use a compact kernel identifier in higher-level protocols.

It is also useful for recursive verification. The recursive verifier should bind the proof to the kernel without observing the entire variable-length kernel digest list directly in the transcript.

For batches of recursive verifications using the same kernel, an optimized MASM API can also avoid rehashing the same kernel digest list for every proof by accepting a precomputed `kernel_commitment`.

## Current behavior

There is currently no `Kernel::commitment()` or equivalent API.

The relevant behavior today is:

- `Kernel` stores a canonical sorted list of procedure digests.
- `ProgramInfo` stores the program hash and the full `Kernel`.
- `Program::hash()` is only the program entrypoint/root digest; it does not include the kernel.
- `PublicInputs::to_air_inputs()` returns fixed public values plus the raw kernel digest felts.
- `Statement::new(...)` receives those raw kernel felts as aux / variable-length input data.
- The recursive verifier mirrors this by loading kernel digests from advice, observing them into the transcript, and also reducing them for the Kernel ROM LogUp boundary correction.

So today the kernel is transcript-bound by observing the raw canonical kernel digest list, not by observing a commitment.

## Proposal

Add a canonical kernel commitment:

- `Kernel::commitment() -> Word`;
- optionally `ProgramInfo::kernel_commitment() -> Word` as a convenience wrapper.

The commitment should be computed as `Poseidon2::hash_elements(flatten(kernel.proc_hashes()))`.

Details:

- each kernel procedure digest is a `Word` of 4 field elements;
- the input is the flat concatenation of all procedure digest elements in canonical kernel order;
- the empty kernel commitment is `Poseidon2::hash_elements([])`;
- use the standard `hash_elements` layout;
- do not introduce a kernel-specific domain separator for now;
- do not append an explicit zero word for odd numbers of digests.

This gives downstream users a stable identifier for a kernel and gives the verifier a fixed-size value to observe instead of the full digest list.

## Transcript direction

Raw kernel digest felts should no longer be observed directly into the transcript.

Instead, the transcript should observe `kernel_commitment`.

The raw kernel digest list is still needed internally for the Kernel ROM LogUp boundary correction, but it should be treated as eval-only data rather than transcript-observed statement data.

Open question: should the transcript also observe `num_kernel_digests`, or is observing `kernel_commitment` sufficient?

## Native verifier behavior

The native Rust verifier should remain safe.

It should not require callers to provide a precomputed `kernel_commitment`. Instead, it should recompute the commitment from the supplied `ProgramInfo` / `Kernel`.

That keeps the public verifier API sound by construction:

- the verifier receives the kernel;
- it computes `Kernel::commitment()`;
- it observes the commitment into the transcript;
- it still uses the raw kernel digest list internally for `eval_external`.

Accepting a precomputed commitment may be useful in MASM, but it should not be part of the safe native verifier API.

## Recursive verifier impact

For recursive verification, this is part of the broader effort to define a clean recursive verifier API.

The low-level MASM verifier may want an optimized unchecked API that receives:

- a precomputed `kernel_commitment`;
- a pointer to the raw kernel digest list in memory;
- the number of kernel digests.

The commitment is used for transcript binding. The raw digest list is used for the Kernel ROM LogUp reduction.

This API is unchecked because the verifier does not prove that the precomputed commitment matches the memory region. If a malicious caller provides a commitment for one kernel and a pointer/count for another, the transcript and LogUp reduction refer to different kernels.

A safe recursive wrapper can avoid this by obtaining the digest list, writing it to memory, computing `kernel_commitment`, and then calling the low-level verifier.

This gives us both:

- a safe wrapper for general use;
- an optimized unchecked path for cases like verifying many proofs with the same kernel.

## Memory-layout note

The recursive verifier should work with a natural-order contiguous list of kernel digest words in memory.

The pointer should be word-aligned. `mem_stream` reads two consecutive words and advances by 8 field elements; the underlying word reads require word alignment. It does not appear to require the starting pointer to be double-word aligned.

For an odd number of digests, an implementation may write an extra zero word after the list so `mem_stream` can read a full final double-word. That extra word is only an implementation convenience and is not part of the `hash_elements` input.

## Maximum kernel count

The maximum number of kernel digests should be handled carefully.

This bound is part of the AIR / multi-AIR definition, not ordinary per-statement data. If it needs to be cryptographically bound, it should likely be covered by the relation or circuit commitment rather than observed as a separate statement element.

## Open questions

1. Is `Poseidon2::hash_elements(flatten(kernel.proc_hashes()))` the right canonical kernel commitment?
2. Should `num_kernel_digests` be observed alongside `kernel_commitment`?
3. How should the maximum number of kernel digests be represented and bound?
4. What shape should the STARK statement API take to separate transcript-observed kernel metadata from eval-only raw kernel felts?
5. What safe recursive verifier wrapper should be exposed for callers that provide kernel digests through advice?

## Acceptance criteria

- Miden VM has a canonical definition of `kernel_commitment`.
- `Kernel::commitment()` is added as the canonical Rust API.
- Native verification recomputes the commitment from `ProgramInfo` / `Kernel`.
- Recursive verification observes the kernel commitment instead of raw kernel digest felts.
- Raw kernel digests remain available for Kernel ROM LogUp boundary correction.
- The issue clearly distinguishes the safe native API, safe recursive wrapper, and optimized unchecked recursive verifier path.

Nashtare:
### Summary

  Moves the multi-AIR outer-LogUp boundary check into the MASM recursive verifier and removes gamma batching from the ACE circuit, makes the ACE layout generic over the number of AIRs, and restructures the
  kernel/public-input surface so the verified statement is fixed-size and caller-bindable with a fully rate-aligned Fiat-Shamir schedule.
  
  Supersedes #3096.

  Closes #3033.
  Closes #3166.
  Closes #3246.
  
  ### Motivation

  The previous design batched the per-AIR LogUp finals into the ACE circuit via a sampled `gamma`, which had no Rust mirror and was unsound as a standalone check. This PR asserts the boundary identity directly in MASM
  against the FS-bound aux-trace finals, eliminates `gamma`, and in the process cleans up the public-input handling that fed it.

  ### Changes

  **1. Outer-LogUp boundary in MASM (`9133e701e`)**

  - The verifier computes the boundary correction `c_total` from the public inputs and asserts `c_total + aux_sum == 0` against the absorbed per-AIR LogUp finals (`aux_trace.masm`, `public_inputs.masm`), mirroring
  `MidenMultiAir::eval_external`.
  - Drops `gamma` batching from ACE entirely (`air/src/ace.rs`, `crates/ace-codegen/`), removing the unsound sampled-gamma path.

  **2. ACE layout generic over AIR count (`e8f9f7e79`)**

  - Generalizes the ACE codegen layout (`keys.rs`, `plan.rs`, `policy.rs`, `pipeline.rs`) from the hardcoded 2-AIR shape to N AIRs (β-fold, per-AIR selectors, prefix-sum offsets), keeping the current Core + Chiplets config
  as the instantiation.

  **3. `kernel_H` public-input restructure + rate-aligned FS (`0d6f2215a`)**

  - Kernel digests enter the transcript through a single Poseidon2 commitment `kernel_H` (`hash_kernel_digests`) instead of an unbounded digest list; callers read a typed 16-felt reduced-inputs block `[kernel_H | pad | 
  program_digest | transcript_state]`.
  - `MidenMultiAir::observe` emits a rate-aligned 6-block FS schedule mirrored by MASM `process_public_inputs`; the scratch-gather absorption path is deleted.
  - The ACE READ section now carries only the 32 stack-i/o felts; the VLPI region, `InputKey::VlpiReduction`, and `AceConfig::num_vlpi_groups` are removed.

  ### Test plan

  - [ ] `cargo test -p miden-air` — relation-digest snapshot + ACE codegen
  - [ ] `cargo test -p miden-core-lib` — stark/crypto suites incl. `ace_read_check` and `variable_length_public_inputs` (cross-checks `kernel_H` / FLPI / `c_total` over 0..255 kernel digests)
  - [ ] Recursive e2e: `stark_verifier_*`, `test_poseidon2_prove_verify`
  - [ ] Codex review of the diff

- Updated `CHANGELOG.md'


crates/test-utils/src/recursive_verifier.rs
```diff
@@ -171,19 +171,9 @@ fn build_advice(
...
-    // 2. Number of kernel procedure digests.
-    let num_kernel_proc_digests = kernel_digests.len();
-    advice_stack.push(num_kernel_proc_digests as u64);
-
-    // 3. Kernel procedure digest elements (each digest padded to 8 elements, reversed).
-    let kernel_advice = build_kernel_digest_advice(kernel_digests);
-    advice_stack.extend_from_slice(&kernel_advice);
-
-    // 4. Auxiliary randomness [beta0, beta1, alpha0, alpha1].
+    // 1. Auxiliary randomness [beta0, beta1, alpha0, alpha1] — consumed first by
```
huitseeker: Could you also update the module banner above? It still says the advice order is `fixed-length PI -> num_kernel_proc_digests -> kernel_digests -> aux randomness`, but this code now puts aux randomness first, then `N`/digests, then the public-input tail. The local comments in `build_advice` look right, so this seems like just stale top-level docs.

huitseeker (approved):
Much nicer, thanks a lot! One inline nit

air/src/ace.rs
```diff
@@ -1,7 +1,7 @@
 //! ACE circuit integration for the multi-AIR (CoreAir + ChipletsAir) proof.
 //!
-//! This module extends the constraint-evaluation DAG produced by
-//! `build_ace_dag_for_air` with the LogUp auxiliary-trace boundary check:
+//! The circuit checks the β-folded constraint composition of the two AIRs. The
+//! LogUp auxiliary-trace boundary identity
```
Al-Kindi-0: Maybe worth being general instead of saying "two AIRs"

air/src/ace.rs
```diff
@@ -10,16 +10,11 @@
...
-//! (`c_bh`, `c_lp`), so they are rebuilt directly inside the DAG as rational
-//! fractions `(n, d)` and folded into a running rational `(N, D)` without any
-//! in-circuit inversion. The kernel-ROM correction depends on the variable-
-//! length kernel digest list which the circuit can't walk, so MASM computes
-//! it (one final `ext2inv`) and hands it in as a single scalar via the
-//! existing `VlpiReduction(0)` input. The final boundary check is the
-//! quadratic identity `(Σ aux_bound + c_kr) · D + N = 0`.
+//! is *not* part of the circuit: the MASM verifier asserts it directly on the
+//! operand stack (see `sys/vm/public_inputs.masm` / `sys/vm/aux_trace.masm`),
+//! mirroring the native verifier's `MidenMultiAir::eval_external` check.
```
Al-Kindi-0: implementation/historical comment that should be removed

air/src/ace.rs
```diff
@@ -31,180 +26,19 @@ use miden_crypto::{
...
-///   `combined + γ · boundary = 0`
-/// where `combined = chip_acc · β_multi + core_acc` is the β-folded sum of the per-AIR
-/// alpha-folded constraint roots, and `boundary` is the cross-AIR LogUp identity built
-/// over both AIRs' aux-bus-boundary slots plus the open-bus rational corrections plus
-/// the kernel-ROM scalar correction.
+///   `combined = 0`
+/// where `combined = chip_acc · β_multi + core_acc - q·v` is the β-folded sum of the per-AIR
+/// alpha-folded constraint roots minus the shared quotient binding. The cross-AIR LogUp
+/// boundary identity is asserted separately by the MASM verifier (it is not batched into
+/// the circuit).
```
Al-Kindi-0: Again same kind of comments, this module shouldn't know about the number of AIRs.
Also better to remove the mentions of the MASM verifier

air/src/ace.rs
```diff
@@ -213,10 +47,8 @@ where
 /// 2. Re-emit each sub-DAG's nodes into a fresh `DagBuilder` configured for the *combined* layout.
 ///    The chiplets sub-DAG's input keys are rewritten so its main/aux/bus-boundary slot indices
 ///    land in the chiplets-half of the combined layout. The core sub-DAG passes through unchanged.
-/// 3. β-fold: `combined = MultiAirBetaCore · core_acc + MultiAirBetaChip · chip_acc`, where the
+/// 3. β-fold: `combined = MultiAirBeta(0) · core_acc + MultiAirBeta(1) · chip_acc - q·v`, where the
```
Al-Kindi-0: We can be general over the AIRs

air/src/ace.rs
```diff
@@ -225,8 +57,8 @@ where
     SymbolicExpressionExt<Felt, EF>: Algebra<EF>,
 {
     assert!(
-        config.is_multi_air,
-        "build_multi_air_ace_circuit requires AceConfig::is_multi_air = true"
+        config.num_airs == 2,
+        "build_multi_air_ace_circuit is the (Core, Chiplets) builder; AceConfig::num_airs must be 2"
```
Al-Kindi-0: Is there a reason for this ?

air/src/ace.rs
```diff
@@ -275,25 +107,47 @@ where
     //   the wire byte order exactly. Padding slots within each AIR's subregion are unreferenced by
     //   the constraints (see eval bodies of CoreAir and ChipletsAir, which only address columns up
     //   to the original width).
+    // - `num_public` covers only the circuit-referenced window of the statement's public values:
+    //   the 32 stack-i/o felts at `pv[4..36]`. The program hash and transcript state are consumed
+    //   by the MASM-side outer-LogUp boundary check and the FS transcript, not by the circuit, so
+    //   they are kept out of the ACE READ section; `Public` keys are shifted by `-PV_WINDOW_START`
+    //   during re-emit below.
```
Al-Kindi-0: I can see the optimization here but is it really worth the implementation/mental overhead?

air/src/ace.rs
```diff
@@ -371,22 +227,17 @@ where
         });
     }
 
-    let mab_core = builder.input(InputKey::MultiAirBetaCore);
-    let mab_chip = builder.input(InputKey::MultiAirBetaChip);
+    let mab_core = builder.input(InputKey::MultiAirBeta(0));
+    let mab_chip = builder.input(InputKey::MultiAirBeta(1));
     let core_term = builder.mul(mab_core, core_acc);
```
Al-Kindi-0: While we are at it, why not make the ACE codegen generic over the AIRs ?

air/src/ace.rs
```diff
@@ -371,22 +227,17 @@ where
...
     let chip_term = builder.mul(mab_chip, chip_acc);
     let combined_acc = builder.add(core_term, chip_term);
+    // SAFETY-CRITICAL invariant: this `sub` must be the *last* operation emitted into the
+    // builder, since the MASM ACE chip's "is the last op zero?" check evaluates that node
+    // as the root.
     let combined_constraint = builder.sub(combined_acc, chip_qv);
```
Al-Kindi-0: I think this and `build_multi_air_ace_circuit` need to be re-worked in order to really for `build_multi_air_ace_circuit` to live up to its name

air/src/lib.rs
```diff
@@ -632,6 +642,45 @@ impl<EF: ExtensionField<Felt>> MultiAir<Felt, EF> for MidenMultiAir {
         Kernel::MAX_NUM_PROCEDURES * WORD_SIZE
     }
 
+    /// Absorb statement-owned public inputs into the Fiat-Shamir challenger.
+    ///
+    /// Replaces the default length-prefixed stream with the rate-aligned schedule
+    /// the MASM recursive verifier mirrors (six 8-felt blocks, 48 felts total):
+    ///
+    /// ```text
+    /// [ kernel_H (4) | 0,0,0,0 ]      Poseidon2 summary of the kernel digests
```
Al-Kindi-0: I would remove `Poseidon2` here and make the comment purposeful

air/src/lib.rs
```diff
@@ -632,6 +642,45 @@ impl<EF: ExtensionField<Felt>> MultiAir<Felt, EF> for MidenMultiAir {
...
+        assert_eq!(air_inputs.len(), NUM_PUBLIC_VALUES, "unexpected public-value count");
+
+        let kernel_h = hash_kernel_digests(aux_inputs);
+        for v in kernel_h {
+            challenger.observe(v);
+        }
+        for _ in 0..WORD_SIZE {
+            challenger.observe(Felt::ZERO);
+        }
+        let program_hash = &air_inputs[PV_PROGRAM_HASH..PV_PROGRAM_HASH + WORD_SIZE];
```
Al-Kindi-0: Why absorbing the ZEROs ? The kernel digest and program hash could be absorbed in one absorb, no ?

air/src/lib.rs
```diff
@@ -659,6 +708,28 @@ impl<EF: ExtensionField<Felt>> MultiAir<Felt, EF> for MidenMultiAir {
...
+/// Computes `kernel_H`, the fixed-size Poseidon2 commitment to the kernel-procedure digests.
+///
+/// This is the canonical [`Kernel::commitment`] value expressed over the flattened digest
+/// felts: the Poseidon2 linear hash (`hash_elements`) of `kernel_felts`, packing eight felts
+/// (two digests) per permutation. The empty digest list yields `hash_elements(&[])`.
+///
+/// The MASM recursive verifier recomputes the same value from the materialized digests
+/// (`sys/vm/public_inputs.masm`); both sides absorb `kernel_H` into the Fiat-Shamir transcript
+/// in place of the unbounded digest list, and callers of the recursive verifier read it back
+/// as the kernel commitment of the verified statement.
```
Al-Kindi-0: Do we need to again talk about the MASM verifier here ?

core/src/program/kernel.rs
```diff
@@ -89,6 +92,16 @@ impl Kernel {
...
+
+    /// Returns the canonical commitment to this kernel: the Poseidon2 linear hash of the
+    /// flattened procedure digests.
+    ///
+    /// This matches the kernel commitment computed by the protocol and is the fixed-size
+    /// identifier observed by the recursive verifier in place of the raw digest list.
+    pub fn commitment(&self) -> Word {
+        let elements: Vec<Felt> = self.0.iter().flat_map(Word::as_elements).copied().collect();
+        hasher::hash_elements(&elements)
+    }
```
Al-Kindi-0, line 104: Seems a duplicate method, no ?

crates/ace-codegen/src/layout/policy.rs
```diff
@@ -138,7 +137,7 @@ impl InputLayout {
         //   3    is_first            EF      Precomputed: (z^N - 1) / (z - 1)
         //   4    is_last             EF      Precomputed: (z^N - 1) / (z - g^{-1})
         //   5    is_transition       EF      Precomputed: z - g^{-1}
-        //   6    gamma               EF      Batching challenge
+        //   6    gamma               EF      Reserved (zero)
```
Al-Kindi-0: Should probably be removed now

crates/ace-codegen/src/pipeline.rs
```diff
@@ -38,18 +38,13 @@ pub enum LayoutKind {
...
     /// Layout policy (Native vs Masm).
     pub layout: LayoutKind,
-    /// Whether this circuit is a multi-AIR combined circuit. When `true`, the
-    /// stark-vars region reserves additional EF slots for the multi-AIR β
-    /// coefficients (one per AIR) and per-AIR lifted selector triples (one per
-    /// AIR). Default: `false` (single-AIR layout).
-    pub is_multi_air: bool,
+    /// Number of AIRs the circuit is built over. With `1` the layout is the plain
+    /// single-AIR layout; with `num_airs >= 2` the stark-vars region reserves
+    /// additional EF slots for the per-AIR multi-AIR β coefficients (one per AIR)
```
Al-Kindi-0, line 45: Hmm, shouldn't we just use beta and then Horner inside the circuit ? I am thinking that this is part of making the ACE pipeline truly multi-AIR

crates/lib/core/asm/stark/public_inputs.masm
```diff
@@ -69,62 +51,56 @@ end
...
+#! The reduced-inputs block is NOT consumed by the ACE circuit; it sits there as a stable
+#! caller-readable summary of the variable-length public inputs after `verify_proof` returns.
+#!
+#! Inputs:  [num_fixed_len_pi, ...]
 #! Outputs: [...]
 #!
 #! Invocation: exec
 pub proc compute_and_store_public_inputs_address
-    # 1) Get a pointer to where OOD evaluations are stored
+    # 1) pi_ptr = OOD_EVALUATIONS_PTR - 8 - num_fixed_len_pi * 2
```
Al-Kindi-0, line 70: This is a clear regression in the comment quality here and in other places in the MASM modules

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -7,44 +7,71 @@ use miden::core::crypto::hashes::poseidon2
...
+#   pi_ptr             -->  [ FLPI: 32 base felts as 32 EF slots ]   stack i/o
+#
+#   anchor             -->  [ aux_rand: beta0,beta1,alpha0,alpha1 ]   AUX_RAND_ND_PTR (loaded ND
+#                                                                     here, verified in step II)
+#
+# The outer-LogUp boundary correction accumulated at `C_TOTAL_PTR` is
+#     c_total = Σ_i 1 / ((α + γ) + msg(kernel_digest_i))    (KernelRomInit)
+#             +     1 / ((α + 2γ) + msg(program_digest))    (BlockHashTable)
+#             +     1 / (α + 3γ)                            (LogPrecompileTranscript init)
+#             −     1 / ((α + 3γ) + msg(transcript_state))  (LogPrecompileTranscript final)
```
Al-Kindi-0: magic constants

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
-    push.43 exec.constants::tmp3 mem_load mul.4 add
-    # => [L]
-    dup exec.constants::absorb_scratch_ptr add exec.constants::tmp1 mem_store
-    u32divmod.8 drop
-    # => [nfb]
+    # Write the zero pad word at reduced_inputs+4 (the FS-alignment pad).
+    padw
+    exec.constants::reduced_inputs_address_ptr mem_load add.4
+    mem_storew_le
+    dropw
```
Al-Kindi-0: The `dropw` followed by a `padw`  for `mem_loadw_le` is a very common LLM pattern, would be good to avoid at least 8 cycles, especially if the loop is hot

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
-    end
-    drop drop
-
-    # scratch[41] = MAX_AUX_INPUTS, scratch[42] = len(kernel_felts) = 4 * num_kernel_proc_digests.
-    # (scratch[43..43+4K] was already filled in natural order by `load_kernel_digests`.)
-    push.MAX_AUX_INPUTS exec.constants::absorb_scratch_ptr add.41 mem_store
-    exec.constants::tmp3 mem_load mul.4 exec.constants::absorb_scratch_ptr add.42 mem_store
-
-    # ---- Block-absorb scratch[0..L], L = 43 + 4 * num_kernel_proc_digests ----
+proc observe_and_fold_kernel
```
Al-Kindi-0: nit: i would split the observe from the fold

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
+        padw dup.4 mem_loadw_le
+        # => [d0, d1, d2, d3, ptr, N, ...]
+        exec.fold_digest_word
+        # => [msg0, msg1, ptr, N, ...]
+        push.BUS_ID_KERNEL_ROM_INIT_PLUS_ONE
+        exec.push_alpha_plus_i_gamma
+        # => [p0, p1, msg0, msg1, ptr, N, ...]
+        ext2add ext2inv
+        # => [t0, t1, ptr, N, ...]
+        exec.add_into_c_total
```
Al-Kindi-0: Would be better to keep the values on the stack, deeper, and then do when fold at the end, instead of the loading the accumulator from memory for each term to accumulate

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
 #! Output: [...]
-proc finalize_kernel_reduction
-    # Load the auxiliary randomness (alpha and beta) from advice.
+proc consume_small_message_sub_from_c_total
+    exec.consume_small_message_compute_inv_term
+    # => [inv_term0, inv_term1, ...]
+    neg swap neg swap
+    # => [-inv_term0, -inv_term1, ...]
+    exec.add_into_c_total
+end
```
Al-Kindi-0: Do we really need both ? We can just compute the requests separately from the responses and do one final subtraction. Also reduces code surface/bloat

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
+#! Output: [inv_term0, inv_term1, ...]
+proc consume_small_message_compute_inv_term
     padw adv_loadw
-    exec.constants::aux_rand_nd_ptr mem_storew_le
-    dropw
+    # => [d0, d1, d2, d3, bus_id, offset, ...]
+
+    # Store the word canonically at reduced_inputs + offset; the word stays on the stack
+    # (mem_storew_le pops only the address). `offset` sits at depth 6 once `base` is on
+    # top, so `dup.6 add` builds the target address.
```
Al-Kindi-0: This is a work-in-progress comment and should be removed. Readers reaching this place do not need to be reminded of the semantics of `mem_storew_le`

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +83,416 @@ end
...
+    exec.constants::reduced_inputs_address_ptr mem_load
+    # => [base, d0, d1, d2, d3, bus_id, offset, ...]
+    dup.6 add
+    # => [target, d0, d1, d2, d3, bus_id, offset, ...]
+    mem_storew_le
+    # => [d0, d1, d2, d3, bus_id, offset, ...]   (mem written canonical)
```
Al-Kindi-0: same for this comment

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -7,44 +7,71 @@ use miden::core::crypto::hashes::poseidon2
...
+#   anchor             -->  [ aux_rand: beta0,beta1,alpha0,alpha1 ]   AUX_RAND_ND_PTR (loaded ND
+#                                                                     here, verified in step II)
+#
+# The outer-LogUp boundary correction accumulated at `C_TOTAL_PTR` is
+#     c_total = Σ_i 1 / ((α + γ) + msg(kernel_digest_i))    (KernelRomInit)
+#             +     1 / ((α + 2γ) + msg(program_digest))    (BlockHashTable)
+#             +     1 / (α + 3γ)                            (LogPrecompileTranscript init)
+#             −     1 / ((α + 3γ) + msg(transcript_state))  (LogPrecompileTranscript final)
+# with `γ = β^16` and `msg(w) = Σ_{i<4} w_i·β^i`, computed using the non-deterministically
+# supplied `α, β` (verified against the FS-sampled values by `generate_aux_randomness` in
```
Al-Kindi-0: Where is this non-determinism verified ? I can't seem to locate it I guess. Also, is the use of non-determinism still needed here. The reason this was used earlier was because we could stream things much more than the proposed implementation in this PR, but maybe now we can do a two pass and non-determinism usage is not justified anymore

Al-Kindi-0 (commented):
Thanks you

I did an initial pass and I left a few comments.
The more general thought that I was having while reviewing is that the code has become more complex and also, it seems, more costly.
Would be good to benchmark the number of cycles it will take to process say 64 kernel procedure digests, but my intuition tells me that it is non-negligible.

So while the idea of moving the VLPI handling from ACE is defensible engineering-wise, performance and maintainability should be also be considered.

air/src/ace.rs
Nashtare: Just as a safeguard given the current config but I'll drop it while refactoring based on follow-up comments.

crates/lib/core/asm/sys/vm/public_inputs.masm
```diff
@@ -56,434 +97,363 @@ end
...
+    movup.2 drop
+    # => [inv_term0, inv_term1, ...]
+end
 
-    # ---- Block-absorb scratch[0..L], L = 43 + 4 * num_kernel_proc_digests ----
+#! Observes the 16-felt reduced-inputs window at `reduced_inputs_ptr + 0..+15` into the FS
+#! transcript as two rate-aligned 8-felt chunks via `mem_stream + permute`:
+#!   Chunk 1: `[kernel_H | program_digest]`   at reduced_inputs+0..+7.
+#!   Chunk 2: `[transcript_state | zero_pad]` at reduced_inputs+8..+15.
+#! The trailing pad word is read back zero from the VM's zero-initialized memory.
```
huitseeker: This pad word is part of the Fiat-Shamir input, but nothing writes it before `mem_stream` absorbs it. I could reproduce this locally by pre-writing a nonzero word at `reduced_inputs_ptr + 12` before `process_public_inputs`; the pad still read back as `16` afterward. Could we store an explicit zero word there before absorbing the reduced-input window?

crates/lib/core/asm/stark/constants.masm
```diff
@@ -86,8 +86,9 @@ const OOD_FIXED_TERM_HORNER_EVALS_PTR = 3223322664
 ### Address storing trace domain generator
 const TRACE_DOMAIN_GENERATOR_PTR = 3223322669
 
-### Address storing a pointer to variable length public inputs
-const VARIABLE_LEN_PUBLIC_INPUTS_ADDRESS_PTR = 3223322670
+### Address storing a pointer to the caller-facing reduced-inputs block
+### [kernel_H | zero_pad | program_digest | transcript_state] (16 felts, just below FLPI).
```
huitseeker: This line still has the old reduced-input order. The current code writes and absorbs `[kernel_H | program_digest | transcript_state | zero_pad]`, and the longer layout comment below says that too. Could this header use the same order so callers do not bind the wrong 16-felt window?

air/src/lib.rs
```diff
@@ -632,6 +642,50 @@ impl<EF: ExtensionField<Felt>> MultiAir<Felt, EF> for MidenMultiAir {
...
+        challenger: &mut C,
+        air_inputs: &[Felt],
+        aux_inputs: &[Felt],
+        _log_trace_heights: &[u8],
+    ) {
+        assert_eq!(air_inputs.len(), NUM_PUBLIC_VALUES, "unexpected public-value count");
+
+        let kernel_h = hash_kernel_digests(aux_inputs);
+        let program_hash = &air_inputs[PV_PROGRAM_HASH..PV_PROGRAM_HASH + WORD_SIZE];
+        let transcript_state = &air_inputs[PV_TRANSCRIPT_STATE..PV_TRANSCRIPT_STATE + WORD_SIZE];
```
adr1anh: If we place these in the `aux_inputs`, the public inputs section of the ACE becomes smaller, since they are unused by the Air.

crates/lib/core/asm/sys/vm/public_inputs.masm
Nashtare: You're right we don't really need it anymore, I'll go with the two pass

crates/ace-codegen/src/pipeline.rs
Nashtare: I agree, but I think to be truly generic we'd first need #3251 to land, as we currently rely on that fixed order of beta coeffs.

Nashtare:
> Thanks you
> 
> I did an initial pass and I left a few comments. The more general thought that I was having while reviewing is that the code has become more complex and also, it seems, more costly. Would be good to benchmark the number of cycles it will take to process say 64 kernel procedure digests, but my intuition tells me that it is non-negligible.
> 
> So while the idea of moving the VLPI handling from ACE is defensible engineering-wise, performance and maintainability should be also be considered.

Thank you for the review @Al-Kindi-0! 
I benchmarked `next` (`process_public_inputs`) vs this branch (`process_public_inputs + compute_outer_logup_correction`), as of commit [bd06067](https://github.com/0xMiden/miden-vm/pull/3256/commits/bd060678a699fdac008e0b3705feea210fb40078) after addressing the main concerns. Results below:

```bash
  ┌─────┬───────┬─────────────┐
  │  N  │ next  │ this branch │
  ├─────┼───────┼─────────────┤
  │   0 │  2015 │        1415 │
  ├─────┼───────┼─────────────┤
  │   1 │  2326 │        1679 │
  ├─────┼───────┼─────────────┤
  │  16 │  4297 │        3577 │
  ├─────┼───────┼─────────────┤
  │  32 │  6585 │        5745 │
  ├─────┼───────┼─────────────┤
  │  64 │ 11161 │       10081 │
  ├─────┼───────┼─────────────┤
  │ 128 │ 20313 │       18753 │
  ├─────┼───────┼─────────────┤
  │ 255 │ 38648 │       36094 │
  └─────┴───────┴─────────────┘

  ```
  
  So it is actually slightly cheaper with the new approach.

crates/ace-codegen/src/layout/policy.rs
Nashtare: It's flagged as `Reserved` now to not mess up the MASM entirely by re-shifting (and one pad slot is required anyway because of word-alignment)

Al-Kindi-0:
> > Thanks you
> > I did an initial pass and I left a few comments. The more general thought that I was having while reviewing is that the code has become more complex and also, it seems, more costly. Would be good to benchmark the number of cycles it will take to process say 64 kernel procedure digests, but my intuition tells me that it is non-negligible.
> > So while the idea of moving the VLPI handling from ACE is defensible engineering-wise, performance and maintainability should be also be considered.
> 
> Thank you for the review @Al-Kindi-0! I benchmarked `next` (`process_public_inputs`) vs this branch (`process_public_inputs + compute_outer_logup_correction`), as of commit [bd06067](https://github.com/0xMiden/miden-vm/pull/3256/commits/bd060678a699fdac008e0b3705feea210fb40078) after addressing the main concerns. Results below:
> 
> ```shell
>   ┌─────┬───────┬─────────────┐
>   │  N  │ next  │ this branch │
>   ├─────┼───────┼─────────────┤
>   │   0 │  2015 │        1415 │
>   ├─────┼───────┼─────────────┤
>   │   1 │  2326 │        1679 │
>   ├─────┼───────┼─────────────┤
>   │  16 │  4297 │        3577 │
>   ├─────┼───────┼─────────────┤
>   │  32 │  6585 │        5745 │
>   ├─────┼───────┼─────────────┤
>   │  64 │ 11161 │       10081 │
>   ├─────┼───────┼─────────────┤
>   │ 128 │ 20313 │       18753 │
>   ├─────┼───────┼─────────────┤
>   │ 255 │ 38648 │       36094 │
>   └─────┴───────┴─────────────┘
> ```
> 
> So it is actually slightly cheaper with the new approach.

Thank you for the investigation, @Nashtare !
After the optimizations it is not surprising and this is also the reason that `next` is not a good baseline to compare against.
In the grand scheme of things, I think we can just push ahead and then maybe comeback to this if it turns out to be a bottleneck.

Al-Kindi-0:
@Nashtare , what is the status here ? It seems to me that all comments are addressed, right ? If yes, let us get it out of the way

Nashtare:
> @Nashtare , what is the status here ? It seems to me that all comments are addressed, right ? If yes, let us get it out of the way

I've implemented a final feature request from @adr1anh for which I'd appreciate his review, otherwise no blocking comments I believe.

PR #3260, opened 2026-06-17
fix: Bound deferred precompile requests

huitseeker:
`AdviceProvider` could keep adding deferred precompile requests without a count limit or calldata byte limit. A program could grow the request  list until the host ran out of memory.

This adds execution options for the maximum request count and total calldata bytes. `AdviceProvider` now rejects additions that would exceed either limit before it changes the stored requests.

Closes 0xMiden/security-issues#14.

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -23,6 +23,10 @@ pub struct ExecutionOptions {
     /// Maximum number of continuations allowed on the continuation stack at any point during
     /// execution.
     max_num_continuations: usize,
+    /// Maximum number of deferred precompile requests allowed during execution.
+    max_precompile_requests: usize,
+    /// Maximum total number of calldata bytes allowed across deferred precompile requests.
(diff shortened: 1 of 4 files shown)


Al-Kindi-0 (approved):
LGTM
It is true that we don't have concrete data point supporting the constants choices but I think this is fine for now and these choices can be changed in the future

processor/src/execution_options.rs
adr1anh, line 22: Does this still make sense with the more general bounds ?

adr1anh (approved):
LGTM. I'll take these into account for the new framework.

processor/src/execution_options.rs
huitseeker: The new limits cap how many deferred requests we keep, and how many total bytes they hold.
  
`max_hash_len_bytes` caps one hash call before we read and hash its input. That protects the host from
one very large request.

I'm tempted to keep both limits but feel free to address in a downstream refactoring.

PR #3261, opened 2026-06-17
fix(processor): stack depth limits include all active contexts

plafer:
Fixes a vulnerability where stack depth limits computation didn't include all active contexts.


huitseeker (approved):
Looks good, thanks a lot!

PR #3262, opened 2026-06-17
Check user doc cycle counts against source of truth (#3111)

### LINKED ISSUE #3111: Check user doc cycle counts against source of truth
User docs repeat cycle counts by hand, so they can drift from the generated core library docs.

Current stale examples:

- `docs/src/user_docs/core_lib/collections.md` says `num_leaves_to_num_peaks` is 67 cycles. `crates/lib/core/docs/collections/mmr.md` says 32.
- `docs/src/user_docs/core_lib/collections.md` says `mmr::unpack` is `162 + 9 * extra_peak_pair`. `crates/lib/core/docs/collections/mmr.md` says `215 + 9 * extra_peak_pair`.
- `docs/src/user_docs/core_lib/crypto/hashes.md` appears stale for Poseidon2 `hash_words` compared with `crates/lib/core/docs/crypto/hashes/poseidon2.md`.

Add a check that fails when user-doc cycle text is out of sync.

Suggested shape:

- Add `scripts/check-user-doc-cycles.sh` and a small parser script.
- Run `MIDEN_BUILD_LIB_DOCS=1 cargo check -p miden-core-lib` first, so `crates/lib/core/docs/**/*.md` is fresh.
- Compare generated `Cycles:` entries against mapped entries in `docs/src/user_docs/core_lib/**/*.md`.
- Start with an explicit mapping file, not fuzzy matching.
- Report the user doc path, procedure, expected text, and actual text on mismatch.

Example mapping:

```toml
[[entry]]
user_doc = "docs/src/user_docs/core_lib/collections.md"
section = "Merkle Mountain Range"
procedure = "unpack"
generated_doc = "crates/lib/core/docs/collections/mmr.md"
generated_procedure = "unpack"
```

Assembly instruction docs need measured fixtures, since they do not have generated MASM docs as a source of truth. Add stable markers in the docs and measure small snippets with `clk`.

Example marker:

```md
<!-- cycle-check: u32popcnt -->
```

Example fixture:

```toml
[[case]]
id = "u32popcnt"
doc = "docs/src/user_docs/assembly/u32_operations.md"
program = "begin clk push.7 u32popcnt drop clk swap sub end"
expected = "32 cycles"
```

----

A stronger long-term fix is to generate the core library user-doc reference tables from MASM comments, instead of only checking cycle strings.

The core library already has this source of truth:

- Source: `crates/lib/core/asm/**/*.masm`
- Generator: `crates/lib/core/build.rs`
- Generated docs: `crates/lib/core/docs/**/*.md`

The user docs could keep hand-written overview text, but import generated procedure tables or per-procedure fragments. Those fragments could live under `docs/src/generated/core_lib/**/*.md` and be rendered with a Docusaurus-friendly version of the existing core-lib doc renderer.

Then CI can regenerate the fragments and fail if `git diff -- docs/src/generated` is non-empty. This would catch stale inputs, outputs, panic notes, descriptions, formulas, and cycle counts.

Suggested boundary: generate reference blocks, not whole user-doc pages. Keep conceptual prose hand-written.

crazywriter1:
## Rationale

User-facing core-lib docs copy cycle counts by hand and drift from generated MASM docs. This PR adds CI to catch that drift via explicit mappings, plus a starting assembly check with `clk` fixtures.

## Summary

- Add `scripts/check-user-doc-cycles.sh`, a Python parser, and an explicit `user-doc-cycle-mappings.toml` to compare user-facing core-lib docs against generated MASM docs (`MIDEN_BUILD_LIB_DOCS=1 cargo check -p miden-core-lib`).
- Add assembly cycle fixtures with `<!-- cycle-check: ... -->` markers and a processor test that measures instruction cost via `clk` (with a baseline program to isolate the target instruction).
- Wire the check into `make check-user-doc-cycles` and a new CI job.
- Fix stale cycle text in user docs (collections, hashes, aead, sys) and update `u32popcnt` from 32 to 38 cycles based on `clk` measurement.

## Test plan

- [x] `make check-user-doc-cycles`
- [x] `cargo test -p miden-processor --lib tests::user_doc_assembly_cycle_fixtures_match_documentation -- --exact`

## Follow-up (out of scope for this PR)

- Expand `user-doc-cycle-mappings.toml` to cover remaining core-lib procedures with cycle text (e.g. `find_key_value`, `log_precompile_request`, `math/u64` helpers) once generated docs expose comparable `Cycles:` entries.
- Add more assembly `cycle-check` markers and `clk` fixtures beyond the initial `u32popcnt` example.
- Longer term: generate core-lib reference tables / cycle fragments from MASM comments (issue suggestion) so CI can catch drift in inputs, outputs, and descriptions—not just cycle strings.

Closes #3111


crazywriter1:
@huitseeker Thanks for the detailed issue. The suggested shape (explicit mapping + `clk` fixtures) mapped cleanly to an initial CI check.

**Rationale:** User docs repeat cycle counts by hand, so they drift from generated core-lib MASM docs. This adds a guardrail before that drift lands on `next`.

**What’s in the PR:**
- `make check-user-doc-cycles` + CI job
- `user-doc-cycle-mappings.toml` → compares user docs vs `MIDEN_BUILD_LIB_DOCS=1` generated docs
- Assembly: `cycle-check` marker + `clk` fixture (starting with `u32popcnt`)

**Heads-up:** `clk` measurement (with baseline for `push.7`/`drop`) gives **38 cycles** for `u32popcnt`, not 32 from the issue example — docs updated to match measurement.

Follow-ups (broader mappings, more assembly fixtures, generated reference blocks) are in the PR description. Happy to iterate on the assembly measurement approach if you prefer something else.

scripts/check-user-doc-cycles.sh
```diff
@@ -0,0 +1,14 @@
...
+cd "${ROOT}"
+
+echo "Regenerating core library docs..."
+MIDEN_BUILD_LIB_DOCS=1 cargo check -p miden-core-lib
+
+echo "Checking user doc cycle counts..."
+python3 scripts/check_user_doc_cycles.py
+
+echo "Checking assembly cycle fixtures..."
+cargo test -p miden-processor tests::user_doc_assembly_cycle_fixtures_match_documentation -- --exact
```
huitseeker: This new Make target fails for me at this exact step. `make check-user-doc-cycles` gets through doc regeneration and the Python comparison, then the fixture test aborts with `thread ... has overflowed its stack` and exits 101. Since the new CI job runs this target directly, it looks like the job will fail unless this command gets the same stack setup as the normal test targets, or the fixture is moved behind the existing nextest wrapper.

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,181 @@
...
+import sys
+import tomllib
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parent.parent
+MAPPINGS = Path(__file__).resolve().parent / "user-doc-cycle-mappings.toml"
+ASSEMBLY_FIXTURES = Path(__file__).resolve().parent / "assembly-cycle-fixtures.toml"
+
+
+def extract_cycles_from_description(description: str) -> str:
```
huitseeker, line 26: Could we avoid adding a second Markdown parser here? The core library already has a source-of-truth path: MASM doc comments generate `crates/lib/core/docs/**` when `MIDEN_BUILD_LIB_DOCS=1 cargo check -p miden-core-lib` runs.

  A simpler check would be:

  1. keep cycle text in the MASM doc comments,
  2. regenerate the core docs in CI,
  3. fail if `crates/lib/core/docs/**` has a diff,
  4. make the user docs link to, include, or stay thin around those generated tables.

  That would let us delete this Python normalizer and `scripts/user-doc-cycle-mappings.toml`, and it avoids maintaining a second Markdown grammar just to compare prose back to prose.

  For the one assembly case, I think the same idea applies: a direct Rust assertion for `u32popcnt` cycle metadata, or one inline MASM test, is enough. The TOML fixture plus parser feels heavier than the check it buys us.

huitseeker (changes requested):

crazywriter1:
@huitseeker Pushed a fix for the stack overflow: the assembly fixture test in `check-user-doc-cycles.sh` now uses `RUST_MIN_STACK=16777216` (same as the Makefile test targets). `make check-user-doc-cycles` should pass end-to-end now.

On the Python parser / mapping TOML feedback: agreed that generated-doc diff + thinner user docs is the better long-term shape. Should I refactor this PR to that approach, or merge this guardrail and follow up separately? Happy to do whichever you prefer.

docs/src/user_docs/assembly/instruction_reference.md
```diff
@@ -114,7 +114,7 @@ _Note: Assertions can be parameterized with an error message (e.g., assert.err="
 | `u32shr` <br /> `u32shr.b`   | `[b, a, ...]` | `[c, ...]`   | 20 <br /> 5 | $c = \lfloor a / 2^b \rfloor$. Undefined if $a \geq 2^{32}$ or $b > 31$.    |
```
huitseeker, line 115: Could we guard this duplicate too? The new fixture only marks `u32_operations.md`, so this table can drift from `38` and `make check-user-doc-cycles` would still pass. A second marker or doc entry here would keep both user-facing copies covered.

docs/src/user_docs/core_lib/collections.md
```diff
@@ -20,7 +20,7 @@ The following procedures are available to read data from and make updates to a M
...
+| unpack                       | Writes the MMR whose peaks hash to `HASH` to the memory location pointed to by `mmr_ptr`. Inputs: `[HASH, mmr_ptr, ...]`. Outputs: `[...]`. Where: `HASH` is the MMR peak hash, the hash is expected to be padded to an even length and to have a minimum size of 16 elements. The advice map must contain a key with `HASH`, and its value is `[num_leaves, 0, 0, 0] \|\| hash_data`, and hash_data is the data used to compute `HASH`. `mmr_ptr` is the memory location where the MMR data will be written, starting with the MMR forest (the total count of its leaves) followed by its peaks. The memory location must be word-aligned.<br /><br />Cycles: 162 + 9 * extra_peak_pair cycles<br />where `extra_peak` is the number of peak pairs in addition to the first 16, i.e. `round_up((num_of_peaks - 16) / 2)` |
```
huitseeker: Could we keep the variable name the same here? The formula says `extra_peak_pair`, but the next sentence defines `extra_peak`, so it is hard to tell which one readers should use. Since this comes from generated docs, the source MASM doc may need the same wording fix.

crazywriter1:
@huitseeker Thanks for the follow-up review addressed both points in the latest push:

1. **`extra_peak_pair` wording:** Renamed `extra_peak` → `extra_peak_pair` in the MASM source (`mmr.masm`), regenerated `crates/lib/core/docs/collections/mmr.md`, and updated the matching user doc entry so the formula and definition use the same variable name.

2. **`instruction_reference.md` coverage:** Added `<!-- cycle-check: u32popcnt -->` to the `u32popcnt` row and a second fixture entry in `assembly-cycle-fixtures.toml`, so both `u32_operations.md` and `instruction_reference.md` are guarded by the cycle check.

On the earlier stack overflow feedback: the assembly fixture test in `check-user-doc-cycles.sh` now runs with `RUST_MIN_STACK=16777216` (same as the Makefile test targets).

Still happy to pivot to the generated-doc diff approach in this PR or a follow-up let me know which you prefer.

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,181 @@
...
+    )[0]
+    block = re.sub(r"<br\s*/?>", "\n", block, flags=re.IGNORECASE)
+    block = re.sub(r"</?li>", "\n", block, flags=re.IGNORECASE)
+    block = re.sub(r"</?ul>", "", block, flags=re.IGNORECASE)
+    block = re.sub(r"</?[^>]+>", "", block)
+    block = block.replace("`", "")
+    block = re.sub(r"\*\*", "", block)
+    block = re.sub(r"\$([^$]+)\$", r"\1", block)
+    block = block.replace("~", "")
+    block = re.sub(r"\s*-\s+", " ", block)
```
huitseeker: This turns `(end_addr - start_addr) / 4` into `(end_addr start_addr) / 4`, so a user doc could lose the `-` and still compare equal to the generated doc. Anchoring this to Markdown list markers, like `^\s*-\s+` after the `<br>` split, would keep formulas checked.

crazywriter1: Thansk for the review. I tightened the normalization so `-` is only stripped when it is acting as a Markdown list marker (`(?m)^\s*-\s+`). That keeps formulas like `(end_addr - start_addr) / 4` intact while still normalizing generated list formatting.

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,181 @@
...
+        expected = case["expected"]
+
+        content = doc_path.read_text(encoding="utf-8")
+        if marker not in content:
+            errors.append(f"{doc_path}: missing marker {marker!r}")
+            continue
+
+        window_start = content.index(marker)
+        window = content[window_start : window_start + 500]
+        if expected not in window:
```
huitseeker: Could this compare the marked row's cycle value exactly? Right now `expected in window` can pass when the doc says `138 cycles`, because `38 cycles` is still a substring. That would let this new guard miss the same kind of drift it is meant to catch for assembly docs.

crazywriter1: Addressed in the latest push: the assembly cycle check now compares against the marked table row only, using a digit-bounded exact match (`(?<!\d)...(?!\d)`), so values like `138 cycles` no longer satisfy an expected `38 cycles`.

docs/src/user_docs/core_lib/crypto/aead.md
```diff
@@ -64,7 +64,7 @@ The padding block is automatically added and encrypted. The tag is stored right
 - Blocks must be stored contiguously in memory
 - `src_ptr` and `dst_ptr` **must be different** (in-place encryption not supported)
 
-**Cycles:** ~77 + 2 * n, where n = number of field elements encrypted (includes the final padding block)
+**Cycles:** 77 + 2 * n, where n = number of field elements encrypted (includes the final padding block). For num_blocks data blocks: n = 8 * (num_blocks + 1).
```
huitseeker: The source MASM docs for `encrypt` and `decrypt` still mark these as `Cycles (estimate):` (see `crates/lib/core/asm/crypto/aead.masm`), but here the user docs now read `**Cycles:**` without any estimate qualifier. The old text used `~77 + 2 * n`, which signaled the approximation, so this edit quietly upgrades an estimate to an exact-looking value.

The new checker does not catch this because `extract_cycles_from_description` strips `(estimate)` via `Cycles(?:\s*\(estimate\))?`, so both forms normalize to the same string. Could the user docs keep the `~` or an `(estimate)` note, and could the checker preserve whether a count is estimated so this kind of drift is caught too?

crazywriter1:
Addressed: restored `~` on AEAD encrypt/decrypt cycle counts, and updated `extract_cycles_from_description` so `Cycles (estimate)` / `~` are preserved in the normalized comparison (dropping the estimate marker no longer compares equal). @huitseeker

docs/src/user_docs/assembly/instruction_reference.md
crazywriter1: Addressed earlier: instruction_reference.md now has a cycle-check marker and a second fixture entry (u32popcnt-instruction-reference).

processor/src/tests/mod.rs
```diff
@@ -1451,3 +1456,117 @@ fn test_assert_message_without_debug_info_reports_error_code() {
...
+        );
+    }
+}
+
+struct AssemblyCycleFixture {
+    id: String,
+    program: String,
+    baseline_program: Option<String>,
+    stack_inputs: Option<Vec<u64>>,
+    expected_cycles: u8,
```
huitseeker: Nit: The `u8` here (and on `measure_program_cycles` at line 1538, plus the `u8::try_from` at line 1571) caps measurable programs at 255 cycles. That is fine for `u32popcnt` at 38, but a future fixture for something like `truncate_stack` (`17 + 11 * overflow_words`) or `hash_elements` would blow past 255 and panic with a confusing "measured cycle count should fit in u8" message instead of a clean assertion failure. Would `u32` or `usize` be safer in this test struct so the cap does not bite later?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,193 @@
...
+
+        marker_at = content.index(marker)
+        # Compare against the marked table row only (not a loose substring window).
+        row_start = content.rfind("\n", 0, marker_at) + 1
+        row_end = content.find("\n", marker_at)
+        if row_end == -1:
+            row_end = len(content)
+        row = content[row_start:row_end]
+        # Reject larger numbers that contain the expected digits (e.g. 138 vs 38).
+        if re.search(rf"(?<!\d){re.escape(expected)}(?!\d)", row) is None:
```
huitseeker: Nit: The marker check scans the whole table row for the expected text. 

That works for `u32popcnt` today because the digit-boundary anchor `(?<!\d)...(?!\d)` rejects `2^{32}` as a match for `38`. But if a future fixture points at a row whose description column also contains the same digit (for example an expected `40` against a row that mentions `2^{40}`), the check could pass even if the cycle column drifted. 

Pulling out just the cycle column of the marked row, or reusing `extract_cycles_from_description` on the row, would make this tighter as the fixture list grows.

scripts/user-doc-cycle-mappings.toml
```diff
@@ -0,0 +1,208 @@
+# Maps user-facing core library docs to generated MASM docs (source of truth for cycles).
+# See scripts/check_user_doc_cycles.py and issue #3111.
+
+[[entry]]
+user_doc = "docs/src/user_docs/core_lib/collections.md"
+section = "Merkle Mountain Range"
+procedure = "add"
+generated_doc = "crates/lib/core/docs/collections/mmr.md"
+generated_procedure = "add"
```
huitseeker: All 31 entries in this file set `generated_procedure` to the same value as `procedure` (none of them rename the procedure on the generated side). If that holds, the field is redundant and `check_user_doc_cycles.py` could just reuse `procedure` at lines 124, 129, and 133. That would drop 31 lines and one concept from the mapping table. Happy to leave it if you anticipate a generated-side rename soon, but wanted to flag it since nothing currently exercises the split.

processor/src/tests/mod.rs
huitseeker: - The `AssemblyCycleFixture` struct and `load_assembly_cycle_fixtures` loader (about 60 lines) exist to support `baseline_program: Option` and `stack_inputs: Option`, but the two fixtures in `assembly-cycle-fixtures.toml` always set `baseline_program` and never set `stack_inputs`. 
- They also run the identical `program`. 
- If those optionality fields stay unused, inlining two `measure_program_cycles(...)` calls in the test body (and reading expected/baseline directly) would let the struct, the loader, the `include_str!`, and the `toml` dev-dep in `processor/Cargo.toml` go. 
- The `begin` branch at line 1554 is also unexercised. 

Fine to keep the shape if you expect a second fixture shape soon, just flagging the current cost.

processor/src/tests/mod.rs
```diff
@@ -1451,3 +1456,117 @@ fn test_assert_message_without_debug_info_reports_error_code() {
...
+}
+
+fn load_assembly_cycle_fixtures() -> Vec<AssemblyCycleFixture> {
+    const FIXTURES: &str = include_str!("../../../scripts/assembly-cycle-fixtures.toml");
```
huitseeker: This test reads `../../../scripts/assembly-cycle-fixtures.toml` from inside `processor/src/tests`, but that file lives outside the `miden-processor` crate. If someone runs the packaged crate tests from crates.io, the repo-level `scripts/` directory will not be there, so this `include_str!` path will fail to compile.

Could we move the fixture under `processor/` or another packaged path, and have the script read that same copy? That keeps the workspace check and packaged tests using the same fixture data.

huitseeker (changes requested):
Thanks, this is much closer now.

  I think there are two remaining issues to fix before merge.

The first issue is in the assembly cycle check. It still proves that the expected text appears somewhere in the marked row. It does not prove that the cycle value itself matches. A later row can pass for the wrong reason if the same number appears in the note text. The fix is to compare the cycle cell, not the whole row.

The second issue is the fixture path in the processor test. The test includes a file from `../../../scripts`, which is outside the `miden-processor` crate. That works in this workspace, but it can fail when the crate is packaged and tested by itself. The fixture should live under a path that is packaged with the crate, and the script should read that same file.

The other comments are cleanup. I would not block on them.

crazywriter1:
> Thanks, this is much closer now.
> 
> I think there are two remaining issues to fix before merge.
> 
> The first issue is in the assembly cycle check. It still proves that the expected text appears somewhere in the marked row. It does not prove that the cycle value itself matches. A later row can pass for the wrong reason if the same number appears in the note text. The fix is to compare the cycle cell, not the whole row.
> 
> The second issue is the fixture path in the processor test. The test includes a file from `../../../scripts`, which is outside the `miden-processor` crate. That works in this workspace, but it can fail when the crate is packaged and tested by itself. The fixture should live under a path that is packaged with the crate, and the script should read that same file.
> 
> The other comments are cleanup. I would not block on them.

Addressed all remaining review points in the latest push:

- The assembly documentation check now extracts and compares only the cycle cell, rather than scanning the full table row.
- Moved `assembly-cycle-fixtures.toml` under `processor/src/tests/`, so it is packaged with `miden-processor`; both the processor test and Python checker use this copy.
- Changed measured and expected cycle counts from `u8` to `u32`.
- Removed the redundant `generated_procedure` field from all mappings.
- Simplified the fixture loader by removing unused optional `stack_inputs` / baseline handling and the unexercised `begin` branch.

The documentation cycle checker and processor assembly fixture test both pass. @huitseeker

crazywriter1:
@huitseeker @Al-Kindi-0 Friendly ping for another look.

The remaining review items from July 15 are addressed in the latest commits:
- assembly check compares the cycle **cell**, not the full row
- `assembly-cycle-fixtures.toml` lives under `processor/src/tests/` (shared by the processor test and the Python checker)
- cycle counts use `u32`; redundant `generated_procedure` removed; fixture loader simplified

Also rebased/merged with current `next` (MMR cycle docs updated for the upstream formulas). `make check-user-doc-cycles` and the processor assembly fixture test both pass locally.

Happy to address anything else that comes up.

docs/src/user_docs/core_lib/collections.md
```diff
@@ -49,6 +49,6 @@ The following procedures are available:
...
+| find_word          | Finds a value in a sorted array of words.<br/><br/>**Inputs:** `[VALUE, start_ptr, end_ptr, ...]`<br/>**Outputs:** `[is_value_found, value_ptr, start_ptr, end_ptr, ...]`<br/><br/>Where `VALUE` is the word to search for, `start_ptr` and `end_ptr` define the array bounds (word-aligned), `is_value_found` is 1 if found and 0 otherwise, and `value_ptr` points to the found value or the insertion point.<br/><br/>**Requirements:**<br/>- Words must be sorted in non-decreasing order<br/>- `start_ptr` and `end_ptr` must be word-aligned<br/>- `start_ptr <= end_ptr`<br/><br/>**Cycles:**<br/>Value exists: 46 cycles<br/>Value doesn't exist and the array is empty: 25 cycles<br/>Value doesn't exist and is smaller than all elements: 151 cycles<br/>Value doesn't exist and is larger than all elements: 149 cycles<br/>Value doesn't exist: 286 cycles |
```
huitseeker: `find_word` still lists 46/25/151/149/286 cycles, while the generated source docs list 47/25/152/150/288. This makes `make check-user-doc-cycles` fail at this row. Could you update the row to the generated values?

docs/src/user_docs/core_lib/sys.md
```diff
@@ -6,7 +6,8 @@ sidebar_position: 7
...
+| ---------------------- | ------------- |
+| truncate_stack         | Removes elements deep in the stack until the depth of the stack is exactly 16. The elements are removed in such a way that the top 16 elements of the stack remain unchanged. If the stack would otherwise contain more than 16 elements at the end of execution, then adding a call to this function at the end will reduce the size of the public inputs that are shared with the verifier.<br/>Input: Stack with 16 or more elements.<br/> Output: Stack with only the original top 16 elements.<br/><br/>Cycles: `17 + 11 * overflow_words`, where `overflow_words` is the number of words needed to drop. |
+| drop_stack_top         | Drops the top 16 values from the stack.<br/><br/>Input: Stack with 16 or more elements.<br/>Output: Stack with the top 16 elements removed. |
+| log_precompile_request | Logs a precompile commitment and removes the helper words produced by the underlying `log_precompile` instruction.<br/><br/>Input: `[COMM, TAG, ...]`<br/>Output: `[...]` (top three helper words `R0`, `R1`, `CAP_NEXT` are dropped internally)<br/><br/>Cycles: 1 (plus cost of the underlying `log_precompile` instruction). |
```
huitseeker: `log_precompile_request` no longer exists in `miden::core::sys`. The generated module docs now export `build_proof_request_key` instead. Could you remove this stale row or replace it with the current procedure?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,222 @@
...
+MAPPINGS = Path(__file__).resolve().parent / "user-doc-cycle-mappings.toml"
+ASSEMBLY_FIXTURES = ROOT / "processor/src/tests/assembly-cycle-fixtures.toml"
+
+
+def extract_cycles_from_description(description: str) -> str:
+    description = html.unescape(description)
+    matches = list(
+        re.finditer(r"\bCycles(?:\s*\((estimate)\))?\s*:?\s*(.*)", description, re.DOTALL)
+    )
+    if not matches:
```
huitseeker, line 31: `extract_cycles_from_description` returns an empty string when `Cycles` is missing, so a mapped procedure passes when both generated and user docs lose their cycle text. Could this return an error instead, with a regression test for two cycle-free rows?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,222 @@
...
+    return text
+
+
+def extract_cycle_cell_from_row(row: str) -> str | None:
+    """Return the cycle cell text from a marked markdown table row.
+
+    Prefer a dedicated Cycles column (instruction_reference style). Fall back to
+    an embedded `*(N cycles)*` fragment in the instruction cell (u32_operations).
+    """
+    cells = [c.strip() for c in row.strip().strip("|").split("|")]
```
huitseeker, line 266: `extract_cycle_cell_from_row` returns the first numeric cell, so a row with an earlier numeric operand can pass when its cycle column drifts. Could this locate the `Cycles` column from the table header, with a test where the earlier cell matches `expected`?

processor/src/tests/mod.rs
```diff
@@ -1504,3 +1509,101 @@ fn test_assert_message_without_debug_info_reports_error_code() {
...
+                expected_cycles,
+            }
+        })
+        .collect()
+}
+
+fn measure_program_cycles(program: &str) -> u32 {
+    use miden_utils_testing::Test;
+
+    const TRUNCATE_STACK: &str = r"@locals(4)
```
huitseeker: nit: `TRUNCATE_STACK` duplicates `miden_utils_testing::TRUNCATE_STACK_PROC`, which this test crate already uses. Could this import the shared constant and remove the local MASM copy?

huitseeker (commented):
Sorry for the late review! This needs a rebase

scripts/check_user_doc_cycles.py
crazywriter1: Fixed in 75db88f47: `extract_cycles_from_description` now raises `MissingCycleTextError` when cycle text is missing (no more empty-string match). Added `test_missing_cycle_text_raises` — run with `python3 scripts/check_user_doc_cycles.py test`.

scripts/check_user_doc_cycles.py
crazywriter1: Fixed in 75db88f47: `extract_cycle_cell_from_row` now reads the `Cycles` column from the table header via `table_header_cells()`. Added `test_cycle_cell_uses_cycles_column_not_operand` where stack operands contain `2`/`3` but the cycle column is `38`.

crazywriter1:
@huitseeker Addressed the two open `check_user_doc_cycles.py` threads in the latest push. Happy to resolve those conversations if the fixes look good.

docs/src/user_docs/core_lib/collections.md
```diff
@@ -19,9 +19,9 @@ The following procedures are available to read data from and make updates to a M
...
-| num_leaves_to_num_peaks      | Given the number of leaves in an MMR, computes the number of peaks (i.e. the number of set bits in `num_leaves`).<br /><br />Inputs: `[num_leaves, ...]`<br />Outputs: `[num_peaks, ...]`<br /><br />Cycles: 67 |
+| pack                         | Computes a commitment to the given MMR and copies the MMR to the Advice Map using the commitment as a key. Inputs: `[mmr_ptr, ...]`. Outputs: `[HASH, ...]`.<br /><br />Cycles: `198 + 6 * extra_peak_pair cycles` where `extra_peak_pair` is the number of peak pairs in addition to the first 16 peaks, i.e. `round_up((num_of_peaks - 16) / 2)` |
```
huitseeker: `num_peaks_to_message_size` clamps counts below 16 up to 16, so `extra_peak_pair` is zero there. This formula becomes negative for common small MMRs and reports the wrong cycle total. Could you clamp it in both MASM comments, regenerate the docs, and sync both rows?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,293 @@
...
+    if "~" in block:
+        is_estimate = True
+    block = block.replace("~", "")
+    block = re.sub(r"(?m)^\s*-\s+", "", block)
+    block = re.sub(r"where:\s*", "where ", block, flags=re.IGNORECASE)
+    block = re.sub(r"[,.\:;]", " ", block)
+    block = re.sub(r"\s+", " ", block.lower()).strip()
+    if is_estimate:
+        return f"estimate {block}"
+    return block
```
huitseeker, line 62: `extract_cycles_from_description("Cycles:")` still returns `""`, so two empty cycle labels compare equal and the check passes without a value. Could empty normalized text raise `MissingCycleTextError` too, with a regression test for empty and whitespace-only values?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,293 @@
...
+
+    header_cells = None
+    if content is not None and row_start is not None:
+        header_cells = table_header_cells(content, row_start)
+
+    if header_cells:
+        for idx, name in enumerate(header_cells):
+            if "cycle" in name and idx < len(cells):
+                text = _normalize_table_cell(cells[idx])
+                if text:
```
huitseeker: `extract_cycle_cell_from_row` trusts the named column only when it is nonempty. A blank `Cycles` cell then falls back to `*(38 cycles)*` in another cell and passes. Could a present cycle column be authoritative, returning `None` when blank, with a regression test?

scripts/check-user-doc-cycles.sh
```diff
@@ -0,0 +1,27 @@
...
+elif command -v python3.11 >/dev/null 2>&1; then
+    PYTHON=python3.11
+else
+    PYTHON=python3
+fi
+if ! "$PYTHON" -c 'import tomllib' 2>/dev/null; then
+    echo "error: Python 3.11+ required (tomllib). Use python3.11 or python3.12." >&2
+    exit 1
+fi
+"$PYTHON" scripts/check_user_doc_cycles.py
```
huitseeker, line 24: The three regression tests in `CheckUserDocCyclesTests` never run here because the script only enters `unittest.main()` with the `test` argument, and no other workflow passes it. Could this target run `"$PYTHON" scripts/check_user_doc_cycles.py test` before the main check?

CHANGELOG.md
```diff
@@ -299,6 +299,7 @@
 
 - Added a Blake3 pure execution benchmark axis and reduced processor benchmark compile time by relaxing forced inlining in execution helpers ([#3289](https://github.com/0xMiden/miden-vm/pull/3289)).
 - Documented that `smt::peek` is a fast, untrusted advice lookup, and that caller code must verify the returned value before relying on it ([#3297](https://github.com/0xMiden/miden-vm/pull/3297)).
+- Added `scripts/check-user-doc-cycles.sh` to verify user-facing core library and assembly instruction cycle counts against generated MASM docs and measured `clk` fixtures ([#3111](https://github.com/0xMiden/miden-vm/issues/3111)).
```
huitseeker: This entry is still under the released `miden-vm v0.25.0` section, while the current unreleased section is `v0.32.0`. Could you move it to the `v0.32.0` Changes section so the published `v0.25.0` record stays unchanged?

scripts/user-doc-cycle-mappings.toml
```diff
@@ -0,0 +1,172 @@
...
+section = "encrypt"
+procedure = "encrypt"
+generated_doc = "crates/lib/core/docs/crypto/aead.md"
+
+[[entry]]
+user_doc = "docs/src/user_docs/core_lib/crypto/aead.md"
+section = "decrypt"
+procedure = "decrypt"
+generated_doc = "crates/lib/core/docs/crypto/aead.md"
+
```
huitseeker: nit: `git diff --check origin/next...HEAD` still reports this extra blank line at EOF. Could you remove it?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,293 @@
...
+                return [
+                    c.strip().lower()
+                    for c in header.strip().strip("|").split("|")
+                ]
+    return None
+
+
+def _normalize_cycle_cell_text(text: str) -> str:
+    text = text.lower()
+    if re.fullmatch(r"\d+(?: cycles?)?", text, re.IGNORECASE):
```
huitseeker: nit: The two full-match branches return the same `text` as the final fallback, and line 213 computes another normalized value that is never read. Could these five dead lines be removed?

huitseeker (changes requested):
Thanks! This PR introduces an enforcement chec, but it still has two false-success paths defeat that check’s  purpose, and CI does not run the tests meant to protect those cases. The incorrect MMR formula and changelog placement are separate documentation errors.

crazywriter1:
Addressed the latest review:

- Empty / whitespace-only cycle text now raises `MissingCycleTextError` (with regression test)
- Blank `Cycles` column is authoritative (no fallback; with regression test)
- Shell target runs `"$PYTHON" scripts/check_user_doc_cycles.py test` before the main check
- Clamped the MMR `extra_peak_pair` formula in MASM + docs
- Moved the CHANGELOG entry to `v0.32.0` Changes
- Removed the mappings.toml EOF blank line and the dead normalize branches

Also merged latest `next` to clear the CHANGELOG conflict.

docs/src/user_docs/core_lib/collections.md
```diff
@@ -19,9 +19,9 @@ The following procedures are available to read data from and make updates to a M
...
+| unpack                       | Writes the MMR whose preimage hashes to `HASH` to the memory location pointed to by `mmr_ptr`. Inputs: `[HASH, mmr_ptr, ...]`. Outputs: `[...]`. Where: `HASH` is the hash of `[num_leaves, 0, 0, 0] \|\| padded_peaks`. The advice map must contain a key with `HASH`, and its value is `[num_leaves, 0, 0, 0] \|\| padded_peaks`. `mmr_ptr` is the memory location where the MMR data will be written, starting with the MMR forest (the total count of its leaves) followed by its peaks. The memory location must be word-aligned.<br /><br />Cycles: `215 + 9 * extra_peak_pair cycles` where `extra_peak_pair` is the number of peak pairs beyond the first 16 peaks (zero when `num_of_peaks <= 16`), i.e. `round_up(max(num_of_peaks - 16, 0) / 2)` |
+| num_leaves_to_num_peaks      | Given the number of leaves in an MMR, computes the number of peaks (i.e. the number of set bits in `num_leaves`).<br /><br />Inputs: `[num_leaves, ...]`<br />Outputs: `[num_peaks, ...]`<br /><br />Cycles: 32 |
```
huitseeker: `num_leaves_to_num_peaks` now contains only `u32popcnt`, which this PR measures at 38 cycles, but its MASM annotation and this row still say 32. The checker passes because both copies agree. Could you measure the wrapper and update all three copies?

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,311 @@
...
+    content = path.read_text(encoding="utf-8")
+    for line in content.splitlines():
+        description = extract_table_row_description(line, procedure)
+        if description is not None:
+            return extract_cycles_from_description(description)
+
+    raise KeyError(f"generated procedure not found: {procedure!r} in {path}")
+
+
+def check_core_lib_mappings() -> list[str]:
```
huitseeker, line 154: `check_core_lib_mappings` proves that two text copies agree, but when `num_leaves_to_num_peaks` is wrong in both copies, this check passes. Could every guarded cycle count require a measured fixture or an explicit exemption?

crazywriter1: Done in 9ef9ac0.

`check_core_lib_mappings` now requires exactly one of `fixture_id` (measured case in `assembly-cycle-fixtures.toml`) or a non-empty `exemption` on every mapping. `num_leaves_to_num_peaks` is wired to fixture `num_leaves_to_num_peaks` (body is `u32popcnt`, measured 38) and the MASM/generated/user docs were updated to 38. Formula / multi-case / estimate entries use explicit exemptions.

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,413 @@
...
+    if description.endswith("|"):
+        description = description[:-1].rstrip()
+    return description
+
+
+def extract_user_procedure_cycles(content: str, section: str | None, procedure: str) -> str:
+    scoped = slice_section(content, section)
+
+    subsection = section or procedure
+    if re.search(rf"^###\s+{re.escape(subsection)}\s*$", content, re.MULTILINE):
```
huitseeker: `scoped` is computed above, but subsection detection still searches all of `content`, and this loop accepts the first line that mentions `cycles` rather than the `Cycles:` field. A same-named heading in another section, or prose that mentions cycles before the field, can make the check read the wrong text. Could this search the scoped subsection and require the actual cycle label?

CHANGELOG.md
```diff
@@ -73,6 +74,7 @@
 #### Changes
 
 - Cached loaded MAST forests in `FastProcessor` so repeated external calls reuse the forest and merge its advice map once ([#3764](https://github.com/0xMiden/miden-vm/pull/3764)).
+- Added `scripts/check-user-doc-cycles.sh` to verify user-facing core library and assembly instruction cycle counts against generated MASM docs and measured `clk` fixtures ([#3111](https://github.com/0xMiden/miden-vm/issues/3111)).
```
huitseeker: This entry is duplicated at line 11 under v0.34.0 (Unreleased). This copy places the change in v0.32.0, which shipped before this PR merged. Please remove this copy and keep the unreleased entry.

scripts/check_user_doc_cycles.py
```diff
@@ -0,0 +1,413 @@
...
+        for idx, name in enumerate(header_cells):
+            if "cycle" in name:
+                if idx >= len(cells):
+                    return None
+                text = _normalize_table_cell(cells[idx])
+                if not text:
+                    return None
+                return _normalize_cycle_cell_text(text)
+
+    for cell in cells:
```
huitseeker: This fallback still scans every cell. If the instruction loses its cycle annotation but the notes cell contains `previously *(38 cycles)*`, `extract_cycle_cell_from_row()` returns `38 cycles` and the check passes. Please restrict the fallback to the instruction cell and add a regression test where only the notes contain the expected count.

processor/src/tests/assembly-cycle-fixtures.toml
```diff
@@ -0,0 +1,31 @@
...
+marker = "u32popcnt"
+program = "clk push.7 u32popcnt drop clk swap sub"
+baseline_program = "clk push.7 drop clk swap sub"
+expected = "38"
+
+# Core-lib `mmr::num_leaves_to_num_peaks` body is solely `u32popcnt`.
+# No assembly doc marker: used by scripts/user-doc-cycle-mappings.toml.
+[[case]]
+id = "num_leaves_to_num_peaks"
+program = "clk push.7 u32popcnt drop clk swap sub"
```
huitseeker: This fixture does not measure the procedure whose documentation it backs. It runs a copied `u32popcnt` snippet, so a later change to `mmr::num_leaves_to_num_peaks` can make the procedure cost diverge while this fixture and both documentation copies remain at 38. Please measure `num_leaves_to_num_peaks` through the core-library test harness and connect that result to the documentation expectation.

huitseeker (changes requested):

PR #3264, opened 2026-06-18
fix: Bound advice Merkle store growth

huitseeker:
`AdviceProvider` could keep adding internal Merkle store nodes without a runtime limit. A program could grow the store until the host ran out of memory.
This adds an execution option for the maximum Merkle store node count. AdviceProvider now rejects initial advice and live Merkle store mutations that would exceed that limit before it changes the stored nodes.

Closes 0xMiden/security-issues#12.

### DIFF
--- a/processor/src/execution_options.rs
+++ b/processor/src/execution_options.rs
@@ -23,6 +23,8 @@ pub struct ExecutionOptions {
     /// Maximum number of continuations allowed on the continuation stack at any point during
     /// execution.
     max_num_continuations: usize,
+    /// Maximum number of internal nodes allowed in the advice provider's Merkle store.
+    max_merkle_store_nodes: usize,
     /// Maximum number of deferred precompile requests allowed during execution.
     max_precompile_requests: usize,
     /// Maximum total number of calldata bytes allowed across deferred precompile requests.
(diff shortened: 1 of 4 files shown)


processor/src/execution_options.rs
adr1anh, line 43: With all of these new bounds, would it start making sense to group them in a common struct?

adr1anh (approved):
Minor comment but this looks good to me

processor/src/host/advice/mod.rs
```diff
@@ -595,11 +632,17 @@ impl AdviceProvider {
     }
 
     /// Extends the [MerkleStore] with the given nodes.
-    pub fn extend_merkle_store<I>(&mut self, iter: I)
+    pub fn extend_merkle_store<I>(&mut self, iter: I) -> Result<(), AdviceError>
     where
         I: IntoIterator<Item = InnerNodeInfo>,
     {
-        self.store.extend(iter);
+        let mut store = self.store.clone();
```
plafer: Do we *need* to clone the store here? This will probably slow down the `MRUPDATE` operation quite a bit.

plafer (approved):
Just a question about whether we need to always clone the store; otherwise LGTM

processor/src/host/advice/mod.rs
huitseeker: Good catch ! Fixed in latest.

processor/src/execution_options.rs
bobbinth: I think the goal is to get rid of all of these in favor of a single setting that puts a bound on the overall advice provider size. The user doesn't really care where the consumption is coming from - they just want to make sure a malicious program doesn't bring down the system.

PR #3266, opened 2026-06-18
refactor: chiplets columns layout

### LINKED ISSUE #3165: Refactor chiplets column layout
Follow up of #3115:

> For a more general refactor, I'd move them [`s_perm`/`chip_clk`] to the front of the array (unfortunately having to modify all the column indices), rename `s_perm -> s_00` and `s0 -> s01`.

_Originally posted by @adr1anh in https://github.com/0xMiden/miden-vm/pull/3115#discussion_r3234223123_

Nashtare:
## Describe your changes

closes #3165 

I'm not super convinced about the clarity this brings though

- Updated `CHANGELOG.md'


processor/src/trace/utils.rs
```diff
@@ -120,11 +196,20 @@ impl<'a> ChipletTraceFragment<'a> {
...
+    /// Returns `None` when this fragment has no prefix space (`col_start < DATA_COL_START`),
+    /// i.e. the band starts inside the data columns.
+    pub fn prefix_mut(&mut self, row: usize) -> Option<&mut ChipletPrefix> {
+        if self.col_start < DATA_COL_START {
+            return None;
+        }
         let row_start = row * self.stride;
-        &mut self.band[row_start..row_start + self.col_start]
+        let cells = &mut self.band[row_start..row_start + DATA_COL_START];
+        let (head, view, tail) = unsafe { cells.align_to_mut::<ChipletPrefix>() };
```
huitseeker: Do we need the typed prefix view here? It looks like the only caller uses this to set `s_01`, so this adds an `unsafe` cast and a public helper struct for one column write. A small `set_s_01(row)` helper, or even a named column constant, would be easier to audit.

huitseeker (approved):
LGTM

adr1anh:
This looks good, I think the uniformity helps understand the chiplet selectors layout. I do agree, though that this might not bring as much clarity as we thought it would I think there might be a way for us to adapt the documentation so that it refers to computed flags expressions instead of the raw selectors. We could still define something like `f_perm=s_0` `f_ctrl=1-s_0` and similarly for the other chiplets. The only place we would refer the raw `s` selectors would be when we map them to the `ChipletsFlags`. 

I’m fine with this PR as it is though, but happy to discuss this more asynchronous if you think it's worth going further in this direction

Nashtare:
@adr1anh I added some light clarification inline for the mental map.

adr1anh (approved):
Thanks for the changes, LGTM

PR #3268, opened 2026-06-18
Remove stripped MastForest serialization mode

### LINKED ISSUE #3263: MastForest and DebugInfo Serialization Cleanup
PR #3221 moved debug metadata out of `MastForest` and into package debug sections, but the forest-level STRIPPED flag and its associated APIs still imply a forest can be stripped of debug info. That implication is now wrong. A package may have debug sections, but a MastForest should hold execution data only.

 What changes

- MastForest serialization describes execution data only. Normal forest bytes carry executable MAST data, roots, advice data, and node hashes needed by trusted readers. Hashless forest bytes omit rebuildable internal node hashes for untrusted validation. Neither mode talks about debug stripping.
- Remove MastForest::write_stripped, MastForest::stripped_size_hint, MastForestWireView::is_stripped, and the FLAG_STRIPPED constant. The normal forest writer is Serializable::write_into / to_bytes. The hashless writer name stays as-is.
- Package serialization owns debug section behavior. Trusted package readers preserve debug sections. Untrusted package readers validate the embedded forest and drop debug sections. Package::debug_info, Package::strip_debug_info, and Package::without_debug_info stay at the package layer.
- HASHLESS keeps its current bit value. Bit 0 becomes reserved. No flag compaction before release.
- No new MAST or package format versions beyond what this branch already uses ([0,0,4] and [6,0,0]).

huitseeker:
Closes #3263.

PR #3221 moved debug info out of `MastForest` and into package debug sections. After that change, the forest level `STRIPPED` flag and APIs still made `MastForest` look like it could carry debug info. It cannot. A `MastForest` now stores execution data only.

This PR removes the stripped forest mode. Normal forest serialization now uses `Serializable::write_into` and `to_bytes` with flag `0x00`. Hashless serialization keeps flag `0x02`. Bit 0 is now reserved.

This also removes `MastForest::write_stripped`, `MastForest::stripped_size_hint`, `MastForestWireView::is_stripped`, and the old size hint code. Tests and callers now use normal serialization or hashless serialization directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - [BREAKING] Moved debug info ownership out of `MastForest` and into package debug sections, adding source-node debug metadata that preserves distinct source occurrences after MAST node deduplication ([#3221](https://github.com/0xMiden/miden-vm/pull/3221)).
 - Cleaned up processor error handling for diagnostics, malformed MAST loading, and binary-value checks ([#3230](https://github.com/0xMiden/miden-vm/pull/3230)).
 - [BREAKING] Bump Plonky3 related dependencies to fix NEON arithmetic bug ([#3272](https://github.com/0xMiden/miden-vm/pull/3272)).
+- [BREAKING] Removed the stripped `MastForest` serialization mode. Normal forest bytes now describe execution data only ([#3268](https://github.com/0xMiden/miden-vm/pull/3268)).
 
 #### Fixes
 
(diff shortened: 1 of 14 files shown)


bitwalker (approved):
LGTM, though my review is somewhat cursory here as I'm less familiar with the MAST serialization internals

plafer (approved):
LGTM

PR #3269, opened 2026-06-18
fix(debug): preserve struct names in debug types

djolertrk:
This preserves concrete struct metadata when emitting VM debug types. Previously, structs that reached debug info without a declared `TypeExpr` fell back to `<anon>` and numeric fields; now the emitter uses the resolved `StructType` name and `field` names when available. This fixes debug dumps for SDK/WIT record types, e.g. `miden:base/core-types@1.0.0/felt`, which now renders by name instead of as `<anon>`.

Useful for debugging tools.

Resolves this:  https://github.com/0xMiden/compiler/issues/1198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.23.4 (TBD)
+
+#### Fixes
+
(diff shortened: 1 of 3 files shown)


huitseeker (approved):
Works for me!

djolertrk:
cc @bitwalker

bitwalker (approved):

PR #3273, opened 2026-06-19
Reject private types in exported MASM surfaces

### LINKED ISSUE #3092: Reject private type declarations in exported procedure signatures

huitseeker:
This rejects private type references in public MASM APIs.

An exported procedure signature or public type could name a private local type. The private type could also be reached through an alias, an enum payload, a same-module absolute path, or a public pub use.

This PR checks public procedure signatures and public type declarations during semantic analysis. It also checks public item re-exports after linking.

Fixes #3092. Obsoletes #3154

### DIFF
--- a/crates/assembly-syntax/src/sema/mod.rs
+++ b/crates/assembly-syntax/src/sema/mod.rs
@@ -329,12 +329,7 @@ fn verify_exported_signature_type_expr(
         },
         TypeExpr::Struct(ty) => {
             for field in ty.fields.iter() {
-                verify_exported_signature_type_expr(
-                    module,
-                    analyzer,
-                    &field.ty,
-                    visiting_types,
-                );
+                verify_exported_signature_type_expr(module, analyzer, &field.ty, visiting_types);
             }
         },
         TypeExpr::Ref(path) => {
@@ -380,12 +375,9 @@ fn verify_exported_signature_type_expr(
             }
 
             match type_decl {
-                TypeDecl::Alias(alias) => verify_exported_signature_type_expr(
-                    module,
-                    analyzer,
-                    &alias.ty,
-                    visiting_types,
-                ),
+                TypeDecl::Alias(alias) => {
+                    verify_exported_signature_type_expr(module, analyzer, &alias.ty, visiting_types)
+                },
                 TypeDecl::Enum(ty) => {
                     for variant in ty.variants() {
                         if let Some(payload_ty) = variant.value_ty.as_ref() {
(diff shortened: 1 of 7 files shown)


crates/assembly/src/assembler.rs
```diff
@@ -54,6 +57,25 @@ use crate::{
...
+            Self::ProcedureSignature => {
+                SemanticAnalysisError::PrivateTypeInExportedSignature { span, defined }
+            },
+            Self::TypeDeclaration => {
+                SemanticAnalysisError::PrivateTypeInExportedType { span, defined }
+            },
+        }
+    }
+}
+
```
plafer: This is copy/pasted from assembly-syntax/src/sema/mod.rs right? We should define it once and import

crates/assembly-syntax/src/sema/mod.rs
```diff
@@ -294,6 +296,157 @@ fn normalize_namespace_path(path: &Path) -> Result<Arc<Path>, PathError> {
...
+            };
+            let Some(export) = module.get(item) else {
+                return;
+            };
+            let Item::Type(type_decl) = export else {
+                return;
+            };
+
+            if !type_decl.visibility().is_public() {
+                analyzer.error(usage.private_type_error(path.span(), type_decl.name().span()));
```
plafer, line 416: Q: the semantic analyzer collects multiple errors, whereas the assembler stops after the first one (i.e. use of `?` in `verify_exported_type_expr()`, propagated up through `verify_exported_signature_type_visibility()`). Should we also accumulate errors at assembly time too? Maybe a follow-up if desired.

plafer (approved):
LGTM - left a coppy/paste improvement and question

crates/assembly-syntax/src/sema/mod.rs
bitwalker: It's trickier to accumulate errors during assembly than during semantic analysis, since errors during linking usually mean we can't really meaningfully proceed. That said, we could accumulate errors in general, and just immediately exit on fatal errors (i.e. we can't meaningfully proceed), but otherwise allow non-fatal errors (e.g. visibility violations) to accumulate and then exit with an error after linking is complete.

I'm planning to rework how diagnostics are handled during parsing/assembly in an upcoming release, but it's relatively low priority for the moment.

bitwalker (approved):

PR #3274, opened 2026-06-19
Ci:Blake3 criterion benchmark

### LINKED ISSUE #3281: Stabilize the Blake3 non-regression benchmark
The Blake3 1-to-1 non-regression benchmark currently runs one measurement per side in CI. That single run is noisy. A small change in runner load or scheduling can look like a regression, or hide one.

We opened #3274 to replace that one-shot measurement with a Criterion benchmark. It keeps the same CI role, but uses repeated samples and reports separate axes for execution, trace building, proving from a trace, and the end-to-end prove path.

This should make the CI comparison less sensitive to run-to-run variance, while keeping the step-level detail needed to update the README benchmark numbers and diagnose where a slowdown comes from.

huitseeker:
The Blake3 benchmark now uses Criterion instead of one manual prove run for measurement precision & stability.
  
CI still compares the pushed commit against its baseline. It still writes the report as a commit comment.

The benchmark reports four axes:
- execute_trace_inputs_sync
- build_trace
- prove_trace_sync
- e2e_prove

e2e_prove runs execution and trace generation for each sample, then measures only the prover span.
The harness proves and verifies the program once before timing proof axes. A proof that does not verify fails the benchmark.
The old Python runner is removed.
Validation matched the old benchmark at the merge base with `origin/next`. The new proof numbers were within the old 5 percent threshold.


benches/blake3-bench/src/bin/blake3-nonregression.rs
```diff
@@ -0,0 +1,844 @@
...
+    log_path: &Path,
+) -> Result<f64, Box<dyn std::error::Error>> {
+    command.current_dir(cwd);
+    for (key, value) in envs {
+        command.env(key, value);
+    }
+    let start = Instant::now();
+    let output = command.output()?;
+    let elapsed = duration_ms(start.elapsed());
+    let mut log = format!("$ {command:?}\n");
```
plafer: This ends up being reported as "wall time", but also includes building the benchmark (e.g. right after a `cargo clean`), right? Would probably then report as "build and run wall time", or somehow avoid including the build time in that measurement

plafer (approved):
LGTM - left a minor question/suggestion

benches/blake3-bench/src/bin/blake3-nonregression.rs
```diff
@@ -0,0 +1,844 @@
...
+        "prove_program_sync" => Some("e2e_prove_wall".to_string()),
+        _ => None,
+    }
+}
+
+fn metric_value(metric: &Metric) -> Option<f64> {
+    metric.mean_ms.or(metric.duration_ms)
+}
+
+fn compare_results(
```
Al-Kindi-0: The report collects multiple axes, but the actual regression decision is based only on `current.primary_metric`. With the default all, a large regression in `execute_trace_inputs_sync`, `build_trace`, or `prove_trace_sync` can be reported but still not fail CI if `e2e_prove` stays within threshold, right ?

README.md
```diff
@@ -92,6 +92,23 @@ A few general notes on performance:
...
+  --rayon-num-threads 16 \
+  --sample-size 10 \
+  --light-sample-size 100 \
+  --measurement-time-secs 1 \
+  --warm-up-time-secs 1 \
+  --bench-axes all \
+  --git-ref "$(git rev-parse HEAD)"
+```
+
+The result is written to `target/blake3-nonregression/result.json`; the harness does not parse the `miden-vm run` or `miden-vm prove` text output. The benchmark records `execute_trace_inputs_sync`, `build_trace`, `prove_trace_sync`, and `e2e_prove`. The `e2e_prove` metric runs execution and trace generation on each sample, but only measures the prover span. The harness also proves and verifies the program once before timing proof-heavy axes.
```
Al-Kindi-0, line 110: `e2e_prove` is a bit confusing, should we have something like `prove_with_built_trace` or something else to distinguish between pure proving and proving plus witness and trace generation ?

benches/blake3-bench/src/bin/blake3-nonregression.rs
```diff
@@ -0,0 +1,844 @@
...
+    cwd: &Path,
+    envs: &BTreeMap<String, String>,
+    log_path: &Path,
+) -> Result<f64, Box<dyn std::error::Error>> {
+    command.current_dir(cwd);
+    for (key, value) in envs {
+        command.env(key, value);
+    }
+    let start = Instant::now();
+    let output = command.output()?;
```
Al-Kindi-0: Would `Command::spawn` be better here ?

benches/blake3-bench/src/bin/blake3-nonregression.rs
Al-Kindi-0, line 1: This module is doing a lot of work, would it make sense to split this into small modules, for example:
- `runner`: command execution, env setup
- `criterion_results`: discovery/parsing
- `compare`: result comparison and regression logic
- `report`: Markdown and GitHub output

Al-Kindi-0 (approved):
Looks good
Left a few comments to improve things hopefully

benches/blake3-bench/src/bin/blake3-nonregression.rs
huitseeker: I kept `Command::output`. These commands are not interactive, and the runner needs the full stdout and stderr so it can write the log files and print the command output after the command exits. spawn would only help if we needed streamed output (It would add pipe handling here without changing the benchmark result).

README.md
huitseeker: I kept `e2e_prove`. The name means that each sample runs execution and trace generation before the measured prover span. The README says this directly. I did not use `prove_with_built_trace` because that sounds like the trace is built outside the sample, which is not true for this axis.

benches/blake3-bench/src/bin/blake3-nonregression.rs
huitseeker: Done!

benches/blake3-bench/src/bin/blake3-nonregression.rs
huitseeker: The compare step now marks the run as a regression if any shared measured axis crosses the threshold.

benches/blake3-bench/src/bin/blake3-nonregression.rs
huitseeker: Blake3 build time should now be out of metrics

Issue #1943, opened 2025-07-02
Merge `AdviceInputs` with `AdviceProvider`


plafer:
Context: https://github.com/0xMiden/miden-vm/pull/1936#discussion_r2180137526

bobbinth:
I would probably hold off on this for now. We may add other functionality to `AdviceProvider` (e.g., some kind of "recording" functionality to make building "replay" traces for the fast processor simpler).

bobbinth:
Closing this as in https://github.com/0xMiden/miden-vm/pull/1953 `AdviceInputs` and `AdviceProvider` are diverging quite a bit, and they are likely to diverge even more in the future.

PR #1395, opened 2024-07-15
feat(processor): expose the process api

bitwalker:
While doing some testing in the compiler where I wanted to set up some VM state and then execute a program on that state, I realized that there is no currently public API that allows one to access the state of the process to do so, short of using the `internals` feature flag of the crate.

Rather than requiring the use of that flag, this commit instead makes public the [Process] struct, but with all of its fields private unless the `internals` flag is enabled. This makes it possible to use the public APIs of the various traits implemented by `Process` to interact with it in a more fine-grained fashion.

It is still expected that most people will use the `execute` or `execute_iter` APIs, but this makes it possible to take control over that yourself when necessary/useful. I should also note that we were making the [Process] struct visible for ourselves under `#[cfg(test)]`, so it is clearly useful in certain cases (namely testing).

PR #3276, opened 2026-06-20
Debug info encoder decoder

marijamijailovic:
This PR adds `ast::types::TypedProcInfo` to `miden-assembly-syntax`: a typed view over procedure signatures that can print the signature, encode argument into stack felts, and decode result felts back. Callers pass the procedure name and its `FunctionType`; `word` and `felt` are built in, while other types such as `account-id` are provided by the caller via the `WitScalarCodec` trait. The `miden call` command (https://github.com/0xMiden/rust-sdk/pull/2179) uses this to print typed signatures and decode results instead of raw feltst. The module was moved from `miden-client` into `miden-assembly-syntax` since the types already live in `miden-vm` (https://github.com/0xMiden/rust-sdk/pull/2179#pullrequestreview-4479854775).

The relevant issue https://github.com/0xMiden/rust-sdk/issues/2098, and context and motivation: Main issue [#1395](https://github.com/0xMiden/miden-client/issues/1395), and see the comment in [#1943](https://github.com/0xMiden/miden-client/pull/1943#issuecomment-4244288925).

### DIFF
--- a/crates/assembly-syntax/src/ast/types/signatures/decode.rs
+++ b/crates/assembly-syntax/src/ast/types/signatures/decode.rs
@@ -7,8 +7,8 @@ use alloc::{
     vec::Vec,
 };
 
-use miden_assembly_syntax::ast::types::Type;
 use miden_core::Felt;
+use midenc_hir_type::Type;
 
 use super::{
     WitScalarCodec,
--- a/crates/assembly-syntax/src/ast/types/signatures/mod.rs
+++ b/crates/assembly-syntax/src/ast/types/signatures/mod.rs
@@ -1,12 +1,12 @@
 //! A typed view of a procedure signature.
 //!
-//! The manifest keeps the high-level (WIT) signature of each export; the VM stack only holds
-//! felts. [`TypedProcInfo`] joins the two: it prints the signature, turns argument text into stack
-//! felts, and result felts back into text.
+//! A procedure carries a high-level (WIT) signature; the VM stack only holds felts.
+//! [`TypedProcInfo`] joins the two: it prints the signature, turns argument text into stack felts,
+//! and result felts back into text.
 //!
 //! The felts follow the canonical ABI: one stack slot per leaf field, so `struct { u8, u8 }` is
 //! two felts. This is the layout of a component export, the kind a caller reaches with `call`, and
-//! the only convention this crate takes.
+//! the only convention this module takes.
 //!
 //! Types that take one token, like `word` and `felt`, go through a [`WitScalarCodec`]. Those two
 //! are built in; the caller adds others with [`TypedProcInfo::with_scalar_codec`].
@@ -19,8 +19,8 @@ use alloc::{
 };
 use core::fmt;
 
-use miden_assembly_syntax::ast::types::{CallConv, FunctionType, Type};
 use miden_core::Felt;
+use midenc_hir_type::{CallConv, FunctionType, Type};
 
 use self::{
     arity::{felt_count, token_count},
@@ -53,18 +53,17 @@ pub use self::{
 pub struct TypedProcInfo {
     /// The name we show, like `add-points`.
     name: String,
-    /// The signature from the package manifest.
+    /// The signature we encode arguments and decode results against.
     signature: FunctionType,
     /// Codecs the user added. We match them by type name when we encode and decode.
     codecs: Vec<Box<dyn WitScalarCodec>>,
 }
 
 impl TypedProcInfo {
-    /// Builds the typed view of the procedure `name`, with the signature its package manifest
-    /// carries.
+    /// Builds the typed view of the procedure `name`, with the signature `signature`.
     ///
-    /// The caller picks the export. It knows which one it wants, and it keeps the signature and
-    /// anything else it reads from that export, like its digest, on the same procedure.
+    /// The caller picks the procedure. It knows which one it wants, and it keeps the signature and
+    /// anything else it reads about it, like its digest, on the same procedure.
     ///
     /// The signature has to be `component-model`. Another convention puts the same value on the
     /// stack in another shape, and we would read and write felts the procedure never sees.
--- a/crates/mast-package/src/typed/encode.rs
+++ b/crates/mast-package/src/typed/encode.rs
@@ -1,0 +1,156 @@
+//! Turns argument tokens into the felts a procedure needs on the stack.
+
+use alloc::{boxed::Box, string::ToString, vec::Vec};
+use core::str::FromStr;
+
+use miden_assembly_syntax::ast::types::Type;
+use miden_core::Felt;
+
+use super::{
+    WitScalarCodec,
+    arity::{felt_count, max_for_bits, signed_range, slot_bits},
+    codec::{codec_for_struct, parse_felt_token},
+    errors::TypedError,
+};
+
+/// Encodes one value of `ty`. It reads exactly as many tokens as
+/// [`token_count`](super::arity::token_count) says, and makes as many felts as the value takes on
+/// the stack.
+pub(super) fn encode_type<I>(
+    tokens: &mut I,
+    ty: &Type,
+    codecs: &[Box<dyn WitScalarCodec>],
+) -> Result<Vec<Felt>, TypedError>
+where
+    I: Iterator,
+    I::Item: AsRef<str>,
+{
+    match ty {
+        Type::Struct(struct_ty) => {
+            if let Some(codec) = codec_for_struct(codecs, struct_ty) {
+                let token = tokens.next().ok_or(TypedError::TokenCountMismatch)?;
+                let felts = codec.encode(token.as_ref())?;
+                // A codec chooses the text form of its type, but not the width. A wrong width
+                // would put a value on the stack that the procedure cannot read. A type with no
+                // width on the stack gives us nothing to check against, so it stops here too,
+                // the same way it does when we decode.
+                let expected = felt_count(ty).ok_or(TypedError::UnsupportedType("struct"))?;
+                if felts.len() != expected {
+                    return Err(TypedError::CodecWidthMismatch {
+                        wit_name: codec.wit_name().to_string(),
+                        expected,
+                        actual: felts.len(),
+                    });
+                }
+                return Ok(felts);
+            }
+            let mut felts = Vec::new();
+            for field in struct_ty.fields() {
+                felts.extend(encode_type(tokens, &field.ty, codecs)?);
+            }
+            Ok(felts)
+        },
+        Type::Array(array_ty) => {
+            let mut felts = Vec::new();
+            for _ in 0..array_ty.len {
+                let chunk = encode_type(tokens, &array_ty.ty, codecs)?;
+                // An element of width zero reads no tokens and makes no felts. Without this
+                // check, a big `len` would loop and never move forward.
+                let empty = chunk.is_empty();
+                felts.extend(chunk);
+                if empty {
+                    break;
+                }
+            }
+            Ok(felts)
+        },
+        Type::List(_) => Err(TypedError::UnsupportedType("list")),
+        Type::Ptr(_) => Err(TypedError::UnsupportedType("pointer")),
+        Type::Function(_) => Err(TypedError::UnsupportedType("function")),
+        Type::Enum(_) => Err(TypedError::UnsupportedType("enum")),
+        Type::Unknown => Err(TypedError::UnsupportedType("unknown")),
+        Type::Never => Err(TypedError::UnsupportedType("never")),
+        primitive => {
+            let token = tokens.next().ok_or(TypedError::TokenCountMismatch)?;
+            encode_primitive(token.as_ref(), primitive)
+        },
+    }
+}
+
+fn encode_primitive(token: &str, ty: &Type) -> Result<Vec<Felt>, TypedError> {
+    let felts = ty.size_in_felts();
+    let bits = ty.size_in_bits() as u32;
+
+    match ty {
+        // The compiler turns WIT `felt` into a named struct, and `FeltCodec` handles that. This
+        // branch is for a plain field element in the signature. Both read the token the same way,
+        // but here we do not go through the codec.
+        Type::Felt => Ok(alloc::vec![parse_felt_token(token)?]),
+        Type::I1 => {
+            let v = match token.to_ascii_lowercase().as_str() {
+                "true" | "1" => 1,
+                "false" | "0" => 0,
+                _ => return Err(TypedError::InvalidBool(token.to_string())),
+            };
+            Ok(alloc::vec![Felt::from_u32(v)])
+        },
+        Type::I8 | Type::I16 | Type::I32 | Type::I64 | Type::I128 => {
+            let value = parse_signed(token, bits, slot_bits(felts), ty)?;
+            Ok(limbs_to_felts(value, felts))
+        },
+        Type::U8 | Type::U16 | Type::U32 | Type::U64 | Type::U128 => {
+            let value = parse_unsigned(token, bits, ty)?;
+            Ok(limbs_to_felts(value, felts))
+        },
+        Type::F64 => {
+            let value = parse_number::<f64>(token, ty)?;
+            Ok(limbs_to_felts(value.to_bits() as u128, felts))
+        },
+        // 256 bits do not fit our `u128` path. We cannot encode them yet.
+        Type::U256 => Err(TypedError::UnsupportedType("u256")),
+        _ => Err(TypedError::UnsupportedType("unsupported primitive")),
+    }
+}
+
+/// Cuts `value` into `n` limbs of 32 bits, one felt each. The low limb comes first, so `felts[0]`
+/// holds bits `0..32`. Every limb is a `u32`, so [`Felt::from_u32`] always works.
+fn limbs_to_felts(value: u128, n: usize) -> Vec<Felt> {
+    debug_assert!(n <= 4, "a value of more than four limbs does not fit the u128 path");
+    (0..n).map(|i| Felt::from_u32((value >> (32 * i)) as u32)).collect()
+}
+
+/// Reads `token` as `T`, naming the type from the signature in the error.
+fn parse_number<T: FromStr>(token: &str, ty: &Type) -> Result<T, TypedError> {
+    token.parse::<T>().map_err(|_| TypedError::InvalidNumber {
+        token: token.to_string(),
+        ty: ty.to_string(),
+    })
+}
+
+/// Reads an unsigned decimal number and checks that it fits in `bits` bits.
+fn parse_unsigned(token: &str, bits: u32, ty: &Type) -> Result<u128, TypedError> {
+    let value = parse_number::<u128>(token, ty)?;
+    if value > max_for_bits(bits) {
+        return Err(int_out_of_range(token, ty));
+    }
+    Ok(value)
+}
+
+/// Reads a signed decimal number and returns the bits that go on the stack. The value has to fit
+/// the `bits` of its type, but it is written sign-extended across the whole `slot`: `-1i8` is the
+/// 32-bit pattern of `-1`, not the byte `0xff`.
+fn parse_signed(token: &str, bits: u32, slot: u32, ty: &Type) -> Result<u128, TypedError> {
+    let signed = parse_number::<i128>(token, ty)?;
+    let (min, max) = signed_range(bits);
+    if signed < min || signed > max {
+        return Err(int_out_of_range(token, ty));
+    }
+    Ok((signed as u128) & max_for_bits(slot))
+}
+
+fn int_out_of_range(token: &str, ty: &Type) -> TypedError {
+    TypedError::IntOutOfRange {
+        token: token.to_string(),
+        ty: ty.to_string(),
+    }
+}
--- a/crates/mast-package/src/typed/encode.rs
+++ b/crates/mast-package/src/typed/encode.rs
@@ -8,7 +8,7 @@ use miden_core::Felt;
 
 use super::{
     WitScalarCodec,
-    arity::{felt_count, max_for_bits, signed_range, slot_bits},
+    arity::{felt_count, max_for_bits, signed_range, slot_bits, token_count},
     codec::{codec_for_struct, parse_felt_token},
     errors::TypedError,
 };
@@ -51,16 +51,12 @@ where
             Ok(felts)
         },
         Type::Array(array_ty) => {
+            if token_count(&array_ty.ty, codecs) == Some(0) {
+                return Ok(Vec::new());
+            }
             let mut felts = Vec::new();
             for _ in 0..array_ty.len {
-                let chunk = encode_type(tokens, &array_ty.ty, codecs)?;
-                // An element of width zero reads no tokens and makes no felts. Without this
-                // check, a big `len` would loop and never move forward.
-                let empty = chunk.is_empty();
-                felts.extend(chunk);
-                if empty {
-                    break;
-                }
+                felts.extend(encode_type(tokens, &array_ty.ty, codecs)?);
             }
             Ok(felts)
         },
--- a/crates/mast-package/src/typed/mod.rs
+++ b/crates/mast-package/src/typed/mod.rs
@@ -4,6 +4,10 @@
 //! felts. [`TypedProcInfo`] joins the two: it prints the signature, turns argument text into stack
 //! felts, and result felts back into text.
 //!
+//! The felts follow the canonical ABI: one stack slot per leaf field, so `struct { u8, u8 }` is
+//! two felts. This is the layout of a component export, the kind a caller reaches with `call`, and
+//! the only convention this crate takes.
+//!
 //! Types that take one token, like `word` and `felt`, go through a [`WitScalarCodec`]. Those two
 //! are built in; the caller adds others with [`TypedProcInfo::with_scalar_codec`].
 
@@ -15,7 +19,7 @@ use alloc::{
 };
 use core::fmt;
 
-use miden_assembly_syntax::ast::types::{FunctionType, Type};
+use miden_assembly_syntax::ast::types::{CallConv, FunctionType, Type};
 use miden_core::Felt;
 
 use self::{
@@ -62,16 +66,26 @@ impl TypedProcInfo {
     /// The caller picks the export. It knows which one it wants, and it keeps the signature and
     /// anything else it reads from that export, like its digest, on the same procedure.
     ///
+    /// The signature has to be `component-model`. Another convention puts the same value on the
+    /// stack in another shape, and we would read and write felts the procedure never sees.
+    ///
     /// The result has [`WordCodec`] and [`FeltCodec`]. Add more with [`Self::with_scalar_codec`].
-    pub fn new(name: impl Into<String>, signature: FunctionType) -> Self {
-        Self {
-            name: name.into(),
+    pub fn new(name: impl Into<String>, signature: FunctionType) -> Result<Self, TypedError> {
+        let name = name.into();
+        if signature.abi != CallConv::ComponentModel {
+            return Err(TypedError::UnsupportedCallConv {
+                procedure: name,
+                abi: signature.abi.to_string(),
+            });
+        }
+        Ok(Self {
+            name,
             signature,
             codecs: alloc::vec![
                 Box::new(WordCodec) as Box<dyn WitScalarCodec>,
                 Box::new(FeltCodec),
             ],
-        }
+        })
     }
 
     /// Adds a [`WitScalarCodec`]. The codec then handles its own WIT type, instead of the normal
@@ -100,6 +114,10 @@ impl TypedProcInfo {
     }
 
     /// Turns `tokens` into the felts the procedure needs on the stack, in parameter order.
+    ///
+    /// The first felt goes on top of the stack, because the first argument sits on top:
+    /// `f(1u32, 2u32)` gives `[1, 2]`. A caller that pushes them one by one starts at the back.
+    /// See `docs/external/src/appendix/calling-conventions.md` in the compiler.
     ///
     /// We check the count first. So if the caller passes the wrong number of arguments, the error
     /// gives the procedure name and both counts.
@@ -140,7 +158,8 @@ impl TypedProcInfo {
             .try_fold(0usize, |total, result| total.checked_add(felt_count(result)?))
     }
 
-    /// Turns the result felts at the start of `stack` into text.
+    /// Turns the result felts at the start of `stack` into text. `stack[0]` is the top of the
+    /// stack. The first result sits there, and so does the first field of a struct.
     ///
     /// `Ok(None)` means the procedure returns nothing. That is the only case that is not a
     /// problem. Everything else is an error the caller can show: a result type we cannot decode,
--- a/crates/mast-package/src/typed/mod.rs
+++ b/crates/mast-package/src/typed/mod.rs
@@ -20,7 +20,7 @@ use miden_core::Felt;
 
 use self::{
     arity::{felt_count, token_count},
-    codec::type_leaf_name,
+    codec::{StructShape, codec_for_struct, field_name, struct_shape, type_leaf_name},
     decode::decode_type,
     encode::encode_type,
 };
@@ -126,8 +126,13 @@ impl TypedProcInfo {
         Ok(felts)
     }
 
-    /// How many felts the procedure leaves on the stack. `Some(0)` if it returns nothing, `None`
-    /// if a result type has no place on the stack.
+    /// Whether the procedure have a result.
+    pub fn returns_value(&self) -> bool {
+        !self.signature.results.is_empty()
+    }
+
+    /// How many felts the procedure leaves on the stack, or `None` if a result type has no
+    /// place on the stack.
     pub fn output_felt_count(&self) -> Option<usize> {
         self.signature
             .results
@@ -142,12 +147,14 @@ impl TypedProcInfo {
     /// a stack that is too short, or felts that are not a valid value of the type. A value that a
     /// codec says is bad is also an error. The user should hear about that, not see raw felts.
     pub fn decode_result(&self, stack: &[Felt]) -> Result<Option<String>, TypedError> {
+        // A result can take no felts and still be a result: an empty struct, or an array of
+        // length zero.
+        if !self.returns_value() {
+            return Ok(None);
+        }
         let total = self
             .output_felt_count()
             .ok_or_else(|| TypedError::UnsupportedResult { procedure: self.name.clone() })?;
-        if total == 0 {
-            return Ok(None);
-        }
         if total > stack.len() {
             return Err(TypedError::ResultStackTooShort {
                 procedure: self.name.clone(),
@@ -182,7 +189,7 @@ impl TypedProcInfo {
 /// Prints the signature, like `add-points(point, point) -> point`.
 impl fmt::Display for TypedProcInfo {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&format_signature(&self.name, &self.signature))
+        f.write_str(&format_signature(&self.name, &self.signature, &self.codecs))
     }
 }
 
@@ -196,35 +203,70 @@ impl fmt::Display for TypedProcInfo {
 /// and we show its last part. We walk the types that hold other types so a struct inside one gets
 /// the same treatment, and we print them the way [`Type`] itself does, so a signature reads like
 /// the source it came from.
-fn format_type(ty: &Type) -> String {
+fn format_type(ty: &Type, codecs: &[Box<dyn WitScalarCodec>]) -> String {
     match ty {
         Type::Struct(struct_ty) => {
-            // `name` outlives the borrow `type_leaf_name` returns.
-            let name = struct_ty.name();
-            match name.as_deref().and_then(type_leaf_name) {
-                Some(leaf) => leaf.into(),
-                None => {
+            // A codec reads one token for the whole struct, not one per field, the same way
+            // `token_count` and `decode_type` treat it. So the signature shows the codec's name
+            // for it, not a field list that would wrongly ask the user for one token per field.
+            if let Some(codec) = codec_for_struct(codecs, struct_ty) {
+                return codec.wit_name().to_string();
+            }
+
+            let fields = struct_ty.fields();
+            // A struct with mixed field names has no shape, and `Display` cannot show an error, so
+            // the signature says so instead of looking correct. `decode_type` is where it becomes
+            // an error.
+            match struct_shape(struct_ty) {
+                // A struct with a name is its name. Its fields are not part of the signature.
+                Some(
+                    StructShape::Tuple { name: Some(name) }
+                    | StructShape::Record { name: Some(name) },
+                ) => name,
+                Some(StructShape::Tuple { name: None }) => {
                     let fields: Vec<String> =
-                        struct_ty.fields().iter().map(|field| format_type(&field.ty)).collect();
+                        fields.iter().map(|field| format_type(&field.ty, codecs)).collect();
                     format!("({})", fields.join(", "))
+                },
+                // A record has a name on every field, so `?` only shows up for a struct whose
+                // fields are partly named. It stands for a field `field_name` reads as
+                // positional: no name, an empty name, or the position itself, like `"0"`.
+                Some(StructShape::Record { name: None }) | None => {
+                    let rendered: Vec<String> = fields
+                        .iter()
+                        .enumerate()
+                        .map(|(i, field)| {
+                            let name = field_name(field, i).unwrap_or("?");
+                            format!("{name}: {}", format_type(&field.ty, codecs))
+                        })
+                        .collect();
+                    let body = rendered.join(", ");
+                    // The name too, so the reader knows which type is bad, not only that one is.
+                    let name = struct_ty.name();
+                    match name.as_deref().and_then(type_leaf_name) {
+                        Some(name) => format!("{name} {{ {body} }}"),
+                        None => format!("{{ {body} }}"),
+                    }
                 },
             }
         },
-        Type::Array(array_ty) => format!("[{}; {}]", format_type(&array_ty.ty), array_ty.len),
-        Type::List(element_ty) => format!("list<{}>", format_type(element_ty)),
-        Type::Ptr(ptr_ty) => {
-            format!("ptr<{}, {}>", ptr_ty.addrspace, format_type(&ptr_ty.pointee))
+        Type::Array(array_ty) => {
+            format!("[{}; {}]", format_type(&array_ty.ty, codecs), array_ty.len)
         },
-        Type::Function(sig) => format_signature("fn", sig),
+        Type::List(element_ty) => format!("list<{}>", format_type(element_ty, codecs)),
+        Type::Ptr(ptr_ty) => {
+            format!("ptr<{}, {}>", ptr_ty.addrspace, format_type(&ptr_ty.pointee, codecs))
+        },
+        Type::Function(sig) => format_signature("fn", sig, codecs),
         primitive => primitive.to_string(),
     }
 }
 
 /// Prints `name(a, b) -> c`. With no results there is no arrow. With more than one result they
 /// go in a tuple.
-fn format_signature(name: &str, sig: &FunctionType) -> String {
-    let params: Vec<String> = sig.params.iter().map(format_type).collect();
-    let results: Vec<String> = sig.results.iter().map(format_type).collect();
+fn format_signature(name: &str, sig: &FunctionType, codecs: &[Box<dyn WitScalarCodec>]) -> String {
+    let params: Vec<String> = sig.params.iter().map(|ty| format_type(ty, codecs)).collect();
+    let results: Vec<String> = sig.results.iter().map(|ty| format_type(ty, codecs)).collect();
 
     let ret = match results.as_slice() {
         [] => String::new(),
--- a/crates/mast-package/src/typed/mod.rs
+++ b/crates/mast-package/src/typed/mod.rs
@@ -1,0 +1,236 @@
+//! A typed view of a procedure signature.
+//!
+//! The manifest keeps the high-level (WIT) signature of each export; the VM stack only holds
+//! felts. [`TypedProcInfo`] joins the two: it prints the signature, turns argument text into stack
+//! felts, and result felts back into text.
+//!
+//! Types that take one token, like `word` and `felt`, go through a [`WitScalarCodec`]. Those two
+//! are built in; the caller adds others with [`TypedProcInfo::with_scalar_codec`].
+
+use alloc::{
+    boxed::Box,
+    format,
+    string::{String, ToString},
+    vec::Vec,
+};
+use core::fmt;
+
+use miden_assembly_syntax::ast::types::{FunctionType, Type};
+use miden_core::Felt;
+
+use self::{
+    arity::{felt_count, token_count},
+    codec::type_leaf_name,
+    decode::decode_type,
+    encode::encode_type,
+};
+
+mod arity;
+mod codec;
+mod decode;
+mod encode;
+mod errors;
+#[cfg(test)]
+mod tests;
+
+pub use self::{
+    codec::{FeltCodec, MIDEN_CORE_TYPES, WitScalarCodec, WordCodec},
+    errors::TypedError,
+};
+
+// TYPED PROCEDURE INFO
+// ================================================================================================
+
+/// The typed signature of one exported procedure.
+///
+/// Build it with [`Self::new`], then use [`Self::encode_args`] for arguments and
+/// [`Self::decode_result`] for results. [`fmt::Display`] prints the signature, like
+/// `add-points(point, point) -> point`.
+pub struct TypedProcInfo {
+    /// The name we show, like `add-points`.
+    name: String,
+    /// The signature from the package manifest.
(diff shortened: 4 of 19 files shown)


crates/mast-package/src/debug_info/typed/decode.rs
```diff
@@ -0,0 +1,169 @@
...
+/// How many felts a value of type `idx` takes on the stack. The decoder reads exactly this many
+/// to render the result. `None` when the count isn't known statically (dynamic arrays, `Unknown`),
+/// the type graph is cyclic/deeper than [`MAX_TYPE_DEPTH`], or a count computation overflows.
+pub(super) fn stack_felt_count(view: &TypedView<'_>, idx: DebugTypeIdx) -> Option<usize> {
+    count_units(view, idx, primitive_felt_count, |c| c.felt_count())
+}
+
+/// Stack felts a primitive leaf occupies.
+fn primitive_felt_count(p: DebugPrimitiveType) -> usize {
+    match p {
```
huitseeker: This size table looks too narrow for the existing debug type model. `DebugPrimitiveType::size_in_felts()` says `I64`/`U64`/`F32`/`F64` take 2 felts and `I128`/`U128` take 4, but this path counts them as 1. Then `output_felt_count()` and struct/array sizing will read too few stack elements for those declared result types.

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,127 @@
...
+                "true" | "1" => 1u64,
+                "false" | "0" => 0,
+                _ => return Err(TypedDebugInfoError::InvalidBool(token)),
+            };
+            Felt::try_from(v)
+                .map(|f| alloc::vec![f])
+                .map_err(|_| TypedDebugInfoError::FeltOutOfRange(token))
+        },
+        DebugPrimitiveType::Void => Ok(Vec::new()),
+        _ => Ok(alloc::vec![parse_felt_token(&token)?]),
```
huitseeker: This also accepts values outside the declared scalar type. For example a `U32` argument token larger than `u32::MAX` still passes as long as it fits in a field element, while signed values and full-width `U64`/`U128` are not encoded with their declared shape. It seems better to parse each primitive by its own width and signedness, or reject the unsupported ones for now.

huitseeker (changes requested):
Typed primitive sizing and encoding do not match declared debug primitive widths

crates/mast-package/src/debug_info/typed/decode.rs
marijamijailovic: > `DebugPrimitiveType::size_in_felts()` says `I64`/`U64`/`F32`/`F64` take 2 felts 

Thanks for the comments, I’ll fix this. I just want to check: is it a bug that ‎`F32` takes 2 felts? Should it take 1?

marijamijailovic:
> Typed primitive sizing and encoding do not match declared debug primitive widths

Hi @huitseeker, thanks for your comment! I’ve updated the PR to correct the felt sizing in encoding/decoding.

marijamijailovic:
I’ve rebased this branch and I’m re‑introducing ‎`size_in_bytes` and ‎`size_in_felts` on ‎`DebugPrimitiveType`. These methods were removed in #3221, but the new typed encoder/decoder needs the per‑primitive byte width and felt count in order to lower argument tokens into stack felts and lift return felts back from the stack.

crates/mast-package/src/debug_info/typed/decode.rs
```diff
@@ -0,0 +1,205 @@
...
+            Some((word.to_hex(), rest))
+        },
+        DebugPrimitiveType::Felt => {
+            let (head, rest) = felts.split_first()?;
+            Some((head.to_string(), rest))
+        },
+        DebugPrimitiveType::Bool => {
+            let (head, rest) = felts.split_first()?;
+            let v = head.as_canonical_u64();
+            Some((if v == 0 { "false".into() } else { "true".into() }, rest))
```
huitseeker: Can this reject any value other than `0` or `1`? The VM treats booleans as exactly those two felts, and `encode_arg` already enforces that on input. With this branch a malformed return felt like `2` renders as `bool(true)`, which can hide bad output instead of falling back to raw felts.

huitseeker (changes requested):
Besides the non-canonical bool, this looks much closer now!

crates/mast-package/src/debug_info/typed/decode.rs
marijamijailovic: Right! Thanks, I am adding strict check.

marijamijailovic:
> Besides the non-canonical bool, this looks much closer now!

@huitseeker Thanks for your comments. I’ve added strict checks for boolean, and also for integer widths (including rejecting out‑of‑range felts), and expanded the typed debug errors so we get more precise diagnostics

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,805 @@
...
+    /// (codec-handled scalars like `word` and `account-id` take one token each; other structs
+    /// take one per leaf field). `None` if any parameter has no statically-known token count
+    /// (e.g. a dynamic array).
+    pub fn expected_arg_count(&self) -> Option<usize> {
+        let view = self.view();
+        self.params.iter().map(|p| arg_token_count(&view, p.type_idx)).sum()
+    }
+
+    /// Type-table indices of the procedure's parameters, in declaration order. Pair with
+    /// [`encode_tokens`] to encode argument tokens parameter by parameter; the orchestration
```
huitseeker: This link breaks the doc job because `encode_tokens` is private, and rustdoc treats private links from public docs as an error. Could this point readers to `Self::encode_arg` instead, or make this plain text?

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,805 @@
...
+
+/// `(return_type, param_types)` from the `Function`-typed debug entry, or `(None, vec![])` if
+/// the entry has no Function type.
+fn extract_signature_types(
+    func: &DebugFunctionInfo,
+    types: &DebugTypesSection,
+) -> (Option<DebugTypeIdx>, Vec<DebugTypeIdx>) {
+    match func.type_idx.and_then(|i| types.get_type(i)) {
+        Some(DebugTypeInfo::Function { return_type_idx, param_type_indices }) => {
+            (*return_type_idx, param_type_indices.clone())
```
huitseeker: This preserves `Some(Void)` as a real return type. The assembler uses `Some(Void)` for functions with no results, so those signatures will print as `foo() -> Void` instead of `foo()`. Could this normalize a `Void` return to `None` here?

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,163 @@
...
+    string::{String, ToString},
+};
+
+use miden_core::serde::Deserializable;
+
+use super::super::{DebugFunctionInfo, DebugFunctionsSection, DebugTypesSection};
+use crate::{Package, SectionId};
+
+/// Reads both debug sections from `package`. `None` if either is missing or fails to decode.
+pub(super) fn read_debug_sections(
```
huitseeker: Could this use `Package::debug_info()` instead of reading `package.sections` directly?

That would remove this local section reader and also inherit the package debug-info trust policy. The serialization work in #3284/#3294 is tightening that path, so it seems better for typed debug info to reuse it rather than decoding `DEBUG_FUNCTIONS` and `DEBUG_TYPES` a second way.

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,805 @@
...
+impl TypedProcInfo {
+    /// Resolves the typed signature of `procedure_name` from `package`'s debug sections.
+    ///
+    /// `None` if the package has no debug info or no entry for `procedure_name`. The procedure
+    /// name is matched on the bare leaf with `_` and `-` treated as equal, so both `get_count`
+    /// and `get-count` resolve the same entry.
+    ///
+    /// The returned info has the built-in [`WordCodec`] registered; add further scalar codecs
+    /// with [`Self::with_scalar_codec`].
+    pub fn from_package(package: &Package, procedure_name: &str) -> Option<Self> {
```
huitseeker: The one API choice is whether `from_package` should keep returning `Option` and silently fall back on debug-info errors, or return `Result<Option<_>, PackageDebugInfoError>` so callers can distinguish missing debug info from untrusted or malformed debug info.

huitseeker (changes requested):
Please add a line to the Changelog too.

marijamijailovic:
> Please add a line to the Changelog too.

Hi @huitseeker, thanks again for the review and comments! I’ve added the changelog entry, and the other changes as well (normalizing the ‎`Void` return type to ‎`None`, using ‎`Package::debug_info`, and fixing the doc link)

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// The compiler emits the same procedure under two names: the `::`-path form (`..."get-count"`)
+/// has the high-level typed signature, the `#`-qualified form (`...@0.1.0#get-count`) the lowered
+/// (felt-flattened) one. Both have a `type_idx`, so prefer the `::`-path form. Otherwise take any
+/// entry with a `type_idx`, then any match at all.
+pub(super) fn find_debug_fn<'a>(
+    funcs: &'a DebugFunctionsSection,
+    procedure_name: &str,
+) -> Option<&'a DebugFunctionInfo> {
+    let kebab = procedure_name.replace('_', "-");
```
huitseeker: Can this keep an exact-name match before the snake-to-kebab fallback? The assembler writes `procedure.path().as_str()` into `DEBUG_FUNCTIONS`, and MASM names can use `_`. With only this rewritten key, `from_package(package, "foo_bar")` will not find a debug entry whose leaf is really `foo_bar`; it only looks for `foo-bar`.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1627,6 +1627,37 @@ pub enum DebugPrimitiveType {
...
+            | Self::I8
+            | Self::U8
+            | Self::I16
+            | Self::U16
+            | Self::I32
+            | Self::U32
+            | Self::F32
+            | Self::Felt => 1,
+            Self::I64 | Self::U64 | Self::F64 => 2,
+            Self::I128 | Self::U128 | Self::Word | Self::U256 => 4,
```
huitseeker: Can `U256` avoid returning `4` here? The core docs and handler use eight 32-bit limbs for a u256 on the stack, so `output_felt_count()` will tell callers to read only half of a `U256` result. Returning `None` is better than something wrong.

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,818 @@
...
+        self
+    }
+
+    /// Number of argument tokens the procedure expects, summed over its parameters
+    /// (codec-handled scalars like `word` and `account-id` take one token each; other structs
+    /// take one per leaf field). `None` if any parameter has no statically-known token count
+    /// (e.g. a dynamic array).
+    pub fn expected_arg_count(&self) -> Option<usize> {
+        let view = self.view();
+        self.params.iter().map(|p| arg_token_count(&view, p.type_idx)).sum()
```
huitseeker: Can this final sum also use checked addition? `count_units` protects each parameter from overflowing internally, but two large fixed-array parameters can still overflow here and turn `expected_arg_count()` into a wrapped count in release builds instead of `None`.

huitseeker (commented):
Thanks for addressing the prior bool/trust-path/changelog concerns.

crates/mast-package/src/debug_info/types.rs
marijamijailovic: For ‎`u256` I kept ‎`size_in_felts` at 8 rather than dropping it to ‎`None`, so its stack layout stays correct. What’s still missing is typed encode/decode, which currently returns ‎`UnsupportedType`/‎`None`. I’d prefer to add full ‎`u256` support in a focused follow‑up so this PR stays scoped to the priority types, tracked in #2098.

crates/mast-package/src/debug_info/typed/mod.rs
marijamijailovic: Added!

crates/mast-package/src/debug_info/typed/lookup.rs
marijamijailovic: Fixed

marijamijailovic:
> Thanks for addressing the prior bool/trust-path/changelog concerns.

Thanks again @huitseeker for all the detailed reviews and suggestions

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,179 @@
...
+    let mut typed_path: Option<&DebugFunctionInfo> = None;
+    let mut typed_any: Option<&DebugFunctionInfo> = None;
+    let mut first_any: Option<&DebugFunctionInfo> = None;
+    for f in &funcs.functions {
+        let Some(s) = funcs.strings.get(f.name_idx as usize) else {
+            continue;
+        };
+        let s = s.as_ref();
+        let leaf = proc_display_name(s);
+        if leaf != procedure_name && leaf != kebab {
```
huitseeker: Can this prefer an exact leaf match before the kebab fallback? Right now exact and kebab matches share the same buckets, so if the debug table has both `foo-bar` and `foo_bar`, a `from_package(package, "foo_bar")` call can still return whichever entry appears first. That still misses the MASM proc whose name really contains `_`.

marijamijailovic: Good catch, and apologies for missing that on the first pass. 

I’ve updated ‎`find_debug_fn` to do two passes: first with the exact ‎`procedure_name`, then with the kebab form. Each pass uses strict leaf equality. That way, if both ‎`foo-bar` and ‎`foo_bar` exist in the table, a ‎`from_package(package, "foo_bar")` query will resolve to the exact ‎`_` leaf .

crates/mast-package/src/debug_info/typed/decode.rs
bitwalker: You should refer to `midenc-hir-types` layout rules for this information, specifically `Type::size_in_felts`. The `F32` type can occupy a single felt, since it is a 32-bit value.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1627,6 +1627,38 @@ pub enum DebugPrimitiveType {
 }
 
 impl DebugPrimitiveType {
+    /// Returns the size of this primitive type in bytes.
+    pub const fn size_in_bytes(self) -> u32 {
```
bitwalker: I would suggest we add a conversion from `DebugPrimitiveType` to `midenc_hir_types::Type`, and then call `midenc_hit_types::Type::size_in_bytes`/`size_in_felts`.

That will ensure that we have a single canonical place where the layout rules are implemented, and avoids any drift between the two. The implementation in `midenc_hir_types` also handles all types, not just primitive types.

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+    pub fn with_scalar_codec(mut self, codec: Box<dyn WitScalarCodec>) -> Self {
+        self.codecs.push(codec);
+        self
+    }
+
+    /// Number of argument tokens the procedure expects, summed over its parameters
+    /// (codec-handled scalars like `word` and `account-id` take one token each; other structs
+    /// take one per leaf field). `None` if any parameter has no statically-known token count
+    /// (e.g. a dynamic array).
+    pub fn expected_arg_count(&self) -> Option<usize> {
```
bitwalker: I find this method confusing, and suggest we rename it, or clarify the documentation (and likely both):

* If a procedure has no type signature associated with it, we can't really provide any information for it at all (aside from its name and possibly its calling convention), i.e. it would be better to just refuse to construct a `TypedProcInfo` at all in those cases, and have APIs that request a `TypedProcInfo` for some procedure return `None` if that information is unavailable.
* The calling conventions in MASM always have a fixed arity, even when a dynamically-sized type is involved - because dynamically-sized types must always be passed by reference, and those references have a fixed representation (a fat pointer, i.e. two felts).

With that in mind, it isn't clear to me in what circumstances this would ever return `None`.

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+    /// Encodes the argument tokens for the parameter (or any type) at `idx` into its stack felts,
+    /// consuming exactly the tokens that type needs from `tokens` and using the registered scalar
+    /// codecs. Codec-handled scalars (`word`, `account-id`, ..) each parse from a single token;
+    /// other structs expect one token per leaf field.
+    ///
+    /// Exposed per parameter rather than as a whole-signature call so the caller owns the argument
+    /// policy — how many are expected (see [`Self::expected_arg_count`] and
+    /// [`Self::param_type_indices`]) and how a wrong count is reported — since turning
+    /// user-supplied strings into arguments is an application concern.
+    pub fn encode_arg<I: Iterator<Item = String>>(
```
bitwalker: Why does this take an iterator of `String`? That seems excessively wasteful in terms of allocations when the value isn't ever used.

I'd modify the signature here to be:


```suggestion
    pub fn encode_arg<I>(
        &self,
        tokens: &mut I,
        idx: DebugTypeIdx
    ) -> Result<Vec<Felt>, TypedDebugInfoError>
    where
        I: IntoIterator,
        <I as IntoIterator>::Item: AsRef<str>,
    {
```

This allows passing a vector of `String`, `&str`, or anything else that is resolvable to a `&str`.

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+        &self,
+        tokens: &mut I,
+        idx: DebugTypeIdx,
+    ) -> Result<Vec<Felt>, TypedDebugInfoError> {
+        encode_tokens(tokens, &self.view(), idx)
+    }
+
+    /// Number of felts the procedure pushes on the stack for its return value. `Some(0)` if
+    /// there is no return type; `None` if the return type has no statically-known felt size.
+    pub fn output_felt_count(&self) -> Option<usize> {
```
bitwalker: Similar to `expected_arg_count`, this should always be resolvable to a value when a procedure type signature is available

crates/mast-package/src/debug_info/typed/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+        }
+        let slice = &stack[..n];
+        let return_ty = self.types.get_type(return_idx);
+        let (rendered, rest) = decode_value(slice, &view, return_idx)?;
+        if !rest.is_empty() {
+            return None;
+        }
+        Some(match return_ty {
+            Some(DebugTypeInfo::Primitive(p)) => {
+                format!("{}({rendered})", format!("{p:?}").to_lowercase())
```
bitwalker: I would suggest we use suffix format, e.g. `42u32`/`-1i32`/`0felt`, etc. - that aligns closer with how typed primitives are typically expressed when written

crates/mast-package/src/debug_info/typed/decode.rs
```diff
@@ -0,0 +1,221 @@
...
+    let shift = 128 - bits;
+    (((value << shift) as i128) >> shift).to_string()
+}
+
+/// `name(body)` for named structs; `{body}` for anonymous or unnamed ones.
+fn wrap_struct(short: &str, body: &str) -> String {
+    if is_anonymous(short) {
+        format!("{{{body}}}")
+    } else {
+        format!("{short}({body})")
```
bitwalker: Why `{ body }`/`name(body)` rather than `{ body }`/`name { body }`? This better matches how the type signatures are written/expressed IMO

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+                if empty {
+                    break;
+                }
+            }
+            Ok(felts)
+        },
+        DebugTypeInfo::Array { count: None, .. } => {
+            Err(TypedDebugInfoError::UnsupportedType("array"))
+        },
+        // No defined encoding for the below type shapes as arguments.
```
bitwalker: This is incorrrect:

* `Pointer` is always a `u32`-typed value, i.e. it fits in a single felt
* `Function` is a reference type, represented as a `u32`-typed index into the global function table written into memory during component initialization.
* `Enum` is either C-like (in which case its representation is a value of its discriminant type), or struct-like (in which case it must be represented as an opaque blob of bytes/felts equal in size to its largest variant + discriminant, because each variant might have a different payload).
* `Unknown` of course has no representation, and I don't think we should even be representing `Unknown` as a variant of `DebugTypeInfo`, as we either have debug type information or we don't, and anywhere that we'd be creating a value of `DebugTypeInfo::Unknown` we should be returning an error instead IMO

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+        // only for a core `Word` primitive.
+        DebugPrimitiveType::Word => WordCodec.encode(&token),
+        DebugPrimitiveType::Void => Ok(Vec::new()),
+        DebugPrimitiveType::Bool => {
+            let v = match token.to_ascii_lowercase().as_str() {
+                "true" | "1" => 1u64,
+                "false" | "0" => 0,
+                _ => return Err(TypedDebugInfoError::InvalidBool(token)),
+            };
+            Felt::try_from(v)
```
bitwalker: You can just use `Felt::from_u32` here - it's impossible for 1 or 0 to overflow `Felt`

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+        },
+        // 256-bit values exceed the `u128` limb path; no typed encoding is defined for them yet.
+        DebugPrimitiveType::U256 => Err(TypedDebugInfoError::UnsupportedType("u256")),
+    }
+}
+
+/// Cuts `value` into `n` 32-bit limbs (low first), one felt each. Each limb is a `u32`, so
+/// [`Felt::from_u32`] never fails.
+fn limbs_to_felts(value: u128, n: usize) -> Vec<Felt> {
+    (0..n).map(|i| Felt::from_u32((value >> (32 * i)) as u32)).collect()
```
bitwalker: I would add a comment here that this extracts the limbs of `value` in order of least-significant to most-significant to make it clear what the expectations are with regard to endianness.

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Reads an unsigned int (decimal or `0x..` hex) and checks it fits in `bits` bits.
+fn parse_unsigned(
+    token: &str,
+    bits: u32,
+    p: DebugPrimitiveType,
+) -> Result<u128, TypedDebugInfoError> {
+    let value = if let Some(hex) = token.strip_prefix("0x").or_else(|| token.strip_prefix("0X")) {
```
bitwalker: In what situations are we emitting hex? Do we need to be supporting that? I'd remove it unless we absolutely require it.

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/// Reads a signed int and returns its two's-complement bits. A `0x..` token is raw bits (must fit
+/// in `bits`); a decimal token is checked against the signed range.
+fn parse_signed(
+    token: &str,
+    bits: u32,
+    p: DebugPrimitiveType,
+) -> Result<u128, TypedDebugInfoError> {
+    if let Some(hex) = token.strip_prefix("0x").or_else(|| token.strip_prefix("0X")) {
```
bitwalker: Same comment here wrt to hex encoded literals

crates/mast-package/src/debug_info/typed/encode.rs
```diff
@@ -0,0 +1,215 @@
...
+    Ok((signed as u128) & max_for_bits(bits))
+}
+
+fn int_out_of_range(token: &str, p: DebugPrimitiveType) -> TypedDebugInfoError {
+    TypedDebugInfoError::IntOutOfRange { token: token.to_string(), ty: p }
+}
+
+/// Reads a decimal or `0x..` hex token into a `Felt` (full field range). Used by the `Felt` arm.
+fn parse_felt_token(s: &str) -> Result<Felt, TypedDebugInfoError> {
+    let v: u64 = if let Some(hex) = s.strip_prefix("0x").or_else(|| s.strip_prefix("0X")) {
```
bitwalker: Same comment wrt to hex encoding. I believe the only place it likely makes sense to support hex is for words

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,235 @@
...
+                }
+            }
+        }
+        typed_path.or(typed_any).or(first_any)
+    })
+}
+
+/// The bare type name from a full WIT path.
+///
+/// Debug type names are full WIT paths with namespace and version, e.g.
```
bitwalker: That's only true in some cases - debug type names _may_ be encoded WIT-style, but not necessarily.

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,235 @@
...
+/// `miden:base/core-types@1.0.0/account-id`. We want the last `/`-segment (`account-id`) for
+/// display and for matching a type regardless of package or version. Empty if `name` is empty or
+/// ends in `/`.
+pub(super) fn wit_type_name(name: &str) -> &str {
+    name.rsplit('/').next().filter(|s| !s.is_empty()).unwrap_or("")
+}
+
+/// The short procedure name from a long compiler name: take the last `::` part, remove the quotes,
+/// and drop anything before `#`. `::"miden:cc/mcc@0.1.0"::"get-count"` becomes `get-count`.
+pub(super) fn proc_display_name(raw: &str) -> &str {
```
bitwalker: This is too sensitive to how the compiler encodes names currently (and specifically how it encodes names for Rust - it does not do so for other languages). I would prefer if we kept the code here as free of compiler-specific details like this as possible.

But you should also use `miden_assembly_syntax::Path` for these types of operations, e.g. `Path::new(raw).split_last()`.

Do you have examples of where `#` is used? It isn't clear to me why we'd want to remove that, since it is technically part of the procedure name.

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,235 @@
...
+    let seg = raw.rsplit("::").next().unwrap_or(raw).trim_matches('"');
+    seg.rsplit('#').next().unwrap_or(seg)
+}
+
+/// Whether a struct's short name is anonymous, so callers render its fields instead of a name.
+/// `<anon>` is the sentinel the compiler writes when it has no name for a struct.
+pub(super) fn is_anonymous(short: &str) -> bool {
```
bitwalker: We don't need to match the compiler behavior here - I would just elide the struct name part for anonymous structs.

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,235 @@
...
+    short.is_empty() || short == "<anon>"
+}
+
+/// The type/field name string at `name_idx`, or `""` when the string table has no entry.
```
bitwalker:
```suggestion
/// The type/field name string at `name_idx`, or the empty string when the string table has no entry.
```

The way this was worded makes it sound like the name would be a literal `""`, rather than just an empty string

crates/mast-package/src/debug_info/typed/lookup.rs
```diff
@@ -0,0 +1,235 @@
...
+pub(super) fn type_name_raw(types: &DebugTypesSection, name_idx: u32) -> &str {
+    types.strings.get(name_idx as usize).map_or("", AsRef::as_ref)
+}
+
+/// Display name of a struct field at `name_idx`, falling back to `f<idx>` when the string table
+/// has no entry.
+pub(super) fn field_name(types: &DebugTypesSection, name_idx: u32) -> String {
```
bitwalker: I think we should print structs with anonymous fields as tuples, rather than attempt to print them as structs. We currently do not emit struct types with a mix of named and unnamed fields (and I don't see any reason to support that). We should treat the debug info as invalid in such cases IMO

bitwalker (changes requested):
Looks good overall, but I have some nits/changes that I think need to be made before we merge this

crates/mast-package/src/debug_info/typed/decode.rs
marijamijailovic: Thanks @bitwalker  for reply and reference to midenc-hir-types.

crates/mast-package/src/debug_info/typed/decode.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+    let ty = view.types.get_type(idx)?;
+    match ty {
+        DebugTypeInfo::Primitive(p) => decode_primitive(felts, *p),
+        DebugTypeInfo::Struct { name_idx, fields, .. } => {
+            let full = type_name_raw(view.types, *name_idx);
+            let short = wit_type_name(full);
+            // A codec that fails to render (invalid value) falls through to the generic
+            // field-by-field rendering below.
+            if let Some(codec) = view.codec_for(short)
```
huitseeker: Can this skip the generic struct fallback when a registered scalar codec rejects the value? For `account-id`, the codec is where the real validity check lives. If it returns `None`, this path still prints the same felts as `account-id(prefix=..., suffix=...)`, so a bad value looks typed instead of falling back to raw felts.

crates/mast-package/src/debug_info/typed/decode.rs
marijamijailovic: Hi @bitwalker, I’ve got a couple of questions from wiring this up to ‎`Type::size_in_felts` that I’d like your input on.

The first is `F32`. It doesn’t exist in `midenc_hir_type::Type`, but `DebugPrimitiveType` has it, so I can’t get its layout from `Type` like the others. Can we handle it like

```
    pub fn size_in_felts(self) -> u32 {
        match self {
            Self::Void => 0,
            // Not part of the canonical type system; a 32-bit value occupies a single felt.
            Self::F32 => 1,
            other => Type::try_from(other)
                .expect("all non-Void/F32 primitives have a canonical type")
                .size_in_felts() as u32,
        }
    }
```

Separately, I noticed [`midenc-hir-type`](https://github.com/0xMiden/midenc-hir-type/tree/next) was archived on Jul 7 and couldn’t find a replacement mentioned anywhere. Is `midenc-hir-type`(as per [comment)](https://github.com/0xMiden/miden-vm/pull/3276#discussion_r3514487831) still the layout reference we should be following for `size_in_bytes`/`size_in_felts`, or has the type system moved somewhere else?

crates/mast-package/src/debug_info/types.rs
marijamijailovic: Thanks, I’ve wired this up to `midenc_hir_types::Type` as suggested: `DebugPrimitiveType` now converts into `Type`, and `size_in_bytes`/`size_in_felts` just delegate to its layout rules.

Since `midenc_hir_types` was archived on Jul 7, is that still the right crate to treat as the layout source of truth, or has the type system moved somewhere else that we should be targeting instead?

crates/mast-package/src/debug_info/typed/mod.rs
marijamijailovic: Agree, with no signature we shouldn’t build a ‎`TypedProcInfo` at all.

I’ve updated ‎`extract_signature_types` to return ‎`None` when the debug entry has no ‎`Function` type, and ‎`from_package` now returns ‎`Ok(None)` in that case. So constructed ‎`TypedProcInfo` should always has a signature.

I also renamed the method to ‎`expected_arg_token_count` and clarified the docs.

crates/mast-package/src/debug_info/typed/mod.rs
marijamijailovic: Applied! Thanks!

crates/mast-package/src/debug_info/typed/mod.rs
marijamijailovic: Same fix as `expected_arg_token_count` applied.

crates/mast-package/src/debug_info/typed/mod.rs
marijamijailovic: Done

crates/mast-package/src/debug_info/typed/decode.rs
marijamijailovic: Agree, done!

crates/mast-package/src/debug_info/typed/encode.rs
marijamijailovic: Thanks! This was wrong, no more hex!

crates/mast-package/src/debug_info/typed/encode.rs
marijamijailovic: No hex!

crates/mast-package/src/debug_info/typed/encode.rs
marijamijailovic: Removed `parse_felt_token`. Since the compiler lowers `felt` to a named struct with a single `inner` field, the generic struct path was rendering values as `felt { inner: 42 }`. 

We now register a `FeltCodec` and treat `felt` as a scalar instead: one decimal token in, one felt out, and it renders as a bare `42` rather than exposing the `inner` field.

crates/mast-package/src/debug_info/typed/lookup.rs
marijamijailovic: Anonymous structs now print only their body (`{ x: 3, y: 4 }` or `(3, 4)`), with no placeholder name.

I’ve kept the `<anon>` check only to stop that sentinel leaking into output. As far as I can see (debuginfo.rs:319, 388), the current assembler emits `<anon>` for nameless structs.

Does it make sense to keep the `<anon>` check on the reader side for now, rather than drop it here?

crates/mast-package/src/debug_info/typed/lookup.rs
marijamijailovic: Structs with unnamed fields now render as tuples: `(3, 4)` for values and `(Felt, Felt)` in signatures, instead of `{ 0: 3, 1: 4 }`.

I haven’t been able to reproduce `<anon>` from Rust: as `#[component]` only accepts named types, `#[export_type]` rejects tuple structs, and internal tuple returns are lowered to out-pointers, so these <anon> only appear in the package from MASM:
- `STRUCT <anon> { 0: felt, 1: felt }` (from `-> (felt, felt)`) renders as `(3, 4)`
- `STRUCT <anon> { x: felt, y: felt }` (from `struct { x, y }`) renders as `{ x: 3, y: 4 }`

So one check elides the `<anon>` name, and the field names decide tuple vs record. Mixed named/unnamed fields are treated as invalid and fall back to raw felts.

Does this line up with how you were picturing anonymous and tuple structs behaving?

crates/mast-package/src/debug_info/typed/encode.rs
marijamijailovic: Right, and my comment was wrong. What I meant is that I’m treating `Pointer`/`Function`/`Enum`/`Unknown` as unsupported for now, not that they have no representation.

I create a small example with a C‑like enum (`CounterStatus { Idle, Running, Done }`) and a struct‑like enum (`CounterEvent { Reset, Set(Felt), Moved(Point) }`) to the counter-contract example and rebuilt. In the .masp debug info:

- the C‑like enum is lowered to its i32 discriminant,
- the struct‑like enum appears as UNKNOWN, and
- there is no ENUM entry at all.

So there’s no DebugTypeInfo::Enum to decode yet, which is why this branch currently returns Unsupported instead of guessing a layout.

Does it sound reasonable to keep these treated as unsupported in this PR and then add full enum support in a follow‑up once the compiler starts preserving enum types in the debug info?

crates/mast-package/src/debug_info/typed/lookup.rs
marijamijailovic: Thanks!

I’ve renamed the helper from wit_type_name` to `type_leaf_name`, since it just extracts the trailing leaf regardless of format (WIT / paths, :: paths, or a plain name).

I confirmed this with a debug dump: core types come through as WIT paths (`miden:base/core-types@1.0.0/felt`), while a local struct is just it name.

crates/mast-package/src/debug_info/typed/lookup.rs
marijamijailovic: Good points, all addressed. I’ve dropped the  ‎`#` handling: ‎`proc_display_name` now just takes the last path component and trims quotes. The lowered names with ‎`#` use flattened types, while we care about keeping the high‑level types. Since lowered names with ‎`#` (like ‎`"@0.1.0#get-count"`) don’t match the bare procedure name (‎`"get-count"`), we naturally resolve to the interface form, which is the one that keeps the high‑level types. With ‎`#` stripping gone, the extra preference logic in ‎`find_debug_fn` became dead and is removed.⁠

Here is [raw_debug_dump_full.txt](https://github.com/user-attachments/files/29861057/raw_debug_dump_full.txt) a MASP dump that shows the duplicate function names that originally led me to handle both ways of extracting the name.

Does this match what you were aiming for here, both on dropping the ‎`#` logic and on using ‎`Path::split_last()`?

marijamijailovic:
Thanks @huitseeker and @bitwalker for all the detailed review comments. I’ve gone through the latest round and addressed the requested changes. I’ve also left a couple of follow‑up questions in the relevant threads.

huitseeker (approved):
This LGTM! Please rebase.

marijamijailovic:
> This LGTM! Please rebase.

Thanks @huitseeker! I've rebased this PR.

marijamijailovic:
Hi @huitseeker, I’ve fixed the ‎`miden-core-fuzz` crate check

huitseeker (commented):
i'm sorry but this will need a rebase once again. Beware that the Changelog format has changed a bit. You can also click options to let maintainers edit this PR.

marijamijailovic:
Hi, I’ve rebased and pushed the latest changes. I don’t see the option to allow maintainer edits anywhere in the PR or any extra PR options in the sidebar.

bitwalker:
I think we will need to wait to determine the disposition of #3398 - whether we will call it quits on the v0.25 release series and move everything to v0.26, or land #3398 on `main` and yank all prior v0.25 releases - either way, it makes significant changes to how debug info is represented that directly affect this PR and the way the `miden call` command is implemented.

We're planning to make a final call on that tomorrow morning, but just wanted to give you a heads up @marijamijailovic

marijamijailovic:
> I think we will need to wait to determine the disposition of #3398 - whether we will call it quits on the v0.25 release series and move everything to v0.26, or land #3398 on `main` and yank all prior v0.25 releases - either way, it makes significant changes to how debug info is represented that directly affect this PR and the way the `miden call` command is implemented.
> 
> We're planning to make a final call on that tomorrow morning, but just wanted to give you a heads up @marijamijailovic

I appreciate the heads-up! I’ll wait for the decision on #3398 and then update this PR as needed.

marijamijailovic:
> I think we will need to wait to determine the disposition of #3398 - whether we will call it quits on the v0.25 release series and move everything to v0.26, or land #3398 on `main` and yank all prior v0.25 releases - either way, it makes significant changes to how debug info is represented that directly affect this PR and the way the `miden call` command is implemented.
> 
> We're planning to make a final call on that tomorrow morning, but just wanted to give you a heads up @marijamijailovic

Hi @bitwalker, is there any update regarding this and the plan for #3398?

marijamijailovic:
Hi @huitseeker and @bitwalker, I’ve now rebased this on top of #3398.

What changed:
- The module moved from `crates/mast-package/src/debug_info/typed/` to `crates/mast-package/src/typed/`.
- It no longer reads the `DEBUG_FUNCTIONS` / `DEBUG_TYPES` sections. It’s now a typed view over the package manifest’s procedure signatures.
- Re‑introducing `size_in_bytes` / `size_in_felts` on `DebugPrimitiveType` is no longer needed, since arity is derived from the manifest type instead (`typed/arity.rs`).

Since the old review threads now point at code that no longer exists, it’s probably cleanest to review the single new commit. The earlier commits are still reachable through the force‑push compare in this PR’s timeline (`f57c182f3` → `d6230b541`) if you’d like to compare.

Thanks again for all your previous comments and suggestions.

crates/assembly-syntax/src/ast/types/signatures/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+/// the source it came from.
+fn format_type(ty: &Type) -> String {
+    match ty {
+        Type::Struct(struct_ty) => {
+            // `name` outlives the borrow `type_leaf_name` returns.
+            let name = struct_ty.name();
+            match name.as_deref().and_then(type_leaf_name) {
+                Some(leaf) => leaf.into(),
+                None => {
+                    let fields: Vec<String> =
```
huitseeker, line 245: I think this means anonymous records lose field names in signatures, appearing as tuples.

`decode_type` treats an unnamed struct with named fields as a record, but this branch formats the same type as `(felt, felt)`, so the signature says tuple while the result is `{ x: ..., y: ... }`. 

The existing `unnamed_record` test checks only the result; asserting the signature there would catch this.

crates/mast-package/src/typed/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///
+    /// `Ok(None)` means the procedure returns nothing. That is the only case that is not a
+    /// problem. Everything else is an error the caller can show: a result type we cannot decode,
+    /// a stack that is too short, or felts that are not a valid value of the type. A value that a
+    /// codec says is bad is also an error. The user should hear about that, not see raw felts.
+    pub fn decode_result(&self, stack: &[Felt]) -> Result<Option<String>, TypedError> {
+        let total = self
+            .output_felt_count()
+            .ok_or_else(|| TypedError::UnsupportedResult { procedure: self.name.clone() })?;
+        if total == 0 {
```
huitseeker: Can this check `self.signature.results.is_empty()` instead? 

A declared empty struct or zero-length array also has `total == 0`, so this returns `Ok(None)` even though the procedure has a result and `decode_type` can render it. 

That breaks the method's promise that `Ok(None)` only means no return value.

huitseeker (commented):
The anonymous type comment is the important one.

crates/mast-package/src/typed/mod.rs
marijamijailovic: Added, thanks for this note! It also notice a bug: `[empty; 5]` was rendering as `[empty()]`, and now prints all five elements ([empty(), empty(), empty(), empty(), empty()]).

crates/assembly-syntax/src/ast/types/signatures/mod.rs
marijamijailovic: Fixed, `format_type` and `decode_type` now share `struct_shape`

marijamijailovic:
Hi @huitseeker, I have a quick question about the `CHANGELOG.md` format.

I see the latest entry is for version 0.28.0, and below that there is `miden-vm 0.25.8`. For this PR, where should I add the changelog entry?

marijamijailovic:
Hi @huitseeker, @bitwalker I’ve rebased this and added the changelog entry under the 0.30.0 release. Could you please take another look and let me know if it’s ready to merge?

crates/assembly-syntax/src/ast/types/signatures/decode.rs
```diff
@@ -0,0 +1,183 @@
...
+                (Some(name), false) => format!("{name} {{ {body} }}"),
+                (None, true) => format!("({body})"),
+                (None, false) => format!("{{ {body} }}"),
+            };
+            Ok((out, cursor))
+        },
+        Type::Array(array_ty) => {
+            let mut cursor = felts;
+            let mut rendered = Vec::new();
+            for _ in 0..array_ty.len {
```
huitseeker, line 85: Could we reject or cap zero-width arrays before this loop? `ArrayType::len` can come from a serialized `usize`, but `felt_count([empty; n])` is zero for any `n`, so the stack-length check does not bound the work here. 

A package declaring `[empty; 1_000_000]` made `decode_result(&[])` spend about 0.67 seconds and 62 MB building the string, and the encoded length can be much larger. This gives a small untrusted manifest a direct memory and CPU denial of service.

crates/mast-package/src/typed/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl TypedProcInfo {
+    /// Builds the typed view of the procedure `name`, with the signature its package manifest
+    /// carries.
+    ///
+    /// The caller picks the export. It knows which one it wants, and it keeps the signature and
+    /// anything else it reads from that export, like its digest, on the same procedure.
+    ///
+    /// The result has [`WordCodec`] and [`FeltCodec`]. Add more with [`Self::with_scalar_codec`].
+    pub fn new(name: impl Into<String>, signature: FunctionType) -> Self {
```
huitseeker: Could `TypedProcInfo` reject signatures whose `abi` is not `ComponentModel`, unless the other layouts are implemented? 

The encoder always flattens aggregate fields separately. That matches the canonical ABI used by the downstream `miden call`, but not `Fast`: the Miden ABI packs `struct { u8, u8 }` into one raw stack part, while this API emits two felts. `C` and `Wasm` have different aggregate rules too. 

The test helper currently labels every fixture `CallConv::Fast`, so it hides this mismatch.

crates/mast-package/src/typed/encode.rs
```diff
@@ -0,0 +1,156 @@
...
+        },
+        Type::Array(array_ty) => {
+            let mut felts = Vec::new();
+            for _ in 0..array_ty.len {
+                let chunk = encode_type(tokens, &array_ty.ty, codecs)?;
+                // An element of width zero reads no tokens and makes no felts. Without this
+                // check, a big `len` would loop and never move forward.
+                let empty = chunk.is_empty();
+                felts.extend(chunk);
+                if empty {
```
huitseeker: A custom codec for a zero-field struct still consumes one token even though it correctly emits zero felts. For `[empty; 2]`, the count check accepts two tokens, this breaks after the first codec call, and `encode_args` returns `TokenCountMismatch`. 

Checking `token_count` or otherwise tracking token progress would preserve the large zero-width-array guard without skipping token-consuming codecs.

huitseeker (commented):
I still think this works. But `TypedProcInfo` accepts calling conventions it cannot encode correctly, so the public API promises more than the implementation delivers. The zero-width array decode also lets a small serialized type description consume unbounded time and memory. There's also an edge case paper cut in custom-codec arrays.

I would restrict the API to Component Model signatures for now, bound zero-width decoding, and make the array loop track token consumption.

crates/assembly-syntax/src/ast/types/signatures/decode.rs
marijamijailovic: Thanks for this. I was thinking about a cap, but I was not sure about the upper limit, so I reject it with a ZeroWidthArray error instead. The test is `an_array_of_elements_that_take_no_felts_is_not_decoded`.

crates/mast-package/src/typed/encode.rs
marijamijailovic: Thanks! The array now skips its elements on token_count == 0, not on an empty chunk, so a codec that reads a token runs once per element and the token count matches.

crates/mast-package/src/typed/mod.rs
marijamijailovic: Thanks for this catch! The tests now use ComponentModel, and TypedProcInfo::new refuses any other convention

marijamijailovic:
Hi @huitseeker @bitwalker, just a quick bump on this PR. Can you look into this, and let me know if there’s anything else blocking merge?

bitwalker (changes requested):
I think the only thing I would like to change before this is merged is to move this code into `miden-assembly-syntax` under the `ast::types` module, maybe as `ast::types::signatures` or something to that effect, and re-export `TypedProcInfo` (or whatever makes the most sense) from `ast::types`.

Nothing implemented here is relevant to `miden-mast-package` itself though - the details are all type signature oriented, and should live as close to that code as possible IMO.

Once this is moved into `miden-assembly-syntax`, it can be merged.

marijamijailovic:
> I think the only thing I would like to change before this is merged is to move this code into `miden-assembly-syntax` under the `ast::types` module, maybe as `ast::types::signatures` or something to that effect, and re-export `TypedProcInfo` (or whatever makes the most sense) from `ast::types`.
> 
> Nothing implemented here is relevant to `miden-mast-package` itself though - the details are all type signature oriented, and should live as close to that code as possible IMO.
> 
> Once this is moved into `miden-assembly-syntax`, it can be merged.

Hi @bitwalker, I have moved `TypedProfInfo` to the `miden-assembly-syntax`.

bitwalker (approved):

bitwalker:
Thanks @marijamijailovic!

huitseeker:
@marijamijailovic Can you make sure every commit in this PR is signed?

marijamijailovic:
> @marijamijailovic Can you make sure every commit in this PR is signed?

Hi @huitseeker, I’ve rebased this branch so that every commit is signed

Issue #3259, opened 2026-06-16
Explore nutype for small checked newtypes


huitseeker:
Several small types in this repo repeat the same pattern. They have a private wrapper, a checked constructor, a small error type, and serde code that must call the same checks. The `nutype` crate may remove some of that code while keeping invalid values out after serde deserialization.

I would start with `ItemIndex` in `crates/assembly-syntax/src/ast/item/index.rs`. It only checks that a `usize` fits in `u16`, then exposes a few simple traits and `Display`. This is a small test case before touching parser or VM types.

`StackInputs` and `StackOutputs` in `core/src/program/stack/inputs.rs` and `core/src/program/stack/outputs.rs` are also strong candidates. Their checked path only rejects more than `MIN_STACK_DEPTH` values, then pads with zeroes. Please keep the current public shape, `Copy` behavior, and binary encoding unless the refactor proves that changing them is harmless.

`Kernel` in `core/src/program/kernel.rs` is a useful second pass. It sorts procedure hashes, rejects duplicates, and caps the length at `MAX_NUM_PROCEDURES`. `nutype` can express the length check and a custom validator, but the sort is a domain rule and should stay explicit.

`SectionId` in `crates/mast-package/src/package/section.rs` is another small string case. It checks the first character, checks the allowed characters, supports constants, and has transparent serde. This looks close to the `nutype` string use case.

`Ident`, `ProcedureName`, and `PathBuf` are worth checking, but they should not be the first step. `Ident` carries a source span and interning behavior. `ProcedureName` has quote parsing rules. `PathBuf` has canonical form, component joining, and path rewrite logic. A refactor should only move the small checked string field into `nutype` if the parser and display behavior stay the same.

The custom Miden binary serde code should stay manual unless a small bridge can call the generated checked constructor. The `nutype` serde support covers `serde`, not the repo specific `Serializable` and `Deserializable` traits.

I would skip `EventName`, `PackageId`, `TargetType`, `Linkage`, `MastNodeId`, and the indexed vector types for now. Some have no validation rule today. Some are enums. Some need context from a forest or container, which does not fit a standalone newtype check.

A good first PR would move only `ItemIndex` or `SectionId`. Keep public behavior and the binary format unchanged. Stop there if the diff is not smaller or clearer.

mgkz0:
Hi, Id like to try this

enlightened88:
I can take this on

enlightened88:
I tried porting ItemIndex to nutype

[...]

PR #3277, opened 2026-06-21
fix: validate SectionId during deserialization

enlightened88:
Section::read_from() constructed SectionId directly via Cow::Owned,
bypassing the validation and canonicalization logic implemented by FromStr.
As a result, deserialization could produce SectionId values that cannot be
constructed through the public API.

Use id_str.parse::<SectionId>() instead so invalid identifiers are rejected
during deserialization and built-in IDs are canonicalized to their borrowed
constants.

### DIFF
--- a/crates/mast-package/src/package/section.rs
+++ b/crates/mast-package/src/package/section.rs
@@ -18,7 +18,7 @@ use serde::{Deserialize, Serialize};
 
 /// A unique identifier for optional sections of the Miden package format
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 #[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
 #[repr(transparent)]
@@ -121,6 +121,17 @@ impl fmt::Display for SectionId {
     }
 }
 
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for SectionId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = alloc::string::String::deserialize(deserializer)?;
+        s.parse::<SectionId>().map_err(serde::de::Error::custom)
+    }
+}
+
 #[derive(Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Section {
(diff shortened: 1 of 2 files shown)


crates/mast-package/src/package/section.rs
```diff
@@ -177,7 +177,9 @@ impl Deserializable for Section {
         let id_str = core::str::from_utf8(id_bytes).map_err(|err| {
             DeserializationError::InvalidValue(format!("invalid utf-8 in section name: {err}"))
         })?;
-        let id = SectionId(Cow::Owned(id_str.to_owned()));
+        let id = id_str.parse::<SectionId>().map_err(|err| {
```
huitseeker, line 191: This fixes the binary package reader, but `SectionId` can still be made invalid through serde when the crate's `serde` feature is on. The derived transparent deserialize path accepts `"1bad"` and builds a value that `SectionId::from_str` would reject. A custom serde deserialize impl that calls `SectionId::from_str` would keep both readers on the same invariant.

huitseeker (changes requested):

crates/mast-package/src/package/section.rs
enlightened88: Fixed, replaced derived Deserialize with a custom impl that delegates to FromStr, and added tests for both paths.

crates/mast-package/src/package/section.rs
```diff
@@ -121,8 +121,19 @@ impl fmt::Display for SectionId {
...
+        D: serde::Deserializer<'de>,
+    {
+        let s = <&str>::deserialize(deserializer)?;
+        s.parse::<SectionId>().map_err(serde::de::Error::custom)
+    }
+}
+
 #[derive(Clone, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize))]
```
huitseeker: This drops serde deserialize for `Section` itself. `SectionId` now validates through its custom impl, but `Section` used to derive `Deserialize` when the `serde` feature was on, and now it only derives `Serialize`. Can this keep `Deserialize` here too so callers can still deserialize full sections through serde?

crates/mast-package/src/package/section.rs
```diff
@@ -121,8 +121,19 @@ impl fmt::Display for SectionId {
     }
 }
 
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for SectionId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = <&str>::deserialize(deserializer)?;
```
huitseeker: This makes serde reads require a borrowed string. For example, `serde_json::from_reader` on a valid `"my_section"` fails with `expected a borrowed string`. Could this deserialize into `String` or `Cow<'de, str>` first, then call `parse::<SectionId>()`?

huitseeker (changes requested):
This might be better addressed through #3259 ?

crates/mast-package/src/package/section.rs
enlightened88: Fixed, restored Deserialize for Section and switched to String::deserialize to avoid the borrowed string requirement.

enlightened88:
> This might be better addressed through #3259 ?

If you'd rather handle this in #3259, I'm happy to do that instead.

huitseeker (approved):
LGTM

enlightened88:
Hey @huitseeker @bitwalker you can merge this commit

PR #3279, opened 2026-06-22
Prevent debug info fuzz OOM

huitseeker:
Debug info deserialization trusted length prefixes before checking the remaining input. A malformed input could make the reader try to allocate a huge string table, which caused the daily fuzz job to run out of memory.

This fixes the issue by checking lengths before allocating debug info strings, byte buffers, and collections. The same bounded reads are used for nested URI and debug variable data.

Regression tests cover the failing fuzz input and the nested string/byte readers.


plafer (approved):
👍

bitwalker (approved):
LGTM

PR #3280, opened 2026-06-22
feat: import midenc-hir-type

huitseeker:
This PR imports midenc-hir-type under crates/midenc-hir-type as a git subtree with its full history.

The crate is now a released workspace crate. It uses the workspace version, currently 0.24.0, so it stays on the same release version as the other crates.

The standalone repo files were removed from the subtree because the workspace already owns those settings. The old midenc-hir-type changelog is now in the workspace changelog as prior history.

The fuzz lockfile now resolves midenc-hir-type to the local 0.24.0 crate.


bitwalker (commented):
I'm not sure I follow why we'd make this crate inherit the workspace version (if anything, I think we need to move away from blanket inheritance of the workspace version for crates that we anticipate will be released on a different schedule than the core VM crates (which IMO is represented by`miden-core`/`miden-processor`/`miden-prover`/`miden-air`). We need to be setting ourselves up for being able to release assembler/package/project-related changes on a more rapid schedule than those core crates, and that will require separate versioning as a prerequisite.

huitseeker:
Added individual release scripts similar to https://github.com/0xMiden/crypto/pull/1066

bitwalker (approved):

bitwalker:
I'd suggest we merge this as soon as it is updated with `next`

PR #3282, opened 2026-06-23
ci: ignore MASM struct field labels in export check

huitseeker:
The release dry run failed because the MASM export check treated struct field names as part of the exported type. That made `struct u256 {u128, u128}` look different from `struct u256 {lo : u128, hi : u128}`.

Those names are debug text: they do not change the type layout, the stack shape, or the MAST root.

This change strips struct field labels before comparing exported types and procedure signatures. The check still fails on changes to field types, field order, field count, struct names, repr attrs, qualified type paths, export digests, auth attrs, and callconv attrs.


bobbinth (approved):
A very light review from me - but looks good! Thank you!

PR #3283, opened 2026-06-23
test: port unentered loop regression to next

huitseeker:
Add the unentered_loop_with_following_block regression from #3278 to the next branch flow-control tests.

The test passes on next without the main-side processor replay fix, proving #3278 does not need to be back-ported to next.


bobbinth (approved):
Looks good! Thank you!

Just to confirm: this also means that the issue raised in https://github.com/0xMiden/protocol/issues/3113 is addressed - right?

huitseeker:
@bobbinth Yes, that means this issue is already structural fixed in the upcoming version of the VM

PR #3284, opened 2026-06-23
feat: serialize trusted trace proving inputs

### LINKED ISSUE #3235: Implement serialization for `TraceProvingInputs`
With #2730 we added the `SparseMastForest`, which gets built by the `ExecutionTracer` and included in the `TraceGenerationContext`, but never settled on a serialization format. I believe @huitseeker had some ideas there. AFAIU this is the biggest missing piece before we can make all of `TraceProvingInputs` serializable.

Additional context: https://github.com/0xMiden/protocol/pull/3035#pullrequestreview-4433715082

huitseeker:
The transaction executor can now serialize `TraceProvingInputs` and send them to a trusted prover. The prover can deserialize the payload, build the trace, prove it, and return a proof.

This adds binary serialization for the trace replay state, sparse MAST replay data, `TraceBuildInputs`, `ProvingOptions`, and `TraceProvingInputs`.

Sparse MAST serialization is a trusted replay format. It preserves the source node IDs, roots, advice data, digest entries, and source forest commitment. It does not implement the untrusted hashless MAST read path (and does not verify node hashes).

Tests cover sparse round trips, malformed sparse payloads, bad serialized forest IDs, trace summary equality after `TraceBuildInputs` round trip, proof generation after `TraceProvingInputs` round trip, and fuzz smoke runs for the new deserialize targets.

Closes #3235


core/src/mast/mod.rs
```diff
@@ -838,7 +839,10 @@ impl<T: ExecutableMastForest + ?Sized> ExecutableMastForest for Arc<T> {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
-#[cfg_attr(all(feature = "arbitrary", test), miden_test_serde_macros::serde_test)]
+#[cfg_attr(
+    all(feature = "arbitrary", test),
+    miden_test_serde_macros::serde_test(binary_serde(true))
+)]
```
plafer, line 845: All the `serde_test` require the `arbitrary` feature to be turned on, but `make test` doesn't turn it on (since `WORKSPACE_TEST_FEATURES` doesn't include that feature). So those tests aren't run in CI I believe. Is this correct/intended?

core/src/mast/serialization/tests.rs
```diff
@@ -679,6 +680,489 @@ fn test_untrusted_hashless_keeps_external_digests_by_node_index() {
...
+        &basic_block_data,
+        expected_root_digest,
+    );
+
+    let restored = SparseMastForest::read_from_bytes(&bytes).unwrap();
+    assert_eq!(restored.get_digest_by_id(root), Some(expected_root_digest));
+}
+
+#[test]
+fn sparse_reader_rejects_trailing_bytes_with_exact_prefix_budget() {
```
plafer, line 948: Both `sparse_reader_rejects_trailing_bytes_with_exact_prefix_budget` and the trailing-byte assertion in `test_trace_proving_inputs_round_trip_...` pass `budget = valid_prefix_len` against a payload of `prefix_len + 1`. That trips the `budget < bytes.len()` branch ("budget is smaller than payload length"), **not** the `has_more_bytes()` trailing-byte check. The `has_more_bytes()` rejection (the one that matters for the default `read_from_bytes` path) is untested.

plafer (approved):
Overall really LGTM!

huitseeker:
@plafer Thanks, both points were right.
- I made `testing` enable `arbitrary` (as it does in core) for every crate that has arbitrary backed serde tests. `make test` now builds and runs those generated tests.
- I also changed both trailing byte tests so the budget covers the full invalid payload. They now reach the trailing byte checks instead of the budget check.

bobbinth:
> Sparse MAST serialization is a trusted replay format. It preserves the source node IDs, roots, advice data, digest entries, and source forest commitment. It does not implement the untrusted hashless MAST read path (and does not verify node hashes).

Question: what would happen if someone sends a malformed MAST forest (i.e., with incorrect hashes) - would the prover error out or crush?

The reason for the question is that we have two separate settings for this:

- Block/batch proof generation - where we trust that the block producer is not malicious and so a remote prover doesn't need to verify the inputs.
- Transaction proof generation - here, any external user could send a request to prove a transaction to a remote prover. So, in theory, the proving inputs could be serialized incorrectly. Though, maybe this is something that could be caught at a higher level because we do need to enforce some limits (or collect some fees) from users who'd like to use remote provers.

huitseeker:
@bobbinth 
> Question: what would happen if someone sends a malformed MAST forest (i.e., with incorrect hashes) - would the prover error out or crush?

The prover would not crash. The bad digest is accepted into the sparse forest. `build_trace` may still succeed, because trace replay uses those stored digests. The STARK proof path does not call `ExecutionTrace::check_constraints()` before proving The likely developer-visible outcome is an invalid proof that fails verification, or a constraint failure only if they explicitly run `check_constraints()`.

From the PR description:

> > Sparse MAST serialization is a trusted replay format. It preserves the source node IDs, roots, advice data, digest entries, and source forest commitment. It does not implement the untrusted hashless MAST read path (and does not verify node hashes).

The closest signal for this scope is code documentation, not an API boundary strong enough to prevent misuse.

I opened https://github.com/0xMiden/miden-vm/issues/3303 for the untrusted serialization.

prover/src/lib.rs
```diff
@@ -52,6 +66,57 @@ impl TraceProvingInputs {
...
+    pub fn read_from_bytes_with_budget(
+        bytes: &[u8],
+        budget: usize,
+    ) -> Result<Self, SerdeDeserializationError> {
+        if budget < bytes.len() {
+            return Err(SerdeDeserializationError::InvalidValue(
+                "TraceProvingInputs byte budget is smaller than payload length".into(),
+            ));
+        }
+        let allocation_budget = budget.min(bytes.len().saturating_mul(4));
```
bobbinth, line 83: Would be good to add a comment why multiplying by 4 is appropriate/sufficient here.

prover/src/lib.rs
```diff
@@ -52,6 +66,57 @@ impl TraceProvingInputs {
...
+        let mut reader = BudgetedReader::new(SliceReader::new(bytes), allocation_budget);
+        let inputs = Self::read_from(&mut reader)?;
+        if reader.has_more_bytes() {
+            return Err(SerdeDeserializationError::InvalidValue(
+                "TraceProvingInputs payload has trailing bytes".into(),
+            ));
+        }
```
bobbinth, line 90: Do we need this check? Having it implies that we wouldn't be able to serialize `TraceProvingInputs` as a part of some bigger object, right?

processor/src/trace/chiplets/ace/trace.rs
```diff
@@ -289,3 +306,395 @@ impl WireBus {
...
+                "ACE wire bus count {} does not match read/eval node wire count {expected_wires}",
+                self.wire_bus.num_wires
+            )));
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(feature = "arbitrary")]
```
bobbinth, line 514: nit: I'd add a section separator before these lines to indicate that the code below is used for testing.

processor/src/continuation_stack.rs
```diff
@@ -341,6 +361,187 @@ impl<F> ContinuationStack<F> {
...
+impl Deserializable for Continuation<MastForestId> {
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        match u8::read_from(source)? {
+            0 => Ok(Self::StartNode(MastNodeId::read_from(source)?)),
+            1 => Ok(Self::FinishJoin(MastNodeId::read_from(source)?)),
+            2 => Ok(Self::FinishSplit(MastNodeId::read_from(source)?)),
+            3 => Ok(Self::FinishLoop(MastNodeId::read_from(source)?)),
+            4 => Ok(Self::FinishCall(MastNodeId::read_from(source)?)),
+            5 => Ok(Self::FinishDyn(MastNodeId::read_from(source)?)),
+            6 => Ok(Self::ResumeBasicBlock {
```
bobbinth, line 435: Probably fine for now, but ideally, we'd have the tag values defined as constants.

processor/src/trace/trace_state.rs
bobbinth, line 1: This file has grown too large - let's split it up and move the content under `src/trace/trace_state` directory. I would split it up along the component lines. For example:

- `src/trace/trace_state/system.rs`
- `src/trace/trace_state/stack.rs`
- `src/trace/trace_state/block_stack.rs`
- etc.

core/src/mast/serialization/mod.rs
```diff
@@ -171,10 +179,16 @@ const MAGIC: &[u8; 4] = b"MAST";
 /// from local structure.
 pub(super) const FLAG_HASHLESS: u8 = 0x02;
 
+/// Flag indicating that the payload uses sparse MAST replay serialization.
+///
+/// Sparse payloads preserve the source forest's [`MastNodeId`] space and therefore cannot be read
+/// through dense [`MastForest`] entry points.
+pub(super) const FLAG_SPARSE: u8 = 0x04;
```
bobbinth, line 186: Question: what's the benefit of combining regular and sparse MAST forest serializations? Could we treat sparse MAST forest as a completely separate object with its own serialization? I think the use cases for these are fairly distinct and I don't foresee the need to serialize sparse MAST forest outside of the `TraceProvingInputs`.

processor/src/trace/mod.rs
bobbinth, line 63: Not related to this PR, but I find the structure of `TraceBuildInputs` and related structs pretty confusing. For example:

Do we actually need `TraceBuildOutput`? It seems like 3 out of 4 fields there are about precompiles and probably should be put into a dedicated struct. And once this is done, maybe it would make sense to dissolve this struct entirely?

Also, `TraceBuildOutput` doesn't need to be `pub(crate)` as it is used only within this module (and its submodules).

processor/src/trace/execution_tracer.rs
```diff
@@ -79,6 +86,136 @@ pub struct TraceGenerationContext {
...
+        };
+        validate_trace_generation_context_invariants(&context)?;
+        validate_trace_generation_context_forest_ids(&context)?;
+        Ok(context)
+    }
+}
+
+fn validate_trace_generation_context_invariants(
+    context: &TraceGenerationContext,
+) -> Result<(), DeserializationError> {
```
bobbinth, line 142: nit: why not not make this a method on `TraceGenerationContext`?

processor/src/trace/execution_tracer.rs
```diff
@@ -79,6 +86,136 @@ pub struct TraceGenerationContext {
...
+                context.max_stack_depth
+            )));
+        }
+    }
+    Ok(())
+}
+
+fn validate_trace_generation_context_forest_ids(
+    context: &TraceGenerationContext,
+) -> Result<(), DeserializationError> {
```
bobbinth, line 168: Same nit as above.

core/src/lib.rs
```diff
@@ -73,6 +75,53 @@ pub mod serde {
             err => err,
         })
     }
+
+    /// Serializable view over a [`VecDeque`].
+    ///
+    /// This uses the same wire shape as `Vec<T>`: a length prefix followed by items in iteration
+    /// order.
+    pub struct SerializableVecDeque<'a, T>(pub &'a VecDeque<T>);
```
bobbinth, line 83: We are doing this because we don't have have serialization implemented for `VecDeque` in `miden-crypto`? If so, should we just implement it it there?

If we'd rather keep it here for now, I'd move it `src/utils/mod.rs`.

core/src/mast/mod.rs
```diff
@@ -907,6 +911,24 @@ impl From<MastNodeId> for u32 {
...
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        Serializable::write_into(&self.0, target);
+    }
+}
+
+impl Deserializable for MastNodeId {
+    fn read_from<R: crate::serde::ByteReader>(
+        source: &mut R,
+    ) -> Result<Self, DeserializationError> {
+        Ok(Self(<u32 as Deserializable>::read_from(source)?))
```
bobbinth, line 924: I'm not sure this is safe. I remember previously we've always avoided deserializing MAST node IDs directly and preferred using `from_u32_with_node_count()`. I believe this was because MAST node ID values were limited to $2^{30} - 1$ and so a full u32 value could result in an invalid MAST node ID.

If this is not a concern now, we should explicitly document this. But also, I'm a bit weary of adding this serialization option as it could be easily misused.

bobbinth (commented):
Thank you! Looks good. Not a full review but I left some comments inline.

huitseeker:
Closing this version because the change is too large as one PR. It has been replaced by three smaller PRs:

- https://github.com/0xMiden/miden-vm/pull/3313
- https://github.com/0xMiden/miden-vm/pull/3314
- https://github.com/0xMiden/miden-vm/pull/3315

These keep the same code changes, split by scope.

core/src/mast/serialization/mod.rs
huitseeker: Addressed in #3313: Sparse payloads now use their own magic and version, and has its own writers/readers.

processor/src/trace/chiplets/ace/trace.rs
huitseeker: Addressed in #3314.

processor/src/continuation_stack.rs
huitseeker: Addressed in #3314

processor/src/trace/trace_state.rs
huitseeker: I agree that `trace_state.rs` is large. I have found large mechanical diffs have a negative impact on some reviewers, so I would rather keep that split out of this stack. Happy to have a component split as a separate cleanup PR.

prover/src/lib.rs
huitseeker: The check is in `read_from_bytes()` because that function reads one full standalone payload. A larger wrapper should call `TraceProvingInputs::read_from(reader)` inside its own reader, then the wrapper owns the final end-of-input check. #3314 documents that, and it is the same pattern used e.g. by `serde_json` or  `bincode` for whole-slice vs reader-based deserialization.

processor/src/trace/mod.rs
huitseeker: #3314 makes `TraceBuildOutput` private. I would leave the larger shape change for a follow-up because it is not needed for serialization.

processor/src/trace/execution_tracer.rs
huitseeker: Addressed in #3314

processor/src/trace/execution_tracer.rs
huitseeker: #3314 moves this into `TraceGenerationContext::validate_invariants()`.

core/src/lib.rs
huitseeker: #3314 moves the `VecDeque` helper to `core::utils`. I did not add it to `miden-crypto` for now because (besides release propagation delays) the only current caller is trace replay.

core/src/mast/mod.rs
huitseeker: Addressed in #3313/#3314:
- #3313 keeps `MastNodeId` serializable but removes the blanket `Deserializable` impl,
- The sparse reader reads raw IDs only [through a local helper that checks the source node count](https://github.com/0xMiden/miden-vm/blob/e2efae93b1c20159f6f7c816d78c79d45640c8ba/core/src/mast/serialization/sparse.rs#L261-L283)
- #3314 uses local trace replay readers instead of a global `MastNodeId::read_from`,

prover/src/lib.rs
huitseeker: Addressed in #3314, which names the multiplier and documents the split between the byte budget and the allocation budget. It's the same pattern as the `MastForest`, `UntrustedMastForest`, `Package`, etc.

PR #3285, opened 2026-06-24
Merge `main` back into `next`

bobbinth:
This PR merges the current `main` into `next`. The main change is that it discards changes from https://github.com/0xMiden/miden-vm/pull/3278 as this is fixed in `next` in other ways (see https://github.com/0xMiden/miden-vm/pull/3283).


Al-Kindi-0:
Just to be precise, I think that the work on the do-while loop semantics removed the buggy part and hence made the fix unnecessary.

Al-Kindi-0 (approved):
LGTM (assuming this has been reviewed previously)

crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -403,80 +406,88 @@ fn register_debug_type(
...
 #[cfg(test)]
 mod tests {
+    use alloc::sync::Arc;
+
     use super::*;
-    use crate::ast::types::{EnumType, Variant};
+    use crate::ast::types::{CallConv, FunctionType};
 
     #[test]
-    fn registers_c_like_enum_debug_type() {
```
bitwalker, line 412: It appears these tests are being dropped when they are not present in `main`, but are present in `next` - there are two sets of tests (one set that is in `main` which aren't in `next`, and one set that are in `next`, but not in `main`) - we want the union of those sets.

bitwalker (changes requested):
Looks like there is one issue that will needed to be addressed before this can be merged

crates/assembly/src/assembler/debuginfo.rs
bobbinth: Thank you for catching this! Should be addressed in 2a0f021fc3066a313702d4377ce18af551ce1996.

bitwalker (approved):

PR #3289, opened 2026-06-25
Reduce processor compile time

### LINKED ISSUE #3271: Measure and reduce processor generic compile cost
The processor execution loop has a broad generic shape. `execute_impl` has type parameters for the VM state, stopping logic, trace recording, MAST storage, and host access. Most node helpers carry the same type parameters, and many of them use `#[inline(always)]`.

Rust usually builds a separate copy of generic code for each concrete type. Here, that can duplicate the same interpreter logic across `FastProcessor`, the replay processor, tracing and no tracing, different stoppers, forest types, and host types.

This cost shows up late in the dependency graph. `miden-test-utils`, `miden-vm`, and several test binaries depend on `miden-processor`, so they can pay for optimized processor code after they pull in the broad processor surface.

#3181 found the first practical workaround for cold test builds. #3221 applied it in commit `221fa59109e150947ff84a51edef20111cd33044`. It keeps `miden-processor` at `opt-level = 0` in `test-dev` and raises `TEST_RUST_MIN_STACK` for nextest. See https://github.com/0xMiden/miden-vm/issues/3181#issuecomment-4742263239.

That workaround hides a large compile-time cost. I measured the same build-only nextest shape in two clean target directories on local branch `issue-3263-serialization-cleanup` at `470500095f`. The adjusted run used the current settings. The comparison run used the same `test-dev` profile, but removed only the `miden-processor` package override and did not set `RUST_MIN_STACK`.

| Build | Total time | Wall time |
| --- | ---: | ---: |
| Current adjusted build | 260.8s | 4m 20s |
| Without the processor opt-level override | 1419.6s | 23m 39s |

The build without the processor override was about 5.4 times slower. This was a build-only run, so `RUST_MIN_STACK` does not explain the difference. The compile-time difference comes from optimizing `miden-processor` and crates that depend on it.

The largest changed units were:

| Unit | Current adjusted | Without override |
| --- | ---: | ---: |
| `miden-processor` test lib | 18.9s | 870.1s |
| `miden-processor` `async_compat` test | 10.5s | 523.8s |
| `miden-processor` lib | 10.4s | 116.2s |
| `miden-test-utils` lib | 177.0s | 507.0s |
| `miden-prover` `async_compat` test | 183.3s | 540.0s |
| `miden-vm-synthetic-bench` lib | 104.5s | 261.9s |

The open question is whether we can reduce the generic surface enough to remove or narrow the `miden-processor` override.

One experiment is to remove the host type parameter from the inner execution state and node helpers where only `BaseHost` is needed. The host is used for source-rich errors and for operations that leave the main loop. It is not part of normal arithmetic execution. Passing `&mut dyn BaseHost` in those inner paths could remove one generic axis. This needs a benchmark because each host call would use a vtable lookup.

A second experiment is to make stopping concrete for `FastProcessor`. Full execution uses the never-stop mode. Step execution uses the step mode. We can replace `S: Stopper` in the fast path with runtime config, such as a small local mode that says whether this run should stop after each step. The replay processor can keep its own stopper path.

A third experiment is to audit `#[inline(always)]` on high-level execution helpers. Small operation helpers may still need forced inline. Node walkers and control-flow helpers should only force inline when a benchmark shows a clear win.

This work must not slow down normal execution. `program_execution_fast/blake3_1to1` should be the main guardrail. Any patch that changes processor generics, dispatch, or inline hints should show that this benchmark does not regress. A compile-time win is not enough if it slows this benchmark down.

A good first patch would:

- Change the inner host parameter from a generic `H: BaseHost` to `&mut dyn BaseHost` where possible.
- Keep public host APIs unchanged.
- Measure the adjusted and no-override build-only timings before and after.
- Measure `program_execution_fast/blake3_1to1` before and after, and do not land the patch if it regresses.
- Keep the `miden-processor` profile override until timings show it is no longer needed.

A follow-up patch can handle stopper concretization and the inline audit separately.

huitseeker:
On top of #3274.

This PR reduces compile time for the processor benchmark path in #3271. The root cause is forced inlining in large generic execution helpers. This change keeps forced inlining on the pure replay path and on the inner operation loop. It changes the source aware path and the basic block helpers to `#[inline]`, so LLVM can decide when to inline them.

The direct Blake3 comparison uses the parent of this work as the base and the current branch head as the tip. The run used all four Blake3 axes, with 10 proof samples, 1000 light samples, 3 seconds of measurement time, and 1 second of warmup.

Base: `1e91767e90ab9e4f8df08991a8ee174b4574b995`.

Tip: `30dee80d26f928213d9a1381c9ccd99901baed36`.

Criterion results:

* `e2e_prove` improved from `9187.2998497 ms` to `9086.9550542 ms`. This is `1.09%` faster.
* `prove_trace_sync` changed from `9288.9703209 ms` to `9573.5592793 ms`. This is `3.06%` slower in this run.
* `execute_trace_inputs_sync` improved from `18.111430051 ms` to `8.24616859 ms`. This is `54.47%` faster.
* `build_trace` improved from `108.006259741 ms` to `56.828919122 ms`. This is `47.38%` faster.

~Proof spans also improved:~ (obsolete, see [below](https://github.com/0xMiden/miden-vm/pull/3289#issuecomment-4806404033))

~* `prove_program_sync` wall time improved from `13186.812667 ms` to `8400.075417 ms`. This is `36.30%` faster.~
~* `build aux traces` improved from `51.48075 ms` to `37.603667 ms`. This is `26.96%` faster.~
~* `commit to aux traces` improved from `1003.236458 ms` to `629.976166 ms`. This is `37.21%` faster.~
~* `commit to main traces` improved from `3460.164125 ms` to `2065.563666 ms`. This is `40.30%` faster.~
~* `commit to quotient poly chunks` improved from `1070.276875 ms` to `715.463625 ms`. This is `33.15%` faster.~
~* `evaluate constraints` improved from `6814.079292 ms` to `4449.177041 ms`. This is `34.71%` faster.~
~* `open` improved from `680.16025 ms` to `408.692417 ms`. This is `39.91%` faster.~

Build time at the final branch tip:

* The adjusted `test-dev` build for the processor test command took `1m18s`, with `real 79.12s`, `user 537.61s`, and `sys 32.76s`.
* The same build without the `miden-processor` profile override took `2m06s`, with `real 127.23s`, `user 723.99s`, and `sys 35.23s`.
* The build without the profile override is still about `1.61x` slower, so this PR keeps the profile override.

Earlier branch measurements show the compile time change across the work:

* The Blake3 benchmark build improved from `5m00s` to `3m46s`.
* The optimized runner build improved from `11m02s` to `3m27s`.
* The adjusted processor test build improved from `6m04s` to `1m18s`.
* The same processor test build without the profile override improved from `24m34s` to `2m06s`.

The code changes are small:

* The source aware execution path is separate from the pure execution loop.
* `NoopHost` trace replay still uses static dispatch.
* Source aware node handlers use `#[inline]` instead of `#[inline(always)]`.
* Basic block helper functions use `#[inline]` instead of `#[inline(always)]`.
* The inner operation batch helper and the pure replay helpers still use `#[inline(always)]`.

The `finalize_clock_cycle` annotations are unchanged because changing them did not improve compile time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v0.25.0 (TBD)
 
+#### Changes
+
+- Added a Blake3 pure execution benchmark axis and reduced processor benchmark compile time by relaxing forced inlining in execution helpers ([#3289](https://github.com/0xMiden/miden-vm/pull/3289)).
+
 ## v0.24.0 (2026-06-24)
 
 #### Enhancements
--- a/benches/blake3-bench/benches/blake3_bench.rs
+++ b/benches/blake3-bench/benches/blake3_bench.rs
@@ -128,7 +128,10 @@ fn blake3_bench(c: &mut Criterion) {
             group.bench_function("execute_sync", |b| {
                 b.iter_batched(
                     || fixture.clone(),
-                    |fixture| black_box(execute_program(&fixture)),
+                    |fixture| {
+                        execute_program(&fixture);
+                        black_box(())
+                    },
                     BatchSize::SmallInput,
                 );
             });
--- a/benches/blake3-bench/benches/blake3_bench.rs
+++ b/benches/blake3-bench/benches/blake3_bench.rs
@@ -3,14 +3,19 @@ use std::{hint::black_box, time::Duration};
 use codspeed_criterion_compat as criterion;
 use criterion::{BatchSize, Criterion, SamplingMode, criterion_group, criterion_main};
 use miden_vm_blake3_bench::{
-    BENCH_GROUP, Blake3Fixture, build_trace, execute_trace_inputs, prove_and_verify_once,
-    prove_span_duration, prove_trace, repo_root_from_manifest,
+    BENCH_GROUP, Blake3Fixture, build_trace, execute_program, execute_trace_inputs,
+    prove_and_verify_once, prove_span_duration, prove_trace, repo_root_from_manifest,
 };
 
-const ALL_AXES: [&str; 4] =
-    ["execute_trace_inputs_sync", "build_trace", "prove_trace_sync", "e2e_prove"];
+const ALL_AXES: [&str; 5] = [
+    "execute_sync",
+    "execute_trace_inputs_sync",
+    "build_trace",
+    "prove_trace_sync",
+    "e2e_prove",
+];
 const PROOF_AXES: [&str; 2] = ["e2e_prove", "prove_trace_sync"];
-const LIGHT_AXES: [&str; 2] = ["execute_trace_inputs_sync", "build_trace"];
+const LIGHT_AXES: [&str; 3] = ["execute_sync", "execute_trace_inputs_sync", "build_trace"];
 
 fn env_usize(name: &str, default: usize) -> usize {
     match std::env::var(name) {
@@ -49,6 +54,7 @@ fn resolve_axes() -> Vec<&'static str> {
         match axis {
             "all" => requested.extend(ALL_AXES),
             "e2e_prove" | "prove" | "prove_program_sync" => requested.push("e2e_prove"),
+            "execute_sync" => requested.push("execute_sync"),
             "execute_trace_inputs_sync" => requested.push("execute_trace_inputs_sync"),
             "prove_trace_sync" => requested.push("prove_trace_sync"),
             "build_trace" => requested.push("build_trace"),
@@ -117,6 +123,16 @@ fn blake3_bench(c: &mut Criterion) {
     if LIGHT_AXES.iter().any(|axis| has_axis(&axes, axis)) {
         let mut group = c.benchmark_group(BENCH_GROUP);
         configure_group(&mut group, light_sample_size, measurement_time_secs, warm_up_time_secs);
+
+        if has_axis(&axes, "execute_sync") {
+            group.bench_function("execute_sync", |b| {
+                b.iter_batched(
+                    || fixture.clone(),
+                    |fixture| black_box(execute_program(&fixture)),
+                    BatchSize::SmallInput,
+                );
+            });
+        }
 
         if has_axis(&axes, "execute_trace_inputs_sync") {
             group.bench_function("execute_trace_inputs_sync", |b| {
--- a/benches/blake3-bench/src/bin/blake3_nonregression/criterion_results.rs
+++ b/benches/blake3-bench/src/bin/blake3_nonregression/criterion_results.rs
@@ -57,10 +57,16 @@ pub(crate) fn collect_result(
 fn select_primary_metric(
     metrics: &BTreeMap<String, Metric>,
 ) -> Result<&str, Box<dyn std::error::Error>> {
-    [PRIMARY_METRIC, "prove_trace_sync", "build_trace", "execute_trace_inputs_sync"]
-        .into_iter()
-        .find(|metric| metrics.contains_key(*metric))
-        .ok_or_else(|| "no supported primary metric found in benchmark result".into())
+    [
+        PRIMARY_METRIC,
+        "prove_trace_sync",
+        "build_trace",
+        "execute_trace_inputs_sync",
+        "execute_sync",
+    ]
+    .into_iter()
+    .find(|metric| metrics.contains_key(*metric))
+    .ok_or_else(|| "no supported primary metric found in benchmark result".into())
 }
 
 fn collect_criterion_metrics(
--- a/benches/blake3-bench/src/lib.rs
+++ b/benches/blake3-bench/src/lib.rs
@@ -133,6 +133,19 @@ pub fn execute_trace_inputs(fixture: &Blake3Fixture) -> TraceBuildInputs {
     }
 }
 
+pub fn execute_program(fixture: &Blake3Fixture) {
+    let mut host = host_for_fixture(fixture);
+    let processor = FastProcessor::new_with_options(
+        fixture.stack_inputs,
+        fixture.advice_inputs.clone(),
+        execution_options(),
+    )
+    .expect("processor advice inputs should fit advice map limits");
+    processor
+        .execute_sync(&fixture.program, &mut host)
+        .expect("failed to execute Blake3 benchmark");
+}
+
 pub fn prove_program(fixture: &Blake3Fixture) {
     let _span = tracing::info_span!("prove_program_sync").entered();
     prove_trace(execute_trace_inputs(fixture));
--- a/processor/src/errors.rs
+++ b/processor/src/errors.rs
@@ -383,7 +383,7 @@ impl OperationError {
     pub fn with_package_source_context(
         self,
         context: PackageSourceDebugContext<'_>,
-        host: &impl BaseHost,
+        host: &(dyn BaseHost + '_),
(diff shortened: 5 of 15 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,13 @@
 
 #### Changes
 
+- Bumped MSRV from 1.90 to 1.96 and replaced local `assert_matches!` macro with `core::assert_matches` ([#3267](https://github.com/0xMiden/miden-vm/pull/3267)).
+- [BREAKING] Unified `OnceLockCompat` behavior across `std` and `no_std` ([#3188](https://github.com/0xMiden/miden-vm/pull/3188)).
+- Added an event-based `miden::core::debug` module providing `print_*` procedures for print-style debugging of the operand stack, memory, advice stack, and advice map ([#3169](https://github.com/0xMiden/miden-vm/issues/3169)).
+- [BREAKING] Removed `debug.*` decorators in favor of `miden::core::debug` procedures, and bumped the MAST wire format to `0.0.4` ([#3201](https://github.com/0xMiden/miden-vm/pull/3201)).
+- Added `miden::core::debug::print_mem_addr` for printing a single memory cell (combine with `locaddr` to print a procedure local) ([#3203](https://github.com/0xMiden/miden-vm/issues/3203)).
+- Improved performances of auxiliary trace generation ([#3119](https://github.com/0xMiden/miden-vm/pull/3119)).
+- Reduced processor benchmark compile time by relaxing forced inlining in execution helpers ([#3289](https://github.com/0xMiden/miden-vm/pull/3289)).
```
bobbinth: Seems like this is adding some extra entries.

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@
 
 #### Changes
 
+- Bumped MSRV from 1.90 to 1.96 and replaced local `assert_matches!` macro with `core::assert_matches` ([#3267](https://github.com/0xMiden/miden-vm/pull/3267)).
+- [BREAKING] Unified `OnceLockCompat` behavior across `std` and `no_std` ([#3188](https://github.com/0xMiden/miden-vm/pull/3188)).
```
plafer: Do we expect this PR to have all these other entries?

plafer (approved):
👍

bobbinth (approved):
Looks good! Thank you!

One question: how did this affect pure execution time - i.e., when we want to execute a program as quickly as possible and don't crate about trace generation or debugging?

huitseeker:
Will fix the Changelog. @bobbinth the question you asked is answered in the PR summary (and in commit messages)

bobbinth:
> Will fix the Changelog. @bobbinth the question you asked is answered in the PR summary (and in commit messages)

Hmmm - I'm not seeing it in the commit summary. The closest I'm seeing is `execute_trace_inputs_sync` but AFAIK, this also collects the data needed for trace generation. I'm looking for pure execution which should be on the order of 2ms - 3ms. Am I missing something?

huitseeker:
I added a Blake3 Criterion axis for the pure execution path, `FastProcessor::execute_sync`, so this now measures execution without trace generation, proof generation, or debug execution.
With Criterion sample size 100, 3s measurement, and 1s warmup:
- Baseline, `origin/next` plus only the new benchmark axis: `2.6462988908 ms`, CI `[2.6407266442, 2.6520347295]`.
- This PR’s code on top of the same benchmark axis: `2.5792452367 ms`, CI `[2.5663538842, 2.5938210681]`.

That is `2.53%` faster, so this does not regress pure execution time in this measurement.

bobbinth:
> Proof spans also improved:
> 
> * `prove_program_sync` wall time improved from `13186.812667 ms` to `8400.075417 ms`. This is `36.30%` faster.
> * `build aux traces` improved from `51.48075 ms` to `37.603667 ms`. This is `26.96%` faster.
> * `commit to aux traces` improved from `1003.236458 ms` to `629.976166 ms`. This is `37.21%` faster.
> * `commit to main traces` improved from `3460.164125 ms` to `2065.563666 ms`. This is `40.30%` faster.
> * `commit to quotient poly chunks` improved from `1070.276875 ms` to `715.463625 ms`. This is `33.15%` faster.
> * `evaluate constraints` improved from `6814.079292 ms` to `4449.177041 ms`. This is `34.71%` faster.
> * `open` improved from `680.16025 ms` to `408.692417 ms`. This is `39.91%` faster.

This part seems a bit strange. AFAICT, the changed code affects only execution/trace generation - but it shouldn't affect proving time by that much. Also, this seems to conflict with the `e2e_prove` benchmark. Basically, how can almost all parts of trace generation improve by 30%+ but `e2e_prove` improves only by 1%? Any ideas why we are measuring such large improvements in individual proof generation steps?

huitseeker:
@bobbinth Oh, these numbers came from traced subspans inside proving — in an old version of the branch, I used `tokio-tracing` spans to measure time and just didn't edit the PR summary. 

The code changes here affect execution and trace generation. They do not change the proof.

I reran this with a deterministic script that compares two commits directly. The script creates a detached worktree for each commit, uses a fresh `CARGO_TARGET_DIR`, disables `RUSTC_WRAPPER`, sets `CARGO_INCREMENTAL=0`, and keeps only the Cargo registry cache. It records cold CodSpeed build time and all Blake3 Criterion axes. It runs on a commit-per commit basis and also on the base-to-tip span.

Full output and the script are here:

https://gist.github.com/huitseeker/7a4ffa519fcbe1a0d71fd14bca3b713f

The full load-bearing change is `a15662e98f79` to `ae1669bd8eba`. This omits the benchmark-only commit and the Changelog commit.

| Metric | Base | Tip | Change |
|---|---:|---:|---:|
| Cold CodSpeed build | 365.944 s | 164.059 s | 201.885 s faster, 55.17% |
| `execute_sync` | 2.653 ms | 2.485 ms | 0.167 ms faster, 6.31% |
| `execute_trace_inputs_sync` | 8.564 ms | 8.001 ms | 0.563 ms faster, 6.58% |
| `build_trace` | 55.519 ms | 56.342 ms | 0.823 ms slower, 1.48% |
| `prove_trace_sync` | 8851.048 ms | 8819.261 ms | 31.787 ms faster, 0.36% |
| `e2e_prove` | 8904.801 ms | 8830.666 ms | 74.135 ms faster, 0.83% |

The reliable read is that this PR cuts cold benchmark build time by about `55%`. It does not show a Blake3 execution regression in this run. The only slower full axis is `build_trace`, and it moves by `1.48%`.

The older PR summary had two kinds of numbers. The full Criterion axes were useful, but they came from an earlier comparison setup. The proof subspan numbers were not a sound way to explain this PR. I will update the PR summary to cite the deterministic full-axis comparison instead.

Across the three load-bearing commits, the build-time win comes mostly from the basic block helper change:

| Step | Cold build change | Main runtime note |
|---|---:|---|
| Processor benchmark compile cost | 7.742 s faster, 2.12% | `build_trace` was 16.98% slower in this step. |
| Source-aware execution inlining | 2.059 s faster, 0.57% | `build_trace` recovered 11.93%. Proof axes moved by less than 1%. |
| Basic block orchestration inlining | 192.085 s faster, 53.93% | `execute_sync` was 2.32% slower in this step, but the overall branch still has `execute_sync` 6.31% faster than the base. |

PR #3290, opened 2026-06-25
ci: skip MASM root check across pre-1.0 release lines

huitseeker:
Make the MASM root change script follow cargo's semver conventions for a release prior to 1.0


bobbinth (approved):
Looks good! Thank you!

PR #3293, opened 2026-06-26
Import precompiles prover into VM workspace

adr1anh:
Depends on #3222.

## Summary

Imports the filtered `precompile-experiments` prover history into the VM workspace as `precompiles-prover` / `miden-precompiles-prover`, then adapts it to the VM-owned deferred precompile tags and types from #3222.

The immediate integration target remains intentionally narrow: Keccak-only synthetic deferred-state proving where the precompile proof public root equals `DeferredState.root`.

## What changed

- Adds `precompiles-prover` as a VM workspace member.
- Uses VM workspace `miden-air`, `miden-core`, `miden-crypto`, and `miden-precompiles` instead of sibling or mirrored VM dependencies.
- Ports the prover crate to the VM p3 0.6 / `miden-lifted-*` 0.27 dependency line.
- Sources framework and Keccak deferred tags from `miden-core` / `miden-precompiles`.
- Adds a synthetic Keccak `DeferredState` path with generic chunk nodes plus `Keccak256Precompile::assert_node(len_bytes, input_digest, expected_digest)`.
- Tests that prover public roots match synthetic `DeferredState.root`, including a proof/verify path.

## Imported history note

The prover was imported with history using a subdirectory-filtered import rather than squashing. The most relevant post-main prover integration commits are the semantic pinning notes, chunk-native Keccak digest commitments, VM deferred tag alignment, and the synthetic Keccak root equality tests.

## Validation

Previously run during integration:

- `cargo check --manifest-path vm/Cargo.toml -p miden-precompiles-prover`
- `cargo check --manifest-path vm/Cargo.toml -p miden-precompiles-prover --examples`
- `cargo test --manifest-path vm/Cargo.toml -p miden-precompiles-prover deferred_state -- --nocapture`
- `cargo test --manifest-path vm/Cargo.toml -p miden-precompiles-prover deferred_tags -- --nocapture`

## Follow-up work

- Compare the synthetic/prover root against a VM/MASM-produced Keccak `DeferredState.root`.
- Wire the full VM execution proof plus precompile proof path.
- Defer EC/uint public semantic migration, curve encoding, and SHA-512 prover support.


precompiles-prover/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+miden-lifted-stark = { version = "0.27", features = ["testing"] }
+miden-stateful-hasher = "0.27"
+p3-air = "0.6.1"
+p3-challenger = "0.6.1"
+p3-dft = "0.6.1"
+p3-field = "0.6.1"
+p3-goldilocks = "0.6.1"
+p3-matrix = "0.6.1"
+p3-symmetric = "0.6.1"
+rand = "0.9"
```
huitseeker: This new crate is already in the default workspace, so its dependency list has to satisfy the repo-wide checks before this can merge. The current CI failure is from this manifest: `rand` needs workspace inheritance, `tracing-subscriber` should use the existing workspace dependency, and `cargo-shear` also reports `itertools` and `p3-goldilocks` as unused here.

precompiles-prover/src/stark_config.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Placeholder relation digest for the precompile chiplet AIR set.
+///
+/// The VM binds `RELATION_DIGEST = Poseidon2::hash_elements([PROTOCOL_ID,
+/// CIRCUIT_COMMITMENT])`. The precompile prover does not yet have a generated
+/// relation/circuit commitment, so for the serialized-proof surface we bind an
+/// explicit empty placeholder digest, `[Felt::ZERO; 4]`, instead of reusing the
+/// VM digest. Replace this with the generated precompile relation digest before
+/// treating proofs as production-secure across AIR upgrades.
+pub const PRECOMPILE_RELATION_DIGEST: RelationDigest = [Felt::ZERO; 4];
```
huitseeker: Could we keep this out of the final `prove` / `verify` path until the real precompile relation digest exists? 

Having a nerdier `custom API is OK, but right now `miden_prover::prove` can produce a `DeferredProof::Stark`, and `miden_verifier::verify` accepts it, but the nested proof is bound to `[0; 4]`. 

That makes the public final-proof surface look like a 96-bit checked proof even though the precompile AIR digest is still a placeholder. A later AIR change could reuse the same transcript domain unless old proofs are rejected some other way.

adr1anh:
Re https://github.com/0xMiden/miden-vm/pull/3293#discussion_r3554390918 — good point. The reviewed line was exposing `[0; 4]` as `pub const PRECOMPILE_RELATION_DIGEST`, which makes the placeholder look like a stable protocol parameter.

I tightened that in 3a6b81276: the placeholder is no longer public from `stark_config`; it’s private to the session prove/verify path. The config constructors still require an explicit relation digest.

I’d prefer to keep the STARK-backed final `prove` / `verify` path enabled for now so we can keep exercising the end-to-end integration. The real fix is the generated precompile/ACE relation digest work, which is already in progress in #3344. Once that lands, we can replace the private placeholder with the generated digest.

precompiles-prover/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+#[cfg(any(test, feature = "std"))]
+extern crate std;
+
+use alloc::string::{String, ToString};
+
+use miden_core::deferred::{DeferredState, TRUE_DIGEST};
+pub use miden_core::proof::{DeferredProof, HashFunction, StarkProof};
+pub use session::{VerifyError, verify_deferred};
+
+pub mod ec;
```
huitseeker: Can these modules stay private or pub(crate) for now?

I can only find external use of `miden_precompiles_prover::{prove_deferred_state, verify_deferred}` from `prover` and `verifier`. 

Exporting `ec`, `hash`, `session`, `stark_config`, and the rest turns the imported prover internals into public API for this crate, which seems a bit soon when we don't have all PRs in this stack ready yet?.

precompiles-prover/src/session/prove.rs
```diff
@@ -0,0 +1,481 @@
...
+use serde::{Serialize, de::DeserializeOwned};
+use serde_wincode::SerdeCompat;
+
+const MAX_STARK_PROOF_BYTES: usize = 64 * 1024 * 1024;
+
+/// Temporary relation digest for the precompile chiplet AIR set.
+///
+/// This is intentionally private: until the generated precompile/ACE relation digest exists,
+/// callers should not treat the all-zero placeholder as a stable protocol parameter.
+const PLACEHOLDER_RELATION_DIGEST: RelationDigest = [Felt::ZERO; 4];
```
huitseeker, line 40: Nit: I know this is made read in #3344, but it would be nice to have this be kept this out of the final proof path until then.

Cargo.toml
```diff
@@ -143,7 +146,7 @@ once_cell       = { version = "1.21", default-features = false }
 paste           = { version = "1.0", default-features = false }
 proptest        = { version = "1.8", default-features = false, features = ["no_std", "alloc"] }
 proptest-derive = { version = "0.7", default-features = false }
-rand            = { version = "0.10", default-features = false }
+rand            = { version = "0.10", default-features = false, features = ["std", "std_rng", "thread_rng"] }
```
huitseeker: Can the `std`, `std_rng`, and `thread_rng` features stay off the workspace `rand` dependency?

This change makes every `rand.workspace = true` consumer request those features. 

In this branch, `cargo tree -p miden-lifted-stark --no-default-features -i rand --edges features` shows `miden-lifted-stark` requesting `rand/std`, `std_rng`, and `thread_rng`, and `miden-verifier --no-default-features` inherits that path through the new precompile prover dependency.

It's easier to add features in the specific crates that need them rather than here.

precompiles-prover/src/deferred/session.rs
```diff
@@ -0,0 +1,367 @@
...
+        if let Some((point, _)) = terms.first() {
+            self.session
+                .constrain_scalar_bound(&point.node, curve.scalar_domain().bound_ptr());
+        }
+
+        let expr_terms = terms
+            .iter()
+            .map(|(point, scalar)| (point.node, scalar.value))
+            .collect::<Vec<_>>();
+        let expr = strategies::straus(&mut self.session, &expr_terms);
```
huitseeker: Could this return a typed translation error, or normalize before calling `straus`?

The curve precompile accepts any non-empty MSM pair list and evaluates it term by term, but this lowering sends every MSM through `strategies::straus()`. That helper panics for `k > 16` and for all-zero scalars.

precompiles-prover/src/session/prove.rs
adr1anh: We need some digest to generate the proving and verification configurations. Since #3344 immediately follows this PR, is already approved, and replaces the placeholder with the generated precompile relation digest, I would prefer to keep the final proof path wired up here rather than temporarily disable and re-enable it. We could use the VM relation digest in the meantime, but it would still only be a placeholder.

precompiles-prover/src/lib.rs
adr1anh: Addressed in `417f69216`: these implementation modules are now `pub(crate)`, leaving the root-level proving and verification API as the supported surface.

Making the modules private causes Rust’s reachability analysis to report a substantial amount of the imported prover stack as dead code, together with related unused imports. Removing all of that cleanly would turn this visibility change into a much larger cleanup, so I added a crate-local `allow(dead_code, unused_imports)` with an explicit reason to keep this commit scoped.

Cargo.toml
adr1anh: Addressed in `417f69216`. The workspace `rand` dependency is back to `default-features = false`, without enabling `std`, `std_rng`, or `thread_rng` globally. `precompiles-prover` now requests only `std_rng` locally for its dev dependency.

`make lint` and `make doc` both pass with this configuration.

precompiles-prover/src/deferred/session.rs
adr1anh: Addressed in `417f69216`, adapted to the latest upstream implementation. The lowering now uses `joint_wnaf` rather than `straus`, so the old `k > 16` panic and its strategy-specific limit no longer apply. The remaining all-zero-scalar panic now returns a typed `UnsupportedMsmAllZeroScalars` translation error, with a regression test.

I did not introduce a new term-count cap because the precompile currently accepts any non-empty pair list and `joint_wnaf` does not require Straus’s 16-term limit. It would still be good to specify an explicit maximum MSM arity as a protocol/resource bound, rather than deriving it from a particular lowering strategy; once specified, that bound should also return a typed error.

precompiles-prover/src/lib.rs
huitseeker: Yep, the allow makes sense here.

precompiles-prover/src/deferred/session.rs
huitseeker: The switch to `joint_wnaf` addresses the Straus-specific part of my comment. I agree that we should not add a term-count cap just because Straus had one.

I still think the duplicate-base issue remains. `Session::ec_msm` still requires a fully merged claim with distinct bases. That means `u1*G + u2*Q` is still not safe when `Q == G`, which is a valid secp256k1 public key.

I also think mixed zero-scalar terms should either be normalized away or rejected before lowering. The new typed error covers the all-zero case, but it does not cover `a*P + 0*Q`.

So I would narrow the request. We do not need an arity cap. We do need either MSM normalization, or an explicit rejection of duplicate canonical bases & zero-scalar terms, with the ECDSA wrapper handling the valid `Q == G` case.

huitseeker (approved):
Thanks a lot for the colossal amount of work!
I opened https://github.com/0xMiden/miden-vm/pull/3414 as potential cleanup, but I think we can work with this PR merged and iterate.

PR #3294, opened 2026-06-26
Serialize MAST commitment inputs in forest payloads

### LINKED ISSUE #3291: Add Commitment Input Sections to MAST Forest Serialization
[Issue #3067](https://github.com/0xMiden/miden-vm/issues/3067) tracks bobbinth's concern from [the #2909 review](https://github.com/0xMiden/miden-vm/pull/2909#pullrequestreview-4186004950). A MAST forest commitment that only uses root nodes does not commit to the shape of the forest. A forest with external nodes can have the same root commitment as a forest with those nodes included.

In [the follow-up comment on #3067](https://github.com/0xMiden/miden-vm/issues/3067#issuecomment-4537778496), bobbinth proposed this commitment rule.

```text
interface_commitment = hash(sorted root digests)
dependency_commitment = hash(sorted external dependency digests)
commitment = hash(interface_commitment || dependency_commitment)
```

[0xMiden/protocol#2977](https://github.com/0xMiden/protocol/issues/2977) uses the same model for public code. Public code commits to public roots and sorted external dependencies.

[0xMiden/protocol#2961](https://github.com/0xMiden/protocol/issues/2961) uses the same ordering rule for account procedure roots. Commitment inputs should not depend on the order in which callers provide roots or components.

Serialized MAST forest wire payloads need to carry the exact input lists used by this commitment. Otherwise, a reader has to recover those lists from fields that were written for other purposes after deserialization has already started.

Dense payloads already carry root IDs and external node digests. Root IDs rebuild the procedure list. External node digests let readers map a node ID to its digest. These fields are needed, but they are not the canonical commitment inputs. Root IDs are not root digests. External digests follow node order, while the commitment uses digest order.

Sparse payloads may omit source nodes. A reader cannot always rebuild the full source root and dependency lists from the visited nodes alone.

Add two canonical commitment input sections to dense and sparse MAST wire payloads.

- Root digests must be sorted by `Word` and strictly increasing.
  
- External dependency digests must be sorted by `Word` and strictly increasing.
  

Readers should reject a payload if either section is out of order or has duplicate digests.

Dense readers should also reject the payload if the sections do not match the roots and external nodes in the payload.

Sparse readers should reject the payload if the root section does not match the sparse roots, if the dependency section is missing any full external node, or if the two sections do not match the stored source forest commitment.

Keep the existing root ID and external node sections. Readers still need them for node ID mapping and forest reconstruction.

Out of scope:

- Changing the commitment rule from #3067.
  
- Making trusted trace replay input an untrusted format. That concern came up in [the #3284 review discussion](https://github.com/0xMiden/miden-vm/pull/3284#issuecomment-4827656030).
  
- Adding advice map data to the MAST commitment. Bobbinth left that as an open question in [the #3067 comment](https://github.com/0xMiden/miden-vm/issues/3067#issuecomment-4537778496).

huitseeker:
On top of ~#3284~ #3313. Fixes #3291

This PR writes sorted MAST root and dependency digests into dense *and* sparse forest payloads. The readers now check those digests before accepting the payload.

It also rejects invalid or too large package type data.


core/src/mast/sparse.rs
```diff
@@ -65,6 +66,12 @@ pub struct SparseMastForest {
 
     /// Cached commitment to the original MAST forest (i.e. a commitment to all roots).
     commitment_cache: Word,
+
+    /// Sorted source root digests used as the forest commitment input.
+    commitment_root_digests: Vec<Word>,
+
+    /// Sorted full external node digests represented in this sparse forest.
+    dependency_digests: Vec<Word>,
```
bobbinth: Why do we need to keep track of this info? Specifically:

- Aren't `dependency_digests` already covered by `digests` (these are all the external nodes in the forests)? - and they are already sorted.
- How are `commitment_root_digests` different from the info provided by `roots`?

core/src/mast/serialization/layout.rs
```diff
@@ -33,6 +33,8 @@ pub(crate) struct ForestLayout {
     node_entry_offset: usize,
     external_digest_offset: usize,
     node_hash_offset: Option<usize>,
+    root_digest_offset: usize,
+    dependency_digest_offset: usize,
```
bobbinth: How is `dependency_digest_offset` different from `external_digest_offset`? Aren't these the same thing?

And same question for `roots_offset` and `root_digest_offset`.

core/src/mast/serialization/layout.rs
```diff
@@ -51,6 +53,7 @@ pub(super) struct MastForestHeader {
     roots_offset: usize,
     basic_block_len: usize,
     external_digests_len: usize,
+    root_digests_len: usize,
```
bobbinth: I'm not sure I fully understand why we need to introduce `root_digests_len`. We already have `roots_count` and `roots_offset` and this gives us MAST node IDs for the roots, right? We just need to make sure the `roots` IDs are sorted in such a way that root digests are sorted in increasing order.

core/src/mast/serialization/mod.rs
```diff
@@ -58,6 +58,10 @@
 //! - digests for all non-external nodes (`Vec<Word>`, ordered by node index)
 //! - lookup is also dense-by-kind: the Nth non-external node uses slot N in this section
 //!
+//! (Commitment input sections)
+//! - root node digests (`Vec<Word>`, sorted by digest)
+//! - external node digests (`Vec<Word>`, sorted by digest)
```
bobbinth, line 65: Similar to the above comments, I'm not sure I understand why we need these new sections. How is data described on line 63 here different from the data described on line 54 above?

bobbinth (commented):
Thank you! Looks good. Not a full review but I left a few comments/questions inline. The main thing is that I didn't fully understand the approach of this PR. It seems to me that that serialized MAST forests already have all the required info - so, new sections should not be needed (both for full and sparse forests). But maybe I missed something?

core/src/mast/sparse.rs
huitseeker: `digests`, `dependency_digests`, `roots`, and `commitment_root_digests` are different views of the same sparse forest. `digests` is keyed by source `MastNodeId` for replay lookups, while `dependency_digests` is the sorted external digest list used by the dependency commitment: [fields](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/sparse.rs#L47-L80). `roots` stores source root IDs. `commitment_root_digests` stores sorted root digests. The sparse reader reads the sorted sections separately and then checks them against the roots and full external nodes: [read](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/sparse.rs#L198-L203), [root check](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/sparse.rs#L226-L246), [dependency check](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/sparse.rs#L269-L287).

core/src/mast/serialization/layout.rs
huitseeker: The offsets point to sections with different order rules. `external_digest_offset` points to external digests ordered by node index, so lookup can find the digest for the Nth external node: [format docs](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/mod.rs#L53-L55). `dependency_commitment_digest_offset` points to the same dependency digest set sorted by digest: [commitment docs](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/mod.rs#L65-L67). The root sections have the same split. `roots_offset` points to root IDs, while `root_commitment_digest_offset` points to sorted root digests: [layout offsets](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/layout.rs#L139-L148). The names now mark the commitment sections explicitly.

core/src/mast/serialization/layout.rs
huitseeker: There is no separate root digest count on the wire. The reader reads `roots_count` once, then uses the same count to size the root ID section and the root commitment digest section: [layout read](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/layout.rs#L229-L236), [length calculation](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/layout.rs#L246-L252). #3294 keeps root IDs in procedure order and writes a separate digest-sorted commitment section. That avoids adding an in-memory root-order rule here. The writer creates that sorted root digest section with `sorted_root_digests`: [write](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/mod.rs#L299-L305), [sort](https://github.com/0xMiden/miden-vm/blob/e0420086a3bb20c8d1ef5b36fc03a80d17708f58/core/src/mast/serialization/mod.rs#L316-L320).

core/src/mast/serialization/mod.rs
huitseeker: The two external digest sections may contain the same values, but they support different access patterns. The structural section is ordered by node index. That follows the earlier goal that a reader can find one node without reading the whole forest. That was a goal in [#2623 r2807884038](https://github.com/0xMiden/miden-vm/pull/2623#discussion_r2807884038), and #2726 kept the same goal for direct `mmap` access and size calculation: [issue](https://github.com/0xMiden/miden-vm/issues/2726), [meeting note](https://github.com/0xMiden/miden-vm/issues/2726#issuecomment-4012893937). The commitment section is sorted by digest, which gives the commitment code stable input order.

core/src/mast/sparse.rs
bobbinth: I'm still not sure I fully understand why we need these distinctions. Is your underlying assumption that we need a Sparse MAST forest to have exactly the same commitment as the source MAST forest? If so, I'm not sure we need this assumption (and relaxing it should simplify a few things).

core/src/mast/serialization/mod.rs
bobbinth: > The two external digest sections may contain the same values, but they support different access patterns. The structural section is ordered by node index. That follows the earlier goal that a reader can find one node without reading the whole forest.

For external nodes, the node index doesn't matter - right? External nodes are always leaf nodes and so, we can sort them in arbitrary order. My assumption was that we'd just sort them by digest, and that this was relatively easy to do at the time when we build MAST forest. If that creates complications, we can always just sort them in memory - there is definitely no need to create two separate sections with the same data but different ordering.

huitseeker:

PR #3297, opened 2026-06-26
fix: document SMT peek advice preimage requirements

huitseeker:
This changes `smt::peek` so it verifies the Merkle path from the requested root to the returned leaf node. It then decodes the advice as an `SmtLeaf`, checks that the leaf belongs to the queried index, and checks that the decoded leaf hash matches the authenticated node.
  
This PR fixes 0xMiden/security-issues#25 for `smt::peek`. [A related crypto PR](https://github.com/0xMiden/crypto/pull/1076) hardens `SmtLeaf::try_from_elements` so future callers get the same leaf index check from the helper itself. This PR does not depend on that change, because it checks the decoded leaf index and hash in `miden-vm`.


bobbinth (commented):
I'm not sure we need to make these changes. `smt::peek` is meant to be an untrusted (but lightweight) operation - similar to reading untrusted data from the advice provider. So, not verifying the Merkle path is intentional and it is assumed that the user code will verify that the data was returned from `smt::peek` was correct.

The usage pattern is usually first call `smt::peek` and then later on call `smt::set` which returns the original value under the modified key, and then to compare that the value returned form `smt::peek` was the same as from `smt::set`. This way, we can frequently avoid doing the expensive Merkle path verification twice.

So, this may be just a documentation issue.

huitseeker:
@bobbinth this is now a documentation PR

bobbinth (approved):
Looks good! Thank you!

PR #3298, opened 2026-06-28
test: simplify MAST node equality checks

### LINKED ISSUE #2358: simplify equality comparison for MastNodes
> do not include any information about decorators, do they?

Yes, correct. My comment was more conceptual in nature. We have these two ways of comparing MAST nodes (with or without decorators) - the current `PartialEq` is kind of in-between. On the one hand, it does include comparison of decorators, on the other hand, this comparison is not always done correctly (hence the need for `semantic_eq()`).

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2323#discussion_r2509140662_

TL;DR: we would like a correct and complete way of comparing two nodes with/without decorators

| Equality kind | Without decorators |With decorators |
|---|---|---|
| semantic  | compare `node.digest` | `semantic_eq` for `JoinNode`, `BasicBlockNode` [^1]|
| structural (representation-aware) | ? | `PartialEq` |
            
            
[^1]: `MastForest::fingerprint_for_node`, `MastNodeFingerprint::fingerprint_from_parts` are strictly speaking incorrect since the `MastForestBuilder` still calls `append_{before_enter,after_exit}`

huitseeker:
Closes #2358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Changes
 
 - Added a Blake3 pure execution benchmark axis and reduced processor benchmark compile time by relaxing forced inlining in execution helpers ([#3289](https://github.com/0xMiden/miden-vm/pull/3289)).
+- Clarified MAST node equality coverage by using structural `PartialEq` directly in merge tests ([#3298](https://github.com/0xMiden/miden-vm/pull/3298)).
 
 ## v0.24.0 (2026-06-24)
 
--- a/core/src/mast/merger/tests.rs
+++ b/core/src/mast/merger/tests.rs
@@ -149,6 +149,24 @@ fn mast_forest_merge_preserves_padded_basic_block_batches() {
     );
 }
 
+#[test]
+fn mast_node_partial_eq_is_structural_for_same_digest_blocks() {
+    let mut forest = MastForest::new();
+
+    let block_a = BasicBlockNodeBuilder::new(vec![Operation::Assert(Felt::from_u32(1))])
+        .add_to_forest(&mut forest)
+        .unwrap();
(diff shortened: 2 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

There is still a slight difference between structural equality and hash-based equality. I think this mostly affects basic blocks because of error codes that are embedded into various operations. So, we could end up with two basic blocks that are structurally different but have the same hash. Not sure if this causes any issues right now, but definitely something to be aware of.

huitseeker:
@bobbinth Thanks, what was missing from `mast_node_partial_eq_is_structural_for_same_digest_blocks` to demonstrate what happens in this case ?

bobbinth:
> what was missing from `mast_node_partial_eq_is_structural_for_same_digest_blocks` to demonstrate what happens in this case ?

The test looks good. I'm wondering more about how these will end up in the MAST forest - i.e., would such two blocks be collapsed somehow because they have the same hash? Or would we keep them as two separate blocks because they are structurally different?

huitseeker:
@bobbinth See `compact_keeps_error_code_bearing_basic_blocks_distinct` for compaction, `test_error_code_bearing_basic_blocks_do_not_dedup_by_digest_only` for construction, merge/compact dedup uses `fingerprint_for_node`, not just `digest()`.

bobbinth:
Ah - I see. So, we don't dedup basic blocks that differ only in error codes. I guess this is the best we can do for now w/o another major refactoring. Maybe worth creating an issue to come back to this in the future because technically, error codes are "debug info".

Al-Kindi-0 (approved):
LGTM

PR #3304, opened 2026-06-29
Generate precompile math MASM during build

huitseeker:
This PR stops storing generated precompile math MASM in the repo. The build now writes those files into Cargo `OUT_DIR`, builds the `miden-precompiles` package from that generated tree, and leaves handwritten MASM under `precompiles/asm`.

The shared precompile IDs, tags, modulus data, and curve data now live in `miden-precompiles-codegen`. The runtime code and the generator both use that crate.

The generated math files under `precompiles/asm/math` are deleted. The old freshness check is deleted too, including `make check-precompile-masm`, `make regenerate-precompile-masm`, the CI job, and the wrapper binary in `miden-precompiles`.

To inspect generated MASM, run:

`cargo run -p miden-precompiles-codegen -- --out target/miden-precompiles-generated-asm`


adr1anh:
I like this direction a lot, it makes a lot more sense now that we bundle the precompile wrappers directly into the core-lib. My only concern is that we're now implementing the specification and arithmetic in the codegen crate which is then re-exported. It's not a big issue, since a bigger clean up of the precompiles crate will be necessary. We could consider maybe splitting the precompile crate into smaller units, namely one with the masm and handlers (which might also allow us to remove the dependency on processor for `EventHandler`).

huitseeker:
The separate `miden-precompiles-codegen` crate is mainly a cycle breaker. `miden-precompiles/build.rs` needs the generator before `miden-precompiles` itself is built. So the generator cannot depend on `miden-precompiles` for the specs. Putting the shared descriptors in the codegen crate lets both sides use them:

- build time: generate MASM from the descriptors
- runtime: use the same descriptors in the precompile handlers

That said, we could do:

- miden-precompiles-spec: no-std descriptors and constants
- miden-precompiles-codegen: std-only MASM generation
- miden-precompiles: runtime handlers and packaged MASM

adr1anh (approved):
Thank you very much

Issue #3301, opened 2026-06-29
Handle debugger events that were previously traces


mooori:
**This issue summarizes [this conversation](https://midengroup.slack.com/archives/C08U8ANH5N3/p1782410213938349).**

The debugger ([`miden-debug`](https://github.com/0xMiden/miden-debug)) handles things like `println` and frame start/end via traces. After the removal of traces, these need to be handled via events. I'm calling them `debugger_events` in the text below.

Events (`emit` instruction) cannot be stripped and host implementations usually [raise an error](https://github.com/0xMiden/miden-vm/blob/049481fd6138dfa54253e3364ca489208267c612/processor/src/host/default.rs#L137-L143) if an event is not handled. However, only the debugger is supposed to handle `debugger_events`. What we want to achieve:

- The debugger handles `debugger_events`
- Hosts other than the debugger don't care about these events. Still they shouldn't raise errors on encountering a `debugger_event`.

#### A: Add no-op handlers to `CoreLibrary`

Most (all?) hosts import the core library and register its event handlers. That way hosts don't need to be updated manually to learn about `debugger_events`. In the debugger itself we can overwrite these no-op handlers.

Currently there are 5 `debugger_events`. They are expected to be relatively stable.

#### B: Add a new library

For example `NoopDebugHandlersLibrary` that defines no-op handlers for all `debugger-events`. Hosts that don't care about `debugger_events` can import this library and register its handlers.

#### C: Each host is responsible for registering the handlers

To facilitate that, the `debugger_events` could be defined centrally, for instance in the `processor` crate.

#### D: Ignore unknown events

Change hosts to ignore unknown events, at the risk of missing events.

#### E: Optional events

Change event semantics to make some events optional. Hosts then shouldn't raise an error if there is no handler for an optional event.

### Notes

With `A` and `B` changing the set of `debugger_events` should be easier: update the library and hosts learn about the new set of `debugger_events` by pulling the latest version of the library. With `C` all hosts need to be updated manually.

`A` to `D` should be relatively small changes. `E` would go deeper, for instance requiring a change to assembly syntax or a convention for event formats.

mooori:
#3305 implements `Option A` as a temporary solution.

In the long term we aim for `Option E` so that new readonly events can be added without the need to update upstream crates.

mooori:
Closing because the temporary workaround is implemented and there is #3324 for a long-term solution.

PR #3305, opened 2026-06-29
feat: add no-op handlers for readonly events in core lib

mooori:
Implements `Option A` from #3301 as temporary solution. This enables all hosts using core library handlers to process debugger events with no-op handlers. Without a handler, hosts raise an `UnknownEvent` error and abort execution.

Adding these handlers to the core library avoids having to add them to all hosts manually.

Event naming is up for debate because as of now the compiler still generates traces instead of events. So from that side there is no restriction on event names.

In the long term we want to encode the readonly property in event names/ids, which is `Option E` in #3301.

/cc @bobbinth, @bitwalker this is the hotfix we discussed today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.24.1 (TBD)
+
+#### Enhancements
+
+- Added no-op handlers for readonly debugger events to `CoreLibrary::handlers`, so hosts that load the core library can execute programs emitting those events without registering no-op handlers manually ([#3305](https://github.com/0xMiden/miden-vm/pull/3305)).
+
 ## v0.24.0 (2026-06-24)
 
 #### Enhancements
--- a/crates/lib/core/src/handlers/readonly.rs
+++ b/crates/lib/core/src/handlers/readonly.rs
@@ -1,0 +1,55 @@
+//! No-op event handlers that never produce advice mutations and are therefore safe to ignore.
+//! Hosts that want to handle these events are expected to replace the no-op handlers.
+//!
+//! This enables communicating readonly events to hosts importing the core library without having
+//! to manually add these no-op handlers. This is a temporary solution. In the long-term, events
+//! themselves should be marked as readonly.
+
+use alloc::{sync::Arc, vec, vec::Vec};
+
+use miden_processor::{
+    ProcessorState,
+    advice::AdviceMutation,
+    event::{EventError, EventHandler, EventName},
+};
+
+// EVENT NAMES
+// ================================================================================================
+//
+// Only the debugger cares about `READONLY_MIDEN_DEBUG_*` events.
+
+/// Marks the start of a debug trace frame.
+pub const READONLY_MIDEN_DEBUG_FRAME_START: EventName =
+    EventName::new("readonly::miden_debug::frame_start");
+/// Marks the end of a debug trace frame.
+pub const READONLY_MIDEN_DEBUG_FRAME_END: EventName =
+    EventName::new("readonly::miden_debug::frame_end");
+/// Emitted when an assertion in a debug trace fails.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## v0.25.0 (TBD)
 
+#### Enhancements
+
+- Added no-op handlers for readonly debugger events to `CoreLibrary::handlers`, so hosts that load the core library can execute programs emitting those events without registering no-op handlers manually ([#3305](https://github.com/0xMiden/miden-vm/pull/3305)).
```
bobbinth, line 7: I believe we want to release this as v0.24.1 patch.

crates/lib/core/src/handlers/readonly.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+// EVENT NAMES
+// ================================================================================================
+//
+// Only the debugger cares about `READONLY_MIDEN_DEBUG_*` events.
+
+/// Marks the start of a debug trace frame.
+pub const READONLY_MIDEN_DEBUG_FRAME_START: EventName =
+    EventName::new("readonly::miden_debug::frame_start");
```
bobbinth, line 23: I would probably go with just `miden::core::debug::frame_start` etc. as this solution is meant to be temporary anyways.

crates/lib/core/src/handlers/readonly.rs
```diff
@@ -0,0 +1,55 @@
...
+struct ReadonlyNoopHandler;
+
+impl EventHandler for ReadonlyNoopHandler {
+    fn on_event(&self, _process: &ProcessorState) -> Result<Vec<AdviceMutation>, EventError> {
+        Ok(vec![])
+    }
+}
+
+/// Returns no-op handlers for all readonly events.
+pub fn readonly_noop_handlers() -> Vec<(EventName, Arc<dyn EventHandler>)> {
```
bobbinth, line 46: nit: maybe I'd name this `noop_debug_handlers()`.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments about naming inline. Also, we should probably target this against `main` so that we can make a patch release with these changes.

CHANGELOG.md
mooori: I've rebased on top of `main` and changed the base branch. The changelog entry is moved to `v0.24.1`.

crates/lib/core/src/handlers/readonly.rs
mooori: The `miden::core::debug` prefix is already used for permanent debug events:

https://github.com/0xMiden/miden-vm/blob/a8706eeca67b677e7b20341ffa25e78aa45655dd/crates/lib/core/src/handlers/debug.rs#L37

Are we fine with adding these temporary handlers to the same namespace? Another option might be `miden::debug::*` since `miden_debug` is the crate that handles these events.

bobbinth: Good point! We could also use `miden::debugger` namespace - but also, given that these are temporary changes, I am fine to keep things as they are.

bobbinth (approved):
Looks good! I think we can probably merge this - but may be good for @bitwalker to take a look as well.

bitwalker (approved):

mooori:
FYI I don't have permission to merge the PR, so please press the button when it's ready.

bobbinth:
> FYI I don't have permission to merge the PR, so please press the button when it's ready.

Done!

@mooori - one thing I realized is that we also want to make these changes in the v0.23 VM (because v0.15 protocol works with v0.23 VM). Could you make another PR that creates a branch off of v0.23.4 tag and makes similar changes there? This way, I'll also release the v0.23.5 VM and that would work with the v0.15 testnet toolchain.

mooori:
Sure, will do it tomorrow morning!

PR #3308, opened 2026-06-29
Clarify `sorted_array` sortedness checks

huitseeker:
`sorted_array` lookup helpers use lower bound advice for speed. They check the local advice result, but they do not prove that the whole memory range is sorted.

This PR documents that callers must prove sorted order before using `find_word`, `find_key_value`, or `find_half_key_value`. It adds `assert_sorted_words`, `assert_sorted_keys`, and `assert_sorted_half_keys` for callers that need to prove that order in MASM.

The new checks scan the range once. They preserve `[start_ptr, end_ptr, ...]`, accept duplicate keys, reject malformed ranges, and reject adjacent entries that are out of order.

Fixes 0xMiden/security-issues#26.

/cc @gswirski


crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -154,11 +164,69 @@ pub proc find_word
...
+    # 2: current_ptr
+    dup loc_store.0
+    dup.1 loc_store.1
+    dup add.4 loc_store.2
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.0 u32mod.4 assertz.err="sorted_array invariant: start_ptr must be word-aligned"
+    loc_load.1 u32mod.4 assertz.err="sorted_array invariant: end_ptr must be word-aligned"
+    loc_load.0 loc_load.1 u32lte assert.err="sorted_array invariant: start_ptr must be <= end_ptr"
+    # => [start_ptr, end_ptr, ...]
```
PhilippGackstatter: Do we need to check u32 validity here, e.g. with `u32assert2`?

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -392,6 +476,114 @@ proc find_partial_key_value
...
+@locals(4)
+pub proc assert_sorted_keys
+    # locals:
+    # 0: start_ptr
+    # 1: end_ptr
+    # 2: current_ptr
+    # 3: use_full_key
+    dup loc_store.0
+    dup.1 loc_store.1
+    dup add.8 loc_store.2
```
PhilippGackstatter: Nit: Imo it would be clearer to use something like `add.DOUBLE_WORD_NUM_ELEMENTS` here. I'm mostly bringing this up because having such constants (that one, but primarily `WORD_NUM_ELEMENTS`) in `miden::core` would be a better place than the tx kernel where we currently define it.

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -392,6 +476,114 @@ proc find_partial_key_value
...
+        loc_load.2 sub.8 loc_load.3 exec.load_key
+        # => [PREVIOUS_KEY, start_ptr, end_ptr, ...]
+
+        loc_load.2 loc_load.3 exec.load_key
+        # => [CURRENT_KEY, PREVIOUS_KEY, start_ptr, end_ptr, ...]
+
+        exec.word::gt
+        not
+        assert.err="sorted_array invariant: keys must be sorted"
+        # => [start_ptr, end_ptr, ...]
```
PhilippGackstatter, line 518: Afair, the protocol kernels always require sortedness _and_ uniqueness, so using `gt not` makes these procedures not useful for miden-protocol. I assume it is that way because uniqueness is not required for `sorted_array`, which is fine. However, if there's no existing use case that actually works without uniqueness, consider making this assertion stronger and require uniqueness (using `word::lt`).

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -392,6 +476,114 @@ proc find_partial_key_value
...
+#! Input:  [start_ptr, end_ptr, ...]
+#! Output: [start_ptr, end_ptr, ...]
+#!
+#! Panics if:
+#! - start_ptr is not word-aligned
+#! - end_ptr is not double-word-aligned with start_ptr
+#! - `start_ptr > end_ptr`
+#! - any adjacent half-key pair is out of order.
+@locals(4)
+pub proc assert_sorted_half_keys
```
PhilippGackstatter, line 547: Seems like this is almost a duplicate of `assert_sorted_keys`, so maybe this could be factored out into a shared helper that takes `is_full_key` as an input.

PhilippGackstatter (approved):
LGTM, nothing blocking.

crates/lib/core/asm/collections/sorted_array.masm
huitseeker: done

crates/lib/core/asm/collections/sorted_array.masm
huitseeker: Good idea. I kept the .4 and .8 literals for now because this file already uses that style for word and double-word steps. I would rather do a small follow-up cleanup across the file

crates/lib/core/asm/collections/sorted_array.masm
huitseeker: Keeping them separate is useful to let each public checker have a direct stack contract and direct panic text. I would be fine with a follow-up refactor if you prefer it in this PR.

crates/lib/core/asm/collections/sorted_array.masm
huitseeker: I could go either way on this, as long as the contract is documented.

crates/lib/core/asm/collections/sorted_array.masm
PhilippGackstatter: Yeah, the custom error is actually one reason that we may still consider not using these procedures in the protocol, even if they enforce uniqueness, so I get that argument. I wish there was a way to provide errors as parameters. Although, if we had VM stacktraces to be able to see which set of calls led to a failure, the need for these would also go down by a lot, and these would be way more useful than good error messages in any case.

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -154,11 +164,71 @@ pub proc find_word
...
+    # 2: current_ptr
+    dup loc_store.0
+    dup.1 loc_store.1
+    dup add.4 loc_store.2
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.0 loc_load.1 u32assert2.err="sorted_array invariant: range pointers must be u32"
+    u32lte assert.err="sorted_array invariant: start_ptr must be <= end_ptr"
+    loc_load.0 u32mod.4 assertz.err="sorted_array invariant: start_ptr must be word-aligned"
+    loc_load.1 u32mod.4 assertz.err="sorted_array invariant: end_ptr must be word-aligned"
```
Al-Kindi-0: nit: i would avoid `loc_load` and instead use `dup` earlier

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -154,11 +164,71 @@ pub proc find_word
...
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.2 loc_load.1 u32lt
```
Al-Kindi-0: same nit here

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -392,6 +478,118 @@ proc find_partial_key_value
...
+    push.1 loc_store.3
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.0 loc_load.1 u32assert2.err="sorted_array invariant: range pointers must be u32"
+    u32lte assert.err="sorted_array invariant: start_ptr must be <= end_ptr"
+    loc_load.0 u32mod.4 assertz.err="sorted_array invariant: start_ptr must be word-aligned"
+    loc_load.1 loc_load.0 sub u32mod.8 assertz.err="sorted_array invariant: key-value range must be double-word aligned"
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.2 loc_load.1 u32lt
```
Al-Kindi-0: same nit here

crates/lib/core/asm/collections/sorted_array.masm
```diff
@@ -392,6 +478,118 @@ proc find_partial_key_value
...
+    push.0 loc_store.3
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.0 loc_load.1 u32assert2.err="sorted_array invariant: range pointers must be u32"
+    u32lte assert.err="sorted_array invariant: start_ptr must be <= end_ptr"
+    loc_load.0 u32mod.4 assertz.err="sorted_array invariant: start_ptr must be word-aligned"
+    loc_load.1 loc_load.0 sub u32mod.8 assertz.err="sorted_array invariant: key-value range must be double-word aligned"
+    # => [start_ptr, end_ptr, ...]
+
+    loc_load.2 loc_load.1 u32lt
```
Al-Kindi-0: same nit here and elsewhere

Al-Kindi-0 (approved):
LGTM

PR #3310, opened 2026-06-30
ci: manage Blake3 non-regression variance

huitseeker:
Fixes the red CI status from < 1% regression on noisy parts of the bench.


Al-Kindi-0 (approved):
LGTM, thank you

PR #3311, opened 2026-06-30
Bind MAST commitments to external dependency digests & advice map

### LINKED ISSUE #3067: Rationalize digests throughout programs
I think we need to think through the whole commitment approach to MAST forest/package more holistically. Specifically, we have:

_Originally posted by @bobbinth in https://github.com/0xMiden/miden-vm/pull/2909#pullrequestreview-4186004950_

huitseeker:
Fixes https://github.com/0xMiden/miden-vm/issues/3067.

This makes MAST forest commitments include external dependency digests.

Before this change, two forests could have the same commitment even when their external dependencies were different. Now the dependency digests are part of the commitment input, so changing an external dependency changes the forest commitment.

This is stacked on #3294 and uses its sorted commitment input format.


core/src/mast/mod.rs
```diff
@@ -240,6 +240,16 @@ impl MastForest {
...
 
+    pub(in crate::mast) fn add_external_node(
+        &mut self,
+        node: MastNode,
+    ) -> Result<MastNodeId, MastForestError> {
+        debug_assert!(node.is_external());
+        let node_id = self.nodes.push(node).map_err(|_| MastForestError::TooManyNodes)?;
+        self.commitment = self.compute_mast_forest_commitment();
+        Ok(node_id)
+    }
```
bobbinth: Why do we need to add this method here? I think we should try to remove the existing mutators from `MastForest` rather than add new ones. I understand that adding it here doesn't change much because the same mutation was previously happening via `MastForestContributor` directly - but I think we should probably try to get rid of `MastForestContributor` at some point in the future to make sure direct mutations of `MastForest` are not possible.

core/src/mast/mod.rs
```diff
@@ -457,6 +469,25 @@ fn compute_nodes_commitment(
...
 
+fn compute_dependency_commitment(nodes: &IndexVec<MastNodeId, MastNode>) -> Word {
+    let mut digests: Vec<Word> = nodes
+        .iter()
+        .filter(|node| node.is_external())
+        .map(MastNodeExt::digest)
+        .collect();
+    digests.sort_unstable();
+    miden_crypto::hash::poseidon2::Poseidon2::merge_many(&digests)
+}
```
bobbinth, line 661: This is fine for now, but ideally, we'd avoid this need for iteration/sorting by making sure external nodes are already sorted and contiguous upon construction. Is this coming in a future PR?

core/src/mast/mod.rs
```diff
@@ -457,6 +469,25 @@ fn compute_nodes_commitment(
...
+
+fn compute_mast_forest_commitment(
+    nodes: &IndexVec<MastNodeId, MastNode>,
+    roots: &[MastNodeId],
+) -> Word {
+    let interface_commitment = compute_nodes_commitment(nodes, roots);
+    let dependency_commitment = compute_dependency_commitment(nodes);
+    miden_crypto::hash::poseidon2::Poseidon2::merge(&[interface_commitment, dependency_commitment])
+}
```
bobbinth, line 672: As mentioned in the original issue, we should also include advice commitment into the forest commitment as well. Otherwise, advice data could be stripped from the forest and there would be no way to tell this.

core/src/mast/mod.rs
```diff
@@ -118,7 +118,7 @@ pub struct MastForest {
     /// Advice map to be loaded into the VM prior to executing procedures from this MAST forest.
     advice_map: AdviceMap,
 
-    /// Commitment to this MAST forest (commitment to all roots).
+    /// Commitment to this MAST forest's roots and external dependencies.
     commitment: Word,
```
bobbinth, line 130: I would maybe create something like `MastForestCommitment` struct that could look like so:

```rust
struct MastForestCommitment {
    commitment: Word,
    interface_commitment: Word,
    dependency_commitment: Word,
    advice_commitment: Word,
}
```

This struct could be built at the time when we construct the MAST forest and then we can add accessors to `MastForest` to read & return relevant fields of this struct.

bobbinth (commented):
Thank you! Looks good. Not a full review yet, but I left some comments which may have impact on the approach.

core/src/mast/mod.rs
huitseeker: The original issue #3067 was rather speculative on whether the advice map should be committed to:

> Open question: the above does not commit to the advice map data included in the forest - should it? If so, the commitment would be:
> ```
> commitment = hash(interface_commitment || dependency_commitment || advice_commitment)
> ```

 Updated, this now commits to the advice map as a separate part of the forest commitment.

core/src/mast/mod.rs
huitseeker: We now store the combined commitment and its three inputs in `MastForestCommitment`. `MastForest` exposes the interface, dependency, advice & final commitment through accessors.

core/src/mast/mod.rs
huitseeker: We now makes finalized dense `MastForest` values canonical. External nodes come first and are strictly sorted by digest. Basic blocks follow. Internal nodes come last, with children before parents. 

Finalization canonicalizes the `MastForest` node order. Serialization expects finalized forests, but still has a temporary fallback for construction-phase forests that were built through the low-level append API.

I think that's all?

core/src/mast/mod.rs
huitseeker: The reader now treats dense order as part of the format. It checks the order while reading instead of sorting the payload into shape. The dependency digest section must match the external prefix exactly.

The function you're commenting on is used in construction-phase forests.

core/src/mast/mod.rs
huitseeker: `MastForestContributor` is still needed because it is the current common builder interface for adding all node types to a MastForest. It is used across core tests, merger code, package tests/serialization/arbitrary support, assembly finalization, and helper builders. We can make `MastForestContributor` operate on another structure, but I'd recommend another PR.

core/src/mast/mod.rs
```diff
@@ -457,6 +610,87 @@ fn compute_nodes_commitment(
...
+            .filter(|node| node.is_external())
+            .map(MastNodeExt::digest)
+            .collect();
+        digests.sort_unstable();
+        digests
+    };
+    miden_crypto::hash::poseidon2::Poseidon2::merge_many(&digests)
+}
+
+pub(in crate::mast) fn compute_advice_commitment(advice_map: &AdviceMap) -> Word {
```
bobbinth: This could be a method on the `AdviceMap` type, right?

I would also add a comment to this describing commitment methodology. I understand that we build a list of elements from concatenating key-value pairs (sorted by key) and than hash this list.

core/src/mast/mod.rs
```diff
@@ -457,6 +610,87 @@ fn compute_nodes_commitment(
...
+    let entry_commitments = advice_map
+        .iter()
+        .map(|(key, values)| {
+            let mut elements = Vec::with_capacity(WORD_SIZE + 2 + values.len());
+            elements.extend_from_slice(key.as_elements());
+            let value_len = values.len() as u64;
+            elements.push(Felt::from_u32(value_len as u32));
+            elements.push(Felt::from_u32((value_len >> 32) as u32));
```
bobbinth: I'm not sure if mixing the length into the hash is really needed (the overall length of the vector is used for one of the capacity elements in `hash_elements()` - but if we do want to include, we could probably safely assume that the length fits into a single field element since these are actual lengths of in-memory structures.

core/src/mast/mod.rs
```diff
@@ -457,6 +610,87 @@ fn compute_nodes_commitment(
...
+            elements.extend_from_slice(values);
+            miden_crypto::hash::poseidon2::Poseidon2::hash_elements(&elements)
+        })
+        .collect::<Vec<_>>();
+
+    miden_crypto::hash::poseidon2::Poseidon2::merge_many(&entry_commitments)
```
bobbinth: nit: instead of using a full path in several places, l'd just import `Poseidon2` at the top.

core/src/mast/node/external.rs
```diff
@@ -155,13 +155,11 @@ impl ExternalNodeBuilder {
...
 impl MastForestContributor for ExternalNodeBuilder {
     fn add_to_forest(self, forest: &mut MastForest) -> Result<MastNodeId, MastForestError> {
-        // Create the node in the forest with Linked variant from the start
-        // Move the data directly without intermediate cloning
         let node_id = forest
             .nodes
             .push(ExternalNode { digest: self.digest }.into())
             .map_err(|_| MastForestError::TooManyNodes)?;
-
+        forest.commitment = forest.compute_mast_forest_commitment();
```
bobbinth, line 162: This is probably fine for now, but recomputing the entire commitment after every external node is added is probably pretty expensive. Do we need to create an issue to cover this or do we already have one?

core/src/mast/serialization/resolved.rs
bobbinth: This means that root digests are not sorted in the serialized representation, right?

bobbinth: It seems like we are doing some double work here: we first iterate over the external nodes and put them into a vector. Then, separately, we iterate over them again and check that they were copied correctly and that they are sorted. Could this note be done as a part of the loop on lines 221-225? If yes, we could probably get rid of the `validate_digest_section()` function entirely.

crates/assembly/src/assembler.rs
```diff
@@ -1126,7 +1126,8 @@ impl Assembler {
                         )));
                     },
                 };
-                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info))
+                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info)
+                    .with_source_library_commitment(lib.interface_digest()))
```
bobbinth: I'm not sure I follow this code: doesn't `StaticLibrary::new()` already set source library commitment to `mast.interface_commitment()`? Or is this because `lib.interface_digest()` digest is not the same as `lib.mast.interface_digest()`?

If so, I would we should document this better and explain why overriding MAST forest commitment is desired here. But also, maybe a better approach is to have dedicated constructors - e.g., something like:

```rust
impl<'a> StaticLibrary<'a> {

    pub(crate) fn from_mast(
        mast: &'a MastForest,
        debug_info: Option<PackageDebugInfo>
    ) -> Self {
       ...
    }

    pub(crate) fn from_library(
        mast: &'a LinkLibrary,
        debug_info: Option<PackageDebugInfo>
    ) -> Self {
       ...
    }

}
```

Or maybe even:

```rust
impl<'a> StaticLibrary<'a> {

    pub(crate) fn from_mast(mast: &'a MastForest) -> Self {
       ...
    }

    pub(crate) fn from_library(mast: &'a LinkLibrary) -> Self {
       ...
    }

    pub(crate) fn with_debug_info(mut self, debug_info: PackageDebugInfo) -> Self {
        ...
    }

}
```

core/src/mast/sparse.rs
```diff
@@ -64,13 +67,14 @@ pub struct SparseMastForest {
...
+    /// Cached commitment to the original MAST forest's roots, external dependencies, and advice
+    /// map.
     commitment_cache: Word,
 
     /// Sorted source root digests used as the forest commitment input.
     commitment_root_digests: Vec<Word>,
 
-    /// Sorted full external node digests represented in this sparse forest.
+    /// Sorted source external node digests used as the dependency commitment input.
     dependency_digests: Vec<Word>,
```
bobbinth, line 84: Question: why do we need to store these instead of using the `MastForestCommitment` struct?

core/src/mast/serialization/mod.rs
```diff
@@ -235,6 +235,20 @@ impl MastForest {
...
+            // direct path below; this fallback prevents public serialization from panicking before
+            // callers have moved through a finalization path.
+            let canonical = MastForest::from_raw_parts(
+                self.nodes.clone(),
+                self.roots.clone(),
+                self.advice_map.clone(),
+            )
+            .expect("dense MAST forest must be valid before serialization");
+            canonical.write_into_with_options(target, hashless);
+            return;
```
bobbinth, line 253: I'm not sure I follow the invariants here - my understanding is that once we have a `MastForest` struct, everything is guaranteed to be in the correct order. Is this not the case?

crates/mast-package/src/package/mod.rs
```diff
@@ -80,8 +80,8 @@ pub struct Package {
     pub name: PackageId,
     /// An optional semantic version for the package
     pub version: Version,
-    /// The content hash of the exported code of this package, formed by hashing the roots of all
-    /// exports in lexicographical order (by digest, not procedure name)
+    /// The code commitment of this package, formed by hashing the package interface commitment
+    /// with the dependency commitment of the underlying MAST forest.
     digest: Word,
```
bobbinth: I was thinking that redefining package commitment would be out of scope for this PR as we still need to align on methodology. One open question for me is whether MAST interface commitment should be included in the package commitment or not.

core/src/mast/mod.rs
bobbinth, line 1: Probably not for this PR, but this file has become quite big and difficult to review. We should probably split it up (e.g., test-related code could go into a separate module).

core/src/mast/mod.rs
```diff
@@ -180,25 +193,162 @@ impl MastForest {
...
+    for &root_id in &parts.roots {
+        if root_id.to_usize() >= node_count {
+            return Err(MastForestError::NodeIdOverflow(root_id, node_count));
+        }
+    }
+
+    Ok(())
+}
+
+fn canonicalize_dense_parts(parts: MastForestParts) -> Result<MastForestParts, MastForestError> {
```
bobbinth, line 229: Could we add some doc comments explaining what this does? Also, would it make sense to put this as a method on `MastForestParts`?

bobbinth (commented):
Looks good! Thank you. Still not a full review, but I left some more comments/questions inline.

core/src/mast/mod.rs
huitseeker: `AdviceMap` now owns `commitment()`, with a doc comment for the key-ordered entry hashing.

core/src/mast/mod.rs
huitseeker: I removed the explicit length fields. Each entry is `key || values`; the key is fixed width, and `hash_elements()` already binds the element count.

core/src/mast/mod.rs
huitseeker: Fixed

core/src/advice/map.rs
```diff
@@ -87,6 +88,26 @@ impl AdviceMap {
...
+            .map(|(key, values)| {
+                let mut elements = Vec::with_capacity(WORD_SIZE + values.len());
+                elements.extend_from_slice(key.as_elements());
+                elements.extend_from_slice(values);
+                Poseidon2::hash_elements(&elements)
+            })
+            .collect::<Vec<_>>();
+
+        Poseidon2::merge_many(&entry_commitments)
+    }
```
bobbinth, line 109: Thinking more about it, there may be an attack vector here which can result in two distinct advice maps reducing to the same digest.

For example, let's say we have the following:

- A map with two key-value entries:
  - `key1` -> `value1`
  - `key2 -> `value2`
- A map with a single entry:
  - `key1` -> `value1` || `key2` || `value2`

Basically, the value in the second map is "extended" with key-value entry from second element of the first map. AFAICT, these would currently result in the same commitment.

Brining back length injection should fix this (though, I haven't thought it through completely).

Another alternative could be to fist hash key-value entries, and then compute the commitment from these hashes.

bobbinth: Ah - n/m, I think this already implements the alternative I suggested.

core/src/mast/node/external.rs
huitseeker: #3311 keeps this "repair" path because that branch still has direct append construction. #3329 removes the production append trait path and makes dense finalization return a canonical MastForest, so this per-append commitment recompute is no longer part of normal construction.

core/src/mast/serialization/resolved.rs
huitseeker: Yes. In #3311, the serialized root digest commitment section is canonical. It must equal the procedure root digests sorted lexicographically.

That section is not the same as the serialized root ID list. [The root ID list is written in forest root order](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L269-L271). It tells the reader which node IDs are roots. It is not a digest commitment input.

#3294 is where these explicit commitment input sections were added to the wire format. [It bumped the MAST wire version to `[0, 0, 5]` for "sorted root digest and sorted external digest commitment input sections"](https://github.com/0xMiden/miden-vm/blob/8fbae2e4c3b8149c5a4ecaf1d39cf64cc888876d/core/src/mast/serialization/mod.rs#L215-L222). [It writes the sorted root digest section and then the sorted external dependency digest section after the node digest data](https://github.com/0xMiden/miden-vm/blob/8fbae2e4c3b8149c5a4ecaf1d39cf64cc888876d/core/src/mast/serialization/mod.rs#L289-L304). [It also made readers compare those serialized sections with the digests derived from the serialized forest](https://github.com/0xMiden/miden-vm/blob/8fbae2e4c3b8149c5a4ecaf1d39cf64cc888876d/core/src/mast/serialization/resolved.rs#L206-L265). #3311 builds on that by using those sections as the canonical input lists for the forest commitment changes.

[The root digest commitment section is written from `sorted_root_digests(self)`](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L309-L310), and [`sorted_root_digests()` sorts those digests before writing them](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L325-L329).

On read, [`validate_commitment_input_sections()` derives the expected root digests from the root IDs, sorts that expected list, rejects duplicate root digests, and compares each serialized digest entry to the expected digest at the same index](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/resolved.rs#L206-L226). The sort is for canonicalizing the expected list, not for checking that the original list was ordered. The validation constrains the wire section by equality against that canonical list. It is not a direct adjacent-pair check over the serialized section.

This aims to match your design requirements as I could gather them:

- [a comment in Issue #3067 asks for sorted root and dependency digest commitments](https://github.com/0xMiden/miden-vm/issues/3067#issuecomment-4537778496).
  
- [PR #2765 asks for external nodes to precede internal nodes and be sorted in the `MastForest`](https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122440986).
  
- [Issue #3043 extends the in-memory order to external nodes, then basic blocks, then internal nodes, with children before parents](https://github.com/0xMiden/miden-vm/issues/3043#issuecomment-4323023661).
  
- [PR #3055 says sorted wire digest sections are less useful if the in-memory forest order differs](https://github.com/0xMiden/miden-vm/pull/3055#pullrequestreview-4249450277).

core/src/mast/mod.rs
huitseeker: #3329 makes the finalization API clearer by adding `MastForest::from_raw_parts_with_id_map()`, which documents that canonicalization returns an ID remap. I kept `canonicalize_dense_parts()` private under MastForest because it is an implementation step inside finalization. `MastForestParts` is still just the crate-local bag of fields used to finish construction.

crates/assembly/src/assembler.rs
huitseeker: The override is still needed because the linker stores libraries by package digest. The static-library import path must use the same key when it records source metadata. [The linker uses](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/crates/assembly/src/linker/mod.rs#L169-L176) `library.package.digest()` [as the library key](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/crates/assembly/src/linker/mod.rs#L169-L176). [The static-library path now uses](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/crates/assembly/src/assembler.rs#L1129-L1130) `lib.package.digest()` [for the source library commitment](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/crates/assembly/src/assembler.rs#L1129-L1130).

This no longer depends on a separate `interface_digest()` method.

crates/mast-package/src/package/mod.rs
huitseeker: I retracted the package digest widening.`Package::digest()` is back to the exported-interface digest.
I also removed the separate package `interface_digest()` API.

core/src/mast/serialization/mod.rs
huitseeker: This is addressed in the stack. #3311 still repairs at the boundary because the branch still has direct append construction. #3329 makes finalized dense MastForest values canonical in memory, so serialization and dependency commitment can trust the external-node prefix.

core/src/mast/mod.rs
huitseeker: I would keep the broad file split out of #3311.  #3329 moves the dense builder into `core/src/mast/dense_builder.rs`, but it does not split all of `core/src/mast/mod.rs`. This PR is already double the recommended size, but I'll stack a PR to make that change (on top of the sequence of #3313, #3294, #3311, #3329).

core/src/mast/serialization/resolved.rs
huitseeker: Yes for dependency digests: because dense wire order puts external nodes first and writes the dependency digest section in that same order, [`validate_dependency_commitment_input_section()`](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/core/src/mast/serialization/resolved.rs#L239-L260) can read external node `i`, read dependency digest `i`, check strict increase, and compare them directly, while root digests still derive from the [serialized root ID list](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/core/src/mast/serialization/mod.rs#L269-L271) because those IDs identify the executable public roots and the [sorted root digest section](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/core/src/mast/serialization/resolved.rs#L211-L236) only commits to their digest set.

core/src/mast/sparse.rs
huitseeker: Because a `SparseMastForest` is not the full input to the source forest commitment, but a replay view whose builder [keeps original node IDs, copies only full visits, and stores some referenced nodes as digest-only entries](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/core/src/mast/sparse.rs#L548-L599), it must also carry the sorted root and dependency digest lists used by the source forest, so deserialization can [recompute the same commitment and check the retained roots and full external nodes against those lists](https://github.com/0xMiden/miden-vm/blob/1454efaaedcac9c46eb481a8ae0c42dd222bd353/core/src/mast/sparse.rs#L153-L177) even though the sparse node set alone cannot recreate the dense forest's commitment input.

core/src/mast/mod.rs
huitseeker: That other PR is #3329 now.

huitseeker:

Issue #3303, opened 2026-06-29
Add untrusted TraceProvingInputs deserialization


huitseeker:
`TraceProvingInputs` can now be serialized for trusted remote proving. This reader is not safe for untrusted bytes.

Today, `SparseMastForest` keeps full-node hashes from the payload. If a payload has the right shape but wrong hashes, the reader can accept it. The error may only show up later as a failed proof or a failed constraint check. The API should make the trusted path and untrusted path clear.

Add an untrusted reader for `TraceProvingInputs`.

The untrusted reader should:

* reject hashes that can be rebuilt from node data,
* rebuild full-node hashes while reading,
* check the source forest commitment from the rebuilt roots,
* reject digest-only entries unless the reader can rebuild their hashes,
* define how external nodes are checked once #3291 is done,
* return a clear error when a rebuilt hash does not match the payload.

The external-node part is blocked by #3291.

Keep the current trusted reader for trusted prover deployments.

Add tests that corrupt:

* a full-node hash,
* a digest-only child hash,
* the source forest commitment,
* external node data after #3291.

The untrusted reader should reject each case. The trusted reader may keep accepting these bytes by design.

PR #3313, opened 2026-06-30
1/3 Serialize sparse MAST forests

huitseeker:
Part of #3235.

This PR adds trusted sparse MAST forest serialization.

Sparse payloads preserve source node IDs, full visited nodes, digest-only entries, roots, the advice map, and the source forest commitment.

Dense MAST readers reject sparse payloads. Sparse readers reject dense payloads, trailing bytes, invalid node IDs, duplicate sparse entries, missing child digests, and full-node digest overlap.

This PR does not serialize trace proving inputs yet.

### DIFF
--- a/core/src/mast/serialization/sparse.rs
+++ b/core/src/mast/serialization/sparse.rs
@@ -38,7 +38,6 @@ const SPARSE_EXTERNAL: u8 = 8;
 /// [`MastForest`] wire format. Callers must only read these bytes from a trusted producer, or after
 /// an outer transport/authentication layer has accepted them.
 fn write_sparse_into<W: ByteWriter>(forest: &SparseMastForest, target: &mut W) {
-    forest.num_nodes().write_into(target);
     write_node_ids(forest.procedure_roots(), target);
     write_sparse_nodes(forest.nodes(), target);
     write_digest_entries(forest.digest_entries(), target);
@@ -161,53 +160,37 @@ impl Deserializable for SparseMastForest {
 fn read_sparse_from<R: ByteReader>(
     source: &mut R,
 ) -> Result<SparseMastForest, DeserializationError> {
-    let source_node_count = source.read_usize()?;
-    if source_node_count > MastForest::MAX_NODES {
-        return Err(DeserializationError::InvalidValue(format!(
-            "sparse source node count {source_node_count} exceeds maximum allowed {}",
-            MastForest::MAX_NODES
-        )));
-    }
-
-    let roots = read_node_ids(source, source_node_count, "procedure root")?;
-    let nodes = read_sparse_nodes(source, source_node_count)?;
-    let digest_entries = read_digest_entries(source, source_node_count)?;
+    let roots = read_node_ids(source, "procedure root")?;
+    let nodes = read_sparse_nodes(source)?;
+    let digest_entries = read_digest_entries(source)?;
     let advice_map = AdviceMap::read_from(source)?;
 
-    SparseMastForest::from_serialized_parts(
-        nodes,
-        digest_entries,
-        source_node_count,
-        roots,
-        advice_map,
-    )
+    SparseMastForest::from_serialized_parts(nodes, digest_entries, roots, advice_map)
 }
 
 fn read_node_ids<R: ByteReader>(
     source: &mut R,
-    source_node_count: usize,
     label: &str,
 ) -> Result<Vec<MastNodeId>, DeserializationError> {
     let count = read_bounded_count(source, u32::min_serialized_size(), label)?;
     let mut ids = Vec::with_capacity(count);
     for _ in 0..count {
-        ids.push(read_node_id(source, source_node_count, label)?);
+        ids.push(read_node_id(source, label)?);
     }
     Ok(ids)
 }
 
 fn read_sparse_nodes<R: ByteReader>(
     source: &mut R,
-    source_node_count: usize,
 ) -> Result<Vec<(MastNodeId, MastNode)>, DeserializationError> {
     let count = read_bounded_count(source, sparse_node_min_size(), "full node count")?;
     let mut nodes = Vec::with_capacity(count);
     let mut previous_id = None;
 
     for _ in 0..count {
-        let id = read_node_id(source, source_node_count, "full node")?;
+        let id = read_node_id(source, "full node")?;
         validate_strictly_increasing_id(previous_id, id, "full node")?;
-        let node = read_sparse_node(source, source_node_count, id)?;
+        let node = read_sparse_node(source, id)?;
         nodes.push((id, node));
         previous_id = Some(id);
     }
@@ -217,7 +200,6 @@ fn read_sparse_nodes<R: ByteReader>(
 
 fn read_digest_entries<R: ByteReader>(
     source: &mut R,
-    source_node_count: usize,
 ) -> Result<Vec<(MastNodeId, Word)>, DeserializationError> {
     let count =
         read_bounded_count(source, sparse_digest_entry_min_size(), "digest-only node count")?;
@@ -225,7 +207,7 @@ fn read_digest_entries<R: ByteReader>(
     let mut previous_id = None;
 
     for _ in 0..count {
-        let id = read_node_id(source, source_node_count, "digest-only node")?;
+        let id = read_node_id(source, "digest-only node")?;
         validate_strictly_increasing_id(previous_id, id, "digest-only node")?;
         let digest = Word::read_from(source)?;
         digests.push((id, digest));
@@ -237,7 +219,6 @@ fn read_digest_entries<R: ByteReader>(
 
 fn read_sparse_node<R: ByteReader>(
     source: &mut R,
-    source_node_count: usize,
     node_id: MastNodeId,
 ) -> Result<MastNode, DeserializationError> {
     let tag = source.read_u8()?;
@@ -254,27 +235,27 @@ fn read_sparse_node<R: ByteReader>(
                 .map(Into::into)
         },
         SPARSE_JOIN => {
-            let first = read_node_id(source, source_node_count, "join first child")?;
-            let second = read_node_id(source, source_node_count, "join second child")?;
+            let first = read_node_id(source, "join first child")?;
+            let second = read_node_id(source, "join second child")?;
             JoinNodeBuilder::new([first, second])
                 .with_digest(digest)
                 .build_linked()
                 .map(Into::into)
         },
         SPARSE_SPLIT => {
-            let on_true = read_node_id(source, source_node_count, "split true child")?;
-            let on_false = read_node_id(source, source_node_count, "split false child")?;
+            let on_true = read_node_id(source, "split true child")?;
+            let on_false = read_node_id(source, "split false child")?;
             SplitNodeBuilder::new([on_true, on_false])
                 .with_digest(digest)
                 .build_linked()
                 .map(Into::into)
         },
         SPARSE_LOOP => {
-            let body = read_node_id(source, source_node_count, "loop body")?;
+            let body = read_node_id(source, "loop body")?;
             LoopNodeBuilder::new(body).with_digest(digest).build_linked().map(Into::into)
         },
         SPARSE_CALL | SPARSE_SYSCALL => {
-            let callee = read_node_id(source, source_node_count, "call callee")?;
+            let callee = read_node_id(source, "call callee")?;
             let builder = if tag == SPARSE_SYSCALL {
                 CallNodeBuilder::new_syscall(callee)
             } else {
@@ -324,11 +305,10 @@ fn sparse_digest_entry_min_size() -> usize {
 
 fn read_node_id<R: ByteReader>(
     source: &mut R,
-    node_count: usize,
     label: &str,
 ) -> Result<MastNodeId, DeserializationError> {
     let raw = u32::read_from(source)?;
-    MastNodeId::from_u32_with_node_count(raw, node_count).map_err(|err| {
+    MastNodeId::from_u32_with_node_count(raw, MastForest::MAX_NODES).map_err(|err| {
         DeserializationError::InvalidValue(format!("invalid {label} id {raw}: {err}"))
     })
 }
--- a/core/src/mast/serialization/sparse.rs
+++ b/core/src/mast/serialization/sparse.rs
@@ -1,9 +1,9 @@
 use alloc::{format, string::ToString, vec::Vec};
 
 use super::{
-    FLAG_HASHLESS, FLAG_SPARSE, MAGIC, MastNodeEntry, VERSION,
+    MastNodeEntry,
     basic_blocks::{BasicBlockDataBuilder, BasicBlockDataDecoder},
-    layout::{OffsetTrackingReader, TrackingReader, WireFlags, read_and_validate_header},
+    layout::{OffsetTrackingReader, TrackingReader},
 };
 use crate::{
     Word,
@@ -15,12 +15,12 @@ use crate::{
     },
 };
 
-const SPARSE_FLAGS: u8 = FLAG_HASHLESS | FLAG_SPARSE;
+const SPARSE_MAGIC: &[u8; 4] = b"SMST";
+const SPARSE_VERSION: [u8; 3] = [0, 0, 0];
 
 fn sparse_mast_forest_min_serialized_size() -> usize {
-    MAGIC.len()
-        + 1
-        + VERSION.len()
+    SPARSE_MAGIC.len()
+        + SPARSE_VERSION.len()
         + usize::min_serialized_size() * 7
         + Word::min_serialized_size()
         + usize::min_serialized_size()
@@ -56,9 +56,8 @@ pub(super) fn write_sparse_into<W: ByteWriter>(forest: &SparseMastForest, target
     let non_external_count =
         entries.iter().filter(|entry| !matches!(entry, MastNodeEntry::External)).count();
 
-    target.write_bytes(MAGIC);
-    target.write_u8(SPARSE_FLAGS);
-    target.write_bytes(&VERSION);
+    target.write_bytes(SPARSE_MAGIC);
+    target.write_bytes(&SPARSE_VERSION);
 
     target.write_usize(forest.procedure_roots().len());
     target.write_usize(forest.num_nodes());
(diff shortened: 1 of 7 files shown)


core/src/mast/sparse.rs
bobbinth, line 54: Not related to this PR, but why not combine these with `nodes` by using `ExternalNode` as a substitute for nodes for which we know only digests?

bobbinth, line 60: Also not related to this PR, but why do we need this here? The use case for `SparseMastForest` is to be used in `TraceProvingInputs`, and in that context, everything related to the advice provider should be covered by `AdviceReplay`. Or am I missing something?

bobbinth: I'm not sure the comment here is accurate (at least as of the state in this PR) - AFAICT, this field is used for validation only. And it is not clear to me if for validation use case it actually provides any meaningful benefit (a malicious actor could use any value for `num_nodes` and there is no way for the deserializer to check if that was correct).

core/src/mast/sparse.rs
```diff
@@ -98,6 +104,157 @@ impl SparseMastForest {
...
+        digests: Vec<(MastNodeId, Word)>,
+        num_nodes: usize,
+        roots: Vec<MastNodeId>,
+        advice_map: AdviceMap,
+        commitment_cache: Word,
+    ) -> Result<Self, crate::serde::DeserializationError> {
+        validate_sparse_node_bound(num_nodes)?;
+
+        let nodes = collect_unique_nodes(nodes, num_nodes)?;
+        let digests = collect_unique_digests(digests, num_nodes)?;
```
bobbinth: As mentioned in the previous comment, we could probably pass `MastForest::MAX_NODES` instead of `num_nodes` as a second parameter here without opening up any extra attack vectors because `max_nodes` is fully controlled by the serializer.

Also, as mentioned in the other comment, in the future, we should probably collapse `digests` into `nodes` to process them more uniformly.

core/src/mast/sparse.rs
```diff
@@ -98,6 +104,157 @@ impl SparseMastForest {
...
+
+    /// Builds a sparse forest from parts decoded from the sparse wire format.
+    pub(in crate::mast) fn from_serialized_parts(
+        nodes: Vec<(MastNodeId, MastNode)>,
+        digests: Vec<(MastNodeId, Word)>,
+        num_nodes: usize,
+        roots: Vec<MastNodeId>,
+        advice_map: AdviceMap,
+        commitment_cache: Word,
+    ) -> Result<Self, crate::serde::DeserializationError> {
```
bobbinth: nit: let's import `DeserializationError` at the top - we usually don't use absolute paths unless there is some special reason for doing so. (would be good to do this consistently throughout the codebase)

core/src/mast/sparse.rs
```diff
@@ -98,6 +104,157 @@ impl SparseMastForest {
...
+        }
+    }
+    Ok(result)
+}
+
+fn validate_full_node_child_digests(
+    nodes: &BTreeMap<MastNodeId, MastNode>,
+    digests: &BTreeMap<MastNodeId, Word>,
+    num_nodes: usize,
+) -> Result<(), crate::serde::DeserializationError> {
```
bobbinth: Would be good to add a brief doc comment to this function.

Also, comment for the future: we could use the same pass to verify that all node digests are correct. This would basically make `SparseMastForest` serialization always untrusted, but I think this may be OK.

core/src/mast/serialization/sparse.rs
bobbinth, line 1: nit: could we add our usual separators to this file to make it a bit easier to navigate (e.g., constants, helper functions, per-struct headers etc.).

core/src/mast/serialization/sparse.rs
```diff
@@ -0,0 +1,400 @@
...
+        + usize::min_serialized_size()
+}
+
+/// Serializes a [`SparseMastForest`] in trusted sparse replay form.
+///
+/// This format carries the digest for each full node and accepts those digests on read. It is
+/// suitable for trusted remote proving inputs, not as an untrusted hashless validation path.
+///
+/// See <https://github.com/0xMiden/miden-vm/issues/3303> for the planned untrusted reader.
+pub(super) fn write_sparse_into<W: ByteWriter>(forest: &SparseMastForest, target: &mut W) {
```
bobbinth: Why is this `pub(super)`? It doesn't seem to be used anywhere outside of this file.

core/src/mast/serialization/sparse.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
+        SparseMastForest::read_from_bytes(bytes)
+    }
+}
+
+/// Reads a trusted sparse replay payload.
+///
+/// The payload carries full-node digests and digest-only entries as replay data. It does not
+/// rebuild those hashes from node structure.
+pub(super) fn read_sparse_from<R: ByteReader>(
```
bobbinth: Same comment as above: why is this `pub(super)` - it doesn't seem to be used anywhere outside of this file.

core/src/mast/serialization/sparse.rs
```diff
@@ -0,0 +1,400 @@
...
+    target.write_bytes(SPARSE_MAGIC);
+    target.write_bytes(&SPARSE_VERSION);
+
+    target.write_usize(forest.procedure_roots().len());
+    target.write_usize(forest.num_nodes());
+    target.write_usize(full_ids.len());
+    target.write_usize(forest.digest_entries().len());
+    target.write_usize(external_full_node_count);
+    target.write_usize(non_external_count);
+    target.write_usize(basic_block_data.len());
```
bobbinth: This is meant to be a kind of a header, right? I'm wondering if this is actually required because the use case for the `SparseMastForest` is pretty narrow (e.g., we do not intend them to save them to disk, or to have any kind of random access to them). So, we should be able to serialize most fields directly. The only exception being the basic blocks which we probably should serialize separately.

In the longer term, if we are able to get rid of `digests`, `advice_map`, `num_nodes` - the serialization could be just:

- Serialize `commitment` data.
- Serialize `roots` (as a plain vector).
- Serialize non-basic block nodes by iterating over `nodes` and serializing `(node_id, node)` directly.
- Serialize basic block nodes (as we do now).

This should be quite a bit simpler than what we have now.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - some applicable now and some probably best left for the future.

Overall, I think this works but we should be able to simplify the code quite a bit by making some opinionated assumptions and also simplifying the structure of the underlying `SparseMastForest` (e.g., it is not clear to me why `nodes` and `digests` are kept separately).

core/src/mast/sparse.rs
huitseeker: Fixed. `clippy::needless_qualification` + our xclippy config allows linting this systematically, but a 6-PR deep stack does not seem like the place.

core/src/mast/sparse.rs
huitseeker: Fixed. The field comment now says `num_nodes` is an upper bound for source IDs, not proof of source forest size.

core/src/mast/serialization/sparse.rs
huitseeker: Fixed

core/src/mast/sparse.rs
huitseeker: Fixed the doc comment in `validate_full_node_child_digests`. I agree that recomputing all node digests would make this closer to an untrusted sparse reader, but that fits the planned untrusted reader work (#3303), not this trusted sparse replay serialization PR.

core/src/mast/serialization/sparse.rs
huitseeker: Fixed.

core/src/mast/serialization/sparse.rs
huitseeker: FIxed

core/src/mast/sparse.rs
huitseeker: `AdviceReplay` covers advice reads in the #3314 replay path. `ExecutionReplay` stores `advice: AdviceReplay`, serializes it, and passes it into `ReplayProcessor`: [replay field](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/trace_state.rs#L303-L324), [replay serialization](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/trace_state.rs#L1878-L1902), [fragment split](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/parallel/mod.rs#L717-L752). `ReplayProcessor` uses `AdviceReplay` as its advice provider: [processor type](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/parallel/processor.rs#L452-L480).

The replay processor does not read advice maps from sparse forests. When replay resolves a `dyn` or external call, it looks up the next sparse forest by `MastForestId` and keeps running: [dyn resolution](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/parallel/processor.rs#L179-L207), [external resolution](https://github.com/0xMiden/miden-vm/blob/42e8e7135493dcf5018dccb10a7d806f4236437c/processor/src/trace/parallel/processor.rs#L209-L240). It does not extend the advice provider from `SparseMastForest.advice_map`.

So the serialized sparse `advice_map` is not used by replay trace generation. It is still present because `SparseMastForest` copied the dense `MastForest` shape, and `ExecutableMastForest` still requires every executable forest to return an advice map: [stored field](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/sparse.rs#L61-L65), [wire field](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/serialization/sparse.rs#L98-L103), [trait method](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/mod.rs#L743-L769). We can do a follow-up after #3314 to remove the sparse `advice_map` field and wire section, then either remove `advice_map` from `ExecutableMastForest` or make sparse forests return an empty map.

huitseeker: The sparse forest is a replay view. `ExecutionTracer::record_visit` records the executed node as `FullVisit` and its children as `DigestOnly`, because a parent row may need both child digests even when only one child runs: [recording rule](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/processor/src/trace/execution_tracer.rs#L209-L230). External nodes are different. Execution can reach them, and the VM resolves them into a target forest before the trace row starts: [external execution](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/processor/src/execution/external.rs#L31-L49), [tracer contract](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/processor/src/tracer.rs#L112-L128). Mixing the two would make skipped child nodes look like resolved external calls.

core/src/mast/sparse.rs
huitseeker: Agreed that `num_nodes` has no security value when bytes come from an untrusted sender. The sender controls it, so it cannot prove the original forest size. Its current value is narrower: it preserves the source forest ID range for sparse roundtrip shape, and the reader uses it to reject IDs outside that declared range: [sparse model](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/sparse.rs#L29-L37), [ID check](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/sparse.rs#L159-L170). If we do not care to preserve that shape, using `MastForest::MAX_NODES` is enough for the trusted reader, but I'm not sure that's the case.

core/src/mast/serialization/sparse.rs
huitseeker: Yes, this is a count header. It is not for disk storage or random access. #3313 uses it because the current sparse wire shape has several variable sections, and the reader needs counts before reading them: [header fields](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/serialization/sparse.rs#L68-L78), [count reads](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/serialization/sparse.rs#L147-L185).

core/src/mast/sparse.rs
bobbinth: > So the serialized sparse `advice_map` is not used by replay trace generation. It is still present because `SparseMastForest` copied the dense `MastForest` shape, and `ExecutableMastForest` still requires every executable forest to return an advice map: [stored field](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/sparse.rs#L61-L65), [wire field](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/serialization/sparse.rs#L98-L103), [trait method](https://github.com/0xMiden/miden-vm/blob/61bb5e80d6d82ef7a67d7e64b518502722c6a437/core/src/mast/mod.rs#L743-L769). We can do a follow-up after #3314 to remove the sparse `advice_map` field and wire section, then either remove `advice_map` from `ExecutableMastForest` or make sparse forests return an empty map.

Yes, makes sense, and both options sound reasonable. Without thinking too much about it, maybe having `ExecutableMastForest::advice_map()` return an `Option` is a better approach.

bobbinth: I'm not sure I follow the distinction. Let's say we have node `A` with children `B` and `C` such that, only child `B` is executed. Currently, we'd record node `C` as just `digest`. But we should be able to record it just as easily as an `External` node. This node will never be executed and so, the VM will not try to fetch the actual underlying MAST for it, but for the purposes of executing `A` and `B` it should be sufficient.

bobbinth (approved):
Looks good! Thank you! I've left some replies inline. Overall, this is fine to merge as is, but I do think we've overcomplicated both the `SparseMerkleTree` and its serialization quite a bit. This would be good to fix in the future. Let's create an issue for this.

core/src/mast/sparse.rs
huitseeker: At the end of the day, if we decide to store non-resolved nodes as `External` nodes, we lose track of the reason those nodes are absent: is it because they are absent of the original forest, or because the replay did not involve them? 

This is going to make debugging some issues with `SparseMastForest`s harder: looking at the `SparseMastForest` as the "cache" of an execution, a cache miss becomes hard to interpret: is it that the value is missing and shouldn't be (the `SparseMastForest` creator messed up) or that execution has a bug and should not query at that key in the first place (the `SparseMastForest` user is messing up)? 

That's a trade-off though, and for that, we should get something in return, namely the ability to not concern ourselves about `SparseMastForest`s as a separate case of `MastForest`, since given that `External` encoding of absent nodes, they can masquerade (from serialization and commitment Pov) as `MastForest`.

Besides not having to specialize the serialization, we can avoid separate digest-sorted sections in the `MastForest` serialization overall: today, dense `MastForest` serialization can recompute digest-sorted commitment inputs from its full node data, but `SparseMastForest` may omit source nodes that still matter to the source forest’s dependency commitment, so sparse serialization needs an explicit source dependency digest list, or an equivalent field, to prove that the sparse payload commits to the same external dependencies as the full source forest — using `External` nodes for that guarantees a simpler model from a commitment PoV.

This approach doe not have your [expressed preference](https://github.com/0xmiden/miden-vm/pull/3284#discussion_r3501991542).

core/src/mast/sparse.rs
huitseeker: Removed in the latest state of the PR

core/src/mast/serialization/sparse.rs
huitseeker: The sparse format no longer has a source-node-count header. It still has section counts because the trusted reader needs to read variable collections without scanning for delimiters: roots, full nodes, digest-only entries, and an advice-map count that must be zero. The writer is explicit about those sections: [writer](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/serialization/sparse.rs#L34-L44). The reader now mirrors that shape directly and rejects non-empty advice before parsing entries: [reader](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/serialization/sparse.rs#L160-L179).

core/src/mast/sparse.rs
huitseeker: `SparseMastForest` now carries an empty advice map. `SparseMastForestBuilder::finalize` does not copy advice from the source forest, and deserialization rejects any non-empty advice-map count before reading advice entries: [empty field docs](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/sparse.rs#L89-L94), [builder finalization](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/sparse.rs#L341-L370), [early sparse advice rejection](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/serialization/sparse.rs#L160-L179), [tests](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/core/src/mast/serialization/tests.rs#L850-L885). #3314 makes this is safe: `ExecutionReplay` carries `AdviceReplay`, serializes it, and `ReplayProcessor` uses it as its `AdviceProvider`: [replay field](https://github.com/0xMiden/miden-vm/blob/0fd51afaa8b9a82bfc91aa617adf6035033ec3fc/processor/src/trace/trace_state.rs#L303-L324), [replay serialization](https://github.com/0xMiden/miden-vm/blob/0fd51afaa8b9a82bfc91aa617adf6035033ec3fc/processor/src/trace/trace_state.rs#L1878-L1902), [processor provider type](https://github.com/0xMiden/miden-vm/blob/0fd51afaa8b9a82bfc91aa617adf6035033ec3fc/processor/src/trace/parallel/processor.rs#L452-L480).

huitseeker: Kept digest-only entries. Digest-only entries are skipped children in the same source forest. `ExecutionTracer::record_visit` records the executed node as full and records its immediate children as digest-only, because a parent row may need child digests even when execution enters only one child: [recording rule](https://github.com/0xMiden/miden-vm/blob/37edd400e3cf03798c57d9eeb7b7e1ff883666a9/processor/src/trace/execution_tracer.rs#L209-L230). This also keeps accidental entry into a pruned child as a `get_node_by_id` miss.

PR #3314, opened 2026-06-30
2/3 Serialize trace proving inputs

huitseeker:
Part of #3235.

This PR serializes trusted trace proving inputs for remote proving.

It writes the trace generation context, sparse MAST forest store, replay data, proving options, and precompile transcript state.

Readers check the stored shape before use, including fragment size, stack depth, and MAST forest IDs.

This PR does not add fuzz targets or proptest coverage.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - [BREAKING] Added `trace`, `trace.CONST`, and `trace.event("...")` assembly as syntactic sugar for emitting optional read-only trace events. This adds variants `Trace` and `TraceImm` to the public enum `miden_assembly_syntax::ast::Instruction` ([#3478](https://github.com/0xMiden/miden-vm/pull/3478)).
 - Added `Mmr::nodes_from(start)`, returning the MMR's nodes at indices `start..` in insertion (postorder) order ([#3585](https://github.com/0xMiden/miden-vm/pull/3585)).
 - [BREAKING] Added `Mmr::from_nodes_unchecked(forest, nodes)`, constructing an MMR from its complete postorder node array without recomputing hashes ([#3585](https://github.com/0xMiden/miden-vm/pull/3585)).
+- Added trusted trace proving input serialization for remote proving ([#3314](https://github.com/0xMiden/miden-vm/pull/3314)).
```
bobbinth: Should we update phrasing here since "trace proving input" is outdated phrasing now?

crates/test-serde-macros/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "miden-test-serde-macros"
+version = "0.30.0"
+description = "Proc macros for serde roundtrip testing in Miden VM"
+readme = "README.md"
```
bobbinth: Could we add a brief `README.md` file for this crate?

crates/test-serde-macros/src/lib.rs
bobbinth, line 1: I did not review this file in detail.

processor/src/trace/trace_state/arbitrary.rs
bobbinth, line 1: I did not review this file in detail.

processor/src/continuation_stack.rs
```diff
@@ -412,6 +432,215 @@ impl<F> ContinuationStack<F> {
...
+            },
+            Self::EnterForest {
+                forest,
+                package_debug_info: _,
+                inline_context_depth,
+            } => {
+                TAG_ENTER_FOREST.write_into(target);
+                forest.write_into(target);
+                inline_context_depth.write_into(target);
+            },
```
bobbinth, line 502: We should document somewhere that we don't serialize `package_debug_info` (which I think is correct) and so, round-tripping any struct that contains `Continuation` may not be exact.

processor/src/continuation_stack.rs
```diff
@@ -412,6 +432,215 @@ impl<F> ContinuationStack<F> {
...
+
+fn read_mast_node_id<R: ByteReader>(source: &mut R) -> Result<MastNodeId, DeserializationError> {
+    Ok(MastNodeId::from(u32::read_from(source)?))
+}
+
+impl Serializable for ContinuationStack<MastForestId> {
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        self.stack.write_into(target);
+        self.source_node_ids.write_into(target);
+    }
```
bobbinth, line 545: Question: why do we need to serialize `source_node_ids`? My understanding is that the rest of the debug info is not serialized, and so, would this cause any issues (or be used somehow) during proving?

processor/src/trace/chiplets/ace/trace.rs
```diff
@@ -293,3 +310,398 @@ impl WireBus {
...
+
+fn read_quad_felt<R: ByteReader>(source: &mut R) -> Result<QuadFelt, DeserializationError> {
+    let c0 = Felt::read_from(source)?;
+    let c1 = Felt::read_from(source)?;
+    Ok(QuadFelt::new([c0, c1]))
+}
+
+fn quad_felt_min_serialized_size() -> usize {
+    Felt::min_serialized_size() * 2
+}
```
bobbinth: Question: why have these as free functions rather than as implementations of `Serializable` / `Deserializable` on relevant structs?

processor/src/trace/mod.rs
bobbinth, line 62: Maybe not for this PR, but should we add a version to `ExecutionWitness`? I'm assuming we'd probably be changing serialization formats for this struct once in a while, and it may be good to know which version of a struct we are dealing with.

For now, we could probably just add a version fields - or maybe even just serialize a single allowed version value (e.g., `1`). This would simplify introducing additional versions in the future.

processor/src/trace/trace_state.rs
```diff
@@ -1387,6 +1527,676 @@ impl StackOverflowReplay {
     }
 }
 
+// SERIALIZATION
+// ================================================================================================
```
bobbinth: This file is already huge - could we move the serialization code into a sub-module? For example, `src/trace/trace_state/serde.rs`.

Also, I'd move `src/trace/trace_state.rs` into `src/trace/trace_state/mod.rs` to keep all related files together - but maybe that's for another PR.

processor/src/trace/trace_state.rs
```diff
@@ -1387,6 +1527,676 @@ impl StackOverflowReplay {
...
+
+/// Serializable view over a [`VecDeque`].
+///
+/// This uses the same wire shape as `Vec<T>`: a length prefix followed by items in iteration order.
+struct SerializableVecDeque<'a, T>(&'a VecDeque<T>);
```
bobbinth: Why do we need this wrapper? Couldn't we implement `Serializable` / `Deserializable` directly on `VecDeque`?

prover/src/lib.rs
```diff
@@ -34,6 +42,129 @@ pub use miden_processor::{
...
+/// proof-generation options. Deserialization checks malformed structure and bounded allocation, but
+/// sparse MAST node and digest maps are accepted as replay data and are not checked against a
+/// source [`miden_core::mast::MastForest`] commitment.
+///
+/// See <https://github.com/0xMiden/miden-vm/issues/3303> for the planned untrusted reader.
+#[derive(Debug)]
+pub struct TraceProvingInputs {
+    witness: ExecutionWitness,
+    hash_fn: HashFunction,
+}
```
bobbinth: I'm not sure we need to introduce this struct. I think having `ExecutionWitness` be serializable should be sufficient and we can then just use the regular `Prover` interface to generate proofs for execution witnesses.

And the way the desired hash function would be communicated to the prover could be done in a number of ways (including being a part of a given prover's settings).

prover/src/lib.rs
```diff
@@ -34,6 +42,129 @@ pub use miden_processor::{
...
+#[instrument("prove_partial_from_trace_sync", skip_all)]
+pub fn prove_partial_from_trace_sync(
+    inputs: TraceProvingInputs,
+) -> Result<(StackOutputs, ExecutionProof), ExecutionError> {
+    let (witness, hash_fn) = inputs.into_parts();
+    let stack_outputs = *witness.claim().stack_outputs();
+    let prover = Prover::new().with_hash_fn(hash_fn);
+    let proof = prover.prove(witness).map_err(ProverError::into_execution_error)?;
+    Ok((stack_outputs, proof))
+}
```
bobbinth: Related to the above comment, I'm not sure we need these functions.

bobbinth (commented):
Looks good! Thank you! Not a super detailed review yet, but I left some comments inline.

processor/src/trace/trace_state.rs
huitseeker: It's an artefact of the history of this PR, which predates the move of crypto here. This satisfies the orphan rule.

processor/src/trace/chiplets/ace/trace.rs
huitseeker: Same as `SerializableVecDeque`: `QuadFelt`/`RowIndex` are foreign

processor/src/continuation_stack.rs
huitseeker: Dropped them!

CHANGELOG.md
```diff
@@ -7,17 +7,21 @@
...
 - Added `LinkMode::Analysis` and `Linker::link_analysis`, which commit resolved modules and call edges and report a static recursion cycle as a nonfatal diagnostic (`LinkAnalysis`) instead of rejecting it. Strict linking is unchanged: it still rejects cycles before MAST is built and rolls back on failure ([#3535](https://github.com/0xMiden/miden-vm/pull/3535)).
 - Added `AdviceMutation::extend_advice_stack_with`, which takes an `IntoIterator<Item = Felt>` so that small host replies no longer have to build an `AdviceStack` first ([#3543](https://github.com/0xMiden/miden-vm/pull/3543)).
+- Added trusted binary serialization of execution witnesses (VM witness plus optional singleton precompile witness) for remote proving, with a wire format version tag and bounded, trusted deserialization ([#3314](https://github.com/0xMiden/miden-vm/pull/3314)).
 
 #### Changes
 
+- [BREAKING] Removed `TraceProvingInputs`, `prove_from_trace_sync`, and
+  `prove_partial_from_trace_sync`. Deserialize `ExecutionWitness` directly and pass it to the
+  regular `Prover` interface ([#3314](https://github.com/0xMiden/miden-vm/pull/3314)).
```
bobbinth: Is this entry correct? I don't think we had `TraceProvingInputs` prior to this PR.

prover/Cargo.toml
bobbinth, line 1: Are the changes in this file necessary? AFAICT, this PR does not change any code in the `prover` crate.

processor/src/trace/trace_state.rs
```diff
@@ -1402,3 +1543,8 @@ mod tests {
         assert!(ops.next().is_none());
     }
 }
+
+#[cfg(feature = "arbitrary")]
+mod arbitrary;
+
+mod serde;
```
bobbinth: nit: I'd move these to the top of the file as this is usually where we put sub-module declarations.

processor/src/trace/trace_state.rs
```diff
@@ -1240,6 +1357,26 @@ impl HasherRequestReplay {
         self.record(HasherOp::UpdateMerkleRoot((old_value, new_value, path, index)));
     }
 
+    pub(crate) fn iter_hash_basic_block_forest_ids(
+        &self,
+    ) -> impl Iterator<Item = MastForestId> + '_ {
```
bobbinth, line 1281: Could we add a brief doc comment explaining what this method does and why it is needed?

processor/src/trace/mod.rs
```diff
@@ -148,6 +221,71 @@ impl VmWitness {
...
+    }
+    let deferred_wire = DeferredStateWire::read_from(source)?;
+    let deferred_state =
+        DeferredState::from_wire(Arc::new(miden_precompiles::registry()), &deferred_wire).map_err(
+            |err| DeserializationError::InvalidValue(format!("invalid deferred state: {err}")),
+        )?;
+
+    let witness = PrecompileWitness::new(deferred_state).map_err(|err| {
+        DeserializationError::InvalidValue(format!("invalid precompile witness: {err}"))
+    })?;
```
bobbinth, line 280: Not related to this PR: what is the purpose of the `DeferredStateWire` struct? That is, why don't we just serialize/deserialize `PrecompileWitness` directly (instead of going through the `DeferredStateWire` struct)?

bobbinth (approved):
Looks good! Thank you! I left a few comments inline - but nothing major.

Also, I didn't do a line-by-line review of all serialization code - so, it may be good if someone else takes a look at it more closely.

adr1anh (approved):
Looks good!

crates/test-serde-macros/Cargo.toml
huitseeker: The crate is gone now. It only supported the property tests, which are part of #3315, so #3314 no longer needs the crate or a README

crates/test-serde-macros/src/lib.rs
huitseeker: This file has been removed from #3314 with the rest of the test macro crate.

processor/src/trace/trace_state/arbitrary.rs
huitseeker: This file has been removed from #3314. The property test coverage now lives in #3315 and works with `ExecutionWitness` directly

processor/src/continuation_stack.rs
huitseeker: The comments on `Continuation` and `ContinuationStack` now say that `package_debug_info` and `source_node_ids` are omitted. Deserialized continuations use `None` for that metadata, so the round trip is intentionally lossy.

processor/src/trace/mod.rs
huitseeker: Done

prover/src/lib.rs
huitseeker: I removed `TraceProvingInputs` and the extra proving entry points. `ExecutionWitness` is serializable and goes through the regular `Prover::prove` interface.

prover/src/lib.rs
huitseeker: Removed with `TraceProvingInputs`. The regular `Prover` API consumes `ExecutionWitness` now.

CHANGELOG.md
huitseeker: Correct, removed.,

prover/Cargo.toml
huitseeker: I reverted the changes to `prover/Cargo.toml` after removing the property test support from #3314.

processor/src/trace/trace_state.rs
huitseeker: The `serde` submodule declaration is now at the top of the file.

processor/src/trace/trace_state.rs
huitseeker: The doc comment now says this returns forest IDs from buffered `HashBasicBlock` operations so deserialization can reject IDs outside `TraceReplay::mast_forest_store.` In streaming mode, those operations are resolved and sent to the trace builder immediately, so the replay does not retain any forest IDs.

processor/src/trace/mod.rs
huitseeker: `DeferredStateWire` is the passive, canonical transport form of the deferred DAG. `PrecompileWitness` contains a hydrated `DeferredState`, _including state tied to a registry_. Keeping a wire type lets proof decoding read the bytes without choosing a registry. 

Verification later rebuilds the state under the installed registry and checks its structure, size limit, canonical form, and root evaluation.

Another advantage is that rehydration builds a fresh `DeferredState` and returns it only after all checks pass. If any check fails, the caller receives an error and cannot observe a partly rebuilt state.

processor/src/trace/chiplets/ace/trace.rs
huitseeker: Small correction to my earlier reply. The replay structs now implement `Serializable` and `Deserializable` directly. `VecDeque` uses the generic implementations in `miden-serde-utils`, so that wrapper is gone. The remaining helpers are only for `RowIndex` and `QuadFelt`. Moving those implementations to `miden-serde-utils` would create a dependency cycle because `miden-core` depends on `miden-serde-utils`, and `miden-air` depends on `miden-core`.

processor/src/trace/trace_state.rs
huitseeker: `VecDeque` now uses the generic `Serializable` and `Deserializable` implementations in `miden-serde-utils`, so I removed `SerializableVecDeque` and serialize these queues directly.

PR #3318, opened 2026-07-02
Remove MastForest compact API

### LINKED ISSUE #3312: Remove `MastForest::compact()` and associated code
After recent refactorings, `MastForest::compact()` should no longer be needed as MAST forests should be built/merged without duplicate nodes (one exception to this is basic block nodes where same set of instructions have different error codes, but `compact()` doesn't merge these ones either).

It is possible that I'm missing something and we still need `compact()` for now, but that's probably an indication of a deeper issues that we should also resolve.

As a part of this, we may be able to get rid of methods like `MastForestContributor::add_to_forest()` - because direct mutations of MAST forest should not be possible. But these methods are heavily used in testing, and so, maybe the best we can do for now is to put them behind the testing flags.

huitseeker:
Removes the public `MastForest::compact()` API and updates the remaining tests to exercise the underlying behavior through explicit `MastForest::merge([&forest])`.

There's no in-repo no production caller of `MastForest::compact()`. The only MAST compaction call sites were tests.

## Why this is safe

Normal MAST construction now goes through `MastForestBuilder`, which interns pending nodes before finalizing the forest. That path should not produce duplicate reachable nodes that need a later public compaction pass.

Duplicates can still enter a `MastForest`, but only through lower-level construction paths that bypass the builder’s interning, for example:

- manually constructing a `MastForest` with `MastForest::new()`
- adding equivalent nodes directly with node builders’ `add_to_forest()`
- marking both equivalent nodes as roots
- calling explicit merge normalization with `MastForest::merge([&forest])`

That last case is still supported because it is merge behavior, not a separate compaction API. The tests that mattered for merge correctness now call `MastForest::merge([&forest])` directly.

The tests remain to guarantee/demonstrate same-digest nodes with different runtime error codes must remain distinct.

Fixes #3312.


bobbinth (approved):
Looks good! Thank you!

PR #3329, opened 2026-07-03
refactor: Build dense MAST forests through finalization

### LINKED ISSUE #3325: Make dense MastForest construction produce canonical node order
Dense `MastForest` values should have one node order in memory.

The ordered objects are the entries in `MastForest.nodes`.

The order is:

1. External nodes first.
  
2. External nodes are sorted by digest.
  
3. Basic block nodes next.
  
4. Internal nodes last.
  
5. Every child node appears before its parent.
  

This is not only a wire rule. It should be true for every dense `MastForest` after normal construction.
## Baseline
This issue is about the code in PR #3311 at commit `1df50c545b262097f5107be1dca391cf06c4c9e1`.

PR #3311 already has a canonicalization helper. `MastForest::from_parts()` [sorts node IDs by node class, sorts external nodes by digest, remaps child and root IDs, and validates the result.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/mod.rs#L172-L191) The sort itself lives in `canonicalize_dense_parts()`[.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/mod.rs#L229-L289)

The [validator checks the same order. It rejects a node class that appears after a later class, external digests that are not strictly increasing, and a child ID that points to the current node or to a later node.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/mod.rs#L308-L350)

[Trusted dense deserialization already expects this order. It validates the node order before it accepts the forest.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/mod.rs#L194-L210)
## Problem
The existing assembly `MastForestBuilder` already produces a `MastForest`. The missing rule is stricter than that.

After this change, normal dense construction must not expose a `MastForest` that is only append ordered. A builder may append nodes internally while it is building, but it must finalize through the canonical `MastForest` order before callers can use the forest.

As of PR #3311, the call graph still has direct append paths into `MastForest`.

`MastForestContributor` [exposes](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/node/mast_forest_contributor.rs#L49-L72) `add_to_forest(self, &mut MastForest)`[.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/node/mast_forest_contributor.rs#L49-L72) `ExternalNodeBuilder::add_to_forest()` [writes directly to](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/node/external.rs#L156-L164) `forest.nodes.push(...)` [and then recomputes the forest commitment.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/node/external.rs#L156-L164)

Other node builders implement the same trait. They can append nodes in caller order. The trait does not require external nodes before basic blocks or internal nodes.

A caller can therefore build an append ordered forest. For example, a caller can add a basic block, then add an external node. The forest exists as a `MastForest`, but `MastForest.nodes` is not in canonical order.

`MastForestMerger` is one normal path that still treats `MastForest` as its mutable build target. [It creates](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/merger/mod.rs#L47-L68) `MastForest::new()` [and returns that forest.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/merger/mod.rs#L47-L68) [It appends remapped nodes with](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/merger/mod.rs#L176-L186) `add_to_forest()`[.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/merger/mod.rs#L176-L186)

The existing assembly `MastForestBuilder` also needs one more property. [Its finalizer must return the ID map after canonicalization, and all source and debug metadata must use the final IDs.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/crates/assembly/src/mast_forest_builder/finalizer.rs#L176-L205) Without that, canonicalization can move nodes after the builder has already recorded metadata.
## Current repair paths
PR #3311 keeps repair paths because append ordered forests can still reach commitment and serialization code.

`compute_dependency_commitment()` [first checks whether dense node order is valid. If it is not valid, it filters all external nodes and sorts their digests before hashing them.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/mod.rs#L614-L632)

`write_into_with_options()` [also checks the memory order. If the forest is out of order, it rebuilds a canonical forest with](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L233-L250) `MastForest::from_raw_parts()` [and writes that forest instead.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L233-L250)

These repair paths keep the bytes and commitments correct. They also show that the in-memory invariant is not yet enforced at the construction boundary.
## What is ordered in PR #3311
In memory, `MastForest.nodes` is canonical only after construction goes through `MastForest::from_parts()` or trusted dense deserialization. A forest built by direct `MastForestContributor::add_to_forest()` calls can be in append order.

On the wire, dense serialization writes canonical node entries. If memory is out of order, PR #3311 canonicalizes before writing.

[The serialized root ID list is not sorted by digest. It preserves root order.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L269-L270)

[The serialized root digest section is sorted by digest.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L309-L329) [The reader recomputes root digests from the root IDs, sorts them, and checks the section.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/resolved.rs#L207-L226)

[The serialized dependency digest section follows the external node prefix.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/mod.rs#L262-L300) [The reader checks that the section is strictly sorted and matches the external nodes.](https://github.com/0xMiden/miden-vm/blob/1df50c545b262097f5107be1dca391cf06c4c9e1/core/src/mast/serialization/resolved.rs#L228-L248)
## Change
Keep append order inside builders only.

When a builder returns a dense `MastForest`, the returned forest must already use canonical node order. Any node ID map returned by that builder must use final IDs after canonicalization.

Move normal dense construction paths that still append into `MastForest` onto a builder or finalization path with this property.

After this change, dense serialization should expect canonical memory order. It should fail if the forest is not canonical. It should not sort or rebuild the forest before writing.

The dependency commitment should read the external node prefix directly. It should not need a fallback that scans and sorts all external nodes.
## Prior comments
@bobbinth has asked for this rule to become an in-memory invariant. A serialization rule alone is not enough.

Relevant comments:

1. [https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122440986](https://github.com/0xMiden/miden-vm/pull/2765#discussion_r3122440986)
  
2. [https://github.com/0xMiden/miden-vm/issues/3043#issuecomment-4323023661](https://github.com/0xMiden/miden-vm/issues/3043#issuecomment-4323023661)
  
3. [https://github.com/0xMiden/miden-vm/issues/2473#issuecomment-4416373894](https://github.com/0xMiden/miden-vm/issues/2473#issuecomment-4416373894)
  
4. [https://github.com/0xMiden/miden-vm/pull/3311#discussion_r3516827869](https://github.com/0xMiden/miden-vm/pull/3311#discussion_r3516827869)

huitseeker:
Closes #3325. On top of #3311. Re-do of #3326 after misclick

Dense `MastForest` values now leave normal construction in canonical node order. Builders may still append while they are building, but they must call finalization before returning a `MastForest`.

This PR adds `DenseMastForestBuilder` for dense construction. `finish()` and `finish_with_id_map()` call `MastForest::from_parts_with_id_map()`, which orders `MastForest.nodes` as external nodes, then basic blocks, then internal nodes. External nodes are sorted by digest. Children stay before parents. The returned ID map uses the final node IDs.

`MastForestMerger` now builds through that builder and remaps root maps after finalization. The assembly `MastForestBuilder` finalizer now applies the final ID map before it builds source and debug metadata.

The old append API is no longer available in normal builds. `add_to_forest()` and `make_root()` are limited to tests and `arbitrary`.

Serialization now expects the in-memory order to be valid. It no longer rebuilds a canonical forest before writing. The dependency commitment reads the external node prefix directly.

Public examples now use `DenseMastForestBuilder`, so docs do not point callers back to the append API.


huitseeker:

PR #3330, opened 2026-07-03
test: check core package version matches crate

huitseeker:
Adds a test that `CoreLib` isn't dragging behind the crate version. We can later edit it to be looser to encode any policy we choose in the future.


bobbinth (approved):
Looks good! Thank you!

Issue #3331, opened 2026-07-03
Assembler panics instead of erroring when the program entrypoint has locals set via the AST API


thep2p:
An executable module's `begin..end` block is compiled as an ordinary procedure and is therefore reachable through `Module::procedures_mut`. The `@locals(..)` grammar can't attach locals to `begin`, so the parser can never produce  an entrypoint with locals. 

However, `Procedure::set_num_locals` can:
```rust
let mut module = /* parse an executable module: begin .. end */
for proc in module.procedures_mut() {
      proc.set_num_locals(4);
}

Assembler::new(source_manager).assemble_program("test", module);
```

This aborts with a panic rather than returning an error, because the invariant is enfourced with an assert in `compile_procedure` [here](https://github.com/thep2p/miden-vm/blob/5e625127a23804f50720dc3f185406441689c70d/crates/assembly/src/assembler.rs#L1356):

```rust
assert!(num_locals == 0, "program entrypoint cannot have locals");
```
Any non-zero value triggers it. I guess the assert was written when the text parser was the only way to construct the module, so _an entry point with locals_ could only mean an internal bug. However, with the programmatic AST path it's reachable from user input and should be a diagnostic error instead. 

Maybe a fix can be replacing the assert with a proper error at assembler boundary. More generally (and this is the same point as #3323): invariants the parser enforces on `.masm` source may be worth re-validating once at AST --> assembler boundary, since `Module`s built or mutated programmatically skip the parser entirely. 

cc @huitseeker

bitwalker:
It's currently implicit that AST producers are required to uphold certain MASM invariants (which are then asserted at various points). However, what is supported by the MASM AST is deliberately a superset of what is permitted in MASM itself, to support lowering from high level languages.

Asserts like this are intentional - they exist because they represent unrecoverable bugs in the producer. This policy is similar to those of other compiler backends, e.g. LLVM, which also assert if you construct the IR incorrectly. While we make it harder to do that with MASM, we cannot prevent all incorrect uses at the AST API level without excessive wasteful error handling boilerplate that can't be meaningfully handled by the caller, and would never hit the error path in well-behaved producers (such as the Miden compiler).

It isn't clear to me that this being caught by asserts is the issue. I do think we can move/copy the assert in this case closer to where the invalid construction is produced. We should also ensure this invariant is documented on `set_num_locals`.

thep2p:
[...]

PR #3332, opened 2026-07-05
fix(assembly): reject procedures with more than 65532 locals

### LINKED ISSUE #3323: @locals overflow fix (#2829) doesn't cover procedures built via the AST API
PR #2838 (closing #2829) added a bounds check for `@locals(..)` in the text parser (then `parser/grammar.lalrpop`; after the CST-parser rewrite the check lives in `crates/assembly-syntax/src/parser/cst/forms.rs`), rejecting values above 65532. That
only covers procedures compiled from `.masm` source.

The same overflow is still reachable through the public AST API. `Procedure::new` and `Procedure::set_num_locals` take a raw `u16` with no validation:

```rust
  // crates/assembly-syntax/src/ast/procedure/procedure.rs
  pub fn set_num_locals(&mut self, num_locals: u16) {
      self.num_locals = num_locals;
  }
```

A `Module` built or mutated this way and passed to `Assembler::assemble_library` /`assemble_program` reaches the same frame-pointer codegen the original issue was about:

```rust
// crates/assembly/src/fmp.rs
let locals_frame = Felt::from_u16(num_locals.next_multiple_of(WORD_SIZE as u16));
```

For `num_locals` in `65533..=65535`, `next_multiple_of(4)` overflows `u16`:
- with overflow checks on (debug default): panics with `attempt to add with overflow`
- with overflow checks off (release default, including the `optimized` CLI profile): wraps to `0`, so the frame pointer is never advanced — every subsequent `loc_load`/`loc_store`/`locaddr` in that procedure resolves to the wrong address

The same rounding is repeated in `crates/assembly/src/instruction/mem_ops.rs` (`local_to_absolute_addr`), where `aligned_num_locals - index_of_local` is then miscomputed once `aligned_num_locals` wraps to 0 (subtract-overflow for index > 0,
a silently wrong zero offset for index 0).

Repro:

```rust
  let mut module = /* parse a module with @locals(1) */;
  for proc in module.procedures_mut() {
      proc.set_num_locals(65535);
  }
  Assembler::new(source_manager).assemble_library("test", module, None::<Box<Module>>);
  // panics in debug; silently corrupts local addressing with overflow checks off
```

cc @huitseeker

thep2p:
## Describe your changes

Closes #3323

### Rationale

`Procedure::new` / `set_num_locals` accept any `u16`, bypassing the parser's 65532 cap (#2829 / #2838). Values in `65533..=65535` then overflow the word-alignment rounding in frame-pointer codegen: panic with overflow checks on, silent wrap to a zero-size frame with them off — every `loc_load`/`loc_store`/`locaddr` in the procedure then resolves to the wrong address.

### Changes

Adds a `MAX_PROC_LOCALS` guard in `compile_procedure`, before the frame-pointer sequence is generated, returning a diagnostic error that reports both the declared and maximum count. One guard covers both affected sites (`fmp.rs`, `mem_ops.rs`), since all assembly paths funnel through it.

### Test plan

- `cargo test -p miden-assembly --features testing num_locals` — 3 new tests: rejection at 65535 and at `MAX_PROC_LOCALS + 1`, acceptance at exactly `MAX_PROC_LOCALS`; rejection tests pin the `TooManyProcedureLocals` diagnostic text.
- Full `miden-assembly` suite, `make clippy`, and `make format-check` pass.

- Updated `CHANGELOG.md`


thep2p:
One difference from #3331: this invariant isn't backed by an assert in release builds. `next_multiple_of` only panics with overflow checks on; with them off (release, including the `optimized` profile) it's documented to wrap, so the frame prologue is emitted as `push(0)` and every local access silently resolves against the caller's frame — assembly still reports success. Some always-on check is needed either way; I went with a diagnostic error since `compile_procedure` already returns `Report`, but happy to make it an assert if that fits the policy better.

bitwalker (approved):

bitwalker:
@thep2p Can you rebase this on `next`?

crates/assembly/src/tests.rs
```diff
@@ -7603,3 +7603,96 @@ fn test_linking_recursive_expansion_via_renamed_aliases() -> TestResult {
 
     Ok(())
 }
+
+#[test]
+fn test_num_locals_above_max_is_rejected() -> TestResult {
```
huitseeker: Nit: These three tests repeat the same module fixture, `parse_module`, `set_num_locals`, and `assemble_library(..., None::<Box<Module>>)`. Could this use a small helper like `assemble_library_with_num_locals(&context, num_locals)` so the boundary cases stay focused on the values under test?

huitseeker (approved):
LGTM! Nice seeing you here @thep2p

crates/assembly/src/assembler.rs
```diff
@@ -1327,6 +1335,20 @@ impl Assembler {
...
+        if num_locals > MAX_PROC_LOCALS {
+            let span = proc_ctx.span();
+            let source_file = proc_ctx.source_manager().get(span.source_id()).ok();
+            return Err(Report::new(AssemblerError::TooManyProcedureLocals {
+                span,
+                source_file,
+                max_locals: MAX_PROC_LOCALS,
+                num_locals,
+            }));
+        }
```
bobbinth: This works but I wonder if a check a bit further upstream could be more robust. Specifically, we could ensure that `ProcedureContext::with_num_locals()` could return an error - though, maybe this would make error reporting a bit more complicated.

bobbinth (approved):
Looks good! Thank you! I left one optional comment inline.

crates/assembly/src/tests.rs
thep2p: Good call! extracted `assemble_library_with_num_locals` in https://github.com/0xMiden/miden-vm/pull/3332/commits/61e414cc2545d73bd1faab22c74fbae2c9597bd3

crates/assembly/src/assembler.rs
thep2p: Done in https://github.com/0xMiden/miden-vm/pull/3332/commits/c5c0fd7e417bf3395f35a27920a7f19bed0ac5bd `with_num_locals` now returns an error. Only wrinkle was the one you anticipated where the span used to be set after the local count, so I swapped the chain order to keep the diagnostic pointing at the procedure. Dropped the old check in                `compile_procedure`.

thep2p:
> @thep2p Can you rebase this on `next`?

@bitwalker applied comments and rebased on `next`

crates/assembly/src/assembler.rs
bobbinth: Thank you!

PR #3319, opened 2026-07-02
perf(air): drop redundant transition guards on op-flag-gated constraints

Nashtare:
## Describe your changes

Address longstanding TODO about transitional constraints. Shaves off a few % from constraint evaluation stage too.

ACE circuit: 5548 -> 5404 gates

## Checklist before requesting a review
- Repo forked and branch created from `next` according to naming convention.
- Commit messages and codestyle follow [conventions](../CONTRIBUTING.md).
- Commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- Relevant issues are linked in the PR description.
- Tests added for new functionality.
- Documentation/comments updated according to changes.
- Updated `CHANGELOG.md'

PR #333, opened 2022-07-26
Use faster u32 instructions in BLAKE3 2-to-1 hashing

itzmeanjan:
This PR uses newly introduced `u32unchecked_rotr.*`, which is faster compared to its checked counterpart --- it improves performance of Blake3 2-to-1 hashing by ~12.41%.

At commit 05bee21 : 5487
At commit 8e9b640 : 4806 [ **latest** ]

PR #3333, opened 2026-07-06
perf(air): merge one-hot gated stack op constraints

Nashtare:
## Summary

Folds the per-op stack rewrite constraints in `air/src/constraints/stack/ops.rs` into a single constraint per next-row stack position, and routes `HORNERBASE`/`HORNEREXT` preservation through the general no-shift mechanism instead of explicit per-element equality checks. Both changes reduce the AIR's base constraint count and the recursive verifier's ACE circuit size without changing behavior or max constraint degree.

- Core AIR base constraints: **338 → 239**
- ACE circuit eval gates: **5548 → 5276**
- Max constraint degree: unchanged at **9**
- `RELATION_DIGEST` changes accordingly (changes the recursive verifier program hash) — constraint artifacts regenerated and committed.

### 1. Fold one-hot gated stack ops (`stack/ops.rs`)

Most ops in this file only rewrite a handful of next-row stack positions (`PAD`, `DUP*`, `CLK`, `SWAP`, `MOVUP*`, `MOVDN*`, `SWAPW`/`SWAPW2`/`SWAPW3`/`SWAPDW`, `CSWAP`/`CSWAPW`, `CALLER`, `SDEPTH`), and were each constrained with their own `builder.when(op_flag)` call. Since op flags are one-hot (each is a product of selectors over the same 7 binary op bits for a distinct opcode), at most one op is ever live on a given row.

This lets all the ops touching a given next-row position `i` be folded into one constraint of the form:

```
next[i] * flag_sum == Σ_op flag_op * source_op
```

where `flag_sum` is the sum of op flags that write position `i`, and the right-hand side sums each op's flag times its source expression. Because flags are one-hot, only the live op's term survives on any row, so this is equivalent to the original per-op constraints but costs one gate per stack position instead of one gate per (op, position) pair.

This also drops a `when_transition()` wrapper that had become redundant after #3319 (op flags already vanish on the boundary row).

### 2. Route `HORNERBASE`/`HORNEREXT` through the no-shift composite flags (`op_flags/mod.rs`, `stack/crypto.rs`)

`HORNERBASE` and `HORNEREXT` only ever rewrite the accumulator at `s[14..16]`; `s[0..14]` is always preserved. Previously this was enforced with 2×14 explicit `assert_eq(s_next[i], s[i])` constraints in `crypto.rs`. Both opcodes are now added to the `no_shift` composite flags at depths `0..14` (and removed from the accumulator's now-shifting depths `14`/`15`), so the general stack transition constraints cover this preservation for free — the per-op code only needs to constrain the accumulator update itself. `stack_route_tests.rs` and doc comments updated to match.

### 3. Regenerate constraint artifacts

Regenerated via:

```
cargo run -p miden-core-lib --features constraints-tools --bin regenerate-constraints -- --write
```

Updates `air/src/config.rs` (gate count / `RELATION_DIGEST`), the corresponding insta snapshot, and `crates/lib/core/asm/sys/vm/{constraints_eval,mod}.masm`. **This changes the recursive verifier's program hash** — anything pinned to the old `RELATION_DIGEST` needs to be regenerated against this branch.

## Testing

- `cargo test -p miden-air` — all unit and `ace_codegen` integration tests pass, including `relation_digest_matches_current_air`.
- `cargo build --workspace` — clean.

## Risk / review notes

- Purely a constraint-encoding refactor: no change to VM semantics, stack ISA, or op behavior — only how the constraints are algebraically expressed and packed into the ACE circuit.
- The `RELATION_DIGEST` bump is expected and load-bearing (smaller circuit); reviewers should confirm no downstream artifact hardcodes the old digest.
- Worth double-checking the folded `flag_sum` / `expected` expressions in `ops.rs` position-by-position against the original per-op constraints, since an error in the fold could silently change behavior only for less-common op combinations that happen to collide at the same stack position.

- Updated `CHANGELOG.md'


huitseeker (commented):
The rewrite now folds many per-op transition constraints into shared per-
position constraints, but the tests still only exercise the `MSTREAM`/`PIPE` cursor case. 

Is there a way to add a table-driven AIR unit tests that build the expected next-stack rows for each affected opcode, and verify both acceptance and rejection after mutating each rewritten position?

air/src/constraints/stack/ops.rs
```diff
@@ -290,26 +381,413 @@ mod tests {
...
+                opcode: opcodes::DUP13,
+                setup: base_setup,
+                expected: Vec::from([(0, s(13))]),
+            },
+            StackOpCase {
+                name: "DUP15",
+                opcode: opcodes::DUP15,
+                setup: base_setup,
+                expected: Vec::from([(0, s(15))]),
+            },
```
adr1anh: Would it make sense for these ops (and other numbered ones below), to extend the vector by iterating over the number of indices?

adr1anh (approved):
Looks really good!

huitseeker (approved):
Lovely, tyvm!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Fixed a panic (or silent miscompile in release builds) when assembling a procedure declaring more locals than the maximum representable during frame-pointer codegen; such procedures are now rejected with a diagnostic error ([#3332](https://github.com/0xMiden/miden-vm/pull/3332)).
 - [BREAKING] Bound dense `MastForest` and package digests to stored roots, external dependencies, and advice, rejected non-canonical dense forest payloads, and moved dense forest construction to `DenseMastForestBuilder` ([#3334](https://github.com/0xMiden/miden-vm/pull/3334)).
 - Made `make clippy` and `make lint` deny warnings so local linting fails on the same Clippy warnings as CI ([#3257](https://github.com/0xMiden/miden-vm/issues/3257)).
+- [BREAKING] Optimize constraint evaluation step by merging one-hot gated stack op constraints ([#333](https://github.com/0xMiden/miden-vm/issues/3333)).
```
huitseeker: Nit: the text says `#333`, but the link points to PR 3333. Should this be `#3333`?

PR #3315, opened 2026-06-30
3/3 test: Add trace proving input fuzz and proptest coverage

huitseeker:
Closes #3235.

This PR adds fuzz targets and proptest coverage for trusted trace proving input serialization.

It adds arbitrary strategies for trace state data and seed tests for trace proving inputs.

It also wires the needed testing features through the workspace.

The full stack now supports trusted trace proving input serialization with sparse MAST forest payloads and fuzz coverage.

PR #3334, opened 2026-07-06
Bind dense MastForest commitments and final order

### LINKED ISSUE #3067: Rationalize digests throughout programs

### LINKED ISSUE #3325: Make dense MastForest construction produce canonical node order

huitseeker:
This PR fixes [#3067](https://github.com/0xMiden/miden-vm/issues/3067), and #3325, and completely *punts* on `SparseMastForest` issues.

It extracts the dense `MastForest` commitment work from [#3311](https://github.com/0xMiden/miden-vm/pull/3311), plus the final dense order work from [#3329](https://github.com/0xMiden/miden-vm/pull/3329).

It does not include the trace proving input serialization stack in [#3313](https://github.com/0xMiden/miden-vm/pull/3313), [#3314](https://github.com/0xMiden/miden-vm/pull/3314), and [#3315](https://github.com/0xMiden/miden-vm/pull/3315). It also does not include the extra commitment input wire sections from [#3294](https://github.com/0xMiden/miden-vm/pull/3294).

  ## What changed

  `MastForest::commitment()` now commits to three parts.

  1. The interface commitment uses sorted procedure root digests.
  2. The dependency commitment uses sorted external node digests.
  3. The advice commitment uses all advice map entries in key order.

The final commitment is the hash of those three commitments.

A finalized dense `MastForest` now stores nodes in final dense order. External nodes come first and are sorted by digest. Basic blocks come next. Other internal nodes come last. Children always come before parents.

*Trusted* dense deserialization now rejects dense payloads that do not match this order. Trusted readers may trust serialized hashes, but they should not build an invalid finalized `MastForest`.

`Package::digest()` now follows the full `MastForest` commitment. Package identity changes when stored advice or external dependencies change.

`MastForest::commitment()` commits to the stored forest. This includes all stored external dependencies and all stored advice entries, even when they are not reachable from exported procedure roots.

  ## Out of scope

- This PR does not change `SparseMastForest`.

- This PR does not add new dense `MastForest` wire sections.

- This PR does not bump the dense `MastForest` wire version.

- This PR does not add trusted or untrusted `TraceProvingInputs` serialization.

- `StaticLibrary::new()` still defaults its source library identity to `mast.interface_commitment()`. A TODO now marks this for review after #3067, because package backed static libraries use `Package::digest()`.


core/src/mast/mod.rs
bobbinth, line 1: Not for this PR, but this file is quite large - I would consider splitting it into a few modules in the future.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
         Self::from_parts(MastForestParts { nodes, roots, advice_map })
     }
 
+    /// Builds a [`MastForest`] from raw parts and returns the node ID remapping applied during
+    /// canonicalization.
+    #[doc(hidden)]
+    pub fn from_raw_parts_with_id_map(
```
bobbinth, line 207: nit: may be good to explain in the doc comment why this is a hidden method. My understanding is that we want to be able to construct the forest from raw parts in assembly tests but we don't want to make `MastForestParts` a part of the public interface.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+                &parts.advice_map,
+            ),
             nodes: parts.nodes,
             roots: parts.roots,
             advice_map: parts.advice_map,
         })
     }
 }
 
+fn validate_mast_forest_parts_bounds(parts: &MastForestParts) -> Result<(), MastForestError> {
```
bobbinth, line 266: nit: I'd move all validation functions into something like a "validation" section (or maybe a separate module) so that they are all logically grouped together.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+            return Err(MastForestError::NodeIdOverflow(root_id, node_count));
+        }
+    }
+
+    Ok(())
+}
+
+fn canonicalize_dense_parts(
+    parts: MastForestParts,
+) -> Result<(MastForestParts, DenseIdMap<MastNodeId, MastNodeId>), MastForestError> {
```
bobbinth: Could we add doc comments explaining what this function is supposed to do?

Also nit: why do we not name this just `canonicalize_parts()`? All `MastForest`s are dense and the `SparceMastForest` is not a "true" MAST forest in my mind (i.e., it is not meaningful outside of the `TraceProvingInputs`).

Another nit: it may be a bit cleaner to make this a method on the `MastForestParts` struct - especially if we move the struct into a separate module.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+
+fn mast_node_order_class(node: &MastNode) -> MastNodeOrderClass {
+    if node.is_external() {
+        MastNodeOrderClass::External
+    } else if node.is_basic_block() {
+        MastNodeOrderClass::BasicBlock
+    } else {
+        MastNodeOrderClass::Internal
+    }
+}
```
bobbinth: nit: I'd make this a method on `MastNode` - i.e., `MastNode::class()`. It could be `pub(crate)` so as to not make it a part of the public API.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+            roots,
+            advice_map: parts.advice_map,
+        },
+        remapping,
+    ))
+}
+
+fn final_dense_node_order(
+    nodes: &IndexVec<MastNodeId, MastNode>,
+) -> Result<Vec<MastNodeId>, MastForestError> {
```
bobbinth, line 341: Would be good to add some inline comments in this function explaining what various code blocks do.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+
+    if emitted_internal_node_count != internal_node_count {
+        let node_id = internal_ids
+            .into_iter()
+            .find(|&node_id| pending_child_counts[node_id.to_usize()] > 0)
+            .expect("non-empty internal cycle must contain a pending node");
+        return Err(MastForestError::InvalidNodeOrder {
+            node_id,
+            reason: "internal nodes must form an acyclic child-before-parent graph".into(),
+        });
```
bobbinth, line 443: It is not very clear to me what this code does (some comments would help). My understanding, is that by the time we get to line `ordered_ids`, we already have all external and basic blocks there in correct order. So, the next step is to add internal nodes and make sure we always add children before parents - right?

If so, couldn't we just loop over `internal_ids` and remove IDs from this vector only if the children of this node are already in the `ordered_ids` vector? It may be a bit less efficient, but should be quite a bit less code.

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+    }
+
+    if ordered_ids
+        .iter()
+        .enumerate()
+        .all(|(index, &old_id)| old_id == MastNodeId::new_unchecked(index as u32))
+    {
+        validate_dense_node_order(&parts.nodes)?;
+        return Ok((parts, remapping));
+    }
```
bobbinth, line 304: Could we add a comment about what this code is doing? My understanding is that if the nodes are already in the correct order, we just return early.

But I also don't understand why we need to call `validate_dense_node_order()` on line 289 - didn't the code on lines 285 - 287 already did this?

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+        .roots
+        .into_iter()
+        .map(|root_id| {
+            remapping
+                .get(root_id)
+                .ok_or(MastForestError::NodeIdOverflow(root_id, node_count))
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+
+    validate_dense_node_order(&nodes)?;
```
bobbinth: Why is this needed? Is it to catch a potential bug in the code above? If so, maybe this should be done via `debug_assert!` or something similar?

core/src/mast/mod.rs
```diff
@@ -151,53 +196,274 @@ impl MastForest {
...
+        return Ok((parts, remapping));
+    }
+
+    let mut nodes = IndexVec::with_capacity(node_count);
+    let empty_forest = MastForest::new();
+    for old_id in ordered_ids {
+        let node = parts.nodes[old_id].clone();
+        let remapped_node =
+            node.to_builder(&empty_forest).remap_children(&remapping).build_linked()?;
+        nodes.push(remapped_node).map_err(|_| MastForestError::TooManyNodes)?;
```
bobbinth: How can we get `TooManyNodes` here? I guess this is possible if the input `parts` are malformed somehow, right?

crates/assembly/src/linker/mod.rs
```diff
@@ -172,9 +172,9 @@ impl Linker {
                 reason: err.to_string(),
             })?;
 
-        match self.libraries.entry(library.mast().commitment()) {
+        match self.libraries.entry(library.package.digest()) {
```
bobbinth: I would implement `LinkLibrary::digest()` method so that how to get a digest of the `LinkLibrary` is an internal concern of that struct.

crates/assembly/src/assembler.rs
```diff
@@ -1126,7 +1126,8 @@ impl Assembler {
                         )));
                     },
                 };
-                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info))
+                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info)
+                    .with_source_library_commitment(lib.package.digest()))
```
bobbinth: Related to the previous comment: adding `LinkLibrary::digest()` would simplify this code as well.

Also, I would consider creating `StaticLibrary::from_linked_library()` constructor so that the logic of how to create a static library from a linked library is encapsulated there.

core/src/mast/merger/mod.rs
```diff
@@ -232,6 +232,27 @@ impl MastForestMerger {
     ) -> Result<MastNodeBuilder, MastForestError> {
         super::build_node_with_remapped_ids(merging_id, src, original_forest, nmap)
     }
+
+    fn remap_finalized_node_ids(
+        mut node_id_mappings: Vec<DenseIdMap<MastNodeId, MastNodeId>>,
+        final_id_remapping: &DenseIdMap<MastNodeId, MastNodeId>,
+    ) -> Vec<DenseIdMap<MastNodeId, MastNodeId>> {
```
bobbinth, line 240: Could we add doc comment explaining what this method is intending to do?

core/src/mast/dense_builder.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/// Construction surface for dense MAST forests.
+///
+/// The builder may append nodes while a forest is under construction. The value returned by
+/// [`Self::finish`] is a finalized [`MastForest`] in final dense order: external nodes sorted by
+/// digest, then basic blocks in construction order, then internal nodes with children before
+/// parents and construction order as the tie-breaker.
+#[derive(Debug, Default)]
+pub struct DenseMastForestBuilder {
```
bobbinth, line 19: Related to one of the previous nits: why is this "dense" - could it not be just `MastForestBuilder`?

core/src/mast/dense_builder.rs
```diff
@@ -0,0 +1,92 @@
...
+    nodes: IndexVec<MastNodeId, MastNode>,
+    roots: Vec<MastNodeId>,
+    advice_map: AdviceMap,
+}
+
+impl DenseMastForestBuilder {
```
bobbinth, line 25: Question (and not sure if this is a good idea): would it make sense to move the logic for sorting nodes by class here? For example, we could have an API here that looks like:

```rust
impl DenseMastForestBuilder {
    /// Validates and arranges parts in correct order
    pub fn canonicalize_nodes() -> Result<Self, MastForestError> {
        ...
    }
}
```

And then, we could have something like:

```rust
let mut builder = DenseMastForestBuilder::from_raw_parts(nodes, roots, advice_map);
builder.canonicalize_nodes();
let forest = builder.finish();
```

core/src/mast/dense_builder.rs
```diff
@@ -0,0 +1,92 @@
...
+    ) -> Result<(), MastForestError> {
+        self.advice_map
+            .merge(advice_map)
+            .map_err(|((key, _prev), _new)| MastForestError::AdviceMapKeyCollisionOnMerge(key))
+    }
+
+    /// Finalizes this builder into a dense [`MastForest`].
+    pub fn finish(self) -> Result<MastForest, MastForestError> {
+        self.finish_with_id_map().map(|(forest, _remapping)| forest)
+    }
```
bobbinth, line 70: nit: we usually call such methods `build()`.

crates/assembly/src/mast_forest_builder/finalizer.rs
```diff
@@ -182,33 +182,48 @@ impl MastForestFinalizer {
...
+            MastForest::from_raw_parts_with_id_map(nodes, roots, advice_map)
+                .map_err(|source| Report::new(MastForestBuilderError::FinalizeForest { source }))?;
+
+        for node_id in node_id_by_ref.values_mut() {
+            *node_id = final_id_remapping.get(*node_id).ok_or_else(|| {
+                Report::new(MastForestBuilderError::FinalizeForest {
+                    source: MastForestError::NodeIdOverflow(*node_id, final_id_remapping.len()),
+                })
+            })?;
+        }
```
bobbinth, line 207: Could we add some comments here explaining what this code does?

crates/utils-core-derive/src/lib.rs
```diff
@@ -260,31 +260,25 @@ fn generate_mast_forest_contributor_impl(
     enum_name: &Ident,
     generics: &syn::Generics,
     variant_names: &[&Ident],
-    variant_fields: &[Ident],
+    variant_types: &[&Type],
```
bobbinth, line 263: Was this a bug previously?

But also, I didn't really review code in this file - maybe someone else should take a look at it?

crates/mast-package/src/package/serialization.rs
bobbinth, line 820: nit (and not for this PR): I'd break out tests into a separate file.

crates/mast-package/src/package/mod.rs
```diff
@@ -166,24 +166,20 @@ impl Package {
...
+                        path: export.path.clone(),
+                        digest: export.digest,
+                    }
+                })?;
             }
         }
 
-        let digest = self.mast.compute_nodes_commitment(node_ids.iter());
-        self.digest = digest;
+        self.digest = self.mast.commitment();
```
bobbinth, line 190: This changes the commitment methodology of the package, right? Previously, we'd commit only to the public interface of the package, but now we are using the underling MAST forest commitment. This may create an issue where two packages with the different set of public procedures (but the same underlying MAST forest) have the same commitment.

How to compute package commitment is still an open question in my mind. I'm thinking it could be:
```
hash(package_interface_commitment || mast_forest_commitment)
```
But we may want to leave this for another PR.

So, my preference would be to keep the old methodology here - i.e., hashing the digests of publicly exported procedures.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -110,11 +110,27 @@ pub struct MastForestBuilder {
...
     pub(crate) fn new(mast: &'a MastForest, debug_info: Option<PackageDebugInfo>) -> Self {
-        Self { mast, debug_info }
+        Self {
+            mast,
+            debug_info,
+            // Direct forest-backed static libraries do not have a package digest, so their source
+            // identity is the full forest commitment. This keeps provenance hints scoped to the
+            // same roots, external dependencies, and advice as package-backed static libraries.
+            source_library_commitment: mast.commitment(),
+        }
```
bobbinth, line 127: As mentioned in another comment, I would probably create dedicated constructors here (e.g., `StaticLibrary::from_mast_forest()`, `StaticLibrary::from_link_library()`) so that the logic of how to construct static library from different objects is encapsulated in this file.

core/src/mast/serialization/mod.rs
```diff
@@ -222,7 +222,12 @@ impl MastForest {
     /// Internal serialization with options.
     ///
     /// Current writers encode normal execution payloads or hashless validation payloads.
+    /// Both forms use the finalized dense node order already stored in the `MastForest`; writers
+    /// validate that order but do not sort nodes while writing.
     fn write_into_with_options<W: ByteWriter>(&self, target: &mut W, hashless: bool) {
+        self.validate_dense_node_order()
+            .expect("dense MAST forest must be in final dense order before serialization");
```
bobbinth, line 229: Why do we need this check? Is this primarily to guard is from potential bugs in the `MastForest` itself?

core/src/mast/serialization/resolved.rs
```diff
@@ -202,6 +204,47 @@ impl<'a> ResolvedSerializedForest<'a> {
         self.node_digest_for_entry(index, entry)
     }
 
+    pub(super) fn validate_dense_node_order(&self) -> Result<(), DeserializationError> {
```
bobbinth, line 207: This function is very similar to the one in the `MastForest` file, right? Is there a way to remove this code duplication? That is, could we maybe delay this validation until the `MastForest` struct is constructed after deserializaiton?

bobbinth (commented):
Looks good! Thank you! I reviewed mostly non-testing code and left some comments inline. Most comments are pretty small - the overall approach is solid.

core/src/mast/mod.rs
huitseeker: Agreed. it should be a follow-up so the review does not mix behavior changes with module layout

core/src/mast/mod.rs
huitseeker: Fixed. The doc comment now says this method is hidden because raw dense parts and builder-local ID remapping are only for internal builders and tests.

core/src/mast/mod.rs
huitseeker: Fixed. The helpers now sit under a Dense forest validation and canonicalization section.

core/src/mast/mod.rs
huitseeker: Fixed. The helper is now `MastNode::order_class()`, scoped to `crate::mast`

core/src/mast/mod.rs
huitseeker: Fixed. The function is now named `canonicalize_parts`, and it has a doc comment that states the order it builds and the ID map it returns.

I did not move it onto `MastForestParts`. I agree that can make sense *if* `MastForestParts` moves into its own module. For this PR, a private helper keeps the change smaller.

core/src/mast/mod.rs
huitseeker: The code now explains the external-node duplicate check and the ready-set logic for internal nodes.

core/src/mast/mod.rs
huitseeker: A repeated scan is simpler, but it can delay a parent that becomes ready early in the same pass. That can let a later unrelated internal node win the construction-order tie-breaker.

The current code keeps a ready set keyed by original node ID. This preserves child-before-parent order and construction-order tie breaks at the same time.

core/src/mast/mod.rs
huitseeker: Fixed. The extra validation on the already-ordered branch is now a `debug_assert!`. The real check is the order computed by `final_dense_node_order`.

core/src/mast/mod.rs
huitseeker: Fixed. This is now a `debug_assert!`, because it only checks the remapping code above it.

core/src/mast/mod.rs
huitseeker: Fixed. The code validates the node count before remapping, so pushing exactly that many remapped nodes cannot exceed the limit. The push now uses `expect` with that reason.

crates/assembly/src/linker/mod.rs
huitseeker: Partly fixed, but I would not call the method `digest()` now. There are two identities:

1. The linker surface identity is `Package::interface_digest()`.
2. The static input identity is the full forest commitment.

The full forest identity is hidden behind `LinkLibrary::commitment()`. If we want the linker surface identity hidden too, I would add `LinkLibrary::interface_digest()` rather than a generic `digest()` method.

crates/assembly/src/assembler.rs
huitseeker: Agree. This is still worth doing. The latest code has more identity logic in `static_libraries_for_builder` than I would like.

This should become `StaticLibrary::from_link_library()` in a follow-up cleanup, or I can do one more small commit.

core/src/mast/dense_builder.rs
huitseeker: I would keep `DenseMastForestBuilder` for now. The repo already has an assembly-level `MastForestBuilder`, and `core::mast` also has `SparseMastForestBuilder`. The word `Dense` makes this builder's scope exact.

If we later split dense MAST code into its own module and retire the competing names, then `MastForestBuilder` may be the better public name.

core/src/mast/merger/mod.rs
huitseeker: Fixed. The method now says that it remaps each source forest's node IDs from merger-local IDs to finalized dense IDs.

core/src/mast/dense_builder.rs
huitseeker: I would keep canonicalization at the `MastForest` finalization boundary. The rule should apply to every path that creates a finalized dense forest, not only to `DenseMastForestBuilder`.

That is why `DenseMastForestBuilder::finish_with_id_map()` calls `MastForest::from_parts_with_id_map()`, and that constructor performs canonicalization.

crates/assembly/src/mast_forest_builder/finalizer.rs
huitseeker: Fixed. The finalizer now explains that source and debug references are recorded against builder-local node IDs and must be rewritten to finalized dense IDs.

crates/utils-core-derive/src/lib.rs
huitseeker: This was not a known behavior bug in released code. It became necessary because this PR removed `add_to_forest` from the `MastForestContributor` trait and made enum dispatch use builder conversion instead.

The macro now reads each variant type and generates `From<VariantType>` impls for the enum.

crates/mast-package/src/package/mod.rs
huitseeker: You are right that this changes what `Package::digest()` means. In this branch it is the full stored `MastForest` commitment. It is not the exported interface commitment anymore. 

The old exported-root identity still exists as `Package::interface_digest()`. The linker uses that for public module surface de-duplication and provenance.

The concern about two packages with different exports and the same forest is valid for `Package::digest()`. The package content digest still includes the manifest, so those packages do not have the same content identity.

This PR needs the full forest identity because #3067 needs stored advice + external dependencies as part of the package identity. If we want a true package commitment of `hash(interface_commitment || mast_forest_commitment)`, that should be a follow-up because it is a third identity, not the old one and not the forest one.

crates/assembly/src/mast_forest_builder.rs
huitseeker: The current code has `StaticLibrary::new()` plus setters for source commitments. 

With the new identity split, dedicated constructors would be clearer:

1. `StaticLibrary::from_mast_forest()` for direct forest inputs.
2. `StaticLibrary::from_link_library()` for package-backed static inputs.

This can be one small cleanup commit if we want to address it in this PR.

core/src/mast/serialization/mod.rs
huitseeker: Yes. This guards against internal bugs and test-only mutation paths that could leave a `MastForest` in invalid order. Serialization should not sort as a fallback, because finalization is where canonical order is established, and it would fail to detect and raise that this invariant broke. The writer only checks the invariant before it writes bytes.

core/src/mast/serialization/resolved.rs
huitseeker: The checks are similar, but they run on different data.

`MastForest` validation checks materialized nodes. `ResolvedSerializedForest` checks wire entries before full materialization. `MastForestWireView::new()` needs that because it exposes random access over trusted cache bytes without building a `MastForest`.

I would not delay this validation until materialization, because `MastForestWireView` may never materialize. I agree that we can reduce *code* duplication later by sharing a small order-state helper between the two checks.

crates/assembly/src/mast_forest_builder.rs
huitseeker: Done

crates/assembly/src/assembler.rs
huitseeker: Fixed. `static_libraries_for_builder()` now delegates to `StaticLibrary::from_link_library()`, so the assembler no longer knows which package identities must feed builder provenance.

core/src/mast/node/mod.rs
```diff
@@ -86,6 +86,17 @@ pub enum MastNode {
 // ------------------------------------------------------------------------------------------------
 /// Public accessors
 impl MastNode {
+    /// Returns the ordering class used by finalized dense MAST forests.
+    pub(in crate::mast) fn order_class(&self) -> super::MastNodeOrderClass {
```
bobbinth, line 90: nit: I'd maybe move the definition of `MastNodeOrderClass` to this file.

core/src/mast/mod.rs
bobbinth: I still find this code a bit too complicated, and I'm not sure if we really care about construction order here. I didn't review it in depth - but it is fine to keep it as is for now.

crates/assembly/src/mast_forest_builder.rs
bobbinth: I'm not sure this change was committed (or at least I'm not seeing it in the current code) - but we can also address this in a follow up.

crates/assembly/src/assembler.rs
```diff
@@ -1126,7 +1125,11 @@ impl Assembler {
                         )));
                     },
                 };
-                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info))
+                Ok(StaticLibrary::new(lib.mast().as_ref(), debug_info)
+                    .with_source_library_commitment(lib.commitment())
+                    .with_alternate_source_library_commitment(
+                        lib.package.interface_digest().into_diagnostic()?,
+                    ))
```
bobbinth, line 1132: Related to the previous comment, the code here seems to have gotten more complicated rather than simpler - but we can fix it in a follow-up PR.

crates/assembly/src/linker/mod.rs
```diff
@@ -126,6 +126,11 @@ pub enum LinkStatus {
 pub struct Linker {
     /// The set of libraries to link against.
     libraries: BTreeMap<Word, LinkLibrary>,
+    /// The statically linked libraries to pass to MAST forest construction.
+    ///
+    /// This index is keyed by full MAST forest commitment, not package digest, so static libraries
+    /// with the same exported procedure roots but different stored advice are retained.
+    static_libraries: BTreeMap<Word, LinkLibrary>,
```
bobbinth, line 133: I didn't review the impact of this change. Not sure if it was really needed for this PR, but in the interest of time, let's keep it in.

cc @bitwalker in case he as some opinions here.

crates/mast-package/src/package/mod.rs
bobbinth: I would have preferred to roll back this change and do another PR to fully close #3067. But in the interest of time, let's keep it in and fix this in a follow-up.

bobbinth (approved):
Looks good! Thank you! I left some more comments inline and some of the prior ones still need to be addressed, but in the interest of time I'll merge this PR and we can fix things in follow-ups.

Let's make sure to capture the follow-ups in https://github.com/0xMiden/miden-vm/issues/3067. There are quite a few things we can improve in regards to code and interface organization, but more importantly, we need to come up with a better package commitment methodology.

Issue #3041, opened 2026-04-22
Align endianness of FRI proof queries and fri_ext2fold4


huitseeker:
The FRI proof is queried in domain order, but fri_ext2fold4 expects an index whose low two bits select the bit-reversed position inside the four-element folded row. 

>
> Let's make an issue for the Fri fold instruction 

 _Originally posted by @adr1anh in [#3039](https://github.com/0xMiden/miden-vm/pull/3039/changes#r3122207573)_

giwaov:
I can take this one if it is still available. Could you assign it to me before I start?

Al-Kindi-0:
Thanks @giwaov , this one needs more thinking and interacts with the recursive verifier in non-trivial ways. I will take it as part of some upcoming work.

PR #3335, opened 2026-07-06
perf(fri): replace mask+shift index math with u32divmod in `verify_query_layer`

Nashtare:
## Describe your changes

`verify_query_layer` computed its FRI fold index with a bitmask (u32and) plus a variable-shift and hand-rolled bit-reversal, and needed a 12-slot locals frame just to keep t_depth alive across the Merkle-verification step. This rewrites that math using u32divmod, which gets the remainder and quotient in one shot.

When running against the Fibonacci examples (small & large):

```bash

  ┌───────────────┬────────┬─────────┬────────┬────────────────┐
  │               │  core  │ bitwise │ memory │ chiplets total │
  ├───────────────┼────────┼─────────┼────────┼────────────────┤
  │ small: before │ 35,873 │ 1,552   │ 13,289 │ 72,535         │
  ├───────────────┼────────┼─────────┼────────┼────────────────┤
  │ small: after  │ 32,957 │ 256     │ 12,965 │ 70,675         │
  ├───────────────┼────────┼─────────┼────────┼────────────────┤
  │ large: before │ 36,890 │ 1,552   │ 14,617 │ 75,879         │
  ├───────────────┼────────┼─────────┼────────┼────────────────┤
  │ large: after  │ 33,974 │ 256     │ 14,293 │ 74,019         │
  └───────────────┴────────┴─────────┴────────┴────────────────┘
  
  ```
  
  
  
 (about 54 fewer cycles and 24 fewer bitwise-chiplet rows per call)

Note: the previous cycle count was erroneous (not 99 but 213).

- Updated `CHANGELOG.md'


adr1anh:
Should we bundle https://github.com/0xMiden/miden-vm/issues/3041 into this PR? IIRC this avoids some of the bit-reversing, and might make these changes redundant.

Nashtare:
Ah I didn't know about this issue. I'm not sure it's worth continuing as @Al-Kindi-0 also has some further ideas reducing cycle count that touches this region? I wouldn't want both work to collide too much so perhaps best to just wait for Al's work to land

Al-Kindi-0:
Yes, `fri_ext2fold4` should be co-designed with the surrounding code in that hot-loop. As mentioned a couple of times last week, a few changes are coming soon to the MASM verifier which will include addressing #3041 , among other things.

PR #3327, opened 2026-07-03
Reorganize precompile ownership and default registry wiring

adr1anh:
## Summary

This reorganizes the precompile split so `miden-precompiles` owns the semantic Rust precompile implementations and official registry, while `miden-core-lib` owns the MASM wrappers, wrapper event handlers, and MASM codegen/build support.

The processor now installs the official in-repo precompile registry by default, so VM/prover/CLI/test-utils entry points no longer need to manually attach it.

## Changes

- Move precompile MASM wrappers into `miden-core-lib`.
  - `precompiles/asm/**` moved to `crates/lib/core/precompiles/asm/**`.
  - The aggregate core-lib package still bundles the `miden::precompiles` namespace for internal wrapper support.
- Move precompile wrapper event handlers into `miden_core_lib::handlers::precompiles`.
- Move wrapper integration tests from `precompiles/tests/integration/**` to `crates/lib/core/tests/precompiles/**`.
- Rename and move MASM codegen support from `miden-precompiles-codegen` to `miden-core-lib-codegen`.
- Simplify `miden-precompiles` by removing standalone `PrecompilesLibrary`, standalone MASP package build/test plumbing, and its dependency on `miden-processor`.
- Move semantic descriptor/constants ownership out of codegen and into `miden-precompiles`.
- Install the official precompile registry by default in `FastProcessor::new` and `FastProcessor::new_with_options`.
- Remove `FastProcessor::with_precompile_registry` and higher-layer manual `.with_precompile_registry(miden_precompiles::registry())` call sites.

## Motivation

This makes the layering clearer:

- `miden-precompiles`: semantic deferred precompile implementations and official registry.
- `miden-core-lib`: MASM wrappers and host event handlers needed by bundled core-lib wrappers.
- `miden-processor`: always has the in-repo precompile registry available by default.
- Higher-level crates: no longer need to remember to install precompiles manually.

This also removes temporary standalone `PrecompilesLibrary` and MASP-package scaffolding from this branch.

## Compatibility

This is intentionally breaking for temporary branch APIs:

- `PrecompilesLibrary` is removed.
- `FastProcessor::with_precompile_registry` is removed.
- Manual registry installation at higher layers is no longer supported or needed.
- `miden::precompiles` remains present in the aggregate package but should be treated as internal for now.

## Validation

Previously ran:

- `make format`
[...]

PR #3336, opened 2026-07-06
Limit precompile math and allow custom verifier registries

huitseeker:
This keeps the precompile math surface focused on secp256k1. It removes the Ed25519 and secp256r1 field and curve modules from Rust, generated MASM, codegen, exports, and tests.

Verification can now use a caller supplied `PrecompileRegistry`. The default verifier still use `miden_precompiles::registry()`. Direct processor users can verify proofs made with a custom registry. They can also set the deferred state budget.

The diff also includes the current `origin/next` verifier and AIR changes needed by `precompiles/crate`.


verifier/src/lib.rs
adr1anh, line 1: With https://github.com/0xMiden/miden-vm/pull/3327, the precompiles crate no longer depends on the processor. This allows the registry to be automatically installed without requiring users to manually specify it.

Moreover, in #3293, [this commit](https://github.com/0xMiden/miden-vm/commit/8abf4c3d0a3cd841c2f25563cacfadf2e5304992) introduces an improved verifier API which avoids introducing functions for all configurations. 

I’m okay with leaving these in here if they're intentionally transitional.

huitseeker: This is fixing something straightforwardly wrong with #3220, that I [reported last week and has not been fixed](https://github.com/0xMiden/miden-vm/pull/3222#discussion_r3486815150): execution can already use a custom precompile registry through direct `FastProcessor` use. Before this commit b3d401f4bc2d83af0a89bd5c17316d38111cc3b3, verification always rebuilt deferred state with the default built-in registry. That meant a proof produced with a custom registry could fail at verification, even when the STARK proof and deferred wire were valid for the registry used during execution.

The fix makes the verifier accept the caller’s registry explicitly. The default verify path stays unchanged and still uses the built-in registry, so normal callers keep the same behavior. New helper functions cover the two real knobs: the precompile registry and the deferred-state verifier budget. The combined helper is the deepest entry point, while the simpler helpers delegate to it.

The helper functions are of course optional, the data structure they drive is the real contribution here. Clearly, #3220, #3227, and #3293, if merged, will be merged separatedly, so they need to stand up on their own. On the management of a custom registry, #3220 doesn't.

huitseeker:
@adr1anh Merged with removal of prover-side custom registries

PR #3340, opened 2026-07-07
dep: bump miden-crypto to v0.28

huitseeker:
Bumps the crypto dependency to 0.28.


bobbinth (approved):
Looks good! Thank you!

I thought we'd also need to update the ECDSA precompile - but seems like all tests are passing. So, maybe that code needs to be updated separately.

adr1anh:
#3342 applies the required changes to support the new public key commitment format.

PR #3341, opened 2026-07-07
Merge next + review fixes for precompiles/crate

huitseeker:
On top of #3222.
This include a fresher merge of next + the fix of a couple of unaddressed review threads from #3222. 

This does not yet include the upcoming #3440 or any downstream changes for ECDSA (but once this is in, that should be the only missing pieces).


adr1anh (approved):
LGTM, Thank you very much!

PR #3343, opened 2026-07-08
perf(precompiles): port optimizations from `precompile-experiments`

Nashtare:
- Updated `CHANGELOG.md'


precompiles-prover/src/stark_config.rs
```diff
@@ -42,6 +42,34 @@ const COMPRESSION_INPUTS: usize = 2;
...
+/// `miden_air::config::pcs_params`, which is fixed at `log_blowup = 3` for its own
+/// degree-8 constraints.
+///
+/// The remaining FRI parameters (folding arity, final degree, PoW bits, query count)
+/// mirror `miden_air::config::pcs_params` as-is; they have not been independently
+/// re-derived for this lower blowup and should get a dedicated security review before
+/// this config is relied on outside benchmarking.
+pub fn precompile_pcs_params() -> PcsParams {
+    PcsParams::new(
+        3,  // log_blowup (must be >= log_quotient_degree)
```
huitseeker, line 62: The doc comment here says `log_blowup = 2` (blowup = 4) and that every chiplet closes at `log_quotient_degree <= 2`, but the literal passed just below is `3`, which is the same value as `miden_air::config::pcs_params`. The referenced `log_quotient_degrees_fit_the_blowup` test also asserts `lqd <= 3`, not `<= 2`. 

Was the lower blowup meant to land here, or is the comment ahead of the code?

Nashtare: The comment (and `log_quotient_degrees_fit_the_blowup` test) are slightly misleading. I aimed at simply decoupling the FRI params from the main VM, keeping the same default ones, but the precompile VM does support a smaller blowup factor unlike the main VM.

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -217,14 +218,15 @@ pub const CELL_SLOPE_AUX: usize = 0;
 pub const CELL_LAMBDA: usize = 1;
 pub const CELL_INV: usize = 2;
 pub const CELL_T: usize = 3;
-/// Row-1 cells.
-pub const CELL_W: usize = 0;
+/// Row-1 (tail) cells: the fused result `y₃`, the `x₁ − x₃` witness `e`,
+/// and `x₃` (cell 2 is dead — the mul-subtracts leave no `w` / `u`
+/// intermediate).
```
adr1anh: Could we describe the current layout directly here—i.e. which slots are occupied or reserved and what each row exposes? I think that would be easier to follow than explaining the slots through the removed `w` / `u` intermediates and where `y₃` used to live.

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -717,9 +723,16 @@ where
...
+        // Col 2 (fractions): the shared tail, two fused mul-subtracts (no
+        // `w` / `u` store, no `x₃` / `y₃` sub op). All fire on the slope-row
+        // window — `e` / `x₃` / `y₃` live on the tail row, read via next. The
+        // x₃ subtract splits by case: generic forms `t = x₁ + x₂` then
+        // `x₃ = λ² − t` (κ_c = 1); double, where `x₁ = x₂`, folds `t = 2x₁`
+        // into `x₃ = λ² − 2·x₁` (κ_c = 2, c = x₁) and lays no `t` add.
+        //
+        // generic's `t = x₁ + x₂`, or double's fused `x₃ = λ² − 2·x₁` —
+        // mutually exclusive, exactly one fires per live op — plus double's
+        // y-equality.
```
adr1anh: Could we simplify this explanation a little? The generic/double split appears to be described once before and once after the blank line. A short description of the current relations would be easier to follow.

precompiles-prover/src/uint/mul/mod.rs
```diff
@@ -137,6 +139,10 @@ pub struct UintMulMsg<E> {
     pub c_ptr: E,
     pub r_ptr: E,
     pub bound_ptr: E,
+    /// `0` for the additive shape `κₐ·a·b + κ_c·c ≡ r`, `1` for the
+    /// subtractive `κₐ·a·b − κ_c·c ≡ r` — a distinct relation the consumer
+    /// demands exactly, so an add provide can never satisfy a sub consume.
```
adr1anh, line 146: Since this changes the `UintMul` tuple from seven to eight fields, could we update the linked UintMul/AIR documentation and relation registry in the same pass? A few places still describe the additive-only relation or the previous EC tail flow.

precompiles-prover/src/uint/mul/mod.rs
```diff
@@ -178,7 +185,13 @@ pub const COL_KAPPA_A: usize = 14;
...
 /// blocks, 0 on padding; gates every bus flag + the γ offset constant.
 pub const COL_ACT: usize = 15;
-pub const NUM_MAIN_COLS: usize = 16;
+/// Subtractive-underflow borrow `∈ {0, 1, 2}` (cycle-constant): the moduli
+/// the canonical reduction adds back when `is_sub` and the product
+/// `< κ_c·c` (up to 2 for the `κ_c = 2` doubling `λ² − 2x₁`). Read on the
+/// p-rows (where `bound(β)` is live) to contribute `borrow·(bound(β)+1)`
+/// to the SZ identity. Pinned to 0 unless `is_sub`.
+pub const COL_BORROW: usize = 16;
+pub const NUM_MAIN_COLS: usize = 17;
```
adr1anh: Note: this crosses the rate-aligned threshold and leads to an extra permutation per row.

precompiles-prover/src/uint/mul/mod.rs
```diff
@@ -401,8 +418,13 @@ impl LiftedAir<Felt, QuadFelt> for UintMulAir {
         // one row. κ_c is read from the term row (next).
         let val_sum: AB::ExprEF = (0..8)
             .fold(AB::ExprEF::ZERO, |acc, m| acc + bp[2 * m].clone() * AB::Expr::from(local[m]));
-        let linear =
-            val_sum.clone() * (sel[ROW_C].clone() * kappa_c_next) - val_sum * sel[ROW_R].clone();
+        // `is_sub` (term row, read via next on the c row) flips the c term:
+        // (1 − 2·is_sub) is +1 for `κₐ·a·b + κ_c·c`, −1 for the subtractive
+        // `κₐ·a·b − κ_c·c`. The result r stays on the −R side either way.
+        let is_sub_next: AB::Expr = next[TERM_CELL_IS_SUB].into();
+        let c_sign: AB::Expr = AB::Expr::ONE - AB::Expr::from(Felt::from(2u32)) * is_sub_next;
```
adr1anh:
```suggestion
        let c_sign: AB::Expr = AB::Expr::ONE - is_sub_next.double();
```

precompiles-prover/src/uint/mul/trace.rs
```diff
@@ -373,7 +462,9 @@ pub(crate) fn build_aux(
             _ if row_kind == ROW_R => -val_sum,
             _ if row_kind == ROW_C => {
                 let kappa_c = main.values[(r + 1) * NUM_MAIN_COLS + TERM_CELL_KAPPA_C];
-                QuadFelt::from(kappa_c) * val_sum
+                let is_sub = main.values[(r + 1) * NUM_MAIN_COLS + TERM_CELL_IS_SUB];
+                let signed = (Felt::ONE - Felt::from(2u32) * is_sub) * kappa_c;
```
adr1anh: Could use `.double()` here as well: `let signed = (Felt::ONE - is_sub.double()) * kappa_c;`

precompiles-prover/src/math.rs
```diff
@@ -91,6 +91,53 @@ pub fn mac_div_rem(
     (q.to(), r.to())
 }
 
+/// `(κₐ·a·b − κ_c·c) mod p` — the subtractive scaled-MAC reduction.
+pub fn mac_sub_reduce(kappa_a: u16, a: U256, b: U256, kappa_c: u16, c: U256, bound: U256) -> U256 {
+    mac_sub_div_rem(kappa_a, a, b, kappa_c, c, bound).1
+}
+
+/// `(q_committed, r, borrow)` of `κₐ·a·b − κ_c·c ≡ r (mod p)`, written
```
adr1anh: `κₐ` and `κ_c` use different subscript styles because Unicode has no subscript c. Could we write both as `κ_a` / `κ_c` for consistency?

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -668,6 +680,18 @@ where
...
+                                // would force `s = 0`, which never holds at a
+                                // 2-torsion point of a **smooth** curve
+                                // (`3x² + a ≠ 0` at a simple cubic root). A
+                                // `y = 0` self-add thus cannot satisfy the
+                                // double case — it can only take `cancel`
+                                // (2·(2-torsion) = ∞). This leans on curve
+                                // smoothness (`4a³ + 27b² ≠ 0`), the same
+                                // anchored-curve well-formedness assumption as
+                                // the `b ≠ 0` guard the disequality MACs rest
+                                // on.
```
adr1anh: Maybe something shorter: at `y = 0`, the slope pin forces `3x² + a = 0`. Together with `x³ + ax + b = 0`, this makes `x` a common root of the curve polynomial and its derivative, contradicting smoothness. Double is therefore impossible here; the point must take `cancel`.

precompiles-prover/src/uint/mul/mod.rs
```diff
@@ -371,7 +381,10 @@ impl LiftedAir<Felt, QuadFelt> for UintMulAir {
 
         let kappa_a: AB::Expr = local[COL_KAPPA_A].into();
         let act: AB::Expr = local[COL_ACT].into();
-        let kappa_c_next: AB::Expr = next[TERM_CELL_KAPPA_C].into();
+        // The pinned signed scale (see `TERM_CELL_KAPPA_C_SIGNED`) — reads
+        // straight from the term row, no in-place `κ_c · (1 − 2·is_sub)`
+        // product on the `c` row's `next` window.
```
adr1anh: Could this just describe the value being read? For example: `// Signed κ_c scale, pinned on the term row and read here via next.`

precompiles-prover/src/uint/mul/mod.rs
```diff
@@ -467,6 +479,20 @@ impl LiftedAir<Felt, QuadFelt> for UintMulAir {
         let is_sub: AB::Expr = local[TERM_CELL_IS_SUB].into();
         builder
             .assert_zero(sel[ROW_TERM].clone() * is_sub.clone() * (AB::Expr::ONE - is_sub.clone()));
+
+        // κ_c_signed = κ_c · (1 − 2·is_sub), pinned term-row-local (no `next`
+        // window needed here — both operands live on this row), so `linear`
+        // above can read it via a single `next`-row witness instead of
+        // multiplying two `next`-row witnesses in place. Degree 2 (`κ_c` ×
+        // `is_sub`), well under the lqd-1 budget.
```
adr1anh: Could we shorten this to the invariant and its purpose? For example: `// Pin κ_c_signed = κ_c·(1 − 2·is_sub) on the term row, keeping the id transition at lqd 1.`

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -211,20 +229,21 @@ pub const ROW_RES: usize = 2;
...
+/// intermediate).
+pub const CELL_Y3: usize = 0;
 pub const CELL_E: usize = 1;
-pub const CELL_U: usize = 2;
 pub const CELL_X3: usize = 3;
-/// Row-2 cells: the result's `y₃`, plus the hosted `r` / scalar-bound /
-/// group ptrs.
-pub const CELL_Y3: usize = 0;
+/// Row-2 (res) cells: the hosted `r` / scalar-bound / group ptrs. Cell 0 is
+/// free — `y₃` moved to the tail row so the `λ·e − y₁` mul-subtract reads
```
adr1anh: Could we describe this in terms of the current role rather than where `y₃` used to live? For example, say that cell 0 is reserved and that the other cells host `r`, `sbound`, and `group`.

precompiles-prover/src/ec/add/trace.rs
```diff
@@ -156,7 +156,7 @@ fn op_block(op: &EcAddOp, mult: ProvideMult, ec: &EcStoreRequires) -> Vec<Felt>
     for (cell, ptr) in transients[NUM_CELLS..2 * NUM_CELLS].iter().enumerate() {
         set(ROW_TAIL, cell, *ptr);
     }
-    set(ROW_RES, 0, transients[8]); // CELL_Y3
+    set(ROW_RES, 0, transients[8]); // res cell 0 (free — y₃ rides the tail row)
```
adr1anh: Could this comment state only what is stored here? For example: `// reserved zero cell`.

precompiles-prover/src/ec/require.rs
```diff
@@ -212,52 +212,50 @@ impl<'a> EcRequire<'a> {
...
                     // double: s ≡ 3x² + a and 2λy ≡ s (the κ's carry the
-                    // tangent constants; shared r_ptr = s), λ = s·(2y)⁻¹,
-                    // the equality certificates x₁ = x₂ / y₁ = y₂, and
-                    // the nonzero witness inv = b·y₁⁻¹ certified by
-                    // inv·y₁ ≡ b — the 2y denominator's invertibility.
+                    // tangent constants; shared r_ptr = s), λ = s·(2y)⁻¹. The
+                    // `x₁ = x₂` / `y₁ = y₂` equalities are enforced natively
+                    // in the AIR (the operands are the same stored point, so
+                    // their coords share ptrs under value-interning). No
+                    // `y₁ ≠ 0` witness: the slope pin `2λy ≡ s = 3x² + a`
```
adr1anh: Could we use the same shorter smoothness explanation here? At `y = 0`, the slope pin and curve equation make `x` a common root of the curve polynomial and its derivative, contradicting smoothness.

precompiles-prover/src/ec/require.rs
```diff
@@ -410,36 +408,46 @@ impl<'a> EcRequire<'a> {
         (group, r, pai)
     }
 
-    /// The live cases' shared tail: `w = λ²`, `t = x₁ + x₂`,
-    /// `x₃ = w − t`, `e = x₁ − x₃`, `u = λ·e`, `y₃ = u − y₁`, with `R`
-    /// minted as a **closure-cert** point — its membership rides this
-    /// block's `EcOnCurveCert` (the group law is closed → on-curve
-    /// operands give an on-curve result), so it pays *no* MAC trio.
+    /// The live cases' shared tail: `x₃ = λ² − t`, `e = x₁ − x₃`,
+    /// `y₃ = λ·e − y₁` — the two mul-subtracts fused (no `w` / `u`
```
adr1anh: Could we describe the current equations here without mentioning the removed intermediates? The performance comparison is useful in the commit message, but the API documentation would be easier to follow if it stated only the current behavior.

precompiles-prover/src/uint/add/mod.rs
```diff
@@ -306,19 +330,51 @@ impl LiftedAir<Felt, QuadFelt> for UintAddAir {
...
+        // gates and would needlessly bloat this constraint's degree once
+        // multiplied by `p_sel`).
+        let mut p_own: AB::ExprEF = -(full_sum + bp[0].clone()) * k.clone();
+        for (j, &(row, cell)) in GAMMA_NEG_SLOTS.iter().enumerate() {
+            if row == ROW_P {
+                let w: AB::ExprEF = bp[j + 1].clone() - bp[j].clone() * t32.clone();
+                p_own -= w * AB::Expr::from(local[cell]);
+            }
+        }
+        builder.assert_zero_ext((id + p_own) * p_sel.clone());
```
adr1anh: Nit, we can sum the contributions to `p_own`, and subtract it from `id` in the constraint

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -84,7 +86,7 @@
 //!
 //! | row | cells 0–3 | emits |
 //! |---|---|---|
-//! | 0 `slope` | `(slope_aux, λ, inv, t)` | slope + predicate certs (local), tail certs (cells @ next) |
+//! | 0 `slope` | `(slope_aux, λ, —, t)` | slope + predicate certs (local), tail certs (cells @ next) |
 //! | 1 `tail`  | `(y₃, e, —, x₃)` | the two fused mul-subtracts + the live result consume (`r`/`group` @ next) |
 //! | 2 `res`   | `(—, r, sbound, group)` | the provide + operand/PAI/group consumes (`p`/`q`/mult @ next) |
 //! | 3 `term`  | `(mult, p, q, —)` | — (hosts only; the constancy gate drops here) |
```
adr1anh: We should be able to save a cell here now right?

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -224,11 +226,11 @@ pub const ROW_RES: usize = 2;
 pub const ROW_TERM: usize = 3;
 
 /// Row-0 cells: `slope_aux` is `d = x₂ − x₁` for `generic`, `s = 3x² + a`
-/// for `double`; `inv` is `generic`'s disequality witness `b·d⁻¹` (the null
-/// ptr on `double`, whose `y₁ ≠ 0` rides the slope pin instead).
+/// for `double`. Cell 2 is unused — `generic`'s disequality now rides `d`'s
+/// own `UintAdd` tuple (`nz = 1`, see the module doc) rather than a
+/// separate witnessed inverse.
```
adr1anh: No need to mention what cell 2 used to do

precompiles-prover/src/ec/require.rs
```diff
@@ -452,7 +444,9 @@ impl<'a> EcRequire<'a> {
         // maximum (> operands), satisfying the strict ordering the cert
         // rests on; a hit reuses its existing certified row and mints = false.
         let (r, mints) = self.store.add_point_cert(group, x3, y3);
-        ([slope_aux, lambda, inv, t, y3, e, null, x3, null], r, mints)
+        // Cell 2 (once the disequality witness `inv`) is unused: `generic`'s
+        // `d ≠ 0` now rides `d`'s own `UintAdd` tuple (`nz = 1`).
```
adr1anh: Same comment about mentioning unused cell

precompiles-prover/src/hash/keccak/round/mod.rs
```diff
@@ -49,12 +50,12 @@ pub const COL_A_HI: usize = 2;
...
-/// Logic intermediate: `r = a OP b` on logic rows, `r = a` otherwise.
-pub const COL_R_LO: usize = 5;
-pub const COL_R_HI: usize = 6;
-/// Destination value: `c = ROL(r, s)` on ROL rows, `c = r` otherwise.
-pub const COL_C_LO: usize = 7;
-pub const COL_C_HI: usize = 8;
+/// Destination value: `c = ROL(a OP b, s)`. The logic intermediate `r`
+/// is no longer committed — the bitwise64 buses (`Logic64`/`Rol64`/the
+/// fused `XorRol64`) relate `a`, `b`, `c` directly, so the round carries
+/// only the operands and the result.
```
adr1anh: No need to mention "no longer committed"

precompiles-prover/src/hash/keccak/round/mod.rs
```diff
@@ -216,19 +197,15 @@ impl LiftedAir<Felt, QuadFelt> for KeccakRoundAir {
...
-/// and numerator) are each deg 3 after the 3-batch recurrence, and the
-/// last-row close gates the constraint by the degree-1 `is_transition` /
-/// `is_last_row` selector — `u·(acc_next − Σ acc) − v` on transitions,
-/// `u·(σ − Σ acc) − v` on the last row — so col 0 lands at deg 5 (was 4
-/// under the older ungated `u·(acc_next + σ·inv_n − Σ acc) − v` form).
-/// Col 1's deg is 3 (ungated fraction column). `log_quotient_degree`
-/// stays 2 (`⌈log₂ 4⌉ = ⌈log₂ 3⌉ = 2`); aux blowup factor = 4.
-const COLUMN_SHAPE: [usize; NUM_AUX_COLS] = [3, 2];
+/// Every closing constraint is degree ≤ 3, so `log_quotient_degree = 1`
+/// (aux blowup factor = 2). Width disregarded (research/logup-flatten).
```
adr1anh: > `research/logup-flatten`

maybe remove ?

precompiles-prover/src/primitives/bitwise64.rs
```diff
@@ -484,13 +510,25 @@ impl Bitwise64Requires {
...
+    /// Populated trace height before power-of-two padding: the busiest
+    /// lane's row count after `build_lanes` balances whole chains across
+    /// [`NUM_LANES`] bands.
+    pub fn populated_rows(&self) -> usize {
+        build_lanes(&self.requests)
+            .iter()
+            .map(|lane| lane.iter().map(|ch| ch.logics.len() + 1).sum::<usize>())
+            .max()
+            .unwrap_or(0)
+    }
```
adr1anh: This function allocates the entire trace just to get the number of rows.

precompiles-prover/src/primitives/bitwise64.rs
```diff
@@ -598,38 +636,76 @@ fn claim_producer(
...
-/// either the ROL cap or one trailing dead `Carrier` holding the uncapped
-/// tail. Pads to `next_power_of_two` with all-zero rows; minimum height 1.
+/// Lowers the recorded requests into chains (`build_chains`), balances
+/// whole chains across [`NUM_LANES`] column bands (`build_lanes`), and
+/// lays each lane independently: its LOGIC links as `Real` rows — every
+/// link's `a` is the prior link's `c`, so the AIR's c-in-next-row holds by
+/// construction — then either the ROL cap or one trailing dead `Carrier`
+/// holding the uncapped tail. Each lane is zero-padded past its content;
+/// the matrix pads to `next_power_of_two` of the busiest lane, minimum 2.
 pub fn generate_trace(requires: Bitwise64Requires) -> RowMajorMatrix<Felt> {
```
adr1anh: Do any of the loops in the trace gen need to be parallelized?

precompiles-prover/src/hash/keccak/round/mod.rs
```diff
@@ -93,6 +93,21 @@ pub fn lane_base(lane: usize) -> usize {
...
+        let base = lane_base(lane);
+        for r in 0..height {
+            let src = &cells[r * LANE_WIDTH..(r + 1) * LANE_WIDTH];
+            let row_start = r * NUM_MAIN_COLS + base;
+            trace[row_start..row_start + LANE_WIDTH].copy_from_slice(src);
+        }
+    }
+    RowMajorMatrix::new(trace, NUM_MAIN_COLS)
+}
+
```
adr1anh, line 125: The memory access pattern here is scattered. We could maybe change the input to be a matrix too, and transpose it.

precompiles-prover/src/transcript/poseidon2/mod.rs
```diff
@@ -395,71 +411,101 @@ where
...
+        // Per-column constraint degree bound (shared across cols 0-2).
         let group_deg = Deg { v: 5, u: 4 };
 
+        // Bus emissions split across 3 columns (rather than one mutex-batched
+        // column) so each column's fraction count stays small. The row
+        // selector is folded into each insert's own multiplicity instead of
+        // an outer batch gate — col 0: in_rate0. col 1: in_rate1 + out_rate0.
+        // col 2: in_cap.
+        let m_in_rate0 = is_init_ext.clone() * neg_in_mult.clone();
+        let m_in_rate1 = is_init_ext.clone() * neg_in_mult;
```
adr1anh: These two are the same, is that intensional?

precompiles-prover/src/ec/add/mod.rs
```diff
@@ -261,13 +261,23 @@ const PCOL_TERM: usize = 3;
...
+// - col 4: the generic case's `d`-subtract + chord certificates.
+// - col 5: the double case's tangent-numerator + slope-pin certificates.
+// - col 6: the generic case's `t`-add + fused `x₃` mul-subtract.
+// - col 7: the shared tail's `e`-subtract + fused `y₃` mul-subtract.
+// - col 8: the double case's fused `x₃` mul-subtract, alone (no partner left to pair).
+// - col 9/10: the closure-cert ptr-ordering Range16 limb pairs.
+// - col 11: the result-membership cert provide, alone.
+const NUM_LOGUP_COLS: usize = 12;
+const AUX_WIDTH: usize = 12;
+const COLUMN_SHAPE: [usize; NUM_LOGUP_COLS] = [1, 2, 2, 2, 2, 2, 2, 2, 1, 2, 2, 1];
```
adr1anh, line 277: Is the degree reduction from 9 to 3 a bit too aggressive, at the cost of verifier time. Would it make sense to delay this commit (64d403bc040aec325550fda8a3f2318a66042bec) until we can benchmark together with the recursive verifier?

precompiles-prover/docs/airs/bitwise64.md
adr1anh, line 1: Was this meant to be deleted?

adr1anh, line 1: This commit 0daf7098e10dcb128bec1b4ca609a118d916adce also seems very aggressive to me. I can definitely see how it helps now, but I'm wondering if this will still be the right approach when we start implementing SHA512 which also relies heavily on u64 operations.

adr1anh (commented):
Thank you for all of these investigations, the tricks and optimizations are very elegant. The prover performance gain is very nice. 

There are some commits that I flagged because they seem quite aggressive in the tradeoff between degree reduction and verifier cost. These might be a bit premature until we have a better idea of how this will affect the recursive verifier. One metric that would have been interesting to note in each optimization was the impact on both proof size (as a proxy for the number of hashes), as well as the native verifier. While we don't have the ACE integration yet, the number of gates would also be a good metric to measure. 

Another thought that occurred to me was that we might want to investigate whether having separate quotients for the main and aux constraints at the STARK level helps alleviate the cost of higher degree aux columns. 

Overall, I think most of the commits here definitely make sense, but we may want to wait for the more aggressive ones until we can evaluate the end to end impact.

precompiles-prover/docs/airs/bitwise64.md
Nashtare: Removed after https://github.com/0xMiden/miden-vm/commit/0daf7098e10dcb128bec1b4ca609a118d916adce which would need reverting if we abort this commit

Nashtare:
@adr1anh For a global overview of the impact of each "perf" commit on some meaningful metrics, here is a compiled artifacts. Note that https://github.com/0xMiden/miden-vm/pull/3347 would significantly improve the ACE circuit results (more than the current precompile VM one on your branch).

https://claude.ai/code/artifact/80451a19-f556-4536-8175-9a23e63a1697

On ACE circuit size, considering #3347

  ```bash
  ┌────────────────────┬───────────────┬────────────┬──────────────┐
  │       metric       │ without #3347 │ with #3347 │    delta     │
  ├────────────────────┼───────────────┼────────────┼──────────────┤
  │ ace_num_inputs     │          3392 │       3138 │ −254 (−7.5%) │
  ├────────────────────┼───────────────┼────────────┼──────────────┤
  │ ace_num_eval_gates │         10308 │       9696 │ −612 (−5.9%) │
  └────────────────────┴───────────────┴────────────┴──────────────┘
  ```

Nashtare:
@adr1anh Looking more specifically at the concerning commits you highlighted + the Poseidon2 flattening one, here are their specific impact, although as we can see the ACE circuit overhead for each is minor (<2%):

| commit reverted | ace_inputs | ace_eval_gates | proof (B) | deferred proof (B) | native verify (ms) |
  |---|---:|---:|---:|---:|---:|
  | *HEAD* | 3392 | 10308 | 392855 | 297764 | 11.37 |
  | `83b5395f8` (poseidon2 lqd lowering) | 3376 (−16) | 10164 (−144) | 393879 (+1024) | 298820 (+1056) | 11.21 |
  | `64d403bc0` (ec/keccak/bpl/transcript flatten) | 3296 (−96) | 10172 (−136) | 382455 (−10400) | 289220 (−8544) | 11.00 |
  | `0daf7098e` (KeccakRound/Bitwise64 merge) | 3260 (−132) | 10112 (−196) | 374184 (−18671) | 280565 (−17199) | 11.19 |

precompiles-prover/src/hash/keccak/round/mod.rs
Nashtare: I actually swapped the loop to row-outer / lane-inner, which ends up being even faster

Nashtare:
Also @adr1anh I have another branch on top of this (currently blocked by https://github.com/0xMiden/miden-vm/pull/3344) that would bring the ACE numbers down: `num_inputs` 3128 → 1728 (−45%), `num_eval_gates` 9612 → 6440 (33%)

adr1anh (approved):
These optimizations are really good. My previous comment mentioned we may want to be careful with overoptimizing for the prover without taking the recursive verifier into account. After async discussion, we are okay with merging these optimizations now, and potentially reverting them later. LGTM!

PR #3344, opened 2026-07-09
Add precompile ACE relation digest

adr1anh:
## Summary

Adds ACE relation digest support for the precompile prover.

This extends `miden-ace-codegen` with first-class preprocessed trace inputs, then uses that support to build and snapshot the 15-chiplet precompile multi-AIR ACE circuit. The precompile prover now binds a generated `PRECOMPILE_RELATION_DIGEST` instead of the previous `[Felt::ZERO; 4]` placeholder.

## Changes

- Add preprocessed trace support to `miden-ace-codegen`:
  - `InputCounts::preprocessed_width`
  - `InputKey::Preprocessed { offset, index }`
  - `preprocessed_curr` / `preprocessed_next` layout regions
  - lowering for `BaseEntry::Preprocessed`
  - tests for preprocessed layout, lowering, and evaluation

- Add `precompiles-prover/src/ace.rs`:
  - builds the combined ACE circuit for `ChipletAir::all()`
  - uses the existing chiplet ordering directly
  - leaves `ChipletMultiAir::eval_external` outside the ACE circuit, matching the VM AIR pattern

- Replace the placeholder `PRECOMPILE_RELATION_DIGEST` with the generated digest:
  - `[12794781063212927046, 9890233791820081176, 1897898501728194857, 10203667064001467379]`

- Add an insta snapshot test for the precompile relation digest:
  - `num_inputs: 2816`
  - `num_eval_gates: 9332`

## Notes

This intentionally uses the given `ChipletAir::all()` ordering for now and does not enforce ascending-height ordering.

The relation digest mirrors the existing VM formula:

```text
Poseidon2::hash_elements([PROTOCOL_ID, ACE_CIRCUIT_HASH...])
```

## Validation

- `cargo test -p miden-ace-codegen`
- `cargo test -p miden-air relation_digest_matches_current_air`
- `cargo check -p miden-precompiles-prover --no-default-features`
- `cargo test -p miden-precompiles-prover --no-default-features precompile_relation_digest_matches_current_air`
- `cargo test -p miden-precompiles-prover precompile_relation_digest_matches_current_air`
- `cargo test -p miden-precompiles-prover precompile_multi_air_ace_circuit_builds`
- `cargo test -p miden-precompiles-prover`
- `make lint`
- `make doc`


precompiles-prover/src/ace.rs
```diff
@@ -0,0 +1,275 @@
...
+            NodeKind::Constant(v) => builder.constant(v),
+            NodeKind::Add(a, b) => builder.add(translation[a.index()], translation[b.index()]),
+            NodeKind::Sub(a, b) => builder.sub(translation[a.index()], translation[b.index()]),
+            NodeKind::Mul(a, b) => builder.mul(translation[a.index()], translation[b.index()]),
+            NodeKind::Neg(a) => builder.neg(translation[a.index()]),
+        };
+        translation.push(new_id);
+    }
+    translation
+}
```
Nashtare: isn't it basically a duplicate of the air/src/ace.rs version?

precompiles-prover/src/ace.rs
```diff
@@ -0,0 +1,275 @@
...
+///    shifting preprocessed/main/aux/bus-boundary slot indices into that AIR's subregion and
+///    tagging row selectors with the AIR's `ChipletAir::all()` instance index.
+/// 3. β-fold all AIR accumulators and subtract the single shared `q·v` quotient binding.
+///
+/// The caller order is exactly [`ChipletAir::all()`]. We intentionally do not sort by height or
+/// proof order here; verifier-side `MultiAirBeta(i)` assignment is expected to handle proof-order
+/// binding outside this circuit.
+pub fn build_precompile_multi_air_ace_circuit(
+    config: AceConfig,
+) -> Result<AceCircuit<QuadFelt>, AceError> {
```
Nashtare, line 26: I think also here quite a few pieces are shared with the air/src/ace.rs module, maybe it'd be worth sharing primitives? non-blocking though

Nashtare (approved):
LGTM overall

huitseeker (commented):
Not sure I get how this works: the combined precompile ACE circuit lowers each chiplet AIR with its local periodic max period, but the final circuit has one global ZK input.  

IIUC the current chiplet set mixes shorter periods with a period-128 Keccak round AIR, are smaller-period columns evaluated at the right point? Does the relation digest bind to a circuit that matches the verifier expression?

adr1anh:
> Not sure I get how this works: the combined precompile ACE circuit lowers each chiplet AIR with its local periodic max period, but the final circuit has one global ZK input.  
> 
> IIUC the current chiplet set mixes shorter periods with a period-128 Keccak round AIR, are smaller-period columns evaluated at the right point? Does the relation digest bind to a circuit that matches the verifier expression?

This is something we had to handle for the Miden AIR already, and uses lifting to handle these mixed sizes. Each trace `i` gets its own `z_i = z^{N_max/n_i}` where `N_max` is the size of the largest subgroup of roots of unity. The columns of the trace evaluated at this point repeat with period `n_i`.


Within the AIR for this trace, we evaluate a periodic column at `z_i' = z_i^{n_i/p} = z^{N_max/p}`, which makes it repeat with a period `p`. Therefore, by providing `z` as part of the ACE input, the circuit computes `z_i'` at which it evaluates the periodic polys, independent of the height of any traces in the AIR.

This ensures the relation digest is constant across all instances. 

Constraints referencing periodic columns generally enforce that the minimum trace height is at least as tall as the maximum period. If we supplied shorter traces, these would be lifted to match the period, but this usually does not satisfy the constraints due to the construction of the constraints. If it did, this would probably be a soundness error, outside of the scope of the ACE/verifier.

huitseeker:
@adr1anh IIUC, the combined ACE circuit gives periodic columns the same point as the lifted STARK verifier: each AIR uses `z_i = z^(N_max / n_i)`, so a period `p` column is evaluated at `z_i^(n_i / p) = z^(N_max / p)`, matching the [lifted verifier rule](https://github.com/0xMiden/p3-miden/blob/v0.5.0/p3-miden-lifted-stark/src/verifier/periodic.rs#L48-L73). 

A concern is that ACE defines `InputKey::ZK` as [`zeta^(N / max_cycle_len)`](https://github.com/0xMiden/miden-vm/blob/40072502e3bcb9007566d987cb3196d521acb83f/crates/ace-codegen/src/layout/keys.rs#L32-L40), then [lowers periodic columns](https://github.com/0xMiden/miden-vm/blob/40072502e3bcb9007566d987cb3196d521acb83f/crates/ace-codegen/src/dag/lower.rs#L272-L309) from that value. 
  
The thing that wasn't clear from the ACE documentation: what `ZK` value does the combined precompile verifier supply (raw `z`, per-AIR `z_i`, or one global `z^(N / max_cycle_len)`?), and what `max_cycle_len` does it use?

Nashtare:
I think @huitseeker raised a good point, although I think the fix is simple and just requires to lower against a global max cycle length `ratio = 128 / p` for every AIR, with the verifier supplying one `ZK = z^(N / 128)`. This should just add a few cached squarings and keeps the single ZK input?

adr1anh:
Thanks @Nashtare!

I addressed this by moving mixed-period handling into `miden-ace-codegen`: batch lowering now derives one global shared period across all AIRs and evaluates each period-`p` column from the single `ZK` basis using `shared_period / p`, so the Miden and precompile circuits follow the same path. While doing this, I also centralized the duplicated multi-AIR layout, DAG re-emission, and β-folding logic into a common builder, leaving both consumers as thin policy wrappers. I rewrote the branch into four commits—preprocessed inputs, shared-period lowering, common multi-AIR codegen, and precompile digest integration—so the periodic correctness fix can be reviewed independently from the refactor; all circuit commitments, gate counts, and input counts remain unchanged.

Nashtare (approved):
Thanks Adrian!

huitseeker (dismissed):
~Thanks a lot, this looks great!~ Looking into the test failure

precompiles-prover/src/stark_config.rs
```diff
@@ -37,6 +37,19 @@ pub type PackedFelt = <Felt as Field>::Packing;
 /// Number of inputs to the Merkle compression function.
 const COMPRESSION_INPUTS: usize = 2;
 
+// DOMAIN-SEPARATED FIAT-SHAMIR TRANSCRIPT
+// ================================================================================================
+
+/// PRECOMPILE_RELATION_DIGEST = Poseidon2::hash_elements([PROTOCOL_ID, CIRCUIT_COMMITMENT]).
+///
+/// Compile-time constant binding the Fiat-Shamir transcript to the precompile chiplet AIR set.
```
huitseeker: The doc says this binds the transcript to "the precompile chiplet AIR set", but the digest is a bit narrower than that wording suggests.

  As I read the code, `crate::ace::build_precompile_multi_air_ace_circuit` builds the ACE circuit in canonical `ChipletAir::all()` order with identity `proof_order`, while the lifted STARK proof derives proof order from trace heights (`crates/lifted-stark/src/order.rs`). Also,
  `ChipletMultiAir::eval_external` remains outside the ACE circuit hash, including the cross-chiplet sigma sum and fixed boundary correction.

  Could we add a short caveat here so future readers do not assume this digest covers the whole precompile relation? For example:

  ```suggestion
  /// Compile-time constant binding the Fiat-Shamir transcript to the canonical per-AIR ACE
  /// circuit generated from `ChipletAir::all()` order. Cross-chiplet external assertions
  /// (`ChipletMultiAir::eval_external`) and trace-height-derived proof order are checked by
  /// the lifted STARK protocol outside this circuit hash.
  /// Keep this in sync with [`crate::ace::build_precompile_multi_air_ace_circuit`].
  ```

huitseeker (approved):
This is fine, just needed a relation update.

PR #3345, opened 2026-07-09
Split Poseidon2 permutations into a separate table

### LINKED ISSUE #3251: refactor: replace hardcoded AIR reordering with a committed permutation tree

Al-Kindi-0:
This PR splits Poseidon2 permutation execution out of `ChipletsAir` into a separate `Poseidon2PermutationAir`.

Main changes:

- Move Poseidon2 permutation rows into a separate table.
- Keep hasher controller rows in `ChipletsAir`.
- Add the ACE registry and AIR-order handling needed to robustly support more than two AIRs.
- Fix a Poseidon2 permutation-link soundness issue by binding controller input/output rows to the same permutation id.
- Add synthetic recursive verifier benchmarks.

Closes #3251.

## Design

The multi-AIR instance order is:

1. `Core`
2. `Chiplets`
3. `Poseidon2Permutation`

The hasher controller remains in `ChipletsAir`. It records the VM-side request and response rows for hasher operations. The Poseidon2 permutation trace moves to `Poseidon2PermutationAir`, where each permutation occupies one 16-row cycle.

The recursive verifier now needs ACE circuits for every possible AIR proof order. The registry keeps all `3! = 6` proof orders. This is intentional: it lets the MASM verifier copy OOD values into the READ section in proof order instead of reordering them at runtime. This is an acceptable solution when the number of AIRs is small (e.g., less than 10).

## Tightening the controller-to-permutation link

The hasher/Poseidon2 link uses keyed input/output messages:

- controller input row: `(perm_id, input_state)`
- controller output row: `(perm_id, output_state)`
- Poseidon2 AIR row 0 removes `(perm_id, input_state)`
- Poseidon2 AIR row 15 removes `(perm_id, output_state)`

Without the controller-side binding, a prover could mount a balanced tuple-swap attack in which it swaps `(perm_id, output)` pairs across controller outputs while keeping the LogUp bus balanced.

The fix has two parts:

- `perm_id` is constant across a Poseidon2 cycle and increments at cycle boundaries.
- The controller constrains the input and output rows of one permutation request to use the same `perm_id`.

The Poseidon2 AIR's transition constraints tie each input state to its output state. The bus ties the controller side to the corresponding Poseidon2 cycle.

## Benchmark Methodology

There are two benchmark comparisons.

### Fixed MASM

Synthetic MASM programs are generated once on the baseline branch, then reused unchanged on both branches. This is the cleaner apples-to-apples comparison because both branches prove the same programs.

### Automatic Pipeline

Each branch runs the normal synthetic benchmark pipeline and calibrates/emits its own MASM programs. This is easier to reproduce, but it is not a perfect apples-to-apples comparison because old `next` snapshots do not contain `poseidon2_permutation_rows`.

### Fixed MASM

| scenario | next padded | split padded | next prove | split prove | improvement |
|---|---:|---:|---:|---:|---:|
| create single P2ID note | 131,072 | 131,072 | 1.63s | 1.55s | 5.1% |
| consume single P2ID note | 131,072 | 131,072 | 1.65s | 1.54s | 6.4% |
| consume two P2ID notes | 262,144 | 131,072 | 2.46s | 1.59s | 35.2% |
| consume CLAIM note L1 | 262,144 | 131,072 | 2.04s | 1.20s | 41.4% |
| consume CLAIM note L2 | 262,144 | 262,144 | 2.06s | 1.83s | 11.1% |
| consume B2AGG note | 1,048,576 | 1,048,576 | 8.30s | 7.39s | 11.0% |

For these fixed programs, `split chiplets + split p2` matches `next chiplets` within 8 or 16 rows.

### Automatic Pipeline

| scenario | next prove | split prove | delta |
|---|---:|---:|---:|
| create single P2ID note | 1.61s | 1.54s | -4.1% |
| consume single P2ID note | 1.60s | 1.55s | -3.4% |
| consume two P2ID notes | 2.39s | 2.19s | -8.5% |
| consume CLAIM note L1 | 2.01s | 1.80s | -10.5% |
| consume CLAIM note L2 | 2.01s | 1.80s | -10.3% |
| consume B2AGG note | 8.10s | 7.26s | -10.3% |

## Recursive Benchmarks

The recursive benchmark verifies multiple proofs for the synthetic `consume single P2ID note` fixture.

### Fixed MASM

| proofs | next avg | split avg | improvement | next padded | split padded |
|---:|---:|---:|---:|---:|---:|
| 2 | 2.39s | 1.83s | 23.5% | 262,144 | 131,072 |
| 3 | 4.76s | 3.28s | 31.1% | 524,288 | 262,144 |
| 4 | 4.81s | 3.61s | 25.0% | 524,288 | 262,144 |
| 5 | 4.85s | 4.94s | -1.9% | 524,288 | 524,288 |
| 6 | 9.61s | 6.58s | 31.6% | 1,048,576 | 524,288 |
| 7 | 9.69s | 7.42s | 23.5% | 1,048,576 | 524,288 |
| 8 | 9.72s | 7.54s | 22.4% | 1,048,576 | 524,288 |

### Automatic Pipeline

| proofs | next avg | split avg | delta | next padded | split padded |
|---:|---:|---:|---:|---:|---:|
| 2 | 2.41s | 1.80s | -25.4% | 262,144 | 131,072 |
| 3 | 4.79s | 3.25s | -32.0% | 524,288 | 262,144 |
| 4 | 4.86s | 3.66s | -24.6% | 524,288 | 262,144 |
| 5 | 4.89s | 4.94s | +1.0% | 524,288 | 524,288 |
| 6 | 9.76s | 6.54s | -33.0% | 1,048,576 | 524,288 |
| 7 | 10.23s | 7.33s | -28.4% | 1,048,576 | 524,288 |
| 8 | 10.96s | 7.53s | -31.3% | 1,048,576 | 524,288 |

The 5-proof case remains the boundary case: the split branch reduces raw trace height, but both branches land on the same padded height, so the proving-time win does not materialize there.


benches/synthetic-bench/benches/recursive_verify.rs
```diff
@@ -0,0 +1,993 @@
...
+    fn poseidon2_permutation_trace_len(&self) -> usize;
+}
+
+// `next` has no separate Poseidon2 AIR. Branches which add one expose an inherent method with this
+// name; inherent methods take precedence over this shim.
+impl TraceLenSummaryExt for TraceLenSummary {
+    fn poseidon2_permutation_trace_len(&self) -> usize {
+        0
+    }
+}
```
huitseeker: This trait and its impl look like dead code on this branch. `TraceLenSummary` already has an inherent `poseidon2_permutation_trace_len` method at `processor/src/trace/utils.rs:250`, and Rust's inherent-vs-trait resolution rules mean the inherent method always wins. The call at line 402 resolves to the inherent method, so the trait shim is never consulted.

The inline comment acknowledges this ("inherent methods take precedence over this shim"), which is exactly why it is unreachable here.

If the shim only exists so the file compiles on `next` (which has no separate Poseidon2 AIR), that cross-branch concern probably belongs in a feature gate or a separate shared file rather than a `#[allow(dead_code)]` trait sitting in the bench.

benches/synthetic-bench/benches/recursive_verify.rs
```diff
@@ -0,0 +1,993 @@
...
+///
+/// `generate_advice_inputs` parses the inner STARK proof and returns the exact advice stack,
+/// Merkle store, and advice-map entries expected by `exec.vm::verify_proof`.
+/// The stack is ordered so its first element is the next value consumed by the VM.
+fn recursive_proof_advice(fixture: &TxProofFixture) -> RecursiveProofAdvice {
+    let pub_inputs = PublicInputs::new(
+        fixture.program_info.clone(),
+        fixture.stack_inputs,
+        fixture.stack_outputs,
+        PrecompileTranscriptState::default(),
```
huitseeker: The recursive advice here uses `PrecompileTranscriptState::default()`, which is only correct when the fixture emits no precompile requests. 

The synthetic P2ID/CLAIM/B2AGG fixtures in the PR all carry zero precompile requests, so this works today. But if someone later points `RECURSION_BENCH_MASM` at a fixture that does emit precompiles, the advice inputs would be built against the wrong transcript.

Could we make the assumption explicit at fixture load, e.g. assert `proof.precompile_requests().is_empty()` (or panic with a clear message pointing at this line) in `load_tx_fixtures` ?

huitseeker (approved):
LGTM!

benches/synthetic-bench/benches/recursive_verify.rs
Al-Kindi-0: fixed

benches/synthetic-bench/benches/recursive_verify.rs
Al-Kindi-0: fixed

Issue #3339, opened 2026-07-07
Merge `0xMiden/crypto` into `miden-vm` with history


huitseeker:
For simplicity and ease of release should move the `0xMiden/crypto` crates into `miden-vm` and keep their source history. Use `git subtree` for the import so the PR records the upstream commit and keeps a link to the old repo history.

## How

Do the migration in reviewable commits. First, import the chosen `0xMiden/crypto` commit with `git subtree` into a staging path, with no source edits. The commit message should name the source repo, source commit, subtree prefix, and whether the import keeps full history or uses `--squash`. We need full walkable history, so do not use `--squash`. 

Then move the publishable crates to their final paths and wire them into the VM workspace.

| Package | Destination |
| --- | --- |
| `miden-crypto` | `crates/crypto` |
| `miden-crypto-derive` | `crates/crypto-derive` |
| `miden-field` | `crates/field` |
| `miden-serde-utils` | `crates/serde-utils` |
| `miden-lifted-air` | `crates/lifted-air` |
| `miden-lifted-stark` | `crates/lifted-stark` |
| `miden-stark-transcript` | `crates/stark-transcript` |
| `miden-stateful-hasher` | `crates/stateful-hasher` |

Keep the package names unchanged. Add each crate to root workspace members and `[workspace.dependencies]`. Remove imported repo files that no longer apply, such as crypto CI, nested lockfiles, release-plz config, toolchain files, and standalone changelogs.

### Non-publish crates

Handle the non-publishable code in the same migration. Move `miden-crypto-fuzz` and `miden-serde-utils/fuzz` into the VM fuzz setup or into named standalone fuzz workspaces. Move useful Wycheproof tests into the VM test layout. Move useful benchmarks into the VM benchmark layout. Keep `tools/zeroize-audit` only if the PR wires it into VM tooling.

### Make & CI

Merge the Make targets that still apply. This likely includes TOML formatting, spell checking if we keep `typos`, workspace lint checks, crypto no-std checks, `wasm32-wasip2` checks, fuzz build checks, useful crypto benchmark targets, and zeroize audit if that tool moves.

[...]

PR #3352, opened 2026-07-10
Cargo workspace publish

huitseeker:
This replaces the file-driven selected release flow with a Cargo workspace publish flow.

The shared workspace-release action now derives publishable crates from `cargo metadata` instead of r`elease-packages.txt`. Empty packages means all publishable workspace crates. A non-empty packages input still supports comma, space, or newline separated selected releases. The action rejects duplicate or invalid selected package names before running Cargo.

The action now runs `cargo publish --workspace --locked`. In dry-run mode it adds `--dry-run`. It also excludes workspace members with `publish = false`, because Cargo fails on those under `--workspace`.

Already-published crate versions are strict by default. If a selected crate version already exists on crates.io, publish fails before upload. The publish workflow adds an explicit `allow_existing` input for manual recovery from partial publish failures. When `allow_existing=true`, already-published crate versions are excluded and Cargo publishes the remaining selected crates.

The release dry-run workflow *always* passes `allow_existing: "true"`. This fixes the `next` dry-run failure where `miden-utils-core-derive` v0.25.0 (e.g.) already existed on crates.io. Dry-runs can now run after a release while still checking any unpublished selected crates.

The publish workflow removes the `packages_file` input and adds `allow_existing`, defaulting to false. Normal releases keep the version-bump guard. Manual reruns can opt into resume behavior.

Removed files:

  - release-packages.txt
  - scripts/check-release-packages.sh
  - scripts/release-selected.sh

One behavior is not preserved: the workflow no longer retries each crate independently inside one run. cargo publish --workspace is a single Cargo operation. If crates.io accepts some crates and then fails or times out, a maintainer can rerun the publish workflow with `allow_existing=true` to skip crates that already reached crates.io.


Mirko-von-Leipzig (approved):
Workflow eyeballs as correct; and approach looks good.

I'm unsure of the underlying need to only publish certain packages; so it does feel overly complex to me personally e.g. why not just publish everything always.

If only publishing certain subsets is desired then that also shouldn't be manually done imo, but rather could just be done by bumping its version, which would still work under _publish all_ flow.

huitseeker:
@Mirko-von-Leipzig The workflow you're pointing at, which does make a ton of sense, is exactly what you get if you happen to not use the `packages` argument.

Mirko-von-Leipzig:
> @Mirko-von-Leipzig The workflow you're pointing at, which does make a ton of sense, is exactly what you get if you happen to not use the `packages` argument.

I get that, I more meant why provide additional knobs/complexity if that covers everything.

huitseeker:
@Mirko-von-Leipzig One of the issues with the reconciliation mode (`cargo publish --workspace` with `allow_existing = "true"`) is that the state of each crate's `Cargo.toml` must either be "already published" or "about to be published and ready to go".

This prevents a patch release of say one crate (e.g. `midenc-hir-type`) while the rest of the code base is working on preparing the release of a larger bunch of crates (e.g. the rest of the VM crates). This may become more acute with #3339.

But happy to remove the functionality if we think it's uneeded for now.

Mirko-von-Leipzig:
> @Mirko-von-Leipzig One of the issues with the reconciliation mode (`cargo publish --workspace` with `allow_existing = "true"`) is that the state of each crate's `Cargo.toml` must either be "already published" or "about to be published and ready to go".
> 
> This prevents a patch release of say one crate (e.g. `midenc-hir-type`) while the rest of the code base is working on preparing the release of a larger bunch of crates (e.g. the rest of the VM crates). This may become more acute with #3339.
> 
> But happy to remove the functionality if we think it's uneeded for now.

No worries, I think keep it. Was more just curious

bitwalker (approved):
I think this looks good - some trial runs with different configurations will say for sure if it is an improvement, but I'm liking this better already.

PR #3353, opened 2026-07-10
Merge next into precompiles branch and fix release order

huitseeker:
On top of #3222. This merges the current `origin/next` into `precompiles/crate`.

The merge resolves the AIR and recursive verifier artifact conflicts by regenerating the constraint artifacts. It also moves `miden-precompiles` before `miden-processor` in `release-packages.txt`, since `miden-processor` now depends on it.

Merge with `merge` not `squash`


adr1anh (approved):
TYVM!

PR #3355, opened 2026-07-10
processor: ensure debug info is available via ResumeContext

bitwalker:
The v0.25.0 release shipped without making debug information attached to `ResumeContext` available outside the `miden-processor` crate itself, effectively making debug info inaccessible. This prevents the debugger from being able to upgrade to v0.25.0, as well as the compiler (whose test suite depends on the debug executor).

This PR makes this information available externally, addressing that oversight.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## TBD
```
bobbinth: I'd update this to `v0.25.1 (2026-07-10)`

bobbinth (approved):
Looks good! Thank you! I'd also increment crate versions to v0.25.1.

Cargo.toml
```diff
@@ -41,7 +41,7 @@ rust-version = "1.96.1"
 # Private workspace members inherit this version. Publishable crates keep their
 # own package versions so release tooling can publish only the crates selected
 # for a release.
-version = "0.25.0"
+version = "0.25.1"
```
bobbinth, line 44: I think we now need to update this for all crates individually.

bitwalker: Interesting, I thought we inherited the workspace version for most crates, and only manually changed the versions for crates that don't share the workspace version. Noted.

bitwalker:
@bobbinth Does the release dry-run failure matter here? I have no idea how to exclude `midenc-hir-type` from the list of crates to publish without the `release package list` lint failing. Presumably there is a way to do it, but I don't know what the expected flow is here.

PR #3356, opened 2026-07-10
refactor: rename module and kernel descriptors

### LINKED ISSUE #2828: Release trace generation relies on wrapping subtraction for clk=0 memory access

huitseeker:
This renames two public types so their names describe what they hold.

`ModuleInfo` is now `ModuleDescriptor`. Its package and linker methods now use the same wording: `module_descriptors`, `try_module_descriptors`, and `kernel_module_descriptor`.

`Kernel` is now `KernelDescriptor`. The `TargetType::Kernel` and `ModuleKind::Kernel` enum variants are unchanged because they name the package and AST kind, not the descriptor type.

Closes #2828 #2944


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - [BREAKING] Optimize constraint evaluation step by merging one-hot gated stack op constraints ([#3333](https://github.com/0xMiden/miden-vm/issues/3333)).
 - [BREAKING] Removed `MastForest::compact`; MAST construction should deduplicate through builders or explicit `MastForest::merge` calls instead ([#3318](https://github.com/0xMiden/miden-vm/pull/3318)).
 - [BREAKING] Changed the ECDSA K256 Keccak public key commitment format to use affine public key coordinates (`qx_le_u32[8] || qy_le_u32[8]`) instead of compressed SEC1 public key bytes, aligning the core wrapper with the `miden-crypto` commitment format discussed in [0xMiden/crypto#1075](https://github.com/0xMiden/crypto/issues/1075). Existing public key commitments must be regenerated with `PublicKey::to_commitment()` ([#3342](https://github.com/0xMiden/miden-vm/pull/3342)).
+- [BREAKING] Renamed module and kernel metadata APIs from `ModuleInfo`/`Kernel` to `ModuleDescriptor`/`KernelDescriptor`, including matching module descriptor method names ([#3356](https://github.com/0xMiden/miden-vm/pull/3356)).
```
bobbinth: This should go into the unreleased v0.26.0 section.

core/src/program/kernel.rs
bobbinth: nit: I'd probably change "kernel" on line 18 to "descriptor".

crates/assembly-syntax/src/module.rs
```diff
@@ -9,11 +9,11 @@ use crate::{
     ast::{self, AttributeSet, ConstantValue, Ident, ItemIndex, ProcedureName, SubmoduleDecl},
 };
 
-// MODULE INFO
+// MODULE DESCRIPTOR
 // ================================================================================================
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ModuleInfo {
+pub struct ModuleDescriptor {
```
bobbinth, line 17: Not related to this PR, but may be good to add a brief doc comment explaining the intent of this struct.

miden-vm/src/cli/verify.rs
```diff
@@ -101,8 +101,8 @@ impl VerifyCmd {
     }
 }
 
-/// Loads a kernel from a file (.masm or .masp) and returns the Kernel.
-fn load_kernel(kernel_path: &PathBuf) -> Result<Kernel, Report> {
+/// Loads a kernel from a file (.masm or .masp) and returns the KernelDescriptor.
+fn load_kernel(kernel_path: &PathBuf) -> Result<KernelDescriptor, Report> {
```
bobbinth: I'd probably update the comment to say more like "Loads a kernel descriptor" and maybe also rename the function itself.

crates/mast-package/src/package/mod.rs
```diff
@@ -434,29 +435,29 @@ impl Package {
             })
     }
 
-    /// Returns an iterator over the module infos of the library.
-    pub fn module_infos(&self) -> impl Iterator<Item = ModuleInfo> {
+    /// Returns an iterator over the module descriptors of the library.
+    pub fn module_descriptors(&self) -> impl Iterator<Item = ModuleDescriptor> {
```
bobbinth: Not from this PR, but we should probably say "package" instead of "library" on line 438.

crates/mast-package/src/package/mod.rs
```diff
@@ -1037,8 +1039,8 @@ where
 
 /// Conversions
 impl Package {
-    /// Get a [Kernel] from this package, if this package contains one.
-    pub fn to_kernel(&self) -> Result<Kernel, Report> {
+    /// Get a [KernelDescriptor] from this package, if this package contains one.
+    pub fn to_kernel(&self) -> Result<KernelDescriptor, Report> {
```
bobbinth: Maybe this should be `to_kernel_descriptor()`?

bobbinth (approved):
Looks good! Thank you! I left some small comments inline.

PR #3357, opened 2026-07-11
Merge main/ into next/

huitseeker:
`next` should include the fixes and release updates that are now on `main`.

  This merges `main` into `next` after v0.25.1. It brings back the Cargo workspace publish changes from #3352 and the `ResumeContext` debug info fix from #3355.


bobbinth (approved):
Looks good! Thank you!

There is a changelog conflict, but I'm afraid to fix it using Github UI because I think this will mess with history.

Also, this PR should be merged using a merge commit (rather than squashed).

PR #3358, opened 2026-07-11
processor: provide means for constructing ResumeContext from package

bitwalker:
While migrating the debugger to v0.25, I discovered that we have no means of constructing an initial `ResumeContext` from  a package, and therefore the debug context for execution never gets initialized properly. The debugger requires the stepping APIs of the processor, which require a `ResumeContext` to call - the only way to construct one is via the `Processor` which only had an API that derived one from a `Program`, which lacks the necessary debug information.

This PR adds an additional API for constructing it from a package, while preserving the existing API - going forward we can probably remove the `Program` one, though it doesn't really cost anything to keep it around.

I've also set everything up so that this can be released as v0.25.2

### DIFF
--- a/processor/src/errors.rs
+++ b/processor/src/errors.rs
@@ -5,7 +5,10 @@ use alloc::{boxed::Box, string::String, sync::Arc, vec::Vec};
 
 use miden_core::program::MIN_STACK_DEPTH;
 use miden_debug_types::{Location, SourceFile, SourceSpan};
-use miden_mast_package::debug_info::{DebugSourceNodeId, PackageDebugInfo};
+use miden_mast_package::{
+    PackageDebugInfoError,
+    debug_info::{DebugSourceNodeId, PackageDebugInfo},
+};
 use miden_utils_diagnostics::{Diagnostic, miette};
 
 use crate::{
@@ -108,6 +111,8 @@ pub enum ExecutionError {
     ProvingError(String),
     #[error(transparent)]
     HostError(#[from] HostError),
+    #[error(transparent)]
+    PackageDebugInfoError(#[from] PackageDebugInfoError),
 }
 
 impl ExecutionError {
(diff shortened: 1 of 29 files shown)


processor/src/errors.rs
```diff
@@ -108,6 +111,8 @@ pub enum ExecutionError {
     ProvingError(String),
     #[error(transparent)]
     HostError(#[from] HostError),
+    #[error(transparent)]
+    PackageDebugInfoError(#[from] PackageDebugInfoError),
```
bobbinth, line 115: This technically makes this a breaking change - but I think it should be fine since nothing is using v0.25 VM yet.

processor/src/fast/mod.rs
```diff
@@ -281,6 +282,27 @@ impl FastProcessor {
         })
     }
 
+    /// Returns the resume context to be used with the first call to `step_sync()`.
+    ///
+    /// This function asserts that `package` is not of executable type - callers should ensure that
+    /// it is before calling
+    pub fn get_initial_resume_context_for_package(
```
bobbinth, line 289: Is the comment here correct? I believe we are asserting here that this is an executable package (via `package.unwrap_program()`).

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments. I'll merge this as is, and if needed, we can fix the comment in the `next` branch.

processor/src/errors.rs
bitwalker: We should mark these errors as non-exhaustive then - adding variants to these error types shouldn't ever cause downstream breakages (you are right that this is technically possible as it stands now though)

processor/src/fast/mod.rs
bitwalker: Indeed, that's a typo

processor/src/errors.rs
bobbinth: Fixed in 0a6742c74122c80c2b351f6f9356134e714b2707.

processor/src/fast/mod.rs
bobbinth: Fixed in 0a6742c74122c80c2b351f6f9356134e714b2707.

PR #3366, opened 2026-07-13
Merge the crypto repo in the VM

### LINKED ISSUE #3339: Merge `0xMiden/crypto` into `miden-vm` with history

huitseeker:
This closes #3339.

This imports the full 0xMiden/crypto history under imported/crypto, then moves the active crates into the VM workspace.

The VM now owns these crates:

- miden-crypto
- miden-crypto-derive
- miden-field
- miden-serde-utils
- miden-stateful-hasher
- miden-stark-transcript
- miden-lifted-air
- miden-lifted-stark

The PR also moves the crypto benches, fuzz targets, and Wycheproof tests into the VM repo, and wires them into the existing CI and release workflows.

## Code-level fixups (not import, move, or scaffolding) in this PR, in detail

I can remove any of these if controversial

The bulk of this PR is one of three things: a git-subtree import of `0xMiden/crypto` as-is, "move" commits that relocate imported crates into their final `crates/` paths without semantic edits, or scaffolding commits that adjust changelog, release management, and CI to cover the newly imported crates.

The commits below are none of those. Each one is a real code-level change to imported crypto or VM code and deserves individual review. They are grouped by theme.

### Hardening against attacker-controlled input

- 91872f4c78 — Reject malleable AEAD ciphertexts. Both `AeadPoseidon2` and `XChaCha` decrypt paths now wrap ciphertext parsing in `read_encrypted_data_strict`, which sizes a `BudgetedReader` to the input and rejects trailing bytes via `has_more_bytes()`. Without this, an attacker could append bytes to a valid ciphertext and have it decrypt successfully, which breaks the "one valid encoding" contract AEAD relies on for tag verification.
- 124e0e9a45 — Reject Falcon signature trailing bytes. `SignaturePoly::read_from_bytes` now scans the tail of the input and rejects non-zero trailing bytes, closing a malleability window on the compressed polynomial encoding.
- b823ff4940 — Harden Falcon deserialization against short buffers. Extracts `next_sig_poly_byte` so each byte read bounds-checks against `SIG_POLY_BYTE_LEN`. The previous form indexed directly and could read past the decoder's logical end on a malformed payload.
- 3162029589 — Cap budgeted zero-sized collections. `BudgetedReader::max_alloc` now returns `0` for zero-sized elements instead of `usize::MAX`, so a length-prefixed `Vec<ZST>` cannot claim `u64::MAX` elements and force a huge allocation before iteration begins. This is the only public-trait divergence from upstream `0xMiden/crypto@next` and is documented in the trait doc.
- 9c6af52df4 — Harden `Randomizable::from_random_bytes`. Every impl now uses `source.get(..N)?` instead of indexing into `source[..N]`. The old form panicked on short slices; the new form returns `None`, which is the contract the trait promises.

### Correctness fixes for hash and serialization

- 70dd4b1099 — Hash empty byte inputs. The algebraic sponge short-circuits on empty `bytes` to absorb the padding element before permutation. Without this, `hash_bytes(&[])` produced a digest inconsistent with the documented padding rule.
- ec2e087dbc — Hash empty domain-separated inputs. Corrects the condition under which a final permutation is applied when the element stream is empty but a domain tag was set. Same class of bug as above, for `hash_elements_in_domain`.
- 4210855139 — Doc clarification for the empty-byte hash behavior above.
- df6b9adeb2 — Restore compact SMT serialization budgets. Adds `min_serialized_size` to `NodeValue` so an empty-subtree-only `NodeValue` can be read under a tight budget. Without this, the recommended `read_from_bytes_with_budget` path rejected a legitimate minimal encoding.
- 383d989174 — Reset `ReadAdapter` buffer position. When the adapter's local buffer was drained to empty during a `read_slice`, the `pos` field was not reset, so the next read would compute a wrong offset. Adds `self.pos = 0` in the empty-buffer branch and a regression test.

### Shared-secret zeroization

- e266ece98e — Refactor `SharedSecret` in `k256` and `x25519` to hold an owned `bytes: [u8; 32]` instead of the upstream library's inner type. Removes a hand-rolled `unsafe` volatile-write block and delegates to the `zeroize` crate's `[u8; 32]` impl.
- 9d7572a10e — Add explicit `impl Drop for SharedSecret { fn drop(&mut self) { self.zeroize(); } }`. The prior commit switched to an owned bytes field, which silently lost the upstream `ZeroizeOnDrop` behavior; this restores it explicitly.

### Build and portability fixes

- 44c1aaef93 — Build the crypto SVE archive from target cfg. The upstream `build.rs` gated the SVE archive compile on `#[cfg(target_feature = "sve")]`, which does not fire in build scripts because they run on the host. This reads `CARGO_CFG_TARGET_ARCH` / `CARGO_CFG_TARGET_FEATURE` instead, which is the supported way. Also switches `std::ffi::c_ulong` to `core::ffi::c_ulong` in the FFI bindings for no-std compatibility, and adds a `rerun-if-changed` for the 256-bit SVE header.
- 688e1feab4 — Qualify word wrapper hex string. The derive macro emitted `pub fn to_hex(&self) -> String` with an unqualified `String`, which fails to compile for downstream no-std crates without `String` in scope. Now emits `alloc::string::String`.
- 40fde3c32f — Avoid caller alloc in word wrapper derive. Wraps the generated impl in `const _: () = { extern crate alloc; ... }` so the derive works for `#![no_implicit_prelude]` consumers. Adds a `no_implicit_prelude` test to pin it.

### Fuzz target quality

- 9d5f838846 — Wire deterministic crypto SMT fuzz target. Removes `rand` from the fuzz harness and derives the split percentage from the input bytes, so the same corpus entry reproduces across runs.
- a2749a4ccd — Compare SMT fuzz mutation errors. The fuzz target now compares errors structurally instead of `expect`-ing both sequential and parallel paths to succeed, so a legitimate error on one path no longer masks a divergence.

### Mixed commits worth noting

- 1fa7566a38 — Harden xchacha keys and docs-only diffs. Bundles two changes: the xchacha `SecretKey` switches to `SilentDebug`, `Zeroizing::new`, and `read_sensitive_array` (real hardening, same class as the SharedSecret work); and a tweak to the docs-only diff filter that was later simplified by 08336ca308. The xchacha part is the load-bearing half.
- 4c56c1fe50 — Document imported crypto safety contracts. Adds the `# Security` doc section on `Deserializable::read_from_bytes` spelling out the trusted-input contract, a comment on the AEAD back-compat early-return, and a comment on the SMT construction path documenting the "caller must discard on apply failure" assumption.
- 03820dba66 — Cover FRIE2F4 overflow shift constraints. Adds VM-side tests for the overflow shift constraint that were missing. Not crypto-related; caught during the migration's test sweep.

### Lockfile and CI hygiene (not security, but not pure scaffolding either)

- adc4520a06, 7a12aab200, a2e0123a07 — Refresh fuzz and root lockfiles after the version unification and path moves. These are mechanical, but they are not "moving" commits in the sense of relocating source — they re-pin dependency versions to keep `--locked` checks green.
- c8dc6ba5d0 — Restore CI after crypto workspace import. Mostly workspace dependency consolidation (`version = "..."` → `workspace = true`), but it also fixed a transient `use syn2 as syn;` duplicate that the migration had introduced. The duplicate was later removed by `a2e0123a07`.

Every other commit on the branch is either the subtree import itself, a pure file move (`chore: integrate crypto crates into workspace`, `chore: move fuzz workspaces under tools`), or CI/changelog/release scaffolding (`ci: *`, `docs: add crypto changelog`, `docs: unify release changelog history`, etc.).


Cargo.toml
```diff
@@ -102,37 +121,89 @@ wincode       = { version = "0.5.5", default-features = false, features = ["allo
...
 env_logger      = { version = "0.11" }
+flume           = { version = "0.12.0", default-features = false }
+hex             = { version = "0.4", default-features = false }
+hkdf            = { version = "0.13", default-features = false }
 insta           = { version = "1.43", default-features = false, features = ["colors"] }
 itertools       = { version = "0.15", default-features = false, features = ["use_alloc"] }
+k256            = { version = "0.14", default-features = false }
 log             = { version = "0.4", default-features = false }
+num             = { version = "0.4", default-features = false }
+num-complex     = { version = "0.4" }
```
Nashtare: I believe default pulls `std`.

```suggestion
num-complex     = { version = "0.4", default-features = false }
```

crates/lifted-stark/Cargo.toml
```diff
@@ -0,0 +1,105 @@
+[package]
+description            = "Lifted STARK prover and verifier (LMCS-based)."
+edition                = "2024"
+homepage.workspace     = true
+license                = "MIT OR Apache-2.0"
```
Nashtare: nit: `edition` / `license` can be derived from workspace

Nashtare: could also inherit authors/categories/keywords

crates/crypto/Cargo.toml
```diff
@@ -0,0 +1,157 @@
+[package]
+authors.workspace      = true
+categories.workspace   = true
+description            = "Miden Cryptographic primitives"
+documentation          = "https://docs.rs/miden-crypto/0.18.0"
```
Nashtare: noticed the version here is stale, is there a way to bump it automatically upon releases?

Al-Kindi-0 (approved):
LGTM, a few questions collected during the review of the commits highlighted in the PR description:


1. Commit `91872f4c78` also rewrites `AeadPoseidon2::key_from_bytes` into a SHA-256-based derivation, which isn't mentioned in the commit message or PR description. Since this means the same key bytes now produce a different key than upstream `miden-crypto` (and the compatibility fallback in `c236e6c1` was later reverted), was this break intentional, and is there any persisted Poseidon2-IES data that would be affected?

2. In `ec2e087dbc`, the empty domain-separated input absorbs an all-zero block, so `hash_elements_in_domain(&[], d) == hash_elements_in_domain(&[ZERO; 8], d)` for any nonzero domain. Should this absorb a padding marker instead, like the bytes-path fix in `70dd4b1099` does?

3. `70dd4b1099` changes the digest of `hash(&[])` from the old all-zero value — is there anywhere the old empty-input digest could have been persisted?

4. A few commits contain more than their messages state (e.g. `b823ff49` also rewrites `ReadAdapter::read_exact`/`buffer_at_least`, and `e266ece9` removes `ExactSizeIterator` from `ReadManyIter`, a public API difference from upstream). Could these be reflected in the commit messages or the PR description?

greenhat (commented):
I checked `miden-field` and its infra, and everything looks fine.

crates/field/Cargo.toml
```diff
@@ -10,7 +10,7 @@ name                   = "miden-field"
 readme                 = "README.md"
 repository.workspace   = true
 rust-version.workspace = true
-version                = "0.29.0"
+version = "0.28.0"
```
adr1anh, line 13: Nit: formatting

crates/lifted-air/Cargo.toml
```diff
@@ -7,7 +7,7 @@ name                   = "miden-lifted-air"
 readme.workspace       = true
 repository.workspace   = true
 rust-version.workspace = true
-version                = "0.29.0"
+version = "0.28.0"
```
adr1anh, line 10: Nit formatting

adr1anh (approved):
Thank you very much, this will make things a lot easier!

huitseeker:
@Al-Kindi-0 Thanks for the review, very useful!

1. The break was not intentional. Commit 91872f4c78 said it fixed AEAD malleability, but the diff also swapped `key_from_bytes` from a canonical-Felt decode to a SHA-256 KDF (`derive_secret_key_from_bytes`) that does rejection sampling on `value < Felt::ORDER`, for reasons that will appear clear below

    I removed the KDF in commit 4392e67269. `key_from_bytes` is back to the upstream one-liner.

    ```rust
    fn key_from_bytes(bytes: &[u8]) -> Result<Self::Key, EncryptionError> {
        if bytes.len() != SK_SIZE_BYTES {
            return Err(EncryptionError::FailedOperation);
        }
        SecretKey::read_from_bytes_with_budget(bytes, SK_SIZE_BYTES)
            .map_err(|_| EncryptionError::FailedOperation)
    }
    ```

    The trailing-bytes rejection from 91872f4c78 is untouched, and `test_decrypt_rejects_trailing_ciphertext_bytes` still covers it.

    Reverting the KDF surfaced the bug which prompted its introduction. `CryptoBox`, the IES layer, feeds HKDF-SHA256 output straight into `key_from_bytes`. HKDF output is 32 random bytes, and canonical decoding rejects any limb greater than or equal to `Felt::ORDER`. Upstream `0xMiden/crypto` has the same latent bug, confirmed against `origin/next` at `7dbb919d1`. 

    The IES bug is fixed at the correct layer in commit `87c5ed5d25`. It adds `AeadScheme::key_from_uniform_bytes` as a separate derivation path. The default impl delegates to `key_from_bytes`. `AeadPoseidon2` overrides it to reduce via `Felt::new_unchecked`. All four `CryptoBox` call sites now route through it. Commit 0bbe048b1d pins the exact reduced elements in the regression test.

    Commits: 4392e67269, 87c5ed5d25, 0bbe048b1d, 612d411167.

2. Commit 0ed971f5bb fixes it. `hash_elements_internal` now absorbs a `ONE` padding marker when the input is empty and the domain is nonzero, mirroring 70dd4b1099. The existing `hash_elements(&[])` path is unchanged because its domain is `ZERO` and the gate does not fire. Each of the three sponge test modules (`poseidon2/test.rs`, `rescue/rpo/tests.rs`, `rescue/rpx/tests.rs`) gets a regression test asserting the collision is gone.

    Commits: 0ed971f5bb.

3. Commit 70dd4b1099 stays as is. The old behavior was a degenerate case that nothing would deliberately persist as a meaningful hash, since it is indistinguishable from a default-constructed `Word`. Any downstream code that relied on it was relying on that collision, which the fix removes. The changelog notes the digest change under Fixes.

4. For b823ff4940, the fix is documentation only. The `ReadAdapter` rewrite is fine. The PR description's "Code-level fixups" section now calls out that b823ff4940 rewrites `ReadAdapter::read_exact` and `buffer_at_least` in `crates/serde-utils/src/byte_reader.rs` as the larger half of the diff.

    For e266ece98e and `ReadManyIter`, the `ExactSizeIterator` impl is restored in commit d1cd1566e6. This was a real public-API regression, not a message issue. 

    Commit d1cd1566e6 also puts `size_hint`'s lower bound back to `self.remaining`. The e266ece98e refactor had changed it to `0` to encode a "do not advertise fallible items as ready" contract. The iterator knows exactly how many items it will yield, so the exact-size contract holds.

    Commits: d1cd1566e6.

Nashtare (approved):

PR #3367, opened 2026-07-13
ci: reduce blake3 nonregression noise

huitseeker:
The Blake3 nonregression workflow was failing on tracing span changes even when the Criterion benchmark stayed under the 5 percent limit. It also posted a commit comment for every pushed run, including passing runs.

This changes the failure rule so only Criterion results can fail the job, and only when the change is over both 5 percent and 1 ms. Tracing spans still appear in the report. The workflow now comments only on failed benchmark runs. The manual trigger is back under the GitHub input limit.

### DIFF
--- a/.github/workflows/blake3-nonregression.yml
+++ b/.github/workflows/blake3-nonregression.yml
@@ -20,6 +20,11 @@ on:
         required: false
         default: "5"
         type: string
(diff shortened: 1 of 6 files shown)


bobbinth (approved):
Not a detailed review - but looks good! Thank you!

PR #3370, opened 2026-07-13
fix(processor): return error on empty OverflowTable instead of panicking

### LINKED ISSUE #3295: trace: OverflowTable methods panic on empty table instead of returning an error
`OverflowTable` in `processor/src/trace/stack/overflow.rs` has multiple methods that panic when called on an empty table:

**`restore_context()` (line 170-177):**
```rust
pub fn restore_context(&mut self) {
    let overflow_stack_for_ctx = self.overflow.swap_remove(self.overflow.len() - 1);
```
If `self.overflow` is empty (start_context was never called), `self.overflow.len()` is 0, so `0usize - 1` wraps to `usize::MAX` and `swap_remove` panics with out-of-bounds access.

**`get_current_overflow_stack_mut()` (line 195-198):**
```rust
fn get_current_overflow_stack_mut(&mut self) -> &mut OverflowStack {
    let len = self.overflow.len();
    &mut self.overflow[RowIndex::from(len - 1)]
}
```
Same underflow pattern — empty overflow table causes `0 - 1`.

**`get_current_overflow_stack()` (line 187-191):**
```rust
self.overflow.as_slice().last()
    .expect("The current context should always have an overflow stack initialized")
```
`.last()` returns `None` on empty table, `expect()` panics.

These are in the trace-recording path used during proving, not the fast execution path. A bug in the trace generator (start_context/restore_context mismatch) would cause a panic during proof generation instead of a recoverable error.

Fix: Return proper errors instead of panicking on empty state, or add precondition checks that survive release builds.

I have a fix ready and can open a PR once assigned.

Sertug17:
## Summary
Fixes #3295  `OverflowTable` methods now return proper errors instead of panicking when called on an empty table.

## Rationale
Preventing panics in the trace-recording path ensures that proof generation failures are recoverable rather than crashing the entire process. This improves debuggability and robustness of the proving pipeline.

## Problem
`OverflowTable` in `processor/src/trace/stack/overflow.rs` had three methods that panic on empty state:

1. `restore_context()` — `0usize - 1` underflow causes `swap_remove` to panic
2. `get_current_overflow_stack_mut()` — same underflow pattern
3. `get_current_overflow_stack()` — `.expect()` panics when `.last()` returns `None`

These panics occur in the trace-recording path during proof generation, turning a recoverable state mismatch into a crash.

## Changes
- `restore_context()`: Returns `OperationError::OverflowTableEmpty` instead of panicking
- `get_current_overflow_stack_mut()`: Returns `Result&lt;&mut OverflowStack, OperationError&gt;` instead of underflow
- `get_current_overflow_stack()`: Returns `Result&lt;&OverflowStack, OperationError&gt;` instead of `.expect()` panic
- Updated `test_restore_context_empty_table` to verify error return path

## Test Plan
1. `cargo test -p miden-processor -- overflow` → verify all tests pass
2. `cargo check -p miden-processor` → verify no compiler errors
3. Review `processor/src/trace/stack/overflow.rs` → verify no panics on empty table

## Testing
- `cargo test -p miden-processor -- overflow` all 7 tests pass
- `cargo check -p miden-processor` clean
- `cargo fmt` applied

## Checklist
- [x] Tests added/updated
- [x] `cargo fmt` applied
- [x] `cargo check` passes
- [x] Fixes #3295


Sertug17:
@huitseeker 

All review comments addressed:

1. `OperationError` import added to `overflow.rs`
2. `restore_context()` now returns `Result&lt;(), OperationError&gt;` instead of panicking
3. `get_current_overflow_stack()` and `get_current_overflow_stack_mut()` return `Result` with `OperationError::Internal`
4. Public API methods (`last_update_clk_in_current_ctx`, `clk_after_pop_in_current_ctx`, `num_elements_in_current_ctx`, `push`, `pop`) preserved with `.unwrap()` internally
5. `fast/mod.rs` — `package.debug_info()` error handled with `map_err` + `ExecutionError::Internal`
6. `execution_tracer.rs` — `restore_context()` result handled with `let _ = ...`
7. `cargo fmt` applied
8. CHANGELOG entry added under `## Unreleased`

`cargo check -p miden-processor` passes clean. `cargo test -p miden-processor -- overflow` passes. PTAL!

processor/src/trace/stack/overflow.rs
```diff
@@ -163,17 +163,24 @@ impl OverflowTable {
...
-    /// # Panics
-    /// - if there is no overflow stack for the current context.
-    /// - if the overflow stack for the current context is not empty.
+    /// # Errors
+    /// Returns `OperationError::OverflowTableEmpty` if there is no overflow stack for the current
+    /// context (i.e., `start_context` was never called).
+    /// Returns `OperationError::Internal` if the overflow stack for the current context is not empty.
     ///   - i.e. this should be checked before calling this function.
-    pub fn restore_context(&mut self) {
+    pub fn restore_context(&mut self) -> Result<(), OperationError> {
```
huitseeker, line 178: I still can't build this head with `cargo check -p miden-processor`. This file returns `OperationError` without importing it, `OperationError::OverflowTableEmpty` is not defined, and `processor/src/errors.rs` also removed the `PackageDebugInfoError` conversion while `processor/src/fast/mod.rs` still uses `package.debug_info()?`.

After those compile errors are fixed, this `Result` also needs to be handled by the callers. The public overflow helpers still call `.last()`, `.num_elements()`, `.push()`, and `.pop()` on the returned value, and `execution_tracer.rs` currently calls `self.overflow_table.restore_context();` without checking the error. That means the trace path can still ignore the error this PR is trying to make recoverable.

processor/src/trace/stack/overflow.rs
```diff
@@ -163,17 +163,24 @@ impl OverflowTable {
...
-        let overflow_stack_for_ctx = self.overflow.swap_remove(self.overflow.len() - 1);
-        assert!(
-            overflow_stack_for_ctx.is_empty(),
-            "the overflow stack for the current context should be empty when restoring a context"
-        );
+        let len = self.overflow.len();
+        if len == 0 {
+            return Err(OperationError::OverflowTableEmpty);
+        }
+        let overflow_stack_for_ctx = self.overflow.swap_remove(len - 1);
```
huitseeker: This removes the current overflow stack before checking whether it is empty. If the stack is not empty, the method returns `Err`, but the table has already lost that context, so the caller cannot recover from the error cleanly.

Could this check look at the last stack first and only remove it after it is known to be empty?

huitseeker (changes requested):

processor/src/trace/stack/overflow.rs
Sertug17: Fixed `use crate::operation::OperationError` is now the correct import path, using the existing `OperationError::Internal` variant (no new variants, no changes to `errors.rs` or `fast/mod.rs`). All callers are handled: public API methods use `.expect()` (invariant guaranteed by constructor), and `execution_tracer.rs` handles the `Result` with `.expect()` since the contract guarantees the stack is empty at that point. `cargo check -p miden-processor` compiles clean.

processor/src/trace/stack/overflow.rs
Sertug17: Good catch fixed. `restore_context()` now checks `self.overflow.as_slice().last().is_empty()` *before* calling `swap_remove`, so if the stack is not empty the method returns `Err` without modifying the table. No data loss on the error path.

Sertug17:
@huitseeker All review comments addressed:

1. **Import fix**: `use crate::operation::OperationError` — compiles clean, no changes to `errors.rs` or `fast/mod.rs`.
2. **No new variant**: Using existing `OperationError::Internal(&'static str)`.
3. **Logic fix**: `restore_context()` checks `last().is_empty()` *before* `swap_remove` no data loss on error path.
4. **Caller handling**: `execution_tracer.rs` handles `Result` with `.expect()`. Public API methods use `.expect()` (constructor guarantees ≥1 stack).
5. **CHANGELOG**: Entry added under `## Unreleased`.

`cargo check -p miden-processor` and `cargo test -p miden-processor -- overflow` both pass clean. PTAL!

processor/src/trace/stack/overflow.rs
```diff
@@ -163,38 +172,56 @@ impl OverflowTable {
...
-            "the overflow stack for the current context should be empty when restoring a context"
-        );
+    /// Returns an error if the overflow table has no stacks (i.e. there is no current context)
+    /// or if the overflow stack for the current context is not empty (i.e. the caller should
+    /// have drained it before calling this function).
+    pub fn restore_context(&mut self) -> Result<(), OperationError> {
+        let len = self.overflow.len();
+
+        // Ensure there is at least one overflow stack to restore.
+        if len == 0 {
```
huitseeker: I think this still lets the bad empty-table state happen. `OverflowTable::new()` starts with one root stack, so `restore_context()` called with no child context has `len == 1`, passes this check, and then `swap_remove(len - 1)` removes the root stack. The call returns `Ok(())`, but the table is now empty, so the next `last_update_clk_in_current_ctx()`, `push()`, or `pop()` still hits an `expect()` panic.

Should this reject `len <= 1` before removing anything?

huitseeker (commented):
This needs a rebase on `next`, which is the development branch for this repo, not `main`. Also, note the Changelog format has changed with #3339.

Sertug17:
@huitseeker Rebased on `next` and updated CHANGELOG entry to `v0.28.0 (unreleased)` / `#### Changes` format per #3339. PTAL!

Sertug17:
Good catch fixed. `restore_context()` now rejects `len <= 1` so the root stack can never be removed. The helpers (`get_current_overflow_stack` / `_mut`) still use `len == 0` since accessing the root stack is valid. Also fixed CHANGELOG duplicates and rebased cleanly on `next`. PTAL!

processor/src/trace/execution_tracer.rs
```diff
@@ -927,7 +927,9 @@ impl Tracer for ExecutionTracer {
         if self.pending_restore_context {
             // Restore context for call/syscall/dyncall: pop the current context's
             // (empty) overflow stack and restore the previous context's overflow state.
-            self.overflow_table.restore_context();
+            self.overflow_table
+                .restore_context()
+                .expect("restore_context should not fail: stack should be empty at this point");
```
huitseeker: This still leaves a panic path in trace generation. `OverflowTable::restore_context()` now returns `OperationError`, but this caller immediately unwraps it with `expect`, so an empty table or non-empty child overflow stack still panics instead of becoming an execution error. 

Could this error be propagated through the trace-finalization path?

Sertug17:
Fixed `finalize_clock_cycle` now returns `Result<(), ExecutionError>` across the `Tracer` trait, so the error propagates instead of panicking. Changes:

- `Tracer::finalize_clock_cycle` signature: `()` → `Result<(), ExecutionError>`
- `ExecutionTracer`: propagates `restore_context()` error via `map_err(|_| ExecutionError::Internal(...))?`
- `FastProcessor` (no-op tracer): returns `Ok(())`
- `ParallelTracer`: returns `result` directly
- Caller in `execution/mod.rs`: `if let Err(e) = ... { return ControlFlow::Break(BreakReason::Err(e)); }`

No more panic paths in trace generation. PTAL! @huitseeker

Sertug17:
@huitseeker Updated for now, `execution_tracer.rs` keeps the `.expect()` on `restore_context()` since the `Tracer` trait currently returns `()` from `finalize_clock_cycle`. Changing the trait signature to `Result<(), ExecutionError>` touches 5 files across 3 implementations + caller chain I'd prefer to do that in a follow up PR to keep this one focused on the `OverflowTable` fix. Happy to open that follow up once this lands. PTAL!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v0.28.0 (unreleased)
 
 #### Changes
+- Replaced panics in `OverflowTable::restore_context()`, `get_current_overflow_stack()`, and `get_current_overflow_stack_mut()` with proper `OperationError` returns ([#3370](https://github.com/0xMiden/miden-vm/pull/3370)).
```
huitseeker, line 7: This diff adds 33 changelog lines: this entry appears 20 times across old releases, and 13 unrelated precompile and AST entries are copied in too. Could we keep only one `#3370` entry under the current `v0.28.0` section and drop the other 32 additions?

processor/src/trace/stack/overflow.rs
```diff
@@ -163,17 +172,27 @@ impl OverflowTable {
...
-        // 1. pop the last overflow stack for the current context, and make sure it is empty.
-        let overflow_stack_for_ctx = self.overflow.swap_remove(self.overflow.len() - 1);
-        assert!(
-            overflow_stack_for_ctx.is_empty(),
-            "the overflow stack for the current context should be empty when restoring a context"
-        );
+    /// Returns an error if the overflow table has no stacks (i.e. there is no current context)
+    /// or if the overflow stack for the current context is not empty (i.e. the caller should
+    /// have drained it before calling this function).
+    pub fn restore_context(&mut self) -> Result<(), OperationError> {
```
huitseeker, line 178: I can't find the `test_restore_context_empty_table` named in the PR body. `cargo test -p miden-processor -- overflow` runs seven existing tests, but none call `OverflowTable::restore_context()`. Could we add focused tests for the root-only error and the non-empty child error, including that each error leaves the table unchanged?

huitseeker (changes requested):

CHANGELOG.md
Sertug17: Fixed reset to origin/next and added a single entry under `v0.28.0 (unreleased)` / `#### Changes`. No more duplicates.

processor/src/trace/stack/overflow.rs
Sertug17: Added three focused unit tests in `overflow.rs`:
- `restore_context_rejects_root_only_table`: confirms `len <= 1` returns error and table is unchanged
- `restore_context_rejects_non_empty_child_stack`: confirms non-empty child returns error and table is unchanged
- `restore_context_succeeds_with_empty_child_stack`: confirms happy path works

All pass: `cargo test -p miden-processor --lib -- restore_context` → 6 passed.

processor/src/trace/stack/overflow.rs
```diff
@@ -203,3 +226,43 @@ impl Default for OverflowTable {
...
+
+    #[test]
+    fn restore_context_succeeds_with_empty_child_stack() {
+        let mut table = OverflowTable::new();
+        table.start_context();
+        // Child stack is empty, restore should succeed.
+        let result = table.restore_context();
+        assert!(result.is_ok(), "restore_context should succeed with empty child stack");
+        // Back to root stack.
+        assert_eq!(table.num_elements_in_current_ctx(), 0);
```
huitseeker, line 266: Could this call `restore_context()` a second time and assert that it fails? The zero-element check is true for both an empty child and the root, so this test would still pass if the first call returned `Ok(())` without removing the child context.

processor/src/trace/execution_tracer.rs
huitseeker: @Sertug17 This is still a live concern

processor/src/trace/stack/overflow.rs
huitseeker: the successful-restore test does not prove the child context was removed.

huitseeker (changes requested):
I think restoration failures remain panics instead of recoverable execution errors.

Sertug17:
@huitseeker All review comments addressed:

- `Tracer::finalize_clock_cycle` signature: `()` → `Result<(), ExecutionError>`
- `ExecutionTracer`: `restore_context()` error propagated via `map_err` no more `.expect()`
- `FastProcessor` (no-op tracer): returns `Ok(())`
- `CoreTraceGenerationTracer`: returns `Ok(())`
- Caller in `execution/mod.rs`: `if let Err(e) = ... { return ControlFlow::Break(BreakReason::Err(e)); }`
- `restore_context()` rejects `len <= 1` (root stack protection)
- Test calls `restore_context()` twice: first succeeds, second asserts `Err`

No panic paths remain in trace generation. Rebased on `next`. PTAL!

huitseeker (commented):
Super, thanks a lot!

Sertug17:
@huitseeker CI is green (the cargo-deny failure is a pre-existing dependency advisory, unrelated to this PR). Could you reapprove when you get a chance? Thanks!

huitseeker (approved):

Sertug17:
Thanks for the thorough review process @huitseeker the trait level propagation change made the fix much stronger than the original. Appreciate the patience through the iterations!

PR #3372, opened 2026-07-14
feat(mast-package): add get_export_node and procedures_with_attribute APIs

### LINKED ISSUE #3320: Deepen package procedure access APIs
0xMiden/protocol#3146 exposed a gap between `miden-vm` package APIs and protocol account code.

Motivating PR comments:

- https://github.com/0xMiden/protocol/pull/3146#discussion_r3516208770
- https://github.com/0xMiden/protocol/pull/3146#discussion_r3516217416
- https://github.com/0xMiden/protocol/pull/3146#discussion_r3516223720

`Package::get_procedure_root_by_path` and `Package::get_procedure_node_by_path` already cover direct path lookup in `crates/mast-package/src/package/mod.rs`. That is enough for one protocol method.

Protocol still has to read package internals for account interface procedures:

- [`crates/miden-protocol/src/account/component/code.rs`](https://github.com/0xMiden/protocol/blob/9ba9e9ed5faaeb35354e77e4f123ea13812353fa/crates/miden-protocol/src/account/component/code.rs#L32-L61) reads package exports and filters `ProcedureExport` attributes.
- [`crates/miden-protocol/src/account/component/mod.rs`](https://github.com/0xMiden/protocol/blob/9ba9e9ed5faaeb35354e77e4f123ea13812353fa/crates/miden-protocol/src/account/component/mod.rs#L185-L190) reads the export node to compute account procedure roots.

The existing direct lookup API is in [`crates/mast-package/src/package/mod.rs`](https://github.com/0xMiden/miden-vm/blob/9631e12d8f9939d9ed8bf89f2da8733be986e56a/crates/mast-package/src/package/mod.rs#L400-L413).

Please add package APIs that hide that shape:

- iterate procedure exports by attribute.
- return the resolved procedure root for an export, using the export node when present.
- avoid direct downstream reads of `Package::manifest`, `ProcedureExport::node`, and `Package::mast_forest()` for this common case.

After that, protocol account code can ask the package for interface procedure roots instead of rebuilding that logic.

enlightened88:
Adds two Package methods that hide internal export-resolution details:
- get_export_node(): non-panicking counterpart to get_export_node_id(), reused to de-duplicate the node-resolution logic that was previously copy-pasted in get_export_node_id(), is_reexport(), and get_procedure_node_by_path().
- procedures_with_attribute(): iterates procedure exports filtered by attribute name, avoiding the need for callers to read Package::manifest and filter ProcedureExport directly.

Closes #3320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - [BREAKING] Add missing constraint in Bitwise chiplet ([#3386](https://github.com/0xMiden/miden-vm/pull/3386)).
 - [BREAKING] Fixed a soundness gap in the chiplets AIR where a chiplet section's first-row initialization was skipped when the preceding section was empty. A program that uses memory but performs no `u32and`/`u32xor` operations produces an empty bitwise section, which caused the memory chiplet to skip its "values not being written must be zero" reset; a malicious prover could exploit this to forge a read of never-written memory. Each section's first row is now identified from the chiplet selectors at the boundary rather than from the previous chiplet's last row, so the initialization holds no matter which preceding sections are empty. The ACE section-start reset was hardened the same way as a precaution ([#3387](https://github.com/0xMiden/miden-vm/pull/3387)).
 - [BREAKING] Optimize periodic columns evaluation for fewer ACE gates ([#3347](https://github.com/0xMiden/miden-vm/pull/3347)).
+- Added `Package::get_export_node()` and `Package::procedures_with_attribute()` APIs ([#3320](https://github.com/0xMiden/miden-vm/issues/3320)).
 
 #### Fixes
 
--- a/crates/mast-package/src/package/mod.rs
+++ b/crates/mast-package/src/package/mod.rs
@@ -424,7 +424,14 @@ impl Package {
 
     /// Resolves the MAST node corresponding to `export`, or `None` if it cannot be found.
     ///
-    /// This is the non-panicking counterpart to [`Package::get_export_node_id`].
+    /// The returned node is the recorded `export.node` only when it points at a procedure
+    /// root in this package's forest whose digest matches `export.digest`; otherwise this
+    /// falls back to a digest-based lookup.
+    ///
+    /// This is the non-panicking counterpart to [`Package::get_export_node_id`], though the
+    /// two take different arguments: `get_export_node_id` resolves an export by path and
+    /// panics if not found, while this method resolves an already-known `&ProcedureExport`
+    /// and treats its recorded `node` as an untrusted hint rather than authoritative.
     pub fn get_export_node(&self, export: &ProcedureExport) -> Option<MastNodeId> {
         export
             .node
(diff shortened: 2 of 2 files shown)


crates/mast-package/src/package/mod.rs
```diff
@@ -419,7 +419,26 @@ impl Package {
...
             .and_then(PackageExport::as_procedure)
-            .and_then(|export| export.node.or_else(|| self.mast.find_procedure_root(export.digest)))
+            .and_then(|export| self.get_export_node(export))
+    }
+
+    /// Resolves the MAST node corresponding to `export`, or `None` if it cannot be found.
+    ///
+    /// This is the non-panicking counterpart to [`Package::get_export_node_id`].
+    pub fn get_export_node(&self, export: &ProcedureExport) -> Option<MastNodeId> {
+        export.node.or_else(|| self.mast.find_procedure_root(export.digest))
```
huitseeker: Since this method is public, should it check that a recorded node actually belongs to this package and still matches `export.digest` before returning it? A caller can pass a cloned or constructed `ProcedureExport`, and today `Some(node)` wins even if that node is from another forest or points at a different digest. That makes the `None if it cannot be found` contract false and can lead to a later panic when the returned id is used with this package's forest.

The package constructor already uses `is_procedure_root_with_exact_digest`; doing the same here before falling back to `find_procedure_root(export.digest)` would keep this API safe for external callers.

huitseeker (changes requested):

crates/mast-package/src/package/mod.rs
enlightened88: Fixed. get_export_node now validates the node against the digest before trusting it, and falls back to the lookup otherwise

crates/mast-package/src/package/mod.rs
```diff
@@ -419,7 +419,29 @@ impl Package {
...
         self.get_export_by_lookup_path(path.as_ref())
             .and_then(PackageExport::as_procedure)
-            .and_then(|export| export.node.or_else(|| self.mast.find_procedure_root(export.digest)))
+            .and_then(|export| self.get_export_node(export))
+    }
+
+    /// Resolves the MAST node corresponding to `export`, or `None` if it cannot be found.
+    ///
+    /// This is the non-panicking counterpart to [`Package::get_export_node_id`].
+    pub fn get_export_node(&self, export: &ProcedureExport) -> Option<MastNodeId> {
```
huitseeker, line 435: Nit: The doc comment says this is the "non-panicking counterpart to `get_export_node_id`", but `get_export_node_id` takes a path and this one takes a `&ProcedureExport`. The contract is a bit different and worth spelling out, since the recorded `export.node` is now treated as an untrusted hint: if it is missing or does not match a procedure root with `export.digest`, the call silently falls back to a digest lookup and may return a different node id than what the caller passed in.

Could the doc say something like "the returned node is only the recorded one when it points at a procedure root whose digest matches `export.digest`; otherwise this falls back to a digest lookup"?

crates/mast-package/src/package/mod.rs
```diff
@@ -2037,6 +2059,85 @@ mod tests {
         assert!(!package.is_reexport("::app::entry"));
     }
 
+    #[test]
+    fn get_export_node_prefers_recorded_node_id() {
```
huitseeker, line 2070: This test passes for the right reason (the digest matches), but it also passes if `export.node` were `None`, because the digest fallback in `get_export_node` would return the same id. So it does not actually prove the recorded-node path is exercised.

The case the second commit (`3144689a57`) specifically guards is: `export.node` is `Some(id)`, `id` is a real procedure root in the forest, but `mast[id].digest() != export.digest`. 

None of the three new tests cover that mismatch path. Would be worth adding one, since it is the regression a future "simplification" of the filter back to `export.node.or_else(...)` would silently reintroduce. 

`build_same_digest_package_exports` later in this module already constructs multiple roots, so it could be reused to make two distinct roots and pick the second one as `export.node` while passing the first one's digest.

huitseeker (commented):
Thanks, this looks good now! there's a few nits to fix, a doc fix and a suggestion for another test (see inline).

crates/mast-package/src/package/mod.rs
enlightened88: Updated the docs to clarify the fallback behavior

crates/mast-package/src/package/mod.rs
enlightened88: added a regression test for that case

huitseeker (approved):
Thanks a lot!

CHANGELOG.md
```diff
@@ -26,6 +27,14 @@
...
+- [BREAKING] Removed the `miden::core::crypto::dsa::eddsa_ed25519` MASM module, Rust handler, docs, and tests. EdDSA support is temporarily removed from core-lib and will be reintroduced once it is supported by the precompiles prover.
+- [BREAKING] Removed the `miden::core::crypto::hashes::sha512` MASM module, Rust handler, docs, and tests. SHA-512 support is temporarily removed from core-lib and will be reintroduced once it is supported by the precompiles prover.
+- [BREAKING] Changed the `miden::core::crypto::dsa::ecdsa_k256_keccak` advice/signature ABI to `QX[8] || QY[8] || SIG_R[8] || SIG_S[8]` as little-endian u32 field elements. Existing 65-byte signature advice must be re-encoded as `(r, s)` limbs without a recovery byte.
+- [BREAKING] Removed the public `miden::core::crypto::dsa::ecdsa_k256_keccak::verify_prehash` and raw `miden::precompiles::crypto::dsa::ecdsa_secp256k1::assert_verify_prehash` ECDSA prehash verifier entrypoints. ECDSA K256 Keccak verification is now exposed only through the high-level `verify` procedure, whose implementation inlines the verifier, loads signature scalars directly from advice, and avoids the raw prehash memory ABI.
```
huitseeker: Those look like a merge error

enlightened88: Fixed

bitwalker (approved):
LGTM

PR #3374, opened 2026-07-14
ci: reduce synthetic tx nonregression noise

huitseeker:
This is a direct follow-up to #3367 that performs the same modificaitons to the synthetic benchmark (given the variance of the GH machines).


Al-Kindi-0 (approved):
LGTM

PR #3375, opened 2026-07-14
assembly: implement missing source provider functionality

bitwalker:
This PR provides missing metadata/functionality needed for implementations of `ProjectSourceProvider` to be able to not only derive project source inputs, but also finalize assembled packages derived from those inputs, in order to wire up metadata, custom sections, etc.

This set of commits is based on `main`, so we can release them as a patch release for v0.25. The changes are purely additive, except in one subtle area: the type parameter of `ProjectAssembler` that specifies what implementation of `PackageCache` it uses has to be `Sized` now (previously the bound was `S: ?Sized + PackageCache`, now it is simply `S: PackageCache). This is required because `TargetAssemblyContext` needs to carry a `dyn` reference to a `PackageRegistry`/`PackageProvider` (supertraits of `PackageCache`), but a reference to an unsized type cannot be cast to a `dyn` reference. This doesn't actually change any of our current uses of these APIs (all instances of `S` have always been `Sized`), but it is technically breaking. However, the only consumer of these APIs currently is the compiler, and assembler internals, and since v0.25 just released, there has basically been no time for new consumers to be implemented - I think this is case where it is safe to release as a patch.

### DIFF
--- a/crates/assembly/src/linker/mod.rs
+++ b/crates/assembly/src/linker/mod.rs
@@ -40,7 +40,7 @@ mod debug;
 mod errors;
 mod library;
 mod module;
-mod namespaces;
+pub mod namespaces;
 mod resolver;
 mod rewrites;
 mod symbols;
@@ -68,12 +68,13 @@ pub use self::{
     callgraph::{CallGraph, CycleError},
     errors::LinkerError,
     library::{LinkLibrary, Linkage},
+    namespaces::NamespaceGraph,
     resolver::{ResolverCache, SymbolResolutionContext, SymbolResolver},
     symbols::{Import, Symbol, SymbolItem},
 };
 use self::{
     module::{LinkModule, ModuleSource},
-    namespaces::{NamespaceGraph, ResolvedImports},
+    namespaces::ResolvedImports,
     resolver::*,
 };
 
--- a/crates/assembly/src/linker/mod.rs
+++ b/crates/assembly/src/linker/mod.rs
@@ -384,7 +384,7 @@ impl Linker {
     /// Returns a new [Linker] instantiated from the provided kernel and kernel info module.
     ///
     /// Note: it is assumed that kernel and kernel_module are consistent, but this is not checked.
-    pub(super) fn with_kernel(
+    pub fn with_kernel(
         source_manager: Arc<dyn SourceManager>,
         kernel_package: Arc<MastPackage>,
     ) -> Result<Self, Report> {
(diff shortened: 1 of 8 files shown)


crates/assembly/src/linker/mod.rs
```diff
@@ -869,7 +871,7 @@ impl Linker {
     }
 
     /// Resolves a [GlobalProcedureIndex] to the known attributes of that procedure
-    pub(super) fn resolve_attributes(&self, gid: GlobalItemIndex) -> AttributeSet {
+    pub fn resolve_attributes(&self, gid: GlobalItemIndex) -> AttributeSet {
```
huitseeker, line 875: Nit: This now shows up in public docs, and `make doc` is failing because `[GlobalProcedureIndex]` is not in scope here.

crates/assembly/src/project.rs
```diff
@@ -709,6 +715,40 @@ where
             _ => Ok(inputs),
         }
     }
+
+    fn apply_post_assembly_hooks(
```
huitseeker, line 701: This block (context construction, extension lookup, provider lookup) looks like a close copy of the one in `load_target_sources` just above. 

Would it be worth pulling the shared piece into a helper that returns both the context and the resolved provider? That way the extension-parse and "unsupported target file type" error live in one place.

crates/assembly/src/project/providers.rs
```diff
@@ -25,17 +26,20 @@ pub struct TargetAssemblyContext<'a> {
     pub dependency_graph: &'a ProjectDependencyGraph,
     /// The current source manager
     pub source_manager: Arc<dyn SourceManager>,
+    /// The current package store of the assembler
+    pub package_registry: &'a dyn PackageRegistryAndProvider,
```
huitseeker, line 30: `package_registry` is threaded into the context here and through five `DependencyGraph` methods, but as far as I can tell nothing reads it yet, it is only forwarded into `TargetAssemblyContext::new`. 

Is this landing ahead of the first out-of-tree consumer that needs it?

crates/assembly/src/project/providers.rs
```diff
@@ -114,4 +120,20 @@ pub trait ProjectSourceProvider {
...
+
+    /// Called after a project target - whose sources were provided via this trait- has been
+    /// assembled to a package, so that the provider can do any language-specific post-processing
+    /// of the assembled package before it is frozen and submitted to the package cache/registry.
+    ///
+    /// The default implementation is a no-op.
+    ///
+    /// The `context` given is the same as given to [`ProjectSourceProvider::provide_sources`].
+    #[allow(unused_variables)]
+    fn post_process_package(
```
huitseeker, line 132: The default no-op is the only implementation right now, `MasmSourceProvider` does not override it. 

Just checking there an out-of-tree consumer planned soon.

huitseeker (approved):
Thanks! LGTM overall, some nits inline

crates/assembly/src/project/providers.rs
bitwalker: No - the compiler requires access to the package registry in order to provide sources

crates/assembly/src/project/providers.rs
bitwalker: The rationale for source providers was for assembly of sources derived from other languages - namely Rust, but I have another language I've been working on as well that would use them as well.

I'm the middle of migrating the compiler to use these, which is why I'm only encountering some of these gaps now, rather than when this was first implemented.

PR #3377, opened 2026-07-14
assembly: link debug function info to source graph

bitwalker:
This PR modifies `DebugFunctionInfo` records such that they are linked to procedure code by `MastNodeId`/`DebugSourceNodeId`, rather than to a MAST root/digest. This allows us to link `DebugFunctionInfo` to the correct MAST and set of debug info for that function (e.g. source locations, debug variables, etc.), and vice versa (resolving debug info to a matching `DebugFunctionInfo` record).

The main thing I'm a little uncertain on is whether `MastNodeId` and `DebugSourceNodId` are needed, or if the latter is redundant. I've attempted to maintain both in the initial pass here, but let me know if I'm going about this the right way or not.

@huitseeker I'm thinking we can simplify how the debug info is structured in general. I think it's reasonable to say that without a source graph + source map, none of the other debug info is particularly useful, and so remove the `Option`-wrapping of all the `PackageDebugInfo` fields. The assembler at this point is going to either produce all of the sections within `PackageDebugInfo`, or none of them (`error_messages` is a bit unique in that we elide it even when emitting debug info, if there are no error messages to store). We could then make `PackageDebugInfo` a bit more ergonomic.

I'm expecting to expand function debug info emission a bit further with call frame information in a subsequent PR as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [BREAKING] Renamed module and kernel metadata APIs from `ModuleInfo`/`Kernel` to `ModuleDescriptor`/`KernelDescriptor`, including matching module descriptor method names ([#3356](https://github.com/0xMiden/miden-vm/pull/3356)).
 - Split package serialization assembly tests into their own module ([#3083](https://github.com/0xMiden/miden-vm/pull/3083)).
+- `DebugFunctionInfo` records are now linked to MAST node/source node, rather than MAST roots, as the latter lack sufficient precision for debug info purposes. ()
 
 #### Fixes
 
--- a/crates/assembly/src/assembler.rs
+++ b/crates/assembly/src/assembler.rs
@@ -424,8 +424,9 @@ impl Assembler {
             .map(|module| module.parse(self.warnings_as_errors, self.source_manager.clone()))
             .collect::<Result<Vec<_>, Report>>()?;
 
+        let emit_debug_info = self.emit_debug_info;
         self.assemble_library_modules(name.into(), root, support, TargetType::Library)?
-            .into_artifact()
+            .into_artifact(emit_debug_info)
     }
 
     /// Assemble a library [`Package`] from the set of modules reachable from `root`.
@@ -453,8 +454,9 @@ impl Assembler {
         // Derive the package name from the namespace of the root module
         let name = root.path().as_str().replace("::", "-");
 
+        let emit_debug_info = self.emit_debug_info;
         self.assemble_library_modules(name.into(), root, support, TargetType::Library)?
-            .into_artifact()
+            .into_artifact(emit_debug_info)
     }
 
     /// Assembles the provided module into a kernel package.
@@ -468,8 +470,9 @@ impl Assembler {
         root: Box<ast::Module>,
         support: impl IntoIterator<Item = Box<ast::Module>>,
     ) -> Result<Box<Package>, Report> {
+        let emit_debug_info = self.emit_debug_info;
         self.assemble_library_modules(name.into(), root, support, TargetType::Kernel)?
-            .into_artifact()
+            .into_artifact(emit_debug_info)
     }
 
     /// Assemble a kernel [`Package`] from a standard Miden Assembly kernel project layout.
@@ -501,8 +504,9 @@ impl Assembler {
             self.warnings_as_errors,
         )?;
 
+        let emit_debug_info = self.emit_debug_info;
         self.assemble_library_modules(name.into(), root, support, TargetType::Kernel)?
-            .into_artifact()
+            .into_artifact(emit_debug_info)
     }
 
     /// Shared code used by both [`Self::assemble_library`] and [`Self::assemble_kernel`].
@@ -599,7 +603,16 @@ impl Assembler {
             .collect::<Result<BTreeMap<_, _>, _>>()?;
 
         let modules = self.package_modules(module_indices);
-        self.finish_library_product(name, mast_forest, source_graph, exports, modules, kind)
+        self.finish_library_product(
+            name,
+            mast_forest,
+            source_graph,
+            source_id_by_ref,
+            node_id_by_ref,
+            exports,
+            modules,
+            kind,
+        )
     }
 
     fn package_modules(&self, module_indices: &[ModuleIndex]) -> Vec<PackageModule> {
@@ -662,6 +675,7 @@ impl Assembler {
                 let resolved = match mast_forest_builder.get_procedure(gid) {
                     Some(proc) => ResolvedProcedure {
                         node: proc.body_node_ref(),
+                        source_node: proc.source_node_ref(),
                         signature: proc.signature(),
                     },
                     // We didn't find the procedure in our current MAST forest. We still need to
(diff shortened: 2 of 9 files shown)


crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -841,6 +840,14 @@ impl Deserializable for DebugFileInfo {
 
 impl Serializable for DebugFunctionInfo {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        target.write_u32(self.node.into());
```
huitseeker, line 843: This changes the `debug_functions` payload but keeps the section version at `2`. Old v2 bytes start with `name_idx` and end with optional `mast_root`; new v2 bytes start with `node` and `source_node`.

Should we bump `DEBUG_FUNCTIONS_VERSION`?

crates/mast-package/src/debug_info/types.rs
```diff
@@ -865,14 +875,25 @@ fn merge_debug_functions(
...
+
+        let node = root_map.map_node(forest_index, &function.node).ok_or(
+            PackageDebugInfoMergeError::MissingExecNodeMapping {
+                forest_index,
+                exec_node: function.node,
+            },
+        )?;
+        let source_node = source_id_map
+            .zip(function.source_node)
+            .map(|(source_id_map, source_node)| source_id_map[&source_node]);
```
huitseeker, line 887: This lookup can panic when a debug function points at a missing source node. The surrounding merge code returns `MissingSourceNodeMapping` for missing graph roots, children, source-map rows, and inline calls, but this new function link bypasses that path.

That can be reached from a `PackageDebugInfo` that was deserialized from a "trusted" package, because the new `source_node` field on `DebugFunctionInfo` isn't bounds-checked in `validate_debug_function` (it only checks `name_idx`, `linkage_name_idx`, `file_idx`, `type_idx`). So an out-of-range `source_node` from any trusted-source tool gets in here and panics the merge.

Could this use `.get(&source_node).copied().ok_or(PackageDebugInfoMergeError::MissingSourceNodeMapping { ... })?` to match the rest of the function? The variant already exists in the error enum.

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1650,6 +1671,37 @@ impl DebugPrimitiveType {
...
+            // the same for u32/f32
+            //
+            // A future release will add F32 to the HIR type system to close this gap
+            Self::F32 => Type::U32,
+            Self::F64 => Type::F64,
+            Self::Felt => Type::Felt,
+            Self::Word => Type::Array(Arc::new(ArrayType::new(Type::Felt, 4))),
+            Self::U256 => Type::U256,
+        })
+    }
```
huitseeker, line 1704: Just flagging: `to_hir_type` has no callers anywhere in-tree.

crates/assembly/src/assembler.rs
```diff
@@ -1103,20 +1128,32 @@ impl Assembler {
...
             #[cfg_attr(not(feature = "std"), expect(unused_mut))]
             let mut debug_info = self.debug_info.clone();
             #[cfg(feature = "std")]
             if let Some(trimmer) = self.source_path_trimmer() {
                 debug_info.trim_paths(&trimmer);
             }
             debug_info
-        });
+        } else {
+            self.debug_info.clone()
```
huitseeker, line 1140: On the `!emit_debug_info` path this now eagerly clones `self.debug_info` (and on the next block, runs `apply_source_debug_options` over the source graph), only for `into_artifact` to early-return at `product.rs:97` and drop it all. Pre-PR both branches used `.then(|| ...)`, so no work happened when debug info was off.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - [BREAKING] Renamed module and kernel metadata APIs from `ModuleInfo`/`Kernel` to `ModuleDescriptor`/`KernelDescriptor`, including matching module descriptor method names ([#3356](https://github.com/0xMiden/miden-vm/pull/3356)).
 - Split package serialization assembly tests into their own module ([#3083](https://github.com/0xMiden/miden-vm/pull/3083)).
+- `DebugFunctionInfo` records are now linked to MAST node/source node, rather than MAST roots, as the latter lack sufficient precision for debug info purposes. ()
```
huitseeker, line 9: We should be fixing the PR URL, and mark as breaking (see serialization comment)

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1721,6 +1773,10 @@ impl DebugFileInfo {
 /// Links source-level function information to the compiled MAST representation.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DebugFunctionInfo {
+    /// The execution node linked to this function info
+    pub node: MastNodeId,
+    /// Source/debug occurrence linked to this function info.
+    pub source_node: Option<DebugSourceNodeId>,
```
huitseeker, line 1779: The two new fields `node` and `source_node` carry indices into other sections, but `validate_debug_function` in `package/mod.rs` only checks `name_idx`/`linkage_name_idx`/`file_idx`/`type_idx`. 

The old `mast_root` field was just a `Word` digest and didn't need bounds-checking, but these do.

Net effect: a `PackageDebugInfo` produced by a different trusted tool can carry `function.node = u32::MAX` or `function.source_node` past the end of the source graph, pass validation, then panic on first read of `mast_forest[node]` (or in `merge_debug_functions`, see the comment over there).

Worth adding bounds checks here too? Something like `function.node.as_u32() as usize >= self.mast.num_nodes()` and, when a source graph is present, the same against `source_graph.nodes().len()`.

crates/mast-package/src/debug_info/types.rs
bitwalker: Yes - it is used by the compiler/debugger naturally

crates/assembly/src/assembler.rs
bitwalker: IMO we should just always do the work, and just ignore emitting the debug info when requested - it needlessly complicates the code to make them optional (especially as both the source graph and debug info are already being produced during assembly), and the overhead of doing the post-processing work anyway is minimal - this is very much a premature optimization for something in the compiler pipeline.

I suspect we don't even need to clone the data structure anyway though - I'll take another look

huitseeker:
@bitwalker Sorry, I meant to answer but go caught doing something else. I don’t think `DebugSourceNodeId` is redundant. With the source graph model introduced in #3221, `MastNodeId` is execution identity, while `DebugSourceNodeId `is source occurrence identity. Multiple source occurrences can point at the same executable MAST node after reduction/dedup, but still have different source spans, asm-op rows, debug vars, inline-call rows, and eventually call-frame info. We encountered concrete cases where `DebugSourceNodeId` is not only essential, but where its absence let to the yanking of releases in the 0.22 and 0.23 lanes.

So anything that wants precise source-level debug lookup needs the source node.

A valid question is whether both fields belong in `DebugFunctionInfo`.

If `PackageDebugInfo` keeps today’s partial-section shape, then I’d keep both with validation that, when source_node is present, it exists and its `exec_node == node`. In that model, `node` is the execution fallback and `source_node` is the precise source/debug link.

If we move to the stricter model you describe, where function info only exists with a source graph/source map, then `source_node` should be the canonical key and node becomes derivable from `debug_info.source_node(source_node).exec_node`. Keeping both forces validation, and I agree it's annoying.

On the broader `PackageDebugInfo` shape: While I can see daylight between `DebugInfo` and `SourceMap` (especially when the producer of the forest is *not* the compiler), I agree with the direction for the sake of simplicity, but I’d make it an invariant around a debug-info bundle rather than just removing every `Option` independently. The useful states seem to be “no debug info” or “complete source debug info”, with `error_messages` allowed to be empty/absent as a special case.

That would make the API easier to reason about and avoid supporting odd states like function records without the source graph needed to interpret them precisely.

So TL;DR: `DebugSourceNodeId` is the important precise link; `MastNodeId` is only needed if we still support partial/debugless function metadata or want a validated cached execution link.

bitwalker:
Closing this in favor of #3398

PR #3380, opened 2026-07-15
fix: long miden-format imports and configurable constant wrapping

### LINKED ISSUE #3363: `miden-format` does not respect max line length for item imports
`miden-format` currently formats this

```
use {
    ACCOUNT_GET_ID_OFFSET,
    ACCOUNT_GET_NONCE_OFFSET,
    ACCOUNT_COMPUTE_COMMITMENT_OFFSET,
    ACCOUNT_GET_CODE_COMMITMENT_OFFSET,
    ACCOUNT_COMPUTE_STORAGE_COMMITMENT_OFFSET,
    ACCOUNT_GET_VAULT_ROOT_OFFSET,
    ACCOUNT_GET_ITEM_OFFSET,
    ACCOUNT_HAS_STORAGE_SLOT_OFFSET,
    ACCOUNT_GET_MAP_ITEM_OFFSET,
    ACCOUNT_GET_ASSET_OFFSET,
    ACCOUNT_GET_NUM_PROCEDURES_OFFSET,
    ACCOUNT_GET_PROCEDURE_ROOT_OFFSET,
    ACCOUNT_HAS_PROCEDURE_OFFSET,
} from miden::protocol::kernel_proc_offsets
```

into this:

```
use {ACCOUNT_GET_ID_OFFSET, ACCOUNT_GET_NONCE_OFFSET, ACCOUNT_COMPUTE_COMMITMENT_OFFSET, ACCOUNT_GET_CODE_COMMITMENT_OFFSET, ACCOUNT_COMPUTE_STORAGE_COMMITMENT_OFFSET, ACCOUNT_GET_VAULT_ROOT_OFFSET, ACCOUNT_GET_ITEM_OFFSET, ACCOUNT_HAS_STORAGE_SLOT_OFFSET, ACCOUNT_GET_MAP_ITEM_OFFSET, ACCOUNT_GET_ASSET_OFFSET, ACCOUNT_GET_NUM_PROCEDURES_OFFSET, ACCOUNT_GET_PROCEDURE_ROOT_OFFSET, ACCOUNT_HAS_PROCEDURE_OFFSET}
    from miden::protocol::kernel_proc_offsets
```

The problem is that the output does not respect the max line length (the line is 412 chars).

What to do when the max line length is reached is subjective and can be fixed in multiple ways. For simplicity, we probably want to choose one. My preference would be that the initial format wouldn't be touched, basically the `Vertical` option of rustfmt's `import_layout` in https://github.com/rust-lang/rustfmt/blob/main/Configurations.md#imports_layout. Imo, it is way easier to edit than horizontal layouts (I sometimes still edit code myself), and the verbosity of it is not a concern to me.

### LINKED ISSUE #3364: `miden-format` expands constants into four lines
`miden-format` currently formats this

```
pub const ON_BEFORE_ASSET_ADDED_TO_ACCOUNT_PROC_ROOT_SLOT = word("miden::protocol::faucet::callback::on_before_asset_added_to_account")
```

into this:

```
pub const ON_BEFORE_ASSET_ADDED_TO_ACCOUNT_PROC_ROOT_SLOT =
    word(
        "miden::protocol::faucet::callback::on_before_asset_added_to_account"
    )
```

Consider a slightly more concise (-1 line) format:

```
pub const ON_BEFORE_ASSET_ADDED_TO_ACCOUNT_PROC_ROOT_SLOT = word(
        "miden::protocol::faucet::callback::on_before_asset_added_to_account"
    )
```

The same would apply to all constants (and maybe word constants?).

Motivation (cc @mmagician):
- https://github.com/0xMiden/protocol/pull/3285#discussion_r3570804656
- https://github.com/0xMiden/protocol/pull/3285#discussion_r3570807816

Feel free to disagree, this is a pretty subjective thing. If there's no consensus we may want to have options for this eventually (not a high priority).

huitseeker:
`miden-format` could still create lines that were too long for item imports.

For item imports, it kept the whole import list on one line and moved only the `from` clause to the next line. A large import list could still be more than 400 characters long.

This PR changes the formatter so long item imports use one item per line.

It also adds an `overflow_delimited_expr` option for long delimited value expressions such as `pub const NAME = word("...")`. By default, long declarations keep the existing rustfmt-like style that moves the full expression to the next line. Projects that prefer the shorter style can set `overflow_delimited_expr = true` to keep `word(` or `event(` on the assignment line when that line still fits.

Closes #3363.
Closes #3364.

### DIFF
--- a/crates/miden-format/src/formatter.rs
+++ b/crates/miden-format/src/formatter.rs
@@ -291,9 +291,8 @@ fn render_item_import_lines(
     }
 
     let specs = import.item_specs().map(render_import_specifier).collect::<Vec<_>>();
-    let items = format!("{{{}}}", specs.join(", "));
     let from = format!("from {path}");
-    let line = format!("{header} {items} {from}");
+    let line = format!("{header} {{{}}} {from}", specs.join(", "));
 
     if line_length(&line) <= config.max_line_length() {
         vec![line]
@@ -309,27 +308,19 @@ fn render_vertical_item_import_lines(
     indent: usize,
     config: &Config,
 ) -> Vec<String> {
-    if specs.is_empty() {
-        return vec![
-            format!("{header} {{}}"),
-            format!("{}{}", indent_string(indent + config.indent_size()), from),
-        ];
-    }
-
-    let mut lines = vec![format!("{header} {{")];
     let item_indent = indent_string(indent + config.indent_size());
-    for spec in specs {
+    let mut lines = vec![format!("{header} {{")];
+    for spec in &specs {
         lines.push(format!("{item_indent}{spec},"));
     }
 
-    let mut closing = format!("{}}}", indent_string(indent));
-    if line_length(&closing) + 1 + line_length(&from) <= config.max_line_length() {
-        closing.push(' ');
-        closing.push_str(&from);
-        lines.push(closing);
+    let indent_str = indent_string(indent);
+    let inline_closing = format!("{indent_str}}} {from}");
+    if line_length(&inline_closing) <= config.max_line_length() {
+        lines.push(inline_closing);
     } else {
-        lines.push(closing);
-        lines.push(format!("{}{}", indent_string(indent + config.indent_size()), from));
+        lines.push(format!("{indent_str}}}"));
+        lines.push(format!("{item_indent}{from}"));
     }
 
     lines
--- a/crates/miden-format/src/formatter.rs
+++ b/crates/miden-format/src/formatter.rs
@@ -290,19 +290,49 @@ fn render_item_import_lines(
         );
     }
 
-    let specs = import.item_specs().map(render_import_specifier).collect::<Vec<_>>().join(", ");
-    let items = format!("{{{specs}}}");
+    let specs = import.item_specs().map(render_import_specifier).collect::<Vec<_>>();
+    let items = format!("{{{}}}", specs.join(", "));
     let from = format!("from {path}");
     let line = format!("{header} {items} {from}");
 
     if line_length(&line) <= config.max_line_length() {
         vec![line]
     } else {
-        vec![
-            format!("{header} {items}"),
-            format!("{}{}", indent_string(indent + config.indent_size()), from),
-        ]
+        render_vertical_item_import_lines(header, specs, from, indent, config)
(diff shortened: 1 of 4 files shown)


crates/miden-format/src/formatter.rs
```diff
@@ -1901,6 +1971,86 @@ const X =
...
+pub const ON_BEFORE_ASSET_ADDED_TO_ACCOUNT_PROC_ROOT_SLOT = word(\"miden::protocol::faucet::callback::on_before_asset_added_to_account\")
+";
+
+        let parse = parse_text(source);
+        assert!(!parse.has_errors(), "{:?}", parse.diagnostics());
+
+        let config = Config::default();
+        let formatted = format_syntax(&config, &parse.syntax());
+        let expected = "\
+pub const ON_BEFORE_ASSET_ADDED_TO_ACCOUNT_PROC_ROOT_SLOT = word(
```
bitwalker, line 2071: I would add an option for this (akin to `rustfmt`'s `overflow_delimited_expr: bool` option). Rust by default breaks those expressions on to a new line if they need to break, but that option allows the expressions to begin on the same line before wrapping.

I personally find the style where just one part of the expression starts on the same line to be less readable - if it has to break I'd rather read the expression on a single line, rather than across them. That said, its obviously a subjective thing, which is why I think it makes for a good candidate for configuration.

bitwalker (changes requested):
Looks good, though I object to the subjective change in the formatting of long declarations - see my comment for details.

In general I think we should avoid treating subjective choices as equivalent to bugs in the formatter, or we'll be perpetually dealing with debates over style. The first (and primary) issue this PR addresses is an actual bug, but the change to how long declarations are formatted when they are wrapped is not a bug, but a stylistic choice. Anything that falls in the latter bucket should probably be made configurable, with few exceptions, so individual projects can choose their preferences on those points.

huitseeker:
@bitwalker kept the existing long declaration wrapping as the default, and made the shorter `word(`/`event(` assignment-line style opt-in via `overflow_delimited_expr`.

PhilippGackstatter (approved):
Looks good, thank you very much!

bitwalker (approved):

PR #3382, opened 2026-07-16
fix(assembly): assert program entrypoint has no locals in set_num_locals

### LINKED ISSUE #3331: Assembler panics instead of erroring when the program entrypoint has locals set via the AST API

thep2p:
## Describe your changes

An executable module's `begin`..`end` block compiles to an ordinary procedure reachable via `Module::procedures_mut`. The `@locals(..)` grammar can't attach locals to `begin`, so the parser can never produce an entrypoint with locals. But `Procedure::set_num_locals` can set them programmatically. When it did, assembly aborted with an `assert!` deep in `compile_procedure`, far from the offending mutation:

```rust
let mut module = /* parse an executable module: begin .. end */;
for proc in module.procedures_mut() {
    proc.set_num_locals(4);
}
Assembler::new(source_manager).assemble_program("test", module); // panics deep in the assembler
```

## Rationale

Per the discussion in #3331, an entrypoint with locals is an unrecoverable producer bug. The MASM AST is deliberately a superset of what MASM source permits. So this stays an assertion rather than becoming a recoverable diagnostic. The change just moves the enforcement closer to where the invalid state is produced and documents it:

- Document the invariant on `Procedure::set_num_locals` and assert it there, at the AST mutation site, so the failure surfaces at the producer boundary instead of deep in the assembler.
- Retain the assembler-side assertion as a backstop for entrypoints constructed directly via `Procedure::new`, which bypasses the setter.
- Add tests covering both the `set_num_locals` and `Procedure::new` programmatic-AST paths.

Closes #3331.

## Test Plan
- Added `test_entrypoint_with_locals_via_setter_panics` and `..._via_constructor_panics`; run `make test-assembly`.

- [x] Updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@
 #### Fixes
 
 - [BREAKING] Bound MMR peak commitments to the leaf count by hashing `[num_leaves, 0, 0, 0] || padded_peaks`, and updated the core library `mmr::pack`/`mmr::unpack` procedures to use the same preimage ([#3388](https://github.com/0xMiden/miden-vm/pull/3388)).
+- Documented the program-entrypoint locals invariant on `Procedure::set_num_locals` and now assert it at that AST mutation site, so setting locals on an executable module's `begin`..`end` block panics at the producer boundary. The existing assembler assertion is retained as a backstop for entrypoints built directly via `Procedure::new` ([#3382](https://github.com/0xMiden/miden-vm/pull/3382)).
 - Validated `SectionId` on deserialization: `Section::read_from()` now rejects invalid identifiers and the `serde` path delegates to `FromStr`, keeping both readers on the same invariant ([#3277](https://github.com/0xMiden/miden-vm/pull/3277)).
 - Fixed `hash_elements_in_domain(&[], d)` colliding with `hash_elements_in_domain(&[ZERO; RATE_WIDTH], d)` for nonzero `d`, by absorbing a `ONE` padding marker on the empty-input branch ([#3366](https://github.com/0xMiden/miden-vm/pull/3366)).
 - Fixed `hash_bytes(&[])` returning `Word::default()`; the empty-bytes input now absorbs a padding marker and permutes, producing a nonzero digest consistent with the 10\* sponge padding rule ([#3366](https://github.com/0xMiden/miden-vm/pull/3366)).
--- a/crates/assembly-syntax/src/ast/procedure/procedure.rs
+++ b/crates/assembly-syntax/src/ast/procedure/procedure.rs
@@ -117,7 +117,18 @@ impl Procedure {
     }
 
     /// Override the number of locals allocated by this procedure.
+    ///
+    /// # Panics
+    ///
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 
 #### Fixes
 
+- Documented the program-entrypoint locals invariant on `Procedure::set_num_locals` and now assert it at that AST mutation site, so setting locals
```
huitseeker: This line leaves trailing whitespace, and `git diff --check origin/next...HEAD` fails on this new changelog entry. Can you trim the spaces here and on the next wrapped line?

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 
 #### Fixes
 
+- Documented the program-entrypoint locals invariant on `Procedure::set_num_locals` and now assert it at that AST mutation site, so setting locals 
+  on an executable module's `begin`..`end` block panics at the producer boundary. The existing assembler assertion is retained as a backstop for 
+  entrypoints built directly via `Procedure::new` ([#3282](https://github.com/0xMiden/miden-vm/pull/3382)).
```
huitseeker: The link target is PR 3382, but the visible text says `#3282`. Should this be `[#3382]`?

huitseeker (commented):
Overall looks ok, but the format of the changelog has significantly moved with the change of #3339, you may have to rebase.

CHANGELOG.md
thep2p: folded the entry into a single unwrapped line

CHANGELOG.md
thep2p: fixed

huitseeker (approved):
Thanks a lot!

PR #3383, opened 2026-07-16
assembly: expose more missing functionality of the assembler

bitwalker:
During migration of the compiler, I found additional missing APIs/functionality:

* Inability to configure the assembler debug info options (they were not exported for some reason)
* Removal of virtual targets from `ProjectAssembler` meant that we effectively always require the top-level project to be written to disk somewhere, when that is only really required for assembly of source dependencies. A lot of compiler test cases are completely virtual, so I've exposed `ProjectAssembler::assemble_source_package` with a slight modification to allow the sources and source provenance to be provided by the caller. This allows the top-level project to be completely virtualized, only requiring dependencies in source form to actually exist on disk somewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Changelog
 
-## v0.25.3 (2026-07-14)
+## v0.25.3 (2026-07-16)
 
 - Add package post-processing hooks to the `ProjectSourceProvider` trait ([#3375](https://github.com/0xMiden/miden-vm/pull/3375)).
+- Expose some assembler configuration methods, and `ProjectAssembler::assemble_source_project` ([#3383](https://github.com/0xMiden/miden-vm/pull/3383)).
 
 ## v0.25.2 (2026-07-12)
 
--- a/crates/assembly/src/assembler.rs
+++ b/crates/assembly/src/assembler.rs
@@ -236,15 +236,10 @@ impl Assembler {
         self
     }
 
-    #[cfg(feature = "std")]
-    pub(crate) fn with_emit_debug_info(mut self, yes: bool) -> Self {
-        self.emit_debug_info = yes;
-        self
-    }
-
-    #[cfg(feature = "std")]
-    pub(crate) fn with_trim_paths(mut self, yes: bool) -> Self {
-        self.trim_paths = yes;
+    /// Configure this assembler based on configuration in `profile`
+    pub fn with_profile(mut self, profile: &miden_project::Profile) -> Self {
+        self.emit_debug_info = profile.should_emit_debug_info();
+        self.trim_paths = profile.should_trim_paths();
         self
     }
 }
--- a/crates/assembly/src/project.rs
+++ b/crates/assembly/src/project.rs
@@ -207,10 +207,19 @@ where
         self
     }
 
+    /// Get the project being assembled
     pub fn project(&self) -> &ProjectPackage {
         self.project.as_ref()
     }
 
+    /// Assemble a target of the current project matching `target_selector`, using the configuration
+    /// under the profile `profile_name`.
+    ///
+    /// Returns an error if:
+    ///
+    /// * `target_selector` cannot find a matching target in the current project
+    /// * `profile_name` is not a profile defined in this project
+    /// * An error occurs during assembly of the project or one of its dependencies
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v0.25.3 (2026-07-14)
```
bobbinth: Let's change the date here.

crates/assembly/src/project.rs
```diff
@@ -748,9 +791,9 @@ where
 // ================================================================================================
 
 #[derive(Clone)]
-struct ResolvedPackage {
-    package: Arc<MastPackage>,
-    linked_kernel_package: Option<Arc<MastPackage>>,
+pub struct ResolvedPackage {
+    pub package: Arc<MastPackage>,
+    pub linked_kernel_package: Option<Arc<MastPackage>>,
 }
```
bobbinth, line 800: nit: would it make sense to expose these via methods? (I'm always a bit weary of structs with public fields).

crates/assembly/src/project.rs
```diff
@@ -244,32 +255,52 @@ where
...
+    /// assertion.
+    pub fn assemble_source_package(
         &mut self,
         package_id: PackageId,
         project: Arc<ProjectPackage>,
         target: &Target,
         profile_name: &str,
         required_lib: Option<ResolvedPackage>,
+        sources: Option<ProjectSourceInputs>,
+        source_provenance: Option<ProjectSourceProvenanceInputs>,
```
bobbinth, line 288: Would it make sense to bundle these into a single struct - e.g., something like:

```rust
pub struct ProjectSource {
    source: ProjectSourceInputs,
    provenance: ProjectSourceProvenanceInputs,
}
```

And then here we'd pass as single `source: Option<ProjectSource>` parameter.

crates/assembly/src/assembler.rs
```diff
@@ -236,17 +236,24 @@ impl Assembler {
         self
     }
 
-    #[cfg(feature = "std")]
-    pub(crate) fn with_emit_debug_info(mut self, yes: bool) -> Self {
+    /// Tells the assembler whether or not to produce packages with debug information
+    pub fn with_emit_debug_info(mut self, yes: bool) -> Self {
         self.emit_debug_info = yes;
         self
     }
```
bobbinth: I didn't realize we still have this. Is there a reason to not always emit debug info? Stripping it out is pretty easy now.

Probably not for this PR, but my preference would be to remove this setting and to always emit debug info (and then the user can clear it from the package, if needed).

bobbinth (approved):
Looks good! Thank you! This is a very shallow review from me, but I left a few comments inline - mostly optional nits and potential changes for the future.

crates/assembly/src/project.rs
bitwalker: This is basically a named tuple, so the fields being public isn't a risk (i.e. we'd literally use a tuple for this, except it would be confusing to identify what the fields are supposed to be).

crates/assembly/src/project.rs
bitwalker: The provenance is optional - but we could do this. I was mainly avoiding it to sidestep adding a new type that is only used by a single API.

crates/assembly/src/assembler.rs
bitwalker: IMO the `with_emit_debug_info`/`with_trim_paths` helpers can be removed - the new `with_profile` should be kept, so that the assembler can make use of configuration options of the profile. Having the assembler avoid adding debug info to the package based on the current profile (which must be specified for project assembly anyway) is essentially free, and avoids users having to manually strip debug info when their project configuration already says don't emit debug info in the first place.

bitwalker: I've removed `with_emit_debug_info`/`with_trim_paths`, since they are now unused and were never public to begin with

crates/assembly/src/project.rs
bitwalker: I'll make a note to revisit this as part of our overall cleanup of the assembler APIs - depending on how that shakes out, it'll be clearer what types we'll need to make the remaining APIs more type safe.

Issue #2282, opened 2025-10-13
precompiles: consider a separate precompile crate


adr1anh:
# Summary

Create a new top-level workspace crate:

'precompiles/'

with package name 'miden-precompiles'.

This crate should become the first-class home for Precompile VM (PVM) work inside 'miden-vm'.

The immediate goal is to create a clean landing zone for existing private/research PVM work so it can be moved into this repo incrementally, reviewed in smaller pieces, and developed on a focused tracking branch before merging into 'next'.

# Why

PVM work is not just "add another chip". It cuts across several parts of the system:

- deferred operation semantics;
- PVM AIR constraints;
- PVM trace layout and witness generation;
- core PVM chips;
- concrete precompile implementations;
- Rust-side handlers and registries;
- MASM wrappers and library exposure;
- tests that check the whole path end-to-end.

Without a dedicated crate, this work tends to spread across 'core', 'processor', 'air', 'prover', 'verifier', and 'miden-core-lib'. That makes it harder to develop quickly and harder to review.

A dedicated 'miden-precompiles' crate gives us one place to develop the PVM subsystem while keeping the boundary with the base VM clear. It also makes it easier to maintain a PVM tracking branch that should not constantly conflict with unrelated VM/API refactors.

# Proposed Direction

Start with one crate: 'miden-precompiles'.

Initially, this crate should contain the PVM-specific implementation work:

- PVM AIR constraints;
- PVM trace layout;
- PVM witness generation;
- the eval chip;
- the chunks chip;
- the permutation chip;
- the permutation-controller chip;
- concrete precompile implementations as they become ready;
- a built-in registry of supported precompiles;
- end-to-end PVM tests.

We should not start by splitting this into many crates. A later split into something like 'miden-precompile-vm' and 'miden-precompile-lib' may make sense, but only after the first end-to-end path is working and we understand the dependency graph better.

# Boundary With The Base VM

For now, assume the generic deferred-state framework stays in 'miden-core'.

That includes shared VM ABI pieces such as:

- deferred tags and nodes;
- transcript/root shape;
- wire format;
- generic deferred-state data structures;
- generic registry interfaces.

The 'processor' should remain generic. It may carry a registry and update deferred state through system events, but it should not know which concrete precompiles exist.

The new 'miden-precompiles' crate should own the concrete PVM/precompile logic: the chips, constraints, trace/witness generation, and concrete precompile implementations.

# MASM Wrappers

[...]

PR #3384, opened 2026-07-16
Shrink precompile prover integration

huitseeker:
Related to #2282.

## Rationale

This PR keeps the part of PR 3293 that is needed for deferred precompile verification in the VM.

It keeps `miden-precompiles-prover` in the workspace.

It exposes `verify_deferred` from the `miden_precompiles_prover` crate root.

It updates the verifier to call that root entry point.

It keeps the precompile benchmark. The benchmark source now lives under `precompiles/benches`, near the precompile API that it exercises. The benchmark target and benchmark-only dependencies live in the private `miden-vm-precompiles-bench` workspace package, so release packaging does not try to resolve unpublished workspace crates from crates.io.

It removes the imported prototype docs and examples from `precompiles-prover`.

## Why the docs are removed

The removed docs came from the standalone prototype. They are not a good review target for this integration PR because they mix current code, old notes, and future work.

Examples:

* `precompiles-prover/docs/framework-migration.md` describes a framework migration spike on `feat-framework-0.26`. It cites commit `8fbd826`, which is not present in this checkout. It also talks about 287 tests and four examples. This PR is not that migration spike.
* `precompiles-prover/BENCHMARKS.md` records one warm benchmark run on an Apple M3 Max with `rustc 1.94.1`. It says to rerun the numbers before treating them as current. It also depends on example binaries that this PR removes. This PR keeps the benchmark code, but drops the stale benchmark write-up.
* `precompiles-prover/docs/architecture.md` says the first deliverable is `Keccak-f[1600]`. The crate now includes uint arithmetic, elliptic curve group operations, transcript evaluation, and multi scalar multiplication. That page is no longer a clean description of the code under review.
* `precompiles-prover/DESIGN.md` is mostly an index into `precompiles-prover/docs/`. Keeping it after removing the docs would leave a broken index. Keeping the whole docs tree would ask reviewers to review a large design package along with the code integration.

The docs should come back after the API and constraints are settled. Then they can describe the code that lands.

## Why the benchmark stays

The benchmark code is useful.

It measures VM programs that call the Keccak and ECDSA precompile wrappers from the core library.

That is a VM performance question. It is not part of the prover design docs.

The benchmark still compares `Blake3_256` and `Poseidon2` proof hash functions. It still uses the same default workload of 100 Keccak calls and 4 ECDSA checks.


adr1anh (approved):
Looks great! thank you

PR #3385, opened 2026-07-17
assembly: use trusted deserialization when resolving preassembled packages

bitwalker:
This PR changes how preassembled packages are loaded from the project dependency graph to use `Package::read_from_bytes_trusted`, as we implicitly trust the package registry/cache. Without this, dependencies loaded from preassembled packages would have their debug information stripped, which would result in the final artifact missing debug info in some cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.25.5 (2026-07-16)
+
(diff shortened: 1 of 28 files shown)


bobbinth (approved):
Looks good! Thank you!

Would be good to also update the changelog and increment crate versions to v0.25.6.

bobbinth (approved):
All looks good! Thank you!

PR #3386, opened 2026-07-17
fix: add missing constraint in bitwise chiplet

Nashtare:
## Describe your changes

My review of #3345  didn't have time to land that it got merged into `next`, but I believe there is a potential issue introduced in the refactoring.
The bitwise chiplet is no longer constrained to an 8-row boundary, and thus a malicious prover could insert a few `is_padding` controller rows (the bus stays balanced because padding emits nothing) to push `bitwise_start` off an 8-boundary, the leading partial cycle then contains a response-emit row whose `(a,b,z)` was never reset, letting it emit a forged u32 AND/XOR. 

@Al-Kindi-0 feel free to correct me if that's an oversight and the concern is actually gated by some other constraint.

- Updated `CHANGELOG.md'


Al-Kindi-0 (approved):
LGTM, thank you for catching this!

PR #3387, opened 2026-07-17
Enforce chiplet first-row init across empty preceding sections

Al-Kindi-0:
Each non-controller chiplet detected its first row from the previous chiplet's `is_last` flag. Any section except the controller can be empty, and an empty predecessor chip has no last row, hence the flag never fires in such cases and the init was silently skipped.

This fix is simply to redefine `next_is_first` for memory and ACE to detect the section boundary directly from the chiplet selectors ("current row precedes the section, next row is its first row"), so it fires regardless of which preceding sections are empty.


adr1anh (approved):
Thanks for the fix, LGTM

PR #3388, opened 2026-07-17
fix: bind MMR commitments to leaf count

huitseeker:
MMR peak commitments now hash the forest leaf count together with the padded peaks. The VM pack and unpack procedures use the same preimage, and unpack rejects advice where the encoded leaf count does not match the committed preimage.

Contains the best parts of 0xmiden/crypto#997 and #3109. Closes 0xmiden/security-issues#18. See also 0xmiden/crypto#1087


Al-Kindi-0 (approved):
LGTM
I think, pending a holistic domain separation solution, this is a good short-term solution.

bobbinth:
Did this address https://github.com/0xMiden/miden-vm/issues/3530 in some way? Or is https://github.com/0xMiden/miden-vm/issues/3530 a broader issue?

huitseeker:
@bobbinth #3530 is a broader issue. The text of the issue touches on the solution implemented in this PR ("an ad hoc commitment format")

PR #3391, opened 2026-07-17
package: fix improper validation of procedure names when extracting module info

bitwalker:
This bug would cause any attempt to read package module info to fail if the procedure name was a quoted identifier (commonly needed for symbols emitted from languages like Rust).

### DIFF
--- a/crates/mast-package/src/package/mod.rs
+++ b/crates/mast-package/src/package/mod.rs
@@ -466,9 +466,9 @@ impl Package {
                     signature,
                     attributes,
                 }) => {
-                    let name = path.last().unwrap();
+                    let name = path.procedure_name().expect("valid procedure name").unwrap();
                     module.add_procedure_with_provenance(
-                        ast::ProcedureName::new(name).expect("valid procedure name"),
+                        name,
                         *digest,
                         signature.clone().map(Arc::new),
                         attributes.clone(),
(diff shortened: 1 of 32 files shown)


crates/mast-package/src/package/mod.rs
```diff
@@ -573,11 +580,23 @@ impl Package {
...
                 },
                 PackageExport::Constant(ConstantExport { path, value }) => {
-                    let name = ast::Ident::new(path.last().unwrap()).expect("valid identifier");
+                    let name =
+                        path.components().next_back().unwrap().expect("valid path component");
+                    let name = if name.is_quoted() {
+                        name.to_ident().expect("non-root path component")
+                    } else {
+                        ast::Ident::new(name.as_str()).expect("valid identifier")
+                    };
```
bobbinth: This pattern repeats 3 times - would it make sense to encapsulate it in some helper method? Maybe something like: `Path::try_last_component_into_ident()` (though, maybe something different would be more approrpirate)?

bobbinth (approved):
Looks good! Thank you! I left one comment inline to improve the maintainability of the code.

crates/mast-package/src/package/mod.rs
bitwalker: I've reworked this a bit

bobbinth (approved):
All looks good! Thank you!

crates/mast-package/src/package/mod.rs
```diff
@@ -561,9 +561,9 @@ impl Package {
                     signature,
                     attributes,
                 }) => {
-                    let name = path.last().unwrap();
+                    let name = path.procedure_name().expect("valid procedure name").unwrap();
```
huitseeker, line 568: This fixes `try_module_infos()`, but the public `Package::module_infos()` method above still does `path.last()` and then `ProcedureName::new(name)`. 

The old module-info path should use the same `procedure_name()` conversion too.

PR #3395, opened 2026-07-18
fix: Reuse debug source map tables during package load

### LINKED ISSUE #3394: Debug source map load time is too high for large packages
Loading a package with debug info can spend most of its time in `DebugSourceMapSection::from_parts_with_inline_calls()`.

This is visible when loading Rust compiled packages into the debugger. A test that used to take a few seconds now takes more than a minute.

The same path affects MASM packages too, but the cost is smaller for the packages I checked.

**What happens**

`DebugSourceMapSection::read_from()` reads the string table and location table from the package.

It then expands every asm op row into owned `String` and `Location` values.

After that, it calls `from_parts_with_inline_calls()`.

That constructor rebuilds those tables from the expanded rows.

As a result, users wait while the package loader builds tables that were already in the file.

**Where**

The extra work is here.

`crates/mast-package/src/debug_info/serialization.rs`

```rust
Ok(Self::from_parts_with_inline_calls(asm_ops, debug_vars, inline_calls))
```

`crates/mast-package/src/debug_info/types.rs`

```rust
let locations = intern_locations(&asm_ops);
let strings = intern_source_map_strings(&asm_ops);
```

The current row type also keeps owned copies of the values.

```rust
pub struct DebugSourceAsmOp {
    pub location: Option<Location>,
    pub context_name: String,
    pub op: String,
}
```

So the package does not get much value from the string and location tables after deserialization.

**Local probe**

Here is the probe source.

```rust
use std::{
    path::{Path, PathBuf},
    time::Instant,
};

use miden_debug_types::{ByteIndex, Location, Uri};
use miden_mast_package::{
    Package,
    debug_info::{
        DebugSourceAsmOp, DebugSourceInlineCall, DebugSourceMapSection, DebugSourceNodeId,
    },
};

fn workspace_path(path: impl AsRef<Path>) -> PathBuf {
    let path = path.as_ref();
    if path.is_absolute() {
        return path.to_path_buf();
    }

    Path::new(env!("CARGO_MANIFEST_DIR"))
        .parent()
        .and_then(Path::parent)
        .expect("crate should live under workspace/crates")
        .join(path)
}

fn make_rows(
    rows: usize,
    unique_strings: usize,
    unique_locations: usize,
    payload_len: usize,
) -> Vec<DebugSourceAsmOp> {
    let payload = "x".repeat(payload_len);
    let contexts = (0..unique_strings)
        .map(|i| format!("test::ctx::{i:06}::{payload}"))
        .collect::<Vec<_>>();
    let ops = (0..unique_strings)
        .map(|i| format!("operation_with_a_moderately_long_name_and_payload_{i:06}_{payload}"))
        .collect::<Vec<_>>();
    let locations = (0..unique_locations)
        .map(|i| {
            let start = (i as u32).saturating_mul(3);
            Location::new(
                Uri::new(format!("file:///synthetic/source_{i:06}_{payload}.masm")),
                ByteIndex(start),
                ByteIndex(start + 2),
            )
        })
        .collect::<Vec<_>>();

    (0..rows)
        .map(|i| {
            DebugSourceAsmOp::new(
                DebugSourceNodeId::from(0),
                i as u32,
                Some(locations[i % unique_locations].clone()),
                contexts[i % unique_strings].clone(),
                ops[(i * 17) % unique_strings].clone(),
                1,
            )
        })
        .collect()
}

fn env_usize(name: &str, default: usize) -> usize {
    std::env::var(name).ok().and_then(|value| value.parse().ok()).unwrap_or(default)
}

#[test]
#[ignore]
fn probe_debug_source_map_constructor_scaling() {
    let payload_len = env_usize("MDSMP_PAYLOAD_LEN", 0);
    let max_rows = env_usize("MDSMP_MAX_ROWS", 200_000);

    for rows in [10_000, 25_000, 50_000, 100_000, 200_000, 500_000, 1_000_000] {
        if rows > max_rows {
            continue;
        }

        let asm_ops = make_rows(rows, rows / 2, rows / 2, payload_len);
        let started = Instant::now();
        let section = DebugSourceMapSection::from_parts_with_inline_calls(
            asm_ops,
            Vec::new(),
            Vec::<DebugSourceInlineCall>::new(),
        );
        let elapsed = started.elapsed();

        eprintln!(
            "rows={rows} locations={} strings={} payload_len={payload_len} constructor_ms={}",
            section.locations().len(),
            section.strings().len(),
            elapsed.as_millis()
        );
    }
}

#[test]
#[ignore]
fn probe_package_debug_info_read() {
    let package_path = std::env::var("MDSMP_PACKAGE").expect("set MDSMP_PACKAGE");
    let package_path = workspace_path(package_path);
    let bytes = std::fs::read(&package_path).expect("package should be readable");

    let started = Instant::now();
    let package = Package::read_from_bytes_trusted(&bytes).expect("package should deserialize");
    let package_elapsed = started.elapsed();

    let started = Instant::now();
    let debug_info = package.debug_info().expect("debug info should deserialize");
    let debug_elapsed = started.elapsed();

    let source_map = debug_info.as_ref().and_then(|debug_info| debug_info.source_map());
    eprintln!(
        "path={} bytes={} package_ms={} debug_ms={} asm_ops={} locations={} strings={} debug_vars={} inline_calls={}",
        package_path.display(),
        bytes.len(),
        package_elapsed.as_millis(),
        debug_elapsed.as_millis(),
        source_map.map_or(0, |source_map| source_map.asm_ops().len()),
        source_map.map_or(0, |source_map| source_map.locations().len()),
        source_map.map_or(0, |source_map| source_map.strings().len()),
        source_map.map_or(0, |source_map| source_map.debug_vars().len()),
        source_map.map_or(0, |source_map| source_map.inline_calls().len()),
    );
}
```

To measure a package:

```bash
MDSMP_PACKAGE=target/test-dev/build/miden-core-lib-3b3a4f7bcaada604/out/assets/miden-core.masp cargo test -p miden-mast-package --test debug_source_map_probe probe_package_debug_info_read -- --ignored --nocapture
```

On my machine, the generated core MASM package printed this:

```text
bytes=4063600 package_ms=77 debug_ms=139 asm_ops=96738 locations=9147 strings=1298 debug_vars=0 inline_calls=0
```

To measure only the source map constructor:

```bash
MDSMP_MAX_ROWS=1000000 MDSMP_PAYLOAD_LEN=512 cargo test -p miden-mast-package --test debug_source_map_probe probe_debug_source_map_constructor_scaling -- --ignored --nocapture
```

On my machine, that printed this:

```text
rows=10000 locations=5000 strings=10000 payload_len=512 constructor_ms=26
rows=25000 locations=12500 strings=25000 payload_len=512 constructor_ms=53
rows=50000 locations=25000 strings=50000 payload_len=512 constructor_ms=118
rows=100000 locations=50000 strings=100000 payload_len=512 constructor_ms=260
rows=200000 locations=100000 strings=200000 payload_len=512 constructor_ms=540
rows=500000 locations=250000 strings=500000 payload_len=512 constructor_ms=1470
rows=1000000 locations=500000 strings=1000000 payload_len=512 constructor_ms=3136
```

**Fix**

Change `DebugSourceAsmOp` to store table indexes or shared values instead of owned `String` and `Location` values.

Keep the tables that `DebugSourceMapSection::read_from()` already read.

Do not call `from_parts_with_inline_calls()` after reading a source map section from bytes.

If `v0.25` can take this small breaking change, change the row type now.

huitseeker:
This fixes a slow path when loading packages with debug info. The deserializer already reads the source map's location and string tables, but it rebuilt those tables from every asm op. Rust compiled packages can have many repeated source strings and locations, so package load became very slow.

The fix keeps the tables from the serialized section and makes each asm op point at those shared values. The deserializer also rejects tables with duplicate or unused entries.

This changes `DebugSourceAsmOp` to store shared `Arc` values. Callers use accessors for location, context_name, and op.
  
Closes #3394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - [BREAKING] Add missing constraint in Bitwise chiplet ([#3386](https://github.com/0xMiden/miden-vm/pull/3386)).
 - [BREAKING] Fixed a soundness gap in the chiplets AIR where a chiplet section's first-row initialization was skipped when the preceding section was empty. A program that uses memory but performs no `u32and`/`u32xor` operations produces an empty bitwise section, which caused the memory chiplet to skip its "values not being written must be zero" reset; a malicious prover could exploit this to forge a read of never-written memory. Each section's first row is now identified from the chiplet selectors at the boundary rather than from the previous chiplet's last row, so the initialization holds no matter which preceding sections are empty. The ACE section-start reset was hardened the same way as a precaution ([#3387](https://github.com/0xMiden/miden-vm/pull/3387)).
 - [BREAKING] Optimize periodic columns evaluation for fewer ACE gates ([#3347](https://github.com/0xMiden/miden-vm/pull/3347)).
+- [BREAKING] Reused serialized debug source map tables during package load, avoiding a slow table rebuild and changing `DebugSourceAsmOp` fields to shared values ([#3395](https://github.com/0xMiden/miden-vm/pull/3395)).
 
 #### Fixes
 
--- a/crates/assembly/src/mast_forest_builder/static_import.rs
+++ b/crates/assembly/src/mast_forest_builder/static_import.rs
@@ -1,4 +1,4 @@
-use alloc::{collections::BTreeMap, sync::Arc, vec::Vec};
+use alloc::{collections::BTreeMap, string::String, sync::Arc, vec::Vec};
 
 use miden_core::{
     Word,
@@ -450,10 +450,10 @@ impl MastForestBuilder {
                 (
                     row.op_idx as usize,
                     AssemblyOp::new(
-                        row.location.clone(),
-                        row.context_name.clone(),
+                        row.location().cloned(),
+                        String::from(row.context_name()),
                         row.num_cycles,
-                        row.op.clone(),
+                        String::from(row.op()),
                     ),
                 )
             })
--- a/crates/assembly/src/project/tests.rs
+++ b/crates/assembly/src/project/tests.rs
@@ -584,7 +584,7 @@ end
             .expect("root package should contain source map")
             .asm_ops()
             .iter()
-            .find(|row| row.context_name == context_name)
+            .find(|row| row.context_name() == context_name)
             .map(|row| row.source_node)
             .unwrap_or_else(|| panic!("missing asm-op row for {context_name}"))
     };
@@ -603,11 +603,7 @@ end
         .into_iter()
         .filter(|&source_node| debug_info.source_node(source_node).unwrap().exec_node == depa_exec)
         .map(|source_node| {
-            debug_info
-                .first_asm_op_for_source_node(source_node)
-                .unwrap()
-                .context_name
-                .as_str()
+            debug_info.first_asm_op_for_source_node(source_node).unwrap().context_name()
         })
         .collect::<Vec<_>>();
     assert!(contexts_for_deduped_exec.contains(&"depa_ctx"));
@@ -635,7 +631,7 @@ end
         round_tripped_debug_info
             .first_asm_op_for_source_node(depa_source)
             .unwrap()
-            .context_name,
+            .context_name(),
         "depa_ctx",
     );
     assert_eq!(
--- a/crates/mast-package/src/debug_info/serialization.rs
+++ b/crates/mast-package/src/debug_info/serialization.rs
@@ -1,6 +1,6 @@
 //! Serialization and deserialization for the debug_info section.
 
-use alloc::{collections::BTreeSet, string::String, sync::Arc, vec::Vec};
+use alloc::{string::String, sync::Arc, vec::Vec};
 
 use miden_core::{
     Word,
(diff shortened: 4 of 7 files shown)


bitwalker (changes requested):
This doesn't actually fix the issue, though it does reduce duplicate allocations which is nice. It also treats duplicate strings as a validation error (which it isn't per se, de-duplication is an optimization, not a validity requirement). This also spends time verifying that debug source locations are used, which is also not necessary and an expensive check to be performing during deserialization.

Fundamentally the problem is that we are doing work during deserialization that we should be doing during packaging. Deserialization has to be very fast, with the debug info in particular, all we need to do is ensure that what gets deserialized isn't broken (i.e. has invalid references, too many records, etc.), it doesn't need to check for unused/duplicate data, because those things don't matter for debugging purposes. We still have debug info rows that hold owned data that doesn't get serialized via the strings tables, and that adds plenty of overhead as well.

I have a larger refactoring of the debug info structure that removes any owned types from the debug info rows themselves, and just stores internal table references. It collapses all of the various debug info sections into `PackageDebugInfo`, and provides APIs to expand the debug info rows into richer types which provide the owned data the debugger needs. I'll be doing some benchmarking here shortly to confirm that it fixes the issue I was observing before I open a PR - but I think that's the path we should take with fixing this (i.e. a more aggressive refactoring of the debug info managed by the assembler).

huitseeker:
@bitwalker Addressed : debug source map deserialization reuses the serialized tables and only rejects broken row references, without treating duplicate or unused table entries as invalid.

huitseeker:
Closing in favor of #3398

Issue #3373, opened 2026-07-14
Bring back `trace` assembly to generate `readonly` events


mooori:
> [...] 
> At the assembly level, we'll bring back the `trace` keyword. We don't even need to tie to the existing event syntax - could be something like:
> 
> ```
> const FOO = trace("foo")
> 
> proc main
>     trace.FOO
> end
> ```
> 
> Under the hood, this will be translated to our current event machinery as follows:
> 
> - We'll have a single new system event - e.g., `SystemEvent::TraceEvent`.
> - As a part of translating `trace` instruction into this event, we'll push two elements onto the stack: the ID of the `SystemEvent::TraceEvent` and the ID of the trace (i.e., the ID corresponding to "foo" in the above example).
> [...] 

 _Originally posted by @bobbinth in [#3324](https://github.com/0xMiden/miden-vm/issues/3324#issuecomment-4964154135)_

The implementation of #3324 in the host and vm is WIP. This issue tracks the `trace` assembly which should produce `[push.FOO, push.SystemEvent::TraceEvent, emit, drop, drop]`.

bobbinth:
I believe this was closed by #3396 - but please feel free to re-open if not.

mooori:
@bobbinth I would like to re-open this issue but don't have the permission. It is about bringing back `trace.FOO` assembly which will then be expanded to `[push.FOO, push.SystemEvent::TraceEvent, emit, drop, drop]`. It's the syntactic sugar for generating trace events mentioned as follow-up in #3396.

Issue #3324, opened 2026-07-03
Enable marking events `readonly`


mooori:
#### Context

Currently it's assumed that events must be handled. Hosts raise an error and abort execution if they encounter an event for which no handler is registered.

The debugger (`miden-debug`) has been using traces for debug-related "events" that only the debugger cares about. After the recent removal of traces, these must become "real" events.

#### Problem

There will be events that only a subset of hosts needs to handle: readonly events. If a host encounters a readonly event and has no handler registered for it, no error should be raised and execution should continue.

Individual hosts are free to register handlers for readonly events as long as these handlers don't produce advice mutations.

**Events should encode whether they are readonly.**

#### Example

Developers can use the SDK's `println!` macro to get a string printed in the debugger. This is handled by an event which triggers the debugger to read a string from memory and display it to the user.

So the debugger has an event handler for the corresponding readonly event, producing side effects (printing) but no advice mutations. Other hosts ignore this event.

#### Proposed long-term solutions

In previous conversations the following solutions have been proposed:

##### A: 1-bit specifier in `EventId`

`EventId` is a `felt` derived from a `Blake3_256` hash of `EventName`. The idea is to modify `EventId` derivation such that one bit of the `EventId` encodes whether the event is readonly.

Challenges:

- Can we find `n` such that bit `event_id[n]` has the same value for all existing readwrite events?
  - If not I think we'd need to change the ids of existing events
- How to modify `EventId` derivation such that bit `n` classifies the event?

##### B: Modify assembly syntax

Such that it encodes whether an emitted event is readonly.

Challenges

- Assembly syntax modifications are difficult

#### Temporary workaround

Currently there are 5 readonly events. In #3305 and #3316 we added no-op handlers for these events to the core library. Most hosts import the core library and its handlers, so due to these no-op handlers they ignore the readonly events and don't abort execution when encountering them. 

#### Follow-up task

- [ ] Remove readonly no-op handlers from the core library.

bobbinth:
> ##### A: 1-bit specifier in `EventId`
> `EventId` is a `felt` derived from a `Blake3_256` hash of `EventName`. The idea is to modify `EventId` derivation such that one bit of the `EventId` encodes whether the event is readonly.
> 
> Challenges:
> 
> * Can we find `n` such that bit `event_id[n]` has the same value for all existing readwrite events?
>   
>   * If not I think we'd need to change the ids of existing events
> * How to modify `EventId` derivation such that bit `n` classifies the event?

[...]

PR #3396, opened 2026-07-18
feat: add trace events (optional readonly events)

mooori:
Implements trace events as described in #3324. Recap:

- Debug events used to be handled via trace decorators.
- Trace decorators were removed recently, so debug trace decorators became events.
- Hosts raise an error and abort execution when encountering an event for which they don't have a handler registered. So for known debug events, we added temporary no-op handlers to the core lib #3305.

We want to let others use their own custom events without worrying that all hosts have handlers for them. That is the purpose of trace events, which are optional readonly events. Since hosts can ignore them, they  must not generate advice mutations.

### Implementation details

A trace event `foo` can be emitted with `push.foo_id, push.sys_event_trace, emit, drop, drop`.

On encountering `sys_event_trace`, the processor invokes the host's on_trace method which must not error on seeing a trace event it has no handler for. Trace handler are stored in another registry separate from event handlers. They account for `sys_event_trace` being on top of the stack.

### Terminology

- Traces previously existed as decorators and are now re-introduced as optional readonly events.
- Under the hood the new traces are events, so at some point the line between traces and events gets a bit blurry.

### Follow-up task

Bring back `trace` assembly #3373 

/cc @bobbinth, @bitwalker


processor/src/fast/basic_block/mod.rs
```diff
@@ -87,6 +88,42 @@ impl FastProcessor {
         }
     }
 
+    /// `trace_id` refers to the event defined by the user, which is on the stack below
+    /// `SystemEvent::TraceEvent`.
+    #[inline(always)]
```
adr1anh: This probably doesn't need `inline`

processor/src/errors.rs
```diff
@@ -176,6 +176,10 @@ pub enum HostError {
     DuplicateEventHandler { event: EventName },
     #[error("attempted to add event handler for '{event}' (reserved system event)")]
     ReservedEventNamespace { event: EventName },
+    #[error("attempted to add trace handler for '{event}' (already registered)")]
+    DuplicateTraceHandler { event: EventName },
+    #[error("attempted to add trace handler for '{event}' (reserved system event)")]
+    ReservedTraceNamespace { event: EventName },
```
adr1anh: I think we can just use "`Event`" for both classic events and trace events.

processor/src/host/handlers.rs
```diff
@@ -174,3 +174,397 @@ impl Debug for EventHandlerRegistry {
...
+pub type TraceError = Box<dyn Error + Send + Sync + 'static>;
+
+// TRACE HANDLER REGISTRY
+// ================================================================================================
+
+/// Registry for maintaining trace handlers.
+#[derive(Default)]
+pub struct TraceHandlerRegistry {
+    handlers: BTreeMap<EventId, (EventName, Arc<dyn TraceHandler>)>,
+}
```
adr1anh, line 219: Not critical, but maybe there would be a way to unify Event/Trace into a common HandlerRegistry?

adr1anh (approved):
This looks good! Only left some minor comments.

processor/src/test_utils/tracing_test_host.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    fn on_event(&mut self, process: &ProcessorState) -> Result<Vec<AdviceMutation>, EventError> {
+        let event_id: u32 = process.get_stack_item(0).as_canonical_u64().try_into().unwrap();
+        self.events.push(event_id);
+        Ok(Vec::new())
+    }
+
+    fn on_trace(&mut self, process: &ProcessorState) -> Result<(), TraceError> {
+        let trace_event_id: u32 = process.get_stack_item(1).as_canonical_u64().try_into().unwrap();
```
huitseeker: I think this can panic for normal trace IDs. 

Trace IDs are `EventId` values, and the docs show [deriving them from names](https://github.com/0xMiden/miden-vm/blob/next/docs/src/user_docs/assembly/events.md), so many valid IDs could be larger than `u32::MAX`.  The same downcast is in `TestHost::on_trace`. 

Could these record `u64` or `EventId` instead of `u32`?

huitseeker (changes requested):

processor/src/host/handlers.rs
mooori: I would lean towards having a separate `TraceHandlerRegistry` because:

- it allows to keep the registries simple; no overhead due to storing event and trace handlers (different type signatures) in the same registry
- with separate registries the returned handler always matches the expected signature, so no need to check or downcast

processor/src/fast/basic_block/mod.rs
mooori: Addressed in 37c0521

processor/src/errors.rs
mooori: Addressed in 37c0521

processor/src/test_utils/tracing_test_host.rs
mooori: Changed all test host `on_event` and `on_trace` handlers to record `u64` in e213e56.

AFAIK this was test-only code and the panic would be triggered only if someone added a test where a non-`u32` event ends up in one of these handlers. So I think nothing critical.

processor/src/test_utils/tracing_test_host.rs
```diff
@@ -0,0 +1,179 @@
...
+
+/// A minimal testing host that records trace events.
+///
+/// Intended only for running self-contained programs built directly from source. It resolves no
+/// external MAST forests, loads no kernels, and produces no advice mutations.
+///
+/// Trace events are triggered by emitting `SystemEvent::TraceEvent` identifier and the id of the
+/// trace event is expected below that on the stack.
+#[derive(Debug, Clone)]
+pub struct TracingTestHost {
```
huitseeker: Could this test use the existing `TestHost` instead? 

This PR already teaches `TestHost` to record both `event_handler` and `trace_handler`, while `TracingTestHost` is only used by `test_trace_event_handling`. Keeping both adds a second public testing host with the same behavior for one call site.

docs/src/user_docs/assembly/events.md
```diff
@@ -33,3 +33,24 @@ push.<felt> emit drop
 **System Events** - Built-in events handled by the VM for memory operations, cryptography, math operations, and data structures.
 
 **Custom Events** - Application-defined events for external services, logging, or custom protocols.
+
+### Trace Events (optional read-only events)
+
+Trace events are a special class of optional, read-only events. Unlike regular custom events, they cannot mutate the advice provider, and emitting one for which the host has no handler registered should not result in an error.
+
+A trace event is emitted by pushing the user trace event ID and then the `sys::trace_event` system event ID before `emit`. There is currently no `emit.<event_id>`-style immediate form for trace events.
```
huitseeker: After `push.MY_TRACE`, `emit.SYS_EVENT` (or `emit.event("sys::trace_event")`) emits the envelope and leaves only the user trace ID to drop. The current wording and manual-only example make that supported spelling look unavailable.

Could this distinguish a dedicated trace immediate from the ordinary immediate that already works?

huitseeker (approved):
LGTM, modulo a few inline nits

bitwalker (approved):

processor/src/test_utils/tracing_test_host.rs
mooori: Removed `TracingTestHost` in 690da9a and made the test use `TestHost` instead.

docs/src/user_docs/assembly/events.md
mooori: Addressed in 3a8fffb. As a follow-up it's planned to bring back `trace` assembly and make it expand to the sequence: #3373.

mooori:
The last nits should be resolved and if you consider it ready, please someone press the merge button (I don't have permission). Feel free to squash-merge if possible. Or if you don't mind a force-push, I can rebase and clean-up the commit history.

miden-vm/tests/integration/operations/decorators/mod.rs
bobbinth, line 1: Not related to this PR, but we should probably move some files around as we no longer have "decorators".

docs/src/user_docs/assembly/events.md
```diff
@@ -33,3 +33,30 @@ push.<felt> emit drop
...
+push.SYS_EVENT
+emit
+drop
+drop
+
+# Since `emit.<event_id>` expands to `push.<event_id>, emit, drop`, this can be shortened too:
+
+push.MY_TRACE
+emit.SYS_EVENT
+drop
```
bobbinth, line 57: This is fine for this PR, but I think we should introduce something like a `trace` keyword to make this more convenient. Basically, we should be able to do something like:

```
trace.MY_TRACE
``` 

And the assembler should expand this into the appropriate sequence of instructions.

processor/src/host/mod.rs
```diff
@@ -107,9 +120,25 @@ pub trait SyncHost: BaseHost {
...
+    /// - System events are handled by the VM before and don't call this method
     fn on_event(&mut self, process: &ProcessorState<'_>)
     -> Result<Vec<AdviceMutation>, EventError>;
+
+    /// Handles a trace event emitted from the VM.
+    ///
+    /// Trace events are optional, read-only events. [`SystemEvent::TraceEvent`] is at stack
+    /// position 0 and the user trace event ID is at position 1 when this handler is called. The
+    /// handler cannot mutate the advice provider. Hosts that do not care about trace events can use
+    /// this default no-op implementation. Hosts are expected to nat raise an error on encountering
```
bobbinth, line 132: typo: "nat" -> "not".

bobbinth (commented):
Looks good! Thank you! I left a few comments inline. The main one is that I think we should follow this up with a PR that adds a `trace` keyword and creates syntactic sugar for using trace events.

docs/src/user_docs/assembly/events.md
mooori: That's #3373 :)

miden-vm/tests/integration/operations/decorators/mod.rs
mooori: Opened #3449 for this.

processor/src/host/mod.rs
mooori: Sent the fix in #3450

PR #3398, opened 2026-07-19
debuginfo: consolidate debug sections for efficiency and easier mx

bitwalker:
This PR consolidates all of the `PackageDebugInfo` sections into `PackageDebugInfo` itself, this:

* Improves sharing of common data such as strings
* Improves the ergonomics of accessing various types of debug info
* Centralizes construction and maintenance of debug info under a single hierarchy of types
* Simplifies MastForestBuilder internals that handle op-specific debug info
* Many of the individual debug info records have been simplified so as to avoid holding any owned/heap-allocated types, with the exception of vectors of certain sub-records attached to `SourceNode` as a deliberate trade-off.

Additionally, this addresses the extremely slow deserialization/loading of `PackageDebugInfo` from `Package`, making it about as efficient as possible given the amount of data involved.

The changes are are breaking for the debugger only - all other components were able to be migrated without any breakage. However, technically this is a breaking change to various debug info types, so we may have to land this in `next` and abandon the v0.25 series - we can discuss here.

Example timing information for a full end-to-end test from the compiler suite:

- Before: ~90s to run one of our end-to-end arithmetic tests (checked addition of u32s), a lot of time spent interning strings, but no single source of overhead per se - a lot of papercuts leading to the overall timing
- After: ~9s to run the same test, with about half that time spent validating the debug info indices post-deserialization.

Running the new `deserialize_core_lib_debug_info` benchmark prior to this commit (compared to after this commit) shows the difference starkly:

```
Finished [`optimized` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 59.95s
     Running benches/deserialize_core_lib_debug_info.rs (target/optimized/deps/deserialize_core_lib_debug_info-e6e718c83de1b76a)
Gnuplot not found, using plotters backend
Benchmarking deserialize_core_lib_debug_info/read_from_bytes: Collecting 100 samples in estimdeserialize_core_lib_debug_info/read_from_bytes
                        time:   [20.843 ms 21.368 ms 21.991 ms]
                        change: [+1628.0% +1673.8% +1727.8%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 15 outliers among 100 measurements (15.00%)
  5 (5.00%) high mild
  10 (10.00%) high severe
```


crates/assembly/src/assembler.rs
```diff
@@ -1138,20 +1133,19 @@ impl Assembler {
...
         kernel: Option<Arc<Package>>,
     ) -> Result<AssemblyProduct, Report> {
         let mast = Arc::new(mast_forest);
         let entry: Arc<Path> = namespace.join(ast::ProcedureName::MAIN_PROC_NAME).into();
         let entry_digest = mast[entrypoint].digest();
-        let entry_source_node = source_graph
-            .unique_root_for_exec_node(entrypoint)
-            .map(|source_id| DebugSourceNodeId::from(u32::from(source_id)));
+        let entry_source_node =
+            debug_info.unique_source_root_for_exec_node(entrypoint).into_diagnostic()?;
```
huitseeker: This now turns an ambiguous source root into an assembly error for executables. 

That can happen when `main` deduplicates to the same executable MAST node as another procedure root: both source roots point at the same `entrypoint`, so `unique_source_root_for_exec_node()` returns `AmbiguousRoot`.

Could this use the entry procedure's known source ref instead of looking it up by executable node?

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -910,6 +604,101 @@ impl Deserializable for DebugFunctionInfo {
...
+        let slice_layout = alloc::alloc::Layout::array::<T>(slice.len()).unwrap();
+        let range = slice.as_ptr_range();
+        let actual_size = range.end.byte_offset_from_unsigned(range.start);
+        let layout_size = slice_layout.size();
+        assert!(
+            actual_size >= layout_size,
+            "expected layout of slice of len {} ({layout_size} bytes) to be <= {actual_size} bytes",
+            slice.len()
+        );
+        let bytes = core::slice::from_raw_parts(ptr, actual_size);
```
huitseeker: This makes the package debug format depend on Rust's in-memory layout for each copied row. 

Some of these rows contain padding or Rust-layout fields like `Option<DebugStringIdx>` / `Option<DebugTypeIdx>`, so this can serialize padding bytes and can change across compiler versions or targets even when the logical fields are the same.

Could we keep the fast path only for row types with a checked plain-old-data layout and explicit byte order, and serialize the rest field-by-field? 
The function table seems especially risky because it includes `Word` plus several `Option` fields.

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,848 @@
...
+            .map(|pos| DebugFileIdx::from(pos as u32))
+    }
+
+    /// Apply `trimmer` to every string in the strings table which represents a file path
+    ///
+    /// If `trimmer` returns `None`, the table entry is left unmodified; otherwise, the table
+    /// entry is replaced with the string that is returned.
+    pub fn trim_file_paths(&mut self, mut trimmer: impl FnMut(&str) -> Option<Arc<str>>) {
+        for file in self.files.iter() {
+            let string = &mut self.strings[file.path_idx];
```
huitseeker: This mutates the shared string table entry used by a file path. 

Since the new format interns file paths, function names, context names, op names, and messages into one table, trimming a path can also rewrite any unrelated debug string that happened to intern to the same value.

Could this add/reuse the trimmed path as a separate string and update only `file.path_idx`? That keeps path trimming local to file records.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,613 +1,337 @@
...
-        }
+        // Read the serialized data into a temporary allocation specifically aligned so that we
+        // can construct references directly into it for certain rows/tables. It is not safe to do
+        // so unless we guarantee the alignment of the entire buffer, because our padding bytes
+        // are emitted based relative to the start of the buffer.
+        let data_len = read_bounded_len(source, "package debug info", 1)?;
+        let data = source.read_slice(data_len)?;
+        // Copy the data to an allocation guaranteed to be u32-aligned - this is sufficient to
+        // guarantee proper alignment of all the contained data.
+        let layout = alloc::alloc::Layout::from_size_align(data_len, align_of::<u32>())
```
huitseeker: This allocation does not look strong enough for the later typed reads. 

The buffer is allocated with `align_of::<u32>()`, but the same reader later casts parts of it to types such as `DebugFunctionInfo`, which contains a `Word`. `Word` can require 16-byte alignment on the Miden wasm target, and this will hit the `ptr.is_aligned()` assert or make the cast invalid there.

There is also a layout mismatch: this pointer is allocated with a custom `Layout`, then converted to `Box<[u8]>`, so it will be freed as a byte slice rather than with the allocation layout used here. 

This could maybe use a small aligned-buffer wrapper that deallocates with the same layout, and align to the max alignment of all zero-copy table rows?

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,848 @@
...
+    ) -> Result<Self, DebugInfoMergeError<MastNodeId, Src>>
+    where
+        Src: 'a,
+    {
+        let mut inputs = inputs.into_iter();
+        let Some((_, base)) = inputs.next() else {
+            return Ok(Default::default());
+        };
+
+        let mut builder = DebugInfoBuilder::from(Box::from(base.clone()));
```
huitseeker: This seeds the merged debug info from the first input without applying `root_map` to its `SourceNode.exec_node` values. The loop below remaps every later input through `root_map.map_node(...)`, so the first forest is the only one left in pre-merge node-id space.

Could the first input go through the same remap path as the rest? 

If `MastForest::merge` canonicalizes or deduplicates nodes from the first forest, the debug graph can point at stale node ids in the merged package, see e.g. https://gist.github.com/huitseeker/9c06e9ecaef18eb600c7b062905cb828

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,848 @@
...
+                    DebugStringIdx::from(u32::try_from(i).expect("invalid string table index"));
+                let new_index = builder.add_string(string.clone());
+                remapped_strings.insert(prev_index, new_index);
+            }
+
+            let mut remapped_types = BTreeMap::<DebugTypeIdx, DebugTypeIdx>::default();
+            for (i, debug_type) in debug_info.types.iter().enumerate() {
+                let prev_index =
+                    DebugTypeIdx::from(u32::try_from(i).expect("invalid types table index"));
+                let debug_type = remap_debug_type_info(
```
huitseeker: This remaps each type before adding that type's old-to-new entry to `remapped_types`. That rejects valid forward references in later inputs, for example a pointer or function type that refers to a type row that appears later in the same table.

The finalizer handles this by pre-seeding the full type map before rewriting any type payloads. Could this merge path do the same?

crates/assembly/src/mast_forest_builder/debug_info.rs
```diff
@@ -0,0 +1,379 @@
...
+};
+
+use super::{MastNodeRef, SourceNodeRef};
+
+/// This visits a type exported or used in a procedure signature, and emits records to the
+/// provided debug types section corresponding to it.
+///
+/// The declared name and type expression can be optionally provided to give additional useful
+/// context to the debug info type produced, e.g. type name, field names, etc.
+pub fn register_debug_type(
```
huitseeker: This free function looks like a near-exact duplicate of `DebugInfoBuilder::register_debug_type`.

Could we delete this file and have the one caller (`assembler/debuginfo.rs`?) invoke `self.debug_info.register_debug_type(...)` directly?

crates/mast-package/src/debug_info/builder.rs
```diff
@@ -0,0 +1,784 @@
...
+        let path_idx = self.add_string(uri);
+        self.add_file_info(
+            DebugFileInfo::new(path_idx)
+                .with_checksum(checksum.unwrap_or(DebugFileInfo::EMPTY_CHECKSUM)),
+        )
+    }
+
+    /// Adds a file to the file table and returns its index.
+    pub fn add_file_info(&mut self, file: DebugFileInfo) -> DebugFileIdx {
+        assert!(
```
huitseeker, line 262: These asserts check invariants on a builder that the merge path (`merge_source_debug`) and the finalizer reach into via `debug_info_mut()`. 

Would it make sense to return `Result<_, IndexedVecError>` (with a new `InvalidIndex` variant) from `add_file_info`, `add_node`, `add_root`, and `add_location_info`?

The callers in the merge path already bubble `Result<_, Report>`, so converting an `IndexedVecError` at the boundary is essentially free.

crates/utils-indexing/src/lib.rs
```diff
@@ -409,6 +445,98 @@ where
...
+        for _ in 0..len {
+            vec.push(deserializer(source)?);
+        }
+
+        Ok(Self { raw: vec, _m: PhantomData })
+    }
+}
+
+/// Reads and validates a serialized length before it is used for allocation.
+fn read_bounded_len<R: ByteReader>(
```
huitseeker, line 503: These two helpers are copies of `miden_core::serde::{read_bounded_len, validate_bounded_len}` It looks like the only reason `utils-indexing` re-exports `miden_serde_utils` and depends on `miden_crypto` transitively is to define them.

Could `IndexVec::read_from_bounded` and any other local caller just use `miden_core::serde::read_bounded_len` directly and we drop these two copies?

huitseeker (changes requested):
Requesting changes for the memory-layout issues.

crates/mast-package/src/debug_info/mod.rs
huitseeker: Fixed ff75e4b36b

crates/mast-package/src/debug_info/mod.rs
huitseeker: FIxed in 2827cf8a11

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,613 +1,340 @@
...
 
-        // Write file table
-        target.write_usize(self.files.len());
-        for file in &self.files {
-            file.write_into(target);
-        }
-    }
-}
+        let functions_len = source.read_u32()?;
+        let functions = source
```
huitseeker: Wondering if it's ok to construct `DebugFunctionInfo` values directly from user-provided bytes.

`DebugFunctionInfo` has `Option<...>` fields and `LineNumber` / `ColumnNumber` values backed by non-zero integers. A bad section can put invalid enum tags or zero values in this byte range, and the cast in `read_aligned_slice_of` creates those Rust values before package validation gets a chance to reject them.

Could this row stay field-decoded, or use an explicit wire row made only from plain integers and validate it before building `DebugFunctionInfo`?

crates/mast-package/src/debug_info/mod.rs
huitseeker: Fixed in 2827cf8a11

crates/mast-package/src/debug_info/serialization.rs
huitseeker: Fixed in 8438d27d35

crates/utils-indexing/src/csr.rs
```diff
@@ -6,7 +6,7 @@
 
 use alloc::vec::Vec;
 
-use miden_crypto::utils::{
+use miden_serde_utils::{
```
huitseeker, line 9: Is `CsrMatrix` still needed after this change?

I checked `origin/main` and this head with `git grep`; outside its own docs, tests, and re-exports, I do not see a caller. If the removed source debug graph was the last real use, could this PR delete `csr.rs` and the `CsrMatrix` / `CsrValidationError` exports instead of keeping this dependency swap?

crates/mast-package/src/package/section.rs
```diff
@@ -73,15 +62,7 @@ impl SectionId {
...
-        matches!(
-            self.as_str(),
-            "debug_types"
-                | "debug_sources"
-                | "debug_functions"
-                | "debug_source_graph"
-                | "debug_source_map"
-                | "debug_error_messages"
-        )
+        self == &Self::DEBUG_INFO
```
huitseeker, line 65: Nit: This now only treats `debug_info` as debug metadata.

Older packages can still contain `debug_types`, `debug_sources`, `debug_functions`, `debug_source_graph`, `debug_source_map`, and `debug_error_messages`. 

The untrusted package reader and `strip_debug_info()` both call `is_debug()`, so those old sections would no longer be stripped and could be reserialized as custom sections.

So definitely a Big Breaking Change requiring purging data, which sounds like a major release

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,910 @@
...
+pub const DEBUG_INFO_VERSION: u8 = 2;
+
+// PACKAGE DEBUG INFO
+// ================================================================================================
+
+/// Trusted package-owned debug information decoded from well-known debug sections.
+pub type PackageDebugInfo = DebugInfo<MastNodeId, DebugSourceNodeId>;
+
+/// Represents debug information bound to a pending/finalized [`miden_core::mast::MastForest`].
+///
```
huitseeker, line 40: Could this add a small `Arbitrary` strategy for `PackageDebugInfo` and use the existing `serde_test(binary_serde(true))` round-trip macro here?

That macro was introduced in #2132 for this exact pattern: once a type has `Arbitrary`, the generated binary round-trip test gives more coverage than hand-written examples. A limited strategy with small tables would still exercise empty tables, shared strings, optional function fields, source nodes, and padding-heavy layouts without making the whole debug-info generator large.

crates/mast-package/src/debug_info/serialization.rs
bitwalker: The structs are all marked `#[repr(C)]`, so the layout is well-defined - the exception to `#[repr(C)]` currently are all `Option`-wrapped integral types, which have a predictable layout for now. We can certainly use our own `Option`-like type though if we expect Rust is going to change the layout of those anytime soon.

The padding bytes aren't really an issue in practice - but I'm about to push a commit which reorganizes the fields of these types, and fixes their size, such that there are zero padding bytes between elements (and between fields of those elements).

crates/mast-package/src/debug_info/builder.rs
bitwalker: I think we want those to be asserts (i.e. produce a panic) - since violating them is necessarily a bug in the code which calls the APIs, rather than an issue with user input. If we return an error, it will obscure the source of the issue, and make it extremely painful to track down the culprit IME.

crates/utils-indexing/src/lib.rs
bitwalker: This crate doesn't depend on `miden-core`, only `miden-serde-utils`, and we don't want it to depend on `miden-core` (I suspect it can't in any case, due to cyclic dependencies). I'm not sure why these helpers were added to `miden-core` rather than `miden-serde-utils` (probably because it would've been too much of a pain to do the release cycle dance), but these can be removed in favor of ones in `miden-serde-utils` once we can move them from `miden-core` to `miden-serde-utils` and have `miden-core` just re-export them.

For now this was the only way to get those helpers where they were needed.

crates/mast-package/src/debug_info/serialization.rs
bitwalker: `Option<NonZeroNN>` has a specific niche layout - a value of `0` is `None`, and all other values in the `NN` bit range are valid `NonZeroNN` values, so those at least are perfectly safe.

crates/utils-indexing/src/csr.rs
bitwalker: I'm not sure if there are other downstream users of that, but I'd make that change on `next` just in case.

I think `miden-utils-indexing` should depend on the narrower `miden-serde-utils` dependency - it has no need for anything else in `miden-crypto`, and the latter is a much heavier dependency.

crates/mast-package/src/package/section.rs
bitwalker: I don't think this is a big breaking change? There aren't any packages built "for real" with v0.25 yet - and those that are (the core lib, the protocol lib) are going to be rebuilt with the new debug info format anyway.

crates/mast-package/src/debug_info/serialization.rs
bitwalker: I've gone ahead and introduced a stable layout equivalent of `Option` for use in these types just to solidify that detail.

None of the types have any padding bytes (either between fields or between elements when laid out sequentially in memory).

crates/mast-package/src/debug_info/serialization.rs
bitwalker: I've changed `LineNumber`/`ColumnNumber` here to use `LineIndex`/`ColumnIndex` instead (the former was just for convenience, so converting them to indices and back sidesteps the issue with those types.

All the other fields that were using `Option` are now using a type with a stable memory layout

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,613 +1,338 @@
...
+            .read_aligned_slice_of::<DebugFunctionInfo>(functions_len as usize)?
```
huitseeker: `OptionC<T>` is `#[repr(u32)]` with two variants (`None = 0`, `Some = 1`). `DebugFunctionInfo` carries three `OptionC<...>` fields and `DebugSourceAsmOp` carries one. `read_aligned_slice_of` followed by `.to_vec()` constructs Rust values directly from the buffer bytes.

If any of the four discriminant words in a row holds a value >= 2 (corrupted trusted-cache file, bit flip on disk, a stale file from a prior format version), we materialize a Rust enum value with an invalid discriminant. The Rust reference is explicit that producing such a value is undefined behavior, even before we ever match on it, and even if it only ever lives inside a `Vec`.

This matters most on the trusted-cache paths (`read_from_bytes_trusted`, the embedded kernel reader at `package/mod.rs:1124`, and `read_from_bytes_unchecked`). Those paths exist precisely so that local on-disk artifacts can skip `validate_debug_info`, and on-disk artifacts do rot.

`Option<NonZeroNN>` was safe here because every bit pattern is a valid inhabitant (0 = None, anything else is a valid `NonZeroNN`). `OptionC<T>` does not have that property.

Could the wire form for these rows stay a plain `u32` discriminant + payload, validated against `{0, 1}` before we ever cast the buffer to typed references? Or scan the discriminant words of each row up front and reject the slice if any are out of range, before the `from_raw_parts` call?

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1060,472 +265,180 @@ impl Deserializable for DebugSourceNodeId {
...
-    pub op: String,
+    /// Source location index in the locations table
+    pub location_idx: OptionC<DebugLocIdx>,
+    /// Assembly context name index in the strings table
+    pub context_name_idx: DebugStringIdx,
+    /// Assembly operation text index in the strings table
+    pub op_name_idx: DebugStringIdx,
     /// Number of VM cycles taken by the operation.
     pub num_cycles: u8,
+    _padding: [u8; 7],
```
huitseeker: With `op_idx: u32` (4) + `location_idx: OptionC<DebugLocIdx>` (8) + `context_name_idx: u32` (4) + `op_name_idx: u32` (4) + `num_cycles: u8` (1) = 21 bytes used, alignment 4. `repr(C)` would have padded `num_cycles` to a 24-byte struct naturally (3 bytes of trailing pad to reach the next 4-byte boundary).

`[u8; 7]` makes the explicit record 28 bytes. `[u8; 3]` (or no `_padding` field at all) would give 24 bytes with the same no-internal-padding property the size-check asserts.

asm_ops rows are the hottest table in source nodes, so 4 bytes/row is meaningful. Worth switching to `[u8; 3]`?

crates/mast-package/src/debug_info/types.rs
```diff
@@ -1060,472 +265,180 @@ impl Deserializable for DebugSourceNodeId {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(C)]
 pub struct DebugErrorMessage {
     /// Runtime error code emitted by the assembled assertion operation.
     pub err_code: u64,
-    /// Human-readable assertion error message from source.
-    pub message: Arc<str>,
+    /// String table index of the error message content
+    pub message: DebugStringIdx,
+    _padding: [u8; 0],
```
huitseeker: A `[u8; 0]` field takes no space and adds no alignment, so it doesn't enforce the zero-init-trailing-pad property the equivalent field on `DebugSourceAsmOp` is going for. `DebugErrorMessage` is already 16 bytes (8-byte align) without it.

If the intent was to absorb the 4 bytes of trailing pad so the wire form has no uninitialized bytes, this wants `[u8; 4]`. If there was no intent, the field, the `_padding: [0; _]` in `new()`, and the size-check const can all go.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -640,10 +595,39 @@ fn serialize_basic_block_error_codes(op_batches: &[OpBatch]) -> Vec<u8> {
...
+        return None;
+    }
+
+    let mut data = Vec::new();
+    data.extend_from_slice(&raw_op_count.to_le_bytes());
+
+    let mut raw_op_prefix = 0_u64;
+    for batch in op_batches {
+        for group_idx in 0..batch.num_groups() {
+            let group_op_count = batch.indptr()[group_idx + 1] - batch.indptr()[group_idx];
```
huitseeker, line 680: It's very hard for me to remember that`raw_op_prefix += (group_op_count - usize::from(has_padding))` does not underflow even if `has_padding` is ever true on an empty group — the builder's `pad_if_needed` should insert a `Noop`.

Adding a `debug_assert!(group_op_count >= usize::from(has_padding))` next to the subtraction would help me and document the invariant.

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,923 @@
...
+                                    forest_index,
+                                    location_idx: idx,
+                                },
+                            )
+                        })
+                        .transpose()?;
+                    debug_vars.push(DebugSourceVar {
+                        op_idx: row.op_idx,
+                        name_idx,
+                        type_id: row.type_id,
```
huitseeker: `row.type_id` is copied through unchanged while every other index in this loop (name_idx, location_idx, exec_node, children) is remapped via the input→merged tables. If a later input package has a `DebugSourceVar` with a non-zero `type_id`, the merged debug info will point at the wrong row of the merged types table.

huitseeker (commented):
Looking good! A few padding-related places I still have questions about

crates/mast-package/src/debug_info/serialization.rs
bitwalker: Yep, good point

crates/mast-package/src/debug_info/types.rs
bitwalker: Yep, we can shrink the padding here

crates/mast-package/src/debug_info/types.rs
bitwalker: Hmm, I don't remember adding padding to this struct, but in any case, yeah we can definitely remove it

crates/mast-package/src/debug_info/mod.rs
bitwalker: Yeah, this should be made a `DebugTypeIdx` and remapped like others - it has been getting treated as a raw `u32` because of how debug vars were being emitted in MASM - but looking through the compiler, nothing is emitting a `type_id` today, and won't ever be able to directly because that is information maintained by the MastForestBuilder during assembly.

I think `DebugVar` needs to change to hold a `Option<Type>` + `Option<TypeExpr>`, which is then converted to `DebugTypeInfo` via `register_debug_type` during assembly. We don't actually need `DebugVarInfo` anymore (not in `miden-core` at least).

I'll make that change here as well, since currently the compiler is the only thing producing these records, and they are basically broken right now.

crates/utils-indexing/src/lib.rs
huitseeker: It would be great to track the move of those helpers, the removal of `CsrMatrix` and `LookupByIdx` in an issue for next

crates/utils-indexing/src/csr.rs
huitseeker: As mentioned [above](https://github.com/0xMiden/miden-vm/pull/3398#discussion_r3633322789) let's track in an issue.

crates/utils-indexing/src/lib.rs
bitwalker: https://github.com/0xMiden/miden-vm/issues/3415 has been created for this

crates/assembly/src/assembler.rs
```diff
@@ -1697,21 +1685,23 @@ impl Assembler {
             }
         }
 
+        #[cfg(false)]
```
huitseeker: We should either have this stay enabled, or the test and error path be removed in this PR. I don't know if there's a value in punting on the decision.

crates/assembly/src/mast_forest_builder/static_import.rs
```diff
@@ -447,20 +421,41 @@ impl MastForestBuilder {
...
+            .map(|row| {
+                let name_idx = self.debug_info.add_string(package_debug_info[row.name_idx].clone());
+                let location_idx = row
+                    .location_idx
+                    .and_then(|loc| package_debug_info.get_location(loc))
+                    .map(|loc| self.debug_info.add_location(loc));
+                DebugSourceVar {
+                    op_idx: row.op_idx,
+                    name_idx,
+                    type_id: row.type_id,
```
huitseeker: Should this remap `type_id` too?

This copies `row.type_id` from the dependency package after remapping the name and location into `self.debug_info`. That index belongs to `package_debug_info.types`, so a statically linked package with typed debug variables can point at the wrong type row.

This looks like the same class of issue as the merge path below. The source package type index needs to be mapped into the destination debug info before it is stored here.

crates/mast-package/src/package/mod.rs
```diff
@@ -709,48 +679,47 @@ impl Package {
...
-                )?;
+            for row in source_node.asm_ops.iter() {
+                self.validate_source_map_row(source_id, source_node, row.op_idx, "assembly op")?;
+                self.validate_string_index(row.context_name_idx, debug_info, || {
+                    format!("debug source node {source_id:?} assembly op context name")
+                })?;
+                self.validate_string_index(row.op_name_idx, debug_info, || {
+                    format!("debug source node {source_id:?} assembly op name")
+                })?;
+                if let Some(location_idx) = row.location_idx.into_option() {
```
huitseeker: A corrupted u32 discriminant on `OptionC` here (and in the other `into_option` callers in `validate_debug_function`) will panic instead of producing a `PackageDebugInfoError`.

The row types `DebugFunctionInfo` and `DebugSourceAsmOp` are `#[repr(C)]` and cast straight off the wire via `read_aligned_slice_of` (see the SAFETY comment in `serialization.rs` around line 653 — it explicitly pushes responsibility for "encoded bytes are valid values of `T`" onto callers). So any u32 can reach `OptionC::discriminant`, including bit-flips or partially-written trusted packages.

`into_option` panics on discriminants other than 0/1, and `validate_debug_info` is the only thing standing between untrusted bytes and the rest of the process. `try_into_option` already returns `InvalidOptionCError` — could the validate path use that and surface it as a normal `PackageDebugInfoError`?

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -809,106 +539,199 @@ impl Deserializable for DebugVariantInfo {
...
+        }
+    }
+}
 
-        let file_idx = source.read_u32()?;
-        let line_raw = source.read_u32()?;
-        let column_raw = source.read_u32()?;
-        let line = LineNumber::new(line_raw).unwrap_or_default();
-        let column = ColumnNumber::new(column_raw).unwrap_or_default();
+impl ByteReader for AlignedSliceReader<'_> {
```
huitseeker: This `ByteReader` impl doesn't override `max_alloc`, so it inherits the `usize::MAX` default. The `BudgetedReader` wrapping it caps the byte budget but not the in-memory expansion.

The result is that `read_bounded_len` in `utils-indexing` returns a `len` bounded only by remaining input bytes divided by `min_serialized_size`. For variable-length records like `DebugTypeInfo`, `min_serialized_size() == 1`, so `len` can equal the section byte length — and `IndexVec::read_from_bounded` (`utils-indexing/src/lib.rs:465`) then calls `Vec::with_capacity(len)`.

Concretely: a 100MB trusted debug section can cause `Vec::with_capacity` to reserve ~4GB for the types table before a single element is decoded, since each `DebugTypeInfo` is ~40 bytes. Same shape applies to the `nodes` table (~128x amplification).

Could `AlignedSliceReader` override `max_alloc` like the test reader at line 801 does, returning `(self.source.len() - self.pos) / element_size`?

crates/assembly/src/assembler/debuginfo.rs
```diff
@@ -1,129 +1,45 @@
...
 
-#[cfg(feature = "std")]
 impl SourcePathTrimmer {
     pub fn new(cwd: PathBuf) -> Self {
         let cwd = cwd.canonicalize().unwrap_or(cwd);
         Self { cwd }
     }
 
+    #[allow(unused)]
     pub fn trim_location(&self, mut location: Location) -> Location {
```
huitseeker: These two `trim_*` methods are dead now that path trimming moved into `PackageDebugInfo::trim_file_paths`. 

Delete?

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,924 @@
...
+impl<Src: SourceNodeIdMarker> DebugInfo<MastNodeId, Src> {
+    /// Merges package-owned source/debug metadata after a [`miden_core::mast::MastForest`] merge.
+    ///
+    /// [`miden_core::mast::MastForest::merge`] remains execution-only. This helper applies the
+    /// returned node mappings to package source/debug sections so callers can merge
+    /// `(MastForest, PackageDebugInfo)` pairs without reattaching debug metadata to the forest.
+    ///
+    /// This also merges the type, source-file, and function tables referenced by source-map
+    /// inline-call rows.
+    pub fn merge_source_debug<'a>(
```
huitseeker, line 716: This duplicates the type remapping logic in `MastForestFinalizer::remap_debug_tables`. `remap_debug_type` in `finalizer.rs` and `remap_debug_type_info` here need to stay in sync when `DebugTypeInfo` changes. The `type_id` remap issue below looks like one way the two paths can drift.
Could we share this remapping code?

I'm also curious about the need to expose this merging code externally - to, presumably, the compiler?

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,924 @@
...
+                                    forest_index,
+                                    location_idx: idx,
+                                },
+                            )
+                        })
+                        .transpose()?;
+                    debug_vars.push(DebugSourceVar {
+                        op_idx: row.op_idx,
+                        name_idx,
+                        type_id: row.type_id,
```
huitseeker: Coming back to this — commit `7a39f789b2` ("make DebugVar actually useful") changed how `type_id` is populated: producers now register the type into the debug-info type table and store the resulting `DebugTypeIdx` here. So `type_id` is no longer a placeholder, it's a live index into `debug_info.types`.

That makes the unremapped copy at `mod.rs:677` more consequential than when I first flagged it: if input package #1 has `type_id: Some(DebugTypeIdx(5))` referring to its own type table, and the merge shifts type indices (which it does at line 525 via `remapped_types`), the copied `type_id` now silently points at the wrong type. `recover_registered_type` in `types.rs` will then hand back whatever type happens to sit at that index in the merged table.

Same fix as before — `type_id` should go through `remapped_types.get(&row.type_id).copied()` like `name_idx` does two lines above.

crates/assembly/src/assembler.rs
bitwalker: @bobbinth ping

crates/mast-package/src/debug_info/mod.rs
bitwalker: I've consolidated the merging/remapping logic for debug info under `DebugInfo::merge_source_debug` and `DebugInfo::merge_tables_into` so that `DebugInfo` owns its own merging implementation.

Exposing this externally is just a natural artifact of `DebugInfo` being defined in a separate library than where the merging happens - you either have to expose internals of `DebugInfo` to allow merging to live elsewhere, which makes invariants harder to enforce, or you expose merging helpers like we do here. I think the latter is preferable, even if in practice the only implementation (currently) is in the assembler.

crates/mast-package/src/debug_info/builder.rs
```diff
@@ -0,0 +1,792 @@
...
+                let element_name = declared_ty.and_then(|t| match t {
+                    TypeExpr::Array(array) => match array.elem.as_ref() {
+                        TypeExpr::Ref(t) => Some(Arc::from(t.inner().as_str())),
+                        _ => None,
+                    },
+                    _ => None,
+                });
+                let element_name = element_name.map(|name| self.add_string(name));
+                let element_decl = declared_ty.and_then(|t| match t {
+                    TypeExpr::Ptr(p) => Some(p.pointee.as_ref()),
```
huitseeker: Looks like this branch was copied from the pointer case. In the `Type::Array` case, `declared_ty` is a `TypeExpr::Array`, so this match never passes the element's declared type into the recursive registration.

Should this match `TypeExpr::Array(array) => Some(array.elem.as_ref())` instead?

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,613 +1,338 @@
...
+        // aligned according to the required minimum alignment requirements of that data so that
+        // we can construct direct references to it during deserialization.
+        let mut output = Vec::<u8>::with_capacity(16 * 1024);
+
+        self.strings.write_into(&mut output);
+
+        output.write_u32(self.files().len().try_into().unwrap());
+        pad_to_align::<DebugFileInfo>(&mut output);
+        unsafe {
+            copy_slice_memory_to_output(self.files().as_slice(), &mut output);
```
huitseeker: Could this stay field-encoded? This copies host-endian integers into the package. An index written as `01 00 00 00` on a little-endian machine is read as `0x01000000` on a big-endian machine. `repr(C)` fixes field order, but not byte order, so this format is target-dependent.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -802,113 +532,217 @@ impl Deserializable for DebugVariantInfo {
...
+}
 
-        let file_idx = source.read_u32()?;
-        let line_raw = source.read_u32()?;
-        let column_raw = source.read_u32()?;
-        let line = LineNumber::new(line_raw).unwrap_or_default();
-        let column = ColumnNumber::new(column_raw).unwrap_or_default();
+impl ByteReader for AlignedSliceReader<'_> {
+    fn max_alloc(&self, element_size: usize) -> usize {
+        (self.source.len() - self.pos) / element_size
```
huitseeker: This still seems to allow the large up-front allocation from the earlier thread. `IndexVec::read_from_bounded` passes the minimum serialized size here, which is `1` for `DebugTypeInfo`, then calls `Vec::with_capacity(len)`. A 100 MB payload can therefore reserve about 4 GB before the first type tag is decoded. Could the initial capacity be capped by the in-memory element size, or grow as rows are validated?

crates/mast-package/src/debug_info/builder.rs
```diff
@@ -0,0 +1,792 @@
...
+// ================================================================================================
+
+impl<Exec: Idx, Src: Idx> DebugInfoBuilder<Exec, Src> {
+    /// Get a reference to the current state of the `DebugInfo` being built
+    pub fn debug_info(&self) -> &DebugInfo<Exec, Src> {
+        &self.debug_info
+    }
+
+    /// Get a mutable reference to the current state of the `DebugInfo` being built
+    pub fn debug_info_mut(&mut self) -> &mut DebugInfo<Exec, Src> {
```
huitseeker, line 197: I think this should stay crate-private or be replaced with narrow methods, to give us a shot at maintaining the construction invariants.

The builder keeps `string_indices`, `location_indices`, and `type_indices` beside this value. A caller can e.g. use this reference to run `trim_file_paths()`, which may append strings without updating `string_indices`; a later `add_string()` can then insert the same string again.

crates/assembly/src/mast_forest_builder/static_import.rs
```diff
@@ -438,30 +443,63 @@ impl MastForestBuilder {
...
-                    AssemblyOp::new(
-                        row.location.clone(),
-                        row.context_name.clone(),
-                        row.num_cycles,
-                        row.op.clone(),
-                    ),
-                )
+                let location_idx = row
+                    .location_idx
+                    .into_option()
```
huitseeker, line 478: `Package::debug_info()` validates `location_idx` before constructing a `StaticLibrary`, so `into_option()` is safe on the current package-loading path. 

Just mentioning it would be easy to use `try_into_option()`here—or documenting that invariant—to keep this code robust if another construction path is added later?

Nit, because this does not appear to be a currently reachable malformed-package panic.

crates/assembly/src/mast_forest_builder/pending_record.rs
```diff
@@ -1,111 +1,45 @@
...
+    /// Multiple source occurrences may point at the same [`MastNodeRef`] when they have identical
+    /// execution content but distinct source metadata.
+    pub(crate) struct SourceNodeRef;
+);
+
+impl SourceNodeIdMarker for SourceNodeRef {}
+
+newtype_id!(
+    /// Stable assembly-time reference to assembly operation metadata.
+    pub(crate) struct AsmOpRef;
```
huitseeker: `AsmOpRef` and `DebugVarRef` look orphaned after this PR.

Safe to delete both blocks?

huitseeker (commented):
Main items:
- Raw struct serialization depends on host endianness/layout.
- The allocation guard can still permit roughly 40× oversized up-front allocations.
- Mutable access to `DebugInfo` can desynchronize its interning caches.
- The array branch forwards the wrong declared element type.

The test looks good!

bitwalker:
> Main items:
> 
> * Raw struct serialization depends on host endianness/layout.

Endianness is a non-issue - we're not targeting POWER or other big-endian platforms, which are extremely niche at this point.

I have to keep reiterating this point, but - the layout is well-defined, aside from endianness (which as mentioned above, is little-endian on every platform we're even thinking of targeting or is in common use today), the layout is specified completely.

Obviously, by serializing/deserializing structs directly, their representation become part of the format specification - that's a given. The one case where there is some fragility, is with `Word`, as it is an array of `Felt`, and we're encoding their internal representation here. If we make changes to that internal representation, it can result in the format being broken. I'm considering that acceptable risk though, as `Felt` is very unlikely to change unless we change the field itself.

> * The allocation guard can still permit roughly 40× oversized up-front allocations.

It isn't clear to me how this can happen - `PackageDebugInfo` gets encoded with the _full_ size of the data, and we then place that buffer in memory and deserialize from it. When `read_bounded_len` calls `check_eor` with the deserialized length, it must not be larger than the remaining size of the `PackageDebugInfo` buffer, or that check will fail. This is also true of the `read_aligned_slice_of` calls.

Do you have a reproducer for this?

> * Mutable access to `DebugInfo` can desynchronize its interning caches.

How so? There are no mutable APIs that allow modifying its internal state, except for through `DebugInfoBuilder`.

crates/assembly/src/mast_forest_builder/pending_record.rs
bitwalker: Indeed

crates/assembly/src/mast_forest_builder/static_import.rs
bitwalker: I'm inclined to allow this to panic, since it would indicate that somehow we're allowing unvalidated/invalid debug info through into an area where it is assumed trusted/valid, and that is exactly where a panic is useful - so you can pinpoint the precise place where it failed, and how execution got there. 

I'd add a comment about it, but it isn't super clear to me where the best place is to document this invariant - we always extract statically-linked debug info from the package, which always validates it, so there aren't any logical places to document that package debug info is presumed valid (the entire assembler presumes that packages are trusted/valid anyway, most validation is incidental, rather than defensive). A comment on `statically_linked_package_debug_info` field is unlikely to be helpful if `into_option` panics - but maybe that's just me.

crates/mast-package/src/debug_info/builder.rs
bitwalker: I think that edge case is perfectly acceptable - the worst case scenario is you end up with debug info that has some duplicate data in it, and that isn't a correctness property required of `DebugInfo`.

The `DebugInfo` type provides no mutable methods that aren't designed explicitly for modifying it in ways that preserve its validity.

The alternative is to require constructing a `DebugInfoBuilder` for the sole purpose of calling `trim_file_paths`, which is an expensive operation.

I do agree that we must avoid providing mutable methods on `DebugInfo` that do _not_ guarantee that validity is preserved - all such methods must be implemented on `DebugInfoBuilder` instead.

crates/mast-package/src/debug_info/serialization.rs
bitwalker: I don't believe it is possible to actually hit this in practice - `PackageDebugInfo` is deserialized from a temporary buffer whose full size is known in advance, and `read_bounded_len` will fail if the deserialized length ever exceeds the size of that buffer, and APIs like `read_aligned_slice_of` and `read_slice` all check (via `check_eor` that the underlying reader has sufficient bytes of input for the requested length), so they would also fail early. `IndexVec` uses `read_bounded_len` before it allocates the `Vec`, so I don't believe it is possible to allocate a buffer larger than the bounds.

If you have a reproducer, I'll definitely make sure we close any such gap though.

crates/mast-package/src/debug_info/serialization.rs
bitwalker: We support exactly zero big-endian targets, and AFAIK, have no intention of supporting them. They are all but dead industry-wide at this point, except in very niche workloads. I think it is perfectly acceptable to require little-endian hosts for running Miden for now.

If it ever became necessary to support a big-endian target, we could do so without any breaking changes, simply by post-processing the deserialized data and reversing the bits of all integer fields contained in a struct that was serialized directly from memory. Expensive, but it would work. I'm willing to bet that eventuality never happens though.l

crates/mast-package/Cargo.toml
```diff
@@ -20,19 +20,22 @@ doctest = false
...
 miden-assembly-syntax.workspace = true
 miden-core.workspace = true
 miden-debug-types.workspace = true
+miden-utils-indexing.workspace = true
 
 # External dependencies
 log.workspace = true
 proptest = { workspace = true, optional = true }
 proptest-derive = { workspace = true, optional = true }
+hashbrown = "0.17"
```
bobbinth, line 37: We are using `hashbrown` here so that we could use the same hash maps in `std` and `no_std` context, right?

crates/assembly/src/assembler/debuginfo.rs
bobbinth, line 1: It may make sense to rename this file as AFAICT, it doesn't hold debug info structs any more. Maybe `path_trimmer.rs` would be more approprpirate?

Also, and maybe not for this PR, but it would be good to add some comments explaining what the path trimmer is actually for.

processor/src/errors.rs
bobbinth, line 1059: Not for this PR, but we should probably split this file into several files (e.g., by moving tests out).

processor/src/continuation_stack.rs
```diff
@@ -89,8 +89,7 @@ impl<F> Continuation<F> {
         }
     }
 
-    #[cfg(any(test, feature = "testing"))]
-    pub(crate) fn exec_node(&self) -> Option<MastNodeId> {
+    pub fn exec_node(&self) -> Option<MastNodeId> {
```
bobbinth, line 92: We are making these public because they are used somewhere downstream by the compiler?

crates/assembly/src/linker/mod.rs
```diff
@@ -763,7 +763,7 @@ impl Linker {
         source_library_commitment: Word,
         mast_root: Word,
         selected_root_id: MastNodeId,
-    ) -> Option<MastNodeId> {
+    ) -> Option<(MastNodeId, Arc<Path>)> {
```
bobbinth: Could we also update the doc comments to describe what the returned `Path` is for?

crates/mast-package/src/package/mod.rs
bobbinth, line 1: This is a huge file - would be good to split this up into logical components (not for this PR).

crates/mast-package/src/debug_info/types.rs
bobbinth, line 1: Same comment as above (and also not for this PR).

bobbinth (approved):
This is an very high-level review from me - but looks good! Thank you! I left some comments inline, but most of them are about generally improving code organization and maintainability and are not for this PR.

Also, my understanding is that after this PR, if a package exports two procedures with identical MAST roots but different fingerprints, we'd get an assembly error - is this correct?

bitwalker:
> Also, my understanding is that after this PR, if a package exports two procedures with identical MAST roots but different fingerprints, we'd get an assembly error - is this correct?

No - I've removed the check that causes assembly to fail if a dynamically-linked procedure would have ambiguous source roots. The caller has no ambiguity in terms of debug info, but the `External` node does not encode which execution node it actually cares about - so at runtime the link is lost, but this should not matter for execution purposes, it just means debugging for that call may be missing info, or refer to a different source  location than you expect. In the future we can use caller debug info to help disambiguate during execution to improve the precision of debugging - but that isn't done her yet.

crates/mast-package/Cargo.toml
bitwalker: Yes, exactly

crates/assembly/src/assembler/debuginfo.rs
bitwalker: I think we could probably move this entirely to where `DebugInfo` itself is defined - it doesn't need to live in the assembler.

processor/src/continuation_stack.rs
bitwalker: By the debugger, but yes, exactly

crates/assembly/src/linker/mod.rs
bitwalker: Actually this function can be removed now

crates/mast-package/src/package/mod.rs
bitwalker: Agreed

bobbinth:
One question I forgot to ask: did this have any meaningful effect on the processor execution times (when running w/o debug info)? I don't think it should, but wanted to double-check.

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,1045 @@
...
+                .map(|tid| {
+                    remapping.ty(tid).ok_or(DebugInfoTableRemapError::MissingType { type_idx: tid })
+                })
+                .transpose()?;
+            let file_idx = remapping.file(function.file_idx).ok_or(
+                DebugInfoTableRemapError::MissingSourceFile { file_idx: function.file_idx },
+            )?;
+            let function_idx = target.add_function(FunctionInfo {
+                mast_root: function.mast_root,
+                source_node: None.into(),
```
huitseeker, line 631: This sets every imported function to `None`, and that caller never applies `remapped_nodes` to the function rows. So adding source-node checks to `test_package_source_debug_merge_remaps_execution_nodes_without_collapsing_sources`; we can see that both fail with `None` instead of their new source IDs. 

`MastForestFinalizer::finalize_source_graph()` already does the missing `set_function_source_node` pass.

crates/mast-package/src/debug_info/mod.rs
```diff
@@ -5,7 +5,1045 @@
...
+                });
+            }
+        }
+        for function in self.functions.iter() {
+            if self.files.get(function.file_idx).is_none() {
+                return Err(DebugInfoTableRemapError::MissingSourceFile {
+                    file_idx: function.file_idx,
+                });
+            }
+            if let Ok(Some(tid)) = function.type_idx.try_into_option()
```
huitseeker: Could these `try_into_option()` errors be returned instead of ignored? `merge_tables_into()` calls this validation, then uses the panicking `into_option()` above. 

So we can panic by setting a decoded function's `type_idx` tag to 2, even though this API returns `Result` and says it validates before changing the target. 

An invalid-option error plus `try_into_option()?` would avoid the panic.

crates/mast-package/src/debug_info/serialization.rs
huitseeker: `check_eor` proves enough encoded bytes remain, but it does not bound the in-memory reservation by `size_of::<DebugTypeInfo>()`.

Reproducer at https://gist.github.com/huitseeker/62e10684360a41177564844e9ac003bc

crates/assembly/src/mast_forest_builder/static_import.rs
huitseeker: I was thinking of the `StaticLibrary::debug_info` field, pointing at the validation; but as mentioned this is a nit.

crates/mast-package/src/debug_info/serialization.rs
huitseeker: A simple fix is to annotate the `impl Serializable` and `impl Deserializable` with `#[cfg(target_endian = "little")]`. Or, if we can't be bothered, add this to the crate root:
```rust
#[cfg(not(target_endian = "little"))]
compile_error!("This crate requires a little-endian target.");
```

crates/mast-package/src/debug_info/builder.rs
huitseeker: Fair, given that the duplication is valid, I overstated this as a validity issue:

huitseeker:
@bobbinth That part is ok:


|Revision|          Five uncontaminated run means|
|-|-|
| Before f7d8549 |                       2.5297 ms|
| PR head d55c2a           |             2.5654 ms|
| Difference|                               +1.41%|


  I benchmarked only `blake3_1to1/execute_sync`

crates/mast-package/src/debug_info/types.rs
```diff
@@ -85,971 +135,262 @@ pub enum PackageDebugInfoMergeError {
...
+impl<T: Copy> Copy for OptionC<T> {}
 
-    for ty in &section.types {
-        output.add_type(remap_debug_type_info(forest_index, ty, &string_map, &type_map)?);
+impl<T: Clone> Clone for OptionC<T> {
+    fn clone(&self) -> Self {
+        match self.discriminant {
+            0 => Self::none(),
+            1 => Self::some(unsafe { self.payload() }.clone()),
+            invalid => Self::invalid(invalid),
```
huitseeker: This is undefined behavior (UB) (introduced in 13e71d5da8), and it is reachable through safe APIs. 

`PackageDebugInfo::read_from` can return an invalid `OptionC`, and `DebugInfo::clone()` recreates it here with uninitialized payload bytes. The serializer later forms a `&[u8]` over the whole `FunctionInfo` row and copies it. That slice includes the uninitialized payload, so the copy is UB and may write unrelated memory contents into the package. Could `invalid()` use `MaybeUninit::zeroed()` like `none()` so every byte copied by the raw serializer is initialized?

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,613 +1,342 @@
...
+        let roots = source.read_aligned_slice_of::<DebugSourceNodeId>(roots_len)?.to_vec();
 
-        let files_len = read_bounded_len(source, "debug_sources files", 1)?;
-        let files = source.read_many_iter(files_len)?.collect::<Result<_, _>>()?;
+        let error_messages_len = source.read_u32()? as usize;
+        let error_messages =
+            source.read_aligned_slice_of::<DebugErrorMessage>(error_messages_len)?.to_vec();
 
-        Ok(Self { version, strings, files })
+        Ok(PackageDebugInfo {
```
huitseeker, line 175: `data_len` consumes the whole envelope, but this inner reader is not checked after the last field. Could this reject when the inner reader still has trailing bytes? `DebugSourceNode::read_from` has the same gap.

crates/utils-indexing/src/lib.rs
```diff
@@ -419,6 +563,49 @@ mod tests {
     newtype_id!(TestId);
     newtype_id!(TestId2);
 
+    struct FixedBudgetReader {
```
huitseeker: Could this test use `BudgetedReader::new(SliceReader::new(&[]), 256)`? Its `max_alloc` already gives the same remaining-budget divided by element-size behavior, including zero-sized elements, so this reader is a not necessary. The instrumented reader in `mast-package` is different because that test records the requested element size.

huitseeker (commented):
I think there's a reachable UB in the `impl Clone for OptionC<T>`

crates/mast-package/src/debug_info/types.rs
bitwalker: It isn't undefined behavior to clone MaybeUninit - it implements the trait itself. The issue of serializing random bytes is something to address via `zeroed` though

crates/assembly/src/mast_forest_builder/finalizer.rs
```diff
@@ -270,27 +258,84 @@ impl MastForestFinalizer {
...
                 node,
-                pending_source_node.op_start,
-                pending_source_node.op_end,
+                pending_source_node.op_start as usize,
+                pending_source_node.op_end as usize,
             );
-            let inserted_id = finalized_nodes
-                .push(SourceNode::new(exec_node, children, op_start, op_end, asm_ops, debug_vars))
+            let inserted_id = debug_info
+                .add_node(SourceNode {
```
huitseeker, line 330: So this is replacing an external placeholder with a concrete non-leaf node, but it leaves the placeholder's zero-child source occurrence attached to the same node ref. 

Finalization emits that stale occurrence for the new `Join`/`Split`/`Call`/`Loop`, and package validation then rejects the child-count mismatch. 

This replacement should rebind stale source occurrences (with a non-leaf replacement test), I think. The existing replacement test only covers a basic block, where zero children still happens to be correct.

crates/assembly/src/project.rs
```diff
@@ -714,11 +807,22 @@ where
...
+        role: InterruptedTargetRole,
+    ) -> Result<ControlFlow<AssemblyInterrupted, ProjectSourceInputs>, Report> {
         let (provider, context) =
             self.get_provider_and_target_assembly_context(&project, target, profile)?;
 
-        let inputs = provider.provide_sources(&context)?;
+        let inputs = match provider.provide_sources_interruptible(&context)? {
+            ControlFlow::Break(_) => {
+                return Ok(ControlFlow::Break(AssemblyInterrupted {
+                    package: project.name().into_inner(),
```
huitseeker, line 818: I think you want `project.target_package_name(target)`.

An executable target named `runner` successfully produces package `demo:runner`, but an immediate provider interruption reports `AssemblyInterrupted.package == demo`.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -1,611 +1,354 @@
...
+        let roots = source.read_aligned_slice_of::<DebugSourceNodeId>(roots_len)?.to_vec();
+
+        let error_messages_len = source.read_u32()? as usize;
+        let error_messages =
+            source.read_aligned_slice_of::<DebugErrorMessage>(error_messages_len)?.to_vec();
 
-        let files_len = read_bounded_len(source, "debug_sources files", 1)?;
-        let files = source.read_many_iter(files_len)?.collect::<Result<_, _>>()?;
+        if source.has_more_bytes() {
+            let remaining_len = data_len.saturating_sub(data_len);
```
huitseeker: `data_len.saturating_sub(data_len)` is always zero, so this error cannot report the trailing data it just detected. 

Could this use the inner reader's actual remaining length and assert the count in tests? The same expression appears in `DebugSourceNode::read_from`.

huitseeker (commented):
Thanks! this looks good apart for a few warts mentioned inline. I have a proposal in https://github.com/0xMiden/miden-vm/pull/3430 which aims at simplifying the (unsafe) management of Rust rows (refactoring to zerocopy pod style). It has no performance downside.

crates/assembly/src/project.rs
bitwalker: We want the root package name here, as the target information that is also provided via `AssemblyInterrupted` can be used to cover the target-specific package name if we actually need it

crates/mast-package/src/debug_info/serialization.rs
bitwalker: This was a typo I mean to subtract the position in the stream

huitseeker (commented):
Thanks for these iterations!  I still think #3430 is a strict improvement in maintainability. Invariants move ot compiler checks:

- Generic `Copy` casts become zerocopy operations (and reject padding and invalid bit patterns at compile time).
- `OptionC` becomes a fully initialized `[u32; 2]`, no `MaybeUninit` logic and unsafe implementations.
- `DebugFunctionInfo` gets an explicit wire type, so field elements are validated instead of treating arbitrary bytes as a valid `Word`.
- Unsafe code is confined to four straightforward operations inside the aligned-buffer owner.

huitseeker (approved):
LGTM! Tough I'd prefer having $3430 in before merge, I can also submit it after

PR #3401, opened 2026-07-20
Use serde-wincode reexport for proof encoding

huitseeker:
The root cause is that `SerdeCompat` implements `wincode` traits for the `wincode` crate used by `serde-wincode`. If a Miden crate also imports `wincode` directly, Cargo can pick two `wincode` versions. Then proof code can ask for one `SchemaRead` or `SchemaWrite` trait while `SerdeCompat` implements the other one.

This change makes each current proof encoding path use `serde_wincode::wincode` with `SerdeCompat`. It removes direct `wincode` dependencies from `miden-prover`, `miden-verifier`, and `miden-test-utils`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.25.7 (2026-07-20)
+
+- Use the `wincode` crate re-exported by `serde-wincode` for proof encoding.
+
 ## v0.25.6 (2026-07-19)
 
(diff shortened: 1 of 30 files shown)


bitwalker (approved):
Looks good!

bobbinth (approved):
Looks good! Thank you!

If this is going into `main`, we should probably also add a changelog entry and bump up crate versions to v0.25.7.

PR #3402, opened 2026-07-20
ci: Add a package version gate

huitseeker:
This PR adds a package version gate for release paths and protected branch pushes.

- The gate runs on pushes to `main` and `next`. It also runs when a maintainer starts the release dry-run, starts the release workflow, or runs the gate by hand from GitHub Actions.
- For each selected publishable crate, it compares the local crate version with the latest version on crates.io. It fails if the local version is strictly older than crates.io.
- If the local version already exists on crates.io, the gate checks whether the package changed since that version was published. If it changed, the gate fails and asks for a version bump. If it did not change, the gate reports it and keeps going.
- When the local version is newer than crates.io, the check runs `cargo-semver-checks`. A semver failure means the public API changed more than the version bump allows.

Release operators can run the same check locally for a faster preview.

Run this to check every publishable crate:

```bash
scripts/check-package-release-plan.sh
```

Run this to check only selected crates:

```bash
scripts/check-package-release-plan.sh miden-core-lib,miden-vm
```

This PR also removes the old check that forced all publishable crates to use the same version. Publishable crates may now use different versions.

<details><summary>Notes</summary>

`cargo-semver-checks` can fail in this repo with `empty expression` before it compares the public API. That happens while it reads Cargo config on a local misconfigured system. To avoid that failure, the scripts run `cargo-semver-checks` from a temporary directory with isolated Cargo state. When they can find the release commit, they compare rustdoc JSON for HEAD with rustdoc JSON for that release.

The scripts skip packages without a normal library target because `cargo-semver-checks` cannot compare them.
</details>


bitwalker (commented):
Haven't reviewed this yet, but wanted to propose we have some way to bypass the check when doing a release in case it fails for spurious reasons, or we specifically want to do the release regardless (e.g. with the intent of yanking prior releases, or because we just did a release and found a bug immediately that required a breaking change). Obviously we'd want such a bypass to be explicit, and rarely used, but I think we likely do want one so we always have a path for doing a release "manually" (i.e. still using our release automation, just without any of the fallible gates meant to catch mistakes that could happen in the normal development/release cycle).

huitseeker:
@bitwalker fwiw, the set of situations where `cargo publish` would succeed but this script would not is non-empty but *very* small. The whole idea of this job is to shift left. But we do have an override, now.

bitwalker (approved):

PR #3404, opened 2026-07-21
AIR constraints compiler

Al-Kindi-0:
Adds a `miden-constraint-compiler` crate that captures the hand-written AIR constraints into one hash-consed IR graph, and two backends that consume it: a Rust-evaluator emitter (globally CSE'd `generated.rs`, regenerated and checked by a `regenerate-evaluator` bin) and an IR-driven ACE lowering. `MidenAir::eval` and ACE lowering now run the compiled pipeline; the hand-written definitions stay the auditable source of truth behind `eval_handwritten`. The final commit switches the emitter to an interleaved schedule (each value emitted at first use, asserts inline): same op counts, but values die immediately, which fixes a host-contention sensitivity on x86.
This results in improvements in the constraints evaluation phase between 20% to 30% depending on the VM and the architecture.

## API compatibility

This intentionally removes the unused generic base/extension-field parameters from the public ACE pipeline. The compiler IR and its Miden consumers rely throughout on the concrete `Felt`/`QuadFelt` pair, and there are no known callers using another field pair. The v0.28 changelog records this as a breaking API change.

## Benchmarks

8-proof recursive prove, stepwise ladder on `next` @ 526abebb3:

| platform | delta |
|---|---|
| Graviton c8g | −8.8% |
| x86 c7i | −7.1% quiet, −12.4% on contended hosts |
| Apple M4 | −5.9% |


adr1anh:
I only took a very brief look, but it might be worth investigating using `.dup()` from the latest P3, as well as trying to keep bindings as `Var` rather than `Expr`.

crates/ace-codegen/src/dag/lower_ir.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    let quotient = build_quotient_recomposition_dag::<Felt, QuadFelt>(&mut builder, layout);
+    let z_pow_n = builder.input(InputKey::ZPowN);
+    let one = builder.constant(QuadFelt::ONE);
+    let vanishing = builder.sub(z_pow_n, one);
+    let q_times_v = builder.mul(quotient, vanishing);
+    let root = builder.sub(acc, q_times_v);
+
+    builder.build(root)
```
huitseeker: Shouldn't we compact this DAG before returning, as the old lowering did? Without that step, dead interned nodes stay at the front: the new node-for-node test fails at node 1 (`Input(IsFirst)` versus `Constant(0)`). That also changes the digest-visible node order this function says it preserves.

crates/miden-constraint-compiler/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+repository.workspace = true
+rust-version.workspace = true
+edition.workspace = true
+
+[dependencies]
+miden-core = { workspace = true }
+miden-crypto = { workspace = true }
+
+[dev-dependencies]
+miden-air = { workspace = true }
```
huitseeker: Could we avoid closing the package cycle here? The release order is now `miden-air` → `miden-ace-codegen` → `miden-constraint-compiler` → `miden-air`, so the workspace dry run cannot stage the unpublished 0.28 crates. Current `next` passes, but this head fails while packaging `miden-air` because `miden-ace-codegen ^0.28` cannot be resolved. 

We should move these cross-crate tests or otherwise break the cycle.

huitseeker (changes requested):

crates/ace-codegen/src/dag/lower_ir.rs
Nashtare:
```suggestion
    let mut dag = builder.build(root);
    dag.compact();
    dag
```

Nashtare (approved):

crates/ace-codegen/src/dag/lower_ir.rs
Al-Kindi-0: done

crates/miden-constraint-compiler/Cargo.toml
Al-Kindi-0: done

huitseeker (approved):
This LGTM!

PR #3405, opened 2026-07-21
SVE2 packed-permutation C kernel for Poseidon2

Al-Kindi-0:
Adds a C-intrinsics SVE2 kernel for the WIDTH=12 two-lane packed Poseidon2 permutation, dispatched at `p3_permute_packed` and gated on aarch64+linux+sve2 in both `build.rs` and the call site; all other targets keep the existing packed path. C over Rust `asm!` because the win is interleaving the SVE2 S-box (elements 0–7) with scalar-GPR work (elements 8–11, internal-round chains) — the compiler schedules across both register domains, which `asm!` blocks prevent (best Rust-asm hybrid: 728 ns vs 662 ns). Same architecture as the existing RPO SVE kernels.

## Benchmarks

- Packed permutation: 874 → 662 ns/logical permutation (Neoverse V2).
- 8-proof recursive prove (stepwise ladder, `next` @ 526abebb3): −13.6% on Graviton c8g; 0.0% on x86 and M4 (kernel cfg-gated out there).


crates/crypto/src/hash/algebraic_sponge/poseidon2/mod.rs
```diff
@@ -45,9 +45,61 @@ fn p3_permute(state: &mut [Felt; STATE_WIDTH]) {
 ))]
 #[inline(always)]
 pub(super) fn p3_permute_packed(state: &mut [miden_field::PackedFelt; STATE_WIDTH]) {
+    #[cfg(all(target_arch = "aarch64", target_os = "linux", target_feature = "sve2"))]
```
huitseeker: I'm not sure if we need `target_os = "linux"`, technically. If Apple eventually adds support for sve2 in M6 that may work.

huitseeker (approved):
LGTM, thanks a lot!

adr1anh:
Codex reported a bug, but I can't test it locally. Would it be possible to check the following example?

```rust
#[test]
fn poseidon2_packed_permutation_double_wrap_regression() {
    const EPS: u64 = (1_u64 << 32) - 1;
    const INPUT: [u64; STATE_WIDTH] = [
        Felt::ORDER - 1,
        EPS,
        Felt::ORDER - 1,
        EPS,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
    ];

    check_with(
        Poseidon2Permutation256::apply_permutation,
        |state| Poseidon2Permutation256.permute_mut(state),
        |element, _lane| Felt::new_unchecked(INPUT[element]),
    );
}```

**Comment by Al-Kindi-0** [comment]
Thanks, this is a valid reproducer. The bug is not specific to SVE2, I ran the same test on an M4 MBP, and it also fails on the current `p3-goldilocks` 0.6.2. It comes from the single-correction Goldilocks addition used by the released Plonky3 NEON implementation and mirrored by the SVE2 C kernel. In your input, the initial MDS first computes

    (P - 1) + (2^32 - 1) = 2^64 - 1

twice, then reaches `add(u64::MAX, u64::MAX)`. The `u64` addition overflows, and adding the Goldilocks epsilon correction (`2^32 - 1`) overflows again. The old helper accounts for only the first carry, so the result is one epsilon short.

On the M4, the first output differs as follows:

    NEON packed:  0xd1579c46d79d8cfd
    scalar:       0x9a1730bf172ea856

Plonky3 has since fixed the NEON helpers by canonicalizing the second operand before addition/subtraction in commit [`84d46121`](https://github.com/Plonky3/Plonky3/commit/84d4612157110ce647d80a9a59336c7543b6205a), merged through Plonky3 [#1921](https://github.com/Plonky3/Plonky3/pull/1921). That commit's regression test directly exercises `add_neon(u64::MAX, u64::MAX)`. The fix landed after 0.6.2, which explains why Miden's currently locked dependency still fails on NEON.

The change we are about to push applies the equivalent canonicalization to the SVE2 C kernel. We verified the SVE2 failure before the change and agreement with the scalar permutation afterward on Graviton4 hardware. This patch does not update Miden's NEON dependency, so the new regression is SVE2-gated; NEON will be corrected when Miden picks up the upstream Plonky3 fix.

**Comment by Nashtare** [comment]
> Plonky3 has since fixed the NEON helpers by canonicalizing the second operand before addition/subtraction in commit [84d46121](https://github.com/Plonky3/Plonky3/commit/84d4612157110ce647d80a9a59336c7543b6205a), merged through Plonky3 https://github.com/Plonky3/Plonky3/pull/1921.

Ah! I thought there was another canonicalization bug that you had found that was still affecting latest P3. Yeah this fix is aimed to be released in 0.6.3 (ideally this week).

File: crates/crypto/src/hash/algebraic_sponge/packed.rs [hunk 2]
```diff
@@ -140,6 +134,33 @@ mod tests {
         }
     }
 
+    /// Checks the SVE2 packed permutation against the scalar implementation for a
+    /// state that triggers a second-order carry in the initial MDS.
+    #[cfg(all(target_arch = "aarch64", target_feature = "sve2"))]
```
adr1anh: Could this could be a test on all platforms?

Al-Kindi-0: This is the NEON case I flagged in [my earlier reply](https://github.com/0xMiden/miden-vm/pull/3405#issuecomment-5114686386):

> I ran the same test on an M4 MBP, and it also fails on the current `p3-goldilocks` 0.6.2.

Every other packed backend — x86 AVX2/512 and wasm SIMD128 — already canonicalizes the second operand before adding, so only the released NEON path still single-corrects (fixed upstream by [`84d46121`](https://github.com/Plonky3/Plonky3/commit/84d4612157110ce647d80a9a59336c7543b6205a), not yet in the dependency). So I've gated it to run on every packed backend except NEON:

    #[cfg(not(all(target_arch = "aarch64", not(target_feature = "sve2"))))]

PR #3406, opened 2026-07-21
Parallelize commitment/DEEP buffer initialization

Al-Kindi-0:
Parallel first-touch initialization for the large proving buffers in the commitment and DEEP phases:

- `par_filled_vec`: chunked parallel first-touch init for the multi-MB sponge-state buffers — `vec![default; n]` pays page faults and zeroing on one thread, a fixed serial cost that does not shrink with thread count. Falls back to a sequential fill without the `concurrent` feature (via `p3-maybe-rayon`).
- Lazy `scratch_states`: single-matrix trees (quotient, FRI rounds) never upsample, so the scratch buffer is allocated on first use.

Isolated paired A/B on Graviton4 (c8g.2xlarge, 8-proof recursive-verifier workload, T=8, interleaved runs, binaries differing only in this change): **−2.6% end-to-end** (median −447 ms), 8/8 pairs favoring the change, non-overlapping distributions.

**Rescoped after review** (thanks @Nashtare, @huitseeker): the earlier revision also restructured the DEEP column reduction into concurrent per-matrix partials. That change only fired when all matrices share one height — which the multi-AIR proving path never satisfies — and kept one full-domain vector per matrix live versus the existing bounded streaming accumulator. It has been dropped; the reduction code is untouched from `next`.


crates/lifted-stark/src/pcs/deep/prover.rs
```diff
@@ -220,31 +221,51 @@ impl<EF> DeepPoly<EF> {
...
+                                            }
+                                        })
+                                        .collect();
+                                    (matrix, packed)
+                                })
+                                .collect::<Vec<_>>()
+                        })
+                        .collect();
+
+                let mut partials: Vec<Vec<EF>> = per_matrix
```
huitseeker: Could we bound how many full-domain partials stay live here? 

`partials` keeps one `n`-element `Vec<EF>` per matrix until the final sum; with 256 supported same-height AIRs, the main and aux trees plus the quotient can create 513 partials, about 8 GiB at `n = 2^20`, while the old path used one 16 MiB accumulator. 

The feature choice is also off: the no-`concurrent` build pays the same cost with no parallel work. 

A fused parallel map/reduce or fixed-size batches would keep the overlap better, I think.

Nashtare (commented):
I'm unsure about the benefits of the changes.

First, this only fires under `uniform_height` (which we never have in practice), and the changes seem to yield relatively noisy benefits or no gains at all, for quite a blowup in memory consumption (because `partials` now holds many full domain vectors whereas we used to hold only 1 at time before)

crates/lifted-stark/src/pcs/deep/prover.rs
Al-Kindi-0: All three points hold, and reverted all those changes and kept only the parallel buffer init

Al-Kindi-0:
I have now reverted the changes and kept only the parallel buffer init which yields on Graviton4 a 2.6% e2e improvement  on the 8-proof recursive workload.

crates/lifted-stark/src/util/mod.rs
```diff
@@ -2,3 +2,28 @@
...
+/// across the pool. Falls back to sequential fill when built without the
+/// parallel feature.
+pub(crate) fn par_filled_vec<T: Copy + Send + Sync>(value: T, len: usize) -> alloc::vec::Vec<T> {
+    use p3_maybe_rayon::prelude::*;
+
+    let mut out = alloc::vec::Vec::with_capacity(len);
+    let spare = &mut out.spare_capacity_mut()[..len];
+    // 1 MiB chunks: coarse enough to amortize scheduling, fine enough to
+    // spread page faults evenly.
+    let chunk = (1 << 20) / size_of::<T>().max(1);
```
adr1anh: should we `const { assert!{size_of::<T>() > 0 }` to avoid the special handling of ZSTs?

crates/lifted-stark/src/util/mod.rs
```diff
@@ -2,3 +2,28 @@
...
+///
+/// For buffers in the tens of MB, sequential initialization is dominated by
+/// first-touch page faults and zeroing on a single thread (a fixed cost that
+/// does not shrink with thread count); chunked parallel writing spreads it
+/// across the pool. Falls back to sequential fill when built without the
+/// parallel feature.
+pub(crate) fn par_filled_vec<T: Copy + Send + Sync>(value: T, len: usize) -> alloc::vec::Vec<T> {
+    use p3_maybe_rayon::prelude::*;
+
+    let mut out = alloc::vec::Vec::with_capacity(len);
```
huitseeker: Could we use `p3_maybe_rayon::iter::repeat_n(value, len).collect()` here? Its parallel collector reserves once and fills separate slots in parallel, while the non-parallel build falls back to `core::iter::repeat_n`. That keeps parallel first-touch and removes this local `unsafe` block.

If the fixed 1 MiB chunks are needed for the perf gain, could we include an A/B against the native collector?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - `FastProcessor` `restore_call_state()` and `restore_context()` now return `OperationError::Internal` instead of panicking on empty stacks ([#3371](https://github.com/0xMiden/miden-vm/pull/3371), fixes [#3296](https://github.com/0xMiden/miden-vm/issues/3296)).
 
 - Opened the `LargeSmtForest` backend API for external implementations: made `LineageMutation::new` and `AppliedLineageMutation::new` public and added `LineageId::as_bytes` and `MutationSet::from_parts`.
+- Parallelized buffer initialization in the commitment and DEEP phases ([#3406](https://github.com/0xMiden/miden-vm/pull/3406)).
```
huitseeker: Could we drop “DEEP” from this entry now that `pcs/deep/prover.rs` was reverted? At this head, `par_filled_vec` is only used by LMCS tree construction, so this could say “Parallelized commitment buffer initialization and made the LMCS upsampling scratch buffer lazy.”

huitseeker (approved):
A few nits inline, but looks good overall

crates/lifted-stark/src/util/mod.rs
Al-Kindi-0: Addressed by removing the helper entirely, the two call sites now use `p3_maybe_rayon::iter::repeat_n`

crates/lifted-stark/src/util/mod.rs
Al-Kindi-0: Done, thanks for the suggestion

CHANGELOG.md
Al-Kindi-0: Done

PR #3407, opened 2026-07-21
Parallelize chiplet trace generation and overlap it with execution

Al-Kindi-0:
Trace building is a significant part of single-threaded work sitting between VM execution and the parallel prover. Three changes made in this PR, none changing the produced trace: 

1. The five chiplets build in parallel (`rayon::join` tree), each capping its own trace length;
2. The hasher's Poseidon2 permutation trace fills via `par_chunks_exact_mut` (each cycle writes a disjoint row range); 
3. `execute_and_build_trace_sync` streams hasher-chiplet requests through a channel and builds the hasher chiplet on a second thread while the VM is still executing. Controlled by `ExecutionOptions::with_overlapped_trace_build` (on by default in the sync prover; `no_std` keeps the sequential path). Also swaps the hasher's two dedup maps from `BTreeMap` to `hashbrown::HashMap`.

## Measured

8-proof recursive prove, stepwise ladder on `next` @ 526abebb3: Graviton c8g −0.5%, x86 c7i −0.4%, M4 ≈0. Small by design: the change targets the serial term of the prove (the part that does not shrink with threads), so the value grows with core count and concurrent proves rather than showing up in a single fixed-thread run. So the best way to benchmark this would be as function of number of cores increase against the baseline `next`.


processor/src/trace/trace_state.rs
```diff
@@ -1038,24 +1135,31 @@ impl HasherRequestReplay {
...
         node_id: MastNodeId,
-        expected_hash: Word,
+        basic_block_node: &BasicBlockNode,
     ) {
-        self.hasher_ops
-            .push_back(HasherOp::HashBasicBlock((forest_id, node_id, expected_hash)));
+        let expected_hash = basic_block_node.digest();
+        self.record_resolved(HasherOp::HashBasicBlock((forest_id, node_id, expected_hash)), || {
+            ResolvedHasherOp::HashBasicBlock((
+                basic_block_node.op_batches().to_vec(),
```
Nashtare: Not sure how heavy this is but is the deep clone really necessary? afaik this is never read by the hasher even

processor/src/trace/trace_state.rs
```diff
@@ -1066,17 +1170,44 @@ impl HasherRequestReplay {
...
+    /// Streamed replays yield nothing here: their ops were already delivered to the concurrent
+    /// builder.
+    pub fn into_resolved_ops<'a>(
+        self,
+        mast_forest_store: &'a [Arc<SparseMastForest>],
+    ) -> impl Iterator<Item = Result<ResolvedHasherOp, ExecutionError>> + 'a {
+        let ops = match self.sink {
+            HasherOpSink::Buffered(ops) => ops,
+            #[cfg(feature = "std")]
+            HasherOpSink::Streamed(_) => VecDeque::new(),
```
Nashtare: shouldn't this directly return an error instead? That would fail later on anyway right?

processor/src/trace/chiplets/hasher/trace.rs
```diff
@@ -373,17 +374,23 @@ pub(super) fn fill_poseidon2_permutation_trace(
...
-            perm_id_felt(perm_id),
-            Felt::new_unchecked(request.multiplicity),
-        );
-        row_idx += HASH_CYCLE_LEN;
-    }
+    // Each cycle is an independent permutation writing a disjoint row chunk;
+    // sequentially this loop is tens of milliseconds of single-thread work on
+    // large traces, so it runs in parallel.
+    rows[..request_count * HASH_CYCLE_LEN]
+        .par_chunks_exact_mut(HASH_CYCLE_LEN)
```
Nashtare: nit: not sure how impactful this is but couldn't we also parallelize also the padding loop below?

processor/src/fast/execution_api.rs
```diff
@@ -136,6 +136,54 @@ impl FastProcessor {
...
+        let mut tracer = ExecutionTracer::new_with_streamed_hasher(
+            self.options.core_trace_fragment_size(),
+            self.options.max_stack_depth(),
+            sender,
+        );
+
+        std::thread::scope(|scope| {
+            // Only the receiver crosses threads; execution (and the host) stay on this one.
+            let hasher = scope
+                .spawn(move || build_hasher_chiplet(receiver.into_iter().map(Ok), MAX_TRACE_LEN));
```
Nashtare: not sure here but aren't tracing spans thread local? if so, then I think this wouldn't appear on a profiling trace

processor/src/trace/trace_state.rs
Al-Kindi-0: It is read and the concurrent builder consumes exactly these op batches to hash the block. The clone is what lets the builder run without forest access: the replay would otherwise carry a borrow of the forest store across the thread.
Buffered (non-overlapped) mode never pays it, i.e., the closure only runs for streamed replays and the measured overlap win includes this cost.

processor/src/trace/chiplets/hasher/trace.rs
Al-Kindi-0: Good idea! Done!

processor/src/trace/trace_state.rs
Al-Kindi-0: I took a deeper look and managed to simplify things. The batch clone is now gone entirely

processor/src/fast/execution_api.rs
Al-Kindi-0: fixed

processor/src/trace/parallel/mod.rs
```diff
@@ -457,168 +488,229 @@ fn initialize_range_checker(
 
 /// Replays recorded operations to populate chiplet traces. Results were already used during
 /// execution; this pass only needs the trace-recording side effects.
+///
+/// The five chiplets are populated from disjoint replays, so they build in parallel. Each
+/// builder caps its own trace length at `max_trace_len`, and the combined length check runs
+/// once at the end, returning the same error an incremental check would. The bound before the
```
huitseeker: Could we keep this as a hard cap on the total chiplet trace? 

Each worker can now allocate up to `max_trace_len` before the combined check runs, so a program that grows several chiplets can allocate several times the `2^29` limit and OOM instead of returning `TraceLenExceeded`. 

This reverses the memory bound added in `eace5a10af`. See https://github.com/0xMiden/protocol/issues/3388 for the same sort of fight between local caps and a global cap. 

A shared budget or exact size checks before materializing the builders would preserve the parallel work without weakening the cap.

processor/src/trace/chiplets/mod.rs
```diff
@@ -20,19 +20,19 @@ use crate::{
 };
 
 mod bitwise;
-use bitwise::Bitwise;
+pub(crate) use bitwise::Bitwise;
 
 mod hasher;
-use hasher::Hasher;
+pub(crate) use hasher::Hasher;
```
huitseeker, line 26: Note: `Chiplets::new` is removed, ok in a major release.

processor/src/trace/trace_state.rs
```diff
@@ -1066,17 +1171,51 @@ impl HasherRequestReplay {
...
+                let node = forest
+                    .get_node_by_id(node_id)
+                    .ok_or(ExecutionError::Internal("invalid node ID in hasher replay"))?;
+                let MastNode::Block(basic_block_node) = node else {
+                    return Err(ExecutionError::Internal(
+                        "expected basic block node in hasher replay",
+                    ));
+                };
+                Ok(ResolvedHasherOp::HashBasicBlock((
+                    basic_block_node.op_batches().iter().map(|batch| *batch.groups()).collect(),
```
huitseeker: Could the buffered replay keep borrowing `op_batches()` and reserve the owned `Vec` for streamed messages? 

This collects and copies a vector on every basic-block visit, even when `hash_basic_block` immediately hits its memoized digest. A 1,000,000-iteration loop took 354–369 ms here versus 316–332 ms at the base, confirmed in reverse order. 

The async prover and the sync opt-out always use this buffered path, so they regress.

huitseeker (commented):
Directionally good, though I suggested a few improvements inline.

processor/src/trace/chiplets/mod.rs
Al-Kindi-0: Agreed and it is intentional

processor/src/trace/trace_state.rs
Al-Kindi-0: Addressed. Resolved basic-block data now has separate borrowed and owned representations: buffered replay borrows `op_batches()` from the finalized MAST forest, so memoized basic-block visits allocate and copy nothing, while streamed replay retains an owned `Vec` because it crosses the thread boundary

processor/src/trace/parallel/mod.rs
Al-Kindi-0: Addressed without adding per-operation synchronization. Before starting the builders, `initialize_chiplets` computes the exact non-hasher row total from replay metadata. The hasher receives only the remaining allowance, an oversized prebuilt hasher is rejected before entering `rayon::join`, and a `debug_assert_eq!` checks the metadata total against the materialized trace across the test suite.

huitseeker (approved):
Super, LGTM!

PR #3411, opened 2026-07-21
ci: speed up test feedback and remove duplicate crypto tests

huitseeker:
# Summary

- Reuse Cargo's built `miden-vm` binary in CLI tests (rather than run escargot on each).
- Give each CLI test its own temporary directory so the tests can run together.
- Stop local tests after the first failure, while CI still reports every failure.
- Remove 774 crypto tests repeated by a second CI job 🤦 .
- Keep the `persistent-forest` and lifted-STARK test runs.

## Why

Each CLI test started Cargo again to build `miden-vm`. This work ran ten times and could wait on Cargo's build lock. Shared output paths also forced the tests to run one at a time.

`make test` also used CI settings on developer machines. It kept starting tests after a failure instead of returning sooner.

The main workspace job already runs the ordinary crypto suite. The crypto job ran the same 774 tests again.

## Effects

Local tests now stop scheduling work after the first failure. GitHub Actions selects the CI settings directly, so CI still runs to the end and reports every failure.

CLI tests no longer start nested Cargo processes or leave shared files in the repository. New CLI tests must keep their files inside their temporary directory to remain safe when run together.

The main workspace job still runs ordinary crypto tests for VM, crypto, and workspace changes. The crypto job still covers the `persistent-forest` and lifted-STARK settings outside the standard suite.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,14 +3,3 @@
 fail-fast = false
 # Put a bound on hanging/slow tests
 slow-timeout = { period = "10m", terminate-after = 1, grace-period = "15s" }
-
-[test-groups]
-serialized = { max-threads = 1 }
-
-[[profile.default.overrides]]
-filter = 'test(cli_test)'
-test-group = 'serialized'
-
-[[profile.ci.overrides]]
-filter = 'test(cli_test)'
-test-group = 'serialized'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,9 +52,9 @@ jobs:
       - name: Install rust
         run: rustup update --no-self-update
       - name: Build tests
-        run: make test-build
+        run: make test-build NEXTEST_PROFILE=ci
       - name: test
-        run: make test
+        run: make test NEXTEST_PROFILE=ci
 
   crypto-test:
     name: crypto test on warpbuild
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,17 +1367,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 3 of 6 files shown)


bitwalker (commented):
I almost always prefer to see all the failures rather than stop on the first one, especially when I expect failures or am evaluating the damage from some change. It's pretty rare that I want to triage tests one at a time without being surgical about what test I'm running - is there a particular advantage to making fail-fast the default behavior? One can always CTRL-C to kill the tests early too.

Alternatively, maybe we can control this with a variable or provide a separate target for the fail-fast behavior?

huitseeker:
@bitwalker If you go through nextest for running tests, this is standard configuration unaffected by this PR. If you go through the `Makefile`, this is already exposed through a variable *and* that variable is exercised by CI, as you can see in the PR: `make test NEXTEST_PROFILE=ci`

bitwalker:
I basically always use `make test`, or `make test-<project>`. In the compiler repo our `cargo make test` task sets `--no-fail-fast`, and for some reason I thought that was also being set for `make test`, but apparently not.

I would recommend we set `--no-fail-fast` by default here, but it's obviously a subjective thing. Remembering to set `NEXTEST_PROFILE=ci` if you want `--no-fail-fast` isn't exactly intuitive, but 🤷

bitwalker (approved):

huitseeker:
@bitwalker I'm happy to take edits! I think the key here is that the number of tests run in this workspace has grown by a lot after #3366 and using `--no-fail-fast` as default in _interactive_ development is a liability. But if you want a `make patient-test` or something that's ok by me.

PR #3412, opened 2026-07-21
Normalize LogUp running sums

adr1anh:
## Summary

For row contribution `t[r]`, let:

```text
sigma = sum_r t[r]
mu    = sigma / n
```

The accumulator is constrained cyclically by:

```text
a[(r+1) mod n] = a[r] - mu + t[r]
```

Summing over all rows cancels the accumulator terms and yields:

```text
n * mu = sigma
```

For multiple AIRs, the external check becomes:

```text
sum_k (n_k * mu_k) + B = 0
```

This centered recurrence is the additive coboundary construction from §3, “A cohomological sumcheck argument,” of [Darlin: Recursive Proofs using Marlin](https://eprint.iacr.org/2021/930), applied to `p[r] = t[r] - mu`.

This avoids closing the full LogUp expression with the last-row selector, which would add one degree. The old implementation avoided that cost by requiring no lookup activity on the last row; the cyclic construction keeps the degree saving without that restriction.

## Performance

The prover performs one additional reduction over the row totals and one division by `n` before the sequential accumulator scan. The reduction uses the existing parallel-iterator shim. Native verification adds one scalar multiplication per AIR, while recursive verification derives proof-order trace-length weights and scales both extension-field coordinates.

There is no increase in auxiliary trace width, committed-value count, or proof size. Against the current three-AIR `next` architecture, removing the terminal constraints reduces every generated proof-order ACE circuit from 5208 to 5176 gates.

## Testing

- 102 `miden-air` tests
- 1561 `miden-processor` tests
- 261 `miden-precompiles-prover` tests
- 362 `miden-vm` tests, including equal/divergent-height recursive verification and precompile prove/verify
- `make check-constraints`
- warnings-as-errors Clippy and nightly rustfmt checks


air/src/lib.rs
```diff
@@ -847,10 +708,12 @@ impl<EF: ExtensionField<Felt>> MultiAir<Felt, EF> for MidenMultiAir {
...
-    /// Cross-AIR LogUp closure: the sum of every committed aux final plus the
-    /// per-trace boundary corrections must vanish. `aux_inputs` carries the program hash and
-    /// deferred root (consumed by the core boundary) followed by the kernel digests
-    /// (consumed by the chiplets boundary).
+    /// Cross-AIR LogUp closure: each AIR commits `sigma_prime_i = sigma_i / n_i`, so the
+    /// trace-length-weighted sum `sum_i(n_i * sigma_prime_i)` plus the per-trace boundary
+    /// corrections must vanish. Boundary corrections are added once and are not trace-length
+    /// scaled. `aux_values` and `log_trace_heights` are parallel in instance order. `aux_inputs`
+    /// carries the program hash and transcript state followed by kernel digests.
+    /// digests.
```
Nashtare:
```suggestion
    /// carries the program hash and transcript state followed by kernel digests.
```

Nashtare (approved):
LGTM, thanks!

Al-Kindi-0 (approved):
LGTM, and this is much more robust than the dead last row solution, thank you!

Al-Kindi-0:
One thing I just remembered, should we cite the Darling paper ?

air/src/lib.rs
adr1anh: Good catch—this duplicate was removed during the rebase, with the wording updated to the current deferred-root terminology.

adr1anh:
Yes—this is the additive coboundary construction from §3, “A cohomological sumcheck argument,” of [Darlin: Recursive Proofs using Marlin](https://eprint.iacr.org/2021/930), applied to `p[r] = t[r] - mu`. I added the citation beside the recurrence in `air/src/lookup/aux_builder.rs` and to the PR summary. Thanks for the reminder!

PR #3413, opened 2026-07-21
feat: open LargeSmtForest backend API for external implementations

JereSalo:
## Description

The `Backend` trait for `LargeSmtForest` is public but cannot be implemented outside this crate, because `compute_mutations` and `apply_mutations` must return `LineageMutation` and `AppliedLineageMutation` values whose fields are private and whose constructors are `pub(crate)`. This PR makes both constructors public and adds a `LineageId::as_bytes` getter so backends can persist lineage identifiers.

It also adds a `MutationSet::from_parts` constructor. An external backend has to produce the forward and reverse mutation sets carried by `AppliedLineageMutation`, and since `MutationSet` fields are private the only way to build them today is to load the full tree into memory and run `Smt::compute_mutations` / `apply_mutations_with_reversion`, which makes update cost scale with tree size instead of change size. With the constructor, a backend can compute mutations against its persisted tree data (affected leaves and stored subtree nodes only) and assemble both sets directly, the same way the built-in RocksDB backend does internally. In the SQLite backend prototype this took a single-entry committed update on a 1,000,000-entry tree from ~30 s to ~1 ms.

We need this change because in 0xMiden/rust-sdk#2315 we want to use the LargeSmtForest but with an SQLite backend. And it's very likely that we'll want to do the same in the web SDK with IndexedDB.

In order to make use of this change we should also bump miden-crypto, so it'd be good to have this in 0.28.1 since miden-protocol currently pins `^0.28`.


JereSalo:
Ideally we would like this in crypto 0.28.1 so we can get to use this soon. Since crypto [has been moved](https://github.com/0xMiden/crypto/pull/1097) to this repository I'm not entirely sure how crypto releases are managed. I pointed this PR to `next`, but if you need me to make any changes regarding that let me know.

Al-Kindi-0 (approved):
LGTM

huitseeker (approved):
This works! This will get published as 0.28.1 and does not need a major bump.

huitseeker:
@JereSalo This will normalize as soon as we release the VM at 0.28, but what you did is just fine.

JereSalo:
Thanks for the review. I had to add one more little change that I noticed was necessary while implementing the `LargeSmtForest` Backend in https://github.com/0xMiden/rust-sdk/pull/2333. It's just a constructor. I updated the PR description accordingly. Let me know what you think.

PR #3414, opened 2026-07-21
Shrink PR 3293 to the ECDSA K256 Keccak production path

huitseeker:
This PR is a small incremental / experimental shrink of `precompiles/vm-precompiles-prover` (#3293). These commits are a bit controversial, so may be tweaked.

- It keeps the path we need for Ethereum ECDSA verification in the VM. 
- This branch removes slightly broad changelog entries that describe the whole deferred precompile redesign as one release unit.
- It also drops the broad `bootstrap`, `eval`, and `integration` prover test files from that branch. The PR still contains prover tests. This replaces the broad end-to-end test surface with more focused tests.
- It leaves #3344 out of scope. This branch does not reimplement the generated ACE relation digest work.
- The shrink keeps the MSM fix because the latest review discussion only removes the Straus-specific arity concern. The duplicate-base concern still applies.  The ECDSA wrapper now handles that valid case. The generic curve precompile and deferred prover also reject MSM terms with zero scalars or duplicate canonical points.


adr1anh:
Thanks, I think the current shape makes sense for this PR.

The zero-scalar and duplicate-canonical-point checks are stricter than the mathematical definition of an MSM, but they are useful compatibility guards for the current deferred witness builder. Relaxing only the native evaluator would let execution register and log an MSM that witness generation cannot currently lower or prove. I would rather keep the execution and proving surfaces aligned here, then relax both atomically.

For the ECDSA path, the practical cases look covered:

- `Q = G` is handled by combining the two coefficients before scalar multiplication.
- For a valid signature, `U2 = r/s` is nonzero.
- If `Q = G` and `U1 + U2 = 0`, the verification point is the identity, so the signature cannot be valid with `r != 0`.
- The remaining theoretical valid case is `Q != G` with `U1 = 0`. Since the prehash is computed internally rather than supplied through the public API, this requires finding a `MSG` whose Keccak digest is exactly `0` or the secp256k1 scalar modulus `n`. That is not a practical malicious-input vector.

An invalid signature can deliberately cause cancellation, but under the current checks registration returns an execution error before deferred witness generation. It does not leave us with an accepted but unprovable deferred state.

I therefore do not think we need to broaden this PR. I opened #3416 to support zero coefficients and repeated canonical bases end-to-end. Once that lands, we can remove these restrictions and the ECDSA-specific `Q = G` workaround.

The only framing distinction I would preserve is that these inputs are **currently unsupported by the witness builder**, rather than mathematically invalid MSM inputs.

adr1anh (approved):
LGTM, see my previous comment for a followup.

precompiles-prover/src/tests/eval.rs
Nashtare, line 1: While I agree that bootstrap.rs and integration.rs were mostly redundant, I feel like removing this module is somewhat a step back, as it had several corruption tests out of which only one remained I believe?

PR #3418, opened 2026-07-23
Improve untrusted MAST overspec warning

### LINKED ISSUE #3425: MAST deserialization logs UntrustedMastForest overspecification at ERROR on a successful path
Running a `miden client` CLI command prints ERROR-level lines from MAST deserialization even though the command succeeds:

```
ERROR miden_core::mast::serialization: UntrustedMastForest expected HASHLESS input; supplied artifact includes wire node hashes, and validation will recompute them and require them to match
```

The command succeeds, so surfacing this at ERROR is misleading — it would probably read better at debug/trace, or handled quietly.

Full report: https://github.com/0xMiden/feedback/issues/149.

huitseeker:
~~This PR changes the untrusted MAST forest overspec log from `error!` to `warn!` and adds caller location to the warning. It also aligns package trusted readers with their names.~~

This PR hardens package deserialization across trust boundaries.

Untrusted package reads now validate the embedded MAST forest and decode package `DebugInfo` within fixed bounds. They validate that debug info against the package and retain it only after that succeeds.

Trusted package reads now mean what the name says. They trust the embedded MAST and manifest links. They also preserve package debug sections while skipping the cross-checks meant for hostile input.

The MAST warning is narrower now. If an untrusted MAST forest includes wire hashes, we log `warn!` with the caller location. The input is still accepted and validated by recomputing hashes. The warning is about wasted untrusted bytes, not corruption.

Local registry cache reads now use trusted package reads for cache reload and repair paths, including legacy artifact checks. The registry still validates package files passed to `publish(path)` before those bytes enter the artifact store. This keeps cache serialization unchanged and avoids routing same-domain cache bytes through the untrusted MAST warning path.

~~`Package::read_from` remains the untrusted reader for packages where debug sections are not needed. It validates the MAST and drops package debug sections.~~

`Package::read_from` and `Package::read_from_bytes` are the untrusted readers. They validate MAST and package debug info before returning the package.

~~`Package::read_from_untrusted` and `Package::read_from_bytes_untrusted` are new. They validate the MAST and package debug sections, then keep debug info.~~

No separate untrusted reader was added. The existing `Package::read_from` and `Package::read_from_bytes` keep that role.

~~`Package::read_from_unchecked` and `Package::read_from_bytes_unchecked` remain the way to trust bytes completely.~~

`Package::read_from_unchecked` and `Package::read_from_bytes_unchecked` were removed. Use `Package::read_from_trusted` and `Package::read_from_bytes_trusted` for same-domain trusted package bytes.

The DebugInfo path now rejects oversized payloads, oversized string and type tables, oversized strings, invalid references, bad spans, invalid layouts, and unsafe lookup shapes. Fuzz targets and regression seeds cover those hostile inputs.

Validation:

- `make check`
- `make test-fast`
- `make lint`
- `make check-features`

Roborev:

- `31284` passed with no issues.

### DIFF
--- a/core/src/mast/untrusted.rs
+++ b/core/src/mast/untrusted.rs
@@ -1,4 +1,5 @@
 use alloc::vec::Vec;
+use core::panic::Location;
 
 use super::{AdviceMap, MastForest, MastForestError, serialization};
 use crate::serde::{BudgetedReader, ByteReader, DeserializationError, SliceReader};
@@ -76,6 +77,18 @@ impl UntrustedMastForestReadOptions {
 }
 
 impl UntrustedMastForest {
+    /// Deserializes an [`UntrustedMastForest`] from a byte reader.
+    ///
+    /// Note: This method does not apply budgeting. For untrusted bytes, prefer
+    /// [`read_from_bytes`](Self::read_from_bytes) or
+    /// [`read_from_bytes_with_options`](Self::read_from_bytes_with_options).
+    #[track_caller]
+    pub fn read_from_reader<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let caller = Location::caller();
+        serialization::read_untrusted_with_flags_and_caller(source, caller)
+            .map(|(forest, _flags)| forest)
+    }
+
     /// Validates the forest by checking structural invariants and recomputing all node hashes.
     ///
     /// This method performs a complete validation of the deserialized forest:
@@ -142,6 +155,7 @@ impl UntrustedMastForest {
     /// // Validate before use
     /// let forest = untrusted.validate()?;
     /// ```
+    #[track_caller]
     pub fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
         Self::read_from_bytes_with_options(bytes, UntrustedMastForestReadOptions::default())
     }
@@ -151,16 +165,20 @@ impl UntrustedMastForest {
     /// The wire byte budget limits wire-driven parsing and collection pre-sizing. The validation
     /// helper-allocation budget is derived from that wire budget and caps tracked hashless helper
     /// allocations such as digest slot tables and rebuilt digest tables.
+    #[track_caller]
     pub fn read_from_bytes_with_options(
         bytes: &[u8],
         options: UntrustedMastForestReadOptions,
     ) -> Result<Self, DeserializationError> {
+        let caller = Location::caller();
         let wire_byte_budget = options.wire_byte_budget(bytes.len());
         let mut reader = BudgetedReader::new(SliceReader::new(bytes), wire_byte_budget);
-        let (forest, _flags) = serialization::read_untrusted_with_flags_and_allocation_budget(
-            &mut reader,
-            options.validation_allocation_budget(wire_byte_budget),
-        )?;
+        let (forest, _flags) =
+            serialization::read_untrusted_with_flags_allocation_budget_and_caller(
+                &mut reader,
+                options.validation_allocation_budget(wire_byte_budget),
+                caller,
+            )?;
         if reader.has_more_bytes() {
             return Err(DeserializationError::InvalidValue(
                 "extra bytes after MastForest payload".into(),
--- a/crates/package-registry-local/src/lib.rs
+++ b/crates/package-registry-local/src/lib.rs
@@ -297,7 +297,7 @@ impl LocalPackageRegistry {
         let Ok(bytes) = fs::read(path) else {
             return false;
         };
-        let Ok(loaded) = MastPackage::read_from_bytes(&bytes) else {
+        let Ok(loaded) = MastPackage::read_from_bytes_trusted(&bytes) else {
             return false;
         };
         let actual_version = Version::new(loaded.version.clone(), loaded.digest());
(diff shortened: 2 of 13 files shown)


crates/project/src/dependencies/graph.rs
```diff
@@ -775,8 +778,8 @@ impl<'a, R: PackageRegistry + ?Sized> ProjectDependencyGraphBuilder<'a, R> {
     ) -> Result<CollectedDependencyNode, Report> {
         let path = path.canonicalize().map_err(|error| Report::msg(error.to_string()))?;
         let bytes = std::fs::read(&path).map_err(|error| Report::msg(error.to_string()))?;
-        let package = MastPackage::read_from_bytes_trusted(&bytes)
-            .map_err(|error| Report::msg(error.to_string()))?;
+        let package =
+            MastPackage::read_from_bytes(&bytes).map_err(|error| Report::msg(error.to_string()))?;
```
bitwalker, line 782: I think this should use `read_from_bytes_trusted` - implicitly the dependency graph is operating on trusted packages (they are referenced either explicitly by the user, assembled on-demand and cached, or are loaded from a user-provided registry).

This is exactly the sort of issue I was running into - some code might defensively say "I'm not sure if I should assume trust, so I'm going to use the safer choice", but this would result in errors getting logged. Now that those errors are warnings, it's marginally better, but it's still essentially out of the control of the caller to say "this package is trusted", and I think it's fair to assume packages are trusted here.

crates/package-registry-local/src/lib.rs
```diff
@@ -224,7 +227,7 @@ impl LocalPackageRegistry {
     ) -> Result<PublishedPackage, LocalRegistryError> {
         let package_path = package_path.as_ref();
         let bytes = fs::read(package_path).map_err(LocalRegistryError::IndexRead)?;
-        let package = MastPackage::read_from_bytes_trusted(&bytes).map_err(|error| {
+        let package = MastPackage::read_from_bytes(&bytes).map_err(|error| {
```
bitwalker, line 231: This is going to result in packages in the registry having no debug info, no? That is definitely not desirable.

crates/mast-package/src/package/serialization.rs
```diff
@@ -253,19 +227,22 @@ impl Package {
...
+            PackageManifest::read_from_trusted(source, &mast)?
+        };
 
         // Read custom sections
         let mut sections = Vec::<Section>::read_from(source)?;
         if !debug_sections_trusted && sections.iter().any(|section| section.id.is_debug()) {
-            log::warn!(
-                "Package read ignored debug sections from an untrusted artifact; use Package::read_from_trusted for local cache/debug reads"
-            );
+            log::warn!("Package read ignored debug sections from an untrusted artifact");
```
bitwalker: It's unclear to me why we treat debug info  like this? It should just be another custom section, just one that various built-in facilities recognize - in other words, what is special about the debug info section that warrants stripping it out?

core/src/mast/untrusted.rs
```diff
@@ -76,6 +77,18 @@ impl UntrustedMastForestReadOptions {
...
 
 impl UntrustedMastForest {
+    /// Deserializes an [`UntrustedMastForest`] from a byte reader.
+    ///
+    /// Note: This method does not apply budgeting. For untrusted bytes, prefer
+    /// [`read_from_bytes`](Self::read_from_bytes) or
+    /// [`read_from_bytes_with_options`](Self::read_from_bytes_with_options).
+    #[track_caller]
+    pub fn read_from_reader<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let caller = Location::caller();
```
bitwalker, line 87: I suspect this is going to point to one of the `Package` methods in the majority of cases where a log message is emitted - we may want to have some way of propagating the caller information from there too

bitwalker (commented):
Looks good, I do have a few questions though (see comments)

crates/project/src/dependencies/graph.rs
huitseeker: Updated to use a narrower reader than `read_from_bytes_trusted`: it validates MAST and manifest data, but preserves trusted local debug sections.

crates/package-registry-local/src/lib.rs
huitseeker: Fixed. Registry publish and load now use the validating debug-preserving package reader.

core/src/mast/untrusted.rs
huitseeker: Added `#[track_caller]` on the package reader entry points too, so the warning now points at the package reader caller instead of only the MAST internals.

crates/mast-package/src/package/serialization.rs
huitseeker: `Package::read_from` stays the untrusted reader. It validates MAST and drops debug sections because as of today we do not know how to verify debug info adversarially (what sort of resource exhaustion could it create in any of its downstream processing, not limited to deserialization?). Keeping it would make unverified debug data look trusted.

bitwalker: That feels to me like a job for `Package::debug_info`, not `Package::read_from` - we're not deserializing the debug section here, and at this point it is no different than any other section.

The trust boundary here is really about the package metadata and the MAST - the parts it actually has to process when this is called. By conflating the two we're stripping out one of the most valuable pieces of the package _silently_.

bitwalker (changes requested):
I think we should stop stripping debug info when deserializing `Package` via `read_from` (i.e. the untrusted path), and instead punt that responsibility to `Package::debug_info` when we actually deserialize that data.

Currently, we're unintentionally stripping debug info in various places, silently, and it will be super confusing.

melnikga:
Related: #3425  same diagnostic, reported from the miden client CLI side

bitwalker (approved):
Approving under the assumption that package debug info is retained as a result of #2460

PR #3422, opened 2026-07-24
Recursive verifier API

### LINKED ISSUE #3250: Define the recursive verifier API for structured Miden VM proof verification
### Background

The current MASM recursive verifier already verifies a Miden VM STARK proof, but its interface is still shaped around STARK verifier internals rather than the VM-level verification contract we want callers to use.

Today, the entrypoint is `exec.vm::verify_proof` in `crates/lib/core/asm/sys/vm/mod.masm`. It expects proof-shape data on the operand stack and reads the rest of the verifier data from advice. Public input handling is delegated to `crates/lib/core/asm/sys/vm/public_inputs.masm`, which loads fixed public inputs, streams the kernel procedure digest list, absorbs these values into the verifier transcript, and computes the kernel-ROM LogUp reduction required by the AIR.

That implementation is the right low-level machinery, but we should define a clearer API boundary on top of it. Callers should not need to understand the flat STARK public input layout, the advice stack ordering, or the internal kernel-ROM reduction.

### Proposed Native API

The high-level verifier should expose one safe entrypoint:

```rust
fn verify(
    proof: MidenProof,
    inputs: VmInputs,
) -> Result<(), VerificationError>
```

where:

```rust
struct VmInputs {
    program_digest: Digest,
    kernel_commitment: Digest,
    stack_inputs: StackInputs,
    stack_outputs: StackOutputs,
}
```

`verify()` should verify the full computation. The proof object may internally contain different forms of deferred evidence, but callers should not have to care which form is used.

The proof type may conceptually contain:

```rust
struct MidenProof {
    vm_stark_proof: VmStarkProof,
    deferred: DeferredProof,
}
```

where `DeferredProof` may represent:

- no deferred work,
- raw deferred state wire,
- a precompile VM proof,
- a future batch/aggregate deferred proof form.

The exact enum shape does not need to be settled here, but the proof type must carry enough deferred evidence for `verify(proof, inputs)` to be safe.

### Mid-Level Verification Primitives

The high-level verifier should be built from smaller primitives.

```rust
fn verify_vm_proof(
    vm_stark_proof: VmStarkProof,
    inputs: VmInputs,
    deferred_commitment: DeferredCommitment,
) -> Result<(), VerificationError>
```

This verifies only the Miden VM STARK proof against structured VM inputs and the deferred commitment. It does not verify the deferred work itself. This primitive is always called after obtaining a deferred commitment from one of the deferred verification paths.

```rust
fn verify_precompile_proof(
    precompile_stark_proof: PrecompileStarkProof,
) -> Result<DeferredCommitment, VerificationError>
```

This verifies a precompile VM proof and returns the deferred commitment it proves.

```rust
fn verify_deferred_state(
    deferred_state_wire: DeferredStateWire,
) -> Result<DeferredCommitment, VerificationError>
```

This validates/evaluates raw deferred state and returns the deferred commitment. This is not a full deferred proof, but it is a valid verification path for proof formats that carry raw deferred state.

The high-level `verify()` function should internally dispatch to the appropriate deferred verifier, then call `verify_vm_proof(...)`.

### Future Batch/Aggregation Scenario

The core API should leave room for batching without defining the batching strategy in this issue.

For example, future APIs may look conceptually like:

```rust
fn verify_batch(
    proofs: impl IntoIterator<Item = (VmStarkProof, VmInputs)>,
    batch_deferred_proof: BatchPrecompileStarkProof,
) -> Result<(), VerificationError>
```

where:

```rust
fn verify_batch_precompile_proof(
    batch_precompile_stark_proof: BatchPrecompileStarkProof,
) -> Result<DeferredCommitmentSet, VerificationError>
```

A batch deferred proof could verify deferred work for multiple VM proofs. Each VM proof would then be verified by calling `verify_vm_proof(...)` with the matching deferred commitment.

This issue should only ensure that the core verifier primitives are compatible with this future direction.

### MASM Recursive Verifier API

The MASM recursive verifier should mirror the native API at the conceptual level.

The low-level MASM primitive should correspond to:

```rust
verify_vm_proof(vm_stark_proof, inputs, deferred_commitment) -> ()
```

Conceptually, a MASM entrypoint such as:

```text
exec.vm::verify_vm_proof
```

should verify the VM proof against:

- program digest,
- kernel commitment,
- stack inputs,
- stack outputs,
- deferred commitment,
- proof/advice data.

The exact MASM layout is still open. Possible shape:

- program digest and kernel commitment on the operand stack,
- pointers to stack inputs and stack outputs in memory,
- deferred commitment on stack or in memory,
- proof transcript data in advice, as today.

The new API should wrap or adapt the current `exec.vm::verify_proof` machinery. Internally, it may still use the flat public input layout and stream individual kernel procedure digests, but caller-facing verification should be expressed in terms of structured VM inputs.

### Deferred Commitment

The deferred commitment commits to all deferred computation requested by the VM execution.

The VM proof is verified against this commitment. The deferred work itself is verified separately, either by a precompile VM proof or by validating raw deferred state.

The high-level `verify()` API should hide this detail from ordinary callers. The mid-level `verify_vm_proof(...)` primitive should accept the deferred commitment explicitly because it verifies only the VM STARK portion.

### Open Questions

- What exact `MidenProof` / `DeferredProof` shape should we expose?
- Should `verify_vm_proof(...)` be public, crate-internal, or available only for advanced recursive/batching use?
- What exact MASM operand stack and memory layout should `exec.vm::verify_vm_proof` use?
- Should the MASM verifier assert `VmInputs` internally, or expose verified public values for a wrapper to assert?
- What exact kernel commitment length-binding rule should be used?
- Should raw deferred state verification remain a supported path, or only a transitional one?
- How should future batch deferred proofs expose the set of verified deferred commitments?

Al-Kindi-0:
Reworks recursive verification around a canonical **execution claim**. Resolves #3250 (and #2045).

The statement a proof attests is now an `ExecutionClaim` — `(program_root, kernel_commitment, stack_inputs, stack_outputs)` — with a domain-tagged commitment bound into the Fiat-Shamir transcript, so a proof is bound to exactly its claim. The deferred obligation `D` is returned by verification, never accepted from callers.

The API changes are for:

**Native** — `Verifier::verify(proof, claim)` and `verify_partial(proof, claim)`; `verify_partial` returns the deferred obligation as a `#[must_use]` `Unsettled` token that the caller settles or re-exposes.

**MASM** — `exec.vm::verify_vm_proof [claim_ptr] -> [D, num_queries, query_pow_bits, deep_pow_bits, folding_pow_bits]`. The caller stages the 40-felt claim in memory; the kernel-digest witness is fetched from the advice map under the claim's kernel commitment and bound-checked before use; the proof is consumed from the advice stack (or fetched by `request_key`). The verifier returns the deferred root together with the proof's **transcript-bound security parameters** and applies **no acceptance policy of its own** — callers grade the parameters (e.g. `sys::vm::conjectured_security_level` + a `>= 96` threshold; shown in the e2e tests). `miden_verifier::recursive` builds the advice.


crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -45,22 +62,25 @@ proc load_security_params
...
-    push.ACCEPTABLE_QUERY_POW_BITS assert_eq.err="query_pow_bits does not match acceptable security policy"
-
-    exec.constants::get_deep_pow_bits
-    push.ACCEPTABLE_DEEP_POW_BITS assert_eq.err="deep_pow_bits does not match acceptable security policy"
-
-    exec.constants::get_folding_pow_bits
-    push.ACCEPTABLE_FOLDING_POW_BITS assert_eq.err="folding_pow_bits does not match acceptable security policy"
+    push.CONJECTURED_BITS_PER_QUERY_FP mul
+    u32shr.16
+    exec.constants::get_query_pow_bits add
```
huitseeker: The old wrapper required `deep_pow_bits = 12` and `folding_pow_bits = 4`, but this grade only reads the query count and query PoW.  The generic verifier only checks that the other two values are below 32, so a proof built with both set to zero still gets 96 for `(27, 17)`. 

That removes both intermediate grinding steps while reporting the same security level which sounds suboptimal. 

Should we keep minimum checks for these fields?

verifier/src/lib.rs
```diff
@@ -224,21 +202,26 @@ fn hydrate_deferred_state(
 }
 
 fn stark_security_level(_proof: &StarkProof) -> u32 {
-    // Mirrors `ExecutionProof::security_level` until the STARK security estimator is available.
+    // TODO: placeholder for the precompile-VM proof's security level. Blocked on the
+    // precompile-VM security estimator (does not exist yet); wire together with the VM-side
+    // native level via `miden_air::config`. `verify` returns `min(vm_level, this)`, so this must
```
huitseeker, line 223: `precompile_pcs_params()` still uses `(num_queries, query_pow_bits) = (27, 16)`, which the estimator added here rates at 95 bits.

core/src/program/domain.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Selectors share one namespace with the `merge_in_domain` values used for MAST control-block
+//! hashing. Those are opcode-sized (`< 256`) while every registered selector is `>= 257`
+//! (`domain_id >= 1`), so the ranges cannot collide. The sequential and two-to-one
+//! constructions share the capacity layout deliberately: a registered id must not be reused as
+//! a merge domain.
+
+use crate::Felt;
+
+/// Registered domain id for the kernel commitment.
+pub const KERNEL_COMMITMENT_DOMAIN_ID: u32 = 32;
```
huitseeker: Nit: the (draft!) registry in 0xMiden/crypto#1026 reserved `0x010000..0x01ffff` for miden-vm; IDs below `0x010000` belong to miden-crypto. 

Using 32 here puts these consensus tags in crypto's range and would allows a future crypto entry to reuse them. _If_ we keep the allocated ranges, wouldn't we need to allocate these three IDs from the VM range?

crates/lib/core/docs/sys/vm.md
```diff
@@ -3,4 +3,5 @@
...
...d from the advice map by<br />content.<br /><br />The caller populates the claim region with its own trusted P, I, O (K's slot may be left<br />unset) and supplies the kernel digest witness. This procedure materializes K from the<br />witness, derives the claim commitment from the completed region, fetches the registered<br />proof from the advice map under `request_key(verifier_root, claim_commitment)` (the verifier<br />root via `procref` of `verify_vm_proof` itself), and verifies it. The derived commitment is<br />both the fetch address and what verification binds, so consistency needs no check: a wrong<br />or substituted package fails verification against the caller's claim.<br /><br />Inputs:  [claim_ptr, kernel_ptr, num_kernel_digests, ...]<br />Outputs: [D, ...]<br /><br />Where:<br />- claim_ptr is the word-aligned address of the caller's claim region (40 felts), with P at +0,<br />I at +8, and O at +24 already populated from the caller's own inputs; K (+4) is written<br />here (and rederived by the verifier from the same witness).<br />- kernel_ptr / num_kernel_digests locate the kernel digest witness.<br />- D is the deferred root bound by the verified statement.<br /> |
```
huitseeker: These generated rows describe the entrypoints backwards. `verify_vm_proof` is now the commitment-first advice-map wrapper, while `verify_vm_proof_from_claim` takes only `claim_ptr`, consumes the proof stream, and returns `[security_level, D]` ..  and neither accepts the kernel pointer and count shown here.

Does this need re-generation? It should also fix the stale `stage_boundary_inputs` row.

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Note that we have removed the wrapper and the core MASM verifier now returns the proof parameters, including the pow_bits.
It is correct that for conjectured security only the query phase pow matters. So the current conjectured security estimator only takes into account the query phase pow.
An upcoming proven security estimator will take all of the pow into account.

A question then is what security notion should we go with for mainnet. If we go with conjectured, then we can set `deep_pow_bits` and `folding_pow_bits` to zero and gain a few ms in proving times.

verifier/src/lib.rs
Al-Kindi-0: Fixed

core/src/program/domain.rs
Al-Kindi-0: Good catch, and fixed it now though these are likely to change when we land the domain separation work

crates/lib/core/docs/sys/vm.md
Al-Kindi-0: Regenerated and should be good now

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+    let heights = miden_trace_heights(&stark)?;
+
+    build_advice(&config, params, &stark, heights, pub_inputs, claim_commitment)
+}
+
+/// Deserializes a wincode-encoded Poseidon2 STARK proof, bounding preallocation by
+/// [`MAX_STARK_PROOF_BYTES`].
+fn deserialize_proof(proof_bytes: &[u8]) -> Result<P2ProofData, RecursiveAdviceError> {
+    let encoding_config = wincode::config::Configuration::default()
+        .with_preallocation_size_limit::<MAX_STARK_PROOF_BYTES>();
```
huitseeker, line 222: `ExecutionProof::from_parts` can hold more than 64 MiB. This setting caps each sequence allocation, not the total input, and ordinary wincode decoding accepts trailing bytes. 

A valid proof prefix with a large suffix can reach `advice_inputs`, while `verify_stark_proof` rejects it at `verifier/src/lib.rs:302`.

It would be good to match the native proof-size check, and to use exact decoding here.

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -494,25 +420,122 @@ mod fast_parallel {
...
+    .expect("Proving failed");
+    let claim = ExecutionClaim::from_program_info(program.into(), stack_inputs, stack_outputs);
+
+    // swap in empty deferred material: it resolves to TRUE, which the statement did not bind
+    let stark = proof.miden_proof();
+    let tampered = ExecutionProof::new(
+        StarkProof::new(stark.bytes().to_vec(), stark.hash_fn()),
+        DeferredProof::empty(),
+    );
+    assert!(verify(tampered, claim).is_err());
```
huitseeker, line 551: Hmm, could we also verify the untampered `proof` before replacing its deferred material? 

This removed the only positive top-level test for a valid `DeferredProof::Stark`; the current assertion still passes if final composed verification is broken, because it checks only the tampered package. 

The inner deferred-proof tests do not exercise the outer VM proof against the returned root.

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,507 @@
...
+    let lmcs = config.lmcs();
+
+    let mut partial_trees = Vec::new();
+    let mut advice_map = Vec::new();
+
+    // DEEP openings (one BatchProof per commitment: main, aux, quotient), then FRI openings
+    // (one per FRI round).
+    for batch_proof in pcs.deep_witnesses.iter().chain(pcs.fri_witnesses.iter()) {
+        let (tree, entries) = batch_proof_to_merkle(lmcs, batch_proof)?;
+        partial_trees.push(tree);
```
huitseeker: Could we create `store` before this loop and extend it from `tree.inner_nodes()` here? `partial_trees` is only retained for the loop below, so inserting each tree immediately removes the intermediate `Vec` and keeps the ownership flow local.

huitseeker (commented):
This overall looks good, though there are some check and test gaps in the PR, see inline.

verifier/src/recursive/mod.rs
Al-Kindi-0: fixed

miden-vm/tests/integration/prove_verify.rs
Al-Kindi-0: fixed

verifier/src/recursive/mod.rs
Al-Kindi-0: done

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,562 @@
...
+    if proof_bytes.len() > MAX_STARK_PROOF_BYTES {
+        return Err(RecursiveAdviceError::ProofTooLarge {
+            size: proof_bytes.len(),
+            max: MAX_STARK_PROOF_BYTES,
+        });
+    }
+
+    let encoding_config = wincode::config::Configuration::default()
+        .with_preallocation_size_limit::<MAX_STARK_PROOF_BYTES>();
+    deserialize_serde_exact::<P2ProofData, _>(proof_bytes, encoding_config)
```
huitseeker, line 223: So, recursive advice rejects proof bytes that native `verify` still accepts. 

Appending trailing bytes to a valid inner STARK proof returns `Ok(96)` natively but reaches `TrailingBytes` here; the precompile verifier uses the same permissive decode as the native path. 

Could we use exact decoding in all three places (ideally a strict one) so they agree on which encodings are valid?

miden-vm/src/lib.rs
```diff
@@ -21,26 +24,18 @@ pub use miden_processor::{
...
-pub fn verify(
-    program_info: ProgramInfo,
-    stack_inputs: StackInputs,
-    stack_outputs: StackOutputs,
-    proof: ExecutionProof,
-) -> Result<u32, VerificationError> {
-    Verifier::new().verify(program_info, stack_inputs, stack_outputs, proof)
+/// Wire-backed deferred proofs are partial/delegable proof material and are rejected here; use
+/// [`Verifier::verify_partial`] to verify and hydrate wire-backed partial proofs.
+pub fn verify(proof: ExecutionProof, claim: ExecutionClaim) -> Result<u32, VerificationError> {
```
huitseeker, line 39: Nit: this changes the public call to `(proof, ExecutionClaim)`, but `miden-vm/README.md:143-175` still lists four arguments and calls `Verifier::new().verify(ProgramInfo::from(program), StackInputs::default(), expected_outputs, proof)`.

huitseeker (approved):
Some nits inline, but overall LGTM

verifier/src/recursive/mod.rs
Al-Kindi-0: fixed

miden-vm/src/lib.rs
Al-Kindi-0: fixed

verifier/src/lib.rs
```diff
@@ -18,16 +18,32 @@ use miden_crypto::stark::{
...
+    C: wincode::config::Config,
+    SerdeCompat<T>: wincode::SchemaRead<'de, C, Dst = T>,
+{
+    let value = <SerdeCompat<T> as wincode::SchemaRead<'de, C>>::get(bytes.by_ref())?;
+    if bytes.is_empty() {
+        Ok(value)
+    } else {
+        Err(wincode::error::trailing_bytes())
+    }
+}
```
bobbinth, line 38: This function feels out of place here. It also looks like a duplicate of a function with the similar name in the `preocmpiles-prover` crate. Should we move it to some "utils" module and de-duplicate?

verifier/src/lib.rs
```diff
@@ -70,49 +88,30 @@ impl Verifier {
...
     /// - The proof's STARK-backed precompile VM proof, if present, does not verify against its
     ///   public root.
     pub fn verify(
         &self,
-        program_info: ProgramInfo,
-        stack_inputs: StackInputs,
-        stack_outputs: StackOutputs,
         proof: ExecutionProof,
+        claim: ExecutionClaim,
     ) -> Result<u32, VerificationError> {
```
bobbinth, line 109: Not related to this PR, but I think we should change the API in a couple of ways to align it more with the MASM API:

1. We shouldn't return the security level as it requires making some opinionated assumptions. Instead we could return the proof parameters (number of queries, grinding factors etc.).
2. `verify()` should return something like `Option<DeferredState>` in case the proof contains unverified deferred state.

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    pub advice_stack: Vec<Felt>,
+    /// Merkle store backing the query openings (`mtree_get` authentication paths).
+    pub store: MerkleStore,
+    /// Content-addressed advice-map entries: query rows (`leaf_hash -> leaf_data`), the ACE
+    /// circuit, and the kernel digest witness under K.
+    pub advice_map: Vec<(Word, Vec<Felt>)>,
+    /// Commitment to the execution claim: the content address (paired with a verifier root) the
+    /// proof stream is registered under.
+    pub claim_commitment: Word,
+}
```
bobbinth, line 81: Do we really need this type? AFAICT, this is basically `AdviceInputs` and `claim_commitment`. If we do want to keep it, it should probably look more like:

```rust
pub struct RecursiveVerifierInputs {
    advice: AdviceInputs,
    claim_commitment: Word,
}
```

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+// ================================================================================================
+
+/// Builds the advice a MASM recursive verifier consumes to verify a Miden VM proof against
+/// its claim.
+///
+/// The proof must be a Poseidon2 proof — the recursive verifier verifies only Poseidon2 STARKs.
+pub fn advice_inputs(
+    proof: &ExecutionProof,
+    claim: &ExecutionClaim,
+) -> Result<RecursiveVerifierInputs, RecursiveAdviceError> {
```
bobbinth, line 139: If we do keep the `RecursiveVerifierInputs` struct, I'd convert this function into the constructor of this struct - something like:

```rust
impl RecursiveVerifierInputs {

    pub fn new(
        proof: &ExecutionProof,
        claim: &ExecutionClaim,
    ) -> Result<Self, RecursiveVerifierInputsError> {
        ...
    }
}
```

verifier/src/recursive/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// for any number of proofs merge into one advice provider in any order. A consumer holding the
+    /// claim commitment fetches the package under this key and verifies it with
+    /// `exec.vm::verify_vm_proof`; the key is addressing, not trust — a package that does not match
+    /// the consumer's claim fails verification.
+    pub fn into_request_package(mut self, verifier_root: Word) -> Self {
+        let key = request_key(verifier_root, self.claim_commitment);
+        let proof_stream = core::mem::take(&mut self.advice_stack);
+        self.advice_map.push((key, proof_stream));
+        self
+    }
```
bobbinth, line 97: Question: why do we want to provide the verifier root separately from struct construction? Building on the previous comment, why not make the constructor look like this:

```rust
impl RecursiveVerifierInputs {

    pub fn new(
        verifier_root: Word,
        proof: &ExecutionProof,
        claim: &ExecutionClaim,
    ) -> Result<Self, RecursiveVerifierInputsError> {
        ...
    }

}
```

crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -449,6 +449,62 @@ pub proc hash_elements
     # => [HASH]
 end
 
+#! Computes the domain-tagged hash of Felt values starting at the specified memory address.
+#!
+#! Mirrors `Poseidon2::hash_elements_in_domain` on the Rust side: the first capacity element
+#! carries the padding rule (`num_elements % 8`), the second carries the domain tag, and the
+#! input is zero-padded to a rate multiple.
```
bobbinth, line 456: I would probably state explicitly that this returns a non-empty word on empty input.

Separately, looking at the Rust's `hash_elements_internal()` implementation now, I am not sure I fully understand the logic. Specifically, my understanding is that on empty input, we'd apply the permutation to the following state:

```
[[1, 0, 0, 0], [0, 0, 0, 0], [0, domain, 0, 0]]
```

If this is correct, wouldn't we get a collision when trying to do `hash_elements_in_domain([1, 0, 0, 0, 0, 0, 0, 0], domain)`?

A potential fix here could be to inject `1` into a "free" capacity register (e.g. `state[CAPACITY_RANGE.start + 2]`) - but this probably requires more investigation.

crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -449,6 +449,62 @@ pub proc hash_elements
...
+#! Outputs: [HASH]
+#!
+#! Where:
+#! - ptr is the memory address of the first element to be hashed; must be word-aligned.
+#! - num_elements is the number of elements to be hashed.
+#! - domain is the domain-separation tag placed in the second capacity element.
+pub proc hash_elements_in_domain
+    # => [ptr, num_elements, domain]
+    swap u32divmod.8 swap
+    # => [num_elements/8, num_elements%8, ptr, domain]
```
bobbinth, line 468: We never check if `num_elements` is a `u32` - isn't this a potential issue?

I also noticed that `prepare_hasher_state` has the same problem.


Also, why not use `prepare_hasher_state` here and the just swap the domain element? It is a bit more cycles, but should result in less code.

crates/lib/core/asm/sys/vm/claim.masm
```diff
@@ -0,0 +1,105 @@
...
+    # => [CLAIM_HASH, ...]
+end
+
+#! Computes the canonical kernel commitment over a raw kernel-procedure digest list.
+#!
+#! Mirrors miden_core::program::KernelDescriptor::commitment: the domain-tagged sequential hash
+#! of the flattened digests, length-bound by the sponge's padding rule, in the descriptor's
+#! canonical order.
+#!
+#! Inputs:  [kernel_ptr, num_kernel_digests, ...]
```
bobbinth, line 61: nit: I'd probably rename `num_kernel_digests` into `num_kernel_procedures` as it makes the variable content a bit more self-evident.

crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -45,22 +48,38 @@ proc load_security_params
...
+#! Must match miden_air::config::conjectured_security_level bit-for-bit (enforced by
+#! cross-tests). Holds no policy: the caller applies its own acceptance threshold to the result.
+#!
+#! PRECONDITION: `num_queries <= 150`, so `num_queries * C` fits in a u32 (the `u32shr` below
+#! requires that). Parameters returned by `verify_vm_proof` satisfy this bound: the generic
+#! verifier enforces it.
+#!
+#! Inputs:  [num_queries, query_pow_bits, ...]
+#! Outputs: [level, ...]
+pub proc conjectured_security_level
```
bobbinth, line 77: nit: I'd rename this into something like `compute_conjectured_security_level` to have the procedure name start with an action verb.

crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -172,34 +191,88 @@ proc store_ace_registry_root
...
-#! block (kernel_H | program_digest | deferred_root | pad) before the transcript starts.
+#! The returned deferred root is bound by the verified statement. The caller must settle it or
+#! include it in its own statement.
 #!
-#! Inputs:  [kernel_ptr, num_kernel_digests, stack_io_ptr, PROG0, PROG1, PROG2, PROG3]
-#! Outputs: []
-pub proc verify_proof
-    # --- Load and validate security parameters from advice stack ---
+#! Inputs:  [claim_ptr, ...]
+#! Outputs: [D, num_queries, query_pow_bits, deep_pow_bits, folding_pow_bits, ...]
```
bobbinth, line 223: Question: why not pass `CLAIM_COMMITMENT` instead of `claim_ptr`? We should be able to write the pre-image of the claim commitment into the local procedure memory pretty easily.

crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -172,34 +191,88 @@ proc store_ace_registry_root
...
+#!
+#! Where:
+#! - claim_ptr is word-aligned; its 40-felt region ends at or before
+#!   `constants::verifier_memory_start`.
+#! - D is the deferred root bound by the verified statement.
```
bobbinth, line 228: I would mention here explicitly that if `D == EMPTY_WORD` then the were not deferred computations.

crates/lib/core/asm/sys/vm/mod.masm
```diff
@@ -172,34 +191,88 @@ proc store_ace_registry_root
...
+#! - num_queries, query_pow_bits, deep_pow_bits, folding_pow_bits are the proof's
+#!   transcript-bound security parameters.
+pub proc verify_vm_proof
```
bobbinth, line 231: It would be good to start adding signatures to procedures.

bobbinth (commented):
Looks good! Thank you! Not a very deep review, but I left some comments inline. Some bigger API refactoring should probably be addressed as a part of https://github.com/0xMiden/miden-vm/pull/3437, but others should be addressed in a follow-up PR.

verifier/src/lib.rs
Al-Kindi-0: Agreed, moved this to `miden-serde-utils`

verifier/src/recursive/mod.rs
Al-Kindi-0: Done

verifier/src/recursive/mod.rs
Al-Kindi-0: Done

verifier/src/recursive/mod.rs
Al-Kindi-0: Added `for_request(verifier_root, proof, claim)` for request packages. I kept `new(proof, claim)` for direct verifier invocation and made the packaging step private.

crates/lib/core/asm/crypto/hashes/poseidon2.masm
Al-Kindi-0: Good point and the `10*` trick wouldn't work with our padding rule. Adopted your suggestion which matches the spirit of the upcoming domain separation RFC

crates/lib/core/asm/crypto/hashes/poseidon2.masm
Al-Kindi-0: Done

crates/lib/core/asm/sys/vm/claim.masm
Al-Kindi-0: Done

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Done

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Done

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Done

Al-Kindi-0:
@bobbinth , will open a small followup PR to address your feedback

crates/lib/core/asm/sys/vm/mod.masm
Al-Kindi-0: Changed

verifier/src/lib.rs
Al-Kindi-0: Agreed, we can do it in the other PR #3437 as it makes more sense there

PR #3423, opened 2026-07-24
Make advice stack order explicit

### LINKED ISSUE #2557: Introduce `AdviceStack` type
_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2547#discussion_r2672991371_

>IMO the need for an advice stack *builder* goes away when we get rid of `adv_push.n`. Building the advice stack should be trivial: you either queue in felts, words or double words. And similarly when consuming data from the advice stack.
>
>Hence, the main problem I see with our advice provider representation is that we don't have an `AdviceStack` type to encode this (I would personally get rid of `AdviceInputs` in favor of a `AdviceProvider` constructor, but that's a separate issue). It would represent the stack with a `VecDeque<Felt>` as you do here, and have 6 methods: `append_{felt,word,dword}()`, and `consume_{felt,word,dword}` (naming subject to change).
>
>This `AdviceStack` API would force the caller to think about how the data will be consumed in-VM: either with `adv_push`, `adv_loadw` or `adv_pipe`. For example, in `...::falcon512::rpo::encode_signature()` we currently construct the stack as a `Vec<Felt>`, and then dump that on the advice stack in `push_falcon_signature()` using `AdviceMutation::extend_stack()`. Doing so makes it hard to be convinced that the data is pushed on the advice stack in the correct order. It would be a lot more clear if instead `AdviceMutation::extend_stack()` took an `AdviceStack` as argument, and `encode_signature()` called a sequence of `append_{felt,word,dword}()` methods. Not only would it make more clear that data consumed by `adv_pipe` in a loop is properly double-word aligned, we would be easily able to compare this method's implementation with the corresponding MASM code to make sure that whatever gets pushed a certain way gets consumed analogously in MASM.
>
>Given the size of this PR though, we should probably do this in a follow-up PR. Though ideally we get rid of the `adv_push.4`-related methods in this PR.

Blocked on #2547

huitseeker:
This PR adds a public `AdviceStack` type. It stores advice values in the same top first order that the VM consumes.

`AdviceInputs` can now accept and return an `AdviceStack`.

`AdviceProvider` now stores its live stack as an `AdviceStack`. `AdviceMutation` can also carry an `AdviceStack`

Core handlers and key tests now build advice with typed calls such as push_word, push_dword, and push_for_adv_pipe. This removes local reversal rules from places where order must be clear.

Closes #2557 

The old `stack`field and the old `extend_stack` helper are removed only in the last commit. Question for @Al-Kindi-0 : should we make `push_for_adv_push` deprecated?

### DIFF
--- a/core/src/advice/stack.rs
+++ b/core/src/advice/stack.rs
@@ -3,6 +3,155 @@ use alloc::{collections::VecDeque, vec::Vec};
 use super::{AdviceInputs, AdviceMap};
 use crate::{Felt, Word, crypto::merkle::MerkleStore};
 
+// ADVICE STACK
+// ================================================================================================
+
+/// Advice stack values ordered from top to bottom.
+///
+/// The front of the stack is the next element consumed by `adv_push`.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct AdviceStack {
+    stack: VecDeque<Felt>,
+}
+
+impl AdviceStack {
+    /// Creates a new empty advice stack.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Returns the number of elements on this advice stack.
+    pub fn len(&self) -> usize {
+        self.stack.len()
+    }
+
+    /// Returns true if this advice stack has no elements.
+    pub fn is_empty(&self) -> bool {
+        self.stack.is_empty()
+    }
+
+    /// Returns an iterator over elements from top to bottom.
+    pub fn iter(&self) -> impl Iterator<Item = &Felt> {
+        self.stack.iter()
+    }
+
+    /// Pushes a single element onto the advice stack.
+    ///
+    /// Elements are consumed in FIFO order: first pushed = first consumed by advice operations.
+    pub fn push_element(&mut self, value: Felt) -> &mut Self {
+        self.stack.push_back(value);
+        self
+    }
+
+    /// Extends the advice stack with raw elements ordered from top to bottom.
+    ///
+    /// Elements are consumed in FIFO order: first element in iter = first consumed.
+    pub fn push_elements<I>(&mut self, values: I) -> &mut Self
+    where
+        I: IntoIterator<Item = Felt>,
+    {
+        self.stack.extend(values);
+        self
+    }
+
+    /// Adds elements for consumption by multiple sequential `adv_push` instructions.
+    ///
+    /// After `repeat.n adv_push end`, the operand stack will have `slice[0]` on top.
+    pub fn push_for_adv_push(&mut self, slice: &[Felt]) -> &mut Self {
+        for elem in slice.iter().rev() {
+            self.stack.push_back(*elem);
+        }
+        self
+    }
+
+    /// Adds a word for consumption by `adv_loadw` or `adv_pushw`.
+    pub fn push_word(&mut self, word: Word) -> &mut Self {
+        self.stack.extend(word.iter().copied());
+        self
+    }
+
+    /// Adds two words for consumption by `adv_pipe`.
+    pub fn push_dword(&mut self, words: [Word; 2]) -> &mut Self {
+        for word in words {
+            self.push_word(word);
+        }
+        self
+    }
+
+    /// Adds elements for sequential consumption by `adv_pipe` operations.
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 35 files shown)


Al-Kindi-0 (approved):
LGTM!

On the question, I would keep `push_for_adv_push`. The type's invariant is not "no transformations" but "every order rule lives in one named, tested place": one helper per consuming instruction (`push_word` for `adv_loadw`/`adv_pushw`, `push_for_adv_pipe` for `adv_pipe`, `push_for_adv_push` for repeated `adv_push`). Deleting it would push `.iter().rev()` back to the call sites, which is exactly the ad-hoc reversal #2557 set out to remove - and it has real callers (e.g., the Falcon tau pair).

huitseeker:
@Al-Kindi-0 Sounds good! It's kept in this PR.

core/src/advice/stack.rs
```diff
@@ -59,129 +70,130 @@ impl AdviceStackBuilder {
...
+    {
+        let values: Vec<Felt> = values.into_iter().collect();
+        for value in values.into_iter().rev() {
+            self.stack.push_front(value);
+        }
+        self
+    }
+
+    /// Prepends a single element to the top of the advice stack.
+    pub fn prepend_element(&mut self, value: Felt) -> &mut Self {
```
bobbinth: nit (feel free to ignore): this is basically a "push" - right?

core/src/advice/stack.rs
```diff
@@ -59,129 +70,130 @@ impl AdviceStackBuilder {
...
-    /// builder.push_word([w0, w1, w2, w3].into());
-    /// // MASM: adv_loadw (or adv_pushw)
-    /// // Result: operand stack = [w0, w1, w2, w3, ...] with w0 on top
-    /// ```
     pub fn push_word(&mut self, word: Word) -> &mut Self {
-        for elem in word.iter() {
-            self.stack.push_back(*elem);
+        self.stack.extend(word.iter().copied());
+        self
+    }
```
bobbinth: Is "push" the right terminology here? My understanding is that this appends element at the bottom of the stack while "push" usually puts the elements at the top of the stack. Maybe a better name would be `append_word` (and some goes for `push_dword`, and also `push_fore_adv_pipe` becomes `append_for_adv_pipe`).

core/src/advice/stack.rs
```diff
@@ -48,7 +59,7 @@ impl AdviceStackBuilder {
         self
     }
 
-    /// Extends the advice stack with raw elements (already ordered top-to-bottom).
+    /// Extends the advice stack with raw elements ordered from top to bottom.
     ///
     /// Elements are consumed in FIFO order: first element in iter = first consumed.
     pub fn push_elements<I>(&mut self, values: I) -> &mut Self
```
bobbinth: Similar to the other comment, I find the "push" terminology a bit confusing here. Overall, my thinking is:

- "push" should put elements on the top of the stack. I think the only method that I'd probably use this for is `prepend_element` which I'd rename into either `push_element` or maybe even just `push`.
- We have methods that put elements on top of the stack "in order" - for these I'd use the "prepend" word (e.g., `prepend_elements`, `prepend_word` etc.).
- We have methods that add elements at the bottom of the stack (also "in order") - for these I'd use "append" word - e.g., `append_elements` (this method), `append_word`, `append_dword` etc.

bobbinth (approved):
Looks good! Thank you! I left a couple of comments inline - mostly about the confusion I had with the push/prepend/append terminology.

core/src/advice/stack.rs
bitwalker: Agreed - I'd for sure be confused that `push` places items on the bottom of the stack

core/src/advice/stack.rs
huitseeker: I renamed `prepend_element` to `push_element`, since this is the single helper that puts one value on top of the advice stack.

core/src/advice/stack.rs
huitseeker: I renamed the bottom-adding helpers to `append_word`, `append_dword`, and `append_for_adv_pipe`.

core/src/advice/stack.rs
huitseeker: `push_element` now adds one value to the top, `prepend_*` keeps ordered top insertion, and `append_*` covers bottom insertion.

PR #3426, opened 2026-07-25
perf(precompiles): add support for GLV strategy 

Nashtare:
## Describe your changes

Add support for GLV strategy when computing MSMs.

Move the secp256k1 GLV endomorphism split from the ECDSA verifier's MASM/advice ABI into the precompiles prover's addition-chain strategy. `ecdsa_k256_keccak::verify` now always logs a plain `u1*G + u2*Q` claim
  (Straus-shaped, no algorithm selector, no GLV advice), and the deferred prover satisfies it internally with a GLV-decomposed wNAF chain -- `phi(P)` is certified in-circuit via its value relation (`x_phi = beta*x_P`, `y_phi
  = y_P`), never trusted host advice.

Performance comparison between the two algorithms: https://claude.ai/code/artifact/b2e7a755-9313-4397-9810-eba22ab4a39e?via=auto_preview

MASM overhead is negligible (12 cycles for `verify`).
 
- Updated `CHANGELOG.md`


crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -100,6 +100,381 @@ pub proc verify
...
+    # => [HAT_2B, HAT_2A, U2, ...]
+    exec.push_lambda_digest
+    # => [LAMBDA, HAT_2B, HAT_2A, U2, ...]
+    exec.assert_recompose
+    # => [...]
+
+    # -- Resolve the 4-term MSM over the pair buffer at loc[24..56]. --
+    push.4 locaddr.24
+    # => [ptr, n, ...]
+    exec.secp256k1::msm_mem
```
huitseeker: I think we want to normalize these four pairs before registering the MSM.

A valid split can contain a zero half (`glv_decompose(1)` returns magnitudes `1` and `0`), and `Q == G` can repeat a signed `G` base. 

`CurvePrecompile::evaluate_msm` rejects zero scalars and repeated points, so both valid cases end in `curve: InvalidPayload`. 

We could omit zero terms and merge equal signed bases, or fall back to Straus.

crates/lib/core/src/dsa.rs
```diff
@@ -37,37 +54,99 @@ pub mod ecdsa_k256_keccak {
...
+        out
+    }
+
+    /// Computes the verifier's `u1`/`u2` ECDSA scalars the same way the MASM verifier does, splits
+    /// each via the secp256k1 GLV endomorphism, and encodes the four signed halves as advice.
+    fn glv_advice_felts(sig: &Signature, msg: Word) -> [Felt; GLV_HALVES_FELTS] {
+        let z = z_from_message(msg);
+        let r = be_bytes_to_le_limbs(sig.r());
+        let s = be_bytes_to_le_limbs(sig.s());
+        let s_inv = scalar_inv_mod_n(s);
```
huitseeker: Should this validate `s` before trying to invert it and return a `Result` `Signature::from_sec1_bytes_and_recovery_id` accepts untrusted bytes without checking `s`, so `s = 0` now panics here instead of letting the verifier reject it. 

`s = n` is also inconsistent: `scalar_inv_mod_n(n)` returns zero even though its docs say unreduced inputs panic.

huitseeker (commented):
Thanks! overall this looks super promising!!

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+#! secp256k1 GLV endomorphism image. On-curve membership (and hence `phi(P) = lambda*P`) is
+#! certified when this digest is later evaluated, because `P` is on-curve and `beta^3 = 1 mod p`
+#! keeps the curve equation satisfied.
+#!
+#! Input:  [bx, by, ...]
+#! Output: [PHI_DIGEST, ...]
+proc phi_point
+    exec.push_beta_digest
+    # => [BETA, bx, by, ...]
+    swapw
```
adr1anh: Is this `swapw` necessary? Base-field multiplication is commutative, so `k1_base::mul` should be able to consume `[BETA, BX]` directly.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+    # => [MAG_LO_U32[4], ...]
+    adv_pushw
+    # => [MAG_HI_U32[4], MAG_LO_U32[4], ...]
+    swapw
+    # => [MAG_LO_U32[4], MAG_HI_U32[4], ...]
+    exec.k1_scalar::load
+    # => [MAG_DIGEST, ...]
+    adv_push
+    # => [sign, MAG_DIGEST, ...]
+    dup dup dup mul assert_eq
```
adr1anh: Could this assertion carry a specific error message so the booleanity requirement is visible at the failure site?

```masm
dup dup dup mul assert_eq.err="GLV sign must be boolean"
```

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+#!
+#! Each inlined half below follows the same shape: pull one signed magnitude from advice, load its
+#! base digest from a fixed local slot, write the MSM pair `(signed_base, mag)` to the pair buffer,
+#! and reduce to `hat = sign ? (0 - mag) : mag` for the term's recompose check.
+#!
+#! Advice: four [`pull_signed_scalar`] draws, in order: u1's `a` half, u1's `b` half, u2's `a`
+#! half, u2's `b` half.
+#! Input:  [BX, BY, U1, U2, ...]
+#! Output: [VERIFY_POINT, ...]
+@locals(60)
```
adr1anh: Could we document the complete local-memory layout here? In particular, the slots for `G`, `φ(G)`, `φ(Q)`, `Q`, `U1`, `U2`, the four MSM pairs, and cached `λ`. With 60 locals and a mix of canonical VALUE and expression digests, an explicit map would make the procedure much easier to audit.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+proc glv_4base_msm
+    # => [BX, BY, U1, U2, ...]
```
adr1anh: Could the stack and local annotations distinguish canonical VALUE digests from expression digests, for example with `_VALUE` and `_EXPR` suffixes?

That provenance matters here because raw digest comparison is only valid for known canonical VALUE digests, while equality involving expression digests must go through the semantic precompile operations. In particular, `QX/QY` and the advised magnitudes are canonical values, while `U1/U2`, signed bases, and `φ(Q)` may be expressions.

crates/lib/core/src/dsa.rs
```diff
@@ -40,34 +44,85 @@ pub mod ecdsa_k256_keccak {
...
+    /// the MASM verifier re-derives and checks this relation, so an incorrect witness fails
+    /// verification rather than forging anything. `PAD` is trailing zero felts rounding the total
+    /// up to a multiple of 8 (`push_for_adv_pipe`'s requirement); the verifier drains them from
+    /// advice unread. The result is advice witness data, not a commitment to the supplied
+    /// signature encoding.
     ///
     /// The public-key elements come from [`SequentialCommit::to_elements()`], matching the
     /// commitment returned by [`public_key_commitment()`].
-    pub fn encode_signature(pk: &PublicKey, sig: &Signature) -> Vec<Felt> {
+    pub fn encode_signature(pk: &PublicKey, sig: &Signature, msg: Word) -> Vec<Felt> {
```
adr1anh: I don’t think the GLV decomposition belongs in the signature encoding. This changes `encode_signature(pk, sig)` to require the message, expands its output with verifier-specific hints, and duplicates the verifier’s computation of `z`, `s⁻¹`, `u1`, and `u2`.

We already have the generic `k1_scalar::inv` procedure backed by the `field_inv` event handler, and the host implementation is available through `UintDomain::K1Scalar`.

Could we instead emit a GLV-decomposition event after MASM has derived `U1` and `U2`? The handler can use `ProcessorState::require_canonical_deferred_node` to obtain their canonical scalar values, run `glv_decompose`, and push the four signed halves to advice. MASM should retain the current boolean and recomposition checks because handler output remains untrusted.

This would preserve the existing signature advice/API boundary, avoid duplicating the ECDSA scalar calculation, and allow malformed `s` values to be rejected by the verifier rather than causing `encode_signature` to panic.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+#! Computes `R = u1*G + u2*Q` via the secp256k1 GLV endomorphism (see the module note above), given
+#! `Q`'s raw base-field coordinate digests.
+#!
+#! Each inlined half below follows the same shape: pull one signed magnitude from advice, load its
+#! base digest from a fixed local slot, write the MSM pair `(signed_base, mag)` to the pair buffer,
+#! and reduce to `hat = sign ? (0 - mag) : mag` for the term's recompose check.
+#!
+#! Advice: four [`pull_signed_scalar`] draws, in order: u1's `a` half, u1's `b` half, u2's `a`
+#! half, u2's `b` half.
+#! Input:  [BX, BY, U1, U2, ...]
```
adr1anh: Since these are specifically `Q`’s coordinates, should this use `QX, QY`? `BX/BY` reads like an arbitrary base-field pair and makes the provenance less obvious.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+#! Output: [VERIFY_POINT, ...]
+@locals(60)
+proc glv_4base_msm
+    # => [BX, BY, U1, U2, ...]
+
+    # -- Gather phase: Q, phi(Q), phi(G) (precomputed), G -- stash G/phi(G)/phi(Q)/Q/U1/U2 to locals
```
adr1anh: Could we reserve locals for the required MSM pair buffer and keep more of `G`, `φ(G)`, `Q`, `φ(Q)`, `U1/U2`, and `λ` stack-threaded?

The current gather/process path introduces six scratch stores, six corresponding loads, and a separate `λ` store/load. I’m not sure all of those round trips are necessary.

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak_glv.masm
```diff
@@ -0,0 +1,409 @@
...
+
+    padw locaddr.56 mem_loadw_le
+    # => [LAMBDA, HAT_2B, HAT_2A, U2, ...]  (reload of term 1's cached LAMBDA, avoids recomputing it)
+    exec.assert_recompose
+    # => [...]
+
+    # -- Resolve the 4-term MSM over the pair buffer at loc[24..56]. --
+    push.4 locaddr.24
+    # => [ptr, n, ...]
+    exec.secp256k1::msm_mem
```
adr1anh: A valid GLV decomposition can contain a zero half. In particular, `glv_decompose(1)` produces `(1, 0)`.

This procedure always registers all four pairs, but `CurvePrecompile::evaluate_msm` rejects every zero-scalar term. Therefore an otherwise valid signature can fail—for example, `s = z` gives `u1 = 1`, and this does not require an exceptional public key, so the repeated-base fallback does not cover it.

I think we need to either omit zero terms, fall back to the 2-base path when a half is zero, choose an equivalent nonzero decomposition, or normalize zero terms in the MSM implementation.

precompiles/src/math/curve/glv.rs
```diff
@@ -0,0 +1,361 @@
...
+/// Computes `a * b mod n`, the secp256k1 scalar-field order.
+pub fn scalar_mul_mod_n(a: Limbs, b: Limbs) -> Limbs {
+    let n = limbs_to_wide(K1Scalar::MODULUS);
+    wide_to_limbs((limbs_to_wide(a) * limbs_to_wide(b)) % n)
+}
+
+/// Computes the multiplicative inverse of `a` modulo `n` (the secp256k1 scalar-field order) via
+/// Fermat's little theorem (`n` is prime): `a^(n-2) mod n`. Panics if `a` is not reduced, or is
+/// zero (which has no inverse).
+pub fn scalar_inv_mod_n(a: Limbs) -> Limbs {
```
adr1anh: Do we need these scalar arithmetic helpers? The uint implementation already exposes multiplication and prime-field inversion through `UintDomain::K1Scalar.mul/inv`, and MASM uses the corresponding generic `k1_scalar::inv` procedure.

This inverse also does not match its documentation: it says unreduced inputs panic, but `a = n` is reduced to zero inside `mod_pow` and returns zero.

The GLV event-handler approach would remove the need for these helpers. If they remain, I think they should reuse the uint-domain implementation, validate canonical input, and return `Option`/`Result` for zero.

adr1anh (changes requested):
The signed GLV algebra looks correct to me:

```text
u ≡ sa·a + λ·sb·b mod n
```

and the implementation consistently applies each sign both to the MSM base and to the scalar-field recomposition. Here `λ` is the scalar-field eigenvalue; `β` is only used to construct `φ(P) = (βx, y)`.

I have two main blocking concerns:

1. Valid zero GLV halves are always included in the 4-term MSM, but the curve precompile rejects zero-scalar terms. This can reject valid signatures.
2. I don’t think GLV hints should become part of `encode_signature`. Generating them through an event from the already-derived `U1/U2` digests would preserve the signature API and avoid duplicating hashing, inversion, and scalar arithmetic.

There is also a mismatch between the PR description and the current branch: the description says selection remains manual and the benchmark uses Straus, while the implementation is now GLV-only and Straus is archived.

The remaining comments are documentation and MASM cleanup suggestions. There are also unrelated duplicate changelog entries, and the branch should be rebased rather than carrying the merge from `next`.

This was a static review; I did not compile or run tests.

Nashtare:
Updated comparison against baseline. We now have GLV outperforming Straus in almost all cases and the delta is even larger for heavy batches (up to 36%)

Screenshot 2026-08-03 at 17 57 38

Nashtare:
@adr1anh I've revamped the approach to account for our latest discussion. The initial overhead of GLV is now basically removed (and the gains are even bigger). This should also make it much easier / cleaner to integrate non-GLV curves like secp256r1 or Curve25519 in the future.

crates/precompiles-prover/src/deferred/session.rs
```diff
@@ -291,7 +312,46 @@ impl<'a> DeferredSessionBuilder<'a> {
...
+        // unaffected either way. Both tables are cached per `(point,
+        // window)` the same way the plain path's are (a recurring base —
+        // the ECDSA generator across a batch of signatures — lays each
+        // table once); sign rides the digit selection inside
+        // `glv_joint_wnaf_with_tables`, not the table's seed, so a shared
+        // base's tables serve every claim's GLV split regardless of sign.
+        let expr = if !curve.endomorphisms().is_empty() {
+            for (base, _) in &expr_terms {
+                self.ensure_wnaf_table(base, MSM_WNAF_WINDOW);
+                self.ensure_wnaf_table_endo(base, MSM_WNAF_WINDOW);
```
huitseeker: `translate_ec_msm` now builds a GLV endomorphism table for every secp256k1 base, but `intro_endo` panics when the base is infinity. Identity is a valid MSM point, so could we retain the plain path for identity?

huitseeker (commented):
I think we just need to touch up the infinity edge case, otherwise lgtm!

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -98,40 +98,67 @@ pub proc verify
...
-
 #! Computes ECDSA scalars U1 = z/s and U2 = r/s, keeping SIG_R for the final check.
 #!
 #! Input:  [Z_U32[8], pubkey_ptr, ...]
 #! Advice: [SIG_R[8] | SIG_S[8] | ...]
-#! Output: [U1_DIGEST, Q_DIGEST, U2_DIGEST, SIG_R_DIGEST, ...]
+#! Output: [U1_DIGEST, U2_DIGEST, SIG_R_DIGEST, pubkey_ptr, ...]
+#!
+#! `pubkey_ptr` rides through unread: the public key is loaded once, by the verification-point
+#! procedure that also needs `Q`'s raw coordinate digests.
```
Al-Kindi-0: Is this still applicable or should we just remove ?

crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm
```diff
@@ -98,40 +98,67 @@ pub proc verify
...
+    # => [BY, pubkey_ptr+16, pubkey_ptr, U1, U2, SIG_R, ...]
+    movup.5
+    # => [pubkey_ptr, BY, pubkey_ptr+16, U1, U2, SIG_R, ...]
+    exec.k1_base::load_mem_stream
+    # => [BX, pubkey_ptr+8, BY, pubkey_ptr+16, U1, U2, SIG_R, ...]
+    movup.4 drop
+    # => [BX, BY, pubkey_ptr+16, U1, U2, SIG_R, ...]
+    movup.8 drop
+    # => [BX, BY, U1, U2, SIG_R, ...]
+    exec.secp256k1::load_digest_pair
```
Al-Kindi-0: Isn't this block just `secp256k1::load_mem` ?

crates/precompiles/src/math/curve/mod.rs
```diff
@@ -327,6 +353,52 @@ impl CurveId {
...
+            }],
+        }
+    }
+
+    /// Returns named fixed base-field constants whose canonical VALUE digests the generated MASM
+    /// needs as comparison literals -- empty for curves with no such constants.
+    ///
+    /// Unlike [`Self::extra_points`], these are never precompile operands: MASM only compares a
+    /// runtime coordinate digest against them, so they are deliberately *not* seeded into
+    /// [`CurvePrecompile::init`] and cost nothing outside the generated constant table.
```
Al-Kindi-0: Are we sure of this ?

crates/precompiles/src/math/curve/glv.rs
```diff
@@ -0,0 +1,309 @@
...
+    let k2 = c1.negate().mul(b1).sub(c2.mul(b2));
+
+    [(k1.neg, wide_to_limbs(k1.mag)), (k2.neg, wide_to_limbs(k2.mag))]
+}
+
+/// Computes `a * b mod n`, the secp256k1 scalar-field order.
+pub fn scalar_mul_mod_n(a: Limbs, b: Limbs) -> Limbs {
+    let n = limbs_to_wide(K1Scalar::MODULUS);
+    wide_to_limbs((limbs_to_wide(a) * limbs_to_wide(b)) % n)
+}
```
Al-Kindi-0, line 172: Is this still needed ?

crates/lib/core/tests/crypto/dsa.rs
```diff
@@ -36,12 +37,61 @@ fn core_ecdsa_k256_keccak_verify_accepts_valid_signature() {
...
+#[test]
+fn core_ecdsa_k256_keccak_verify_accepts_glv_base_repeating_public_keys() {
+    let one = core::array::from_fn(|i| u32::from(i == 0));
+    let lambda_squared = scalar_mul_mod_n(SECP256K1_LAMBDA, SECP256K1_LAMBDA);
+
+    for (name, secret_scalar) in [
+        ("Q == G", one),
+        ("Q == -G", negate_scalar_mod_n(one)),
+        ("Q == phi(G)", SECP256K1_LAMBDA),
+        ("Q == phi^2(G)", lambda_squared),
```
Al-Kindi-0, line 111: Should we also cover `Q == -phi(G)` and `Q == -phi^2(G)`?

crates/precompiles-prover/src/tests/deferred_session.rs
```diff
@@ -125,6 +125,25 @@ fn deferred_session_inputs_reject_duplicate_base_msm() {
     assert!(state.register(msm).is_err(), "duplicate-base MSM must be rejected");
 }
 
+#[test]
+fn deferred_session_lowers_msm_with_identity_base_without_panicking() {
+    let mut state = state();
+    let curve = CurveId::Secp256k1;
+    let identity = CurvePrecompile::identity_node(curve);
+    let generator = CurvePrecompile::generator_node(curve);
+    let one = UintPrecompile::value_node(curve.scalar_domain(), limbs(1));
```
Al-Kindi-0: Does this test still pass with `limbs(17)` in place of `limbs(1)`?

crates/precompiles-prover/src/session/strategies.rs
```diff
@@ -290,6 +345,94 @@ pub fn joint_wnaf(session: &mut Session, terms: &[(EcNode, U256)], w: usize) ->
...
+/// Panics if `terms` is empty or any scalar is zero.
+pub fn glv_joint_wnaf_with_tables(
+    session: &mut Session,
+    terms: &[(&WnafTable, Option<&WnafTable>, U256)],
+) -> EcExprPtr {
+    assert!(!terms.is_empty(), "an MSM needs at least one base");
+
+    let mut signed_terms: Vec<(&WnafTable, U256, bool)> = Vec::with_capacity(terms.len() * 2);
+    for &(plain, endo, k) in terms {
+        assert_ne!(k, U256::ZERO, "glv_joint_wnaf_with_tables terms must have a nonzero scalar");
```
Al-Kindi-0, line 386: Should zero scalars be treated like identity bases, basically evaluate to the identity rather than reject?

CHANGELOG.md
```diff
@@ -80,6 +80,8 @@
 - Bound deferred precompile STARK proofs to the generated precompile ACE relation digest ([#3344](https://github.com/0xMiden/miden-vm/pull/3344)).
 - [BREAKING] Split Poseidon2 permutation rows out of `ChipletsAir` into `Poseidon2PermutationAir`, and updated the recursive verifier ACE registry for three AIRs ([#3345](https://github.com/0xMiden/miden-vm/pull/3345)).
 - [BREAKING] Optimize periodic columns evaluation for fewer ACE gates ([#3347](https://github.com/0xMiden/miden-vm/pull/3347)).
+- [BREAKING] Optimize periodic columns evaluation for fewer ACE gates ([#3347](https://github.com/0xMiden/miden-vm/pull/3347)).
```
Al-Kindi-0: Should be removed

Al-Kindi-0 (approved):
LGTM

crates/precompiles-prover/src/deferred/session.rs
```diff
@@ -277,6 +298,12 @@ impl<'a> DeferredSessionBuilder<'a> {
                     reason: "zero-scalar terms are not supported",
                 });
             }
+            if self.session.is_pai(&point.node) {
+                return Err(DeferredSessionError::UnsupportedMsm {
```
huitseeker, line 302: nit: `CurvePrecompile::evaluate_msm` accepts identity bases, but this rejects the resulting deferred state during proof generation even though the plain-table fallback below supports it. 

Any way of having those paths agree on how they behave on identity is ok by me (as long as they do).

crates/precompiles/src/math/curve/mod.rs
```diff
@@ -327,6 +353,33 @@ impl CurveId {
...
     }
 
+    /// Returns named fixed point constants for this curve beyond identity/generator, e.g. a GLV
+    /// endomorphism image -- empty for curves with no such fixed derived points.
+    ///
+    /// Every entry here needs both (a) seeding into [`CurvePrecompile::init`]'s initial
+    /// deferred-DAG node set, so the point's digest is resolvable with no live registration, and
+    /// (b) a baked-in MASM constant/wrapper proc, emitted by `crates/lib/core/codegen`. Both
+    /// consumers read this single list so they cannot drift apart.
+    pub fn extra_points(self) -> Vec<(&'static str, CurvePoint)> {
```
huitseeker: nit: `extra_points()` seeds and generates `push_phi_generator`, but no production code uses it now that GLV runs inside the prover. 

Is it worth removng this public/init surface and keeping `φ(G)` as a test vector?

crates/precompiles-prover/src/session/strategies.rs
```diff
@@ -290,6 +345,94 @@ pub fn joint_wnaf(session: &mut Session, terms: &[(EcNode, U256)], w: usize) ->
...
+/// this in-circuit certification closes, not a prover-strategy one).
+///
+/// Builds each base's tables fresh. When a base recurs across many calls —
+/// the generator across a batch of signatures — build its tables once with
+/// [`wnaf_table`] / [`wnaf_table_endo`] and drive the same ladder with
+/// [`glv_joint_wnaf_with_tables`] instead.
+///
+/// Returns the combined MSM expression. Panics if `terms` is empty, any
+/// scalar is zero, or `curve` has no GLV endomorphism.
+pub fn glv_joint_wnaf(
```
huitseeker: nit: `glv_joint_wnaf()` is used only by one test, while production calls `glv_joint_wnaf_with_tables()`. Could the test use the production path so this uncached wrapper and its documentation can be removed?

crates/precompiles/src/math/curve/mod.rs
```diff
@@ -327,6 +353,33 @@ impl CurveId {
...
+        match self {
+            Self::Secp256k1 => vec![("PHI_GENERATOR", phi_generator())],
+        }
+    }
+
+    /// Returns this curve's fixed GLV endomorphisms -- empty for curves with none. A slice, not
+    /// an `Option`: a higher-dimensional GLV curve would list several, each merging into the same
+    /// base's term the same way one does.
+    pub fn endomorphisms(self) -> &'static [Endomorphism] {
```
huitseeker: nit: `endomorphisms()` returns a slice, but every caller uses at most one entry and the AIR stores one `beta` and `lambda`. Could this return `Option<Endomorphism>`?

huitseeker (approved):
LGTM

Nashtare:
@adr1anh Merging is blocked on your review request 🙂

adr1anh (approved):

PR #3427, opened 2026-07-25
feat(debug): add initial support for debugging inline calls

djolertrk:
Addresses https://github.com/0xMiden/compiler/issues/1284.

### DIFF
--- a/core/src/operations/debug_metadata/inline_call.rs
+++ b/core/src/operations/debug_metadata/inline_call.rs
@@ -1,0 +1,59 @@
+use alloc::sync::Arc;
+
+use miden_debug_types::FileLineCol;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// Debug information describing one source-level function in an active inline call chain.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DebugInlineCallInfo {
+    name: Arc<str>,
+    linkage_name: Option<Arc<str>>,
+    declaration: FileLineCol,
+    call_site: FileLineCol,
+}
+
+impl DebugInlineCallInfo {
+    pub fn new(
+        name: impl Into<Arc<str>>,
+        declaration: FileLineCol,
+        call_site: FileLineCol,
+    ) -> Self {
+        Self {
+            name: name.into(),
+            linkage_name: None,
+            declaration,
+            call_site,
+        }
+    }
+
+    pub fn with_linkage_name(mut self, linkage_name: impl Into<Arc<str>>) -> Self {
+        self.linkage_name = Some(linkage_name.into());
+        self
+    }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn linkage_name(&self) -> Option<&str> {
+        self.linkage_name.as_deref()
+    }
+
+    pub fn declaration(&self) -> &FileLineCol {
+        &self.declaration
+    }
+
+    pub fn call_site(&self) -> &FileLineCol {
+        &self.call_site
+    }
+
+    pub fn set_declaration(&mut self, declaration: FileLineCol) {
+        self.declaration = declaration;
+    }
+
+    pub fn set_call_site(&mut self, call_site: FileLineCol) {
+        self.call_site = call_site;
+    }
+}
--- a/crates/assembly-syntax/src/ast/instruction/inline_call.rs
+++ b/crates/assembly-syntax/src/ast/instruction/inline_call.rs
@@ -48,12 +48,4 @@ impl DebugInlineCallInfo {
     pub fn call_site(&self) -> &FileLineCol {
         &self.call_site
     }
-
-    pub fn set_declaration(&mut self, declaration: FileLineCol) {
-        self.declaration = declaration;
-    }
-
-    pub fn set_call_site(&mut self, call_site: FileLineCol) {
-        self.call_site = call_site;
-    }
 }
--- a/crates/assembly-syntax/src/ast/instruction/mod.rs
+++ b/crates/assembly-syntax/src/ast/instruction/mod.rs
@@ -283,6 +283,8 @@ pub enum Instruction {
 
     // ----- debug decorators --------------------------------------------------------------------
     DebugVar(DebugVarInfo),
+    DebugInlineCall(miden_core::operations::DebugInlineCallInfo),
+    DebugInlineCallClear,
 
     // ----- event decorators --------------------------------------------------------------------
     Emit,
@@ -299,7 +301,7 @@ impl Instruction {
     /// Some instructions (like [`DebugVar`](Self::DebugVar)) are compiler-internal and have
     /// no surface syntax. They should be skipped during pretty-printing.
     pub const fn has_textual_representation(&self) -> bool {
-        !matches!(self, Self::DebugVar(_))
+        !matches!(self, Self::DebugVar(_) | Self::DebugInlineCall(_) | Self::DebugInlineCallClear)
     }
 }
 
--- a/crates/assembly-syntax/src/ast/instruction/mod.rs
+++ b/crates/assembly-syntax/src/ast/instruction/mod.rs
@@ -1,5 +1,6 @@
 pub mod advice;
 mod debug_var;
+mod inline_call;
 mod print;
 
 use alloc::vec::Vec;
@@ -8,6 +9,7 @@ use core::ops::Range;
 pub use self::{
     advice::SystemEventNode,
     debug_var::{DebugVarInfo, DebugVarLocation},
+    inline_call::DebugInlineCallInfo,
 };
 use crate::{
     Felt,
@@ -282,8 +284,18 @@ pub enum Instruction {
     ProcRef(InvocationTarget),
 
     // ----- debug decorators --------------------------------------------------------------------
+    /// Records a source variable location at this point in the generated operation stream.
     DebugVar(DebugVarInfo),
-    DebugInlineCall(miden_core::operations::DebugInlineCallInfo),
+    /// Appends one source function to the inline call chain active for subsequent operations.
+    ///
+    /// Frames are emitted from innermost to outermost. This compiler-internal marker has no VM
+    /// execution semantics; the assembler stores the active chain on each generated source-node
+    /// occurrence until [`DebugInlineCallClear`](Self::DebugInlineCallClear) is encountered.
+    DebugInlineCall(DebugInlineCallInfo),
+    /// Clears the inline call chain for subsequent generated operations and control nodes.
+    ///
+    /// This marks the point at which source attribution returns to the containing non-inlined
+    /// function. It is compiler-internal and does not emit a VM operation.
     DebugInlineCallClear,
 
     // ----- event decorators --------------------------------------------------------------------
--- a/crates/assembly/src/basic_block_builder.rs
+++ b/crates/assembly/src/basic_block_builder.rs
@@ -23,7 +23,7 @@ use miden_mast_package::debug_info::{
 use crate::{
     ProcedureContext,
     assembler::BodyWrapper,
-    mast_forest_builder::{MastForestBuilder, MastNodeRef},
+    mast_forest_builder::{MastForestBuilder, MastNodeUse},
 };
 
 // PENDING ASM OP
@@ -338,14 +338,14 @@ impl BasicBlockBuilder<'_> {
     ///
     /// This consumes all operations in the builder, but does not touch the operations in the
     /// epilogue of the builder.
-    pub(crate) fn make_basic_block(&mut self) -> Result<Option<MastNodeRef>, Report> {
+    pub(crate) fn make_basic_block(&mut self) -> Result<Option<MastNodeUse>, Report> {
         if !self.ops.is_empty() {
             let ops = core::mem::take(&mut self.ops);
             let asm_ops = core::mem::take(&mut self.asm_ops);
             let debug_vars = core::mem::take(&mut self.debug_vars);
             let inline_calls = core::mem::take(&mut self.inline_calls);
 
-            let basic_block_node_ref = self.mast_forest_builder.ensure_block_ref(
+            let basic_block_node_use = self.mast_forest_builder.ensure_block_use(
                 ops,
                 asm_ops,
                 debug_vars,
@@ -353,7 +353,7 @@ impl BasicBlockBuilder<'_> {
                 vec![],
             )?;
 
-            Ok(Some(basic_block_node_ref))
+            Ok(Some(basic_block_node_use))
         } else {
             Ok(None)
         }
@@ -366,7 +366,7 @@ impl BasicBlockBuilder<'_> {
     /// - Operations contained in the epilogue of the builder are appended to the list of ops which
     ///   go into the new BASIC BLOCK node.
     /// - The builder is consumed in the process.
-    pub fn try_into_basic_block(mut self) -> Result<Option<MastNodeRef>, Report> {
+    pub fn try_into_basic_block(mut self) -> Result<Option<MastNodeUse>, Report> {
         self.ops.append(&mut self.epilogue);
         self.make_basic_block()
     }
--- a/crates/assembly/src/basic_block_builder.rs
+++ b/crates/assembly/src/basic_block_builder.rs
@@ -11,11 +11,14 @@ use miden_assembly_syntax::{
     diagnostics::Report,
 };
 use miden_core::{
-    Felt,
+    Felt, Word,
     events::SystemEvent,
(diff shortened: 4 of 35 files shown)


djolertrk:
Rebased on top of `origin/bitwalker/package-debug-info-consolidation` branch

djolertrk:
cc @bitwalker

crates/assembly/src/basic_block_builder.rs
```diff
@@ -116,6 +133,7 @@ impl BasicBlockBuilder<'_> {
 impl BasicBlockBuilder<'_> {
     /// Adds the specified operation to the list of basic block operations.
     pub fn push_op(&mut self, op: Operation) {
+        self.record_active_inline_calls(self.ops.len() as u32);
```
huitseeker: `record_active_inline_calls` only covers basic-block ops. 

`Call`/`Exec` and structured control flow create MAST nodes directly, so their inline chain is dropped. A marked `add` had one row while `call` and `if` paths had none. 

Could we store the chain on control-node occurrences and test both paths?

core/src/operations/debug_metadata/inline_call.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    pub fn declaration(&self) -> &FileLineCol {
+        &self.declaration
+    }
+
+    pub fn call_site(&self) -> &FileLineCol {
+        &self.call_site
+    }
+
+    pub fn set_declaration(&mut self, declaration: FileLineCol) {
```
huitseeker: nit: `set_declaration` and `set_call_site` add public mutation APIs that neither this repository nor the companion compiler producer uses. Since callers already provide both values to `new`, could we remove these setters until a consumer needs them?

crates/assembly-syntax/src/ast/instruction/inline_call.rs
bitwalker, line 1: I don't believe we need to define this type here as it isn't inherently part of core - it should be able to go directly in `miden-assembly-syntax` instead, since it is directly tied to assembly and the package format.

crates/assembly-syntax/src/ast/instruction/mod.rs
```diff
@@ -283,6 +283,8 @@ pub enum Instruction {
 
     // ----- debug decorators --------------------------------------------------------------------
     DebugVar(DebugVarInfo),
+    DebugInlineCall(miden_core::operations::DebugInlineCallInfo),
+    DebugInlineCallClear,
```
bitwalker, line 299: Could we document the semantics of this instruction (both really, but this is the one that is harder to reason about)?

bitwalker (changes requested):

crates/assembly/src/basic_block_builder.rs
djolertrk: Yes, thanks!

core/src/operations/debug_metadata/inline_call.rs
djolertrk: yes, thanks!

crates/assembly-syntax/src/ast/instruction/inline_call.rs
djolertrk: it does make sense. addressed

crates/assembly-syntax/src/ast/instruction/mod.rs
djolertrk: yes, thanks!

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +494,92 @@ impl MastForestBuilder {
...
+    ///
+    /// `exec` reuses the callee's execution node instead of creating a control node. Clone its
+    /// source-occurrence tree so the call-site chain is visible for every executed operation while
+    /// leaving the shared MAST nodes and other source occurrences unchanged.
+    pub(crate) fn record_exec_inline_calls(
+        &mut self,
+        node_ref: MastNodeRef,
+        inline_calls: &[DebugSourceInlineCall],
+    ) -> Result<(), Report> {
+        if inline_calls.is_empty() {
```
huitseeker, line 587: `record_exec_inline_calls` skips empty chains, so I'm afraid that a plain `exec` after a decorated `exec` of the same target could reuse the decorated source occurrence and shows stale inline frames.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +494,92 @@ impl MastForestBuilder {
...
+        for op_idx in source_node.op_start..source_node.op_end {
+            inline_calls.extend(active_inline_calls.iter().map(|inline_call| {
+                DebugSourceInlineCall {
+                    op_idx,
+                    callee_idx: inline_call.callee_idx,
+                    loc_idx: inline_call.loc_idx,
+                }
+            }));
+        }
+        inline_calls.extend(source_node.inline_calls);
```
huitseeker: `clone_source_occurrence_with_inline_calls` puts the caller chain before the callee's existing chain. The debugger reads rows in innermost-to-outermost order, so nested `exec` frames are reversed. 

Is that what we want, and could we clarify with an add a nested-chain test?

crates/assembly/src/instruction/procedures.rs
```diff
@@ -44,8 +53,9 @@ impl Assembler {
         &self,
         mast_forest_builder: &mut MastForestBuilder,
         asm_op: AssemblyOp,
+        inline_calls: Vec<DebugSourceInlineCall>,
     ) -> Result<Option<MastNodeRef>, Report> {
-        let dyn_node_ref = mast_forest_builder.ensure_dyn_node_ref(false, asm_op)?;
+        let dyn_node_ref = mast_forest_builder.ensure_dyn_node_ref(false, asm_op, inline_calls)?;
```
huitseeker: `dynexec` stores these rows only on the `DYN` node. The processor then selects the target's independent source root, so an inlined `hir.exec_indirect` loses its caller chain throughout the target body. 

Shouldn't this context follow the selected target?

crates/assembly/src/mast_forest_builder.rs
bitwalker, line 587: A decorated `exec` becomes this execution ref's global latest/history occurrence, while the empty-chain return records nothing. Later positional source selection therefore moves the decoration to a different `exec` or even a later `call` of the same callee. Could we track the selected `SourceNodeRef` per use instead of deriving it from global latest history?

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +494,92 @@ impl MastForestBuilder {
...
+                false,
+            )?);
+        }
+
+        let mut inline_calls = Vec::with_capacity(
+            source_node.inline_calls.len()
+                + active_inline_calls.len()
+                    * source_node.op_end.saturating_sub(source_node.op_start) as usize,
+        );
+        for op_idx in source_node.op_start..source_node.op_end {
```
bitwalker, line 736: For a dynamically linked target, the source occurrence is an `External` node with range `0..0`, so this loop emits no inline rows. Execution then switches to the target package's source graph, which has no knowledge of the caller chain. Could we add an explicit representation or runtime propagation mechanism for the external boundary?

crates/assembly/src/mast_forest_builder.rs
bitwalker: The target occurrence's existing rows describe frames inside the chain active at the `exec` site. Appending them after the active caller rows produces outer-to-inner ordering, while the public contract and debugger expect innermost-to-outermost. Could we emit the target rows before the caller rows for each operation?

bitwalker (changes requested):
Looks good overall, but automated review found some issues I didn't notice during manual review, and I think we should address them before merging

crates/assembly/src/mast_forest_builder.rs
djolertrk: hmm well that is true. every `exec` now records its exact source occurrence, including empty chains, so a plain `exec` cannot inherit stale inline frames. there is a test for it

crates/assembly/src/mast_forest_builder.rs
djolertrk: this is fixed now. added a nested `exec` regression test as well.

crates/assembly/src/instruction/procedures.rs
djolertrk: yeah, this is fixed now. added test for this as well. thank you!

crates/assembly/src/mast_forest_builder.rs
djolertrk: yeah, fixed

crates/assembly/src/mast_forest_builder.rs
djolertrk: true, fixed

crates/assembly/src/mast_forest_builder.rs
djolertrk: fixed

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +526,100 @@ impl MastForestBuilder {
...
+    /// `exec` reuses the callee's execution node instead of creating a control node. Clone its
+    /// source-occurrence tree so the call-site chain is visible for every executed operation while
+    /// leaving the shared MAST nodes and other source occurrences unchanged.
+    pub(crate) fn record_exec_inline_calls(
+        &mut self,
+        target: MastNodeUse,
+        inline_calls: &[DebugSourceInlineCall],
+    ) -> Result<MastNodeUse, Report> {
+        let mut cloned = BTreeMap::new();
+        let decorated_source_ref = self.clone_source_occurrence_with_inline_calls(
```
huitseeker, line 601: This clones the full callee source-occurrence tree for every `exec`, even when `inline_calls` is empty. A linear chain of procedures where each body executes the previous procedure twice makes the debug graph grow as `S(n) = 2S(n - 1) + 1`, so a small ordinary MASM program can exhaust memory during assembly despite having no inline markers.

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +526,100 @@ impl MastForestBuilder {
...
+                        u32::from(self.nodes[source_node.exec_node].kind.is_external()),
+                    ) as usize,
+        );
+        if source_node.op_start == source_node.op_end
+            && self.nodes[source_node.exec_node].kind.is_external()
+        {
+            inline_calls.extend(source_node.inline_calls.iter().copied());
+            inline_calls.extend(active_inline_calls.iter().map(|inline_call| {
+                DebugSourceInlineCall {
+                    op_idx: 0,
```
huitseeker: The package validator permits a zero-width external occurrence to store its boundary rows at `source_node.op_start`, but this rewrites them to index `0`. `SourceInlineCallContext::for_operation_zero` also queries only index `0`, so a valid external occurrence with a nonzero start either becomes invalid when cloned or loses its inherited context at runtime.

huitseeker (changes requested):
Plain exec may grow the debug graph exponentially.

crates/assembly/src/mast_forest_builder.rs
djolertrk: hmm, that is true! plain `exec` now reuses its exact `MastNodeUse` when the inline chain is empty, so it performs no source-tree cloning, which is I think sufficient for solving this

crates/assembly/src/mast_forest_builder.rs
djolertrk: Fixed

djolertrk:
> This pull request was closed without review because none of the linked open issues is assigned to @djolertrk.
> 
> Our [contribution guidelines](https://github.com/0xMiden/miden-vm/blob/main/CONTRIBUTING.md#contribution-quality) require contributions to link to an issue assigned to the author. Please ask to be assigned before opening a new pull request.

Hmm, I think this is an issue :) The https://github.com/0xMiden/compiler/issues/1284 is assigned to myself, and in future, we will have issues in repos other than VM, which will require changes in VM...

bitwalker:
@huitseeker This was improperly closed by the quality checks automation - it completely disregarded the fact that a PR is being actively reviewed, that @djolertrk is a regular contributor, and that there _is_ an issue motivating this PR, just not in this repo. I'd suggest we relax some of the criteria if those conditions apply (especially if a maintainer has performed a review already, at that point there shouldn't be any automatic actions taken on the PR IMO).

crates/assembly/src/mast_forest_builder.rs
```diff
@@ -492,6 +526,106 @@ impl MastForestBuilder {
...
+
+        let mut inline_calls = Vec::with_capacity(
+            source_node.inline_calls.len()
+                + active_inline_calls.len()
+                    * core::cmp::max(
+                        source_node.op_end.saturating_sub(source_node.op_start),
+                        u32::from(self.nodes[source_node.exec_node].kind.is_external()),
+                    ) as usize,
+        );
+        if source_node.op_start == source_node.op_end
```
bitwalker: **[P1] Preserve external-boundary identity across node deduplication**

This classifies a zero-width source occurrence using the execution node's current kind, but an `External` draft may already be deduplicated onto a concrete same-digest node or may be replaced by one later. In the first ordering, the empty `0..0` loop drops the active inline chain; in the reverse ordering, the boundary row remains after replacement and package validation rejects it because the finalized node is no longer external. Please preserve the external-boundary property on the source occurrence itself, or remap it to the concrete source occurrence, rather than consulting the mutable execution-node kind.

processor/src/execution/external.rs
```diff
@@ -88,10 +90,15 @@ where
     tracer.record_mast_forest_resolution(resolved_node_id_new_forest, &new_mast_forest);
 
     let old_package_debug_info = current_package_debug_info.clone();
+    let inline_context_depth = inline_call_contexts.len();
+    inline_call_contexts.push(inline_call_context);
```
bitwalker: **[P1] Avoid retaining context entries during source-unaware execution**

For source-unaware fast execution and replay, this pushes `None` into the caller-owned vector. `execute_impl_pure` then uses a separate local vector, and its `EnterForest` arm restores only the forest, so the caller-owned entry is never truncated. Every externally linked `exec` therefore grows this vector until the execution or fragment ends, which is unbounded in the number of calls and can exhaust memory in a loop. Please either avoid mutating this state in the pure path or consistently thread and restore the same vector.

processor/src/fast/step.rs
```diff
@@ -40,6 +41,19 @@ impl ResumeContext {
...
+    pub fn next_source_node_id(&self) -> Option<DebugSourceNodeId> {
+        self.continuation_stack
+            .peek_continuation_with_source_node_id()
+            .and_then(|(_, source_node_id)| source_node_id)
+    }
+
+    /// Returns dynamic-boundary inline contexts active for the next operation, ordered from the
+    /// innermost boundary to the outermost.
+    pub fn inherited_inline_call_contexts(&self) -> impl Iterator<Item = &SourceInlineCallContext> {
+        self.inline_call_contexts.iter().rev().filter_map(Option::as_ref)
```
bitwalker: **[P2] Account for a pending `EnterForest` before exposing contexts**

After an external target's final `END` clock, `StepStopper` returns a `ResumeContext` while `EnterForest` is still pending. The next step consumes that non-clock continuation and truncates the context stack before executing the next operation, so this accessor currently reports a stale external frame despite its next-operation contract. Please derive the effective depth after leading non-clock `EnterForest` continuations, or normalize the context stack before constructing the resume context.

bitwalker (changes requested):
Still a few latent issues unfortunately, but we're getting close!

djolertrk:
> Still a few latent issues unfortunately, but we're getting close!

Thank you for the review!

crates/assembly/src/tests.rs
```diff
@@ -2940,6 +2940,425 @@ fn invalid_debug_variable_type_returns_error_instead_of_panicking() -> TestResul
     Ok(())
 }
 
+fn replace_nops_with_inline_call_markers(
```
huitseeker: nit: `replace_nops_with_inline_call_markers` repeats the marker rewrite in `replace_nops_with_named_inline_call_markers` and has only two callers. Could those callers pass alternating markers to the named helper so this copy can be removed?

huitseeker (approved):
LGTM!

crates/assembly/src/tests.rs
djolertrk: Hmm, yes! :)

bitwalker (approved):

huitseeker:
@djolertrk Sadly we can't edit this branch, so it's key to maintain it up to date, could I ask you for yet another rebase?

djolertrk:
> @djolertrk Sadly we can't edit this branch, so it's key to maintain it up to date, could I ask you for yet another rebase?

ofc :) @huitseeker

djolertrk:
@huitseeker rebased again, thanks!

PR #3424, opened 2026-07-24
Cleanup unused public surface

huitseeker:
## Summary

- Removes unused convenience methods, constants, test helpers, and error factories.
- Removes the unused `ColMatrix` type and its stale range-check source file.
- Narrows test-only helpers to the smallest visibility needed by their callers.

## Downstream use

The removed names were checked against the current `next` branches of:

- `0xMiden/compiler`;
- `0xMiden/protocol`;
- `0xMiden/rust-sdk`;
- `0xMiden/node`.

No use was found. This includes the direct and transitive `miden-crypto` paths used by protocol,
client, and node. The scan also covered merged precompile work from #3293 and the current #3398
branch.

PR #3428, opened 2026-07-25
Restore StorageUpdates API compatibility and reduce benchmark noise

huitseeker:
PR #3424 removed (unused!) public `StorageUpdates` methods _usable_ by existing users. This restores them so miden-crypto 0.28.1 remains compatible with 0.28.0 (that way the next release of `miden-crypto` doesn't have to be 0.29)

The Blake3 benchmark now reports possible regressions without failing CI or posting commit comments (it's too noisy).

The synthetic transaction benchmark repeats a possible regression with the run order reversed. It fails only when the same metric crosses the limit twice. The error includes both results.


bobbinth (approved):
Not a deep review, but looks good! Thank you!

PR #3430, opened 2026-07-26
Limit unsafe code in debug info serialization

huitseeker:
PR #3398 casts fixed debug info tables between bytes and Rust rows. The casts accept any `Copy` type. `Copy` does not ensure that every byte pattern is valid or that every byte may be read, so each row needs a manual safety check.

The shared string table and fast decoding from #3398 are unchanged. Simple rows now derive `zerocopy::FromBytes`, `IntoBytes`, `Immutable`, and `KnownLayout`. `DebugFunctionInfo` uses a separate wire row because its field elements need validation. `OptionC` stores two initialized `u32` values instead of a `MaybeUninit<T>` payload. The four remaining unsafe operations only manage the aligned byte allocation.

Unsafe code occurrences in `debug_info/serialization.rs` and `debug_info/types.rs` fall from 20 to 4.

`OptionC` was added by #3398. In this branch, it is limited to `Idx` types and `as_ref` is removed. All current uses are index values, and `miden-protocol` does not call `as_ref`.

The roundtrip tests pass. A deterministic debug info fixture also has the same serialized bytes on #3398 and this branch.

We measured `PackageDebugInfo::read_from_bytes` and `Package::debug_info()`. The second measurement matches the benchmark in the #3398 summary. The patches and full command are in the [benchmark gist](https://gist.github.com/huitseeker/a361a9e5e8e6848db8d26a254179366e).

```bash
MIDEN_DEBUG_INFO_BENCH_SECONDS=15 \
  cargo bench --profile optimized \
  --bench deserialize_core_lib_debug_info -- --noplot
```

| Revision | `read_from_bytes` | `Package::debug_info()` |
| --- | ---: | ---: |
| `main` at `f7d8549000` | 17.924 ms | 18.406 ms |
| PR #3398 at `962d427284` | 0.521 ms | 1.475 ms |
| This change at `d029cf4336` | 0.532 ms | 1.472 ms |

On `Package::debug_info()`, #3398 and this change are both about 92% faster than
`main`. The central estimate for this change was 0.2% lower than #3398. Its
`read_from_bytes` estimate was 2.2% higher.

The speedup from #3398 remains without the broad raw memory casts. This result
does not tell us how much of that speedup comes from string interning.

### DIFF
--- a/crates/mast-package/src/debug_info/serialization.rs
+++ b/crates/mast-package/src/debug_info/serialization.rs
@@ -791,7 +791,7 @@ mod tests {
     use core::cell::Cell;
 
     use miden_assembly_syntax::ast::DebugVarLocation;
-    use miden_core::Word;
+    use miden_core::{Felt, Word};
     use miden_debug_types::{ByteIndex, ColumnNumber, LineNumber, Location, Uri};
 
     use super::*;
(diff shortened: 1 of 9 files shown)


crates/mast-package/src/debug_info/types.rs
```diff
@@ -231,53 +249,48 @@ impl SourceNodeIdMarker for DebugSourceNodeId {}
...
+#[repr(transparent)]
+#[derive(
+    Copy,
+    Clone,
+    zerocopy::FromBytes,
+    zerocopy::Immutable,
+    zerocopy::IntoBytes,
+    zerocopy::KnownLayout,
+)]
+pub struct OptionC<T: Idx> {
```
bitwalker: I would rename this as `OptionalIndex` rather than `OptionC`, since it now bakes in the assumption that the input type implements `Idx` and is representable as a `u32`

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -20,90 +23,109 @@ use super::{
...
+    assert!(align_of::<DebugSourceNodeId>() <= POD_BUFFER_ALIGNMENT);
+    assert!(align_of::<DebugErrorMessage>() <= POD_BUFFER_ALIGNMENT);
+    assert!(align_of::<DebugSourceAsmOp>() <= POD_BUFFER_ALIGNMENT);
 };
 
+/// Wire form of an optional index. Byte arrays make little-endian encoding explicit, while keeping
+/// invalid discriminants representable for later validation.
+#[repr(C)]
+#[derive(Clone, Copy, zerocopy::FromBytes, Immutable, IntoBytes, KnownLayout)]
+struct WireOptionU32 {
```
bitwalker: This is extremely redundant IMO (i.e. it duplicates `OptionC`, solely to represent `u32` as raw bytes). We already require `#[cfg(target = "little_endian")]`, so `u32` is already explicitly little-endian by definition, so it is not at all clear to me that this makes the code clearer and easier to reason about.

crates/mast-package/src/debug_info/serialization.rs
```diff
@@ -20,90 +23,109 @@ use super::{
...
+#[derive(Clone, Copy, zerocopy::FromBytes, Immutable, IntoBytes, KnownLayout)]
+struct WireDebugFunctionInfo {
+    mast_root: [[u8; 8]; 4],
+    source_node: WireOptionU32,
+    type_idx: WireOptionU32,
+    linkage_name_idx: WireOptionU32,
+    name_idx: [u8; 4],
+    file_idx: [u8; 4],
+    line: [u8; 4],
+    column: [u8; 4],
```
bitwalker: The vast majority of these fields either already implement `zerocopy` traits, or are integer types which are concretely little-endian, as we only implement serialization for little-endian targets. I would make this struct a mirror of `DebugFunctionInfo`, but with `mast_root` represented as an array of `u64` (corresponding to the canonical form of the felts of the word) - that solves the primary risk with serializing `DebugFunctionInfo` directly, which is the reliance on the internal representation of `Felt`.

bitwalker (commented):
I'm largely in favor of this, with a couple of nits before we merge this into #3398

crates/mast-package/src/debug_info/serialization.rs
huitseeker: Simplified in afd152c

crates/mast-package/src/debug_info/serialization.rs
huitseeker: OK. `WireDebugFunctionInfo` now mirrors `DebugFunctionInfo`, with only `mast_root` stored as `[u64; 4]`. `Decode` still validates each field element. I also added a byte-level test that pins the v2 layout, which is really optional at this point.

PR #3433, opened 2026-07-27
perf: enable simd128 Plonky3 backend for WASM builds

### LINKED ISSUE #3104: Add wasm32 SIMD backend to p3-goldilocks to close the WASM-vs-native prove gap
## Context

The Miden VM, via `miden-air → p3-goldilocks` (Plonky3 0.5.x), uses Goldilocks (`p ≡ 2^64 − 2^32 + 1`) as its STARK field. Plonky3 ships hand-tuned SIMD backends for `x86_64` (AVX2 / AVX-512 IFMA) and `aarch64` (NEON) but falls back to scalar code on `wasm32-unknown-unknown`.

For the in-browser prover (the path used by `@miden-sdk/miden-sdk` and downstream wallet integrations), this scalar fallback is the single largest contributor to the WASM-vs-native prove-time gap.

## Measurements

Bench: [`miden-prover-bench`](https://github.com/WiktorStarczewski/miden-prover-bench) — 10 send/consume cycles per scheme, public notes, testnet, MacBook with M-series Apple Silicon, `navigator.hardwareConcurrency = 10`. The bench inserts a 1.5s cooldown between cycles to suppress sustained-power thermal drift; numbers below are medians from steady-state cycles.

### Native Rust client (`miden-client` 0.14.5, aarch64-apple-darwin, release + LTO + cgu=1)

```
Cycles summary — variant ecdsa
phase    | n  |   avg ms |   med ms |   min ms |   max ms
---------+----+----------+----------+----------+----------
send     | 10 |    658.7 |    642.9 |    630.4 |    759.2
consume  | 10 |    646.9 |    647.8 |    638.2 |    653.3

Cycles summary — variant falcon
phase    | n  |   avg ms |   med ms |   min ms |   max ms
---------+----+----------+----------+----------+----------
send     | 10 |   1702.0 |   1694.9 |   1621.1 |   1842.0
consume  | 10 |   1696.1 |   1691.6 |   1639.2 |   1756.1
```

### WASM-MT release in Chrome (`@miden-sdk/miden-sdk` 0.14.6 dual-build, MT subpath, `+simd128` with LLVM auto-vectorization disabled, `optimize_for_size` patched out of `rollup-plugin-rust`)

```
Cycles summary — variant mt (ECDSA)
phase    | n  |   med ms |   p25 ms |   p75 ms |   iqr ms
---------+----+----------+----------+----------+----------
send     | 10 |   3989.1 |   3866.3 |   4122.6 |    256.4
consume  | 10 |   3739.4 |   3535.3 |   4238.4 |    703.1

Cycles summary — variant mt (Falcon)
phase    | n  |   med ms |   p25 ms |   p75 ms |   iqr ms
---------+----+----------+----------+----------+----------
send     | 10 |   6621.9 |          |          |    357.0
consume  | 10 |   6628.8 |          |          |    225.0
```

### WASM tax (median ratio, WASM-MT release / native)

| Phase | Native | WASM-MT | Ratio |
|---|---|---|---|
| ECDSA send | 643 ms | 3989 ms | **6.20×** |
| ECDSA consume | 648 ms | 3739 ms | **5.77×** |
| Falcon send | 1695 ms | 6622 ms | 3.91× |
| Falcon consume | 1692 ms | 6629 ms | 3.92× |

ECDSA's WASM tax is markedly larger than Falcon's. The reason is structural — see decomposition below.

## Why the ECDSA tax is so large

ECDSA signature verification in Miden's MASM (\`miden-vm/crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm\`) is **deferred via host precompile**:

\`\`\`
pub proc verify_prehash_impl
    emit.ECDSA_VERIFY_EVENT
    # => [pk_ptr, digest_ptr, sig_ptr, ...]

    # Read verification result from advice stack (provided by the host)
    adv_push.1
    ...
\`\`\`

The actual secp256k1 elliptic-curve math runs off-circuit; the VM only emits an event and reads the boolean result back. So ECDSA's WASM prove time is **almost entirely STARK machinery** — trace generation, LDE, FRI commitments, RPO/RPO Merkle hashing — with a thin Keccak + advice-handling wrapper. Every one of those phases is Goldilocks-bottlenecked.

For comparison, the Falcon MASM (\`falcon512_poseidon2.masm\`) is 700 lines doing real polynomial verification over GF(12289), in-circuit. Its WASM prove time = STARK machinery (Goldilocks-bottlenecked) + in-circuit Falcon polynomial work (operates over a 14-bit prime, fits cleanly in \`i32x4\` lanes — already benefits from \`+simd128\` because WASM has native 32×32→64 widening multiply).

This explains the asymmetric WASM tax: Falcon has a portion of its prove that already benefits from current WASM SIMD; ECDSA does not.

## Proposed fix

Add a \`wasm32\` SIMD backend to \`p3-goldilocks\` mirroring the existing \`x86_64\` and \`aarch64\` siblings. New file:

\`\`\`
p3-goldilocks/src/wasm32/wasm_simd.rs
\`\`\`

Wired in \`lib.rs\` via:

\`\`\`rust
#[cfg(all(target_arch = \"wasm32\", target_feature = \"simd128\"))]
mod wasm32;
\`\`\`

The hard primitive is the missing widening 64×64 multiply: WASM v128 has no \`i64x2\` widening mul. Workaround: Karatsuba on 32-bit halves using \`i64x2.extmul_low_i32x4_u\` / \`i64x2.extmul_high_i32x4_u\`, which *are* native — they provide the 32×32→64 widening primitive. Reassemble high/low halves of the 128-bit product, then apply Goldilocks's Solinas-style reduction (same logic as the AVX2 / NEON backends, just reconstructed from the 32-bit widening primitive).

Reference impls in the same repo:
- \`p3-goldilocks/src/x86_64/avx2.rs\` — uses \`_mm256_mul_epu32\` + \`_mm256_srli_epi64\`
- \`p3-goldilocks/src/aarch64/neon.rs\` — uses \`vmull_u32\` (native widening 32×32→64) and \`vqdmulhq_s32\`

The \`aarch64\` impl is the closer template for a \`wasm32\` port: same lane width, similar instruction flavor, similar reduction structure. The wasm32 impl differs only in how the widening primitive is reconstructed.

## Estimated impact

Based on the data above:

| Phase | Today (WASM-MT release) | Projected post-PR | Expected speedup |
|---|---|---|---|
| ECDSA send | 4.0 s | 2.0–2.5 s | **1.6–2.0×** |
| ECDSA consume | 3.7 s | 1.9–2.3 s | **1.6–2.0×** |
| Falcon send | 6.6 s | 4.5–5.5 s | 1.2–1.5× |
| Falcon consume | 6.6 s | 4.5–5.5 s | 1.2–1.5× |

ECDSA gets the bigger relative speedup because it's almost pure STARK machinery (Goldilocks-bottlenecked). Falcon's improvement is smaller because a larger fraction of its prove is the in-circuit Falcon-prime polynomial work that already benefits from current \`+simd128\`.

Native target (post-PR) is unaffected by definition — this only changes the wasm32 codegen path.

All downstream Plonky3 consumers benefit, not just Miden — the broader open-source ecosystem (RiscZero, SP1, etc., to the extent they consume \`p3-goldilocks\`) sees the same wins on browser deployment.

## References

- web-sdk MT-WASM PR: [0xMiden/web-sdk#134](https://github.com/0xMiden/web-sdk/pull/134)
- Bench harness + measurement scripts: https://github.com/WiktorStarczewski/miden-prover-bench
- Native Rust bench (cross-checked native baseline): companion harness using \`miden-client\` 0.14.5 directly
- Plonky3 Goldilocks AVX2 impl: https://github.com/Plonky3/Plonky3/blob/main/goldilocks/src/x86_64/avx2.rs
- Plonky3 Goldilocks NEON impl: https://github.com/Plonky3/Plonky3/blob/main/goldilocks/src/aarch64/neon.rs
- ECDSA MASM (deferred-precompile evidence): \`miden-vm/crates/lib/core/asm/crypto/dsa/ecdsa_k256_keccak.masm\`

Nashtare:
## Describe your changes

Activate the Plonky3 SIMD128 backend for WASM builds.
Gives ~17% speed-up for Blake3 / ~30% for Poseidon2 (running the 4 ECDSAs / 100 Keccak precompile example).

closes #3104
(note that there is still a 1.81x / 2.87x difference with a native run for Blake3 / Poseidon2 respectively)

- Updated `CHANGELOG.md`


Nashtare:
I'm tracking additional improvements opportunities on the Plonky3 side, when they land we may integrate them in a patch release for miden-vm to benefit from those quickly.

bobbinth:
> note that there is still a 1.81x / 2.87x difference with a native run for Blake3 / Poseidon2 respectively

To check my understanding: this means that generating proofs natively using Poseidon2 is about 3x faster than generating them in the browser on the same machine, right?

Nashtare:
Correct (although this was measured on a single thread, I'm not familiar with WASM multi-threading capabilities, I don't think there are any?). Note that the comparison is with NEON vectorization on my M4 pro which is also fairly optimized. An AVX2 machine would have a smaller gap between native and WASM.

There's an open PR on Plonky3 that speeds-up things further with WASM (~30% for Poseidon2 hashing / 28% to 43% faster `sum_array` and 1.8x to 2.6x faster `dot_product`) so this would help narrow the gap.

bobbinth:
> although this was measured on a single thread, I'm not familiar with WASM multi-threading capabilities, I don't think there are any?

I believe @WiktorStarczewski was able to turn on multi-threading in WASM, but I think it is done further downstream (at the environment level). The speedup was quite significant, IIRC.

WiktorStarczewski:
> > although this was measured on a single thread, I'm not familiar with WASM multi-threading capabilities, I don't think there are any?
> 
> I believe @WiktorStarczewski was able to turn on multi-threading in WASM, but I think it is done further downstream (at the environment level). The speedup was quite significant, IIRC.

Yes it's working nicely (https://github.com/0xMiden/web-sdk#single-threaded-vs-multi-threaded-proving), the performance gain was about 3-4x.

> There's an open PR on Plonky3 that speeds-up things further with WASM (~30% for Poseidon2 hashing / 28% to 43% faster sum_array and 1.8x to 2.6x faster dot_product) so this would help narrow the gap.

Awesome!

.cargo/config.toml
```diff
@@ -160,4 +160,4 @@ xclippy-fix = [
 ]
 
 [target.wasm32-unknown-unknown]
-rustflags = ['--cfg', 'getrandom_backend="unsupported"']
+rustflags = ['--cfg', 'getrandom_backend="unsupported"', '-C', 'target-feature=+simd128']
```
huitseeker, line 163: This file does not propagate. Cargo only loads `.cargo/config.toml` from the working directory and its parents, and web-sdk builds the published Miden crates as dependencies.  The default ST artifact has no `+simd128`; the MT artifact already sets it. 

So the change turns on the Plonky3 backend for this workspace's no-std build, but neither shipped browser artifact changes. 

I think a README update indicating how downstream can turn this on would definitely be appreciated.

huitseeker (commented):
Would be great to have a WASM runtime check for the simd backend.

`make build-no-std` only compiles it, and the test jobs run native AVX or NEON code. A small test that compares packed Goldilocks and Poseidon2 results on the scalar path would catch target-only arithmetic bugs before release.

.cargo/config.toml
Nashtare: I've updated the readme and added a CI job for this. Let me know if the changes now look good to you!

.github/workflows/build.yml
```diff
@@ -69,6 +69,29 @@ jobs:
...
+        uses: ./.github/actions/cleanup-runner
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install --no-self-update
+          rustup target add wasm32-wasip1
+      - name: Install wasmtime
+        run: |
+          curl -sSfL https://github.com/bytecodealliance/wasmtime/releases/download/v45.0.2/wasmtime-v45.0.2-x86_64-linux.tar.xz \
+            | tar -xJ
+          echo "$PWD/wasmtime-v45.0.2-x86_64-linux" >> "$GITHUB_PATH"
```
huitseeker: Another option is to reuse the pinned `cargo-binstall` setup already used in `lint.yml`. This removes the manual download and keeps the Wasmtime version exact.

```suggestion
      - name: Install cargo-binstall
        uses: taiki-e/install-action@a6b2e2dcd845ddd7f509ce4f3ed3d922b80cc5d9 # v2.84.0
        with:
          tool: cargo-binstall
      - name: Install wasmtime
        run: cargo binstall --no-confirm --strategies crate-meta-data wasmtime-cli@45.0.2
```

.github/workflows/build.yml
```diff
@@ -69,6 +69,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - name: Cleanup large tools for build space
+        uses: ./.github/actions/cleanup-runner
+      - name: Install Rust toolchain
+        run: |
+          rustup toolchain install --no-self-update
+          rustup target add wasm32-wasip1
```
huitseeker, line 86: Could we use the pinned `dtolnay/rust-toolchain` action here? It installs the exact workspace toolchain and the WASI target in one step.

```suggestion
      - name: Install Rust toolchain
        uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master
        with:
          toolchain: 1.96.1
          targets: wasm32-wasip1
```

README.md
```diff
@@ -54,6 +54,8 @@ In the coming months we plan to finalize the design of the VM and implement supp
 
 Miden VM is written in pure Rust and can be compiled to WebAssembly. Rust's `std` standard library is linked by default for most crates. To compile to one of the two `wasm32` supported targets, use `cargo`'s `--no-default-features` flag to ensure Rust's standard library isn't linked (*i.e. compiling in `no_std`).
 
+This workspace's [`.cargo/config.toml`](.cargo/config.toml) sets `-C target-feature=+simd128` for the `wasm32-unknown-unknown` target, enabling Plonky3's SIMD128 backend for faster hashing (Blake3, Poseidon2) in WASM. Cargo only applies `.cargo/config.toml` to builds run from this workspace (or a directory nested under it); downstream consumers building Miden as a dependency (e.g. via `web-sdk`) must set this flag themselves, for example with `RUSTFLAGS="-C target-feature=+simd128"` or their own `.cargo/config.toml`, to get the same speedup.
```
huitseeker, line 57: /cc @WiktorStarczewski

huitseeker (commented):
Overall, looks good, some detailed simplifications inline

.github/workflows/build.yml
Nashtare: This is actually making zizmor check to go :x:, see https://github.com/0xMiden/miden-vm/actions/runs/30558687524/job/90925753853?pr=3433.

huitseeker: I think you can use `pinact` to figure out the correct pin for this action.

bobbinth (approved):
Looks good! Thank you! I'll merge as is. @huitseeker - may be good for you to still take a look and if there are any outstanding comments, we can address them in a follow-up.

PR #3437, opened 2026-07-29
feat(prover): add composable prover interface

### LINKED ISSUE #3420: Introduce a configurable Prover API
## Context

PR [#3293 — Import precompiles prover into VM workspace](https://github.com/0xMiden/miden-vm/pull/3293) integrates the Precompile VM into the standard execution-proof flow. As part of that work, verification moves to a first-class, configurable `Verifier`: persistent verification policy belongs to the verifier, while each call supplies the proof and public inputs being verified.

Proving has the same shape, but is still exposed through a growing family of free functions that receive execution and proving options on every call:

```rust
let (outputs, proof) = prove_sync(
    &program,
    stack_inputs,
    advice_inputs,
    &mut host,
    execution_options,
    proving_options,
)?;
```

This leaves an unnecessary asymmetry between proving and verification and makes the proving API increasingly parameter-heavy as we add final, partial, synchronous, asynchronous, and trace-based proving paths.

## Proposal

Introduce a configurable `Prover` type that owns the persistent policy used for execution and proof generation. Program-specific data remains explicit at each invocation.

```rust
let prover = Prover::new()
    .with_execution_options(execution_options)
    .with_hash_fn(HashFunction::Poseidon2);

let (outputs, proof) = prover.prove_sync(
    &program,
    stack_inputs,
    advice_inputs,
    &mut host,
)?;
```

`Prover::new()` should use the canonical execution and proving defaults.

The API should support both installing a complete proving profile and overriding individual settings:

```rust
let prover = Prover::new()
    .with_proving_options(proving_options);
```

```rust
let prover = Prover::new()
    .with_hash_fn(HashFunction::Poseidon2);
```

These forms should compose:

```rust
let prover = Prover::new()
    .with_proving_options(proving_options)
    .with_hash_fn(HashFunction::Poseidon2);
```

Builder methods are applied in order, so the last value configured for a setting wins. In the example above, `with_hash_fn` overrides the hash function from `proving_options`.

This gives callers a way to pass around complete proving profiles while keeping common configuration changes concise.

## Proving methods

The existing proving modes should become methods on `Prover`:

```rust
prover.prove(...).await;
prover.prove_sync(...);

prover.prove_partial(...).await;
prover.prove_partial_sync(...);

prover.prove_from_trace(trace_inputs);
prover.prove_partial_from_trace(trace_inputs);
```

The trace-based methods do not need a `_sync` suffix because trace construction and STARK proof generation are inherently synchronous.

Moving proving options onto `Prover` also removes the need for `TraceProvingInputs`, whose purpose is currently to pair `TraceBuildInputs` with a one-off `ProvingOptions` value:

```rust
// Current
prove_from_trace_sync(TraceProvingInputs::new(
    trace_inputs,
    proving_options,
))?;

// Proposed
prover.prove_from_trace(trace_inputs)?;
```

## Final and partial proving

“Partial” refers to the deferred proof material, not to the Miden VM execution trace or its STARK proof.

Both final and partial proving produce a complete Miden VM proof bound to the execution’s deferred root:

- Final proving uses the retained `DeferredState` to generate the corresponding Precompile VM STARK proof.
- Partial proving preserves the deferred state as a canonical `DeferredProof::Wire`, allowing that work to be proved later by a delegated or batching prover.

This remains meaningful for trace-based proving because `TraceBuildInputs` contains both the data needed to build the Miden VM trace and the deferred-state witness accumulated during execution. The witness cannot be reconstructed from the deferred root alone.

Thus:

```rust
prover.prove_from_trace(trace_inputs)
```

produces final deferred proof material, while:

```rust
prover.prove_partial_from_trace(trace_inputs)
```

produces a complete Miden VM proof while preserving the deferred witness for later proving.

### Consistency with `Verifier`

The proof lifecycle becomes easier to discover and understand:

```rust
Prover::new()
    .with_hash_fn(...)
    .prove_sync(...);

Verifier::new()
    .with_max_deferred_elements(...)
    .verify(...);
```

Both types own persistent policy, and both receive operation-specific data through their methods.

### Separate policy from invocation data

Execution limits and proving parameters usually remain constant across multiple proofs. Programs, inputs, and hosts vary per invocation.

Representing these separately makes that distinction explicit and avoids repeatedly threading configuration through every proving call.

### A coherent namespace for proving operations

Final, partial, sync, async, and trace-based proving are closely related operations. A `Prover` gives them a natural home and makes them discoverable through the type’s methods rather than through a growing crate-level function namespace.

### Easier extension

As proving configuration grows, focused builder methods can be added without expanding every proving function signature or introducing additional `prove_with_*` variants.

Complete `ProvingOptions` values remain useful for named or shared profiles, while focused methods such as `with_hash_fn` cover the common case.

## Follow-up

If this API works well, we should consider applying the same pattern more broadly to execution and other VM APIs.

In particular, a follow-up discussion could explore:

- a configurable `Executor` owning execution policy;
- whether `Prover` should compose such an executor;
- consistent configuration conventions across `Executor`, `Prover`, and `Verifier`;
- a stateful execution-session API for stepped or resumable execution.

Those ideas build naturally on this proposal, but the immediate goal is to give proving the same smooth, configuration-owning API shape introduced for verification in #3293.

adr1anh:
## Summary

Closes #3420.

This PR replaces the intermediate partial-proof APIs with a composable execution-proof lifecycle
built around the canonical `miden_core::program::ExecutionClaim`.

- Execution produces an in-memory `ExecutionWitness` containing a VM witness and optional hydrated
  precompile witness.
- VM-first proving returns either a complete VM-only proof or a deferred proof carrying passive
  precompile wire for later delegated proving.
- Full local proving proves all retained precompile work directly from the hydrated witness.
- Precompile proving may run separately and complete an existing VM proof without reproving it.
- Verification checks artifact shape, verifies every required STARK, and reports any authenticated
  precompile root that remains outstanding.
- Synchronous end-to-end proving retains #3407's overlapped execution/hasher trace-building path.
- Recursive VM verification uses the canonical claim encoding and commitment from #3422.

## Public lifecycle

```rust
impl Prover {
    /// Proves the VM and returns either a VM-only Complete proof or a Deferred proof.
    pub fn prove(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;

    /// Proves the VM and all retained precompile work directly from the in-memory witness.
    pub fn prove_full(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;

    /// Proves one singleton or merged hydrated precompile witness.
    pub fn prove_precompile(
        &self,
        witness: &PrecompileWitness,
    ) -> Result<PrecompileProof, ProverError>;
}

impl Verifier {
    pub fn verify(
        &self,
        claim: &ExecutionClaim,
        proof: &ExecutionProof,
    ) -> Result<VerificationOutcome, VerificationError>;

    pub fn verify_precompile(
        &self,
        proof: &PrecompileProof,
        expected_root: DeferredRoot,
    ) -> Result<u32, VerificationError>;
}
```

`ExecutionProof` has two public lifecycle states:

```rust
pub enum ExecutionProof {
    Deferred {
        vm: VmProof,
        precompile: DeferredStateWire,
    },
    Complete {
        vm: VmProof,
        precompile: Option<PrecompileProof>,
    },
}
```

A deferred proof contains a verified VM statement plus passive wire for the remaining precompile
work. The wire is transport data: proof decoding does not hydrate it or require a precompile
registry. Hydration is an explicit operation performed before delegated precompile proving.

`ExecutionProof::complete` transitions a deferred envelope by attaching a `PrecompileProof`. It does
not cryptographically verify either proof or establish compatibility. `Verifier` owns artifact-shape
acceptance, expected-root coverage, and cryptographic verification.

For deferred proofs, `Verifier::verify` verifies the VM STARK and returns its authenticated
precompile root as an outstanding obligation. The passive wire is not part of this verification.
For complete proofs, the verifier checks the VM/precompile shape, verifies every required STARK, and
returns the minimum component security level.

`VerificationOutcome` is `#[must_use]` because successful VM verification may still leave an
outstanding precompile obligation.

### Complete local proving

```rust
let witness = processor.execute_for_proving_sync(&program, &mut host)?;
let claim = witness.claim();
let proof = Prover::new().prove_full(witness)?;
let outcome = Verifier::new().verify(&claim, &proof)?;
assert!(outcome.is_complete());
```

### VM-first proving and later completion

```rust
let witness = processor.execute_for_proving_sync(&program, &mut host)?;
let claim = witness.claim();
let prover = Prover::new();
let proof = prover.prove(witness)?;

let outcome = Verifier::new().verify(&claim, &proof)?;
assert!(!outcome.is_complete());

let ExecutionProof::Deferred { precompile: wire, .. } = &proof else {
    unreachable!("this execution contains deferred precompile work");
};
let witness = precompile_witness_from_wire(wire)?;
let precompile = prover.prove_precompile(&witness)?;
let proof = proof.complete(precompile)?;

assert!(Verifier::new().verify(&claim, &proof)?.is_complete());
```

### Verifying a shared precompile proof directly

```rust
let security_level = Verifier::new().verify_precompile(&precompile_proof, expected_vm_root)?;
```

A shared precompile proof may contain compatible extra roots. Root order and duplicate multiplicity
are retained because they determine the aggregate precompile STARK statement.

## Trust boundaries

The lifecycle separates three operations:

1. **Transport:** canonical encoding preserves the supplied proof or wire representation under fixed
   allocation ceilings.
2. **Hydration:** `precompile_witness_from_wire` validates passive wire using the bundled precompile
   registry and constructs a hydrated `PrecompileWitness`.
3. **Verification:** `Verifier` validates artifact shape, checks expected-root coverage, and verifies
   the relevant VM and precompile STARKs.

Transport decoding does not claim that an artifact is structurally or cryptographically valid.
Representable malformed shapes can round-trip and are rejected when passed to `Verifier`. Generic
serde support follows the same representation-preserving contract and does not replace verifier
validation.

## Migration

- Use the canonical `miden_core::program::ExecutionClaim`; the branch-local competing claim is
  removed. Its encoding, commitment, and recursive-verification semantics follow #3422.
- `TraceWitness` is now private `TraceReplay`.
- `ExecutionTrace` is now `VmTrace`.
- `execute_trace_inputs*` is now `execute_for_proving*` and returns `ExecutionWitness`.
- Use `Prover::{prove, prove_full, prove_precompile}` instead of the removed async, partial, and
  trace-input free proving APIs.
- `ProvingOptions` is removed. Configure a `Prover` with `with_hash_fn`.
- The temporary free `prove_sync(&Prover, ...)` remains for optimized overlapped synchronous
  orchestration; later Executor work can own this path.
- Replace the old `ExecutionProof`/`DeferredProof` envelope with `VmProof`, `PrecompileProof`, and
  `ExecutionProof::{Deferred, Complete}`.
- Hydrate deferred wire with `miden_vm::precompile_witness_from_wire` before delegated precompile
  proving.
- `ExecutionProof::complete` attaches a precompile proof and returns `ExecutionProofError` only for an
  invalid lifecycle transition; use `Verifier` to establish proof validity.
- Use borrowed `Verifier::verify(&claim, &proof)` and inspect its `#[must_use]`
  `VerificationOutcome` instead of the removed free and partial verification APIs.
- Use `Verifier::verify_precompile` to verify a shared precompile proof against one expected VM root.
- `ExecutionProof::{to_bytes, read_from_bytes}` provide registry-independent canonical transport.

## Transport and resource policy

- `ExecutionProof::to_bytes` is infallible and preserves the public enum representation.
- `ExecutionProof::read_from_bytes` rejects invalid discriminants, non-canonical encodings, trailing
  bytes, and payloads exceeding fixed decoding ceilings without hydrating deferred wire.
- Standalone `VmProof` and `PrecompileProof` decoders reject trailing bytes.
- Each encoded STARK payload has a fixed 64 MiB allocation and verification ceiling.
- Precompile proof root counts are bounded before allocating root storage or folding roots.
- Precompile witness merging accepts ordered singleton witnesses, preserves duplicates, and applies
  fixed aggregate deferred-state and root-count ceilings.
- `StarkProof` and `VmProof` report their shortest encoded sizes so budgeted sequence decoding accepts
  compact proofs.
- `PrecompileProof` reports the shortest transport-representable encoding. Empty-root shapes can be
  transported but are rejected by `Verifier`; the shortest verifier-acceptable singleton encoding is
  therefore larger than the transport minimum.
- Proof artifacts support serde representation round trips and an execution-proof serde fuzz target.
  Callers remain responsible for bounding generic serde payloads before allocation.
- #3458 tracks configurable protocol-level policy for complete envelopes, decoder preallocation, and
  limits below the fixed library safety ceilings.

## Recursive verification

Recursive VM verification consumes the canonical claim commitment from the operand stack. The proof
stream is stored in the advice map under `proof_request_key(verifier_root, claim_commitment)` and is
loaded by the verifier request flow. Successful recursive VM verification authenticates the carried
precompile root, which remains an obligation to be settled separately when non-empty.

## Remaining work / non-goals

- `Prover::prove_vm` remains private; this PR does not define a general delegated VM-worker transport
  protocol.
- #3458 owns configurable aggregate resource policy beyond the fixed library safety ceilings.
- This PR does not add the future `Executor`, a protocol settlement envelope, or a new registry
  identity model.

## Suggested review order

1. Proof transport and lifecycle states: `core/src/proof.rs`, `core/src/deferred/witness.rs`.
2. Verifier-owned artifact and STARK validation: `verifier/src/lib.rs`.
3. Execution witness and trace terminology: `processor/src/trace/`, `processor/src/fast/`.
4. Proving implementation: `prover/src/prover.rs`.
5. Façade and lifecycle flows: `miden-vm/src/`, `miden-vm/tests/integration/prove_verify.rs`.
6. Resource limits, serde/fuzz coverage, and documentation.

## Validation

- `RUSTFLAGS="-D warnings" cargo +stable xclippy`
- `cargo +nightly fmt --all --check`
- Commit hooks, including `cargo-shear`, pass.
- `miden-core`: 238 passed, 1 skipped.
- `miden-precompiles-prover`: 283 passed, 20 skipped.
- `miden-verifier`: 10 passed.
- Focused delegated and recursive `miden-vm` tests: 3 passed.
- `miden-prover` async compatibility: 2 passed.

## Checklist

- [x] Uses the canonical execution claim and recursive commitment semantics.
- [x] Reconciles façade, CLI, benchmarks, precompile prover, recursive verifier, and tests.
- [x] Removes superseded APIs instead of restoring them.
- [x] Keeps changed documentation within the 100-character convention.
- [x] Keeps all branch commits signed.
- [x] Preserves upstream changelog history and adds one #3437 lifecycle entry.
- [x] Documents transport, hydration, and verification as separate trust boundaries.
- [x] Merges `next` through `3b233fbb4` and preserves the factored ACE registry exports.


core/src/proof.rs
```diff
@@ -125,106 +126,170 @@ impl Deserializable for HashFunction {
...
+    fn write_into<W: ByteWriter>(&self, target: &mut W) {
+        self.proof.write_into(target);
+        self.precompiles_root.write_into(target);
     }
+}
 
-    /// Returns the deferred proof material associated with the Miden VM proof.
-    pub const fn deferred_proof(&self) -> &DeferredProof {
-        &self.deferred
+impl Deserializable for VmProof {
```
huitseeker, line 149: `Vec<VmProof>` uses `Deserializable::min_serialized_size()` before allocation, and the default is the 64-byte in-memory size even though the shortest encoding is 34 bytes. 

I think we need to override the minima for `StarkProof` and `VmProof` (plus optionally an all-settled batch round-trip test) to keep canonical encodings readable.

verifier/src/lib.rs
```diff
@@ -39,236 +37,218 @@ pub use exports::*;
...
-            wire,
-            max_deferred_elements,
-        )?),
-        DeferredProof::Empty | DeferredProof::Stark { .. } => {
-            Err(VerificationError::UnsupportedDeferredProof)
-        },
+    fn verify_precompiles(&self, proof: &PrecompilesProof) -> Result<u32, VerificationError> {
+        miden_precompiles_prover::verify_deferred(
+            proof.proof(),
+            proof.aggregate_root(),
```
huitseeker: Could we bound the constituent roots *before* folding them? 

`with_max_proof_size` only limits the STARK bytes. An untrusted proof can add _tons_ of extra roots, pass the shape checks, and make this call run one Poseidon2 fold per root before the byte-size check. 

I think we need a complete-proof byte limit or root-count limit at deserialization.

prover/src/prover.rs
```diff
@@ -0,0 +1,267 @@
...
+        witnesses: Vec<Option<PrecompilesWitness>>,
+    ) -> Result<Option<PrecompilesProof>, ProverError> {
+        let mut states = witnesses.into_iter().flatten().map(PrecompilesWitness::into_state);
+        let Some(mut aggregate) = states.next() else {
+            return Ok(None);
+        };
+
+        // Each witness was bounded during execution, but the first state's remaining budget is not
+        // a valid batch limit. Disable it; batch aggregation currently has no separate element cap.
+        aggregate.set_max_elements(usize::MAX);
```
huitseeker: Could this keep a real aggregate limit instead of resetting it to `usize::MAX`?  😱

verifier/src/lib.rs
```diff
@@ -57,217 +60,260 @@ pub mod recursive;
...
-/// # Errors
-/// Returns an error if:
-/// - The provided proof does not prove a correct execution of the claim.
-/// - The proof carries wire-backed deferred proof material, which is a partial/delegable form.
-/// - The proof's STARK-backed deferred proof, if present, does not verify against its public root.
-pub fn verify(proof: ExecutionProof, claim: ExecutionClaim) -> Result<u32, VerificationError> {
-    Verifier::default().verify(proof, claim)
+/// Result of fully verifying an execution proof and all supplied STARKs.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct VerificationOutcome {
```
huitseeker, line 299: Could we keep the old `#[must_use]` guard on this outcome? `verify()` now returns `Ok` for `ExecutionProof::Deferred`, so `Verifier::verify(&claim, &proof)?;` compiles and silently drops the outstanding precompile root. 

Marking `VerificationOutcome` as `#[must_use = "verification may leave an outstanding precompile obligation"]` would make callers inspect or explicitly discard it.

verifier/src/lib.rs
```diff
@@ -57,217 +60,260 @@ pub mod recursive;
...
-
-        Ok(precompile_security_level
-            .map(|level| miden_security_level.min(level))
-            .unwrap_or(miden_security_level))
+        claim: &ExecutionClaim,
+        proof: &ExecutionProof,
+    ) -> Result<VerificationOutcome, VerificationError> {
+        match proof {
+            ExecutionProof::Deferred { vm, precompile } => {
+                validate_deferred(vm, precompile)
```
huitseeker: nit: Could we expose one proof-structure validation entry point from `miden-core` and call it here? The helpers in this crate duplicate the invariants already checked in `core/src/proof.rs`, leaving about 40 lines of shape checks that can drift apart.

core/src/proof.rs
huitseeker: I think this concern is still live

verifier/src/lib.rs
huitseeker: This concern is still live

huitseeker (commented):
I think the aggregate limit issue is still unsolved as well.

bobbinth:
> ```rust
> impl Prover {
>     pub fn prove(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;
>     pub fn prove_full(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;
>     pub fn prove_vm(&self, witness: VmWitness) -> Result<VmProof, ProverError>;
>     pub fn prove_precompile(
>         &self,
>         witness: &PrecompileWitness,
>     ) -> Result<PrecompileProof, ProverError>;
> }
> ```

In the above, what's the difference between `prove()`, `prove_full()`, and `prove_vm()`?

adr1anh:
> > ```rust
> >
> > impl Prover {
> >
> >     pub fn prove(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;
> >
> >     pub fn prove_full(&self, witness: ExecutionWitness) -> Result<ExecutionProof, ProverError>;
> >
> >     pub fn prove_vm(&self, witness: VmWitness) -> Result<VmProof, ProverError>;
> >
> >     pub fn prove_precompile(
> >
> >         &self,
> >
> >         witness: &PrecompileWitness,
> >
> >     ) -> Result<PrecompileProof, ProverError>;
> >
> > }
> >
> > ```
> 
> 
> 
> In the above, what's the difference between `prove()`, `prove_full()`, and `prove_vm()`?

Here, `prove` would be better named `prove_partial`, which returns an ExecutionProof with a unproven precompile state. If there are no precompiles, then it returns a complete proof, with an empty precompile proof.

The `prove_full` will also run the precompile prover to produce a full proof. 

In contrast, `prove_vm` is a convenience function which proves a VM witness, without access to the precompile witness. I understood that we might have a VM-only delegated prover, which this prover would invoke.

core/src/proof.rs
adr1anh: Fixed. StarkProof now reports the minimum size of its encoded byte vector plus hash-function tag, and VmProof composes that with the deferred-root encoding. This gives VmProof its actual 34-byte minimum rather than its 64-byte in-memory size, so budgeted Vec<VmProof> decoding continues to accept compact canonical encodings.

I also added a regression test that checks the reported StarkProof and VmProof minima against their shortest canonical encodings.

verifier/src/lib.rs
adr1anh: Fixed. ExecutionProof now exposes validate_structure() from miden-core, and Verifier::verify calls it before cryptographic verification.

The duplicate verifier-local deferred/complete shape helpers have been removed, so constructors, completion, and verification now share the same structural invariant implementation.

verifier/src/lib.rs
adr1anh: Fixed via the root-count option suggested here.

PrecompileProof now has a hard library safety ceiling of 4,096 ordered constituent roots. Binary decoding reads and checks the root count before collecting the root vector, and PrecompileProof::from_parts enforces the same invariant for proofs constructed directly in memory. aggregate_root() can therefore never perform an unbounded number of folds.

The same ceiling is applied to precompile-witness decoding and merging. Order and duplicate multiplicity remain unchanged, and the exact boundary is covered by tests.

This is an allocation/folding safety ceiling, not the final protocol acceptance policy. #3458 will add configurable lower root limits together with complete-envelope and early inner-STARK byte limits.

prover/src/prover.rs
adr1anh: Fixed. PrecompileWitness::merge now requires an explicit aggregate max_elements instead of resetting the accumulator to u32::MAX.

The first witness is rejected if it already exceeds the supplied aggregate limit, and subsequent states are imported through the existing insertion accounting, so the same limit applies cumulatively to the complete merged state. Deduplicated nodes remain budget-free, while newly imported and aggregate nodes consume the remaining budget.

Existing lifecycle call sites currently pass DEFAULT_MAX_DEFERRED_ELEMENTS. The stacked #3458 work will decide where the configurable aggregate policy is owned across the prover, decoder, verifier, and facade.

verifier/src/lib.rs
adr1anh: Fixed as suggested. VerificationOutcome is now marked must_use with the message "verification may leave an outstanding precompile obligation". This preserves successful deferred verification while warning callers that silently discard the authenticated outstanding precompile root.

adr1anh:
@huitseeker Agreed that the full aggregate policy was still unresolved. I have split the work into:

1. Direct safety fixes in this PR:
   - hard constituent-root ceiling enforced before root-vector allocation and aggregate-root folding;
   - explicit aggregate element bound for witness merging;
   - rejection of an already-oversized merge accumulator.
2. A stacked follow-up in #3458:
   - complete execution-proof envelope limits;
   - early inner-STARK byte limits;
   - configurable acceptance limits below the hard ceilings;
   - consistent prover/verifier/facade and ingestion policy.

This removes the unbounded operations introduced by this PR without defining the broader resource-policy interface in the middle of the prover lifecycle review.

core/src/proof.rs
```diff
@@ -125,107 +129,151 @@ impl Deserializable for HashFunction {
...
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
-        self.miden.write_into(target);
-        self.deferred.write_into(target);
+        self.proof.write_into(target);
+        self.roots.write_into(target);
     }
 }
 
-impl Deserializable for ExecutionProof {
+impl Deserializable for PrecompileProof {
```
huitseeker, line 190: `PrecompileProof` still inherits the 56-byte in-memory default for `min_serialized_size()`, although its smallest valid encoding is 35 bytes. 

This makes the recommended budgeted decoder reject valid collections: two singleton proofs encode in 71 bytes, but `Vec<PrecompileProof>::read_from_bytes_with_budget(..., 71)` fails with `requested 2 elements but reader can provide at most 1`. 

We should override the minimum for the required one-root shape (and optionally add a budgeted two-proof round-trip test).

prover/README.md
```diff
@@ -1,48 +1,90 @@
...
+Malformed cross-artifact `ExecutionProof` values can therefore serialize and decode, but full
+verification rejects inconsistent structure or invalid STARKs.
+
+Ordinary façade callers can decode with the standard bundled registry via
+`miden_vm::read_execution_proof_from_bytes(bytes)`. Custom-precompile callers use
+`ExecutionProof::read_from_bytes(bytes, registry, max_elements)` directly.
+
+The deferred-wire canonical decode-and-reencode policy remains unchanged. The outer execution-proof
+decoder now rejects trailing bytes and encodings that do not round-trip exactly. Hydration takes an
+explicit deferred-element bound, and decoding retains its per-allocation ceiling. Merging resets the
```
huitseeker: This paragraph now describes the opposite of `74a5bed7`: `PrecompileWitness::merge` requires `max_elements`, enforces it across the whole merged state, and caps the list at `MAX_PRECOMPILE_ROOTS`. 

The same stale policy appears in `docs/src/design/deferred/semantics.md`, and `docs/src/design/stack/precompiles.md` still shows the old one-argument call. 

Can you update all three together?

core/src/proof.rs
```diff
@@ -234,146 +282,284 @@ impl Deserializable for ExecutionProof {
...
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for PrecompileProof {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        #[derive(Deserialize)]
+        struct Fields {
+            proof: StarkProof,
+            roots: Vec<DeferredRoot>,
+        }
+
+        let Fields { proof, roots } = Fields::deserialize(deserializer)?;
```
huitseeker: nit: The 4,096-root check in `from_parts` runs only after this `Vec` has been fully deserialized, so a serde/Postcard caller can reserve and parse more roots than the cap before rejection. 

There's no in-repo serde consumer of `PrecompileProof`, and the standard `ExecutionProof` transport uses the early-checked binary decoder, so this is not currently reachable here. 

Still, could we either use a bounded visitor or document this serde path as trusted-only?

miden-vm/tests/integration/prove_verify.rs
```diff
@@ -69,18 +71,41 @@ fn assert_recursive_verify(
...
-            # Initial stack: [CLAIM_COMMITMENT].
-            dupw
-            procref.vm::verify_vm_proof exec.sys::build_proof_request_key
-            adv.push_mapval dropw
+            # Initial stack: [claim_ptr].
+
+            # Copy the claim encoding P | K | I | O (40 felts) into the claim region
+            # (claim_ptr = 4096); the kernel digest witness travels in the advice map.
+            push.40 push.4096
+            exec.copy_advice_to_mem
```
huitseeker: I think here `generate_advice_inputs` puts the claim commitment on the operand stack and only the proof stream on advice. This stale copy consumes the first 40 proof felts, causing both recursive CI tests to fail. Without the copy, this should work calling`verify_vm_proof` directly?

core/src/proof.rs
```diff
@@ -125,261 +129,436 @@ impl Deserializable for HashFunction {
...
+    fn read_from<R: ByteReader>(source: &mut R) -> Result<Self, DeserializationError> {
+        let proof = StarkProof::read_from(source)?;
+        let precompile_root = DeferredRoot::read_from(source)?;
+        Ok(Self::from_parts(proof, precompile_root))
     }
 
-    /// Returns the deferred proof material associated with the Miden VM proof.
-    pub const fn deferred_proof(&self) -> &DeferredProof {
-        &self.deferred
+    fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
```
Al-Kindi-0, line 156: Should this also reject trailing bytes, like `ExecutionProof::read_from_bytes` does?

core/src/proof.rs
```diff
@@ -125,261 +129,436 @@ impl Deserializable for HashFunction {
...
+                max: MAX_PRECOMPILE_ROOTS,
+            }));
+        }
+        let roots = source
+            .read_many_iter::<DeferredRoot>(root_count)?
+            .collect::<Result<Vec<_>, _>>()?;
+        Self::from_parts(proof, roots).map_err(invalid_proof_shape)
     }
 
     fn read_from_bytes(bytes: &[u8]) -> Result<Self, DeserializationError> {
```
Al-Kindi-0, line 205: Should this also reject trailing bytes, like `ExecutionProof::read_from_bytes` does?

Al-Kindi-0 (approved):
LGTM

adr1anh:
I want to call out that the latest push is larger than a normal review-fix round.

The two commits that were already on the PR when it was approved remain unchanged:

- `fce89c4f0` — initial composable proving implementation
- `12fee12d2` — documentation and fixed proof limits

I rewrote only the local history above those commits. The previous local history mixed architectural simplification, review fixes, an upstream merge, and benchmark compatibility into several unsigned or overly broad commits. The new top is signed and grouped by rationale.

More importantly, the new commits do change part of the public contract. I do not want the existing approval to be interpreted as approval of that changed contract. I am asking for a focused re-review of the commits below, particularly the first and fourth.

## Commit-by-commit rationale

### `1f4222d9a` — `refactor(prover): simplify delegated proof lifecycle`

This consolidates the work previously spread across `8cf6f4fc8`, `53155ac12`, and `372c98888`.

The main change is to separate three concerns more explicitly:

1. proof and witness transport preserves the supplied artifact;
2. hydration validates deferred wire under a selected registry;
3. the verifier owns proof structure and cryptographic acceptance.

In particular:

- `ExecutionProof::Deferred` carries passive `DeferredStateWire`;
- generic proof decoding does not hydrate that wire or require a registry;
- `ExecutionProof::complete` performs a lifecycle transition by attaching a proof, not by cryptographically validating it;
- `Verifier` is the authority that checks VM/precompile compatibility and verifies the relevant STARKs;
- shared precompile proofs can settle an individual VM obligation while retaining their complete ordered root list;
- duplicated proof-shape checks and intermediate wrappers were removed.

This removes a substantial amount of API and transport machinery while keeping the same end-to-end proving flows.

The tradeoff is important: a deferred envelope can temporarily contain wire unrelated to the VM root, because deferred verification authenticates only the root carried by the VM STARK. The wire becomes trusted prover input only after explicit hydration, and a completed proof remains invalid until `Verifier` accepts it.

I think this is a cleaner trust boundary, but it is a real contract change and is the main place where I would appreciate renewed review.

### `5ac4fc5de` — `fix(verifier): restore recursive proof request flow`

This is an independent correctness fix for the recursive verifier tests.

The previous helper placed the claim commitment on the operand stack and the proof stream in the advice map. The test then copied 40 felts from the advice tape as if the claim encoding were still there, consuming the first part of the proof stream and causing both recursive verification tests to fail.

The fix:

- uses `generate_request_inputs`;
- leaves the claim commitment on the operand stack;
- derives the proof request key;
- loads the proof stream from the advice map;
- calls `verify_vm_proof` through the intended request flow.

This commit does not depend on accepting the broader contract change. Reverting it would reintroduce the recursive verification failure, so I consider this one required regardless of what we decide about the surrounding refactor.

### `9c20f4638` — `fix(prover): tighten delegated proof transport contract`

This consolidates the documentation work from `231367588` and the validation work from `7009f991f`.

It:

- documents the distinction between passive transport, hydration, structural validation, and cryptographic verification;
- documents the supported local, deferred, delegated, and merged proving flows;
- bounds merged precompile witnesses instead of resetting aggregate accounting;
- keeps ordered roots and duplicate multiplicity explicit;
- aligns the crate READMEs and design documentation with the implemented lifecycle.

This commit is necessary if we keep the passive-wire/verifier-owned model from `1f4222d9a`. If we reject that model, this commit should be reverted with it rather than partially retained with contradictory documentation.

### `5280b8f73` — merge of `origin/next`

This is only the upstream synchronization boundary. Its resolved tree matches the previous local merge.

`next` advanced once more immediately after the push to `77282b479` (`Add the factored precompile VM ACE registry`). That currently leaves one merge conflict in `crates/precompiles-prover/src/lib.rs`; I will refresh this merge and resolve that conflict before the PR is merged.

### `b7db10c89` — `fix(prover): complete delegated proof review follow-ups`

This collects the non-benchmark review and hardening work that was previously mixed into `265ef4e8`.

It addresses the recent review concerns around untrusted proof artifacts:

- reports encoded minima for `StarkProof` and `VmProof`;
- keeps budgeted collections of compact proofs decodable;
- rejects trailing bytes in standalone VM and precompile proof decoders;
- imposes fixed safety ceilings on STARK bytes and precompile root counts before large allocation or folding;
- applies the same hard ceilings to direct in-memory construction and verifier preflight;
- adds direct verification of a shared `PrecompileProof` against one expected VM obligation;
- extends lifecycle tests with reordered roots, missing duplicate roots, mutated VM roots, trailing proof bytes, and buffered/overlapped trace-fragment stress;
- adds serde round-trip coverage and an execution-proof serde fuzz target.

One point deserves explicit reviewer confirmation: `PrecompileProof::min_serialized_size()` is `3`, because transport decoding deliberately preserves an empty-root representation and leaves rejection to the verifier. The requested smallest structurally valid singleton encoding is `35` bytes. The current value is conservative and the 71-byte two-proof budget regression succeeds, but it reflects the new “transport preserves representation” contract rather than the previous structural-validity contract.

If we retain the old interpretation, I should instead reject empty roots during decoding and report `35`.

The serde/fuzz work is also a scope expansion relative to the earlier PR description. I think it is worthwhile because these artifacts now cross generic transport boundaries, but it can be dropped or moved to a follow-up if that is preferable.

### `18ceff2bf` — `fix(bench): preserve legacy prover benchmark metrics`

This is intentionally separate because it is operational compatibility rather than prover behavior.

The API rename changed the benchmark spelling from `execute_trace_inputs_sync` to `execute_for_proving_sync`. This commit accepts both spellings while emitting the stable current metric name, so existing scheduled/manual benchmark invocations and stored Criterion paths continue to work.

This is independently revertible and can be moved to a follow-up without affecting proof correctness.

## What changed in the contract

The main contract delta is:

| Previously described | Current implementation |
|---|---|
| Deferred proof retains a hydrated `PrecompileWitness` | Deferred proof retains passive `DeferredStateWire` |
| Proof serialization may hydrate/validate deferred state | Serialization preserves the artifact representation |
| Construction/completion establishes structural compatibility | Completion attaches artifacts; the verifier establishes compatibility |
| Core and verifier share structural validation | Verifier owns acceptance policy |
| Generic serde was a non-goal | Proof artifacts have serde round trips and fuzz coverage |

My rationale is that transport, hydration, and verification are different trust boundaries:

- transport should be bounded and canonical without requiring a registry;
- hydration should establish that deferred wire is meaningful under a selected registry;
- verification should establish the public VM/precompile claims cryptographically.

The cost is that the in-memory artifact types carry fewer static guarantees before verification. I think that tradeoff is reasonable, but it should be agreed explicitly rather than entering under the prior approval.

I will update the PR description so it states this contract accurately.

## Rollback options

These commits were separated so we retain options:

1. **Keep the new contract.** Review `1f4222d9a`, `9c20f4638`, and `b7db10c89` as one coherent lifecycle/trust-boundary change.
2. **Return to the previously approved contract.** Reset to `12fee12d2`, retain the recursive fix, and port only the narrow decoder/resource corrections required by the outstanding comments.
3. **Reduce peripheral scope.** Drop `18ceff2bf` and/or move serde fuzzing to follow-up PRs without changing the proving lifecycle.
4. **Keep only the correctness fix.** `5ac4fc5de` can stand independently if review capacity does not allow reopening the API contract today.

My preference is option 1 because the resulting API is smaller and the trust boundaries are clearer, but I am comfortable reverting to option 2 if reopening the contract at this stage creates too much review risk.

## Validation

The final source tree matches the previously tested local branch. Validation completed:

- workspace Clippy with warnings denied;
- nightly rustfmt;
- `miden-core`: 238 passed;
- `miden-verifier`: 10 passed;
- focused delegated and recursive VM tests: 3 passed;
- prover async compatibility: 2 passed.

I would especially appreciate feedback on:

1. passive `DeferredStateWire` versus hydrated `PrecompileWitness` in `ExecutionProof::Deferred`;
2. verifier-owned structural validation;
3. whether empty-root transport shapes should remain representable;
4. whether serde/fuzz and benchmark compatibility should remain in this PR.

core/src/deferred/witness.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        for witness in &witnesses {
+            if witness.roots.len() != 1 {
+                return Err(PrecompileWitnessError::NonSingleton);
+            }
+        }
+
+        let roots = witnesses.iter().map(|witness| witness.roots[0]).collect::<Vec<_>>();
+        let mut witnesses = witnesses.into_iter();
+        let mut state = witnesses.next().expect("non-empty witness input was checked above").state;
```
huitseeker: `PrecompileWitness::merge` keeps the first state's orphan nodes but imports only root-reachable nodes from later states. This makes the fixed element budget depend on witness order. I can make `[A, B]` fail while `[B, A]` succeeds. Could we compact the first state and add that regression?

prover/src/prover.rs
```diff
@@ -0,0 +1,271 @@
...
+                prove_stark(
+                    &config,
+                    core_matrix,
+                    chiplets_matrix,
+                    poseidon2_matrix,
+                    &public_values,
+                    &aux_inputs,
+                )
+            },
+            HashFunction::Keccak => {
```
huitseeker, line 131: The old integration suite covered all five VM hash configurations, but the new `assert_prove_verify` calls cover only Blake3 and Poseidon2. Could we retain a table-driven public `Prover::prove_full`/`Verifier::verify` round trip across every supported variant?

huitseeker (commented):
The proving boundaries ar much neater.
One issue: `PrecompileWitness::merge` retains unreachable nodes only from its first input, so the fixed deferred-state budget and merge success depend on witness order. 
Happy to stamp once this is addressed!

huitseeker (approved):
This looks great, thanks a lot!

README.md
```diff
@@ -41,7 +41,10 @@ Miden VM is a fully-featured virtual machine. Despite being optimized for zero-k
...
 - **Fast processor execution mode.** In addition to the trace-generating processor used for proof generation, Miden VM includes a fast processor that can execute programs at up to 320 MHz, enabling among other things rapid program testing and debugging.
-- **Precompiles.** Miden VM supports [precompiles](./docs/src/design/stack/precompiles.md), allowing programs to defer expensive computations to the host while still producing auditable evidence inside the STARK proof. This enables efficient verification of operations like signature schemes and hash functions that would otherwise be prohibitively expensive to execute natively in the VM.
+- **Precompiles.** Miden VM supports
+  [precompiles](./docs/src/design/stack/precompiles.md), allowing programs to defer expensive
+  computations to the host. VM verification authenticates the outstanding deferred root; an
+  aggregate precompile proof can settle compatible deferred executions later.
```
bobbinth, line 47: I think this description could be improved a bit - and maybe it deserves two separate bullet points (one about deferring computations and another about having a VM with specialized circuits). I'd also list the current set of precompiles we have (Keccak, ECDSA).

Also, we should update the "Planned features" section below since we already have recursive proof verification and, arguably, "better dubbing" as well.

@huitseeker maybe worth including this in the general README revamp as we also need to update the project structure section and probably refresh the benchmarks section too (would be great to include WASM performance there).

README.md
```diff
@@ -109,7 +112,13 @@ RAYON_NUM_THREADS=16 cargo run --profile optimized -p miden-vm-blake3-bench --bi
...
 ```
 
-The result is written to `target/blake3-nonregression/result.json`; the harness does not parse the `miden-vm run` or `miden-vm prove` text output. The benchmark records `execute_trace_inputs_sync`, `build_trace`, `prove_trace_sync`, and `e2e_prove`. The `e2e_prove` metric runs execution and trace generation on each sample, but only measures the prover span. The harness also proves and verifies the program once before timing proof-heavy axes.
+The result is written to `target/blake3-nonregression/result.json`; the harness does not parse
+the `miden-vm run` or `miden-vm prove` text output. The benchmark records
+`execute_for_proving_sync`, `build_trace`, `prove_trace_sync`, and `e2e_prove`. It also accepts the
+historical `execute_trace_inputs_sync` axis input, normalizing both spellings to the stable
+`execute_for_proving_sync` metric key. The `e2e_prove` metric runs execution and trace generation
+on each sample, but only measures the prover span. The harness also proves and verifies the program
+once before timing proof-heavy axes.
```
bobbinth, line 121: nit: we usually don't split paragraphs into multiple lines in README files.

PR #3438, opened 2026-07-29
refactor: decouple native handlers from processor state

adr1anh:
Native event and trace handlers currently depend on `miden-processor` and receive a processor-specific state view. This PR introduces `miden-event-handler` so native handlers can use a read-only context without depending on the processor implementation.

Handlers, host callbacks, and registry dispatch take `EventContext` by value. The context is `Copy + Clone` and borrows VM state for the invocation. Event handlers return `AdviceMutation` values that the processor validates and applies atomically; trace handlers remain read-only.

Addresses #3761.

## Changes

- Add portable handler traits, invocation metadata, checked stack/memory/advice reads, and typed read errors in `miden-event-handler`. The crate depends on `miden-core` and supports `no_std` with allocation.
- Make native stack reads payload-relative: event reads omit the event ID, and trace reads omit the trace marker and ID. Semantic IDs come from `context.id()`. Stack reads zero-extend beyond the logical depth.
- Expose current-context and explicit root-context memory reads. Scalar and word reads preserve missing-versus-zero behavior; bulk reads zero-fill uninitialized memory and leave output unchanged on failure.
- Migrate core-library handlers, synchronous/asynchronous hosts, registrations, and the Wasm host adapter to the shared context. Registration accepts concrete handlers, functions, and typed closures; existing shared handles use `registration::EventHandler::shared` or its trace equivalent.
- Keep Wasm v1 behavior unchanged, including the raw event envelope visible to guest stack reads.
- Use fixed per-invocation limits for bundled native crypto events: `MAX_KECCAK_INPUT_BYTES` is 1 MiB and `MAX_AEAD_PLAINTEXT_BYTES` is 16 MiB of serialized plaintext field elements, excluding padding and tag. Admission checks precede allocation and cryptographic work. Remove the configurable execution-option hash limit; retain the independent aggregate advice budget.
- Have generated field-inverse wrappers materialize operand hints and pass a field selector plus eight limbs to the inverse event. The handler needs no deferred-state access. MASM still proves `original X_DIGEST × INV_DIGEST = one`; both supported field modules are regenerated from their owning template.
- Fix system hash-input reads to zero-extend below the logical stack depth, preventing stale backing-buffer values from affecting hashes. This fix remains a separate commit.

## Migration

This is a breaking source migration: native handler implementations, custom hosts, and their helper reads must move to `EventContext` together. Import the portable traits, context, errors, and mutations directly from `miden-event-handler`. The raw `ProcessorState` view remains available for processor inspection outside callbacks.

Replace raw event payload offset `n` with `n - 1`, or `n - 2` for traces. Use semantic invocation metadata and current/root memory reads, and update concrete memory-error handling. Convert the `u32` clock where downstream APIs still require `RowIndex`. See the [handler migration guidance](https://github.com/0xMiden/miden-vm/blob/c321a9ce684cec015085ce0574af1b1bd273206c/crates/event-handler/README.md).

Protocol requires a coordinated follow-up in its executor/prover/mock hosts, transaction-event decoding, kernel-memory helpers, and LinkMap view. Source inspection of [protocol `b5cbb343`](https://github.com/0xMiden/protocol/tree/b5cbb343d30e098a0d027663b6e2183252201393) found the required reads covered by the new context. This PR does not change protocol; downstream compilation and runtime validation remain outstanding.

## Validation

- 172 focused nextest tests passed, covering portable context contracts, native crypto handlers, inverse-expression and forged-witness regressions, sync/async dispatch and proving, mutation atomicity, logical-stack zero extension, and Wasm handlers including compiled Rust guests.
- A standalone handler library depending directly on `miden-event-handler` with default features disabled passed the `no_std` check for `wasm32-unknown-unknown`.
- `make lint` passed: workspace/all-target/all-feature Clippy with warnings denied, including the affected Wasm benchmark target, nightly formatting checks, and cargo-shear with 28 advisory warnings.
- `git diff --check` passed. Generated MASM and documentation changes were inspected for scope.
- The ECDSA VM cycle baseline was measured at 1,353 cycles, up from 1,325.

### DIFF
--- a/core/src/execution.rs
+++ b/core/src/execution.rs
@@ -5,6 +5,12 @@ use miden_utils_diagnostics::{Diagnostic, miette};
 
 use crate::Felt;
 
+mod options;
+pub use options::{ExecutionOptions, ExecutionOptionsError};
+
+/// The minimum length of an execution trace required to support range checks.
+pub const MIN_TRACE_LEN: usize = 64;
+
 /// Identifies an execution context.
 #[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
 pub struct ContextId(u32);
--- a/core/src/execution.rs
+++ b/core/src/execution.rs
@@ -1,0 +1,131 @@
+use alloc::string::String;
+use core::fmt::{self, Display, LowerHex};
+
+use miden_utils_diagnostics::{Diagnostic, miette};
+
+use crate::Felt;
+
+/// Identifies an execution context.
+#[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd)]
+pub struct ContextId(u32);
+
+impl ContextId {
+    pub const fn root() -> Self {
+        Self(0)
+    }
+
+    pub const fn is_root(&self) -> bool {
+        self.0 == 0
+    }
+}
+
+impl From<u32> for ContextId {
+    fn from(value: u32) -> Self {
+        Self(value)
+    }
+}
+
+impl From<ContextId> for u32 {
+    fn from(context_id: ContextId) -> Self {
+        context_id.0
+    }
+}
+
+impl From<ContextId> for u64 {
+    fn from(context_id: ContextId) -> Self {
+        context_id.0.into()
+    }
+}
+
+impl From<ContextId> for Felt {
+    fn from(context_id: ContextId) -> Self {
+        Felt::from_u32(context_id.0)
+    }
+}
+
+impl Display for ContextId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Identifies a memory address.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct MemoryAddress(u32);
+
+impl MemoryAddress {
+    pub const fn new(address: u32) -> Self {
+        Self(address)
+    }
+}
+
+impl From<u32> for MemoryAddress {
+    fn from(address: u32) -> Self {
+        Self(address)
+    }
+}
+
+impl From<MemoryAddress> for u32 {
+    fn from(address: MemoryAddress) -> Self {
+        address.0
+    }
+}
+
+impl Display for MemoryAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        Display::fmt(&self.0, f)
+    }
+}
+
+impl LowerHex for MemoryAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        LowerHex::fmt(&self.0, f)
+    }
+}
+
+impl core::ops::Add<MemoryAddress> for MemoryAddress {
+    type Output = Self;
+
+    fn add(self, rhs: MemoryAddress) -> Self::Output {
+        Self(self.0 + rhs.0)
+    }
+}
+
+impl core::ops::Add<u32> for MemoryAddress {
+    type Output = Self;
+
+    fn add(self, rhs: u32) -> Self::Output {
+        Self(self.0 + rhs)
+    }
+}
+
+/// Lightweight error type for memory operations.
+///
+/// This enum captures error conditions without expensive source context. Execution engines can add
+/// source context when converting it into their top-level execution error.
+#[derive(Debug, thiserror::Error, Diagnostic)]
+pub enum MemoryError {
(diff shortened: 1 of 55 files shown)


bitwalker:
I think this probably can't be in core, since event handlers necessarily need to be able to access information about the processor state. That said, if we switch to Wasm-compiled event handlers, the actual Rust interface for those event handlers wouldn't need to reference the real processor types at all - a combination of host-defined functions (that provide the APIs we want event handlers to be able to call to query information) and `#[repr(C)]` data types for passing queried data (and event information) between host and handler, would make it so that virtually all of this could live just above `miden-core` in a dedicated crate.

@greenhat Is working on a PoC of this functionality at the moment, so it might make sense for us to make these changes in the context of landing that, and kill two birds with one stone.

greenhat:
> @greenhat Is working on a PoC of this functionality at the moment, so it might make sense for us to make these changes in the context of landing that, and kill two birds with one stone.

Yes, the PoC is in https://github.com/0xMiden/miden-vm/pull/3664

crates/event-handler/src/context.rs
```diff
@@ -0,0 +1,550 @@
...
+pub struct Invocation {
+    kind: InvocationKind,
+    id: EventId,
+    clock: u32,
+    context_id: ContextId,
+}
+
+impl Invocation {
+    /// Creates custom-event invocation metadata.
+    #[doc(hidden)]
```
adr1anh: We can make this public

crates/event-handler/src/context.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    fn advice_map(&self) -> &AdviceMap;
+
+    fn merkle_node(&self, root: Word, index: NodeIndex) -> Result<Word, MerkleReadError>;
+
+    fn merkle_path(&self, root: Word, index: NodeIndex) -> Result<MerklePath, MerkleReadError>;
+
+    fn has_merkle_path(&self, root: Word, index: NodeIndex) -> bool;
+
+    fn has_merkle_root(&self, root: Word) -> bool;
```
adr1anh: Are we sure we aren't missing anything useful?

crates/event-handler/src/context.rs
```diff
@@ -0,0 +1,550 @@
...
+            return Err(MemoryReadError::AddressOutOfBounds { address: start });
+        }
+        if end > u32::MAX as u64 {
+            return Err(MemoryReadError::AddressOutOfBounds { address: end });
+        }
+        if start > end {
+            return Err(MemoryReadError::InvalidRange { start, end });
+        }
+        Ok(start as u32..end as u32)
+    }
```
adr1anh: I'd like to remove, and replace with something more generic. Deprecate and use memory_range. Can we use MemoryAddress as input, allowing for safe conversions

crates/event-handler/src/context.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    /// Returns true when the advice map contains `key`.
+    pub fn advice_map_contains(&self, key: &Word) -> bool {
+        self.advice_map().contains_key(key)
+    }
+
+    /// Returns the advice-map value for `key`.
+    pub fn advice_map_entry(&self, key: &Word) -> Option<&[Felt]> {
+        self.advice_map().get(key).map(AsRef::as_ref)
+    }
```
adr1anh: Maybe we can just do this through `.map().vvv()`

core/src/execution.rs
```diff
@@ -0,0 +1,137 @@
...
+        Display::fmt(&self.0, f)
+    }
+}
+
+// MEMORY ADDRESS
+// ================================================================================================
+
+/// Identifies an element address in VM memory.
+#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct MemoryAddress(u32);
```
adr1anh: We should be able to have try_from* and free functions to create checked ranges.

crates/lib/core/src/handlers/falcon_div.rs
```diff
@@ -37,8 +37,9 @@ pub const FALCON_DIV_EVENT_NAME: EventName =
 /// - Returns an error if the divisor is ZERO.
 /// - Returns an error if either a0 or a1 is not a u32.
 pub fn handle_falcon_div(context: &EventContext) -> Result<Vec<AdviceMutation>, EventError> {
-    let dividend_hi = context.stack_item(1).as_canonical_u64();
-    let dividend_lo = context.stack_item(2).as_canonical_u64();
+    let mut dividend = [ZERO; 2];
+    context.read_stack(1, &mut dividend);
+    let [dividend_hi, dividend_lo] = dividend.map(|value| value.as_canonical_u64());
```
adr1anh: just do two reads for each value no?

Issue #3439, opened 2026-07-30
Support LeanSig as a hash-based post-quantum application signature scheme


partylikeits1983:
## Motivation

Miden currently supports Falcon512-Poseidon2 verification as a post-quantum signature option. Falcon is lattice-based and stateless. We should explore LeanSig as a complementary hash-based post-quantum signature scheme at the application/account authentication layer.

The main motivations are:

- The Ethereum Foundation's PQ team is actively developing and testing LeanSig/leanXMSS for lean Ethereum.
- Hash-based signatures rely primarily on the security of the underlying hash function, providing a comparatively conservative post-quantum assumption.
- Supporting both hash-based and lattice-based signatures gives applications cryptographic diversity instead of relying on a single post-quantum family.
- LeanSig's synchronized epoch model maps naturally onto blockchain applications that already maintain ordered account or protocol nonces.

Relevant EF work:

- https://pq.ethereum.org/
- https://github.com/leanEthereum/leanSig
- https://eprint.iacr.org/2025/055.pdf

The recently disclosed improved attack against HAWK is a slight additional motivator for this diversity. The attack exploits a HAWK-specific lattice automorphism and reduces the demonstrated HAWK-256 key-recovery cost from approximately 2^64 to 2^38 operations. It remains an exponential attack and does not affect Falcon, other NIST post-quantum candidates, or lattice cryptography generally.

This should therefore not be interpreted as evidence that Falcon is vulnerable. It is simply a reminder that maintaining diversity across post-quantum assumptions is useful.

Reference: https://www.anthropic.com/research/discovering-cryptographic-weaknesses

## Proposed Miden instantiation

Add an experimental Miden-native LeanSig implementation consisting of:

- A verifier in the Miden core library, implemented in MASM.
- Matching Rust key generation, signing, verification, serialization, and advice encoding under `miden-crypto::dsa`.
- A persisted monotonic epoch/nonce cursor that rejects reused or skipped one-time keys.
- End-to-end tests in which the Rust signer generates a signature consumed by the MASM verifier.

The construction should retain LeanSig's generalized-XMSS flow and lifetime-2^32 target-sum parameters:

- Dimension: 46
- Winternitz base: 8
- Target sum: 200
- Merkle depth: 32
- SHAKE128 secret-key PRF

[...]

PR #3440, opened 2026-07-30
feat(crypto): add Miden-native Poseidon2 LeanSig

partylikeits1983:
WIP LeanSig implementation for Miden.

Unlike the current EF LeanSig instantiation, this uses the Goldilocks field and Miden-native Poseidon2 rather than KoalaBear Poseidon1, so it is not wire-compatible. LeanSig itself is not tied to Poseidon1, and BLAKE3 is also being considered.

Tracking issue: #3439


Nashtare:
I may be lacking context (haven't interacted with leanSig in a while), but haven't they reverted back to Poseidon1 after the recent findings against Poseidon2?

PR #3342, opened 2026-07-08
core-lib: align ECDSA public key format with miden-crypto 0.28

adr1anh:
Depends on #3340.

## Describe your changes

PR #3340 bumps `miden-crypto` to 0.28, where ECDSA `PublicKey::to_elements()` / `to_commitment()` use affine public key coordinates: `qx_le_u32[8] || qy_le_u32[8]`.

This PR updates the core ECDSA K256 Keccak wrapper to use that format consistently:

- Advice public key chunk becomes `PK[16]`.
- Public key commitment uses `PublicKey::to_commitment()`.
- Precompile request calldata stores `pk_affine_coordinate_bytes[64] || digest[32] || sig[65]`.
- MASM commitment computation hashes 16 public key felts.
- Tests use `pk.to_elements()` / `pk.to_commitment()`.

This is a format/ABI change for the core ECDSA wrapper: callers should provide public key advice via `pk.to_elements()` and public key commitments via `pk.to_commitment()`.

## Validation

```sh
cargo test -p miden-core-lib --test core-lib crypto::ecdsa_k256_keccak --features testing
```

PR #3444, opened 2026-07-31
docs: clarify limited crate releases

huitseeker:
Recent crypto 0.28.1 was published from https://github.com/0xmiden/crypto/ instead of here, leading to a few recovery PRs (#3342 #3343) so taking this occasion to ensure the docs are easier to follow


bobbinth (approved):
Looks good! Thank you!

PR #3445, opened 2026-08-01
ci: preserve baseline Blake3 benchmark harness

Al-Kindi-0:
## Describe your changes

Preserve each benchmark ref's existing Blake3 harness, and install the current harness only when the selected ref predates it. This prevents the non-regression workflow from compiling a baseline against a harness written for a newer public API.

This fixes the deterministic baseline build failure in https://github.com/0xMiden/miden-vm/actions/runs/30693565170. The harness difference between d02eb7adc and 752e8dea7 is limited to the untimed proof-validation call; the measured benchmark paths are unchanged.

## Validation

- Simulated the install step with baseline d02eb7adc and current 752e8dea7 and confirmed both harnesses were preserved.
- Confirmed the fallback installs the harness for a ref that predates it.
- Ran cargo check -p miden-vm-blake3-bench --lib against both refs.
- Parsed the workflow YAML and checked the staged diff.

- Branch created from next.
- Commit message and style follow repository conventions.
- No changelog entry: this only fixes CI benchmark setup.

### DIFF
--- a/.github/workflows/blake3-nonregression.yml
+++ b/.github/workflows/blake3-nonregression.yml
@@ -327,25 +327,28 @@ jobs:
           git worktree add --detach "${WORKTREE_DIR}/baseline" "${BASELINE_SHA}"
           git worktree add --detach "${WORKTREE_DIR}/current" "${CURRENT_SHA}"
 
-      - name: Install current Blake3 benchmark harness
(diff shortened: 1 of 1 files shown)


bobbinth (approved):
Looks good! Thank you! I'll merge as is - but @huitseeker do take a look in case you'd have any comments.

PR #3446, opened 2026-08-02
Tracking PR for v0.29.0 release

bobbinth:
This is a tracking PR for v0.29.0 release.


crates/lib/core/asm/miden-project.toml
```diff
@@ -1,13 +1,13 @@
...
-version = "0.26.0"
+version = "0.29.0"
 
 [lib]
 namespace = "miden::core"
 path = "mod.masm"
 
 [dependencies]
-miden-precompiles = { version = "0.26", linkage = "dynamic" }
+miden-precompiles = { version = "0.29", linkage = "dynamic" }
```
bobbinth, line 10: How does `miden-precompiles` project get built? I couldn't find a TOML file for it, and it seems like it may be generated dynamically in the `build.rs`? Also, it seems like the version defined here doesn't matter too much as it gets overridden in the `build.rs` as well?

huitseeker: `crates/lib/core/build.rs` generates a dependency-free aggregate MASM project containing both `miden::core` and `miden::precompiles`; the generated manifest uses `miden-core-lib`’s Cargo version instead of the checked-in MASM manifest’s versions.

bobbinth:
Interesting that CodSpeed is showing a 10% regression in `apply_mutations`. The last PR merged was just a typo fix - so, shouldn't have changed anything performance-wise. Or are we comparing this against `main`?

huitseeker:
@bobbinth not _that_ interesting — the codspeed baselines were only set up correctly in #3448

PR #3447, opened 2026-08-02
Recursive verifier review followups

Al-Kindi-0:
Follow-up to #3422 addressing issues found during review.

- Authenticate recursive-verifier claims by commitment and validate kernel witness bounds.
- Fix the empty domain-separated field-list hash collision in Rust and MASM.
- Consolidate exact schema decoding and simplify the recursive-verifier Rust API.
- Add negative and end-to-end coverage for claim, kernel-witness, and request-key binding.


benches/synthetic-bench/benches/recursive_verify.rs
```diff
@@ -602,18 +576,16 @@ fn recursive_proof_advice(fixture: &TxProofFixture) -> RecursiveProofAdvice {
         fixture.stack_outputs,
     );
     let verifier_inputs = generate_advice_inputs(&fixture.proof, &claim).expect("recursive advice");
+    let claim_commitment = verifier_inputs.claim_commitment;
 
-    let advice_stack = AdviceStack::try_from_values(verifier_inputs.advice_stack())
+    let advice_stack = AdviceStack::try_from_values(verifier_inputs.advice_stack().iter().copied())
```
huitseeker: Nit: Could this benchmark use `RecursiveVerifierInputs::new(...).into_parts()` directly?

huitseeker (approved):
LGTM!

benches/synthetic-bench/benches/recursive_verify.rs
Al-Kindi-0: Done

core/src/program/request.rs
```diff
@@ -0,0 +1,24 @@
...
+pub const PROOF_REQUEST_DOMAIN_TAG: Felt = domain_selector(PROOF_REQUEST_DOMAIN_ID, 1);
+
+/// Returns the advice-map key addressing a proof package for `claim_commitment` under the
+/// verifier identified by `verifier_root`.
+///
+/// The key is `H_tag(claim_commitment ‖ verifier_root)` (one rate block, domain-separated). It
+/// is a lookup address, not a trust anchor: the verifier re-checks the retrieved package against
+/// its statement, so a wrong package fails verification. Both inputs are values the requester owns
+/// rather than values taken from advice.
+pub fn proof_request_key(verifier_root: Word, claim_commitment: Word) -> Word {
```
bobbinth, line 17: Would it make sense to associate this function with `ExecutionProof`? Something like:

```rust
impl ExecutionProof {

    pub fn get_advice_key(verifier_root: Word, claim_commitment: Word) -> Word {
        ...
    }

}
```

verifier/src/recursive/mod.rs
```diff
@@ -60,46 +60,97 @@ type P2Config = config::Poseidon2Config;
...
-        let key = request_key(verifier_root, self.claim_commitment);
-        let proof_stream = core::mem::take(&mut self.advice_stack);
-        self.advice_map.push((key, proof_stream));
+    /// # Errors
+    ///
+    /// Returns an error if the proof cannot be converted into verifier advice.
+    pub fn new(
+        proof: &ExecutionProof,
+        claim: &ExecutionClaim,
+    ) -> Result<Self, RecursiveVerifierInputsError> {
```
bobbinth: Question: what are the use cases when we want to `RecursiveVerifierInputs` but don't have the verifier root? Basically, why do we need to have this `new()` constructor vs. just always using  `for_request()`?

crates/lib/core/asm/sys/mod.masm
```diff
@@ -36,3 +48,27 @@ end
...
+#! The key is the domain-tagged hash of `claim_commitment ‖ verifier_root` (exactly one rate
+#! block, so a single permutation with no memory). It is a lookup address, not a trust anchor: the
+#! verifier re-checks the retrieved package, so a wrong package fails verification. Both inputs are
+#! program-owned: the verifier's MAST root comes from `procref`, while the claim commitment
+#! comes from an authenticated verifier output or the program's own inputs. Mirrors
+#! `miden_core::program::proof_request_key`.
+#!
+#! Inputs:  [VERIFIER_ROOT, CLAIM_COMMITMENT, ...]
+#! Outputs: [PROOF_REQUEST_KEY, ...]
+pub proc proof_request_key
```
bobbinth: This is only relevant for recursive verification, right? If so, I'd put this under `miden::core::sys::vm` namespace.

Also, I'd rename the procedure to something like `build_proof_request_key`.

crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -460,28 +476,30 @@ end
...
-    push.0.0 movup.5 dup.5
-    # => [num_elements%8, domain, 0, 0, ptr, end_pairs_addr, num_elements%8]
+    push.0 movdn.2
+    # => [ptr, num_elements, pad_inputs_flag=0, domain]
+    exec.prepare_hasher_state
+    # => [R0, R1, [num_elements%8, 0, 0, 0], ptr, end_pairs_addr, num_elements%8, domain]
 
-    padw padw
+    # Move domain into the second capacity element.
+    movup.15 movdn.9 movup.10 drop
```
bobbinth: nit: we could write this as `movup.15 swap.9 drop`.

crates/lib/core/asm/crypto/hashes/poseidon2.masm
```diff
@@ -491,9 +509,10 @@ pub proc hash_elements_in_domain
     dup.10 eq.0 not and
     # => [empty_and_domain_nonzero, R0, R1, C, ptr, end_pairs_addr, num_elements%8]
     if.true
-        # set rate[0] = 1, apply one permutation, and squeeze
-        drop push.1
-        # => [R0 = (1, 0, 0, 0), R1, C, ptr, end_pairs_addr, num_elements%8]
+        # Mark empty input in the third capacity element. This lane is zero for every non-empty
+        # input, so the marker cannot be reproduced by a full rate block.
+        movup.10 drop push.1 movdn.10
```
bobbinth: nit: this could be `push.1 swap.10 drop`

bobbinth (approved):
Looks good! Thank you! Not a very deep review (I mostly skipped testing code) but I left a few small comments inline.

core/src/program/request.rs
Al-Kindi-0: I would keep it as a free function with a forward thinking towards the PVM side of things.

verifier/src/recursive/mod.rs
Al-Kindi-0: Good point, the `new` variant was used to accommodate benches and tests where it was more straightforward to use it rather than the `for_request` variant.
Removed `new` and updated the sites that used to use it.

crates/lib/core/asm/sys/mod.masm
Al-Kindi-0: Forward thinking towards the PVM where this is will also be useful

crates/lib/core/asm/crypto/hashes/poseidon2.masm
Al-Kindi-0: Done

crates/lib/core/asm/crypto/hashes/poseidon2.masm
Al-Kindi-0: Done

PR #3448, opened 2026-08-02
ci: fix codspeed and run all jobs

huitseeker:
- split Codspeed into synthetic (60m) and SMT (30m) jobs
- run build, test, and lint jobs without changed-area skips
- generate core.masp from the workspace-locked embedded artifact
- exercise make packages in every release dry-run


bobbinth (approved):
Looks good! Thank you! (though, not a deep review)

PR #3453, opened 2026-08-03
Clarify push syntax stack order

### LINKED ISSUE #2572: Remove `push.a.b.c.d` syntax

Al-Kindi-0:
Closes #2572

### DIFF
--- a/docs/src/user_docs/assembly/io_operations.md
+++ b/docs/src/user_docs/assembly/io_operations.md
@@ -29,16 +29,16 @@ push.4660.22136.36882.43981
 
 In both case the values must still encode valid field elements.
 
-#### Word literal syntax
+#### Sequential pushes and word literals
 
-The `push.[a,b,c,d]` syntax provides a convenient way to push a word (4 field elements) onto the stack. The elements are pushed such that the first element `a` ends up on top of the stack:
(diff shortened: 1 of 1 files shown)


adr1anh (approved):
Looks good

PR #3455, opened 2026-08-03
Cover same-clock access invariant

### LINKED ISSUE #2390: Investigate if we should disallow memory operations when `(ctx, adr, clk)` is the same
It seems to me that the only situation where we would want to allow two memory operations to happen for the same `(ctx, adr, clk)` is when one is a word-read and the other is an element-read. Even this situation seems a bit strange, so we should investigate if we should disallow same memory operations for the same `(ctx, adr, clk)` as this can break the soundness of the memory argument.

_Originally posted by @plafer in https://github.com/0xMiden/miden-vm/pull/2322#discussion_r2550466016_

Al-Kindi-0:
Closes #2390


docs/src/design/chiplets/memory.md
```diff
@@ -182,7 +182,7 @@ where:
 - `idx0` and `idx1` are selector columns used to identify which element in the word is being accessed. Specifically, the index within the word is computed as `idx1 * 2 + idx0`.
   - However, when `ew` is set to $1$ (indicating that a word is accessed), these columns are meaningless and are set to $0$.
 - `clk` contains clock cycle at which the memory operation happened. Values in this column must increase monotonically for a given context and memory word but there can be gaps between two consecutive values of up to $2^{32}$.
-  - When the context and word address are unchanged, `clk` may stay the same, but only read operations are allowed.
+  - When the context and word address are unchanged, `clk` may stay the same, but only read operations are allowed. This includes any combination of word and element reads; a write may not share the same `(ctx, word_addr, clk)` tuple with any other access.
```
huitseeker, line 185: nit: Could we update the `Memory::read` and `Memory::read_word` error docs in `processor/src/trace/chiplets/memory/mod.rs` too? They still say any repeated same-clock address returns an error, but the implementation allows read/read just as this line says. The matching `MemorySegmentTrace` docs have the same stale claim.

huitseeker (approved):
LGTM!

docs/src/design/chiplets/memory.md
Al-Kindi-0: Done

PR #3456, opened 2026-08-03
fix(crypto): wipe secret-key temporaries

Jr-kenny:
Ports [0xMiden/crypto#1061](https://github.com/0xMiden/crypto/pull/1061) now that the crypto crates live in this workspace. The original gap is tracked in [0xMiden/miden-vm#3533](https://github.com/0xMiden/miden-vm/issues/3533).

## What was still leaking

`SecretKey` wipes its stored Falcon basis on drop, but the work around it was leaving extra copies of the same secret material behind. `Polynomial<F>` claimed `ZeroizeOnDrop` without a `Drop` implementation, so those coefficients were never actually wiped. Signing, serialization, deserialization, public-key derivation, and rejected key-generation retries all created more secret polynomials or encoded key buffers that dropped normally.

The same encoded-key pattern also showed up in the ECDSA and EdDSA secret-key comparison and serialization paths.

## What changed

`FalconFelt` now implements `Zeroize`, and secret-carrying polynomials are wrapped in `Zeroizing` where they are created. The unbacked `ZeroizeOnDrop` marker is gone. The FFT chain in `SecretKey::read_from`, the encoded buffers in `write_into` and `generate_seed`, the public-key derivation clones, and rejected `ntru_gen` candidates are all covered.

Secret-key equality stays available to the crate's tests, but `Eq` and `PartialEq` are no longer exposed for Falcon, ECDSA, or EdDSA secret-key wrappers in production builds. ECDSA and EdDSA serialization now wipe their temporary encoded copies too.

I ported the security changes onto the current `next` code instead of copying the old crypto files across, so the newer `miden-vm` refactors stay intact. The `BigInt` and `Complex64` residue documented in [0xMiden/miden-vm#3533](https://github.com/0xMiden/miden-vm/issues/3533) is still outside this change because those paths need the same manual volatile treatment already used by `LdlTree`.

## Checks I ran

The focused regression failed before the fix because `FalconFelt` did not satisfy `Zeroize`. It passes with this change, along with the full crypto suite and the build combinations used by this code.

```text
cargo test -p miden-crypto test_zeroize_wipes_falcon_felt_polynomial
cargo test -p miden-crypto dsa                         # 53 passed
cargo test -p miden-crypto --lib                       # 677 passed
cargo check -p miden-crypto --all-targets --all-features
cargo clippy -p miden-crypto --all-targets --all-features -- -D warnings
cargo check -p miden-crypto --no-default-features --target wasm32-unknown-unknown
cargo +nightly fmt --all --check
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## v0.29.0 (Unreleased)
 
+#### Changes
+
+- [BREAKING] Removed `Eq` and `PartialEq` from Falcon, ECDSA-k256, and EdDSA secret-key types in
+  non-test builds so production comparisons cannot materialize encoded private-key copies
+  ([0xMiden/crypto#1061](https://github.com/0xMiden/crypto/pull/1061)).
+
+#### Fixes
+
+- Wiped Falcon secret-polynomial and encoded-key temporaries during key generation,
+  deserialization, serialization, public-key derivation, and signing. ECDSA-k256 and EdDSA
+  serialization temporaries are also wiped
+  ([0xMiden/crypto#1061](https://github.com/0xMiden/crypto/pull/1061)).
+
 ## v0.28.0 (2026-08-01)
 
 #### Features
--- a/crates/crypto/src/dsa/ecdsa_k256_keccak/mod.rs
+++ b/crates/crypto/src/dsa/ecdsa_k256_keccak/mod.rs
@@ -21,7 +21,8 @@ use crate::{
     ecdh::k256::{EphemeralPublicKey, SharedSecret},
     utils::{
         ByteReader, ByteWriter, Deserializable, DeserializationError, Serializable,
-        read_sensitive_array, zeroize::ZeroizeOnDrop,
+        read_sensitive_array,
+        zeroize::{ZeroizeOnDrop, Zeroizing},
     },
 };
 
@@ -96,20 +97,25 @@ impl SecretKey {
 // which ensures that the secret key material is securely zeroized when dropped.
 impl ZeroizeOnDrop for SecretKey {}
 
+#[cfg(test)]
 impl PartialEq for SecretKey {
     fn eq(&self, other: &Self) -> bool {
         use subtle::ConstantTimeEq;
-        self.to_bytes().ct_eq(&other.to_bytes()).into()
+        let self_bytes = Zeroizing::new(self.inner.to_bytes());
+        let other_bytes = Zeroizing::new(other.inner.to_bytes());
+        self_bytes[..].ct_eq(&other_bytes[..]).into()
     }
 }
 
+#[cfg(test)]
 impl Eq for SecretKey {}
 
 // SIGNING KEY
 // ================================================================================================
 
 /// A secret key for ECDSA signature verification over the secp256k1 curve.
-#[derive(Clone, Eq, PartialEq, SilentDebug, SilentDisplay)] // Safe as SecretKey has const-time eq
+#[derive(Clone, SilentDebug, SilentDisplay)]
+#[cfg_attr(test, derive(Eq, PartialEq))] // Safe as SecretKey has const-time eq in tests
 pub struct SigningKey(SecretKey);
 
 impl SigningKey {
@@ -170,7 +176,8 @@ impl Deserializable for SigningKey {
 // ================================================================================================
 
 /// A secret key for ECDH key-exchange over the secp256k1 curve.
-#[derive(Clone, Eq, PartialEq, SilentDebug, SilentDisplay)] // Safe as SecretKey has const-time eq
+#[derive(Clone, SilentDebug, SilentDisplay)]
+#[cfg_attr(test, derive(Eq, PartialEq))] // Safe as SecretKey has const-time eq in tests
 pub struct KeyExchangeKey(SecretKey);
 
 impl KeyExchangeKey {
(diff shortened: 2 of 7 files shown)


crates/crypto/src/dsa/falcon512_poseidon2/keys/secret_key.rs
```diff
@@ -198,12 +198,16 @@ impl SecretKey {
...
     /// Derives the public key corresponding to this secret key using h = g /f [mod ϕ][mod p].
     fn compute_pub_key_poly(&self) -> PublicKey {
-        let g: Polynomial<FalconFelt> = self.secret_key[0].clone().into();
-        let g_fft = g.fft();
-        let minus_f: Polynomial<FalconFelt> = self.secret_key[1].clone().into();
-        let f = -minus_f;
-        let f_fft = f.fft();
-        let h_fft = g_fft.hadamard_div(&f_fft);
+        let g: Zeroizing<Polynomial<FalconFelt>> =
+            Zeroizing::new(self.secret_key[0].clone().into());
```
huitseeker: `Zeroizing` starts after `self.secret_key[0].clone()` has allocated a full `Polynomial<i16>`. The by-value conversion only reads that clone to build the `FalconFelt` polynomial, then the clone drops without a wipe. The same happens for `minus_f` below, so public-key derivation still leaves two secret copies. Could these use the existing `From<&Polynomial<i16>>` conversion instead, so the input clones are never created?

crates/crypto/src/dsa/falcon512_poseidon2/math/polynomial.rs
```diff
@@ -643,7 +643,10 @@ impl<F: Zeroize> Zeroize for Polynomial<F> {
     }
 }
 
-impl<F: Zeroize> ZeroizeOnDrop for Polynomial<F> {}
+// `ZeroizeOnDrop` cannot be provided for `Polynomial<F>`: a `Drop` impl bounded on
+// `F: Zeroize` is rejected by the compiler (E0367), and bounding the struct itself would
+// exclude `Polynomial<Complex64>` used by the signing path. Callers holding secret
+// polynomials must wrap them in `Zeroizing` instead.
```
huitseeker, line 649: Removing the marker makes this a caller-enforced rule, but `to_complex_fft` still starts with `basis.clone()`. That creates four plain `Polynomial<i16>` allocations on every key construction and every signature; they drop unwiped before the acknowledged `Complex64` residue. Could `to_complex_fft` map from the borrowed basis entries and avoid those copies?

Jr-kenny: Sorry, I missed this copy path in the original change. `basis.clone()` was creating four plain `Polynomial<i16>` allocations before the FFT work. I changed `to_complex_fft` to destructure the borrowed array, so each `map` runs from the borrowed entries instead. I verified the focused zeroization test and the crypto gates locally on the Mac.

crates/crypto/src/dsa/falcon512_poseidon2/keys/secret_key.rs
Jr-kenny: Sorry, I missed these plain secret copies in the original change. `.clone().into()` was allocating the `Polynomial<i16>` before `Zeroizing` took ownership, for both `g` and `minus_f`. I changed those conversions to `Polynomial::from(&self.secret_key[0])` and `Polynomial::from(&self.secret_key[1])`, so the conversion now reads the existing secret polynomials by reference. I verified the focused zeroization test and the crypto gates locally on the Mac.

crates/crypto/src/dsa/ecdsa_k256_keccak/mod.rs
```diff
@@ -448,9 +455,10 @@ impl Signature {
 
 impl Serializable for SecretKey {
     fn write_into<W: ByteWriter>(&self, target: &mut W) {
-        let mut buffer = Vec::with_capacity(SECRET_KEY_BYTES);
-        let sk_bytes: [u8; SECRET_KEY_BYTES] = self.inner.to_bytes().into();
-        buffer.extend_from_slice(&sk_bytes);
+        let mut buffer = Zeroizing::new(Vec::with_capacity(SECRET_KEY_BYTES));
```
huitseeker, line 458: nit: This keeps two guarded copies of the same 32-byte key and allocates a `Vec` only to pass it to `write_bytes`. Could we write `sk_bytes` directly? The array is already wrapped in `Zeroizing`, so this would keep the wipe and remove the extra allocation and copy.

crates/crypto/src/dsa/falcon512_poseidon2/keys/secret_key.rs
```diff
@@ -274,7 +278,9 @@ impl SecretKey {
     fn generate_seed(&self, message: &Word) -> [u8; 32] {
         let mut buffer = Vec::with_capacity(1 + SK_LEN + Word::SERIALIZED_SIZE);
         buffer.push(LOG_N);
-        buffer.extend_from_slice(&self.to_bytes());
+        // Bind the serialized key so the temporary holding it is wiped, not just `buffer`.
+        let sk_bytes = Zeroizing::new(self.to_bytes());
```
huitseeker, line 282: nit: Could `buffer` be wrapped in `Zeroizing` when it is created? That would remove the manual wipe below and keep the buffer covered if this function ever gains an early return or a panic path.

Jr-kenny:
Addressed both nits in cb19f83:

1. ECDSA `SecretKey::write_into` now writes `sk_bytes` directly. The intermediate `Vec` is gone, so there is only one guarded copy of the 32-byte key.

2. Falcon `generate_seed` now creates `buffer` as `Zeroizing<Vec<_>>` instead of a plain `Vec` plus a manual wipe at the end. Early returns and panics stay covered.

Verified in Codespace on this branch:
```
cargo check -p miden-crypto --lib
cargo test -p miden-crypto dsa   # 53 passed
```

Tracked under 0xMiden/miden-vm#3533.

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 ## v0.29.0 (Unreleased)
 
+#### Changes
+
+- [BREAKING] Removed `Eq` and `PartialEq` from Falcon, ECDSA-k256, and EdDSA secret-key types in
```
huitseeker, line 7: `Polynomial<F>` also loses its public `ZeroizeOnDrop` impl. A downstream `T: ZeroizeOnDrop` bound compiles at the base and fails here with E0277, so could this breaking entry mention that trait-implementation removal too?

Jr-kenny:
Addressed in 656d029c44e6126e775596bba515c43547f99884.

The breaking CHANGELOG entry now also notes that `Polynomial<F>` loses its public `ZeroizeOnDrop` impl, since a downstream `T: ZeroizeOnDrop` bound that compiled at the base will fail with E0277 here.

huitseeker (changes requested):
I'm afraid this has been invalidated by #3769, and now needs a rebase 😱. So sorry!

